PR #67, opened 2016-08-03
Fixed and tested Interfaces.TimeData save parameter

vincepandolfo:
- Added methods to initialize spatial data and retrieve results from TimeData objects. 
- Added a test for the save parameter.

A little issue: The final result is at `timesteps - 2` while one would expect it to be at `timesteps - 1` since the Devito calculated values start from 0.

My guess is that the result when `save=False` is not the 100th timestep, but the 99th. I can't see why that would happen though.

EDIT: Added issue https://github.com/opesci/devito/issues/69 regarding this problem

### DIFF
--- a/devito/interfaces.py
+++ b/devito/interfaces.py
@@ -265,6 +265,7 @@ class TimeData(DenseData):
             return
         else:
             super(TimeData, self).__init__(*args, **kwargs)
+            self._full_data = self._data.view() if self._data else None
             time_dim = kwargs.get('time_dim')
             self.time_order = kwargs.get('time_order', 1)
             self.save = kwargs.get('save', False)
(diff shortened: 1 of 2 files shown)


devito/interfaces.py
```diff
@@ -295,9 +296,37 @@ def _allocate_memory(self):
...
 
+        self._full_data = self._data.view()
+
         if self.pad_time:
             self._data = self._data[self.time_order:, :, :]
 
+    def init_data(self, timestep, data):
+        """Function to initialize the initial time steps
+
+        :param timestep: Time step to initialize. Must be negative since calculated timesteps start from 0.
```
FabioLuporini: long line

devito/interfaces.py
```diff
@@ -295,9 +296,37 @@ def _allocate_memory(self):
...
+        :param data: :class:`numpy.ndarray` containing the initial spatial data
+        """
+        if self._full_data is None:
+            self._allocate_memory()
+
+        assert timestep < 0, "Timestep must be negative"
+        assert data.shape == self._full_data[0].shape, "Data must have the same shape as the spatial data"
```
FabioLuporini: long line

tests/test_save.py
```diff
@@ -0,0 +1,41 @@
...
+
+    return ui
+
+
+def run_simulation(save=False, dx=0.01, dy=0.01, a=0.5, timesteps=100):
+    nx, ny = int(1 / dx), int(1 / dy)
+    dx2, dy2 = dx**2, dy**2
+    dt = dx2 * dy2 / (2 * a * (dx2 + dy2))
+
+    u = TimeData(name='u', shape=(nx, ny), time_dim=timesteps, time_order=1, space_order=2, save=save, pad_time=save)
```
FabioLuporini: long line

FabioLuporini:
if @navjotk  is happy with this, I think we can merge

PR #68, opened 2016-08-03
Moved initialization logic from Operator to Propagator

vincepandolfo:
Some initialization logic for the Propagator was part of the Operator code. 

This PR moves it down to the Propagator.

### DIFF
--- a/devito/operator.py
+++ b/devito/operator.py
@@ -165,12 +165,14 @@ class Operator(object):
         self.stencils = [Eq(expr_indexify(eqn.lhs), expr_indexify(eqn.rhs))
                          for eqn in self.stencils]
 
+        for name, value in factorized.items():
+            factorized[name] = expr_indexify(value)
+
         # Apply user-defined substitutions to stencil
         self.stencils = [eqn.subs(args) for eqn, args in zip(self.stencils, substitutions)]
-        self.propagator = Propagator(self.getName(), nt, shape, spc_border=spc_border,
-                                     time_order=time_order, forward=forward,
-                                     space_dims=self.space_dims,
-                                     compiler=self.compiler, profile=profile,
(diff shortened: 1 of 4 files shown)


devito/operator.py
```diff
@@ -165,12 +165,14 @@ def __init__(self, nt, shape, dtype=np.float32, stencils=[],
         self.stencils = [Eq(expr_indexify(eqn.lhs), expr_indexify(eqn.rhs))
                          for eqn in self.stencils]
 
+        for name in factorized.keys():
+            self.propagator.factorized[name] = expr_indexify(factorized[name])
```
FabioLuporini: is this identical to : 
```
for name, value in factorized.items():
    self.propagator.factorized[name] = expr_indexify(value)
```
?

vincepandolfo: Yes. And it also doesn't make much sense to have it there. I'll change this

devito/propagator.py
```diff
@@ -35,10 +37,11 @@ class Propagator(object):
                       a list stating block sizes for each dimension. Set block size to None to skip blocking on that dim
     """
 
-    def __init__(self, name, nt, shape, spc_border=0, time_order=0,
+    def __init__(self, name, nt, shape, stencils, factorized=[], spc_border=0, time_order=0,
```
FabioLuporini: Never use ``[]`` for initialisation. Use ``None`` and then in the constructor you can do

```
self.factorized = factorized or []
```

Explanation here: http://docs.python-guide.org/en/latest/writing/gotchas/

FabioLuporini:
Apart from these minor things, this should be merged ASAP.

Also, since you're touching this file, can you break the doc strings longer than 89 characters ? Thanks

mlange05:
Looks good to me.

FabioLuporini:
Merging this.

Issue #71, opened 2016-08-04
Initialization of `SymbolicData` objects


vincepandolfo:
Since #67 brought to light some issues with the initialization of SymbolicData objects (TimeData objects specifically) we need to find a coherent and uniform interface to initialize SymbolicData objects.

A possible solution for TimeData would be to have the `data` property reference `_full_data` instead of `_data`. In this way when `self.initializer` is called from `initialize` it will get as an argument the full data array instead of the data array with the hidden indexes.

This would also have the effect of removing the time padding from the user point of view.

Initial discussion with @mlange05 here: https://github.com/opesci/devito/commit/09dc1557d51455c6be6fe6e887223e5d568282ee#commitcomment-18517910

mlange05:
I like the idea of having the `data` property referencing `_full_data` under the hood for time-buffered `TimeData` objects. I also note that this probably aligns with issue #69, where the obvious solution is to keep a private running timestep counter that we can modulo on to get the right time-frame in `data`.

Also, I believe the idea of adding the "initializer function" as a constructor argument was floated at some point, which I think sounds like a good idea? As does the idea of removing the `pad_time` argument? Just a few suggestions for further discussion...

navjotk:
Agree on the initializer function and `pad_time` above.

vincepandolfo:
Having `data` reference `_full_data` doesn't quite work since `Operator.apply` calls `param.data`, but it needs `_data` to work since when `save=True` the generated code tries to access the hidden indexes.

PR #73, opened 2016-08-04
Removed get_data and init_data, changed TimeData.initialize()

vincepandolfo:
This PR tries to address some of the problems discussed in issue #71 .

- It is now possible to initialize TimeData objects using an `initializer` function. (Note: This is also the only way to initialize them if `save=True`). `test_save.py` has been modified to demonstrate this.
- Removed `get_data` and `init_data`

### DIFF
--- a/tests/test_save.py
+++ b/tests/test_save.py
@@ -26,10 +26,9 @@ def run_simulation(save=False, dx=0.01, dy=0.01, a=0.5, timesteps=100):
     dt = dx2 * dy2 / (2 * a * (dx2 + dy2))
 
     u = TimeData(
-        name='u', shape=(nx, ny), time_dim=timesteps,
+        name='u', shape=(nx, ny), time_dim=timesteps, initializer=initializer,
         time_order=1, space_order=2, save=save, pad_time=save
     )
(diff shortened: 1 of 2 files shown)


FabioLuporini:
So the only thing that I think we also sort of discussed is that the ``initializer`` function could be provided directly through the constructor. Is this still correct ?

mlange05:
Looks good to me. I would like to add the `initializer` function to the constructor, although that's not a critical requirement.

tests/test_save.py
```diff
@@ -35,9 +39,9 @@ def run_simulation(save=False, dx=0.01, dy=0.01, a=0.5, timesteps=100):
     op.apply()
 
     if save:
-        return u.get_data(timesteps - 2)
+        return u.data[timesteps - 1, :]
     else:
-        return u.get_data()
+        return u.data[0, :]
```
navjotk: Do you want to use the `%` operator here to make sure this works even if the number of timesteps above is changed? That would also make the code more self-explanatory.

FabioLuporini:
Now it is in the constructor. Good

navjotk:
Merged

Issue #65, opened 2016-08-02
Profiling broken


navjotk:
The ability to profile code, a feature that was previously working is now broken on master.

vincepandolfo:
Fixed in #80

PR #74, opened 2016-08-06
Overhauled and improved profiling

vincepandolfo:
The aim of this PR is to improve the profiling capabilities of Devito (also fixing issue #65 )

* Profiling information are now stored in a numpy array in the Propagator instead of being printed to stdout
* Profiles the kernel, the time loop stencils (before and after space loops) and the space loops.
* The logic for executing `Propagator.cfunction` has been moved down to the propagator. This should also make easier to execute it when the user uses the low level `Propagator` interface directly.

### DIFF
--- a/devito/operator.py
+++ b/devito/operator.py
@@ -3,7 +3,7 @@ from sympy import (Eq, Function, Indexed, Symbol, lambdify, preorder_traversal,
                    solve, symbols)
 from sympy.abc import t
 
-from devito.compiler import IntelMICCompiler, get_compiler_from_env
+from devito.compiler import get_compiler_from_env
 from devito.interfaces import SymbolicData, TimeData
 from devito.propagator import Propagator
 
@@ -186,6 +186,22 @@ class Operator(object):
         return self.input_params + [param for param in self.output_params
                                     if param not in self.input_params]
 
+    @property
+    def total_kernel_time(self):
+        return self.propagator.timings[0]
+
+    @property
+    def total_space_loops_time(self):
+        return self.propagator.timings[1]
+
+    @property
+    def total_pre_loops_time(self):
+        return self.propagator.timings[2]
+
+    @property
+    def total_post_loops_time(self):
+        return self.propagator.timings[3]
+
     def apply(self, debug=False):
         """:param debug: If True, use Python to apply the operator. Default False.
 
@@ -194,18 +210,12 @@ class Operator(object):
         if debug:
             return self.apply_python()
 
-        f = self.propagator.cfunction
-
         for param in self.input_params:
             if hasattr(param, 'initialize'):
                 param.initialize()
+
         args = [param.data for param in self.signature]
-        if isinstance(self.compiler, IntelMICCompiler):
-            # Off-load propagator kernel via pymic stream
-            self.compiler._stream.invoke(f, *args)
-            self.compiler._stream.sync()
-        else:
-            f(*args)
+        self.propagator.run(args)
 
         return tuple([param.data for param in self.output_params])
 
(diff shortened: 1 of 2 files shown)


devito/propagator.py
```diff
@@ -304,21 +329,37 @@ def generate_loops(self, loop_body):
...
-            )]
-        else:
-            start_time_stmt = []
-            end_time_stmt = []
-
-        initial_block = omp_single + ([cgen.Block(time_stepping + time_loop_stencils_b)]
-                                      if time_stepping or time_loop_stencils_b else [])
-        initial_block = initial_block + start_time_stmt
-        end_block = end_time_stmt + omp_single + ([cgen.Block(time_loop_stencils_a)]
-                                                  if time_loop_stencils_a else end_time_stmt)
```
mloubout: omp_single should have been inside the parenthesis here :
`(omp_single + [cgen.Block(time_stepping + time_loop_stencils_b)] if time_stepping or time_loop_stencils_b else [])`
to avoid empty omp sigle section breaking the compilation. Does your change fix it?

vincepandolfo: Ok I didn't know that. I think my changes breaks it.

devito/operator.py
```diff
@@ -186,6 +186,22 @@ def signature(self):
...
+    @property
+    def total_space_loops_time(self):
+        return self.propagator.timings[1]
+
+    @property
+    def total_pre_loops_time(self):
+        return self.propagator.timings[2]
+
+    @property
+    def total_post_loops_time(self):
```
FabioLuporini: could we document pre_loops and post_loops ? what do they mean?

mloubout: pre_loop and post_loop are operations done outside of the spacial loop :
for (t=...)
    pre_loop
    for (x=.., y=...)
        stencils, factors,...
    endforxyz
    post_loop
endfort

devito/propagator.py
```diff
@@ -85,12 +85,7 @@ def __init__(self, name, nt, shape, stencils, factorized=None, spc_border=0, tim
 
         # Settings for performance profiling
         self.profile = profile
-
-        if self.profile:
-            self.add_local_var("time", "double")
-            self.pre_loop.append(cgen.Statement("struct timeval start, end"))
-            self.pre_loop.append(cgen.Assign("time", 0))
-            self.post_loop.append(cgen.PrintStatement("time: %f\n", "time"))
+        self.timings = np.zeros(4)
```
mlange05: I'm not convinced passing an array here is the best thing to do, since it requires a property to the Operator for each timing we might add later, and it also makes the code slightly confusing. Instead, could we not use a `ctypes.struct`, to which we can just add fields and that we return to the user as a single object? This would also make it easier to inspect in IPython, I think.

vincepandolfo: The problem with that is that I would have to redefine `struct timeval` in ctypes but the definition of `struct timeval` is not platform independent

navjotk:
This needs to be tested on MIC.

vincepandolfo:
Works on MIC

devito/operator.py
navjotk: Please confirm what I am saying but I believe pre and post loops were old names for these variables which were later changed to `time_loop_stencils`. During some rebase/merge conflict resolution these old names seem to have crept back in and right now both versions of these names exist in the code creating redundancy. AFAIK, these old names `pre_loop` and `post_loop` are not being used anywhere in the code and I have been meaning to open a PR to remove them (again). 

If you can confirm the above, please feel free to remove this.

vincepandolfo: To be honest I have not seen those names anywhere in the code. I used them just because they were much shorter than `time_loops_stencils`. 

Right now I am working on completely redoing the profiler by making a profiler class. I will close this PR until the new version is ready (it should be ready tomorrow) if it's not urgent to have it merged.

ggorman:
How did you test this?

vincepandolfo:
@pvelesko did

PR #76, opened 2016-08-07
Gradient test

mloubout:
Rewrite of the Acoustic and TTI operators and codegen to separate operator, model and data.
Gradient test is also back in even so it may be too memory expensive for travis.


mloubout:
Currently fails because of memory and permission error for deployement

mloubout:
Only fails because of memory on linux travis, apple passes

mloubout:
@navjotk @mlange05 
This one need review. 

Time_substitution is still there as this was created from Zhang.

I made a change to `operator` to output the structure (`TimeData`,...) instead of the data. It allow to pass back to gradient directly this structure and avoid useless memory copies. There may be a better way to do it than what I did but this is the main idea.

Finally i moved most of the `u,v,rec`,... to the operators in `fwi_operators` instead of `Acoustic_codegen` as most of it are temporary or input variable there is no reason to have as a property of the operator as it's only needed at computation if called.

examples/Acoustic_codegen.py
```diff
@@ -64,60 +50,37 @@ def damp_boundary(damp):
...
-        self.m.data[:] = self.model.vp**(-2)
-
-        self.damp = DenseData(name="damp", shape=self.model.vp.shape, dtype=self.dtype)
+        self.damp = DenseData(name="damp", shape=self.model.get_shape_comp(), dtype=self.dtype)
 
         # Initialize damp by calling the function that can precompute damping
         damp_boundary(self.damp.data)
-
-        self.src = SourceLike(name="src", npoint=1, nt=self.nt, dt=self.dt, h=self.h,
+        self.src = SourceLike(name="src", npoint=1, nt=data.traces.shape[1], dt=self.dt, h=self.model.get_spacing(),
```
FabioLuporini: We should really try hard to have lines shorter than 90 characters

examples/Acoustic_codegen.py
```diff
@@ -64,60 +50,37 @@ def damp_boundary(damp):
...
+    def Forward(self, save=False):
+        fw = ForwardOperator(self.model, self.src, self.damp, self.data, time_order=self.t_order,
+                             spc_order=self.s_order, save=save)
+        u, rec = fw.apply()
+        return rec.data, u
 
     def Adjoint(self, rec):
-        self.rec.data[:] = rec
-        adj = AdjointOperator(self.m, self.rec, self.damp, self.srca, time_order=self.t_order, spc_order=self.s_order)
+        adj = AdjointOperator(self.model, self.damp, self.data, rec, time_order=self.t_order, spc_order=self.s_order)
```
FabioLuporini: same

FabioLuporini:
I'll go through this properly later, but we should start having smaller, lighter PRs

examples/Acoustic_codegen.py
mloubout: On it

mloubout:
@FabioLuporini 
This one is really hard to make smaller. In practice there is not much as it is only on test (gradient), but it required some rewrite to be able to run it properly without blowing up everything.

mloubout:
The inversion repo is based on this branch so @vincepandolfo (because you worked already on inversion an looked a bit a this branch) and @mlange05 can we get this one going.

devito/compiler.py
```diff
@@ -90,7 +91,7 @@ def __init__(self, *args, **kwargs):
         self.ldflags = ['-shared']
 
         if self.openmp:
-            print "WARNING: Disabling OpenMP because clang does not support it."
+            print("WARNING: Disabling OpenMP because clang does not support it.")
```
vincepandolfo: Use `logger` module here

devito/compiler.py
```diff
@@ -147,7 +147,7 @@ def __init__(self, *args, **kwargs):
         if self.openmp:
             self.ldflags += ['-qopenmp']
         else:
-            print "WARNING: Running on Intel KNL without OpenMP is highly discouraged"
+            print("WARNING: Running on Intel KNL without OpenMP is highly discouraged")
```
vincepandolfo: `logger`

devito/compiler.py
```diff
@@ -206,7 +206,7 @@ def jit_compile(ccode, basename, compiler=GNUCompiler):
     """
     src_file = "%s.cpp" % basename
     lib_file = "%s.so" % basename
-    print "%s: Compiling %s" % (compiler, src_file)
+    print("%s: Compiling %s" % (compiler, src_file))
```
vincepandolfo: `logger`

examples/TTI_codegen.py
```diff
@@ -62,70 +50,26 @@ def damp_boundary(damp):
...
-                                  ndim=len(dimensions), dtype=self.dtype, nbpml=nbpml)
-        if dm_initializer is not None:
-            self.dm = DenseData(name="dm", shape=self.model.vp.shape, dtype=self.dtype)
-            self.dm.data[:] = np.pad(dm_initializer, tuple(pad_list), 'edge')
+        srccoord = np.array(self.data.source_coords, dtype=self.dtype)[np.newaxis, :]
+        self.src = SourceLike(name="src", npoint=1, nt=data.traces.shape[1],
+                              dt=self.dt, h=self.model.get_spacing(),
+                              coordinates=srccoord, ndim=len(self.damp.shape),
+                              dtype=self.dtype, nbpml=nbpml)
+        self.src.data[:] = data.get_source()[:, np.newaxis]
```
vincepandolfo: Address the case when source is not in data but passed as parameter

mloubout: If the source is passed as aparameter it is added to the data (line 30) so this will also do what it is supposed to

tests/test_gradient.py
```diff
@@ -0,0 +1,117 @@
+import numpy as np
+import pytest
+from numpy import linalg
+
+from examples.Acoustic_codegen import Acoustic_cg
+from examples.containers import IGrid, IShot
+
+
+class Test_Gradient(object):
```
FabioLuporini: One shouldn't use `_` for class names -- I *think* it's quite a common rule in Python

tests/test_gradient.py
```diff
@@ -0,0 +1,117 @@
...
+
+    @pytest.fixture(params=[2])
+    def time_order(self, request):
+        return request.param
+
+    @pytest.fixture(params=[2])
+    def space_order(self, request):
+        return request.param
+
+    def test_Grad(self, Acoustic):
```
FabioLuporini: ``test_grad`` (no G)

tests/test_gradient.py
```diff
@@ -0,0 +1,117 @@
...
+import pytest
+from numpy import linalg
+
+from examples.Acoustic_codegen import Acoustic_cg
+from examples.containers import IGrid, IShot
+
+
+class Test_Gradient(object):
+    @pytest.fixture(params=[(70, 80), (60, 70, 80)])
+    def Acoustic(self, request, time_order, space_order):
```
FabioLuporini: maybe ``acoustic`` ?

There's no convention at the moment

tests/test_adjointA.py
```diff
@@ -7,28 +7,28 @@
 
 
 class Test_AdjointA(object):
```
FabioLuporini: No `_` in class names (see also below)

PR #78, opened 2016-08-08
Auto tuning v3

sheino:
Adding auto tuning of block sizes feature, which is made to be more general and potentially could be easily extended.

Added `AutoTuner` object which takes `Operator` and `at_report_dir_path`  and `blocked_dims` as parameters.
To use auto tuning results pass initialised `auto_tuner`  instance as arg for`cache_blocking`  to `Operator`.

Added tests and auto tuning example file.
Updated README.md
Added `info` function to `logger` as @FabioLuporini suggested

Changed cache blocking logic, which is as follows:
Now  `cache_blocking`  has to be either single int instance which will block all dims except inner most or list of same size as spacial domain explicitly specifying which dims to block and which block sizes to use.

If 0 is passed as ant arg for `cache_blocking` then best guess based on architecture is used for block size

When cache blocking\auto tuning all block sizes are replaced by variables which are passed as arguments to the function and loop limits are calculated on run time


sheino:
@navjotk @FabioLuporini @mlange05 Inviting comments!

Particulary whether decide_block_sizes function in propagator is not to complex.

navjotk:
So is the "default behaviour" back to blocking all dimensions?

sheino:
if you pass single block size value, yes. it blocks all dimensions. Should I change it?

navjotk:
Not saying you should. Just commenting on the fact that the last cache blocking related PR made that the default behaviour and this one seems to be going back to the previous default behaviour. Do we have a definite answer now on what we want the default behaviour to be? (I don't - but I think we should resolve this before changing it again)

devito/at_controller.py
```diff
@@ -0,0 +1,245 @@
...
+        # runs function for each block_size
+        for block in block_list:
+            times.append((block, self._time_blocks(f, args, block)))
+
+        logger.info("Auto tuning using brute force complete")
+
+        times = sorted(times, key=lambda element: element[1])  # sorts the list of tuples based on time
+        self._write_b_report(times)  # writes the report
+
+    def _estimate_b_run_time(self, f, args, minimum, maximum):
```
FabioLuporini: I guess ``b`` stands for black -- if that's the case, I'd just use ``block``

devito/at_controller.py
```diff
@@ -0,0 +1,245 @@
...
+
+        for block in block_sizes:  # appends the block size
+            if block is not None:
+                args += [int(block)]
+
+        f(*args)  # run function
+
+        return time.clock() - start  # returns elapsed time
+
+    def _write_b_report(self, times):
```
FabioLuporini: Same ``b``

devito/at_controller.py
```diff
@@ -0,0 +1,245 @@
...
+import cpuinfo
+
+import logger
+
+# global vars
+final_report_name = "final_report.txt"
+default_at_dir = path.join(path.dirname(path.realpath(__file__)), "At Report")
+
+
+def get_at_block_size(f_name, time_order, spc_border, shape, block_dims, at_report_dir=None):
```
FabioLuporini: This perhaps should be a static function in the Autotuner ?

devito/at_controller.py
```diff
@@ -0,0 +1,245 @@
...
+
+                blocks_str = line.split(' ')[5]
+                block_split = blocks_str[1:len(blocks_str) - 2].split(',')
+
+                return [int(block) if block != "None" else None for block in block_split]
+
+    return None  # returns none if no matching block size was found
+
+
+def get_optimal_block_size(cache_blocking, shape, load_c):
```
FabioLuporini: same ?

FabioLuporini:
Very good job.

One minor thing, one major thing, and one observation for the future.

Minor 1: these lines are too long, seriously, they make the code unreadable. We should really make them shorter than at least 90 (PEP says 80 !), by using more variables, being more pythonic or simply by splitting them over multiple lines

Major: the autotune option has made the ``Propagator`` code even more chaotic than it already was, because of all those ``if then else`` which basically change the behaviour depending on whether we are autotuning or not. I *think* this a symptom that we should rather have a different Propagator class, or maybe extend the existing one. 

Observation for the future: we will have to unpick this template-based method of generating code. Let's keep this in mind.

mlange05:
I echo @FabioLuporini's concerns that this is way too invasive to the Propagator and Operator. Do we really have to change the behaviour that much for auto-tuning, and can we not generalise most of this stuff to get rid of all the `if auto_tune:` pieces? Conceptually, the Propagator in particular should have very little to no awareness of auto-tuning and just be invoked repeatedly by the auto-tuner, no?

sheino:
I don't see how's it is that invasive for Operator as it only uses auto_tune flag in apply function to intercept the compiled function and its args. Does not even save `auto_tune` as a var.

For Propagator however, i think `decide_block_sizes` function could potentially be less complex,
however I think all `if auto_tune` statemens in `generate_loops_blocking` are necessary, because we need to replace most loop limit variables with strings that allow calculation of block size on spot. Thats pretty much the only place where Propagator does need to know about auto_tuning.

Alternative could be to write a different method for generating loops which would get called if auto tuning. This would allow us to eliminate 3-4 out of 5 `if auto_tune` statements in Propagator, however it the code for generating blocked loop would be almost duplicated.

sheino:
Also @FabioLuporini about the lines, im more than happy to shorten them howerver if it really bothers you I think we should include this rule in our PEP 8 evaluation, as I just try to make sure lines are bellow 120 as with current rule

devito/at_controller.py
```diff
@@ -0,0 +1,245 @@
...
+            blocks = []
+            for block in self.op.propagator.block_sizes:
+                blocks = blocks + [random.randrange(minimum, maximum)] if block else blocks + [None]
+
+            timing_run += self._time_blocks(f, args, blocks)
+
+        logger.info("Estimated runtime: %f minutes." % float(
+            timing_run / 5 * ((maximum - minimum) * (maximum - minimum)) / 600))
+
+    def _time_blocks(self, f, args, block_sizes):
```
mlange05: I think this Python-level timing becomes obsolete with PR #80, since you can pull this information straight out of the Propagator now, and the timings are created actually come from the kernel itself.

devito/operator.py
```diff
@@ -207,6 +214,11 @@ def apply(self, debug=False):
             if hasattr(param, 'initialize'):
                 param.initialize()
         args = [param.data for param in self.signature]
+
+        # if auto tuning block sizes return compiled function and its arguments to AutoTuner
+        if auto_tune:
+            return f, args
```
mlange05: This hot-wiring is really not great! Is this only required to feed your own custom timer functions, or is there a deeper reason?

examples/auto_tune_example.py
```diff
@@ -0,0 +1,103 @@
...
+        model = IGrid()
+        model.shape = dimensions
+        model.create_model(origin, spacing, true_vp)
+
+        Acoustic = Acoustic_cg(model, data, dm_orig, None, nbpml=10, t_order=2, s_order=2)
+        fw = ForwardOperator(Acoustic.m, Acoustic.src, Acoustic.damp, Acoustic.rec, Acoustic.u,
+                             time_order=time_order, spc_order=space_order, cache_blocking=cache_blocking)
+
+        at = AutoTuner(fw, at_report_dir)
+        at.auto_tune_blocks(5, 26)
```
mlange05: I don't like that we are creating a full copy of an existing example/test case just for the purpose of auto-tuning. If somebody updates the original, this version will be outdated! Can we not add auto-tuning to one of the existing examples, and make it optional via command-line arguments, or something similar?

devito/at_controller.py
sheino: Are you talking about estimating run time of auto tuning or `time_blocks` function itself?

If the later one I don't see how Propagator would know run time when loop limits are variable?

FabioLuporini:
@sheino I agree we should add the PEP 8 rule

what I meant by "unpicking" is that we are copy-pasting code at the moment more than writing an actual compiler. 

As for the rest, I still think we are not doing enough object oriented programming here. If we had a separate class, we could for example just overwrite that generate_loops_something method, or perhaps split it into smaller (private) methods and then overwriting them accordingly. All the branches would eventually disappear. 

What happens for example if we wanted a Propagator that generates code for GPUs ? or that applies some other magic ? I think we need a proper hierarchy and it seems quite quick to do

examples/auto_tune_example.py
```diff
@@ -0,0 +1,103 @@
...
+receiver_coords[:, 2] = location[2]
+data.set_receiver_pos(receiver_coords)
+data.set_shape(nt, 101)
+
+at_report_dir = path.join(path.dirname(path.realpath(__file__)), "At Report")
+cache_blocking = [True, True, False]  # not blocking outer most dim
+
+# auto tuning over all space time orders
+for time_order in xrange(2, 4, 2):
+    for space_order in xrange(2, 16, 2):
```
mloubout: Once again, why is time order 4 still there. It is unconditionally unstable.

examples/auto_tune_example.py
mloubout: I agree, specially considering I'll have to rewrite this completely for my PR.

examples/auto_tune_example.py
sheino: `range` function does not iterate through the last element so 4 is never used.

mloubout: Then why put a loop?

examples/auto_tune_example.py
sheino: Should I just include some flags to `Acoustic_cg init()` or put them at lower level. Like `acoustic.fw(auto_tune=True)` 
?

mloubout: I like the second option better, this is dalso how I set up the saving options for Acoustic an TTI

examples/auto_tune_example.py
sheino: well its unstable now, does it mean that its gonna always stay this way? as I thought its only temporaraly, thus easy to change loop limits later on.

mloubout: I am not sure I understand this.
4th order in time will always be unstable with the current implementation. There is formulations that are stable for 4th order in time but still only uses 3 time step with a different PDE. So it would still come back to the current time stepping sheme but with 2 stencils instead of 1 (4 instead of 2 for TTI) inside the spacial loops.

devito/at_controller.py
vincepandolfo: I think he refers to the `_time_blocks` function where you time the execution of the function in Python. In PR #80 we introduced a more advanced way of doing that, so you might want to use it to time the execution.

sheino:
@navjotk @FabioLuporini @mlange05 inviting comments again as I updated the logic in response to previous comments and discussion on slack

devito/at_controller.py
```diff
@@ -0,0 +1,223 @@
...
+        if maximum - minimum > 2:  # No point to estimate if diff is below 2.
+            self._estimate_run_time(minimum, maximum)
+
+        times = []  # list where times and block sizes will be kept
+        block_list = []  # used to make sure we do not test the same block sizes
+        org_block_sizes = self.op.propagator.block_sizes
+        blocks = list(self.op.propagator.block_sizes)
+
+        for x in range(minimum, maximum):
+            blocks[0] = x if org_block_sizes[0] else None
```
FabioLuporini: can ``org_block_sizes`` ever be an empty list ?

FabioLuporini: Forget it, I think I understand.

Indexing into lists directly as you do here can be difficult to understand sometimes. Maybe use a ``namedtuple`` ?

devito/at_controller.py
```diff
@@ -0,0 +1,223 @@
...
+                    blocks[1] = y if org_block_sizes[1] else None
+
+                    if len(blocks) > 2:
+                        for z in range(minimum, maximum):
+                            blocks[2] = z if org_block_sizes[2] else None
+                            block_list = block_list + [list(blocks)] if blocks not in block_list else block_list
+                    else:
+                        block_list = block_list + [list(blocks)] if blocks not in block_list else block_list
+            else:
+                block_list = block_list + [list(blocks)] if blocks not in block_list else block_list
```
FabioLuporini: you mean (and similarly above) ``block_list += ...``

FabioLuporini: I also wonder whether you can skip most of these if-then-else by using a `set` instead of a `list` or a `tuple`

devito/at_controller.py
```diff
@@ -0,0 +1,223 @@
...
+        :param minimum: int - minimum tune range
+        :param maximum: int - maximum tune range
+        """
+        timing_run = 0  # estimating run time
+        for i in range(0, 5):
+            logger.info('Estimating auto-tuning runtime...sample %d/5' % (i + 1))
+
+            blocks = []
+            for block in self.op.propagator.block_sizes:
+                blocks = blocks + [random.randrange(minimum, maximum)] if block else blocks + [None]
```
FabioLuporini: you sure u don't need brackets here before the `if` and after the `else` ?

devito/at_controller.py
```diff
@@ -0,0 +1,223 @@
...
+        try:
+            full_report_text = ["%s %f\n" % (str(block), timee) for block, timee in times]
+
+            # Cache blocking dimensions
+            cb_dims = [True if block else False for block in self.op.propagator.block_sizes]
+            cb_dims = str(cb_dims).replace(" ", '')
+            shape = str(self.op.shape).replace(" ", '')
+
+            # Writes all auto tuning information into full report
+            with open(path.join(self.report_dir, "%s_time_o_%s_spc_bo_%s_shp_%s_b_%s_report.txt" %
```
FabioLuporini: file name should be a global attribute of the class ? the string template I mean

devito/operator.py
```diff
@@ -83,9 +83,12 @@ class Operator(object):
                      If not provided, the compiler will be inferred from the
                      environment variable DEVITO_ARCH, or default to GNUCompiler.
     :param profile: Flag to enable performance profiling
-    :param cache_blocking: Flag to enable cache blocking
-    :param block_size: Block size used for cache clocking. Can be either a single number used for all dimensions or
-                      a list stating block sizes for each dimension. Set block size to None to skip blocking on that dim
+    :param cache_blocking: Block sizes used for cache clocking. Can be either a single number used for all dimensions
+                        except outer most or a list stating block sizes for each dimension.
```
FabioLuporini: under-indented

devito/operator.py
```diff
@@ -83,9 +83,12 @@ class Operator(object):
...
+                        Set block size to None to skip blocking on that dim
+    :param at_report: string - indicating path to auto tuning report directory.
+                    If used together with cache blocking, block sizes will be retrieved
```
FabioLuporini: under-indented

devito/propagator.py
```diff
@@ -119,6 +117,11 @@ def run(self, args):
 
         f = self.cfunction
 
+        # appends block sizes if cache blocking
+        for block in self.block_sizes:
+            if block is not None:
+                args.append(block)
```
FabioLuporini: that is:
``args += [block for block in self.block_sizes if block]``
or (no creation of new lists)
``args.extend([...])``
?

devito/propagator.py
```diff
@@ -396,8 +402,8 @@ def generate_space_loops_blocking(self, loop_body):
             loop_limits = self._space_loop_limits[spc_var]
 
             if block_size is not None:
+                upper_limit_str = block_var + "+" + orig_var + "block"
```
FabioLuporini: I think we should use string templates *always* because it really helps understanding what's being crafted

Like: ``"%s+%sblock"``

devito/propagator.py
```diff
@@ -435,24 +441,22 @@ def generate_space_loops_blocking(self, loop_body):
...
                 loop_limits = self._space_loop_limits[spc_var]  # Full loop limits
+                lower_limit_str = str(loop_limits[0])
+                upper_limit_str = str(loop_limits[1])
 
                 if block_size is not None:
                     if weights[orig_var] < 0:
                         # already blocked loop limits
-                        loop_limits = (loop_limits[0], loop_limits[1] - (loop_limits[1] - loop_limits[0]) % block_size)
+                        upper_limit_str = "%d - (%d %% %s)" % (loop_limits[1], loop_limits[1] - loop_limits[0],
+                                                               orig_var + "block")
```
FabioLuporini: I've seen this code already. Maybe use a lambda function?

FabioLuporini:
I've added some small comments but nothing too serious

Lines must be below 90, this should be fixed. I'm gonna file an issue now, we must update the pep8 rule

gj !

devito/at_controller.py
sheino: set is a good idea :)

devito/at_controller.py
sheino: well its not used anywhere else except this method. SO no reason to extract it as that file is just for reference. Not used in data retrieving

devito/at_controller.py
sheino: `namedtuple` elements are immutable, thus I would have to create new tuple on each iteration instead of just modifying it on a fly

devito/operator.py
```diff
@@ -194,21 +198,16 @@ def signature(self):
...
-        for param in self.input_params:
-            if hasattr(param, 'initialize'):
-                param.initialize()
+        self.propagator.run(self.get_args())
 
-        args = [param.data for param in self.signature]
-        self.propagator.run(args)
-
-        return tuple([param for param in self.output_params])
+        return tuple([param.data for param in self.output_params])
```
mloubout: This line is the one breaking your branch.
Replace by ` return tuple([param for param in self.output_params])` as it is in master

devito/propagator.py
sheino: I think we discussed with Gabio that is not really necessary

devito/at_controller.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def __init__(self, operator, at_report_dir=None):
+        """
+        Object responsible for auto tuning block sizes.
+        :param operator: Operator object.
+        :param at_report_dir: string - indicating path to auto tuning report directory.
+                              If not set default at report directory is used
+        :raises ValueError: if operator is not of Operator type
+        """
+        global final_report_name, default_at_dir
```
mlange05: I don't think the report name should be global but a property of the AT. What if I need two AT instances with separate files?

devito/at_controller.py
```diff
@@ -0,0 +1,225 @@
...
+
+        from operator import Operator
```
mlange05: Imports go at the top, unless we have a good reason to defer the import.

devito/at_controller.py
```diff
@@ -0,0 +1,225 @@
...
+        if not isinstance(operator, Operator):
+            raise ValueError("Auto tuner requires Operator object to be passed as an argument")
+
+        self.op = operator
+        self.op.propagator.profile = True
+        self.op.apply()  # want to run function at least once to set the arguments
+
+        self.report_dir = at_report_dir if at_report_dir is not None else default_at_dir
```
mlange05: Why not simply `self.report_dir = at_report_dir or default_at_dir`? Or even derive the default dir at this stage to avoid more unnecessary globals?

devito/at_controller.py
```diff
@@ -0,0 +1,225 @@
...
+        :param shape: list - shape of the data buffer
+        :param block_dims: list - indicating which dims are blocked.
+        :param at_report_dir: string - indicating path to auto tuning report directory.
+                              If not set default at report directory is used
+        :raises ValueError: if auto tuning report is not found
+        """
+        global final_report_name, default_at_dir, model_descr_template
+
+        report_dir = at_report_dir if at_report_dir is not None else default_at_dir
+        final_report_path = path.join(report_dir, final_report_name)
```
mlange05: Can the file path not be set as a property in the constructor instead of re-deriving it?

devito/at_controller.py
```diff
@@ -0,0 +1,225 @@
...
+        :param f_name: str - function name. Used for naming
+        :param time_order: int - time order of kernel
+        :param spc_border: int - space border of kernel
+        :param shape: list - shape of the data buffer
+        :param block_dims: list - indicating which dims are blocked.
+        :param at_report_dir: string - indicating path to auto tuning report directory.
+                              If not set default at report directory is used
+        :raises ValueError: if auto tuning report is not found
+        """
+        global final_report_name, default_at_dir, model_descr_template
```
mlange05: `model_descr_template` really doesn't need to be a global.

devito/at_controller.py
```diff
@@ -0,0 +1,225 @@
...
+        :param load_c: int - load count
+        :return: optimal block size
+        """
+        cache_s = int(cpuinfo.get_cpu_info()['l2_cache_size'].split(' ')[0])  # cache size
+        core_c = cpuinfo.get_cpu_info()['count']  # number of cores
+
+        # assuming no prefetching, square block will give the most cache reuse.
+        # We then take the cache/core divide by the size of the inner most dimension in which we do not block.
+        #  This gives us the X*Y block space, of which we take the square root to get the size of our blocks.
+        # ((C size / cores) / (4 * length inner most * kernel loads)
```
mlange05: This description should really go in the docstring above to make it available to everyone without looking at the code.

devito/at_controller.py
```diff
@@ -0,0 +1,225 @@
...
+
+        return None  # returns none if no matching block size was found
+
+    @staticmethod
+    def get_optimal_block_size(shape, load_c):
+        """
+        Gets optimal block size based on architecture
+        :param shape: list - shape of the data buffer
+        :param load_c: int - load count
+        :return: optimal block size
```
mlange05: While we're at docstrings, I believe we are using a convention of:
```
<One-line or short description of the function>
<empty line>
<paramter list>
<empty line>
<Additional comments, including more detailed descriptions of features>
```

devito/at_controller.py
```diff
@@ -0,0 +1,225 @@
...
+        :param minimum: int - minimum tune range
+        :param maximum: int - maximum tune range
+        """
+        timing_run = 0  # estimating run time
+        for i in range(0, 5):
+            logger.info('Estimating auto-tuning runtime...sample %d/5' % (i + 1))
+
+            blocks = []
+            for block in self.op.propagator.block_sizes:
+                blocks += [random.randrange(minimum, maximum)] if block else [None]
```
mlange05: Could turn this into a list comprehension to make it more pythonic. :snake:

devito/at_controller.py
```diff
@@ -0,0 +1,225 @@
...
+                    final_report.write(str_to_write)  # writes the string
+                return
+            else:
+                with open(self.final_report_path, 'r') as final_report:  # reads all the contents
+                    lines = final_report.readlines()
+
+                # checks whether entry already exist and updates it. Otherwise appends to the end of the file
+                entry_found = False
+
+                for i in range(1, len(lines)):
```
mlange05: Have you considered using a non-custom file format to utilise external packages, like `json` or numpy's `csv` reader? I personally like `json` which effectively pretty-prints and reads nested Python dictionaries, which can be made to be nicely human readable and the read in result data can be directly accessed from the resulting dict. For example you could do:
```
profile = {'timestep': <my time stamp>,
           'block_size_range': ...
           'data' = {(3, 6. None): <timing>,
                     (6, None, 7): <timing>,
                      ...
                     }
           }

File: devito/propagator.py [hunk 28]
```diff
@@ -480,6 +483,48 @@ def generate_space_loops_blocking(self, loop_body):
 
         return [loop_body] + full_remainder if full_remainder else [loop_body]
 
+    def get_number_of_loads(self):
+        """Gets number of loads in the kernel"""
+        return len(self._kernel_dic_oi["load_all_list"])
```
mlange05: Does this still exist with the recent `Profiler` changes? Should we not look at the `Profiler` instead?

devito/at_controller.py
sheino: It's a static method which I inteded to be used without initialising auto tuner thus It need to be rederived.

devito/propagator.py
sheino: Asked @vincepandolfo and he confirmed that it still exists and `Profiler` does not provide that info

devito/at_controller.py
sheino: I didn't think about it as at the time i wrote this function we were still using isat and the cache_blocking logic was different.
Is it neccessary considering that this is working ok?

devito/at_controller.py
sheino: for some reason That i fail to understand if i put this on top. I get an error that import has failed

devito/at_controller.py
```diff
@@ -0,0 +1,211 @@
...
+        :param op: Operator object.
+        :param blocked_dims: list indicating which dims are blocked and we want to tune
+                             Default all except inner most.
+        :param at_report_dir: string - indicating path to auto tuning report directory.
+                              If not set current directory is used
+        :param blocked_dims: list indicating which dimensions we want to auto tune
+        :raises ValueError: if operator is not of Operator type
+        """
+
+        from operator import Operator
```
mlange05: Relative imports are evil! This should be `from devito.operator import Operator` and go to the top.

devito/at_controller.py
```diff
@@ -0,0 +1,211 @@
...
+        default_at_dir = path.join(path.dirname(path.realpath(__file__)), "At Report")
+        self.report_dir = None
+        self.final_report_path = None
+        self.set_report_path(at_report_dir or default_at_dir)
+        self.model_desc_template = "%s %d %d %s %s"
+
+        if not path.isdir(self.report_dir):  # Creates report dir if does not exist
+            mkdir(self.report_dir)
+
+    def set_report_path(self, report_path):
```
mlange05: This little utility routine seems superfluous. Why not do this all in-place above?

devito/at_controller.py
```diff
@@ -0,0 +1,211 @@
...
+        :param minimum: int (optional) - minimum value for auto tuning range. Default 5
+        :param maximum: int (optional) - maximum value for auto tuning range. Default 20
+        :raises ValueError: if  minimum is >= maximum
+        """
+        if minimum >= maximum:
+            raise ValueError("Invalid parameters. Min tune range has to be less than Max")
+
+        self.op.propagator.cache_blocking = [0 if dim else None
+                                             for dim in self.blocked_dims]
+        self.op.apply()  # want to run function at least once to set the arguments
```
mlange05: I don't quite understand the need for this. Can't we just call `get_args` on the Operator/Propagator for this without running the entire setup?

devito/at_controller.py
```diff
@@ -0,0 +1,211 @@
...
+
+            blocks = []
+            blocks += [random.randrange(minimum, maximum) if block else None
+                       for block in self.op.propagator.block_sizes]
+
+            self.op.propagator.block_sizes = blocks
+            timing_run += self.get_execution_time()
+
+        print timing_run / 5
+        print (timing_run / 5 * 400) / 60
```
mlange05: Debug leftover? If not, please use the logger.

devito/at_controller.py
```diff
@@ -0,0 +1,211 @@
...
+                                       "shape, block dims, best block size\n")
+                    final_report.write(str_to_write)  # writes the string
+                return
+            else:
+                # reads all the contents
+                with open(self.final_report_path, 'r') as final_report:
+                    lines = final_report.readlines()
+
+                # checks whether entry already exist and updates it.
+                #  Otherwise appends to the end of the file
```
mlange05: Comment indentation

devito/at_controller.py
sheino: I have no idea why but i get an import error if I put in on top

mlange05: It's because you are using relative imports, which is shadowed by Python's builtin `operator` module. This is the reason why relative imports are evil.

FabioLuporini:
this has been around for a while, but it looks like we're pretty close to merge. Let's do it quickly!

devito/at_controller.py
sheino: Even if i put `from devito.operator import Operator` to the top I still get import error

devito/at_controller.py
sheino: Comment was not clear enough. We need to run this at least once so that block size arg when explicitly set woud not be owerwritten. As it is set based on cache_blocking argument
Im gonna change into propagator.run.

devito/at_controller.py
FabioLuporini: May be a cyclic dependency. Paste the error ?

mloubout:
Can we add to `acoustic_example` how we would use it in addition to the test?

sheino:
Sure, do you want it to be simple and just commented out as we dont want to run auto tuning every time we want to test acoustic or controlled by significant amount of flags

devito/at_controller.py
sheino: Indeed it was, but fixed now.

examples/acoustic_example.py
```diff
@@ -73,6 +73,7 @@ def source(t, f0):
 data.set_shape(nt, 101)
 Acoustic = Acoustic_cg(model, data)
 (rec, u) = Acoustic.Forward(save=True)
+# (rec, u) = Acoustic.Forward(save=True, use_at_blocks=True)
```
mlange05: Debug?

examples/Acoustic_codegen.py
```diff
@@ -62,28 +61,43 @@ def damp_boundary(damp):
...
+    def Forward(self, save=False, cache_blocking=None, use_at_blocks=False):
         fw = ForwardOperator(self.model, self.src, self.damp, self.data,
                              time_order=self.t_order, spc_order=self.s_order,
-                             save=save)
+                             save=save, cache_blocking=cache_blocking)
+        at = AutoTuner(fw)
+
+        if use_at_blocks:
+            fw.propagator.cache_blocking = at.block_size
+
```
mloubout: remove `at = Autotuner(fw)`
`self.at` inside if

sheino: My bad. Good spot

PR #80, opened 2016-08-09
Overhauled and improved profiling

vincepandolfo:
- Added a Profiler class to manage the profiling of the code.
- Moved the logic for running the `cfunction` to `Propagator`.
- Added profiling for time loop stencils, space loop and kernel.
- Modified FunctionDescriptor and FunctionManager to accept struct
  parameters (and also fixed to also add scalar parameters that 
  weren't added by the manager)

### DIFF
--- a/devito/profiler.py
+++ b/devito/profiler.py
@@ -3,13 +3,6 @@ from ctypes import Structure, byref, c_double
 from cgen_wrapper import Block, Statement, Struct, Value
 
 
-def _get_timings_class(fields):
-    """Returns a :class:`ctypes.Structure` subclass defining our structure
-    """
-
-    return type("Timings", (Structure,), {"_fields_": fields})
-
-
 class Profiler(Structure):
     """The Profiler class is used to manage profiling information for Devito
     generated C code.
@@ -31,7 +24,7 @@ class Profiler(Structure):
         :returns: A list of :class:`cgen.Generable` with the added profiling
                   code.
         """
-        self.fields.append(name)
+        self.fields.append((name, c_double))
 
         omp_flag = omp_flag or []
 
@@ -51,6 +44,14 @@ class Profiler(Structure):
         return init + block + end
 
     @property
+    def get_timings_class(self):
+        """Returns a :class:`ctypes.Structure` subclass defining our structure
+
+        :returns: A :class:`Timings` class definition
+        """
+        return type("Timings", (Structure,), {"_fields_": self.fields})
+
+    @property
     def as_cgen_struct(self):
         """Returns the :class:`cgen.Struct` relative to the profiler
 
@@ -58,7 +59,7 @@ class Profiler(Structure):
         """
         fields = []
 
-        for name in self.fields:
+        for name, _ in self.fields:
             fields.append(Value("double", name))
 
         return Struct("profiler", fields)
@@ -69,14 +70,7 @@ class Profiler(Structure):
 
         :returns: The pointer
         """
-        fields = []
-
-        for name in self.fields:
-            fields.append((name, c_double))
-
-        timings_class = _get_timings_class(fields)
-
-        self.timings = timings_class()
+        self.timings = self.get_timings_class()
 
         return byref(self.timings)
 
--- a/devito/profiler.py
+++ b/devito/profiler.py
@@ -1,0 +1,88 @@
+from ctypes import Structure, byref, c_double
+
+from cgen_wrapper import Block, Statement, Struct, Value
+
+
+def _get_timings_class(fields):
+    """Returns a :class:`ctypes.Structure` subclass defining our structure
+    """
+
+    return type("Timings", (Structure,), {"_fields_": fields})
+
+
+class Profiler(Structure):
+    """The Profiler class is used to manage profiling information for Devito
+    generated C code.
+    """
+
+    def __init__(self):
+        self.name = "timings"
+        self.fields = []
+
+    def add_profiling(self, block, name, omp_flag=None):
+        """Function to add profiling code to the given :class:`cgen.Block`.
+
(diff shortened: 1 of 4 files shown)


vincepandolfo:
Here's an example code (from test_save.py) generated with profiling on, for reference:

```
#include <cassert>
#include <cstdlib>
#include <cmath>
#include <iostream>
#include <fstream>
#include <vector>
#include <cstdio>
#include <string>
#include <inttypes.h>
#include <sys/time.h>
#include <math.h>
struct profiler
{
  double loop_stencils_a;
  double loop_stencils_b;
  double loop_body;
  double kernel;
} ;
extern "C" int Operator(float *u_vec, struct profiler *timings)
{
  float (*u)[100][100] = (float (*)[100][100]) u_vec;
  {
    struct timeval start_kernel, end_kernel;
    gettimeofday(&start_kernel, NULL);
    for (int i3 = 0; i3<100; i3+=1)
    {
      {
        struct timeval start_loop_stencils_b, end_loop_stencils_b;
        gettimeofday(&start_loop_stencils_b, NULL);
        {
          gettimeofday(&end_loop_stencils_b, NULL);
          timings->loop_stencils_b += (end_loop_stencils_b.tv_sec - start_loop_stencils_b.tv_sec) * 1000000u + (end_loop_stencils_b.tv_usec - start_loop_stencils_b.tv_usec);
        }
      }
      struct timeval start_loop_body, end_loop_body;
      gettimeofday(&start_loop_body, NULL);
      {
        for (int i1 = 1; i1<99; i1++)
        {
          #pragma GCC ivdep
          for (int i2 = 1; i2<99; i2++)
          {
            u[i3][i1][i2] = 1.11022302462516e-16F*u[i3 - 1][i1][i2] + 2.5e-1F*u[i3 - 1][i1][i2 - 1] + 2.5e-1F*u[i3 - 1][i1][i2 + 1] + 2.5e-1F*u[i3 - 1][i1 - 1][i2] + 2.5e-1F*u[i3 - 1][i1 + 1][i2];
          }
        }
      }
      {
        gettimeofday(&end_loop_body, NULL);
        timings->loop_body += (end_loop_body.tv_sec - start_loop_body.tv_sec) * 1000000u + (end_loop_body.tv_usec - start_loop_body.tv_usec);
      }
      {
        struct timeval start_loop_stencils_a, end_loop_stencils_a;
        gettimeofday(&start_loop_stencils_a, NULL);
        {
          gettimeofday(&end_loop_stencils_a, NULL);
          timings->loop_stencils_a += (end_loop_stencils_a.tv_sec - start_loop_stencils_a.tv_sec) * 1000000u + (end_loop_stencils_a.tv_usec - start_loop_stencils_a.tv_usec);
        }
      }
    }
    {
      gettimeofday(&end_kernel, NULL);
      timings->kernel += (end_kernel.tv_sec - start_kernel.tv_sec) * 1000000u + (end_kernel.tv_usec - start_kernel.tv_usec);
    }
  }
  return 0;
}
```

devito/propagator.py
```diff
@@ -165,6 +181,10 @@ def cfunction(self):
         return self._cfunction
 
     @property
+    def timings(self):
+        return self.profiler.as_dictionary
```
mlange05: Accessing this property fails if we ran with `profile=False`. Can we make sure this is handled more gracefully, eg. return `None` or similar.

devito/profiler.py
```diff
@@ -0,0 +1,88 @@
...
+
+        init = [
+            Statement("struct timeval start_%s, end_%s" % (name, name))
+        ] + omp_flag + [Statement("gettimeofday(&start_%s, NULL)" % name)]
+
+        end = omp_flag + [
+            Block([
+                Statement("gettimeofday(&end_%s, NULL)" % name),
+                Statement(("%(sn)s->%(n)s += (end_%(n)s.tv_sec - start_%(n)s.tv_sec) * " +
+                           "1000000u + (end_%(n)s.tv_usec - start_%(n)s.tv_usec)") %
```
mlange05: We seem to return values that are way too big, ie. `* 1000000`. What's the rational behind this, and/or should we be dividing the final value by this again?

devito/profiler.py
```diff
@@ -0,0 +1,88 @@
+from ctypes import Structure, byref, c_double
+
+from cgen_wrapper import Block, Statement, Struct, Value
+
+
+def _get_timings_class(fields):
```
mlange05: Could make this a class property, but that's just a minor style thing.

devito/profiler.py
vincepandolfo: I used the formula that was used in the old profiler. I don't know where it comes from.

mlange05:
Apart from the in-line comments I think this looks really great and very useful.

devito/propagator.py
vincepandolfo: Addressed in https://github.com/opesci/devito/pull/80/commits/8b21672a3e415e0008fc0fdbf9578cc3d0f4f4b8

devito/profiler.py
mlange05: Fair enough, but the returned result still looks wrong. Can people please chime in on why and/or how to do this right, @navjotk, @felippezacarias, @pvelesko, @FabioLuporini?

FabioLuporini: Do we want times in seconds, milliseconds, or microseconds ? It boils down to this, but the formula looks correct

Currently it seems we are printing times in microseconds (not sure why)

vincepandolfo: Ok I get why it is done. The result is in microseconds.

mlange05: Ok cool. Can we return the final result in seconds, please?

mlange05:
OK, this looks great now. Good to go once travis is green.

PR #81, opened 2016-08-10
Refinement over use of memmap

pp1336:
Add some comments.
Add the ability to have files of repeated name, which happened in acoustic_example(though in this case it may not cause an issue, but similar cases can happen in the future).
Add the ability to retain created file, default behaviour is to delete all files created.

Please ignore changes at places other than devito/interface.MemmapManager, those are just merge conflict.

### DIFF
--- a/devito/interfaces.py
+++ b/devito/interfaces.py
@@ -5,8 +5,8 @@ from sympy import Function, IndexedBase, as_finite_diff
 from sympy.abc import h, p, s, t, x, y, z
 
 from devito.finite_difference import cross_derivative, first_derivative
-from tools import aligned
 from devito.memmap_manager import MemmapManager
+from tools import aligned
 
 __all__ = ['DenseData', 'TimeData', 'PointData']
 
--- a/devito/interfaces.py
+++ b/devito/interfaces.py
@@ -478,17 +478,28 @@ class MemmapManager():
     """Class for managing all memmap related settings"""
     # used to enable memmap as default
     _use_memmap = False
+    # determine whether created files are deleted by default
+    _delete_file = True
     # flag for registering exit func
     _registered = False
     _default_disk_path = os.path.join(gettempdir(), "devito_disk")
     # contains str name of all memmap file created
-    _created_data = set()
+    _created_data = {}
+    # unique id
+    _id = 0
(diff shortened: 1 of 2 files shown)


vincepandolfo:
Can you fix the git history for this PR? It's pretty difficult to see what changes you made. Rebase your changes on top of master

pp1336:
Please just look at changes in devito/interface.MemmapManager. I will do a rebase when previous PR are merged.

devito/interfaces.py
```diff
@@ -244,42 +439,64 @@ def __init__(self, *args, **kwargs):
...
 
         :param shape: Shape of the spatial data
+        :return: indices used for axis.
         """
         _indices = [t, x, y, z]
+
         return _indices[:len(shape) + 1]
 
 
 class MemmapManager():
```
pp1336: Just here

vincepandolfo:
Can you rebase this PR properly before too much stuff gets merged in master making it even more difficult?

pp1336:
Branch rebased.

devito/interfaces.py
```diff
@@ -474,11 +474,17 @@ class MemmapManager():
...
     # flag for registering exit func
     _registered = False
+    # default directory to store memmap file
     _default_disk_path = os.path.join(gettempdir(), "devito_disk")
     # contains str name of all memmap file created
-    _created_data = set()
+    _created_data = {}
+    # unique id
+    _id = 0
+    # exit code used for normal exiting
```
mlange05: This class is getting increasingly complex, so I would propose to move it to it's own file.

devito/interfaces.py
```diff
@@ -510,22 +524,27 @@ def setup(data_self, *args, **kwargs):
...
-            try:
-                os.remove(f)
-            except OSError:
-                print("error removing " + f + " it may be already removed, skipping")
-                pass
+            if MemmapManager._created_data[f]:
+                try:
+                    os.remove(f)
+                except OSError:
+                    print("error removing " + f + " it may be already removed, skipping")
```
mlange05: This should be using the logger. And there is no need for `pass` here.

pp1336:
Rebased again.

PR #82, opened 2016-08-10
Fixed bug introduced in #80

vincepandolfo:
Value parameters were added twice to function signature


mlange05:
Since it changes the MIC code, has this been run and verified with pyMIC?

vincepandolfo:
It does not change it. It is the same code that was already in function_manager.

FabioLuporini:
looks ok

mlange05:
Looks good. Merging...

Issue #84, opened 2016-08-12
Memory is not freed/keeps getting reallocated when creating SymbolicData objects


vincepandolfo:
As demonstrated by this snippet:
```
import gc
import numpy as np
import resource
from sympy import Eq

from devito.interfaces import DenseData
from devito.operator import Operator


def init(data):
    data = np.zeros(data.shape)


def test_memory(nx=1000, ny=1000):

    for i in range(10):
        gc.collect()

        u = DenseData(name='u', shape=(nx, ny), dtype=np.float64, space_order=2, initializer=init)

        eqn = Eq(u, u + 3)

        op = Operator(stencils=eqn, shape=(nx, ny), nt=10, spc_border=1)

        op.apply()

        print(resource.getrusage(resource.RUSAGE_SELF).ru_maxrss)
```

Output:
```
51084
64720
80364
88188
96020
103848
111692
119524
127348
135184
```

The memory for the `u` object keeps getting reallocated.

This is probably caused by the caching of Symbolic objects.

vincepandolfo:
The issue seems to be that the Symbolic cache keeps adding `u` to the cache instead of reusing the cached instance

mlange05:
Ok, so this is partly by design, since the actual `TimeData` constructors get called explicitly by the user, while the cache is designed to catch SymPy's implicit re-instantiation. So, I think the right way to do this is to enforce clean deletion.

However, the clean deletion problem goes further than just the symbol cache, since the refcount for `u` goes up during the internal symbolic manipulation in the Operator/Propagator; it's 11 once we have built the Operator. Investigation continues...

vincepandolfo:
I have a fix. It's ugly, but fixes the problem

mlange05:
Ok, I can confirm that this does indeed keep the memory volume down. I'm still not convinced we should cache in the natural constructor, and the fix is indeed very hacky, but it's a good start.

vincepandolfo:
A possible solution would be to have a method `clean` that needs to be called explicitly by the user since there's no guarantee on if and when `__del__` is called.

mlange05:
Yes, that sounds like a good intermediate solution. I'd love to get this fixed properly, but in the meantime this should work. Can you please implement it and submit a PR?

mlange05:
[...]

PR #86, opened 2016-08-15
Added clean() method to `SymbolicData` and `Operator`

vincepandolfo:
This provides a solution to issue #84 

The clean method needs to be explicitly called by the user. 
Otherwise, the symbolic cache keeps expanding and keeping a reference to every data object ever created, blowing up the memory usage.

### DIFF
--- a/devito/interfaces.py
+++ b/devito/interfaces.py
@@ -1,6 +1,7 @@
 import atexit
 import os
 import sys
+import weakref
 from signal import SIGABRT, SIGINT, SIGSEGV, SIGTERM, signal
 from tempfile import gettempdir
 
@@ -72,7 +73,7 @@ class SymbolicData(Function):
     def __init__(self):
         """Initialise from a cached instance by shallow copying __dict__."""
         original = _SymbolCache[self.__class__]
-        self.__dict__ = original.__dict__.copy()
+        self.__dict__ = original().__dict__.copy()
 
     @classmethod
     def _cached(cls):
@@ -85,7 +86,7 @@ class SymbolicData(Function):
 
         :param obj: Object to be cached.
         """
-        _SymbolCache[cls] = obj
+        _SymbolCache[cls] = weakref.ref(obj)
 
     @classmethod
     def indices(cls, shape):
--- a/tests/test_symbolic_data.py
+++ b/tests/test_symbolic_data.py
@@ -3,7 +3,8 @@ import pytest
 from sympy import Derivative, as_finite_diff, simplify
 from sympy.abc import h, t, x, y, z
 
-from devito import DenseData, TimeData
+from devito import DenseData, TimeData, clear_cache
+from devito.interfaces import _SymbolCache
 
 
 @pytest.fixture
@@ -60,3 +61,12 @@ def test_second_derivatives_space(derivative, dimension, order):
     s_expr = as_finite_diff(u.diff(dimension, dimension), indices)
     assert(simplify(expr - s_expr) == 0)  # Symbolic equality
     assert(expr == s_expr)  # Exact equailty
+
+
+def test_clear_cache(nx=1000, ny=1000):
+    for i in range(10):
+        clear_cache()
+
+        DenseData(name='u', shape=(nx, ny), dtype=np.float64, space_order=2)
+
+        assert(len(_SymbolCache) == 1)
--- a/devito/__init__.py
+++ b/devito/__init__.py
@@ -1,5 +1,19 @@
 from __future__ import absolute_import
+import gc
+
+from sympy.core import cache
+
 from devito.interfaces import *  # noqa
+from devito.interfaces import _SymbolCache
 from devito.operator import *  # noqa
 from devito.finite_difference import *  # noqa
 from devito.iteration import *  # noqa
+
+
+def clear_cache():
+    cache.clear_cache()
+    gc.collect()
+
+    for key, val in _SymbolCache.items():
+        if val() is None:
+            del _SymbolCache[key]


mlange05:
Generally this looks good. We could move the test into the existing `test_symbolic_data.py` to keep the test clutter down, but that's just a small observation.

Also, for the record, this is a **temporary fix**, and the problem is **not** that we add new `TimeData` objects to the cache if they have the same `name` attribute as an existing one, as the test seems to suggest. The real problem is that the ref count for our data objects goes up under symbolic manipulation, so that calling `__del__` never gets invoked on our symbolic objects, which needs to be fixed ASAP.

vincepandolfo:
Fails on travis since no test calls `clean()`. Apparently there are 867 objects in the symbolic cache at the end of the test suite.

`__del__` is never going to get called if we keep a reference in the symbolic cache. We need to find a solution to that first.

mloubout:
This explains why test_gradient had memory errors on TRAVIS

mlange05:
The solution to that is using `weakrefs`, but there seems to be more going on than just the one extra ref from the symbol cache.

vincepandolfo:
I tried using `weakref` but that seems to break sympy. I'll do more testing today

tests/test_symbolic_data.py
```diff
@@ -60,3 +61,18 @@ def test_second_derivatives_space(derivative, dimension, order):
...
+    data = np.zeros(data.shape)
+
+
+def test_symbol_cache(nx=1000, ny=1000):
+    target_len = len(devito.interfaces._SymbolCache)
+
+    for i in range(10):
+        u = DenseData(name='u', shape=(nx, ny), dtype=np.float64, space_order=2, initializer=init)
+
+        assert(len(devito.interfaces._SymbolCache) == target_len + 1)
```
mlange05: Should we not simply add the `delete()` to all other tests, rather than this hack?

vincepandolfo: This would make this solution somewhat less temporary.
Also, if we are going to use `weakref`, the `weakref` object will still be stored in the SymbolCache, increasing the length.

devito/interfaces.py
```diff
@@ -72,7 +73,20 @@ def __new__(cls, *args, **kwargs):
...
         original = _SymbolCache[self.__class__]
-        self.__dict__ = original.__dict__.copy()
+        self.__dict__ = original().__dict__.copy()
+
+    def clean(self):
+        """Cleans up a :class:`SymbolicData` object by removing
+        it from the symbolic cache.
+        """
+        if self._cached():
+            self._clean()
```
mlange05: Ok, now that we implement `__del__` I think we don't need to expose a `clean` function on the objects. Instead I would move this logic to `__del__`.

devito/operator.py
```diff
@@ -193,6 +197,15 @@ def signature(self):
         return self.input_params + [param for param in self.output_params
                                     if param not in self.input_params]
 
+    def clean(self, clean_params=False):
+        """Cleans up all the data object parameters in the operator signature
```
mlange05: Slightly misleading docstring, since the primary functionality is to clear the sympy cache.

devito/propagator.py
```diff
@@ -205,6 +209,9 @@ def save(self, save):
             for i, t_var in enumerate(reversed(self.time_steppers)):
                 self.t_replace[self.time_dim - i*self._time_step] = t_var
 
+            for i in range(1, len(self.time_steppers)):
+                self.t_replace[self.time_dim + i*self._time_step] = self.time_steppers[i - abs(self._time_step)]
```
mlange05: I don't think this should be part of this PR.

tests/test_symbolic_data.py
```diff
@@ -60,3 +61,18 @@ def test_second_derivatives_space(derivative, dimension, order):
     s_expr = as_finite_diff(u.diff(dimension, dimension), indices)
     assert(simplify(expr - s_expr) == 0)  # Symbolic equality
     assert(expr == s_expr)  # Exact equailty
+
+
+def init(data):
```
mlange05: Why do we need an initializer for this test?

tests/test_symbolic_data.py
mlange05: I think this assert now needs to changed back to something absolute.

tests/test_symbolic_data.py
```diff
@@ -60,3 +61,18 @@ def test_second_derivatives_space(derivative, dimension, order):
...
+
+def test_clean_data(nx=1000, ny=1000):
+    target_len = len(devito.interfaces._SymbolCache)
+
+    for i in range(10):
+        u = DenseData(name='u', shape=(nx, ny), dtype=np.float64, space_order=2, initializer=init)
+
+        assert(len(devito.interfaces._SymbolCache) == target_len + 1)
+
+        u.clean()
```
mlange05: Not needed anymore.

devito/interfaces.py
vincepandolfo: `__del__` is not inherited. We would have to make a new `__del__` function for every SymbolicData subclass if we move this logic there

tests/test_symbolic_data.py
vincepandolfo: There's no guarantee on how many elements will be in `_SymbolCache` when this function is called (since there's no guarantee on when `__del__` will be called on other tests objects). Changing it to something absolute won't work.

devito/propagator.py
vincepandolfo: I'll move this to a different PR

devito/interfaces.py
mlange05: Ok, so I don't think that's true. I just build a custom `__del__` on `_SymbolicData` locally and it works fine with your test, provided I clear the sympy cache properly.

vincepandolfo: I must have misread the documentation. Should I remove `clear_params` from `Operator.clear`?

mlange05:
Why did you remove the test? It can be made to work, if you manually clear the sympy cache at the start, you always get `len(_SymbolCache)` of 0 initially and 1 afterwards.

vincepandolfo:
I thought by "not needed anymore" you meant the test and not the `u.clean()` line... lol

vincepandolfo:
Also, do you mean manually clear `_SymbolCache`? Clearing the sympy cache doesn't work for me

mlange05:
No, I had an inherited `__del__` on all `_SymbolicData` objects, so that GC kicks in once the sympy cache is cleaned. You can force that at the start of the test, before the loop, which will clean the symbol cache implicitly due to the `__del__` calling `_clean()`.

devito/interfaces.py
mlange05: No, so to recap, `Operator.__del__` invokes `Operator.clear`, so that we auto-clear the sympy cache at the end of each test/inversion loop, right? Can we confirm that works?

The `clear_params` then might still be useful, just make sure you call the removal explicitly, ie. `del param`.

vincepandolfo:
I added `__del__` to SymbolicData but it doesn't seem to work

vincepandolfo:
Manually cleaning the Sympy cache at the start doesn't work. There's no guarantee that GC will kick in (it doesn't on my machine, for example) and there's also no guarantee that it will collect at every iteration (if we don't force it to)

devito/interfaces.py
vincepandolfo: Yes, that works.

mlange05:
This looks great now, very concise and clean. Merging...

PR #90, opened 2016-08-19
MCells

sheino:
Added simple method for logging performance metrics. Prints MCells and GFlops if profiling.
Rebased onto PR89, so should go in after it.

### DIFF
--- a/devito/propagator.py
+++ b/devito/propagator.py
@@ -14,6 +14,7 @@ from devito.compiler import (IntelMICCompiler, get_compiler_from_env,
                              get_tmp_dir, jit_compile_and_load)
 from devito.function_manager import FunctionDescriptor, FunctionManager
 from devito.iteration import Iteration
+from devito.logger import logger
 from devito.profiler import Profiler
 from tools import flatten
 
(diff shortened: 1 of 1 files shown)


devito/propagator.py
```diff
@@ -126,6 +129,20 @@ def run(self, args):
...
+        if self.profile:
+            self.log_performance()
+
+    def log_performance(self):
+        """Logs performance metrics if profiling"""
+        shape_str = str(self.shape).replace(', ', ' x ')
+        cb_str = ", blocks - %s " % str(self.cache_blocking) if self.cache_blocking else ' '
+        time = self.timings['kernel']
+        iteration_space = map(lambda dim: dim - self.spc_border * 2, self.shape)
+        mcells = int(round(self.nt * np.prod(iteration_space)) / (float(time) * 10 ** 6))
```
vincepandolfo: Time is already float so there's no need for the explicit conversion here

devito/propagator.py
```diff
@@ -15,6 +15,7 @@
 from devito.function_manager import FunctionDescriptor, FunctionManager
 from devito.iteration import Iteration
 from devito.profiler import Profiler
+from devito.logger import logger
```
vincepandolfo: This should go before line 17

mlange05:
Neat. Good to go once travis is green.

Issue #43, opened 2016-07-18
Generic symbolic dimensions  


mlange05:
The default symbols used to represent the different spatial and temporal dimensions work fine for our current examples and test cases, but restrict users that may want to implement other, more complicated equations or time-stepping schemes. A more dynamic and customisable approach will facilitate further implementations and should allow us to significantly reduce the complexity of the `Propagator`. This, however, requires careful design which can be discussed in this issue.

My current thinking is based around the idea of introducing some type of `Dimension` class that, similar to our `SymbolicData` objects, associates SymPy dimension symbols (`x`, `y`, `z`, `t`) with meta-data about dimension sizes and loop iteration spaces. Such a model might work something like this:
- A Propagator maintains a list of `Dimensions`, which it can query to generate the appropriate C loop code, reducing the complexity of the Propagator code.
- A `Dimension` might yield multiple `IterationSpace` objects that represent sub-loops for loop blocking. These can be created according to various blocking schemes and across multiple dimensions.
- Individual stencil equations coming into the Operator can be scanned for `Dimension` symbols, which allows us to capture _free and bound variables_ and will allow us to automatically insert the stencil at the right depth in the loop nest. This will help with sanity checks for _operator merging_ and inserting sparse operations, ie. source and receiver terms, more gracefully.
- Additional SymPy expressions might be used to define custom stepping schemes for individual iteration spaces, which will allow user to implement custom time-stepping algorithms.

As stated above, this issue is intended for discussing and drafting a possible implementation, so comments and suggestions are highly welcome.

vincepandolfo:
Should a `Dimension` object represent a loop on top of the ones defined by the `Iteration` objects?

If yes, I was thinking of a structure like this:
 - `self.iterations` list of `Iteration` objects for the sub loops
 - `self.stencils` list of stencils with the main code for the Dimension loop
 - `self.pre_loop_stencils` list of stencils to be executed before the dimension loop
 - `self.post_loop_stencils` list of stencils to be executed after the dimension loop

mlange05:
No, a dimension in this case is a purely symbolic object that represents an iteration space, ie. `x` or `y` or `t` in our current examples. Thus it behaves like a symbol that we we use to index our symbolic data / function objects, but it knows about the bounds in a certain spatial dimension for example, including things like padding layers. As such it would be able to generate `Iteration` objects that represent actual loops in the code over a certain iteration space.

[...]

PR #92, opened 2016-08-19
New dimension-agnostics code generator 

mlange05:
This merge introduces an alternative "Propagator-style" code generator that is proof of concept that we can treat dimensions generically at the lower ("indexed") API level. This is a first step towards issue #43 and will enable the future integration of non-standard loop nests that break the current pre-defined time-pre-stencil-post loop construct, as well as further integration with other generic compiler-style abstractions/packages. Crucially, this PR only contains a very trivial prototype, as well as all the outside trimmings to make this work seamlessly with the current outer symbolics layer.

In detail we now have:
* A `Dimension` class that behaves like a symbol, but can carry additional information about iteration spaces. We now provide the default space and time dimensions `x`, `y`, `z` and `t` in this format.
* A new `Expression` class that wraps an individual SymPy expression and extracts our known symbols (dimensions, data functions) and presents them to the code generator.
* Adjusted the `Iteration` class to use `Expression` objects when handling the loop body.
* A new `StencilKernel` class (name up for debate), that generates utilises `Expression` and `Iteration` objects to auto-generate the C code, while generically inferring loop dimensions and iteration spaces from the symbol dimensions.
* A new `IndexedData` utility class that annotates the `Indexed` expressions generated from our data functions with an additional pointer back to the original function (`SymbolicData` object). This allows us to access data shapes and indices even from expressions in the low-level API.
* Slightly re-written the caching mechanism for `SymbolicData` objects, putting most of the functionality into a separate `CachedSymbol` class. This cleans up the type-specific constructors, and now stores the actual created base indices (`Dimension` objects) on the cached symbolic data object.

This PR creates the opportunity for a lot more automation and a significant **clean-up or re-implementation of the Propagator**, as well as the missing composability for things like advanced Born operators. It requires a lot more work towards making dimensions and iteration spaces "smarter" to re-create the current Propagator features, like padding and buffering behaviour in time. But it also provides the opportunity for a more flexible code generation process, since, for example, `Dimension` objects might in the future determine whether to "bake in" loop dimensions or not. Due to the required structural changes in the outer symbolics **I would like to get this reviewed and merged ASAP** to further explore and improve this.


vincepandolfo:
`automodule` directives for the new files need to be added to `docs/devito.rst` for the documentation to be generated

devito/dimension.py
```diff
@@ -0,0 +1,22 @@
+from sympy import Symbol
+
+__all__ = ['Dimension', 'x', 'y', 'z', 't']
+
+
+class Dimension(Symbol):
```
FabioLuporini: Just to be picky: how about ``LoopIndex`` or ``IndexSymbol`` or ``LoopDimension`` ? 

Because I know understand what a ``Dimension`` is, but first time I read the name it felt weird for a "Dimension" to be a "Symbol"

devito/dimension.py
```diff
@@ -0,0 +1,22 @@
...
+    """Index object that represent iteration spaces"""
```
FabioLuporini: typo?

devito/interfaces.py
```diff
@@ -125,19 +128,25 @@ def __init__(self, *args, **kwargs):
...
+        :param shape: Optional, shape of the spatial data to
+                      automatically infer dimension symbols.
+        :return: Dimension indices used for each axis.
+        """
+        dimensions = kwargs.get('dimensions', None)
+        if dimensions is None:
+            # Infer dimensions from default and data shape
+            _indices = [x, y, z]
+            shape = kwargs.get('shape')
+            dimensions = _indices[:len(shape)]
```
FabioLuporini: So is ``shape`` always in ``kwargs`` ? because otherwise ``len(...)`` may fail

devito/interfaces.py
```diff
@@ -17,7 +18,34 @@
 _SymbolCache = {}
 
 
-class SymbolicData(Function):
+class CachedSymbol(object):
+    """Base class for symbolic objects that caches on the class type."""
+
+    @classmethod
```
FabioLuporini: Probably a silly question

I'm not sure I understand why are the methods below class methods ? It seems like that they are used on instances later on (e.g., in SymbolicData)

FabioLuporini: Forget it, I understand

devito/iteration.py
```diff
@@ -1,38 +1,50 @@
...
-from devito.codeprinter import ccode
+
+from devito.expression import Expression
+from devito.tools import filter_ordered
 
 __all__ = ['Iteration']
 
 
-class Iteration(object):
+class Iteration(Expression):
```
FabioLuporini: Why do you need to inherit from ``Expression``?

devito/iteration.py
```diff
@@ -1,38 +1,50 @@
...
     """Iteration object that encapsualtes a single loop over sympy expressions.
 
-    :param stencils: SymPy equation or list of equations that define the
-                     stencil used to create the loop body.
+    :param expressions: Single or list of SymPy expressions or :class:`Expression`
+                        objects that define the loop body.
     :param variable: Symbol that defines the name of the variable over which
```
FabioLuporini: I know this was there but probably we need a better name

devito/iteration.py
FabioLuporini: Also, do we want to call this ``IterationSpace`` ?

devito/dimension.py
mlange05: Yes, the nomenclature here is debatable, and I'm not insisting on `Dimension` at all. I would argue though that at the top-level, ie in `f(t, x, y)` the dimension does not yet define a loop, and that is the most common place users will define it, so `LoopDimension` or `LoopIndex` seem slightly too specific to me.

devito/stencilkernel.py
```diff
@@ -0,0 +1,131 @@
...
+from devito.interfaces import SymbolicData
+from devito.iteration import Iteration
+from devito.logger import error
+from devito.tools import filter_ordered
+
+__all__ = ['StencilKernel']
+
+
+class StencilKernel(object):
+    """Code generation class, alternative to Propagator"""
```
FabioLuporini: Based on this and the description of this PR, in the long term we expect the ``Propagator`` to be completely replaced by the ``StencilKernel`` right?

devito/iteration.py
mlange05: So that I can type check every item in `Propagator/StencilKernel.stencils`. They both should provide the same interface functions, like `ccode` and `susbtitute`.

mlange05: And yes, `IterationSpace` sounds ok too. Although I kind of see the `Dimension` as defining the iteration space, since it will eventually know about PML padding and buffering, while the `Iteration` represents a concrete loop in the code tree/list. But again, nomenclature is totally open for debate.

devito/dimension.py
FabioLuporini: Ah yes you're right 

I'm just happy with Dimension!

devito/iteration.py
mlange05: Agreed. Suggestions?

FabioLuporini:
I like all this

You can go quickly through my comments, and then for me it can land

devito/stencilkernel.py
mlange05: Ultimately yes, but whether we retro-fit this type of internals into the current Propagator or whether we re-implement all the missing current features is up for discussion. This is really just intended as a proof-of-concept that we can be generic across dimensions.

devito/iteration.py
FabioLuporini: Ah, I see! makes sense

devito/iteration.py
FabioLuporini: index ?

devito/stencilkernel.py
FabioLuporini: OK

PR #93, opened 2016-08-22
Openmp profiling

pp1336:
An addition to PR#89, allows profiling under OpenMP settings.

### DIFF
--- a/devito/profiler.py
+++ b/devito/profiler.py
@@ -1,22 +1,28 @@
-from ctypes import Structure, byref, c_double
+from ctypes import Structure, byref, c_double, c_longlong
 
 from cgen_wrapper import Block, Statement, Struct, Value
+from devito.logger import error
 
 
-class Profiler(Structure):
+class Profiler(object):
     """The Profiler class is used to manage profiling information for Devito
     generated C code.
     """
+    TIMINGS = 1
+    OIS = 2
 
     def __init__(self):
-        self.name = "timings"
-        self.fields = []
+        self.t_fields = []
+        self.o_fields = []
         self.timings = None
+        self.ois = None
+        self.gflops = None
+        self.oi_defaults = {}
 
-    def add_profiling(self, block, name, omp_flag=None):
+    def add_profiling(self, code, name, omp_flag=None):
         """Function to add profiling code to the given :class:`cgen.Block`.
 
-        :param block: A list of :class:`cgen.Generable` with the code to be
+        :param code: A list of :class:`cgen.Generable` with the code to be
                       profiled.
         :param name: The name of the field that will contain the timings.
         :param omp_flag: OpenMP flag to add before profiling operations,
@@ -24,7 +30,9 @@ class Profiler(Structure):
         :returns: A list of :class:`cgen.Generable` with the added profiling
                   code.
         """
-        self.fields.append((name, c_double))
+        self.t_fields.append((name, c_double))
+        self.o_fields.append((name, c_longlong))
+        self.oi_defaults[name] = self.get_oi(name, code)
 
         omp_flag = omp_flag or []
 
@@ -43,41 +51,77 @@ class Profiler(Structure):
             ])
         ]
 
-        return init + block + end
+        return init + code + end
 
-    @property
-    def get_timings_class(self):
-        """Returns a :class:`ctypes.Structure` subclass defining our structure
-
-        :returns: A :class:`Timings` class definition
+    def get_oi(self, code):
+        """Calculates the total operation intensity of the code provided.
+        If needed, injects the code with operations for runtime calculation.
         """
-        return type("Timings", (Structure,), {"_fields_": self.fields})
+        # TODO: Add the fixed OIs to self.ois_defaults. Inject code to calculate
+        #       OIs in loops
+        pass
 
     @property
-    def as_cgen_struct(self):
+    def get_class(self, choice):
+        """Returns a :class:`ctypes.Structure` subclass defining our structure
+        for Ois or Timings
+
+        :param choice: Profiler.OIS or Profiler.TIMINGS
+
+        :returns: A class definition
+        """
+        if choice == Profiler.OIS:
+            name = "Ois"
+            fields = self.t_fields
+        elif choice == Profiler.TIMINGS:
+            name = "Timings"
+            fields = self.o_fields
+        else:
+            error("Wrong choice")
+
+        return type(name, (Structure,), {"_fields_": fields})
+
+    @property
+    def as_cgen_struct(self, choice):
         """Returns the :class:`cgen.Struct` relative to the profiler
 
         :returns: The struct
         """
         fields = []
 
-        for name, _ in self.fields:
-            fields.append(Value("double", name))
+        if choice == Profiler.TIMINGS:
+            for name, _ in self.t_fields:
+                fields.append(Value("double", name))
+        elif choice == Profiler.OIS:
+            for name, _ in self.o_fields:
+                fields.append(Value("long long", name))
+        else:
+            error("Wrong choice")
 
         return Struct("profiler", fields)
 
     @property
-    def as_ctypes_pointer(self):
-        """Returns a pointer to the ctypes structure
+    def timings_as_ctypes_pointer(self):
+        """Returns a pointer to the ctypes structure for the timings
 
         :returns: The pointer
         """
-        self.timings = self.get_timings_class()
+        self.timings = self.get_class(Profiler.TIMINGS)
 
         return byref(self.timings)
 
     @property
-    def as_dictionary(self):
+    def ois_as_ctypes_pointer(self):
+        """Returns a pointer to the ctypes structure for the OIs
+
+        :returns: The pointer
+        """
+        self.ois = self.get_class(Profiler.OIS)
+
+        return byref(self.ois)
+
+    @property
+    def timings_as_dictionary(self):
         """Returns the timings as a python dictionary
 
         :returns: A dictionary containing the timings
@@ -87,3 +131,31 @@ class Profiler(Structure):
 
         return dict((field, getattr(self.timings, field))
                     for field, _ in self.timings._fields_)
+
+    @property
+    def ois_as_dictionary(self):
+        """Returns the operation intensities as a dictionary
+
+        :returns: A dictionary containing the OIs
+        """
+        if not self.ois:
+            return {}
+
+        return dict((field, getattr(self.ois, field)) for field, _ in self.ois._fields_)
+
+    @property
+    def get_gflops(self):
+        """Returns the GFLOPS of the profiled codes
+
+        :return: A dictionary containing the GFLOPS
+        """
+        if self.gflops:
+            return self.gflops
+
+        self.gflops = {}
+
+        for name, oi in self.ois_as_dictionary.items():
+            self.gflops[name] = ((self.oi_defaults[name] + oi) /
+                                 (self.timings_as_dictionary[name] * 10**9))
+
+        return self.gflops
--- a/devito/profiler.py
+++ b/devito/profiler.py
@@ -70,67 +70,76 @@ class Profiler(object):
         if name not in self.flops_defaults:
             self.flops_defaults[name] = 0
 
+        loads = {"stores": 0}
+
         for elem in code:
             if isinstance(elem, Assign):
-                assign_oi, assign_flops = self._get_assign_oi(elem, size)
-                self.oi[name] += assign_oi
+                assign_flops = self._get_assign_flops(elem, loads)
+                self.oi[name] += assign_flops
                 self.flops_defaults[name] += assign_flops
             elif isinstance(elem, For):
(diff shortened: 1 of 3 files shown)


devito/propagator.py
```diff
@@ -308,7 +308,11 @@ def generate_loops(self, loop_body):
         omp_master = [cgen.Pragma("omp master")] if self.compiler.openmp else []
         omp_single = [cgen.Pragma("omp single")] if self.compiler.openmp else []
         omp_parallel = [cgen.Pragma("omp parallel")] if self.compiler.openmp else []
-        omp_for = [cgen.Pragma("omp for schedule(static)")] if self.compiler.openmp else []
+        reduction_list = ["kernel", "loop_body"] if self.compiler.openmp else []
```
vincepandolfo: This should be managed by the Profiler instead of being hard coded.

devito/propagator.py
```diff
@@ -308,7 +308,11 @@ def generate_loops(self, loop_body):
...
+        omp_for = [cgen.Pragma("omp for schedule(static)"\
+                   + (self.profiler.get_loop_reduction("+", reduction_list)\
+                   if self.profile else ""))]\
+                  if self.compiler.openmp else []
```
vincepandolfo: This should be managed by the Profiler as well, so that we can do a different reduction on timing since that doesn't work with OpenMP and `+` is not the right reduction here I believe.

devito/propagator.py
pp1336: Well as we discussed just now, the other parts of the profiling code are also hard coding the these stuff, so I think this is a more general issue that deserves another PR.

devito/propagator.py
pp1336: the + is only for flops, I did know that timing was not ok with openmp guess a little miscommunication here, I will add the change to handle loop timing

devito/profiler.py
```diff
@@ -1,30 +1,86 @@
...
+        OpenMP settings.
+        :param op: The reduction operator.
+        :param varaiables: A list of string, each is the name of a variable
+                           to be reduced.
+        :returns: String representing the reduction pragma
+        """
+        if len(variables) <= 0:
+            return ""
+        return " reduction(" + op + ":" + self.loop_temp_prefix + variables[0]\
+               + "".join(["," + self.loop_temp_prefix + v for v in variables[1:]]) + ")"
```
mlange05: Using `+` to concatenate strings is considered bad style by many. String formatting (`%s`) is much cleaner and more readable.

devito/profiler.py
```diff
@@ -1,30 +1,86 @@
...
+        :param value: String represeting the initial value of temporary
+                      varaibles.
+        :param variables: A list of string, each is the name of a variable to
+                          be declaried
+        :returns: String representing the declariation statement
+        """
+        if len(variables) <= 0:
+            return ""
+        return "long long " + self.loop_temp_prefix + variables[0] + " = " + value +\
+               "".join([", " + self.loop_temp_prefix + v + " = " + value for v in variables[1:]])
```
mlange05: String formatting

devito/profiler.py
```diff
@@ -1,30 +1,86 @@
...
+
+    def get_loop_flop_update(self, variables):
+        """Function to generate a list of statement representing the amount
+        of flops update per iteration of the loop.
+        :param variables: A list of string, each is the suffix of the
+                          temperary variables used for update
+        :returns: A list of string representing statements used as updating
+                  statements
+        """
+        return [self.f_name + "->" + v + " += " + self.loop_temp_prefix + v for v in variables]
```
mlange05: String formatting

mlange05:
Ok, the code looks great, except for a few minor string formatting comments. One thing I found when running this though is that `gflops` seem to double with two threads, which I don't think is correct. OI and timings look sensible though.

devito/profiler.py
```diff
@@ -1,30 +1,86 @@
...
-    def __init__(self):
-        self.name = "timings"
-        self.fields = []
-        self.timings = None
+    def __init__(self, openmp=False):
+        self.openmp = openmp
+        self.t_name = "timings"
+        self.o_name = "oi"
+        self.f_name = "flops"
+        self.loop_temp_prefix = "temp_"
```
FabioLuporini: Feels like ``t_name, o_name, ..., loop_temp_prefix`` should actually be class (read-only) attributes ?
Also, the names you chose are not very explanatory

devito/profiler.py
```diff
@@ -1,30 +1,86 @@
...
+        """
+        if len(variables) <= 0:
+            return ""
+        return "long long " + self.loop_temp_prefix + variables[0] + " = " + value +\
+               "".join([", " + self.loop_temp_prefix + v + " = " + value for v in variables[1:]])
+
+    def get_loop_flop_update(self, variables):
+        """Function to generate a list of statement representing the amount
+        of flops update per iteration of the loop.
+        :param variables: A list of string, each is the suffix of the
```
FabioLuporini: Leave a blank line here before ``:param variables: ...``

devito/profiler.py
```diff
@@ -1,30 +1,86 @@
...
+        """
+        if len(variables) <= 0:
+            return ""
+        return " reduction(" + op + ":" + self.loop_temp_prefix + variables[0]\
+               + "".join(["," + self.loop_temp_prefix + v for v in variables[1:]]) + ")"
+
+    def get_loop_temp_var_decl(self, value, variables):
+        """Function to generate the declariation and initialisation of the
+        temporary variables used for reduction for loop profiling.
+        :param value: String represeting the initial value of temporary
```
FabioLuporini: Leave a blank line here before ``:param value: ...``

devito/profiler.py
```diff
@@ -1,30 +1,86 @@
...
+        self.oi = {}
+        self._timings = None
+        self._flops = None
+        self.flops_defaults = {}
 
-    def add_profiling(self, block, name, omp_flag=None):
+    def get_loop_reduction(self, op, variables):
+        """Function to generate reduction pragma, used for profiling under
+        OpenMP settings.
+        :param op: The reduction operator.
```
FabioLuporini: Leave a blank line here before ``:param op: ...``

devito/profiler.py
```diff
@@ -39,50 +95,208 @@ def add_profiling(self, block, name, omp_flag=None):
...
-    def get_timings_class(self):
+    def get_oi_and_flops(self, name, code, size):
+        """Calculates the total operation intensity of the code provided.
+        If needed, lets the C code calculate it.
+
+        :param name: The name of the field to be populated
+        :param code: The code to be profiled.
+        """
+        if name not in self.oi:
+            self.oi[name] = 0
```
FabioLuporini: Consider using a ``defaultdict(int)`` instead of a ``{}`` ?

``from collections import defaultdict``

devito/profiler.py
```diff
@@ -39,50 +95,208 @@ def add_profiling(self, block, name, omp_flag=None):
...
+        if name not in self.flops_defaults:
+            self.flops_defaults[name] = 0
```
FabioLuporini: same concept as above

devito/profiler.py
```diff
@@ -39,50 +95,208 @@ def add_profiling(self, block, name, omp_flag=None):
...
+                assign_flops = self._get_assign_flops(elem, loads)
+                self.oi[name] += assign_flops
+                self.flops_defaults[name] += assign_flops
+            elif isinstance(elem, For):
+                for_flops = self._get_for_flops(name, elem, loads)
+                self.oi[name] += for_flops
+            elif isinstance(elem, Block):
+                block_oi, block_flops = self._get_block_oi_and_flops(name, elem, loads)
+                self.oi[name] += block_oi
+                self.flops_defaults[name] += block_flops
```
FabioLuporini: What about AugmentedAssignments, such as Incr, Decr (i.e., +=, -=) ... ?

Maybe we should also have a default ``else`` branch which prints a warning if hit ?

devito/profiler.py
```diff
@@ -39,50 +95,208 @@ def add_profiling(self, block, name, omp_flag=None):
...
+                    cur_load = ""
+                    brackets += 1
+                    idx += 1
+                elif char is ' ' and brackets == 0 and len(cur_load) > 0:
+                    loads[cur_load] = True
+                    cur_load = ""
+                    idx += 1
+                else:
+                    cur_load += char
+                    idx += 1
```
FabioLuporini: Ah, maybe I see now. This looks very hacky. I think we should make a better use of the cgen hierarchy (by suitably handling Assign, Incr, Decr etc)

devito/profiler.py
```diff
@@ -39,50 +95,208 @@ def add_profiling(self, block, name, omp_flag=None):
...
         """Returns the timings as a python dictionary
 
         :returns: A dictionary containing the timings
         """
-        if not self.timings:
+        if not self._timings:
+            return {}
+
+        return dict((field, getattr(self._timings, field))
+                    for field, _ in self._timings._fields_)
```
FabioLuporini: I'm not sure I understand this, maybe add an example as a comment (perhaps as method doc ?)

devito/profiler.py
```diff
@@ -39,50 +95,208 @@ def add_profiling(self, block, name, omp_flag=None):
...
+        """
+        if not self._flops:
             return {}
 
-        return dict((field, getattr(self.timings, field)) for field, _ in self.timings._fields_)
+        return dict((field,
+                     float(
+                         getattr(self._flops, field) + self.flops_defaults[field]
+                     )/self.timings[field]/10**9)
+                    for field, _ in self._flops._fields_)
```
FabioLuporini: Same as before

devito/profiler.py
```diff
@@ -39,50 +95,208 @@ def add_profiling(self, block, name, omp_flag=None):
...
+
+    @property
+    def flops(self):
+        if self._flops:
+            return dict((field, int(getattr(self._flops, field))) for field, _ in self._flops._fields_)
+        else:
+            return {}
```
FabioLuporini: Looks more complicated than it really needs to be ?

pp1336:
I will handle the formatting issue, @vincepandolfo do you want to handle the others since you probably know those code better

devito/profiler.py
pp1336: I was just following the previous style, i dont mind changing it.

devito/profiler.py
vincepandolfo: The only `cgen` objects we use now are `For`, `Block` and `Assign`. That's why it is structured this way.

devito/profiler.py
vincepandolfo: What would better names be? I agree on having the as read-only attributes.

devito/profiler.py
vincepandolfo: Why? I'm not familiar with `defaultdict`

devito/profiler.py
vincepandolfo: What's that you don't get?

FabioLuporini: ah, the content of the resulting dict

vincepandolfo: `self._timings` is a ctypes struct. `self._timings._fields_` contains a list of tuples `(name, type)`. I get the name from there and the value with `getattr`.

It's simply a conversion from a ctypes struct to a `dict`

devito/profiler.py
FabioLuporini: If you use a ``defaultdict``, an entry which is not in the dict gets assigned a default value (0 in the case of int). Look at this:

```
In [1]: from collections import defaultdict

In [2]: d = defaultdict(int)

In [3]: d[1]
Out[3]: 0

In [4]: d[2] += 1

In [5]: d
Out[5]: defaultdict(<type 'int'>, {1: 0, 2: 1})
```

Similar behaviour with a normal dict and using the ``setdefault`` method ({}.setdefault.__doc__ is quite self-explanatory)

FabioLuporini:
@pp1336 we need to merge this ASAP. Please join slack.

devito/profiler.py
FabioLuporini: Maybe ``_C_timings``, ``_C_oi``, ``_C_flops``, ``_C_temporary``. Plus add a comment that says that attributes starting with ``_C_`` are simply used for code generation ?

devito/profiler.py
FabioLuporini: The method is still hacky, though

The right way of doing this, I think, is to have a visitor that traverses cgen expressions, not analysing strings textually

vincepandolfo: The string is the only thing I get from cgen. There's no expression.

pp1336:
vincepandolfo and I have addressed most comments. We agree that we are good with the profiler as it is now, and that the profiler might require further improvements in the future, when we have time.

pp1336:
OK so we are just waiting 81 to be merge, then we will rebase this branch and merge.

pp1336:
Looks like all tests under Linux passed, failed due to flake8. All tests for mac failed.

PR #95, opened 2016-08-22
Enforcing 90 line length rule

sheino:
Enforcing 90 line length rule as @FabioLuporini suggested and updating existing files accordingly.

Close issue 83 and 64 after merging.


mlange05:
I'm not the greatest fan of this for minor aesthetic reasons, but it's probably best in the long run. So merging...

PR #97, opened 2016-08-23
CSE (v2)

vincepandolfo:
The problem with this one seems to be a loss of precision in the results. The tests that fail do so for that reason. I'm opening a PR anyway because I can't really pinpoint the problem by myself.

It needs to be integrated with `Iteration` objects as well but it's probably better to wait for #92 for that.

### DIFF
--- a/devito/operator.py
+++ b/devito/operator.py
@@ -64,9 +64,11 @@ def expr_indexify(expr):
 
 
 def expr_cse(expr):
-    """Performs common subexpression elimination on expressio
+    """Performs common subexpression elimination on expressions
 
-    :param expr: Sympy expression on which CSE needs to be performed
+    :param expr: Sympy equation or list of equations on which CSE needs to be performed
+
+    :return: A list of the resulting equations after performing CSE
     """
     expr = expr if isinstance(expr, list) else [expr]
 
--- a/devito/operator.py
+++ b/devito/operator.py
@@ -1,7 +1,8 @@
 import numpy as np
-from sympy import (Eq, Function, Indexed, Symbol, lambdify, preorder_traversal,
+from sympy import (cse, Eq, Function, Indexed, IndexedBase, Symbol, lambdify, preorder_traversal,
                    solve, symbols)
 from sympy.abc import t
+from sympy.utilities.iterables import numbered_symbols
 
 from devito.compiler import get_compiler_from_env
 from devito.interfaces import SymbolicData, TimeData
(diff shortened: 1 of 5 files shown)


devito/propagator.py
```diff
@@ -267,6 +267,22 @@ def sympy_to_cgen(self, stencils):
...
+        dtypes = []
+
+        for e in preorder_traversal(expr):
+            if hasattr(e, 'dtype'):
+                dtypes.append(e.dtype)
+
+        if dtypes:
+            return np.find_common_type(dtypes, [])
+        else:
+            return np.float64
```
mlange05: Does the Propagator not have a `dtype` argument? Should we not be defaulting to this one?

devito/propagator.py
```diff
@@ -282,12 +298,16 @@ def convert_equality_to_cgen(self, equality):
             s_lhs = ccode(self.time_substitutions(equality.lhs).xreplace(self._var_map))
             s_rhs = self.time_substitutions(equality.rhs).xreplace(self._var_map)
 
+            if s_lhs.find("temp") != -1:
+                s_lhs = "double" + " " + s_lhs
```
mlange05: Is this hard-coding the type of temps? Are we not deriving their type above?

devito/propagator.py
vincepandolfo: Yes, we should. I overlooked that argument.

devito/propagator.py
vincepandolfo: I couldn't find a way to convert the dtype to the proper string easily so I left it this way while working on the actual CSE.

devito/operator.py
```diff
@@ -62,6 +63,64 @@ def expr_indexify(expr):
     return expr.xreplace(replacements)
 
 
+def expr_cse(expr):
+    """Performs common subexpression elimination on expressio
```
FabioLuporini: typo

FabioLuporini:
Great stuff!

I think we should rename ``factorized`` as ``temporaries`` or something else, but definitely not ``factorized``

I think it's time to move all functions/methods that visit and manipulate sympy expressions into a proper class (proper classes) in a separate module (``expressions.py`` ?) Not sure if this should come with this PR, but it should happen soon!

mlange05:
@FabioLuporini Just FYI, PR #92 moves all expression traversal into a separate `Expression` class to aggregate them. In it's current state this is only done on the low-level "indexed" expressions, but it could be extended to cover all expression types.

FabioLuporini:
@mlange05 I see, good

mlange05:
One more thing; is CSE always performed or should we have a flag to swtich it on.off (default on, of course)?

FabioLuporini:
I think we should have a flag at least for performance comparison (code gen time, and also performance of the generated code with different backend compilers)

vincepandolfo:
Yes, that sounds like a sensible idea. Should we integrate this with `Iteration` already or is it better to wait for #92 to land?

mlange05:
Let's leave the integration with `Iteration` for another PR and land this one first. I think we want the flag though.

FabioLuporini:
Assuming all tests pass -- and after rebasing -- is it correct to say that this PR can land ?

vincepandolfo:
Right now all the tests are running with `cse` off since the default value for the flag is `False`. Should I add more tests that use `cse` or change the default value?

mloubout:
I think a test checking with/without cse produce the same result would be nice.

FabioLuporini:
@vincepandolfo remind me why the flag is off by default? shouldn't it be on?

I agree with @mloubout : we should have a few tests, that we know stress cse quite hard, to check the numerical correctness

vincepandolfo:
The flag is on by default now

tests/test_cse.py
```diff
@@ -0,0 +1,55 @@
...
+
+    receiver_coords = np.zeros((101, 3))
+    receiver_coords[:, 0] = np.linspace(50, 950, num=101)
+    receiver_coords[:, 1] = 500
+    receiver_coords[:, 2] = location[2]
+    data.set_receiver_pos(receiver_coords)
+    data.set_shape(nt, 101)
+    Acoustic = Acoustic_cg(model, data)
+    (rec, u) = Acoustic.Forward(save=True, cse=cse)
+
```
mloubout: `save=True` is not necessary unless you want to compare the actual wavefield with
`assert np.isclose(np.linalg.norm(u- u_cse), 0.0)`
If not set to `False` as Travis really don't like large arrays

vincepandolfo: Good to know. Will change now

PR #100, opened 2016-08-26
Check for timings > 0 when calculating GFLOPS

vincepandolfo:
`Profiler.gflops` can sometimes fail if the time is recorded as 0 for some entries. It now returns `NaN` instead.

### DIFF
--- a/devito/profiler.py
+++ b/devito/profiler.py
@@ -312,7 +312,8 @@ class Profiler(object):
             return {}
 
         return dict((field,
-                     float(
+                     (float(
                          getattr(self._C_flops, field) + self.flops_defaults[field]
                      )/self.timings[field]/10**9)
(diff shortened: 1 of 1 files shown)


devito/profiler.py
```diff
@@ -312,7 +312,8 @@ def gflops(self):
             return {}
 
         return dict((field,
-                     float(
+                     (float(
                          getattr(self._C_flops, field) + self.flops_defaults[field]
                      )/self.timings[field]/10**9)
+                     if self.timings[field] > 0.0 else float("nan"))
```
mlange05: Why not `np.nan`? Otherwise good.

vincepandolfo: Didn't want to add another import when this works just as good

mlange05: Ok, cool.

PR #102, opened 2016-08-28
Reduction clause onto the simd line.

pvelesko:
reduction clause needs to go onto each #pragma simd aligned  line otherwise vectorization is broken

### DIFF
--- a/devito/propagator.py
+++ b/devito/propagator.py
@@ -101,6 +101,13 @@ class Propagator(object):
         # Profiler needs to know whether openmp is set
         self.profiler = Profiler(self.compiler.openmp)
 
+        # suffix of temp variable used for loop reduction of corrosponding struct
+        # member in Profiler struct when openmp is on
+        self.reduction_list = ["kernel", "loop_body"] if self.compiler.openmp else []
+        self.reduction_clause = (self.profiler
+                                 .get_loop_reduction("+", self.reduction_list)
+                                 if self.profile else "")
+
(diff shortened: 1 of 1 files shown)


FabioLuporini:
I think I'm missing something... why just adding a TODO?

pvelesko:
Mathias asked to make a PR. I don't have time to fix this

pvelesko:
@pp1336  can you change this back to float64 so it would pass travis and merge?

pp1336:
sure

FabioLuporini:
what's stopping this from merge?

vincepandolfo:
I've cleaned up the git history. Should be good to go when travis is green

(removed the TODO commit, and the two precision ones)

examples/acoustic_example.py
```diff
@@ -71,7 +71,7 @@ def source(t, f0):
 receiver_coords[:, 2] = location[2]
 data.set_receiver_pos(receiver_coords)
 data.set_shape(nt, 101)
-Acoustic = Acoustic_cg(model, data)
+Acoustic = Acoustic_cg(model, data, auto_tune=True)
```
mlange05: Why are we now defaulting to `auto_tune=True`? Isn't that still computationally expensive? Also, how would a user over-write that?

vincepandolfo: Can this be removed from this PR and discussed somewhere else? Shouldn't be related anyway

Issue #94, opened 2016-08-22
Easy-to-replicate performance benchmarks


mlange05:
For publication and performance regression purposes we should have one or two pre-defined benchmark tests that can be re-run with ease to test and verify new features and optimisations. These benchmarks should be fully automated and rely on a pre-defined example setup. Following PR #80 and with a set of benchmarking utilities provided in the [opesci/opescibench](https://github.com/opesci/opescibench) repo we should now be able to provide this.

The proposed development spec:
- Identify appropriate example tests that execute in a reasonable time frame without being too small or trivial, and re-structure them to ensure they can be fully parametrized.
- Add either `<example>_bench.py` or add `--benchmark` option to standalone mode of the test itself.
- Create a benchmark class `<Example>Executor` that inherits from `opescibench.Executor` and defines how to execute the test runs and collects performance data (runtime, GFlops), as well as meta-data (OI).
- Define a `<Example>Benchmark class that inherits from`opescibench.Benchmark`, which allows it to invoke repeated performance runs and store the collected data and meta-data in JSON files. The specific performance relevant parameters should be configurable, for example:

```
python <example>_bench.py --benchmark --compiler clang --cse --cache-blocking 8 16 ...
```
- Use the `Benchmark` class and custom `matplotlib` code to pre-define various types of plots, including roofline, bar comparisons and parallel scaling graphs. This should be invoked with arguments `--plot --plottype` or similar and the relevant parameters to compare and plot, for example:

```
python <example>_bench.py --plot --plottype roofline --cse --compiler clang gcc intel
```
- The final command line options and some default settings to benchmark execution and plotting should be documented in a separate README file in the `examples` directory.

Please note that `opescibench` is intended to provide most of the core functionalities detailed here, ranging from repeated execution and data collection to various utilities for pretty plotting. It is, however, very much a work in progress and more work is probably required on providing more (any) documentation and enhancing the current features.

mlange05:
Following the recent addition of a reproducible benchmark infrastructure, I believe all that is required to close this issue now is a short example of how to run a baseline benchmark of, say the acoustic equation, and create a roofline plot. @FabioLuporini, could you please provide a short README file with example commands of how to run and plot the baseline test you have been executing recently?

mlange05:
Ok, except for the need for more in-depth documentation this has been addressed a long time ago, so closing for now. Documentation will be described in a separate issue.

PR #105, opened 2016-08-30
Benchmark diffusion

mlange05:
This merge re-writes parts of the diffusion examples and adds a very basic benchmark/plotting infrastructure via the [opescibench](https://github.com/opesci/opescibench) utility package. This is a first step towards issue #94, and demonstrates the basic usage of the benchmarking utility. Importantly, we can now also run, bench, and plot various parametrisations of the test via the command line for quick and easy reproducibility. In detail:
* Normal run mode allows visualisation for verification purposes; run with `python example_diffusion.py run -m numpy --show`. This is independent of `opescibench` and can be executed anywhere.
* Benchmark runs that record and store performance data using opescibench. The parametrisation is still very basic, but we can add more performance related parameters later.
* Plot mode creates a simple bar chart comparison between the different modes specified.
* Associated README contains example commands to recreate benchmarks and plotting runs.

Please note that more Devito-related performance parameters, threading, CSE, loop clocking, etc. are still missing, and should ultimately be configurable from the command line as well.


examples/example_diffusion.py
```diff
@@ -0,0 +1,243 @@
...
+def test_diffusion2d(spacing=0.01, timesteps=1000):
+    ui = ring_initial(spacing=spacing)
+    u, _ = execute_devito(ui, spacing=spacing, timesteps=timesteps)
+    assert(u.max() < 2.4)
+    assert(np.linalg.norm(u, ord=2) < 13)
+
+
+if __name__ == "__main__":
+    description = """
+Example script demosntrating several approaches to solving the
```
FabioLuporini: typo

examples/example_diffusion.py
```diff
@@ -0,0 +1,243 @@
...
+2D diffusion equation. The various modes are pure Python,
+vectorized numpy, a lambdified SymPy equation and the Devito API.
+
+Please not that the default settings for spacing, and timesteps are
```
FabioLuporini: typo

examples/example_diffusion.py
```diff
@@ -0,0 +1,243 @@
...
+    parser.add_argument('--show', action='store_true',
+                        help="Show animation of the solution field")
+    parser.add_argument('-i', '--resultsdir', default='results',
+                        help='Directory containing results')
+    parser.add_argument('-o', '--plotdir', default='plots',
+                        help='Directory to store generated plots')
+
+    args = parser.parse_args()
+    if len(args.spacing) > 2:
+        raise ValueError("Too many arguments encountered for --spacing")
```
FabioLuporini: I *think* there's an option to have the ``ArgumentParser`` automatically checking the value of a parameter.

FabioLuporini:
Apart from the very very minor things I pointed out, this can land for me.

I also have another comment, concerning this:
``
Please note that more Devito-related performance parameters, threading, CSE, loop clocking, etc. are still missing, and should ultimately be configurable from the command line as well.
``

Do you think that Devito should have its own default ``ArgumentParser``? because presumably we wanna play with these options in all possible Devito codes. If that's the case, do you know if ArgumentParsers can be nested ?

vincepandolfo:
`ArgumentParser`s can be nested using the `parents` parameter

https://docs.python.org/2/library/argparse.html#parents

FabioLuporini:
Rebase + merge ?

mlange05:
Ok, rebased and updated. I ended up not integrating it with `benchmark.py` because the comparison runs (python, numpy and sympy) are not really compatible with all the options the script provides. The diffusion benchmark does come with it's own documentation in the form of a README file though, and if someone could spare the time I'd appreciate a second look/review of this before merging (it's been a while since the first one).

FabioLuporini:
Even though diffusion doesn't use the benchmarking infrastructure, can we put the relevant files in an appropriate sub-directory, like we did for acoustic and tti?

I also see we're losing a test although you set up travis to run it as an example. What about having a "mock" test that only thing it does is invoking the example?  Or if you're worried that this test is too big, what about splitting it into smaller tests?

mlange05:
Good point about the sub-directory, I moved it. But I don't think we're losing a test, since I'm invoking the diffusion via py-test explicitly. I used to invoke all examples through py-tests, but we're running acoustic/TTI through `benchmark.py test` now, so there is no need.

FabioLuporini:
OK, I see.

Then what we probably need is a Makefile (or something along these lines), which we can use to run a ``make test``. Otherwise one, to test locally, needs to remember and invoke multiple commands -- a simple ``make test`` which picks the tests in ``tests`` as well as a number of examples would be quite handy.

Not a blocker to this PR, though, but perhaps something we need? Or am I stupidly forgetting about something?

mlange05:
Ok, rebased and finally green, so merging...

Re: Single testing command - I agree, but I don't see this as a priority. As I said, the ability got lost with the introduction of the `benchmark.py` script. We might just want to add a few small `test_***` definitions to the acoustic and TTI examples to get `py.test -vs examples` to work again.

PR #106, opened 2016-08-30
Tti fix

mloubout:
Small typos from 90 lines changes and model initialization names fix.

PR #103, opened 2016-08-30
CSE bug fixes

vincepandolfo:
This PR solves:
 - an issue with the collection of terms that change during the execution of the loop

For example, this could happen:
```
double temp = u[t2][i1][i2]
u[t2][i1][i2] = ….
U[t2][i1][i2] = temp + ….
```

After this fix, this is what we'd get:
```
double temp;
temp = u[t2][i1][i2]
u[t2][i1][i2] = ….
temp = u[t2][i1][i2]
U[t2][i1][i2] = temp + ….
```

- an issue with temporaries depending on other temporaries causing compiler errors (this slows down the code quite a bit)

Thanks to @mloubout for pointing out the issues

EDIT: Also changed tti_operators.ForwardOperator not to use `factorized`.

EDIT2: Fixed another bug that caused some temporaries to be reverted when not needed, making the second issue pointed out in the original post disappear in the process

PR #107, opened 2016-08-31
Benchmark for TTI example

vincepandolfo:
This is on top of #102 #103 and #106 

Allows users to benchmark and plot `tti_example.py` using `opescibench`.

```
usage: tti_benchmark.py [-h] [-o] [-d DIMENSIONS DIMENSIONS DIMENSIONS]
                      [-s SPACING SPACING] [-t TN] [-c] [-a]
                      [-cb CACHE_BLOCKING CACHE_BLOCKING] [-r RESULTSDIR]
                      [-p PLOTDIR] [--max_bw MAX_BW] [--max_flops MAX_FLOPS]
                      [{run,bench,plot}] [{gnu,intel,clang,mic,knl}]

Example script for TTI.

positional arguments:
  {run,bench,plot}      Script mode. Either 'run', 'bench' or 'plot'
  {gnu,intel,clang,mic,knl}

optional arguments:
  -h, --help            show this help message and exit
  -o, --omp             Enable OpenMP
  -d DIMENSIONS DIMENSIONS DIMENSIONS, --dimensions DIMENSIONS DIMENSIONS DIMENSIONS
                        Dimension of the grid
  -s SPACING SPACING, --spacing SPACING SPACING
                        Spacing on the grid
  -t TN, --tn TN        Number of timesteps
  -c, --cse             Benchmark with CSE on and off. Enables CSE when
                        execmode is run
  -a, --auto_tuning     Benchmark with auto tuning on and off. Enables auto
                        tuning when execmode is run
  -cb CACHE_BLOCKING CACHE_BLOCKING, --cache_blocking CACHE_BLOCKING CACHE_BLOCKING
                        Uses provided block sizes when AT is off
  -r RESULTSDIR, --resultsdir RESULTSDIR
                        Directory containing results
  -p PLOTDIR, --plotdir PLOTDIR
                        Directory containing plots
  --max_bw MAX_BW       Maximum bandwidth of the system
  --max_flops MAX_FLOPS
                        Maximum FLOPS of the system

```

Next step after this would be moving all this code to a different script and have the user select `tti` or `acoustic`, since the two examples take exactly the same parameters.

### DIFF
--- a/examples/TTI_codegen.py
+++ b/examples/TTI_codegen.py
@@ -3,6 +3,7 @@ from __future__ import print_function
 
 import numpy as np
 
+from devito.at_controller import AutoTuner
 from examples.tti_operators import *
 
 
@@ -62,11 +63,20 @@ class TTI_cg:
                               dtype=self.dtype, nbpml=nbpml)
         self.src.data[:] = data.get_source()[:, np.newaxis]
 
-    def Forward(self, save=False, cse=True, cache_blocking=None):
+    def Forward(self, save=False, cse=True, auto_tuning=False):
         fw = ForwardOperator(self.model, self.src, self.damp, self.data,
                              time_order=self.t_order, spc_order=self.s_order,
-                             profile=True, save=save, cse=cse,
-                             cache_blocking=cache_blocking)
+                             profile=True, save=save, cse=cse)
+
+        if auto_tuning:
+            fw_new = ForwardOperator(self.model, self.src, self.damp, self.data,
+                                     time_order=self.t_order, spc_order=self.s_order,
+                                     save=save, cse=cse)
+
+            at = AutoTuner(fw_new)
+            at.auto_tune_blocks(self.s_order + 1, self.s_order * 4 + 2)
+            fw.propagator.cache_blocking = at.block_size
+
         u, v, rec = fw.apply()
         return (rec.data, u.data, v.data,
                 fw.propagator.gflops, fw.propagator.oi)
--- a/examples/TTI_codegen.py
+++ b/examples/TTI_codegen.py
@@ -76,7 +76,7 @@ class TTI_cg:
                                      profile=True, save=save, cse=cse, compiler=compiler)
 
             at = AutoTuner(fw_new)
-            at.auto_tune_blocks(self.s_order + 1, self.s_order * 4 + 2)
+            at.auto_tune_blocks(3, 30)
             fw.propagator.cache_blocking = at.block_size
 
         u, v, rec = fw.apply()
--- a/examples/TTI_codegen.py
+++ b/examples/TTI_codegen.py
@@ -63,15 +63,15 @@ class TTI_cg:
                               dtype=self.dtype, nbpml=nbpml)
         self.src.data[:] = data.get_source()[:, np.newaxis]
 
-    def Forward(self, save=False, cse=True, auto_tuning=False):
+    def Forward(self, save=False, cse=True, auto_tuning=False, compiler=None):
         fw = ForwardOperator(self.model, self.src, self.damp, self.data,
                              time_order=self.t_order, spc_order=self.s_order,
-                             profile=True, save=save, cse=cse)
+                             profile=True, save=save, cse=cse, compiler=compiler)
 
         if auto_tuning:
             fw_new = ForwardOperator(self.model, self.src, self.damp, self.data,
                                      time_order=self.t_order, spc_order=self.s_order,
-                                     save=save, cse=cse)
+                                     profile=True, save=save, cse=cse, compiler=compiler)
 
             at = AutoTuner(fw_new)
             at.auto_tune_blocks(self.s_order + 1, self.s_order * 4 + 2)
--- a/examples/TTI_codegen.py
+++ b/examples/TTI_codegen.py
@@ -63,10 +63,12 @@ class TTI_cg:
                               dtype=self.dtype, nbpml=nbpml)
         self.src.data[:] = data.get_source()[:, np.newaxis]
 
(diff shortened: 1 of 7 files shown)


examples/tti_example.py
```diff
@@ -1,59 +1,173 @@
...
+            label = "TTI, CSE: %s, AT: %s" % (key["cse"], key["auto_tuning"])
+            mflops_dict[label] = gflops * 1000
+            oi_dict[label] = oi_value
+
+        name = "TTI - Dimensions: %s, Spacing: %s.pdf" % \
+            (parameters["dimensions"], parameters["spacing"])
+
+        plotter = Plotter()
+        plotter.plot_roofline(
+            name, mflops_dict, oi_dict, args.max_bw, args.max_flops)
```
vincepandolfo: It's exactly the same script, I've just added the argument parsing and benchmarking. When run with `python examples/tti_example.py` it behaves in exactly the same way. In what way is this too heavy?

mloubout: I mean non-opesci people need to be able to read it easily if they want to play with it. I know that it runs exactly the same, but the main is unreadabble for a typical user (they can barely read it without everything you added).

vincepandolfo: They can safely ignore the main. I want to move all that stuff out anyway when doing the same thing for `acoustic`, as I said in the first message.

mloubout: I understand. I was just pointing out that if we want to advertize what we do well, it is better to split this examples into "one script , one purpose" so that people can easily find what they need without having to filter out half an example themselves

pvelesko:
Might want to change MIC to KNC since both KNL and KNC are of MIC(Many integrated cores) architecture

FabioLuporini:
The only minor thing I can see is about the help message which looks weird. For example, take 
``-d DIMENSIONS DIMENSIONS DIMENSIONS``
or 
``-cb CACHE_BLOCKING CACHE_BLOCKING``

there should be a way of printing something less aggressive than "DIMENSIONS DIMENSIONS DIMENSIONS", such as "dim1 dim2 dim3"

examples/tti_benchmark.py
```diff
@@ -0,0 +1,117 @@
...
+
+    if args.execmode == "run":
+        run(**parameters)
+
+    if args.execmode == "bench":
+        if Benchmark is None:
+            raise ImportError("Could not find opescibench utility package.\n"
+                              "Please install from https://github.com/opesci/opescibench")
+
+        if parameters["auto_tuning"]:
```
FabioLuporini: why do u need this ``if``?

examples/tti_benchmark.py
```diff
@@ -0,0 +1,117 @@
...
+            parameters["auto_tuning"] = [True, False]
+
+        if parameters["cse"]:
```
FabioLuporini: why do u need this ``if``

examples/tti_benchmark.py
```diff
@@ -0,0 +1,117 @@
...
+        oi = bench.lookup(params=parameters, measure="oi")
+
+        for key, gflops in gflops.items():
+            oi_value = oi[key]
+            key = dict(key)
+            label = "TTI, CSE: %s, AT: %s" % (key["cse"], key["auto_tuning"])
+            mflops_dict[label] = gflops * 1000
+            oi_dict[label] = oi_value
+
+        name = "TTI - Dimensions: %s, Spacing: %s.pdf" % \
```
FabioLuporini: please avoid with spaces in file names

examples/tti_benchmark.py
vincepandolfo: I need to define the parameter space for auto tuning

examples/tti_benchmark.py
vincepandolfo: Same as above

FabioLuporini:
Now numerical correctness is checked. Good

examples/TTI_codegen.py
```diff
@@ -62,12 +63,25 @@ def damp_boundary(damp):
...
+                                     profile=True, save=save, cse=cse, compiler=compiler)
+
+            at = AutoTuner(fw_new)
+            at.auto_tune_blocks(self.s_order + 1, self.s_order * 4 + 2)
+            fw.propagator.cache_blocking = at.block_size
+
         u, v, rec = fw.apply()
-        return (rec.data, u.data, v.data)
+        return (rec.data, u.data, v.data,
+                fw.propagator.gflops, fw.propagator.oi)
```
mloubout: More of a personal view, but I don't think`gflops` and `oi` should be default output.  The forward operator is there for modelling purpose, and extra output just make its use messy. Just a thought doesn't need to be changed now, but I will for Users at some point.

vincepandolfo: If this code is part of the examples, shouldn't it just serve as a reference for the users when they write their own Devito powered code? Does it really matter what it returns?

Then, if they want to use the code provided with the examples, they can customize it as they see fit

mloubout: As I said, fine with me like that, I'll make different operators anyway.

FabioLuporini:
if the latest fix makes TTI working properly, I think we should merge this ASAP !

FabioLuporini:
@vincepandolfo and I keep finding bugs here and there, so will wait a little more before merging -- experimentation is in progress

FabioLuporini:
@vincepandolfo I think we should merge this and then proceed with restructuring the examples directory as we discussed (common frontend + porting acoustic to using ``Benchmark`` + individual subdirectories for each example)

vincepandolfo:
I agree. If no one has any comments on this I think you can proceed with the merge

FabioLuporini:
Merged.

PR #109, opened 2016-09-02
few more FD shortcut

mloubout:
Just a few extra finite-difference shortcuts used to have density in the operators.

### DIFF
--- a/devito/finite_difference.py
+++ b/devito/finite_difference.py
@@ -152,6 +152,8 @@ def first_derivative(*args, **kwargs):
     # Finite difference weights from Taylor approximation with this positions
     c = finite_diff_weights(1, ind, dim)
     c = c[-1][-1]
+    if side == 0:
+        side = 1
     # Diagonal elements
     for i in range(0, len(ind)):
             var = [a.subs({dim: ind[i]}) for a in args]
--- a/devito/finite_difference.py
+++ b/devito/finite_difference.py
@@ -7,7 +7,8 @@ from sympy import finite_diff_weights, symbols
 
 from devito.dimension import x, y
 
-__all__ = ['second_derivative', 'cross_derivative']
+__all__ = ['first_derivative', 'second_derivative', 'cross_derivative',
+           'left', 'right', 'centered']
 
 
 # Explicitly derived Finite Difference coefficients for
@@ -28,6 +29,24 @@ fd_coefficients = {
 
 # Default spacing symbol
 h = symbols('h')
+
+
+class Side(object):
+    """Class encapsulating the side of the shift for derivatives."""
+
+    def __init__(self, side):
+        self._side = side
+
+    def __eq__(self, other):
+        return self._side == other._side
+
+    def __repr__(self):
+        return {-1: 'left', 0: 'centered', 1: 'right'}[self._side]
+
+
+left = Side(-1)
+right = Side(1)
+centered = Side(0)
 
 
 def second_derivative(*args, **kwargs):
@@ -138,9 +157,6 @@ def first_derivative(*args, **kwargs):
     dim = kwargs.get('dim', x)
     diff = kwargs.get('diff', h)
     order = kwargs.get('order', 1)
-    left = -1
-    centered = 0
-    right = 1
     side = kwargs.get('side', centered)
     deriv = 0
     sign = 1
--- a/devito/finite_difference.py
+++ b/devito/finite_difference.py
@@ -127,6 +127,7 @@ def first_derivative(*args, **kwargs):
     :param dims: 2-tuple of symbols defining the dimension wrt. which
        to differentiate, eg. `x`, `y`, `z` or `t`.
     :param diff: Finite Difference symbol to insert, default `h`.
+    :param side: Side of the shift for the first derivatives.
     :returns: The cross derivative
 
     Example: Deriving the first-derivative of f(x)*g(x) wrt. x via:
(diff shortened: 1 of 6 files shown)


mlange05:
Looks nice, but could we have some simple tests for this? Maybe compare against sympy differentiation?

mloubout:
This is only centered first order FD. It is tested within sympy but I can make an extra test if necessary.

mloubout:
@opesci/opesci-dev 
Space orders had some weird stuff going on (order/2, order and order/4 interleaving for no reason). This is no fixed, plus tiny rewrite of 2D TTI with this ones (I can move it into another PR)

devito/finite_difference.py
```diff
@@ -137,18 +137,23 @@ def first_derivative(*args, **kwargs):
     dim = kwargs.get('dim', x)
     diff = kwargs.get('diff', h)
     order = kwargs.get('order', 1)
-    side = kwargs.get('side', 1)
+    side = kwargs.get('side', 0)
```
FabioLuporini: Instead of -1, 0, 1, which are magic numbers, can we not have proper constants ?

devito/finite_difference.py
```diff
@@ -137,18 +137,23 @@ def first_derivative(*args, **kwargs):
...
     deriv = 0
     # Stencil positions for non-symmetric cross-derivatives with symmetric averaging
     if side == 1:
         ind = [(dim + i * diff) for i in range(-int(order / 2) + 1 - (order % 2),
                                                int((order + 1) / 2) + 2 - (order % 2))]
-    else:
+    elif side == -1:
         ind = [(dim - i * diff) for i in range(-int(order / 2) + 1 - (order % 2),
                                                int((order + 1) / 2) + 2 - (order % 2))]
+    else:
```
FabioLuporini: is this the case ``side==0`` ?
if that's the case, can you change the value of side to 1 here (again: better off if no magic numbers are used), rather than below?

devito/finite_difference.py
mloubout: What do you mean proper constants?

devito/finite_difference.py
mloubout: Good point, will do.

devito/finite_difference.py
vincepandolfo: Words that mean something instead of numbers. It's definitely not easy to know what `1` or `0` is in that context

mloubout: I guess it's fine, I liked it like that because it was defining the sign in front of the output and avoid str comparison.

FabioLuporini: no need for string comparison. As @vincepandolfo was saying, I was just suggesting to have something like ``right`` or ``left`` instead of ``1`` or ``-1``, or perhaps better words. This would improve the code

mloubout: Done

FabioLuporini: to avoid string comparison and to make it cleaner, we should have something like (u may be able to find better names):

``
class Position(object):

    """
    Here a brief description
    """

    def __init__(self, position):
        self._position = position

    def __eq__(self, other):
        return self._position == other._position

    def __repr__(self):
        # Can return a suitable representation here

left = Position(-1)
right = Position(1)
centered = Position(0)
``

Then you can import ``left, right, centered`` where you need them. This approach is easily scalable for when you'll need more than just -1, 0, 1

mloubout: I have no idea what this does and how to use/do it.

devito/finite_difference.py
```diff
@@ -137,20 +137,27 @@ def first_derivative(*args, **kwargs):
     dim = kwargs.get('dim', x)
     diff = kwargs.get('diff', h)
     order = kwargs.get('order', 1)
-    side = kwargs.get('side', 1)
+    side = kwargs.get('side', "centered")
```
navjotk: I think what @FabioLuporini was asking for was:
```python
left = -1
centered = 0
right = 1
kwargs.get('side', centered)
...
if side == right:
...
```

String comparison definitely should not be going into the codebase

mloubout: what?

navjotk: I don't believe @FabioLuporini  was asking you to move to string comparisons. Only to use constants with more descriptive names, rather than numbers like 0 and 1 which don't mean anything to an average programmer who is focussed on another part of the code.

mloubout:
Changed according to @navjotk. If any other concern please look at it as this is critical to have proper discretization of all equations.

navjotk:
I plan to merge this as soon as I see travis green

mlange05:
We're still using "magic numbers in `interface.py`. What was wrong with what @FabioLuporini suggested, wrapping the "side" or "position" ids in a custom type? 

Also, I think the new derivatives introduced in this PR are still untested?

mloubout:
I've been doing changes back and forth for a week because everyone says something different. Second, this is `as_finit_diff` based (fully tested in sympy don't see the point of re-writing a test here) and generating proper coefficients.
If someone has a better way to do it, feel free to change it, but master is doing BS finite-differences currently so it can't last f orever.

mlange05:
Ok, so the point of testing is to test a feature, not its implementation. If I were to change the way we compute the derivative to something manual tomorrow and break stuff in the process, travis currently would not complain, which is bad! However, the test is trivial and I'll push it in a second.

mlange05:
Ok, I just pushed a simple little test and the constant wrapper class @FabioLuporini suggested. @mloubout, can you please take a final look and make sure this still works as expected? If so this is good to go, I think.

mloubout:
All good for me

FabioLuporini:
Good to go!

PR #113, opened 2016-09-08
New parser for Profiler

vincepandolfo:
The old parser didn't work well with CSE, causing wrong results to be produced.

This new parser should do a much better job and is also much easier to understand.

Also fixed a problem from #102 , only `flops->loop_body` and `flops->kernel` were updated. The problem never showed up before because in the examples we use them, the old parser couldn't count flops properly and had `0` as a result.


devito/profiler.py
```diff
@@ -167,69 +179,70 @@ def _get_for_flops(self, name, loop, loads):
...
-            loads[cur_load] += 1
+        string = assign.lvalue + " " + assign.rvalue
+
+        to_ignore = [
+            "int",
+            "float",
+            "double",
+            "F",
+            "i",
+            "t",
```
mlange05: `t` and `i` should probably also come from `Iteration` in the future, but this is good for now.

mlange05:
I can believe this, and the code is indeed much nicer now. Can we have a simple test though that ensures we're not counting any special functions or indexing operations? Also, @FabioLuporini, can you have a look as well to ensure this works as expected now?

devito/profiler.py
FabioLuporini: actually, I agree

devito/profiler.py
```diff
@@ -81,18 +86,23 @@ def add_profiling(self, code, name, byte_size=4, omp_flag=None):
                           Defaults to 4.
         :param omp_flag: OpenMP flag to add before profiling operations,
                          if needed.
+        :param to_ignore: List of strings containing the labels of
+                          symbols used as loop variables
```
FabioLuporini: aren't the loop variables fixed, and therefore known already?

devito/profiler.py
```diff
@@ -167,69 +179,70 @@ def _get_for_flops(self, name, loop, loads):
...
+            "t",
+            "fabsf",
+            "e",
+            "temp",
+            "p",  # This one shouldn't be here.
+                  # It should be passed in by an Iteration object.
+                  # Added only because tti_example uses it.
+        ] + to_ignore
+
+        symbols = re.findall(r"[a-z_]+\d?", string)
```
FabioLuporini: short comment for people who don't know reg exprs or that have to re-learn reg exprs every time they see something slightly more advanced ?

FabioLuporini:
much better! We should merge this very soon.

As the loop variables are given (in the end, we set them!), it seems that the ``to_ignore`` variable can be removed. We might just be able to import this information from ``Iteration``

Once this is sorted out, we can proceed with landing

devito/profiler.py
vincepandolfo: While it's probably possible to extract some loop variables from the `cgen.For` objects, some others (like the time steppers) are not declared in the For statement and need to be passed in using `to_ignore`. 

We might as well have them all passed in `to_ignore`

FabioLuporini: Whenever a new Iteration object is instantiated, we could update a global data structure (eg belonging to the class, not to the instances) which keeps track of all Iteration objects instantiated. Then the Profiler could just have to query Iteration (e.g. through a static method) to get all used indices

I agree it's probably pointless to extract them from cgen

mlange05: I don't think a global registry approach would work, since it is quite conceivable to have multiple `Operators` in an application, each with a distinct set of `Iterations` and it's own profiler.

FabioLuporini: Sorry: I was talking about loop indices, not actual ``Iteration`` objets (I confused the names), as in the docstring of ``to_ignore`` I see "List of strings containing the labels of symbols used as loop variables"

mlange05: Hmm, in that case, yes that would be possible. I'm still not sure though what we would gain from a global approach over querying the relevant `Iteration` and `Expression` hierarchy explicitly in the `Operator `and passing the information to it's profiler - once the internals always use them that is, of course.

FabioLuporini: You're right . How difficult would this be?

mlange05: Hard to say, but we'll need to either re-invent or retro-fit almost all of the `Propagator` internals to the `Iteration`/`Expression` classes, including pragmas, loop blocking, etc. I'll update issue #43, and get on it as soon as I can. In the meantime, I think the `to_ignore` approach is a good enough placeholder, so it shouldn't stop this merge.

FabioLuporini: OK, I do agree !

mlange05:
I think this is good to go now. Merging...

PR #119, opened 2016-09-12
AT fix

mloubout:
Profiler switched on at declaration to have proper number outputed. Reduction clause are ignored if the profiler is turned on inside `at_controller` afterward.

### DIFF
--- a/examples/Acoustic_codegen.py
+++ b/examples/Acoustic_codegen.py
@@ -64,7 +64,7 @@ class Acoustic_cg:
         if auto_tune:  # auto tuning with dummy forward operator
             fw = ForwardOperator(self.model, self.src, self.damp, self.data,
                                  time_order=self.t_order, spc_order=self.s_order,
-                                 save=False, profile=False)
+                                 save=False, profile=True)
             self.at = AutoTuner(fw)
             self.at.auto_tune_blocks(self.s_order + 1, self.s_order * 4 + 2)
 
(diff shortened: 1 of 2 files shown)


examples/acoustic_example.py
```diff
@@ -3,15 +3,16 @@
...
 model = IGrid()
 model0 = IGrid()
 model1 = IGrid()
 model.shape = dimensions
 model0.shape = dimensions
 model1.shape = dimensions
-origin = (0., 0.)
-spacing = (20., 20.)
+origin = (0., 0., 0.)
+spacing = (20., 20., 20.)
```
FabioLuporini: This is fixed by #115 so we should first merge that, and then rebase this on top of it

FabioLuporini:
Apart from the minor comment, this should be merged ASAP

examples/Acoustic_codegen.py
```diff
@@ -64,7 +64,7 @@ def damp_boundary(damp):
         if auto_tune:  # auto tuning with dummy forward operator
             fw = ForwardOperator(self.model, self.src, self.damp, self.data,
                                  time_order=self.t_order, spc_order=self.s_order,
-                                 save=False)
+                                 save=False, profile=False)
```
vincepandolfo: `profile` defaults to `False`. Why is this needed?

mloubout: Should be True my bad.

mloubout:
This one is good to go I think

mlange05:
Yes, it is. Merging...

PR #120, opened 2016-09-13
Examples reorganization and benchmark

vincepandolfo:
- removed `AcousticWave2D.py` (doesn't use Devito)
- `acoustic_example` can now be benchmarked using `benchmark.py` (exactly like TTI in #107 )
- removed `+x` permission on `.py` files
- reorganized the structure of the example folder


mlange05:
While we're shuffling and renaming things, can we come up with a more descriptive filename than `Acoustic_codegen` and `TTI_codegen`? Consistent capitalisation would also be nice.

vincepandolfo:
What would a good name be? `acoustic_setup`?

examples/acoustic/acoustic_example.py
```diff
@@ -0,0 +1,79 @@
...
+    rec, u, gflops, oi = Acoustic.Forward(
+        save=True, cse=cse, auto_tuning=auto_tuning, compiler=compiler
+    )
+
+    if not full_run:
+        return gflops, oi, [rec, u.data]
+
+    Acoustic.Adjoint(rec)
+    Acoustic.Gradient(rec, u)
+    Acoustic.Born(dm)
```
mloubout: Please put back the "print" or "log" statement so that people know what's going on when running it.

examples/tti/tti_example.py
```diff
@@ -55,7 +55,7 @@ def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0), tn=250.0,
     rec, u, v, gflops, oi = TTI.Forward(
         cse=cse, auto_tuning=auto_tuning, cache_blocking=cache_blocking, compiler=compiler
     )
-    return gflops, oi, rec, u, v
+    return gflops, oi, [rec, u, v]
```
mloubout: As pointed before, outputing gflops and OI at all time is not necessarly the best way. Makes practical usage more complicated than necessary. Jut my point of view

mloubout:
I am gonna ask again. Can we separate benchmark/profle output (OI, GFLOPS, ..) from the application outputs (rec, u). This has to be usable in a simple way. We can't have a very good API that just spill out everything. I however do not have a solution for it but I know that a lot of users will be pretty unhappy if they have to spend their time sorting out outputs.

mlange05:
The problem with separating benchmark and application output is really that the current `Forward` et al. wrappers swallow the `Operator` object by handling input and output arrays explicitly. To fix this, I believe we would need to drop these wrappers altogether and let the outer example code use the operator objects directly. This would allow the profiler info to be accessed from the outer benchmarking layers as well, since we only need to query the operator object. In order for the application/example code to only deal with arrays instead of function objects, however, we would need to extend the operator to handle the input/output in array form.

FabioLuporini:
Apart from Mathias' comments, which are reasonable but seem to require some thinking, is there any other blocker to this ?

mlange05:
Looks good to me - although as pointed out above we need to re-think the example structure a little further, but that is beyond this PR. It does need a manual rebase now though.

PR #122, opened 2016-09-13
Auto tuning now done inside Operator

vincepandolfo:
Instead of having the user manually create an `Operator` to perform autotuning and then assign the tuned block sizes by hand, this PR makes the whole process easier by performing AT inside the operator.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,10 +99,5 @@ Note:
 
 Example usage:
 ```
-op = Operator(...)
-at = AutoTuner(op, [True, True, False], <at_report_directory_path>)
-at.auto_tune_blocks(min_block_size, max_block_size)
-
-#using auto tuned block size
-new_op = Operator(..., cache_blocking=at.block_size)
+op = Operator(..., auto_tuning=True, blocked_dims=[True, True, False], at_range=(4, 32))
 ```
--- a/devito/at_controller.py
+++ b/devito/at_controller.py
@@ -2,12 +2,11 @@ import random
 from os import mkdir, path
 
 import logger
-from devito.operator import Operator
 
 
 class AutoTuner(object):
 
-    def __init__(self, op, blocked_dims=None, at_report_dir=None):
+    def __init__(self, op, blocked_dims, at_report_dir=None):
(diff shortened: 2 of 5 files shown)


mlange05:
How do you use stored results from a previous AT run with this?

navjotk:
Having the autotuner separate from the `Operator` was a suggestion from me based on the idea that the `GridSearch` in scikit-learn is a good API to model this behaviour on. 
**http://scikit-learn.org/stable/modules/grid_search.html**

vincepandolfo:
Autotuning can still be done "The old way" if needed

mloubout:
This looks like overkill to me. If you want auto_tuning you have to run it every time? And how do you pass the result from one operator to another?

vincepandolfo:
As I said, the old way of doing it still works.

mlange05:
Ok several things here:
* Adding a second API that is missing a key feature (ie. run with previous results) does not sound like a good idea to me, since it only adds confusion. In fact, making AT part of the Operator was rejected in the original AT PR, since the AT process offers a lot of specific parameters that the Operator does not need to know anything about.
* The way AT is run in the current example is not great either, since it seems to always run `auto_tun_blocks`, regardless of whether the full-range AT has run before or not. This needs fixing, so that an average user never needs to run AT more than once for a given operator.
* The "cloning" of the Operator in this PR could potentially create an unnecessary symbolic overhead, and so I really don't think the Operator needs to know about AT at all, at least not for the brute-force approach we are currently taking. This might change if we chose to integrate AT with the main time loop itself, but there is a long list of design changes that need to happen before this.

mloubout:
> The way AT is run in the current example is not great either, since it seems to always run auto_tun_blocks, regardless of whether the full-range AT has run before or not. This needs fixing, so that an average user never needs to run AT more than once for a given operator.

This was not the case at original merge. AT was done in `__init__` of `Acoustic_codegen` and then just reused in every operators

vincepandolfo:
It seems this is not a needed feature nor an improvement. Closing.

PR #123, opened 2016-09-13
Removing redundant concept of pre and post kernel steps

navjotk:
Fairly non-controversial. Removing the redundant lines of code pointed out in https://github.com/opesci/devito/issues/121


mloubout:
What do you mean redundant lines?

FabioLuporini:
I like removing stuff. I'm happy with this

PR #124, opened 2016-09-15
Reduction fix when cache_blocking

vincepandolfo:
When cache_blocking the profiler would return the wrong number of flops. 

This PR fixes this issue, while also reorganising the profiling to reduce the coupling between `Propagator` and `Profiler`.


vincepandolfo:
Now part of #126

PR #126, opened 2016-09-15
Fixed OI calculation when blocking + profiler reorganization

vincepandolfo:
This is on top of #124.

As per title.

EDIT: #124 now bundled with this.


mlange05:
This looks good to me, although I would like a second opinion (@FabioLuporini?). Once confirmed this is ready to go in.

devito/profiler.py
```diff
@@ -115,133 +108,251 @@ def add_profiling(self, code, name, byte_size=4, omp_flag=None, to_ignore=None):
...
+        oi = float(num_flops) / (self.byte_size*(len(loads) + loads["stores"] - 1))
+        oi_low = oi / (self.byte_size*load_val_sum)
+        num_loads = load_val_sum - loads["stores"]
+
+        return oi, oi_low, num_loads, self.temps
+
+    def _get_loop_reduction(self, variable):
+        return " reduction(+:%s)" % variable
+
+    def _add_reduction(self, pragma, name):
```
FabioLuporini: Can we not just squash this method and ``get_loop_reduction`` into a single method ``_handle_reduction`` ?

mloubout:
Does this fix the reduction problem I pointed out in slack?

FabioLuporini:
According to @vincepandolfo it should, but I haven't followed specifically that problem

devito/profiler.py
vincepandolfo: Yes

FabioLuporini:
Once travis is green, this can probably land .

mlange05:
Good to go now, merging...

PR #127, opened 2016-09-15
Benchmark saves only loop_body data

vincepandolfo:
`benchmark.py` used the data for the whole kernel, while only the main loop is needed (according to discussion with @mloubout on Slack)

### DIFF
--- a/examples/benchmark.py
+++ b/examples/benchmark.py
@@ -133,7 +133,7 @@ if __name__ == "__main__":
                     np.isclose(res[i], last_res[i])
 
     elif args.execmode == "bench":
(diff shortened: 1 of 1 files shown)


FabioLuporini:
Yes, it's correct, because we want to model a very specific kernel.

It'd be nice to report also how much time is spent in the rest of the kernel (or what percentage of time). I think this is easily doable -- could you add it?

FabioLuporini:
OK, thanks!

FabioLuporini:
I think this is quite uncontroversial, so merging...

FabioLuporini:
Merged.

PR #128, opened 2016-09-15
Separation of SourceLike, fwi operators in acoustic

mloubout:
Separation of Source/Receiver calsss from the operators.


examples/acoustic/acoustic_example.py
```diff
@@ -24,7 +24,7 @@ def source(t, f0):
     return (1-2.*r**2)*np.exp(-r**2)
 
 
-def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0), tn=250.0,
+def run(dimensions=(150, 150, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
```
vincepandolfo: Spacing is passed as a two elements tuple in `benchmark.py`. Can you modify that as well and do this same thing for tti_example?

mloubout: Spacing should be same size as dimensions

vincepandolfo: That's why it needs to be changed in `benchmark.py` and `tti_example.py` as well

examples/source_type.py
```diff
@@ -0,0 +1,116 @@
...
+        return sum([b.subs(subs) * u.indexed[(t, ) + idx]
+                    for idx, b in zip(index_matrix, self.bs)])
+
+    def read(self, u):
+        """Iteration loop over points performing grid-to-point interpolation."""
+        interp_expr = Eq(self.indexed[t, p], self.grid2point(u))
+        return [Iteration(interp_expr, index=p, limits=self.shape[1])]
+
+    def read2(self, u, v):
+        """Iteration loop over points performing grid-to-point interpolation."""
```
FabioLuporini: Using the same docstring as ``read(...)`` -- specify what differs here please

examples/source_type.py
```diff
@@ -0,0 +1,116 @@
...
+        interp_expr = Eq(self.indexed[t, p], self.grid2point(u) + self.grid2point(v))
+        return [Iteration(interp_expr, index=p, limits=self.shape[1])]
+
+    def add(self, m, u, t=t):
+        """Iteration loop over points performing point-to-grid interpolation."""
```
FabioLuporini: Using the same docstring as ``read(...)`` -- specify what differs here please

examples/tti/tti_example2D.py
```diff
@@ -46,7 +46,7 @@ def source(t, f0):
 data.set_shape(nt, 301)
 
 TTI = TTI_cg(model, data, None, t_order=2, s_order=spc_order, nbpml=10)
-(rec, u, v) = TTI.Forward()
+(rec, u, v, gflops, oi, timings) = TTI.Forward()
```
FabioLuporini: This works even if you don't put parentheses around the LHS

mlange05:
I think this is good to go now, so merging...

Issue #96, opened 2016-08-22
Numpy dependency


navjotk:
In trying to install devito on the yemoja cluster I realised the following:
- Our `requirements.txt` does not list numpy as a dependency
- Travis.yaml does not download/install numpy in any special way
- The tests pass on travis
- Running the same tests on the cluster, with numpy `1.8`, 3 tests fail. These same tests pass on version `1.11` of numpy.

Which version of numpy do we need?
Should this be added to requirements.txt? 
How does travis know which version of numpy to use?

mlange05:
Is this fixed now by PR #129?

mlange05:
@navjotk Is this now fixed by PR #129?

navjotk:
Yes it is, thanks. Closing.

PR #129, opened 2016-09-16
Adding explicit numpy requirement

mlange05:
Making our version dependency on `numpy` explicit to address issue #96.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,8 @@
-git+git://github.com/inducer/cgen
+numpy>=1.11.1
 sympy>=0.7.6
+pytest
 flake8>=2.1.0
 isort
 py-cpuinfo
-pytest
+git+git://github.com/inducer/cgen
 git+git://github.com/inducer/codepy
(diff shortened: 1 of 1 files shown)


FabioLuporini:
uncontroversial

mloubout:
Shouldn't `pytools`  be in it?
And `numpy` 1.1 needs a separate installation of `mpmath`

mlange05:
`pytools` isn't imported by the core package, but `mpmath` is indeed missing. Good spot, thanks.

mloubout:
>pytools isn't imported by the core package
What do you mean? `cgen` is using `pytools` but only `cgen` in requirements is not installing it

mlange05:
`pip` should take care of any dependencies when it installs `cgen` (the dependency is defined in cgen's `setup.py`). We aren't and shouldn't be responsible for tracking the dependencies of other packages, only our own.

mloubout:
OK

navjotk:
Weird travis failures

mlange05:
Travis-OSX builders sporadically die on our TTI example, probbaly due to the high memory demand crippling the VM. It's all green now though.

PR #130, opened 2016-09-16
Profiler: Byte size fix

vincepandolfo:
The `dtype` was never passed to the Profiler, so it always calculated OI based on 4 bytes long floats. This fixes it by having the propagator pass its dtype to the profiler.

Also changed name from `byte_size` to `float_size`


FabioLuporini:
@mlange05 shall we merge this?

mlange05:
Yes, good to go once green again.

PR #132, opened 2016-09-22
Many AT fixes

FabioLuporini:
There are two spurious commits in this PR, which fix the ``acoustic`` example when benchmarking. 
Apart from those, with this PR a number of issues concerning the AT are fixed:

- Avoid unacceptably long AT runs, because too many block sizes were attempted. Now a heuristic is used: only square blocks whose size is a multiple of 2 are picked. This can be refined by extending the corresponding method ``at_controller.py:AutoTuner:_filter`` 
- Avoid executing useless sample runs. 
- Avoid executing a dry run, which was just a hack to populate some lists and make things work. The dry run was also using block sizes that could go beyond the limits imposed by ``minimum`` and ``maximum``, so this might also result in a C-level bug. 
- Fill in output arrays with some random values, to ensure that the AT works on initialized data. Otherwise, the AT output is just garbage, as some runs (some block sizes) would end up with a lower memory traffic.

Also add a logger function for AT

### DIFF
--- a/devito/at_controller.py
+++ b/devito/at_controller.py
@@ -92,25 +92,23 @@ class AutoTuner(object):
         at_nt = 3
         self.op.propagator.nt = at_nt
         self.op.propagator.profile = True
-        self.op.propagator.cache_blocking = [0 if dim else None
-                                             for dim in self.blocked_dims]
 
         info_at("Start. Mode: brute force")
 
-        times = []  # list where times and block sizes will be kept
         block_list = set()  # used to make sure we do not test the same block sizes
-        block = list(self.blocked_dims)
+        mask = [i if i else None for i in self.blocked_dims]
+        block = [None for i in mask]
 
         for x in range(minimum, maximum):
-            block[0] = self.blocked_dims[0] and x
+            block[0] = mask[0] and x
 
             if len(block) > 1:
                 for y in range(minimum, maximum):
-                    block[1] = self.blocked_dims[1] and y
+                    block[1] = mask[1] and y
 
                     if len(block) > 2:
                         for z in range(minimum, maximum):
-                            block[2] = self.blocked_dims[2] and z
+                            block[2] = mask[2] and z
                             block_list.add((tuple(block)))
                     else:
                         block_list.add(tuple(block))
@@ -128,6 +126,7 @@ class AutoTuner(object):
         info_at("Number of block sizes that will be attempted: %d" % len(block_list))
 
         # runs function for each block_size
+        times = []
         self.op.propagator.cache_blocking = list(block_list[0])
         for block in block_list:
             self.op.propagator.block_sizes = list(block)
@@ -150,10 +149,12 @@ class AutoTuner(object):
 
             * block sizes that are not a multiple of 2 are ditched;
             * only square blocks are retained.
+
+        If no block sizes match the heuristic, then the original list is returned.
         """
-        block_list = [b for b in block_list if all(i % 2 == 0 for i in b if i)]
-        block_list = [b for b in block_list if all(i == b[0] for i in b if i)]
-        return block_list
+        filtered_list = [b for b in block_list if all(i % 2 == 0 for i in b if i)]
+        filtered_list = [b for b in filtered_list if all(i == b[0] for i in b if i)]
+        return filtered_list or block_list
 
     def get_execution_time(self):
         """Runs and times the function
--- a/devito/at_controller.py
+++ b/devito/at_controller.py
@@ -1,4 +1,7 @@
+from __future__ import absolute_import
+
 from os import mkdir, path
+from operator import itemgetter
 
 import numpy as np
 
@@ -133,7 +136,7 @@ class AutoTuner(object):
             times.append((block, self.get_execution_time()))
 
         # sorts the list of tuples based on time
-        times = sorted(times, key=lambda element: element[1])
+        times = sorted(times, key=itemgetter(1))
 
         info_at("Finish.")
         info_at("Estimated runtime for %s and %d time steps: %f hours" %
--- a/devito/at_controller.py
+++ b/devito/at_controller.py
@@ -1,5 +1,7 @@
 from os import mkdir, path
 
+import numpy as np
+
 from devito.logger import info_at, error
 from devito.operator import Operator
 
@@ -25,7 +27,7 @@ class AutoTuner(object):
         self.nt_full = self.op.nt
 
         default_blocked_dims = ([True] * (len(self.op.shape) - 1))
-        default_blocked_dims.append(False)  # By default we don't auto tune inner most dim
+        default_blocked_dims.append(None)  # By default don't autotune innermost dim
         self.blocked_dims = blocked_dims or default_blocked_dims
 
         default_at_dir = path.join(path.dirname(path.realpath(__file__)), "At Report")
@@ -92,39 +94,43 @@ class AutoTuner(object):
         self.op.propagator.profile = True
         self.op.propagator.cache_blocking = [0 if dim else None
                                              for dim in self.blocked_dims]
-        # want to run function at least once to make sure block_sizes are not
-        # overwritten
-        self.get_execution_time()
+
         info_at("Start. Mode: brute force")
 
         times = []  # list where times and block sizes will be kept
         block_list = set()  # used to make sure we do not test the same block sizes
-        blocks = list(self.op.propagator.block_sizes)
+        block = list(self.blocked_dims)
 
         for x in range(minimum, maximum):
-            blocks[0] = x if blocks[0] else None
+            block[0] = self.blocked_dims[0] and x
 
-            if len(blocks) > 1:
+            if len(block) > 1:
                 for y in range(minimum, maximum):
-                    blocks[1] = y if blocks[1] else None
+                    block[1] = self.blocked_dims[1] and y
 
-                    if len(blocks) > 2:
+                    if len(block) > 2:
                         for z in range(minimum, maximum):
-                            blocks[2] = z if blocks[2] else None
-                            block_list.add((tuple(blocks)))
+                            block[2] = self.blocked_dims[2] and z
+                            block_list.add((tuple(block)))
                     else:
-                        block_list.add(tuple(blocks))
+                        block_list.add(tuple(block))
             else:
-                block_list.add(tuple(blocks))
+                block_list.add(tuple(block))
 
         # filter off some of the block sizes, heuristically
-        block_list = self._filter(block_list)
+        block_list = sorted(self._filter(block_list))
+
+        # populate output arrays with random values, to make sure that loads
+        # and stores are performed
+        for param in self.op.output_params:
+            param._data = np.random.rand(*param.data.shape).astype(self.op.dtype)
 
         info_at("Number of block sizes that will be attempted: %d" % len(block_list))
 
         # runs function for each block_size
-        for block in sorted(block_list):
-            self.op.propagator.block_sizes = block
+        self.op.propagator.cache_blocking = list(block_list[0])
+        for block in block_list:
+            self.op.propagator.block_sizes = list(block)
             times.append((block, self.get_execution_time()))
 
         # sorts the list of tuples based on time
--- a/devito/at_controller.py
+++ b/devito/at_controller.py
@@ -1,4 +1,3 @@
-import random
 from os import mkdir, path
 
 from devito.logger import info_at, error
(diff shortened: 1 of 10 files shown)


FabioLuporini:
I know that some tests are still failing, but I've opened the PR so that people can start looking into it

FabioLuporini:
Now there should be only one failing test, which has to do with the fact that I changed the precision of Acoustic from float64 to float32. 

I don't remember what it should be. @mloubout enlighten us please

mloubout:
`gradient_test` should fail. if `adjoint_test` fails I'l look into the numbers to validate it or not.

devito/at_controller.py
```diff
@@ -91,71 +92,69 @@ def auto_tune_blocks(self, minimum=5, maximum=20):
...
+
+        # filter off some of the block sizes, heuristically
+        block_list = sorted(self._filter(block_list))
 
-        logger.info("Total number of block size permutations = %d" % len(block_list))
-        if maximum - minimum > 2:  # No point to estimate if diff is below 2.
-            self._estimate_run_time(minimum, maximum, len(block_list))
+        # populate output arrays with random values, to make sure that loads
+        # and stores are performed
+        for param in self.op.output_params:
```
mlange05: Why do we populate `output` and not `input_params`?

devito/at_controller.py
```diff
@@ -91,71 +92,69 @@ def auto_tune_blocks(self, minimum=5, maximum=20):
...
 
-            self.op.propagator.block_sizes = blocks
-            timing_run += self.get_execution_time()
-
-        logger.info("Estimated runtime: %f minutes." % float(timing_run / 5 * n / 60))
+        If no block sizes match the heuristic, then the original list is returned.
+        """
+        filtered_list = [b for b in block_list if all(i % 2 == 0 for i in b if i)]
+        filtered_list = [b for b in filtered_list if all(i == b[0] for i in b if i)]
+        return filtered_list or block_list
```
mlange05: Looks great for now, but I wonder if we should just generate the block list according to some heuristic, instead of creating a list and then trimming it. This is not a show-stopper though, just a thought.

devito/at_controller.py
```diff
@@ -91,71 +92,69 @@ def auto_tune_blocks(self, minimum=5, maximum=20):
...
         # runs function for each block_size
+        times = []
+        self.op.propagator.cache_blocking = list(block_list[0])
         for block in block_list:
-            self.op.propagator.block_sizes = block
+            self.op.propagator.block_sizes = list(block)
             times.append((block, self.get_execution_time()))
 
         # sorts the list of tuples based on time
         times = sorted(times, key=lambda element: element[1])
```
mlange05: Technically not code that was touched, but while we're cleaning house: this should be `operator.itemgetter` instead of a `lambda` function. Fixing this is optional.

devito/at_controller.py
FabioLuporini: probably yes. 

BTW, I'm trimming it twice to highlight how the two rules in the docstring are applied

devito/at_controller.py
FabioLuporini: Right, I'll fix it

devito/at_controller.py
FabioLuporini: Ah, good point.

So, I thought that the output tensor are usually 0-valued (they represent the ``output`` of the computation indeed), while the input tensors will be read in the kernels, so they should never be 0-valued (unless being output tensors too -- ie, the RW access descriptor in pyop2). That's why, but perhaps I've missed something ?

FabioLuporini:
@mloubout Later today I'll introduce the minor fixes pointed out by @mlange05 ; once we know how to deal with the acoustic example, this gets a green light and shall be merged

devito/at_controller.py
mlange05: Hmm, ok, you're right, the ones you're after are the RW ones, really. We don't have particularly sophisticated access descriptors yet, so I guess this is fine then.

examples/acoustic/Acoustic_codegen.py
```diff
@@ -63,17 +63,16 @@ def damp_boundary(damp):
...
                 auto_tuning=False, cse=True, compiler=None):
+
         if auto_tuning:
-            cache_blocking = self.at.block_size
+            fw = ForwardOperator(self.model, self.src, self.damp, self.data,
+                                 time_order=self.t_order, spc_order=self.s_order,
+                                 profile=True, save=False, cse=cse, compiler=compiler)
+            at = AutoTuner(fw)
+            at.auto_tune_blocks(self.s_order + 1, self.s_order * 4 + 2)
+            cache_blocking = at.block_size
```
FabioLuporini: So this means that we have a conceptual bug

If we can reuse an AT with different operators ("they use the same stencil roughly"), then it means AT should be associated not with an operator, but rather with a different abstraction that captures the key computational properties of a set of operators. 

I understand your point, but that's just a hack given the structure that we currently have.

devito/at_controller.py
```diff
@@ -91,71 +92,69 @@ def auto_tune_blocks(self, minimum=5, maximum=20):
...
+                            block_list.add((tuple(block)))
                     else:
-                        block_list.add(tuple(blocks))
+                        block_list.add(tuple(block))
             else:
-                block_list.add(tuple(blocks))
+                block_list.add(tuple(block))
+
+        # filter off some of the block sizes, heuristically
+        block_list = sorted(self._filter(block_list))
```
FabioLuporini: dumb question time: is it even legal to have 1 ? I say ``minimum`` is a function of the stencil shape, and 1 breaks this limit in most circumstances...

Can you also elaborate on "usually work better". By how much ?

examples/acoustic/Acoustic_codegen.py
mlange05: There is currently no abstraction that groups operators other than "the user application", and the base `devito.Operator` doesn't know anything about AT (see rejected PR #122). However, within the user application space, which `examples` are really, it is up to the application developer which assumptions can be made. The overarching abstraction you are looking for is probably "Acoustic" vs. "TTI", but that is a design problem within the examples.

devito/at_controller.py
FabioLuporini: Done

examples/acoustic/Acoustic_codegen.py
FabioLuporini: Fair enough.
However, that commit also fixes a bug, which is: the dummy operator needs to be told about ``cse`` and, especially, ``compiler`` (otherwise, it runs without openmp). Now, these two flags aren't available in __init__. I can be OK with performing AT in __init__ as before, but someone has to tell me how do I get them

devito/at_controller.py
mloubout: Changed since your fix. All I mean is (n, n, None) is not always the best block size, but yest without filtering it's just not feasible.

examples/acoustic/Acoustic_codegen.py
mloubout: All I am saying is, running AT for Forward, adjoint, .... and even when you cal forward again is just an overkill, while aving an "Acoustic" AT makes more sense, as you only need to run AT once for one stencil-shape/stencil and reuse it. Same for `cse`, ... you give it to `init_` acoustic and everything in it use the same setup.

FabioLuporini: I got that, it's just that I don't know how to (cleanly) do the following:
"Same for cse, ... you give it to init_ acoustic and everything in it use the same setup"

mloubout: I agree that doing it cleanly is not easy, but currently, having to run AT at every single `forward` call is overkill.

examples/acoustic/Acoustic_codegen.py
```diff
@@ -63,17 +63,16 @@ def damp_boundary(damp):
...
-            self.at.auto_tune_blocks(self.s_order + 1, self.s_order * 4 + 2)
-
     def Forward(self, save=False, cache_blocking=None,
                 auto_tuning=False, cse=True, compiler=None):
+
         if auto_tuning:
-            cache_blocking = self.at.block_size
+            fw = ForwardOperator(self.model, self.src, self.damp, self.data,
+                                 time_order=self.t_order, spc_order=self.s_order,
+                                 profile=True, save=False, cse=cse, compiler=compiler)
```
mloubout: Should be `save=save` and the same for the other `fw`.

devito/at_controller.py
```diff
@@ -91,71 +95,69 @@ def auto_tune_blocks(self, minimum=5, maximum=20):
...
+        mask = [i if i else None for i in self.blocked_dims]
+        block = [None for i in mask]
 
         for x in range(minimum, maximum):
-            blocks[0] = x if blocks[0] else None
+            block[0] = mask[0] and x
 
-            if len(blocks) > 1:
+            if len(block) > 1:
                 for y in range(minimum, maximum):
```
mloubout: `for y in range(minimum, x + 1):`
block size should be considered commutative meaning if (x, y, None) is tested (y, x, None) should be ignored.

examples/acoustic/Acoustic_codegen.py
FabioLuporini: I understand. However, we should also stop hacking the code base, so I'm looking forward to systematic solutions...

examples/acoustic/Acoustic_codegen.py
FabioLuporini: right ! fixing this...

devito/at_controller.py
FabioLuporini: I can buy this. However, someone still has to convince me about the need for testing non-square blocks

examples/acoustic/Acoustic_codegen.py
mloubout: It doesn't. Forward takes `save` as an input to be able to run Gradient, this `save=False` is killing everything.

devito/at_controller.py
mloubout: TTI stencil is not squared

devito/at_controller.py
FabioLuporini: Is (n, 1, None) equivalent/similar to an ``omp pragma collapse`` ?

devito/at_controller.py
FabioLuporini: ok, I'm changing ``for y in range(minimum, maximum):`` into ``for y in range(minimum, x + 1):``

as for the non-squared block sizes: I agree they might be better -- but by how much ? is the gain ever significant ? I'm happy to relax the filtering once we have compelling reasons to do so

examples/acoustic/Acoustic_codegen.py
FabioLuporini: OK, I'm gonna now push a change that moves AT back into __init__. For this, ``cse`` and ``compiler`` are passed to ``Acoustic_cg``. 

Many parameters, however, are now passed to *both* the constructor (__init__) and the operators methods. Can we not just pass them to __init__, and then use ``self.`` in all methods? or do you want a default behaviour + per-operator knobs, as it would be now?

FabioLuporini:
OK, pushed a few fixes here and there that address both your previous comments and the failing tests (I'm making test_gradient as `xfail`since now we make use of float32 instead of float64)

I wanna merge this ASAP. Then, I'll file a new PR to merge the performance_measurement branch. Once both are in, we should focus again on having a working TTI as well as performance (acoustic, tti)

devito/at_controller.py
mloubout: No really. Collapse just allow to parallelize over the second dimension too, while `(n, 1, None) ` blocks only one direction. If `n` is too big some threads do nothing.

examples/acoustic/Acoustic_codegen.py
```diff
@@ -63,16 +64,18 @@ def damp_boundary(damp):
...
-    def Forward(self, save=False, cache_blocking=None,
-                auto_tuning=False, cse=True, compiler=None):
-
-        if auto_tuning:
+        if auto_tuning:  # auto tuning with dummy forward operator
             fw = ForwardOperator(self.model, self.src, self.damp, self.data,
                                  time_order=self.t_order, spc_order=self.s_order,
                                  profile=True, save=False, cse=cse, compiler=compiler)
             at = AutoTuner(fw)
             at.auto_tune_blocks(self.s_order + 1, self.s_order * 4 + 2)
```
mloubout: `self.at`

FabioLuporini: good spot

will fix this tomorrow morning, then merge and finally move straight the next branch

FabioLuporini:
Allright, merging as soon as travis is happy

FabioLuporini:
Merging...

FabioLuporini:
Merged.

PR #134, opened 2016-09-23
Adapting new `RooflinePlotter` API from opescibench in benchmarks

mlange05:
Resulting plots have now more customisation and look cooler. Requires opescibench PR #3.

### DIFF
--- a/examples/benchmark.py
+++ b/examples/benchmark.py
@@ -12,11 +12,11 @@ from examples.acoustic.acoustic_example import run as acoustic_run
 from examples.tti.tti_example import run as tti_run
 
 try:
-    from opescibench import Benchmark, Executor, Plotter
+    from opescibench import Benchmark, Executor, RooflinePlotter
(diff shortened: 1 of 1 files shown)


examples/benchmark.py
```diff
@@ -12,11 +12,11 @@
 from examples.tti.tti_example import run as tti_run
 
 try:
-    from opescibench import Benchmark, Executor, Plotter
+    from opescibench import Benchmark, Executor, RooflinePlotter
 except:
```
mloubout:
```
from opescibench import Benchmark, Executor
from opescibennch.plotter import RooflinePlotter
```
?

mlange05: No need, `opescibench` is a standalone package, at least from a Python point-of-view, so we don't need to be explicit about sub-modules.

FabioLuporini:
Yes, I like the new interface. This can land for me.

PR #136, opened 2016-10-04
Performance measurement fixes

FabioLuporini:
This PR drops the useless C-level flop and data movement counting and replaces it with Python-level counting.

The profiler now only computes the flop count. 
The data movement can be estimated in three ways (ideal, ideal_with_store, realistic ; documented in the code). Basically, the propagator knows how many fields are accessed, their size as well as the size of the iteration space, how they are accessed (read, write, read/write), and the data type, so one can compute multiple estimates for data movement (default: "realistic"). 
Flops and data are used to calculate the OI. 

A significant number of bugs are fixed. One in particular concerned the calculation of loads and stores by the profiler -- basically, CSE temporaries were counted as actual stores, thus resulting in a totally unreasonable OI. Other bugs were not counting increments as actual increments, computing the wrong OI for the ``kernel`` code section, and a few more. 

This PR also ships a number of minor changes/fixes (eg, fix broken tests, introduce a .gitignore, etc).

Comments welcome.

PS: ``profiler.py`` needs some more "purging", which I plan to do in a subsequent PR, through which I'll try to restore some dignity to the code base.


FabioLuporini:
@mlange05 There's a flake8 issue about assigning lambda functions. Can I just set it to ignore this rule ?

mloubout:
AT only writes 0.00 GFlops/s

devito/propagator.py
```diff
@@ -202,19 +205,154 @@ def timings(self):
...
+
+        :param mode: Several estimates are possible: ::
+
+            * ideal: also known as "compulsory traffic", which is the minimum
+                number of bytes to be moved (ie, models an infinite cache)
+            * ideal_with_stores: like ideal, but a data item which is both read
+                and written is counted twice (load + store)
+            * realistic: assume that all datasets, even those that do not depend
+                on time, need to be re-loaded at each time iteration
+        """
```
mloubout: Can we make `ideal_with_stores` the default behaviour? `realistic` is more like the lower bound and I makes more sense (to me) to compare with the upper bound.

FabioLuporini:
I didn't get your comment about "AT only writes 0.00 GFlops/s"?

You talking about a bug ? If so, how can I reproduce it ?

devito/propagator.py
mloubout: As  I said, TTI in this branch is not correct, benchmarking it is not usefull. ANd the main problem is to have acoustic running fully properly. But yes I can run it.

mloubout:
> @mloubout : I didn't get your comment about "AT only writes 0.00 GFlops/s"?
> You talking about a bug ? If so, how can I reproduce it ?

When I run your branch, rofiling works fine but auto_tuning prints out the time, OI, MCELLS properly but allways 0.00 GFLops
Just run acoustic_example and you should see it (with openmp)

mloubout:
```
    run(dimensions=(512,512,512), spacing=(10,10,10), tn=4000, full_run=True, auto_tuning=True, space_order=4, time_order=2)
  File "time_rtm.py", line 94, in run
    g = Acoustic.Gradient(rec, u, cache_blocking=None)
  File "/scratch/slim/shared/OPESCI/devito/examples/acoustic/Acoustic_codegen.py", line 105, in Gradient
    self.data.reinterpolateD(rec, self.dt_out, self.dt),
  File "/scratch/slim/shared/OPESCI/devito/devito/operator.py", line 168, in apply
    :param profile: Flag to enable performance profiling
  File "/scratch/slim/shared/OPESCI/devito/devito/propagator.py", line 139, in run
    """Calculates MCELLS
  File "/scratch/slim/shared/OPESCI/devito/devito/propagator.py", line 407, in log_performance
    Therefore reverse the list of dimensions, making the last variable in
  File "/scratch/slim/shared/OPESCI/devito/devito/propagator.py", line 216, in oi
    def gflops(self):
ZeroDivisionError: float division by zero
```
This sometimes happen too

FabioLuporini:
> When I run your branch, rofiling works fine but auto_tuning prints out the time, OI, MCELLS properly but allways 0.00 GFLops
Just run acoustic_example and you should see it (with openmp)

This should have now been fixed

FabioLuporini:
The division by zero error should now be fixed too

FabioLuporini:
@mlange05 @navjotk any objections ? if @mloubout tomorrow morning is happy, I'd like to merge this.

tests/test_oi.py
```diff
@@ -10,7 +10,7 @@ def test_oi_1(self):
         '''OI = (ADD+MUL)/[(LOAD+STORE)*word_size]; word_size=8(double),4(float)
         Equation = v1[i2][i1] = 3*v2[i2][i1] + 2*v3[i2][i1];
         '''
-        load = 3.0
+        load = 2.0
```
mlange05: Since we have multiple ways of counting loads, should the OI tests be extended to cover them all?

devito/propagator.py
mlange05: I'm inclined to side with @mloubout on this one, since the _algorithmic limit_ (`ideal_with_stores`) is probably more interesting to users who are concerned about the equation and discretisation orders, while `realistic` is probably more interesting from a performance optimisation/debug perspective to figure out how to get close to the theoretical limit. Since Devito promises (at least in theory) _automated optimisation_ the default external behaviour should probably be the algorithmic limit, while benchmarking and performance analysis always have to consider both. The whole thing sounds like an ongoing discussion though.

In any case, the OI choice should probably be exposed in the benchmark scripts, but this can be done in a follow-up PR.

mloubout:
Looks good to me. Good to go.

devito/propagator.py
FabioLuporini: cool. Mathias, we can talk about how the docstrings can be improved and how to expose the OI knob to the users after this PR.

FabioLuporini:
OK. Merging...

PR #137, opened 2016-10-05
Space dims fix

navjotk:
This PR adds a test that exposes the bug related to setting the `space_dims` parameter in the Propagator. 
It also includes a fix for the bug.

### DIFF
--- a/tests/test_propagator.py
+++ b/tests/test_propagator.py
@@ -1,6 +1,7 @@
 import numpy as np
-from devito import x, y, z
+
 import devito.cgen_wrapper as cgen
+from devito import x, y, z
 from devito.propagator import Propagator
 
 
--- a/tests/test_propagator.py
+++ b/tests/test_propagator.py
@@ -1,5 +1,5 @@
 import numpy as np
-from sympy.abc import x, y, z
+from devito import x, y, z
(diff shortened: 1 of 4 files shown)


tests/test_propagator.py
```diff
@@ -44,6 +44,16 @@ def test_4d(self):
         f(data, arr)
         assert(arr[4][3][2][1] == 122)
 
+    def test_space_dims_2d(self):
+        space_dims = (z, x)
+        propagator = Propagator("process", 1, (4, 3), [], space_dims=space_dims)
```
mlange05: This looks like it only tests prescribed `space_dims`. Should we also test the defaulting behaviour?

navjotk:
The test failing now is the due to the bug that is exposed in the code that guesses the dimensions at  https://github.com/opesci/devito/blob/master/devito/operator.py#L211

mlange05:
Ok, I fixed the dimension guess bug by not using sets and thus enforcing dimension index ordering. If I can get some confirmation that the fix looks ok, I think we can merge this.

FabioLuporini:
Looks reasonable

mlange05:
Great, thanks. Merging...

PR #138, opened 2016-10-06
4th order time steping

mloubout:
This PR does two thing 
1. FINITE DIFFERENCES
   a) The finite differences shortcuts in `interfaces` have been made consistent. It turns out `as_finite_diff` and `finite_difference_weights` return the coefficient with a different sign which was breaking a lot of adjoint assumptions. Center second order ones are correct so untouched, but I can switch it to `second_derivative` instead of `as_finit_diff` for coherency if you want.
   b) Higher order finite-difference shortcut for the biharmonbic operator have been added
   c) TTI is finally fully stable, non-dispersive and produces coherent results
2. 4th order in time
   The 4th order in time implementation of acoustic time-steping has been added.
   It passes the adjoint test, but the currently the gradient (no idea why)
3. TO DO
   a) Comments and docstring (I'm on it)
   b) figure out why `test_cse` now fails.
   c) I will rebase once @FabioLuporini PR is in EDIT : done


FabioLuporini:
Great stuff, sounds like a lot of new functionalities are being put in. 

I've started adding some comments, but then I stopped because the ``diff`` is too verbose. Rebase this on the new master, then we can take another look. Thanks

mloubout:
As I said, this need rebase after your PR, and docstring fix. It will be a lot easier to read after that as oly things in examples are changed.

devito/operator.py
```diff
@@ -133,7 +134,9 @@ def expr_cse(expr):
     new_stencils = []
 
     for stencil in stencils:
-        new_stencils.append(stencil)
+        new_stencils.append(Eq(stencil.lhs,
+                               collect(stencil.rhs,
+                                       [temp.lhs for temp in to_keep])))
```
mloubout: This is the factorization. Mainly once the `temp` are found by cse, I force the stencils to be factorized by it as it chould be some sort of `common factors`. It seems to work quite well for acoustic wich end up doing a radial factorization (distance to (x,y,z)).

examples/acoustic/fwi_operators.py
```diff
@@ -11,27 +11,44 @@ class ForwardOperator(Operator):
     def __init__(self, model, src, damp, data, time_order=2, spc_order=6,
                  save=False, **kwargs):
         nrec, nt = data.shape
-        dt = model.get_critical_dt()
+        s, h = symbols('s h')
         u = TimeData(name="u", shape=model.get_shape_comp(), time_dim=nt,
-                     time_order=time_order, space_order=spc_order, save=save,
+                     time_order=2, space_order=spc_order, save=save,
```
mloubout: I wouldn't yet, but yes you can. 4th order in time for acoustic is a bit peculiar and only needs second order time derivatives.

examples/acoustic/fwi_operators.py
```diff
@@ -49,13 +66,36 @@ def __init__(self, model, src, damp, data, time_order=2, spc_order=6,
...
+            dt = 1.73 * model.get_critical_dt()
+
+        # Create the stencil by hand instead of calling numpy solve for speed purposes
+        # Simple linear solve of a v(t+dt) + b u(t) + c v(t-dt) = L for v(t-dt)
+        stencil = 1.0 / (2.0 * m + s * damp) * \
+            (4.0 * m * v + (s * damp - 2.0 * m) *
+             v.forward + 2.0 * s ** 2 * (Lap + s ** 2 / 12.0 * Lap2))
+
+        # Add substitutions for spacing (temporal and spatial)
+        subs = {s: dt, h: model.get_spacing()}
```
mloubout: This is to avoid `solve` that is very slow for complex stencil. Im practice it only needs to solve 
```
a u(t+1) + b u(t) + c u(t-1) + L u(t) =0
```
but solve goes inside `L` for every term looking for simplification.

examples/acoustic/fwi_operators.py
```diff
@@ -93,36 +127,56 @@ def __init__(self, model, damp, data, recin, time_order=2, spc_order=6, **kwargs
...
+        else:
+            Lap = v.laplace
+            Lap2 = 1 / m * v.laplace2
+            Lap2u = - 1 / m**2 * u.laplace2
+            # PDE for information
+            # eqn = m * v.dt2 - Lap - s**2 / 12 * Lap2 + damp * v.dt
+            dt = 1.73 * model.get_critical_dt()
+            gradient_update = Eq(grad, grad -
+                                 (u.dt2 -
+                                  s ** 2 / 12.0 * Lap2u) * v.forward)
```
mloubout: This would indeed clean up everything. But the point of this PR was not a cleanup PR but only to add the 4th order functionallity we have bneen missing for month. I can try to look at it but it would require a major combined rewrite of `Acoustic_codegen` and `fwi_operators` all together.

devito/operator.py
FabioLuporini: OK, Two things:
1) this is not the right place to do these sort of things. CSE (semantic of the function) has nothing to do with factorization -- they are completely different transformations
2) as you're aware, i'm working on this symbolic engine to handle factorization automatically "behind the scenes", which will cope well with acoustic, tti and potentially many other problems. I'd encourage to remove this bit (perhaps you can cherry-pick the commit onto a temporary branch) and simply wait for this functionality, which will land with my next PR, hopefully by the end of the week

examples/acoustic/fwi_operators.py
FabioLuporini: Can you simply leave time_order, but adding the default value 2 in init ?

examples/acoustic/fwi_operators.py
FabioLuporini: I think that'd be very useful. I realise the importance of application code and features (I really do!), but writing good quality code is now as important

mloubout: I would still prefer to merge it first and then do a full rewrite PR. We all are gonna be presenting this in the next few weeks and it would be better to have this in as people we look into it.

examples/acoustic/fwi_operators.py
mloubout: This one is tricky, as the operator is called with `time_order=2` while there is no 4th order time derivatives. I'll try to think of something

examples/acoustic/fwi_operators.py
FabioLuporini: I then think we need a referee for this :) @mlange05 thoughts?

devito/operator.py
```diff
@@ -114,12 +114,13 @@ def __init__(self, nt, shape, dtype=np.float32, stencils=[],
             factorized[name] = dse_indexify(value)
 
         # Apply user-defined subs to stencil
-        self.stencils = [eqn.subs(subs[0]) for eqn in self.stencils]
+        # self.stencils = [eqn.subs(subs[0]) for eqn in self.stencils]
```
FabioLuporini: Can we drop this line?

devito/propagator.py
```diff
@@ -55,6 +55,7 @@ class Propagator(object):
                            number used for all dimensions except inner most or a list
                            explicitly stating block sizes for each dimension
                            Set cache_blocking to None to skip blocking on that dim
+                           Set cache_blocking to 0 to use best guess based on architecture
```
FabioLuporini: We've dropped this capability (because it was broken), so you can also drop this line

examples/acoustic/acoustic_example.py
```diff
@@ -24,33 +24,34 @@ def source(t, f0):
...
 
     # Define seismic data.
     data = IShot()
 
     f0 = .010
-    dt = model.get_critical_dt()
+    if time_order == 4:
+        dt = 1.73 * model.get_critical_dt()
+    else:
+        dt = model.get_critical_dt()
```
FabioLuporini: duplicate code ? I've seen this above

examples/containers.py
```diff
@@ -89,6 +116,7 @@ def get_shape_comp(self):
 
 
 class ISource:
+    """Source class, currently not implemented"""
```
FabioLuporini: You maybe can then raise a ``NotImplemented`` exception ?

examples/acoustic/acoustic_example.py
mloubout: And as I said, this is usage dependent. Can run 2nd order forward then 4th order gradient. This can't be generalized.

FabioLuporini: Ah. Okay then for now, but we might rethink this together in the future

mloubout:
This should be good to go

FabioLuporini:
This can probably go in, although I still hope we can find some time together to work out how to factorise the redundant code (even after merging).

I'd also like comments from both @navjotk and @mlange05 before merging

tests/test_symbolic_data.py
```diff
@@ -58,7 +58,7 @@ def test_derivatives_space(derivative, dimension, order):
     # Establish native sympy derivative expression
     width = int(order / 2)
     indices = [(dimension + i * h) for i in range(-width, width + 1)]
-    s_expr = as_finite_diff(u.diff(dimension), indices)
+    s_expr = -as_finite_diff(u.diff(dimension), indices)
```
mlange05: Why is this needed all of a sudden?

mlange05:
Ok, generally I think this is great and good to go (I'd like to know about the sign flip in the derivative test though).

However, I also do share @FabioLuporini's concerns about the the need for an internal re-structuring to avoid the special-casing and code repetition, but I think that poses a separate problem that we should address in a separate PR and the new SLIM repo.

tests/test_symbolic_data.py
mloubout: For some reason I haven't found yet (as said in the original message) `finite_diff_weights` and `as_finite_diff` return the opposute sign.

mloubout:
Can this be merged?

tests/test_symbolic_data.py
mlange05: Uhh, but should we not change the implementation then to correct this, instead of adjusting the test?

mlange05:
Cool, looks good to go, so merging once travis is happy.

mloubout:
@mlange05 Travis fails because of time out for single threaded (looks like Travis is now limited to 1h runs)

FabioLuporini:
I think it's time to merge this. I hope, however, that we can have a discussion about how to avoid duplicated code ASAP. Merging...

FabioLuporini:
Merged.

PR #139, opened 2016-10-07
Clean up prior introduction of symbolic engine

FabioLuporini:
This PR factorizes all functions that inspect and manipulate sympy expressions in a new module called ``symbolics.py``. These functions are also renamed by using the prefix ``dse_`` (devito symbolic engine). 

I'm planning to add more functionalities to the symbolic engine in the following weeks

### DIFF
--- a/devito/operator.py
+++ b/devito/operator.py
@@ -1,146 +1,15 @@
 import numpy as np
-from sympy import (Add, Eq, Function, Indexed, IndexedBase, Symbol, cse,
-                   lambdify, preorder_traversal, solve, symbols)
-from sympy.utilities.iterables import numbered_symbols
+
+from sympy import Eq, solve
 
 from devito.compiler import get_compiler_from_env
 from devito.dimension import t, x, y, z
-from devito.interfaces import SymbolicData, TimeData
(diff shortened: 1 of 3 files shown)


devito/operator.py
```diff
@@ -1,148 +1,16 @@
...
-from sympy.utilities.iterables import numbered_symbols
+
+from sympy import Eq, solve
 
 from devito.compiler import get_compiler_from_env
 from devito.dimension import t, x, y, z
-from devito.interfaces import SymbolicData, TimeData
+from devito.interfaces import TimeData
 from devito.propagator import Propagator
+from devito.symbolics import *
```
mlange05: Should we try to keep imports explicit to keep the namespace clean?

FabioLuporini:
Now travis fixed

As for the imports: I can do that, but that'd mean importing 5 modules, which is > 80% of symbolics. I can do this tomorrow, when I'll also incorporate feedback from @navjotk and @mloubout

FabioLuporini:
OK, I fixed the import thing, Travis is happy, and there's not much more to say about this, so merging...

PR #142, opened 2016-10-14
Fix usage of symbolic data grids in cache blocking test

mlange05:
Previously the test would hijack `DenseData` objects by enforcing an
artificial time dimension `t` without the use of the dedicated `TimeData`
grid class. This effectively forces Devito to assume a 4-dimensional
space grid, which is currently not supported. This commit fixes this by
making the time dimension implicit through the use of `TimeData` objects.

As a side note, we might also want to separately test cache blocking 
without the time dimension on a `DenseData` grid, but that is left for
another PR.

### DIFF
--- a/tests/test_cache_blocking.py
+++ b/tests/test_cache_blocking.py
@@ -3,14 +3,14 @@ import pytest
 from sympy import Eq
 
 from devito.dimension import t, x, y, z
-from devito.interfaces import DenseData
+from devito.interfaces import TimeData
 from devito.operator import SimpleOperator
 
 
 class Test_Cache_Blocking(object):
 
     # Full range testing.   This syntax tests all possible permutations of parameters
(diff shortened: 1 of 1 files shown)


mlange05:
Two things:
a) "I think this fix is just as hackish as what I did. This just has the hack in a different location." - I disagree with that, since this change only touches the test and rectifies clear misuse of the the `DenseData` object without introducing the need for a new (and untested) feature (see comments on PR #135).

b) I completely agree, we should be able to get the correct _indexed_ object from a _symbolic_ object without having to prescribe the indices, and indeed we can via `output_grid.indexify()`. However, this exposes the next bug, which looks like some of the the default indices not being replaced correctly in the Propagator.

However, at this point I do not have enough time or patience to dive into thus any further, so I'm closing this PR. It has admirably demonstrated though, at least to my mind, that a _re-write-and-replace_ approach is the way to go "cleaning up the Propagator" (ie. `SteniclKernel` and issue #43).

Issue #144, opened 2016-10-19
Parallel test runs on travis


mlange05:
Should be fairly uncontroversial and needs to happen soon so that we catch OpenMP-related bugs before merging them.

mlange05:
Fixed in PR #145.

PR #145, opened 2016-10-19
Travis parallel testing

mlange05:
This merge adds a new build matrix to the travis setup that includes a thread-parallel GCC builder to address issue #144. It also seems to fix recent OSX builder instabilities by updating the OXC builder to xcode8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,5 @@
 sudo: false
 
-os:
-  - linux
-  - osx
-
 # language: python
 language: c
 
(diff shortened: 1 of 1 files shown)


FabioLuporini:
looks uncontroversial

mlange05:
Yes, but the last commit (OSX fix) doesn't actually work. So, I think for now we want to disable the OSX tests and file an issue until someone properly debugs and tests the intermittent TTI failures.

mlange05:
Ok, all green, so merging...

PR #150, opened 2016-10-25
Vectorise `vp**(-2)` via NumPy

mlange05:
Profiling the model startup on acoustic suggests that the expression `vp**(-2)` takes up large chunks of time, presumably because it does not use NumPy vectorisation. Replacing it with `1. / vp * vp` seems to be much faster. 

The change in `IGrid.padm()` gives approximately 2x speedup for the initialisation of `m` in `fwi_operators.py:ForwardOperator` and the change in `acoustic_example.py:run()` reduces it's impact from 30% to around 5% in a single-timestep run on a 256^3 grid.

### DIFF
--- a/examples/acoustic/acoustic_example.py
+++ b/examples/acoustic/acoustic_example.py
@@ -42,7 +42,7 @@ def run(dimensions=(150, 150, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
     # Smooth velocity
     initial_vp = smooth10(true_vp, dimensions)
 
-    dm = true_vp**-2 - initial_vp**-2
+    dm = 1. / (true_vp * true_vp) - 1. / (initial_vp * initial_vp)
 
     model.create_model(origin, spacing, true_vp)
 
--- a/examples/containers.py
+++ b/examples/containers.py
@@ -71,7 +71,7 @@ class IGrid:
         return self.origin
 
     def padm(self):
-        return self.pad(self.vp**(-2))
+        return self.pad(1./(self.vp * self.vp))
 
     def pad(self, m):
         pad_list = []


examples/acoustic/acoustic_example.py
```diff
@@ -42,7 +42,7 @@ def run(dimensions=(150, 150, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
     # Smooth velocity
     initial_vp = smooth10(true_vp, dimensions)
 
-    dm = true_vp**-2 - initial_vp**-2
+    dm = 1. / true_vp * true_vp - 1. / initial_vp * initial_vp
```
mloubout: This is always 0 without parenthesis on the denominator.

mlange05: Totally true. Sorry, should be fixed now. Interestingly, all tests pass despite this mistake...

mloubout:
Good to go for me

examples/acoustic/acoustic_example.py
mlange05: That is actually very bad(!), since I can now "optimise" the model and completely break it without knowing. It's also relevant to this PR though, so just a point to ponder...

examples/acoustic/acoustic_example.py
```diff
@@ -42,7 +42,7 @@ def run(dimensions=(150, 150, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
     # Smooth velocity
     initial_vp = smooth10(true_vp, dimensions)
 
-    dm = true_vp**-2 - initial_vp**-2
```
FabioLuporini: If true_vp isn't changed in place by smooth10, you can compute the value before (once and for all), instead of doing it on very grid point

examples/containers.py
```diff
@@ -71,7 +71,7 @@ def get_origin(self):
         return self.origin
 
     def padm(self):
-        return self.pad(self.vp**(-2))
```
FabioLuporini: I wonder if this is true as well

examples/acoustic/acoustic_example.py
mlange05: Ok two points here:
a) This is a general design question that warrants a separate github issue. I can see points for and against making Devito-internal allocation optional.
b) For the benchmarking examples this should be fixed regardless of that discussion, since the startup time can limit our benchmarking with _really big_ grids. The fix would effectively involve deferring the initialisation of `m` by separating the code that generates the `vp` and passing it into the `Forward` operator initialisation routines.

FabioLuporini: OK, go then

mlange05:
Cool, merging...

PR #153, opened 2016-11-04
Make benchmark.py usable even w/o plotter

FabioLuporini:


mlange05:
I'm assuming this is to get around the fact that `matplotlib` isn't available on benchmark machines? In that case I would prefer to fix the import in `opescibench` with `try`/`except` constructs, rather than putting a workaround here.

FabioLuporini:
Well, in principle it's because of that, not sure if there's any other reasons this could fail?

FabioLuporini:
@mlange05 I'm closing this without merging, then.

FabioLuporini:
@mlange05 : could you fix the issue in opescibench then? I'm not sure how I would handle things in case of an Import error

The alternative is that we resurrect this PR and add matplotlib as compulsory requirement of opescibench

mlange05:
Ok, I just pushed a tentative fix for this into `opscibench-master`. Please check this and let me know if it works or not.

PR #152, opened 2016-11-02
Introduce expression rewriter

FabioLuporini:
** Premise ** I'm filing this PR so that people can start looking into it, *but* it still lacks a comprehensive performance evaluation. This will hopefully be integrated soon.

This PR extends the symbolic engine in Devito. Besides a cleaner interface, the most important addition is the ``factorizer``, a rewrite operator which gets applied right after CSE. Basically, using the ``collect_nested`` function (a contribution of this PR, since sympy's built-ins just suck for our purposes), constants and common temporaries are sought in each expression composing the stencil (ie the stencils themselves plus any sub-expression captured by CSE) and factorized.

These are the flop reductions I got.
Problem: acoustic; each line is a different space order ([2->16, step=2]).
```
Rewriter: 28 --> 23 flops (Gain: 1.22 X)
Rewriter: 40 --> 30 flops (Gain: 1.33 X)
Rewriter: 52 --> 37 flops (Gain: 1.41 X)
Rewriter: 64 --> 44 flops (Gain: 1.45 X)
Rewriter: 76 --> 51 flops (Gain: 1.49 X)
Rewriter: 88 --> 58 flops (Gain: 1.52 X)
Rewriter: 100 --> 65 flops (Gain: 1.54 X)
Rewriter: 112 --> 72 flops (Gain: 1.56 X)
```

Problem: TTI; each line is a different space order ([2->6, step=2]).
```
Rewriter: 389 --> 327 flops (Gain: 1.19 X)
Rewriter: 935 --> 839 flops (Gain: 1.11 X)
Rewriter: 1783 --> 1667 flops (Gain: 1.07 X)
``` 

The gain dropping with TTI is due to CSE, which "hides" some factorisation opportunities. A way to work around this would consist of searching "across" different expressions and somehow rearranging the temporaries. This requires further thinking and some legwork, so it's not gonna be shipped with this PR.

Some new tests are introduced.

TODO:
- Run experiments to collect runtimes and understand how the roofline plots change

PR #154, opened 2016-11-04
From cse to dse

FabioLuporini:
This PR extends the symbolic engine in Devito. Besides a cleaner interface, the most important addition is the factorizer, a rewrite operator which gets applied right after CSE. Basically, using the collect_nested function (a contribution of this PR, since sympy's built-ins just suck for our purposes), constants and common temporaries are sought in each expression composing the stencil (ie the stencils themselves plus any sub-expression captured by CSE) and factorized.

These are the flop reductions I got.
Problem: acoustic; each line is a different space order ([2->16, step=2]).

```
Rewriter: 28 --> 23 flops (Gain: 1.22 X)
Rewriter: 40 --> 30 flops (Gain: 1.33 X)
Rewriter: 52 --> 37 flops (Gain: 1.41 X)
Rewriter: 64 --> 44 flops (Gain: 1.45 X)
Rewriter: 76 --> 51 flops (Gain: 1.49 X)
Rewriter: 88 --> 58 flops (Gain: 1.52 X)
Rewriter: 100 --> 65 flops (Gain: 1.54 X)
Rewriter: 112 --> 72 flops (Gain: 1.56 X)
```

Problem: TTI; each line is a different space order ([2->6, step=2]).
```
Rewriter: 389 --> 327 flops (Gain: 1.19 X)
Rewriter: 935 --> 839 flops (Gain: 1.11 X)
Rewriter: 1783 --> 1667 flops (Gain: 1.07 X)
```

The gain dropping with TTI is due to CSE, which "hides" some factorisation opportunities. A way to work around this would consist of searching "across" different expressions and somehow rearranging the temporaries. This requires further thinking and some legwork, so it's not gonna be shipped with this PR.

Some new tests are introduced.

** EDIT ** : 
1) run experiments to collect runtimes and understand how the roofline plots change : DONE (see 
[acoustic_dim(1024,1024,1024)_so[2,4,6,8,10,12,14,16]_to[2]_arch[bdwb0ds].pdf](https://github.com/opesci/devito/files/608846/acoustic_dim.1024.1024.1024._so.2.4.6.8.10.12.14.16._to.2._arch.bdwb0ds.pdf))
2) drop the "DROP BEFORE MERGE" commits : DONE

** EDIT ** : instead of having two PRs #152 and #154, we now only have #154 which is based on the old #152. Further, I'm introducing a commit ensuring that the generated code actually reflects what produced by the DSE (Sympy was unpicking the DSE's factorisations right before code generation following a simple ``xreplace``...)


FabioLuporini:
@mlange05 I've updated this PR (see main message above)

performance results still WIP, but now I think I have the right infrastructure to start collecting them.

FabioLuporini:
Some flake8 fixes introduced. These hopefully will make the tests pass.

FabioLuporini:
Performance results on a dual socket broadwell
[acoustic_dim(1024,1024,1024)_so[2,4,6,8,10,12,14,16]_to[2].pdf](https://github.com/opesci/devito/files/607076/acoustic_dim.1024.1024.1024._so.2.4.6.8.10.12.14.16._to.2.pdf)

Also rebased on master. I'm now quite happy with this PR. 

Any more thoughts @mlange05 @navjotk @mloubout ?

FabioLuporini:
Pushed a tiny enhancement to benchmark.py

@mlange05 travis is now unhappy *probably* because some of your changes, as it fails to import opescibench

mlange05:
Yes, sorry, that was my fault. Fixed and rebased now.

devito/symbolics.py
```diff
@@ -7,15 +7,25 @@
...
 """
 
+from __future__ import absolute_import
+
+from collections import OrderedDict
+
 import numpy as np
-from sympy import (Add, Eq, Function, Indexed, IndexedBase, Symbol, cse,
-                   lambdify, numbered_symbols, preorder_traversal, symbols)
+import sympy
```
mlange05: Can we move whatever else we need from this to the bulk include below?

devito/symbolics.py
```diff
@@ -82,90 +92,206 @@ def dse_indexify(expr):
...
-                break
-
-    return to_keep + new_stencils
+    """
+    Transform expressions to reduce their operation count.
+    """
+
+    # Do more factorization sweeps if the expression operation count is
+    # greater than this threshold
+    FACTORIZER_THS = 15
```
mlange05: Just a general question: What's the role of this and does changing it make DSE more aggressive? If so, do we expose this at some point?

FabioLuporini: The factorisation engine is makes use of heuristics, so recursively doing more sweeps may result in a better operation count reduction. However, it might be the case that for insane stencils (dunno, say TTI order 14-16?), the symbolic processing time becomes too high if we apply factorisation recursively to all expressions performing more than ``FACTORIZER_THS`` flops (not only do the temporaries become individually more expensive, but also the total number of temporaries > ``FACTORIZER_THS`` increases!). So this is a knob that we might want to expose (somehow), in the future, to work around this problem. A long symbolic processing time wouldn't be a problem when we do long runs and we care about performance (these are supposed to last for many hours or days), although *it might be* a problem at development or debugging time.

I don't think there's an urgency to expose this parameter in this PR

devito/symbolics.py
FabioLuporini: nope, because they would conflict with ``class Mul`` and ``class Add`` at the bottom of this module (I'm extending the sympy hierarchy, as explained in the comments below)

devito/symbolics.py
mlange05: Cool, sounds good. Thanks.

devito/symbolics.py
mlange05: Ahh ok. Not a fan, but good enough for now.

mlange05:
Looks great, merging...

devito/symbolics.py
navjotk: How was this value chosen?

FabioLuporini: empirically

PR #155, opened 2016-11-05
Python3 compatibility and Travis housekeeping

mlange05:
This merge cleans up the travis build matrix to use the proper builtin Python images and test Python 3.3, as well 2.7. It also fixes the various compatibility issues with Python3.


FabioLuporini:
I think this can just go in

PR #156, opened 2016-11-21
More work towards a recursive code generator

mlange05:
Ok, this is a bit of a blanket PR for the next steps in the re-development of the low-level code generator (currently `StencilKernel`). It adds a set of features that allow us to switch the diffusion example over to the new API and includes in detail:

* Additional tests for the low-level "indexed" API in stencil kernels
* Compatibility with the high-level API through "indexification" in `Expression`
* Automatic derivation of loop bounds from the stencil indices
* Automatic derivation and substitution of loop variables in `Iteration`
* Global symbol substitution in stencil kernel
* `Dimension` objects define "buffered" loops, eg. over dimension "time"

Note: `StencilKernel` objects now behave equivalently to `Operator/Propagator` objects in terms of "baked-in" loop limits. However, these now need to be defined via the default `Dimension` objects (`x`, `y`, `t` or `p`). Generalisation for "open" loop limits will come on a follow-on PR.

### DIFF
--- a/devito/dimension.py
+++ b/devito/dimension.py
@@ -9,10 +9,17 @@ class Dimension(Symbol):
     def __new__(cls, name, **kwargs):
         newobj = Symbol.__new__(cls, name)
         newobj.size = kwargs.get('size', None)
+        newobj._count = 0
         return newobj
 
     def __str__(self):
         return self.name
+
+    def get_varname(self):
+        """Generates a new variables name based on an internal counter"""
+        name = "%s%d" % (self.name, self._count)
+        self._count += 1
+        return name
 
 
 # Set of default dimensions for space and time
--- a/devito/dimension.py
+++ b/devito/dimension.py
@@ -4,11 +4,18 @@ __all__ = ['Dimension', 'x', 'y', 'z', 't', 'p']
 
 
 class Dimension(Symbol):
-    """Index object that represents iteration spaces"""
+    """Index object that represents a problem dimension and thus
+    defines a potential iteration space.
+
+    :param size: Optional, size of the array dimension.
+    :param buffered: Optional, boolean flag indicating whether to
+                     buffer variables when iterating this dimension.
+    """
 
     def __new__(cls, name, **kwargs):
         newobj = Symbol.__new__(cls, name)
         newobj.size = kwargs.get('size', None)
+        newobj.buffered = kwargs.get('buffered', None)
         newobj._count = 0
         return newobj
 
--- a/devito/dimension.py
+++ b/devito/dimension.py
@@ -1,6 +1,6 @@
 from sympy import Symbol
 
-__all__ = ['Dimension', 'x', 'y', 'z', 't']
+__all__ = ['Dimension', 'x', 'y', 'z', 't', 'p']
 
 
 class Dimension(Symbol):
@@ -27,3 +27,4 @@ x = Dimension('x')
 y = Dimension('y')
 z = Dimension('z')
 t = Dimension('t')
+p = Dimension('p')
--- a/devito/expression.py
+++ b/devito/expression.py
@@ -1,9 +1,10 @@
 from __future__ import absolute_import
 
-import cgen
 from collections import defaultdict
 from operator import attrgetter
-from sympy import Eq, IndexedBase, Indexed, preorder_traversal
+
+import cgen
+from sympy import Eq, Indexed, IndexedBase, preorder_traversal
 
 from devito.codeprinter import ccode
 from devito.dimension import Dimension
--- a/devito/expression.py
+++ b/devito/expression.py
@@ -2,6 +2,7 @@ import cgen
 from sympy import Eq, IndexedBase, preorder_traversal
 
 from devito.codeprinter import ccode
+from devito.interfaces import SymbolicData
 from devito.symbolics import dse_indexify
 from devito.tools import filter_ordered
 
@@ -19,6 +20,9 @@ class Expression(object):
         self.functions = []
         # Traverse stencil to determine meta information
         for e in preorder_traversal(self.stencil):
+            if isinstance(e, SymbolicData):
+                self.dimensions += list(e.indices)
+                self.functions += [e]
             if isinstance(e, IndexedBase):
                 self.dimensions += list(e.function.indices)
                 self.functions += [e.function]
--- a/devito/expression.py
+++ b/devito/expression.py
@@ -2,6 +2,7 @@ import cgen
 from sympy import Eq, IndexedBase, preorder_traversal
 
 from devito.codeprinter import ccode
+from devito.symbolics import dse_indexify
 from devito.tools import filter_ordered
 
 __all__ = ['Expression']
@@ -47,3 +48,7 @@ class Expression(object):
         :returns: List of unique data objects required by the expression
         """
         return self.functions
+
+    def indexify(self):
+        """Convert stencil expression to "indexed" format"""
+        self.stencil = dse_indexify(self.stencil)
--- a/devito/expression.py
+++ b/devito/expression.py
@@ -1,7 +1,11 @@
+from __future__ import absolute_import
+
 import cgen
-from sympy import Eq, IndexedBase, preorder_traversal
+from collections import defaultdict
+from sympy import Eq, IndexedBase, Indexed, preorder_traversal
 
 from devito.codeprinter import ccode
+from devito.dimension import Dimension
 from devito.interfaces import SymbolicData
 from devito.symbolics import dse_indexify
 from devito.tools import filter_ordered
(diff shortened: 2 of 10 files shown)


devito/expression.py
```diff
@@ -18,6 +26,9 @@ def __init__(self, stencil):
         self.functions = []
         # Traverse stencil to determine meta information
         for e in preorder_traversal(self.stencil):
+            if isinstance(e, SymbolicData):
```
FabioLuporini: *Not* as part of this PR and I know this is me being picky:
as of sympy 1.0.0, all sympy objects have an ``is_Type`` property to find out what they are. Even Indexed and IndexedBase are added this property. Maybe we should extend our hierarchy to do the same? eg is_SymbolicData?

devito/expression.py
```diff
@@ -46,4 +57,29 @@ def signature(self):
...
+        """Convert stencil expression to "indexed" format"""
+        self.stencil = dse_indexify(self.stencil)
+
+    @property
+    def index_offsets(self):
+        """Collect all non-zero offsets used with each index in a map
+
+        Note: This assumes we have indexified the stencil expression."""
+        offsets = defaultdict(list)
+        for e in preorder_traversal(self.stencil):
```
FabioLuporini: I think you'll soon be able to use this: https://github.com/opesci/devito/blob/from-cse-to-dse/devito/symbolics.py#L339 to get Indexed thus avoiding a nest level (it needs generalisation though).
Also: for ``Indexed``, ``e.args[1:]`` is actually ``e.index`` or ``e.indices`` (I'm sure there's an attribute)

devito/iteration.py
```diff
@@ -72,3 +111,18 @@ def signature(self):
         """
         signatures = [e.signature for e in self.expressions]
         return filter_ordered(chain(*signatures))
+
+    def indexify(self):
+        """Convert all enclosed stencil expressions to "indexed" format"""
+        for e in self.expressions:
+            e.indexify()
```
FabioLuporini: Does this modify the expressions in place ? Maybe it's time to change this behaviour

devito/stencilkernel.py
```diff
@@ -29,11 +39,22 @@ def __init__(self, stencils, name="Kernel", compiler=None):
         stencils = stencils if isinstance(stencils, list) else [stencils]
         self.expressions = [Expression(s) for s in stencils]
 
+        # Lower all expressions to "indexed" API
+        for e in self.expressions:
+            e.indexify()
```
FabioLuporini: Same as before

devito/stencilkernel.py
```diff
@@ -29,11 +39,22 @@ def __init__(self, stencils, name="Kernel", compiler=None):
...
+
+        # Apply supplied substitutions
+        if subs is not None:
+            for expr in self.expressions:
+                expr.substitute(subs)
```
FabioLuporini: Same. I think we should always construct new objects

devito/expression.py
mlange05: Agreed, but after this PR.

FabioLuporini: sure, that's why I started with "Not" :)

devito/expression.py
mlange05: Neat. Let's land #154 first, so I can integrate this into this PR.

devito/dimension.py
```diff
@@ -1,22 +1,36 @@
...
+__all__ = ['Dimension', 'x', 'y', 'z', 't', 'p']
 
 
 class Dimension(Symbol):
-    """Index object that represents iteration spaces"""
+    """Index object that represents a problem dimension and thus
+    defines a potential iteration space.
+
+    :param size: Optional, size of the array dimension.
+    :param buffered: Optional, buffer variables when iterating this dimension.
```
navjotk: What's the type of this?

examples/diffusion/example_diffusion.py
```diff
@@ -120,15 +120,18 @@ def execute_devito(ui, spacing=0.01, a=0.5, timesteps=500):
     # Note: This should be made simpler through the use of defaults
     u = TimeData(name='u', shape=(nx, ny), time_order=1, space_order=2)
     u.data[0, :] = ui[:]
+    u.indices[0].size = timesteps
+    u.indices[1].size = nx
+    u.indices[2].size = ny
```
navjotk: How is the order between `t`, `x` and `y` decided here? Hardcoded?

tests/test_stencil_arithmetic.py
```diff
@@ -2,66 +2,129 @@
...
-    k = Dimension(name='k', size=5)
-    l = Dimension(name='l', size=6)
-
-    a = DenseData(name='a', dimensions=(i, j, k, l))
-    a.data[:] = 2.
-    ai = a.indexify()
-    b = DenseData(name='b', dimensions=(j, k))
-    b.data[:] = 3.
-    bi = b.indexify()
+    clear_cache()
```
navjotk: Is this required?

tests/test_stencil_arithmetic.py
```diff
@@ -2,66 +2,129 @@
...
+
+@pytest.mark.parametrize('expr, result', [
+    ('Eq(a[k, l], a[k - 1 , l] + 1.)',
+     np.meshgrid(np.arange(2., 8.), np.arange(2., 7.))[1]),
+    ('Eq(a[k, l], a[k, l - 1] + 1.)',
+     np.meshgrid(np.arange(2., 8.), np.arange(2., 7.))[0]),
+])
+def test_arithmetic_indexed_increment(i, j, k, l, expr, result):
+    """Tests point-wise increments with stencil offsets in one dimension"""
+    clear_cache()
```
navjotk: same as above

tests/test_stencil_arithmetic.py
```diff
@@ -2,66 +2,129 @@
...
+@pytest.mark.parametrize('expr, result', [
+    ('Eq(a[k, l], b[k - 1 , l] + 1.)', np.zeros((5, 6)) + 3.),
+    ('Eq(a[k, l], b[k , l - 1] + 1.)', np.zeros((5, 6)) + 3.),
+    ('Eq(a[k, l], b[k - 1, l - 1] + 1.)', np.zeros((5, 6)) + 3.),
+    ('Eq(a[k, l], b[k + 1, l + 1] + 1.)', np.zeros((5, 6)) + 3.),
+])
+def test_arithmetic_indexed_stencil(i, j, k, l, expr, result):
+    """Test point-wise arithmetic with stencil offsets across two
+    functions in indexed expression format"""
+    clear_cache()
```
navjotk: same

devito/expression.py
FabioLuporini: Now u can do it :)

tests/test_stencil_arithmetic.py
FabioLuporini: and if so, why

examples/diffusion/example_diffusion.py
```diff
@@ -120,15 +120,18 @@ def execute_devito(ui, spacing=0.01, a=0.5, timesteps=500):
...
     eqn = Eq(u.dt, a * (u.dx2 + u.dy2))
     stencil = solve(eqn, u.forward)[0]
-    op = Operator(stencils=Eq(u.forward, stencil), subs={h: spacing, s: dt},
-                  nt=timesteps, shape=(nx, ny), spc_border=1, time_order=1)
+    op = StencilKernel(Eq(u.forward, stencil), subs={h: spacing, s: dt})
+
     # Execute the generated Devito stencil operator
     tstart = time.time()
-    op.apply()
+    op.apply(u)
```
mlange05: This is diffusion, so there are no `m` or `damp` terms. The core change here though is that the kernel accepts all numpy arrays that have the same dimension as `u.data` and is not bound to the symbol itself anymore.

devito/stencilkernel.py
FabioLuporini: Fair enough, but do you actually mean ``2n`` instead of ``n^2`` ? Shouldn't reconstructing an object simply be O(number of nodes)?

FabioLuporini: As a paraphrase of this discussion, only to see if I get it right: am I correct if I say that once you pass an object to, say, an ``Iteration``, you basically lose it because it gets transformed in place?

FabioLuporini: Sorry, I completely misunderstood what's going on here. That's OK, forget about my comments (apart from the 2n vs n^2 which you may clarify?)

devito/iteration.py
FabioLuporini: sry, forget about this

devito/stencilkernel.py
FabioLuporini: sry, forget about this

devito/stencilkernel.py
mlange05: Hmm, not sure about the `2n` vs. `n^2` to be honest. We could probably write it in `2n` if we make sure not to throw any partial results away, but I see no immediate need for this at this stage.

tests/test_stencil_arithmetic.py
mlange05: Ok, this originally fixed a suspected caching bug, but after several re-writes it now seems fine without the fix. Good catch.

mlange05:
Ok, everyone happy, so merging...

PR #157, opened 2016-11-21
More compiler presets

mlange05:
This merge adds version-specific compiler presets for the GNU toolchain `gcc-4.9` and `gcc-5` and switches travis to use both. It also adds a `CustomCompiler` class which reads the `CC`, `CFLAGS`, `LD` and `LDFLAGS` flags if `DEVITO_ARCH` is not set.

EDIT: This merge now also switches back to C99 compilation during JIT.

### DIFF
--- a/devito/compiler.py
+++ b/devito/compiler.py
@@ -165,9 +165,9 @@ class CustomCompiler(Compiler):
 
     def __init__(self, *args, **kwargs):
         super(CustomCompiler, self).__init__(*args, **kwargs)
-        self.cc = environ.get('CC', 'g++')
+        self.cc = environ.get('CXX', 'g++')
         self.ld = environ.get('LD', 'g++')
-        self.cflags = environ.get('CFLAGS', '-O3 -g -fPIC -Wall').split(' ')
+        self.cflags = environ.get('CXXFLAGS', '-O3 -g -fPIC -Wall').split(' ')
         self.ldflags = environ.get('LDFLAGS', '-shared').split(' ')
 
 
--- a/devito/compiler.py
+++ b/devito/compiler.py
@@ -36,6 +36,8 @@ class Compiler(GCCToolchain):
         * :data:`self.libraries`
         * :data:`self.library_dirs`
         * :data:`self.defines`
+        * :data:`self.src_ext`
+        * :data:`self.lib_ext`
         * :data:`self.undefines`
         * :data:`self.pragma_ivdep`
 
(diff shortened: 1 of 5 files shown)


devito/compiler.py
```diff
@@ -152,12 +154,33 @@ def __init__(self, *args, **kwargs):
...
 # environment variable DEVITO_ARCH. Developers should add
 # new compiler classes here and provide a description in
 # the docstring of get_compiler_from_env().
 compiler_registry = {
     'gcc': GNUCompiler, 'gnu': GNUCompiler,
+    'gcc-4.9': partial(GNUCompiler, version='4.9'),
+    'g++-4.9': partial(GNUCompiler, version='4.9'),
+    'gcc-5': partial(GNUCompiler, version='5'),
+    'g++-5': partial(GNUCompiler, version='5'),
     'clang': ClangCompiler, 'osx': ClangCompiler,
```
mlange05: Sure, but that's because your system has non-standard names. That's really what the `CustomCompiler` is for.

navjotk: I really don't think there are any standard names here.
Anything beyond g++ and gcc should go into CC and CXX.

devito/compiler.py
```diff
@@ -152,12 +154,33 @@ def __init__(self, *args, **kwargs):
...
+
+    :param openmp: Boolean indicating if openmp is enabled. False by default
+
+    Note: Currently honours CC, CFLAGS, LD and LDFLAGS, with defaults similar
+    to the default GNU settings.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super(CustomCompiler, self).__init__(*args, **kwargs)
+        self.cc = environ.get('CC', 'g++')
```
ggorman: By convention CC and CFLAGS are used for the C compiler, while CXX and CXXFLAGS are used for C++.

mlange05: True. Should we still be using C++ though? Our kernels are trivial C and we have stopped linking with complex reader routines ages ago.

mloubout:
Seems to work fine with Intel compiler too.

FabioLuporini:
OK

mloubout:
Can we use this PR to have the libraries extensions fixed?  Currently it is `lib.so` for everyhting on every system. Shouldn't we use the system extension instead (makes also interfacing with other languages easier)

mlange05:
Good point about toolchain-specific lib extensions. I've made the file extensions generic and added the OSX variant (`.dylib`) to our preset, but for Windows compatibiltiy (`.dll'`) we'd need to add an entirely new compiler preset, so I leave that for another PR. Similarly, the `sys.platform` auto-detection requires further stuff to correctly identify the right compiler toolchain, which again is non-trivial and therefore left for a separate PR.

Following this change, can someone with an OSX system please verify that this still works?

mloubout:
No problem for the Windows part, ust having OSX and Linux proper extension is good enough for my point of view. I'll test the OSX part.  FIx looks good.

FabioLuporini:
this looks mergeable

mlange05:
Agreed, merging...

PR #159, opened 2016-11-24
Pip install, license and housekeeping

mlange05:
This merge adds a license file (MIT) and a `setup.py` to allow Devito to be installed via pip, as well as other minor housekeeping changes.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2016, Imperial College, London
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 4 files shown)


devito/cgen_wrapper.py
```diff
@@ -1,6 +1,6 @@
 import ctypes
 
-from cgen import *
+from codepy.cgen import *
```
navjotk: Are you sure you are not moving backward here? `codepy.cgen` is deprecated as seen [here](https://github.com/inducer/codepy/blob/master/codepy/cgen/__init__.py)

mlange05: Good point. Fixed now.

FabioLuporini:
OK

PR #161, opened 2016-12-01
Tti fix again

mloubout:
Trying to fix the TTI code.

- cosine and sine have now an extra 'Taylor' mode
- `cos` and `sin` are replaced by `cosf` and `sinf` in float32
- Initialization of `Igrid()` fixed to include azimuth (`phi`)

- Still random segmentaion faults for some spacial orders


devito/finite_difference.py
```diff
@@ -163,14 +163,13 @@ def first_derivative(*args, **kwargs):
     # Stencil positions for non-symmetric cross-derivatives with symmetric averaging
     if side == right:
         ind = [(dim + i * diff) for i in range(-int(order / 2) + 1 - (order % 2),
-                                               int((order + 1) / 2) + 2 - (order % 2))]
+                                               int(order / 2) + 2)]
```
FabioLuporini: Just a simplification?

devito/interfaces.py
```diff
@@ -159,6 +159,12 @@ def _indices(cls, **kwargs):
                 dimensions = [symbols("x%d" % i) for i in range(1, len(shape) + 1)]
         return dimensions
 
+    def _cached_init(self):
```
FabioLuporini: Add comment along the lines of "# FIXME : ..."

examples/tti/tti_example.py
```diff
@@ -23,10 +23,11 @@ def setup(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
 
     model = IGrid(origin, spacing,
                   true_vp,
-                  .4*np.ones(dimensions),
-                  -.1*np.ones(dimensions),
-                  -np.pi/7*np.ones(dimensions),
-                  np.pi/5*np.ones(dimensions))
+                  rho=None,
```
FabioLuporini: Remind me pls why ``rho`` is None?

examples/tti/tti_operators.py
```diff
@@ -21,58 +21,64 @@ class ForwardOperator(Operator):
     `Bhaskara` uses a rational approximation.
     """
     def __init__(self, model, src, damp, data, time_order=2, spc_order=4, save=False,
-                 trigonometry='normal', **kwargs):
+                 trigonometry='normal', u_ini=None, **kwargs):
```
FabioLuporini: what's ``u_ini``?

examples/tti/tti_operators.py
```diff
@@ -84,65 +90,80 @@ def __init__(self, model, src, damp, data, time_order=2, spc_order=4, save=False
...
-                                     ang2, dim=x, side=centered, order=spc_brd) +
-                   first_derivative(Gxp, ang0,
-                                    ang3, dim=y, side=left, order=spc_brd) -
-                   first_derivative(Gxp, ang1, dim=z, side=left, order=spc_brd))
-            Gzz = (-first_derivative(Gzr, ang1,
-                                     ang2, dim=x, side=centered, order=spc_brd) +
-                   first_derivative(Gzr, ang1,
-                                    ang3, dim=y, side=left, order=spc_brd) +
-                   first_derivative(Gzr, ang0, dim=z, side=left, order=spc_brd))
+            Gxx = (-first_derivative(Gxp * ang0 * ang2,
```
FabioLuporini: I guess this modified hunk is just reformatting?

examples/tti/tti_operators.py
```diff
@@ -84,65 +90,80 @@ def __init__(self, model, src, damp, data, time_order=2, spc_order=4, save=False
...
-                                     ang2, dim=x, side=left, order=spc_brd) -
-                    first_derivative(Gxp2, ang0,
-                                     ang3, dim=y, side=centered, order=spc_brd) +
-                    first_derivative(Gxp2, ang1, dim=z, side=centered, order=spc_brd))
-            Gzz2 = (first_derivative(Gzr2, ang1,
-                                     ang2, dim=x, side=left, order=spc_brd) -
-                    first_derivative(Gzr2, ang1,
-                                     ang3, dim=y, side=centered, order=spc_brd) -
-                    first_derivative(Gzr2, ang0, dim=z, side=centered, order=spc_brd))
+            Gxx2 = (first_derivative(Gxp2 * ang0 * ang2,
```
FabioLuporini: Same as before?

tests/test_tti.py
```diff
@@ -0,0 +1,118 @@
...
+        # Actual adjoint test
+        log("Difference between acoustic and TTI with all coefficients to 0 %f" % res)
+        assert np.isclose(res, 0.0, atol=1e-1)
+
+
+if __name__ == "__main__":
+    t = TestTTI()
+    request = type('', (), {})()
+    request.param = (120, 140, 150)
+    prop = t.tti(request, 4)
```
FabioLuporini: I'm happy we have this new text. How long does it take roughly to execute?

FabioLuporini:
Have you checked that the trigonometric approximations work?

FabioLuporini:
Do you recommend to use this "TTI version" in ``test_rewriter.py`` too ?

examples/tti/tti_example.py
mloubout: Because it's not implemented yet (here) but is on other branches.

examples/tti/tti_operators.py
mloubout: initialize the wavefield to non-zero. Only used for `test_tti` so that I can do only a few time step from an already propagated wavefield.

examples/tti/tti_operators.py
mloubout: same

tests/test_tti.py
mloubout: about  3 ,minutes for 3D order 8.

examples/tti/tti_operators.py
FabioLuporini: OK

mloubout:
>Have you checked that the trigonometric approximations work?
I have tested the approximation an the error should be less than 1% for any appproximation and angle which is more than sufficient.

examples/tti/tti_example.py
FabioLuporini: just a curiosity: I guess ``rho`` isn't here yet because the system is still unstable with the current operators?

mloubout: No, it's in my acoustic, but for tti, there is no point adding more paramterers untill we get the "simple" case `rho=1` running in a decent time. It will be as astable a s the current one just a bit more costly computationally.

devito/interfaces.py
mloubout: Not sure what this does and what to put in the docstring

mlange05: Ok, so as mentioned in the chat, this fix will most definitely cause memory bleeds due to SymPy's caching causing us the invalidate the "original" symbol instance as soon as we form a derivative. This means we will never deallocate any memory, so this needs to be pulled.

However, I would argue that fixing the caching bug should not be part of this PR and instead we can simply `xfail` the new TTI test, which will hopefully trigger on this in travis.

devito/propagator.py
```diff
@@ -425,7 +425,9 @@ def sympy_to_cgen(self, stencils):
                 sub = str(ccode(self.time_substitutions(expr).xreplace(self._mapper)))
                 if self.dtype is np.float32:
                     factors.append(cgen.Assign(name, (sub.replace("pow", "powf")
-                                                      .replace("fabs", "fabsf"))))
+                                                      .replace("fabs", "fabsf")
+                                                      .replace("cos", "cosf")
+                                                      .replace("sin", "sinf"))))
```
mlange05: Not to be fixed as part of this PR, but these replacements should really live in the `CodePrinter`.

examples/acoustic/Acoustic_codegen.py
```diff
@@ -29,16 +29,13 @@ def __init__(self, model, data, source, nbpml=40, t_order=2, s_order=2,
         self.model.nbpml = nbpml
         self.model.set_origin(nbpml)
 
-        def damp_boundary(damp):
+        def damp_boundary(damp, nbp):
```
mlange05: A small docstring or comment explaining what `nbp` is? Otherwise this looks quite cryptic.

examples/acoustic/fwi_operators.py
```diff
@@ -20,13 +20,15 @@ class ForwardOperator(Operator):
      required for the time marching scheme
     """
     def __init__(self, model, src, damp, data, time_order=2, spc_order=6,
-                 save=False, **kwargs):
+                 save=False, u_ini=None, **kwargs):
```
mlange05: Docstring entry

examples/tti/TTI_codegen.py
```diff
@@ -22,13 +22,12 @@ def __init__(self, model, data, source, t_order=2, s_order=2, nbpml=40):
         self.model.nbpml = nbpml
         self.model.set_origin(nbpml)
 
-        def damp_boundary(damp):
+        def damp_boundary(damp, nbp):
```
mlange05: Same as above, a small docstring or comment to help readability.

examples/tti/tti_example2D.py
```diff
@@ -13,9 +13,12 @@
...
 
-model = IGrid(origin, spacing, true_vp, 0.1*(true_vp - 2),
-              0.08 * (true_vp - 2), np.pi/5*np.ones(dimensions),
-              0*np.ones(dimensions))
+model = IGrid(origin, spacing, true_vp,
+              rho=None,
+              epsilon=0.1*(true_vp - 2),
+              delta=0.08 * (true_vp - 2),
+              theta=np.pi/5*np.ones(dimensions),
+              phi=0*np.ones(dimensions))
```
mlange05: Nice, using keyword args makes this much nicer to read. The spacing between factors is inconsistent though and `0*np.ones()` should be `np.zeros()`.

examples/tti/tti_operators.py
```diff
@@ -21,58 +21,64 @@ class ForwardOperator(Operator):
...
     def __init__(self, model, src, damp, data, time_order=2, spc_order=4, save=False,
-                 trigonometry='normal', **kwargs):
+                 trigonometry='normal', u_ini=None, **kwargs):
         nt, nrec = data.shape
         nt, nsrc = src.shape
         dt = model.get_critical_dt()
         u = TimeData(name="u", shape=model.get_shape_comp(),
                      time_dim=nt, time_order=time_order,
-                     space_order=spc_order,
+                     space_order=spc_order/2,
```
mlange05: Ok, this looks to odd to me and definitely requires some type of explanation in a comment or docstring somewhere. Why exactly is this required?

tests/test_tti.py
```diff
@@ -0,0 +1,118 @@
...
+from numpy import linalg
+
+from devito.logger import log
+
+from examples.acoustic.Acoustic_codegen import Acoustic_cg
+from examples.containers import IGrid, IShot
+from examples.tti.TTI_codegen import TTI_cg
+
+
+class TestTTI(object):
```
mlange05: Ok, I know this is how we do it in the acoustic test, but I see absolutely no reason to wrap this test in a class, since we are not storing anything on the object. The only thing it does is force us to use the `request.params` notation instead of setting individual parameter values explicitly like this:
```
@pytest.mark.parametrize('dimension', [(120, 140), (120, 140, 150)])
@pytest.mark.parametrize('space_order', [4])
def test_tti(dimension, space_order):
    ...
```

tests/test_tti.py
mlange05: Indeed this test was very much needed. Well done and many thanks.

examples/tti/tti_operators.py
mloubout: This is so that the discretization is consistent with acoustic.  We apply d/dx to d/dx so using `spc_order/2` makes the second derivative (Laplacian) of order `spc_order`.

tests/test_tti.py
mloubout: Good point will make the change.

mlange05:
Ok, changes look good, many thanks. I'd like the caching fix in PR #164 to go in first though, so that we can set the new test live straight away.

mloubout:
Works for me

mloubout:
@mlange05 only fail is `test_sumbolic_caching`. Don't know why or how to fix that.

devito/finite_difference.py
mlange05: O, it looks like this isn't quite as harmless as it looks, since it breaks the first derivative at first order, ie.
```
In [1]: from devito import *
In [2]: u = DenseData(name='u', shape=(4, 5))
In [3]: u.dx
Out[3]: 0
```
where `master` outputs `-u(x, y)/h + u(x + h, y)/h`. This is unfortunately also missed by our unit test that do not test for `order=1`, which is bad!

tests/test_rewriter.py
```diff
@@ -80,13 +81,14 @@ def test_tti_rewrite_basic():
     output1 = tti_operator(dse=None).apply()
     output2 = tti_operator(dse='basic').apply()
 
-    assert np.allclose(output1[0].data, output2[0].data, rtol=10e-6)
-    assert np.allclose(output1[1].data, output2[1].data, rtol=10e-6)
+    print(np.max(output1[0].data - output2[0].data))
```
mlange05: Debug leftover?

PR #164, opened 2016-12-09
Fix and test cache aliasing

mlange05:
This merge introduces a range of tests for the various cases of symbol caching we do on top of SymPy, as well as a fix for the current aliasing bug (thanks @navjotk).

One `xfailed` test also highlights another remaining caching bug, where we fail to cache new instances of the same function object as SymPy does. Presumably this causes `clear_cache()` to fail if we are running multiple test instances in a single session, but this will be dealt with separately.


FabioLuporini:
OK

PR #165, opened 2016-12-13
Test first derivatives

mlange05:
Small update and test renaming for the derivatives test to cover the first order space derivatives


FabioLuporini:
OK

PR #166, opened 2016-12-13
Allow open loop limits in StencilKernels

mlange05:
This merge further hones the new kernel API to allow generic open loop bounds for dimension loops. The open-loop mode becomes the default for stencil kernels, where dimension sizes are variable and inferred during `.apply()` from the shape of the given array data. The existing "bake-in" behaviour can be recovered by setting explicit iteration sizes on the `Dimension` objects. To demonstrate, the diffusion example now uses generic loop bounds, while setting `x.size = nx; y.size = ny` creates the previous kernel type with constant loop bounds.

An implementation note: Since "buffered" variables, eg. "time", don't allow us to infer the full size of the `t` dimension from data, we now require the actual number of timesteps to be supplied at execution time through keyword args, eg. `t = nt`.   

Development note: We are currently taking a simplifying short-cut where the open loop limit is assumed to be the same variable as the dimension size. This is insufficient for generic loop-blocking and will be further refined in follow-on work.


devito/dimension.py
```diff
@@ -28,6 +29,16 @@ def get_varname(self):
         self._count += 1
         return name
 
+    @property
+    def ccode(self):
+        """C-level variable name of this dimension"""
+        return "%s_size" % self.name if self.size is None else "%d" % self.size
```
FabioLuporini: shall we establish a convention for all methods that produce C code ? as in, calling them ``cname``, ``cdecl``, ``...``

This actually makes me think that we might benefit from a common superclass that represent "Objects that can produce c code"

devito/dimension.py
```diff
@@ -28,6 +29,16 @@ def get_varname(self):
...
+
+    @property
+    def decl(self):
```
FabioLuporini: same as above

FabioLuporini:
Sorry it took so much time to review this.

I think this is good to go, I don't have any major objections. The only think that I start wondering, as you can see from the comments above, is whether we should introduce a proper hierarchy for objects that can return a C representation of themselves . This could be useful. It doesn't need to be addressed in this PR

devito/dimension.py
mlange05: Yes, in principle I agree. But I think this will be best enforced through the visitor pattern that we will need for the loop engine (DLE; discussed offline). The node type then defines and enforces the relevant code components, eg. `ccode`, `cdecl` and the code generation is done via a dedicated visitor class. So I would defer this suggestion until we have that in place.

FabioLuporini: Agreed

PR #167, opened 2017-01-03
DSE - code motion and other features

FabioLuporini:
This PR introduces:

* the concept of ``DSE transformation step`` (see the ``dse_transformation`` decorator in symbolics.py)
* a complete new DSE transformation step: code motion. This is implemented in ``_replace_time_invariants``. Time-invariant sub-expressions are sought and given a score. If the score is greater than an empirically determined threshold, then the sub-expression at hand is "lifted" outside the time loop and evaluated in a temporary
* taylor-based approximation of sin/cos. The default choice, however, is still Bhaskara, as tests start failing with Taylor. I wonder if this is due to the remarkably bad approximation that Taylor gives as one gets far from default center, 0
* refactoring and improvements of various nature for the DSE
* extended propagator to emit time-invariant code outside of the time loop
* new tests

I run some tests to verify the impact of the new transformation (single socket, 8 threads, icc O2 xHost, pinning, etc):

* TTI space order 4: runtime drops from 127s to 91s (~1.4X speed-up)
* TTI space order 10: runtime drops from 373s to 322s (1.16X speed-up)

If *no* code motion is performed and sin/cos are replaced with the respective Bhaskara approximations, the execution time in TTI space order 4 only drops from 127s to 105s. 
If code motion is performed and *no* Bhaskara approximation is introduced, than the run-time is pretty much the same (92s in TTI space order 4) *but* the compile time is much higher (icc takes more to vectorize loops with sin/cos). For example, in TTI space order 10, the compilation time drops from 74s to 16s thanks to Bhaskara

### DIFF
--- a/devito/dse/extended_sympy.py
+++ b/devito/dse/extended_sympy.py
@@ -97,13 +97,13 @@ def eval_taylor_sin(expr):
     try:
         Float(expr)
         return v.doit()
-    except TypeError, ValueError:
+    except (TypeError, ValueError):
         return v
 
 
 def eval_taylor_cos(expr):
     expr_square = Mul(expr, expr, evaluate=False)
-    b =  1.0-.5*expr_square*(1.0-expr_square/12.0)
+    b = 1.0-.5*expr_square*(1.0-expr_square/12.0)
     v = 1.0 + Mul(-0.5,
                   Mul(expr, expr, evaluate=False),
                   1.0 + Mul(expr, expr, -1/12.0, evaluate=False),
@@ -111,7 +111,7 @@ def eval_taylor_cos(expr):
     try:
         Float(expr)
         return b, v.doit()
-    except TypeError, ValueError:
+    except (TypeError, ValueError):
         return v
 
 
--- a/devito/dse/extended_sympy.py
+++ b/devito/dse/extended_sympy.py
@@ -39,6 +39,28 @@ class Add(sympy.Add, UnevaluatedExpr):
     pass
 
 
+class taylor_sin(TrigonometricFunction):
+
+    """
+    Approximation of the sine function using a Taylor polynomial.
+    """
+
+    @classmethod
+    def eval(cls, arg):
+        return 0.0 if arg == 0.0 else eval_taylor_sin(arg)
+
+
+class taylor_cos(TrigonometricFunction):
+
+    """
+    Approximation of the cosine function using a Taylor polynomial.
+    """
+
+    @classmethod
+    def eval(cls, arg):
+        return 1.0 if arg == 0.0 else eval_taylor_cos(arg + 1.5708)
+
+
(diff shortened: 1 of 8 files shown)


mloubout:
What is 10th order doing there. I already said it multiple timers, for TTI, only runs multiple of 4 orders.

FabioLuporini:
I remembered multiple of 2 . Doesn't really matter in this context though. Any sufficiently high random order suites the purpose here.

mloubout:
Even if it doesn't really matter, the point is we keep running test with this orders that should not be used. So evenso it does not impact your result, it means noone will remember it should not be used if people don't make sure to use it properly wether or not it is for a proper geophysical application. I don't want to ebd up with shotty results in a few month because everyone forgot it again.

FabioLuporini:
Yes, so it's best if we add checks at the beginning of TTI which prevent the execution of what u believe should never be run

ggorman:
I'm supporting @mloubout's point here. It feels like we are leaving behind a landmine for someone to step on...

How about making use of the python warnings module?

FabioLuporini:
I also agree with his point, that's why I proposed to add checks for when one tries to run TTI

mloubout:
Another thing. There is with that PR, no possibilty to disable dse. This is sometimes required for non implemented case (I sended you a private message withone case ). Can we add a `no-dse` case to have everything covered?

FabioLuporini:
@mloubout If you instantiate operators passing ``dse=None``, no DSE transformations is applied

mlange05:
OK, so basically this looks really great - although a little complicated for novices. ;) Tying the allocation of temporaries directly into the Propagator worries me slightly, since it makes the 2.0 re-write harder, but in the interest of time I think we can merge this now and unpick later. Also, the TTI space order discussion is orthogonal to this PR, so I think this is good to go.

FabioLuporini:
Thanks for merging this. The reason temporaries are allocated in the propagator is that the DSE is orthogonal to code generation, as it's just pure expression manipulation. 

I think we have to be prepared in 2.0 to create a general mechanism to allocate temporaries at a certain loop nest depth. With such a mechanism, allocation of temporaries for stencils, additional expressions (eg sparse point interpolation), time invariants etc would be handled uniformly with minimal pain

mlange05:
I agree, I was just pointing out that the "allocation" code used for generic FT on symbols and for temporaries should all come from the same place.

Also, I just realised this, but are we actually freeing temporaries in the kernel too, or are we leaking memory here?

FabioLuporini:
Good spot. I'm 100% I handled that, and in fact now that I look at the code, it's missing other stuff. I must have dropped a commit during an interactive rebase.

I'm rebasing #170 to include what's been shamefully dropped here, so that (amongst the other things missing) we don't leak memory

PR #170, opened 2017-01-14
Various DSE improvements

FabioLuporini:
A number of improvements of various nature

Performance-wise, the most important thing is the introduction of the MACROs to prevent denormal numbers at run-time. This greatly improve performance (I might show some numbers in the next few days)

This must go after #167

### DIFF
--- a/devito/dse/manipulation.py
+++ b/devito/dse/manipulation.py
@@ -1,0 +1,76 @@
+"""
+Routines to construct new SymPy expressions transforming the provided input.
+"""
+
+from sympy import Indexed, S
+
+from devito.dse.extended_sympy import Add, Mul
+
+
+def unevaluate_arithmetic(expr):
+    """
+    Reconstruct ``expr`` turning all :class:`sympy.Mul` and :class:`sympy.Add`
+    into, respectively, :class:`devito.Mul` and :class:`devito.Add`.
+    """
+    if expr.is_Float:
+        return expr.func(*expr.atoms())
+    elif isinstance(expr, Indexed):
+        return expr.func(*expr.args)
+    elif expr.is_Symbol:
+        return expr.func(expr.name)
+    elif expr in [S.Zero, S.One, S.NegativeOne, S.Half]:
+        return expr.func()
+    elif expr.is_Atom:
+        return expr.func(*expr.atoms())
+    elif expr.is_Add:
+        rebuilt_args = [unevaluate_arithmetic(e) for e in expr.args]
+        return Add(*rebuilt_args, evaluate=False)
+    elif expr.is_Mul:
+        rebuilt_args = [unevaluate_arithmetic(e) for e in expr.args]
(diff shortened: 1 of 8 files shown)


examples/benchmark.py
```diff
@@ -34,6 +34,9 @@
     parser.add_argument(dest="execmode", nargs="?", default="run",
                         choices=["run", "test", "bench", "plot"],
                         help="Exec modes")
+    parser.add_argument("--bench-mode", "-bm", dest="benchmode", default="all",
+                        choices=["all", "at", "dse"],
+                        help="Choose what to benchmark (all, at, dse).")
```
mlange05: Need to drop explicit `--auto-tuning` then, no?

devito/dse/manipulation.py
```diff
@@ -0,0 +1,76 @@
...
+def rxreplace(exprs, mapper):
+    """
+    Apply Sympy's xreplace recursively.
+    """
+
+    replaced = []
+    for i in exprs:
+        old, new = i, i.xreplace(mapper)
+        while new != old:
+            old, new = new, new.xreplace(mapper)
```
mlange05: I'm not entirely sure why this is needed. Does `xreplace` not capture all instances or are we simply applying our own `mapper` to `mapper`?

examples/benchmark.py
FabioLuporini: Good spot. Actually, what I think I have to do is to turn this ``at`` into ``blocking``. Because in theory the DSE itself could be autotuned (although at present there should be no need for it). In other words, the confusion here arises from the fact that so far we assume AT==blocking although they two different things

devito/dse/manipulation.py
FabioLuporini: I don't remember the actual examples, but I can definitely look them up. Basically, if you have an outer match and an inner match, then ``xreplace`` only applies to the outer match. Stay tuned for the example

FabioLuporini:
OK, I'll share details about the question above and tweak this PR (travis failing, AT, ...) soon

devito/dse/manipulation.py
FabioLuporini: OK so this has to do with the way CSE() was written time ago (I simply copy-pasted it in the DSE from old code developed over the last summer)... and it's at the bottom of my priority list to rewrite it now

basically some of the replacements can contain replaceable parts themselves (ie, the ``mapper`` is actually a graph), so ``rxreplace`` ensures we don't miss these inner replacements

FabioLuporini:
Fixed AT->blocking in ``benchmark.py``

FabioLuporini:
Also fixed:
- Python 3.5 compatibility (hopefully we'll have a green Travis soon)
- memory leaks and other less relevant code generation issues (which should have been in #167, but erroneously got dropped..)

mlange05:
Ok, looks good. Merging....

PR #171, opened 2017-01-16
Move profiling to DSE

FabioLuporini:
This PR makes a step forward towards Devito 2.0

The obsolete, hacky, and bug-prone profiler.py module is drastically trimmed, and its key functionality of "counting flops in the generated code" is moved to the DSE. It is indeed possible to count the number of operations by applying the DSE function ``estimate_cost``, which in turn wraps Sympy's  ``count_ops``, to any Sympy expression.

The operation of counting the memory accesses performed by a Sympy expression is also moved to the DSE (from Propagator).

Summing up, now the DSE returns a data structure containing:
- the transformed expressions
- a mapper to identify time-invariant expressions, out of the transformed expressions
- operations count
- memory accesses count

This PR depends on #170

### DIFF
--- a/devito/propagator.py
+++ b/devito/propagator.py
@@ -251,7 +251,8 @@ class Propagator(object):
             niters_per_section[key] = with_time_loop(niters)
 
         key = LOOP_BODY.name
-        niters = reduce(operator.mul, self.shape)
+        niters = reduce(operator.mul,
+                        [j - i for i, j in self._space_loop_limits.values()])
         niters_per_section[key] = with_time_loop(niters)
 
         for i, subsection in enumerate(self.time_loop_stencils_a):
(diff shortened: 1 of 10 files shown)


devito/propagator.py
```diff
@@ -261,89 +272,36 @@ def niters(self):
...
-            reads = set(flatten([e.rhs.atoms(Indexed) for e in expressions]))
-            writes = set(flatten([e.lhs.atoms(Indexed) for e in expressions]))
-            reads = set([access(s) for s in reads if filter(s)])
-            writes = set([access(s) for s in writes if filter(s)])
-            if mode == 'ideal':
-                return len(set(reads) | set(writes))
-            else:
-                return len(reads) + len(writes)
+    @property
+    def dataspace(self):
```
mlange05: Docstring

PR #173, opened 2017-01-19
Compiler: Adding `march=native` to `GNU/CustomCompiler` to enable SSE

mlange05:
On GCC version 4.8.4 our compiles otherwise fail with the recent denormals fix due to SSE not being enabled.


FabioLuporini:
OK

PR #175, opened 2017-01-19
fix gnu flag

mloubout:
OSX compatibility of `-march=native` with the extra `-mno-avx`.

### DIFF
--- a/devito/compiler.py
+++ b/devito/compiler.py
@@ -76,13 +76,26 @@ class GNUCompiler(Compiler):
         self.version = kwargs.get('version', None)
         self.cc = 'gcc' if self.version is None else 'gcc-%s' % self.version
         self.ld = 'gcc' if self.version is None else 'gcc-%s' % self.version
-        self.cflags = ['-O3', '-g', '-march=native', '-mno-avx', '-fPIC', '-Wall',
-                       '-std=c99', '-Wno-unused-result', '-Wno-unused-variable']
+        self.cflags = ['-O3', '-g', '-march=native', '-fPIC', '-Wall', '-std=c99',
+                       '-Wno-unused-result', '-Wno-unused-variable']
         self.ldflags = ['-shared']
 
         if self.openmp:
             self.ldflags += ['-fopenmp']
         self.pragma_ivdep = [Pragma('GCC ivdep')]
+
+
+class GNUCompilerNoAVX(GNUCompiler):
+    """Set of compiler flags for GCC but with AVX suppressed. This is
+    a work around for a known gcc bug on MAC OS."
+
+    :param openmp: Boolean indicating if openmp is enabled. False by default
+    """
+
+    def __init__(self, *args, **kwargs):
+        super(GNUCompilerNoAVX, self).__init__(*args, **kwargs)
+        self.cflags = ['-O3', '-g', '-march=native', '-mno-avx', '-fPIC', '-Wall',
+                       '-std=c99', '-Wno-unused-result', '-Wno-unused-variable']
 
 
 class ClangCompiler(Compiler):
@@ -191,6 +204,7 @@ compiler_registry = {
     'gcc': GNUCompiler, 'gnu': GNUCompiler,
     'gcc-4.9': partial(GNUCompiler, version='4.9'),
     'gcc-5': partial(GNUCompiler, version='5'),
+    'gcc-noavx': GNUCompilerNoAVX, 'gnu-noavx': GNUCompilerNoAVX,
     'clang': ClangCompiler, 'osx': ClangCompiler,
     'intel': IntelCompiler, 'icpc': IntelCompiler,
     'icc': IntelCompiler,
--- a/devito/compiler.py
+++ b/devito/compiler.py
@@ -76,8 +76,8 @@ class GNUCompiler(Compiler):
         self.version = kwargs.get('version', None)
         self.cc = 'gcc' if self.version is None else 'gcc-%s' % self.version
         self.ld = 'gcc' if self.version is None else 'gcc-%s' % self.version
-        self.cflags = ['-O3', '-g', '-march=native', '-fPIC', '-Wall', '-std=c99',
-                       '-Wno-unused-result', '-Wno-unused-variable']
+        self.cflags = ['-O3', '-g', '-march=native', '-mno-avx', '-fPIC', '-Wall',
+                       '-std=c99', '-Wno-unused-result', '-Wno-unused-variable']
         self.ldflags = ['-shared']
 
         if self.openmp:
@@ -175,7 +175,7 @@ class CustomCompiler(Compiler):
         super(CustomCompiler, self).__init__(*args, **kwargs)
         self.cc = environ.get('CC', 'gcc')
         self.ld = environ.get('LD', 'gcc')
-        default = '-O3 -g -march=native -fPIC -Wall -std=c99'
+        default = '-O3 -g -march=native -fPIC -mno-avx -Wall -std=c99'
         self.cflags = environ.get('CFLAGS', default).split(' ')
         self.ldflags = environ.get('LDFLAGS', '-shared').split(' ')
 


mloubout:
So I guess this is quie controversial and my "fix" is probably not the way to go. 
I will still let this PR open as I do think we need to make sure we have proper portability on different system. I however do not know what is the correct way to fix this.

FabioLuporini:
definitely we do *not* want anything avx-specific 

remind us what compile line failed and what system was that?

mloubout:
```
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T//ccAwyOSD.s:563:no such instruction: `vmovss %xmm2, 8(%rsp)'
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T//ccAwyOSD.s:566:no such instruction: `vmovss 4(%rbx), %xmm1'
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T//ccAwyOSD.s:567:no such instruction: `vcvttss2si %xmm0, %r14d'
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T//ccAwyOSD.s:568:no such instruction: `vmulss LC6(%rip), %xmm1,%xmm0'
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T//ccAwyOSD.s:569:no such instruction: `vmovss %xmm1, (%rsp)'
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T//ccAwyOSD.s:572:no such instruction: `vmovss 8(%rsp), %xmm2'
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T//ccAwyOSD.s:574:no such instruction: `vcvttss2si %xmm0, %ecx'
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T//ccAwyOSD.s:575:no such instruction: `vmovss (%rsp), %xmm1'
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T//ccAwyOSD.s:577:no such instruction: `vcvtsi2ss %r14d, %xmm0,%xmm0'
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T//ccAwyOSD.s:579:no such instruction: `vmovss 16(%rsp), %xmm9'
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T//ccAwyOSD.s:580:no such instruction: `vmovss 24(%rsp), %xmm8'
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T//ccAwyOSD.s:581:no such instruction: `vmovss 32(%rsp), %xmm7'
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T//ccAwyOSD.s:583:no such instruction: `vmovss 40(%rsp), %xmm6'
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T//ccAwyOSD.s:584:no such instruction: `vmulss LC7(%rip), %xmm0,%xmm0'
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T//ccAwyOSD.s:587:no such instruction: `vaddss %xmm0, %xmm2,%xmm2'
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T//ccAwyOSD.s:588:no such instruction: `vcvtsi2ss %ecx, %xmm0,%xmm0'
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T//ccAwyOSD.s:591:no such instruction: `vmulss LC12(%rip), %xmm2,%xmm2'
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T//ccAwyOSD.s:592:no such instruction: `vmulss LC7(%rip), %xmm0,%xmm0'
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T//ccAwyOSD.s:594:no such instruction: `vaddss %xmm0, %xmm1,%xmm1'
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T//ccAwyOSD.s:595:no such instruction: `vmulss %xmm1, %xmm2,%xmm0'
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T//ccAwyOSD.s:596:no such instruction: `vmulss (%r12), %xmm0,%xmm0'
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T//ccAwyOSD.s:597:no such instruction: `vdivss (%rcx,%rdx,4), %xmm0,%xmm0'
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T//ccAwyOSD.s:598:no such instruction: `vaddss 0(%r13,%rax,4), %xmm0,%xmm0'
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T//ccAwyOSD.s:599:no such instruction: `vmovss %xmm0, 0(%r13,%rax,4)'
FAILED compiler invocation: gcc-mp-4.8 -O3 -g -march=native -fPIC -Wall -std=c99 -shared -fopenmp /var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T/devito-502/67620a6ffbdb08b4a053d46cf5ec68add6295d16.c -o /var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T/devito-502/67620a6ffbdb08b4a053d46cf5ec68add6295d16.dylib
```

Fals on OSX with all GCC versions I have (4.4, 4.8, 4.9, 5)

FabioLuporini:
Can you please attach the output of the following commands on your system:

``
sysctl machdep.cpu.features
sysctl machdep.cpu.brand_string
``

mloubout:
```
decon:examples mloubout$ sysctl machdep.cpu.features sysctl machdep.cpu.brand_string
machdep.cpu.features: FPU VME DE PSE TSC MSR PAE MCE CX8 APIC SEP MTRR PGE MCA CMOV PAT PSE36 CLFSH DS ACPI MMX FXSR SSE SSE2 SS HTT TM PBE SSE3 PCLMULQDQ DTES64 MON DSCPL VMX SMX EST TM2 SSSE3 CX16 TPR PDCM SSE4.1 SSE4.2 x2APIC POPCNT AES PCID XSAVE OSXSAVE TSCTMR AVX1.0
machdep.cpu.brand_string: Intel(R) Core(TM) i5-2500S CPU @ 2.70GHz
```

FabioLuporini:
what happens if you keep `-march=native` but you also add `-Wa,-q` ?

mloubout:
```
gcc: unrecognized option '-q'
cc1: error: unrecognized command line option "-Wa"
FAILED compiler invocation: gcc -O3 -g -march=native -fPIC -Wall -std=c99 -Wa -q -shared /var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T/devito-502/397d774d123ec8439e959b4b461cceaa582a9d17.c -o /var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T/devito-502/397d774d123ec8439e959b4b461cceaa582a9d17.so
```

FabioLuporini:
This is obviously a compiler-system problem 

Nobody will ever complain about performance on OSX with gcc. So I think we should just drop `gcc` support for OSX and recommend to use `clang`.

mloubout:
>Nobody will ever complain about performance on OSX with gcc. So I think we should just drop gcc support for OSX and recommend to use clang.

Not sure "don;t want to support it" is the way to go neither, and no one ever use clang

FabioLuporini:
> Not sure "don;t want to support it" is the way to go neither, and no one ever use clang

Isn't clang just the default compiler fronted on OSX ?

What I meant is: we could drop the vectorization option on OSX+gcc, and just leave O3

navjotk:
I would have a problem with dropping gcc support on OSX. Clang doesn't support OpenMP so gcc is the only way I have to test parallel code on my macbook which is my primary development machine. 

Dropping support here would cut me off from the most common use case - parallel execution.

FabioLuporini:
Sorry, what I meant was "dropping the vectorization flag" , not dropping gcc support

navjotk:
Ok thanks for clarifying.

I think this is the time to discuss (again) a good way to easily add/remove individual flags without having to specify all flags (which would be the case if we use `CustomCompiler` with `Cflags` for this) just to remove a single flag from the invocation string. 

Of course, creating a `OsxGccCompiler` class is an option here. However, there might be scenarios where installing devito on some cluster requires something like this. SysAdmins can't be expected to create new classes to get the base functionality of a library on their system. Can we think of an easier way to do this - remembering that the end users of this this piece of code are SysAdmins and not application developers? I am thinking global config here. Open debate :)

mlange05:
Ok, here is my view on things:
a) Adding/removing options via some non-standard flag sounds dangerous to me since our "best guess" defaults are a moving target and we would thus require users to keep up with our commit history.

b) HPC clusters usually have highly specialised environments with much more detailed "best known" configurations than we could ever guess. So on a cluster I would always advice sys-admins to use custom compiler settings from their own toolchain, which is particularly easy, since we follow custom conventions by reading `CFLAGS`, `LDFLAGS`, etc. Since clusters also usually rely on environment modules to customise package settings having long-ish lines for setting `CFLAGS` is also no problem, so for sys-admins there is really no need to add/remove a single option.

c) For stuff to "Just Work" on common setups is the priority of our custom presets, and since this issue hits two of our main developers, adding a preset for `OsxGccCompiler` sounds the most sensible thing to do to me. I'd also add warning about disabling AVX, but as mentioned above nobody will care too much about raw performance on this configuration.

P. S.
> remembering that the end users of this this piece of code are SysAdmins and not application developers?

I really don't think this is true, since sys-admins should only feed us good general compiler settings for their system. I also don't think that the global-config discussion, ie. how to set any flag/option, has anything to with this.

FabioLuporini:
I think that if we want to add a new class with the avx vectorization flag, that *must* have "AVX" in the name and a docstring saying something along the lines of "this is a work around for a known gcc bug on MAC OS". The name ``OsxGccCompiler`` would otherwise be misleading, as people might use it on old OSX-based platforms (where there's no avx) and get segfaults.

mlange05:
I think this is the related link:
http://stackoverflow.com/questions/10327939/error-no-such-instruction-while-assembling-project-on-mac-os-x

navjotk:
Let's say I create a class called `GCCNoAVX` that adds this flag. I still want to use the `CustomCompiler` class since the name of the actual compiler executable should come from the environment variable (as pointed out, my system uses non-standard names since `gcc` maps to clang). However, `CustomCompiler` inherits from `GCCCompiler` and not this new class I create. Please suggest a clean way to do this because I think our current compiler setup is quite specific to certain systems.

mlange05:
`CustomCompiler` does not inherit it's settings from `GCCCompiler`, it merely defaults to GCC settings whenever it can't find the appropriate env variable. So if you want to use a custom setting with some parts of GCC simply replace the parts that are custom with the appropriate env flag (`CC`, `CFLAGS`, etc).

mlange05:
Ok, I've added the workaround prefix. Can developers with Mac OS + GCC please try running with `DEVITO_ARCH=gcc-noavx` to confirm this works?

mloubout:
Works on mine

navjotk:
Mine too

mlange05:
Cool, merging this then...

PR #176, opened 2017-01-23
Forward/backward now field property

mloubout:
This PR aims to fix the time substitution so that the symbolic interface has a rigorous inversion setup.
I didn't change much, I only added a property to the `TimeData` object called `Taxis` (open to discussion for naming)
This property allows two things:
- Keep track of what is a forward wavefield and what is an adjoint wavefield. This has the advantage to allow mixed operation such as `u.dt * v.dt` where `u` is forward and `v` is adjoint. Previous implementations were not handling this properly, but with this new property the time derivatives are taken `left` for `u` and `right` for v which is the proper thing to do.

- This does not change the symbolic manipulation, the only difference is that now, `TimeData` has its own `indexify()` function doing the time substituion according to `taxis`


mloubout:
So this randomly fails at the gradient, trying to figure out why.

devito/interfaces.py
```diff
@@ -15,6 +15,28 @@
 __all__ = ['DenseData', 'TimeData', 'PointData']
 
 
+# This class encapsulates the type of time data in two case
+# Forward : means the time derivative has to be taken to the left (t-2, t-1, t
+# Backward : means the time derivative has to be taken to the right (t+2, t+1, t)
+# Keeping track of this information allows to have interleaving forward and backward
+# wavefield and is specially handy for the computation of the gradient computing
+# the correlation of a forward and adjoint wavefield
```
FabioLuporini: This comment should probably be inside ``class Taxis`` and turned into a docstring ("""...""")

FabioLuporini:
I think a better way of doing this is to have a common superclass ``TimeData`` and then two subclasses, named for example ``BackwardTimeData`` and ``ForwardTimeData``. These two subclasses would share most of the superclass code. This would also easily implementing runtime checks: for example, it'd become easy to raise an error if one tries to invoke ``.backward(..)`` on a ``ForwardTimeData``.

The ``TimeData`` API would *not* need an additional argument (we're already passing way too many arguments) to specify if it's either backward vs forward; You would *not* need anymore an additional attribute (self.taxis); ``indexify`` could just be implemented differently in  ``BackwardTimeData`` and ``ForwardTimeData``.

If you're not familiar on how to do this, I'm happy to help. Just let me know. 

Secondly, we start having many "indexifiable" classes. Probably we need a common ancestor class for all these, but I'm interested in @mlange05 thoughts on how to build a proper hierarchy here

mloubout:
>  it'd become easy to raise an error if one tries to invoke .backward(..) on a ForwardTimeData.

There is no reason why this should raise an error.. Backward/Forward is not a field property, this is a computational property that says how it's been/ how it will be computed.
Separating it would just incur a lot of code replication that doesn't seems to be required.  And also I do not like sperating the two type for the main reason that , as I said, the only difference is in which order it has been computed, but should still have all the same properties/class functions.

concerning the number of argument,, I don't see how we can reduce it, every `timedata` argument can;t be avoided

FabioLuporini:
> There is no reason why this should raise an error.. Backward/Forward is not a field property, this is a computational property that says how it's been/ how it will be computed.

Can you invoke ``backward`` or ``forward`` no matter the value of ``taxis``?

> Separating it would just incur a lot of code replication that doesn't seems to be required.

No, because you're gonna share code through the superclass

> And also I do not like sperating the two type for the main reason that , as I said, the only difference is in which order it has been computed, but should still have all the same properties/class functions.

You are already "faking" two different types by specifying the behaviour based on taxis. 

> concerning the number of argument,, I don't see how we can reduce it, every timedata argument can;t be avoided

Simply because most ``TimeData`` in a program *share the same arguments*. For example, same ``time_order``, same ``space_order``, same ``time_dim``, same ``save``... So what we could/should have is, for example, a ``TimeDataSpace`` class, from which one can draw ``TimeData`` objects all with same ``time_order``, ``space_order`` etc.

For example:

`
tds = TimeDataSpace(space_order=2, time_order=2, time_dim=...)
u = TimeData('u', <other u-specific attributes>, tds)
u = TimeData('v', <other v-specific attributes>, tds)
`
Same for ``DenseData``.

This is just an example of *how it could be possible to do things* -- it does *not* want to be a suggestion. Just a thought.

devito/interfaces.py
```diff
@@ -429,7 +452,20 @@ def _allocate_memory(self):
         self._full_data = self._data.view()
 
         if self.pad_time:
-            self._data = self._data[self.time_order:, :, :]
+            self._data = self._data[self.time_order:, ...]
+
+    def indexify(self):
+        """Convert base symbol and dimensions to indexed data accesses
+
```
mlange05: OK, I really don't think hijacking `indexify()` for this purpose is the way to go, because this function has a very specific use case and that is to translate out equivalents of `sympy.Function` to `sympy.Indexed`. Everything else is unintended "magic behaviour from a user's perspective.

tests/test_gradient.py
```diff
@@ -112,7 +110,7 @@ def test_grad(self, acoustic):
             error2[i] = np.absolute(.5*linalg.norm(d - rec)**2 - F0 - H[i] * G)
             # print(F0, .5*linalg.norm(d - rec)**2, error1[i], H[i] *G, error2[i])
             # print('For h = ', H[i], '\nFirst order errors is : ', error1[i],
-            #       '\nSecond order errors is ', error2[i])
+            #        '\nSecond order errors is ', error2[i])
```
mlange05: Dead debug prints should generally be removed from the code base.

mlange05:
In fact, reading this through again and thinking of the ongoing re-write it would probably be a really good idea to create small self-contained test case that highlights the expected stencil structure of the time discretisation for both cases without relying on any of the complicated example operators.

mloubout:
>Can you invoke backward or forward no matter the value of taxis?

Yes you can

>Simply because most TimeData in a program share the same arguments. For example, same time_order, same space_order, same time_dim, same save... So what we could/should have is, for example, a TimeDataSpace class, from which one can draw TimeData objects all with same time_order, space_order etc.

Once again, no. You can have a mix of time order/space orders in the same operators.

mloubout:
>symbolic derivative expansion in some way, then this should probably go in either finite_difference.py and/or the .dt properties. Or am I missing some subtlety here?

That was the original idea, but everything in `densedata` breaks if `t` is replaced by something else.

devito/interfaces.py
mloubout: >Everything else is unintended "magic behaviour from a user's perspective.

Still better than non-working and wrong magic behaviour. Current status time handling is bad at best and need to be fixed.

FabioLuporini:
> Once again, no. You can have a mix of time order/space orders in the same operators.

Once again, yes. You could, and let me emphasize *could*, simply have more than one ``TimeDataSpace`` in your program, through which determining the space_order and/or the time_order of each Data object used in an operator.

devito/interfaces.py
FabioLuporini: Agreed. However, Michael is pointing out that this is not the right fix. So maybe we should think about how to properly do this.

mlange05:
> That was the original idea, but everything in densedata breaks if t is replaced by something else.

Great. Can you then please provide a Minimal Failing Example of the original idea (ie. something that is small and independent of the example operators) to demonstrate how this should work in theory and how it breaks now? I'm asking because once 3.0 lands we will hopefully be able to remove `TimeData` objects altogether (or at least most of it's special-casing), because in the new design time is "just another dimension". In order to get this right we need test cases that expose flaws and demonstrate how things should work conceptually.

mloubout:
I'll see what I can do....

mloubout:
>Great. Can you then please provide a Minimal Failing Example of the original idea (ie. something that is small and independent of the example operators) to demonstrate how this should work in theory and how it breaks now? I'm asking because once 3.0 lands we will hopefully be able to remove TimeData objects altogether (or at least most of it's special-casing), because in the new design time is "just another dimension". In order to get this right we need test cases that expose flaws and demonstrate how things should work conceptually.

The problem is in operator 
```/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T/devito-502/eeaa5c9f631bcc967e36e0554406340d20b5483f.c: In function 'init':
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T/devito-502/eeaa5c9f631bcc967e36e0554406340d20b5483f.c:23: warning: ignoring #pragma ivdep 
CustomCompiler: compiled /var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T/devito-502/eeaa5c9f631bcc967e36e0554406340d20b5483f.c [0.30 s]
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T/devito-502/322c1ef74195eb7484ec984dbfb2e7c9829f6150.c: In function 'init':
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T/devito-502/322c1ef74195eb7484ec984dbfb2e7c9829f6150.c:22: warning: ignoring #pragma ivdep 
CustomCompiler: compiled /var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T/devito-502/322c1ef74195eb7484ec984dbfb2e7c9829f6150.c [0.21 s]
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T/devito-502/7d776a40bc7d13c3b09b56a81e1e7da33c51de50.c: In function 'init':
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T/devito-502/7d776a40bc7d13c3b09b56a81e1e7da33c51de50.c:22: warning: ignoring #pragma ivdep 
CustomCompiler: compiled /var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T/devito-502/7d776a40bc7d13c3b09b56a81e1e7da33c51de50.c [0.10 s]
CustomCompiler: compiled /var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T/devito-502/cbada689a4b0b2940dfdfa9d181bc7b6f106e328.c [0.10 s]
[s, z, t, y, x]
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/examples/acoustic/acoustic_example.py", line 102, in <module>
    run(full_run=True, auto_tuning=False, space_order=6, time_order=2)
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/examples/acoustic/acoustic_example.py", line 87, in run
    auto_tuning=auto_tuning, compiler=compiler
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/examples/acoustic/Acoustic_codegen.py", line 78, in Forward
    compiler=compiler, profile=True, u_ini=u_ini)
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/examples/acoustic/fwi_operators.py", line 73, in __init__
    **kwargs)
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/devito/operator.py", line 100, in __init__
    sym_undef.remove(d)
KeyError: t - s

Process finished with exit code 1
```
If your time dimension is not `t` but `t-s`

mloubout:
So mainly the problem is, almost everywhere, time handling is hard coded as :
-you update at `t`
- every index is >`t` for forward and <`t` for backward

so this is just unmanageable

mloubout:
Also, with `forward/backward` as a field property, it will greatly improve the finite_difference capabilities. For every odd order derivatives (d/dx, d/dx^3) the forward and adjoint have a sign difference as if you apply d/dx in forward mode, you need to apply (d/dx)^T in adjoint mode. For even order, it is the same as the stencil is symmetric, but for odd order the stencil is antisymmetric and currently requires to add a `-` sign by hand in the adjoint stencil.

mloubout:
@mlange05 @navjotk @FabioLuporini 
New commit that does not touch indexify anymore. Everything is broken but should show what I'm trying to do. I'll keep working on it.

mloubout:
I still think that `forward/backward` need to be a `kwarg` for `StencilKernel` as it may not be possible in some case (gradient with boundary method that reverse propagate the forward wavefield) to infer it from the stencil and/or fields.

devito/visitors.py
```diff
@@ -413,8 +413,7 @@ def visit_Iteration(self, o, subs={}, offsets=defaultdict(set)):
             init = []
             for i, off in enumerate(filter_ordered(offsets[o.dim])):
                 vname = "%s%d" % (o.dim.name, i)
-                # Add modulo to value so that t-modulo is always positive
-                value = o.dim.parent + off + o.dim.modulo
+                value = o.dim.parent + off
```
mloubout: why did you remove that? The `+modulo` is making sure indices are always cyclic. without it the first 3 time steps are not which is maybe why this one change broke the adjoint test.

mloubout:
#210 will take what's needed from here and then I'll close this.

PR #177, opened 2017-01-27
Add DSE and a new AST infrastructure to StencilKernel 

mlange05:
This merge represents a massive step towards Devito-3.0.It integrates the DSE with the current `StencilKernel` and adds a new flexible AST node and tree-visitor infrastructure to represent and manipulate arbitrary loop nests. In particular this merge adds:
* Perform symbolic DSE optimisation on raw incoming SymPy expressions before building `Iteration`s
* Add a hierarchy of AST nodes that include `Expression` and `Iteration` among others
* Add base classes for tree traversal visitors:
  * `Visitor` for passive tree traversal, eg. finding things in the AST
  * `Transformer` for manipulating and re-building a tree
* `FindSymbol` visitor, allowing us to infer the kernel signature more elegantly
* `FindSection` visitor and `Timer` node, allowing us to automate section-specific profiling in C
* `SubstituteExpression` transformer that performs symbolic substitution on `Expression` nodes in a tree
* `ResolveIterationVariable` transformer that creates and substitutes explicit iteration variables for `Dimension`s, including explicit buffer variables (eg. `t0`, `t1`, `t2`) for buffered variables


FabioLuporini:
What can I say, this is a joint PR from me and Michael, so maybe someone else should review this code. 

I again skimmed through it. There's a lot of polishing that needs to be done, but the PR is already massive. I therefore think this should land ASAP, since it will be followed by a stream of smaller PRs

devito/nodes.py
```diff
@@ -0,0 +1,273 @@
...
+        return c.Block(self.header + body + self.footer)
+
+    @property
+    def children(self):
+        return (self.body,)
+
+
+class Expression(Node):
+
+    """Class encpasulating a single stencil expressions"""
```
ggorman: expressions -> expression

devito/nodes.py
```diff
@@ -0,0 +1,273 @@
...
+    def __init__(self, lname, gname, body):
+        """
+        Initialize a Timer object.
+
+        :param lname: Timer name in the local scope.
+        :param gname: Name of the global struct tracking all timers.
+        :param body: Timed block of code.
+        """
+        self._name = lname
+        # TODO: need omp master pragma to be thread safe
```
ggorman: ...err, why? And I am more than a little scared if we are concerned about thread safety at this point.

mlange05: The new infrastructure doesn't deal with OpenMP yet, so these question will all be addressed in upcoming PRs.

mlange05:
Signed off in the chat by @ggorman. So merging now...

PR #178, opened 2017-01-27
Versioneer: Adding basic version tracking with versioneer

mlange05:
Adds basic versioning infrastructure to improve pip installs.


FabioLuporini:
isn't the versioneer pip-installable ?

mlange05:
It is, but it still installs its own script into the package on setup. I think that's for release/deployment workflows, since these need version info, but don't install dependencies.

mloubout:
As this looks like an instal/use PR, can we add a link to opesci.org to the README.

mlange05:
> As this looks like an instal/use PR, can we add a link to opesci.org to the README.

Yes, done.

FabioLuporini:
I wonder whether we should put all these third-party packages in a separate folder.. </picky-mode>

This can land for me

PR #179, opened 2017-01-28
Toward Devito 3.0: extended StencilKernel, introduce DLE

FabioLuporini:
*** Note ***
This PR has been opened to promote discussion, but it still lacks:
- flake8 fixes, isort fixes
- a full run of the test suite (let's see what Travis says...)
- more tests (either tomorrow or Monday)
*************

*** EDIT 1 ***
- passing all tests
- added some DLE tests
*************

Now ``StencilKernel`` is a very special node in an Iteration/Expression tree -- a node that can be JIT-compiled and that is allowed to call the optimization engines. As part of this generalization, more nodes and visitors have been introduced. Further, some performance information is now emitted.

Moreover, the PR introduces the DLE -- Devito Loop Engine. This is an analogous of DSE, but for loop optimization (eg, tiling, parallelization). Only one transformation step is at this stage applied, called ``create_elemental_functions``. This basically can move inner loops into separate functions. This, *for example*, will be essential when applying loop tiling to TTI, as the presence of a deep loop nest (body + reminder loops) including a huge number of stencil statements make the compilation time simply blow up. 

Creating "elemental functions" may also be useful to retain some data alignment (stay tuned for more details)

### DIFF
--- a/devito/dle/inspection.py
+++ b/devito/dle/inspection.py
@@ -1,0 +1,26 @@
+from collections import OrderedDict
+
+from devito.visitors import FindSections
+
+__all__ = ['retrieve_iteration_tree']
+
+
+def retrieve_iteration_tree(node):
+    """Return a list of all :class:`Iteration` sub-trees rooted in ``node``.
+    For example, given the Iteration tree:
+
+        .. code-block::
+           Iteration i
+             expr0
+             Iteration j
+               Iteraion k
+                 expr1
+             Iteration p
+               expr2
+
+    Return the list: ::
+
+        [(Iteration i, Iteration j, Iteration k), (Iteration i, Iteration p)]
+    """
+
+    return FindSections().visit(node).keys()
--- a/devito/dle/inspection.py
+++ b/devito/dle/inspection.py
@@ -1,5 +1,3 @@
-from collections import OrderedDict
-
 from devito.visitors import FindSections
 
 __all__ = ['retrieve_iteration_tree']
--- a/devito/nodes.py
+++ b/devito/nodes.py
@@ -254,6 +254,14 @@ class Iteration(Node):
         return c.For(loop_init, loop_cond, loop_inc, c.Block(loop_body))
 
     @property
+    def is_Open(self):
+        return self.dim.size is not None
+
+    @property
+    def is_Closed(self):
+        return not self.is_Open
+
+    @property
     def children(self):
         """Return the traversable children."""
         return (self.nodes,)
--- a/devito/nodes.py
+++ b/devito/nodes.py
@@ -14,7 +14,8 @@ from devito.dse.inspection import terminals
 from devito.interfaces import SymbolicData
 from devito.tools import as_tuple, filter_ordered
 
-__all__ = ['Node', 'Block', 'Expression', 'Function', 'Iteration', 'Timer']
+__all__ = ['Node', 'Block', 'Expression', 'Function', 'Iteration', 'List',
+           'TimedList']
 
 
 class Node(object):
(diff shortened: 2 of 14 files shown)


devito/nodes.py
```diff
@@ -80,20 +86,50 @@ def __init__(self, header=None, body=None, footer=None):
...
+    is_List = True
+
+    _wrapper = c.Collection
+
+
+class Element(Node):
+
+    """A generic node that is worth identifying in an Iteration/Expression tree.
+
+    It corresponds to a single :class:`cgen.Statement`.
```
mlange05: In that case why not mirror the naming and call it `Statement`? "Element" is a fairly overloaded term, but I'm really just being pedantic here.

devito/visitors.py
```diff
@@ -162,6 +163,11 @@ def visit_tuple(self, o, ret=None, queue=None):
             ret = self.visit(i, ret=ret, queue=queue)
         return ret
 
+    def visit_Node(self, o, ret=None, queue=None):
+        for i in o.children:
+            ret = self.visit(i, ret=ret, queue=queue)
+        return ret
```
mlange05: This looks odd to me. Don't we need to perform some kind of reduction here over `ret`, or is the return value negligible?

devito/visitors.py
```diff
@@ -179,17 +185,34 @@ def visit_Expression(self, o, ret=None, queue=None):
...
-    """Find all :class:`SymbolicData` and :class:`IndexedData`
-    instances in an Iteration/Expression tree.
+    """Find symbols in an Iteration/Expression tree.
+
+    :param mode: Drive the search for symbols. Accepted values are: ::
+
+        * 'with-data' (default): all :class:`SymbolicData` and :class:`IndexedData`
+          instances are collected.
+        * 'free-symbols': all free symbols appearing in :class:`Expression` are
+          collected.
```
mlange05: Ok, so we technically also return `Dimension` objects to denote open-loop sizes. Admittedly this is a bit hacky, so we should either mention it in the docstring, or maybe even have a special mode for this? I suspect this will get touched up further once we do more blocking/tiling.

devito/visitors.py
```diff
@@ -227,6 +249,40 @@ def visit_Iteration(self, o, found=False, multi=False):
...
+    def default_retval(cls):
+        return cls.Cost(0, 0)
+
+    """
+    Estimate the number of floating point operations and memory accesses per
+    loop iteration in an Iteration/Expression tree.
+    """
+
+    def visit_object(self, o):
+        return self.default_retval()
```
mlange05: Is this not default behaviour? Do we need to make it explicit?

devito/dle/inspection.py
```diff
@@ -0,0 +1,24 @@
...
+                 expr1
+             Iteration p
+               expr2
+
+    Return the list: ::
+
+        [(Iteration i, Iteration j, Iteration k), (Iteration i, Iteration p)]
+    """
+
+    return FindSections().visit(node).keys()
```
FabioLuporini: Agreed, will do in the next PR

devito/visitors.py
FabioLuporini: It should be updated in-place in the children

devito/visitors.py
FabioLuporini: ah right !

devito/visitors.py
FabioLuporini: Agreed

FabioLuporini:
I feel dumb but I don't manage to make Travis happy because of ``isort``.

Travis says: ```$ isort --check-only **/*.py
ERROR: /home/travis/build/opesci/devito/tests/test_visitors.py Imports are incorrectly sorted.```

But when I run the same command on my machine, I get no complaints.

mlange05:
Have you updated `isort` locally?

FabioLuporini:
yes!

mlange05:
Ok, good to go after off-line discussion. Merging...

PR #180, opened 2017-01-31
Outer iteration merge and other niceties

mlange05:
This merge includes a set of utility routines and visitors for the `StencilKernel` that bring us closer to being able to handle the FWI examples. Most notably it includes:
* Pretty-print an AST with a custom visitor. This visitor has a `verbose` mode that:
  * If `False` returns a nicely indented short description string of the loop structure.
  * If `True` incorporates all relevant meta-information about the core node types, allowing us to use it as the base of the kernel hash key.
* A new `MergeOuterIteration` transformer that recursively merges matching `Dimension`s so that we can group expressions under them. This is very useful for source terms.
* Fix to include the zero-offset for each `Dimension` in `Expression.index_offset`, which allows us determine "dimensions used in an expression" for loop insertion.
* As mentioned above, build the hash key from the verbose AST pretty-string.
* Add utility dimension `r` to indicate depth of spatial dimension for point source coordinate fields.

### DIFF
--- a/devito/visitors.py
+++ b/devito/visitors.py
@@ -348,7 +348,7 @@ class ResolveIterationVariable(Transformer):
             # For buffered dimensions insert the explicit
             # definition of buffere variables, eg. t+1 => t1
             init = []
-            for off in [0] + offsets[o.dim]:
+            for off in filter_ordered(offsets[o.dim]):
                 vname = o.dim.get_varname()
                 value = o.dim + off
                 modulo = o.dim.buffered
--- a/devito/visitors.py
+++ b/devito/visitors.py
@@ -190,12 +190,12 @@ class PrintAST(Visitor):
 
     def visit_Block(self, o):
         self._depth += 1
-        header = self.visit(o.header)
-        body = self.visit(o.body)
-        footer = self.visit(o.footer)
+        if self.verbose:
+            body = [self.visit(o.header), self.visit(o.body), self.visit(o.footer)]
+        else:
+            body = [self.visit(o.body)]
         self._depth -= 1
-        return self.indent + "<%s>\n%s" % (o.__class__.__name__,
-                                           '\n'.join([header, body, footer]))
+        return self.indent + "<%s>\n%s" % (o.__class__.__name__, '\n'.join(body))
 
     def visit_Iteration(self, o):
         self._depth += 1
--- a/devito/visitors.py
+++ b/devito/visitors.py
@@ -436,12 +436,20 @@ class MergeOuterIterations(Transformer):
     """
 
     def is_mergable(self, iter1, iter2):
+        """Defines if two :class:`Iteration` objects are mergeable.
+
+        Note: This currently does not(!) consider data dependencies
+        between the loops. A deeper analysis is required for this that
+        will be added soon.
+        """
         return (iter1.dim == iter2.dim and
                 iter1.index == iter2.index and
                 iter1.limits == iter2.limits)
 
     def merge(self, iter1, iter2):
-        # Create a new merged loop
+        """Creates a new merged :class:`Iteration` object from two
+        loops along the same dimension.
+        """
         newexpr = iter1.nodes + iter2.nodes
         return Iteration(newexpr, dimension=iter1.dim,
                          limits=iter1.limits)
--- a/devito/visitors.py
+++ b/devito/visitors.py
@@ -22,8 +22,8 @@ __all__ = ['EstimateCost', 'FindSections', 'FindSymbols',
            'ResolveIterationVariable', 'Transformer', 'printAST']
 
 
-def printAST(node):
-    return PrintAST().visit(node)
+def printAST(node, verbose=True):
+    return PrintAST(verbose=verbose).visit(node)
 
 
 class Visitor(object):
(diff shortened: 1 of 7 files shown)


devito/visitors.py
```diff
@@ -143,6 +149,51 @@ def always_rebuild(self, o, *args, **kwargs):
         return o._rebuild(*new_ops, **okwargs)
 
 
+class PrintAST(Visitor):
```
FabioLuporini: Can we add a default visit_Node that at least does proper indentation and uses the class name as return value ?

devito/visitors.py
```diff
@@ -143,6 +149,51 @@ def always_rebuild(self, o, *args, **kwargs):
...
+
+    _depth = 0
```
FabioLuporini: This should probably be an object attribute

FabioLuporini: Actually, you might just get rid of it and simply propagate the depth as a parameter along the call tree

devito/visitors.py
```diff
@@ -363,9 +414,55 @@ def visit_Iteration(self, o, subs={}, offsets={}):
...
+    def is_mergable(self, iter1, iter2):
+        return (iter1.dim == iter2.dim and
+                iter1.index == iter2.index and
+                iter1.limits == iter2.limits)
+
+    def merge(self, iter1, iter2):
+        # Create a new merged loop
+        newexpr = iter1.nodes + iter2.nodes
+        return Iteration(newexpr, dimension=iter1.dim,
+                         limits=iter1.limits)
```
FabioLuporini: offsets too? Probably just use ``args_frozen``

devito/visitors.py
```diff
@@ -363,9 +414,55 @@ def visit_Iteration(self, o, subs={}, offsets={}):
...
+class MergeOuterIterations(Transformer):
+    """
+    :class:`Transformer` that merges subsequent :class:`Iteration`
+    objects iff their dimenions agree.
+    """
+
+    def is_mergable(self, iter1, iter2):
+        return (iter1.dim == iter2.dim and
+                iter1.index == iter2.index and
+                iter1.limits == iter2.limits)
```
FabioLuporini: offsets too?

devito/visitors.py
```diff
@@ -363,9 +414,55 @@ def visit_Iteration(self, o, subs={}, offsets={}):
...
-        # since we're mapping into a list...
-        offsets.update(o.index_offsets)
+    def visit_Expression(self, o, subs={}, offsets=defaultdict(set)):
+        """Collect all offsets used with a dimension"""
+        for dim, offs in o.index_offsets.items():
+            offsets[dim].update(offs)
         return o
+
+
+class MergeOuterIterations(Transformer):
```
FabioLuporini: Using this visitor is legal as long as there are no particular dependencies across the merged expressions. We know this is always true for us... see below for more info about this

tests/test_visitors.py
```diff
@@ -205,3 +237,73 @@ def test_transformer_add_replace(exprs, block2, block3):
...
+
+def test_merge_iterations_flat(exprs, iters):
+    """Test outer loop merging on a simple two-level hierarchy:
+
+    for i                       for i
+        for j              \        for j
+            expr0       === \           expr0
+    for i               === /       for k
+        for k              /            expr1
+            expr1
```
FabioLuporini: Note: it's always legal to merge these two trees as long as ``expr0`` is not a reduction along ``for i`` *AND* ``expr0.lhs`` doesn't appear in ``expr1.rhs``. By reduction "along for i" I mean something like:

```
for i
  for j
    a[j] += c[j]*3
```

The DSE provides the ``TemporariesGraph``, an object that could be used to this purpose. I'd leave a comment in ``StencilKernel`` to remember to introduce this check prior to merging.

mloubout:
This looks great and should allow to completely remvo the *** `damp` field to replace it by a `boundary` class.

devito/visitors.py
mlange05: `offsets` are used to construct `limits` so they are implied - but yes, I should leave a comment to explain that.

devito/visitors.py
mlange05: Yes, in fact we need to also pick up generic `cgen.Generable` objects here to make the `Block` printing nice. I'll add that...

FabioLuporini: The printing of ``cgen`` object should probably only happen in verbose mode (or equivalent).  The most useful thing, I think, is to see the high level tree structure without the "cgen pollution". For example, think about TTI: you can have thousands cgen statements, and easily lose the focus of what's happening if we print *everything*

tests/test_visitors.py
FabioLuporini: @mlange05 Sorry, I explained myself really badly: what I meant was to leave a TODO comment in ``StencilKernel``, so that we remember to deal with this in the near future *after* this PR has been merged !

mlange05: Yes, I know. In fact I'll probably file an issue for this to track it better.

devito/visitors.py
FabioLuporini: Ah I see

PR #182, opened 2017-02-02
FindSymbols: Filter symbol names and sort them

mlange05:
This merge should fix the spurious breakages in current master. 
It adds a new utility routine `filter_sorted()` that behaves like
`filter_ordered` but also enforces sorting according to a key.


FabioLuporini:
OK

mlange05:
This _should_ fix the `master` build, so it goes in first. Merging...

PR #183, opened 2017-02-03
DLE: OpenMP and SIMD

FabioLuporini:
This PR mainly introduces two transformation steps in the DLE:
- Generation of pragmas for OpenMP parallelism
- Generation of pragmas to trigger compiler auto-vectorization

The OpenMP parallelization looks way simpler than that in the Propagator, but it should be effective in exactly the same way. I talked to @pvelesko , and it turns out the original OpenMPization through the propagator had been done by a number of people, or perhaps mostly by @navjotk ? I'm interested in your thoughts.

Some generated code follows, from our Diffusion example:
```
#include "stdlib.h"
#include "math.h"
#include "sys/time.h"
#include "xmmintrin.h"
#include "pmmintrin.h"

struct profiler
{
  double loop_t_0;
} ;

int Kernel(const int t_size, float *u_vec, const int x_size, const int y_size, struct profiler *timings)
{
  float (*u)[x_size][y_size] = (float (*)[x_size][y_size]) u_vec;
  #pragma omp parallel
  {
    /* Flush denormal numbers to zero in hardware */
    _MM_SET_DENORMALS_ZERO_MODE(_MM_DENORMALS_ZERO_ON);
    _MM_SET_FLUSH_ZERO_MODE(_MM_FLUSH_ZERO_ON);
  }
  {
    struct timeval start_loop_t_0, end_loop_t_0;
    gettimeofday(&start_loop_t_0, NULL);
    for (int t = 0; t < t_size - 1; t += 1)
    {   
      int t0 = (t) % 2;
      int t1 = (t + 1) % 2;
      {   
        #pragma omp parallel for schedule(static)
        for (int x0 = 1; x0 < x_size - 1; x0 += 1)
        {   
          #pragma ivdep
          for (int y0 = 1; y0 < y_size - 1; y0 += 1)
          {   
            u[t1][x0][y0] = 2.5e-1F*(u[t0][x0][y0 - 1] + u[t0][x0][y0 + 1] + u[t0][x0 - 1][y0] + u[t0][x0 + 1][y0]);
          }   
        }
      }
    }
    gettimeofday(&end_loop_t_0, NULL);
    timings->loop_t_0 += (double)(end_loop_t_0.tv_sec-start_loop_t_0.tv_sec)+(double)(end_loop_t_0.tv_usec-start_loop_t_0.tv_usec)/1000000;
  }
  return 0;
}
```

Intel's ``#pragma ivdep`` does suffice to trigger auto-vectorization (it also speeds up compilation time over ``#pragma simd``). If another compiler is used instead, the standard ``#pragma omp simd`` is generated

### DIFF
--- a/devito/dle/transformer.py
+++ b/devito/dle/transformer.py
@@ -14,7 +14,7 @@ from devito.dle.manipulation import compose_nodes
 from devito.dse import terminals
 from devito.interfaces import ScalarData, SymbolicData
 from devito.logger import dle, dle_warning
-from devito.nodes import Element, Function, Iteration, List
+from devito.nodes import Denormals, Element, Function, Iteration, List
 from devito.tools import as_tuple, flatten
 from devito.visitors import FindSections, FindSymbols, IsPerfectIteration, Transformer
 
@@ -31,7 +31,8 @@ def transform(node, mode='basic'):
                  are composed. Available modes are: ::
 
                     * 'noop': Do nothing [S]
-                    * 'advanced': Split elemental functions and apply loop blocking [S]
+                    * 'advanced': Add code to avoid denormal numbers, split elemental
+                                  functions, apply loop blocking [S]
                     * 'blocking': Apply loop blocking [T]
                     * 'split': Identify and split elemental functions [T]
 
@@ -109,12 +110,15 @@ class State(object):
 
         self.elemental_functions = ()
         self.arguments = ()
+        self.includes = ()
 
-    def update(self, nodes=None, elemental_functions=None, arguments=None):
+    def update(self, nodes=None, elemental_functions=None, arguments=None,
+               includes=None):
         self.nodes = as_tuple(nodes) or self.nodes
         self.elemental_functions = as_tuple(elemental_functions) or\
             self.elemental_functions
         self.arguments += as_tuple(arguments)
+        self.includes += as_tuple(includes)
 
 
 class Arg(object):
@@ -155,6 +159,7 @@ class BlockingArg(Arg):
 class Rewriter(object):
 
     triggers = {
+        '_avoid_denormals': ('advanced',),
         '_create_elemental_functions': ('split', 'advanced',),
         '_loop_blocking': ('blocking', 'advanced')
     }
@@ -170,6 +175,7 @@ class Rewriter(object):
 
         self._analyze_and_decorate(state)
 
+        self._avoid_denormals(state, mode=mode)
         self._create_elemental_functions(state, mode=mode)
         self._loop_blocking(state, mode=mode)
 
@@ -237,6 +243,17 @@ class Rewriter(object):
                 nodes = transformer.visit(nodes)
 
         state.update(nodes=nodes)
+
+    @dle_transformation
+    def _avoid_denormals(self, state, **kwargs):
+        """
+        Introduce nodes in the Iteration/Expression tree that will generate macros
+        to avoid computing with denormal numbers. These are normally flushed away
+        when using SSE-like instruction sets in a complete C program, but when
+        compiling shared objects specific instructions must instead be inserted.
+        """
+        return {'nodes': (Denormals(),) + state.nodes,
+                'includes': ('xmmintrin.h', 'pmmintrin.h')}
 
     @dle_transformation
     def _create_elemental_functions(self, state, **kwargs):
--- a/devito/dle/transformer.py
+++ b/devito/dle/transformer.py
@@ -21,7 +21,7 @@ from devito.visitors import (FindNodeType, FindSections, FindSymbols,
                              IsPerfectIteration, Transformer)
 
 
-def transform(node, mode='basic'):
+def transform(node, mode='basic', compiler=None):
     """
     Transform Iteration/Expression trees to generate highly optimized C code.
 
@@ -37,6 +37,7 @@ def transform(node, mode='basic'):
                                   functions, apply loop blocking [S]
                     * 'blocking': Apply loop blocking [T]
                     * 'split': Identify and split elemental functions [T]
+                    * 'simd': Add pragmas to trigger compiler auto-vectorization [T]
 
                  If ``mode`` is a tuple, the last entry may be used to provide optional
                  arguments to the DLE transformations. Accepted key-value pairs are: ::
@@ -49,6 +50,8 @@ def transform(node, mode='basic'):
                                     force the blocking of this loop, the ``blockinner``
                                     flag should be set to True.
                     * 'openmp': True to emit OpenMP code, False otherwise.
+    :param compiler: Compiler class used to perform JIT compilation. Useful to
+                     introduce compiler-specific vectorization pragmas.
     """
 
     if isinstance(node, Sequence):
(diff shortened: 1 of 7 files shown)


navjotk:
I had a look at the current OpenMP use in master - while I did the first version, the current version is doing more. I am not even sure I can argue for the correctness of the generated code (in master) right now (see note below). 

The main difference that this implementation achieves w.r.t. how this is being done currently is in removing the `#pragma omp master` clauses. The reasoning for this was twofold:
1. We only want to measure timings on a single thread so as to minimise the impact the act of observation has on the phenomenon we are observing. 
2. We want to record the 'before' and 'after' times on the same thread. 

Since the easiest way to identify a single unique thread is the `#pragma omp master` clause, it was used here. 

Is there something new in OpenMP that makes this unnecessary?

Note: The current (in master) code uses `#pragma omp single` in some places where, in my opinion, it should have been `#pragma omp master`. This is the reason I am unsure about the correctness of the timings this produces.

devito/nodes.py
```diff
@@ -415,6 +415,20 @@ def name(self):
...
 
 
+class Denormals(Block):
+
+    """Macros to make sure denormal numbers are flushed in hardware."""
+
+    def __init__(self, header=None, body=None, footer=None):
+        b = [Element(c.Comment('Flush denormal numbers to zero in hardware')),
+             Element(c.Statement('_MM_SET_DENORMALS_ZERO_MODE(_MM_DENORMALS_ZERO_ON)')),
+             Element(c.Statement('_MM_SET_FLUSH_ZERO_MODE(_MM_FLUSH_ZERO_ON)'))]
```
navjotk: Can we just call these functions directly from python (`ctype`), instead of adding them to the code every time it is generated?

Not sure which would be considered better.

devito/dle/transformer.py
```diff
@@ -438,6 +481,71 @@ def heuristic(dim_size):
...
+                # Heuristic: if at least two parallel loops are available and the
+                # physical core count is greater than self.thresholds['collapse'],
+                # then omp-collapse the loops
+                if psutil.cpu_count(logical=False) < self.thresholds['collapse'] or\
+                        len(candidates) < 2:
+                    n = candidates[0]
+                    mapper[n] = List(header=omplang['par-for'], body=n)
+                else:
+                    nodes = candidates[:2]
+                    mapper.update({n: List(header=omplang['par-for-collapse2'], body=n)
```
navjotk: Is this still the best choice when number of parallel loops available >= 3, or is there an implied assumption here that number of parallel loops <= 2 ?

mlange05:
Code itself looks good to me. Are we ready to merge this?

devito/dle/transformer.py
FabioLuporini: It's an implied assumption that the innermost loop won't be blocked because it will be used as vector dimension, but good catch

devito/dle/transformer.py
```diff
@@ -447,3 +555,23 @@ def _summary(self, mode):
...
+    'par-for': c.Pragma('omp parallel for schedule(static)'),
+    'par-for-collapse2': c.Pragma('omp parallel for collapse(2) schedule(static)'),
+    'simd-for': c.Pragma('omp simd')
+}
+
+"""
+Compiler-specific language
+"""
+complang_ALL = {
+    'IntelCompiler': {'ignore-deps': c.Pragma('ivdep')}
```
FabioLuporini: that's actually the reason I preferred to have a separate dict. In theory, the dict here has everything concerning the language, while the dict in ``IntelCompiler`` contains all of the compilation flags. So, slightly different distinction, but debatable

devito/nodes.py
FabioLuporini: no, because apparently these need be re-hit any time a new parallel region is created (and threads re-created/re-initialised). Not sure why that is the case. Also, these need to be called by **each** thread, while in Python only the master thread would hit them

FabioLuporini:
> Do you think we need tests for this?

Basically this is already tested since travis also tries the case OMP_NUM_THREADS>1 , which triggers the corresponding openmp additions

> I see that the generated code is calling a function in the innermost loop. I suppose this is what you call elemental functions in the code. I probably missed the PR that introduced this. Can you point me to the data that showed that this improves performance?

The elemental functions are just there to improve compilation speed, because with tiling you have remainder loops, and having the huge stencil (eg tti) appearing multiple times in a deep loop nest simply makes the compilation time blow up. I did see no regressions from introducing them. However, in an upcoming PR, I also introduced a threshold to trigger the elemental functions pass iff the number of statements in the body is above a certain value (will be determined empirically trying TTI)

FabioLuporini:
> Code itself looks good to me. Are we ready to merge this?

If @navjotk is happy with my replies, this is good to go. I took note of the minor observations he rightly made, and I will deal with them once the bulk of Devito 3.0 is in.

devito/dle/transformer.py
navjotk: But I am talking about the case where we have more than 4 loops in total (the `StencilKernel` does not assume any maximum number of loops so it could generate a 10 loop nest in theory). 
Even though the innermost loop is being vectorised, we might have more loops available to be parallelised and collapsed. 

So I believe that the `StencilKernel` does not treat the 4 loop case as special. I am asking whether it is the same with this bit of code as well?

devito/dle/transformer.py
navjotk: Keeping all compiler specific settings in a single file/class has the added advantage that a community member can contribute an "adapter" for another compiler without having to understand the complete logical flow to realise that compiler specific code actually goes in two places, not one.

devito/nodes.py
navjotk: > these need to be called by each thread, while in Python only the master thread would hit them

This makes sense

> these need be re-hit any time a new parallel region is created (and threads re-created/re-initialised)

This doesn't, to me. The generated code usually has 2 or more parallel regions while these functions are only being called once in the beginning.

navjotk:
> Basically this is already tested since travis also tries the case OMP_NUM_THREADS>1 , which triggers the corresponding openmp additions

So do you believe that the existing tests are sufficient? I believe we only have regression-like tests and no unit tests that test the transformer you are adding in this PR.

devito/nodes.py
FabioLuporini: already fixed in an upcoming PR. I only recently found out about that

navjotk:
> The elemental functions are just there to improve compilation speed, because with tiling you have remainder loops, and having the huge stencil (eg tti) appearing multiple times in a deep loop nest simply makes the compilation time blow up. I did see no regressions from introducing them. However, in an upcoming PR, I also introduced a threshold to trigger the elemental functions pass iff the number of statements in the body is above a certain value (will be determined empirically trying TTI)

I'd like to see the data from the performance comparison that you carried out, when you have the time.  </OT>

FabioLuporini:
> So do you believe that the existing tests are sufficient? I believe we only have regression-like tests and no unit tests that test the transformer you are adding in this PR.

what I meant is that the transformer pass gets called anytime with OMP_NUM_THREADS>1, so it is tested

question is: is it undertested? maybe, maybe just like the other passes... I'll write more tests in the future

I'm trying to find a good compromise between quality, maintainability, extensibility, modularity and...me that needs to run tests ASAP

FabioLuporini:
> I'd like to see the data from the performance comparison that you carried out, when you have the time.

okay, not now.

So we agree this PR stays around until I find the time to do this

FabioLuporini:
> I'd like to see the data from the performance comparison that you carried out, when you have the time.
> 
> okay, so we agree this PR stays around until I find the time to do this

Unless the following numbers from a very specific case satisfy you. This is acoustic forward, space order 2, time order 2, 3D grid 256**3. Average of 3 runs, variance negligible:

1.43 seconds **w/** elemental functions
1.42 seconds **w/o** elemental functions

navjotk:
> Average of 3 runs, variance negligible
As an aside, it is considered more resilient to report the minimum run time from a set of runs, not the average.

FabioLuporini:
You're right. So minimum runtimes follow:

1.42 seconds **w/** elemental functions
1.42 seconds **w/o** elemental functions

PR #184, opened 2017-02-06
Devito3.0 further improvements

FabioLuporini:
This PR introduces:
- a new transformation step in the Devito Loop Engine to emit nontemporal stores
- a simple auto-tuning system through ``StencilKernel`` that doesn't depend on the old AT system
- extensions to benchmark.py

This PR should go in **after** #183

### DIFF
--- a/devito/stencilkernel.py
+++ b/devito/stencilkernel.py
@@ -16,7 +16,7 @@ from devito.dimension import Dimension
 from devito.dle import transform
 from devito.dse import indexify, rewrite
 from devito.interfaces import SymbolicData
-from devito.logger import error, info
+from devito.logger import error, info, warning
 from devito.nodes import Block, Expression, Function, Iteration, TimedList
 from devito.profiler import Profiler
 from devito.tools import as_tuple
@@ -187,6 +187,17 @@ class StencilKernel(Function):
         for d in d_args:
             arguments[d.name] = dim_sizes[d]
 
+        # Retrieve the data type of the arrays
+        try:
+            dtypes = [i.data.dtype for i in f_args]
+            dtype = dtypes[0]
+            if any(i != dtype for i in dtypes):
+                warning("Found non-matching data types amongst the provided"
+                        "symbolic arguments.")
+            dtype_size = dtype.itemsize
+        except IndexError:
+            dtype_size = 1
+
         # Add profiler structs
         if self.profiler:
             cpointer = self.profiler.as_ctypes_pointer(Profiler.TIME)
@@ -196,7 +207,7 @@ class StencilKernel(Function):
         self.cfunction(*list(arguments.values()))
 
         # Summary of performance achieved
-        info("="*71)
+        info("="*79)
         for itspace, profile in self.sections.items():
             # Time
             elapsed = self.profiler.timings[profile.timer]
@@ -204,12 +215,12 @@ class StencilKernel(Function):
             niters = reduce(operator.mul, [i.size or dim_sizes[i] for i in itspace])
             flops = float(profile.ops*niters)
             gflops = flops/10**9
-            # Memory (FIXME: need to tweak the calculation below once padding is in)
-            traffic = profile.memory*niters
+            # Compulsory traffic
+            traffic = profile.memory*niters*dtype_size
 
             info("Section %s with OI=%.2f computed in %.2f s [Perf: %.2f GFlops/s]" %
                  (str(itspace), flops/traffic, elapsed, gflops/elapsed))
-        info("="*71)
+        info("="*79)
 
     @property
     def _cparameters(self):
--- a/devito/stencilkernel.py
+++ b/devito/stencilkernel.py
@@ -20,7 +20,7 @@ from devito.logger import error, info, warning
 from devito.nodes import Block, Expression, Function, Iteration, TimedList
 from devito.profiler import Profiler
 from devito.tools import as_tuple
-from devito.visitors import (EstimateCost, FindSections, FindSymbols,
+from devito.visitors import (EstimateCost, FindNodeType, FindSections, FindSymbols,
                              IsPerfectIteration, MergeOuterIterations,
                              ResolveIterationVariable, SubstituteExpression,
                              Transformer, printAST)
@@ -29,6 +29,11 @@ __all__ = ['StencilKernel']
 
 
 class StencilKernel(Function):
+
+    """
+    Cache of auto-tuned StencilKernels.
+    """
+    _AT_cache = {}
 
     """A special :class:`Function` to evaluate stencils through just-in-time
     compilation of C code.
@@ -136,6 +141,9 @@ class StencilKernel(Function):
         if len(args) <= 0:
             args = self.parameters
 
+        # Perform auto-tuning if the user requests it and loop blocking is in use
+        maybe_autotune = kwargs.get('autotune', False)
+
         # Map of required arguments and actual dimension sizes
         arguments = OrderedDict([(arg.name, arg) for arg in self.parameters])
         dim_sizes = {}
@@ -169,7 +177,7 @@ class StencilKernel(Function):
                         dim_sizes[dim] = data.shape[i]
                 else:
                     assert dim.size == data.shape[i]
-        # Add user-provided block sizes, if any
+        # Add user-provided loop blocking sizes (if any)
         dle_arguments = OrderedDict()
         for i in self._dle_state.arguments:
             dim_size = dim_sizes.get(i.original_dim, i.original_dim.size)
@@ -179,6 +187,8 @@ class StencilKernel(Function):
                     dle_arguments[i.argument] = i.value(dim_size)
                 except TypeError:
                     dle_arguments[i.argument] = i.value
+                    # User-provided block size available, do not autotune
+                    maybe_autotune = False
             else:
                 dle_arguments[i.argument] = dim_size
         dim_sizes.update(dle_arguments)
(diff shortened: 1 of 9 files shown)


navjotk:
I don't think `StencilKernel` is the right place to do autotuning. Block sizes are only the first of a large number of auto-tunable parameters. 

1. It is a violation of Separation of Concerns - the job of the `StencilKernel` is to accept Stencils and generate code. 
2. Cache block sizes are the first of many parameters that would need to be autotuned. The AT infrastructure would benefit from being in a common place e.g. implementing different search techniques like gradient descent or a stochastic search algorithm.

FabioLuporini:
You're probably right. I think we can move this to a separate module once we have something better than "block size" + brute force. Even better, as soon as we have a backend doing autotuning for us...

What other parameters do you have in mind that might benefit from tuning? I can think of different combinations of DLE transformations, but again, ``StencilKernel``, or the user, drives the DLE (ie what transformations to apply).

navjotk:
We probably want to do domain decomposition soon at the devito level, i.e. a single shot split across multiple MPI ranks. Deciding the ideal split between ranks is an auto-tuning problem too, I believe.

FabioLuporini:
> We probably want to do domain decomposition soon at the devito level, i.e. a single shot split across multiple MPI ranks. Deciding the ideal split between ranks is an auto-tuning problem too, I believe.

Do you expect uniform distribution to be way sub-optimal (why?), is this based on some ideas that you have in mind, or what else? </ OT>

navjotk:
I think we are moving tangentially from the objective of this thread and the PR. My original concerns about this being a violation of Separation of Concerns are still unanswered. 

However, answering your question:
Some reasons that uniform splits aren't the best:
1. Oversplitting - We need to be doing enough computations to overlap with communications so splitting a 100 point domain on to 100 nodes would mean all nodes spend more time waiting for their neighbours to complete computation than actually doing their own computation.
2. Different zones in the velocity model - This one's largely a guess, correct me if I am wrong. There might be some zones of the velocity model where the parameters are unchanging, while in some other zones they vary enormously. In the zones where the parameters are staying constant, I expect low level optimisations to kick in where the stencil values aren't actually being fetched from memory, but simply from cache/registers. The computations in these zones can be expected to be completed much more quickly than in the zones with large variations where large memory traffic is required and expected. This sort of thing is especially important for TTI where we have the anisotropy parameters apart from just the velocity model.

FabioLuporini:
> I think we are moving tangentially from the objective of this thread and the PR. My original concerns about this being a violation of Separation of Concerns are still unanswered.

I'm happy to leave this PR dangling until I have performance results that I can show in Atlanta. Until then, I can't spend time on this matter. I will address it, just not this week. Although this is not necessarily a PR-stopper, I'm totally fine with this. (the other PR can probably go in, though)

> Oversplitting - We need to be doing enough computations to overlap with communications so splitting a 100 point domain on to 100 nodes would mean all nodes spend more time waiting for their neighbours to complete computation than actually doing their own computation.

I'm not sure this needs an AT -- feels more like a sizing problem

> Different zones in the velocity model - This one's largely a guess, correct me if I am wrong. There might be some zones of the velocity model where the parameters are unchanging, while in some other zones they vary enormously. In the zones where the parameters are staying constant, I expect low level optimisations to kick in where the stencil values aren't actually being fetched from memory, but simply from cache/registers. The computations in these zones can be expected to be completed much more quickly than in the zones with large variations where large memory traffic is required and expected. This sort of thing is especially important for TTI where we have the anisotropy parameters apart from just the velocity model.

This was my hypothesis too until I found out about denormal numbers (see Seigen channel on Slack for info). This is just to say that before committing to implementing fancy AT systems, we should first verify that our problems do suffer from problems like load-imbalance.

navjotk:
> I'm happy to leave this PR dangling until I have performance results that I can show in Atlanta. Until then, I can't spend time on this matter. I will address it, just not this week. Although this is not necessarily a PR-stopper, I'm totally fine with this. (the other PR can probably go in, though)

I think this violates the [software design principles](https://en.wikipedia.org/wiki/Single_responsibility_principle) that we decided to follow this time. If you think I am mistaken, ignore me. 

> I'm not sure this needs an AT -- feels more like a sizing problem

I am not sure I understand what sizing problem you refer to here. If I understand correctly, the ideal size would be different on different systems for different stencils. Are there any ways to reliably calculate that without actually running the program with different sizes?

> This was my hypothesis too until I found out about denormal numbers (see Seigen channel on Slack for info).

I read it again right now and I am not sure what denormal numbers have to do with the current discussion. If anything, they only add weight to my argument that the values in the input matrices of a stencil have a direct impact on the execution time. Hence denormal values in one region of the model would lead to a load imbalance if the split was made uniformly across dimensions.

> This is just to say that before committing to implementing fancy AT systems, we should first verify that our problems do suffer from problems like load-imbalance.

I am not sure that making our code modular enough to make it easy to extend/change the implementation later counts as committing to implementing fancy AT systems. If anything, the additional modularity makes it easy for us to test hypotheses like the above.

devito/stencilkernel.py
```diff
@@ -116,12 +119,16 @@ def __init__(self, stencils, **kwargs):
         nodes = SubstituteExpression(subs=subs).visit(nodes)
 
         # Apply the Devito Loop Engine for loop optimization and finalize instantiation
-        dle_state = transform(nodes, mode=dle)
+        dle_state = transform(nodes, mode=set_dle_mode(dle, self.compiler),
+                              compiler=self.compiler)
         body = dle_state.nodes
```
mloubout: I would put 
```
parameters = FindSymbols().visit(nodes)
parameters += [i.argument for i in dle_state.arguments]
```
in a `get_parameter(self, nodes, dle_state)` finction an replace by `parameters = kwrgs.get("parameters", get_parameter(nodes, dle_stat)`

FabioLuporini:
> I think this violates the software design principles that we decided to follow this time. If you think I am mistaken, ignore me.

I agree AT maybe should live in its own class, perhaps taking as input a ``StencilKernel``. I was merely saying that the current ``_autotune`` method is actually very short and simple, so it might have been considered a non-blocker for the PR. But, as I said, I understand your point. You'll probably just have to wait until next week, when I'll have more time to work on this kind of engineering things. Otherwise, I'm happy if you wish to take the lead!

> I am not sure I understand what sizing problem you refer to here. If I understand correctly, the ideal size would be different on different systems for different stencils. Are there any ways to reliably calculate that without actually running the program with different sizes?

I meant that if the user knows the problem size, it should be almost straightforward to determine the approximate number of nodes that you need to run it. Otherwise one strong-scales to find this...and you usually you've got to strong scale anyway. Auto-tuning oversplitting at the MPI level sounds like an overkill, although I might just be completely wrong.

> I read it again right now and I am not sure what denormal numbers have to do with the current discussion. If anything, they only add weight to my argument that the values in the input matrices of a stencil have a direct impact on the execution time. Hence denormal values in one region of the model would lead to a load imbalance if the split was made uniformly across dimensions.

My hypothesis was that load imbalance could arise because, as you also said, some sections of the grid would have taken longer to compute. To support this hypothesis, I said that "if values don't change, it *might* be that you don't have to load them from memory". Then I realised that the slowing factor was denormal numbers, not fewer load/stores. But the denormal numbers problem now has gone. So, in a stencil code like ours, assuming uniform distrubution, I don't know what could go so wrong to end up with a crazy load imbalance. 

> I am not sure that making our code modular enough to make it easy to extend/change the implementation later counts as committing to implementing fancy AT systems. If anything, the additional modularity makes it easy for us to test hypotheses like the above.

Agreed.

devito/stencilkernel.py
FabioLuporini: In what case would that be useful? 

Note that ``parameters`` is the list of the C function parameters (ie, those used for code generation)

FabioLuporini:
The latest push should bring Travis in green land

navjotk:
> I meant that if the user knows the problem size, it should be almost straightforward to determine the approximate number of nodes that you need to run it. Otherwise one strong-scales to find this...and you usually you've got to strong scale anyway. Auto-tuning oversplitting at the MPI level sounds like an overkill, although I might just be completely wrong.

The ideal problem size depends on the amount of computation that you can perfectly overlap with communication. Since communication is such a fickle aspect to predict, any theoretical estimates would be off by a huge factor. The only way I see around this is to run with a few different split sizes (maybe around a theoretically calculated number) and choose whichever performs best. Isn't that exactly what autotune is for? 

> To support this hypothesis, I said that "if values don't change, it might be that you don't have to load them from memory". Then I realised that the slowing factor was denormal numbers, not fewer load/stores.

So denormal numbers can be one reason for a load imbalance across the grid. That still leaves room for other causes of load imbalance, like the one I pointed out earlier.

devito/stencilkernel.py
mloubout: In every case where devito is not run from python such as our UBC Julia interface. Can't write any `ccall` if you don';t know the argument list, so specifying it makes is easier

mlange05:
FWIW, the diff to #183 looks good to me. The one thing I'd point out to keep in mind is potential cross-fire with the old AT that wraps the `Operator`/`StencilKernel`, but that is an integration problem for another day. It'd be interesting to see if the outputs agree through. Good to go from me once #183 is in.

devito/stencilkernel.py
mlange05: Yes, we should provide an explicit interface for external codes to retrieve and set the parameter order (we don't care about the ordering internally). I'll file a separate issue for this and make sure we address it before unleashing 3.0 to the wider public.

PR #187, opened 2017-02-08
Tweaks, hacks and fixes to make Acoustic-Forward run with Devito-3.0

mlange05:
**WARNING: This merge contains temporary hacks that are necessary due to legacy restrictions. It also attempts to clean a few things for acoustic test, but it does not attempt a full clean-up of our seismic examples. A full clean-up of the examples and a more rigorous generalisation of the temporary hacks will be attempted soon.**

So, this PR really contains a series of "as-clean-as-currently-possible" changes and fixes to adjust the Acoustic Forward runs to the 3.0 API. It introduces a `legacy` flag on the acoustic examples classes that allows us to switch between the two modes seamlessly and thus compare and contrast. The 3.0 route is not exhaustively tested yet, but it should allow us to run preliminary performance benchmarks to get an overview of how 3.0 will do and fix issues before they arise.

In particular we do:
* New and improved treatment of buffered dimensions (eg. time)
* Lift src/rec and wavefield creation to the `Acoustic_cg` wrappers to avoid call signature clashes
* Minor clean up of `test_adjoint` and `test_gradient`.
* Lift necessary time substitutions to the wrapper layer
* Tweaks to the interpolation symbolics

### DIFF
--- a/devito/stencilkernel.py
+++ b/devito/stencilkernel.py
@@ -96,9 +96,12 @@ class StencilKernel(Function):
         body = dle_state.nodes
         parameters = FindSymbols('with-data').visit(nodes)
         parameters += [i.argument for i in dle_state.arguments]
+
         # Add all dimensions used in expressions symbols to arguments.
         # This is required to ensure that we can safely perform data casts.
         dimensions = FindSymbols('dimensions').visit(nodes)
+        # For buffered dimensions, ensure the parent is also present
+        dimensions += [d.parent for d in dimensions if d.is_Buffered]
         parameters += filter_ordered([d for d in dimensions if d.size is None],
                                      key=operator.attrgetter('name'))
         super(StencilKernel, self).__init__(name, body, 'int', parameters, ())
@@ -138,12 +141,21 @@ class StencilKernel(Function):
             for i, dim in enumerate(f.indices):
                 # Infer open loop limits
                 if dim.size is None:
+                    # First, try to find dim size in kwargs
+                    if dim.name in kwargs:
+                        dim_sizes[dim] = kwargs[dim.name]
+
                     if dim in dim_sizes:
                         # Ensure size matches previously defined size
-                        assert dim_sizes[dim] == data.shape[i]
+                        if not dim.is_Buffered:
+                            assert dim_sizes[dim] == data.shape[i]
                     else:
                         # Derive size from grid data shape and store
                         dim_sizes[dim] = data.shape[i]
+
+                    # Ensure parent for buffered dims is defined
+                    if dim.is_Buffered and dim.parent not in dim_sizes:
+                        dim_sizes[dim.parent] = dim_sizes[dim]
                 else:
                     if not isinstance(dim, BufferedDimension):
                         assert dim.size == data.shape[i]
--- a/devito/stencilkernel.py
+++ b/devito/stencilkernel.py
@@ -94,8 +94,13 @@ class StencilKernel(Function):
         dle_state = transform(nodes, mode=set_dle_mode(dle, self.compiler),
                               compiler=self.compiler)
         body = dle_state.nodes
-        parameters = FindSymbols().visit(nodes)
+        parameters = FindSymbols('with-data').visit(nodes)
         parameters += [i.argument for i in dle_state.arguments]
+        # Add all dimensions used in expressions symbols to arguments.
+        # This is required to ensure that we can safely perform data casts.
+        dimensions = FindSymbols('dimensions').visit(nodes)
+        parameters += filter_ordered([d for d in dimensions if d.size is None],
+                                     key=operator.attrgetter('name'))
         super(StencilKernel, self).__init__(name, body, 'int', parameters, ())
 
         # DLE might have introduced additional headers
@@ -133,15 +138,7 @@ class StencilKernel(Function):
             for i, dim in enumerate(f.indices):
                 # Infer open loop limits
                 if dim.size is None:
-                    if isinstance(dim, BufferedDimension):
-                        # Check if provided as a keyword arg
-                        size = kwargs.get(dim.name, None)
-                        if size is None:
-                            error("Unknown dimension size, please provide "
-                                  "size via Kernel.apply(%s=<size>)" % dim.name)
-                            raise RuntimeError('Dimension of unspecified size')
-                        dim_sizes[dim] = size
-                    elif dim in dim_sizes:
+                    if dim in dim_sizes:
                         # Ensure size matches previously defined size
                         assert dim_sizes[dim] == data.shape[i]
                     else:
--- a/examples/acoustic/Acoustic_codegen.py
+++ b/examples/acoustic/Acoustic_codegen.py
@@ -77,8 +77,13 @@ class Acoustic_cg(object):
                              save=save, cache_blocking=cache_blocking, dse=dse,
                              compiler=compiler, profile=True, u_ini=u_ini)
 
-        u, rec = fw.apply()
-        return rec.data, u, fw.propagator.gflopss, fw.propagator.oi, fw.propagator.timings
+        if isinstance(fw, StencilKernel):
+            fw.apply()
+            return None, None, None, None, None
+        else:
+            u, rec = fw.apply()
+            return (rec.data, u, fw.propagator.gflopss,
+                    fw.propagator.oi, fw.propagator.timings)
 
     def Adjoint(self, rec, cache_blocking=None):
         adj = AdjointOperator(self.model, self.damp, self.data, self.src, rec,
--- a/examples/acoustic/Acoustic_codegen.py
+++ b/examples/acoustic/Acoustic_codegen.py
@@ -55,11 +55,6 @@ class Acoustic_cg(object):
         # Initialize damp by calling the function that can precompute damping
         damp_boundary(self.damp.data, nbpml)
 
-        if len(self.damp.shape) == 2 and self.src.receiver_coords.shape[1] == 3:
-            self.src.receiver_coords = np.delete(self.src.receiver_coords, 1, 1)
-        if len(self.damp.shape) == 2 and self.data.receiver_coords.shape[1] == 3:
-            self.data.receiver_coords = np.delete(self.data.receiver_coords, 1, 1)
-
         if auto_tuning:  # auto tuning with dummy forward operator
             fw = ForwardOperator(self.model, self.src, self.damp, self.data,
                                  time_order=self.t_order, spc_order=self.s_order,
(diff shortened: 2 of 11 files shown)


devito/stencilkernel.py
```diff
@@ -112,14 +116,25 @@ def __init__(self, stencils, **kwargs):
 
         # Now resolve and substitute dimensions for loop index variables
         subs = {}
-        nodes = ResolveIterationVariable().visit(nodes, subs=subs)
+        variable_map = defaultdict(int)
```
FabioLuporini: since this is not used later, I'd allocate it in the visitor directly

FabioLuporini:
Once #183  and #184 will be in, we might end up with conflicts in `stencilkernel.py`. Not a surprise. It shouldn't be anything serious though. However, my current local branch is based up #184, so I guess we might want to come up with a merge order.

Otherwise looks good to me. As I wrote above, the only thing I noticed is that the presence of ``variable_map`` in ``StencilKernel.__init__``looks quite odd, and should probably rather be moved within the visitor itself.

examples/acoustic/Acoustic_codegen.py
```diff
@@ -55,33 +55,93 @@ def damp_boundary(damp, nbp):
...
-            fw = ForwardOperator(self.model, self.src, self.damp, self.data,
+            nt, nrec = self.data.shape
+            nsrc = self.source.shape[1]
+            ndim = len(self.damp.shape)
+            h = self.model.get_spacing()
+            dtype = self.damp.dtype
+            nbpml = self.model.nbpml
+            dt = self.model.get_critical_dt()
+            if self.t_order == 4:
+                dt *= 1.73
```
navjotk: This logic should probably be moved inside `self.model.get_critical_dt()`, especially since you call it from multiple places.

mlange05:
Yes, this should sit on top of #183 and #184, and I'm happy to do the rebase. In fact there are still some DSE/DLE related breakages that I'm hoping will be fixed when I rebase. I filed the PR because I believe the bulk of the changes is correct now and I wanted the review to get going.

examples/acoustic/Acoustic_codegen.py
```diff
@@ -55,33 +55,93 @@ def damp_boundary(damp, nbp):
...
 
         if auto_tuning:
             cache_blocking = self.at.block_size
-        fw = ForwardOperator(self.model, self.src, self.damp, self.data,
+
+        # Create the forward wavefield
+        u = TimeData(name="u", shape=self.model.get_shape_comp(), time_dim=nt,
+                     time_order=2, space_order=self.s_order, save=save,
+                     dtype=self.damp.dtype)
+        u.pad_time = save
```
navjotk: I suppose you haven't removed my hack related to padding, even though the new version doesn't need it. This might lead to inconsistencies in how results are to be interpreted. I am not sure whether you took care of that.

devito/stencilkernel.py
mlange05: I tried that, but I got some bizarre pseudo-caching that caused breakages and then I ran out of patience. I'll try again and investigate though.

examples/acoustic/fwi_operators.py
```diff
@@ -20,66 +22,69 @@ class ForwardOperator(Operator):
...
-        self.propagator.add_devito_param(source.coordinates)
-        self.propagator.add_devito_param(rec)
-        self.propagator.add_devito_param(rec.coordinates)
+        op.input_params += [src, src.coordinates, rec, rec.coordinates]
+        op.output_params += [rec]
+        op.propagator.time_loop_stencils_a = src.add(m, u) + rec.read(u)
+        op.propagator.add_devito_param(src)
+        op.propagator.add_devito_param(src.coordinates)
+        op.propagator.add_devito_param(rec)
+        op.propagator.add_devito_param(rec.coordinates)
```
navjotk: I don't know why this is being called manually. I believe the Operator does this [automatically](https://github.com/opesci/devito/blob/master/devito/operator.py#L129). Not that it matters at this point.

examples/acoustic/fwi_operators.py
```diff
@@ -20,66 +22,69 @@ class ForwardOperator(Operator):
...
+        # from the Operator/Propagator structure and is used here for
+        # backward compatibiliy. We need re-examine this apporach carefully!
+
+        # Shift time indices so that LHS writes into t only,
+        # eg. u[t+2] = u[t+1] + u[t]  -> u[t] = u[t-1] + u[t-2]
+        stencils = [e.subs(t, t + solve(eqn.lhs.args[0], t)[0])
+                    if isinstance(e.lhs, TimeData) else e
+                    for e in stencils]
+        # Apply time substitutions as per legacy approach
+        time_subs = {t + 2: t + 1, t: t + 2, t - 2: t, t - 1: t + 1, t + 1: t}
```
navjotk: Aren't the above two operations self negating? 

What happens if you just leave both of them out?

examples/acoustic/fwi_operators.py
```diff
@@ -226,7 +231,7 @@ def __init__(self, model, damp, data, recin, u, time_order=2, spc_order=6, **kwa
         # Insert receiver term post-hoc
         self.input_params += [grad, rec, rec.coordinates]
         self.output_params = [grad]
-        self.propagator.time_loop_stencils_b = rec.add(m, v, t + 1)
+        self.propagator.time_loop_stencils_b = rec.add(m, v, u_t=t + 1, p_t=t + 1)
```
navjotk: Not able to understand what is being done here

examples/source_type.py
```diff
@@ -79,37 +80,58 @@ def sym_coord_bases(self):
...
+    @property
+    def index_matrix(self):
+        """Spatial indexing expressions for all grid points
+        contributing to spatial interpolation.
+        """
+        return [tuple([idx + ii + self.nbpml for ii, idx
+                       in zip(inc, self.sym_coord_indices)])
+                for inc in self.increments]
+
+    def point2grid(self, u, m, u_t=t, p_t=t):
```
navjotk: I really hope we have a plan to internalise these interpolations into the library code soon.

examples/acoustic/Acoustic_codegen.py
mlange05: Yes, there are many those things around but I concentrated on the essentials in this PR. I will attempt a throurgh re-design/clean-up of the examples within the next few PRs.

examples/acoustic/Acoustic_codegen.py
mlange05: I can't remove any legacy hacks at this point, since the new version is not fully functional yet. So we have to keep both APIs alive for the moment.

mlange05: Ok, fair enough, I'll try some more low-level verification. Currently I'm not even testing any 3.0 kernels in travis, but we might want to enable that for a select few. Which test would best verify this one?

examples/source_type.py
mlange05: Yes, these need a complete re-think. I'll file an issue with API suggestions.

examples/acoustic/fwi_operators.py
mlange05: Yeah, but the `time_loop_stencils` were except from this, since they follow a slightly different code path than the main stencils.

examples/acoustic/Acoustic_codegen.py
navjotk: I don't think there is a test that explicitly verifies this. The legacy code will put the result of the first time step in `t=0`. Your new code, I believe, will put it in `t=2` for `time_order=2`. 
This will need to be manually verified which parts of the code expect the first timestep to be in `t=0` and which in `t=2`. 
Since the tests run for a small number of time steps, the difference this produces will not show up.

examples/acoustic/fwi_operators.py
```diff
@@ -20,66 +22,69 @@ class ForwardOperator(Operator):
...
+        op.propagator.time_loop_stencils_a = src.add(m, u) + rec.read(u)
+        op.propagator.add_devito_param(src)
+        op.propagator.add_devito_param(src.coordinates)
+        op.propagator.add_devito_param(rec)
+        op.propagator.add_devito_param(rec.coordinates)
+
+    else:
+        # Create stencil expressions for operator, source and receivers
+        eqn = Eq(u.forward, stencil)
+        src_add = src.point2grid(u, m, u_t=t, p_t=time)
```
mlange05: Ok, so `src.point2grid` creates multiple expressions to perform interpolation from `src` to `u`. The time indices in those expressions differ between those expressions because `u` is buffered and thus requires one of the modulo-derived time indices (`t0`, `t1`, `t2`), while `src` actually has an unrolled `time` dimensions (formerly `i3`). Since the `StencilKernel` does no implicit time substitutions under the hood the idea is to explicitly use "real time" (`time`) and "modulo time" (`t`) according to how data is meant to e accessed.

mloubout: This make sense. My question then is, is this only a indexing name or oes it specify which time step to read in source and write in u? Because if it does `u(t) = u(t) + src(t)` it is a bit confusing as it reads like if `u` and `src` are on different time axis.

mlange05: No the time step is explicit. You can just as well use, for example, `u_t=t + 1` and `u_p=time-1` or any other expression you like. I'm hoping to generalise this even further in the coming PRs to separate the rhs expression from the interpolation weights, so that the symbol creation (including time indexing) is left to the user.

mloubout: I also think the'dt**2/m` should be an weight expression input instead of hard coded to enable more general source/receiver expressions,

mlange05: Yes, exactly that's what I'm after. This will follow in one of the next PRs.

examples/acoustic/fwi_operators.py
mlange05: The short answer is: unfortunately no, I get a while bunch of NaNs in the receiver output, and the time indices do not align. A more in-depth analysis of this problem will hopefully follow soon.

mlange05:
Ok, this is now rebased and travis is happy. @navjotk are you happy for this to be merged? I haven't addressed the `u.pad_time` issue yet, since this PR is simply trying to emulate current `Operator`/`Propagator` - redoing this properly will probably be intertwined with sorting out the time indexing hacks.

FabioLuporini:
It's a go for me

navjotk:
I don't understand how you can be sure you are emulating the current behaviour without being sure you are reading the correct time indices. 
If you know you are reading the same elements as before (in time), this is good to go. Without that verification (either automated as a test or atleast manually done once) this could be making all our results invalid and we wouldn't know.

mlange05:
Oh yes, sorry I must have misunderstood you. I have definitely verified that we're accessing the exact same time indices as before (many times even, but by inspection), otherwise `test_adjoint` and others would not work. I've verified this locally for now, because I don't want to overload travis with additional copies of the heavy tests just yet, but they do run. I'll add some more rigorous testing of time indices once we've figured out the right generalisation.

mlange05:
So I guess we're good to go then, so merging...

PR #188, opened 2017-02-08
Devito3.0: more and more features

FabioLuporini:
This PR brings Devito3.0 steps closer to feature completeness.

Support for running several Acoustic operators is introduced. This mainly consists of properly dealing with the DSE-produced temporaries during the various DLE passes as well as for code generation (eg, inferring the right data types). However, to run the Acoustic operators, we'll need #187 .

Performance profiling has been extended, improved, and fixed where necessary (fixes were minor).

On top of all this, a number of improvements of various nature have been introduced.

### DIFF
--- a/devito/stencilkernel.py
+++ b/devito/stencilkernel.py
@@ -264,6 +264,8 @@ class StencilKernel(Function):
         best block sizes (when loop blocking is in use). The block sizes tested
         are those listed in ``options['at_blocksizes']``, plus the case that is
         as if blocking were not applied (ie, unitary block size)."""
+        if not self._dle_state._applied_blocking:
+            return
 
         at_arguments = arguments.copy()
 
--- a/devito/stencilkernel.py
+++ b/devito/stencilkernel.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import
 
 import operator
-from collections import OrderedDict, namedtuple
+from collections import OrderedDict, defaultdict, namedtuple
 from ctypes import c_double, c_int
 from functools import reduce
 from hashlib import sha1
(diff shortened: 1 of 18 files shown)


devito/stencilkernel.py
```diff
@@ -299,6 +474,32 @@ def argtypes(self):
 }
 
 """
+StencilKernel options
+"""
+options = {
+    'at_squeezer': 3,
+    'at_blocksize': [8, 16, 32]
+}
```
mloubout: What's the reason for this sizes?

FabioLuporini: nothing particularly solid: they simply looked reasonable sizes given the expected size of the parallel loop and the expected number of threads. And in my initial experiments (in Devito3.0) they resulted in performance improvements, while slightly different block sizes didn't change much the situation. I'll introduce more options in the future.

examples/benchmark.py
```diff
@@ -182,46 +192,56 @@ def run(self, *args, **kwargs):
             warning("Could not load any results, nothing to plot. Exiting...")
             sys.exit(0)
 
-        gflopss = bench.lookup(params=parameters, measure="gflopss", event="loop_body")
-        oi = bench.lookup(params=parameters, measure="oi", event="loop_body")
-        time = bench.lookup(params=parameters, measure="timings", event="loop_body")
+        gflopss = bench.lookup(params=parameters, measure="gflopss", event="main")
+        oi = bench.lookup(params=parameters, measure="oi", event="main")
+        time = bench.lookup(params=parameters, measure="timings", event="main")
```
mlange05: Is this backwards compatible? I'm assuming we want to run 2.0.1 vs. 3.0 in like-for-like performance comparisons, so should we have a switch here?

examples/benchmark.py
```diff
@@ -170,7 +180,7 @@ def run(self, *args, **kwargs):
         bench = Benchmark(
             name=args.problem, resultsdir=args.resultsdir, parameters=parameters
         )
-        bench.execute(BenchExecutor(), warmups=0)
+        bench.execute(BenchExecutor(), warmups=0, repeats=1)
```
mlange05: Did you mean to commit that?

mlange05:
Apart from the backward-compatibility in `benchmark.py` this looks good to me.

examples/benchmark.py
FabioLuporini: It's not backward compatible for plotting, you're right. But I guess for comparison we're just interested in raw numbers, or at least something simpler to compare than roofline plots? Anyway, good point: I will add a switch

examples/benchmark.py
FabioLuporini: nope, that was just for testing purposes. Good spot

examples/benchmark.py
FabioLuporini: done

examples/benchmark.py
FabioLuporini: done

FabioLuporini:
@navjotk : I added the TODO as discussed yesterday

mlange05:
All looks good. Merging...

PR #194, opened 2017-02-11
Tweak TTI to run via Devito3.0

mlange05:
This merge applies the same type of trickery to allow TTI to be run through the new `StencilKernel` interface, hidden behind a `legacy` flag for now. It comes with a small set of bug fixes and the necessary test modifications and also hooks up profiling information for the acoustic and TTI wrappers for the new API.

Travis does not test the new code path yet due to the heavy nature of all tests involving TTI, but I have verified correctness of the relevant tests locally. This merge is intended to allow us to like-for-like test the current with the 3.0 API and gather performance data ahead of the switch. Support for 3.0 on acoustic and TTI is currently still on a "best effort" basis, until we have a better handle on some of the outstanding issues.

### DIFF
--- a/devito/nodes.py
+++ b/devito/nodes.py
@@ -297,6 +297,10 @@ class Iteration(Node):
         """
         loop_body = [s.ccode for s in self.nodes]
 
+        # Disregard offsets for buffered dimensions
+        if self.dim.is_Buffered:
+            self.offsets = [0, 0]
+
         # Start
         if self.offsets[0] != 0:
             val = "%s + %s" % (self.limits[0], -self.offsets[0])
--- a/devito/nodes.py
+++ b/devito/nodes.py
@@ -298,12 +298,11 @@ class Iteration(Node):
         loop_body = [s.ccode for s in self.nodes]
 
         # Disregard offsets for buffered dimensions
-        if self.dim.is_Buffered:
-            self.offsets = [0, 0]
+        offsets = [0, 0] if self.dim.is_Buffered else self.offsets
 
         # Start
-        if self.offsets[0] != 0:
-            val = "%s + %s" % (self.limits[0], -self.offsets[0])
+        if offsets[0] != 0:
+            val = "%s + %s" % (self.limits[0], -offsets[0])
             try:
                 val = eval(val)
             except (NameError, TypeError):
@@ -313,8 +312,8 @@ class Iteration(Node):
         loop_init = c.InlineInitializer(c.Value("int", self.index), ccode(val))
 
         # Bound
-        if self.offsets[1] != 0:
-            val = "%s - %s" % (self.limits[1], self.offsets[1])
+        if offsets[1] != 0:
+            val = "%s - %s" % (self.limits[1], offsets[1])
             try:
                 val = eval(val)
             except (NameError, TypeError):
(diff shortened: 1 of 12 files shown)


devito/nodes.py
```diff
@@ -297,6 +297,10 @@ def ccode(self):
         """
         loop_body = [s.ccode for s in self.nodes]
 
+        # Disregard offsets for buffered dimensions
+        if self.dim.is_Buffered:
+            self.offsets = [0, 0]
```
FabioLuporini: Should this be done here, or in the constructor? modifying object values in ``ccode`` is weird

FabioLuporini: And is it always the case that ``offsets == [0,0]`` with ``BuffereredDimension``s ?

FabioLuporini:
Apart from the thing in ``nodes.py``, this is for me good to go.

examples/tti/TTI_codegen.py
```diff
@@ -48,27 +51,95 @@ def damp_boundary(damp, nbp):
...
+
+        u.pad_time = save
+        v.pad_time = save
+
+        if u_ini is not None:
+            u.data[0:3, :] = u_ini[:]
+            v.data[0:3, :] = u_ini[:]
+
+        # Create source symbol
+        p_src = Dimension('p_src', size=nsrc)
```
FabioLuporini: One more thing: could in principle this be created internally directly by ``SourceLike`` ?

examples/tti/TTI_codegen.py
```diff
@@ -48,27 +51,95 @@ def damp_boundary(damp, nbp):
...
+        src = SourceLike(name="src", dimensions=[time, p_src], npoint=nsrc, nt=nt,
+                         dt=dt, h=h, ndim=ndim, nbpml=nbpml, dtype=dtype,
+                         coordinates=self.source.receiver_coords)
+        src.data[:] = .5 * self.source.traces[:]
+
+        # Create receiver symbol
+        p_rec = Dimension('p_rec', size=nrec)
```
FabioLuporini: Same as before

FabioLuporini:
I also just added one more small question about ``SourceLike``and ``Dimension``

examples/tti/TTI_codegen.py
mlange05: Yes, I was planning to do this as part of the next steps towards cleaning the examples. Will be addressed soon.

devito/nodes.py
mlange05: I agree we shouldn't set this here. Would explicitly ignoring it with something like `offsets = self.offsets if not self.dim.is_Buffered else [0, 0]` and then use `offsets` instead of `self.offsets` work?

Regarding generality, for now we always ignore offsets on `BufferedDimensions`, but not doing this might hold a key to fixing the `pad_time` problem and relates to the overall time indexing issue - so it's likely to change in the future.

FabioLuporini: Yes, do you as propose

FabioLuporini:
OK then

mlange05:
Ok, I addressed the comments and added a few more hookups to propagate `dle` and `legacy` to the benchmarking script.

FabioLuporini:
I understand from offline  discussions that everybody is happy about landing this, so merging...

FabioLuporini:
Merged.

PR #201, opened 2017-02-26
Seismic model cleanup

mlange05:
Small clean-up change that trims the `Model` in our seismic examples. Most of this is cosmetics to improve readability:
* Rename `IGrid` to `Model` and move to `seismic` sub-directory
* Let the `Model` provide the symbols `m` and `damp` and manage the associated data
* Initialise `Model` with `nbpml` and provide `shape` as well as `shape_pml` - This makes operator construction cleaner, since we now provide all the grid-related meta-data through the model.
* Let `set_vp` perform the automated `m` update
* Make things more Pythonic :snake: by turning various `get`/`set` methods into properties


examples/tti/tti_operators.py
```diff
@@ -7,57 +7,52 @@
...
         theta = 0
 
     if model.phi is not None:
-        phi = DenseData(name="phi", shape=model.get_shape_comp(),
-                        dtype=damp.dtype, space_order=spc_order)
+        phi = DenseData(name="phi", shape=model.shape_pml,
+                        dtype=model.dtype, space_order=spc_order)
         phi.data[:] = model.pad(model.phi)
         parm += [phi]
     else:
```
mloubout: Siuldn't all of theses be moved to `model` too? The same way m is handled.

examples/seismic/model.py
```diff
@@ -0,0 +1,126 @@
...
+            damp[:, -(i + 1)] += val
+        else:
+            damp[i, :, :] += val
+            damp[-(i + 1), :, :] += val
+            damp[:, i, :] += val
+            damp[:, -(i + 1), :] += val
+            damp[:, :, i] += val
+            damp[:, :, -(i + 1)] += val
+
+
```
mloubout: As we do have an explicit expression for this, could we change it to a separate loop instead and avoid a useless computation on zeros? Not sure if it should be part of this PR but worth mentioning.

examples/acoustic/Acoustic_codegen.py
```diff
@@ -95,40 +63,37 @@ def Forward(self, save=False, cache_blocking=None, auto_tuning=False,
...
+                         dt=self.dt, h=h, ndim=ndim, nbpml=nbpml, dtype=dtype,
                          coordinates=self.source.receiver_coords)
         src.data[:] = self.source.traces[:]
 
         # Create receiver symbol
         p_rec = Dimension('p_rec', size=nrec)
         rec = SourceLike(name="rec", dimensions=[time, p_rec], npoint=nrec, nt=nt,
-                         dt=dt, h=h, ndim=ndim, nbpml=nbpml, dtype=dtype,
+                         dt=self.dt, h=h, ndim=ndim, nbpml=nbpml, dtype=dtype,
                          coordinates=self.data.receiver_coords)
```
mloubout: O hadn't realized, but when did the "remove 2nd dim" for 2D disapeared. In practice, even for 2D it is quite usual to have still 3D coordinates wit h0 everywhere in z which would break in our case as dimensions won't match.

examples/seismic/model.py
```diff
@@ -0,0 +1,126 @@
...
+        return tuple(d + 2*self.nbpml for d in self.shape)
+
+    @property
+    def critical_dt(self):
+        """Critical computational time step value from the CFL condition."""
+        # For a fixed time order this number goes down as the space order increases.
+        #
+        # The CFL condtion is then given by
+        # dt <= coeff * h / (max(velocity))
+        coeff = 0.38 if len(self.shape) == 3 else 0.42
```
navjotk: Since we have the FD discretization in the symbolics, can we delegate this down to the symbolic object where it automatically calculates this based on the actual discretization? I would like to get rid of these magic numbers which only work for particular discretizations and hence are implicit assumptions that would be hard to spot later.

tests/test_adjointA.py
```diff
@@ -60,16 +61,15 @@ def test_acoustic(dimensions, time_order, space_order):
...
     data = IShot()
     src = IShot()
 
     f0 = .010
+    dt = model.critical_dt
     if time_order == 4:
-        dt = 1.73 * model.get_critical_dt()
-    else:
-        dt = model.get_critical_dt()
+        dt *= 1.73
```
navjotk: Same as the previous comment. This magic number could probably go away as well, in the same step.

navjotk:
I opened an issue regarding my comments since they don't need to be addressed in this PR. I am all for merging this ASAP since I need it for other stuff that I am working on too.

examples/seismic/model.py
```diff
@@ -0,0 +1,126 @@
...
+        else:
+            damp[i, :, :] += val
+            damp[-(i + 1), :, :] += val
+            damp[:, i, :] += val
+            damp[:, -(i + 1), :] += val
+            damp[:, :, i] += val
+            damp[:, :, -(i + 1)] += val
+
+
+class Model:
```
FabioLuporini: new style classes

FabioLuporini: New style classes

examples/tti/tti_operators.py
```diff
@@ -157,9 +152,9 @@ def ForwardOperator(model, u, v, src, rec, damp, data, time_order=2,
     if legacy:
         kwargs.pop('dle', None)
 
-        op = Operator(nt, m.shape, stencils=stencils, subs=[subs, subs],
+        op = Operator(nt, model.shape_pml, stencils=stencils, subs=[subs, subs],
                       spc_border=spc_order/2, time_order=time_order,
```
FabioLuporini: Is this ``spc_order/2`` still fine @mloubout ?

examples/tti/tti_operators.py
mlange05: Yes, change pushed.

examples/seismic/model.py
mlange05: Yes, but I think we might pick this up in a separate PR.

examples/acoustic/Acoustic_codegen.py
mlange05: Yes, maybe, but the Devito examples are primarily for demonstrating our capabilities. So these sort of "real-life hacks" are just making the examples harder to read and understand, especially if dimensions disappear under the hood somewhere.

examples/seismic/model.py
mlange05: I totally agree. But we'll probably want to do this in a separate PR.

mlange05:
Ok, I think all comments have been addressed now. @mloubout can you please check the last commit on the branch and make sure TTI still works as you'd expect?

examples/acoustic/Acoustic_codegen.py
```diff
@@ -95,40 +63,37 @@ def Forward(self, save=False, cache_blocking=None, auto_tuning=False,
...
 
         if legacy and auto_tuning:
             cache_blocking = self.at.block_size
 
         # Create the forward wavefield
-        u = TimeData(name="u", shape=self.model.get_shape_comp(), time_dim=nt,
+        u = TimeData(name="u", shape=self.model.shape_pml, time_dim=nt,
```
mloubout: naming? `shape_pml` sounds like shape of the pml not full domain size.

mloubout:
Looks good to me.

examples/acoustic/Acoustic_codegen.py
mlange05: Good point, thanks! I've changed it to `shape_domain` for clarity.

mlange05:
Ok, travis is happy, so merging...

PR #202, opened 2017-02-28
Add flag to make first touch optional

navjotk:
Skip unnecessary JIT calls since our NUMA support isn't complete anyway.


FabioLuporini:
looks good to me

PR #209, opened 2017-03-09
Test for Born, probably gonna break memory but smaller makes no sense

mloubout:
New test for adjoint between Born and gradient. Needs about 4Gb of mem so it may fail.


mloubout:
Ok, test seems to not overload memory.
@ggorman here is the trust test :)

FabioLuporini:
Question: given that the grid dimensions look small, why should this test need up to 4GBs of DRAM? what's consuming this much memory?

mloubout:
It is now reduced to ~700Mb. This is a gradient calculation that needs to store the full forward wavefield.

PR #210, opened 2017-03-15
Adjoint3 0

mloubout:
Still a lot to do to migrate everything to 3.0 but 

1- Time axis flag is in (thanks to @mlange05 )
2 - Adjoint test passes with everything using StencilKernel

So I am creating the PR for the sake of it and also because there is a major crash in travis after the first test, and I have no idea where it comes from (runs fine on my machine).

EDIT:
Segfault seems fixed. StencilKernel throws error (dse I think) with
`save=True` in forward
Born

### DIFF
--- a/devito/finite_difference.py
+++ b/devito/finite_difference.py
@@ -6,12 +6,31 @@ from operator import mul
 from sympy import finite_diff_weights, symbols
 
 from devito.dimension import x, y
+from devito.logger import error
 
 __all__ = ['first_derivative', 'second_derivative', 'cross_derivative',
            'left', 'right', 'centered']
 
 # Default spacing symbol
 h = symbols('h')
+
+
+class Transpose(object):
+    """Class that defines if the derivative is itself or adjoint (transpose).
+    This only matter for odd order derivatives that requires
+    a minus sign for the transpose."""
+    def __init__(self, transpose):
+        self._transpose = transpose
+
+    def __eq__(self, other):
+        return self._transpose == other._transpose
+
+    def __repr__(self):
+        return {1: 'direct', -1: 'transpose'}[self._transpose]
+
+
+direct = Transpose(1)
+transpose = Transpose(-1)
 
 
 class Side(object):
@@ -25,6 +44,19 @@ class Side(object):
 
     def __repr__(self):
         return {-1: 'left', 0: 'centered', 1: 'right'}[self._side]
+
+    def adjoint(self, matvec):
+        if matvec == direct:
+            return self
+        else:
+            if self == centered:
+                return centered
+            elif self == right:
+                return left
+            elif self == left:
+                return right
+            else:
+                error("Unsupported side value")
 
 
 left = Side(-1)
@@ -139,7 +171,8 @@ def first_derivative(*args, **kwargs):
     dim = kwargs.get('dim', x)
     diff = kwargs.get('diff', h)
     order = int(kwargs.get('order', 1))
-    side = kwargs.get('side', centered)
+    matvec = kwargs.get('matvec', direct)
+    side = kwargs.get('side', centered).adjoint(matvec)
     deriv = 0
     # Stencil positions for non-symmetric cross-derivatives with symmetric averaging
     if side == right:
@@ -159,4 +192,4 @@ def first_derivative(*args, **kwargs):
     for i in range(0, len(ind)):
             var = [a.subs({dim: ind[i]}) for a in args]
             deriv += c[i] * reduce(mul, var, 1)
-    return deriv
+    return matvec._transpose*deriv
--- a/devito/interfaces.py
+++ b/devito/interfaces.py
@@ -550,9 +550,6 @@ class TimeData(DenseData):
 
         self._full_data = self._data.view()
 
-        if self.pad_time:
-            self._data = self._data[self.time_order:, :, :]
-
     @property
     def dim(self):
         """Returns the spatial dimension of the data object"""
--- a/devito/interfaces.py
+++ b/devito/interfaces.py
@@ -510,9 +510,7 @@ class TimeData(DenseData):
             self.save = kwargs.get('save', False)
             self.pad_time = kwargs.get('pad_time', False)
 
-            if self.save:
-                time_dim += self.time_order
-            else:
+            if not self.save:
                 time_dim = self.time_order + 1
                 self.indices[0].modulo = time_dim
 
--- a/devito/stencilkernel.py
+++ b/devito/stencilkernel.py
@@ -196,7 +196,6 @@ class OperatorBasic(Function):
 
         # Add profiler structs
         arguments.update(self._extra_arguments())
-
         return arguments, dim_sizes
 
     @property
@@ -460,7 +459,7 @@ class OperatorCore(OperatorBasic):
                     if IsPerfectIteration().visit(j) and j not in mapper:
                         # Insert `TimedList` block. This should come from
                         # the profiler, but we do this manually for now.
-                        lname = 'loop_%s_%d' % (j.index, i)
+                        lname = 'loop_%s_%d' % (j.index, len(mapper))
                         mapper[j] = TimedList(gname=self.profiler.t_name,
                                               lname=lname, body=j)
                         self.profiler.t_fields += [(lname, c_double)]
--- a/examples/acoustic/Acoustic_codegen.py
+++ b/examples/acoustic/Acoustic_codegen.py
@@ -229,7 +229,7 @@ class Acoustic_cg(object):
         u = TimeData(name="u", shape=self.model.shape_domain, time_dim=nt,
                      time_order=2, space_order=self.s_order,
                      dtype=self.model.dtype)
-        U = TimeData(name="U", shape=self.model.shape_domain, time_dim=nt,
+        U = TimeData(name="ul", shape=self.model.shape_domain, time_dim=nt,
                      time_order=2, space_order=self.s_order,
                      dtype=self.model.dtype)
 
--- a/examples/acoustic/Acoustic_codegen.py
+++ b/examples/acoustic/Acoustic_codegen.py
@@ -4,9 +4,10 @@ from __future__ import print_function
 import numpy as np
 
 from devito.at_controller import AutoTuner
-from devito.dimension import Dimension, time
+from devito.dimension import Dimension
+from devito.interfaces import DenseData, TimeData, CachedSymbol
 from examples.acoustic.fwi_operators import *
-
+from examples.source_type import SourceLike
 
 class Acoustic_cg(object):
     """
@@ -159,7 +160,6 @@ class Acoustic_cg(object):
     def Gradient(self, recin, u, cache_blocking=None, auto_tuning=False,
                 dse='advanced', dle='advanced', compiler=None, legacy=True):
         nt, nrec = self.data.shape
-        nsrc = self.source.shape[1]
         ndim = len(self.model.shape)
         h = self.model.get_spacing()
         dtype = self.model.dtype
@@ -240,7 +240,6 @@ class Acoustic_cg(object):
                             cache_blocking=cache_blocking, dse=dse,
                             dle=dle, compiler=compiler, profile=True,
                             legacy=legacy)
-
         if legacy:
             born.apply()
             return (rec.data, u, U, born.propagator.gflopss,
--- a/examples/acoustic/Acoustic_codegen.py
+++ b/examples/acoustic/Acoustic_codegen.py
@@ -107,12 +107,50 @@ class Acoustic_cg(object):
             summary = fw.apply(autotune=auto_tuning)
             return rec.data, u, summary.gflopss, summary.oi, summary.timings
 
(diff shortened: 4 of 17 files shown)


mloubout:
Ok, mostly everything there with `StencilKernel`, only problem now:
1- dse does NOT like Born and Gradient stencil
2- Gradient stencils are split in different time loops => can;t compile and even if it does will compute BS
3- time loop limits incorrect in Born.

FabioLuporini:
Please file suitable issues to allow me to fix the wrong DSE behaviour

devito/finite_difference.py
```diff
@@ -31,6 +32,24 @@
...
+    def __eq__(self, other):
+        return self._transpose == other._transpose
+
+    def __repr__(self):
+        return {1: 'direct', -1: 'transpose'}[self._transpose]
+
+
+direct = Transpose(1)
+transpose = Transpose(-1)
+
```
FabioLuporini: Perhaps expressible in a more concise way as:
`from collections import namedtuple

Transpose = namedtuple('Transpose', 'transpose')
DIRECT = Transpose(1)
TRANSPOSE = Transpose(-1)
`
?

devito/finite_difference.py
```diff
@@ -43,6 +62,19 @@ def __eq__(self, other):
...
+        else:
+            if self == centered:
+                return centered
+            elif self == right:
+                return left
+            elif self == left:
+                return right
+            else:
+                error("Unsupported side value")
+
```
FabioLuporini: You can probably use a ``dict`` ?

devito/interfaces.py
```diff
@@ -551,7 +551,7 @@ def _allocate_memory(self):
         self._full_data = self._data.view()
 
         if self.pad_time:
-            self._data = self._data[self.time_order:, :, :]
+            self._data = self._data[self.time_order:, ...]
```
FabioLuporini: ?

devito/stencilkernel.py
```diff
@@ -123,7 +123,6 @@ def __init__(self, stencils, **kwargs):
         # Track the DSE and DLE output, as they may be useful later
         self._dse_state = dse_state
         self._dle_state = dle_state
-
```
FabioLuporini: ?

devito/stencilkernel.py
```diff
@@ -141,9 +140,9 @@ def apply(self, *args, **kwargs):
         # Map of required arguments and actual dimension sizes
         arguments = OrderedDict([(arg.name, arg) for arg in self.parameters])
         dim_sizes = {}
-
         # Traverse positional args and infer loop sizes for open dimensions
         f_args = [f for f in arguments.values() if isinstance(f, SymbolicData)]
+
```
FabioLuporini: ?

devito/stencilkernel.py
```diff
@@ -220,13 +219,15 @@ def apply(self, *args, **kwargs):
     def _profile_sections(self, nodes):
         """Introduce C-level profiling nodes within the Iteration/Expression tree."""
         mapper = {}
+        count = 0
```
FabioLuporini: don't need this, just use ``len(mapper)``

tests/test_tti.py
```diff
@@ -67,8 +67,8 @@ def test_tti(dimensions, space_order):
     t0 = 0.0
     tn = 350.0
     nt = int(1+(tn-t0)/dt)
-    last = (nt - 1) % 3
-    indlast = [(last + 1) % 3, (last+2) % 3, last % 3]
+    last = (nt) % 3
+    indlast = [(last + 2) % 3, (last+1) % 3, last % 3]
```
FabioLuporini: what's happening here?

devito/finite_difference.py
mloubout: Doesn't seems to be a problem for `Side`, `TimeAxis`,.....

devito/interfaces.py
mloubout: Top line only correct for 2D

devito/stencilkernel.py
mloubout: really...

tests/test_tti.py
mloubout: Just computing in what order is the the wavefield after computation (cyclic indices)

devito/finite_difference.py
FabioLuporini: Because I only thought about this now

tests/test_tti.py
navjotk: This should ideally be a property of the data object.

examples/tti/tti_operators.py
```diff
@@ -41,70 +41,94 @@ def ForwardOperator(model, u, v, src, rec, data, time_order=2,
...
+        Hp = -(.5*Gxx + .5*Gxx2 + .5 * Gyy + .5*Gyy2)
+        Hzr = -(.5*Gzz + .5 * Gzz2)
 
     else:
         Gx1p = (ang0 * u.dxr - ang1 * u.dy)
         Gz1r = (ang1 * v.dxr + ang0 * v.dy)
         Gxx1 = (first_derivative(Gx1p * ang0, dim=x,
-                                 side=left, order=spc_brd) -
+                                 side=right, order=spc_brd,
+                                 matvec=transpose) -
```
FabioLuporini: Remind me what's the purpose of ``matvec=transpose`` here?

FabioLuporini:
OK, so, in order to land this, what we need is the following
- some documentation of the tests/operators that fail if ``dse != None`` (GitHub issues with instructions to reproduce?)
- documentation of the hacks used "to make things work" as I see some commit messages mention "huge hacks" 
- (optional) use of `namedtuple` for `Side`, `TimeAxis` etc

examples/tti/tti_operators.py
mloubout: the expression of the derivative is Dx^T.Dx . It used to be done "by hand" by knowing what the transpose is supposed to be, but this way it is actually readable and implement what you would write on paper.

mloubout:
All `legacy=True`  and related will be removed once people are happy as it is not compatible anymore with the memory allocation.

FabioLuporini:
Thanks for documenting the issues. 

This branch however needs to be rebased on master and Travis isn't happy anymore.

One last question: what hacks have been used here, if any? I might be just confused by old commit messages

mloubout:
The only hack used is puting a random id on the model Data to avoid problem wth wrong `m` used.

Travis seems to fail on `openmp` property for the compiler, will look into it,

examples/seismic/model.py
```diff
@@ -57,21 +57,25 @@ def __init__(self, origin, spacing, vp, nbpml=0, dtype=np.float32,
         self.spacing = spacing
         self.nbpml = nbpml
         self.dtype = dtype
-
+        # Generate a random id with the model.
+        # This is needed for the Gradient test that uses two different models
+        # Model(m) and Model(m0) that uses the same value of m without id
```
mlange05: Ok, I think this here really exposes a severe design flaw that requires an API re-think and a separate PR to fix. I've filed issue #220 to track this.

FabioLuporini: Does this become mergeable after #223  lands?

mloubout: There is still some problems with DSE, but can be part of another PR I guess
The failing TRAVIS is a bug when `compiler=None`, I'll fix it

FabioLuporini: Yes, hopefully I'll address those by the hand of the week

mloubout:
So this is Green. I think most of the issues related to this PR need separate PRs as this one is already quite heavy. Please add final review/requests.
@FabioLuporini for the use of `dict` still not sure how to do that, will look into it and file different PR.

FabioLuporini:
This needs be rebased first. Then, it looks mergeable to me. As we agreed, I'll address the DSE bugs in a separate PR.

mloubout:
@mlange05 @FabioLuporini 
Can i remove `Operator` and 'Propagator` and all the related tests?

FabioLuporini:
I think we're not quite there yet (because of my issues...). I also think that should go in the "final PR", the one officially leading to 3.0.

mloubout:
Ok, then I have no idea how to fix that PR as the memory allocation for `StencilKernel` is segfaulting most of `Operator stuff`.

mloubout:
@mlange05 @navjotk @FabioLuporini 

Current status is
- everything green except `gradient_test` due to the two `Model` issue.
- `dse` crashing so set to `None` everywhere

examples/acoustic/Acoustic_codegen.py
```diff
@@ -59,6 +61,7 @@ def __init__(self, model, data, source, nbpml=40, t_order=2, s_order=2,
             self.at = AutoTuner(fw)
             self.at.auto_tune_blocks(self.s_order + 1, self.s_order * 4 + 2)
 
+    # Forward modelling
```
mlange05: Python docstrings would be better, since documentation-generators pick them up. Same for the other classes below.

examples/acoustic/fwi_operators.py
```diff
@@ -1,10 +1,9 @@
-from sympy import Eq, solve, symbols
+from sympy import Eq, symbols
 
 from devito.dimension import t, time
-from devito.interfaces import DenseData, TimeData
-from devito.operator import *
+from devito.interfaces import Backward, Forward
 from devito.stencilkernel import StencilKernel
-from examples.source_type import SourceLike
+from devito.operator import *
```
mlange05: Can we not be explicit with imports here?

tests/test_cache_blocking.py
```diff
@@ -38,6 +39,7 @@ def test_cache_blocking_edge_cases(self, shape, time_order,
         self.cache_blocking_test(shape, time_order, spc_border, cache_blocking)
 
     def cache_blocking_test(self, shape, time_order, spc_border, cache_blocking):
+        clear_cache()
```
mlange05: Why? We're not touching this test at all, why do we need to introduce that?

tests/test_save.py
```diff
@@ -33,14 +33,15 @@ def run_simulation(save=False, dx=0.01, dy=0.01, a=0.5, timesteps=100):
     a, h, s = symbols('a h s')
     eqn = Eq(u.dt, a * (u.dx2 + u.dy2))
     stencil = solve(eqn, u.forward)[0]
-    op = Operator(stencils=Eq(u.forward, stencil), subs={a: 0.5, h: dx, s: dt},
-                  nt=timesteps, shape=(nx, ny), spc_border=1, time_order=1)
+    op = StencilKernel(stencils=Eq(u.forward, stencil), subs={a: 0.5, h: dx, s: dt},
```
mlange05: Just a sidenote, but why are we flipping this test in this PR? Should this not have been done in a separate one?

tests/test_timestepping.py
```diff
@@ -49,7 +49,7 @@ def test_backward(b):
     eqn = Eq(b.backward, b - 1.)
     StencilKernel(eqn, dle=None, dse=None, time_axis=Backward)()
     for i in range(b.shape[0]):
-        assert np.allclose(b.data[i, :], 1. + i, rtol=1.e-12)
+        assert np.allclose(b.data[i, :], 2. + i, rtol=1.e-12)
```
mlange05: Again, why are we even touching this?

tests/test_cache_blocking.py
mloubout: This test segfault (some CMemor error) without it.

tests/test_save.py
mloubout: `save=True` not supported anymore for Operator due to new memory allocation

tests/test_timestepping.py
mloubout: UAtomatic time-loop border change number of time step compared to before.

mlange05:
Also, can you please file a separate issue for the multi-model problem, ideally highlighting this in a MFE?

mloubout:
Will file this issue, what's MFE?

mlange05:
"Minimum failing example" - sorry, I should have been clearer.

FabioLuporini:
> dse crashing so set to None everywhere

This is reasonable given that I'll still need some time to fix those issues properly (don't wanna come up with hacky solutions)

FabioLuporini:
As long as both Michael -- since this PR is mostly touching the operators, stuff that he's rewritten a few times -- and Travis are happy, I'm also happy. This can probably be merged ?

mlange05:
Yes, I agree. Just give me a few minutes to touch up the docstrings, imports and check that segfault.

tests/test_timestepping.py
mlange05: Ok

tests/test_cache_blocking.py
mlange05: Oh wow, confirmed, this does look positively evil. This test is very high on the consolidation list for the actual operator flip though, so we will probably deal with it then. I'm happy to go ahead then.

mlange05:
All good, will merge once travis completes it's bonus lap.

PR #211, opened 2017-03-23
Reverse time interface for 3.0

mlange05:
After long deliberation and experimentation we finally introduce the ability to add reverse-time operators through the new 3.0 API. To the outside this looks pretty similar to the previous approach where an operator argument `StencilKernel, eq, subs=subs, time_axis=Forward/Backward` determines the direction of the time loop.

Under the hood this now behaves slightly different, where the `Dimension.reverse` on the dedicated `time` dimension is set according to the given parameter. This ensures that:
a) We can use the same `time`/`t`symbols in forward and backward runs
b) Custom dimensions may also be inverted if desired

We also introduce a new high-level utility object `TimeAxis` that describes the particular mode of "time" to use in an operator and will hopefully be extended soon.

This change also includes a fix to the `partial_order` utility function that is used to schedule loops according to dimension dependencies in the index signatures. It also forces the `StencilKernel` to create one global partial ordering across encountered clusters to ensure semantic rigour.

### DIFF
--- a/devito/interfaces.py
+++ b/devito/interfaces.py
@@ -4,7 +4,7 @@ import numpy as np
 from sympy import Function, IndexedBase, Symbol, as_finite_diff, symbols
 from sympy.abc import h, s
 
-from devito.dimension import d, p, t, x, y, z
+from devito.dimension import d, p, t, x, y, z, time
 from devito.finite_difference import (centered, cross_derivative,
                                       first_derivative, left, right,
                                       second_derivative)
@@ -515,7 +515,9 @@ class TimeData(DenseData):
         dimensions = kwargs.get('dimensions', None)
         if dimensions is None:
             # Infer dimensions from default and data shape
-            _indices = [t, x, y, z]
+            save = kwargs.get('save', None)
+            tidx = time if save else t
+            _indices = [tidx, x, y, z]
             shape = kwargs.get('shape')
             dimensions = _indices[:len(shape) + 1]
         return dimensions
@@ -538,15 +540,17 @@ class TimeData(DenseData):
     def forward(self):
         """Symbol for the time-forward state of the function"""
         i = int(self.time_order / 2) if self.time_order >= 2 else 1
+        _t = self.indices[0]
 
-        return self.subs(t, t + i * s)
+        return self.subs(_t, _t + i * s)
 
     @property
     def backward(self):
         """Symbol for the time-backward state of the function"""
         i = int(self.time_order / 2) if self.time_order >= 2 else 1
+        _t = self.indices[0]
 
-        return self.subs(t, t - i * s)
+        return self.subs(_t, _t - i * s)
 
     @property
     def dt(self):
--- a/devito/interfaces.py
+++ b/devito/interfaces.py
@@ -555,22 +555,24 @@ class TimeData(DenseData):
     @property
     def dt(self):
         """Symbol for the first derivative wrt the time dimension"""
+        _t = self.indices[0]
         if self.time_order == 1:
             # This hack is needed for the first-order diffusion test
-            indices = [t, t + s]
+            indices = [_t, _t + s]
         else:
             width = int(self.time_order / 2)
-            indices = [(t + i * s) for i in range(-width, width + 1)]
+            indices = [(_t + i * s) for i in range(-width, width + 1)]
 
-        return as_finite_diff(self.diff(t), indices)
+        return as_finite_diff(self.diff(_t), indices)
 
     @property
     def dt2(self):
         """Symbol for the second derivative wrt the t dimension"""
-        width_t = int(self.time_order/2)
-        indt = [(t + i * s) for i in range(-width_t, width_t + 1)]
+        _t = self.indices[0]
+        width_t = int(self.time_order / 2)
+        indt = [(_t + i * s) for i in range(-width_t, width_t + 1)]
 
-        return as_finite_diff(self.diff(t, t), indt)
+        return as_finite_diff(self.diff(_t, _t), indt)
 
 
 class CoordinateData(SymbolicData):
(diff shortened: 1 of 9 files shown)


mloubout:
Look good. Few things
1- Put the ` value = o.dim.parent + off + modulo` in `visitor, non-cyclic indices without it
2- This still look quite similar to my  PR so does this break our operators too?

devito/interfaces.py
```diff
@@ -538,35 +562,39 @@ def dim(self):
...
+        return self.subs(_t, _t - i * s)
 
     @property
     def dt(self):
         """Symbol for the first derivative wrt the time dimension"""
+        _t = self.indices[0]
         if self.time_order == 1:
             # This hack is needed for the first-order diffusion test
-            indices = [t, t + s]
+            indices = [_t, _t + s]
```
mloubout: Only true for forward isn't it?

devito/nodes.py
```diff
@@ -345,28 +345,33 @@ def ccode(self):
...
+            end = self.limits[1]
 
-        # Increment
-        loop_inc = '%s += %s' % (self.index, self.limits[2])
+        # For reverse dimensions flip loop bounds
+        if self.dim.reverse:
+            loop_init = c.InlineInitializer(c.Value("int", self.index), ccode(end))
+            loop_cond = '%s >= %s' % (self.index, ccode(start))
+            loop_inc = '%s -= %s' % (self.index, self.limits[2])
+        else:
```
mloubout: Shouldn't it be `end-1` to `start-1`. Forward loop last index is `end-1`

mlange05:
Ok, I'm really not sure 1 is required, and I'd like to have a go at trying to flip gradient without it. We can still put it in once we actually flip the examples.
For 2 I'm not quite sure what you mean - the `Operator` generally still works (not test has been flipped), but I haven't tried Adjoint or Gradient with the new API yet (next PR, hopefully a rebased #210).

devito/operator.py
```diff
@@ -100,6 +101,9 @@ def __init__(self, nt, shape, dtype=np.float32, stencils=[],
         for d in dimensions:
             sym_undef.remove(d)
 
+        # Replace all occurences of the `time` dimension with `t`
+        self.stencils = [eqn.subs(time, t) for eqn in self.stencils]
+
```
mloubout: IS this always correct or will it break `save` case?

mloubout:
>Ok, I'm really not sure 1 is required, and I'd like to have a go at trying to flip gradient without it. We can still put it in once we actually flip the examples.

It is to make sure we never take modulo of a negative number.

devito/interfaces.py
mlange05: Yes, but this change is only concerned with allowing different time symbols here. Actually changing the timestepping scheme here requires a little more thought and a lot more testing. Now that we have the `TimeAxis` object though we should aim to implement the classic Forward/Backward/Crank Nicholson timesteppers, as documented [here](https://en.wikipedia.org/wiki/Finite_difference_method).

devito/nodes.py
mlange05: Well, technically no, since the automatic loop bound derivation should capture this. But you are right this is insufficiently tested and might be broken. I'll ad some tests.

devito/operator.py
mlange05: Good catch, I'm not sure. It doesn't break any existing tests, but I'll have closer look.

mlange05: Ok, this still looks to work just as intended. I believe it is due to the fact that the treatment of time symbol `t` in the `Operator` is driven by a flag.

devito/nodes.py
mlange05: Ah you were partially right: the backward loop needs to start at `end - 1`, but the rest is picked up by the automatic bound inference. I've added tests to confirm that for forward and backward loops with a second order time stencil.

mlange05:
So, regarding point 1: we never take the modulo of a negative value since the automatic loop bounds ensure that we start from `1` if we have a `u[t - 1, ...]` term in our stencil.

devito/operator.py
```diff
@@ -82,16 +82,17 @@ def __init__(self, nt, shape, dtype=np.float32, stencils=[],
             dimensions += [i for i in retrieve_dimensions(eqn.rhs) if i not in dimensions]
 
         # Time dimension is fixed for now
-        time_dim = t
+        time_dim = [t, time]
```
navjotk: This is VERY confusing. I have read the entire diff and still don't know which time symbol is which.

mlange05: You wouldn't from the diff, since it's already been added a while ago in `dimension.py`. The core idea is that `time` is unrolled and `t` is indexed, but that will be explained in accompanying documentation. It's also now aligned with the `save` flag on the `TimeData` constructor, so the user does not actually need to know this.

mloubout:
>So, regarding point 1: we never take the modulo of a negative value since the automatic loop bounds ensure that we start from 1 if we have a u[t - 1, ...] term in our stencil.

This is one case, no guaranty it will always be the case. Doing what I say in point 1 will give the exact same results and make sure that noone ever end up in a weird case.

mlange05:
Oh sorry I didn't make this clearer: Yes the boundary detection guarantees that this will work for any offset on any dimension (`u[t - n, ...]` entails we start the time loop from `n`). It simply inspects the stencil and pulls out the furthest offset in each direction and then adjusts the loop bounds accordingly. This ensures that it will never access data outside of allocated memory for space loops and similarly that we never take the modulo of something negative for time.

mloubout:
Ok :D

mlange05:
Cool. Everyone happy with this then?

mloubout:
Yes, let's merge this and I'll rebase mine to see what it does

mlange05:
Great, done.

PR #212, opened 2017-03-27
Forward compat with StencilKernel

mloubout:
This is the first step to make out Julia framework compatible with `master` instead of some unmaintainable (for me) branch.

Input gladly accepted, not much change to devito itself, only a few extra functions/conditions


devito/compiler.py
```diff
@@ -76,7 +76,7 @@ def __init__(self, *args, **kwargs):
         self.version = kwargs.get('version', None)
         self.cc = 'gcc' if self.version is None else 'gcc-%s' % self.version
         self.ld = 'gcc' if self.version is None else 'gcc-%s' % self.version
-        self.cflags = ['-O3', '-g', '-march=native', '-fPIC', '-Wall', '-std=c99',
+        self.cflags = ['-O3', '-g', '-fPIC', '-Wall', '-std=c99',
                        '-Wno-unused-result', '-Wno-unused-variable']
```
mloubout: This will go back to `native` I just leave it there for now to make my life easier

devito/compiler.py
```diff
@@ -287,3 +295,14 @@ def jit_compile_and_load(ccode, basename, compiler=GNUCompiler):
     jit_compile(ccode, basename, compiler=compiler)
 
     return load(basename, compiler=compiler)
+
+
+def jit_compile_only(ccode, basename, compiler=GNUCompiler):
```
mlange05: I don't really see the point of this function - it simply invokes `jit_compile` and returns one of the input arguments. Can you not just use `jit_compile` instead?

devito/compiler.py
```diff
@@ -244,8 +244,16 @@ def jit_compile(ccode, basename, compiler=GNUCompiler):
     :param compiler: The toolchain used for compilation. GNUCompiler by default.
     :return: Path to compiled lib
     """
+
     src_file = "%s.%s" % (basename, compiler.src_ext)
-    lib_file = "%s.%s" % (basename, compiler.lib_ext)
+    from sys import platform
```
mlange05: Unless there is a good reason to hide it here, the import should to go the top of the file.

devito/stencilkernel.py
```diff
@@ -65,6 +65,7 @@ class StencilKernel(Function):
     """
     def __init__(self, stencils, **kwargs):
         name = kwargs.get("name", "Kernel")
+        self.name = name
```
mlange05: Could just do this in the above line and update all occurrences of `name` with `self.name` in the rest of the function.

devito/nodes.py
```diff
@@ -341,7 +341,7 @@ def ccode(self):
         loop_body = [s.ccode for s in self.nodes]
 
         # Disregard offsets for buffered dimensions
-        offsets = [0, 0] if self.dim.is_Buffered else self.offsets
+        offsets = self.offsets
```
mlange05: You could just drop this line entirely (and the now outdated comment) and replace all occurrences of `offsets` with `self.offsets` in the rest of the function.

devito/stencilkernel.py
```diff
@@ -511,6 +518,22 @@ def cfunction(self):
             self._cfunction.argtypes = self.argtypes
 
         return self._cfunction
+   
+    @property
+    def cfunctionJ(self):
```
mlange05: Ok, so the basic idea behind the compilation toolchain is to provide a cascade that can be intercepted at any point - but obviously you require the entry point between compile and load. So, why don't we split up the current cascade to expose that instead of adding another version of `cfunction`? I'm thinking we could turn `basename` into a property and provide a utility function `compile()` that simply does `jit_compile(self.ccode, self.basename, compiler=self.compiler)`. In the above `cfunction` we can then replace `jit_compile_and_load()` with
```
self.compile()
self.lib = load(self.basename, compiler=self.compiler)
```

devito/compiler.py
mlange05: Is anybody still working on #175?

mloubout: Not that I am aware of

devito/compiler.py
mloubout: Will do

devito/compiler.py
mloubout: Yes it does. It was mainly to differentiate with `compile_and_load`

devito/stencilkernel.py
mloubout: Caught by my laziness, will do :)

devito/stencilkernel.py
FabioLuporini: sounds better

devito/stencilkernel.py
```diff
@@ -97,7 +98,8 @@ def __init__(self, stencils, **kwargs):
 
         # Introduce C-level profiling infrastructure
         self.sections = OrderedDict()
-        nodes = self._profile_sections(nodes)
+        if self.profiler:
```
FabioLuporini: Instead of polluting the code with this sort of 
``if self.profiler`` "do this" ``else`` "do something else"
I think there's a better way, but I'm interested in @mlange05 thought as well.

So here's what I propose.

We add a ``__new__`` method to ``StencilKernel``. If ``profile == True``, then we simply return a standard ``StencilKernel``, which as by default now, **does profile the C code**. Otherwise, we return a ``StencilKernelBasic`` -- please, find a better name -- which does "less" then a standard ``StencilKernel``. In particular, it does **not** perform C-level profiling. This is easily achievable by having an empty ``_profile_sections()`` and by tweaking the other functions accordingly. 

This should be useful for other future features as well (not just profiling), while leaving the implementation cleaner. 

@mloubout Of course, it's not just these three ``if self.profile`` that were worrying me, but rather the fact that if we don't deal with this **general** problem now (the fact that we may wanna have a ``StencilKernel`` for normal execution and a ``StencilKernelBasic`` for debug/... mode), then the situation might escalate badly rapidly

mloubout: I get your point, but not really the aim of this PR

FabioLuporini: I'll do this then..

Btw, curiosity: what's the reason you explocitly don't want C-level profiling?

mlange05: @fablup Sounds good to me. This should indeed allow us to have a much more detailed "debug" mode, which sounds great.

mloubout: >Btw, curiosity: what's the reason you explocitly don't want C-level profiling?

Don't need it for Julia, and a pain to handle from julia even f we wanted it.

mloubout: @mlange05 @FabioLuporini 
Again, profiling/bug is not the point of this PR. Can we focus on this and make another one for debug if there is a better way to do it.

FabioLuporini: Yes, but the point is that *if* one PR introduces a problem, then it should also fix it -- it shouldn't be the case that one PR introduces a problem which *someone* maybe/one day/in the future/perhaps/... will fix

This is an approach I was trying to avoid.

mloubout: I'm fine doing it part of this PR. The Julia side is ready and I need  to merge this first then the Julia side.

FabioLuporini: To help you with this exercise, I pushed a commit that does most of the stuff in the branch ``stencil-kernel-basic``, based on top of this one. I didn't even try to run it, it's just there to hopefully help you understanding what's the structure that I'd like to obtain eventually

devito/stencilkernel.py
```diff
@@ -513,6 +583,22 @@ def cfunction(self):
         return self._cfunction
 
     @property
+    def cfunctionJ(self):
```
FabioLuporini: There's again some code duplication with ``cfunction``, which should be factored in a common routine

Maybe this is yet another symptom that what you want is a specialized ``StencilKernel``; that is, that you should extend ``StencilKernelBasic`` (or ``OperatorBasic``, as I called it in the new ``stencil-kernel-basic`` branch) to make things properly structured

devito/stencilkernel.py
```diff
@@ -217,6 +218,70 @@ def apply(self, *args, **kwargs):
 
         return summary
 
+    def argument_list(self, **kwargs):
```
mloubout: This is not duplicated, data initialization is also removed and dim_size check are different

devito/stencilkernel.py
mloubout: Once again, not the same.

mloubout:
>pushed a commit that does most of the stuff in the branch stencil-kernel-basic

And just doing it here would have been to simple?

This PR is not a bout pretty. We NEED this to run devito and work with it here and I'm not gonna wait a month and tell people, oh no it's not pretty enough so find something else. It passes every tests, there is nothing wrong with the code and a lot of the comments are nitpicking.

FabioLuporini:
> And just doing it here would have been to simple?

Mmhh...let me think. Why did I not do that? Why do branches even exist? Not sure

> This PR is not a bout pretty. We NEED this to run devito and work with it here and I'm not gonna wait a month and tell people, oh no it's not pretty enough so find something else. It passes every tests, there is nothing wrong with the code and a lot of the comments are nitpicking.

I don't blame you, one day you'll understand that this wasn't about prettiness and that "it works != it's correct" (just like maths btw).

Now, back to the PR.

I worked on this a little more and pushed a new commit **into this branch** (as it seemed you prefer). The commit counts lots of edits but most of them is actually just shuffling code around. Please, take a look and review. I changed a few names and a few interfaces: now from Julia you should call ``compile`` (rather than ``cfunctionJ`` and ``arguments`` (rather than ``arguments_list``). ``arguments`` returns two values, but probably you guys just care about the first one. Can you do the rest of the surgery in Julia and let me know how I can help to fix/improve any outstanding issues? I clearly care as much as you do about merging this ASAP!

Oh, I was about to forget: if you call ``arguments`` with an empty ``kwargs`` you should get exactly the same behaviour that you originally encoded in ``arguments_list``. This simple tweak, plus method overloading, allowed me to capture lots of duplicated code.

Let me know how this goes.

FabioLuporini:
Again, I forgot.

Just pass the special keyword ``external`` to obtain an Operator that doesn't do C-level profiling (i.e., ``Stencilkernel(..., external=True)``). This is just a temporary hack. When we'll have a global configuration mechanism, this probably won't even be needed.

mloubout:
Well, nothing work anymore, I guess back to a week of debuging.

FabioLuporini:
> OrderedDict are unusable with Julia ...

This shouldn't be a big deal, I'll take a look today

> The compile function only gives back the library name and I need the Kernel name too.

Is the function name essentially ``stencilkernel.name`` ? if so, can you get it from there ?

FabioLuporini:
I'm also wondering ( @mlange05 ) how could we write a test to stress the Julia interface and make sure that when we touch ``StencilKernel``, we don't break it

FabioLuporini:
@mloubout Take a look at the latest commit, which adds a special class for Julia support. Now you should automatically get a list of tuples as in your original design

mloubout:
@FabioLuporini All good and works. Thanks a lot.

FabioLuporini:
@mlange05 are you happy with this? I think we can rebase on master and merge, if you are OK with the latest changes

devito/compiler.py
```diff
@@ -290,14 +272,30 @@ def load(basename, compiler=GNUCompiler):
...
     :param compiler: The toolchain used for compilation. GNUCompiler by default.
-    :return: The compiled library.
+
+    :return: The name of the compilation unit.
     """
-    jit_compile(ccode, basename, compiler=compiler)
 
-    return load(basename, compiler=compiler)
+    hash_key = sha1(str(ccode).encode()).hexdigest()
+    basename = path.join(get_tmp_dir(), hash_key)
```
mlange05: Why are we doing this in the compiler? Is the `basename` not a property of the Operator/Kernel that we might need to access the compiled libs externally?

mlange05: Oh I see, it's returned. Is there a rationale for not having this as a property instead? I'm thinking that caching might need this...?

mlange05:
Apart from the `basename` property question this looks good to me. Certain parts of the argument handling need a re-think maybe, but that is beyond this PR (and in fact is made easier through this). So, good to go from me.

devito/compiler.py
FabioLuporini: I wanted ``compiler.py`` to have the freedom to decide how to encode things like the filename etc. 

But yes, you're probably right about caching. Perhaps we can simply keep track of the returned basename in the Operator itself?

mlange05: Ok, I just wanted to make sure we're aware of this - it's not urgent enough to hold this PR any  further. We'll re-visit when/if we actually implement file-level operator caching, for now object-level should suffice. So, all in all good to merge.

FabioLuporini:
Alright, I just rebased this on top of master. If Travis is happy, I'll merge it.

FabioLuporini:
Merged.

PR #214, opened 2017-03-30
Change the type of block sizes to np.int32

### LINKED ISSUE #213: Errors when importing, running examples/benchmarks
I have installed devito (devito==2.0.1+378.gf88a31d) as described in the README: `pip install --user git+https://github.com/opesci/devito.git`. No errors during installation.

`import devito` results in: `ImportError: No module named 'devito.dse'`

`python examples/benchmark.py test -P tti -a -d 20 20 20 -n 5` (taken from .travis.yml) results in: `ValueError: unable to map dtype 'int64'`

`python example_diffusion.py run -m numpy devito --spacing 0.001 --timesteps 500` in the `examples/diffusion/` directory results in `ImportError: No module named 'devito.dse'`

Running `example_diffusion.py` from the base directory of the project (rather than `examples/diffusion/`) runs successfully.

navjotk:
- They don't need to be that big
- Breaks `cgen` on 32 bit machines

This is expected to fix #213

### DIFF
--- a/devito/propagator.py
+++ b/devito/propagator.py
@@ -739,7 +739,7 @@ class Propagator(object):
 
         for i, j in zip(self.block_sizes, self.space_dims):
             if i is not None:
-                self.fd.add_value_param("%sblock" % self._mapper[j], np.int64)
+                self.fd.add_value_param("%sblock" % self._mapper[j], np.int32)
 
     def add_inner_most_dim_pragma(self, inner_most_dim, space_dims, loop_body):
         """


mloubout:
Pretty sure this is gonna break Julia. I'll have to check.

navjotk:
For the future maybe we want tests to make sure we don't break the Julia interface by mistake.

navjotk:
Merging after verbal sign-off

PR #217, opened 2017-04-06
misssing dependency

mloubout:
Added missing dependencies to the `setup.py`


setup.py
```diff
@@ -10,7 +10,7 @@
       description="Finite Difference DSL for symbolic computation.",
       long_descritpion="""Devito is a new tool for performing
       optimised Finite Difference (FD) computation from high-level
-      symbolic problem definitions. Devito performs automated code
+      symbolic problem definitions. Devito performs automated codes
```
FabioLuporini: I think ``code`` was fine here

mloubout: I have no idea where this change comes from, don;t remember making it  :) Will put it back.

PR #219, opened 2017-04-06
all orders second derivative

mloubout:
Small PR that allow any order for the second derivative while the old version had hard-codded order [2,4,6,8,10,12,14,16] only.


FabioLuporini:
I think this is just OK once Travis is happy

mloubout:
All good?

mlange05:
All good. Merging...

Issue #220, opened 2017-04-07
Separating symbols and data


mlange05:
The current design of Devito function symbols is centered around the `SymbolicData` objects that always have an associated `.data` property. The original rationale for this is that we want control over data allocation, so that we may implement data layout changes (padding, vector folding, etc.). On the other hand, various use cases require us to be able to assign different data arrays to a particular symbol on operator invocation (`Operator/StencilKernel.apply()`). As an example we might want to do:
```
shape  = (...)
m0 = DenseData(name='m0', shape=shape)
m1 = DenseData(name='m1', shape=shape)

eqn = some_equation_involving_m
op = Operator(eqn, subs=...)

op.apply(m=m0)
op.apply(m=m1)
```

In the current and previous seismic examples (`gradient_test.py`) this problem is avoided by creating the `m` symbol during the operator setup and copying the data into `m.data`, without any operator re-use. In order to avoid the potentially costly copy operation this implies, we need to adjust our API to allow multiple data objects to be associated with symbols *at runtime*.

mlange05:
Some possible suggestions (although this is not an exhaustive list):

a) We _could_ turn `DenseData/TimeData` into "pure symobl" objects that still carry data layout and shape information (required for the compiler to adhere to things like folding or padding). But instead of having a single `.data` property these would provide an `.allocate()` function that returns `TensorData` objects to encapsulate actual data. This would put the job of managing and associating data into the users hands, but allow us to do enough sanity checking that the data provided at `.apply()` agrees with all assumptions made by the operator during codegen.

b) Somehow over-write our function symbol caching so that multiple explicit invocations of `DenseData(name='m', ...) don't alias. This would allow us to natively inject different `m` objects into the operator with `op.apply(m=m0)` and `op.apply(m=m1)`.

c) Another variation that somewhat combines a) and b), where we allow aliasing at operator invocation time, ie. we allow:
```
m0 = DenseData(name='m0')
m1 = DenseData(name='m1')
...
op.apply(m0=m0)
op.apply(m0=m1)
```

Food for though, other suggestions or opinions are very welcome.

navjotk:
I agree with the general rationale behind this issue and this is indeed a problem. 

[...]

PR #223, opened 2017-04-19
Allow symbol data to be substituted at the time of applying the Operator

navjotk:
In response to #220 
This PR adds:
- The ability to do `op.apply(m=m1)`
- A test for this

There is one tiny caveat in this implementation though. In order to do `op.apply(m=m1)`, both `m` and `m1` need to have the same `name` property. In case this is not true, the expected syntax is `op.apply(m=m1.data)`. 
Edit: Never mind, the above has been fixed as well. `op.apply(m=m1)` would work regardless of the symbol names.

### DIFF
--- a/devito/stencilkernel.py
+++ b/devito/stencilkernel.py
@@ -130,9 +130,8 @@ class OperatorBasic(Function):
 
         # Have we been provided substitutes for symbol data?
         # Only SymbolicData can be overridden with this route
-        r_args = [f_n for f_n, f in arguments.iteritems()
-                  if isinstance(f, SymbolicData)]
-        o_vals = OrderedDict([arg for arg in kwargs.iteritems() if arg[0] in r_args])
+        r_args = [f_n for f_n, f in arguments.items() if isinstance(f, SymbolicData)]
+        o_vals = OrderedDict([arg for arg in kwargs.items() if arg[0] in r_args])
 
         # Replace the over-ridden values with the provided ones
         for argname in o_vals.keys():
--- a/devito/stencilkernel.py
+++ b/devito/stencilkernel.py
@@ -24,6 +24,7 @@ from devito.tools import (SetOrderedDict, as_tuple, filter_ordered, filter_sorte
 from devito.visitors import (FindNodes, FindSections, FindSymbols, FindScopes,
                              IsPerfectIteration, MergeOuterIterations,
                              ResolveIterationVariable, SubstituteExpression, Transformer)
+from devito.exceptions import InvalidArgument
 
 __all__ = ['StencilKernel']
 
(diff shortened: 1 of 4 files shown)


tests/test_stencil_arithmetic.py
```diff
@@ -140,3 +140,24 @@ def test_arithmetic_indexed_open_loops(i, j, k, l, expr, result):
     eqn = eval(expr)
     StencilKernel(eqn)(fa)
     assert np.allclose(fa.data[1, 1:-1, 1:-1], result[1:-1, 1:-1], rtol=1e-12)
+
+
+def test_override(i, j, k, l):
+    """Test that the call-time overriding of Operator arguments works"""
+    a = symbol(name='a', dimensions=(i, j, k, l), value=2., mode='indexed').base.function
+    a1 = symbol(name='a', dimensions=(i, j, k, l), value=3., mode='indexed').base.function
```
mlange05: I wonder why these two symbols not cache on the SymPy symbol-level? The test seems to work though, so I'm happy to go ahead.

tests/test_stencil_arithmetic.py
```diff
@@ -140,3 +140,24 @@ def test_arithmetic_indexed_open_loops(i, j, k, l, expr, result):
...
+    op(a=a2)
+    shape = [d.size for d in [i, j, k, l]]
+
+    assert(np.allclose(a.data, np.zeros(shape) + 5))
+    assert(np.allclose(a1.data, np.zeros(shape) + 6))
+    assert(np.allclose(a2.data, np.zeros(shape) + 7))
+
+
+if __name__ == "__main__":
+    test_override(i(), j(), k(), l())
```
mlange05: I don't like hijacking the main for one particular test. Why don't you use `py.test -k` to run individual tests instead?

devito/stencilkernel.py
```diff
@@ -127,10 +128,22 @@ def arguments(self, *args, **kwargs):
         arguments = OrderedDict([(arg.name, arg) for arg in self.parameters])
         dim_sizes = {}
 
+        # Have we been provided substitutes for symbol data?
+        # Only SymbolicData can be overridden with this route
+        r_args = [f_n for f_n, f in arguments.items() if isinstance(f, SymbolicData)]
```
FabioLuporini: ``if f.is_SymbolicData`` ?

tests/test_stencil_arithmetic.py
mloubout: >I wonder why these two symbols not cache on the SymPy symbol-level? 
Does this mean some symbols may never clear their data, would be problematic

mlange05:
Ok, so the cache misses on new `SymbolicData` instantiations with the same name are actually a known problem that was flagged in PR #164. There is an xfailing test already in `test_symbol_caching.py` that highlights this. I think filing a separate issue for this should be ok, so we can go ahead and land this PR.

tests/test_stencil_arithmetic.py
mlange05: Ping

devito/stencilkernel.py
mlange05: Ping

navjotk: The objects here that are not `SymbolicData` are `BufferedDimension` objects. These objects don't have this property. This leads to `AttributeError: 'BufferedDimension' object has no attribute 'is_SymbolicData'`

mlange05:
Looks good to me. Will merge once travis is happy after update.

PR #224, opened 2017-04-20
Allow pragma to be set by env variable

navjotk:
When using the `CustomCompiler`, the pragma used for ignoring vector dependencies is the one for the intel compiler. This change allows the pragma to be set with an environment variable.
This continues the thought behind having a compiler completely configurable by environment variables.

### DIFF
--- a/devito/compiler.py
+++ b/devito/compiler.py
@@ -192,7 +192,7 @@ class CustomCompiler(Compiler):
         default = '-O3 -g -march=native -fPIC -mno-avx -Wall -std=c99'
         self.cflags = environ.get('CFLAGS', default).split(' ')
         self.ldflags = environ.get('LDFLAGS', '-shared').split(' ')
-        self.pragma_ivdep = [Pragma(environ.get('DEVITO_PRAGMA', 'GCC ivdep'))]
+        self.pragma_ivdep = [Pragma(environ.get('DEVITO_IVDEP', 'GCC ivdep'))]
         if self.openmp:
             self.ldflags += environ.get('OMP_LDFLAGS', '-fopenmp').split(' ')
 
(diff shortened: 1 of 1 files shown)


devito/compiler.py
```diff
@@ -192,7 +192,7 @@ def __init__(self, *args, **kwargs):
         default = '-O3 -g -march=native -fPIC -mno-avx -Wall -std=c99'
         self.cflags = environ.get('CFLAGS', default).split(' ')
         self.ldflags = environ.get('LDFLAGS', '-shared').split(' ')
-
+        self.pragma_ivdep = [Pragma(environ.get('DEVITO_PRAGMA', 'GCC ivdep'))]
```
mlange05: Shouldn't the env name be slightly more specific? Maybe `DEVITO_IVDEP` or `DEVITO_PRAGMA_IVDEP`?

navjotk: I reiterate this sentiment on the use of `t` vs `time` in the entire codebase :)

FabioLuporini:
OK

does this impact StencilKernel too?i wouldn't think so but maybe Im wrong

devito/compiler.py
mlange05: I'm not sure the above comment was meant to be made here... ?

mlange05:
Compiler presets are agnostic to the operator classes, so yes, this will affect `StencilKernel` as well.

mlange05:
Good to go, so merging...

FabioLuporini:
> Compiler presets are agnostic to the operator classes, so yes, this will affect StencilKernel as well.

Sure, but I was just wondering whether that `.pragma_ivdep` was even used in 3.0, as I don't remember I needed it. Anyway, fine.

PR #229, opened 2017-04-28
`PointData` generalisation

mlange05:
This merge moves the `PointData` class into it's own sub-module and provides an explicit API for supporting `point.interpolate(expr)` and `point.inject(expr)` methods to perform sparse-point interpolation of custom user-defined expressions in and out of grid variables. This is a pre-requisite for providing more general source/receiver terms in our seismic examples. 

Note, the merge does not yet integrate this new API with our actual examples, since there is more required to make this shine (to come in a separate PR). The current test set is also limited, in that it only interpolates the expression "1" at the moment. Input with more advanced expressions and their analytical solutions would be greatly appreciated.

Other details in this PR:
* We now store the sparse point coordinates as plan `DenseData` objects, removing the previous `CoordinateData` class.
* Moved `PointData` to it's own sub-module, since all it contains is the automated symbolic derivation of the interpolation expressions, which is conceptually quite far removed from the rest of things residing in `interfaces.py`.
* A small DLE bug-fix to avoid access to an empty `mapper` dict

### DIFF
--- a/devito/pointdata.py
+++ b/devito/pointdata.py
@@ -192,5 +192,5 @@ class PointData(DenseData):
         # Substitute coordinate base symbols into the coefficients
         subs = OrderedDict(zip(self.point_symbols, self.coordinate_bases))
         return [Eq(field.subs(vsub),
-                   field.subs(vsub) + expr.subs(subs) * b.subs(subs))
+                   field.subs(vsub) + expr.subs(subs).subs(vsub) * b.subs(subs))
                 for b, vsub in zip(self.coefficients, idx_subs)]
--- a/devito/pointdata.py
+++ b/devito/pointdata.py
@@ -1,7 +1,9 @@
-from sympy import Matrix, symbols
+from sympy import Function, Matrix, symbols
 from sympy.abc import h
 
+from collections import OrderedDict
 from devito.dimension import d, p, t
+from devito.dse.inspection import indexify, retrieve_indexed
 from devito.interfaces import DenseData
 from devito.logger import error
 
@@ -25,6 +27,7 @@ class PointData(DenseData):
     def __init__(self, *args, **kwargs):
         if not self._cached():
             self.nt = kwargs.get('nt')
+            self.h = kwargs.get('h')
             self.npoint = kwargs.get('npoint')
             self.ndim = kwargs.get('ndim')
             kwargs['shape'] = (self.nt, self.npoint)
@@ -66,9 +69,8 @@ class PointData(DenseData):
         # Grid indices corresponding to the corners of the cell
         x1, y1, z1, x2, y2, z2 = symbols('x1, y1, z1, x2, y2, z2')
         # Coordinate values of the sparse point
-        px, py, pz = symbols('px, py, pz')
+        px, py, pz = self.point_symbols
         if self.ndim == 2:
-            self._increments = ((0, 0), (0, 1), (1, 0), (1, 1))
             A = Matrix([[1, x1, y1, x1*y1],
                         [1, x1, y2, x1*y2],
                         [1, x2, y1, x2*y1],
@@ -80,8 +82,6 @@ class PointData(DenseData):
                         [px*py]])
 
         elif self.ndim == 3:
-            self._increments = ((0, 0, 0), (0, 1, 0), (1, 0, 0), (0, 0, 1),
-                                (1, 1, 0), (0, 1, 1), (1, 0, 1), (1, 1, 1))
             A = Matrix([[1, x1, y1, z1, x1*y1, x1*z1, y1*z1, x1*y1*z1],
                         [1, x1, y2, z1, x1*y2, x1*z1, y2*z1, x1*y2*z1],
                         [1, x2, y1, z1, x2*y1, x2*z1, y2*z1, x2*y1*z1],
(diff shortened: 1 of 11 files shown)


mloubout:
Looks good to me, will have to look at how to combine it with the examples.

FabioLuporini:
All is good. 
Random thought: maybe this is the right occasion to factor out the Devito data hierarchy (currently: interfaces.py + pointdata.py) in a separate folder ?

FabioLuporini:
> Random thought: maybe this is the right occasion to factor out the Devito data hierarchy (currently: interfaces.py + pointdata.py) in a separate folder ?

Forget this. I don't like the idea I think

devito/pointdata.py
```diff
@@ -0,0 +1,197 @@
...
+                        [px*py],
+                        [px*pz],
+                        [py*pz],
+                        [px*py*pz]])
+        else:
+            error("Interpolation coefficients not implemented for "
+                  "%d dimensions." % self.ndim)
+
+        # Map to reference cell
+        reference_cell = {x1: 0, y1: 0, z1: 0, x2: h, y2: h, z2: h}
```
navjotk: Instead of hard-coding `h` here, `h` could be a property of each dimension and be pulled in from there. 
Hard-coded `h` has to be removed eventually. This line makes it harder to remove.

devito/pointdata.py
```diff
@@ -0,0 +1,197 @@
...
+            p = Matrix([[1],
+                        [px],
+                        [py],
+                        [pz],
+                        [px*py],
+                        [px*pz],
+                        [py*pz],
+                        [px*py*pz]])
+        else:
+            error("Interpolation coefficients not implemented for "
```
navjotk: I think we need more than a log message here. Exception would be my recommendation.

devito/pointdata.py
```diff
@@ -0,0 +1,197 @@
...
+        return symbols('px, py, pz')
+
+    @property
+    def point_increments(self):
+        """Index increments in each dimension for each point symbol"""
+        if self.ndim == 2:
+            return ((0, 0), (0, 1), (1, 0), (1, 1))
+        elif self.ndim == 3:
+            return ((0, 0, 0), (0, 1, 0), (1, 0, 0), (0, 0, 1),
+                    (1, 1, 0), (0, 1, 1), (1, 0, 1), (1, 1, 1))
```
navjotk: This could be re-written in a form such that it works for any number of dimensions

tests/test_point_data.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+    points = points_horizontal
+    spacing = a.data[1, 1]
+    a.data[:] = 0.
+    expr = points.inject(a, Function('FLOAT')(1.))
+    StencilKernel(expr, dse=None, dle=None, subs={h: spacing})(t=1)
+    assert np.allclose(a.data[1:-1, 4:6], 1., rtol=1.e-6)
+
+
+def test_inject_from_field_2d(a, points_horizontal):
```
navjotk: No tests for 3D?

devito/pointdata.py
mlange05: Good catch, thanks!

devito/pointdata.py
mlange05: In theory I agree, but without the `Grid` abstraction defining what the spacing term, a dimension cannot know if it is a spatial dimension and thus cannot decide whether it should use `h` or something else. So until we have a `Grid` abstraction to define spatial dimensions explicitly, I think we should stick with the Highlander principle for the spacing symbol in spatial dimension (There can only be one - it is `h` and comes from sympy).

devito/pointdata.py
mlange05: True, but I would argue beyond the scope of this PR.

tests/test_point_data.py
mlange05: Yes, will add, thanks.

devito/pointdata.py
mloubout: > (There can only be one - it is h and comes from sympy).

I actually was about to make an update to make `spacing` a tuple `(hx, hy, hz)` to improve generality and variable grids. It is a small update and I will look at ow to make it compatible with this. This in my opinion still independent of a `Grid` abstraction as it is only one parameter that does not care about indices or origin of the grid.

mlange05: Yeah, the `h` assumption is hard-coded in many places atm, so picking it apart only here will have little effect. I'd hold off until we are building a proper `Grid` abstraction and can test the non-square/non-cube cases for stencils and point interpolation - hopefully this will be very soon.

mlange05:
Ok, exceptions added, tests are generalised to 2D and 3D and rebased on master. Please review.

devito/pointdata.py
navjotk: AFAIK everything else in core devito is independent of the number of dimensions and with this you are introducing a restriction on the supported dimensionality. Maintaining generality is the scope of every PR.

FabioLuporini: > everything else in core devito is independent of the number of dimensions

In "core devito" yes, but the data types provided by ``interfaces.py`` are currently constrained to maximum 3D AFAIK. This is still a limitation of the old, legacy Devito. 

Said that, yes, it'd be good to generalise that, though not sure if strictly necessary for this PR.

navjotk: a) I remember there being tests for the data types with 5/6 dimensions so I don't know at what point this restriction to 3D was introduced, if at all. 
b) Assuming that the old code is broken, it is still no justification for writing more broken code.
c) Assuming that the old code is broken, it will still be corrected at some point. Why make it harder to fix by increasing the number of places where the assumption of only 2 and 3 dimensions is hard-coded?

FabioLuporini:
> a) I remember there being tests for the data types with 5/6 dimensions so I don't know at what point this restriction to 3D was introduced, if at all.

Maybe, as I said, "AFAIR", because I remember seeing the explicit Dimensions ``t, x, y, z`` in some of the constructors of TimeData, etc.

> b) Assuming that the old code is broken, it is still no justification for writing more broken code.

Agreed, but I'd use "non-general" rather than "broken"

> c) Assuming that the old code is broken, it will still be corrected at some point. Why make it harder to fix by increasing the number of places where the assumption of only 2 and 3 dimensions is hard-coded?

Agreed

devito/pointdata.py
FabioLuporini: > a) I remember there being tests for the data types with 5/6 dimensions so I don't know at what point this restriction to 3D was introduced, if at all.

Maybe; as I said, "AFAIR", because I remember seeing the explicit Dimensions ``t, x, y, z`` in some of the constructors of TimeData, etc.

> b) Assuming that the old code is broken, it is still no justification for writing more broken code.

Agreed, but I'd use "non-general" rather than "broken"

> c) Assuming that the old code is broken, it will still be corrected at some point. Why make it harder to fix by increasing the number of places where the assumption of only 2 and 3 dimensions is hard-coded?

Agreed

mlange05: > a) I remember there being tests for the data types with 5/6 dimensions so I don't know at what point this restriction to 3D was introduced, if at all.

Stencil operations over dense grids are still as generic as ever, with as many dimensions as the user cares to provide.

>  b) Assuming that the old code is broken, it is still no justification for writing more broken code.

Nothing in either code is "broken" - Sparse point interpolation, a specific edge-feature that existed in the old and the new code, is merely not generalised beyond 2D/3D and clearly labelled as such. None of the core features are impacted by this.

> c) Assuming that the old code is broken, it will still be corrected at some point. Why make it harder to fix by increasing the number of places where the assumption of only 2 and 3 dimensions is hard-coded?

The coefficients and steppings have always been hardcoded, so nothing is "made harder". Sparse point interpolation is still a fairly isolated feature, so it can be generalised at any point without impacting any other features - as soon as someone finds enough time and motivation to do so.

mlange05:
Ok, merging...

PR #232, opened 2017-05-04
The Operator is dead - long live the Operator!

mlange05:
EDIT: Stopgap fixes for DSE failures are in, so we can run new-style API with full optimisations.
**THIS IS NOW READY TO MERGE.**

This branch removed the legacy constructs `Operator`, `Propagator`,`AutoTuner` and `FunctionaManager` and renames the `StencilKernel` to become the new `Operator`. It also removes the `legacy` flag from the seismic tests and examples and removes old tests that were hard-coded specifically for legacy-operators/propagators.

Caveats:
* This removes `test_oi.py`, which was hard-coded for `Propagators`. Do we have an equivalent test for the new-style OI derivation?
* `test_rewriter.py` used the implicit symbol return signature of old-style operators. The test has been tweaked to be compatible with the new style, but this is not particularly nice. It will be fixed at a later stage, after the seismic/tti operators have been improved.
* Also `test_rewriter.py`: One sub-test uses `operator.stencils`, which we don;t record anymore. This has been temporarily disabled.


FabioLuporini:
Lovely. I will also have to contribute to this PR, because some stuff in the DSE inspection utilities can now be dropped.

I also seem to remember, but perhaps I'm wrong, that some stuff in devito/tools.py could be dropped. I'll check this.

navjotk:
Does the new `Operator` have a python-only debug mode?

mlange05:
Alas, no. Was the python-debug mode ever really used by people? Is this a feature we care about?

Also, for what it's worth, we do still have an example that does the "lambdify-equivalent" of a Devito operator for reference.

mlange05:
Ok, stopgap DSE fixes are in and the branch has been rebased on master. Should be ready to go once travis is happy. Do we regard the lack of a Python-debug mode as a show-stopper? @navjotk

FabioLuporini:
I think this can be merged

navjotk:
>Also, for what it's worth, we do still have an example that does the "lambdify-equivalent" of a Devito operator for reference.

Where is this test?

mlange05:
`examples/diffusion/example_diffusion.py:80`

tests/test_operator.py
```diff
@@ -1,43 +0,0 @@
...
-        input_grid = DenseData(name="input_grid", shape=(3, 2), dtype=np.float64)
-        input_grid.data[:] = np.arange(6, dtype=np.float64).reshape((3, 2))
-        output_grid = DenseData(name="output_grid", shape=input_grid.shape,
-                                dtype=input_grid.dtype)
-        eq = Eq(output_grid.indexed[t, x], input_grid.indexed[t, x] + 3)
-        op = SimpleOperator(input_grid, output_grid, [eq])
-        op.apply(debug=True)
-        assert(output_grid.data[2][1] == 8)
-
-    def test_compare(self):
```
navjotk: I think this test is quite unique in that it is the only one that directly tests the correctness of the generated code against a "reference implementation". This should definitely be rewritten for the new operator.

mlange05:
Ok, as discussed in the dev call, while the Python-debug feature and the like-for-like test are clearly desirable, they require a non-trivial amount of work and are not critical at this point. I'll file an issue to make sure we re-instate this.

mlange05:
And with that we are good to go. So long, old friend. Merging...

PR #233, opened 2017-05-04
allow constant physical params

mloubout:
This is a small improvement of `Model` so that constant parameters can be used.

An example `example_constant` is added to show it.

This is not very pretty I would say, so if you have a better way to do it, let me know.

### DIFF
--- a/examples/seismic/model.py
+++ b/examples/seismic/model.py
@@ -50,16 +50,20 @@ class Model(object):
     :param m: The square slowness of the wave
     :param damp: The damping field for absorbing boundarycondition
     """
-    def __init__(self, origin, spacing, vp, nbpml=0, dtype=np.float32,
-                 rho=None, epsilon=None, delta=None, theta=None, phi=None):
+    def __init__(self, origin, spacing, shape, vp, nbpml=0, dtype=np.float32,
+                 epsilon=None, delta=None, theta=None, phi=None):
         self.vp = vp
         self.origin = origin
         self.spacing = spacing
         self.nbpml = nbpml
         self.dtype = dtype
+        self.shape = shape
         # Create square slowness of the wave as symbol `m`
-        self.m = DenseData(name="m", shape=self.shape_domain, dtype=self.dtype)
-        self.m.data[:] = self.pad(1 / (self.vp * self.vp))
+        if isinstance(vp, np.ndarray):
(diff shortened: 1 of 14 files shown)


examples/seismic/model.py
```diff
@@ -50,61 +50,72 @@ class Model(object):
     :param m: The square slowness of the wave
     :param damp: The damping field for absorbing boundarycondition
     """
-    def __init__(self, origin, spacing, vp, nbpml=0, dtype=np.float32,
-                 rho=None, epsilon=None, delta=None, theta=None, phi=None):
+    def __init__(self, origin, spacing, shape, vp, nbpml=0, dtype=np.float32,
```
mlange05: I'm not sure why we need to make this explicit? The information is given by `vp.shape`, so why not pull it from there? Forcing the user to supply this explicitly only makes sense if we were to make `vp` optional, or am I missing a use case here?

mlange05:
Apart from the additional `shape` parameter to the `Model` this looks nice.

examples/seismic/model.py
mloubout: This is necessary if the field is constant as only a vlaue is passed and `vp.shape = ()`

mlange05: Ahhhhh, I see. Ok then, good to go from me.

examples/seismic/model.py
```diff
@@ -50,61 +50,72 @@ class Model(object):
...
         self.dtype = dtype
+        self.shape = shape
         # Create square slowness of the wave as symbol `m`
-        self.m = DenseData(name="m", shape=self.shape_domain, dtype=self.dtype)
-        self.m.data[:] = self.pad(1 / (self.vp * self.vp))
+        if isinstance(vp, np.ndarray):
+            self.m = DenseData(name="m", shape=self.shape_domain, dtype=self.dtype)
+            self.m.data[:] = self.pad(1 / (self.vp * self.vp))
+        else:
+            self.m = 1/vp**2
```
navjotk: Looks like you're storing both `vp` and `m` in memory. Since `m=1/vp**2`, isn't this like storing two copies of the same information?

mloubout: Yes it is, I guess `vp` can be dropped, needs some changes in the functions accessing it.

Issue #228, opened 2017-04-25
StencilKernel dimensions check


mloubout:
This about a bug I just found out trying to figure out why `adjointJ` wasn't working anymore.

`f_args = [f for f in arguments.values() if isinstance(f, SymbolicData)]` Line 144 in `StencilKernel` produces something like :

`[damp(x, y), m(x, y), rec(time, p_rec), rec_coords(p_rec, d), src(time, p_src), src_coords(p_src, d), u(t + s, x, y)]`

and form this, the dimensions are inferred as : 

`{t: 3, d: 2, y: 130, time: 215, x: 120}`

However, if the order is changed to:
`[u(t + s, x, y), damp(x, y), m(x, y), rec(time, p_rec), rec_coords(p_rec, d), src(time, p_src), src_coords(p_src, d)]`

the dimesion size become:

`{d: 2, y: 90, time: 3, x: 80, t: 3, x_block: 8}`

and `time` has the wrong size.

This happens in Born because one of the field is called `U` and that end up first in the list because it's an upper case. renaming it `ub` fixed it but this is a very dangerous bug to leave in there.

mlange05:
Fixed via #234, I believe. @mloubout can you confirm and close please?

mloubout:
Looks good, closing.

PR #234, opened 2017-05-05
Fix wrong size inferred for buffered dimension

navjotk:
Issue #228 points out a case where changing the name of a `SymbolicData` object changes the order in which dimension sizes are inferred and this causes the size of the `time` dimension to be inferred incorrectly. 

This was being caused because of the way sizes are being inferred. Particularly, when a `BufferedDimension` is encountered, a parent dimension was [added](https://github.com/opesci/devito/blob/master/devito/stencilkernel.py#L166) to the list of `Dimension`s if not already present. This last condition introduced a requirement on the order in which the `Dimension`s are presented to this loop since if the parent `Dimension` appears later in the list of dimensions, its size would already be set by the child `Dimension` and its real size would never be captured. 

This PR fixes that by moving the check for parent dimension sizes outside the loop that traverses the stencil expression. This way, we have ensured that the entire expression has been checked for the presence of the parent `Dimension`. 

Caveat: This PR introduces an extra loop of the order of the number of dimensions in the expression but removes the dependency on the ordering of the dimensions.

### DIFF
--- a/devito/stencilkernel.py
+++ b/devito/stencilkernel.py
@@ -161,13 +161,15 @@ class OperatorBasic(Function):
                     else:
                         # Derive size from grid data shape and store
                         dim_sizes[dim] = shape[i]
-
-                    # Ensure parent for buffered dims is defined
-                    if dim.is_Buffered and dim.parent not in dim_sizes:
-                        dim_sizes[dim.parent] = dim_sizes[dim]
                 else:
                     if not isinstance(dim, BufferedDimension):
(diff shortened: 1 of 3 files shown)


mlange05:
Looks very nice, thanks! Can we add a small test for the size derivation somewhere (maybe `test_stencil_arithmetic.py`)?

FabioLuporini:
Thanks for this

mlange05:
Very neat, thanks again. Merging...

Issue #227, opened 2017-04-25
Seismic example re-structuring


mlange05:
The current seismic examples (Acoustic and TTI) both use explicit wrappers for `Forward`, `Adjoint`, `Gradient` and `Born` wrapper methods that create new operators for every kernel invocation. This prohibits re-use in tight loops, such as checkpoint-based RTM, because every application of the wrapper bypasses the operator-level caching and triggers a full DSE/DLE/codegen/compilation cascade. We should re-structure the examples to, at the very least, utilise pre-setup operators for the different invocations, or even allow a seamless argument injection into the pre-defined operator objects.

The seismic utility wrappers also previously instantiated all symbolic data object at invocation time, which causes costly memory-copy operations. Some of this has moved to a slightly higher level, but a more concrete model of data flow for the various types of grid data and the various desired use cases (adjoint test, gradient test, RTM, FWI, etc.) is required, in particular with focus on which data lives at which temporal granularity and who instantiates it (user-level, operator-wrapper).

mlange05:
As an example let's consider the checkpointing case with all data created by the user (either explicitly or through seismic-specific abstractions/utility classes:
```
# Data allocation
model = Model(velocity, ...)
src = RickerSource(src_coordinates, time_series)
rec = Receivers(rec_coordinates)
u = TimeData(u, ...)
v = TimeData(v, ...)

# Operator instantiation, wrappers return type Operator/StencilKernel
op_fwd = ForwardOperator(model=model, u=u, source=src, receivers=rec, ...)
op_grad = GradientOperator(model=model, v=v, ...)

# Incremental forward loop
ckp = Checkpointer()  # Something clever that handles checkpointing/scheduling
nsteps = 0
while nsteps > 0:
   nsteps = ckp.how_many_timesteps_should_I_run()
   op_fwd(u=u, m=model.m, time_size=nsteps, ...)
   ckp.do_your_thing(u)

for all checkpoints:
    u = ckp.retrieve_forward_wavefield()
    op_grad(v=v, u=u, m=model.m, ...)
```

The key questions arising from this are
* Can we make `op_fwd`/`op_grad` pure `Operator/StencilKernel` objects?
* When/where do we instantiate the wavefields? Should these have an additional abstraction to capture ABCs, or can we hide the PML on the `Model`??
* When/where are sources/receivers instantiated and populated?

Note: The above is a rough draft and likely to change.

mloubout:
[...]

PR #239, opened 2017-05-09
Source/receiver clean-up

mlange05:
This merge aims to reduce the parameter clutter that currently exists in the seismic examples and represents a first step towards solving issue #227. It primarily introduces a wrapper class for seismic source and receiver/shot symbols that replaces `SourceLike` and a short-hand notation to prescribe time indices in low-level "indexed" notation. In particular it brings:

* New class `PointSource`/`Receiver`/`Shot` that wraps `PointData` and can be created with a minimal set of parameters.
* New ellipsis short-hand notation that allows time-index substitutions like `a.indexed[t + 1, ...]` -> `a[t + 1, x, y, z]`. This is then used to enforce the right time index in source/receiver terms.
* Reduce clutter in Acoustic and TTI wrappers and kernel definitions by passing compiler and runtime arguments through to the operator via `**kwargs` rather than defining and repeating them at every layer.

Please note that this does not yet touch the data flow in the seismic examples, which is left for another PR.


mloubout:
The first thing is that I do not really like the separatation `Source/Receiver` as depending of the case, the Receivers can be a Point source and the Source a Receiver (shot gather) object (e.g in `Adjoint`). I would go for one object called "SeismicSource" or something like for all. 

Will come up with detailed comments later.

examples/acoustic/fwi_operators.py
```diff
@@ -34,25 +34,28 @@ def ForwardOperator(model, u, src, rec, data, time_order=2, spc_order=6,
...
 
-    dse = kwargs.get('dse', 'advanced')
-    dle = kwargs.get('dle', 'advanced')
+    # Construct expression to inject source values
+    # Note that src and field terms have differing time indices:
+    #   src[time, ...] - always accesses the "unrolled" time index
+    #   u[ti + 1, ...] - accesses the forward stencil value
+    ti = u.indices[0]
+    source = src.inject(field=u.indexed[ti + 1, ...], offset=model.nbpml,
+                        expr=src.indexed[time, ...] * dt * dt / m)
```
mloubout: I prefer the previous notation with `field=v, u_t=ti`, writting `.indexed` seems odd and not very adequate at the symbolic stencil level.

examples/acoustic/fwi_operators.py
```diff
@@ -82,25 +84,23 @@ def AdjointOperator(model, v, srca, rec, data, time_order=2, spc_order=6,
...
-    # Add substitutions for spacing (temporal and spatial)
-    subs = {s: dt, h: model.get_spacing()}
+    eqn = Eq(v.backward, stencil)
 
-    dse = kwargs.get('dse', 'advanced')
-    dle = kwargs.get('dle', 'advanced')
+    # Construct expression to inject receiver values
+    ti = v.indices[0]
+    receivers = rec.inject(field=v.indexed[ti - 1, ...], offset=model.nbpml,
+                           expr=rec.indexed[time, ...] * dt * dt / m)
```
mloubout: same.

examples/acoustic/fwi_operators.py
```diff
@@ -34,25 +34,28 @@ def ForwardOperator(model, u, src, rec, data, time_order=2, spc_order=6,
...
+    source = src.inject(field=u.indexed[ti + 1, ...], offset=model.nbpml,
+                        expr=src.indexed[time, ...] * dt * dt / m)
 
-    # Create stencil expressions for operator, source and receivers
-    eqn = Eq(u.forward, stencil)
-    src_add = src.point2grid(u, m, u_t=u.indices[0] + 1, p_t=time)
-    rec_read = Eq(rec, rec.grid2point(u, t=u.indices[0]))
-    stencils = [eqn] + src_add + [rec_read]
+    # Create interpolation expression for receivers
+    receivers = Eq(rec, rec.interpolate(expr=u, offset=model.nbpml))
```
mloubout: Does this also required `indexed` if not at the current time step?

examples/acoustic/fwi_operators.py
```diff
@@ -137,25 +137,20 @@ def GradientOperator(model, v, grad, rec, u, data, time_order=2, spc_order=6,
...
-    # Add Gradient-specific updates. The dt2 is currently hacky
-    #  as it has to match the cyclic indices
+    eqn = Eq(v.backward, stencil)
 
-    dse = kwargs.get('dse', 'advanced')
-    dle = kwargs.get('dle', 'advanced')
+    # Add expression for receiver injection
+    ti = v.indices[0]
+    receivers = rec.inject(field=v.indexed[ti - 1, ...], offset=model.nbpml,
+                           expr=rec.indexed[time, ...] * dt * dt / m)
```
mloubout: Can offset be a "self" property instead, carrying it around looks error prone as it is a lot easier to forget it at one place and destroy everything, at least as a property, even if missed, everything will still be consistent (no different offset for src and rec for example if mistake)

examples/acoustic/fwi_operators.py
```diff
@@ -137,25 +137,20 @@ def GradientOperator(model, v, grad, rec, u, data, time_order=2, spc_order=6,
...
 
-    # Create stencil expressions for operator, source and receivers
-    eqn = Eq(v.backward, stencil)
-    rec_add = rec.point2grid(v, m, u_t=t - 1, p_t=time)
-    stencils = [eqn] + [gradient_update] + rec_add
-    return Operator(stencils=stencils, subs=subs, dse=dse, dle=dle,
-                    time_axis=Backward, name="Gradient")
+    return Operator(stencils=[eqn] + [gradient_update] + receivers,
+                    subs={s: dt, h: model.get_spacing()},
+                    time_axis=Backward, name='Gradient', **kwargs)
```
mloubout: stencils = ...
Operator(stencil= stencil,...)

easier to read in my opinion.

examples/acoustic/fwi_operators.py
FabioLuporini: True. Other possible shorthand notations?

examples/acoustic/fwi_operators.py
FabioLuporini: nitpicking

FabioLuporini:
The restructuring looks very good. Let's address @mloubout 's comments though

examples/acoustic/fwi_operators.py
mlange05: Ok, so first thing to note is that the previous `point2grid` function call was simply hiding some very ugly notation that derived the correct indices by hand and hard-codes the expression `src * dt * dt / m` - therefore it cannot really be generalised (see `source_type.py:99-109`). This is in contrast is raw **Devito low-level notation** (ie. our backdoor), as indicated by the `indexed` keyword.

The reason I can only provide the ellipsis shorthand in low-level notation (and believe me I tried doing it on the high-level!), is simply because over-writing `__getitem__` on a `sympy.Function` causes a multitude of low-level SymPy failures in things like `sympy.solve`, because we all of a sudden provide an iteration protocol, although `Function` objects are not actually iterable. I tried things like raising a failure through `__iter__`, but alas no luck. If someone can provide another idea on how to prevent `Function` iteration, we could move this trick to the high-level, so further suggestions are welcome...

FabioLuporini: ugh, OK, I understand. 
Random thought: what about overloading a different operator ?

FabioLuporini:
> If the PointData class is doing all the heavy lifting of interpolation, I don't think the PointSource, Receiver and Shot classes are even necessary. They are not doing anything at all. The application developer can use PointData directly and be quite happy.

Well, PointSource is actually doing some potentially useful type checking AFAICT. There are two names are language cosmetics

navjotk:
I can't see any type checking. Point me to what you are talking about? 
> The other two are language cosmetics.
Unnecessary (and potentially confusing) language cosmetics as pointed out above since sources and receivers switch roles during a single run.

FabioLuporini:
> I can't see any type checking. Point me to what you are talking about?

Lines 25~32 in ``examples/seismic/source.py``; I meant **value** (not type) checking

OK for the other two

examples/seismic/source.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def __new__(cls, name, coordinates, data=None, ntime=None, **kwargs):
+        p_dim = kwargs.get('dimension', Dimension('p_%s' % name))
+        npoint = coordinates.shape[0]
+        if data is None:
+            if ntime is None:
+                error('Either data or ntime are required to'
+                      'initialise source/receiver objects')
+        else:
+            ntime = ntime or data.shape[0]
```
navjotk: Allowing the shape of the data to be different from the provided number of timesteps might be opening up to a range of potential mistakes by app developers.

navjotk:
>Lines 25~32 in examples/seismic/source.py; I meant value (not type) checking

Lines 25~32 are initialisation code which belongs in `PointData` anyway.

examples/acoustic/fwi_operators.py
mloubout: I understand that you need the `indexed`, but can it be hidden inside the `inject` function instead of having the low-level notation at this place?

mloubout:
>If the PointData class is doing all the heavy lifting of interpolation, I don't think the PointSource, Receiver and Shot classes are even necessary. They are not doing anything at all. The application developer can use PointData directly and be quite happy

I still think the `Source/Receiver` classes are necessary as it will be extended to extra properties (time interpolation for example) that do not belong to `PointData`. The same way we have a `Model` class I think it's important to separate the `devito` object from the application/example class.

FabioLuporini:
I lost track of what needs to be fixed for this PR to land. I seem to understand (and please correct me if I'm wrong) it's mostly just nitpicking. Can we merge?

mloubout:
@mlange05 @FabioLuporini 
I made an update that brings back the indexing inside `inject/interpolate` that I find more readable and less low-level inside the operator. I can revert it if you don't like it.

tests/test_adjointA.py
```diff
@@ -100,7 +100,6 @@ def source(t, f0):
     # Actual adjoint test
     term1 = np.dot(srca.reshape(-1), time_series)
     term2 = linalg.norm(rec) ** 2
-    print(np.finfo(np.float32).eps)
```
mlange05: Uninteded debug statement?

mloubout: yes so removed.

mlange05: Ok, thanks. I'll rebase the branch in a minute.

examples/seismic/source.py
mlange05: I'm not sure I understand this comment. This is merely an initialisation short-hand to allow users to omit `ntime` if they provide a-priori data. I can add an assert in or before the `PointData` initialiser to make things safer though.

mlange05:
> If the PointData class is doing all the heavy lifting of interpolation, I don't think the PointSource, Receiver and Shot classes are even necessary. They are not doing anything at all. The application developer can use PointData directly and be quite happy.

Ok, in my view, the `PointSource`/`Receiver` has two additional jobs to the internalized `PointData`:
* Create a custom dimension based on name, so that source and receiver objects don't share the default point dimension `p`. Since this encodes a naming convention depending on the user-given name, I'd argue that step should not be internalised.
* Perform the copy into `.data` if an initial data set is provided. This helps for creating source objects which come with a pre-existing `time_series` and allows us to reduce the argument clutter by inferring dimension sizes from the data. The reason I would like to keep the `.data` copy external to the core is that the `SymbolicData` objects primarily act as data allocators (similar to `numpy.empty`), with the user responsible for any data copies, since implicit data copies can easily kill application performance in a non-obvious way.

> Lines 25~32 are initialisation code which belongs in PointData anyway.

Sure, we can internalise the argument derivation based on the coordinates array (lines 27-28), but the rest is bound to `data`, so, follwoing the argument above, I'd like to keep this separate. 

> I still think the Source/Receiver classes are necessary as it will be extended to extra properties (time interpolation for example) that do not belong to PointData. The same way we have a Model class I think it's important to separate the devito object from the application/example class.

Agreed. The primary intention here is to replace the outdated `IShot` utility with a Devito-based class that provides the hook-point for further user extensions (file readers, data manipulation, etc). On top of the reasons outlined above, this would provide a nice example for application developers on how to specialise the raw data containers Devito provides.

> Unnecessary (and potentially confusing) language cosmetics as pointed out above since sources and receivers switch roles during a single run.

I disagree here. On top of the functional component (interpolation and injection) these `PointData` objects also store the coordinates of the sparse points and thus define "source geometry" and "acquisition geometry". The key thing to note here is that only the former changes in adjoint runs, not the latter. So, I personally think that explicitly using aliased class names to distinguish the geometric role of the object, while providing the same functionality on both types will help us demonstrate the role reversal very clearly. The follow up PR #243 attempts to make this clearer.

mlange05:
> @mlange05 @FabioLuporini
I made an update that brings back the indexing inside inject/interpolate that I find more readable and less low-level inside the operator. I can revert it if you don't like it.

I would argue that using low-level index adjustment is more general than argument-driven substitution, but I'm happy to go with the substitution route. If you don't mind I'll rebase your commits though and pull the Ellipsis commit into a separate branch - if we're not using it, it's an unused and untested feature after all.

FabioLuporini:
> I would argue that using low-level index adjustment is more general than argument-driven substitution, but I'm happy to go with the substitution route. If you don't mind I'll rebase your commits though and pull the Ellipsis commit into a separate branch - if we're not using it, it's an unused and untested feature after all.

What about keeping both?

mlange05:
We have no use-case and no tests for it, so there is no reason to merge the feature. It's a simple change though, so it can be resurrected if needed.

FabioLuporini:
> We have no use-case and no tests for it, so there is no reason to merge the feature. It's a simple change though, so it can be resurrected if needed.

OK

mlange05:
Ok, rebased on master, and _I believe_ all concerns either answered or addressed. Are we all happy with this (in particular @navjotk)?

examples/seismic/source.py
navjotk: I am not sure what the behaviour would be if `ntime` and the time dimension in the data structure don't match up. Do we through an exception? Is there a test for that?

In general, I think we should move away from using `assert`s and use exceptions instead.

navjotk:
> The reason I would like to keep the .data copy external to the core

I don't think an end-user would differentiate between devito classes supplied in the devito directory and those supplied within the examples directory if they seem to do the same thing. In fact, a pip install would even ignore the examples directory, rendering any sample code that uses wrappers from the examples directory unusable. 

> On top of the functional component (interpolation and injection) these PointData objects also store the coordinates of the sparse points and thus define "source geometry" and "acquisition geometry". 

`PointData` objects have everything necessary to store the coordinates. 

> The key thing to note here is that only the former changes in adjoint runs, not the latter

Not sure what you mean since in the adjoint, the source becomes the receiver and the receiver becomes a source.

mlange05:
> > The reason I would like to keep the .data copy external to the core

> I don't think an end-user would differentiate between devito classes supplied in the devito directory and those supplied within the examples directory if they seem to do the same thing. In fact, a pip install would even ignore the examples directory, rendering any sample code that uses wrappers from the examples directory unusable.

Ok, so:
a) pip also ignores tests, so through this route users only get the bare Devito core, but not the full-blown "all the trimmings, including docs, tests and examples" version. If we want to change that, we should have a separate discussion.

b) The examples are intended to highlight the basic principles of how one might (or might not) work with the core functionalities provided by Devito to solve individual FD problems or even build high-level applications. As such they are very much different from the core functionalities and aim to illustrate the intended use of core functionalities (in this case symbolic objects) of any kind. If this demonstration includes adding small wrappers (like eg. `AcousticForward` for operators, or `PointSource`/`Receiver` for symbolic data) to keep notation clean, then that is still a "user-space" choice rather than a core API decision.

c) This brings us to the actual question behind this: Devito allocates the memory, but who is responsible for filling it with the required data? Ie. who copies the pre-created data into `.data`? I would argue the user should always be responsible, since there are many cases where the data isn't pre-arranged and users want to only handle raw buffers, for example via a seg-y file reader. However, in the seismic examples we habitually use pre-created data to fill source/receiver symbols, so having a utility that performs the data copy to hide notational overheads seems reasonable to me, given that the utility lives in the "user space", ie. `examples`.

mlange05:
> > On top of the functional component (interpolation and injection) these PointData objects also store the coordinates of the sparse points and thus define "source geometry" and "acquisition geometry".

> PointData objects have everything necessary to store the coordinates.

Yes, the question of providing multiple aliases for the seismic-specific wrapper is a purely notational one though and comes down to taste, I guess. The intention here is to mirror the existing naming conventions for symbol and variable names (more below). 

> > The key thing to note here is that only the former changes in adjoint runs, not the latter

> Not sure what you mean since in the adjoint, the source becomes the receiver and the receiver becomes a source.

Yes, but the "source" symbol in the adjoint operator is still named `srca` and _acts_ as a receiver by _interpolating_ into the sparse-point data. Conversely, the "acquisition geometry" is still stored on a variable named `rec` (in both cases forward and adjoint), but the `rec` symbol changes role in that it _interpolates_ for forward runs and _injects_ the provided values into the wavefield in adjoint runs. I also note that the convention of using the "geometric role" (ie. "at source location" and "at receiver location") is still the guiding metric for deciding the symbol and variable names in the examples.

I also note that in explaining the final point, we really should focus on providing more verbose documentation of the seismic utilities, preferably in some notebook/tutorial.

navjotk:
> pip also ignores tests

Tests are written for us - the developers of core devito. Are these wrappers written for devito developers or app developers (users of devito)?

> stored on a variable named rec

That is an implementation decision best left to the user (application developer). Why would we mandate what the app developer chooses to call their variables? They key point is, the same object that behaved as a receiver first, behaves as a source later. Why introduce complications by separating Sources and Receivers when their roles in the problem are not separated? What is the problem you are trying to solve with this?

> I also note that in explaining the final point, we really should focus on providing more verbose documentation of the seismic utilities, preferably in some notebook/tutorial.

Examples are supposed to be (a supplement to) documentation in themselves. If the examples require documentation of their own, maybe we are doing something very wrong.

mlange05:
> > pip also ignores tests
> Tests are written for us - the developers of core devito. Are these wrappers written for devito developers or app developers (users of devito)?

They are primarily written to allow tests and examples to be clear, concise and illustrative. If users select to use them further up, they are free to do so, although, of course, they might not be as rigorously tested as core components.

> That is an implementation decision best left to the user (application developer). Why would we mandate what the app developer chooses to call their variables?

Indeed, any application developer is free to make their own decisions on how to implement that. Yet we also have to make that decision if we are to provide a full illustrative set of examples, and in this case I believe the existing variable naming is actually quite intuitive.

> They key point is, the same object that behaved as a receiver first, behaves as a source later. Why introduce complications by separating Sources and Receivers when their roles in the problem are not separated? What is the problem you are trying to solve with this?

The "problem" I am trying to solve is to clearly demonstrate the dual role of sources and receivers to users of all backgrounds. In this case by extending the existing variable naming convention, which mirrors the geometric role, to the data instantiation part, ie. sources are (in the forward base case) instantiated with pre-computed data (time series), while receivers merely require the number of timesteps to allocate an empty data buffer. In the adjoint case that role changes, so that a "source" is now initialised empty, while "receivers" (the sparse points located at "acquisition geometry" coordinates) comes either pre-populated or with pre-computed data. I believe this closely mirror the previous conventions and intents used in our examples - and I actually have found this convention very helpful myself.

> Examples are supposed to be (a supplement to) documentation in themselves. If the examples require documentation of their own, maybe we are doing something very wrong.

Maybe we are, maybe we are not... I believe to demonstrate the use of Devito in complex problems, such as seismic inversion, to a broad scientific audience of varying backgrounds it is actually imperative to provide more than a pure-code example, hence my repeated call for notebook-style tutorials. If these tutorials then manage complexity with the introduction and subsequent use of domain-specific utilities, that is even better, since it manages cognitive load.

FabioLuporini:
> If the examples require documentation of their own, maybe we are doing something very wrong

I don't think so; eg, well documented notebooks are extremely useful (see http://www.firedrakeproject.org/documentation.html for example)

ggorman:
Guys - this discussion is meandering all over the place. 5 days ago in this discussion @FabioLuporini  asked if it was ready for merging and if we were now mostly just nitpicking - the answer is yes, we are nitpicking. If there are broader issues about the state of documentation, tests, examples etc then lets have a separate issue and PR for that specific issue.

I don't want to start nitpicking myself over specific issues mentioned above so I'm going to limit myself to just one. Source and receivers are two different things - just because they can switch roles doesn't mean that they are the same - if you disagree you are going to find all your swivel chairs replaced by three legged stools tomorrow... The distinction becomes clear when you consider data handling (i.e. not synthetic data you create on the fly for testing).

Now, enough with the nitpicking.

FabioLuporini:
I think @ggorman made it pretty clear. Travis is green, so @mlange05 whenever you're happy, please merge.

Issue #26, opened 2016-06-07
Centralised Configuration (Was Configuration File)


navjotk:
Over time, devito is using quite a few environment variables (4, if I am not mistaken). Since devito is being made as a library, it doesn't make sense to accept command line flags since the actual executable will not always be in our control (since the "end-user" of devito would be writing the executables). The remaining options to take in configuration parameters (like the choice of compiler, compiler flags, whether to enable debug/profiling mode or not) are Environment variables and a configuration file. Environment variables can become difficult to track when the number increases (and at 4, it is already complex). They also make debugging issues harder - setting up the project on a fresh machine will also be much quicker with a default configuration file. 

Maybe we should move to reading these parameters from a config file?

navjotk:
4adbae7f68680c84519a9b85c23761159b6ad459 offers an implementation of this. Requesting review and comments.

navjotk:
I will make here a list of parameters that I see in the code that could potentially come from a config file:
- Compiler
- Flags
- OpenMP on/off [Can potentially be the BACKEND var]
- Profiling on/off
- Cache blocking on/off
- Block size
- Auto tuning on/off
- Path to isat installation
- Path prefix for generated isat files
- Dependent or Independent search for autotuning
- Auto tune search range
- Path prefix for memmap files
- Dtype (32 bit/64 bit) [Debatable]
- Enable MIC mode or not [Can potentially be the BACKEND var]
- The path for JIT files (in case the user does not want them to go to temp)

navjotk:
This is now needed even for first touch to work.

mlange05:
OK, to clarify, the need for _a priori_ knowledge arises from the use of a first touch loop being added to memory allocation (see PR #135), which requires us to emulate the same looping pattern if we are using explicit loop blocking.

As a starting point for this I would propose that the internal engine expects some type of `parameters` dictionary, which holds defaults and can in addition be populated via either well-documented environment variables or a fixed config file. We should also be very explicit about which features/flags we need to expose through this mechanism and why, in order to avoid the previous situation where an unnecessarily large number of flags was required to even get baseline performance.

navjotk:
These are the required parameters as I see it:
- The `compiler` object
- The `cache_blocking` flag
- Block sizes

mlange05:
Hmmm, but how does AT work then - assuming that will be the preferred MO of most users, rather than explicit `cache_blocking`? Do we enable AT via config and then let it write block sizes into the global parameters dictionary?

[...]

PR #241, opened 2017-05-10
Global configuration

navjotk:
This PR:
- Introduces a global configuration dictionary named `parameters` since we like to take inspiration from Firedrake. 
- Changes the logger to pick up its log-level from this configuration rather than setting the default in the logger itself. This is provided as an example of the intended use of this dictionary plus @mloubout seemed to want something like this anyway. 
- The `Forward`, `Adjoint` etc. methods of `Acoustic` had a superfluous parameter `data` which is also removed in this PR. 

Nearly a year after this issue was raised in #26 

In the next iteration, the config can be extended to read from a file/environment variables but I believe this should do for the current purposes.

### DIFF
--- a/devito/logger.py
+++ b/devito/logger.py
@@ -3,6 +3,7 @@
 import logging
 import sys
 from contextlib import contextmanager
+from devito.parameters import parameters
 
 __all__ = ('set_log_level', 'set_log_noperf', 'log',
            'DEBUG', 'INFO', 'AUTOTUNER', 'DSE', 'DSE_WARN', 'DLE', 'DLE_WARN',
@@ -34,7 +35,8 @@ logging.addLevelName(DSE_WARN, "DSE_WARN")
 logging.addLevelName(DSE, "DLE")
 logging.addLevelName(DSE_WARN, "DLE_WARN")
 
-logger.setLevel(INFO)
+init_log_level = eval(parameters["log_level"])
+logger.setLevel(init_log_level)
 
 NOCOLOR = '%s'
 RED = '\033[1;37;31m%s\033[0m'
--- a/devito/logger.py
+++ b/devito/logger.py
@@ -17,6 +17,14 @@ logger = logging.getLogger('Devito')
 _ch = logging.StreamHandler()
 logger.addHandler(_ch)
 
+log_level = "log_level"
+
+
+def parameters_updated(key, value):
+    if key == log_level:
+        logger.setLevel(eval(value))
+
+
 # Add extra levels between INFO (value=20) and WARNING (value=30)
 DEBUG = logging.DEBUG
 AUTOTUNER = 19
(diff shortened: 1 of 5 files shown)


FabioLuporini:
All good for me

mloubout:
I guess I'm a bit slow, but can we have a small example on how to use the `parameter` dictionary?

All good to me.

devito/logger.py
```diff
@@ -34,7 +35,8 @@
 logging.addLevelName(DSE, "DLE")
 logging.addLevelName(DSE_WARN, "DLE_WARN")
 
-logger.setLevel(INFO)
+init_log_level = eval(parameters["log_level"])
```
mlange05: Why do we need `eval` here? Shouldn't the `parameters` dictionary be initiated and accessible before we even get here? Or am I missing something here?

navjotk: You usually use the following to set the level of the log: `setLevel(INFO)`. `INFO` here is the *name* of the constant, not the value. If we are doing `setLevel(level)`, the variable `level` would need to contain the *value* of the constant `INFO` to achieve the same effect. This would require that a user of devito be expected to open the source of python's in-built `logging` module (since that is where we inherit the log-level constants from) to see what value the constant was set to and use that in our parameters file. The constant value can potentially change between python versions (which is the reason for using constants in the first place), and hence might produce unexpected results sometimes. Using `eval` we can use the constant names which are guaranteed to be consistent across versions and yet accept the log-level as a string (which is how it would come to us if the log-level was set in a config file or an environment variable).

devito/parameters.py
```diff
@@ -0,0 +1,47 @@
...
+
+        if isinstance(value, Parameters):
+            child_update = lambda x: self._updated(*x)
+            value.update_functions.push(child_update)
+
+        # Tell everyone we've been updated
+        self._updated(key, value)
+
+    def _updated(self, key, value):
+        """ Call any provided update functions so everyone knows we've been updated
```
navjotk: The logger would need to register a watcher. I can add that as an example as well.

navjotk:
> Let's hope the "next iteration" doesn't take another year. 

The first implementation of this was actually done a year ago but was ignored (see linked issue). Maybe if this one isn't ignored, the next one won't take as long ;-)

devito/parameters.py
```diff
@@ -0,0 +1,47 @@
+"""The parameters dictionary contains global parameter settings."""
+
+__all__ = ['Parameters', 'parameters']
```
navjotk: done

devito/parameters.py
navjotk: done

FabioLuporini:
Ready to merge?

mlange05:
Yes, looks good and works now. Merging...

Issue #216, opened 2017-04-05
Interface to Revolve Checkpointer


jhueckelheim:
**Update: The interface has changed, see my comment below.**

The Revolve Checkpointing library needs an interface to work with Devito. All the points below are my current working assumption and up for discussion. If you don't know what this is about, read the following block of text, else goto Devito.

> # Recap: Checkpointing
>
> The adjoint computation of an unsteady nonlinear primal function requires the full primal trajectory in reverse temporal order. Storing this can exceed the available memory. In that case, Checkpointing can be used to store the state only at carefully selected points in time. From there, the forward computation can be restarted to recompute lost sections of the trajectory when they are needed during the adjoint computation. This is always a tradeoff between memory and runtime. The classic and provably optimal way to do this for a known number of time steps is [Revolve](http://dl.acm.org/citation.cfm?id=347837.347846). For a previously unknown number of time steps, there's an [optimal online version](http://dl.acm.org/citation.cfm?id=1654955). For a two-stage storage system (e.g. memory+HDD) there's also an optimal [multistage algorithm](https://hal.inria.fr/hal-01147155/document).
> 
> We are working on a python checkpointing library that wraps around an open source C++ checkpointing library. That library only provides a schedule, i.e. it decides when to store checkpoints. On top of that, we are creating a library (the "Checkpointer", "Revolver", naming suggestions welcome) that performs the actual data and computation management. This library needs to split the forward computation in a number of chunks, store a copy of state variables at the end of each chunk, interleave multiple calls to forward and adjoint operators, and inject data back into the forward operator to resume the computation from a previously stored state. Using this in Devito requires some communication between the Checkpointer and Devito operators.

# Devito

 - users need to pass `forward_operator` and `reverse_operator` to the Checkpointer.
 - These objects need to have
    - an `input_arguments()` routine that returns a list of references to Symbol objects that are read
      by this operator (this may not change throughout the computation),
    - equally an `output_arguments()` list of Symbol objects that are modified,
    - an `apply(time_steps, **kwargs)` routine that performs the prescribed number of time steps. `apply` must accept whatever data is within the symbols as the intial state for that computation, and the forward operator may not have side-effects and only access the symbols given in `kwargs`. This can probably be relaxed for the reverse operator if needed.
 - The Symbol objects need
    - an `nbytes` property that reports the size of this object in memory, and
    - a `data` property with a `deep_copy` routine.
    

# Checkpointer

[...]

PR #242, opened 2017-05-10
Gradient test based example without wrappers

navjotk:
This PR adds an example in the `examples` folder based on `test_gradient`, however it reproduces this test without any wrappers on the operators. This way the operators are available directly in the example code to play around with, for feature development.

This is especially helpful for #216 since this can be the basis for the example that is required to build the interface to the revolve checkpointer. 

This can serve as an example of how something complex like the gradient test can be implemented with pure Operators with (relative) ease. 

This PR also adds a utility function to the `CMemory` class (the one that wraps devito allocated memory) to do a quick `memset`. This is required to be able to reuse the same block of memory for multiple passes.

This PR includes the changes from #241 so any review should keep that in mind.

### DIFF
--- a/examples/acoustic/gradient_example.py
+++ b/examples/acoustic/gradient_example.py
@@ -1,0 +1,178 @@
+import numpy as np
+from examples.seismic import Model
+from devito import Dimension, time, TimeData, DenseData
+from examples.source_type import SourceLike
+from examples.acoustic.fwi_operators import ForwardOperator, GradientOperator
+from numpy import linalg
+
+
+def smooth10(vel, dimensions):
+    out = np.zeros(dimensions)
+    out[:] = vel[:]
+    for a in range(5, dimensions[-1]-6):
+        if len(dimensions) == 2:
+            out[:, a] = np.sum(vel[:, a - 5:a + 5], axis=1) / 10
+        else:
+            out[:, :, a] = np.sum(vel[:, :, a - 5:a + 5], axis=2) / 10
+    return out
+
+
+def source(t, f0):
+    r = (np.pi * f0 * (t - 1./f0))
+    return (1-2.*r**2)*np.exp(-r**2)
+
+
+def run(dimensions=(50, 50, 50), spacing=(15.0, 15.0, 15.0), tn=750.0,
+        time_order=2, space_order=4, nbpml=40, dse='noop', dle='noop',
+        autotuning=False, compiler=None, cache_blocking=None):
+    ndim = len(dimensions)
+    origin = tuple([0.] * ndim)
+    f0 = .010
+
+    t0 = 0.0
+    nsrc = 1
+    nrec = 101
+    # True velocity
+    true_vp = np.ones(dimensions) + .5
+    true_vp[:, :, int(dimensions[2] / 2):] = 2.
+
+    # Smooth velocity - we use this as our initial m
+    initial_vp = smooth10(true_vp, dimensions)
+    m0 = initial_vp**-2
+    # Model perturbation
+    dm = true_vp**-2 - initial_vp**-2
+    model = Model(origin, spacing, true_vp.shape, true_vp, nbpml=nbpml)
+    dt = model.critical_dt
+    if time_order == 4:
+        dt *= 1.73
+    dtype = model.dtype
+    h = model.get_spacing()
+    nt = int(1+(tn-t0)/dt)
+
+    # Source geometry
+    time_series = np.zeros((nt, nsrc))
+    time_series[:, 0] = source(np.linspace(t0, tn, nt), f0)
+
+    location = np.zeros((nsrc, 3))
+    location[0, 0] = origin[0] + dimensions[0] * spacing[0] * 0.5
+    location[0, 1] = origin[1] + dimensions[1] * spacing[1] * 0.5
+    location[0, 2] = origin[1] + 2 * spacing[2]
+
+    # Receiver geometry
+    receiver_coords = np.zeros((nrec, 3))
+    receiver_coords[:, 0] = np.linspace(0, origin[0] + dimensions[0] * spacing[0],
+                                        num=101)
+    receiver_coords[:, 1] = origin[1] + dimensions[1] * spacing[1] * 0.5
+    receiver_coords[:, 2] = location[0, 2]
+
+    # Create source symbol
+    p_src = Dimension('p_src', size=nsrc)
+    src = SourceLike(name="src", dimensions=[time, p_src], npoint=nsrc, nt=nt,
+                     dt=dt, h=h, ndim=ndim, nbpml=nbpml, dtype=dtype,
+                     coordinates=location)
+    src.data[:] = time_series
+
+    # Create receiver symbol
+    p_rec = Dimension('p_rec', size=nrec)
+
+    # Receiver for true model
+    rec_t = SourceLike(name="rec_t", dimensions=[time, p_rec], npoint=nrec, nt=nt, dt=dt,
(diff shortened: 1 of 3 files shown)


examples/acoustic/gradient_example.py
```diff
@@ -0,0 +1,178 @@
...
+                        dtype=model.dtype)
+    u_nosave.pad_time = False
+
+    # Forward wavefield where all timesteps are stored
+    # With checkpointing this should go away <----
+    u_save = TimeData(name="u_s", shape=model.shape_domain, time_dim=nt,
+                      time_order=time_order, space_order=space_order, save=True,
+                      dtype=model.dtype)
+    u_save.pad_time = True
+
```
mloubout: As discussed with @mlange05 and in the meeting, I am not a big fan of all this initializations in the main script. Specially having objects like `Dimensions` that should not be seen by the end user.

FabioLuporini: Remind me what that ``pad_time`` does?

FabioLuporini:
Looks good to me

examples/acoustic/gradient_example.py
mloubout: Outdated reminder of the previous Operator, should be removed.

navjotk:
- `Dimension` objects are meant to be seen by the application developer, who is the target audience for this example. They are part of devito's front-end API. This serves as an example of how to use them.

- If someone who is generally aware of how pde-constrained inversion works but not in the seismic context wants to write a solver using devito but for another pde, the examples using the seismic wrappers wouldn't be very helpful to them since they don't share the same vocabulary. This example tells them how to do the most essential things in inversion without any wrappers, which is what someone needs to see when they want to use devito for another domain for which wrappers don't exist.

examples/acoustic/gradient_example.py
navjotk: Done

FabioLuporini:
Ready to land?

mlange05:
> If someone who is generally aware of how pde-constrained inversion works but not in the seismic context wants to write a solver using devito but for another pde, the examples using the seismic wrappers wouldn't be very helpful to them since they don't share the same vocabulary. This example tells them how to do the most essential things in inversion without any wrappers, which is what someone needs to see when they want to use devito for another domain for which wrappers don't exist.

Ok, I'm sorry to say this, but I agree with @mloubout that this increased complexity is unnecessary and I actually think that this merge does not look particularly helpful (at least to me, and I might be wrong here). My rationale is that the gradient test is possibly the most complicated one we currently have (at least from an implementation point of view), and we are using wrappers in the original test to hide some of this complexity for a reason! If the intention is to highlight the more low-level concepts of the current Devito API (dimensions, point sources, etc.) to new user not familiar with seismic inversion, this can be done with a much simpler set of examples (eg. `adjointA.py`, maybe even `adjointJ.py`, or other example) without throwing the full complexity of the gradient test at them - it might also be easier accessible if it were done in a notebook or website tutorial with full-sentence explanations and type-set equations). And without the educational argument I see very little reason to have to maintain two equivalent versions of an extremely complicated test, where one is quite challenging to read. Possibly up for discussion in the next dev call though?

I would also like to point out that we are currently preparing for a major release, so the objective should be to reduce existing complexity and enhance documentation material, not replicate and add complexity. But again, up for discussion.

examples/acoustic/gradient_example.py
```diff
@@ -0,0 +1,176 @@
...
+    # Gradient symbol
+    grad = DenseData(name="grad", shape=model.shape_domain, dtype=model.dtype)
+    # Reusing u_nosave from above as the adjoint wavefield since it is a temp var anyway
+    gradop = GradientOperator(model, u_nosave, grad, rec_g, u_save, time_order=time_order,
+                              spc_order=space_order, cache_blocking=cache_blocking,
+                              dse=dse, dle=dle, compiler=compiler, profile=True)
+
+    # Clear the wavefield variable to reuse it
+    # This time it represents the adjoint field
+    u_nosave.data.fill(0)
```
mlange05: I take it `.data.fill` is a custom Devito function? Why not stick to native Python instead `u_nosave.data[:] = 0.`?

examples/acoustic/gradient_example.py
```diff
@@ -0,0 +1,176 @@
...
+    src = SourceLike(name="src", dimensions=[time, p_src], npoint=nsrc, nt=nt,
+                     dt=dt, h=h, ndim=ndim, nbpml=nbpml, dtype=dtype,
+                     coordinates=location)
+    src.data[:] = time_series
+
+    # Create receiver symbol
+    p_rec = Dimension('p_rec', size=nrec)
+
+    # Receiver for true model
+    rec_t = SourceLike(name="rec_t", dimensions=[time, p_rec], npoint=nrec, nt=nt, dt=dt,
```
mlange05: If the objective is to not use seismic-specific utilities, should we not rebase this on #239 and use internal `PointData` instead?

examples/acoustic/gradient_example.py
navjotk: It's a wrapper for [numpy.ndarray.fill](https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.fill.html) and is [faster](http://stackoverflow.com/a/13052254)

navjotk:
> this increased complexity

No complexity has increased. This is literally the simplest way for someone to write a similar test for a different equation. Can you write this test in a simpler way using the **Devito** API (wrappers are not the Devito API)? Show me. The wrappers are only for seismic. Do we not want people to try other equations? Are you saying this test is not a useful demo? This test is essential if you want to carry out an inversion problem. If you want complexity to decrease, maybe look at the API and not the example that demonstrates its use. 

> does not look particularly helpful (at least to me

That's because you don't understand the gradient test

> this can be done with a much simpler set of examples (eg. adjointA.py, maybe even adjointJ.py, or other example)

Maybe add some more examples to build up to this. That is no reason to throw away a perfectly good example. The difference between adjointJ.py and this is literally 5 lines of code. Are you arguing over 5 lines of code?

>done in a notebook or website tutorial with full-sentence explanations and type-set equations

Doesn't have to be one or another. You are free to do that if you think it is necessary. 

> enhance documentation material

In a situation of a complete lack of documentation, examples like this are the only documentation we have. If you would like to see some documentation, maybe write some?

Overall I think this is just nitpicking since this PR doesn't touch any devito code. Adding an example (where examples are seriously lacking) is a positive step. At no point have you given a single reason to NOT have this. 

You are free to junk this PR as well, if that makes you happy :-)

FabioLuporini:
I think it's good to have a complete set of examples of increasing complexity. 

One observation though:

> The difference between adjointJ.py and this is literally 5 lines of code

Can we then avoid duplicate code (if we are, I haven't checked)?

FabioLuporini:
@navjotk can you take a look at my question above?

navjotk:
The "duplicated code" that you talk about is setup code. Setting the grid sizes, putting the sources and receivers. If you see the example that this PR proposes, no execution has happened until line 109. We are only setting up the environment in which the test will be conducted until then. There is no reason to hide that behind wrappers except maybe if you want to demonstrate this in a notebook where you could just import a pre-prepared test environment. For anything remotely realistic you would want to set this up manually since this is the stuff that changes from one situation to another. Hence it is actually educational to leave multiple versions of the setup code lying around in the examples so that people see how different things can be achieved. 

TLDR: What you see as duplicate code is the numerical analysis equivalent of variable definitions (`float a;` etc). You want to define the variables every time you use them. It would be counter-productive to write a wrapper that defines variables for you.

FabioLuporini:
Right, all clear

Issue #186, opened 2017-02-06
Caching of generated code


navjotk:
On our current FWI runs, most of the time is spent by the workers compiling code, running the DSE transformations etc. This is obviously needless repetition that we should be avoiding by caching these results. Although I call it caching of generated code in the issue title, I mean to suggest caching at a higher level than just the code since even the DSE need not be called. Feel free to suggest a different title for the issue.

mlange05:
Devito-3.0 (a.k.a. `StencilKernel`) already caches based on a detailed representation of the generated IR tree (see PR #180) . Simply switching the First Touch loops to this should help a lot. For anything higher level than that generating rigorous cache keys becomes tricky, since you need to fix the exact input plus the exact option combination.

mlange05:
OK amendment to the above: 3.0 `StencilKernel` generates the cache key according to the final IR state. We currently still force-generate and compile every time, but it's easy to switch once we're satisfied the cache keys behave as we'd expect.

mlange05:
Picking up this issue again, I think we need to un-muddle what we mean by _caching_ here:
* `StencilKernel` operators actually do cache generated code on an object basis. This means that once an operator is created it can be re-used through multiple `.apply()` calls without incurring any DSE/DLE/codegen/compilation overheads.
* The current seismic examples negate this fact by creating new operators for each invocation, thus forcing a complete DSE/DLE every time.
* Creating a generic caching mechanism that reliably pre-empts DSE (the first stage of creating an operator) requires us to encapsulate some measure of the input state and all potential transformations, given a particular parameter combination. *This seems incredibly hard to do well!*

Given that, I would argue that we should re-write the wrappers in `fwi_operators.py`/`tti_operators.py` to, at the very least, utilise a pre-setup operator across different invocations. In an ideal world, the 3.0 API should be powerful enough to completely remove these wrappers.

With that in mind I think this issue should be renamed to something like "Operator-caching in seismic examples".

mlange05:
@navjotk It might also be useful to have a pseudo-code outline of the type of FWI/RTM loop you had in mind, so that we can discuss when and where different symbolic objects and operators need to be created to avoid re-compilation/re-allocation.

navjotk:
The kind of caching you talk about already existed. The code was generated at operator creation time and multiple apply calls would call the same code. AFAIK this behaviour has remained the same since 2.0. 

This issue was about doing the kind of caching that you call *incredibly hard*.

[...]

PR #243, opened 2017-05-11
Acoustic utility cleanup

mlange05:
Ok, bit of a large merge with lots to discuss, I'm sure. The core idea is to clean up the acoustic wave examples and tests and in particular to clear up the roles of the various utilities involved in building complex test, such as the gradient and adjoint tests. A particular rationale for the various changes follows, but please keep in mind that names and docstrings are all up for discussion, so comments are welcome.

* The previously created `PointSource`/`Receiver` wrappers around `PointData` are now used to replace `IShot()`, since they effectively encapsulate all the data necessary to completely define the source and acquisition geometries, but are based on native Devito classes.
* Rename the `Acoustic_cg` class -> `AcousticWaveSolver` and reduce some of the parameter clutter. The core idea is that the "solver" now encapsulates the two primary solver settings: spatial and temporal discretisation order and provides operator objects created from a general description of the seismic scenario from a model, source and acquisition geometries.
* The solver is also used to cache the different types of operator, so that we can easily re-use them within loops and on different pieces of data (eg. different models). The operators are provided as cached properties so that one may re-write any test/example with raw operators instead of the slightly higher-level wrappers provided by the solver by using `solver.op_fwd(u=..., src=..., rec=..., m=...)` instead of `rec, _, _ = solver.forward(save=False)`.
* The solver-level wrappers ensure the correct data initialisation, depending on the calling context. This allows user to optionally insert or retrieve explicit symbol-data objects, such as the wavefield in a forward run, or let the wrapper allocate and delete them on-the-fly. For this to work properly it is important to note which fields are read-only (eg. source in forward), write-only (eg. rec in forward, which can e re-used if necessary) and increment (eg. the wavefield in forward runs, which needs resetting before re-use).
* The pure operator definitions now also act almost as pure "template operators", in that they create the appropriate data types they need to create the `Operator`, but do not actually touch any data on them. This entails that the constructor/factory function does not incur any hidden memory allocations and simply needs the description of model and source/acquisition geometry to create the right `shape`s for `SymboliData` objects. _We might want to re-think this current hard-requirement to provide a `shape` in the future._
* Operator constructors and solver-level wrappers now pass compiler options, such as `dse`, `dle`, etc) straight to the relevant low-level `Operator` class via `**kwargs`.
* Small tweaks and fixes to the `Operator.arguments` function that hopefully fix the keyword-based overrides and native numpy array injection.
* Loads of various renamings and documentation attempts to improve the general readability and adhere to some common Python practices. Comments and suggestions for even better documentation are much appreciated.

This PR depends on #239. It is is primarily intended to clarify the roles of operators, symbols and data in the new 3.0 paradigm and is aimed at solving issue #227. There is obviously more that one could do to improve the clarity of the example code, but this represents (I hope) the most crucial structural concerns.


navjotk:
>Small tweaks and fixes to the Operator.arguments function that hopefully fix the keyword-based overrides and native numpy array injection.

What is broken that this is trying to fix?

>...simply needs the description of model and source/acquisition geometry to create the right shapes for SymboliData objects...

Source/acquisition geometry changes between iterations so if the Operator depends on these, it would have to be reconstructed between iterations.

Another thing, the parameter `dt` is provided at Operator creation time and finds its way all the way down to the generated code. This parameter is highly likely to change between iterations, again requiring that the Operator be reconstructed.

navjotk:
Ping

devito/operator.py
```diff
@@ -190,14 +190,23 @@ def arguments(self, *args, **kwargs):
         # Insert loop size arguments from dimension values
         d_args = [d for d in arguments.values() if isinstance(d, Dimension)]
         for d in d_args:
-            arguments[d.name] = dim_sizes[d]
+            if d not in dim_sizes and d.is_Buffered:
+                # Only the parent is defined, so use it
+                arguments[d.name] = dim_sizes[d.parent]
+            else:
+                arguments[d.name] = dim_sizes[d]
```
FabioLuporini: `arguments[d.name] = dim_sizes.get(d, dim_sizes[d.parent])` ?

examples/acoustic/wavesolver.py
```diff
@@ -0,0 +1,224 @@
...
+        # Pick m from model unless explicitly provided
+        if m is None:
+            m = m or self.model.m
+
+        # Execute operator and return wavefield and receiver data
+        if save:
+            summary = self.op_fwd_save.apply(src=src, rec=rec, u=u, m=m, **kwargs)
+        else:
+            summary = self.op_fwd.apply(src=src, rec=rec, u=u, m=m, **kwargs)
+        return rec, u, summary
```
FabioLuporini: Let's see if I get this right: it'd make no sense to subclass``AcousticWaveSolver`` with a class that implicitly has ``save = True`` because in general we'll need to apply Forward in both ways, for the same solver, right?

examples/tti/tti_example.py
```diff
@@ -61,25 +61,24 @@ def setup(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
...
+    return TTI_cg(model, data, src, t_order=time_order, s_order=space_order)
 
 
 def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
         time_order=2, space_order=4, nbpml=10, dse='advanced', dle='advanced',
-        auto_tuning=False, compiler=None, cache_blocking=None):
-    if auto_tuning:
+        autotune=False, compiler=None, cache_blocking=None):
+    if autotune:
         cache_blocking = None
```
FabioLuporini: This looks like leftover from ``legacy``

FabioLuporini: Probably we can drop ``cache_blocking`` everywhere

FabioLuporini:
I think overall this is great as, at least to my eyes, it makes everything more readable and easier to understand.

mloubout:
Overall good to me.

examples/acoustic/wavesolver.py
mlange05: It's entirely context-dependent, but yes, there are cases in our test base where we run both modes with the same solver setup. The same, I believe, is true for checkpointing, where we interchange between both modes frequently.

examples/tti/tti_example.py
mlange05: It probably is. I tried not to touch TTI any more than I have to in this PR, but I'm happy to clean this up a little further during the rebase.

mlange05:
> > Small tweaks and fixes to the Operator.arguments function that hopefully fix the keyword-based overrides and native numpy array injection.

> What is broken that this is trying to fix?

1) If numpy arrays are provided instead of `TimeData` objects the size of `t` is derived wrongly, so we pull it from the parent. 
2) `f_args` and `args` can go out of sync if we provide numpy arrays, so we only iterate over `f_args` and pull the name from there. It also adds a small sanity check at the end of the argument derivation.

Arguably, the entire argument derivation routine needs a more complete set of tests and a possible re-write/clean-up - but that is left for another PR.

> Source/acquisition geometry changes between iterations so if the Operator depends on these, it would have to be reconstructed between iterations.

I'm not entirely sure what you mean here. If you are referring to multiple RTM iterations with changing source/acquisition geometry, then supplying new `src`/`rec`/`srca` objects at runtime through either the solver wrappers or on plain operators is both possible: `solver.forward(rec=new_rec, src=new_src` or `solver.op_fwd(src=new_src, rec=new_rec, u=my_u, m=my_model.m)`. 

I note that we currently don't test that case though, so, if you have an example of this, it would be great to have either a test or an example for this (or both). @navjotk 

> Another thing, the parameter dt is provided at Operator creation time and finds its way all the way down to the generated code. This parameter is highly likely to change between iterations, again requiring that the Operator be reconstructed.

Indeed, we should be able to supply individual symbols at runtime, just like we allow deferred dimension sizes. This is quite easy to do with the new API (in theory), but I don't think it's implemented yet. Would you mind raising an issue for this?

ggorman:
Can we pick up @navjotk's comments from 6 days ago?
> Source/acquisition geometry changes between iterations so if the Operator depends on these, it would have to be reconstructed between iterations.

@navjotk is correct that there are cases where geometry varies throughout time. How is that handled here? 

> Another thing, the parameter dt is provided at Operator creation time and finds its way all the way down to the generated code. This parameter is highly likely to change between iterations, again requiring that the Operator be reconstructed.

It feels like a premature optimisation if it is being baked in - particularly for src/rec's. However, without doing benchmarking on real problems it's just pure speculation to argue. I suggest that we stick with whatever is simpler and revisit when we are doing performance analysis on realistic loads.

mlange05:
> > Source/acquisition geometry changes between iterations so if the Operator depends on these, it would have to be reconstructed between iterations.

> @navjotk is correct that there are cases where geometry varies throughout time. How is that handled here?

Two ways are possible, and I guess it depends on context: are the sources/receivers moving or are we loading a completely new set from file (or any other location). In the former case it would probably make more sense to update the `.coordinates.data` array under an existing symbol-object (could be done as part of a kernel), whereas a new set would probably just create a new `src`/`rec` symbol object and then use runtime matching like described above. In any case, the user can chose to have full control by providing it's own object (which he is free to re-create or update) or to rely on the wrappers to do the obvious default thing: use the initially provided source object, it's read-only for forward, and create a new set of receivers at the known "acquisition" coordinates and return it (with equivalents for all other operators, of course). And if that is not flexible enough you can fallback to using the `Operator` object directly, again as explained above via `solver.op_fwd(src=my_src, rec=my_rec, u=my_u, m=my_model.m)`.

mlange05:
> It feels like a premature optimisation if it is being baked in - particularly for src/rec's. However, without doing benchmarking on real problems it's just pure speculation to argue. I suggest that we stick with whatever is simpler and revisit when we are doing performance analysis on realistic loads.

`dt` has always been baked in all previous versions dating back to the inception of Devito - although 3.0 should make it easy to leave it as a runtime parameter and require the user to supply a value when invoking the kernel. Practically speaking we simply have no test-case yet, and therefore haven't actually implemented delayed value-matching for plain parameter symbols. Again, please file an issue.

The performance benchmarking argument is also true for things like loop bounds - but luckily 3.0 provides a backdoor to "bake them in" on-demand by setting the `.size` value on a `Dimension`. Performance benchmarks are forthcoming, but I of course also encourage volunteers. :wink:

navjotk:
> when we are doing performance analysis on realistic loads.

Issue #186 was raised because repeated JIT was dominating the runtime in full FWI runs that I tried with the notebook. This PR supposedly solves it by eliminating JIT between shots/FWI iterations. I claim that this doesn't solve the issue since there are still parameters baked in which change during an FWI run.

> if you have an example of this

It is the FWI notebook on the branch called FWInotebook. 

> multiple RTM iterations

RTM doesn't have iterations. FWI has multiple iterations.

ggorman:
@navjotk - I agree that #186 needs to be addressed, but this PR didn't claim to fix #186 as far as I can see.

Having said that, what I observed previously was that N workers would all compile the same code (identical dt and all). This got particularly bad because for any iterative method the operators are called many times (for the same dt) and the recompilation overhead was ridiculous. So, if caching was fixed for even constant dt the number of compiles would be reduced by a factor of (number of workers)*(number of calls to operators per iteration) - a large step in the right direction.

I suggest that a separate branch&PR to deal with #186 explicitly.

mlange05:
Ok, I believe the separation here should be between process-local and process-global caching - and both mechanisms require us to not "bake in" variable parameters. However, process-local caching can then easily be achieved by not throwing the operator object away, which is what this PR attempts to do while keeping the notational niceness of the "solver" wrappers. If you want process-global caching so that multiple MPI ranks pick up pre-compiled code operator code from disk (or some other shared location), that is another story and I believe issue #186 is the place to discuss that requirement further, since it's non-trivial.

devito/operator.py
mlange05: Alas, no. `d not in dim_sizes and not d.is_Buffered` can also be true and has to do nothing.

FabioLuporini: if ``d`` were not in ``dim_sizes``, then it must be ``d.is_Buffered``, because otherwise ``dim_sizes[d]`` (the ``else`` branch) would fail, right? So perhaps what I wrote is wrong, but there's something weird going on in the ``if`` condition?

mlange05: Ahhh ok, I think I understand now. Since it's a function call now, `d.parent` always gets evaluated, but that obviously fails in many cases. Even better though, removing the change entirely seems to have fixed the problem now (post-rebase), most likely due to the other bug-fix in a follow-up commit. I'll drop this for now and see if travis agrees.

mlange05:
Ok, rebased and all comments addressed, I believe. I also added a warning to run TTI with space orders that are multiples of 4 to avoid errors when running the examples through `benchmark.py`. Can I get a quick ack to indicate that everyone is happy? @navjotk @FabioLuporini

FabioLuporini:
ACK

navjotk:
> `f_args` and `args` can go out of sync if we provide numpy arrays, so we only iterate over f_args and pull the name from there. It also adds a small sanity check at the end of the argument derivation.

This breaks when the overriding symbol does not have the same name as the overriden symbol. It's the obvious case if you want to avoid aliasing. Two symbols will never have the same name.

PR #249, opened 2017-05-19
Foreign

mloubout:
Initialization fix.

`_profile_sections` return a list the same way `OperatorCore` does. 

URGENT PR, needs quick merge as current master breaks Julia.

### DIFF
--- a/devito/operator.py
+++ b/devito/operator.py
@@ -454,6 +454,9 @@ class OperatorForeign(OperatorBasic):
     A special :class:`OperatorBasic` for use outside of Python.
     """
 
+    def __init__(self, stencils, **kwargs):
+        super(OperatorForeign, self).__init__(stencils, **kwargs)
+
     def arguments(self, *args, **kwargs):
         arguments, _ = super(OperatorForeign, self).arguments(*args, **kwargs)
         return arguments.items()
--- a/devito/operator.py
+++ b/devito/operator.py
@@ -453,8 +453,6 @@ class OperatorForeign(OperatorBasic):
     """
     A special :class:`OperatorBasic` for use outside of Python.
     """
-    def __init__(self, stencils, **kwargs):
-        super(OperatorForeign, self).__init__(stencils, **kwargs)
 
     def arguments(self, *args, **kwargs):
         arguments, _ = super(OperatorForeign, self).arguments(*args, **kwargs)
@@ -467,9 +465,9 @@ class OperatorCore(OperatorBasic):
     C code evaluating stencil expressions, can also execute the computation.
     """
 
-    def __init__(self, stencils, **kwargs):
+    def __init__(self, expressions, **kwargs):
         self.profiler = Profiler(self.compiler.openmp)
-        super(OperatorCore, self).__init__(stencils, **kwargs)
+        super(OperatorCore, self).__init__(expressions, **kwargs)
 
     def __call__(self, *args, **kwargs):
         self.apply(*args, **kwargs)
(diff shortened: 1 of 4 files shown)


mloubout:
Can we merge this?

devito/operator.py
```diff
@@ -453,6 +453,8 @@ class OperatorForeign(OperatorBasic):
     """
     A special :class:`OperatorBasic` for use outside of Python.
     """
+    def __init__(self, stencils, **kwargs):
```
FabioLuporini: Why is this needed?

FabioLuporini:
Can u explain the introduction of the constructor? Did u need anything in particular?

We'll merge this on Monday after the above is sorted out

More importantly, we need a way to ensure that nothing breaks in Julia when we merge stuff here in Devito. Perhaps tests on your side?

mloubout:
For the constructor, for some reason, without it there is a bug saying that `init` requires 2 input and only got one.

For the Julia side, I'm looking into proper testing so that stuff don't break.

FabioLuporini:
So, I'm OK with the ``_profile_sections`` fix.

However, I cannot understand why you added the constructor since AFAICT it's exactly identical to that of the superclass, so nothing should change.

Can you try again without it? If it doesn't work, can you show me how you invoke it on the Julia side?
And btw, the name of the first argument should be ``expressions``, not ``stencils``

mloubout:
>Can you try again without it? If it doesn't work, can you show me how you invoke it on the Julia side?

This is not a Julia problem, it crashes on pure python too without it, and as said above it complains about number of inputs.
`TypeError: __init__() takes exactly 2 arguments (1 given)`
>And btw, the name of the first argument should be expressions, not stencils

This is the name that was there in master, I didn't touch it.

FabioLuporini:
No, ``master`` has ``expressions`` because I changed it all over the place, I think. Where did I miss it? 
Which makes me wonder whether you are on latest master.

Please, paste here the exact Python code to reproduce this issue

mloubout:
Line  464 of `devito/operator.py` :
```
class OperatorCore(OperatorBasic):
    """
    A special :class:`OperatorBasic` that, besides generation and compilation of
    C code evaluating stencil expressions, can also execute the computation.
    """

    def __init__(self, stencils, **kwargs):
        self.profiler = Profiler(self.compiler.openmp)
        super(OperatorCore, self).__init__(stencils, **kwargs)

**Comment by FabioLuporini** [comment]
Good catch with ``stencils``

> This is not a Julia problem, it crashes on pure python too without it, and as said above it complains about number of inputs.

I still cannot reproduce this. Please provide code

**Comment by mloubout** [comment]
Just add `external=True` to any operator declaration.

**Comment by mloubout** [comment]
Ok, so after some tests, it only crashes if the stencils are passed with a keyword `stencils=` but not if only called `Operator(eqn, kwargs)`

**Comment by FabioLuporini** [comment]
Which is expected behaviour, as ``stencils`` (or better, ``expressions``) is the only compulsory parameter of an ``Operator``, and it must be provided at the beginning of the argument list.

Remove that additional constructor, and I'll merge this straight away once TRAVIS confirms is happiness.

Can you also switch the leftover ``stencils`` to ``expressions`` ? Thanks

File: examples/acoustic/operators.py [hunk 2]
```diff
@@ -54,7 +54,7 @@ def ForwardOperator(model, source, receiver, time_order=2, space_order=4,
     # Create interpolation expression for receivers
     rec_term = rec.interpolate(expr=u, u_t=ti, offset=model.nbpml)
 
-    return Operator(stencils=eqn + src_term + rec_term,
+    return Operator(eqn + src_term + rec_term, external=True,
```
mloubout: Nop, just saw that, leftover debug

FabioLuporini:
Alright, merging...

FabioLuporini:
Merged.

PR #253, opened 2017-05-22
Refactoring

### LINKED ISSUE #193: StencilKernel's AT needs to be restructured
ATM, ``StencilKernel`` has a private method to auto-tune kernels (only block sizes). We should move AT to a separate class, probably taking ``StencilKernel``s as input and returning the optimal block sizes

### LINKED ISSUE #246: exploit parameters dictionary
Some parameters around Devito should be moved to the global ``parameters`` state that has recently been merged in master. The ``compiler`` itself could be moved in there; DSE and DLE config options and thresholds should also be moved.

Example:
`
parameters = {
  'dse': {
    ...
  }
  'dle': {
    ..
  }
  'compiler': {
    ..
  }
  ...
}
`

Clearly, not all values in ``parameters`` need to be dictionaries as above

FabioLuporini:
Code refactoring

fixes #193 

EDIT: with the latest edits, this also:

fixes #246


devito/parameters.py
```diff
@@ -45,3 +47,4 @@ def _updated(self, key, value):
 parameters = Parameters()
 # Default log level set to INFO
 parameters["log_level"] = 'INFO'
+parameters["autotuning"] = os.environ.get('DEVITO_ATMODE', 'basic')
```
navjotk: I was envisioning this as the place to have defaults only. To be followed up with a function that fills up values from env vars and a corresponding function to do the same from a config file. 

If you prefer to do it this way, could you also do it for the line above (log level)?

examples/acoustic/gradient_example.py
```diff
@@ -98,23 +97,20 @@ def run(dimensions=(50, 50, 50), spacing=(15.0, 15.0, 15.0), tn=750.0,
...
+                         spc_order=space_order, save=True, dse=dse, dle=dle)
 
     fw_nosave = ForwardOperator(model, u_nosave, src, rec_t, time_order=time_order,
-                                spc_order=space_order, save=False,
-                                cache_blocking=cache_blocking, dse=dse, dle=dle,
-                                compiler=compiler, profile=True)
+                                spc_order=space_order, save=False, dse=dse, dle=dle)
 
     # Calculate receiver data for true velocity
-    fw_nosave.apply(autotuning=autotuning, rec_t=rec_t)
```
navjotk: How is this set now?

FabioLuporini: pass in ``autotune={True,False}``

devito/parameters.py
FabioLuporini: I think you're right, having functions is better

But how do we decide whether to populate from "config file" rather than "env_vars+defaults"? Do we use a special env var ?

examples/acoustic/gradient_example.py
navjotk: So do you want to change it to `autotune=` so that the example shows how to set autotuning?

devito/parameters.py
navjotk: Open to ideas, but special env var seems like the best option to me too.

examples/acoustic/gradient_example.py
FabioLuporini: OK, so, I thought about this and eventually I opted for leaving it as is. I think we can safely show autotuning in another example

devito/parameters.py
FabioLuporini: Turned out this was more tangled than I thought.

I eventually opted for a generalization effort, using the global config as expected all over Devito, introducing some nice features, and refactoring the code where needed. Please take a look; it should be much better now. Thanks for pointing it out

examples/acoustic/gradient_example.py
navjotk: There is no reason to change this but alright, as you wish :)

devito/compiler.py
```diff
@@ -86,7 +78,6 @@ def __init__(self, *args, **kwargs):
 
         if self.openmp:
             self.ldflags += ['-fopenmp']
-        self.pragma_ivdep = [Pragma('GCC ivdep')]
```
navjotk: How are these pragmas set now?

FabioLuporini: I had to drop the ``GCC ivdep`` pragma because, unlike the ``ivdep`` pragma from Intel, it doesn't stack with the ``omp simd`` pragma, which **forces** loop vectorisation (ivdep is only a suggestion to skip dependency analysis, to speed up compilation time -- which for us is already extremely short with GCC).

FabioLuporini: This comment only stands for this ``GCC ivdep``

The other pragmas now are part of the "compiler language" under the DLE (grep "Pragma" in /dle for more info)

FabioLuporini:
Happy Travis means happy us, so merging...

Issue #254, opened 2017-05-22
outdated examples


FabioLuporini:
``example_gradient`` and ``gradient_example`` need to be updated to 3.0

BTW, introduce name convention? (eg, drop "example" ?)

Note that some initial tweaks are already in #253

navjotk:
`example_constant` (not `example_gradient`) and `gradient_example` were introduced by PRs #233 and #242 respectively and broken by #243.

I have been trying to fix `gradient_example` and have commented on #243 about the problem I am facing. 

I think @mloubout is the best person to work on fixing `example_constant` since I don't yet understand how that example is supposed to work.

FabioLuporini:
OK, so I'm also appending @mloubout to the "Assignees" list

mloubout:
On it, should I just commit it to the PR or make a different one.

FabioLuporini:
if it's not a major rewrite, just add it; otherwise new PR would be better, but no worries it's OK in either way

mloubout:
Most of it in #256

mloubout:
Closing?

FabioLuporini:
Yes, closed.

Issue #230, opened 2017-04-28
Multi-model


mloubout:
This issue highlights the problem we currently have working with multiple `Model` instances.

A good example is as follow. In the case you know what the true velocity model is, you would want to compare the data generated with the true model with the data generated by your initial or inverted model to look at the similarity in the data and how well.badly it fits. The code would look like

```
m1 = Model(true_vp, ...)
m2 = Model(smooth)_vp,...)

wave_true = Acoutic(m1,...)
wave_smooth = Acoutic(m2,...)

rec_true = wave_true.Forward()
rec_smooth = wave_Smooth.Forward()

# Check differences
initial_objective = .5*np.norm(rec_true - rec_smooth)
#Plot it or whatever QC
```
The current issue is that `wave_smooth` uses the velocity model from `m1` leading ti `initia_objective=0` as `rec_smooth=rec_true`, while it shoulde be quite different.

mlange05:
OK, I think this issue needs to be unpicked at various levels:

a) Should `m1.m` and `m2.m` have different symbolic names or should they both be called `m` in the discretised equation object? If it is the former, the simplest fix seems to be to allow the name to be set through the `Model` constructor.

b) This example relies on the automatic derivation of data during operator invocation, which most likely will fail to detect aliased symbols (ie. two instances of symbolic data `m` that are different but have the same name ). Whether this is even a bug or not raises the rather fundamental question posed in issue #225 of whether Devito should mimick SymPy here and cache those two instances as one, or allow explicitly created aliases.

c) What is the rationale for binding a specific model to an operator at initialisation time, rather than invocation time? The fixes in PR #223, and indeed the entire ethos of the 3.0 API re-write, are based around being able to re-use a single operator for multiple invocations with varying data. This leads to the general discussion in #227 of should we maybe instead aim to do something like this:
```
m1 = Model(...)
m2 = Model(...)
forward = acoustic.Forward(...)
rec_true = forward(m=m1)
rec_smooth = forward(m=m2)
```

mloubout:
Concerning c), I understand the design, but it make the 'Model' object obsolete, and this class/module will eventually contain a lot more (interpolation on new grid, model constraints) that will be very tricky to handle if we remove it. The main idea (in my understanding) is to have a 'Model' that encapsulates the physical properties and should allow to have interleaving ones.

mlange05:
I'm not sure I understand your comment on #230? Can you elaborate please why that would make the `Model` class obsolete?

[...]

Issue #225, opened 2017-04-25
New `SymbolicData` instances with the same name don't cache


mlange05:
Assuming a symbolic data object object `u = DenseData(name='u', shape=...)`, auto/sympy-generated new instances of `u(x, y)`, such as `u(x, y)` and `u(x+h, y)`, all cache on the original `u` object. However, an explicitly created new instance `u2 = DenseData(name='u', shape=...)` does not, as exemplified by the test `test_cache_function_new` in `test_symbol_caching.py`. If this causes memory leaks during `clear_cache()` is currently unclear.

The larger question is, do we want explicit new instances with the same `name` attribute to cache "magically", or should the user always be in explicit control over data objects?

navjotk:
Sympy treats the cache as an optional enhancement for speed. Its behaviour is completely consistent whether the cache is used or not. We should be doing the same.
Also, the most common case of using an operator multiple times is:
```
m=DenseData()
eq=[eqn_involving_m]
op=Operator(eq)
for model in batch: #batch either comes from MPI or is being read from a file
    m1=DenseData()
    m1.data=model
    op.apply(m=m1)
```

There is a valid reason to use multiple data objects with the same symbol name and any name-based caching makes that impossible while providing no advantage. Sympy doesn't restrict the use of multiple symbols with the same name. Why do we think it is a good idea?

mlange05:
> Sympy treats the cache as an optional enhancement for speed. Its behaviour is completely consistent whether the cache is used or not. We should be doing the same.

I'm not entirely sure I understand this sentiment, but consistency with SymPy is exactly the problem that brings this about. Consider the following:
```
In [1]: from sympy import Function

In [2]: from devito import x, y

In [3]: f0 = Function('f')(x, y)

In [4]: f1 = Function('f')(x, y)

In [5]: f0.foo
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-5-c896f72c0131> in <module>()
----> 1 f0.foo

AttributeError: 'f' object has no attribute 'foo'

In [6]: f1.foo
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-6-69dad83151cf> in <module>()
----> 1 f1.foo

AttributeError: 'f' object has no attribute 'foo'

In [7]: f0.foo = 'bar'

In [8]: f1.foo
Out[8]: 'bar'
```

[...]

PR #255, opened 2017-05-22
Test for overriding with different symbol name

navjotk:
This test highlights the bug that is the blocker for #254 .

I believe this test needs to pass regardless of how issues #225 and #230 are solved. 

Edit: Note that travis is supposed to fail on this PR until the underlying issue is fixed.


FabioLuporini:
Closing this as test was already incorporated in #257

PR #256, opened 2017-05-22
Outdated ex

mloubout:
Outdated examples rewrite.

Point to any example/test I may have missed.
IN PROGRESS.

EDIT: next ones:
- ~~gradient_example~~=> Is how it should be from my understanding.
- TTI the way acoustic is done => DONE
- tti_example (2D and 3D) => DONE
- tests fix => Mostly done

EDIT May 26th:
- seismic organization changed, acoustic and tti inside seismic for more clarity and allow space for cfd,... in a better way
- Changed acoustic setup and removed special casing.


mloubout:
@navjotk I still don't really get the point of `gradient_example` specially considering that with the new interface, it makes it quite messy to enroll everything.

mloubout:
@FabioLuporini I need you to lokk at `test_tti_clusters_to_graph()` as I am not sure how to rewrite it with the new interface.

mloubout:
All seems good, only Travis fail is in `test_dse/test_tti_clusters_to_graph ` 

` DSEException: Found redundant node in the TemporariesGraph.`

@FabioLuporini any ideas what that is?

FabioLuporini:
I'll take a look later, thanks

Can you meanwhile rebase this on top of master ?

mloubout:
>Can you meanwhile rebase this on top of master ?

Waiting for #253 to land

FabioLuporini:
@mloubout How old is the EDIT in the comment above? do  you still see the DSE failure?

mloubout:
EDIT is from earlier today. The DSE failure was still there at "waiting for 253"

I'll rebase on master once #253 is in and let you know.

FabioLuporini:
@mlange05 thoughts?

examples/seismic/acoustic/acoustic_example.py
```diff
@@ -27,56 +27,26 @@ def source(t, f0):
...
+    location = np.zeros((1, ndim), dtype=np.float32)
+    location[0, :-1] = [origin[i] + dimensions[i] * spacing[i] * .5
+                        for i in range(ndim-1)]
+    location[0, -1] = origin[-1] + 2 * spacing[-1]
+    # Receivers locations
+    receiver_coords = np.zeros((dimensions[0], ndim), dtype=np.float32)
+    receiver_coords[:, 0] = np.linspace(0, origin[0] +
+                                        (dimensions[0]-1) * spacing[0],
+                                        num=dimensions[0])
+    receiver_coords[:, 1:] = location[0, 1:]
```
FabioLuporini: So this generalizes 2D/3D right?

examples/seismic/acoustic/acoustic_example.py
```diff
@@ -111,7 +81,7 @@ def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
 
     solver = setup(diemnsions=dimensions, spacing=spacing, tn=tn,
                    time_order=time_order, space_order=space_order, nbpml=nbpml,
-                   full_run=full_run, **kwargs)
+                   **kwargs)
```
FabioLuporini: Rearrange so that ``solver = ... `` stays on two lines rather than three?

examples/seismic/tti/operators.py
```diff
@@ -36,58 +45,58 @@ def ForwardOperator(model, u, v, src, rec, data, time_order=2,
         # Derive stencil from symbolic equation
         Gyp = (ang3 * u.dx - ang2 * u.dyr)
         Gyy = (first_derivative(Gyp * ang3,
-                                dim=x, side=centered, order=spc_brd,
+                                dim=x, side=centered, order=space_order,
```
FabioLuporini: This has nothing to do with this PR per se, but I think the interface to take a derivative should be simplified. Here it's taking up three lines !
Shouldn't parameters like ``dim, side, ...`` always be specified? if so, why not making this explicit rather than using key-value pairs (i.e., ``first_derivative(Gyp*ang3, x, centered, space_order,)``)

examples/seismic/tti/tti_example2D.py
```diff
@@ -44,17 +41,16 @@ def source(t, f0):
...
 receiver_coords[:, 1] = location[0, 1]
-data.set_receiver_pos(receiver_coords)
-data.set_shape(nt, 101)
+rec = Receiver(name='rec', ntime=nt, coordinates=receiver_coords)
 
-TTI = TTI_cg(model, data, src, t_order=2, s_order=spc_order)
-rec, u, v, gflopss, oi, timings = TTI.Forward()
+solver = AnisotropicWaveSolver(model, source=src, receiver=rec,
+                               time_order=time_order,
+                               space_order=space_order)
```
FabioLuporini: Can reduce to two lines, from three?

examples/seismic/tti/tti_example3D.py
```diff
@@ -49,34 +46,33 @@ def setup(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
...
     receiver_coords[:, 1] = origin[1] + dimensions[1] * spacing[1] * 0.5
     receiver_coords[:, 2] = location[0, 1]
-    data.set_receiver_pos(receiver_coords)
-    data.set_shape(nt, 101)
+    rec = Receiver(name='rec', ntime=nt, coordinates=receiver_coords)
 
-    return TTI_cg(model, data, src, t_order=time_order, s_order=space_order)
+    return AnisotropicWaveSolver(model, source=src, receiver=rec,
+                                 time_order=time_order,
+                                 space_order=space_order)
```
FabioLuporini: Same?

examples/seismic/tti/wavesolver.py
```diff
@@ -0,0 +1,112 @@
...
+            rec = Receiver(name='rec', ntime=self.receiver.nt,
+                           coordinates=self.receiver.coordinates.data)
+
+        # Create the forward wavefield if not provided
+        if u is None:
+            u = TimeData(name='u', shape=self.model.shape_domain,
+                         save=save, time_dim=self.source.nt,
+                         time_order=self.time_order,
+                         space_order=self.space_order,
+                         dtype=self.model.dtype)
```
FabioLuporini: Again, some line trimming possible?

examples/seismic/tti/wavesolver.py
```diff
@@ -0,0 +1,112 @@
...
+        # Create the forward wavefield if not provided
+        if v is None:
+            v = TimeData(name='v', shape=self.model.shape_domain,
+                         save=save, time_dim=self.source.nt,
+                         time_order=self.time_order,
+                         space_order=self.space_order,
+                         dtype=self.model.dtype)
+        # Pick m from model unless explicitly provided
+        if m is None:
```
FabioLuporini: probably don't need all these ``if X is None``; isn't ``X= X or self.model.X`` enough?

examples/seismic/tti/operators.py
```diff
@@ -148,4 +155,5 @@ def ForwardOperator(model, u, v, src, rec, data, time_order=2,
     stencils += rec.interpolate(expr=u, u_t=ti, offset=model.nbpml)
     stencils += rec.interpolate(expr=v, u_t=ti, offset=model.nbpml)
 
-    return Operator(stencils, subs=subs, dse=dse, dle=dle)
+    return Operator(stencils=stencils, subs=subs,
+                    time_axis=Forward, name='ForwardTTI', **kwargs)
```
FabioLuporini: Perhaps no need to specify ``time_axis=Forward`` as ``Forward`` is the default?

examples/seismic/tti/tti_example2D.py
```diff
@@ -44,17 +41,16 @@ def source(t, f0):
...
+rec, u, v, summary = solver.forward()
```
FabioLuporini: This whole ``test_ttiexample2D.py`` should have a structure similar to ``test_ttiexample3D.py``, where you have ``def source`` ``def setup`` ``def run``

examples/seismic/acoustic/acoustic_example.py
mloubout: yes

FabioLuporini:
In ``tti/operators.py`` it's clear how the operator definition requires different syntax for 2D and 3D (is 3D a generalization of 2D or are there some non-trivial subtleties that I'm neglecting?)
Anyway, I wonder whether it's better to have two subfolders ``2D`` and ``3D`` under ``tti`` with the relevant files in there. I think this would be helpful for new users going through the examples

mloubout:
>is 3D a generalization of 2D or are there some non-trivial subtleties that I'm neglecting?)

Well 3D has one extra physical parameter (phi) so the PDE solved is not really the same, that's why I separated it.

mloubout:
>Also, do I get it right: for acoustic we only have 3D, while for TTI we have both 2D and 3D

yes because acoustic is the same equation/parameters not matter the number of dimensions.

examples/seismic/tti/operators.py
mloubout: This parameters are optional, default is `x, centered, 1`

mloubout:
>I wonder whether it's better to have two subfolders 2D and 3D under tti
There is already quite some layers of subfolders so I would avoid it

examples/seismic/tti/wavesolver.py
mloubout: not here (from the combinations I tried)

examples/seismic/tti/wavesolver.py
mloubout: No idea if it's the same

examples/seismic/acoustic/acoustic_example.py
FabioLuporini: excellent

examples/seismic/tti/operators.py
FabioLuporini: Yeah, I saw that, but it's not exactly the point I was trying to make

examples/seismic/tti/wavesolver.py
FabioLuporini: Try; it should work. Here ``X`` is either None or some object, so it should be OK

FabioLuporini:
@mlange05 I'm basically happy with this PR, can you take a look too? I'd like to merge this ASAP because it's "the last of the bulks of changes" that we needed after 3.0

examples/seismic/tti/wavesolver.py
mlange05: This should get a little better once we have a `Grid` abstraction.

examples/seismic/tti/operators.py
mloubout: I think key value pair are better for the readability.

FabioLuporini:
Merged.

Issue #251, opened 2017-05-19
Back to time bounds


mloubout:
This is a bug I came accross helping emma.

The time loop bounds are always derived from the presence of a source or some receivers. The issue is that for a trivial stencil such as :

```
u = TimeData(name='u', shape=(100, ), time_order=1, space_order=2, time_dim=100)
# Derive the stencil according to devito conventions
eqn = Eq(u.dt,  u.dx)
stencil = solve(eqn, u.forward)[0]
op = Operator([Eq(u.forward, stencil)], subs={h: dx, s: dt})

# Execute the generated Devito stencil operator
op.apply()
```

will only execute 2 time step (`Section main<2,99>`) unless `u` is declared with `save=True`. It turns out, `TimeData` object do not have any recollection of the dimension of the time axis once initialized as is is turned into the shape of `u` which is `(2,100)`.

FabioLuporini:
Because the time dimension in ``shape`` is automatically recasted to 2 when ``save = False``. Look at the code here in ``interfaces.py``:

`
            if not self.save:
                time_dim = self.time_order + 1 
                self.indices[0].modulo = time_dim
            else:
                if time_dim is None:
                    error('Time dimension (time_dim) is required'
                          'to save intermediate data with save=True')
                    raise ValueError("Unknown time dimensions")
            self.shape = (time_dim,) + self.shape
`

And as far as I can tell from a simple ``git log -L510,518:interfaces.py``, you and @mlange05 touched that piece of code a few times. So I'm not sure what's the behaviour you would expect at this point

mloubout:
Yes, mainly @mlange05 and I kinda broke that part, I am just not sure how to keep track of the time axis size properly.

mlange05:
Ok, so the problem is that a `TimeData` object with buffered time dimension can not / should not know about the full extent of time (it's buffered, it can go on forever). This means that an explicit value for `time` needs to be provided at call-time rather than object creation. The explicit dimension overrides seem broken as well at the moment though, and I'm working on a branch to fix this. I've also added a warning in that branch that `time_dim` will be ignored if `save=False`, and we'll probably want a run-time error forcing users to provide a `time` or `t` value instead.

FabioLuporini:
BTW, can we not drop ``time_dim`` and simply pass the information as first entry of the shape? In the end, this is a ``TimeData`` object...

FabioLuporini:
Also, this might impact #236 ( @navjotk )

mlange05:
[...]

Issue #250, opened 2017-05-19
Arguments for apply()


mloubout:
So there is a little "bug" in operator line 152.

```
        for argname in o_vals.keys():
            if not arguments[argname].shape == o_vals[argname].shape:
                raise InvalidArgument("Shapes must match")
```

The little problem with this is, any passed argument needs to have the exact same shape, including number of time steps. I kinda of defeat the purpose of generic dimensions if there sizes can't be changed.

AN example would be

```
u = TimeData(name='u', shape=(100, ), time_order=1, space_order=2, time_dim=100)
# Derive the stencil according to devito conventions
eqn = Eq(u.dt, u.dx)
stencil = solve(eqn, u.forward)[0]
op = Operator([Eq(u.forward, stencil)], subs={h: dx, s: dt})

# Execute the generated Devito stencil operator
op.apply()
```

the doing 

```
u2 = TimeData(name='u', shape=(100, ), time_order=1, space_order=2, time_dim=200)
op.apply(u=u2)
```

produces

```
InvalidArgumentTraceback (most recent call last)<ipython-input-7-3e325d1513f1> in <module>()
      6 # stencil = solve(eqn, u2.forward)[0]
      7 # op = Operator([Eq(u2.forward, stencil)], subs={h: dx, s: dt})
----> 8 op.apply(u=u2)
/Users/mloubout/Dropbox/London/CodeGen/devito/devito/operator.pyc in apply(self, *args, **kwargs)
    478         """Apply the stencil kernel to a set of data objects"""
    479         # Build the arguments list to invoke the kernel function
--> 480         arguments, dim_sizes = self.arguments(*args, **kwargs)
    481 
    482         # Invoke kernel function with args
/Users/mloubout/Dropbox/London/CodeGen/devito/devito/operator.pyc in arguments(self, *args, **kwargs)
    152         for argname in o_vals.keys():
    153             if not arguments[argname].shape == o_vals[argname].shape:
--> 154                 raise InvalidArgument("Shapes must match")
    155 
    156             arguments[argname] = o_vals[argname]
InvalidArgument: Shapes must match
```

FabioLuporini:
I wonder if this our usual caching @mlange05

mloubout:
Well actually, thinking of it, it would be odd and error prone to pass arguments with differet sizes to `apply` as there is other objects ,`m, src,rec` that may not be compatible with this size leading to segfaults or errors.

FabioLuporini:
[...]

Issue #236, opened 2017-05-05
Strange behaviour of TimeData constructor


navjotk:
The behaviour of `TimeData` is essentially broken unless the `shape` property is directly provided in the constructor. If a time dimension is passed explicitly, the shape is incorrectly inferred. If it is not passed, the functional indices are incorrectly inferred which breaks `schedule_expressions` in `StencilKernel`. 

```
>>> from devito import TimeData, t, time, x, y, z
>>> a = TimeData(name="a", dimensions=(x, y, z), time_dim=100)
>>> a.indexed.function
a(x, y, z)
>>> a.shape
(2, None, None, None)
>>> b = TimeData(name="b", dimensions=(time, x, y, z), time_dim=100)
>>> b.indexed.function
b(time, x, y, z)
>>> b.shape
(2, None, None, None, None)
```

mloubout:
> `b = TimeData(name="b", dimensions=(time, x, y, z), time_dim=100)`

This is an odd way to declare it. You create a time data object with two time dimensions. If you add `time` in dimension you don't have to respecify it in `time_dim`.

navjotk:
```
>>> c=TimeData(name="c", dimensions=(time, x, y, z))
>>> c.indexed.function
c(time, x, y, z)
>>> c.shape
(2, None, None, None, None)
```

mloubout:
As I said, `TimeData` is `dimension + time_dim` if you put `time` in dimensions you add a second time dimension. The behavior s not broke as it does what it is supposed to, it main be a design flow if that's not the best way to do it.

navjotk:
My first example was without putting in the time dimension explicitly. In that case, `a.indexed.function` (which is the symbolic representation) does not contain a time dimension. Since the `Operator` uses this symbolic representation to figure out the dimensionality of the object, I can't do this to the `a` object in my example code: `a=1` or `a[time, x, y, z]=1`. `Operator` croaks about not finding a `time` dimension since it uses `a.indexed.function` to find the dimensions. 

Essentially there is no way to instantiate a `TimeData` object purely from `Dimension`s (and not passing a shape) and use the object in an `Operator`.

FabioLuporini:
@mlange05 This can also probably be closed

navjotk:
At least write something about this in the docstring before closing the issue.

FabioLuporini:
I thought this was "fixed" because a merged PR (#257) was referencing this issue.

navjotk:
[...]

PR #257, opened 2017-05-23
Various fixes to runtime argument derivation

mlange05:
The primary goal of this PR is to fix the derivation of dimension sizes so that we can supply the number of timesteps as `op(time=no_timesteps)` or `op(t=no_timesteps)`. The PR _should_ make that part of `Operator.arguments()` easier to understand, and also fixes various other issues in the process: issues #236, #250 #251 and the problem outlines in #255. In particular it does:

* Add a separate test for symbol overrides via numpy arrays
* Fix the `TimeData` constructor so that the `dimensions` parameter now works as the `shape` one - ie. user-supplied argument for `shape` and `dimensions` ignore time and we implicitly add the time dimension based on `save`/`time_order`/`time_dim`.
* We now explicitly warn if the user sets `time_dim` on a buffered time object (`save=False`), since the `TimeData` objects do not retain that information, and defaulting to the buffer size makes little sense.
* In the same vain, we remove a test that checks for defaulting to the size of the time buffer if no `time_dim` is supplied - instead this code will now error and prompt the user to supply a value for `time` at runtime.
* Dimension size checks now check for `<=` than the actual data size. This means we can now iterate over subsets of our data, for example for processing small time-slices. 
* Use the "target" name of a substituted symbol, so that `op(a=a1)` works if `a1` has a different name to `a`.

Final developer note: This PR fixes a bunch of issues and slightly improves the `arguments` function in `operator.py`. However, a more comprehensive re-write would still be desirable to improve readability, modularity and maintainability and avoid users having to scroll through a huge backtrace to get to explicit warning messages. This is left for another PR though due to time constraints.

### DIFF
--- a/devito/interfaces.py
+++ b/devito/interfaces.py
@@ -482,6 +482,8 @@ class TimeData(DenseData):
 
     :param name: Name of the resulting :class:`sympy.Function` symbol
     :param shape: Shape of the spatial data grid
+    :param dimensions: The symbolic dimensions of the function in addition
+                       to time.
     :param dtype: Data type of the buffered data
     :param save: Save the intermediate results to the data buffer. Defaults
                  to `False`, indicating the use of alternating buffers.
@@ -494,7 +496,9 @@ class TimeData(DenseData):
     Note: The parameter :shape: should only define the spatial shape of the
     grid. The temporal dimension will be inserted automatically as the
     leading dimension, according to the :param time_dim:, :param time_order:
-    and whether we want to write intermediate timesteps in the buffer.
+    and whether we want to write intermediate timesteps in the buffer. The
+    same is true for explicitly provided dimensions, which will be added to
+    the automatically derived time dimensions symbol.
     """
 
     is_TimeData = True
@@ -533,15 +537,10 @@ class TimeData(DenseData):
                       automatically infer dimension symbols.
         :return: Dimension indices used for each axis.
         """
-        dimensions = kwargs.get('dimensions', None)
-        if dimensions is None:
-            # Infer dimensions from default and data shape
-            save = kwargs.get('save', None)
-            tidx = time if save else t
-            _indices = [tidx, x, y, z]
-            shape = kwargs.get('shape')
-            dimensions = _indices[:len(shape) + 1]
-        return dimensions
+        save = kwargs.get('save', None)
+        tidx = time if save else t
+        _indices = DenseData._indices(**kwargs)
+        return tuple([tidx] + list(_indices))
 
     def _allocate_memory(self):
         """function to allocate memmory in terms of numpy ndarrays."""
--- a/devito/interfaces.py
+++ b/devito/interfaces.py
@@ -8,7 +8,7 @@ from devito.dimension import t, x, y, z, time
 from devito.finite_difference import (centered, cross_derivative,
                                       first_derivative, left, right,
                                       second_derivative)
-from devito.logger import debug, error
+from devito.logger import debug, error, warning
 from devito.memmap_manager import MemmapManager
 from devito.memory import CMemory, first_touch
 
@@ -507,11 +507,16 @@ class TimeData(DenseData):
         if not self._cached():
             super(TimeData, self).__init__(*args, **kwargs)
             self._full_data = self._data.view() if self._data else None
-            time_dim = kwargs.get('time_dim')
+            time_dim = kwargs.get('time_dim', None)
             self.time_order = kwargs.get('time_order', 1)
             self.save = kwargs.get('save', False)
 
             if not self.save:
+                if time_dim is not None:
+                    warning('Explicit time dimension size (time_dim) found for '
+                            'TimeData symbol %s, despite \nusing a buffered time '
+                            'dimension (save=False). This value will be ignored!'
+                            % self.name)
                 time_dim = self.time_order + 1
                 self.indices[0].modulo = time_dim
             else:
--- a/devito/operator.py
+++ b/devito/operator.py
@@ -138,8 +138,15 @@ class OperatorBasic(Function):
         # Will perform auto-tuning if the user requested it and loop blocking was used
         maybe_autotune = kwargs.get('autotune', False)
 
+        # Initialise argument map and a map of dimension names to values
         arguments = OrderedDict([(arg.name, arg) for arg in self.parameters])
-        dim_sizes = {}
+        dim_sizes = dict([(arg.name, arg.size) for arg in self.parameters
+                          if isinstance(arg, Dimension)])
+
+        # Override explicitly provided dim sizes from **kwargs
+        for name, value in kwargs.items():
+            if name in dim_sizes:
+                dim_sizes[name] = value
 
         # Have we been provided substitutes for symbol data?
         # Only SymbolicData can be overridden with this route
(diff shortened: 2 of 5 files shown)


mloubout:
Looks pretty good overall. 
Small question. In the case of a single stencil such ass `Eq(u.forwd, ...)` with Buffered time dimensions (no save). Is it required to specify the number of time steps at `apply`?

devito/operator.py
```diff
@@ -150,62 +157,76 @@ def arguments(self, *args, **kwargs):
...
+                    # We know the dimension size, check if data shape agrees
+                    if not dim_sizes[dim.name] <= shape[i]:
+                        error('Size of dimension %s was determined to be %d, '
+                              'but data for symbol %s has shape %d.'
+                              % (dim.name, dim_sizes[dim.name], fname, shape[i]))
+                        raise InvalidArgument('Wrong data shape encountered')
+
+        # Make sure we have defined all buffered dimensions and their parents,
+        # even if they are not explicitly given or used.
+        d_args = [d for d in arguments.values() if isinstance(d, Dimension)]
```
FabioLuporini: ``dim_sizes`` is already a ``dict`` of ``Dimension``s. I wonder if you could drop ``d_args``, change ``dim_sizes`` to a dict from ``name -> obj`` , rather than ``name -> size`` (as it is now) -- in which case the name should probably be changed to something else -- and simply reuse it here in place of ``d_args``.

devito/operator.py
```diff
@@ -150,62 +157,76 @@ def arguments(self, *args, **kwargs):
...
+                        error('Size of data argument for %s is greater than the size '
+                              'of dimension %s: %d' % (fname, dim.name, dim.size))
+                        raise InvalidArgument('Wrong data shape encountered')
                     else:
-                        # Derive size from grid data shape and store
-                        dim_sizes[dim] = shape[i]
-                else:
-                    if not isinstance(dim, BufferedDimension):
-                        assert dim.size == shape[i]
+                        continue
```
FabioLuporini: I cannot say I fully understand what's going on with this ``continue`` here. Why can the ``if-else`` below be skipped ?

mlange05:
> Small question. In the case of a single stencil such ass Eq(u.forwd, ...) with Buffered time dimensions (no save). Is it required to specify the number of time steps at apply?

Yes, it is. The previous method of defaulting to the buffer size (ie. 2 or 3, depending on `time_order`) seems very arbitrary and error-prone, since it's not necessarily what users would expect. On top of that, I would argue that it is almost always a good idea to supply the number of timesteps during `apply()` rather than use the defaulting mechansm, since it's the one parameter that changes most frequently - but of course that kind of comes down to taste, I guess.

devito/operator.py
mlange05: Ok, so I'm not sure I understand you comment completely. `dim_sizes` is now `name->value`, whereas is was `object->value` before. The reason for that change is, so that we can easily override via `kwargs`, without having to derive the object from the name. The reason `name->object` won't work is that the `Dimension` object doesn't (and I argue shouldn't) store the particular value for a single invocation, since on the next invocation that value might be different. Imagine `op(time=5); do_some_checkpointing; op(time=3)`. As a corollary, when we actually need the `Dimension` object, we pull it from `arguments` based on name - but I guess we could do the same for `kwargs` replacement and make `dim_sizes` `object->value` again. I'm not sure which way is best, tbh...

devito/operator.py
mlange05: Because in this case the `dim` has already been resolved at the code level, ie. `dim` is not part of the function signature because we hard-coded values in the kernel. In that case all we care about is that the data we run over adheres to the shape we hard-coded into the kernel.

mloubout:
Good for me, was just for information :)

devito/interfaces.py
```diff
@@ -503,11 +507,16 @@ def __init__(self, *args, **kwargs):
...
             self.time_order = kwargs.get('time_order', 1)
             self.save = kwargs.get('save', False)
 
             if not self.save:
+                if time_dim is not None:
+                    warning('Explicit time dimension size (time_dim) found for '
+                            'TimeData symbol %s, despite \nusing a buffered time '
+                            'dimension (save=False). This value will be ignored!'
+                            % self.name)
                 time_dim = self.time_order + 1
```
navjotk: Is this line still required?

tests/test_operator.py
```diff
@@ -150,17 +150,13 @@ class TestArguments(object):
...
-        a1 = symbol(name='a', dimensions=(i, j, k, l), value=3.,
-                    mode='indexed').base.function
-        a2 = symbol(name='a', dimensions=(i, j, k, l), value=4.,
-                    mode='indexed').base.function
-        eqn = Eq(a, a+3)
-        op = Operator(eqn)
+        a = symbol(name='a', dimensions=(i, j, k, l), value=2.)
+        a1 = symbol(name='a1', dimensions=(i, j, k, l), value=3.)
+        a2 = symbol(name='a2', dimensions=(i, j, k, l), value=4.)
+        op = Operator(Eq(a, a + 3))
```
navjotk: I think this test should go back to what it was until we decide to explicitly prohibit multiple symbols of the same name. The change you are making here has already been added as a new test in #255 . Just merge that here.

devito/operator.py
FabioLuporini: OK, sorry, my comment was just nitpicking, and probably I was also wrong

devito/operator.py
FabioLuporini: Oh, I see

FabioLuporini:
OK, good to go

FabioLuporini:
@mlange05 I took the liberty to rebase your branch on top of ``master``. I had to fix small conflicts, but I shouldn't have broken anything. Let's wait for Travis' opinion.

FabioLuporini:
Merging...

FabioLuporini:
Merged.

navjotk:
Ummm....my comments above have been ignored.

FabioLuporini:
My bad, I completely missed them and this morning when I was suggesting to merge nobody complained so I didn't re-check.

OK, let's address them now (I can file a separate PR for those)


> Is this line still required?

Looks like it can be squashed with the line below -- is that what you meant ? Agreed, in that case



> I think this test should go back to what it was until we decide to explicitly prohibit multiple symbols of the same name. The change you are making here has already been added as a new test in #255 . Just merge that here.

Can we then: 
- revert the example
- merge #255 by x-failing the test?

mloubout:
> revert the example
> merge #255 by x-failing the test?

Well it's already in, so we can just close 255 no?

navjotk:
Still need the test for overriding symbols with the same name. That test was removed in this PR and changed to the one that already existed in #255 .

FabioLuporini:
> Still need the test for overriding symbols with the same name. That test was removed in this PR and changed to the one that already existed in #255 .

So is this test now passing, while in your PR it's failing ?

FabioLuporini:
> Looks like it can be squashed with the line below -- is that what you meant ? Agreed, in that case

In fact, I'm wrong; ``time_dim`` is also used a few lines below

navjotk:
> So, this test is now passing, while in your PR it's failing. Why is that the case?

Because this PR fixes the issue while my PR only highlights it with a/the same test.

PR #259, opened 2017-05-23
test_foreign

mloubout:
New test for `external=True`

This will ensure that 
- There is no initialization/compilation bug
- It actually produces the same result than `OperatorCore`


FabioLuporini:
Excellent, this can land. @mlange05 please merge if you're happy

PR #261, opened 2017-05-24
Fix doc generator problems

navjotk:
Due to some problems with the configuration of the documentation generator, the last merge build failed on master. This attempts to fix those issues.


FabioLuporini:
yes go go

PR #262, opened 2017-05-24
Test Born accuracy

mloubout:
This is the gradient test for Born.

I also added some docstring for the gradient test for clarity.


FabioLuporini:
Can we not reuse the code from some of the examples ? Or should we not move this to the examples, and then derive a test as we usually do?

mloubout:
>Can we not reuse the code from some of the examples / should we not move this to the examples, and then derive a test as we usually do?

I gave it a try, probably some F***8 left but let me know if you like the way it is now.

tests/test_gradient.py
```diff
@@ -2,123 +2,60 @@
...
-
+    rec, u, _ = wave.forward()
     # Compute receiver data and full wavefield for the smooth velocity
-    rec0, u0, _ = wave.forward(m=model0.m, save=True)
+    rec0, u0, _ = wave.forward(m=m0, save=True)
 
     # Objective function value
     F0 = .5*linalg.norm(rec0.data - rec.data)**2
-
+    print(F0)
```
FabioLuporini: use logger

FabioLuporini:
Much better.

I'd like @mlange05 to take a look at this though.

examples/acoustic/operators.py
```diff
@@ -24,9 +24,9 @@ def ForwardOperator(model, source, receiver, time_order=2, space_order=4,
                  time_order=time_order, space_order=space_order, save=save,
                  dtype=model.dtype)
     src = PointSource(name='src', ntime=source.nt, ndim=source.ndim,
-                      npoint=source.npoint)
+                      npoint=source.npoint, coordinates=source.coordinates.data)
```
FabioLuporini: what's happening here? @mlange05 ?

tests/test_gradient.py
mloubout: debug remaining

examples/acoustic/operators.py
mloubout: all src/ rec allways at position (0, 0) without it.

examples/acoustic/acoustic_example.py
```diff
@@ -24,55 +27,94 @@ def source(t, f0):
...
+        location[0, 1] = origin[1] + dimensions[1] * spacing[1] * 0.5
+        location[0, 2] = origin[1] + 2 * spacing[2]
+
+        # Receiver coordinates
+        receiver_coords = np.zeros((dimensions[0], 3))
+        receiver_coords[:, 0] = np.linspace(0, origin[0] +
+                                            (dimensions[0] - 1) * spacing[0],
+                                            num=dimensions[0])
+        receiver_coords[:, 1] = origin[1] + dimensions[1] * spacing[1] * 0.5
+        receiver_coords[:, 2] = location[0, 2]
```
FabioLuporini: TTI has ``tti_example2D.py`` and ``tti_example.py`` for 3D.
I think we should be consistent

mloubout: Sure, but more part of #256

examples/acoustic/operators.py
mloubout: @mlange05 what's the point of this two extra initialization, there is already a `source` and `receiver` in the input that can be directly used.

examples/acoustic/example_constant.py
```diff
@@ -60,34 +57,31 @@ def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
...
 
     info("Applying Adjoint")
-    Acoustic.Adjoint(rec, dse=dse, dle=dle)
+    solver.adjoint(rec, dse=dse, dle=dle)
     info("Applying Born")
-    Acoustic.Born(dm, dse=None, dle=dle)
+    solver.born(dm, dse=None, dle=dle)
     info("Applying Gradient")
-    Acoustic.Gradient(rec, u, dse=dse, dle=dle)
+    solver.gradient(rec.data, u, dse=dse, dle=dle)
```
FabioLuporini: Why do you sometimes need the `.data` , and some other times (`solver.` above) the object?

examples/acoustic/example_constant.py
```diff
@@ -60,34 +57,31 @@ def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
...
     if not full_run:
-        return gflopss, oi, timings, [rec, u.data]
+        return summary.gflopss, summary.oi, summary.timings, [rec, u.data]
 
     info("Applying Adjoint")
-    Acoustic.Adjoint(rec, dse=dse, dle=dle)
+    solver.adjoint(rec, dse=dse, dle=dle)
     info("Applying Born")
-    Acoustic.Born(dm, dse=None, dle=dle)
+    solver.born(dm, dse=None, dle=dle)
```
FabioLuporini: Why is `dse = None` here?

examples/acoustic/example_constant.py
```diff
@@ -60,34 +57,31 @@ def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
...
+    rec = Receiver(name='rec', ntime=nt, coordinates=receiver_coords)
 
-    Acoustic = Acoustic_cg(model, data, src, nbpml=nbpml, t_order=time_order,
-                           s_order=space_order, dse=dse, dle=dle)
+    solver = AcousticWaveSolver(model, source=src, receiver=rec,
+                                time_order=time_order, space_order=space_order)
 
     info("Applying Forward")
-    rec, u, gflopss, oi, timings = Acoustic.Forward(save=full_run, dse='basic', dle=dle)
+    rec, u, summary = solver.forward(save=full_run, dse='basic', dle=dle)
```
FabioLuporini: Why `dse='basic'` here?

examples/acoustic/wavesolver.py
```diff
@@ -108,9 +108,15 @@ def forward(self, src=None, rec=None, u=None, m=None, save=False, **kwargs):
...
-            summary = self.op_fwd_save.apply(src=src, rec=rec, u=u, m=m, **kwargs)
+            summary = self.op_fwd_save.apply(src=src, rec=rec, u=u, m=m,
+                                             src_coords=src.coordinates,
+                                             rec_coords=rec.coordinates,
+                                             **kwargs)
         else:
-            summary = self.op_fwd.apply(src=src, rec=rec, u=u, m=m, **kwargs)
+            summary = self.op_fwd.apply(src=src, rec=rec, u=u, m=m,
+                                        src_coords=src.coordinates,
+                                        rec_coords=rec.coordinates,
```
FabioLuporini: Why do you have to pass the coordinates explicitly when you could derive the directly in the callee from the actual parameters, ie `src` and `rec` ?

FabioLuporini: This comment applies to other places as well

examples/acoustic/wavesolver.py
```diff
@@ -181,7 +187,9 @@ def gradient(self, recin, u, v=None, grad=None, m=None, **kwargs):
         if m is None:
             m = m or self.model.m
 
-        summary = self.op_grad.apply(rec=rec, grad=grad, v=v, u=u, m=m, **kwargs)
+        summary = self.op_grad.apply(rec=recin.data, grad=grad, v=v, u=u,
```
FabioLuporini: Why pass in `recin.data` rather than `recin` ?

examples/acoustic/wavesolver.py
mloubout: That the main point, the coordinates are completlly ignored by the `aplly` if not explicitly passed.

FabioLuporini: Ah, yes, I understand

examples/acoustic/acoustic_example.py
```diff
@@ -1,18 +1,21 @@
...
+    out[:] = vel[:]
+    nz = shape[-1]
 
-    for a in range(5, nx-6):
-        out[a, :] = np.sum(vel[a - 5:a + 5, :], axis=0) / 10
+    for a in range(5, nz-6):
+        if len(shape) == 2:
+            out[:, a] = np.sum(vel[:, a - 5:a + 5], axis=1) / 10
+        else:
+            out[:, :, a] = np.sum(vel[:, :, a - 5:a + 5], axis=2) / 10
```
mlange05: Could we use the `...` operator to get rid of the special-casing here?

examples/acoustic/acoustic_example.py
```diff
@@ -24,55 +27,94 @@ def source(t, f0):
...
+def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
+        time_order=2, space_order=4, nbpml=40, full_run=False, **kwargs):
 
-    # Create wave solver from model, source and receiver definitions
-    solver = AcousticWaveSolver(model, source=src, receiver=rec,
-                                time_order=time_order,
-                                space_order=space_order, **kwargs)
+    solver = setup(diemnsions=dimensions, spacing=spacing, tn=tn,
+                   time_order=time_order, space_order=space_order, nbpml=nbpml,
+                   full_run=full_run, **kwargs)
```
mlange05: `full_run` is no longer required, no?

examples/acoustic/example_constant.py
```diff
@@ -60,34 +57,31 @@ def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
...
     if not full_run:
-        return gflopss, oi, timings, [rec, u.data]
+        return summary.gflopss, summary.oi, summary.timings, [rec, u.data]
 
     info("Applying Adjoint")
-    Acoustic.Adjoint(rec, dse=dse, dle=dle)
+    solver.adjoint(rec, dse=dse, dle=dle)
     info("Applying Born")
-    Acoustic.Born(dm, dse=None, dle=dle)
+    solver.born(dm, dse=dse, dle=dle)
```
mlange05: Why not use `**kwargs` to pass any compiler options?

examples/acoustic/wavesolver.py
mlange05: That is a bug. The work around looks clean enough for now, but we should really fix this in `Operator.apply()`.

tests/test_adjointA.py
```diff
@@ -2,99 +2,25 @@
...
-    time_series = np.zeros((nt, 1), dtype=np.float32)
-    time_series[:, 0] = source(np.linspace(t0, tn, nt), f0)
-
-    # Define source and receivers and create acoustic wave solver
-    src = PointSource(name='src', data=time_series, coordinates=location)
-    rec = Receiver(name='rec', ntime=nt, coordinates=receiver_coords)
-    srca = PointSource(name='srca', data=time_series, coordinates=location)
-    solver = AcousticWaveSolver(model, source=src, receiver=rec,
-                                time_order=time_order, space_order=space_order)
+    solver = setup(dimensions=dimensions, time_order=time_order,
```
mlange05: I'm not sure I like this, since you're effectively hiding the entire parameterisation of the problem setup, making it virtually impossible to understand what's going without having to chase up this function. I'd have preferred it if we could unpick the coordinate and ricker-wavelet construction to reduce clutter, but leave the main setup information in place. This just looks like we're sweeping things under the rug.

Anyway, just a personal opinion, not a showstopper for this PR.

tests/test_adjointJ.py
```diff
@@ -2,96 +2,27 @@
...
-    time_series = np.zeros((nt, 1))
-    time_series[:, 0] = source(np.linspace(t0, tn, nt), f0)
-
-    # Define source and receivers and create acoustic wave solver
-    src = PointSource(name='src', data=time_series, coordinates=location)
-    rec = Receiver(name='rec', ntime=nt, coordinates=receiver_coords)
-    solver = AcousticWaveSolver(model0, source=src, receiver=rec,
-                                time_order=2, space_order=space_order)
+    solver = setup(dimensions=dimensions,
+                   space_order=space_order)
```
mlange05: Same as above...

tests/test_gradient.py
```diff
@@ -2,123 +2,61 @@
...
+        \nabla \Phi(m0) = <J^T \delta d, dm> \\
+        \delta d = F(m0+ h dm) - D \\
+
+    with F the Forward modelling operator.
+    :param dimensions: size of the domain in all dimensions
+    in number of grid points
+    :param time_order: order of the time discretization scheme
+    :param space_order: order of the spacial discretization scheme
+    :return: assertion that the Taylor properties are satisfied
+    """
```
mlange05: Very nice, let's do more of these. :+1:

tests/test_adjointA.py
mloubout: Did that cuz fabio preferred it like that. I don't really care how it's done as long as it does not change too often.

navjotk: I'm with @mlange05 on this. I would prefer the setup in place as well - with utility functions/classes to make it concise.

examples/acoustic/acoustic_example.py
mloubout: Still need special casing for src/rec so won't make much of a difference.

tests/test_adjointA.py
mloubout: Will be part of a cleanup PR then, with everyone agreeing. This one is only about an extra test and the coordinates fixes that go with it.

examples/acoustic/example_constant.py
mloubout: will do in #256

examples/acoustic/acoustic_example.py
mloubout: will do in #256 , the cleanup-pr

FabioLuporini:
@mlange05 There's room for some clean up (e.g., getting rid of some special casing), but with @mloubout we agreed this will happen soon directly in #256 

Merging now..

FabioLuporini:
Merged.

PR #264, opened 2017-05-25
tests: Prevent garbage collection in test_dse

FabioLuporini:
Turns out the issue in ``test_dse.py`` wasn't due to "our caching bug" (ie #225 ), but to premature garbage collection . The fact that all tests start passing with the changes of this PR points (strongly) in that direction

I think this is good, because it means the DSE isn't magically affected by the caching bug we're all aware of -- at least, not in incredibly crazy subtle ways. 

I'm keen to hear what u think


FabioLuporini:
Navjot approved, so merging ...

FabioLuporini:
Merged.

PR #266, opened 2017-05-26
Speedup schedule expressions

FabioLuporini:
Much faster expression scheduler (noticeable in TTI -- significantly shorter waiting time between the DSE and the DLE outputs to terminal)

Achieved through a different implementation of `is_index`, now also supported by some tests

### DIFF
--- a/devito/dse/graph.py
+++ b/devito/dse/graph.py
@@ -203,17 +203,11 @@ class TemporariesGraph(OrderedDict):
         """
         if key not in self:
             return False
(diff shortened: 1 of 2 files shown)


FabioLuporini:
Apparently, this makes Travis time dropping from roughly 23s to 15s (ballpark numbers)

FabioLuporini:
PR been around for a while and approved, so merging...

FabioLuporini:
Merged.

PR #272, opened 2017-05-30
Fix docstrings for `TimeData` and `DenseData`

mlange05:
Smal tweaks and fixes to the docstrings in `interfaces.py`, partially in response to comments on #236. Also some tentative fixes to the deploy script that hopefully will fix the docs build.


FabioLuporini:
yes

PR #273, opened 2017-05-30
Read children data when passed override arguments

navjotk:
`PointData` are the first instance of a `SymbolicData` class holding other `SymbolicData` objects inside. This led to the problem that when `PointData` objects were passed to override `apply` arguments, the coordinates of these points which are stored as `DenseData` objects inside the `PointData` object were being ignored. 

This PR addresses that problem by adding a new base class called `CompositeData` which represents any `SymbolicData` class that might contain other `SymbolicData` objects inside. The base class provides a method `CompositeData.children()` which `Operator.arguments()` uses to construct the list of arguments including child objects when applicable. 

Putting this PR out so people aren't writing temporary fixes for something that this PR is already addressing. 

~~TODO before merging: Add a test~~


mloubout:
Would need to run the tests without the coordinate argument in wavesolver, currently only checking that the hack works.

WOuld also need to check that the argument of the forward run correspond to `solver.receiver` not to 0

navjotk:
The aliasing issue again. Suppose the name of the source symbol was `src` when setting up the operator. Now I create another source `src1` to change some details. When I call `Operator.apply(src=src1)`, the coordinates of `src1` are stored in `src1_coords` while the coordinates of the original source were in `src_coords`. At apply-time, there is no easy way to tell the operator that `src1_coords` actually corresponds to `src_coords`. 

I added an xfailing test to show this problem.

navjotk:
> currently only checking that the hack works.

I would argue that this is not a hack but a good fix to the incomplete implementation. 

Also, since this PR only introduces a single change to fix that issue, it only includes tests corresponding to that change. 

I see testing whether the receiver (which hasn't been overriden) as orthogonal to this PR.

mloubout:
>Also, since this PR only introduces a single change to fix that issue, it only includes tests corresponding to that change.

This change is not tested as the coordinates are explicitly passed inside `wavesolver`. The test would pass with or without the change in `operator.

>I see testing whether the receiver (which hasn't been overriden) as orthogonal to this PR.

I don't. The main reason of the hack is that even `apply()` with no argument does not use the coordinates in `self.receiver` unless explicitly given to `apply`

tests/test_operator.py
```diff
@@ -232,6 +234,48 @@ def test_dimension_size_override(self, nt=100):
         op(a=a, time=5)
         assert(np.allclose(a.data[0], 4.))
 
+    def test_override_composite_data(self):
+        dimensions = (100, 100)
+        time_order = 2
+        space_order = 4
+        solver = setup(dimensions=dimensions, time_order=time_order,
```
mlange05: I dont think we need the full complexity of an example setup to test this. In fact it makes the test really unreadable, since it presumes knowledge of an acoustic example, when all we want to test is whether `PointData.coordinates` is used correctly. So, could we not instead do something like (pseudo-code):
```
src1 = PointData(name='src1', coordinates=(1, 1))
src2 = PointData(name='src2', coordinates=(2, 2))
op = Operator(src1.inject(1.))
op(src1=src2)
```
and then test where the `1` values were injected in a grid of otherwise `0`s?

devito/operator.py
```diff
@@ -153,6 +153,11 @@ def arguments(self, *args, **kwargs):
         r_args = [f_n for f_n, f in arguments.items() if isinstance(f, SymbolicData)]
         o_vals = OrderedDict([arg for arg in kwargs.items() if arg[0] in r_args])
 
+        # Are any passed symbols composite? i.e. are they composed of other symbols?
+        for sym in o_vals.values():
```
mlange05: So, if we want to deal with the aliasing problem, I think we need to match the children against their "original names" from the original composite object. The original symbol, however, is lost in the line above, since it requires the name we use to match the original symbols. So we probably need to rewrite the entire section a little, so that we can match everything explicitly by name. How about an explicit loop like this (untested):
```
for name, arg in kwargs:
    if name in arguments:
        # Override the original symbol
        o_vals[name] = arg

        original = arguments[name]
        if original.is_Composite:
            for orig, child in zip(original._children, args._children):
            o_vals[orig.name] = child
```

FabioLuporini:
** random thought **
Can we revisit for a moment the assumption that we need these two objects to be nested? Why not providing them separately, as any other ``SymbolicData`` ? Can the binding between the two be done simply through ``PointData`` name? If that's the case, in ``apply`` we could raise an error if coordinates were missing

If we go down the CompositeData path instead, then the class should probably be moved to ``interfaces.py``

mlange05:
> Can we revisit for a moment the assumption that we need these two objects to be nested? Why not providing them separately, as any other SymbolicData ? Can the binding between the two be done simply through PointData name? If that's the case, in apply we could raise an error if coordinates were missing

Well, the problem is that conceptually a sparse point encapsulates data that is specific to another set of data (the coordinates), so the binding is implicitly assumed by the user, I would say. Requiring the coordinates to be supplied explicitly is what we do now (albeit we don't error if not supplied), which is the root-cause of this bug and stands in contrast to that assumption, I would argue. What's more, making the connection between data and coordinates depend on (user-driven) naming conventions sounds very dangerous to me, since we already allow override symbols to use different symbol names - hence the "aliasing problem" mentioned in earlier comments.

> If we go down the CompositeData path instead, then the class should probably be moved to interfaces.py

I totally agree with that. In fact, thinking of future implementations of staggered vector variables, this could become very useful indeed as a general concept - although that's conjecture on my part.

tests/test_operator.py
```diff
@@ -232,6 +232,26 @@ def test_dimension_size_override(self, nt=100):
         op(a=a, time=5)
         assert(np.allclose(a.data[0], 4.))
 
+    def test_override_composite_data(self):
+        original_coords = (1., 1.)
+        new_coords = (2., 2.)
+        p_dim = Dimension('p_src')
+        ndim = len(original_coords)
```
mlange05: Optional style nitpicks: All these variables could be simply used in-place in the `PointData` constructors below.

tests/test_operator.py
```diff
@@ -232,6 +232,26 @@ def test_dimension_size_override(self, nt=100):
...
+        u = TimeData(name='u', time_order=2, space_order=2, shape=(10, 10))
+        src1 = PointData(name='src1', dimensions=[time, p_dim], npoint=1, nt=10,
+                         ndim=ndim, coordinates=original_coords)
+        src2 = PointData(name='src1', dimensions=[time, p_dim], npoint=1, nt=10,
+                         ndim=ndim, coordinates=new_coords)
+        op = Operator(src1.inject(u, src1))
+
+        # Move the source from the location where the setup put it so we can test
+        # whether the override picks up the original coordinates or the changed ones
```
mlange05: Should this maybe be the docstring of this test? Again nitpicking.

mloubout:
If this is working, can we please remove the hardcoded coordiantes in the `wavesolvers`

devito/pointdata.py
```diff
@@ -36,6 +52,7 @@ def __init__(self, *args, **kwargs):
             self.coordinates = DenseData(name='%s_coords' % self.name,
                                          dimensions=[self.indices[1], d],
                                          shape=(self.npoint, self.ndim))
+            self._children.append(self.coordinates)
```
FabioLuporini: This should probably be done in the superclass, through its constructor

devito/pointdata.py
```diff
@@ -10,7 +10,23 @@
...
+class CompositeData(DenseData):
+    """
+    Base class for DenseData classes that have DenseData children
+    """
+
+    is_CompositeData = True
+
+    def __init__(self, *args, **kwargs):
+        super(CompositeData, self).__init__(self, *args, **kwargs)
+        self._children = []
```
FabioLuporini: This should be instantiated in the constructor (e.g., through ``kwargs['children']``) and made immutable (e.g., a tuple rather than a list)

devito/pointdata.py
```diff
@@ -10,7 +10,23 @@
 __all__ = ['PointData']
 
 
-class PointData(DenseData):
+class CompositeData(DenseData):
```
FabioLuporini: This should be moved to ``interfaces.py``

navjotk: Done

navjotk:
> If this is working, can we please remove the hardcoded coordiantes in the wavesolvers

Done

> This should be instantiated in the constructor (e.g., through kwargs['children']) and made immutable (e.g., a tuple rather than a list)

While I agree with the principle behind this suggestion of making it immutable, it is hard to achieve. I say that because I tried and failed.  

We want to store references to the children. However, the children can usually not be constructed before the parent since they depend on some property of the parent being initialised first. That means I can't pass them to the parent constructor. In this case it is the `Dimension` symbol that is not available in the constructor before calling the super constructor. I feel that this problem is not specific to this case only and might be generally applicable to any `CompositeData` classes.

Also, tiny detail. The list is a better choice semantically. Maybe an immutable list.

examples/seismic/acoustic/wavesolver.py
```diff
@@ -188,8 +182,7 @@ def gradient(self, rec, u, v=None, grad=None, m=None, **kwargs):
             m = m or self.model.m
 
         summary = self.op_grad.apply(rec=rec, grad=grad, v=v, u=u,
-                                     rec_coords=rec.coordinates, m=m,
-                                     **kwargs)
+                                     m=m, **kwargs)
```
FabioLuporini: nitpicking: single line?

examples/seismic/acoustic/wavesolver.py
```diff
@@ -228,6 +221,5 @@ def born(self, dmin, src=None, rec=None, u=None, U=None, m=None, **kwargs):
 
         # Execute operator and return wavefield and receiver data
         summary = self.op_born.apply(dm=dmin, u=u, U=U, src=src, rec=rec,
-                                     m=m, src_coords=src.coordinates,
-                                     rec_coords=rec.coordinates, **kwargs)
+                                     m=m, **kwargs)
```
FabioLuporini: nitpicking: single line?

devito/pointdata.py
```diff
@@ -36,6 +36,7 @@ def __init__(self, *args, **kwargs):
             self.coordinates = DenseData(name='%s_coords' % self.name,
                                          dimensions=[self.indices[1], d],
```
navjotk: Best if @mlange05 answer since he wrote this code but I suppose this is the index for the dimension that counts the number of sparse points contained in a `PointData` instance. Sometimes this dimension is explicitly provided, sometimes implicitly assumed because `npoints` was passed.

FabioLuporini:
> While I agree with the principle behind this suggestion of making it immutable, it is hard to achieve. I say that because I tried and failed.

gotcha. Let me think for a moment about this...

devito/pointdata.py
mlange05: Its the spatial dimension of the coordinates, ie. the `Dimension` object that determines whether we deal with 2D or 3D data. Its never iterated over, since the accesses are unrolled in the interpolation expression, but it's needed to allocate the coordinates data correctly.

mlange05: Ooops, sorry I misread. @navjotk is right, it's the number of points (arbitrary default `p`), while `d` is the spatial dimension.

FabioLuporini:
@navjotk Probably overthinking, so feel free to simply ditch what I write below.

Two potential (different) solutions:
1) multiple inheritance. ``CompositeData`` doesn't need to inherit from ``DenseData``, so we could rather have a ``CompositeSymbol`` (perhaps a child of ``CachedSymbol`` ?) and then call its constructor explicitly at the right time
2) Use a lambda function. Then the actual ``DenseData`` instantiation is performed within ``CompositeData``'s constructor, when ``indices`` has become available

As I said, this is probably overthinking; I'm not even sure that what I wrote even works, take it as food for thought ;)

FabioLuporini:
I still don't like this way of adding to children, but we can postpone improving this to another PR... feel free to merge. When u do so, can you close the issue corresponding to this PR, and file another Enhancement issue to keep in mind that this should/could be improved? Thanks

mlange05:
Ok, since everybody is happy (for now), I'll go ahead and merge this...

Issue #274, opened 2017-05-31
Dimensions argument not passed between run() and setup() in acoustic example


dymcc:
I make use of the acoustic example setup code to do experiments for my master's project with Fabio.
The latest version (#256) of `acoustic_example.py` has a spelling mistake in the newly introduced `run()` / `setup()` logic which makes the `setup()` function always use the default value for the `dimensions` argument.

Python is very sneaky about this... 

https://github.com/opesci/devito/blob/cc772c7f282fe6989af735c3c73a42f3187e7de9/examples/seismic/acoustic/acoustic_example.py#L81

mloubout:
Thank you for the info, good catch. Will fix it asap.

dymcc:
Didn't see your comment until after submitting the pull request :shipit:

FabioLuporini:
Closing this, as the relevant PR has been merged

PR #275, opened 2017-05-31
Correct spelling mistake in acoustic example setup

dymcc:
#274 
The keyword names now match and the correct dimensions argument is propagated through to the setup() call


FabioLuporini:
How could we miss this @mloubout ? Through the tests, I mean

I clearly also approve this, thanks

mloubout:
>How could we miss this @mloubout ? Through the tests, I mean

The test call directly setup so it was fine, only the call to setup in the example had the typo.

FabioLuporini:
Uncontroversial, so merging...

PR #279, opened 2017-06-06
Dse fix

navjotk:
Small fix to enable tests to pass on python 3.4.3 . 

I moved to this python version on my local machine to get past the matplotlib issues and the tests weren't passing without this change.


FabioLuporini:
Thanks, go ahead once Travis is green

PR #282, opened 2017-06-14
Interfaces clean up

FabioLuporini:
Fixes potential leak by now only storing a reference to CMemory
Avoid redundant memset
Drop useless, borked memmap functionality


devito/interfaces.py
```diff
@@ -270,13 +269,11 @@ def __init__(self, *args, **kwargs):
...
             self.space_order = kwargs.get('space_order', 1)
-            initializer = kwargs.get('initializer', None)
-            if initializer is not None:
-                assert(callable(initializer))
-            self.initializer = initializer
+            self.initializer = kwargs.get('initializer', None)
+            if self.initializer is not None:
+                assert(callable(self.initializer))
             self.numa = kwargs.get('numa', False)
-            self._data = kwargs.get('_data', None)
```
mlange05: You're dropping the `_data` backdoor here, which could be used to inject remotely allocated memory (eg. from Julia). I know we're not really using this right now, but let's keep in mind that we need to re-instate something similar if the Julia layer needs to provide it's own pointer, for example to provide non-copy data pointers.

FabioLuporini: I can reinstate it if you think it could actually be useful. I thought they could easily work around that by using `initializer`?

FabioLuporini:
@mloubout are you using the ``_data`` backdoor in Julia?

mloubout:
Still playing around trying to figure out how to do it, so not currently no.

devito/interfaces.py
mlange05: Well, not really, `initializer` is effectively some kernel, `_data` is a raw pointer. Anyway, not a showstopper, just something to keep in mind...

mlange05:
Ok, cool. Merging...

Issue #281, opened 2017-06-14
Reverse flag for time


mloubout:
Just an issue a bout the reverse flag.

Currently, the time loop is written at the first invocation of `apply()` with the current `reverse` property of the global `time dimension`. This is problematic if anither operator with a different flag has been created in the mean as it would change the time axis.

FabioLuporini:
This is gonna disappear automagically once Navjot's arguments handling magic is in. So assigned to him

mlange05:
@FabioLuporini @navjotk Not necessarily, since we'll probably keep an explicit flag to indicate direction, as discussed in #247. The problem with this bug here is that the `Iteration` objects for time in both forward and backward operators read their direction from a global `time` dimension - and what's more they do this during `.ccode`, which is delayed until the first `operator.apply()`. So the first operator reads the dimension direction after the second one has changed it. The fix should be easy, and I'll try to come up with a test for the problem.

navjotk:
#283 offers a fix for this but maintains the stateful `reverse` flag on the global object representing the time dimension. This can potentially cause issues if two `Operator` constructors are running in parallel - one with `reverse = True` and one with `reverse = False`. I argue that having state on a global object is so evil, it should be avoided at all costs. 

There have been various suggestions on how to avoid global state at the design level. I am compiling them along with the associated discussions here to enable further discussion:

### Suggestion 1:
>`_schedule_expressions` could pass the reverse flag to the Iteration constructor from the `Operator`'s `reverse` property, without ever having to store it on the `Dimension`.

@mlange05 's response: `schedule_expression` creates (multiple) `Iteration` objects per `Dimension`, so somehow bypassing the parent object (dimension) when injecting additional information into the derivative object (iteration) defies the point of deriving it from the dimension in the first place. At the very least this sounds counter-intuitive, and no better than the current state hackery, at least IMHO.

My response: Deriving information from the parent object does not require storing any information on the parent. The same could easily be achieved by having a dict mapping `Dimension` objects to their direction (among many other possibilities). I don't see how creating multiple `Iteration` objects per `Dimension` makes anything different.

### Suggestion 2:

> Have a time_r dimension internally, which the Operator switches to, transparently, when it receives a time_axis=Backward

[...]

PR #283, opened 2017-06-15
Fix iteration direction

mlange05:
This merge adresses #281 by storing the value of the `reverse` flag per iteration object (on creation), so that changing the direction on the (global) direction object does not affect the code generated for a particular loop. The problem was that we previously always consulted the dimension object at code generation time, which can be delayed, so that the direction of `time` might have changed already.

The PR also includes an additional test for this edge case and removes the `dle=None` and `dse=None` suppression in the timestepping tests.


mloubout:
Looks good, I'll give it a try as soon as I find time :)

navjotk:
This does fix the problem but I am inclined to argue that the root cause for the issue is in [operator](https://github.com/opesci/devito/blob/master/devito/operator.py#L77). 
The `time` dimension is a global object and ideally we shouldn't be changing state on a global object at all. 

If, due to some convoluted requirements, a piece of code is creating two `Operator`s in parallel - one forward and one reverse, this problem might resurface since there is no lock on `time.reverse`. 

I think the best solution here would be to remove `time.reverse` or not have `time` as a global object since having state on a global object is inherently issue-prone.

mlange05:
> The time dimension is a global object and ideally we shouldn't be changing state on a global object at all.

Generally I agree that we shouldn't be changing state on `time`, which is why I originally proposed having the second, derived dimension object `time_r` to denote "reverse time". This, however, was rejected as "too confusing from a user perspective", so we stayed with a single dimension, forcing ourselves to commit dynamic global state-change crimes.

> I think the best solution here would be to remove time.reverse or not have time as a global object since having state on a global object is inherently issue-prone.

You'll have to provide a little more detail than that. If I don't have a global `time` dimension, how do I figure out that I can merge two outer loops, for example if I have two synchronised state updates of `u(time, ...)` and `v(time, ...)` in a coupled equation (for example see Burgers tutorial)? Alternatively, if I don't have `time.reverse` how do I recognise that I run a `gradient` operator backwards and inject `rec.data` values using the correct time indices (note that the original time index on `rec` is foward)?

navjotk:
`_schedule_expressions` could pass the `reverse` flag to the `Iteration` constructor from the `Operator`'s `reverse` property, without ever having to store it on the `Dimension`, maybe?

We could safely ignore my objections if you think that it is impossible for two `Operator` constructors to be running in parallel in the same python environment.

FabioLuporini:
My suggestion.

Allow Operators to accept dimensions at construction time (through ``kwargs``), with the following syntax (definitely improvable):
`Operator(..., t=REVERSE)`
``REVERSE`` is a special object (there can be many...) that is telling the Operator to traverse all iteration loops using dimension `t` in the reverse direction. The constructor propagates this information to schedule_expression, which will then know how to construct the ``t`` loops. This way, no internal state is changed. Also, the solution becomes completely general (we'd not be specialising time...)

mlange05:
Ok, let's go through these in detail:
> _schedule_expressions could pass the reverse flag to the Iteration constructor from the Operator's reverse property, without ever having to store it on the Dimension, maybe?

`schedule_expression` creates (multiple) `Iteration` objects per `Dimension`, so somehow bypassing the parent object (dimension) when injecting additional information into the derivative object (iteration) defies the point of deriving it from the dimension in the first place. At the very least this sounds counter-intuitive, and no better than the current state hackery, at least IMHO.

> We could safely ignore my objections if you think that it is impossible for two Operator constructors to be running in parallel in the same python environment.

Always and trivially true due to Python's [GIL](https://wiki.python.org/moin/GlobalInterpreterLock). Python's memory management is inherently unsafe, so the interpreter can never be threaded, and the two constructors will always execute serially.

> Edit: Second idea: Have a time_r dimension internally, which the Operator switches to, transparently, when it receives a time_axis=Backward

Trickier than it might seem, since it effectively comes down to applying a `subs({time: time_r})` over all received expressions. But what about `t`, or even some other derived `t_s` (eg. for sub-sampling)? What about `time` in the src/rec terms? I think this approach is really only safe if the user has full control of choosing the symbols for each term, but that makes the API slightly messier due to various built-in variants of "time".

> Allow Operators to accept dimensions at construction time (through kwargs), with the following syntax (definitely improvable):
Operator(..., t=REVERSE)

Hmmm ok, I don't really like this approach either, since it effectively encapsulates the functionality of generic "derived" dimensions but it forces us to encode each transformation in a special-case object, just like `REVERSE`. A more natural way would be to say `time_r = DerivedDimension(expr=time_end - time)` (or anything similar that permits generic expressions rather than hard-coding individual transformations) and then do `Operator(... subs={time: time_r})`. But at that point we are back to whether we should even use a derived dimension to encapsulate "reverse-time" or not and how to deal with granularity (see above discussion).

However, to make a long story short, this discussion is really quite pointless because:
**In standard Python, two operator constructors will never execute in parallel due to GIL!**

FabioLuporini:
> However, to make a long story short, this discussion is really quite pointless because:
> In standard Python, two operator constructors will never execute in parallel due to GIL!

Sure, I was just saying to avoid changing global state 

> Hmmm ok, I don't really like this approach either, since it effectively encapsulates the functionality of generic "derived" dimensions but it forces us to encode each transformation in a special-case object, just like REVERSE [...]

I'm not sure I share your concern here. As in, in theory yes, I agree with what you say, but... How many such objects would you have, in practice? Just a few, for starters just <= 1 (as the default is the forward direction)

Anyway, said that, I'm OK with this PR. So please just land it :)

navjotk:
I was convinced but then I came across this [threading issue in the cached property module](https://github.com/pydanny/cached-property/issues/6) that makes me wonder what is the difference between our code that makes this same issue a concern for them but not for us? 

Reading further, I am inclined to argue that the GIL imposes limits on _parallelism_ but not _concurrency_. It is still possible to have threads through the _multiprocessing_ and _threading_ modules. Although the GIL might prevent simultaneous execution, interleaved execution between two python threads is easily achieved through these modules.

Hence, I insist that global state will lead to race conditions.

mlange05:
Fair enough. Back to the drawing board then....

FabioLuporini:
This PR was a 2 lines change which is fixing an issue that seems 100 orders of magnitudes more likely than threading issues in the python interpreter, AFAICT.
So I suggest we merge it and file an another issue. So yes, I'm fine with replacing an issue with another issue, as long as the latter is *much less* likely than the former

navjotk:
Rebase and pass travis

FabioLuporini:
Happy, but before merge please file an issue and a draft of the potential solutions that we'll be discussing in the (not necessarily near) future

mlange05:
@navjotk Since you raised the concerns in this PR, can you please file the issue?

FabioLuporini:
Issue filed, so we could merge this once Travis is green

PR #285, opened 2017-06-22
Yask integration phase2

FabioLuporini:
* Support for data movement through YaskGrids, which basically wrap YASK Grid objects and act like numpy arrays. Why wrapping was preferred over inheritance? see comments in code, but basically to avoid wasting twice as much memory, because of the difference in storage layouts (what YASK uses and what the users wanna see)
* Add tests!
* Update to more recent YASK API
* Handle many more types of grids
* Many other small features

Next step: restructuring our source directory in a way that YASK acts like a different backend, not just a different sequence of DLE passes.

### DIFF
--- a/devito/dle/backends/yask.py
+++ b/devito/dle/backends/yask.py
@@ -1,3 +1,4 @@
+import numpy as np
 from sympy import Indexed
 
 from devito.dimension import LoweredDimension
@@ -13,6 +14,8 @@ try:
     fac = yask.node_factory()
 except ImportError:
     yask = None
+
+__all__ = ['YaskRewriter', 'yaskarray']
 
 
 class YaskRewriter(BasicRewriter):
@@ -117,3 +120,22 @@ class sympy2yask(object):
                 raise NotImplementedError
 
         return run(expr)
+
+
+class yaskarray(np.ndarray):
+
+    """
+    An implementation of a ``numpy.ndarray`` suitable for the YASK storage layout.
+
+    This subclass follows the ``numpy`` rules for subclasses detailed at: ::
+
+        https://docs.scipy.org/doc/numpy/user/basics.subclassing.html
+    """
+
+    def __new__(cls, array):
+        # Input array is an already formed ndarray instance
+        # We first cast to be our class type
+        return np.asarray(array).view(cls)
+
+    def __array_finalize__(self, obj):
+        if obj is None: return
--- a/devito/dle/backends/yask.py
+++ b/devito/dle/backends/yask.py
@@ -1,10 +1,12 @@
+import os
+
 import numpy as np
 from sympy import Indexed
 
 from devito.dimension import LoweredDimension
 from devito.dle import retrieve_iteration_tree
 from devito.dle.backends import BasicRewriter, dle_pass
-from devito.logger import dle_warning
+from devito.logger import dle, dle_warning
 from devito.visitors import FindSymbols
 
 try:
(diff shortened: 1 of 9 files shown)


devito/dle/backends/yask.py
```diff
@@ -1,27 +1,190 @@
...
+            self.grid.set_elements_in_slice(val, start, stop)
+        elif all(i == j-1 for i, j in zip(shape, self.shape)):
+            debug("YaskGrid: Setting full-array to given scalar via single grid sweep")
+            self.grid.set_all_elements_same(val)
+        else:
+            debug("YaskGrid: Setting block to given scalar via index [%s]" % str(index))
+            self.grid.set_elements_in_slice_same(val, start, stop)
+
+    def __getslice__(self, start, stop):
+        if stop == sys.maxint:
```
mlange05: Where does this behaviour come from? Is this some kind of default behaviour? Not a showstopper, just curious...

devito/dle/backends/yask.py
```diff
@@ -93,27 +258,164 @@ def nary2binary(args, op):
...
+        import yask_compiler as yc
+        # YASK compiler factories
+        cfac = yc.yc_factory()
+        nfac = yc.yc_node_factory()
+    except ImportError:
+        _force_exit("Python YASK compiler bindings")
+
+    try:
+        # Set directory for generated code
+        path = os.path.join(os.environ['YASK_HOME'], 'src', 'kernel', 'gen')
```
mlange05: Should `YASK_HOME` come from our `parameters` dictionary? Not necessarily required now, but probably good to fixup before we set this live.

devito/dle/backends/yask.py
```diff
@@ -93,27 +258,164 @@ def nary2binary(args, op):
...
+def _force_exit(emsg):
+    """
+    Handle fatal errors.
+    """
+    raise DLEException("YASK Error [%s]. Exiting..." % emsg)
+
+
+# Generic utility functions
+
+def convert_multislice(multislice, shape, mode='get'):
```
mlange05: Would be good to have a docstring detailing the index conventions. I can see `+1` and `-1` terms below, so it would be nice to document why and/or how indexing/slicing conventions differ.

devito/interfaces.py
```diff
@@ -302,7 +302,23 @@ def _indices(cls, **kwargs):
 
     def _allocate_memory(self):
         """Allocate memory in terms of numpy ndarrays."""
+        from devito.parameters import configuration
```
mlange05: Should this import be at the top of the file? This is called often, so no point keeping it in the function.The import below is fine though.

tests/test_yask.py
```diff
@@ -0,0 +1,96 @@
+import numpy as np
+from sympy import Eq
+
+import pytest  # noqa
+
+from devito import Operator, DenseData, x, y, z
+from devito.dle import YaskGrid
+from devito.parameters import configuration, defaults
+
+pexpect = pytest.importorskip('yask_compiler')  # Run only if YASK is available
```
mlange05: I take it this needs upstream changes before we can set it live in travis? Just curious...

devito/dle/backends/yask.py
FabioLuporini: I found it in some "official docs", which now I can't find again, argh. 
Here's a pointer though : https://stackoverflow.com/questions/25597815/slice-endpoints-invisibly-truncated

In practice, this is something you need to do to match NumPy's behaviour. I should have added a comment

devito/dle/backends/yask.py
FabioLuporini: Yes, this is how it will work
When YASK will be ready, probably we'll update the installation instructions to ship a virtual env, with these magic variables automatically set

devito/dle/backends/yask.py
```diff
@@ -93,27 +258,164 @@ def nary2binary(args, op):
...
+        path = os.path.join(os.environ['YASK_HOME'], 'src', 'kernel', 'gen')
+        if not os.path.exists(path):
+            os.makedirs(path)
+    except KeyError:
+        _force_exit("Missing YASK_HOME")
+
+    # Create a new stencil solution
+    soln = cfac.new_solution("Hook")
+    soln.set_step_dim("t")
+    soln.set_domain_dims(*[str(i) for i in dimensions])  # TODO: YASK only accepts x,y,z
```
FabioLuporini: Correct. Yes, will do.

devito/dle/backends/yask.py
FabioLuporini: OK

devito/interfaces.py
FabioLuporini: cyclic imports unfortunately, which didn't seem trivial to fix. So I'll attack them in the future

However, I would expect _allocate_memory to be called only a few times, on startup?

tests/test_yask.py
FabioLuporini: Yes, because YASK must be available to run these tests

FabioLuporini:
@mlange05 I've significantly improved the documentation, as per your suggestions.

If you're fine with this, I'd wait for Travis to be happy again, then I'd like to merge this PR

mlange05:
Nice, thanks. Good to go once green.

FabioLuporini:
OK, merging...

FabioLuporini:
Merged.

PR #286, opened 2017-06-27
Generalized loop blocking

FabioLuporini:
This PR does three main things:

1- it ships a range of (relatively small) improvements all over the codebase
2- bug fixes
3- it "generalizes" the loop blocking algorithm

1 - Improvements
=============
1.1 Simpler and more effective DLE ``create_elemental_functions()``
1.2 More powerful DLE inspection routines (eg, ``filter_iterations()``)
1.3 Better, refreshed documentation
1.4 Introduction of ``OperatorDebug`` to ease C-level debugging
1.5 Refactoring. Profiling now in a separate module; dropped legacy functions (now unused); more visitors, enhanced Transformer; updated tests

2 - Bug fixes
=========
2.1 Always attempt to generate parametric array sizes
2.2 Fix data type construction at code generation time (see changes in ``cgen_utils.py``)
2.3 Attach pragmas to ``Iteration``, not through ``List`` objects
2.4 Fix DLE's data dependence analysis (subtle, rare bug arising due to the other changes in this PR)
2.5 Minor fixes to some visitors (handle some corner cases)
2.6 DSE: Introduce ``cluster.reschedule()`` to honor original expression ordering (fixes subtle, rare bug)
2.7 Reuse existing ``function`` instance (type: ``AbstractSymbol``) when calling ``.indexed`` on a data object

3 - Generalized loop blocking
=======================
This is the main contribution of the PR. 
Here's the problem: *if* cross-stencil redundancies are detected by the DSE (also read the note at the bottom), then what you see [here](https://gist.github.com/FabioLuporini/0f8be298523d10b6978d55e1e5553f8f) at the top is the resulting loop nest; that is, this is the loop nest in **input** to the DLE. Capturing cross-stencil redundancies increases the memory footprint, so the only option to avoid an explosion in occupied memory is to apply loop blocking **across** the two loop nests that you see at the link (the first loop nest is pre-evaluating the cross-stencil redundancies into vector temporaries). This required two major changes to the loop blocking algorithm:

* The introduction of ``IterationFold``, as well as routines to "fold" and "unfold" iterations. Folding (which has **nothing** to do with YASK's vector folding) is the process by means of which a sequence of consecutive loop nests are mapped onto a "single abstract loop nest". You see these routines in the new Python module ``blocking_utils.py``. The usual loop blocking algorithm is then applied to ``IterationFold``s, rather than ``Iteration``s.
* Changing the original blocking algorithm so that the remainder loops are also blocked. This is necessary because the created vector temporaries are of size ``[x_block_size][y_block_size][z_block_size]``, and if a remainder loop went from 0 to say ``x_size-1``, rather than to ``x_block_size-1``, then we would end up accessing illegal memory

The (crazy) structure of the loop nest returned by the DLE is shown at the link above.

***FINAL NOTE***
The cross-stencil redundancies detection algorithm is now available in Devito and all tests pass if activated, but at the moment it is still off by default. This is because of two reasons: 1) a complete performance evaluation is missing; 2) a few minor (performance) fixes are still required. To turn the pass on, just switch the following line in ``dse/symbolics.py``:

`'_extract_time_varying': ('future',),` --> `'_extract_time_varying': ('advanced',),`

### DIFF
--- a/devito/dle/inspection.py
+++ b/devito/dle/inspection.py
@@ -3,7 +3,7 @@ from devito.visitors import FindSections
 __all__ = ['filter_iterations', 'retrieve_iteration_tree']
 
 
-def retrieve_iteration_tree(node):
+def retrieve_iteration_tree(node, mode='normal'):
     """Return a list of all :class:`Iteration` sub-trees rooted in ``node``.
     For example, given the Iteration tree:
 
@@ -19,9 +19,24 @@ def retrieve_iteration_tree(node):
     Return the list: ::
 
         [(Iteration i, Iteration j, Iteration k), (Iteration i, Iteration p)]
-    """
 
-    return [i for i in FindSections().visit(node).keys() if i]
+    :param node: The searched Iteration/Expression tree.
+    :param mode: Accepted values are 'normal' (default) and 'superset', in which
+                 case iteration trees that are subset of larget iteration trees
+                 are dropped.
+    """
+    assert mode in ('normal', 'superset')
+
+    trees = [i for i in FindSections().visit(node) if i]
+    if mode == 'normal':
+        return trees
+    else:
+        match = []
+        for i in trees:
+            if any(set(i).issubset(set(j)) for j in trees if i != j):
+                continue
+            match.append(i)
+        return match
 
 
 def filter_iterations(tree, key=lambda i: i, stop=lambda i: False):
--- a/devito/dle/inspection.py
+++ b/devito/dle/inspection.py
@@ -1,6 +1,7 @@
 from devito.visitors import FindSections
+from devito.tools import as_tuple
 
-__all__ = ['filter_iterations', 'retrieve_iteration_tree']
+__all__ = ['filter_iterations', 'retrieve_iteration_tree', 'is_foldable']
 
 
 def retrieve_iteration_tree(node, mode='normal'):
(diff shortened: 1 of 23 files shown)


devito/dle/inspection.py
```diff
@@ -19,31 +20,75 @@ def retrieve_iteration_tree(node):
     Return the list: ::
 
         [(Iteration i, Iteration j, Iteration k), (Iteration i, Iteration p)]
+
+    :param node: The searched Iteration/Expression tree.
+    :param mode: Accepted values are 'normal' (default) and 'superset', in which
+                 case iteration trees that are subset of larget iteration trees
```
mlange05: Typo: `larger`

mlange05:
Ok, confirmed offline that the "crazy" loop nests only happen with non-default cross-stencil redundancy capture. So performance regression test aren't required yet, and will come with the appropriate PR. And with that it's a "good to go" from me.

FabioLuporini:
I fixed the docstrings. I'll wait for Travis to be happy, then I'll hit merge

FabioLuporini:
@navjotk do you have anything to suggest?

FabioLuporini:
OK, merging...

FabioLuporini:
Merged.

PR #287, opened 2017-06-28
Update example to a smoother version

navjotk:
For the scipy paper, the advection example is initialised to a non-smooth function which leads to the diffusive solution as pointed out by the reviewer. 

Instead of getting into explaining the shortcomings of that solution as the reviewer asked, it might just be better to use an example which is more compatible with finite difference, i.e. a smooth initial condition.

In this PR @jhueckelheim and I worked that out to an example that gives an acceptable solution. The smooth function comes from:

Krakos, Joshua Ambre. Unsteady adjoint analysis for output sensitivity and mesh adaptation. Diss. Massachusetts Institute of Technology, 2012. **Pg 68**


mlange05:
Ok, if we change the initial condition in the code, we need to clearly state and document that change in the accompanying text. As far as I can see this PR still states that we are using a "hat function", which is now sort of wrong. We should then also properly document the deviation from the original and provide a reference as to why.

Having said that, I am also wondering if we maybe also should submit this change to the upstream tutorial, possibly alongside some remarks as to why this is a better solution? I don't really like deviating from the original setup, since the point of our tutorial is primarily API documentation, not a discussion of numerical schemes. 

Moreover, this particular tutorial is only the beginning of a series, which progresses to  non-linear advection and the Burgers equation on the same non-smooth "hat function", but with a much smoother result that the linear advection case. I'm wondering therefore if the discontinuity is possibly intended to demonstrate that a non-linear formulation improves results? And further, should we then not be using this new smooth function throughout all other tutorials?

As for the reviewer comment, I am not sure that we necessarily need to change the example, since he is explicitly asking for a discussion and explanation. The point he is making is valid, as is the point about the Laplace example, yet changing that to an implicit scheme is not great for our purposes either. However, this point is probably best discussed offline or on slack.

So, all in all, I'm sceptical about this one, but I'd appreciate community input. @ggorman @mloubout?

jhueckelheim:
I think we should separate two issues here.

One is about the tutorial. Now that you mention the other examples in the tutorial that rely on this function, I agree that the hat is suitable, along with a discussion on numerical schemes etc. I am happy to provide this text if required. @navjotk and I coded this up yesterday and used the PR to share the code in time for the scipy paper, but had no expectation of this actually being merged. Sorry if this is not very clean.

The other is about the scipy paper itself, where there's no need for consistency with the tutorial. This is about showcasing Devito. The hat is a case where FD is not an appropriate method, and Devito fails here not because of anything specific to Devito, but because FD is the wrong choice, and because this problem is known to be hard.

If we keep this example and go into the discussion that the reviewer is asking about, we are in my opinion distracting the reader with a problem for which Devito is clearly unsuitable. Since we are free to select the test case, why not choose one where Devito is appropriate?

And yes, community input welcome.

mloubout:
I would side with @jhueckelheim on this one. Even if we are following the tutorials for illustration purpose, this examples should not be bounded to a specific initial condition or other. Adapting the problem to showcase properly Devito on a fully "supported" test case would have more impact.

The full discussion on the numerical schemes is indeed major, I would however argue that this is something we could include in the notebook that can be extended and detailed rather than adding a lengthy explanation in the paper that would hinder the message and its impact.

My understanding of the review is that we either need to add an explanation or make it more suitable to avoid mis-understandings and changing the non-smooth initial state to a smooth one seems like the easiest and most succinct way to do it.

mlange05:
Ok, following offline discussion with @janhu, I think we can and should do both:
* We leave the non-smooth "copy" of the original tutorial in place, so as to be compatible with all the next tutorial steps up to the Burgers' equation.
* We add a small bit of discussion at the end of the tutorial to explain why the non-smooth initial condition is insufficient and demonstrate the smooth version. We can also add a disclaimer that this is advanced stuff that goes beyond API demonstration.
* We update to the smooth version in the paper, since it's a standalone example, and refer to this notebook and the LeVeque textbook for further discussion and insights.
* As a bonus we can explore using higher-order techniques on the non-smooth initial condition to demonstrate some of the more subtle points mentioned in the in the LeVeque textbook.

mlange05:
Ok, changes look good. It would be very nice to have an actual definition of the smooth initial condition though, since I need to update this in the paper too. Also, I'm getting warnings about notebook versions (v3 instead of v4), which suggests you might want to update your `jupyter`.

mlange05:
Also, a minor nitpick, not to be addressed in this PR, is that we do not need to replicate the _without_ and _with BC_ stuff, as well as printing the generated C code. Instead we should drop these and focus on analysing the oscillations and how they behave with higher order derivatives (as discussed offline). But again this is just ideas for another PR, not for this one.

jhueckelheim:
Yes. I also think it would be a good idea to show that higher order stencils will cause oscillations with the non-smooth IC, and remove the artificial diffusion with the smooth IC. I tried this earlier and failed due to some miscommunication between me and Devito. I added the definition of the smooth bump. Updating Jupyter seems to be less straightforward, working on it.

mlange05:
Ok, cool. As I said, does not need to be this PR. Also, what's A in the IC definition and what's the value we use?

jhueckelheim:
I added the complete IC definition, I hope this is now unambiguous.

mloubout:
Nitpicking.

In the introduction

>This is the case for systems with a smooth solution if $\Delta x$ and $\Delta y$ are sufficiently small. But if the solution is non-smooth we can't expect much.

This is not quite true. The main problem is that the error made on the approximation is led by the derivatives (first, second,...) of the fonction at the current point. As  it is not differentiable at the point, the error grows to infinity. SO the main issue is that the Taylor expansion is not defined at all pints, not the grid size.

mlange05:
Uhhh, can you please also add a one-liner to `docs/tutorials.rst` to point to the new tutorial on the website and say why it exists on?

mlange05:
Ok, good to go. Merging....

PR #288, opened 2017-07-05
Yask integration phase3

FabioLuporini:
This PR does two main things:

- It introduces the concept of ``backend`` in Devito
- It restructures YASK into a separate directory (representing one such ``backend``), and split what was a single file into multiple modules, thus making everything much more readable

Thanks to this new backend concept, we'll be able to specialize YASK operators and data types in a very natural way.

Some code was extracted and re-adapted from Firedrake's PyOP2

Note that most edits are just moving stuff around. The actually new chunks of code are essentially in:
- ``devito/.__init__.py``
- ``devito/backends.py``
- ``devito/base.py``

EDIT: I forgot to say. We'll be able to automagically switch backend on startup by simply setting the env variable ``DEVITO_BACKEND`` to a proper value (this is documented in the code, ATM)


FabioLuporini:
@mlange05 also take a look in particular at ``ad1746c``

FabioLuporini:
Introduced a further backend, ``foreign``. This is what tools like SLIM will use. This means on the Julia side it won't be necessary to add the ``external=True`` flag to an Operator; rather, just ``export DEVITO_BACKEND=foreign`` on startup.

This implies that you cannot have **both** "external" and "core" operators in the same program anymore: either one or the other. I could relax this constraint, but it'd introduce a special case. I don't think it's needed (I've asked Mathias privately already) -- I don't see why it should be -- but I'm otherwise able to reintroduce this feature if really required...

.travis.yml
```diff
@@ -51,6 +51,8 @@ script:
   - export PYTHONPATH=$PYTHONPATH:$PWD
   - flake8 --builtins=ArgumentError .
   - py.test -vs tests/
+  - DEVITO_BACKEND=yask py.test -vs tests/test_yask.py
```
FabioLuporini: Note these changes to Travis

FabioLuporini:
@mloubout @philippwitte 
Another thing I'm deprecating here are are explicit imports from Devito's "internal" modules.
In SLIM you always have things like:
`
from devito.interfaces import DenseData, TimeData
from devito.operator import Operator
from devito.dimension import t
`
from now on, you should use:
`
from devito import DenseData, TimeData, Operator, t
`
This does **not** mean that what you have will stop working once this PR is in. However, it's a (small)  change you should do in the near future, so that in a few months time you'll be able to benefit from things like YASK.

FabioLuporini:
@mlange05 Should I wait for conda to land?

FabioLuporini:
Now Travis shouldn't fail anymore. 

Rebased on master.

tests/test_operator.py
```diff
@@ -430,19 +433,22 @@ def test_directly_indexed_expression(self, fa, ti0, t0, exprs):
         assert trees[1][-1].nodes[0].expr.rhs == eqs[1].rhs
 
 
+@pytest.mark.skipif(configuration['backend'] != 'foreign',
```
mloubout: Allways skipped in TRAVIS, need to be actually tested.

FabioLuporini: doesn't this line do its purpose: https://github.com/opesci/devito/blob/80b4d6dbae041ec2d78be99323d4d7f806af1ae0/.travis.yml#L54

mloubout: It does, my bad sorry.

FabioLuporini:
This has been open for 6 days w/o getting significant comments, so I might just merge it later today unless some last minute comments pop up

devito/operator.py
```diff
@@ -365,17 +369,21 @@ def _schedule_expressions(self, clusters, ordering):
 
         return List(body=processed)
 
-    def _insert_declarations(self, dle_state, parameters):
-        """Populate the Operator's body with the required array and
-        variable declarations, to generate a legal C file."""
+    def _specialize(self, nodes, elemental_functions):
+        """Transform the Iteration/Expression tree into a backend-specific
+        representation, such as code to be executed on a GPU or through a
```
mlange05: Nitpicking: Not sure you want to mention GPUs here, since we don't support them.

FabioLuporini:
OK, merging...

FabioLuporini:
Merged.

PR #289, opened 2017-07-05
Moved things around in the arguments derivation to make it simpler

navjotk:
This took a while but I think I have something functional now. All tests pass. 

Posting this mainly for review. I understand if this branch in its current form is not considered mergeable. Tell me how to make it so and I'll do it.


mloubout:
This allocates arrays for the arguments for 'OperatorForeign' and shouldn't.
Doing some digging to figure out why.

devito/arguments.py
```diff
@@ -0,0 +1,183 @@
...
+        super(TensorArgument, self).__init__(name, source)
+        self.dtype = dtype
+        self._value = self._default_value = self.source
+
+    @property
+    def value(self):
+        if isinstance(self._value, np.ndarray):
+            return self._value
+        else:
+            return self._value.data
```
mloubout: This is what allocate for operator foreign. The cal to '.data' always call allocate if the array is not initialized.

devito/arguments.py
```diff
@@ -0,0 +1,183 @@
...
+
+        verify = verify and all([d.verify(v) for d, v in zip(self.source.indices,
+                                                             value.shape)])
+        if verify:
+            self._value = value
+
+        return self._value is not None and verify
+
+
+class DimensionArgProvider(RuntimeArgProvider):
```
FabioLuporini: Like other classes, this one should implement ``ccode`` and ``decl`` (which need to be moved from ``dimension.py``).

The base class of the hierarchy could be an abstract class and raise a ``NotImplemented`` for ``ccode`` and ``decl`` (this would follow Python convention for abstract classes/methods)

devito/arguments.py
```diff
@@ -0,0 +1,183 @@
+from cached_property import cached_property
+import cgen as c
+import numpy as np
+
```
FabioLuporini: I think this Python module needs a docstring explaining what's going on here

FabioLuporini: I feel like some of the type/variable names below don't follow a convention? 
E.g., Arg vs Argument, Prov vs Provider

devito/arguments.py
```diff
@@ -0,0 +1,183 @@
...
+
+class RuntimeArgProvider(object):
```
FabioLuporini: docstring

devito/arguments.py
```diff
@@ -0,0 +1,183 @@
...
+
+    @property
+    def rtargs(self):
+        raise NotImplemented()
+
+
+class RuntimeArgument(object):
```
FabioLuporini: docstring

devito/arguments.py
```diff
@@ -0,0 +1,183 @@
...
+        raise NotImplemented()
+
+    def reset(self):
+        self._value = self._default_value
+
+    def verify(self, kwargs):
+        raise NotImplemented()
+
+
+class ScalarArgument(RuntimeArgument):
```
FabioLuporini: docstring

devito/arguments.py
```diff
@@ -0,0 +1,183 @@
...
+        cast = c.Initializer(c.POD(self.dtype,
+                                   '(*restrict %s)%s %s' % (self.name, shape, alignment)),
+                             '(%s (*)%s) %s' % (c.dtype_to_ctype(self.dtype),
+                                                shape, '%s_vec' % self.name))
+        return cast
+
+    def verify(self, value):
+        if value is None:
+            value = self._value
+
```
FabioLuporini: nitpicking: excessive spacing ?

devito/arguments.py
```diff
@@ -0,0 +1,183 @@
...
+    def reset(self):
+        self._value = self._default_value
+
+    @property
+    def value(self):
+        return self._value
+
+    @cached_property
+    def rtargs(self):
```
FabioLuporini: what's rtags?

devito/arguments.py
```diff
@@ -0,0 +1,183 @@
...
+        # TODO: Create proper Argument objects - with good init values
+        if self.size is not None:
+            return []
+        else:
+            size = ScalarArgument("%s_size" % self.name, self, max)
+            return [size]
+
+    # TODO: Do I need a verify on a dimension?
```
FabioLuporini: All these TODOs should probably be cleared up before merging

devito/arguments.py
```diff
@@ -0,0 +1,183 @@
...
+    def verify(self, value):
+        verify = True
```
FabioLuporini: I've seen similar code in another class. Can we factorize something in one of the superclasses?

devito/arguments.py
```diff
@@ -0,0 +1,183 @@
...
+
+        if self.size is not None:
+            # Assuming the only people calling my verify are symbolic data, they need to
+            # be bigger than my size if I have a hard-coded size
+            verify = (value >= self.size)
+        else:
+            # Assuming self._value was initialised as maxint
+            if value is not None and self._value is not None:
+                value = self.reducer(self._value, value)
+            if hasattr(self, 'parent'):
```
FabioLuporini: what's going on here?

devito/dle/backends/basic.py
```diff
@@ -94,11 +94,14 @@ def _create_elemental_functions(self, state, **kwargs):
...
 
+                args = flatten([p.rtargs for p in parameters])
+
                 # Produce the new function
-                functions.append(Function(name, root, 'void', parameters, ('static',)))
+                functions.append(Function(name, root, 'void', args, ('static',)))
 
             # Transform the main tree
             imapper = {}
+
```
FabioLuporini: leftover?

devito/nodes.py
```diff
@@ -473,9 +474,19 @@ def __init__(self, name, body, retval, parameters=None, prefix=('static', 'inlin
         self.name = name
         self.body = as_tuple(body)
         self.retval = retval
-        self.parameters = as_tuple(parameters)
         self.prefix = prefix
 
+        if all([isinstance(x, RuntimeArgProvider) for x in parameters]):
```
FabioLuporini: nitpicking: use ``i`` as iteration variable, to follow common Devito convention

devito/nodes.py
```diff
@@ -473,9 +474,19 @@ def __init__(self, name, body, retval, parameters=None, prefix=('static', 'inlin
...
+            args = flatten([x.rtargs for x in parameters])
+        else:
+            assert(all([isinstance(x, RuntimeArgument) for x in parameters]))
+            args = parameters
+
+        self.parameters = as_tuple(args)
+
+        self.t_args = [x for x in args if isinstance(x, TensorArgument)]
```
FabioLuporini: can you expand the names ``t_args`` and ``s_args`` ?

FabioLuporini: Following Devito's and Sympy's conventions, can we add ``is_TensorArgument``-like fields to Argument types ? So that we don't need to use ``isinstance`` **and** we implicitly convey the message that what we have at this point is an object of **that** hierarchy

devito/operator.py
```diff
@@ -14,9 +14,9 @@
 from devito.dimension import Dimension, time
 from devito.dle import (compose_nodes, filter_iterations,
                         retrieve_iteration_tree, transform)
-from devito.dse import clusterize, indexify, rewrite, q_indexed
-from devito.interfaces import SymbolicData, Forward, Backward
-from devito.logger import bar, error, info
+from devito.dse import (clusterize, indexify, rewrite, q_indexed)
```
FabioLuporini: nitpicking, convention: on a single line, unnecessary brackets?

devito/operator.py
```diff
@@ -121,89 +122,56 @@ def __init__(self, expressions, **kwargs):
...
 
+        self.symbolic_data = [x for x in parameters if isinstance(x, SymbolicData)]
+        dims = [x for x in parameters if isinstance(x, Dimension)]
+
+        d_parents = [x.parent for x in dims if hasattr(x, 'parent')]
+        self.dims = list(set(dims + d_parents))
+        assert(all(isinstance(x, RuntimeArgProvider)
+                   for x in self.symbolic_data + self.dims))
+        parameters = list(set(parameters + d_parents))
+        print(parameters)
```
FabioLuporini: leftover

devito/operator.py
```diff
@@ -121,89 +122,56 @@ def __init__(self, expressions, **kwargs):
...
         # Finish instantiation
         super(OperatorBasic, self).__init__(self.name, nodes, 'int', parameters, ())
 
+    def _reset_args(self):
```
FabioLuporini: what's the purpose of this? I can kinda imagine it, but it needs a proper docstring

devito/operator.py
```diff
@@ -121,89 +122,56 @@ def __init__(self, expressions, **kwargs):
         # Track the DLE output, as it might be useful at execution time
         self._dle_state = dle_state
 
+        self.symbolic_data = [x for x in parameters if isinstance(x, SymbolicData)]
```
FabioLuporini: I think these can be a property.

We should try to avoid class pollution as much as possible

devito/operator.py
```diff
@@ -121,89 +122,56 @@ def __init__(self, expressions, **kwargs):
...
+        dims = [x for x in parameters if isinstance(x, Dimension)]
+
+        d_parents = [x.parent for x in dims if hasattr(x, 'parent')]
+        self.dims = list(set(dims + d_parents))
```
FabioLuporini: same as above

devito/nodes.py
FabioLuporini: This ~can~ should be a property. Let's avoid class pollution as much as possible. You'll see a similar comment later on

devito/nodes.py
```diff
@@ -473,9 +474,19 @@ def __init__(self, name, body, retval, parameters=None, prefix=('static', 'inlin
...
+        self.s_args = [x for x in args if isinstance(x, ScalarArgument)]
```
FabioLuporini: propertizable as well

devito/nodes.py
```diff
@@ -484,31 +495,14 @@ def __repr__(self):
...
-                  if isinstance(f, (SymbolicData, TensorFunction))]
-        shapes = [(f, ''.join(["[%s]" % ccode(i) for i in f.symbolic_shape[1:]]))
-                  for f in handle]
-        casts = [c.Initializer(c.POD(v.dtype,
-                                     '(*restrict %s)%s %s' % (v.name, shape, alignment)),
-                               '(%s (*)%s) %s' % (c.dtype_to_ctype(v.dtype),
-                                                  shape, '%s_vec' % v.name))
-                 for v, shape in shapes]
-        return casts
+        t_args = [x for x in self.parameters
```
FabioLuporini: calling ``t_args`` something that in the constructor has identical name, but different semantics

devito/operator.py
```diff
@@ -121,89 +122,56 @@ def __init__(self, expressions, **kwargs):
...
-            if d.is_Buffered:
-                if dim_sizes[d.parent.name] is None:
-                    dim_sizes[d.parent.name] = dim_sizes[d.name]
-                if dim_sizes[d.name] is None:
-                    dim_sizes[d.name] = dim_sizes[d.parent.name]
 
+        new_params = {}
+        for k, v in kwargs.items():
+            if isinstance(v, CompositeData):
+                orig_param = [x for x in self.symbolic_data if x.name == k][0]
```
FabioLuporini: nitpicking, convention: iteration variable should be ``i``

devito/operator.py
```diff
@@ -121,89 +122,56 @@ def __init__(self, expressions, **kwargs):
...
+        dle_arguments = self._dle_arguments(dim_sizes)
+        dim_sizes.update(dle_arguments)
+
+        for d, v in dim_sizes.items():
+            assert(dim_names[d].verify(v))
+        arguments = OrderedDict([(x.name, x.value) for x in self.parameters])
+        arguments.update(self._extra_arguments())
+
+        self._reset_args()
+        self._print_args(arguments)
```
FabioLuporini: leftover?

devito/operator.py
```diff
@@ -121,89 +122,56 @@ def __init__(self, expressions, **kwargs):
...
+        dim_names = OrderedDict([(d.name, d) for d in self.dims])
+        dle_arguments = self._dle_arguments(dim_sizes)
+        dim_sizes.update(dle_arguments)
+
+        for d, v in dim_sizes.items():
+            assert(dim_names[d].verify(v))
+        arguments = OrderedDict([(x.name, x.value) for x in self.parameters])
+        arguments.update(self._extra_arguments())
+
+        self._reset_args()
```
FabioLuporini: this should be documented

devito/operator.py
```diff
@@ -217,29 +185,19 @@ def arguments(self, *args, **kwargs):
...
-        for name, arg in arguments.items():
-            if isinstance(arg, SymbolicData) or isinstance(arg, Dimension):
-                raise ValueError('Runtime argument %s not defined' % arg)
-        return arguments, dim_sizes
+        return dle_arguments
+
+    def _print_args(self, arguments, log=debug):
+        arg_str = []
+        for k, v in arguments.items():
+            if hasattr(v, 'shape'):
```
FabioLuporini: Is this ``hasattr`` strictly necessary? We can probably do better

devito/operator.py
```diff
@@ -217,29 +185,19 @@ def arguments(self, *args, **kwargs):
...
-        arguments.update(self._extra_arguments())
-
-        # Sanity check argument derivation
-        for name, arg in arguments.items():
-            if isinstance(arg, SymbolicData) or isinstance(arg, Dimension):
-                raise ValueError('Runtime argument %s not defined' % arg)
-        return arguments, dim_sizes
+        return dle_arguments
+
+    def _print_args(self, arguments, log=debug):
```
FabioLuporini: this should be a free, utility function

devito/pointdata.py
```diff
@@ -31,7 +31,6 @@ def __init__(self, *args, **kwargs):
             self.ndim = kwargs.get('ndim')
             kwargs['shape'] = (self.nt, self.npoint)
             super(PointData, self).__init__(self, *args, **kwargs)
-
```
FabioLuporini: leftover?

tests/test_dle.py
```diff
@@ -34,6 +34,7 @@ def simple_function(a, b, c, d, exprs, iters):
     #         expr0
     #         expr1
     symbols = [i.base.function for i in [a, b, c, d]]
+    print(symbols)
```
FabioLuporini: leftover

devito/arguments.py
FabioLuporini: let's minimize the use of ``hasattr``

devito/operator.py
navjotk: No this is intentional. To print the arguments to the log in debug mode. This way we can tell if arguments were passed correctly looking at travis logs.

devito/arguments.py
navjotk: This is duck-typing. I believe that is the preferred way in python?

devito/nodes.py
navjotk: Can you help me understand what difference it makes?

navjotk:
@FabioLuporini also look at the travis logs for this one. Only fails in python 3.x . It is related to the multiple inheritance and the MRO but I am not sure why it only errors out in 3.x since nothing related to the MRO is supposed to have changed after 2.4 or something. Some help?

devito/operator.py
navjotk: It could be a `numpy.ndarray` or it could be from the `SymbolicData` heirarchy or it could be one of those wrapper objects you are building for yask allocated memory. The purpose of using `hasattr` instead of class names is that if it has a shape, it is a tensor so we shouldn't print the full tensor - only the L2 norm.

devito/arguments.py
navjotk: Done

devito/arguments.py
navjotk: Done

devito/arguments.py
navjotk: Done

devito/arguments.py
navjotk: Done

devito/arguments.py
navjotk: Are there some guidelines on what's considered excessive?

devito/arguments.py
navjotk: `Dimension` is not a part of the heirarchy that has those methods though. That is the `RuntimeArgument` heirarchy. This is the `RuntimeArgProvider` heirarchy.

devito/arguments.py
navjotk: Added docstring. It is actually Runtime Arguments.

devito/arguments.py
navjotk: I can delete the TODO comments on this branch, if that's what you mean. Actually addressing the TODOs would make this PR too big to process.

devito/arguments.py
navjotk: Please help me see the factorization opportunity

devito/operator.py
navjotk: Done

devito/operator.py
navjotk: Done

tests/test_dle.py
navjotk: Done

devito/pointdata.py
navjotk: Done

devito/arguments.py
FabioLuporini: no, that was just my personal taste, you can ignore whatever I mark as "nitpicking" :)

devito/nodes.py
FabioLuporini: There are two good reasons to do it in Devito, and probably two more in general.
For the sake of Devito:
- convention. It's the approach I've used all over the codebase during the 3.0 rewrite
- marking the attribute as read-only. If it's not a property, you could modify if from the outside
In general (ie, might be useful here in the future, or maybe not):
- we may want to do something sophisticated at getting/setting time
- we may have to recompute things on-the-fly because the object state has changed meanwhile

devito/nodes.py
FabioLuporini: nitpicking, convention: don't need list comprehension within `all`
i.e., you may just write `all(isinstance(...) ...)`

devito/operator.py
FabioLuporini: OK. Can you add a comment?

FabioLuporini: Ah, perhaps a more pythonic way ?
`
try:
    v.shape
    ...
except AttributeError:
    ...
`

FabioLuporini:
> Some help?

On it

devito/arguments.py
navjotk: I will remove `arguments` from `OperatorForeign` either in this PR or the next one. Can you change the Julia interface to use `Operator.parameters` instead?

The whole arguments derivation logic does not belong in the base class.

FabioLuporini:
> @FabioLuporini also look at the travis logs for this one. Only fails in python 3.x . It is related to multiple inheritance and the MRO but I am not sure why it only errors out in 3.x since nothing related to the MRO is supposed to have changed after 2.4 or something. Some help?

Googling I found this:

> The problem occurs because we forward up our inheritance chain any kwargs we don't use. And in python 3, anything that was not eaten by the time we call the highest class, i.e. object will cause a error. I.e. calling super on object must call with empty kwargs.

So I tried adding this to ``RuntimeArgProvider`` (the superclass):

```
        pass
```

And now it works.

Looks like you're ignoring some keyword arguments

devito/nodes.py
navjotk: Does not justify the usage of the term "class-pollution" though.

devito/arguments.py
```diff
@@ -0,0 +1,272 @@
...
+    """ Class used to decorate Symbolic Data objects with behaviour required for runtime
+        arguments.
+    """
+
+    @cached_property
+    def rtargs(self):
+        return [TensorArgument(self.name, self, self.dtype)]
+
+
+class ScalarFunctionArgProvider(RuntimeArgProvider):
```
FabioLuporini: ``ScalarFunction`` and ``TensorFunction`` represent data created internally by Devito (probably I should update the docstrings...), so we shouldn't be expecting them as Operator argument?

devito/arguments.py
```diff
@@ -0,0 +1,272 @@
...
+    """ Class used to decorate Tensor Function objects with behaviour required for runtime
+        arguments.
+    """
+
+    @cached_property
+    def rtargs(self):
+        return [TensorArgument(self.name, self, self.dtype)]
+
+
+def log_args(arguments, log=debug):
```
FabioLuporini: shouldn't this simply be ``debug`` ?

devito/nodes.py
```diff
@@ -473,9 +473,21 @@ def __init__(self, name, body, retval, parameters=None, prefix=('static', 'inlin
...
+            args = flatten([i.rtargs for i in parameters])
+        else:
+            assert(all(isinstance(i, RuntimeArgument) for i in parameters))
+            args = parameters
+
+        self.parameters = as_tuple(args)
+
+        # At this point, all objects in args should be objects of the RuntimeArgument
+        # heirarchy. Separate the tensor arguments from the scalar ones
+        self.t_args = [i for i in args if i.is_TensorArgument]
```
FabioLuporini: (reminder) Propertisable for the reasons already discussed

devito/operator.py
```diff
@@ -121,89 +122,62 @@ def __init__(self, expressions, **kwargs):
         # Track the DLE output, as it might be useful at execution time
         self._dle_state = dle_state
 
+        self.symbolic_data = [i for i in parameters if isinstance(i, SymbolicData)]
```
FabioLuporini: (reminder) propertisable, same below

devito/operator.py
```diff
@@ -121,89 +122,62 @@ def __init__(self, expressions, **kwargs):
...
+                   for i in self.symbolic_data + self.dims))
+        parameters = list(set(parameters + d_parents))
+
         # Finish instantiation
         super(OperatorBasic, self).__init__(self.name, nodes, 'int', parameters, ())
 
-    def arguments(self, *args, **kwargs):
-        """
-        Return the arguments necessary to apply the Operator.
+    def _reset_args(self):
```
FabioLuporini: Does this need to be a method? Is it going to be overridden by the subclasses? If not, I'd just inline it

devito/arguments.py
navjotk: They do go in as `Function` arguments, though.

devito/arguments.py
navjotk: Be default it will write to the `debug` log level, but that can be over-riden with the parameter. What's wrong here?

devito/operator.py
navjotk: For readability, among other reasons, it is considered a good practice to split large methods into smaller ones.

devito/nodes.py
navjotk: @FabioLuporini You haven't continued the discussion here. I am not convinced that it is wrong to not use getters here.

FabioLuporini: I had missed this

Yes, I agree class pollution is very wrong here. However, I think my first two points are still valid, in particular the convention one.

devito/arguments.py
FabioLuporini: Yeah, only private ("static") functions . Do we still need them as Providers?

devito/arguments.py
FabioLuporini: oh, nothing, I was just wondering why one would want something different than ``debug`` ?

devito/operator.py
FabioLuporini: OK. I was just wondering whether you had in mind to override it at some point, because I saw it was being called from only a single place and it was consisting of only two lines

devito/arguments.py
navjotk: The DLE tests won't pass without it.

devito/operator.py
navjotk: No

devito/nodes.py
navjotk: I argue that the current codebase uses object attributes in exactly the same way as I have used here, in multiple places. I have included links to some of them but there are many more instances. That should settle your convention argument since this style is currently in use. The other arguments should apply equally to all these instances I list below. Overall, I think this is just bike-shedding. This does not make a difference either way.

https://github.com/opesci/devito/blob/master/devito/nodes.py#L184

https://github.com/opesci/devito/blob/master/devito/dse/clusterizer.py#L21

https://github.com/opesci/devito/blob/master/devito/dse/graph.py#L137

https://github.com/opesci/devito/blob/master/devito/dse/graph.py#L143

https://github.com/opesci/devito/blob/master/devito/dse/graph.py#L148

https://github.com/opesci/devito/blob/master/devito/dle/backends/yask.py#L106

https://github.com/opesci/devito/blob/master/devito/operator.py#L81

https://github.com/opesci/devito/blob/master/devito/operator.py#L557

devito/arguments.py
FabioLuporini: what error do you get?

devito/nodes.py
```diff
@@ -484,31 +496,14 @@ def __repr__(self):
...
-        for v in self.parameters:
-            if isinstance(v, Dimension):
-                cparameters.append(v.decl)
-            elif v.is_ScalarFunction:
-                cparameters.append(c.Value('const int', v.name))
-            else:
-                cparameters.append(c.Value(c.dtype_to_ctype(v.dtype),
-                                           '*restrict %s_vec' % v.name))
-        return cparameters
+        return [v.decl for v in self.parameters]
```
navjotk: It is related to this and ccasts below. Each argument should be responsible for providing its own ccode, decl and (if applicable) ccast according to the changed approach (which is arguably cleaner than this if-else hierarchy). 

When the DLE creates elemental functions, their arguments are `ScalarFunction` and `TensorFunction` objects that don't support this. Hence they were modified to also provide `Argument` objects so the behaviour can be consistent.

devito/arguments.py
navjotk: Replied [here](https://github.com/opesci/devito/pull/289/files#r126137438)

devito/nodes.py
FabioLuporini: Oh I see, OK, thanks.

devito/arguments.py
mloubout: Let's do it in the next one and make sure for now that this data is `None` for OperatorForeign.

navjotk:
What remains here?

mloubout:
>What remains here?

None data for Foreign

navjotk:
> None data for Foreign

Can't be done. `OperatorForeign` shouldn't be calling `arguments`. What you need is called `parameters`.

mloubout:
> What you need is called parameters

`parameters` is not ordered, can't use it, and doesn't change the fact that data is still allocated

navjotk:
> parameters is not ordered

Why do you say that? Even in the [current master](https://github.com/opesci/devito/blob/master/devito/nodes.py#L476), parameters is a tuple and hence ordered. 

 > doesn't change the fact that data is still allocated

As I said before, don't call `arguments()` and it won't be allocated. The only reason I haven't removed it from the class already is because I am waiting for your confirmation that you aren't using it any more.  This method is only meant to be called by `Operator.apply()`.

mloubout:
parameters : 

```
arguments.ScalarArgument object at 0x10c148e90>, <devito.arguments.ScalarArgument object at 0x105a20d50>, <devito.arguments.TensorArgument object at 0x10c194910>, <devito.arguments.ScalarArgument object at 0x10bf21350>, <devito.arguments.ScalarArgument object at 0x10bf21ed0>, <devito.arguments.TensorArgument object at 0x10c1946d0>, <devito.arguments.ScalarArgument object at 0x10bf21050>, <devito.arguments.TensorArgument object at 0x10c194690>, <devito.arguments.TensorArgument object at 0x10c1485d0>, <devito.arguments.ScalarArgument object at 0x10c13d3d0>, <devito.arguments.TensorArgument object at 0x10c194b10>, <devito.arguments.ScalarArgument object at 0x10bec6d10>, <devito.arguments.ScalarArgument object at 0x10bf11110>, <devito.arguments.TensorArgument object at 0x10c148650>, <devito.arguments.ScalarArgument object at 0x10bf21bd0>, <devito.arguments.ScalarArgument object at 0x10c13dd10>, <devito.arguments.TensorArgument object at 0x10c148710>)

```

Can't use that from Julia

navjotk:
```
[(x.name, x.dtype) for x in op.parameters]
```

mloubout:
Still doesn't make my life easier..... need to figure out how to change the Julia argparser with this update.

devito/operator.py
```diff
@@ -121,89 +122,62 @@ def __init__(self, expressions, **kwargs):
...
+        dle_arguments = self._dle_arguments(dim_sizes)
+        dim_sizes.update(dle_arguments)
 
+        for d, v in dim_sizes.items():
+            assert(dim_names[d].verify(v))
+        arguments = OrderedDict([(x.name, x.value) for x in self.parameters])
+        arguments.update(self._extra_arguments())
+
+        # Clear the temp values we stored in the arg objects since we've pulled them out
+        # into the OrderedDict object above
```
mloubout: Dumb question, does this means some fields are allocated multiple times?

FabioLuporini:
> This method is only meant to be called by Operator.apply().

Can we then make it private?

> [(x.name, x.dtype) for x in op.parameters]

I think what Mathias is asking is a method that does exactly what `arguments` used to do for them up until this PR. Even better, we leave ``arguments`` as is, and we change the names of our methods internally, so nothing will affect the Julia side

FabioLuporini:
> What remains here?

There still are some unanswered comments about propertising what now are class public writable fields.

devito/nodes.py
```diff
@@ -473,9 +473,21 @@ def __init__(self, name, body, retval, parameters=None, prefix=('static', 'inlin
         self.name = name
         self.body = as_tuple(body)
         self.retval = retval
-        self.parameters = as_tuple(parameters)
         self.prefix = prefix
 
+        if all(isinstance(i, ArgumentProvider) for i in parameters):
```
FabioLuporini: Can you use ``is_ArgumentProvider`` here?

devito/operator.py
```diff
@@ -121,89 +122,62 @@ def __init__(self, expressions, **kwargs):
...
-                    dim_sizes[d.name] = dim_sizes[d.parent.name]
+        for x in self.t_args + self.dims + self.s_args:
+            x.reset()
+
+    def arguments(self, *args, **kwargs):
+
+        new_params = {}
+        for k, v in kwargs.items():
+            if isinstance(v, CompositeData):
+                orig_param = [x for x in self.symbolic_data if x.name == k][0]
```
FabioLuporini: list comprehension with iteration variable `i`

FabioLuporini: also need an assert or a check, or a comment, since it seems you're sure there's at least one element

devito/operator.py
```diff
@@ -121,89 +122,62 @@ def __init__(self, expressions, **kwargs):
...
-                if dim_sizes[d.parent.name] is None:
-                    dim_sizes[d.parent.name] = dim_sizes[d.name]
-                if dim_sizes[d.name] is None:
-                    dim_sizes[d.name] = dim_sizes[d.parent.name]
+        for x in self.t_args + self.dims + self.s_args:
+            x.reset()
+
+    def arguments(self, *args, **kwargs):
+
+        new_params = {}
```
FabioLuporini: This whole block needs a comment

devito/operator.py
FabioLuporini: Can/should we not have a mechanism analogous to ``verify`` for scalar/tensors/dimensions for ``CompositeData`` objects ?

devito/operator.py
```diff
@@ -217,29 +191,9 @@ def arguments(self, *args, **kwargs):
...
-        dim_sizes.update(dle_arguments)
-
-        # Insert loop size arguments from dimension values
-        d_args = [d for d in arguments.values() if isinstance(d, Dimension)]
-        for d in d_args:
-            arguments[d.name] = dim_sizes[d.name]
-
-        # Might have been asked to auto-tune the block size
-        if maybe_autotune:
-            arguments = self._autotune(arguments)
```
FabioLuporini: Where's the autotuning logic now?

FabioLuporini:
Since this is adding some seriously important stuff, shouldn't we add some specific tests ? Should these go in ``test_operator``, or are they already there?

devito/operator.py
navjotk: They use the same mechanism as the other objects. This section only pulls out the children and adds them to kwargs. The rest of the method treats them the same as any other symbolic data arguments.

devito/nodes.py
navjotk: No, because if they are not, they would be `Argument`s and won't have this field.

devito/operator.py
navjotk: I don't see why that would happen. We just store pointers to all the arguments. If some of the argument objects haven't been allocated already, they need to be allocated before we can get a pointer. However, I don't see why that would happen multiple times.

navjotk:
> Can we then make it private?

As I have mentioned in my above comments, I want to remove this method from `OperatorForeign` and only have it in the runnable version. Waiting for @mloubout to change his code accordingly. 


> There still are some unanswered comments about propertising what now are class public writable fields.

Yes, it's been 4 days since I posted my comment. If you want to continue that discussion, can you please reply?

> I think what Mathias is asking is a method that does exactly what arguments used to do for them up until this PR

Until this PR `arguments` was doing at least three things:
1. Providing the signature for the generated C function
2. Providing default values for these arguments
3. Processing any provided values for the arguments in the form of `kwargs`, i.e. verification/validation

The reason this PR was necessary was that doing all this in one method makes that method unsustainably complex and breaking this method down into its constituents was the entire purpose of this PR. 

This necessarily means that we don't have a method that does all three any more. If he wants the signature (i.e. `parameters`), there is a method for that, and if he wants default values, there is a separate method for that (i.e. `arguments`)

FabioLuporini:
> The reason this PR was necessary was that doing all this in one method makes that method unsustainably complex and breaking this method down into its constituents was the entire purpose of this PR.

I agree. But then:



> This necessarily means that we don't have a method that does all three any more. If he wants the signature (i.e. parameters), there is a method for that, and if he wants default values, there is a separate method for that (i.e. arguments)

My point was: why don't we provide such a method in OperatorForeign (and just there) to ease their life?

FabioLuporini:
> Yes, it's been 4 days since I posted my comment. If you want to continue that discussion, can you please reply?

I think there's something broken with GitHub, because I cannot see your reply, and it seems you can't see some of my comments (did you see the question about autotuning?). Let me investigate

devito/nodes.py
navjotk: @FabioLuporini Can you not see the above?

devito/nodes.py
FabioLuporini: Ah, because they're two different hierarchies, right?

devito/arguments.py
```diff
@@ -0,0 +1,272 @@
...
+                                                shape, '%s_vec' % self.name))
+        return cast
+
+    def verify(self, value):
+        if value is None:
+            value = self._value
+
+        verify = self.source.shape == value.shape
+
+        verify = verify and all([d.verify(v) for d, v in zip(self.source.indices,
```
FabioLuporini: reminder

devito/arguments.py
```diff
@@ -0,0 +1,272 @@
...
+                                                             value.shape)])
+        if verify:
+            self._value = value
+
+        return self._value is not None and verify
+
+
+class RuntimeArgProvider(object):
+
+    """ Abstract base class for any object that, post code-generation, might resolve
```
FabioLuporini: reminder

devito/operator.py
FabioLuporini: reminder

devito/nodes.py
FabioLuporini: I only see it now ! Let me go through it

devito/operator.py
```diff
@@ -121,89 +122,62 @@ def __init__(self, expressions, **kwargs):
         # Track the DLE output, as it might be useful at execution time
         self._dle_state = dle_state
 
+        self.symbolic_data = [i for i in parameters if isinstance(i, SymbolicData)]
+        dims = [i for i in parameters if isinstance(i, Dimension)]
+
+        d_parents = [d.parent for d in dims if hasattr(d, 'parent')]
```
FabioLuporini: ``if d.is_Buffered`` ?

devito/nodes.py
navjotk: Yes

devito/operator.py
navjotk: Need not be just buffered. There's an entire class of derived `Dimension` objects that are being envisioned for the near future. Time subsampling , for example.

devito/nodes.py
FabioLuporini: I disagree, but I'm OK to leave it as is

Could you however expand the names into ``tensor_args`` and ``scalar_args`` (this definitely is convention)?

devito/operator.py
FabioLuporini: Which sounds like a subclass of Buffered ? Or, definitely, we can factor out a class characterising derived dimensions

devito/operator.py
mloubout: The rephrase, this sound like you allocate temp, copy somewhere else the deallocate.

mloubout:
>As I have mentioned in my above comments, I want to remove this method from OperatorForeign and only have it in the runnable version. Waiting for @mloubout to change his code accordingly.

@navjotk  as I told you, this doesn't remove a method from `OperatorForeign` it removes `OperatorForeign`. There is no according changes, can't use this from Julia.

mloubout:
>If he wants the signature (i.e. parameters), there is a method for that, and if he wants default values, there is a separate method for that (i.e. arguments)

default values allocates everything now, dimensions sizes are not accessible separately.

devito/operator.py
```diff
@@ -121,89 +122,62 @@ def __init__(self, expressions, **kwargs):
...
+
+        new_params = {}
+        for k, v in kwargs.items():
+            if isinstance(v, CompositeData):
+                orig_param = [x for x in self.symbolic_data if x.name == k][0]
+                for orig_child, new_child in zip(orig_param.children, v.children):
+                    new_params[orig_child.name] = new_child
+        kwargs.update(new_params)
+
+        for ta in self.t_args:
```
mloubout: breaks examples if explicit `time.size=nt`

devito/operator.py
navjotk: I haven't used the word allocate. What makes you think something is being allocated here?

navjotk:
With my commit from last night, all comments have been addressed.

devito/operator.py
```diff
@@ -128,88 +140,80 @@ def __init__(self, expressions, **kwargs):
         super(Operator, self).__init__(self.name, nodes, 'int', parameters, ())
         self.elemental_functions = elemental_functions
 
-    def arguments(self, *args, **kwargs):
+    def _reset_args(self):
+        """ Reset any runtime argument derivation information from a previous run
         """
-        Return the arguments necessary to apply the Operator.
+        for x in self.t_args + self.dims + self.s_args:
```
FabioLuporini: I can do it in my next rebase if you want

examples/seismic/tti/tti_example3D.py
```diff
@@ -57,18 +57,19 @@ def setup(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
     rec = Receiver(name='rec', ntime=nt, coordinates=receiver_coords)
 
     return AnisotropicWaveSolver(model, source=src, time_order=time_order,
-                                 space_order=space_order, receiver=rec, dse=dse)
+                                 space_order=space_order, receiver=rec, dse=dse, dle=dle,
+                                 compiler=compiler)
```
FabioLuporini: compiler ?

devito/operator.py
FabioLuporini: has this been fixed?

examples/seismic/tti/tti_example3D.py
navjotk: This was being passed to `Operator.apply()` from [here](https://github.com/opesci/devito/blob/master/examples/benchmark.py#L32) and eaten up since the previous iteration of `arguments` didn't croak on unknown `kwargs` being passed, it simply ignored them and moved on. When I added the `raise Exception()`, I saw incorrect arguments being passed in a lot of tests/examples. I simply moved this argument from the `apply` call to the constructor call.

examples/seismic/tti/tti_example3D.py
```diff
@@ -12,8 +12,8 @@ def source(t, f0):
     return np.exp(-2*s**2)*np.cos(2*np.pi*s)
 
 
-def setup(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
-          time_order=2, space_order=4, nbpml=10, dse='advanced'):
+def setup(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0, time_order=2,
+          space_order=4, nbpml=10, dse='advanced', dle='advanced', compiler=None):
```
navjotk: Same

examples/benchmark.py
```diff
@@ -67,9 +67,9 @@
     devito.add_argument("-a", "--autotune", action="store_true",
                         help=("Benchmark with auto tuning on and off. " +
                               "Enables auto tuning when execmode is run"))
-    devito.add_argument("-cb", "--cache_blocking", nargs=2, type=int,
-                        default=None, metavar=("blockDim1", "blockDim2"),
-                        help="User provided block sizes when auto-tuning is off")
+    # devito.add_argument("-cb", "--cache_blocking", nargs=2, type=int,
```
navjotk: I saw that and thought the merge between our PRs would be easier this way. I needed to remove it as well since `arguments` won't silently ignore extra arguments now - it raises Exceptions.

examples/seismic/tti/tti_example3D.py
FabioLuporini: OK, so the *right* fix is to drop ``compiler``, even from ``benchmark.py``, because that's not how it is set now

examples/benchmark.py
FabioLuporini: OK, I'll take care of that

devito/nodes.py
```diff
@@ -485,8 +485,8 @@ def __init__(self, name, body, retval, parameters=None, prefix=('static', 'inlin
 
         # At this point, all objects in args should be objects of the RuntimeArgument
         # heirarchy. Separate the tensor arguments from the scalar ones
-        self.t_args = [i for i in args if i.is_TensorArgument]
-        self.s_args = [i for i in args if i.is_ScalarArgument]
+        self.tensor_args = [i for i in args if i.is_TensorArgument]
+        self.scalar_args = [i for i in args if i.is_ScalarArgument]
```
FabioLuporini: Thanks

examples/seismic/tti/tti_example3D.py
navjotk: Unrelated to this PR

FabioLuporini: no problem, I have a stack of PRs coming, so I'll just fix it there

FabioLuporini:
@navjotk I think we can merge this once Travis is green ? 
Are we 100% sure that SLIM still works once this PR is in?

navjotk:
The test that @mloubout wrote was (inadvertently?) merged in master already and this code does pass it. Whether that is a sufficient condition for SLIM to be fully functional, only @mloubout knows.

FabioLuporini:
So we'll wait for you, @mloubout

FabioLuporini:
OK, the time has come. Thanks @navjotk for this massive contribution and for being patient with me and Mathias

examples/seismic/tti/tti_example3D.py
mloubout: >compiler=compiler

I removed it at some point as it creates problem when default is `compiler=None`

FabioLuporini:
Merged.

PR #290, opened 2017-07-05
Test emptiness of array object

mloubout:
Extra small test. The main point of `OperatorForeign' is to bypass devito memory allocation for 'DenseData' objects. This new test makes sure of it.

I do know that this is not the main aim of Devito but this is necessary for the Julia interface to avoid duplicate allocation, specially for large object such as the full wavefield for the gradient.

### DIFF
--- a/tests/test_operator.py
+++ b/tests/test_operator.py
@@ -446,3 +446,12 @@ class TestForeign(object):
         arg, _ = op.arguments(a=b)
         op2.cfunction(*list(arg.values()))
         assert(np.allclose(a.data[:], b.data[:]))
+
+    def test_empty_arrays(self):
+        shape = (11, 11)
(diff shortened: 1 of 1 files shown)


FabioLuporini:
OK, thanks

FabioLuporini:
If you have an extra minute to spare, I'd come up with a better name for the test

mloubout:
@navjotk as long as #289 does the right thing, fine with me. I'll leave that open for now as a reminder.

navjotk:
This test is already in master https://github.com/opesci/devito/blob/master/tests/test_operator.py#L448

Could this PR be closed?

mloubout:
Done in @FabioLuporini PR. CLosing.

PR #292, opened 2017-07-06
Force sympy 1.0

mloubout:
This PR is not a solution but a hack so that people don't have problem installing devito while we figure out how to handle sympy 1.1

Only change the setup to force 1.0

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,6 +20,6 @@ setup(name='devito',
       author_email='opesci@imperial.ac.uk',
       license='MIT',
       packages=find_packages(exclude=['docs', 'examples', 'tests']),
-      install_requires=['numpy', 'sympy', 'mpmath', 'cgen', 'codepy',
+      install_requires=['numpy', 'sympy==1.0', 'mpmath', 'cgen', 'codepy',
                         'psutil', 'py-cpuinfo'],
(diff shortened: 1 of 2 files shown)


mloubout:
Oups, TRAVIS uses requirement not setup, will fix it.

At least TRAVIS shows the problem with `sympy 1.1`

setup.py
```diff
@@ -20,6 +20,6 @@
       author_email='opesci@imperial.ac.uk',
       license='MIT',
       packages=find_packages(exclude=['docs', 'examples', 'tests']),
-      install_requires=['numpy', 'sympy', 'mpmath', 'cgen', 'codepy',
+      install_requires=['numpy', 'sympy==1.0', 'mpmath', 'cgen', 'codepy',
```
FabioLuporini: Shouldn't this list match that in requirements.txt? Or can we just drop this @mlange05

FabioLuporini:
OK. @mlange05 I think Mathias is right and we should merge this. We should also add an issue though

setup.py
mloubout: Also is in the requirements

FabioLuporini:
OK, merging...

FabioLuporini:
Merged.

PR #293, opened 2017-07-11
CSRE finalization

FabioLuporini:
Final tweaks making the Cross-Stencil Redundancies Elimination algorithm fully operative, but still not automatically enabled yet. For that, still need one more set of changes.

### DIFF
--- a/devito/core/autotuning.py
+++ b/devito/core/autotuning.py
@@ -1,9 +1,14 @@
+from __future__ import absolute_import
+
 from collections import OrderedDict
 from itertools import combinations
+from functools import reduce
+from operator import mul
+import resource
 
 from devito.logger import info, info_at
 from devito.nodes import Iteration
-from devito.visitors import FindNodes
+from devito.visitors import FindNodes, FindSymbols
 
 __all__ = ['autotune']
 
@@ -24,6 +29,7 @@ def autotune(operator, arguments, tunable, mode='basic'):
 
     # Squeeze dimensions to minimize auto-tuning time
     iterations = FindNodes(Iteration).visit(operator.body)
+    dim_mapper = {i.dim.name: i.dim for i in iterations}
     squeezable = [i.dim.parent.name for i in iterations
                   if i.is_Sequential and i.dim.is_Buffered]
 
@@ -34,14 +40,21 @@ def autotune(operator, arguments, tunable, mode='basic'):
     if mode == 'aggressive':
         blocksizes = more_heuristic_attempts(blocksizes)
 
+    # How many temporaries are allocated on the stack?
+    # Will drop block sizes that might lead to a stack overflow
+    functions = FindSymbols('symbolics').visit(operator.body +
+                                               operator.elemental_functions)
+    stack_shapes = [i.shape for i in functions if i.is_TensorFunction and i._mem_stack]
+    stack_space = sum(reduce(mul, i, 1) for i in stack_shapes)*operator.dtype().itemsize
+
     # Note: there is only a single loop over 'blocksize' because only
     # square blocks are tested
     timings = OrderedDict()
-    for blocksize in blocksizes:
+    for bs in blocksizes:
         illegal = False
         for k, v in at_arguments.items():
-            if k in blocksize:
-                val = blocksize[k]
+            if k in bs:
+                val = bs[k]
                 handle = at_arguments.get(mapper[k].original_dim.name)
                 if val <= mapper[k].iteration.end(handle):
                     at_arguments[k] = val
@@ -54,17 +67,39 @@ def autotune(operator, arguments, tunable, mode='basic'):
         if illegal:
             continue
 
+        # Make sure we remain within stack bounds, otherwise skip block size
+        dim_sizes = {}
+        for k, v in at_arguments.items():
+            if k in bs:
+                dim_sizes[mapper[k].argument.symbolic_size] = bs[k]
+            elif k in dim_mapper:
+                dim_sizes[dim_mapper[k].symbolic_size] = v
+        try:
+            bs_stack_space = stack_space.xreplace(dim_sizes)
+        except AttributeError:
+            bs_stack_space = stack_space
+        try:
+            if int(bs_stack_space) > options['at_stack_limit']:
+                continue
+        except TypeError:
+            # We should never get here
+            info_at("Couldn't determine stack size, skipping block size %s" % str(bs))
+            continue
+
         # Add profiler structs
         at_arguments.update(operator._extra_arguments())
 
         operator.cfunction(*list(at_arguments.values()))
         elapsed = sum(operator.profiler.timings.values())
-        timings[tuple(blocksize.items())] = elapsed
-        info_at("<%s>: %f" %
-                (','.join('%d' % i for i in blocksize.values()), elapsed))
+        timings[tuple(bs.items())] = elapsed
+        info_at("<%s>: %f" % (','.join('%d' % i for i in bs.values()), elapsed))
 
-    best = dict(min(timings, key=timings.get))
-    info('Auto-tuned block shape: %s' % best)
+    try:
+        best = dict(min(timings, key=timings.get))
+        info("Auto-tuned block shape: %s" % best)
+    except ValueError:
+        info("Auto-tuning request, but couldn't find legal block sizes")
+        return arguments
 
     # Build the new argument list
     tuned = OrderedDict()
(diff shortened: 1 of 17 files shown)


devito/nodes.py
```diff
@@ -350,18 +358,28 @@ def ccode(self):
...
 
+        # Append unbounded indices, if any
+        if self.uindices:
+            uinit = ['%s = %s' % (i.index, ccode(i.start)) for i in self.uindices]
+            loop_init = c.Line(', '.join([loop_init] + uinit))
+            ustep = ['%s = %s' % (i.index, ccode(i.step)) for i in self.uindices]
+            loop_inc = c.Line(', '.join([loop_inc] + ustep))
+
+        # Create For header+body
         handle = c.For(loop_init, loop_cond, loop_inc, c.Block(loop_body))
```
mloubout: isn't `loop_cond` undefined for `uindices` only?

devito/nodes.py
```diff
@@ -350,18 +358,28 @@ def ccode(self):
...
             loop_cond = '%s >= %s' % (self.index, ccode(start))
             loop_inc = '%s -= %s' % (self.index, self.limits[2])
         else:
-            loop_init = c.InlineInitializer(c.Value("int", self.index), ccode(start))
+            loop_init = 'int %s = %s' % (self.index, ccode(start))
             loop_cond = '%s < %s' % (self.index, ccode(end))
             loop_inc = '%s += %s' % (self.index, self.limits[2])
 
+        # Append unbounded indices, if any
+        if self.uindices:
```
mloubout: What are these? Do  you have a example (curiosity).

FabioLuporini: Unbounded indices (see also the docstrings of `Iteration`  and `UnboundedIndex`) are special indices that "belong" to an iteration, but are used to traverse a Dimension in a non-linear fashion; that is, instead of increasing in ``+= 1`` steps, their value can change at each iteration in any way (e.g., using ``%``, ``+= k``, etc).

An UnboundedIndex is still related to an Iteration though, because for example it derives from a ``BufferedDimension`` or any future (subclass of) ``Dimension`` object that we plan to introduce; hence, they're now generated in the loop header directly, instead of "floating at random places within the code".

Example:
Previously we had something like
```
for (t = 0; t < N; t++) {
  t1 = t0 % 2;
}
```

Now instead we generate:
```
for (t = 0, t1 = t0 % 2; t < N; t++, t1 = t0 % 2) {
}
```

To make it simple, this simple change makes compiler analysis simpler and faster. Functionally, there is no difference.

devito/nodes.py
FabioLuporini: An UnboundedIndex has no stop condition; in fact, it's unbounded, so no need to change ``loop_cond`` based on the values of the ``uindices``

FabioLuporini:
> Simple questions,

Appreciate this

devito/nodes.py
```diff
@@ -592,3 +633,67 @@ def __init__(self, expr, dtype):
...
+def ntags():
+    return len(known_properties) - ntags.n_original_properties
+ntags.n_original_properties = len(known_properties)  # noqa
+
+
+class UnboundedIndex(object):
+
+    """
+    A generic loop iteration index used to traverse an iteration space in a
+    non-linear way (e.g., non-unitary increment).
```
navjotk: Since `Iteration` has a built-in way to have non-unitary increments, ~I think~ having that as the example here is confusing ~at the very least~ if not wrong (non-unitary != non-linear).

devito/visitors.py
```diff
@@ -201,7 +200,8 @@ def visit_Iteration(self, o):
         self._depth -= 1
         if self.verbose:
             detail = '::%s::%s::%s' % (o.index, o.limits, o.offsets)
-            props = '[%s] ' % ','.join(o.properties) if o.properties else ''
+            props = [str(i) for i in o.properties]
```
navjotk: Convention reminder: Expand variable name

devito/core/autotuning.py
```diff
@@ -24,6 +29,7 @@ def autotune(operator, arguments, tunable, mode='basic'):
 
     # Squeeze dimensions to minimize auto-tuning time
     iterations = FindNodes(Iteration).visit(operator.body)
+    dim_mapper = {i.dim.name: i.dim for i in iterations}
```
FabioLuporini: yeah..any suggestion on how to do this?

devito/core/autotuning.py
```diff
@@ -34,14 +40,21 @@ def autotune(operator, arguments, tunable, mode='basic'):
...
+    functions = FindSymbols('symbolics').visit(operator.body +
+                                               operator.elemental_functions)
+    stack_shapes = [i.shape for i in functions if i.is_TensorFunction and i._mem_stack]
+    stack_space = sum(reduce(mul, i, 1) for i in stack_shapes)*operator.dtype().itemsize
+
     # Note: there is only a single loop over 'blocksize' because only
     # square blocks are tested
     timings = OrderedDict()
-    for blocksize in blocksizes:
+    for bs in blocksizes:
```
FabioLuporini: Ah, for loop iteration variables, I accepted abbreviations or simple indices such as ``i``

devito/dle/backends/basic.py
```diff
@@ -35,75 +40,84 @@ def _create_elemental_functions(self, state, **kwargs):
...
-                # Add scalar arguments
-                handle = filter_sorted(required - seen, key=attrgetter('name'))
-                args.extend([(i.name, ScalarFunction(name=i.name, dtype=np.int32))
-                             for i in handle])
 
-                # Track info to transform the main tree
-                call, parameters = zip(*args)
-                mapper[view] = (List(header=noinline, body=FunCall(name, call)), [root])
+                # Retrieve scalar arguments
+                not_required.update({e.output for e in expressions if e.is_scalar})
```
FabioLuporini: Yes

devito/dle/blocking_utils.py
```diff
@@ -134,43 +139,46 @@ def optimize_unfolded_tree(unfolded, root):
     """
     processed = []
     for i, tree in enumerate(unfolded):
+        assert len(tree) == len(root)
         otree = []
-        stmts = []
+        oroot = []
```
FabioLuporini: I wanted to, but I start breaking too many lines into two if I expand these names, so I preferred the contracted form. I'll see if I can rethink them though, because you're right, ``stmts`` is quite obvious, but ``oroot`` is definitely not

devito/dle/blocking_utils.py
```diff
@@ -134,43 +139,46 @@ def optimize_unfolded_tree(unfolded, root):
...
-                shape += j.output_function.shape[len(otree):]
-                j.output_function.update(shape=shape, onstack=True)
+            for j in exprs:
+                j_shape = shape + j.output_function.shape[len(otree):]
+                j.output_function.update(shape=j_shape, onstack=True)
+
+        # Substitute iteration variables within the folded trees
+        otree = compose_nodes(otree)
+        replaced = xreplace_indices([j.expr for j in exprs], mapper, only_rhs=True)
+        subs = [j._rebuild(expr=k) for j, k in zip(exprs, replaced)]
```
FabioLuporini: It's always been that way, that's the only main reason. The reason was that ``_rebuild`` isn't an action or a property of an Iteration/Expression tree node, but rather a feature that all nodes of the hierarchy have, so I was simply trying to distinguish that. Might change in a future cleanup PR

devito/nodes.py
FabioLuporini: Good catch, I'll fix the docstring

devito/core/autotuning.py
FabioLuporini: I think the right thing to do is to change the test so that we intercept ``autotune=True`` and we call the autotuner explicitly from there. In fact, the test in ``test_dle.py`` is supposed to test the autotuning feature, not if it's called...
Then from ``test_operator.py`` we could test *if* the autotuner gets called -- but how ? Anyway, not part of this PR, but something to keep in mind (suggestions welcome)

FabioLuporini:
@navjotk Updated

devito/core/autotuning.py
navjotk: Yeah as I mentioned in my first comment, when I forgot to call the autotuner from `arguments`, all tests still passed so I am only pointing out that there needs to be a test to catch that.

devito/core/autotuning.py
navjotk: You didn't accept anything except `i`, so maybe change this to `i`?

FabioLuporini: Yep, I will. BTW, to clarify what I wrote above, "for loop iteration variables" -- in particular, for list comprehensions

mloubout:
something I just noticed in `autotuning.py` : 

```
 at_arguments = arguments.copy()
```

Deep copy of wavefields is not an option anywhere, thos should not be there.

FabioLuporini:
> Deep copy

It's not a deep copy. However, a "clean" field must be used prior to an AT run. This takes a negligible amount of time as you can verify by yourself trying it out.

If we move AT to runtime, than even this overhead will disappear

FabioLuporini:
OK, merging...

FabioLuporini:
Merged.

PR #294, opened 2017-07-11
Make Conda install the preferred way to install

mlange05:
Since we increasingly are adding strong version dependencies, we should encourage users to actually use environment managers when installing Devito. So, this PR makes Conda the preferred method by adding an explicit `environment.yml` file, updating Travis to use miniconda and updating the README accordingly.

This PR also adds a description of the tutorials to the README and removes the outdated `memmap_manager` entry from the docs to fix master breakages. It also requires a small tolerance relaxation due to failures with other numpy versions (I suspect).


README.md
```diff
@@ -14,30 +14,47 @@ detailed [API documentation](http://www.opesci.org/devito).
...
+pip install -e .
 ```
 
-Alternatively Devito can be be installed manually from GitHub via:
+If you don't want to use the Conda environment manager, Devito can
+also be installed directly from GitHub via pip:
 ```
-git clone https://github.com/opesci/devito.git
-cd devito && pip install --user -r requirements.txt
+pip install --user git+https://github.com/opesci/devito.git
```
FabioLuporini: But we still need to run through the ``requirements.txt``, don't we?

README.md
```diff
@@ -14,30 +14,47 @@ detailed [API documentation](http://www.opesci.org/devito).
...
+
+To install the editable development version Devito, including examples,
+tests and tutorial notebooks, please run the following commands:
 ```
-pip install --user git+https://github.com/opesci/devito.git
+git clone https://github.com/opesci/devito.git
+cd devito
+conda env create -f environment.yml
+source activate devito
+pip install -e .
```
FabioLuporini: How difficult is it to automate the Conda installation and these instructions in an ``install_devito.sh`` script?

mlange05: The whole point of using Conda is to not have to maintain a separate install script. Except for maybe the final self-install to get the PYTHONPATH right, all of this is boilerplate that you would probably have to do with an install script as well. The magic really is in the new `environment.yml` which gives a complete description of the env we want - and I believe you can expand that with a certain amount of customization at a later stage.

README.md
mlange05: Not necessarily, since `requirements.txt` is more of a "suggestion for a tested env configuration", while all the hard dependencies are in `setup.py`. The problem with the pure pip install is that our `requirements.txt` gets buried somewhere in the local Python install dir, so this is not that easy to predict or automate.

mlange05:
Ok, merging...

PR #295, opened 2017-07-13
DSE restructuring

FabioLuporini:
This should go in AFTER #293 

I'm just restructuring things in the DSE a bit to make everything cleaner and easier to extend

PR #296, opened 2017-07-14
CSRE enhancements

FabioLuporini:
This PR adds some features on top of #295

### DIFF
--- a/examples/benchmark.py
+++ b/examples/benchmark.py
@@ -52,8 +52,8 @@ if __name__ == "__main__":
 
     devito = parser.add_argument_group("Devito")
     devito.add_argument("-dse", default="advanced", nargs="*",
-                        choices=["noop", "basic", "factorize", "approx-trigonometry",
-                                 "glicm", "advanced", "speculative"],
+                        choices=["noop", "basic", "advanced", "speculative",
+                                 "aggressive"],
                         help="Devito symbolic engine (DSE) mode")
     devito.add_argument("-dle", default="advanced", nargs="*",
                         choices=["noop", "advanced", "speculative"],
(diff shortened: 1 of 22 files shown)


FabioLuporini:
Now rebased on master. 

Comments more than welcome

FabioLuporini:
Some more fixes added

FabioLuporini:
And now with stronger dle analysis and more tests

FabioLuporini:
Excellent, Travis green.

Thoughts @mlange05 @navjotk  ?

tests/test_adjointA.py
```diff
@@ -29,4 +29,4 @@ def test_acoustic(dimensions, time_order, space_order, fix_dim):
     term1 = np.dot(srca.data.reshape(-1), solver.source.data)
     term2 = linalg.norm(rec.data) ** 2
     print(term1, term2, ("%12.12f") % (term1 - term2), term1 / term2)
-    assert np.isclose(term1 / term2, 1.0, atol=0.001)
```
mlange05: What's happening here? Are we dropping accuracy somewhere?

examples/benchmark.py
```diff
@@ -52,8 +52,8 @@
 
     devito = parser.add_argument_group("Devito")
     devito.add_argument("-dse", default="advanced", nargs="*",
-                        choices=["noop", "basic", "factorize", "approx-trigonometry",
-                                 "glicm", "advanced", "speculative"],
+                        choices=["noop", "basic", "advanced", "speculative",
+                                 "aggressive"],
```
mlange05: I realise these are in flux, but we will need some serious developer documentation for all these modes soon. Just a reminder though, not as part of this PR.

tests/test_adjointA.py
FabioLuporini: It's because the receiver loops are now parallelised

examples/benchmark.py
FabioLuporini: > Also, as mentioned, some developer-level docs are becoming more and more necessary to follow the flow of these optimisations. I know, low priority, but still important to keep in mind.

Yes, I'm planning to write it up while also writing the paper

tests/test_adjointA.py
mlange05: That still seems like a lot. @mloubout can we live with this?

FabioLuporini: OK, so, I double checked and you're right, it turns out this change was actually unnecessary so I removed it.

The reason it was there was briefly discussed on Slack, when I was talking about parallelising the sources loop (not the receiver loops). 

So this is not a problem any more

mlange05: Excellent. Green light from me then. :smile:

FabioLuporini:
@mlange05 OK, I've just rebased it on master, so I'll merge it as soon as Travis confirms his happiness

FabioLuporini:
Merging...

FabioLuporini:
Merged.

PR #297, opened 2017-07-17
Constant velocity doesn't work anymore

mloubout:
More of an issue, but easier to see the diff of where it breaks here.

Added the constant example to TRAVIS

The new `argument` functions now fails with Invalid argument for the constant example.

```
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/devito/operator.py", line 191, in arguments
    raise InvalidArgument("Unknown arguments passed: " + ", ".join(kwargs.keys()))
devito.exceptions.InvalidArgument: Unknown arguments passed: dle, m, dse
```

### DIFF
--- a/devito/operator.py
+++ b/devito/operator.py
@@ -182,6 +182,7 @@ class Operator(Function):
         autotune = autotune and kwargs.pop('autotune', False)
 
         # Make sure we've used all arguments passed
+        # Breaks for constant velocity
         if len(kwargs) > 0:
             raise InvalidArgument("Unknown arguments passed: " + ", ".join(kwargs.keys()))
 
--- a/devito/operator.py
+++ b/devito/operator.py
@@ -182,7 +182,6 @@ class Operator(Function):
         autotune = autotune and kwargs.pop('autotune', False)
 
         # Make sure we've used all arguments passed
-        # Breaks for constant velocity
         if len(kwargs) > 0:
             raise InvalidArgument("Unknown arguments passed: " + ", ".join(kwargs.keys()))
 
--- a/examples/seismic/acoustic/acoustic_example.py
+++ b/examples/seismic/acoustic/acoustic_example.py
@@ -99,4 +99,4 @@ def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
 
 
 if __name__ == "__main__":
-    run(full_run=False, autotune=False, space_order=6, time_order=2)
+    run(full_run=True, autotune=False, space_order=6, time_order=2)
--- a/examples/seismic/acoustic/constant_example.py
+++ b/examples/seismic/acoustic/constant_example.py
@@ -85,7 +85,7 @@ def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
 
     initial_vp = 1.8
     dm = (initial_vp**2 - solver.model.m.data) * np.ones(solver.model.shape_domain,
-                                                          dtype=np.float32)
+                                                         dtype=np.float32)
     info("Applying Forward")
     m0 = ConstantData(name="m", value=.25, dtype=np.float32)
     rec, u, summary = solver.forward(save=full_run)
--- a/examples/seismic/acoustic/constant_example.py
+++ b/examples/seismic/acoustic/constant_example.py
@@ -79,13 +79,14 @@ def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
         time_order=2, space_order=4, nbpml=40, full_run=False, **kwargs):
 
     solver = setup(dimensions=dimensions, spacing=spacing, nbpml=nbpml, tn=tn,
-                   space_order=space_order, time_order=time_order, **kwargs)
+                   space_order=space_order, time_order=time_order,
+                   dse='noop', dle='noop', **kwargs)
 
     initial_vp = 1.8
-    dm = (initial_vp**2 - solver.model.m) * np.ones(solver.model.shape_domain,
-                                                    dtype=np.float32)
+    dm = (initial_vp**2 - solver.model.m.value) * np.ones(solver.model.shape_domain,
+                                                          dtype=np.float32)
     info("Applying Forward")
-    rec, u, summary = solver.forward(save=full_run)
+    rec, u, summary = solver.forward(save=full_run, m=.25)
 
     if not full_run:
         return summary.gflopss, summary.oi, summary.timings, [rec, u.data]
@@ -99,4 +100,5 @@ def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
 
 
 if __name__ == "__main__":
-    run(full_run=True, autotune=False, space_order=6, time_order=2)
+    run(full_run=True, autotune=False, space_order=6, time_order=2,
+        dimensions=(50, 50), tn=500.)
--- a/examples/seismic/acoustic/constant_example.py
+++ b/examples/seismic/acoustic/constant_example.py
@@ -55,7 +55,12 @@ def setup(dimensions=(50, 50, 50), spacing=(15.0, 15.0, 15.0), tn=500.,
     # Define source and receivers and create acoustic wave solver
     src = PointSource(name='src', data=time_series, coordinates=location)
     rec = Receiver(name='rec', ntime=nt, coordinates=receiver_coords)
-    return AcousticWaveSolver(model, source=src, receiver=rec,
+
+    # Note: A potential bug in GCC requires using less aggressive loop
+    # transformations. In particular, the DLE `_avoid_denormals` pass is skipped
+    # by using `advanced-safemath`, rather than the default DLE mode `advanced`.
+
+    return AcousticWaveSolver(model, source=src, receiver=rec, dle='advanced-safemath',
                               time_order=time_order, space_order=space_order, **kwargs)
 
 
--- a/examples/seismic/acoustic/constant_example.py
+++ b/examples/seismic/acoustic/constant_example.py
@@ -6,21 +6,6 @@ from examples.seismic import Model, PointSource, Receiver
 from devito import ConstantData
 
 
-# Velocity models
-def smooth10(vel, shape):
-    out = np.ones(shape, dtype=np.float32)
-    out[:] = vel[:]
-    nz = shape[-1]
-
-    for a in range(5, nz-6):
-        if len(shape) == 2:
-            out[:, a] = np.sum(vel[:, a - 5:a + 5], axis=1) / 10
-        else:
-            out[:, :, a] = np.sum(vel[:, :, a - 5:a + 5], axis=2) / 10
-
-    return out
-
-
 # Set up the source as Ricker wavelet for f0
 def source(t, f0):
     r = (np.pi * f0 * (t - 1./f0))
@@ -56,7 +41,6 @@ def setup(dimensions=(50, 50, 50), spacing=(15.0, 15.0, 15.0), tn=500.,
     if time_order == 4:
         dt *= 1.73
     t0 = 0.0
-    tn = tn
     nt = int(1+(tn-t0)/dt)
 
     # Set up the source as Ricker wavelet for f0
@@ -72,25 +56,26 @@ def setup(dimensions=(50, 50, 50), spacing=(15.0, 15.0, 15.0), tn=500.,
     src = PointSource(name='src', data=time_series, coordinates=location)
     rec = Receiver(name='rec', ntime=nt, coordinates=receiver_coords)
     return AcousticWaveSolver(model, source=src, receiver=rec,
-                              time_order=time_order, space_order=space_order,
-                              **kwargs)
+                              time_order=time_order, space_order=space_order, **kwargs)
 
 
 def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
         time_order=2, space_order=4, nbpml=40, full_run=False, **kwargs):
 
     solver = setup(dimensions=dimensions, spacing=spacing, nbpml=nbpml, tn=tn,
-                   space_order=space_order, time_order=time_order,
-                   **kwargs)
+                   space_order=space_order, time_order=time_order, **kwargs)
 
     initial_vp = 1.8
     dm = (initial_vp**2 - solver.model.m.data) * np.ones(solver.model.shape_domain,
                                                          dtype=np.float32)
     info("Applying Forward")
-    m0 = ConstantData(name="m", value=.25, dtype=np.float32)
+    # Default model.m
     rec, u, summary = solver.forward(save=full_run)
-    rec, u, summary = solver.forward(save=full_run, m=m0)
-    rec, u, summary = solver.forward(save=full_run, m=.25)
+    # With  a new m as ConstantData
+    m0 = ConstantData(name="m", value=.25, dtype=np.float32)
+    solver.forward(save=full_run, m=m0)
+    # With a new m as a scalar value
+    solver.forward(save=full_run, m=.25)
 
     if not full_run:
         return summary.gflopss, summary.oi, summary.timings, [rec, u.data]
@@ -105,4 +90,4 @@ def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
 
 if __name__ == "__main__":
     run(full_run=True, autotune=False, space_order=6, time_order=2,
-        dimensions=(50, 50), tn=500.)
+        dimensions=(50, 50), tn=500., nbpml=0)
--- a/examples/seismic/acoustic/constant_example.py
+++ b/examples/seismic/acoustic/constant_example.py
@@ -3,6 +3,7 @@ import numpy as np
 from devito.logger import info
 from examples.seismic.acoustic import AcousticWaveSolver
 from examples.seismic import Model, PointSource, Receiver
+from devito import ConstantData
 
 
 # Velocity models
@@ -80,12 +81,15 @@ def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
 
     solver = setup(dimensions=dimensions, spacing=spacing, nbpml=nbpml, tn=tn,
                    space_order=space_order, time_order=time_order,
-                   dse='noop', dle='noop', **kwargs)
+                   **kwargs)
 
     initial_vp = 1.8
-    dm = (initial_vp**2 - solver.model.m.value) * np.ones(solver.model.shape_domain,
+    dm = (initial_vp**2 - solver.model.m.data) * np.ones(solver.model.shape_domain,
                                                           dtype=np.float32)
     info("Applying Forward")
+    m0 = ConstantData(name="m", value=.25, dtype=np.float32)
+    rec, u, summary = solver.forward(save=full_run)
+    rec, u, summary = solver.forward(save=full_run, m=m0)
     rec, u, summary = solver.forward(save=full_run, m=.25)
 
     if not full_run:
--- a/examples/seismic/acoustic/constant_example.py
+++ b/examples/seismic/acoustic/constant_example.py
@@ -1,18 +1,21 @@
 import numpy as np
 
 from devito.logger import info
-from examples.seismic.acoustic.wavesolver import AcousticWaveSolver
+from examples.seismic.acoustic import AcousticWaveSolver
 from examples.seismic import Model, PointSource, Receiver
 
 
 # Velocity models
 def smooth10(vel, shape):
-    out = np.ones(shape)
-    out[:, :] = vel[:, :]
-    nx = shape[0]
+    out = np.ones(shape, dtype=np.float32)
+    out[:] = vel[:]
+    nz = shape[-1]
 
-    for a in range(5, nx-6):
-        out[a, :] = np.sum(vel[a - 5:a + 5, :], axis=0) / 10
+    for a in range(5, nz-6):
+        if len(shape) == 2:
+            out[:, a] = np.sum(vel[:, a - 5:a + 5], axis=1) / 10
+        else:
+            out[:, :, a] = np.sum(vel[:, :, a - 5:a + 5], axis=2) / 10
 
     return out
 
@@ -24,36 +27,14 @@ def source(t, f0):
     return (1-2.*r**2)*np.exp(-r**2)
 
 
-def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
-        time_order=2, space_order=4, nbpml=40, dse='advanced', dle='advanced',
-        full_run=False):
+def setup(dimensions=(50, 50, 50), spacing=(15.0, 15.0, 15.0), tn=500.,
+          time_order=2, space_order=4, nbpml=10, vp=1.5, **kwargs):
+
     ndim = len(dimensions)
     origin = tuple([0.]*ndim)
     spacing = spacing[:ndim]
-
-    # True velocity
-    true_vp = 2.
-
-    # Smooth velocity
-    initial_vp = 1.8
-
-    dm = 1. / (true_vp * true_vp) - 1. / (initial_vp * initial_vp)
-    model = Model(origin, spacing, dimensions, true_vp, nbpml=nbpml)
-    dm = np.ones(model.shape_domain, dtype=np.float32)*dm
-
-    # Define seismic data.
-    f0 = .010
-    dt = model.critical_dt
-    if time_order == 4:
-        dt *= 1.73
-    t0 = 0.0
-    nt = int(1+(tn-t0)/dt)
-
-    # Source geometry
-    time_series = np.zeros((nt, 1))
-
-    time_series[:, 0] = source(np.linspace(t0, tn, nt), f0)
-
+    # Velocity model, two layers
+    true_vp = vp
     # Source location
     location = np.zeros((1, ndim), dtype=np.float32)
     location[0, :-1] = [origin[i] + dimensions[i] * spacing[i] * .5
@@ -65,25 +46,57 @@ def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
                                         (dimensions[0]-1) * spacing[0],
                                         num=dimensions[0])
     receiver_coords[:, 1:] = location[0, 1:]
+
+    # Define seismic data
+    model = Model(origin, spacing, dimensions, true_vp, nbpml=int(nbpml))
+
+    f0 = .010
+    dt = model.critical_dt
+    if time_order == 4:
+        dt *= 1.73
+    t0 = 0.0
+    tn = tn
+    nt = int(1+(tn-t0)/dt)
+
+    # Set up the source as Ricker wavelet for f0
+    def source(t, f0):
+        r = (np.pi * f0 * (t - 1./f0))
+        return (1-2.*r**2)*np.exp(-r**2)
+
+    # Source geometry
+    time_series = np.zeros((nt, 1), dtype=np.float32)
+    time_series[:, 0] = source(np.linspace(t0, tn, nt), f0)
+
+    # Define source and receivers and create acoustic wave solver
     src = PointSource(name='src', data=time_series, coordinates=location)
     rec = Receiver(name='rec', ntime=nt, coordinates=receiver_coords)
+    return AcousticWaveSolver(model, source=src, receiver=rec,
+                              time_order=time_order, space_order=space_order,
+                              **kwargs)
 
-    solver = AcousticWaveSolver(model, source=src, receiver=rec,
-                                time_order=time_order, space_order=space_order)
 
+def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
+        time_order=2, space_order=4, nbpml=40, full_run=False, **kwargs):
+
+    solver = setup(dimensions=dimensions, spacing=spacing, nbpml=nbpml, tn=tn,
+                   space_order=space_order, time_order=time_order, **kwargs)
+
+    initial_vp = 1.8
+    dm = (initial_vp**2 - solver.model.m) * np.ones(solver.model.shape_domain,
+                                                    dtype=np.float32)
     info("Applying Forward")
-    rec, u, summary = solver.forward(save=full_run, dse=dse, dle=dle)
+    rec, u, summary = solver.forward(save=full_run)
 
     if not full_run:
         return summary.gflopss, summary.oi, summary.timings, [rec, u.data]
 
     info("Applying Adjoint")
-    solver.adjoint(rec, dse=dse, dle=dle)
+    solver.adjoint(rec, **kwargs)
     info("Applying Born")
-    solver.born(dm, dse=dse, dle=dle)
+    solver.born(dm, **kwargs)
     info("Applying Gradient")
-    solver.gradient(rec, u, dse=dse, dle=dle)
+    solver.gradient(rec, u, **kwargs)
 
 
 if __name__ == "__main__":
-    run(full_run=True, space_order=6, time_order=2)
+    run(full_run=True, autotune=False, space_order=6, time_order=2)
--- a/examples/seismic/acoustic/operators.py
+++ b/examples/seismic/acoustic/operators.py
@@ -162,7 +162,7 @@ def GradientOperator(model, source, receiver, time_order=2, space_order=4, **kwa
     subs = dict([(t.spacing, dt)] + [(time.spacing, dt)] +
                 [(i.spacing, model.get_spacing()[j]) for i, j
                  in zip(v.indices[1:], range(len(model.shape)))])
-    return Operator([eqn] + [gradient_update] + receivers,
+    return Operator([eqn] + receivers + [gradient_update],
                     subs=subs,
                     time_axis=Backward, name='Gradient', **kwargs)
 
--- a/examples/seismic/acoustic/operators.py
+++ b/examples/seismic/acoustic/operators.py
@@ -40,9 +40,9 @@ def ForwardOperator(model, source, receiver, time_order=2, space_order=4,
 
     s = t.spacing
 
-    stencil = 1 / (2 * m + s * damp) * (
-        4 * m * u + (s * damp - 2 * m) * u.backward +
-        2 * s**2 * (u.laplace + s**2 / 12 * biharmonic))
+    stencil = 1.0 / (2.0 * m + s * damp) * (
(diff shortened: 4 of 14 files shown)


navjotk:
As I mentioned before, this is because incorrect arguments were being passed to `apply` before and being ignored while not throwing any error. 

The error message shows that `dle, m, dse` are the three unrecognised arguments and that makes sense to me.

FabioLuporini:
so let's fix the example ?

navjotk:
Unable to execute anything from master on my machine

mloubout:
>so let's fix the example ?

The example is how it should be

navjotk:
> The example is how it should be

No it isn't. `Operator.apply()` does not take arguments like `dse=...` or `dle=...` but you are passing them [here](https://github.com/opesci/devito/blob/master/examples/seismic/acoustic/constant_example.py#L74). Which is what the error is telling you.

FabioLuporini:
Can any of you fix the example?

mloubout:
dse/dle removed, still problem with m as it is not an input for constant velocity. Not sure how to do it without if/else everywhere

FabioLuporini:
News here?

mloubout:
Still working on it. The isssue is actually broader. The problem is, how to make Devito aware of constants changes and regenerate accordingly (spaing, dt, ...)

FabioLuporini:
14 days passed and we still have a broken example

mloubout:
Ok, so after trying a lot of different things.

As it is now, this constant are hard coded (which is not a bad thing) but forgotten by devito once the operator is cached as a property. This leads to two problems
1) Devito doesn't recognize `m` as a parameter anymore for constant velocity. Tried a lot of thing, couldn't male it work so far.
2) If something changes, such as `dt` or any spacing, Devito would completely ignores it too.

I suppose the idea was to recreate a new operator for each change an only cache for  multiple source locations.

I'd say we need an new class `ConstantData` (name open to debate) to handle constant parameters, but I couldn't implement one ( I guess I still don't quite get what's going on in interface anymore with Functions, Tensor, Data). 

I'm open to ideas, completely stuck now.

devito/operator.py
```diff
@@ -280,6 +283,7 @@ def cfunction(self):
         if self._cfunction is None:
             self._cfunction = getattr(self._lib, self.name)
             argtypes = [c_int if isinstance(v, ScalarArgument) else
+                        c_float if isinstance(v, ConstantArgument) else
```
FabioLuporini: Just a reminder that here we're hardcoding assumptions about the data types of ScalarArgument and ConstantArgument

examples/seismic/acoustic/constant_example.py
```diff
@@ -24,66 +27,76 @@ def source(t, f0):
     return (1-2.*r**2)*np.exp(-r**2)
 
 
-def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
-        time_order=2, space_order=4, nbpml=40, dse='advanced', dle='advanced',
-        full_run=False):
+def setup(dimensions=(50, 50, 50), spacing=(15.0, 15.0, 15.0), tn=500.,
```
FabioLuporini: Are you *also* trying to refactor this file so that it looks like as similar as possible to the other examples?

examples/seismic/acoustic/constant_example.py
```diff
@@ -24,66 +27,76 @@ def source(t, f0):
...
-    true_vp = 2.
-
-    # Smooth velocity
-    initial_vp = 1.8
-
-    dm = 1. / (true_vp * true_vp) - 1. / (initial_vp * initial_vp)
-    model = Model(origin, spacing, dimensions, true_vp, nbpml=nbpml)
-    dm = np.ones(model.shape_domain, dtype=np.float32)*dm
+    # Define seismic data
+    model = Model(origin, spacing, dimensions, true_vp, nbpml=int(nbpml))
```
FabioLuporini: why could ``nbpml`` be a non-float ?

examples/seismic/acoustic/operators.py
```diff
@@ -162,7 +163,7 @@ def GradientOperator(model, source, receiver, time_order=2, space_order=4, **kwa
     subs = dict([(t.spacing, dt)] + [(time.spacing, dt)] +
                 [(i.spacing, model.get_spacing()[j]) for i, j
                  in zip(v.indices[1:], range(len(model.shape)))])
-    return Operator([eqn] + [gradient_update] + receivers,
+    return Operator([eqn] + receivers + [gradient_update],
```
FabioLuporini: why?

examples/seismic/acoustic/wavesolver.py
```diff
@@ -38,6 +40,11 @@ def __init__(self, model, source, receiver,
         # Cache compiler options
         self._kwargs = kwargs
 
+        if ~isinstance(model.m, DenseData):
```
FabioLuporini: can this be dropped with the adoption of ``ConstantData`` ?

examples/seismic/acoustic/wavesolver.py
```diff
@@ -102,15 +111,21 @@ def forward(self, src=None, rec=None, u=None, m=None, save=False, **kwargs):
                          space_order=self.space_order,
                          dtype=self.model.dtype)
 
+        kwargs.update({'u': u})
+
         # Pick m from model unless explicitly provided
-        if m is None:
-            m = m or self.model.m
+        if m is not None:
```
FabioLuporini: Definitely, somehow, we should catch all this ``if m is not None: ...`` **before** this point, otherwise we end up with lots of duplicated code

tests/test_adjointA.py
```diff
@@ -30,3 +31,33 @@ def test_acoustic(dimensions, time_order, space_order, fix_dim):
     term2 = linalg.norm(rec.data) ** 2
     print(term1, term2, ("%12.12f") % (term1 - term2), term1 / term2)
     assert np.isclose(term1 / term2, 1.0, atol=0.001)
+
+
+@pytest.mark.parametrize('space_order', [8])
+@pytest.mark.parametrize('time_order', [2])
+@pytest.mark.parametrize('dimensions', [(60, 70), (60, 70, 80)])
+@pytest.mark.parametrize('fix_dim', [True, False])
+def test_acoustic_consant(dimensions, time_order, space_order, fix_dim):
```
FabioLuporini: typo

tests/test_adjointA.py
```diff
@@ -30,3 +31,33 @@ def test_acoustic(dimensions, time_order, space_order, fix_dim):
     term2 = linalg.norm(rec.data) ** 2
     print(term1, term2, ("%12.12f") % (term1 - term2), term1 / term2)
     assert np.isclose(term1 / term2, 1.0, atol=0.001)
+
+
+@pytest.mark.parametrize('space_order', [8])
```
FabioLuporini: why using ``parametrize`` if it's just a single fixed value

tests/test_operator.py
```diff
@@ -30,6 +30,22 @@ def symbol(name, dimensions, value=0., mode='function'):
     return s.indexify() if mode == 'indexed' else s
 
 
+class TestAPI(object):
+
+    @classmethod
+    def setup_class(cls):
+        clear_cache()
+
+    def test_code_generation(self):
```
FabioLuporini: This test, called **_code_generation**, was meant to verify the presence of the constants in the generated code (e.g. by searching for strings or inspecting ASTs through tools.pprint). Could you add such a test? Probably it can be generalised to non-constant data objects

The current test implementation, instead, compares computed values, which is also useful, but probably should be called differently (besides being added a docstring)

examples/seismic/acoustic/wavesolver.py
FabioLuporini: Probably not, fine

FabioLuporini: btw, any reason to use ~isinstance instead of !isinstance

examples/seismic/acoustic/constant_example.py
FabioLuporini: And why are you changing these values?

examples/seismic/acoustic/wavesolver.py
FabioLuporini: And, most importantly: why not simply letting Devito fails if one tries to pass in a DenseData when ``m`` was instead set to a constant in the problem specification? In such a case, where would Devito die? @mlange05 any thoughts ?

devito/operator.py
mloubout: Yes, will change. This is also what is breaking dse/dle probalby.
All the `required` in `create_elemental_function` are casted as `ScalarArguments` and `m` may en up in it.
I'll try to remove the hard-coded types.

examples/seismic/acoustic/constant_example.py
mloubout: Copy pasted the `acoustic-example` to have same setup.

examples/seismic/acoustic/constant_example.py
mloubout: `test_adjoint` adds `spc_order/2` to `nbpml` and with some pythons versions it's not an int anymore.

examples/seismic/acoustic/operators.py
mloubout: How it should be, I keep finding stuff like that (my fault).

examples/seismic/acoustic/wavesolver.py
mloubout: Habit :)
This  is gonna go away anyway tto swich to `compare_type(self.model.m, m)`, just need to figure out how to access Class type of m.

examples/seismic/acoustic/wavesolver.py
mloubout: It woould crash at `cfunction` as the inout type wouldn't match (ndpointer instead of const float and vis-versa)

tests/test_operator.py
mloubout: >e.g. by searching for strings or inspecting ASTs through tools.pprint)
I am honeslty completly lost trying to got through `visitor` and figure out how it works.

mloubout:
@FabioLuporini @mlange05 
Ok so everything should be in, the only issue left is the dse/dle case that makes `ConstantData` fail:
```
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/examples/seismic/acoustic/constant_example.py", line 104, in <module>
    dimensions=(50, 50), tn=500.)
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/examples/seismic/acoustic/constant_example.py", line 89, in run
    rec, u, summary = solver.forward(save=full_run, m=.25)
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/examples/seismic/acoustic/wavesolver.py", line 115, in forward
    summary = self.op_fwd_save.apply(**kwargs)
  File "/Volumes/Users/mloubout/anaconda/lib/python2.7/site-packages/cached_property.py", line 26, in __get__
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/examples/seismic/acoustic/wavesolver.py", line 55, in op_fwd_save
    space_order=self.space_order, **self._kwargs)
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/examples/seismic/acoustic/operators.py", line 64, in ForwardOperator
    time_axis=Forward, name='Forward', **kwargs)
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/devito/backends.py", line 75, in __call__
    return t(*args, **kwargs)
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/devito/core/operator.py", line 58, in __new__
    obj.__init__(*args, **kwargs)
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/devito/operator.py", line 89, in __init__
    clusters = rewrite(clusters, mode=set_dse_mode(dse))
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/devito/dse/symbolics.py", line 54, in rewrite
    processed.extend(modes[mode]().run(cluster))
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/devito/dse/backends/common.py", line 96, in run
    self._pipeline(state)
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/devito/dse/backends/advanced.py", line 20, in _pipeline
    self._extract_time_invariants(state)
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/devito/dse/backends/common.py", line 29, in wrapper
    for c in state.clusters]))
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/devito/dse/backends/advanced.py", line 36, in _extract_time_invariants
    processed, found = xreplace_constrained(cluster.exprs, make, rule, costmodel)
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/devito/dse/manipulation.py", line 179, in xreplace_constrained
    ret, _ = run(root)
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/devito/dse/manipulation.py", line 153, in run
    children = [run(a) for a in expr.args]
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/devito/dse/manipulation.py", line 150, in run
    base, flag = run(expr.base)
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/devito/dse/manipulation.py", line 153, in run
    children = [run(a) for a in expr.args]
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/devito/dse/manipulation.py", line 153, in run
    children = [run(a) for a in expr.args]
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/devito/dse/manipulation.py", line 153, in run
    children = [run(a) for a in expr.args]
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/devito/dse/manipulation.py", line 148, in run
    return expr, rule(expr)
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/devito/dse/queries.py", line 70, in <lambda>
    return lambda e: not e.is_Number and graph.time_invariant(e)
  File "/Users/mloubout/Dropbox/London/CodeGen/devito/devito/dse/graph.py", line 246, in time_invariant
    elif not i.base.function.is_SymbolicData:
AttributeError: 'Symbol' object has no attribute 'base'

```

examples/seismic/acoustic/operators.py
FabioLuporini: Oh no, I mean, why are you changing the order

mloubout: That's what I meant, that's the order it should be, I must have messed it up debugging at some point and forget about it, sorry.

devito/arguments.py
```diff
@@ -85,6 +91,44 @@ def verify(self, value):
         return self._value is not None
 
 
+class ConstantArgument(Argument):
+
+    """ Class representing scalar arguments that a kernel might expect.
```
FabioLuporini: Shouldn't then this inherit from the ``ScalarArgument``, thus reusing some of its code (e.g, ``def dtype``)

mloubout:
@FabioLuporini @mlange05 
All green but `test_code_generation` that fails in Python 3. It looks liek the argument order is changed with this Python version breaking the test.

FabioLuporini:
> All green but test_code_generation that fails in Python 3. It looks liek the argument order is changed with this Python version breaking the test.

Random thought: could it be due to this ``set`` here: https://github.com/opesci/devito/blob/master/devito/operator.py#L121 ?
Could you verify that? Btw, we'll have to drop that ``set`` anyhow as soon as we have MPI

devito/nodes.py
```diff
@@ -528,7 +528,6 @@ def __init__(self, name, body, retval, parameters=None, prefix=('static', 'inlin
         self.body = as_tuple(body)
         self.retval = retval
         self.prefix = prefix
-
```
FabioLuporini: ?

examples/seismic/acoustic/constant_example.py
```diff
@@ -1,18 +1,22 @@
...
 
 from devito.logger import info
-from examples.seismic.acoustic.wavesolver import AcousticWaveSolver
+from examples.seismic.acoustic import AcousticWaveSolver
 from examples.seismic import Model, PointSource, Receiver
+from devito import ConstantData
 
 
 # Velocity models
 def smooth10(vel, shape):
```
FabioLuporini: is this the same routine used in the other examples?

mloubout:
Hum odd, the second TRAVIS passes. 

>Could you verify that?

I don't have any python3 installed on my machinf for now, will look into it but won't be quick.

examples/seismic/acoustic/constant_example.py
```diff
@@ -24,66 +28,81 @@ def source(t, f0):
...
+    # Define seismic data
+    model = Model(origin, spacing, dimensions, true_vp, nbpml=int(nbpml))
 
-    # Define seismic data.
     f0 = .010
     dt = model.critical_dt
     if time_order == 4:
         dt *= 1.73
     t0 = 0.0
+    tn = tn
```
FabioLuporini: leftover

examples/seismic/acoustic/constant_example.py
```diff
@@ -24,66 +28,81 @@ def source(t, f0):
...
 
-    solver = AcousticWaveSolver(model, source=src, receiver=rec,
-                                time_order=time_order, space_order=space_order)
 
+def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
+        time_order=2, space_order=4, nbpml=40, full_run=False, **kwargs):
+
+    solver = setup(dimensions=dimensions, spacing=spacing, nbpml=nbpml, tn=tn,
+                   space_order=space_order, time_order=time_order,
+                   **kwargs)
```
FabioLuporini: no need for new line

examples/seismic/acoustic/constant_example.py
```diff
@@ -24,66 +28,81 @@ def source(t, f0):
...
-    receiver_coords[:, 0] = np.linspace(0, origin[0] +
-                                        (dimensions[0]-1) * spacing[0],
-                                        num=dimensions[0])
-    receiver_coords[:, 1:] = location[0, 1:]
+    # Define source and receivers and create acoustic wave solver
     src = PointSource(name='src', data=time_series, coordinates=location)
     rec = Receiver(name='rec', ntime=nt, coordinates=receiver_coords)
+    return AcousticWaveSolver(model, source=src, receiver=rec,
+                              time_order=time_order, space_order=space_order,
+                              **kwargs)
```
FabioLuporini: probably no need for new line

examples/seismic/acoustic/constant_example.py
```diff
@@ -24,66 +28,81 @@ def source(t, f0):
...
+
+    initial_vp = 1.8
+    dm = (initial_vp**2 - solver.model.m.data) * np.ones(solver.model.shape_domain,
+                                                         dtype=np.float32)
     info("Applying Forward")
-    rec, u, summary = solver.forward(save=full_run, dse=dse, dle=dle)
+    m0 = ConstantData(name="m", value=.25, dtype=np.float32)
+    rec, u, summary = solver.forward(save=full_run)
+    rec, u, summary = solver.forward(save=full_run, m=m0)
+    rec, u, summary = solver.forward(save=full_run, m=.25)
```
FabioLuporini: why overriding the previous results ?

examples/seismic/acoustic/wavesolver.py
```diff
@@ -211,10 +219,34 @@ def born(self, dmin, src=None, rec=None, u=None, U=None, m=None, **kwargs):
...
-            m = self.model.m
+        kwargs = self.check_input(m, **kwargs)
 
+        kwargs.update({'dm': dmin})
         # Execute operator and return wavefield and receiver data
-        summary = self.op_born.apply(dm=dmin, u=u, U=U, src=src, rec=rec, m=m, **kwargs)
+        summary = self.op_born.apply(**kwargs)
         return rec, u, U, summary
+
+    def check_input(self, m, **kwargs):
```
FabioLuporini: @mlange05 I'd like your input here

examples/seismic/tti/wavesolver.py
```diff
@@ -88,28 +90,85 @@ def forward(self, src=None, rec=None, u=None, v=None, m=None,
...
-        if len(m.shape) == 2:
-            summary = op.apply(src=src, rec=rec, u=u, v=v, m=m, epsilon=epsilon,
-                               delta=delta, theta=theta, **kwargs)
-        else:
-            summary = op.apply(src=src, rec=rec, u=u, v=v, m=m, epsilon=epsilon,
-                               delta=delta, theta=theta, phi=phi, **kwargs)
+        summary = op.apply(src=src, rec=rec, u=u, v=v, **kwargs)
         return rec, u, v, summary
+
+    def check_input(self, m, epsilon, delta, theta, phi, **kwargs):
```
FabioLuporini: @mlange05 again

FabioLuporini: This is madness, I'm not sure how but this needs to go

tests/test_adjointA.py
```diff
@@ -30,3 +31,30 @@ def test_acoustic(dimensions, time_order, space_order, fix_dim):
     term2 = linalg.norm(rec.data) ** 2
     print(term1, term2, ("%12.12f") % (term1 - term2), term1 / term2)
     assert np.isclose(term1 / term2, 1.0, atol=0.001)
+
+
+@pytest.mark.parametrize('dimensions', [(60, 70), (60, 70, 80)])
+@pytest.mark.parametrize('fix_dim', [True, False])
+def test_acoustic_constant(dimensions, fix_dim):
```
FabioLuporini: Can you not parametrize the test over ``setup`` (``setup``, ``setup_c``)? Isn't it the same function body? ( I haven't checked thoroughly though)

tests/test_tti.py
```diff
@@ -36,10 +36,10 @@ def test_tti(dimensions, space_order):
 
     model = Model(origin, spacing, dimensions,
                   true_vp, nbpml=nbpml,
-                  epsilon=0.0 * true_vp,
-                  delta=0.0 * true_vp,
-                  theta=0.0 * true_vp,
-                  phi=0.0 * true_vp)
+                  epsilon=None,
```
FabioLuporini: Isn't ``None`` default?

examples/seismic/acoustic/constant_example.py
mloubout: Just running every cases to make sure it doesn't break, the value of the output `rec` is not required here but tested in the adjoint test.

examples/seismic/tti/wavesolver.py
mloubout: ?

tests/test_tti.py
mloubout: It is, but don;t want people to have to dig into Model to see it, this is still TTI and wanna make sure it shows at setup.

examples/seismic/acoustic/constant_example.py
FabioLuporini: So remove the assignment. Just do
``solver.forward(...)``
otherwise it feels like you forgot to do something

tests/test_tti.py
FabioLuporini: But this is a test, people look at examples, not tests. Anyway, since this is a test, I don't really mind...

mloubout:
Still fails in 3D:
```
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T/devito-502/568c4e1e7cfebb825af642c323bdd39feb2f9211.c:15:201: error: conflicting types for 'm'
 void f_0(const int x_start, const int x_finish, const int y_start, const int y_finish, const int z_start, const int z_finish, float *restrict damp_vec, const float m, float *restrict u_vec, const int m, const int time, const int time_size, const int x_size, const int y_size, const int z_size)
                                                                                                                                                                                                         ^
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T/devito-502/568c4e1e7cfebb825af642c323bdd39feb2f9211.c:15:165: note: previous definition of 'm' was here
 void f_0(const int x_start, const int x_finish, const int y_start, const int y_finish, const int z_start, const int z_finish, float *restrict damp_vec, const float m, float *restrict u_vec, const int m, const int time, const int time_size, const int x_size, const int y_size, const int z_size)
                                                                                                                                                                     ^
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T/devito-502/568c4e1e7cfebb825af642c323bdd39feb2f9211.c: In function 'f_0':
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T/devito-502/568c4e1e7cfebb825af642c323bdd39feb2f9211.c:23:0: warning: ignoring #pragma omp simd [-Wunknown-pragmas]
       #pragma omp simd aligned(damp,u:16)
 ^
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T/devito-502/568c4e1e7cfebb825af642c323bdd39feb2f9211.c: In function 'Forward':
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T/devito-502/568c4e1e7cfebb825af642c323bdd39feb2f9211.c:56:99: error: 'm_vec' undeclared (first use in this function)
           f_0(x_block,x_block + x_block_size,y_block,y_block + y_block_size,3,z_size - 3,damp_vec,m_vec,u_vec,m,time,time_size,x_size,y_size,z_size);
                                                                                                   ^
/var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T/devito-502/568c4e1e7cfebb825af642c323bdd39feb2f9211.c:56:99: note: each undeclared identifier is reported only once for each function it appears in
FAILED compiler invocation: gcc -O3 -g -march=native -mno-avx -fPIC -Wall -std=c99 -Wno-unused-result -Wno-unused-variable -shared -fopenmp /var/folders/z0/tqrtpj3j5lz5vmpm09lx4mbw0000gp/T/devito-502/568c4e
```

mloubout:
And yes, it is still broken on `master` too.

mlange05:
Confirmed, the `list(set(...))` constructs in `operator.py` are non-deterministically broken in master. Will file a separate fix in a minute.

examples/seismic/acoustic/wavesolver.py
mlange05: Yeah I don't think this routine is necessary at all. The `wavesolver` should just pass the objects down and let the devito argument engine pick up any changes in the type signature. 

If we want to go further and cause re-compilation, I would propose to improve the operator cashing to take the type signature into account. But as far as I understand it, this is not yet required, and so beyond this PR? @mloubout

examples/seismic/tti/wavesolver.py
mlange05: Again, `wavesolver` should not need to know about this.

examples/seismic/model.py
```diff
@@ -65,7 +65,7 @@ def __init__(self, origin, spacing, shape, vp, nbpml=0, dtype=np.float32,
             self.m = DenseData(name="m", shape=self.shape_domain, dtype=self.dtype)
             self.m.data[:] = self.pad(1 / (self.vp * self.vp))
         else:
-            self.m = 1/vp**2
+            self.m = ConstantData(name="m", dtype=np.float32, value=1/vp**2)
```
mlange05: We probably want a utility routine here that allocated the right type based on inspecting the argument. This way we can re-use it for all of the TTI-related fields just the same way.

FabioLuporini:
@mloubout I pushed a potential fix for the 3D case, let me know how it goes

FabioLuporini:
There still is a segfault, but I wonder whether this is actually an issue of the model. The segfault disappears in "safe math mode". I'll explain once you can confirm that the numerical results with ``dle = 'noop'`` are meaningful, otherwise it'd be pointless

examples/seismic/acoustic/wavesolver.py
```diff
@@ -83,17 +84,19 @@ def forward(self, src=None, rec=None, u=None, m=None, save=False, **kwargs):
         :param u: (Optional) Symbol to store the computed wavefield
         :param m: (Optional) Symbol for the time-constant square slowness
         :param save: Option to store the entire (unrolled) wavefield
-
+s
```
mlange05: ?

devito/operator.py
```diff
@@ -182,6 +182,7 @@ def arguments(self, **kwargs):
         autotune = autotune and kwargs.pop('autotune', False)
 
         # Make sure we've used all arguments passed
+        # Breaks for constant velocity
```
FabioLuporini: Drop this

examples/seismic/acoustic/acoustic_example.py
```diff
@@ -99,4 +99,4 @@ def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
 
 
 if __name__ == "__main__":
-    run(full_run=False, autotune=False, space_order=6, time_order=2)
+    run(full_run=True, autotune=False, space_order=6, time_order=2)
```
FabioLuporini: Oh no, let's not start again changing these things

examples/seismic/acoustic/wavesolver.py
```diff
@@ -211,10 +225,17 @@ def born(self, dmin, src=None, rec=None, u=None, U=None, m=None, **kwargs):
                          save=False, time_order=self.time_order,
                          space_order=self.space_order, dtype=self.model.dtype)
 
+        kwargs.update({'src': src})
+        kwargs.update({'rec': rec})
+        kwargs.update({'u': u})
+        kwargs.update({'U': U})
+        kwargs.update({'m': self.model.m})
```
FabioLuporini: @mlange05 any thoughts on this kwargs updates? Shouldn't we at least avoid changing this parameter in-place ?

tests/test_adjointA.py
```diff
@@ -12,6 +14,7 @@
 @pytest.mark.parametrize('dimensions', [(60, 70), (60, 70, 80)])
 @pytest.mark.parametrize('fix_dim', [True, False])
 def test_acoustic(dimensions, time_order, space_order, fix_dim):
+    set_log_level('DEBUG')
```
FabioLuporini: drop

tests/test_adjointA.py
FabioLuporini: This is still unanswered

tests/test_adjointA.py
```diff
@@ -30,3 +33,30 @@ def test_acoustic(dimensions, time_order, space_order, fix_dim):
...
+    else:
+        time.size = None
+    # Run forward and adjoint operators
+    rec, _, _ = solver.forward(save=False)
+    solver.adjoint(rec=rec, srca=srca)
+
+    # Actual adjoint test
+    term1 = np.dot(srca.data.reshape(-1), solver.source.data)
+    term2 = linalg.norm(rec.data) ** 2
+    print(term1, term2, ("%12.12f") % (term1 - term2), term1 / term2)
```
FabioLuporini: turn into ``debug(...)``

tests/test_adjointA.py
```diff
@@ -30,3 +33,30 @@ def test_acoustic(dimensions, time_order, space_order, fix_dim):
...
+    assert np.isclose(term1 / term2, 1.0, atol=0.001)
+
+
+if __name__ == "__main__":
```
FabioLuporini: drop?

examples/seismic/acoustic/acoustic_example.py
mloubout: No idea whi changed that, but the example is supposed to run fully

tests/test_adjointA.py
mloubout: Different parameter, one does not have multiple space_order. Only there to check constant works properly and produces correct result

tests/test_adjointA.py
mloubout: This mains are very useful, don;t like to drop it

examples/seismic/acoustic/wavesolver.py
mloubout: I find it nicer than all the `if` but can change it back.

tests/test_adjointA.py
mlange05: You can easily run these with `py.test .... -k <some patter to identify single test>`. Using this instead of `__main__` functions means that you also etst your on test setup when you do dev runs, and pytest bugs like the silent `setup()` would be caught straight away...

examples/seismic/acoustic/wavesolver.py
mlange05: Yeah, I don't understand why this is necessary either.  What's wrong with the previous `op.apply(src=src, rec=rec, u=u...)` approach?

FabioLuporini:
So this is still segfaulting. 
@mloubout : can you tell me how the arithmetic is varying when you comment out:
``self._avoid_denormals(state)`` in ``devito/dle/backends/advanced.py`` ?
Apparently if we comment this out it doesn't segfault.

FabioLuporini:
I think this is either:
- a bug in the numerics,
- or a GCC bug (Intel works like a charm after my latest fix)
The generated code looks sane... not sure what could I miss.

I'll leave this to you @mlange05 @mloubout 
My (temporary?) fix consists of a slightly less aggressive DLE mode, which doesn't drop denormal numbers. This is really mysterious though, so I strongly encourage to have a look at the numerics to confirm that this is actually a GCC bug.

tests/test_adjointA.py
mloubout: In michael's PR

tests/test_adjointA.py
mlange05: FWIW, we can further parametrize this once the model is a single switch (PR #308). We can then also build explicit test-parameter matrices to manage complexity.

examples/seismic/acoustic/acoustic_example.py
mlange05: :grimacing: `full_run=True` tends to completely kill laptops with only 4GB mem :cry:, so it's kind of an evil landmine to have lying around, especially with no killswitch.  I'd much prefer some basic option processing with safe defaults (eg. `python acoustic_example.py --full -so 12 --autotune` or similar). It's fine for this PR, but worth considering in one of the next ones, because I will eventually delete this test in a fit of rage if it takes out my machine again. :smiling_imp:

mloubout: I can reduce dedault tn to keep it at ok level memorywise. All operators are now tested so this one is not really mandatory indeed so can be skipped in tests. Would still kepp it as example of our wrappers

FabioLuporini:
I think I'm happy, although I'd have preferred someone to look at what happens to the numerics as per my comments above.

Finally: @mlange05 do I have the honor to hit merge?

mlange05:
@FabioLuporini Yes, please. You can hit the button. 

I also had a quick look at the numerics issue, but couldn't figure it out on the spot. Maybe file an issue to track this and we'll all vow to get to the bottom of this ASAP?

FabioLuporini:
Merged.

vahid-honarbakhsh:
Hi
How can I replace a simple 2-layer model in the example with any other model,for example 2D Marmousi data set؟
Thanks,

PR #300, opened 2017-07-17
Support variable grids, only 1D-3D

mloubout:
Small modification to support 3D spacing such as `spacing=(10,10,15)` (coarser in depth).

Input appreciated.


FabioLuporini:
I think this is more your territory @mlange05

mlange05:
Ok, this is a really nice start towards having a better "Grid" abstraction. However, since we are now creating the dimension->spacing symbol map in multiple places, I would suggest extending the `Dimension` class with a `dim.spacing`attribute/property so that we can define dimensions like this (in `dimension.py`):
```
x = Dimension('x', spacing='h_x')
y = Dimension('y', spacing='h_y')
...
```
This means , instead of getting `h_x` from `x` via a map we can simply do `x.spacing` or `dim.spacing` wherever we need the spacing symbol, including in the `subs` dictionary where we replace it again.

I also note that, in the future we might also be able to auto-create the `subs` dictionary from a higher-level "Grid" class, which now only needs to know about dimensions and verify that all fields agree on space dimensions, etc. But that is probably best left for a separate PR because there might be complicated details.

mloubout:
Thanx for the Dimension hint, will ne one it as soon as I have fixed all the tests.

mloubout:
Also, wait for my Green for this one, I need to do some compatibility on julia side

mlange05:
Ok, this is shaping up nicely. Can we also do the same trick for `time/t` and `s` to stay consistent? After that we should be good to go (we'll get rid of the complex `subs` expressions in the next PR).

devito/dimension.py
```diff
@@ -22,6 +22,7 @@ def __new__(cls, name, **kwargs):
         newobj = Symbol.__new__(cls, name)
         newobj.size = kwargs.get('size', None)
         newobj.reverse = kwargs.get('reverse', False)
+        newobj.spacing = kwargs.get('spacing', Symbol('h'))
```
FabioLuporini: you mean ``Symbol('h_%s' % name)`` ?

devito/finite_difference.py
```diff
@@ -115,39 +112,43 @@ def cross_derivative(*args, **kwargs):
        ``f(h + x, -h + y)*g(h + x, -h + y)) / h**2``
     """
     dims = kwargs.get('dims', (x, y))
-    diff = kwargs.get('diff', h)
+    diff = kwargs.get('diff', (dims[0].sapcing, dims[1].sapcing))
```
FabioLuporini: typo

devito/finite_difference.py
```diff
@@ -115,39 +112,43 @@ def cross_derivative(*args, **kwargs):
...
              for i in range(-int(order / 2) + 1 - (order < 4),
                             int((order + 1) / 2) + 2 - (order < 4))]
 
     # Finite difference weights from Taylor approximation with this positions
-    c1 = finite_diff_weights(1, ind1r, dims[0])
-    c1 = c1[-1][-1]
-    c2 = finite_diff_weights(1, ind1l, dims[0])
-    c2 = c2[-1][-1]
+    c11 = finite_diff_weights(1, ind1r, dims[0])
+    c11 = c11[-1][-1]
```
mlange05: Optional: We could combine the two lines to give `c11 = finite_diff_weights(1, ind1r, dims[0])[-1][-1]`. Same for all below.

devito/interfaces.py
```diff
@@ -184,7 +184,9 @@ def _mem_heap(self):
 
     def indexify(self):
         """Create a :class:`sympy.Indexed` object from the current object."""
-        indices = [a.subs({h: 1, s: 1}) for a in self.args]
+        indices = [a.subs({x.spacing: 1, y.spacing: 1, z.spacing: 1, s: 1})
```
mlange05: Should probably be generalised at some point, but good for now.

examples/cfd/example_diffusion.py
```diff
@@ -133,7 +133,9 @@ def execute_devito(ui, spacing=0.01, a=0.5, timesteps=500):
     # Derive the stencil according to devito conventions
     eqn = Eq(u.dt, a * (u.dx2 + u.dy2))
     stencil = solve(eqn, u.forward)[0]
-    op = Operator(Eq(u.forward, stencil), subs={h: spacing, s: dt})
+    op = Operator(Eq(u.forward, stencil), subs={x.spacing: spacing,
+                                                y.spacing: spacing,
+                                                s: dt})
```
mlange05: Should this be `time.spacing: dt`?

examples/cfd/test_01_convection.ipynb
```diff
@@ -398,7 +413,9 @@
     "# Now combine the BC expressions with the stencil to form operator\n",
     "expressions = [Eq(u.forward, stencil)]\n",
     "expressions += [bc_left, bc_right, bc_top, bc_bottom]\n",
-    "op = Operator(expressions=expressions, subs={h: dx, s:dt},\n",
+    "op = Operator(expressions=expressions, subs={x.spacing: dx,\n",
+    "                                             y.spacing: dy,\n",
+    "                                             t.spacing:dt},\n",
```
mlange05: Nitpick: Space after `t.spacing:`.

examples/seismic/acoustic/acoustic_example.py
```diff
@@ -99,4 +99,5 @@ def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
 
 
 if __name__ == "__main__":
-    run(full_run=True, autotune=False, space_order=6, time_order=2)
+    run(full_run=False, autotune=False, space_order=6, time_order=2, dimensions=(50, 60),
```
mlange05: Why are we changing the way the test is run here? Is this just a debug leftover?

examples/seismic/tti/operators.py
```diff
@@ -152,7 +152,8 @@ def ForwardOperator(model, source, receiver, time_order=2, space_order=4,
                            offset=model.nbpml)
     stencils += src.inject(field=v, u_t=ti + 1, expr=src * dt * dt / m,
                            offset=model.nbpml)
-    stencils += rec.interpolate(expr=u, u_t=ti, offset=model.nbpml)
-    stencils += rec.interpolate(expr=v, u_t=ti, offset=model.nbpml)
-
+    stencils += rec.interpolate(expr=u + v, u_t=ti, offset=model.nbpml)
```
mlange05: Lolz, we (me and @FabioLuporini) had been wondering whether this was supposed to be two `rec` expressions or not. :wink:

mloubout: Supposed to be interpolate the sum, saw it so changed it here even so not really core of the PR.

examples/seismic/acoustic/acoustic_example.py
mloubout: debug, will remove

examples/cfd/example_diffusion.py
mloubout: yes it should, leftover (so tedious PR... :) )

devito/interfaces.py
mloubout: yes, and s shouldn't be there, will fix

examples/seismic/tti/operators.py
FabioLuporini: incidentally, the DSE should sum them up automatically

mloubout:
Corresponding PR ready in SLIM for julia, good to go to me.

mlange05:
Ok, some unwanted debug output seems to have crept into the seismic modelling notebook. I'll re-run locally to clean up and then merge.

PR #303, opened 2017-07-19
cached-property install

mloubout:
Added missing entry in the `setup.py` file.

Also for some reason install wants `cached-property` instead of `cached_property`


FabioLuporini:
Merged.

PR #305, opened 2017-07-20
Abc separated

mloubout:
This still needs some work, but it's quite cool so leave it there (also for setup/organization comments).

Main point
- `damp` is finally gone, this was a useless waste of computation and memory
- new `boundary` class that currently only contains absorbing boundary conditions `ABC`. Only compatible with acoustic wave equation (only one we have so oK) for now


mloubout:
Small update, still no idea WTF is the problem with sympy and divisions, still working on it

mloubout:
Sympy issue I need to figure out, closing fornow

mloubout:
`sympy` issue found, reopening

Added free-surface. Will be improved once #360 is in.

examples/seismic/boundaries.py
```diff
@@ -0,0 +1,132 @@
...
+        :return:
+        """
+        return [self.abc_eq.subs({y: self.p_abc}),
+                self.abc_eq.subs({y: self.full_shape[0] - 1 - self.p_abc})]
+
+    def damp_z(self):
+        """
+        Dampening profile along z
+        This seems to need a different Dimension to avoid beeing
+        stuck in an infinite loop inside /devito/tools.py", line 123 while loop on queue
```
mloubout: @fablup if you have any idea on this when you find time (no priority/urgency at all)

FabioLuporini:
Once rebased, remind me what is that this PR needs to go in, @mloubout ?

mloubout:
So his PR is an attempt to separate the PDE and the BC parts, this is crazy inefficient currently so shouldn't be merge. The stencils in this PR go as follows:

```
stencil = Eq(forward, sloce(eq,forward)
stencil += ABC
stencil += free_surface if free_surface (reflection at the top
```
this should "higlhilght" the potential to split PDE and boundary conditons.

Also one dle issue as in 3D there is three "indices" that are `abc., y, z` , `x, abc, z` and `x, y, abc` and dle is stuck in an infinite while loop trying to schedule it

FabioLuporini:
> Also one dle issue as in 3D there is three "indices" that are abc., y, z , x, abc, z and x, y, abc and dle is stuck in an infinite while loop trying to schedule it

Any chance this is reproducible in a small, self-contained example on top of master? Lower priority, though, so no worries..

mloubout:
>Any chance this is reproducible in a small, self-contained example on top of master?

Not sure, this is a really odd case, will see what I can do

mloubout:
CLosing, needs full rethinking. The main issue is that there is a loop over the size of the ABC layer, but this size does not correspond to any data and can't be infered at apply()

FabioLuporini:
> The main issue is that there is a loop over the size of the ABC layer

Just out of curiosity, could you provide some pseudo-code of what you'd like to be generated?

mloubout:
>some pseudo-code of what you'd like to be generated?

something like:
```
for(abc=1; abc<nbpm;abc++)
     u[t][abc][y][z] = ... # apply_abc
```

but not field has `abc` size its passed as `None` at apply

PR #308, opened 2017-07-24
WIP: Seismic model utility

mlange05:
Just floating this as an idea here... The main purpose of this PR is to pave the way for the more advanced seismic imaging/inversion tutorials; but it does introduce enough to be discussed separately, I think. It's nowhere near done, and I'm happy to extend the idea *iff* we achieve some sort of consensus.

The core idea is that we create and perturb seismic models in many places throughout the examples - and although the models are near trivial (two layer, etc.) they eat up 20-30 lines in each instance, which is bad for reaadability. Instead I believe we can provide a demo utility that creates a particular variant of various preset models (including Marmousi, for example) to use (near) interchangeably in examples and tutorials. Of course, we would not run Marmousi on travis, but it would then (at least in most cases) require only a single-line change in a notebook to switch from a two-layer trivial model to a real seismic use case.

To discuss:
* The extend to which we can generalise a conceptual preset like "two layers" (eg. across 2D/3D) for general use in the test case. Or should we just use this for demo/tutorial purposes?
* Do we have other presets that are simple, but complex enough to be of interest (I think I saw a circle somewhere...)?
* Is a plain utility function with a keyword enough, or do we want something more sophisticated?

More details in this PR:
* For convenience and readability separate the TTI utility fields into a dedicated `AnisotropicModel` class.
* To apply individual smoothing on a preset model we now need to update `model.m.data` from model.vp`, so I have finally found a legitimate reason for a Python "setter" method :open_mouth:.
This would then allow us to do perturbations like this:
```
model = demo_model('marmousi2D')

model0 = demo_model('marmousi2D')
model0.vp[:] = scipy.ndimage.gaussian_filter(model0.vp, ...)
```


examples/seismic/acoustic/acoustic_example.py
```diff
@@ -23,34 +23,23 @@ def smooth10(vel, shape):
...
-    # Define seismic data
-    model = Model(origin, spacing, dimensions, true_vp, nbpml=int(nbpml))
+    model = demo_model('layers', ratio=3, shape=dimensions,
+                       spacing=spacing, nbpml=nbpml)
 
     # Derive timestepping from model spacing
-    dt = model.critical_dt
-    if time_order == 4:
-        dt *= 1.73
+    dt = model.critical_dt * 1.73 if time_order == 4 else 1.0
```
mloubout: Not sure what the if does, apply to the * only or to dt?

If dt, should be `else model. critical_dt`

examples/seismic/model.py
```diff
@@ -1,9 +1,78 @@
...
+        vp = kwargs.pop('vp', 1.5)
+
+        return Model(vp=vp, origin=origin, shape=shape,
+                     spacing=spacing, nbpml=nbpml, **kwargs)
+
+    elif preset.lower() in ['layers', 'twolayer', '2layer']:
+        # A two-layer model in a 2D or 3D domain with two different
+        # velocities split across the height dimension:
+        # The top part of the domain has 1.5 km/s,
+        # the bottom part of the domain has 1.5 km/s.
```
mloubout: typo, 2.5

examples/seismic/model.py
```diff
@@ -1,9 +1,78 @@
...
+        shape = kwargs.pop('shape', (101, 101))
+        spacing = kwargs.pop('spacing', tuple([10. for _ in shape]))
+        origin = kwargs.pop('origin', tuple([0. for _ in shape]))
+        nbpml = kwargs.pop('nbpml', 10)
+        ratio = kwargs.pop('ratio', 2)
+
+        # Define a velocity profile in km/s
+        v = np.empty(shape, dtype=np.float32)
+        v[:] = 1.5  # Top velocity (background)
+        v[..., shape[-1] / ratio:] = 2.5  # Bottom velocity
```
mloubout: `int(shape[-1]/ratio)`, sometime non-integer and throw warning

tests/test_adjointA.py
```diff
@@ -4,24 +4,66 @@
...
+    t0 = 0.0  # Start time
+    tn = 500.  # Final time
+    nrec = 130  # Number of receivers
+
+    # Create model from preset
+    model = demo_model(spacing=[15. for _ in dimensions],
+                       shape=dimensions, nbpml=10, **(presets[mkey]))
+
+    # Derive timestepping from model spacing
+    dt = model.critical_dt * 1.73 if time_order == 4 else 1.0
```
mloubout: Same as acoustic example

examples/seismic/model.py
```diff
@@ -120,6 +196,13 @@ def __init__(self, origin, spacing, shape, vp, nbpml=0, dtype=np.float32,
             self.phi = 0
 
     @property
+    def dim(self):
```
mloubout: love this one, this `len` everywhere wasn't really smart indeed

tests/test_dse.py
```diff
@@ -13,53 +13,41 @@
...
-from examples.seismic.tti.tti_example3D import setup
+from examples.seismic import demo_model, RickerSource, GaborSource, Receiver
+from examples.seismic.tti import AnisotropicWaveSolver
 
 
 # Acoustic
 
 def run_acoustic_forward(dse=None):
     dimensions = (50, 50, 50)
-    origin = (0., 0., 0.)
```
mloubout: I would still keep origin, even if not used for readability, just personal opinion, can ignore.

mloubout:
Flake8 and good to go for me

tests/test_dse.py
mlange05: In outward-facing examples, yes, maybe, but probably not in internal tests like this.

mlange05:
Ok, there are still a few bits and pieces to be done documentation-wise, but we can do that later. Let's get this landed, so we can focus on the tutorial notebooks.

There are still some odd error with Py3, but I'll go ahead and fix them and merge.

mlange05:
Ok, fixed, merging...

Issue #298, opened 2017-07-17
time-order and indices


mloubout:
So I realized that everywhere in the `acoustic` and `tti` wrapper there is : 
```
u = TimeData(name='u', shape=self.model.shape_domain,
                         save=save, time_dim=self.source.nt,
                         time_order=self.time_order,
                         space_order=self.space_order,
                         dtype=self.model.dtype)

```
`time_order` is supposed to be allways set to `2` for the wavefield. I don;t know when this changed but it now creates weird time loop indices as it set temp `t` for `t+2` and `t-2`.

Also, the Buffered indices are now `time, time+1, time+2` instead of `time-1, time, time+1`.

mlange05:
Ok,
a) It would be really good if you could explain why the wavefield has to be 2nd order, even for the 4th order scheme. There are no annotations in the code that state this, so it is really easy to miss. 
b) We're testing with 4th order time - why are none of the tests triggering on this?

mloubout:
for a)
http://www.sciencedirect.com/science/article/pii/S0168927406001012?via%3Dihub

Mainly, 4th order taylor is unconditionaly unstable so
- apply d/dt^2 to the wave equation
- replace some terms
- end up with a double laplacian but no 4th order time derivative
- still 4th order in time
b)
tricky, mix of 4th and 2nd order everywhere

mloubout:
>Also, the Buffered indices are now time, time+1, time+2 instead of time-1, time, time+1.

any update?

For the 4th order I'm looking into fixing it in a nice way.

FabioLuporini:
This commit/line **could** be the cause: https://github.com/opesci/devito/pull/293/commits/39afb019b853b1f3ec97394d554d13e23890b45a#diff-a106a375c221953a1c637f16ecad1c9eR557

mloubout:
>Also, the Buffered indices are now time, time+1, time+2 instead of time-1, time, time+1.

Turns out, I made that change, this is to make sure modulo don't get caught in weird behavior forcing any modulo-ed index to be positive.

I'll try to make the 4th order as a "special kernel" the way TTI is handled as soon as I find time to do it.

FabioLuporini:
@mloubout fixed?

FabioLuporini:
@mloubout fixed?

mloubout:
Yes by #473, closing

PR #314, opened 2017-07-31
Seismic interpolation cleanup

mlange05:
So it turns out that all the explicit time-index replacements via the outdated `u_t` and `p_t` arguments are pretty unnecessary. This merge removes them and uses `field=u.forward / v.backward` instead, as well as defaulting to the `time` dimension in the `PointData` constructor.


FabioLuporini:
OK

mlange05:
I can't actually see any use case that you can't use with `u.forward/backward`, except for when you sample/interpolate the src/rec data at a time offset. For this very specific case you can still use a simple `subs` in the operator definition code. In the meantime, I think it's much more dangerous to have outdated and unused backdoors lying around that clutter up the interface.

mloubout:
>I can't actually see any use case that you can't use with u.forward/backward

Current 4th order, `u.forward` is `t+2` which is not the one you want (`t, t+1, t-1`  only in propagation once #298 is fixed). I actually have no idea why TRAVI is green as 4th order in time should fail with this implementation.

This also gonna be problematic for elastic (first order coupled) where either `forward` or `backward` is gonna create problem.

mlange05:
Is this not a separate problem? Regardless of fixing the 4th order `u`/`v` discretization, should the time index we update with the stencil not be the same as the time index used in src/rec expressions? If so, I would say this is an unrelated issue.

> This also gonna be problematic for elastic (first order coupled) where either forward or backward is gonna create problem.

You need to be way more specific than that. In the meantime you can manually sub indices on the input to `interpolate` and the discretized output epxpression, so unless we can specify what the problem is there should be enough flexibility in the interface to create whatever you need.

mloubout:
>should the time index we update with the stencil not be the same as the time index used in src/rec expressions? 

This depends on the numerical scheme used. We currently used simple forward Euler, but some more advanced schemes may change that.

>e you can manually sub indices on the input to interpolate

This is what I was trying to avoid when added `u_t, p_t`. May sound trivial for you or me but most users won't have any idea what's going on wihtout it.

examples/seismic/acoustic/operators.py
```diff
@@ -46,15 +46,11 @@ def ForwardOperator(model, source, receiver, time_order=2, space_order=4,
     eqn = [Eq(u.forward, stencil)]
 
     # Construct expression to inject source values
-    # Note that src and field terms have differing time indices:
-    #   src[time, ...] - always accesses the "unrolled" time index
-    #   u[ti + 1, ...] - accesses the forward stencil value
-    ti = u.indices[0]
-    src_term = src.inject(field=u, u_t=ti + 1, offset=model.nbpml,
-                          expr=src * dt**2 / m, p_t=time)
+    src_term = src.inject(field=u.forward, expr=src * dt**2 / m,
```
mloubout: doesn't this work even without changing `pointdata`?

mlange05: Yes, but now the feature is unused and untested, and hence assumed to be broken. I really don't get your apprehension here - it's a simple `subs` that can easily added back in by hand for special cases without having a dedicated API in core abstractions, eg. `src.inject(field=u.subs(t, ti + 2), ...)`. If those special cases become frequent, it's also easy to put it back in - but until that time it's just API clutter.

mlange05:
> This is what I was trying to avoid when added u_t, p_t. May sound trivial for you or me but most users won't have any idea what's going on wihtout it.

Ok, I hear you. I was referring to using explicit `subs` for development purposes, if and when we play with staggered and "weird setups". If we happen to find a good reason to do these `subs` often and regularly, we can easily put the short-cut back in - but until such a time having an unused backdoor seems overkill, especially we have a still reasonably simple developer-level workaround.

Also, if you're happy with my reasoning, can you please change the review status?

mloubout:
Well I need to make a new Julia branch and make compatible with this first because if we merge that SLIM crashes, and fined time to do it.

mlange05:
@mloubout Any update on down-streaming this change to Julia?

Also, in order to avoid this sort of delay for interface changes, do you maybe want to make the Julia install use explicit Devito version tags? We've recently done several releases, so you can always run on a safe version and update Julia, whenever you think it needs updating and have the time to align the two.

mloubout:
@mlange05 
SO I actually need at least `p_t` as I am working with kernel that require:
`rec.inject(v.backward, p_t=time+shift)` with shift betwee ~-200:+200

mlange05:
Ok, fine. I'll rebase and leave the backdoors in. But please be warned that after this merge they will effectively be untested and can thus break at any point without us catching it.

mlange05:
@mloubout Can you please verify this still works for you, so I can finally merge it? Thanks

mloubout:
Good to go.

mlange05:
Great, merging...

PR #319, opened 2017-08-03
Yask integration phase6

FabioLuporini:
Redesign of the YASK backend, though no new functionalities are added.

Needed because now I understand more of how the thing is supposed to work

EDIT: kills some duplicated code


FabioLuporini:
Merged.

PR #321, opened 2017-08-03
Helper readme for running benchmarks

FabioLuporini:


examples/PERFORMANCE.md
```diff
@@ -0,0 +1,181 @@
...
+auto-tuner aggressiveness through:
+```
+DEVITO_AUTOTUNING=aggressive
+```
+
+### Choice of the backend compiler
+
+For each Operator, Devito generates C code, which then gets compiled into a
+shared object through the user-selected "backend compiler". Supported backend
+compilers are `gcc`, `icc`, `llvm`. In theory, any C compiler would work, but
```
mlange05: We recognise keywords `clang` or `osx`, but not `llvm`.

examples/PERFORMANCE.md
```diff
@@ -0,0 +1,181 @@
...
+
+ * At the moment, there is no support for MPI parallelism. This is perhaps the
+   biggest limitation, because it significantly affects execution on
+   multi-socket nodes.  The ideal setting on multi-socket nodes would be to
+   have 1 MPI process per socket, and OpenMP (or, why not, MPI itself in
+   shared-memory mode) for the processes within a socket. One can still use
+   OpenMP across all available sockets (the default case if `OMP_NUM_PROCS` is
+   unset and more than one sockets are available), but the final performance
+   will be very far from the attainable machine peak, due to the well known
+   NUMA effect.
```
mlange05: We do have a parameter `numa` that can be set on `DenseData`/`TimeData` objects to force first-touch, but I don't think is enabled by default, or accessible through global parameters. Should we maybe add this and mention it here?

examples/PERFORMANCE.md
FabioLuporini: fixed

FabioLuporini:
@mlange05 I sadly just noticed that `numa ` still attempts to use the old Propagator :(

Also, there's no simple work around to the fact that due to auto-tuning, the effect of `numa` will anyway be totally unpredictable, so... 

The right and only solution is one MPI rank per socket. I vote to drop `numa`.

FabioLuporini:
Merged.

PR #322, opened 2017-08-03
ConstantData

FabioLuporini:
Open to debate

### DIFF
--- a/devito/dle/backends/basic.py
+++ b/devito/dle/backends/basic.py
@@ -89,11 +89,10 @@ class BasicRewriter(AbstractRewriter):
 
                 # Retrieve tensor arguments
                 for i in fsymbols:
-                    if i.is_SymbolicFunction:
-                        handle = "(%s*) %s" % (c.dtype_to_ctype(i.dtype), i.name)
-                    else:
-                        handle = "%s_vec" % i.name
-                    args.append((handle, i))
+                    if i.is_TensorFunction:
+                        args.append(("(%s*)%s" % (c.dtype_to_ctype(i.dtype), i.name), i))
+                    elif i.is_SymbolicData:
+                        args.append(("%s_vec" % i.name, i))
 
                 # Retrieve scalar arguments
                 not_required.update({i.output for i in expressions if i.is_scalar})
--- a/devito/interfaces.py
+++ b/devito/interfaces.py
@@ -755,6 +755,13 @@ class IndexedData(IndexedBase):
         obj.function = self.function
         return obj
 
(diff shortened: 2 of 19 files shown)


FabioLuporini:
@mlange05 this test is currently failing: ``TestArithmetic.test_indexed_open_loops``
it looks like this is due to some sort of weird effects of using ``eval`` on expressions; if you run the single test, it works like a charm. Otherwise it fails, probably because it's picking some older ``DenseData``, though I'm not sure...

devito/interfaces.py
```diff
@@ -94,18 +94,51 @@ def __init__(self, \*args, \*\*kwargs):
...
+          ------------------                 ------------------
+          |                |                 |                |
+    ScalarFunction  TensorFunction     ConstantData           |
+                                                              |
+                                                ----------------------------
+                                                |             |            |
+                                            DenseData      TimeData  CompositeData
+                                                                           |
+                                                                       PointData
+
```
navjotk: This is very helpful. Thank you for this.

devito/dle/backends/basic.py
```diff
@@ -89,11 +89,10 @@ def _create_elemental_functions(self, state, **kwargs):
...
                 for i in fsymbols:
-                    if i.is_SymbolicFunction:
-                        handle = "(%s*) %s" % (c.dtype_to_ctype(i.dtype), i.name)
-                    else:
-                        handle = "%s_vec" % i.name
-                    args.append((handle, i))
+                    if i.is_TensorFunction:
+                        args.append(("(%s*)%s" % (c.dtype_to_ctype(i.dtype), i.name), i))
+                    elif i.is_SymbolicData:
+                        args.append(("%s_vec" % i.name, i))
```
navjotk: Isn't this just `ccode`?

FabioLuporini: good point, I'll try that !

FabioLuporini: turns out this is the string for the function call, not the function declaration
we could definitely add it along with ``decl`` ``ccode`` etc, but not part of this PR

FabioLuporini:
I suspect the Python3.5 bug here is due to #323

FabioLuporini:
Confirmed: depends on #323

FabioLuporini:
With the latest rebase, this should have become Green

FabioLuporini:
Nope. Still a few minor issues. Will take care tomorrow...

mlange05:
Just had another look through the recent changes, and I'm still happy. Good to go.

FabioLuporini:
Merged.

PR #323, opened 2017-08-04
Yet another symbolic cache bug, plus potential fix

FabioLuporini:
We should reopen #225 

@navjotk : you're gonna love this

EDIT: added potential fix


FabioLuporini:
Merged.

PR #325, opened 2017-08-09
Small fixes for Python3 environments

mlange05:
1) Pre-load `libc` in `memory.py` to avoid race-conditions on unclean shutdowns in Python3, eg.:
```
ImportError: sys.meta_path is None, Python is likely shutting down
```
2) Use `filter_ordered` instead of `list(set(...))` during parameter derivation in `operator.py` (the former is deterministic, the latter is not).

### DIFF
--- a/devito/operator.py
+++ b/devito/operator.py
@@ -118,10 +118,10 @@ class Operator(Function):
         self.symbolic_data = [i for i in parameters if isinstance(i, SymbolicData)]
         dims = [i for i in parameters if isinstance(i, Dimension)]
         dims_parents = [d.parent for d in dims if d.is_Buffered]
-        self.dims = list(set(dims + dims_parents))
+        self.dims = filter_ordered(dims + dims_parents)
         assert(all(isinstance(i, ArgumentProvider) for i in self.symbolic_data +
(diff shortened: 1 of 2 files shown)


devito/operator.py
```diff
@@ -118,10 +118,10 @@ def __init__(self, expressions, **kwargs):
         self.symbolic_data = [i for i in parameters if isinstance(i, SymbolicData)]
         dims = [i for i in parameters if isinstance(i, Dimension)]
         dims_parents = [d.parent for d in dims if d.is_Buffered]
-        self.dims = list(set(dims + dims_parents))
+        self.dims = filter_ordered(dims + dims_parents)
```
FabioLuporini: is this exactly the same as ``dimensions`` at line 99?

FabioLuporini:
yes, go ahead

devito/operator.py
mlange05: Yeah, that entire derivation logic seems to have redundancies. I'll have a look in a separate branch.

mlange05:
Ok, merging now, as it fixes non-deterministic travis failures in other branches.

PR #327, opened 2017-08-10
Ricker and Gabor wavelet source utility

mlange05:
Not yet ready, but filing now to request input from the seismic experts for rigorous docstring documentation.

This PR adds a new utility class to our seismic example toolkit that allows us to define sources and source geometries more concisely. The idea is to be able to create pure sources by simply giving the paramaters (frequency, etc.) instead of having `def source` or `def ricker` functions lying around all over the place. It also comes with a handy `.show()` functionality to quickly verify the source signal.

Additional fixes/cleanups:
- Rename `setup()` -> `acoustic_setup()`; this was cauasing unwanted namespace clashes with pytest, causing unwanted dummy allocations before test runs.
- Use devito's logger for displaying test norms, etc.
- `Model` provides `.domain_size` utility, so we don;t have to derive it all over the place
- Define source and receiver coordinates in-place, instead of pre-defining and copying into devito memory.
- **EDIT: Also integrate new src/rec approach with modelling tutorial and show the wavelet.**
- **EDIT: Add simple option processing for standalone acoustic examples.**
- **EDIT: Added Gabor wavelet utility and integrated with TTI examples.**


examples/seismic/source.py
```diff
@@ -34,15 +36,67 @@ def __new__(cls, name, ntime=None, npoint=None, ndim=None,
...
+        if not self._cached():
+            super(RickerSource, self).__init__(*args, **kwargs)
+
+    def wavelet(self, f0, t):
+        """
+        Create Ricker wavelet with a peak frequency f0 at time t.
+
+        :param f0: Peak frequency
+        :param t: Discretized values of time
+        """
```
mlange05: @mloubout Can you please provide more appropriate dostrings for these? Also, how does our implementation differ from more general ones? Should things like wavelength or amplitude be configurable?

tests/test_adjointA.py
```diff
@@ -17,16 +18,16 @@ def test_acoustic(dimensions, time_order, space_order, fix_dim):
...
     # Run forward and adjoint operators
     rec, _, _ = solver.forward(save=False)
     solver.adjoint(rec=rec, srca=srca)
 
     # Actual adjoint test
     term1 = np.dot(srca.data.reshape(-1), solver.source.data)
     term2 = linalg.norm(rec.data) ** 2
-    print(term1, term2, ("%12.12f") % (term1 - term2), term1 / term2)
+    info("L1 norm: %f, L2 norm: %f, L1-L2: %12.12f, L1/L2: %f"
+         % (term1, term2, term1 - term2, term1 / term2))
```
mlange05: Are these terms correct, or can we do better? Why are we printing these anyway? Some form of documentation would be great here.

mloubout: L1, bad naming, sounds like L1 norm

mloubout: Printed for debug

mlange05: Well what is it then? And why do you need it for debug? Either it's a useful diagnostic, in which case we should document it (at least in place), or it's pure developer-debug level, in which case it should not be in master.

examples/seismic/acoustic/acoustic_example.py
```diff
@@ -20,67 +20,52 @@ def smooth10(vel, shape):
...
+    nt = int(1 + (tn-t0) / dt)
+    time = np.linspace(t0, tn, nt)
 
-    # Set up the source as Ricker wavelet for f0
-    def source(t, f0):
-        r = (np.pi * f0 * (t - 1./f0))
-        return (1-2.*r**2)*np.exp(-r**2)
+    # Define source geometry (center of domain, just below surface)
+    src = RickerSource(name='src', ndim=ndim, f0=0.01, time=time)
+    src.coordinates.data[0, :] = np.array(model.domain_size) * .5
```
mloubout: WHy is the x coordinate multiple values?
np.array(model.domain_size) git the dimension in all direction right?

examples/seismic/source.py
mloubout: >wavelength
Ricker wavelet is defined by its peak frequency and time axis, wavelength is automatic from that (1/f)

Not sure what you mean by more appropriate, this are the conventional names of this variables. AMplitude could be added yes.

tests/test_adjointA.py
mloubout: >Either it's a useful diagnostic

It is, allows  to see the values didn't change between two runs.

examples/seismic/acoustic/acoustic_example.py
mlange05: Overwritten in the next line; first put everything in the center then adjust the dimension we care about.

tests/test_adjointA.py
mlange05: That is neither useful documentation, nor a good reason to keep stuff in master. What do these values mean, and why should I care?

mloubout: Well remove if you want, let's just hope we won;t go back to what happen last time we did it, which is, all pass with BS values.

mlange05: As discussed offline, I don;t want to remove it, I just want a more informative logs tring.

examples/seismic/source.py
mlange05: Ok, fair enough. The source definition in the TTI examples seems to be slightly different though. Can we generalise this so that we can create both with the same source object, but different parameters?

mloubout: Yes the source in TTI has a narower bandwith to avoid dispesion, but it can be replace byt the Ricker one without any problem.

mlange05:
@mloubout Ok, I generalised the wavelet preset utilities to include a `GaborSource` utility as well, and integrated this with the TTI example. With this in place, I think we are good to go. Can you please take one last look and confirm you are happy?

tests/test_adjointA.py
```diff
@@ -18,18 +19,18 @@ def test_acoustic(dimensions, time_order, space_order, fix_dim):
...
-        time.size = None
+    # Set fixed ("baked-in") time dimension if requested
+    time.size = solver.source.nt if fix_dim else None
+
     # Run forward and adjoint operators
     rec, _, _ = solver.forward(save=False)
     solver.adjoint(rec=rec, srca=srca)
 
-    # Actual adjoint test
+    # Adjoint test: Verify <Ax,y> matches  <x, A^T, y> closely
```
mloubout: <x, A^Ty>

My bad, typo here.

tests/test_adjointA.py
```diff
@@ -18,18 +19,18 @@ def test_acoustic(dimensions, time_order, space_order, fix_dim):
...
     term1 = np.dot(srca.data.reshape(-1), solver.source.data)
     term2 = linalg.norm(rec.data) ** 2
-    print(term1, term2, ("%12.12f") % (term1 - term2), term1 / term2)
+    info('<Ax,y>: %f, <x, A^T, y>: %f, difference: %12.12f, ratio: %f'
```
mloubout: same

mlange05:
Ok, typos fixed and all else green. Merging...

Issue #318, opened 2017-08-01
Messy parameters derivation in operator.py


FabioLuporini:
*** Some of the following may be debatable or I may be missing something ***

``parameters`` computed twice in ``Operator.__init__``. 

Also, for ``self.symbolic_data``, we could do the following:
- we already have a ``self.output`` attribute; we could then add ``self.input``
- infer ``self.symbolic_data`` from these two
This seems more natural to me

~Finally, we're doing: 
``parameters = list(set(parameters + dims_parents))``
but this breaks deterministic ordering (because of ``set``), which is a problem in the case of MPI execution and operator caching (though these features are still unused)~

~I'm temporarily marking this as a bug because of this last item~
EDIT: fixed

FabioLuporini:
Reminder: `operator._extra_arguments` can/should be dropped. All arguments should naturally end up in `parameters`

FabioLuporini:
Reminder: some methods suffixed with ``args`` other with ``arguments``

FabioLuporini:
Reminder: in the constructor, we're computing both ``dimensions`` and a bit later ``dims``, which should just be the same

FabioLuporini:
@mlange05 BTW, if you really find the time to revisit the derivation logic in ``operator.py.__init__``, here's a list of known things to look at. Ignore the first comment though ("Reminder: ``operator._extra_arguments`` can/should be ...")

navjotk:
Can we have an updated, definitive TODO list here?

FabioLuporini:
I'd say let's start with avoiding computing same information twice (look up ``parameters``, ``dimensions``, ``dims``, ...etc)

FabioLuporini:
I don't have a definitive tODO list -- I've been using this issue to keep track of the things that we need to do, in random order

navjotk:
> though these features are still unused

Non-existent would be more accurate?

> I'm temporarily marking this as a bug because of this last item

I believe this is fixed so this issue shouldn't be a bug any more. 

I'll wait for you to get an updated, definitive TODO on this before I get started specifically on this issue.

FabioLuporini:
> I believe this is fixed so this issue shouldn't be a bug any more.

Yes, changed to enhancement

I think I provided enough information here to understand what needs to/could be done. I can otherwise do this once I'll be back from holidays.

FabioLuporini:
I'm quite happy now about this, so closing ( @navjotk )

PR #329, opened 2017-08-11
Misc clean up, necessary for more Yask integration

FabioLuporini:
Many Operator-related improvements, with some refactoring.

Also extends the hierarchy in ``interfaces.py``


FabioLuporini:
@navjotk here I centralised arguments code generation inside ``nodes.Function``, which to me makes much more sense than having code generation methods attached to ``Arguments``.
In the future we may think of creating an actual tree visitor to do code generation (low priority now)

devito/arguments.py
```diff
@@ -79,10 +74,6 @@ def __init__(self, name, provider, reducer, default_value=None):
         super(ScalarArgument, self).__init__(name, provider, default_value)
         self.reducer = reducer
 
-    @property
-    def decl(self):
-        return c.Value('const %s' % c.dtype_to_ctype(self.dtype), self.name)
-
     def verify(self, value):
```
navjotk: I don't understand why `decl` is being removed here.

devito/arguments.py
```diff
@@ -186,14 +177,6 @@ def rtargs(self):
...
 
     @property
-    def ccode(self):
-        """C-level variable name of this dimension"""
-        if self.size is not None:
-            return "%d" % self.size
-        else:
-            return self.rtargs[0].ccode
-
-    @property
```
navjotk: Why is this being removed?

devito/arguments.py
```diff
@@ -102,23 +93,7 @@ class TensorArgument(Argument):
...
-    def ccast(self):
-        alignment = "__attribute__((aligned(64)))"
-        shape = ''.join(["[%s]" % ccode(i) for i in self.provider.symbolic_shape[1:]])
-
-        cast = c.Initializer(c.POD(self.dtype,
-                                   '(*restrict %s)%s %s' % (self.name, shape, alignment)),
-                             '(%s (*)%s) %s' % (c.dtype_to_ctype(self.dtype),
-                                                shape, '%s_vec' % self.name))
-        return cast
+        super(TensorArgument, self).__init__(name, provider, provider)
```
navjotk: Same. Why is this being removed?

devito/dimension.py
```diff
@@ -31,10 +33,10 @@ def __str__(self):
...
     def symbolic_size(self):
         """The symbolic size of this dimension."""
-        try:
-            return Number(self.ccode)
-        except ValueError:
-            return Symbol(self.ccode)
+        if self.size is not None:
+            return Number(self.size)
+        else:
+            return Symbol('%s%s' % (self.name, Dimension._size_suffix))
```
navjotk: This seems to be undoing the work I did with arguments.

devito/nodes.py
```diff
@@ -547,16 +549,51 @@ def __repr__(self):
...
+        for i in self.parameters:
+            if i.is_ScalarArgument:
+                argtypes.append(numpy_to_ctypes(i.dtype))
+            elif i.is_TensorArgument:
+                argtypes.append(ctypeslib.ndpointer(dtype=i.dtype, flags='C'))
+            else:
+                argtypes.append(ctypes.c_void_p)
+        return argtypes
+
+    @property
```
navjotk: Very ugly hacky method. Should be moved inside each argument

devito/nodes.py
```diff
@@ -547,16 +549,51 @@ def __repr__(self):
...
+        decls = []
+        for i in self.parameters:
+            if i.is_ScalarArgument:
+                decls.append(c.Value('const %s' % c.dtype_to_ctype(i.dtype), i.name))
+            elif i.is_TensorArgument:
+                decls.append(c.Value(c.dtype_to_ctype(i.dtype),
+                                     '*restrict %s_vec' % i.name))
+            else:
+                decls.append(c.Value('void', '*_%s' % i.name))
+        return decls
```
navjotk: Same as above. Undoes the cleanup I did earlier and brings us back to this ugly hacky if-else mess.

devito/nodes.py
```diff
@@ -547,16 +549,51 @@ def __repr__(self):
...
+                lvalue = c.POD(i.dtype, '(*restrict %s)%s %s' % (i.name, shape, align))
+                rvalue = '(%s (*)%s) %s' % (c.dtype_to_ctype(i.dtype), shape,
+                                            '%s_vec' % i.name)
+                ccasts.append(c.Initializer(lvalue, rvalue))
+            elif i.is_PtrArgument:
+                ctype = ctypes_to_C(i.dtype)
+                lvalue = c.Pointer(c.Value(ctype, i.name))
+                rvalue = '(%s*) %s' % (ctype, '_%s' % i.name)
+                ccasts.append(c.Initializer(lvalue, rvalue))
+        return ccasts
```
navjotk: Same as above. This is not a cleanup. Quite the contrary.

mlange05:
@navjotk Ok, I think I can provide some context here from previous discussions. The problem with the current argument derivation hierarchy is that it breaks abstraction and encapsulation, because it contains a whole bunch of _compile-time_ properties (types, declarations, etc.), despite encapsulating _runtime_ behaviour. I believe this PR is trying to decouple the codegen-specific things from the runtime argument hierarchy, because these things will eventually need to be sub-classed for different compiler backends (eg. YASK). The intent here, I believe, is to make the argument derivation engine agnostic to the compiler backend, so as to avoid having to have YASK-specific argument types. As such, I think this is actually much cleaner, since it increases the overall degree of encapsulation and draws rigorous lines between our internal abstractions - which is ultimately the goal of object-orientation.

Admittedly, these discussions would benefit from clearer developer-level documentation of the internal hierarchies, but they are still very much in flux.

FabioLuporini:
@mlange05 is right about the rationale behind this.

Moving everything inside the node hierarchy is a first step towards building a tree visitor doing code generation. Once the tree visitor will be ready, we won't even have those ``if/then/else``.

BTW, we can argue about lots of things, like these edits making the structure more or less clean, but definitely this is not hacky.

FabioLuporini:
**UPDATE**

Before this PR, code generation was annoyingly distributed over multiple modules: ``operator.py``, the ``Argument`` classes in ``arguments.py``, the node classes in ``nodes.py``. Now, this has been fixed. There's a new tree visitor responsible for code generation. It basically aggregates the ``ccode``, ``cdecl``, ``ctop``, ... . I added a test for this visitor.

The second update is that I refactored the crazy messy ``Operator`` initialization, which fixes issue #318.

FabioLuporini:
@navjotk this needs another round of reviewing

devito/dimension.py
navjotk: This should be using the `rtargs` framework.

devito/operator.py
```diff
@@ -165,17 +145,17 @@ def arguments(self, **kwargs):
...
-        for ta in self.tensor_args:
-            assert(ta.verify(kwargs.pop(ta.name, None)))
-
-        for d in self.dims:
+        # Derivation
+        for i in self.parameters:
+            if i.is_TensorArgument:
+                assert(i.verify(kwargs.pop(i.name, None)))
+            else:
+                i.verify(kwargs.pop(i.name, None))
```
navjotk: This is changing the tree structure of the arguments derivation. Why is this required? Especially in this PR?

FabioLuporini: Can you provide an example of what gets broken here? As all tests are passing, hence probably there's a flaw in our testing infrastructure.

navjotk: The arguments derivation is a WIP, including tests. 

- You removed the tree structure by putting it all in one loop
- Scalar Arguments aren't being called any more
- Dimensions returned to the assumption that there is only one argument per dimension. Undoing my previous cleanup regarding this

FabioLuporini: Again, I need an example, or **better**, a test to catch what's breaking here, otherwise the only option is to go ahead and "fix" this in a second moment, when the test will be available

My concerns follow:
* ``self.parameters`` includes both tensor and scalar aguments. So ``ScalarArgument``s are actually being called (``else`` branch).
* I don't see how I could have altered the tree structure, since you also had several loops
* I don't understand the point about dimensions, as it seems the same loop to me ?

I totally understand the arguments stuff was WIP; but still, this can't be my fault if no tests caught this "issue" (?)

FabioLuporini:
Addressed @navjotk 's comments after offline discussion

FabioLuporini:
Merging...

FabioLuporini:
Merged.

devito/operator.py
```diff
@@ -444,13 +406,21 @@ def _retrieve_stencils(self, expressions):
...
-    @property
-    def _cparameters(self):
-        return super(Operator, self)._cparameters
+    def _retrieve_symbols(self, expressions):
+        """
+        Retrieve the symbolic functions read or written by the Operator,
+        as well as all traversed dimensions.
+        """
+        input = flatten(terminals(i) for i in expressions)
+        input = filter_sorted([i.base.function for i in input], key=attrgetter('name'))
```
mloubout: This breaks if there  is a `Dimension` in `terminals`

PR #332, opened 2017-08-16
Revive first touch

navjotk:
As mentioned [previously](https://github.com/opesci/devito/issues/330#issuecomment-322825716), our data initialisation is currently broken because the `a[:] = 0` syntax only seems to run single core. We are currently using this syntax to initialise all arrays before they are used. We should change to a different method of initialisation that uses our own JIT infrastructure. Luckily such a function existed - `first_touch` but it was broken since it was never updated in the shift from Devito 2.0 to 3.0. 

In this PR:
1. I changed the name of the parameter from `numa` to `first_touch`.
2. Updated the `first_touch` function so it becomes compatible with the Devito 3.0 API
3. Wrote a test for the function

The test currently fails at a line in [profiling.py](https://github.com/opesci/devito/blob/first_touch_again/devito/profiling.py#L40). Once this is fixed, I believe this might be ready for merge. 

Note that this PR only revives a broken functionality. It doesn't change any default behaviour. To use first touch, you would still have to pass the `first_touch` parameter as in the test. This PR will enable performance comparisons between the two methods.


devito/memory.py
```diff
@@ -11,7 +11,7 @@
 from devito.dimension import t
 from devito.logger import error
 from devito.tools import convert_dtype_to_ctype
-
+import devito
```
FabioLuporini: question: symptom of circular dependency?

navjotk: Having a `from devito import ....` here would introduce a circular dependency, yes.

navjotk:
> This should be wired in with our configuration system.

Considering this isn't even functional until the issue with `profiling.py` is fixed, how this can be configured is not the most relevant thing at this point. Do you have anything to say on how to fix the broken profiler instead?

mlange05:
@navjotk I pushed a fix to the problem you pointed out, which should fix the test.

I agree that this should be wired in with the parameters dictionary, so that we can switch it on universally
(as originally intended). We might also want to add that to the OpenMP travis builder then (or even have a separate one; might be overkill though).

I also note that there might also be a problem with re-compilation, since we don't do full symbolic operator caching yet, although this specific operator will be called a lot. Not part of this PR though - let's see how this plays out in benchmarks first.

tests/test_memory.py
```diff
@@ -0,0 +1,9 @@
+from devito import DenseData
+import pytest
+import numpy as np
+
+
+@pytest.mark.parametrize('shape', [(20, 20), (20, 20, 20), (20, 20, 20, 20)])
+def test_first_touch(shape):
```
FabioLuporini: I think a better test is to do something similar to ``test_cache_blocking_structure``, in which I search the AST for the presence of stuff like openmp pragmas etc.

devito/memory.py
```diff
@@ -8,10 +8,9 @@
 import numpy as np
 from sympy import Eq
 
-from devito.dimension import t
 from devito.logger import error
 from devito.tools import convert_dtype_to_ctype
-
+import devito
```
FabioLuporini: Any idea of how we could avoid the circular dependencies that we would have with ``from devito import Operator`` ?

I think this is a symptom that we should do things differently

devito/interfaces.py
```diff
@@ -24,6 +24,16 @@
 # manipulation with the correct shapes, pointers, etc.
 _SymbolCache = {}
 
+# This global stores whether newly allocated memory should be initialised
+# using Devito's own JIT engine. It will otherwise be initialised using
+# numpy.
+Global_First_Touch = None
```
FabioLuporini: This is already in ``configuration``, why having a global here would be any useful?

devito/interfaces.py
```diff
@@ -24,6 +24,16 @@
...
+
+def set_global_first_touch(flag):
```
FabioLuporini: perhaps unnecessary? or where is it called?

devito/interfaces.py
```diff
@@ -47,6 +57,8 @@ def __repr__(self):
 Backward = TimeAxis('Backward')
 
 
+
```
FabioLuporini: leftover

devito/interfaces.py
```diff
@@ -402,7 +414,10 @@ def __init__(self, *args, **kwargs):
             self.initializer = kwargs.get('initializer', None)
             if self.initializer is not None:
                 assert(callable(self.initializer))
-            self.numa = kwargs.get('numa', False)
+            first_touch = kwargs.get('first_touch', None)
+            if first_touch is None:
+                first_touch = Global_First_Touch
+            self.first_touch = first_touch
```
FabioLuporini: these 4 lines could probably be rewritten as:
``self.first_touch = kwargs.get('first_touch', Global_First_Touch)``

devito/interfaces.py
navjotk: In [parameters](https://github.com/opesci/devito/blob/451849e46a50114215e875680a262665fdceddd2/devito/parameters.py#L153), as is the convention for the other configuration parameters.

devito/interfaces.py
navjotk: Fixed

devito/interfaces.py
navjotk: Fixed.

devito/interfaces.py
mlange05: I don't think this has anything to do with convention. The callback mechanism for the `log_level` is required to update the logger object if the value is changed at runtime. Since the FT variable is completely read-only and we're not "undoing" first touch on anything previously allocated, I don't see the need for a callback mechanism here. The only thing it does is some minor type-checking; is this really required here, or should this be part of the `Parameter` class?

navjotk: If I read the parameters object directly in `interfaces.py`, it would be like `configuration['first_touch']`, i.e. the magic string "first_touch" would be used in `interfaces.py` as well as `parameters.py`. This way, the string stays confined to a single place.

devito/memory.py
navjotk: `operator.py` imports `SymbolicData`, `CompositeData`, `Forward` and `Backward` from `interfaces.py`. I don't think we can remove all 4 of these dependencies. 

`interfaces.py` imports `CMemory` and `first_touch` from `memory.py`. These seem pretty obvious dependencies to me. 

If we do a `from devito import Operator` to initialise our own arrays (and why shouldn't we?), we introduce a dependency from `memory.py` to `operator.py` and hence completing the circle. 

I don't see any way to restructure this that would not introduce a circular dependency. Any ideas?

tests/test_memory.py
navjotk: The `first_touch` function does not manipulate ASTs, unlike the DLE method for cache blocking. `first_touch` is a one-line unit that only deals in `Operator`s and arrays - which is being tested by this unit test (see updated test). I see no reason to complicate the function for this reason alone.

devito/memory.py
mlange05: Yes, so I think this clearly indicates that we are doing way too much in one super-generic `interfaces` submodule. If we are serious about separating the various data type hierarchies for codegen, arguments and symbolics, we should refactor them into their own submodules, so that the operator only depends on the relevant interfaces, while the functional implementations (`Time/DenseData` et al.) can happily import an operator.

Clearly this is beyond this PR, but we should keep it in mind.

FabioLuporini:
> extra global parameter unnecessary

is it necessary again to avoid circular dependencies?

navjotk:
> is it necessary again to avoid circular dependencies?
It is unrelated to circular dependencies. Can you please update your review @FabioLuporini ?

FabioLuporini:
> Can you please update your review @FabioLuporini ?

Yes, happy now

devito/memory.py
FabioLuporini: This is useful information, perhaps it's useful to file an issue with exactly this description to remember about these circular dependencies

navjotk:
Shall merge when green

navjotk:
> is it necessary again to avoid circular dependencies?

I didn't know this before but now I see what you mean :)

FabioLuporini:
> I didn't know this before but now I see what you mean :)
> 
> Edit: I tried to do without the global but that leads to a different circular dependency. I believe no one (including myself) has a strong opinion for or against the global approach so I will still merge it with the global approach unless someone objects before this goes green.

Fine. In the end, we use a similar patter with ``logger`` and ``compiler``, so... that's acceptable. ``False`` is probably a better init value than ``None``, though.

I think we'll have to untangle this dependency mess by moving everything inside ``parameters``. This should be fairly trivial with the dse and dle registries thanks to the latest rewrites. IOW, I *think* ``parameters.py`` should be independent of the other Devito modules, so that we can import ``configuration`` from any module. Not something to do in this PR if you don't have time, of course; but probably we need to keep track of it through an issue...

navjotk:
Issue created. Travis green. Merging.

PR #333, opened 2017-08-17
Seismic: Fix "Explicit time dimension" warning for forward wavefields

mlange05:
Should be fairly uncontroversial and fix the flood of "Explicit time dimension" warnings generated by our seismic wavesolver wrappers for forward operators.


mlange05:
Ok, merging...

PR #336, opened 2017-08-30
Remove critical circular dependencies

### LINKED ISSUE #334: Interfaces.py is too prone to circular dependencies
While working on #332 we realised that there are several scenarios that lead to circular dependencies which can be traced back to how bulky `interfaces.py` has become. This file should (probably) be broken down into smaller parts so that the parts can individually be imported without creating circular dependencies.

FabioLuporini:
fixes #334 

This makes it possible to access ``configuration`` from basically everywhere in Devito, w/o having to use local imports


FabioLuporini:
Merging...

FabioLuporini:
Merged.

PR #338, opened 2017-09-04
Yask backend now capable of jitting and executing

FabioLuporini:
Depends on (and currently includes!)  #329


devito/interfaces.py
```diff
@@ -357,6 +362,17 @@ class SymbolicData(AbstractSymbol):
...
+    def _data_buffer(self):
+        """Reference to the actual data. This is *not* a view of the data.
+        This method is for internal use only."""
+        return self.data
+
+    @abc.abstractproperty
+    def data(self):
+        """The value of the data object."""
+        return
+
```
navjotk: What is going on here?

FabioLuporini: Python's abstract methods are supposed to return None (see for example https://stackoverflow.com/questions/4382945/abstract-methods-in-python)

This ensures that all derived (concrete) classes will have an implementation for ``data``

FabioLuporini:
all clear @navjotk ?

FabioLuporini:
Merging (offline discussion cleared this out)...

FabioLuporini:
Merged.

Issue #337, opened 2017-09-03
Interface needs cleanup and robustness rewrite


mloubout:
To make it simple, we currently consider that any dimension except the first one is a spatial dimension (in `TimeData`) which make it a pain if you add a dimension. e.g if you solve multiple PDE at once in 2D you have `u[t,x,y,p_src]` but Devito interpret it as a 3D field and tries to take `u.dz2` in the laplacian.

This one should be "easy" as long as we consoder only 2 or 3D problems. If not, I need to reqrite `Dimension` to seperate spatial, temporal and local ones.

chuckyount:
FYI, I had this problem in YASK. In the new version of YASK under development, I'm requiring the user to more explicitly create each dimension as 'step' (i,e., temporal), 'domain' (i.e., spatial), or 'misc' (i.e., local). This allows me to have grids like you have mention here, and mix 3d, 2d, and 1d grids in the same problem, like when using Cerjan arrays for ABC.

FabioLuporini:
> if you solve multiple PDE at once in 2D you have u[t,x,y,p_src]

Can you please elaborate? What value is the notation ``u[t,x,y,p_src]`` trying to access?

mloubout:
>What value is the notation u[t,x,y,p_src] trying to access?

The wavefield for the source number `p_src`, instead of looping over sources in python you solve all at once

mlange05:
Yes, the new auto-generation of derivative shortcuts should allow this. It's currently ownly restricted to space, but I'm working on time.

mloubout:
PointData is hard-coded as time-dependent, and non-compatible with non-conventional dimensions

mloubout:
Open #374 for the pointdata part, this one can be closed I think wiht #369

FabioLuporini:
Is this still an issue @mlange05  @mloubout

mloubout:
Yes, I need to rewrite the "time free" case for the new devito.

mloubout:
FIxed, closing.

PR #340, opened 2017-09-05
Interface robustness improvments

mloubout:
Very small PR that removes the assumption on the order of the dimensions in `TimeData` and `DenseData` thagt was hard-coded to `t,x,y,z`


mloubout:
This fixes part of the problems of #337

devito/dse/inspection.py
```diff
@@ -21,7 +21,10 @@ def terminals(expr, discard_indexed=False):
 
     # Use '.name' for quickly checking uniqueness
     junk = flatten([i.free_symbols for i in indexed])
-    junk = [i.name for i in junk]
+    junk = [i.name for i in junk] + \
+           [i.parent.name if
+            (i.is_Buffered if hasattr(i, "is_Buffered") else False)
```
FabioLuporini: ``... + [i.parent.name for i in junk if isinstance(i, BufferedDimension)]``

mloubout: Thanks

Just not sure if this should be here. I hacked it because it work but seems odd and out of place.

FabioLuporini:
OK this needs to be rebased on latest master (many things can now be dropped, I think, as part of the latest bug fixes in master). Then we can start thinking about how to improve it

mloubout:
As agreed with @FabioLuporini I will put back mandatory time as first dimension as the loop scheduling is too messy to handle with random dimensions ordering

mlange05:
Ok, for discussion's sake, I'm not a fan of hard-coding all the "space dimensions" specifically either, but as long as we have no clear abstractions to separate Time/Space/Misc dimensions there is little we can do. This sort of separation has a long dependency chain as well though, since it directly affects how we sanity-check and default things should we ever build a generic "grid" abstraction. In fact, I think the two features are very tightly coupled, since a grid can only ever provide/check/know about "space dimensions", while things like time and misc should be optional additions for individual fields. Just my 2 cents' worth though...

tests/test_operator.py
```diff
@@ -569,7 +569,7 @@ def test_equations_mixed_densedata_timedata(self, shape, dimensions):
         assert len(trees) == 2
         assert all(trees[0][i] is trees[1][i] for i in range(3))
 
-        # Verify both operators produce the same result
+        # Verify both operators produce the same resultsx
```
FabioLuporini: typo

devito/interfaces.py
```diff
@@ -694,13 +693,24 @@ def __init__(self, *args, **kwargs):
                             'dimension (save=False). This value will be ignored!'
                             % self.name)
                 time_dim = self.time_order + 1
-                self.indices[0].modulo = time_dim
+                # Get time index
+                _t = [var for var in self.indices if var in (t, time)][0]
```
FabioLuporini: can we assume this is the outermost left dimension or not?

tests/test_operator.py
mloubout: Darn

FabioLuporini:
> which is why there is a new test.

I cannot find it

devito/interfaces.py
mloubout: For our purpose yes, but I think it is reductive to inforce it, so I left it as a possibility to put the dimensions in any order with a warning stating you need to know what you are doing, as in some case, the order may not matter

FabioLuporini: One seriously big (purely technical) issue is that all those ``if t/time is not indices/dimensions then ...``  make the code a mess

mloubout:
>This PR is touching a piece of code that was already quite awful

Let's make it clear, this piece of code is not supposed to be pretty CS, it is supposed to be the math link between Devito and whatever people want to do with it. 

> What features are you exactly trying to get

Not enforcing our specific setup to any PDE people may want to solve

>WHy do you need a time dimension different than t/time

I don't there is just no way to know wether it's `time` or `t`, we can force it to be one or the other

>this needs to be rebased on master

It;s like 5 times less commits than most of you guys PR, common

FabioLuporini:
> it is supposed to be the math link between Devito and whatever people want to do with it

so it's even more important

> Not enforcing our specific setup to any PDE people may want to solve

Concrete examples would be useful so that I can understand what generalisation you are looking for. Is it just to allow people to use dims different than ``x/y/z`` ?

FabioLuporini:
> It;s like 5 times less commits than most of you guys PR, common

It's not that. Being on top of master makes the ``diff`` simpler to read

devito/interfaces.py
```diff
@@ -716,10 +726,16 @@ def _indices(cls, **kwargs):
...
-        tidx = time if save else t
-        _indices = DenseData._indices(**kwargs)
-        return tuple([tidx] + list(_indices))
+        # Get dimensions
+        dimensions = kwargs.get('dimensions', None)
+        # Add time index if not specified
+        if dimensions is None or (t not in dimensions and time not in dimensions):
+            save = kwargs.get('save', None)
+            tidx = time if save else t
+            _indices = DenseData._indices(**kwargs)
```
FabioLuporini: does ``super(TimeData, self)._indices(**kwargs)`` work ?

mloubout: This was there before, shows as diff because of the if

mloubout:
>Is it just to allow people to use dims different than x/y/z
no just to allow any order including time

>so it's even more important
It's important to be readable for non-CS people, not fancy

mloubout:
>It's not that. Being on top of master makes the diff simpler to read

only diff is `interfaces` and the laplacian in the test, waste of time for the exact same diff

FabioLuporini:
> It's important to be readable for non-CS people, not fancy

Yeah, I should stop making observations only to make the code fancier :)

FabioLuporini:
@mloubout and I discussed about this offline and we agreed this needs first a wider discussion with the rest of the development team to be done properly

mloubout:
Needs a fresh start from recent updates, closing.

PR #341, opened 2017-09-06
Fix loop scheduling

FabioLuporini:
Fixes and improves the loop ordering construction, which is in turn used to build the AST (thus impacting the resulting loop tree structure).

A test is added.

@mloubout Can you try this please?


FabioLuporini:
OK, approved by @mlange05 , and @mloubout is happy. So merging...

FabioLuporini:
Merged.

PR #342, opened 2017-09-06
Separate the case where Dimensions have a known size into its own class

navjotk:
The cases where the size of a `Dimension` is fixed beforehand are rare enough, and different enough from the case of open-size `Dimension`s to warrant a different class for `FixedDimension`s. 
This reduces the complexity of the arguments derivation etc. and eases further extension.

Edit: Note that this is PR 1/3 on the way to checkpointing.


mloubout:
Not sure i understand why we need this. It seems to make high-level usage of devito more intrcated

navjotk:
It removes a lot of nested "if-else"s which makes it easier to reason about and build more complex stuff. The argument derivation logic becomes more complex (and more separated between the two cases) as I introduce 3 arguments per dimension in my next PR.
As I understand it, this feature is not really being used and was largely included as an experiment @mlange05 wanted to conduct to see which way is faster (baking in loop limits or keeping them in). The feature remains usable but is moved out of the way with this PR. 

If there is real need for the feature and we wish to go back to the previous user-facing API, we could still figure out some dynamic-dispatch voodoo to make it look the same at the higher level.

mloubout:
>If there is real need for the feature and we wish to go back to the previous user-facing API

Wouldn't it be better to benchmark and test on this branch, and if really necessary then add it to master. The user-facing API is what every user uses and the main point of Devito is that it is easy to use, if we start making the interface complicated so  that we can do some benchmarking on our own, this is gonna hurt us

mlange05:
Ok, a few nuggets of context:

> It seems to make high-level usage of devito more intrcated

I'm not sure "intricate" is the right word, but it definitely makes changing the nature of a dimension more "explicit", which IMHO is a good thing.

> As I understand it, this feature is not really being used and was largely included as an experiment @mlange05 wanted to conduct to see which way is faster (baking in loop limits or keeping them in)

That sounds fairly wrong - the feature of "baked in" dimension sizes in was _kept_ not _added_ for backward compatibility, as "baking in" was the only _modus operandi_ before the introduction of dimensions and the 3.0 re-write. The (set of) experiment(s) of whether this matters should have been done a long time ago, and is probably somewhat problem and compiler-dependent. But alas, time is a precious commodity... :frowning: :hourglass: 

> Wouldn't it be better to benchmark and test on this branch, and if really necessary then add it to master.

Totally irrelevant, as the experiment is trivial to perform for a particular test cases in both branches. Nothing in this PR change the mechanism of "baking in", it only changes the argument verification and the API with which to achieve it - which is still is pretty much a one-liner in both cases.

tests/test_adjointJ.py
```diff
@@ -48,8 +48,8 @@ def test_acousticJ(dimensions, space_order):
 
     # Compute initial born perturbation from m - m0
     dm = model.m.data - model0.m.data
-    du, _, _, _ = solver.born(dm, m=model0.m)
 
+    du, _, _, _ = solver.born(dm, m=model0.m)
```
mlange05: Why is this line being touched?

tests/test_adjointA.py
```diff
@@ -58,7 +59,8 @@ def test_acoustic(mkey, dimensions, time_order, space_order, nbpml, fix_dim):
                                 space_order=space_order)
 
     # Set fixed ("baked-in") time dimension if requested
-    time.size = solver.source.nt if fix_dim else None
+    if fix_dim:
+        devito.time = FixedDimension('time', size=solver.source.nt)
```
mlange05: Hmmm, not sure I completely like this way of overriding the global time dimension, but at least it is (painfully) explicit what is happening. I'd be interested in what "dynamic dispatch voodoo" was envisioned here, but it's ok for now IMHO.

devito/dimension.py
```diff
@@ -31,10 +30,24 @@ def __str__(self):
...
-            return self.rtargs[0].as_symbol
+        return self.rtargs[0].as_symbol
+
+
+class FixedDimension(FixedDimensionArgProvider, Dimension):
+
+    is_Fixed = True
+    """This class defines the behaviour of a dimension whose size is fixed
+       at the time of problem definition and can thus be baked into generated
+       code """
```
mlange05: Spacing seems off here: `"""` should be in next line with a blank before `__init__`.

devito/arguments.py
```diff
@@ -79,6 +78,10 @@ def __init__(self, name, provider, reducer, default_value=None):
         super(ScalarArgument, self).__init__(name, provider, default_value)
         self.reducer = reducer
 
+    @property
+    def as_symbol(self):
+        return Symbol(self.name)
```
FabioLuporini: This is ``master`` already, so this needs rebasing?

devito/arguments.py
```diff
@@ -175,51 +212,41 @@ def dtype(self):
...
-            return []
-        else:
-            size = ScalarArgument("%s_size" % self.name, self, max)
-            return [size]
-
-    @property
-    def decl(self):
-        return self.rtargs[0].decl
+        size = ScalarArgument("%s_size" % self.name, self, max)
+        return [size]
```
FabioLuporini: inlinable

devito/arguments.py
```diff
@@ -175,51 +212,41 @@ def dtype(self):
...
     # TODO: Can we do without a verify on a dimension?
     def verify(self, value):
         verify = True
+        if value is None:
+            if self.value is not None:
+                return True
 
-        if value is None and self._value is not None:
-            return verify
+            try:
```
FabioLuporini: This try-catch is probably a symptom that it'd be better to have a ``BufferedDimensionArgProvider``. If that's the case, then we could also have a superclass for these DimensionArgProvider classes

devito/arguments.py
```diff
@@ -175,51 +212,41 @@ def dtype(self):
...
+                value = self.parent.value
+                if value is None:
+                    return False
+            except AttributeError:
+                return False
 
-        if value is not None and value == self._value:
-            return verify
+        try:
```
FabioLuporini: exactly the same point as above

devito/arguments.py
```diff
@@ -175,51 +212,41 @@ def dtype(self):
...
+            parent_value = self.parent.value
```
FabioLuporini: naming inconsistency: called "value" above, now "parent_value"

devito/arguments.py
```diff
@@ -175,51 +212,41 @@ def dtype(self):
...
+        if value == self.value:
+            return True
+
+        # Derived dimensions could be linked through constraints
+        # At this point, a constraint needs to be added that enforces
+        # dim_e - dim_s < SOME_MAX
+        # Also need a default constraint that dim_e > dim_s (or vice-versa)
+        verify = verify and all([a.verify(v) for a, v in zip(self.rtargs, (value,))])
+        if verify:
+            self._value = value
```
FabioLuporini: I still don't like the fact that we are changing state in a method called ``verify``. Is this changing any time soon ?

devito/dle/backends/advanced.py
```diff
@@ -162,7 +162,7 @@ def _loop_blocking(self, state, **kwargs):
                     # Build Iteration over blocks
                     dim = blocked.setdefault(i, Dimension("%s_block" % i.dim.name))
                     block_size = dim.symbolic_size
-                    iter_size = i.dim.size or i.dim.symbolic_size
+                    iter_size = i.dim.size if i.dim.is_Fixed else i.dim.symbolic_size
```
FabioLuporini: At this point I think having just ``i.dim.symbolic_size`` (i.e., with the ``if ... else``) should work. Could you try this?

devito/nodes.py
```diff
@@ -277,13 +277,6 @@ def __init__(self, nodes, dimension, limits, index=None, offsets=None,
         else:
             self.limits = list((0, limits, 1))
 
-        # Replace open limits with variables names
```
FabioLuporini: 👍

devito/operator.py
```diff
@@ -201,7 +201,9 @@ def _dle_arguments(self, dim_sizes):
         dle_arguments = OrderedDict()
         autotune = True
         for i in self.dle_arguments:
-            dim_size = dim_sizes.get(i.original_dim.name, i.original_dim.size)
+            dim_size = dim_sizes.get(i.original_dim.name,
+                                     i.original_dim.size if i.original_dim.is_Fixed
```
FabioLuporini: What do you think of adding a ``size`` property to the "main" ``Dimension`` class, returning ``None``, so that we don't need ``if ... else`` any time we check for the presence of a ``FixedDimension``?

devito/operator.py
```diff
@@ -295,7 +297,13 @@ def _schedule_expressions(self, clusters, ordering):
                 needed = entries[index:]
 
                 # Build and insert the required Iterations
-                iters = [Iteration([], j.dim, j.dim.size, offsets=j.ofs) for j in needed]
+                iters = []
+                for j in needed:
```
FabioLuporini: I think this should just work with a single list comprehension as before, using ``symbolic_size``, and w/o using any ``if ... else``. If it doesn't, can you point me to one failing example?

devito/profiling.py
```diff
@@ -143,10 +143,10 @@ def summary(self, dim_sizes, dtype):
             gflops = flops/10**9
 
             # Compulsory traffic
-            datashape = [i.dim.size or dim_sizes[dims[i].name] for i in itspace]
+            datashape = [i.dim.size if i.dim.is_Fixed
+                         else dim_sizes[dims[i].name] for i in itspace]
             dataspace = reduce(operator.mul, datashape)
             traffic = profile.memory*dataspace*dtype().itemsize
```
FabioLuporini: leave the blank line pls

tests/test_adjointJ.py
navjotk: Fixed (please confirm)

tests/test_adjointA.py
navjotk: Maybe an issue?

devito/arguments.py
navjotk: Fixed (confirm)

devito/arguments.py
navjotk: The next PR will extend this anyway. Do you really want it inline for a couple of days? :)

devito/arguments.py
navjotk: I agree but can we consider that outside the scope of this PR?

devito/arguments.py
navjotk: No I haven't worked on that but can we wait until all 3 PRs are in? (This is 1/3)

devito/dle/backends/advanced.py
navjotk: Do you mean without the `if...else`?

devito/operator.py
navjotk: Could that wait for a future cleanup or do you want it in this PR?

devito/operator.py
navjotk: This is changed again in the next PR. Goes back to being a one-liner but written in terms of limits. Can you wait for PR 2/3 before you come to a conclusion on what's the best way?

devito/dimension.py
navjotk: Fixed

devito/arguments.py
navjotk: Fixed

devito/profiling.py
navjotk: Fixed

devito/arguments.py
FabioLuporini: nope :)

devito/arguments.py
FabioLuporini: Alright

devito/arguments.py
FabioLuporini: Yes

devito/operator.py
FabioLuporini: Can wait

devito/operator.py
FabioLuporini: Unlike the previous ones, this bit of is of fundamental importance, so could you shed more light on this?

FabioLuporini:
I agree about postponing most of my comments, but:
- I wanna clean up at least the loop in ``_schedule_expressions``
- perhaps best if you add an issue to keep track of all these things

devito/operator.py
navjotk: Fixed

devito/operator.py
```diff
@@ -302,7 +304,7 @@ def _schedule_expressions(self, clusters):
                 needed = entries[index:]
 
                 # Build and insert the required Iterations
-                iters = [Iteration([], j.dim, j.dim.size, offsets=j.ofs) for j in needed]
+                iters = [Iteration([], j.dim, j.dim.symbolic_size, offsets=j.ofs) for j in needed]
```
FabioLuporini: OK now

FabioLuporini:
I think this is mergeable. Any last minute thought, @mlange05 ?

Issue #247, opened 2017-05-18
Add Devito features required for checkpointing


navjotk:
To introduce checkpointing, Devito requires some additional features.

Following from the last discussion, this is to document the changes that will be made in Devito to enable the use of checkpointing:

1. **Start and stop Operators at a given time-step**: Add optional parameters to `Operator.apply()` called `tstart` and `tend` to (re)start the Operator from a given time-step and run for a given number of time steps. 

2. **Store last accessed time step in data objects**: This is largely related to issue #69 where, when a `TimeData` object has a buffered time dimension, it is not clear which buffer holds the field computed at the last time-step. Currently we are working around this by externally tracking the number of time-steps that were run on the object. This should be internalised so that external code can reliably know how to access the last computed field. There was a proposal to do the same for `PointData` objects as well but I don't remember the argument for that - could potentially be discussed (again). 

3. **Deep copy routines inside symbolic data objects**: All symbolic data objects to have a `save` routine to save their state (data as well as the internal time counter added by the above, to start with) in a memory location provided by the caller. There would also be an accompanying `restore` method to restore state from a provided memory location. A corollary to this would be that a symbol should be aware of and be capable of reporting the amount of memory required to save its state. This knowledge would eventually find its way into the checkpointer which would be responsible for allocating memory to store the checkpoints.

navjotk:
I discussed point 1 with @FabioLuporini today and we agreed that the cleanest way to implement this would be to extend the parameter that is currently used to pass the number of timesteps, `t`. The current code assumes that the loop will run from `t=0` to `t=<provided value>`. We could extend that code to also accept 2-tuples and default to `(0, <provided value>)` in case a scalar is passed. (All this extends naturally to all dimensions)

This leads to my question: how does this relate to the `direction` parameter? We could either:
- Assume that the first element in the tuple represents the loop starting value and the second the end value. This makes the direction parameter redundant since reversing the order of this tuple would achieve reversing any loop. 
- Assume that the first element of the tuple is always less than the second and the direction of traversal in this space is provided by the direction parameter. 


Thoughts?

FabioLuporini:
I prefer the former of the two, I think. And yes, I agree the ``direction`` parameter would have to be dropped in that case, which would make me even happier

navjotk:
Due to implementation complexities, I believe it is much better to implement `2.` above although in theory I do prefer `1.`. I am currently working on that.

mlange05:
[...]

Issue #69, opened 2016-08-03
Final timestep index not clearly defined when `save=False` in TimeData


vincepandolfo:
This is the stencil generated for test_diffusion.py:

```
extern "C" int Operator(float *u_vec)
{
  float (*u)[100][100] = (float (*)[100][100]) u_vec;
  {
    int t0;
    int t1;
    for (int i3 = 0; i3<19; i3+=1)
    {
      {
        t0 = (i3)%(2);
        t1 = (t0 + 1)%(2);
      }
      {
        for (int i1 = 1; i1<99; i1++)
        {
          #pragma GCC ivdep
          for (int i2 = 1; i2<99; i2++)
          {
            u[t1][i1][i2] = 1.11022302462516e-16F*u[t0][i1][i2] + 2.5e-1F*u[t0][i1][i2 - 1] + 2.5e-1F*u[t0][i1][i2 + 1] + 2.5e-1F*u[t0][i1 - 1][i2] + 2.5e-1F*u[t0][i1 + 1][i2];
          }
        }
      }
    }
  }
  return 0;
}
```

The values of t0 and t1 alternate, so it's not clearly defined what index contains the final timestep. If the number of timesteps was even, the final result would be at index 1, while in this code the final result is at index 0

navjotk:
Which timestep is last written to is a property and also the decision of the Operator that was run on it. I believe we should keep in mind that these data objects would be used for multiple operations and would exist between multiple Operator apply calls. Therefore the last time step written to should be seen as a piece of information the Operator controls and not the data.

vincepandolfo:
There's no logic in the operator to do that as of right now though

mloubout:
@navjotk , this data may be used for different operators. As we originally designed it, the PDE (operator) is a solver that get it's input (nt, src/rec pos, h,...) from the data and the model. The data should however not be aware of any time indexing in the C code, and the last time step is the last index in data. In case of cyclic index, the data need to know which index is which time for reuse.

navjotk:
Yeah now I understand and appreciate this problem better. I think the way to go forward would really be for the data structure to be aware of these cyclic indices indeed. Overriding the array access and redirecting it to the right location is what I suggest.  This is probably what you were originally thinking of too, @vincepandolfo

mlange05:
[...]

PR #345, opened 2017-09-07
Introduce start and end arguments for dimensions

navjotk:
Finally, this is the critical PR that introduces the new devito feature from #247 , letting the user specify the start and end for any dimension. 

This is PR 2/3 on the way to checkpointing.


FabioLuporini:
I'll review this properly once rebased on master+#342 

However, skimming through it, I can already anticipate this lacks testing. We need many

devito/arguments.py
```diff
@@ -193,14 +189,15 @@ class DimensionArgProvider(ArgumentProvider):
...
     def reset(self):
-        self._value = self._default_value
+        for i in self.rtargs:
+            i.reset()
 
     @property
     def value(self):
-        return self._value
+        child_values = tuple([i.value for i in self.rtargs])
+        return child_values if all(i is not None for i in child_values) else None
```
FabioLuporini: In what case can they all be ``None`` ?

devito/arguments.py
```diff
@@ -209,8 +206,23 @@ def dtype(self):
...
     @cached_property
     def rtargs(self):
-        size = ScalarArgument("%s_size" % self.name, self, max)
-        return [size]
+        size = ScalarArgument(self.size_name, self, max)
+        start = ScalarArgument(self.start_name, self, max, 0)
+        end = ScalarArgument(self.end_name, self, max)
+        return [size, start, end]
+
+    def promote(self, value):
```
FabioLuporini: docstring

devito/arguments.py
```diff
@@ -209,8 +206,23 @@ def dtype(self):
...
+        if not isinstance(value, tuple):
+            size, start, end = self.rtargs
+            value = (value, start.default_value, value)
```
FabioLuporini: if not a tuple, then it's something else, and we're returning it. What is it supposed to be? can we assert it?

devito/arguments.py
```diff
@@ -209,8 +206,23 @@ def dtype(self):
...
+        else:
+            if len(value) == 2:
+                # Assume we've been passed a (start, end) tuple
```
FabioLuporini: is "Assume" the right word ? isn't it always the case if we have a 2-tuple ?

devito/arguments.py
```diff
@@ -209,8 +206,23 @@ def dtype(self):
...
+        size = ScalarArgument(self.size_name, self, max)
+        start = ScalarArgument(self.start_name, self, max, 0)
+        end = ScalarArgument(self.end_name, self, max)
+        return [size, start, end]
+
+    def promote(self, value):
+        if not isinstance(value, tuple):
+            size, start, end = self.rtargs
+            value = (value, start.default_value, value)
+        else:
```
FabioLuporini: can we just have ``elif``s ? are we handling all cases ?

devito/arguments.py
```diff
@@ -226,10 +238,13 @@ def verify(self, value):
             except AttributeError:
                 return False
 
+        value = self.promote(value)
```
FabioLuporini: Probably we need a comment here

devito/dimension.py
```diff
@@ -32,6 +32,30 @@ def symbolic_size(self):
         """The symbolic size of this dimension."""
         return self.rtargs[0].as_symbol
 
+    @property
+    def symbolic_extent(self):
+        """Return the extent of the loop over this dimension.
```
FabioLuporini: I wouldn't use the work "loop" here

FabioLuporini: BTW, random thought, I think we should start calling these objects ``Iterators``, or at least something better than ``Dimension``. Or we should derive Iterators from Dimensions... don't know. Just a random thought

devito/operator.py
```diff
@@ -231,7 +231,7 @@ def compile(self):
         """
         if self._lib is None:
             # No need to recompile if a shared object has already been loaded.
-            return jit_compile(self.ccode, self._compiler)
+            return jit_compile(self.ccode, configuration['compiler'])
```
FabioLuporini: This won't work with YASK. It needs to be ``self._compiler``. Perhaps this is post-merge/post-rebase leftover/error

devito/operator.py
```diff
@@ -240,7 +240,7 @@ def cfunction(self):
         """Returns the JIT-compiled C function as a ctypes.FuncPtr object."""
         if self._lib is None:
             basename = self.compile
-            self._lib = load(basename, self._compiler)
+            self._lib = load(basename, configuration['compiler'])
```
FabioLuporini: same

devito/nodes.py
```diff
@@ -394,6 +394,7 @@ def bounds(self, start=None, finish=None):
         available (either statically known or provided through ``start``/
         ``finish``). ``None`` is used as a placeholder in the returned 2-tuple
         if a limit is unknown."""
+        lower = None
```
FabioLuporini: likewise ``upper = None`` ?

devito/operator.py
```diff
@@ -456,7 +456,7 @@ def apply(self, **kwargs):
         self.cfunction(*list(arguments.values()))
 
         # Output summary of performance achieved
-        summary = self.profiler.summary(dim_sizes, self.dtype)
+        summary = self.profiler.summary(arguments, self.dtype)
```
FabioLuporini: so can we drop ``dim_sizes`` ?

devito/profiling.py
```diff
@@ -137,17 +137,25 @@ def summary(self, dim_sizes, dtype):
...
+                else:
+                    start = arguments[dims[i].start_name]
+                    end = arguments[dims[i].end_name]
+                itershape.append(i.extent(finish=end, start=start))
             iterspace = reduce(operator.mul, itershape)
             flops = float(profile.ops*iterspace)
             gflops = flops/10**9
 
             # Compulsory traffic
             datashape = [i.dim.size if i.dim.is_Fixed
```
FabioLuporini: perhaps this is movable inside the ``for i in itspace`` loop above?

devito/arguments.py
navjotk: Each runtime argument is independent and gets its value independent of the others. When it doesn't have a value, it is `None`, so they can all be `None` if none of them have a value. Is that your question? 
The code above, however, returns a tuple of values if all arguments have a known value, otherwise it returns `None`.

devito/arguments.py
navjotk: The condition is on whether `value` is a `tuple` or not so technically, yes, we are handling all the cases. What is your concern here?

devito/arguments.py
navjotk: There are many ways to interpret a 2-tuple in order to promote it to a 3-tuple. I just chose one as convention and stated that assumption here. It is an assumption.

devito/arguments.py
navjotk: Saying what?

devito/dimension.py
navjotk: We already derive `Iteration`s from `Dimension`s. How are `Iterator`s different from `Iteration`s?

devito/operator.py
navjotk: Indeed

devito/operator.py
navjotk: I am working towards dropping it, yes. I think this might be controversial with @mloubout though. I believe `arguments`, the other return value has everything we need.

devito/nodes.py
navjotk: Tests pass without it. They don't pass without this. Something missing in the tests? :)

devito/arguments.py
navjotk: This method is promoting a scalar to a tuple so everything that is not a tuple is treated like a scalar.

devito/arguments.py
FabioLuporini: gotcha now

devito/arguments.py
FabioLuporini: OK. Perhaps make it private with ``_`` ?

devito/arguments.py
FabioLuporini: I **think** so... unless it's too complicated, could you sketch (pseudocode is fine) a concrete example in which this happens ?

devito/arguments.py
FabioLuporini: when/why we need promotion

devito/nodes.py
FabioLuporini: perhaps...

devito/arguments.py
navjotk: Let's say we are talking of the `time` dimension, it has three `Argument`s - `time_size`, `time_start`, `time_end`. If `time_size.value=100`, `time_start.value=50` and `time_end.value=60`, this method would return `(100, 50, 60)`. If any of them doesn't have a value (and hence is `None`), this method would return `None`.

devito/profiling.py
navjotk: TBH I don't understand the reason for calculating the datashape and iterspace separately. They seem to resolve to the same number always. Either that or my changes to the code in this section are based on incorrect assumptions.

FabioLuporini: the itershape differs in general because of the stencil

navjotk: The itershape is the product of the size of all dimensions, and so is the datashape. Correct? Always?

FabioLuporini: no it's the product of the extents. Previously ``dim_sizes`` was being used, which in turn was used to call ``extents``

navjotk: Which one? Itershape or datashape? What is the other one?

FabioLuporini: itershape

devito/operator.py
mloubout: What d you mean drop dim sizes?

devito/profiling.py
navjotk: And how is datashape different?

FabioLuporini: You're touching boundary data through the stencil

FabioLuporini: Pls add a comment

devito/dimension.py
FabioLuporini: It's just that at this point I'm not sure I like the name ``Dimension`` given what these objects can do...but this is more than debatable

navjotk:
Temporarily closing

devito/arguments.py
FabioLuporini: You wrote:

> Each runtime argument is independent and gets its value independent of the others. When it doesn't have a value, it is None, so they can all be None if none of them have a value. Is that your question?
> The code above, however, returns a tuple of values if all arguments have a known value, otherwise it returns None.
> 

Do you think it'd be useful to add it as a comment in the code ?

devito/operator.py
```diff
@@ -81,7 +81,6 @@ def __init__(self, expressions, **kwargs):
         self.dtype = self._retrieve_dtype(expressions)
         self.input, self.output, self.dimensions = self._retrieve_symbols(expressions)
         stencils = self._retrieve_stencils(expressions)
-
```
FabioLuporini: please leave a blank line

devito/operator.py
```diff
@@ -151,14 +150,22 @@ def arguments(self, **kwargs):
...
+        runtime_dimensions = [d for d in self.dimensions if not d.is_Fixed]
         for d in runtime_dimensions:
             d.verify(kwargs.pop(d.name, None))
         for i in self.parameters:
             if i.is_ScalarArgument:
                 i.verify(kwargs.pop(i.name, None))
-
-        dim_sizes = OrderedDict([(d.name, d.value) for d in runtime_dimensions])
+        dim_sizes = {}
+        for d in runtime_dimensions:
```
FabioLuporini: what's going on here?

devito/operator.py
```diff
@@ -151,14 +150,22 @@ def arguments(self, **kwargs):
...
+            if d.value is not None:
+                _, d_start, d_end = d.value
+                # The variable is called size but what it expects is extent
+                # Since this is used for loop blocking
```
FabioLuporini: I don't like the fact that ``Operator.arguments`` does anything specific to "loop blocking". Can we abstract things a little bit ?

devito/profiling.py
FabioLuporini: reminder that this still needs to be addressed.

Also, since you're touching the profiling engine, can you verify that, for example, acoustic or tti still produce the same "performance summary" when running from ``master`` and from this ``branch`` ? Can you paste here some data?

devito/dimension.py
```diff
@@ -36,6 +36,30 @@ def symbolic_size(self):
     def size(self):
         return None
 
+    @property
+    def symbolic_extent(self):
+        """Return the extent of the loop over this dimension.
+        Would be the same as size if using default values """
+        _, start, end = self.rtargs
```
FabioLuporini: I still think  ``namedtuple`` would be better

devito/operator.py
navjotk: It is not specifically for loop blocking. It is simply calculating the extent of the dimension - which is a general feature.

devito/arguments.py
```diff
@@ -299,7 +339,27 @@ class ObjectArgProvider(ArgumentProvider):
 
     @cached_property
     def rtargs(self):
-        return [PtrArgument(self.name, self)]
+        return (PtrArgument(self.name, self),)
+
+
+class ArgumentEngine(object):
+    def extract_dimension_offsets(self, stencils):
```
FabioLuporini: what's going on here (docs) ?

FabioLuporini: I might have a vague idea, one or two concrete examples would be of help

devito/operator.py
```diff
@@ -150,15 +157,22 @@ def arguments(self, **kwargs):
...
-            d.verify(kwargs.pop(d.name, None))
+            d.verify(kwargs.pop(d.name, None), self.argument_engine, enforce=True)
         for i in self.parameters:
             if i.is_ScalarArgument:
-                i.verify(kwargs.pop(i.name, None))
-
-        dim_sizes = OrderedDict([(d.name, d.value) for d in runtime_dimensions])
+                i.verify(kwargs.pop(i.name, None), self.argument_engine, enforce=True)
+        dim_sizes = {}
+        for d in runtime_dimensions:
```
FabioLuporini: perhaps this ``extent`` calculation could be a ``Dimension`` property?

devito/operator.py
```diff
@@ -94,6 +95,12 @@ def __init__(self, expressions, **kwargs):
         # Wrap expressions with Iterations according to dimensions
         nodes = self._schedule_expressions(clusters)
 
+        # Initialise Argument Engine
+        self.argument_engine = ArgumentEngine()
+
+        # Extract dimension offsets
+        self.argument_engine.extract_dimension_offsets([i.stencil for i in clusters])
```
FabioLuporini: with this ``ArgumentEngine``, are you simply trying to extract the outermost offsets along each dimension? An example; given the expression
```
A[i-3, j+1] - B[i+4, j+5] + C[i-2, j+3]
```
Return:
```
i: (-3, 4)
j: (1, 5)
```

devito/operator.py
```diff
@@ -304,7 +316,8 @@ def _schedule_expressions(self, clusters):
                 needed = entries[index:]
 
                 # Build and insert the required Iterations
-                iters = [Iteration([], j.dim, j.dim.symbolic_size, offsets=j.ofs)
+                iters = [Iteration([], j.dim, j.dim.limits,
+                                   offsets=j.ofs if j.dim.is_Fixed else (0, 0))
```
FabioLuporini: why?

tests/test_operator.py
```diff
@@ -289,6 +303,63 @@ def test_override_composite_data(self):
         arg_name = src1.name + "_coords"
         assert(np.array_equal(args[arg_name], np.asarray((new_coords,))))
 
+    def test_start_end_2d(self):
```
FabioLuporini: docstrings, for this and all other new tests

devito/arguments.py
```diff
@@ -52,37 +53,38 @@ def as_symbol(self):
...
 
+    def __repr__(self):
+        return self.name
+
 
 class ScalarArgument(Argument):
 
     """ Class representing scalar arguments that a kernel might expect.
         Most commonly used to pass dimension sizes
+        enforce determines whether any reduction will be performed or not.
```
FabioLuporini: use doc-friendly notation:
```
:param enforce: Determine ...
```
Also, add an example? Not straightforward to imagine what's the use case here

FabioLuporini: Q: should we provide a difference type of ``Argument``, rather than adding a parameter such as ``enforce`` ?

devito/arguments.py
```diff
@@ -52,37 +53,38 @@ def as_symbol(self):
...
 
     is_ScalarArgument = True
 
-    def __init__(self, name, provider, reducer, default_value=None):
+    def __init__(self, name, provider, reducer=lambda old, new: new, default_value=None):
         super(ScalarArgument, self).__init__(name, provider, default_value)
         self.reducer = reducer
 
-    def verify(self, value):
+    def verify(self, value, engine, enforce=False):
```
FabioLuporini: unused ``engine``

devito/operator.py
navjotk: Dimensions are global objects shared between different Operators. This is an intrinsic property of an Operator.

devito/operator.py
navjotk: If you're hardcoding the limits, hard-code the offsets as well. Else, the offset will be accounted for in the runtime limit calculation.

FabioLuporini: no but I mean, why `j.ofs` wouldn't return ``(0, 0)`` if ``j.is_Fixed`` is True?

devito/operator.py
FabioLuporini: I disagree; that'd be a property of a *specific Operator invocation*, not of the Operator itself. And also, so would be ``.limits``. Finally, let's not forget that a Dimension **is** an ArgProvider, so... We can argue whether that should be the case, but that's another story

devito/operator.py
navjotk: Because it would return the offset in that dimension which depends on the stencil radius?

devito/operator.py
navjotk: So if it is a property of the invocation, it belongs right here, no? 
I don't understand what that ArgProvider thing is about but it seems off topic

devito/operator.py
navjotk: Is there a question?

devito/arguments.py
navjotk: For consistency, all verify methods accept engine.

devito/arguments.py
navjotk: I don't think your suggestion has anything to do with what is happening here. The argument decides whether the reducer will be called or not. It is still a scalar, but when I am auto-deriving, I want to use a reducer to get the value that everyone agrees on, while if I am using a user-provided value, I want to skip all the reduction operations and just use the provided value directly.

devito/operator.py
FabioLuporini: yes a confirmation that this what you're trying to do

navjotk: Yes but it doesn't return those values, it caches them on the engine object.

devito/operator.py
FabioLuporini: anyway, we can drop this because of #373

devito/operator.py
FabioLuporini: OK, then I think there should be ways of doing this without this engine, which is only over-complicating things

a few lines above you have the ``stencils`` variable, which you can inspect to get the max/min offsets along any Operator dimension (you might even add helper methods to ``Stencil`` directly?).  What happens if you try using this?

devito/operator.py
FabioLuporini: Being an ``ArgProvider``, it seemed kinda natural to have an ``extent`` method.

I'm just concerned that the ``.arguments`` method is becoming more and more complicated, so it would be useful (for many purposes) to "offload" some bit of functionalities to other modules, if/where possible. Mine was just a suggestion in this direction (perhaps a conceptually wrong one, but I'm not that sure).

navjotk:
The biggest problem in this branch currently is that `test_dse.py/test_tti_rewrite_speculative` fails because there are some hidden assumptions about all loops running from 0 to `dimension_size` in the speculative rewrite passes of the DSE that I haven't been able to find after a week of looking. Can @FabioLuporini please look at making this test pass? More specifically, this problem is most probably localised within the "inter-stencil redundancies" pass because disabling that pass makes everything else work fine in speculative mode.

FabioLuporini:
> after a week of looking



> The biggest problem in this branch currently is that test_dse.py/test_tti_rewrite_speculative fails because there are some hidden assumptions about all loops running from 0 to dimension_size in the speculative rewrite passes of the DSE that I haven't been able to find after a week of looking. Can @FabioLuporini please look at making this test pass? More specifically, this problem is most probably localised within the "inter-stencil redundancies" pass because disabling that pass makes everything else work fine in speculative mode.

if you get stuck on something where I might help for more than a few hours, you should ask, not wait for 8x5=40 hours. Anyway, sure, I'll take a look. I'll do that once the branch has been rebased and we're sure that all other tests pass. Can you address the other comments, meanwhile?

navjotk:
I did ask you when you were here. Twice. The first time you just explained to me how the pass works instead of helping me. The second time you said (and I'm paraphrasing) you didn't have the (mental) bandwidth to help me so I went back to trying to figure it out myself.

devito/profiling.py
FabioLuporini: reminder. This still needs to be addressed.

devito/operator.py
FabioLuporini: any updates ?

FabioLuporini:
I'm planning to take a look at the failing test case this afternoon

devito/operator.py
navjotk: This will go in first

devito/operator.py
navjotk: And that's exactly what the ArgumentEngine is slated to do

devito/operator.py
navjotk: The ArgumentEngine will provide a runtime environment for the operator. I see only a cleaner interface because of its introduction - there is no overcomplication. The stencil radius is a property that is computed at compile (operator creation) time and used at runtime (apply). The ArgumentEngine is self-sufficient in what it needs to apply the operator and hence stores these properties on itself.

FabioLuporini: > What happens if you try using this?

? -- point is, if this suffices, we can just store this info and forget about all the rest

tests/test_autotuner.py
```diff
@@ -102,7 +102,7 @@ def test_timesteps_per_at_run():
         op = Operator(stencil, dle=('blocking', {'blockalways': True}))
         op(infield=infield, outfield=outfield, autotune=True)
         out = [i for i in buffer.getvalue().split('\n') if 'AutoTuner:' in i]
-        expected = options['at_squeezer'] - to
```
FabioLuporini: why is this necessary?

devito/profiling.py
ggorman: Let me pick this apart. @FabioLuporini are you saying that the original code is wrong? Or is this functionally the same as the previous code but you have a different feeling for how it should be styled? If so, can we distinguish between advisory and required changes and style changes can potentially be put into a separate PR and applied consistently. Essentially (and particularly for large PR's) we don't want to have moving goalposts.

FabioLuporini: I originally pointed out something about style (the ``dataspace`` calculation). Then I thought that even more important would be to verify that the "Performance Summary" is correct. For example, if we run ``acoustic`` on ``master`` and then on this branch, does Devito produce/display the same numbers (itershape, datashape, ..)? In the past, @mloubout and I had to do some serious investigation to find out why the performance numbers were off, so I just wanna make sure that we're still producing exactly the same reports as ``master`` (hence my "reminder" from a few days ago).

BTW, the style change was a one-line edit. I'm not such a great believer in "this low-priority thing will be improved in a subsequent PR", because it rarely happens in practice. So it's much better to do it straight away. This is the philosophy we've adopted since the Devito3.0 rewrite (Mathias can confirm)

devito/operator.py
ggorman: @FabioLuporini - do you mean "drop the patch" or "drop the review comment"?

devito/operator.py
ggorman: What's the conclusion from this discussion? Are there changes required to this PR or is the increasing complexity of .arguments being highlighted for some TLC (ie should an issue be articulated explaining the problem and we can assign the task to someone)?

devito/operator.py
FabioLuporini: I was assuming #373 would go in first (it's ready, while this one is not); if #373 goes in first, we can drop the ``is_Fixed`` check above, because #373 removes ``FixedDimension``

FabioLuporini:
After a thorough offline discussion, we're in this situation:

- The branch ``int_start_end_fabio`` has a commit that should fix some of the current Travis failures
- However, even with this patch, the current PR still breaks code generation in various places, in particular due to dropping offsets on the floor during ``_schedule_expressions`` (the routine responsible for constructing loops). I think I convinced @navjotk that this is wrong, but still we need a solution to the following problem:

Say we want an Operator to run for 20 iterations along the time dimension, starting at timestep 50 (doesn't have to be time, but just to make the discussion more concrete... as time is the driving use case of the PR). We could easily generate code such as:
```
for (t = t_start + 1 ; t < t_end - 1; t++) {
  ...
}
```
The offsets +-1 are necessary to avoid out-of-bounds array accesses along the time dimension.
So, what does it mean "to start executing at timestep 50" ? Does it mean ``t_start`` should be ``== 49``? Or should we actually start from ``t=51``? What convention should we follow?

The PR as is in its current state was trying to work around this issue by dropping the offsets in the generated code (perhaps I'm missing some details here, please @navjotk  clarify as necessary). But, as I said, we do agree this is in general wrong (we do need offsets, for a number of reasons).

So I guess we need a convention before a fix ( @ggorman @mlange05 @mloubout warmly invited to contribute).

mlange05:
Ok, I generally think this ties in with the upcoming switch to properly using the concept of a "domain", as termed in YASK. This domain does not include the "boundary"/"padding" region that arises due to the stencil and effectively constitutes the part of the iteration that is written to. I think the user should address iteration indices with respect to the "domain", which might be different to the actual data indices in memory. To illustrate, some ASCII art:

```
 boundary             domain                     boundary
  | -- | ---------------------------------------- | -- |
   read                write                       read

Data indices:
  0    1               51     71                 101  102
User indices
       0               50     70                 100
```

I think the example code @FabioLuporini showed above would exactly achieve this, no?

navjotk:
I would call the offline discussion inconclusive. 
I have since merged the branch with the DSE fixes and added a test to show that Iterations are still perfectly capable of producing offsets if you choose to use them for some bit of the code.

With code like: 
```
for (t = t_start + 1 ; t < t_end - 1; t++) {
  ...
}
```
we are lying about which timestep we are running since, being passed `t_start=50`, we would start with `t=51`. To make this more transparent to the user, this PR now generates code like:
```
for (t = t_start; t < t_end; t++) {
  ...
}
``` 
where `t_start` and `t_end` already carry the effect of the offset at run time. 

I do not agree that it is wrong to do this since logically we are still doing the same thing. 

Specifically, I don't see why certain DSE passes should break because of this, when it works fine with the lower level DSE optimisations. I continue to insist that this is a bug in the DSE being exposed as a side-effect of the PR. 

I am not in favour of dragging along this PR into a never-ending discussion.

@mlange05 I don't see your point

FabioLuporini:
> and added a test to show that Iterations are still perfectly capable of producing offsets if you choose to use them for some bit of the code.

Of course they are: you haven't touched the ``Iteration`` implementation. When I say that you're dropping the offsets on the floor, what I mean is that now an ``Operator`` (through ``_schedule_expressions``), which is the only entity in Devito constructing ``Iteration``s (apart from some DLE passes), is "dropping them", but ignoring the information carried by a ``cluster``. This is the incriminating line:

https://github.com/opesci/devito/pull/345/files#diff-de8fe5c64c6d55d55d54343f37e2397cR321

in particular, we're not using ``j.ofs`` anymore (don't forget that ``is_Fixed`` is now totally irrelevant).

> Specifically, I don't see why certain DSE passes should break because of this, when it works fine with the other DSE optimisations. I continue to insist that this is a bug in the DSE being exposed as a side-effect of the PR.

We basically spent the whole of yesterday talking about this, about why we need offsets in Iterations. I thought I had convinced you, but I may be wrong. Different ``Iterations `` over the same dimension, say ``x``, may require slightly different iteration ranges. For example, in the same kernel, I may have:
```
for (x = 1 ; x < x_size -1 ; x++)
...
...
for (x = 2 ; x < x_size -2 ; x++)
```
It's logically still the `x` dimension (it is and it must be); but the loop bounds of the first loop lead to a slightly larger iteration range because of DSE-induced optimizations. 

So you cannot just replace both of them with a single ``x_start``; it won't work, because the first loop/iteration differs from the second one. You might use ``x_start`` and ``x_start + 1``; or you might write a completely new implementation (but why) where you have ``x_start_1`` and ``x_start_2`` (with the property that ``x_start_2 == x_start_1 + 1``; or whatever. 

So, again, this issue is *not* exposing any DSE bug. The DSE works as it's been conceived. We might argue about how bad or cool is it, but I don't see why this should be called a DSE bug.

On the other hand, the bug is in the current PR implementation (as I hope I have clarified)

> @mlange05 I don't see your point

I do. And given that and what I just wrote, I don't see why a solution in which we write loops such as:
```
for (x = 1 + x_start; x < x_end -1 ; x++)
...
...
for (x = 2 + x_start ; x < x_end -2 ; x++)
```
wouldn't just work (compare to the other snippet above; I added ``x_start`` here to both loops). ``x_start`` is user-provided at apply time (i.e., ``op.apply(x_start=50)``, as expected, I think, by the current API proposal). Michael basically said: ignore the offsets.

mlange05:
> we are lying about which timestep we are running since, being passed t_start=50, we would start with t=51.

FWIW, I think this is only true for time due to the fact that we currently treat time as a boundary-less dimension (because of `save`, etc). In this case `j.ofs` should always be 0 for time and we might just force this behaviour. For other dimension types, eg. the space dimensions, "lying about" the iteration index is actually the aim here, although I admit that we haven't switched to the "domain" concept yet and are currently still treating everything as absolute indices (from a user perspective).

Regardless of this, the mechanism to consider offsets will be needed as soon as we do switch, and is already required due to the inter-stencil redundancy stuff (if I understand @FabioLuporini correctly).

>  (don't forget that is_Fixed is now totally irrelevant).

Is this fact affecting the current discussion now? If so, should #373 be held on account of this PR, or should we re-evaluate this PR in light of #373? @navjotk

FabioLuporini:
> Regardless of this, the mechanism to consider offsets will be needed as soon as we do switch, and is already required due to the inter-stencil redundancy stuff (if I understand @FabioLuporini correctly).

So I wouldn't say "required", strictly speaking, because things as they are now on master are working as they're meant to be. Just for completeness of the discussion: ``Cluster``s, through ``Stencil``s, carry the offset information prior to loop construction; ``Cluster``s are originally created by the ``Operator``, based on the user-provided equations, and subsequently "modified" by the DSE (which may add or drop clusters). 

The offset/domain/... mechanism, through a ``Grid``, will be fundamental for YASK.

mloubout:
Shouldn't this PR also fix/be related #69 ?
The one thing to remember is that, if the `start/end` are called multiple times in a row, the only thing that should matter is that you get the same thing as running the full time/space at once, if not there is a big problem.

>we are lying about which timestep we are running since, being passed t_start=50, we would start with t=51.

this will depend on the kernel/PDE/discretization, ignoring offset is risky down the road and I think whatever `tstart` or `xstart` and same for end should correspond to the ones computed so for acoustic:

```
for (t = t_start-1; t < t_end; t++) {
  u[t+1] = .... (first computed is t_start, last is t_end)
}
t_start_new = t_end+1
t_end_new=....
```

but with symbolic check to insure the value passed are the correct one and don't go out of bounds.

mlange05:
> > we are lying about which timestep we are running since, being passed t_start=50, we would start with t=51.

> FWIW, I think this is only true for time due to the fact that we currently treat time as a boundary-less dimension 

@navjot Would it make things easier if we introduced explicit `TimeDimension` types now? I have them in a separate WIP branch atm, but they can be fast-tracked. With this you could simply do something like
```
iters = [Iteration([], j.dim, j.dim.limits, offsets=(0, 0) if j.dim.is_Time else j.ofs)
```

FabioLuporini:
> @Navjot Would it make things easier if we introduced explicit TimeDimension types now? I have them in a separate WIP branch atm, but they can be fast-tracked. With this you could simply do something like

If we do 0-offsets in time, can we already guarantee that we generate something like:
```
u[t+k] = f(u[t+k-1], u[t+k-2], ..., u[t])
```
rather than, for example:
```
u[t] = f(u[t-1], u[t-2], ..., u[t-k])
```
ie, we don't end up with OOB accesses?

mlange05:
OK, #382 has landed, so this branch needs a rebase/merge. @navjotk

devito/profiling.py
```diff
@@ -137,17 +137,19 @@ def summary(self, dim_sizes, dtype):
...
             gpoints = iterspace/10**9
 
             # Compulsory traffic
-            datashape = [dim_sizes[dims[i].name] for i in itspace]
+            datashape = [(arguments[dims[i].end_name] - arguments[dims[i].start_name])
+                         for i in itspace]
             dataspace = reduce(operator.mul, datashape)
             traffic = profile.memory*dataspace*dtype().itemsize
-
+            traffic = float(profile.memory*dataspace*dtype().itemsize)
```
FabioLuporini: redundant line, use float(...) in the line above

navjotk: merge leftover. fixed.

navjotk:
Can this be merged?

FabioLuporini:
please go

Issue #344, opened 2017-09-07
Code quality issues related to checkpointing work


navjotk:
This issue is to keep track of things being put off for until after checkpointing is completely in. 

~https://github.com/opesci/devito/pull/342#discussion_r137460092~

~https://github.com/opesci/devito/pull/342#discussion_r137458662~

~https://github.com/opesci/devito/pull/342#discussion_r137459227~

~https://github.com/opesci/devito/pull/342#discussion_r137457801~

~https://github.com/opesci/devito/pull/342#discussion_r137449295~

Edit: 

~https://github.com/opesci/devito/pull/354/files/324bdaf2843f577032f8604a0b1de3d09d5f6cf4#r141003338~

FabioLuporini:
ANy leftovers here, @navjotk ?

FabioLuporini:
Any leftovers here, @navjotk ?

PR #346, opened 2017-09-11
Revert "Separate the case where Dimensions have a known size into its own class"

navjotk:
Reverts opesci/devito#342

### DIFF
--- a/devito/arguments.py
+++ b/devito/arguments.py
@@ -147,40 +147,6 @@ class ArgumentProvider(object):
         raise NotImplemented()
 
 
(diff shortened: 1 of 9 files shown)


navjotk:
The previous PR broke major functionality in Julia and was clearly under-tested. Reverting for now. Shall be back with better testing.

FabioLuporini:
I wonder whether you can address #344 since you're going to spend some more time on this

Perhaps we should also temporarily close #345 and reopen it when it's time, because it's gonna need a rebase anyway, so...

PR #349, opened 2017-09-12
YASK: Support for loops with indirect accesses

FabioLuporini:
~NOT ready for merge yet, although we're basically there.~

Now we can execute through YASK any Devito kernels mixing sources/receivers-like loop nests (w/ indirect memory accesses) and parallel loops (ie, loops that can be "offloaded" to YASK).

In the generated code ...
... The offloadable loops are replaced by a function call to YASK
... All other loops remain "explicit" in the generated code, but array accesses are replaced with YASK grid accesses. This involves some shared pointer magic in the compiler and in the generated code.

Tests are also added

### DIFF
--- a/devito/arguments.py
+++ b/devito/arguments.py
@@ -127,6 +127,7 @@ class PtrArgument(Argument):
         super(PtrArgument, self).__init__(name, provider, provider.value)
 
     def verify(self, value):
+        self._value = value or self._value
         return True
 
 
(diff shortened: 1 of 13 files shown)


FabioLuporini:
This needs a review and then should be merged.

FabioLuporini:
I'm merging it as, anyhow, this code is gonna evolve pretty quickly and... let's not forget this is still a prototype

PR #350, opened 2017-09-12
acoustic don't ignore at anymore

mloubout:
TIny fix to use AT in acoustic (kwarg was ignored so far)


examples/seismic/acoustic/acoustic_example.py
```diff
@@ -52,25 +52,25 @@ def acoustic_setup(dimensions=(50, 50, 50), spacing=(15.0, 15.0, 15.0),
 
 def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
         time_order=2, space_order=4, nbpml=40, full_run=False, **kwargs):
-
+    autotune = kwargs.pop('autotune', False)
```
FabioLuporini: move ``autotune=False`` to the function signature so you don't have to pop it here. Same in TTI

mloubout:
I mainly only replicated the TTI example setup.

FabioLuporini:
> Will try to make it nicer

No worries. Just add ``autotune`` to the signature as per my comment above, then we can merge this

mloubout:
An issue is left, something is buggy wit hautotuning with `save=False`

- benchmark ignores AT for `forward` but standalone example doesn't
- autotuning fails with `save=True` when not ignored as the `sequntial` is empty (my change)
-  autotuning runs the ful ltime axis with `save=True` once timestep count is removed

examples/seismic/acoustic/acoustic_example.py
```diff
@@ -51,7 +51,8 @@ def acoustic_setup(dimensions=(50, 50, 50), spacing=(15.0, 15.0, 15.0),
 
 
 def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
-        time_order=2, space_order=4, nbpml=40, full_run=False, **kwargs):
+        time_order=2, space_order=4, nbpml=40, full_run=True,
```
FabioLuporini: shouldnt ``full_run`` be False by default?

devito/core/autotuning.py
```diff
@@ -93,7 +93,9 @@ def autotune(operator, arguments, tunable):
         operator.cfunction(*list(at_arguments.values()))
         elapsed = sum(operator.profiler.timings.values())
         timings[tuple(bs.items())] = elapsed
-        info_at("<%s>: %f" % (','.join('%d' % i for i in bs.values()), elapsed))
+        iter_time = [s.extent(finish=at_arguments['time_size']) for s in sequential][0]
```
FabioLuporini: I think a better way is to check as soon as possible that ``sequential`` contains only one loop and then extract it straight away

FabioLuporini: ie right after line 33

FabioLuporini:
> autotuning fails with save=True when not ignored as the sequntial is empty

I think I know why it's because of ``i.dim.is_Buffered``

> autotuning runs the ful ltime axis with save=True once timestep count is removed

Probably because we're attempting to replace ``t_size`` only instead of also ``time_size`` (or the other way around)

FabioLuporini:
@mloubout I'll need this stuff next week, so I pushed a few improvements and rebased on master.

The only thing I don't understand is why you replaced ``kwargs`` with ``autotune=autotune`` in the acoustic example. I think we should match what we do in ``tti_example``.

Once this is sorted out, and once both you and Travis are happy, we merge this. I recommend before next week.

mloubout:
> I think we should match what we do in `tti_example`.

That is what's in it. Specially if we add `dse/dle` to the examples parseargs, the operators don't accept it as input due to `devito/operator/arguments`

Will finish this one today

FabioLuporini:
Fixes and more tests added. As discussed offline with @mloubout , I'm gonna merge this if Travis gets green

FabioLuporini:
Mering...

FabioLuporini:
Merged.

PR #351, opened 2017-09-14
New tti again

mloubout:
Ok, so I rewrote the TTI kernel (again yes I know)

It is quite faster/ less operations and still produce good results (I think).

Current "issues":
- dse has to be `basic` in 2D
- dse has to be `aggressive` in 3D

I also added a bunch of tti preset models

### DIFF
--- a/examples/seismic/model.py
+++ b/examples/seismic/model.py
@@ -52,6 +52,36 @@ def demo_model(preset, **kwargs):
         return Model(vp=v, origin=origin, shape=shape,
                      spacing=spacing, nbpml=nbpml, **kwargs)
 
+    elif preset.lower() in ['layerstti', 'twolayertti', '2layertti']:
+        # A two-layer model in a 2D or 3D domain with two different
+        # velocities split across the height dimension:
+        # By default, the top part of the domain has 1.5 km/s,
+        # and the bottom part of the domain has 2.5 km/s.\
+        shape = kwargs.pop('shape', (101, 101))
+        spacing = kwargs.pop('spacing', tuple([10. for _ in shape]))
+        origin = kwargs.pop('origin', tuple([0. for _ in shape]))
+        nbpml = kwargs.pop('nbpml', 10)
+        ratio = kwargs.pop('ratio', 2)
+        vp_top = kwargs.pop('vp_top', 1.5)
+        vp_bottom = kwargs.pop('vp_bottom', 2.5)
+
+        # Define a velocity profile in km/s
+        v = np.empty(shape, dtype=np.float32)
+        v[:] = vp_top  # Top velocity (background)
+        v[..., int(shape[-1] / ratio):] = vp_bottom  # Bottom velocity
+
+        epsilon = .3*(v - 1.5)
+        delta = .2*(v - 1.5)
+        theta = .5*(v - 1.5)
+        phi = None
+        if len(shape) > 2:
+            phi = .1*(v - 1.5)
+
+        return Model(vp=v, origin=origin, shape=shape,
+                     spacing=spacing, nbpml=nbpml,
+                     epsilon=epsilon, delta=delta, theta=theta, phi=phi,
+                     **kwargs)
+
     elif preset.lower() in ['circle']:
         # A simple circle in a 2D domain with a background velocity.
         # By default, the circle velocity is 2.5 km/s,
@@ -97,6 +127,89 @@ def demo_model(preset, **kwargs):
 
         return Model(vp=v, origin=origin, shape=v.shape,
                      spacing=spacing, nbpml=20)
+
+    elif preset.lower() in ['marmousitti', 'marmousi2dtti']:
+
+        dimensions = (201, 201, 70)
+        shape = (201, 70)
+        spacing = (10., 10.)
+        origin = (0., 0.)
+        nbpml = kwargs.pop('nbpml', 20)
+
+        # Read 2D Marmousi model from opesc/data repo
+        data_path = kwargs.pop('data_path', None)
+        if data_path is None:
+            error("Path to opesci/data not found! Please specify with "
+                  "'data_path=<path/to/opesci/data>'")
+            raise ValueError("Path to model data unspecified")
+        path = os.path.join(data_path, 'marmousi3D/vp_marmousi_bi')
+
+        # velocity
+        vp = 1e-3 * np.fromfile(os.path.join(data_path,'marmousi3D/MarmousiVP.raw'),
+                                dtype='float32', sep="")
+        vp = vp.reshape(dimensions)
+        vp = vp[101, :, :]
+        # Epsilon, in % in file, resale between 0 and 1
+        epsilon = np.fromfile(os.path.join(data_path, 'marmousi3D/MarmousiEps.raw'),
+                              dtype='float32', sep="") * 1e-2
+        epsilon = epsilon.reshape(dimensions)
+        epsilon = epsilon[101, :, :]
+        # Delta, in % in file, resale between 0 and 1
+        delta = np.fromfile(os.path.join(data_path, 'marmousi3D/MarmousiDelta.raw'),
+                            dtype='float32', sep="") * 1e-2
+        delta = delta.reshape(dimensions)
+        delta = delta[101, :, :]
+        # Theta, in degrees in file, resale in radian
+        theta = np.fromfile(os.path.join(data_path, 'marmousi3D/MarmousiTilt.raw'),
+                            dtype='float32', sep="")
+        theta = np.float32(np.pi / 180 * theta.reshape(dimensions))
+        theta = theta[101, :, :]
+
+        return Model(vp=vp, origin=origin, shape=shape,
+                     spacing=spacing, nbpml=nbpml,
+                     epsilon=epsilon, delta=delta, theta=theta,
+                     **kwargs)
+
+    elif preset.lower() in ['marmousitti3d', 'marmousi3dtti']:
+
+        dimensions = (201, 201, 70)
+        spacing = (10., 10., 10.)
+        origin = (0., 0., 0.)
+        nbpml = kwargs.pop('nbpml', 20)
+
+        # Read 2D Marmousi model from opesc/data repo
+        data_path = kwargs.pop('data_path', None)
+        if data_path is None:
+            error("Path to opesci/data not found! Please specify with "
+                  "'data_path=<path/to/opesci/data>'")
+            raise ValueError("Path to model data unspecified")
+        path = os.path.join(data_path, 'marmousi3D/vp_marmousi_bi')
+
+        # Velcoity
+        vp = 1e-3 * np.fromfile(os.path.join(data_path, 'marmousi3D/MarmousiVP.raw'),
+                                dtype='float32', sep="")
+        vp = vp.reshape(dimensions)
+        # Epsilon, in % in file, resale between 0 and 1
+        epsilon = np.fromfile(os.path.join(data_path, 'marmousi3D/MarmousiEps.raw'),
+                              dtype='float32', sep="") * 1e-2
+        epsilon = epsilon.reshape(dimensions)
+        # Delta, in % in file, resale between 0 and 1
+        delta = np.fromfile(os.path.join(data_path, 'marmousi3D/MarmousiDelta.raw'),
+                            dtype='float32', sep="") * 1e-2
+        delta = delta.reshape(dimensions)
+        # Theta, in degrees in file, resale in radian
+        theta = np.fromfile(os.path.join(data_path, 'marmousi3D/MarmousiTilt.raw'),
+                            dtype='float32', sep="")
+        theta = np.float32(np.pi / 180 * theta.reshape(dimensions))
+        # Phi, in degrees in file, resale in radian
+        phi = np.fromfile(os.path.join(data_path, 'marmousi3D/Azimuth.raw'),
+                          dtype='float32', sep="")
+        phi = np.float32(np.pi / 180 * phi.reshape(dimensions))
+
+        return Model(vp=vp, origin=origin, shape=dimensions,
+                     spacing=spacing, nbpml=nbpml,
+                     epsilon=epsilon, delta=delta, theta=theta, phi=phi,
+                     **kwargs)
 
     else:
         error('Unknown model preset name %s' % preset)
--- a/examples/seismic/model.py
+++ b/examples/seismic/model.py
@@ -42,10 +42,10 @@ def demo_model(preset, **kwargs):
         v[:] = 1.5
         epsilon = .3*np.ones(shape)
         delta = .2*np.ones(shape)
-        theta = .2*np.ones(shape)
+        theta = .7*np.ones(shape)
         phi = None
         if len(shape) > 2:
-            phi = .1*np.ones(shape)
+            phi = .35*np.ones(shape)
 
         return Model(vp=v, origin=origin, shape=shape,
                      spacing=spacing, nbpml=nbpml,
--- a/examples/seismic/model.py
+++ b/examples/seismic/model.py
@@ -31,6 +31,27 @@ def demo_model(preset, **kwargs):
         return Model(vp=vp, origin=origin, shape=shape,
                      spacing=spacing, nbpml=nbpml, **kwargs)
 
+    elif preset.lower() in ['constanttti']:
+        # A constant single-layer model in a 2D or 3D domain
+        # with velocity 1.5km/s.
+        shape = kwargs.pop('shape', (101, 101))
+        spacing = kwargs.pop('spacing', tuple([10. for _ in shape]))
+        origin = kwargs.pop('origin', tuple([0. for _ in shape]))
+        nbpml = kwargs.pop('nbpml', 10)
+        v = np.empty(shape, dtype=np.float32)
+        v[:] = 1.5
+        epsilon = .3*np.ones(shape)
+        delta = .2*np.ones(shape)
+        theta = .2*np.ones(shape)
+        phi = None
+        if len(shape) > 2:
+            phi = .1*np.ones(shape)
+
+        return Model(vp=v, origin=origin, shape=shape,
+                     spacing=spacing, nbpml=nbpml,
+                     epsilon=epsilon, delta=delta, theta=theta, phi=phi,
+                     **kwargs)
+
     elif preset.lower() in ['layers', 'twolayer', '2layer']:
         # A two-layer model in a 2D or 3D domain with two different
         # velocities split across the height dimension:
@@ -70,8 +91,8 @@ def demo_model(preset, **kwargs):
         v[:] = vp_top  # Top velocity (background)
         v[..., int(shape[-1] / ratio):] = vp_bottom  # Bottom velocity
 
-        epsilon = .3*(v - 1.5)
-        delta = .2*(v - 1.5)
+        epsilon = .1*(v - 1.5)
+        delta = .05*(v - 1.5)
         theta = .5*(v - 1.5)
         phi = None
         if len(shape) > 2:
--- a/examples/seismic/tti/operators.py
+++ b/examples/seismic/tti/operators.py
@@ -26,12 +26,12 @@ def ForwardOperator(model, source, receiver, time_order=2, space_order=4,
 
     Stencil goes as follow
     For Hp compute Hz and Hp = laplace - Hz
-    For Hz compute Hp and Hz = laplace - Hp
+    For Hz compute directly Hz
 
     This guaranties that the diagonal of the operator will
     not be empty due to odd-even coupling. The FD operator
     is still self-adjoint for stability as both the laplacian and
-    the implementation of HP and HZ are self-adjoints
+    the implementation of Hp and Hz are self-adjoints
     """
     dt = model.critical_dt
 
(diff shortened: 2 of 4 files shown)


FabioLuporini:
I'll push the DSE fixes on top of this if that's OK with you

examples/seismic/tti/operators.py
```diff
@@ -17,6 +17,137 @@ def ForwardOperator(model, source, receiver, time_order=2, space_order=4,
...
+    stencils += rec.interpolate(expr=u + v, offset=model.nbpml)
+    # Add substitutions for spacing (temporal and spatial)
+    subs = dict([(t.spacing, dt)] + [(time.spacing, dt)] +
+                [(i.spacing, model.get_spacing()[j]) for i, j
+                 in zip(u.indices[1:], range(len(model.shape)))])
+    # Operator
+    return Operator(stencils, subs=subs, name='ForwardTTI', **kwargs)
+
+
+def ForwardOperator_legacy(model, source, receiver, time_order=2, space_order=4,
```
FabioLuporini: So here's a rule of thumb: any time you do a copy-paste which contains more than one line, that there's a non-negligible chance that things can be done in a much better way. 

Do I get it right : the difference between ``ForwardOperator`` and ``ForwardOperator_legacy`` is only in the symbolics, while the other lines are exactly the same. If that's the case, you have to refrain from copy-pasting and think about how we can factorize some code. In this case, one possibility is to have one function ``ForwardOperator`` that does the setup, and then two other functions -- one function per operator -- that take the same parameters (the symbols ``u, m, ...`` and return the symbolic equations. In fact, you could apply this same reasoning to the 2D specifications. 

Then you could have ``ForwardOperator(...)`` taking an additional name, representing what type of operator one wants to execute (give them significant names, not "legacy"); you can also have a dictionary mapping names to functions. Then you can imagine the rest; if the provided name is not in the dictionary, you raise an error; otherwise, you fetch the corresponding function

examples/seismic/tti/tti_example.py
```diff
@@ -2,33 +2,27 @@
...
     t0 = 0.0
     nt = int(1 + (tn-t0) / dt)
     time = np.linspace(t0, tn, nt)
 
     # Define source geometry (center of domain, just below surface)
-    src = GaborSource(name='src', ndim=model.dim, f0=0.01, time=time)
+    src = RickerSource(name='src', ndim=model.dim, f0=0.015, time=time)
     src.coordinates.data[0, :] = np.array(model.domain_size) * .5
-    src.coordinates.data[0, -1] = model.origin[-1] + 2 * spacing[-1]
+    # src.coordinates.data[0, -1] = model.origin[-1] + 2 * spacing[-1]
```
FabioLuporini: leftover

examples/seismic/tti/tti_example.py
```diff
@@ -45,7 +39,7 @@ def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
     autotune = kwargs.pop('autotune', False)
     solver = tti_setup(dimensions, spacing, tn, time_order, space_order, nbpml, **kwargs)
 
-    if space_order % 4 != 0:
+    if space_order % 2 != 0:
```
FabioLuporini: ?

examples/seismic/tti/tti_example.py
```diff
@@ -62,22 +56,24 @@ def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
                         help="Enable autotuning for block sizes")
     parser.add_argument("-to", "--time_order", default=2,
                         type=int, help="Time order of the simulation")
-    parser.add_argument("-so", "--space_order", default=4,
+    parser.add_argument("-so", "--space_order", default=8,
                         type=int, help="Space order of the simulation")
-    parser.add_argument("--nbpml", default=10,
+    parser.add_argument("--nbpml", default=40,
```
FabioLuporini: why?

examples/seismic/tti/wavesolver.py
```diff
@@ -51,6 +47,20 @@ def op_fwd_save(self):
                                receiver=self.receiver, time_order=self.time_order,
                                space_order=self.space_order, **self._kwargs)
 
+    @cached_property
+    def op_fwd_legacy(self):
```
FabioLuporini: again, you see, this is a combinational explosion of possibilities

tests/test_dse.py
```diff
@@ -134,6 +134,7 @@ def test_tti_rewrite_basic(tti_nodse):
     assert np.allclose(tti_nodse[1].data, rec.data, atol=10e-3)
 
 
+@pytest.mark.xfail(reason="DSE mode not supported for TTI")
```
FabioLuporini: OK, this needs fixing before merging. And I know it's on me.

tests/test_tti.py
```diff
@@ -73,7 +73,8 @@ def ricker_source(t, f0):
                                   time_order=2, space_order=space_order)
 
     solver_tti = AnisotropicWaveSolver(model, source=source, receiver=receiver,
-                                       time_order=2, space_order=space_order)
+                                       time_order=2, space_order=space_order,
+                                       dse='basic')
```
FabioLuporini: I guess we won't need to have ``dse`` here as soon as the advanced mode bug is fixed

examples/seismic/tti/wavesolver.py
```diff
@@ -113,3 +123,66 @@ def forward(self, src=None, rec=None, u=None, v=None, m=None,
             summary = op.apply(src=src, rec=rec, u=u, v=v, m=m, epsilon=epsilon,
                                delta=delta, theta=theta, phi=phi, **kwargs)
         return rec, u, v, summary
+
+    def forward_legacy(self, src=None, rec=None, u=None, v=None, m=None,
```
FabioLuporini: again, so much duplication

examples/seismic/tti/tti_example.py
```diff
@@ -62,22 +56,24 @@ def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
...
                         type=int, help="Number of PML layers around the domain")
     args = parser.parse_args()
 
     # 3D preset parameters
```
FabioLuporini: misleading comment

examples/seismic/tti/tti_example.py
```diff
@@ -62,22 +56,24 @@ def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
...
+        tn = 500.0
+        dse = 'basic'
     else:
-        dimensions = (50, 50, 50)
-        spacing = (20.0, 20.0, 20.0)
-        tn = 250.0
+        dimensions = (100, 100, 100)
+        spacing = (10.0, 10.0, 10.0)
+        tn = 400.0
+        dse = 'aggressive'
```
FabioLuporini: Can you not just change the default values in the function signature ? 
I wouldn't set ``dse`` here; you can add it as a parser option if you really want.

Also, can we not fetch some of these params from the model ?

examples/seismic/tti/tti_example.py
mloubout: ABC coefficients designed for 40, so better leave it that way, can put it back to 10.

examples/seismic/tti/tti_example.py
mloubout: only shows the full diff from size changes, dse here due to current 2d/3d issues with dse, will go away once figured out

examples/seismic/tti/wavesolver.py
mloubout: Different Kernel, different function, that was my idea

examples/seismic/tti/tti_example.py
mloubout: >Can we not fetch some of these params from the model ?
Well no, this params create the model
>Can you not just change the default values in the function signature ?
Nothing changed from before, only the dse due to the current 2d/3d behavior bugs

examples/seismic/tti/wavesolver.py
mloubout: This one is not changeable, as it is a cached property both need to exist or can't change kernel once the operator created

examples/seismic/model.py
```diff
@@ -98,6 +149,89 @@ def demo_model(preset, **kwargs):
         return Model(vp=v, origin=origin, shape=v.shape,
                      spacing=spacing, nbpml=20)
 
+    elif preset.lower() in ['marmousitti', 'marmousi2dtti']:
```
FabioLuporini: So this is for TTI and the other for Acoustic ?

examples/seismic/tti/operators.py
```diff
@@ -35,124 +379,42 @@ def ForwardOperator(model, source, receiver, time_order=2, space_order=4,
...
         (4.0 * m * v + (s * damp - 2.0 * m) *
-         v.backward + 2.0 * s**2 * (delta * Hp + Hzr))
-
-    # Add substitutions for spacing (temporal and spatial)
+         v.backward + 2.0 * s ** 2 * (delta * H0 + Hz))
     first_stencil = Eq(u.forward, stencilp)
     second_stencil = Eq(v.forward, stencilr)
     stencils = [first_stencil, second_stencil]
 
+    # SOurce and receivers
```
FabioLuporini: Sources and receivers

examples/seismic/tti/operators.py
```diff
@@ -35,124 +379,42 @@ def ForwardOperator(model, source, receiver, time_order=2, space_order=4,
...
     stencils += src.inject(field=u.forward, expr=src * dt * dt / m,
                            offset=model.nbpml)
     stencils += src.inject(field=v.forward, expr=src * dt * dt / m,
                            offset=model.nbpml)
     stencils += rec.interpolate(expr=u + v, offset=model.nbpml)
+    # Add substitutions for spacing (temporal and spatial)
     subs = dict([(t.spacing, dt)] + [(time.spacing, dt)] +
                 [(i.spacing, model.get_spacing()[j]) for i, j
                  in zip(u.indices[1:], range(len(model.shape)))])
+    # Operator
```
FabioLuporini: useless

examples/seismic/tti/operators.py
```diff
@@ -35,124 +379,42 @@ def ForwardOperator(model, source, receiver, time_order=2, space_order=4,
...
     return Operator(stencils, subs=subs, name='ForwardTTI', **kwargs)
+
+
+kernels = {('shifted', 3): kernel_shited_3d, ('shifted', 2): kernel_shited_2d,
```
FabioLuporini: Docstring ?

examples/seismic/tti/wavesolver.py
```diff
@@ -27,33 +27,51 @@ def __init__(self, model, source, receiver,
...
         self.dt = self.model.critical_dt
-        if self.time_order == 4:
-            self.dt *= 1.73
 
         # Cache compiler options
         self._kwargs = kwargs
 
     @cached_property
-    def op_fwd(self):
+    def op_fwd_shifted(self):
```
FabioLuporini: So instead of having 4 *properties* ``op_fwd_shifted, op_fwd_shifted_save, op_fwd_centered, op_fwd_centered_save``, can we not have one *method* taking two optional parameters ``kernel`` and ``save`` ?

examples/seismic/tti/wavesolver.py
```diff
@@ -27,33 +27,51 @@ def __init__(self, model, source, receiver,
...
+    @cached_property
+    def op_fwd_centered_save(self):
+        """Cached operator for forward runs with unrolled wavefield"""
+        return ForwardOperator(self.model, save=True, source=self.source,
+                               receiver=self.receiver, time_order=self.time_order,
+                               space_order=self.space_order,
+                               kernel='centered', **self._kwargs)
+
+    def operator(self, kernel, save):
+        func_name = 'op_fwd_' + '%s' % kernel + ('save' if save else '')
```
FabioLuporini: Better:
``func_name = "op_fwd_%s%s" % (kernel, 'save' if save else '')``
But, as I said above, I'd rather have a single method to be called as
``self.fwd_op(kernel, save)``

tests/test_tti.py
```diff
@@ -93,8 +94,8 @@ def ricker_source(t, f0):
                       .5 * u_tti.data.reshape(-1) - .5 * v_tti.data.reshape(-1))
     res /= linalg.norm(u.data.reshape(-1))
     log("Difference between acoustic and TTI with all coefficients to 0 %f" % res)
-    assert np.isclose(res, 0.0, atol=1e-1)
+    assert np.isclose(res, 0.0, atol=1e-4)
```
FabioLuporini: really?

examples/seismic/model.py
mloubout: yes

examples/seismic/tti/wavesolver.py
mloubout: No idea how to make it a cached property that way

tests/test_tti.py
mloubout: Yes,centred is more accurate for weak (small) anisotropy

examples/seismic/tti/operators.py
mloubout: docstrin in each kerel

examples/seismic/tti/operators.py
mloubout: Don't agree, help separating setup parts

examples/seismic/model.py
FabioLuporini: should we then not have two different models, one for acoustic and one for tti (to be put in their respective folders) ? The common parts could then be left in a common module

examples/seismic/tti/operators.py
FabioLuporini: You can leave blank lines, but fair enough, it's just nitpicking

examples/seismic/tti/operators.py
FabioLuporini: 👍

tests/test_tti.py
FabioLuporini: centered is what we had before?

examples/seismic/model.py
mloubout: Well technically TTI is acoustic...
But separating model would be not very helpful in my opinion, havibg a set of "physical" example model in the same place makes it eaiser to play around

examples/seismic/tti/wavesolver.py
mloubout: 10K sources same model, no need to regenerate the operator

examples/seismic/tti/wavesolver.py
mloubout: I think it is linked to cached/not_cahed properties

tests/test_tti.py
mloubout: Shifted is what we had before

examples/seismic/model.py
FabioLuporini: I'm inclined to disagree. However...

let's at least be more consistent with names. 
``constant`` --> ``constant_acoustic` `
``constanttti`` --> ``constant_tti``
and so on and so forth.
This will make the 1-to-1 relationship with the ``acoustic`` and ``tti`` problems more clear

examples/seismic/tti/wavesolver.py
FabioLuporini: OK, I see.

What you're looking for is called "memoization" (not a Python-specific concept).
Here a Python example:
https://stackoverflow.com/questions/1988804/what-is-memoization-and-how-can-i-use-it-in-python

It's much much simpler than it seems/sounds

I recommend to adopt the decorator solution (e.g., @memoized) and put it into ``devito/utils.py`` . If too complicated, we can look at this together early next week

mloubout: > and put it into devito/utils.py
Then it's not an example anymore, couldn't disagree more this is gonna  make `examples` a nightmare. Remember that most user will NOT be computer scientist and have about 5-10 min to look at Devito, if I need more than that knowing the software we are shooting ourselves in the leg.
Examples are showcasing what we can do with devito and some repetition is not a problem as you see what's waht at first glance without having to dig into stuff.

examples/seismic/tti/wavesolver.py
mloubout: >But, as I said above, I'd rather have a single method

No

examples/seismic/tti/wavesolver.py
FabioLuporini: > Then it's not an example anymore

The subject was the ``memoizer`` ... and ``utils`` was definitely the right place for it. A ``memoizer`` is a general thing.

So... Are you really saying that putting a @memoize decorator on top of a function would make an example less readable?

@mlange05, can you take a look at this?

examples/seismic/tti/tti_example.py
```diff
@@ -41,14 +35,14 @@ def tti_setup(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
 
 
 def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
-        time_order=2, space_order=4, nbpml=10, **kwargs):
+        time_order=2, space_order=4, nbpml=10, kernel='centered', **kwargs):
     autotune = kwargs.pop('autotune', False)
```
FabioLuporini: you've undone ``autotune`` in the signature

FabioLuporini:
Have you changed the ``shifted`` TTI version operators in this PR ? The operation counts returned are different that those from master.

FabioLuporini:
@mloubout : pushed a potential fix here:  ``new-tti-again-dsefixes ``

examples/seismic/tti/wavesolver.py
mlange05: :+1: for `@memoize` - that's exactly what the solver object is effectively doing. It should be a copy-and-past job to provide the decorator (properly linked of course).

mloubout: > It should be a copy-and-past job to provide the decorator

So repalce a copy paste by a copy paste.....

examples/seismic/tti/tti_example.py
mloubout: Not part of the PR so yes

mloubout:
>Have you changed the shifted TTI version operators in this PR ? The operation counts returned are different that those from master.

Shifted uses double the previous order.

FabioLuporini:
> Shifted uses double the previous order.

So if I run this:
```
python tti_example.py -k shifted -so 4
```
You're saying I instead get ``so=8`` (and I can now appreciate that). So we need a warning/info message to tell the user what's happening, and why. Or we should explicitly forbid the combination ``shifted`` + ``so=4``

mloubout:
>Or we should explicitly forbid the combination shifted + so=4

What do you mean forbid? what order would you run?

The main difference is that previously I was halving the space_order in the init function. Now I leave it alone, so it end up with 4th order of 4th order and is 8th order overall

examples/seismic/model.py
```diff
@@ -52,7 +74,37 @@ def demo_model(preset, **kwargs):
         return Model(vp=v, origin=origin, shape=shape,
                      spacing=spacing, nbpml=nbpml, **kwargs)
 
-    elif preset.lower() in ['circle']:
+    elif preset.lower() in ['layerstti', 'twolayertti', '2layertti']:
```
FabioLuporini: consistency with the -isotropic names: put a '-' in all names right before 'tti'

FabioLuporini:
> The main difference is that previously I was halving the space_order in the init function. Now I leave it alone, so it end up with 4th order of 4th order and is 8th order overall

What I don't like is the fact that for the same value of ``-so``, ``shifted`` and ``centered`` seem to be using different discretization orders

FabioLuporini:
@mloubout I took the liberty of pushing a further tiny fix to the DSE

mloubout:
>What I don't like is the fact that for the same value of -so, shifted and centered seem to be using different discretization orders

Fixed it, space_order is redefine properly according to the Kernel used.

examples/seismic/model.py
```diff
@@ -98,6 +150,89 @@ def demo_model(preset, **kwargs):
...
+        # Read 2D Marmousi model from opesc/data repo
+        data_path = kwargs.pop('data_path', None)
+        if data_path is None:
+            error("Path to opesci/data not found! Please specify with "
+                  "'data_path=<path/to/opesci/data>'")
+            raise ValueError("Path to model data unspecified")
+        path = os.path.join(data_path, 'marmousi3D/vp_marmousi_bi')
+
+        # Velcoity
+        vp = 1e-3 * np.fromfile(os.path.join(data_path, 'marmousi3D/MarmousiVP.raw'),
```
FabioLuporini: How does one get these ``.raw`` files ? Should we add an optional requirement to Devito to run these examples ? ( @mlange05 )

examples/seismic/tti/tti_example.py
```diff
@@ -64,20 +58,24 @@ def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
                         type=int, help="Time order of the simulation")
     parser.add_argument("-so", "--space_order", default=4,
                         type=int, help="Space order of the simulation")
-    parser.add_argument("--nbpml", default=10,
+    parser.add_argument("--nbpml", default=40,
```
FabioLuporini: remind me why is that?

tests/test_tti.py
```diff
@@ -61,7 +61,7 @@ def ricker_source(t, f0):
                                   time_order=2, space_order=space_order)
     rec, u1, _ = acoustic.forward(save=False)
 
-    tn = 50.0
+    tn = 100.0
```
FabioLuporini: why?

examples/seismic/model.py
mloubout: https://github.com/opesci/data/tree/master/marmousi3D

mloubout: Can add requirement if needed

examples/seismic/tti/tti_example.py
mloubout: Even if 10 doesn't break anything, ABC (pml) coefficients are designed for 10, and on top of that, kinda make sure we "always" run at least the full model for high order (so 8 with shifted is already out with nbpml=10)

tests/test_tti.py
mloubout: 50 doesn't cover much more than the source start, I am not a big fan of running/benchmarking on zeros only. even 100 is small but bigger would mess up TRAVIS

mloubout:
>I think the last standing item here is the memoizer

I looked at it, still no idea how it works/what to do

examples/seismic/tti/tti_example.py
```diff
@@ -64,20 +58,24 @@ def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
...
                         type=int, help="Number of PML layers around the domain")
+    parser.add_argument("-k", dest="kernel", default='centered',
+                        type=str, help="Choice of finite-difference kernel")
```
mlange05: If it's a choice, you should make that explicit with `choices=['a', 'b', 'c']` instead of `type=str`. This way the script documentation will list the possible options.

examples/seismic/model.py
```diff
@@ -31,7 +31,29 @@ def demo_model(preset, **kwargs):
...
+        phi = None
+        if len(shape) > 2:
+            phi = .35*np.ones(shape)
+
+        return Model(vp=v, origin=origin, shape=shape,
+                     spacing=spacing, nbpml=nbpml,
+                     epsilon=epsilon, delta=delta, theta=theta, phi=phi,
+                     **kwargs)
+
+    elif preset.lower() in ['layers-isotropic', 'twolayer-isotropic',
```
FabioLuporini: single line?

examples/seismic/model.py
```diff
@@ -98,6 +150,89 @@ def demo_model(preset, **kwargs):
         return Model(vp=v, origin=origin, shape=v.shape,
                      spacing=spacing, nbpml=20)
 
+    elif preset.lower() in ['marmousitti', 'marmousi2dtti']:
```
FabioLuporini: based on the convention you used for marmousi 3d, you should have here:

``marmousi-tti marmousi-tti2d marmousi2d-tti``

examples/seismic/model.py
FabioLuporini: @mlange05 perhaps we can have it optional ? and update the error message above accordingly?

examples/seismic/tti/tti_example.py
```diff
@@ -64,20 +58,24 @@ def run(dimensions=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
...
+    parser.add_argument("-dse", dest="dse", default='advanced',
+                        type=str, help="Choice of dse backend")
     args = parser.parse_args()
 
     # 3D preset parameters
     if args.dim2:
```
FabioLuporini: dumb question: should we not add these (dimensions, shaping, tn) to the arg parser too ?

examples/seismic/model.py
```diff
@@ -98,6 +150,89 @@ def demo_model(preset, **kwargs):
...
+
+        dimensions = (201, 201, 70)
```
FabioLuporini: why do we need both ``dimensions`` and ``shape``? btw, usually (not always, unfortunately, but most of the times yes, I think) we use ``dimensions`` to represent the symbolic dimensions ``x, y, z``; here it's just numbers

examples/seismic/tti/wavesolver.py
```diff
@@ -27,33 +25,23 @@ def __init__(self, model, source, receiver,
...
         self.dt = self.model.critical_dt
-        if self.time_order == 4:
-            self.dt *= 1.73
 
         # Cache compiler options
         self._kwargs = kwargs
 
-    @cached_property
-    def op_fwd(self):
+    @memoized
```
FabioLuporini: excellent

examples/seismic/tti/wavesolver.py
```diff
@@ -63,9 +51,15 @@ def forward(self, src=None, rec=None, u=None, v=None, m=None,
...
         :param m: (Optional) Symbol for the time-constant square slowness
         :param save: Option to store the entire (unrolled) wavefield
+        :param kernel: type of discretization, centered or shifted
 
         :returns: Receiver, wavefield and performance summary
         """
+
+        # Space order needs to be halved in the shifted case to have an
+        # overall space_order discretization
+        self.space_order = self.space_order / 2 if kernel == 'shifted' \
```
FabioLuporini: should we rather to this straight away in the constructor?

examples/seismic/tti/wavesolver.py
```diff
@@ -84,7 +78,7 @@ def forward(self, src=None, rec=None, u=None, v=None, m=None,
         # Create the forward wavefield if not provided
         if v is None:
             v = TimeData(name='v', shape=self.model.shape_domain,
-                         save=save, time_dim=self.source.nt,
+                         save=save, time_dim=self.source.nt if save else None,
```
FabioLuporini: same here: should we not compute this information once and for all in the constructor?

tests/test_adjointJ.py
```diff
@@ -17,7 +17,7 @@ def test_acousticJ(dimensions, space_order):
     spacing = [15. for _ in dimensions]
```
FabioLuporini: you see this should be called ``shape`` not ``dimensions``

devito/__init__.py
```diff
@@ -11,7 +11,7 @@
 from devito.logger import error, warning, info  # noqa
 from devito.parameters import (configuration, init_configuration,  # noqa
                                env_vars_mapper)
-
+from devito.tools import *  # noqa
```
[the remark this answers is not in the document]
mlange05: Nope, intentional. Needed to use `@memoized` in external `example` directory without absolute imports.

devito/tools.py
```diff
@@ -9,6 +10,8 @@
 
 import numpy as np
 
+__all__ = ['memoized']
```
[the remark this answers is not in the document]
mlange05: Also, intentional. Trying to reduce namespace clutter for `from devito import *`. Can/should be extended as needed.

examples/seismic/model.py
mlange05: A legacy convention from the early seismic prototypes, before `Dimension` objects were introduced. Should probably be changed everywhere (loads more outside TTI), but probably beyond this PR; and should disappear once a `Grid` class gets introduced.

examples/seismic/tti/wavesolver.py
mlange05: In this case no, as save is an invocation time parameter, and many tests run through both sides of this conditional with the same `solver` object.

tests/test_adjointJ.py
mlange05: See above.

examples/seismic/tti/wavesolver.py
mloubout: Runtime argument, the constructor can't know which kernel will be run.

examples/seismic/model.py
mloubout: too long

examples/seismic/tti/tti_example.py
mloubout: We could, not quite part of this PR, already "too much" non-tti related stuff

FabioLuporini:
We can probably land this once rebased on master. We'll leave my nitpicking for another time. It'd be nice if, in another PR, we could apply memoization to acoustic too

mloubout:
> It'd be nice if, in another PR, we could apply memoization to acoustic too

I would agree. I personally don't mind holding that one off for acoustic. Unfortunately, this PR contains DSE fix that are necessary to run TTI, so I am fine fixing acoustic over the weekend (meaning me), but can't wait much longer

FabioLuporini:
Yes, please, just rebase this branch on top of master; then I'll merge it as soon as Travis is green

FabioLuporini:
OK, merging...

FabioLuporini:
Merged.

PR #352, opened 2017-09-18
yask: More fixes and change semantics of data.shape

FabioLuporini:
Two critical fixes in the **YASK backend**:

- The following is now supported, as expected in Devito:
```
u0 = TimeData(name='u', shape=S, dimensions=D, ...)
u1 = TimeData(name='u', shape=S, dimensions=D, ...)
``` 
Clearly u0 and u1 cannot be used in the same Operator. Previously, attempting to create two shared object with same name/shape/dimensions would have raised an error

- A data object shape now corresponds to the "domain size"; the halo, inferred from ``space_order``, is "constructed" on top of the shape, not by removing inner points. This matches the YASK behavior. This represents an API divergence points w.r.t. the default backend. However, we expect the default backend to adopt this semantics as well in the (close) future, in particular when the Grid abstraction will become available. The reasons why this was necessary have been discussed at length offline (Slack #yask and #general channels).

There's also some refactoring going on with this PR. 

Again, this PR has many changes because the YASK backend is still a prototype, which is continuously evolving. Once the YASK backend will be released, PRs will go back being more sane.

### DIFF
--- a/devito/interfaces.py
+++ b/devito/interfaces.py
@@ -469,7 +469,7 @@ class DenseData(TensorData):
                 error("Creating symbolic data objects requries either"
                       "a 'shape' or 'dimensions' argument")
                 raise ValueError("Unknown symbol dimensions or shape")
-            _indices = [x, y, z]
+            _indices = (x, y, z)
             shape = kwargs.get('shape')
             if len(shape) <= 3:
                 dimensions = _indices[:len(shape)]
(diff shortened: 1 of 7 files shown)


FabioLuporini:
Can any of @navjotk or @mlange05 just sign this off ? The YASK backend is continuously evolving and this PR is *not* touching any code outside of ``/yask``, so...

FabioLuporini:
Thanks! merging...

FabioLuporini:
Merged.

PR #354, opened 2017-09-19
Bring back FixedDimension

navjotk:
Reverts opesci/devito#346

### DIFF
--- a/devito/dimension.py
+++ b/devito/dimension.py
@@ -1,26 +1,25 @@
 from sympy import Number, Symbol
-from devito.arguments import DimensionArgProvider
+from devito.arguments import DimensionArgProvider, FixedDimensionArgProvider
 
-__all__ = ['Dimension', 'x', 'y', 'z', 't', 'p', 'd', 'time']
+__all__ = ['Dimension', 'FixedDimension', 'x', 'y', 'z', 't', 'p', 'd', 'time']
 
 
 class Dimension(Symbol, DimensionArgProvider):
+
+    is_Buffered = False
+    is_Lowered = False
+    is_Fixed = False
 
     """Index object that represents a problem dimension and thus
     defines a potential iteration space.
 
     :param size: Optional, size of the array dimension.
     :param reverse: Traverse dimension in reverse order (default False)
-    :param buffered: Optional, boolean flag indicating whether to
-                     buffer variables when iterating this dimension.
+    :param spacing: Optional, symbol for the spacing along this dimension.
     """
-
-    is_Buffered = False
-    is_Lowered = False
 
     def __new__(cls, name, **kwargs):
         newobj = Symbol.__new__(cls, name)
-        newobj.size = kwargs.get('size', None)
         newobj.reverse = kwargs.get('reverse', False)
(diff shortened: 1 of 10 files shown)


tests/test_dimension.py
```diff
@@ -0,0 +1,17 @@
...
+@pytest.mark.xfail
+def test_incorrect_usage():
+    m = DenseData(name="m", shape=(10, 10))
+    devito.x = FixedDimension(name='x', size=10, spacing=devito.x.spacing)
+    assert(devito.x in m.indices)
+
+
+def test_correct_usage():
+    myx = FixedDimension(name='x', size=10, spacing=devito.x.spacing)
+    m = DenseData(name="m", dimensions=(myx, y))
```
mloubout: `m.laplace` ?

tests/test_adjointA.py
```diff
@@ -58,7 +59,8 @@ def test_acoustic(mkey, dimensions, time_order, space_order, nbpml, fix_dim):
                                 space_order=space_order)
 
     # Set fixed ("baked-in") time dimension if requested
-    time.size = solver.source.nt if fix_dim else None
+    if fix_dim:
+        devito.time = FixedDimension('time', size=solver.source.nt)
```
mloubout: Incorrect usage according to the test.

devito/dimension.py
```diff
@@ -31,10 +30,38 @@ def __str__(self):
...
+
+    def __new__(cls, name, **kwargs):
+        newobj = super(FixedDimension, cls).__new__(cls, name)
+        newobj._size = kwargs.get('size', None)
+        return newobj
+
+    @property
+    def symbolic_size(self):
+        """The symbolic size of this dimension."""
+        return Number(self.size)
```
mloubout: This is the only difference with a non-fixed dimension (unless I am missing something). Wouldn't be easier/nicer to only have a "conditional" property?

devito/operator.py
```diff
@@ -198,7 +198,9 @@ def _dle_arguments(self, dim_sizes):
         dle_arguments = OrderedDict()
         autotune = True
         for i in self.dle_arguments:
-            dim_size = dim_sizes.get(i.original_dim.name, i.original_dim.size)
+            dim_size = dim_sizes.get(i.original_dim.name,
+                                     i.original_dim.size if i.original_dim.is_Fixed
```
FabioLuporini: Perhaps we don't need this ``if``, because if ``False`` it's gonna return ``None`` anyway?

devito/dimension.py
```diff
@@ -31,10 +30,38 @@ def __str__(self):
     @property
     def symbolic_size(self):
         """The symbolic size of this dimension."""
-        try:
-            return Number(self.size)
-        except TypeError:
-            return self.rtargs[0].as_symbol
+        return self.rtargs[0].as_symbol
```
FabioLuporini: Suggestion for the future: perhaps we wanna return a ``namedtuple`` when it comes to the ``Dimension``'s ``rtargs``

devito/operator.py
FabioLuporini: @navjotk

PR #355, opened 2017-09-19
yask: Fixes to make the acoustic wave equation test take off

FabioLuporini:
This will have to be rebased and should go *after* #352

### DIFF
--- a/tests/test_yask.py
+++ b/tests/test_yask.py
@@ -162,6 +162,20 @@ class TestOperatorExecution(object):
         assert all(np.all(u.data.with_halo[1, :, i, :] == 0) for i in range(space_order))
         assert all(np.all(u.data.with_halo[1, :, :, i] == 0) for i in range(space_order))
 
+    def test_increasing_multi_steps(self):
+        """
+        Apply the trivial equation ``u[t+1,x,y,z] = u[t,x,y,z] + 1`` for 11
+        timesteps and check that all grid domain values are equal to 11 within
+        ``u[1]`` and equal to 10 within ``u[0]``.
+        """
+        u = TimeData(name='yu4D', shape=(8, 8, 8), dimensions=(x, y, z), space_order=0)
+        u.data.with_halo[:] = 0.
+        op = Operator(Eq(u.forward, u + 1.), subs={t.spacing: 1})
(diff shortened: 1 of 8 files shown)


FabioLuporini:
Rebased; this is the next one in the YASK line

tests/test_yask.py
```diff
@@ -293,51 +307,102 @@ class TestOperatorRealAcoustic(object):
...
+        results when running the same model as in ``test_adjointA.py``.
+        """
+        dt = model.critical_dt
+        u.data[:] = 0.0
+        eqns = [Eq(u.forward, stencil)]
+        eqns += src.inject(field=u.forward, expr=src * dt**2 / m, offset=model.nbpml)
+        eqns += rec.interpolate(expr=u, offset=model.nbpml)
+        op = Operator(eqns, subs=subs)
+        op.apply(u=u, m=m, damp=damp, src=src, rec=rec, t=1)
+        # TODO: come up with proper asserts, but u.data looks sane already
```
navjotk: The assert could just be a known value of the L2 norm of `u` associated with the model/damp/.... setup

mloubout: > # TODO: come up with proper asserts, but u.data looks sane already
with t=1? everything will still be zero in `u` so not much to assert on. 

One possibility would be to assert against non-yask solution.
As it looks like everything is nicely setup, you could also run adjointTest on a single space_orde/time_order combination.

FabioLuporini: I cannot run adjoint yet because of fix required in YASK

> That would work too for more time step but would restrict it to a the part of the domain where we are sure both YASK and non-YASK have solution value (remove PML part for example)

I can think of something along these lines

FabioLuporini: BTW, to check correctness I was doing exactly that (only using 1-norm instead of 2-norm). Not sure why I initially opted to leave this kind of check out. Now implemented using 2-norms and comparing to the values obtained from an analogous test from test_adjointA

FabioLuporini:
Addressed comments and rebased. Will merge once travis is happy

FabioLuporini:
Comments addressed and approved, so merging...

FabioLuporini:
Merged.

PR #356, opened 2017-09-21
Move benchmark.py to /seismic and improvements

FabioLuporini:


mlange05:
Overall, this looks alright, so fine by me.

One suggestion for future improvements though: I think we might want to consider using the [click](http://click.pocoo.org/5/) package to improve the argument handling for the different modes, as the complexity of this has gone slightly out of hand and is making this hard to read and understand.

FabioLuporini:
> we might want to consider using the click package

agreed.

Shall we create a GitHub issue for all those "nice-to-have-but-kinda-low-priority" features that a UROP first-year student could perhaps do?

FabioLuporini:
Merged.

Issue #358, opened 2017-09-23
Examples cleanup


FabioLuporini:
- Have a single ``ArgumentParser``, used by both acoustic and TTI and any other future seismic equations. Perhaps to be moved to a ``utils.py`` module under ``examples/seismic``
- Use memoization in acoustic, in the same way we did for TTI
- Move legacy leftovers of the following sort to ``model``:
```
    # 3D preset parameters
     if args.dim2:
        shape = (150, 150)
         spacing = (15.0, 15.0)
         tn = 750.0
     else:
        shape = (50, 50, 50)
        spacing = (20.0, 20.0, 20.0)
        tn = 250.0
```

mloubout:
On it

FabioLuporini:
Also:
- raise an error if one passes an non-existing model name to ``demo_model()``. I just bumped into this issue

mloubout:
`demo_model` where is this one coming from?

FabioLuporini:
as I wrote above, my mistake. However, you could still improve the ``demo_model`` docstring in ``examples/seismic/model.py``, as it still shows a description for ``layer2d`` and ``marmousi2d`` that are now unrecognised names.

Thanks

mloubout:
We can close this no?

FabioLuporini:
once #360 is in?

FabioLuporini:
Done, closed

PR #360, opened 2017-09-25
Examples cleanup

mloubout:
Cleanup according to #358 

> have a single ArgumentParser, used by both acoustic and TTI and any other future seismic equations. Perhaps to be moved to a utils.py module under examples/seismic

Not sure about this one, it would make it easier to run/benchmark for us but will start to make the example inaccessible for a lot of users IMO. 

> Use memoization in acoustic, in the same way we did for TTI

Done

> Move legacy leftovers of the following sort to model:

This will end up in `if-else` somewhere anyway, not sure it should/where it sould be moved.

`demo_model` docstring updated

### DIFF
--- a/examples/seismic/acoustic/operators.py
+++ b/examples/seismic/acoustic/operators.py
@@ -75,7 +75,7 @@ def ForwardOperator(model, source, receiver, time_order=2, space_order=4,
     # Create symbols for forward wavefield, source and receivers
     u = TimeData(name='u', shape=model.shape_domain, dtype=model.dtype,
                  save=save, time_dim=source.nt if save else None,
-                 time_order=time_order, space_order=space_order)
+                 time_order=2, space_order=space_order)
     src = PointSource(name='src', ntime=source.nt, ndim=source.ndim,
                       npoint=source.npoint)
     rec = Receiver(name='rec', ntime=receiver.nt, ndim=receiver.ndim,
@@ -115,7 +115,7 @@ def AdjointOperator(model, source, receiver, time_order=2, space_order=4, **kwar
     m, damp = model.m, model.damp
 
     v = TimeData(name='v', shape=model.shape_domain, save=False,
-                 time_order=time_order, space_order=space_order,
+                 time_order=2, space_order=space_order,
                  dtype=model.dtype)
     srca = PointSource(name='srca', ntime=source.nt, ndim=source.ndim,
                        npoint=source.npoint)
@@ -158,10 +158,10 @@ def GradientOperator(model, source, receiver, time_order=2, space_order=4, **kwa
     grad = DenseData(name='grad', shape=model.shape_domain,
                      dtype=model.dtype)
     u = TimeData(name='u', shape=model.shape_domain, save=True,
-                 time_dim=source.nt, time_order=time_order,
+                 time_dim=source.nt, time_order=2,
                  space_order=space_order, dtype=model.dtype)
     v = TimeData(name='v', shape=model.shape_domain, save=False,
-                 time_order=time_order, space_order=space_order,
+                 time_order=2, space_order=space_order,
                  dtype=model.dtype)
     rec = Receiver(name='rec', ntime=receiver.nt, ndim=receiver.ndim,
                    npoint=receiver.npoint)
@@ -210,10 +210,10 @@ def BornOperator(model, source, receiver, time_order=2, space_order=4, **kwargs)
 
     # Create wavefields and a dm field
     u = TimeData(name="u", shape=model.shape_domain, save=False,
-                 time_order=time_order, space_order=space_order,
+                 time_order=2, space_order=space_order,
                  dtype=model.dtype)
     U = TimeData(name="U", shape=model.shape_domain, save=False,
-                 time_order=time_order, space_order=space_order,
+                 time_order=2, space_order=space_order,
                  dtype=model.dtype)
     dm = DenseData(name="dm", shape=model.shape_domain,
                    dtype=model.dtype)
--- a/examples/seismic/acoustic/operators.py
+++ b/examples/seismic/acoustic/operators.py
@@ -14,7 +14,7 @@ def laplacian(field, time_order, m, s):
     :param field:  Symbolic TimeData object, solution to be computed
     :param time_order: time order
     :param m: square slowness
-    :param s: symbol of for the time-step
+    :param s: symbol for the time-step
     :return: H
     """
     if time_order == 2:
@@ -34,7 +34,7 @@ def iso_stencil(field, time_order, m, s, damp, **kwargs):
     :param field: Symbolic TimeData object, solution to be computed
     :param time_order: time order
     :param m: square slowness
-    :param s: symbol of for the time-step
+    :param s: symbol for the time-step
     :param damp: ABC dampening field (DenseData)
     :param kwargs: forwad/backward wave equation (sign of u.dt will change accordingly
     as well as the updated time-step (u.forwad or u.backward)
--- a/examples/seismic/acoustic/operators.py
+++ b/examples/seismic/acoustic/operators.py
@@ -47,7 +47,7 @@ def iso_stencil(field, time_order, m, s, damp, **kwargs):
     # Define time sep to be updated
     next = field.forward if kwargs.get('forward', True) else field.backward
     # Define PDE
-    eq = m * field.dt2 - H + kwargs.get('q', 0)
+    eq = m * field.dt2 - H - kwargs.get('q', 0)
     # Add dampening field according to the propagation direction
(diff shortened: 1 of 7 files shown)


FabioLuporini:
The edits are uncontroversial, thanks. I have one question though:

> Not sure about this one, it would make it easier to run/benchmark for us but will start to make the example inaccessible for a lot of users IMO.

why would it make the examples less accessible?

mloubout:
It would split the examples. I find it easier to have one file, one example and avoid mixing different ones.

examples/seismic/acoustic/wavesolver.py
```diff
@@ -38,21 +42,14 @@ def __init__(self, model, source, receiver,
         # Cache compiler options
         self._kwargs = kwargs
 
-    @cached_property
-    def op_fwd(self):
+    @memoized
+    def op_fwd(self, save=False):
```
mlange05: We probably want to use `@memoized` for all other operator types too, for consistency. Unless I'm missing something, this will also allows us to drop the dependency on `cached_property` completely. 

We probably also want to drop the `save` keyword when calling the memoized operators, since it's not actually used and just swallowed with no effect by `kwargs`.

examples/seismic/acoustic/wavesolver.py
```diff
@@ -27,7 +28,10 @@ def __init__(self, model, source, receiver,
         self.source = source
         self.receiver = receiver
 
-        self.time_order = time_order
+        self.time_order = 2
```
mlange05: I don't think this is right, because the value forced here is ultimately the value passed to the operator constructors, so we will always run 2nd order. Do you mean to always set the `time_order=2` for the wavefield symbols `u` and `v`, instead, even if we're running 4th order?

examples/seismic/acoustic/wavesolver.py
mloubout: I initially tried `memoized` everywhere, but it seems to not like functions without `kwargs`.

>We probably also want to drop the save keyword when calling the memoized 

Wouldn't it ignore it? We would still need a new operator and generate new code if `save` changes.

examples/seismic/acoustic/wavesolver.py
mloubout: >so we will always run 2nd order

For this wave equation this is the case. There is some high-order solvers but these won;t come from "pure FD" but from some multi-level time-stepping schemes.

We are still running a second order in time wave-equation, but it's formulation allows a 4th order discretization error (O(dt^4) but only d/dt^2). It would be risky in my opinion to leave it a free parameter that would lead people to do bad stuff, specially now that I put back `solve`, the time discretization (u.dt2) would break everything (unconditionally unstable for order 4 FD in time).

examples/seismic/acoustic/operators.py
```diff
@@ -1,9 +1,62 @@
...
+    :param time_order: time order
+    :param m: square slowness
+    :param s: symbol of for the time-step
+    :return: H
+    """
+    if time_order == 2:
+        biharmonic = 0
+    elif time_order == 4:
+        biharmonic = field.laplace2(1/m)
+    else:
```
FabioLuporini: Perhaps we should capture such an error much earlier

mloubout: Not sure, one may to to create new operators with higher order, this is here to ensure that if you use ours, these two orders are the only accepted ones.

examples/seismic/acoustic/operators.py
```diff
@@ -1,9 +1,62 @@
...
+
+
+def iso_stencil(field, time_order, m, s, damp, **kwargs):
+    """
+    Stencil for the acoustic isotropic wave-equation:
+    u.dt2 - H + damp*u.dt = 0
+    :param field: Symbolic TimeData object, solution to be computed
+    :param time_order: time order
+    :param m: square slowness
+    :param s: symbol of for the time-step
```
FabioLuporini: of for

examples/seismic/acoustic/operators.py
```diff
@@ -1,9 +1,62 @@
...
 from examples.seismic import PointSource, Receiver
 
 
+def laplacian(field, time_order, m, s):
+    """
+    Spacial discretization for the isotropic acoustic wave equation. For a 4th
+    order in time formulation, the 4th order time derivative is replaced by a
+    double laplacian:
+    H = (laplacian + s**2/12 laplacian(1/m*laplacian))
+    :param field:  Symbolic TimeData object, solution to be computed
```
FabioLuporini: two consecutive empty spaces

examples/seismic/acoustic/operators.py
```diff
@@ -1,9 +1,62 @@
...
+    :param time_order: time order
+    :param m: square slowness
+    :param s: symbol of for the time-step
```
FabioLuporini: of for

examples/seismic/acoustic/wavesolver.py
```diff
@@ -16,7 +14,11 @@ class AcousticWaveSolver(object):
     :param model: Physical model with domain parameters
     :param source: Sparse point symbol providing the injected wave
     :param receiver: Sparse point symbol describing an array of receivers
-    :param time_order: Order of the time-stepping scheme (default: 2)
+    :param time_order: Order of the time-stepping scheme (default: 2, choices: 2,4)
+                       time_order=4 will not implement a 4th order FD discretization
+                       of the time-derivative as it is unstable. It implement instead
```
FabioLuporini: implements

mloubout:
>I'm not sure I understanding why defaulting time_order=2 all over the place

Because this is the math

http://www.hl107.math.msstate.edu/pdfs/rein/HighANM_final.pdf

FabioLuporini:
> Because this is the math

I haven't checked, but is it clear from the code and/or docstrings that this example/operators are trying to reproduce the method in that paper? ie do you explicitly mention the paper?

examples/seismic/acoustic/wavesolver.py
```diff
@@ -98,7 +94,7 @@ def forward(self, src=None, rec=None, u=None, m=None, save=False, **kwargs):
         if u is None:
             u = TimeData(name='u', shape=self.model.shape_domain, save=save,
                          time_dim=self.source.nt if save else None,
-                         time_order=self.time_order,
+                         time_order=2,
```
FabioLuporini: squash some lines here

examples/seismic/acoustic/wavesolver.py
```diff
@@ -16,7 +14,12 @@ class AcousticWaveSolver(object):
     :param model: Physical model with domain parameters
     :param source: Sparse point symbol providing the injected wave
     :param receiver: Sparse point symbol describing an array of receivers
-    :param time_order: Order of the time-stepping scheme (default: 2)
+    :param time_order: Order of the time-stepping scheme (default: 2, choices: 2,4)
+                       time_order=4 will not implement a 4th order FD discretization
+                       of the time-derivative as it is unstable. It implements instead
+                       a 4th order accurate wave-equation with only second order
+                       time derivative.
+                       http://www.hl107.math.msstate.edu/pdfs/rein/HighANM_final.pdf
```
FabioLuporini: Please say explicitly what this links represents

examples/seismic/acoustic/wavesolver.py
```diff
@@ -135,7 +128,7 @@ def adjoint(self, rec, srca=None, v=None, m=None, **kwargs):
         # Create the adjoint wavefield if not provided
         if v is None:
             v = TimeData(name='v', shape=self.model.shape_domain,
-                         save=False, time_order=self.time_order,
+                         save=False, time_order=2,
```
FabioLuporini: I'm leaving the comment here, but this applies probably to other places as well. If you can, try to minimize the number of lines to instantiate a ``TimeData``. Sounds like stupid, but it makes the code incredibly verbose

mloubout:
Should be green, TRAVIS/conda problem with python 2.7

FabioLuporini:
@mlange05 , while Travis attempts again to get green, do you have any further enquiries ?

mlange05:
Generally I'm ok with this, and I think it's good to go.

One comment, that should not affect this PR, is that instead of unpicking the components of the individual operators are now quite separate and have the 2nd/4th order diff in various places. The alternative would be to create bespoke operators for each case, ie. `ForwardOperatorTO2`/`ForwardOperatorTO4` - this would result in some code replication but would make the documentation of each operator more explicit, as all of the  pieces for each operator are in a single documented place. This one comes largely down to taste, and while I was originally aiming for the replicated case, I'm fine with this approach too.

FabioLuporini:
Unfortunately Travis/Conda are still angry...

mlange05:
Ok, all happy, merging....

PR #361, opened 2017-09-27
ConstantData new_cls fix

mloubout:
Tiny fix for `ConstantData` that was skipping `eval=false` in the `new options`.
Was making sympy solve fail.

### DIFF
--- a/devito/interfaces.py
+++ b/devito/interfaces.py
@@ -179,7 +179,8 @@ class AbstractSymbol(Function, CachedSymbol):
 
     def __new__(cls, *args, **kwargs):
         if cls in _SymbolCache:
-            newobj = Function.__new__(cls, *args)
+            options = kwargs.get('options', {})
+            newobj = Function.__new__(cls, *args, **options)
             newobj._cached_init()
         else:
             name = kwargs.get('name')


FabioLuporini:
OK, let's merge this

FabioLuporini:
Oh, wait: are you not in this way dropping all other ``kwargs`` on the floor?

mloubout:
Not droping anything, only passing extra options when there is some

FabioLuporini:
OK, what I meant was: are you not dropping the other values in ``kwargs`` ? 
Look at how it's done here: https://github.com/opesci/devito/pull/361/files#diff-9609ccc7839f217f15b36193cbc5425fR282

mloubout:
>OK, what I meant was: are you not dropping the other values in kwargs

No I'm not. As you see in the line you point out `kwargs` contains the normal ones + the `options` dictionary required to call `Function.__new__` like here:
https://github.com/opesci/devito/pull/361/files#diff-9609ccc7839f217f15b36193cbc5425fR193

this is done correctly is not cached but was missing if cached.

FabioLuporini:
Merged.

PR #365, opened 2017-09-29
Sympy-1.1 update

mlange05:
This merge adds the necessary compatibility changes to make Devito work with the new `sympy-1.1` and updates the dependencies everywhere.

**Note: This is not backwards compatible with `sympy-1.0` and all environments have to be upgraded!**

### DIFF
--- a/devito/dse/extended_sympy.py
+++ b/devito/dse/extended_sympy.py
@@ -41,7 +41,16 @@ class FrozenExpr(Expr):
 
 
 class Eq(sympy.Eq, FrozenExpr):
-    pass
+
+    """
+    A customized version of :class:`sympy.Eq` which suppresses
+    evaluation.
+    """
+
+    def __new__(cls, *args, **kwargs):
+        kwargs['evaluate'] = False
+        obj = sympy.Eq.__new__(cls, *args, **kwargs)
+        return obj
 
 
 class Mul(sympy.Mul, FrozenExpr):
--- a/devito/__init__.py
+++ b/devito/__init__.py
@@ -12,6 +12,7 @@ from devito.logger import error, warning, info  # noqa
 from devito.parameters import (configuration, init_configuration,  # noqa
                                env_vars_mapper)
 from devito.tools import *  # noqa
+from devito.dse import *  # noqa
 
 
 def clear_cache():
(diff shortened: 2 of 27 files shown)


devito/dse/extended_sympy.py
```diff
@@ -41,7 +41,16 @@ def xreplace(self, rule):
...
-    pass
+
+    """
+    A customized version of :class:`sympy.Eq` which suppresses
+    evaluation.
+    """
+
+    def __new__(cls, *args, **kwargs):
+        kwargs['evaluate'] = False
+        obj = sympy.Eq.__new__(cls, *args, **kwargs)
```
FabioLuporini: Just of out curiosity, where did it break w/o this ?

mloubout:
Should `Eq` be imported from devito instead of `sympy` too in:
`examples/seismic/acoustic/operators.py`
`examples/seismic/tti/operators.py`
`devito/pointdata.py`

And any other one I may have missed?

devito/dse/extended_sympy.py
mlange05: The reason is subtle and only applies for a few corner-cases in `test_operator.py`; effectively degenerate expressions like `Eq(a, a + 3)` will now immediately evaluate to `False`, so we need to suppress evaluation.

mlange05:
@mloubout Good call, I have updated as many as I could find. It technically doesn't matter, as only some degenerate expressions are affected (see comment above), but it's good for consistency.

devito/dse/extended_sympy.py
mloubout: Add `Eq` to `_all_` and should be good

FabioLuporini:
Oh, perhaps time to replace ``q_indexed`` (in ``dse/queries``) with ``.is_Indexed`` all over the codebase ?

FabioLuporini:
I think this should be the next one to go

mlange05:
@FabioLuporini Agreed. I just rebased and tried to get rid of most of the `q_indexed` usages, although we still need it for the search engine. Should be good to go once travis is happy.

FabioLuporini:
Thanks. Let's merge this as soon as Travis is happy

FabioLuporini:
OK, merging...

FabioLuporini:
Merged.

PR #364, opened 2017-09-28
Misc YASK updates

FabioLuporini:
These make the tests in ``test_yask.py`` pass when running on top of ~YASK:v2_alpha:9b87b4b21~ YASK:v2_alpha:ac9f784bf7d , which changed quite a few things w.r.t. to the ``develop`` branch

It also ships the required tweaks after the sympy1.1 switch

PR #366, opened 2017-10-02
Travis setup for YASK testing

mlange05:
Depends on #364. 

This merge adds a dedicated YASK builder to our travis setup that pulls the current YASK development branch and executes a dedicated small test suite with it while skipping all other tests. This will allow us to gradually enable more of our regular test base on the YASK-backend as it develops.

One thing I note is that despite `-j3` the YASK compile times are quite severe, making the test suite quite slow;  but that might need to be addressed upstream.


FabioLuporini:
Yes, good to go

mlange05:
Ok, YASK is ready, merging...

PR #368, opened 2017-10-02
dse: Fix aliases detection

FabioLuporini:
Hopefully uncontroversial

### DIFF
--- a/devito/dse/aliases.py
+++ b/devito/dse/aliases.py
@@ -44,7 +44,7 @@ def collect(exprs):
     # - that are non-scalar
     candidates = OrderedDict()
     for expr in exprs:
-        if q_indexed(expr):
+        if q_indexed(expr.lhs):
             continue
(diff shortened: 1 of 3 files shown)


FabioLuporini:
@mlange05 this fixes the issue we bumped into a few days ago when I asked you to check whether things were working properly in TTI with dse aggressive mode at higher SO order, to eventually find out there was a bug in ``master``, rather than your branch, with SO=16

FabioLuporini:
In fact, I decided this wasn't enough and pushed a DSE test to monitor the operation counts in TTI. This test will probably increase the TRAVIS running time by about 2 mins (only 80secs on my workstation), but I think it's far too important and should really go in. Open to debate, though... let me know what you think @mlange05

mlange05:
Yes, looks alright to me; slightly increased travis times, but still doable it seems

FabioLuporini:
OK, then merging...

FabioLuporini:
Merged.

PR #369, opened 2017-10-02
Introducing a new Grid abstraction

mlange05:
Requires PR #365.

This PR fundamentally changes the way we create discretized function objects (`Dense/TimeData`) by introducing a new `Grid` container class that provides, stores and manages discretization meta-data. With the new API users can either provide a full description of the discretisation (ie. dimensions and shape) or the define a grid which defines spatial dimensions and a "domain shape" from which all field/function objects can be derived.

This PR also introduces the concept of a `SpaceDimension`, which is simply a `Dimension` object that defines the spatial shape of the domain. We now create all spatial derivative short-cuts (eg. `dx, dx2, dxy, etc`) on-the-fly and for each `SpaceDimension`, which means **users can finally customise space itself!** :open_mouth:


mlange05:
Please note, that this is the first of a series of re-structuring PRs that will strongly affect the user API. There is lots more work to be done of top of this, eg. renaming our primary user-facing data symbols or flipping the concpet of "domain" and how to deal with stencil boundaries. These will all be addressed in follow-up PRs to keep things digestible.

FabioLuporini:
Looking good; the diff is huge, so I'll do a proper review once #365 is in !

FabioLuporini:
oh btw, if we have this ``SpaceDimension`` type, should we also rename ``BufferedDimension`` --> ``TimeDimension`` ?

devito/grid.py
```diff
@@ -0,0 +1,50 @@
...
+    to discretize :class:`Function`s.
+
+    :param shape: Shape of the domain region in grid points.
+    :param extent: Physical extent of the domain in m; defaults to a
+                   unit box of extent 1m in all dimensions.
+    :param origin: Physical coordinate of the origin of the domain;
+                   defaults to 0. in all dimensions.
+    :param dtype: Default data type to be inherited by all Functions
+                  created from this :class:`Grid`.
+    """
```
mloubout: Convention is usually (n,d,o)
n : number of grid points
d : spacing
o : origin

the get extent from this three. The current way will give wrong spacing if `origin` is non-zero.

mloubout:
I'll have a deeper look at it with available time (SINBAD consortium meeting this week).

DOes this mean we ca nget rid of `FixedDimension`...? 😬

devito/finite_difference.py
```diff
@@ -190,3 +191,34 @@ def first_derivative(*args, **kwargs):
...
+    """
+    Create generic arbitrary order derivative expression from a
+    single :class:`Function` object. This methods is essentially a
+    dedicated wrapper around SymPy's `as_finite_diff` utility for
+    :class:`devito.Function` objects.
+
+    :param function: The symbol representing a function.
+    :param order: Derivative order, eg. 2 for a second derivative.
+    :param dim: The dimension for which to take the derivative.
+    :param width: Order of the coefficient discretization and thus
```
mloubout: Nitpicking,

 I know it says what it does, but `width` is an odd name, would call it fd_order or something like that

devito/grid.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @property
+    def dim(self):
+        """Problem dimension, or number of spatial dimensions."""
+        return len(self.shape)
+
+    @property
+    def spacing(self):
+        """Spacing between grid points in m."""
+        return as_tuple(np.array(self.extent) / np.array(self.shape))
```
mloubout: `(np.array(self.extent) - np.array(self.origin)) / (np.array(self.shape) -1)`
if kept with origin/extend/shape

devito/interfaces.py
```diff
@@ -440,44 +445,126 @@ class DenseData(TensorData):
...
-                self.shape = tuple([d.size for d in dimensions])
+            self.grid = kwargs.get('grid', None)
+
+            if self.grid is None:
+                self.shape_domain = kwargs.get('shape', None)
+                self.dtype = kwargs.get('dtype', np.float32)
+                if self.shape_domain is None:
+                    error("Creating a Function requires either 'shape'"
+                          "or a 'grid' argument")
+                    raise ValueError("Unknown symbol dimensions or shape")
```
mloubout: `self.grid = Grid(kwargs.get('shape', None))` with raise exception?

May lead to issues if using different dimensions for different objects.

devito/interfaces.py
```diff
@@ -440,44 +445,126 @@ class DenseData(TensorData):
...
+
+                # First derivative, right
+                dxr = partial(first_derivative, order=self.space_order,
+                              dim=dim, side=right)
+                setattr(self.__class__, 'd%sr' % dim.name,
+                        property(dxr, 'Symbol for the first derivative '
+                                 'wrt %s with a right stencil' % dim.name))
+
+                # Second derivative
+                dx2 = partial(generic_derivative, order=2, dim=dim,
```
mloubout: `second_derivative`?

devito/interfaces.py
```diff
@@ -440,44 +445,126 @@ class DenseData(TensorData):
...
+                                 'wrt the %s dimension' % dim.name))
+
+                # Fourth derivative
+                dx4 = partial(generic_derivative, order=4, dim=dim,
+                              width=max(int(self.space_order / 2), 2))
+                setattr(self.__class__, 'd%s4' % dim.name,
+                        property(dx4, 'Symbol for the fourth derivative '
+                                 'wrt the %s dimension' % dim.name))
+
+                for dim2 in self.space_dimensions:
```
mloubout: This will create duplicates properties (e.g `dxy`/`dyx`), not really a problem just unnecessary

devito/interfaces.py
```diff
@@ -440,44 +445,126 @@ class DenseData(TensorData):
...
-            shape = kwargs.get('shape')
-            if len(shape) <= 3:
-                dimensions = _indices[:len(shape)]
-            else:
-                dimensions = [Dimension("x%d" % i) for i in range(1, len(shape) + 1)]
+        else:
+            if dimensions is not None:
+                warning("Creating Function with 'grid' and 'dimensions' "
+                        "argument; ignoring the 'dimensions' and using 'grid'.")
+            dimensions = grid.dimensions
```
mloubout: should be only line needed if `grid` created in `init` if missing.

examples/seismic/model.py
```diff
@@ -303,37 +303,33 @@ class Model(object):
...
-        self.spacing = spacing
         self.shape = shape
         self.nbpml = int(nbpml)
-        self.dtype = dtype
 
-        # Ensure same dimensions on all inpute parameters
-        assert(len(origin) == len(spacing))
-        assert(len(origin) == len(shape))
+        shape_pml = np.array(shape) + 2 * self.nbpml
+        extent = tuple(np.array(spacing) * (shape_pml))
```
mloubout: `tuple(origin + np.array(spacing) * (shape_pml -1)`

mlange05:
@FabioLuporini Ok, rebased on master for a cleaner diff.

> oh btw, if we have this SpaceDimension type, should we also rename BufferedDimension --> TimeDimension ?

Yes and no. We should definitely introduce a dedicated `SpaceDimension`, but it will not be what we now have as `BufferedDimension`, as it is the "SteppingDimension" (YASK term) that has the true `time` dimension as a parent. Having `time` as a `TimeDimension` class will allow us generically add the time derivative shortcuts, and may also help in doing derived dimensions to facilitate time sub-sampling of wavefields.

mlange05:
> DOes this mean we ca nget rid of FixedDimension...?

@mloubout No, but it does mean that if we were to introduce something like a `FixedSpaceDimension` type we can now specify this on the grid **once(!)**, and then have all functions magically inherit the fixed dimensions. This effectively circumvents the limitations of the new API due to the fact that our default dimensions are currently shared globally. Hopefully we can remove all global instantiations of dimensions in a follow-up PR.

devito/grid.py
mlange05: Yeah, so I don't think the current convention is particularly intuitive, tbh. My primary concern is that the `Grid` primarily defines a physical domain, given by the origin (bottom left in 2D) and the extent (or "width") of the domain, so that the top right corner is defined as `origin + extent`. I'll make this clearer in the docstring.

The other thing I would really like is that the `spacing` is a diagnostic property determined by the spatial extent of the domain and the number of either cells or nodes (currently nodes) that define the grid shape. This effectively makes it much easier to create equivalent physical grids with different discretisations without the need for the user to carefully check that `spacing * shape` always comes out to the same exact `extent`. After all, a unit-cube always has `extent=(1., 1., 1.)`, no matter what `shape` I discretise it with - and I believe this mirrors general PDE-conventions more closely.

devito/finite_difference.py
mlange05: Agreed, will change.

devito/grid.py
mlange05: As mentioned above, I think we have different interpretations of `extent`. I see it purely as the "width" of the domain, not the top-right-back coordinate.

devito/interfaces.py
mlange05: If you're using different dimensions for two objects you should either create tow-different grids and make sure they agree on all dimensions you want them to agree on, or explicitly provide `dimensions` and `shape` when creating your function objects.

Also, not providing a `shape` to a `Grid`, which ultimately defines the base of the spatial data discretisation within the physical domain, simply doesn't make sense to me.

devito/interfaces.py
mlange05: Both are equivalent, although I probably should rename `order` to `derivative_order`, or something clearer.

devito/interfaces.py
mlange05: Hmmm, good point. It probably will, but they should be symbolically equivalent, so I don't think we need to worry about it too much.

examples/seismic/model.py
mlange05: Again, different interpretations of `extent`.

mlange05: Uhhh, and I tried the `- 1`, but things started to break. So we might need to go through the model code carefully to make sure that everything treats nodes and cells accordingly.

devito/interfaces.py
mlange05: Ahhh, now I get you. However, I don't think that is a good idea, since, for example, a point source has dimensions `(d, npoints)`, which means I would now have to auto-create a grid for that dimension pair. As a grid is actually meant to represent a physical space, that makes very little sense though, which is why we need to keep the ability to explicitly specify a function with a `(shape, dimensions)` pair.

devito/grid.py
```diff
@@ -0,0 +1,50 @@
...
+__all__ = ['Grid']
+
+
+class Grid(object):
+
+    """
+    A cartesian grid that encapsulates a physical domain over which
+    to discretize :class:`Function`s.
+
+    :param shape: Shape of the domain region in grid points.
```
FabioLuporini: Is it worth start mentioning the distinction between domain and halo? or is it too premature?

devito/grid.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def __init__(self, shape, extent=None, origin=None, dimensions=None,
+                 dtype=np.float32):
+        self.shape = as_tuple(shape)
+        self.extent = as_tuple(extent or tuple(1. for _ in shape))
+        self.origin = as_tuple(origin or tuple(0. for _ in shape))
+        self.dtype = dtype
+
+        # TODO: Raise proper exceptions and logging
+        assert(self.dim == len(self.origin) == len(self.extent) == len(self.spacing))
```
FabioLuporini: I'm not sure how this couldn't be the case? as in, given how the various fields are computed

devito/interfaces.py
```diff
@@ -422,11 +424,14 @@ class DenseData(TensorData):
     """Data object for spatially varying data acting as a :class:`SymbolicData`.
 
     :param name: Name of the symbol
-    :param dtype: Data type of the scalar
-    :param shape: The shape of the tensor
+    :param shape: Domain shape of the associated data for this :class:`Function`.
+                  Note that this does not include the boundary padding added due
+                  the stencil radius for space dimensions.
```
FabioLuporini: Yeah, previously in the other comment I meant something like this

devito/interfaces.py
```diff
@@ -440,44 +445,126 @@ class DenseData(TensorData):
...
             self.name = kwargs.get('name')
-            self.shape = kwargs.get('shape', None)
-            if self.shape is None:
-                dimensions = kwargs.get('dimensions')
-                self.shape = tuple([d.size for d in dimensions])
+            self.grid = kwargs.get('grid', None)
+
+            if self.grid is None:
+                self.shape_domain = kwargs.get('shape', None)
+                self.dtype = kwargs.get('dtype', np.float32)
```
FabioLuporini: nitpicking: move after the ``if``

devito/finite_difference.py
```diff
@@ -190,3 +191,34 @@ def first_derivative(*args, **kwargs):
...
+def second_cross_derivative(function, dims, order):
+    """
+    Create a second order order cross derivative for a given function.
+
+    :param function: The symbol representing a function.
+    :param dims: Dimensions for which to take the derivative.
+    :param order: Discretisation order of the stencil to create.
+    """
+    first = second_derivative(function, dim=dims[0], width=order)
+    return second_derivative(first, dim=dims[1], order=order)
```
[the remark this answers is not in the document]
mlange05: Are you sure? This is only used for `u.dx2y2` and mirrors the old convention; it;s not replacing the original `cross_derivative` function.

devito/interfaces.py
```diff
@@ -440,44 +445,126 @@ class DenseData(TensorData):
...
+                setattr(self.__class__, 'd%s' % dim.name,
+                        property(dx, 'Symbol for the first derivative '
+                                 'wrt the %s dimension' % dim.name))
+
+                # First derivative, left
+                dxl = partial(first_derivative, order=self.space_order,
+                              dim=dim, side=left)
+                setattr(self.__class__, 'd%sl' % dim.name,
+                        property(dxl, 'Symbol for the first derivative '
+                                 'wrt %s with a left stencil' % dim.name))
```
FabioLuporini: for homogeneity :
```
Symbol for the first derivative wrt the %s dimension with a left stencil
```
Perhaps more correct
```
Return the symbolic first-order left-sided approximation wrt the %s dimension
```
Note that in the case of a centered approximation the approximation is second order

FabioLuporini: and, probably, also the other automatically generated docstrings below could be tweaked a bit

devito/interfaces.py
```diff
@@ -440,44 +445,126 @@ class DenseData(TensorData):
...
         """
+
+        grid = kwargs.get('grid', None)
         dimensions = kwargs.get('dimensions', None)
-        if dimensions is None:
-            # Infer dimensions from default and data shape
-            if 'shape' not in kwargs:
-                error("Creating symbolic data objects requries either"
-                      "a 'shape' or 'dimensions' argument")
+        if grid is None:
```
FabioLuporini: nitpicking: move ``dimensions = kwargs.get(...)`` inside this first ``if``

devito/interfaces.py
```diff
@@ -440,44 +445,126 @@ class DenseData(TensorData):
...
+                warning("Creating Function with 'grid' and 'dimensions' "
+                        "argument; ignoring the 'dimensions' and using 'grid'.")
+            dimensions = grid.dimensions
         return dimensions
 
+    @property
+    def shape_data(self):
+        """
+        Full allocated shape of the data associated with this
+        :class:`Function`.
```
FabioLuporini: single line?

devito/interfaces.py
```diff
@@ -501,145 +588,19 @@ def initialize(self):
...
-    def dz4(self):
-        """Symbol for the fourth derivative wrt the z dimension"""
-        width_h = max(int(self.space_order / 2), 2)
-        indz = [(z + i * z.spacing) for i in range(-width_h, width_h + 1)]
-
-        return self.diff(z, z, z, z).as_finite_difference(indz)
-
-    @property
     def laplace(self):
         """Symbol for the second derivative wrt all spatial dimensions"""
```
FabioLuporini: not part of this PR, I know, but since we're fixing this stuff... the docstring, again: this returns something more than "just a Symbol"

devito/interfaces.py
```diff
@@ -650,6 +611,7 @@ class TimeData(DenseData):
     :param shape: Shape of the spatial data grid
     :param dimensions: The symbolic dimensions of the function in addition
                        to time.
+    :param shape: Domain shape of the associated data for this :class:`Function`.
```
FabioLuporini: docstring different than that used in ``DenseData``

devito/interfaces.py
```diff
@@ -684,24 +646,32 @@ class TimeData(DenseData):
...
                     error('Time dimension (time_dim) is required'
                           'to save intermediate data with save=True')
                     raise ValueError("Unknown time dimensions")
-            self.shape = (time_dim,) + self.shape
+
+    @property
+    def shape_data(self):
+        """
+        Full allocated shape of the data associated with this
+        :class:`TimeFunction`.
```
FabioLuporini: single line?

devito/grid.py
mlange05: I actually tried that, but flipping the automated treatment of the stencil boundary is tricky, so I left if for another PR. So the documentation of these concepts (stencil boundary, halo, etc.) should come with the actual features, I think.

devito/grid.py
mlange05: All of them could be user-provided, and I'm just asserting against user-level idiocy (pebkac).

devito/interfaces.py
mlange05: Actually, I should remove this one and add it back in once that statement is actually true (not in this PR).

devito/interfaces.py
mlange05: Not according to emacs, but sure, why not. :wink:

devito/interfaces.py
mlange05: Agreed, it's actually a symbolic expression that's returned.

devito/interfaces.py
mlange05: Actually no, since I want the `dtype` to be able to override the `grid.dtype` (one might want to have different dtypes per function), but the defaulting value changes from `int32` to `grid.dtype`.

devito/interfaces.py
mlange05: No, it's used in the else branch too to warn about implicit overrides.

mlange05:
Ok, I _believe_ I have addressed all comments, so could we re-review this, please? @mloubout @FabioLuporini

devito/finite_difference.py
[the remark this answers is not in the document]
mloubout: Oh ok then

devito/grid.py
mloubout: The I would call it `width`, extent sounds like "how far" and needs to know the origin to know the actual size, the "-1" is till required

examples/seismic/model.py
mloubout: Even with the definition of extent, domain starts at `origin` and finishes at `origin+extent` in n points, so that's (n-1) intervals.

devito/grid.py
mlange05: Yeah no, `extent` includes, for example in 3D, the height of the grid, and generally speaking `width != height`. The point about extent is that it is, as far as I interpret the word usually, a size relative to an origin - and since our domain has an "origin", the extent (relative size) of the domain in combination with the absolute coordinates of the origin defines the entire physical space that our domain covers. 

In fact, in your interpretation "extent" seems to be an absolute coordinate, and I'm not entirely sure how you justify that?

examples/seismic/model.py
mlange05: Ahhh yes, you are right. I know what I did wrong previously. Will fix.

devito/grid.py
mloubout: This are more practical considerations. In practice two things happen:

1- People choose a `space_order` according to some empirical/theoretical results on dispersion and from it, choose the optimal spacing, so that is the parameter they will know/want to input

2- The actual computational domain is defined by physical constraints such as in seismic case, what is the maximum/minimum src/rec position, so the "extent" the way you define it is not the known, origin/end is what is known.

Which is why the usual origin/spacing + either end/npoints is the convention.

devito/grid.py
```diff
@@ -0,0 +1,77 @@
...
+            |     |            |
+            |     |   DOMAIN   | extent[1]
+            |     |            |
+            |     |            |
+            |     |  extent[0] |
+            |     x------------x
+            |  origin
+            |
+            |----------------------->
+                       y
```
mloubout: nitpicking
convention is x horizontal, y vertical (matters for angles orientation for tti for example).

mlange05: Ooops, completely true. Will fix.

devito/grid.py
mlange05: Ah ok, now this is starting to make sense. The problem though is that this is a somewhat specific convention in seismic, I believe. As far as I'm aware, more generic physics modelling (eg. CFD) starts from some description of the physical domain, which might have geometric features like inlets/outlets on a channel, and discretizes this to arrive at the grid spacing. 

Since we want to keep Devito a generic FD engine, I would therefore argue we keep this convention on the `Grid`, while the `Model`, which is clearly specific to seismic cases, implements the convention you described. After all it is quite trivial to switch between the two.

mloubout: >As far as I'm aware, more generic physics modelling (eg. CFD) starts from some description of the physical domain, which might have geometric features like inlets/outlets on a channel, and discretizes this to arrive at the grid spacing.

The grid spacing would still be constrained by the discretization order, but fine with me I guess

mloubout: still needs `(np.array(self.shape) -1)` to get the correct spacing.

mlange05: True, just pushed.

devito/grid.py
```diff
@@ -69,7 +69,7 @@ def dim(self):
     @property
     def spacing(self):
         """Spacing between grid points in m."""
-        return as_tuple(np.array(self.extent) / np.array(self.shape))
+        return as_tuple(np.array(self.extent) / np.array(self.shape) - 1)
```
mloubout: Parenthesis at wrong place : 
 `as_tuple(np.array(self.extent) / (np.array(self.shape) - 1))`

mlange05: Argh, totally true.... Thanks!

mlange05:
Ok, everyone happy, so merging....

PR #371, opened 2017-10-05
Switching to and fixing Python3.6

mlange05:
As it turns out our `python3` travis builder has been silently reverting to `python2.7` for some time, meaning a few small bugs have crept in. Since the old trick of separating the two environments doesn't seem to work any more, this PR will switch to Python-3.6 for testing exclusively. The Python version is now also hard-defined in the `environment.yml`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,21 +14,21 @@ env:
 matrix:
   include:
     - os: linux
-      python: "2.7"
+      python: "3.6"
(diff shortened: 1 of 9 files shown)


mlange05:
Ok, YASK issues resolved on travis. Can I please get some confirmation that everyone is happy with this change? I would like to merge this next.

mlange05:
Ok, discussed offline with @FabioLuporini and @ggorman. So I'm going ahead and merge this now...

Issue #372, opened 2017-10-05
FixedDimension has to go


mloubout:
Full discussion on slack

Code-generation API can not create interface object that don't make sense for cleanup purposes.

mlange05:
Ok, first of all, polemic titles like this are not helping anyone. `FixedDimension` objects will most definitely stay, _if_ (and only if) the concept of fixed dimensions even survives at all.

Now for the technical explanation:
a) The introduction of a specialised `FixedDimension` type **is not merely "for cleanup purposes"!** The concept of fixed dimensions is naturally incompatible with certain aspects of more advanced features that we are working on, eg. checkpointing or MPI support. As such, having regular dimensions and fixed ones present as the same object type throughout the code generation engine is making developing these features extremely difficult. Hence separating the types makes a lot of sense, and paves the way for these new features.

b) The real problem with fixed dimensions is that the default dimension used by function symbols are currently managed globally, and thus runtime injection of specialised dimensions is tricky at best. This is the reason that the fixed-dimension test in `test_adjointA.py` was recently disabled. A more comprehensive solution to this problem is to not use global dimension symbols as defaults, but build them on a per-case basis. This can be done via the newly introduced `Grid` classes and I'm working on a new API to facilitate just that. 

c) **Important:** The concept of a "fixed dimension", ie. hard-coding loop bounds, is **a purely low-level compiler feature** and not part of any formal mathematical abstraction Devito provides. It is largely kept around for legacy purposes and was introduced as an implementation shortcut in Devito-2.0 (Propagator). As such, **no high-level domain-specific application should depend on this feature!** After all, Devito operators are supposed to be black boxes where the engine decides how best to solve the given mathematical definition in the fastest way, leaving low-level implementation details, such as loop bounds, to the compiler engine. **This is the essence of software abstraction!**

mloubout:
>FixedDimension objects will most definitely stay, if (and only if) the concept of fixed dimensions even survives at all.

Since when?

>features that we are working on, eg. checkpointing or MPI support. As such, having regular dimensions and fixed ones present as the same object type throughout the code generation engine is making developing these features extremely difficult

After spending three weeks telling me that the features we are working on is not an argument for not having it, the argument is because your features need it?

>The concept of a "fixed dimension", ie. hard-coding loop bounds, is a purely low-level compiler feature and not part of any formal mathematical abstraction

[...]

PR #373, opened 2017-10-06
 Die, FixedDimension, die, die, die!

mlange05:
I hope the title is self-explanatory. Should address issue #372.


FabioLuporini:
Yes, there's no reason to have them and it's now just a maintenance burden  .

tests/test_dimension.py
```diff
@@ -1,21 +0,0 @@
...
-    devito.x = FixedDimension(name='x', size=10, spacing=devito.x.spacing)
-    assert(devito.x in m.indices)
-
-
-@skipif_yask
-def test_correct_usage():
-    myx = FixedDimension(name='x', size=10, spacing=devito.x.spacing)
-    grid = Grid(shape=(10, 10), dimensions=(myx, y))
-    m = Function(name="m", grid=grid)
-    assert(myx in m.indices)
```
mloubout: Maybe not for this PR, but should we still have at least one test (no idea what) for Dimensions?

mlange05: We have a whole bunch of tests that use explicitly specified and grid-derived dimensions all over the test base - this test was really just specific to the fixed dimension API and it's recent change. I do agree though that we are lacking tests for some more exotic dimension combinations, in particular with respect to mixing space/time and auxiliary dimensions - but yes that is definitely stuff for future PRs.

mloubout: > do agree though that we are lacking tests for some more exotic dimension combinations,

We made some with @fablup previously as part of some dse fixes. I think there is some left in #340 that include mixed dimensions laplacian. I can dig it up, and can be part of a new PR

mlange05: Excellent. I'm still working on adding dedicated `TimeDimension`  types, so these will come in handy.

mloubout: #340 is actually quite small, the `test_operator` with `b.laplace` would the go to one for this PR to make sure we catch correct dimensions for auto-shortcuts

FabioLuporini:
> I do agree though that we are lacking tests for some more exotic dimension combinations, in particular with respect to mixing space/time and auxiliary dimensions

There are a few tests in ``test_operator.TestScheduler`` checking that code generation is appropriate with different types of dimensions, different order, etc. I agree this is far from being a complete coverage, though. And they're also not tailored to the new Dimension types...

mlange05:
Ok, I take it everyone is happy with this then? @navjotk?

navjotk:
Wait for 345 to go in first.

mlange05:
Ok, with final approval this is good to go in, I think. Merging once travis is happy...

mlange05:
Ok, merging...

PR #375, opened 2017-10-08
Switch to `nbval` for notebook testing on travis

mlange05:
This solution is much more powerful than the previous `pytest-ipynb`, allowing is to use a normal naming convention and deliberately skip individual cells in the notebooks. It also updates the metadata on all notebooks to Python3.


mlange05:
Yes, and it is essentially required for figures, as they are not yet diffable (for a discussion see https://github.com/computationalmodelling/nbval/issues/6). Also, `#NBVAL_SKIP`, is another very useful keyword that allows us to skip computation that is too heavy for travis, eg. a fill FWI run.

mlange05:
Ok, merging once travis is happy...

mlange05:
Travis green, merging....

Issue #328, opened 2017-08-11
numerics of acoustic with constant velocity 


FabioLuporini:
We get segfaults when denormal numbers are dropped by the DLE. This seems to be a bug of GCC; things works like a charm with ICC. Hence, we're currently using the ``dle=advanced-safemath`` mode, rather than ``dle=advanced``. This needs to be investigated further.

mloubout:
I played wit it and I am not able to reproduce it even with GCC, Will keep digging.

mloubout:
Looks like fixed, no issue with #377, closing

PR #377, opened 2017-10-10
Acoustic examples merged

mloubout:
Small PR  that merged constant example as an argument to avoid useless duplicate.

Also avoids segfaults that were coming from `nbpml=0` in `constant_example` as the rec interpolation was trying to access a point outside the domain.

@FabioLuporini , it runs fine with GCC(4.9) and any `dse` so not sure what to say about #328


.travis.yml
```diff
@@ -83,7 +83,7 @@ script:
       python examples/seismic/benchmark.py test -P tti -so 4 -a -d 20 20 20 -n 5;
       python examples/seismic/benchmark.py test -P acoustic -a;
       python examples/seismic/acoustic/acoustic_example.py --full;
-      python examples/seismic/acoustic/constant_example.py --full;
+      python examples/seismic/acoustic/acoustic_example.py -c;
```
FabioLuporini: should we keep both ``--full`` and ``-c`` ?

mloubout: Yes, two different things, `full` is to run all operators (born, adjoint, gradient)

examples/seismic/acoustic/acoustic_example.py
```diff
@@ -52,16 +56,24 @@ def acoustic_setup(shape=(50, 50, 50), spacing=(15.0, 15.0, 15.0),
...
     dm = np.float32(initial_vp**2 - solver.model.m.data)
     info("Applying Forward")
     rec, u, summary = solver.forward(save=full_run, autotune=autotune)
 
+    if constant:
+        # With  a new m as Constant
+        m0 = Constant(name="m", value=.25, dtype=np.float32)
+        solver.forward(save=full_run, m=m0)
+        # With a new m as a scalar value
+        solver.forward(save=full_run, m=.25)
```
mlange05: Should the non-constant code path still be executed as well, or should this go into an `else` clause here?

examples/seismic/acoustic/acoustic_example.py
```diff
@@ -95,6 +107,8 @@ def run(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
     parser.add_argument("-dle", default="advanced",
                         choices=["noop", "advanced", "speculative"],
                         help="Devito loop engine (DSE) mode")
+    parser.add_argument("--c", default=False, dest='constant', action='store_true',
```
mlange05: Can we make the argument signature `'-c', '--constant'` for consistency? You won't need to specify `dest='constant'` then.

examples/seismic/acoustic/acoustic_example.py
mloubout: >Should the non-constant code path still be executed as well

This is not a non-constant path.
The rest (adjoint, born and gradient) is generated for constant/non-constant according to m, this `if` only adds an extra "test" for the constant case that is done all over the place for the non-constant case making sure we runtime args works.

examples/seismic/acoustic/acoustic_example.py
mloubout: sure

examples/seismic/acoustic/acoustic_example.py
mlange05: Ok

mloubout:
good to go?

mlange05:
Yes, just need travis to catch up and then we can merge.

mloubout:
Didn't see master changed, sorry.

FabioLuporini:
> Didn't see master changed, sorry.

That was me pushing directly to ``master`` a single word change in ``PERFORMANCE.md``

mlange05:
Ok, all good. Merging...

PR #378, opened 2017-10-13
Turn `Constant` into a `Symbol`, not `Function`

mlange05:
First part of an internal data type shake up that should make `Constant` objects more compatible with the various backends. The primary idea is to
* Split our internal caching data types into `Function`s, symbols representing tensor/array data that carry indices, and `Symbol`s, dimension-free, scalar symbols.
* By providing `AbstractSymbol` in addition to `AbstractFunction`, we can now add our internal `Symbol` types (spacing symbols, iteration size variables, etc.) to our caching hierachy, but also better emulate the expected meta-data of functions. This means:
  * We now have to initialise them as `Symbol(name='symbols_name')`, and `s0 = Symbol(name='s'); s1 = Symbol(name='s')` don't cache anymore.
  * We can provide all the meta-data expected of functions, eg. `dtype`, `indices`, etc...
* Turn `Constant` into a `Symbol`, not `Function`. This is notationally nicer and more intuitive, and allows the DSE to effectively treat it as a `Function`.

Note that the final point is the aim of the game here, as this PR prepares things for generic grids, where we don't necessarily have to substitute spacing symbols.

Please also note that we can probably convert `Dimension` types to the new `AbstractSymbol` types, but we might want to leave that for another PR.

### DIFF
--- a/devito/function.py
+++ b/devito/function.py
@@ -10,7 +10,7 @@ from devito.memory import CMemory, first_touch
 from devito.cgen_utils import INT, FLOAT
 from devito.dimension import d, p, t, time, x, y, z
 from devito.arguments import ConstantArgProvider, TensorFunctionArgProvider
-from devito.types import SymbolicFunction, AbstractSymbol
+from devito.types import SymbolicFunction
 from devito.finite_difference import (centered, cross_derivative,
                                       first_derivative, left, right,
                                       second_derivative, generic_derivative,
@@ -44,26 +44,21 @@ Forward = TimeAxis('Forward')
 Backward = TimeAxis('Backward')
 
 
-class Constant(SymbolicFunction, ConstantArgProvider):
+class Constant(sympy.Symbol, ConstantArgProvider):
 
     """
-    Data object for constant values.
+    Symbol representing constant values in symbolic equations.
     """
 
     is_Constant = True
     is_Scalar = True
 
-    def __new__(cls, *args, **kwargs):
-        kwargs.update({'options': {'evaluate': False}})
-        return AbstractSymbol.__new__(cls, *args, **kwargs)
+    is_SymbolicFunction = False
 
     def __init__(self, *args, **kwargs):
-        if not self._cached():
-            self.name = kwargs.get('name')
-            self.shape = ()
-            self.indices = ()
-            self.dtype = kwargs.get('dtype', np.float32)
-            self._value = kwargs.get('value', 0.)
+        self.name = kwargs.get('name')
+        self.dtype = kwargs.get('dtype', np.float32)
+        self._value = kwargs.get('value', 0.)
 
     @property
     def data(self):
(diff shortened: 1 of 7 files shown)


devito/function.py
```diff
@@ -44,26 +43,21 @@ def __repr__(self):
 Backward = TimeAxis('Backward')
 
 
-class Constant(SymbolicFunction, ConstantArgProvider):
+class Constant(AbstractSymbol, ConstantArgProvider):
 
     """
-    Data object for constant values.
+    Symbol representing constant values im symbolic equations.
```

devito/types.py
```diff
@@ -33,15 +33,15 @@ class Basic(object):
...
+             AbstractFunction
+    <see diagram in AbstractFunction.__doc__>
 
     All derived :class:`Basic` objects may be emitted through code generation
     to create a just-in-time compilable kernel.
     """
 
     # Top hierarchy
-    is_AbstractSymbol = False
+    is_AbstractFunction = False
```
FabioLuporini: We need ``is_AbstractSymbol`` too right ?

devito/types.py
```diff
@@ -94,7 +94,60 @@ def _cached_init(self):
         self.__dict__ = original().__dict__
 
 
-class AbstractSymbol(sympy.Function, CachedSymbol):
+class AbstractSymbol(sympy.Symbol, CachedSymbol):
```
FabioLuporini: is_AbstractSymbol = True

devito/function.py
```diff
@@ -44,26 +43,21 @@ def __repr__(self):
...
     """
 
     is_Constant = True
     is_Scalar = True
 
-    def __new__(cls, *args, **kwargs):
-        kwargs.update({'options': {'evaluate': False}})
-        return AbstractSymbol.__new__(cls, *args, **kwargs)
+    is_SymbolicFunction = False
```
FabioLuporini: This derives from ``AbstractSymbol``, so it should be unnecessary to set this to False, right?

devito/types.py
```diff
@@ -265,12 +315,13 @@ class Scalar(SymbolicData, ScalarArgProvider):
 
     def __init__(self, *args, **kwargs):
         if not self._cached():
-            self.name = kwargs.get('name')
-            self.shape = ()
-            self.indices = ()
             self.dtype = kwargs.get('dtype', np.float32)
 
     @property
+    def shape(self):
```
FabioLuporini: isn't this inherited?

devito/visitors.py
```diff
@@ -11,12 +11,13 @@
 from operator import attrgetter
 
 import cgen as c
-from sympy import Symbol
+# from sympy import Symbol
```
FabioLuporini: leftover

mlange05:
I believe I have addressed the relevant comments. Are we happy with this now?

Also, I realise that this requires further updates to the docstrings and user-documentation, but I think this is still partially in flux.

FabioLuporini:
Yes, good to go, thanks.

mlange05:
Ok, cool. Merging...

FabioLuporini:
@mlange05 merging...?

mlange05:
Lolz, yes merging. Got sidetracked again...

PR #380, opened 2017-10-17
Added draft of devito dask FWI tutorial

ggorman:
Changed temporary directory name to use mkdtemp rather than getuid(). The issue is that when using Devito in distributed mode we need to guarantee the filenames are unique or a race condition arises.


FabioLuporini:
This should have been two different commits. Next time, `git add -p`

examples/seismic/tutorials/04_dask.ipynb
```diff
@@ -0,0 +1,269 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "import dask\n",
```
FabioLuporini: This is not a hard requirement of Devito. We should catch the exception if an ImportError arises at this point.

Also, we should state somewhere that `dask` is "an optional requirement of Devito, to..." ( @mlange05 ?)

examples/seismic/tutorials/04_dask.ipynb
```diff
@@ -0,0 +1,269 @@
...
+    "from distributed import Client\n",
```
FabioLuporini: same as above

examples/seismic/tutorials/04_dask.ipynb
```diff
@@ -0,0 +1,269 @@
...
+    "\n",
+    "from scipy import optimize\n",
```
FabioLuporini: same as above

devito/compiler.py
```diff
@@ -194,12 +194,14 @@ def get_tmp_dir():
...
-    if not path.exists(tmpdir):
-        mkdir(tmpdir)
-
-    return tmpdir
+    global _devito_compiler_tmpdir
+    try:
+        path.exists(_devito_compiler_tmpdir)
+    except:
+        suffix = "devito-%s" % getuid()
+        _devito_compiler_tmpdir = mkdtemp(prefix=gettempdir(), suffix=suffix)
```
FabioLuporini: lots of tests failing, probably because you're missing a `/` somewhere here. How could this work on your machine?

examples/seismic/tutorials/04_dask.ipynb
```diff
@@ -0,0 +1,269 @@
...
+    "    solver = AcousticWaveSolver(true_model, src, rec, space_order=4)\n",
+    "\n",
+    "    # Generate synthetic receiver data from true model\n",
+    "    true_d, _, _ = solver.forward(src=src, m=true_model.m)\n",
+    "\n",
+    "    return src, true_d, nt, solver\n",
+    "\n",
+    "# Create FWI gradient kernel \n",
+    "def fwi_gradient_i(work):\n",
+    "    clear_cache()\n",
```
FabioLuporini: comment why this is necessary

examples/seismic/tutorials/04_dask.ipynb
```diff
@@ -0,0 +1,269 @@
...
+    "    g[-25:,:] = 0.\n",
+    "\n",
+    "    return f, g\n",
+    "\n",
+    "def sum_pairs_g(a, b):\n",
+    "    return a[0]+b[0], a[1]+b[1]\n",
+    "\n",
+    "def sum_g(client, futures):\n",
+    "    # Simple tree reduction\n",
+    "    while len(futures) > 1:  # while there is more than one element left\n",
```
FabioLuporini: you can drop this comment ..

examples/seismic/tutorials/04_dask.ipynb
```diff
@@ -0,0 +1,269 @@
...
+    "    work = [(param, i) for i in range(parameters['nshots'])]\n",
+    "    gi = client.map(fwi_gradient_i, work)\n",
+    "    f, g = sum_g(client, gi).result()\n",
+    "    client.close()\n",
+    "    \n",
+    "    return f, g.flatten().astype(np.float64)\n",
+    "\n",
+    "def fwi(model, param): \n",
+    "    result = optimize.minimize(fwi_gradient, model.m.data.flatten().astype(np.float64), args=(param, ),\n",
+    "            method='L-BFGS-B', jac=True, bounds=tuple((0.08, 0.25) for i in range(model.m.data.flatten().size)),\n",
```
FabioLuporini: out of curiosity, remind me what does ``L-BFGS-B``  stand for?

FabioLuporini: why ``(0.08, 0.25)`` as bounds?

mlange05:
Ok, so to be honest, I'm unclear as to what the purpose of this notebook is. If it is intended as an actual tutorial for the website, it needs significant amounts of documentation and hand-holding, because in its current form it is pretty impenetrable for someone coming in cold on this. A single cell setting up nearly all utility functions with bare minimum documentation is not particularly helpful for making this easy-to-understand.

If, on the other hand it is intended as a work-in-progress/proof-of-concept from which to derive a hand-holding tutorial at a later stage, I would recommend turning it into an example script (eg. `example_fwi_dask.py`) with a command-line interface to adjust things like model, nshots or optimization algorithm. This way we can add further parameterizations at a later stage and use this to explore/benchmark parallel execution on clusters.

examples/seismic/tutorials/04_dask.ipynb
mloubout: creates new `u0` each call but is not cleared automatically for some reason, so memory overload without it.

mloubout:
@ggorman I know getting a pretty result is nice, but this not how TV constraint work in any way. Constraints are part of the optimization problem and updates, not a filter for the final result, showing this will only say we have no idea what we're doing.

mlange05:
Ok, so this is looking much better and more informative now. Notebook is clean, pretty and gree (travis), so I'm happy. I leave the method discussion to @ggorman and @mloubout.

devito/compiler.py
```diff
@@ -194,12 +194,13 @@ def get_tmp_dir():
 
     :return: Path to a devito-specific tmp directory
     """
-    tmpdir = path.join(gettempdir(), "devito-%s" % getuid())
+    global _devito_compiler_tmpdir
```
FabioLuporini: why does it need to be ``global``?

environment.yml
```diff
@@ -15,6 +15,8 @@ dependencies:
   - psutil>=5.1.0
   - sphinx
   - sphinx_rtd_theme
+  - dask
```
FabioLuporini: @mlange05 are these, and should these be, compulsory requirements ?

mlange05: Not really, since the use of Dask is not actually part of core Devito. Having it in our "recommended environment" but not in the "hard dependencies" in `setup.py` is thus appropriate, I think.

mlange05:
Except for the global compiler path commented on above, this looks ready to go (minus one final merge-from-master). What's the verdict on this then? @FabioLuporini @ggorman

devito/compiler.py
mlange05: Ping @ggorman

FabioLuporini:
I'm happy to land this, as long as we clarify on the use of ``global``

devito/compiler.py
mlange05: Ping again, @ggorman. This is ready to land otherwise, so please respond.

mloubout:
nitpicking, units would be helpful on error plot (absolute, %, relative?)

devito/compiler.py
ggorman: What I want here is for _devito_compiler_tmpdir to be globally cached so that it only creates one unique tmp dir per process. get_tmp_dir() is not a class member so I cannot instead use memoize.

mlange05:
> nitpicking, units would be helpful on error plot (absolute, %, relative?)

Agreed, but needs to be retrofitted to all use cases of `plot_image`. Will fix in upcoming tweaks

> Also image not rendered in cell 2 for some reason

It renders correctly, it just happens to be a constant values throughout.

mlange05:
Merging...

PR #381, opened 2017-10-17
Sparse generalization

mloubout:
Attempt to generalize `SparseFunction`, Source/Rec are untouched as this ones are specific to the seismic setup.
This is open to discussion/naming, but this is needed as the current one has too many hard-coded issues.
Summary:
 - `time` is not mandatory anymore
 - `coordinates` can be associated with any dimension, and falls back to default if not (a coordinate is by definition a values along a dimension, not a magic number). 
- Extra function to assign instead of inject a value at point positions :**unsafe function are we cannot check at symbolic level if two values of the coordinate will assign to the same index of the wavefield**
- Extra options to take cumulative sum instead of current time value in interpolate

### DIFF
--- a/devito/function.py
+++ b/devito/function.py
@@ -458,25 +458,26 @@ class SparseFunction(CompositeFunction):
 
     def __init__(self, *args, **kwargs):
         if not self._cached():
-            self.nt = kwargs.get('nt')
+            self.nt = kwargs.get('nt', 0)
             self.npoint = kwargs.get('npoint')
             self.ndim = kwargs.get('ndim')
-            kwargs['shape'] = (self.nt, self.npoint)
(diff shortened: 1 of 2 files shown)


mlange05:
Generally I like what this PR is trying to achieve, but I think it needs to be consolidated with #382, as this will negate various assumptions made in this branch. I'll take a closer look at this soon.

mloubout:
Same as #340, needs fresh restart, closing

mlange05:
> Same as #340, needs fresh restart, closing

Yes, in both cases, can we start by adding a clear test or example to illustrate the intent? This way potential implementation choices and API changes can be discussed with the whole team beforehand, instead of trial-and-error approaches that lead to longstanding discussion PRs? Thanks.

mloubout:
First case is fixed (by you)
This one I have the idea, just need to rewrite for new API

PR #383, opened 2017-10-18
misc tweaks

FabioLuporini:
- Minor improvements to benchmark.py (will need for the upcoming tutorial)
- Better core.autotuning
- Better configuration system
- More env vars supported to drive YASK performance tuning
- Less verbose default logger. Now you'll see DSE/DLE output reports only if logging level = DEBUG


devito/parameters.py
```diff
@@ -91,8 +97,23 @@ def initialize(self):
     'DEVITO_DEBUG_COMPILER': 'debug_compiler',
 }
 
+configuration = Parameters("Devito-Configuration")
+"""The Devito configuration parameters."""
+
+configuration.add('travis_test', 0, [0, 1], lambda i: bool(i))
```
mlange05: `travis_test` one is a leftover and can actually be removed.

devito/parameters.py
```diff
@@ -91,8 +97,23 @@ def initialize(self):
...
+# Set the Instruction Set Architecture (ISA)
+default_isa = 'cpp'
+ISAs = [None, 'cpp', 'avx', 'avx2', 'avx512', 'knc']
+configuration.add('isa', None, ISAs)
+# Set the CPU architecture (only codename)
+default_platform = 'intel64'
+PLATFORMs = [None, 'intel64', 'sandybridge', 'ivybridge', 'haswell',
+             'broadwell', 'skylake', 'knc', 'knl']
```
mlange05: Do we still support KNC?

devito/parameters.py
```diff
@@ -136,3 +157,54 @@ def init_configuration():
...
+
+
+def print_state():
+    """Print the current configuration state."""
+    from devito.logger import info
+    for k, v in configuration.items():
+        info('%s: %s' % (k, v))
+
+
+def infer_cpu():
```
mlange05: Probably obvious from name, but docstring would still be nice.

devito/parameters.py
FabioLuporini: really? I can drop this entire line? Remind me what was it used for? it's been around for ages (and I think I moved it more than once xD)

devito/parameters.py
FabioLuporini: well, what can I say...there's an ``IntelMICCompiler``, and yask supports it, so I'd leave it there... for the time being. I guess..

devito/parameters.py
FabioLuporini: Done

FabioLuporini:
Green => merging

FabioLuporini:
Merged.

devito/parameters.py
mlange05: For context, this was introduced with the FWI tutorial to control runtime on travis. It's been removed with the introduction of `nbval` which allows us to simply skip cells.

PR #384, opened 2017-10-23
yask: Support for Constant objects

FabioLuporini:
plus a few minor tweaks/additions

### DIFF
--- a/tests/test_yask.py
+++ b/tests/test_yask.py
@@ -5,8 +5,8 @@ import pytest  # noqa
 
 pexpect = pytest.importorskip('yask')  # Run only if YASK is available
 
-from devito import (Eq, Operator, Function, TimeFunction, SparseFunction, Backward,
-                    time, t, x, y, z, configuration, clear_cache)  # noqa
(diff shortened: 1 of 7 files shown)


tests/test_yask.py
```diff
@@ -358,6 +358,30 @@ def test_capture_vector_temporaries(self):
...
+        assert 'run_solution' in str(op)
+        op.apply(yu4D=u, c=c, t=11)
+        # Check YASK did its job and could read constant grids w/o problems
+        assert np.all(u.data[0] == 20.)
+        # Check the Constant could be read correctly even in Devito-land, i.e.,
+        # outside of run_solution
+        assert p.data[0][0] == 3.
+        # Check re-executing with another constant gives the correct result
+        c2 = Constant(name='c', value=5.)
+        op.apply(yu4D=u, c=c2, t=3)
```
mlange05: There are a few different ways to use constants, eg. by supplying explicit values, and/or creating  `Constant` without initial value. All these work though (tested locally), so that is optional.

FabioLuporini: I'll add them in my current local branch , good spot

FabioLuporini:
Merged.

PR #385, opened 2017-10-23
Preliminaries to timeslot optimization

FabioLuporini:
Lots of edits, but, in practice very, very, very little is actually changing.

The reason you see lots of edits is that I've moved a few things around, or (in the case of the DLE edits) I've improved the structuring so that compiler passes work at the level of individual abstract syntax trees, rather than "lists of abstract syntax trees" as was before. So the edits are due to
previously:
```
def pass(ASTs):
    for i in ASTs:
         do_something(i)
```
now:
```
def pass(i):
    do_something(i)
```
Perhaps you don't appreciate it from this description, but the code is now much nicer in the DLE

The main change of this PR is that now data dependence analysis (leading to decoration of ASTs with properties such as PARALLEL, SEQUENTIAL, etc.) is performed **before** the DLE, through the ``flow.py`` module. This will allow me, in a subsequent PR, to implement the "timeslot optimization" (see #yask channel on Slack or documentation of the new WRAPPABLE property). Indeed, the timeslot optimization requires data dependence analysis, but, logically, has nothing to do with the DLE. Likewise, there was no reason for data dependence analysis to occur at such a late stage as the DLE.

I've also refactored code here and there, and refreshed obsolete documentation.

The data dependence analysis should also now be more robust.

### DIFF
--- a/devito/nodes.py
+++ b/devito/nodes.py
@@ -11,7 +11,7 @@ from sympy import Eq, Indexed, Symbol
 from devito.cgen_utils import ccode
 from devito.dse import as_symbol, retrieve_terminals
 from devito.stencil import Stencil
-from devito.tools import as_tuple, filter_ordered, flatten
+from devito.tools import as_tuple, filter_ordered, filter_sorted, flatten
 from devito.arguments import ArgumentProvider, Argument
 import devito.types as types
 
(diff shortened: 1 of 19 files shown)


FabioLuporini:
thoughts @mlange05 ?

devito/nodes.py
```diff
@@ -6,14 +6,14 @@
 from collections import Iterable, OrderedDict
 
 import cgen as c
-from sympy import Eq
+from sympy import Eq, Indexed, Symbol
```
mlange05: Is there a reason we're not using our internal types here?

FabioLuporini: So this goes back to a question I asked you online -- "Can we assume that we always **only** see our own data types" -- for which there was no certain answer. So I picked the sympy ones, since it's not harmful... I could try using our types (other PR) and see what happens

FabioLuporini:
Merging, then

FabioLuporini:
Merged.

PR #386, opened 2017-10-23
Add example of checkpointing using pyrevolve

navjotk:
PR 3/3. This is the last in the checkpointing series of PRs. Here we add an example of using pyrevolve in combination with devito to use checkpointing based on the revolve algorithm. 

In more detail, this PR:
- Updates the `gradient_example.py` to more recent API conventions and adds it to travis
- Adds a new example called `checkpointing_example.py`, based on the above, which reproduces the same result while respecting a limit on the maximum memory consumption
- Adds a `size` property to Devito types so they report their size in memory (this is the only change in the core of devito that this PR has)
- Updates the `GradientOperator` to have a `save=False` mode
- Adds 2 new tests that use and test the checkpointing infrastructure

Although this might be the first time most of you are seeing the code for the checkpointing implementation, we have already discussed this to the point of knowing that this a suboptimal implementation. In particular:
- The load and save procedures violate the zero-cost-checkpointing assumption of Revolve (this possibly requires a complete re-think of the checkpointing implementation and maybe even the pyRevolve API)
- In the "worst case" where checkpointed steps are far away, we would store 2 timesteps (for a second-order-in-time formulation) per checkpoint. This implementation stores 3, i.e. an extra timestep per checkpoint. This can be fixed relatively easily. 
- In the "best case", where consecutive steps are being checkpointed, storing 2 timesteps per checkpoint (remember this implementation stores 3 timesteps/checkpoint) involves storing multiple copies of the same timestep, another wasteful side-effect of the current implementation. I'm currently in the process of resolving this but it's quite complex, requiring work on the theory as well as the implementation.
- Of course, all this is still in memory and more work needs to go into pyRevolve to support multi-stage checkpointing. That is another parallel WIP. This, however, depends on the solution to 1. above. 

Despite the shortcomings, having an(y) implementation of checkpointing means that it is now possible to solve a new set of problems that we couldn't before.


devito/types.py
```diff
@@ -287,6 +289,10 @@ def _mem_heap(self):
         in a C module, False otherwise."""
         return False
 
+    @property
+    def size(self):
```
FabioLuporini: docstring

examples/checkpointing/checkpoint.py
```diff
@@ -0,0 +1,65 @@
...
+        i_ptr_hi = 0
+        for s in self.symbols:
+            i_ptr_hi = i_ptr_hi + s.size
+            s.data[:] = ptr[i_ptr_lo:i_ptr_hi].reshape(s.shape)
+            i_ptr_lo = i_ptr_hi
+
+    @property
+    def size(self):
+        """The memory consumption of the data contained in a checkpoint."""
+        return sum([s.size for s in self.symbols])
```
FabioLuporini: times number of bytes per element?

examples/checkpointing/checkpointing_example.py
```diff
@@ -0,0 +1,143 @@
...
+        error2[i] = np.absolute(.5*linalg.norm(d - rec_t.data)**2 - F0 - H[i] * G)
+
+    # Test slope of the  tests
+    p1 = np.polyfit(np.log10(H), np.log10(error1), 1)
+    p2 = np.polyfit(np.log10(H), np.log10(error2), 1)
+    assert np.isclose(p1[0], 1.0, rtol=0.1)
+    assert np.isclose(p2[0], 2.0, rtol=0.1)
+
+
+class CheckpointedGradientExample(object):
```
FabioLuporini: Why do we need a class for this? aren't free functions enough?

examples/seismic/acoustic/operators.py
```diff
@@ -149,8 +150,8 @@ def GradientOperator(model, source, receiver, time_order=2, space_order=4, **kwa
 
     # Gradient symbol and wavefield symbols
     grad = Function(name='grad', grid=model.grid)
-    u = TimeFunction(name='u', grid=model.grid, save=True, time_dim=source.nt,
-                     time_order=2, space_order=space_order)
+    u = TimeFunction(name='u', grid=model.grid, save=save, time_dim=source.nt if save
```
FabioLuporini: @mloubout why was ``save=True`` before? Was it just an oversight?

devito/arguments.py
```diff
@@ -77,14 +77,31 @@ class ScalarArgument(Argument):
     def __init__(self, name, provider, reducer=lambda old, new: new, default_value=None):
         super(ScalarArgument, self).__init__(name, provider, default_value)
         self.reducer = reducer
+        self._frozen = False
+
+    def reset(self):
+        super(ScalarArgument, self).reset()
+        self._frozen = False
```
FabioLuporini: what's ``frozen`` for?

devito/arguments.py
```diff
@@ -77,14 +77,31 @@ class ScalarArgument(Argument):
...
 
     def verify(self, value, enforce=False):
-        # Assuming self._value was initialised as appropriate for the reducer
+        # If we're being passed a null value, ignore
         if value is not None:
             if self._value is not None and not enforce:
-                self._value = self.reducer(self._value, value)
+                # We already have a value and the value passed
+                # here is not enforced (i.e. optional)
```
FabioLuporini: can you remind me the use case for enforced?

examples/checkpointing/checkpoint.py
```diff
@@ -0,0 +1,65 @@
+from pyrevolve import Checkpoint, Operator
+from devito import TimeFunction
+
+
+class DevitoOperator(Operator):
```
FabioLuporini: Why not a fist-citizen class in ``operator.py`` ? 

A name such as ``CheckpointOperator`` would also be much, much more informative

FabioLuporini: Other question (also based on what I read in ``apply.__doc__``). Does pyrevolve depend on Devito? I guess no -- in fact, it should be the other way round. So how can pyrevolve know about ``Operators`` and ``apply``? Or am I getting this completely wrong?

examples/checkpointing/checkpoint.py
```diff
@@ -0,0 +1,65 @@
...
+    """Class to wrap devito operators so they conform to the pyRevolve API
+    """
+    def __init__(self, op, args, argnames):
+        self.op = op
+        self.args = args
+        self.argnames = argnames
```
FabioLuporini: can't the argname be obtained directly from the arg?

examples/checkpointing/checkpoint.py
```diff
@@ -0,0 +1,65 @@
...
+            pyRevolve.Operator.apply() without caring about these extra arguments while
+            this method passes them on correctly to devito.Operator
+        """
+        args = self.args.copy()
+        args[self.argnames['t_start']] = t_start
+        args[self.argnames['t_end']] = t_end
+        self.op.apply(**args)
+
+
+class DevitoCheckpoint(Checkpoint):
```
FabioLuporini: so, again, why don't we put all this stuff in a ``checkpointing.py`` module, rather than as an example?

examples/checkpointing/checkpoint.py
```diff
@@ -0,0 +1,65 @@
...
+    """Devito's concrete implementation of the Checkpoint abstract base class provided by
+       pyRevolve. Holds a list of symbol objects that hold data.
+    """
+    def __init__(self, symbols):
+        """Intialise a checkpoint object. Upon initialisation, a checkpoint
+        stores only a reference to the symbols that are passed into it.
+        The symbols must be passed as a mapping symbolname->symbolobject."""
+        assert(all(isinstance(s, TimeFunction) for s in symbols))
```
FabioLuporini: if these are functions, the variable should be called ``functions``, not ``symbols``

examples/checkpointing/checkpoint.py
```diff
@@ -0,0 +1,65 @@
...
+        self._dtype = symbols[0].dtype
```
FabioLuporini: we need stronger checks

examples/checkpointing/checkpoint.py
```diff
@@ -0,0 +1,65 @@
...
+        """Overwrite live-data in this Checkpoint object with data found at
+        the ptr location."""
+        i_ptr_lo = 0
+        i_ptr_hi = 0
+        for s in self.symbols:
+            i_ptr_hi = i_ptr_hi + s.size
+            ptr[i_ptr_lo:i_ptr_hi] = s.data.flatten()[:]
+            i_ptr_lo = i_ptr_hi
+
+    def load(self, ptr):
```
FabioLuporini: just a thought that is not strictly related to this PR: we'll need to time up these methods. We'll soon need a ``@timed_method`` decorator to time critical methods such as these ones, jit-compilation, and others

examples/seismic/acoustic/gradient_example.py
```diff
@@ -142,9 +102,10 @@ def run(dimensions=(50, 50, 50), tn=750.0,
     # Test slope of the  tests
     p1 = np.polyfit(np.log10(H), np.log10(error1), 1)
     p2 = np.polyfit(np.log10(H), np.log10(error2), 1)
+    print(p1, p2)
```
FabioLuporini: leftover, or use logger

examples/checkpointing/checkpointing_example.py
```diff
@@ -0,0 +1,143 @@
...
+from numpy import linalg
+from math import floor
+from devito import Function, TimeFunction
+from examples.seismic import Receiver, demo_model, RickerSource
+from examples.seismic.acoustic import ForwardOperator, GradientOperator, smooth10
+from examples.checkpointing.checkpoint import DevitoCheckpoint, DevitoOperator
+from pyrevolve import Revolver
+
+
+def setup(shape, tn, spacing, time_order, space_order, nbpml):
```
FabioLuporini: if this is seismic/acoustic based, then it should probably live in the seismic folder 

much better: move this to /seismic ; add here the most trivial example that one can come up with

examples/checkpointing/checkpoint.py
navjotk: > Does pyrevolve depend on Devito? 
No
> So how can pyrevolve know about Operators and apply? 
pyRevolve provides an `Operator` ABC that has an apply method. This is a concrete implementation of the `pyrevolve.Operator` class that wraps the `devito.Operator` class.

examples/checkpointing/checkpoint.py
navjotk: This is a `checkpointing.py` module, it is just called `checkpoint.py` (if it is the name you are concerned about, I am happy to change). Checkpointing is not a feature that devito provides. Pyrevolve provides the feature to implement checkpointing with just a few lines, given you have the appropriate operators (that devito provides). This is an example of using devito and pyrevolve together. Just like `examples/seismic/operators.py` etc., `examples/checkpointing/` is an appropriate place for it, no?

examples/checkpointing/checkpoint.py
navjotk: There is a related benchmarking script I have left for a future PR.

examples/checkpointing/checkpoint.py
navjotk: If you're using `numpy` all over, it is more convenient to track (size in bytes) as (number of elements, dtype) since in creating a numpy array you would need to know the number of elements and the dtype - not the size in bytes.

examples/checkpointing/checkpointing_example.py
navjotk: This is the most trivial (and yet useful) example I can think of, of using a devito operator with pyrevolve. "The most trivial example" lives in the examples folder of pyrevolve.

examples/checkpointing/checkpointing_example.py
navjotk: For benchmarking purposes, we split the gradient test into 3 steps - setup, gradient calculation and verification. The only step we want to time is the gradient calculation (and hence run multiple times). We use this class's objects to store the objects that were setup in step 1 and will be used by the gradient calculation.

examples/seismic/acoustic/operators.py
navjotk: It wasn't possible to calculate the gradient without having `u(..., save=True...)` before. This ability is being brought in by this PR (or was brought in by the previous PR, depending on how you want to look at it).

devito/arguments.py
navjotk: Added docstring to `verify` for explanation

devito/arguments.py
navjotk: Added docstring to `verify` for explanation

devito/types.py
navjotk: Fixed

examples/checkpointing/checkpoint.py
navjotk: done

examples/seismic/acoustic/gradient_example.py
navjotk: done

examples/checkpointing/checkpoint.py
navjotk: Changed so it becomes more clear

examples/checkpointing/checkpoint.py
navjotk: Done

devito/types.py
navjotk: Done

examples/checkpointing/checkpoint.py
FabioLuporini: I wasn't concerned about the name. OK, I have a different view about this, but I may well be wrong:

> Checkpointing is not a feature that devito provides

I thought checkpointing is a feature that Devito **provides**; it provides it through ``pyrevolve``, but that's a different story. This way, any Devito users would be able to use checkpointing, as long as a checkpointable Operator were used. Or, perhaps, all Operators are checkpointable... anyway, let me think about this

examples/checkpointing/checkpoint.py
```diff
@@ -17,49 +26,49 @@ def apply(self, t_start, t_end):
...
-        self.revolver = None
+        stores only a reference to the objects that are passed into it."""
+        assert(all(isinstance(o, TimeFunction) for o in objects))
+        self.objects = objects
 
     @property
     def dtype(self):
-        return self._dtype
+        try:
+            return self.objects[0].dtype
```
FabioLuporini: What I meant in the original comment was whether we should check that **all** objects/functions have **same** dtype

examples/seismic/acoustic/operators.py
FabioLuporini: oh I see, of course

examples/checkpointing/checkpoint.py
navjotk: What are you suggesting here anyway?

examples/checkpointing/checkpoint.py
navjotk: Done

examples/checkpointing/checkpoint.py
```diff
@@ -0,0 +1,74 @@
+from pyrevolve import Checkpoint, Operator
+from devito import TimeFunction
+
+
+class CheckpointOperator(Operator):
+    """Devito's concrete implementation of the ABC pyrevolve.Operator. This class wraps
+       devito.Operator so it conforms to the pyRevolve API. pyRevolve will call apply
+       with arguments t_start and t_end. Devito calls these arguments t_s and t_e so
+       the following dict is used to perform the translations between different names.
+       :param op: The devito.Operator object that this object will wrap
```
FabioLuporini: If this is a wrapper, why inheriting from ``Operator`` ?

examples/checkpointing/checkpointing_example.py
```diff
@@ -0,0 +1,142 @@
...
+    rec_s = Receiver(name='rec_s', grid=model.grid, ntime=nt, npoint=nrec,
+                     coordinates=rec_t.coordinates.data)
+
+    # Receiver for Gradient run
+    rec_g = Receiver(name="rec_g", grid=model.grid, ntime=nt,
+                     coordinates=rec_s.coordinates.data)
+
+    # Create the forward wavefield to use (only 3 timesteps)
+    u = TimeFunction(name="u", grid=model.grid, time_order=time_order,
+                     space_order=space_order, save=False, dtype=model.dtype)
```
mlange05: `dtype` parameter is no longer needed, as it's inferred from `model`.

examples/checkpointing/checkpointing_example.py
```diff
@@ -0,0 +1,142 @@
...
+
+    v = TimeFunction(name="v", grid=model.grid, time_order=time_order,
+                     space_order=space_order, save=False, dtype=model.dtype)
```
mlange05: Drop `dtype` as above.

examples/seismic/acoustic/gradient_example.py
```diff
@@ -2,119 +2,79 @@
...
+    rec_s = Receiver(name='rec_s', grid=model.grid, ntime=nt, npoint=nrec,
+                     coordinates=rec_t.coordinates.data)
 
     # Create the forward wavefield to use (only 3 timesteps)
     # Once checkpointing is in, this will be the only wavefield we need
-    u_nosave = TimeFunction(name="u", shape=model.shape_domain, time_dim=nt,
-                            time_order=time_order, space_order=space_order, save=False,
-                            dtype=model.dtype)
+    u_nosave = TimeFunction(name="u", grid=model.grid, time_order=time_order,
+                            space_order=space_order, save=False, dtype=model.dtype)
```
mlange05: Drop `dtype`

examples/seismic/acoustic/gradient_example.py
```diff
@@ -2,119 +2,79 @@
...
 
     # Forward wavefield where all timesteps are stored
     # With checkpointing this should go away <----
-    u_save = TimeFunction(name="u", shape=model.shape_domain, time_dim=nt,
+    u_save = TimeFunction(name="u", grid=model.grid, time_dim=nt,
```
mlange05: `dtype`

examples/seismic/acoustic/gradient_example.py
```diff
@@ -2,119 +2,79 @@
...
     # Receiver for Gradient
-    # Confusing nomenclature because this is actually the source for the adjoint
-    # mode
-    rec_g = Receiver(name="rec", coordinates=rec_s.coordinates.data,
-                     data=rec_s.data - rec_t.data)
+    rec_g = Receiver(name="rec", coordinates=rec_s.coordinates.data, grid=model.grid,
+                     data=rec_s.data - rec_t.data, dt=dt)
     # Gradient symbol
-    grad = Function(name="grad", shape=model.shape_domain, dtype=model.dtype)
+    grad = Function(name="grad", grid=model.grid, dtype=model.dtype)
```
mlange05: Yep, you guessed it.

examples/checkpointing/checkpoint.py
```diff
@@ -0,0 +1,74 @@
...
+        return self._dtype
+
+    def save(self, ptr):
+        """Overwrite live-data in this Checkpoint object with data found at
+        the ptr location."""
+        i_ptr_lo = 0
+        i_ptr_hi = 0
+        for o in self.objects:
+            i_ptr_hi = i_ptr_hi + o.size
+            ptr[i_ptr_lo:i_ptr_hi] = o.data.flatten()[:]
```
mlange05: Note for future development: This looks to be the identical use case to the `dump`/`load` routnes used in the Dask example. We may want to lift that utility to the `relevant `Function` types.

mlange05:
One final note though: I'm still very concerned that the unrolled and much-more-complex-than-necessary version of the gradient test live in `examples`, while the clean `Wavesolver`-based test is buried in tests. A curious new user is much more likely to stumble upon the complex example, negating the primary point of of our high-level utilities.  In my eye this is simply a very bad practice and risks alienating new users.

mlange05:
I think the checkpointing wrappers are best placed in the `examples` folder (I don't care where exactly there), as checkpointing itself is an application-level algorithm and does not really match the FD abstractions in core Devito. The current interfaces actually convey that modularity quite well, I think.

environment.yml
```diff
@@ -18,4 +18,5 @@ dependencies:
   - pip:
     - "git+https://github.com/inducer/cgen"
     - "git+https://github.com/inducer/codepy"
+    - "git+https://github.com/opesci/pyrevolve"
```
mlange05: I just realised that Cython itself takes ages to install via pip. Can you please add `cython` to the conda part of our env to speed up the travis install? Thanks

examples/checkpointing/checkpoint.py
navjotk: Note that it is *inheriting* from `pyrevolve.Operator` while wrapping `devito.Operator`.

FabioLuporini: ah, right!

examples/checkpointing/checkpointing_example.py
navjotk: Fixed

examples/checkpointing/checkpointing_example.py
navjotk: Fixed

examples/seismic/acoustic/gradient_example.py
navjotk: Fixed

examples/seismic/acoustic/gradient_example.py
navjotk: Fixed

examples/seismic/acoustic/gradient_example.py
navjotk: Fixed

environment.yml
navjotk: I have fixed it but please confirm that this is what you were talking about.

mlange05: :+1:

navjotk:
I believe I have addressed everything. Tests seem to be passing (after pulling out the dse fix into its own PR). I will merge as soon as I see the green light from travis.

FabioLuporini:
> I will merge as soon as I see the green light from travis.

Yes please. Thanks

navjotk:
This is finally green. There was an actual bug exposed by the failing yask tests (fixed in 643ea47) and some tests were simply outdated and updated in this PR.

FabioLuporini:
Not sure why you're trying to bring in `devito/algorithm.py`, which was probably in a branch of mine before being renamed into `flow.py`. Perhaps a leftover

navjotk:
Thanks for pointing that out. Indeed it was leftover from a rebase.

tests/test_yask.py
```diff
@@ -192,7 +192,7 @@ def test_increasing_multi_steps(self):
         u = TimeFunction(name='yu4D', grid=grid, space_order=0)
         u.data.with_halo[:] = 0.
         op = Operator(Eq(u.forward, u + 1.))
-        op(yu4D=u, t=12)
+        op(yu4D=u, t=11)
```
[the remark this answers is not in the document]
navjotk: If you read the docstring of the test and the asserts below, it seems to me (please confirm) that you always wanted to execute 11 time steps but were compensating for the fact that, before this change, passing `t=12` would actually run 11 time steps, i.e. you were passing `required_time_steps + 1` earlier. Now that this has been included inside arguments, you can just pass `required_time_steps`.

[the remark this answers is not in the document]
FabioLuporini: oh I see. Yes, that' s quite plausible. Thanks

PR #389, opened 2017-10-26
Reorganisation of the intermediate representations

FabioLuporini:
OK, this may look like a punch in the eye to some of you, but (just like other recent PRs) it actually changes nothing in how things work. It's just moving things around, which also has the positive effect of untangling some circular dependencies. 

Commits should be quite self-explanatory. However, in essence, what this PR does is to gather together the various IR representations into a new folder (``ir``). Different IRs are used during different stages of the compilation process. 

Also, a ``symbolics`` folder now contains "all stuff built on top of sympy`` that were previously spread throughout the code base. In particular, some functions/classes/... have been moved from the ``DSE`` to ``symbolics``.

EDIT: why all this? Because in the next PR, I'll need to add proper data dependence analysis, which greatly benefits from having cleanly organised IRs


FabioLuporini:
Instead of looking at the edits, it's probably more effective taking a look at the new structure of the ``devito`` directory tree

mlange05:
As far as I can tell this all looks good and makes nice sense. Good to go from me.

mlange05:
Argh, this needs a rebase, but is otherwise good to go. @FabioLuporini

FabioLuporini:
Merged.

PR #391, opened 2017-10-27
Enable dynamic backend switching

mlange05:
Shuffles the configuration options around a little to enable us to switch to the YASK backend dynamically from within a single Python interpreter - for example an Azure notebook session.

### DIFF
--- a/devito/__init__.py
+++ b/devito/__init__.py
@@ -42,6 +42,17 @@ configuration.add('debug_compiler', 0, [0, 1], lambda i: bool(i))
 configuration.add('backend', 'core', list(backends_registry),
                   callback=init_backend)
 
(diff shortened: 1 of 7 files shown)


FabioLuporini:
Yes, thanks.

I wonder, however, whether we could also move the leftover configuration options from ``parameters.py`` to ``__init__.py`` (ISA, platform), as well as ``infer_cpu()`` (or perhaps even better: this one to ``tools.py``). At this point, I think they have no right to be in ``parameters.py`` any more

mlange05:
@FabioLuporini Ok, moved things around. I also switched `cgen` to a (quicker) conda install. Should be good now.

FabioLuporini:
Yes, good to go

mlange05:
Ok, merging...

PR #392, opened 2017-10-28
Abc seperated

mloubout:
Annnnnnd  it's back

Still not working the way I want, but highlight the idea if people have better way to do it. Isuues remain the same


mloubout:
Closing until cleaned up

PR #394, opened 2017-11-01
Dask: Adjust `SparseFunction` API usage with grids

mlange05:
The Dask tutorial actually breaks on executing the FWI cells without this.


mlange05:
Ok, approved and green. Going in now....

PR #395, opened 2017-11-01
Some arguments hot fixes

navjotk:
Discovered a couple of bugs in the arguments code while debugging the failing tests on the other branch. Since these bugs are now blockers, separating the fixes into their own PR first.


mlange05:
Ok, looks good to me and seems to fix the rather glaring bug in my demo presentation this afternoon.

Since this is somewhat time-sensitive now, I'd like to fast-track this for the tutorial demo. So please shout now if you have any objections to this going in.

navjotk:
Note that 386 will include some tests that would have found these issues.

PR #398, opened 2017-11-03
Configuration: Fix dynamic OpenMP switch

mlange05:
When setting the `openmp` option explicitly from Python, we need to re-initialise the `Compiler` object to update the linker flags, etc. **This now enable seamless switching between sequential and threaded mode in notebooks** via `configuration['openmp'] = True`!  @ggorman @FabioLuporini  

Also adding a small fix for a debug leftover that has crept in somehow.


FabioLuporini:
yes, thanks

mlange05:
Great, merging...

PR #401, opened 2017-11-06
IR: Vector/IterationInstance types

FabioLuporini:
This is the first PR out of a series of PRs that will overhaul data dependence analysis in Devito. 

Why
====
When all PRs will be in, data dependence analysis in Devito will be vastly more robust than it is now. The Operator's expression scheduler, in particular, will benefit from this ( @mloubout this should make you happy). The loop merger will also be much smarter, being capable of detecting groupable expressions across independent expressions. And it will be much simpler to write new data dependence testers, required to determine whether certain transformations will honor the data dependences.

How
====
By computing "flow", "anti", and "output" dependences, through actual ``distance vectors analysis``. What's a distance vector? An extremely intuitive explanation is here: http://www.cs.utah.edu/~mhall/cc/Utah-L1.pdf -- but just google for it and you'll find plenty of resources. It's actually a very standard compiler concept.

Alternatives
=========
Could we do this differently? Yes
Could we reuse some existing software? Maybe
Was this necessary? Maybe
Data dependence analysis is hard. I couldn't find "easy-to-use" Python packages for this. One option was to use the Integer Set Library (ISL), but:
* I first needed to understand the problem, so this whole exercise I solved was particularly useful to me in order to understand all various sides of the problem
* I still think ISL is an overkill given what we need. ISL is a very general tool, whereas we can leverage domain properties and the peculiar data access pattern of our stencils to perform rapid, effective, _conservative_ data dependence analysis
* Whatever external tool we might have decided to use, we would have still needed to translate our IR into some suitable format -- again, this looked like an overkill.

In the future, if necessary, we might revisit some of these assumptions, I think.

About this PR
==========
Apart from some minor fixes and improvements (necessary for the upcoming PRs), this PR introduces two new types, ``Vector`` and ``IterationInstance`` (a Vector subclass), and the concept of distance between two vectors. I anticipate already that this concept of ``distance`` will be revisited in the subsequent PRs. Although tests are shipped, these new types are not used anywhere in Devito yet.


devito/ir/analysis/basic.py
```diff
@@ -0,0 +1,182 @@
...
+        precedes (< 0), equals (== 0), or succeeds (> 0) the i-th component of
+        ``other``.
+        """
+        return self.distance(other)
+
+
+class IterationInstance(Vector):
+
+    """A representation of the iteration space point accessed by a
+    :class:`Indexed` object."""
```
mlange05: This sounds a little vague. Would a little more detail help understand why this is needed?

FabioLuporini: All the docstrings are already carefully expanded/modified in the future PRs!

reference branch is ``optimize-timeslots-final``. However, it's too premature to copy those docstrings here (due to missing functionalities).

mlange05: Ah ok. Thanks

mlange05:
Up next, once travis has caught up.

FabioLuporini:
Green, merging...

FabioLuporini:
Merged.

PR #403, opened 2017-11-07
Test op fix

mloubout:
Small test fix.

The test was actually failing but wasn't caught as only three time-steps were ran and the modulo was hiding it (modulo 3)

`a` needs to be reinitialized to get correct result.


FabioLuporini:
Thanks, merging.

FabioLuporini:
Merged.

PR #404, opened 2017-11-07
general Sparse

mloubout:
Rewrite of the previously done generalization with the new API


mlange05:
Ok, I wonder if we should completely drop the `nt` argment and remove any hard-coded knowledge of time in `SparseFunctions`? This was only added to aid the seismic-specific concept of sources and receivers, but does not really generalise as a concept. Instead we could/should allow explicit `dimensions`/`shape` arguments just like `Function` objects and use the seismic `Source`/`Receiver` classes to insert the time-specific data layout and behaviour. Would that still work for `.inject(...)` expressions?

mloubout:
I guess that would work, but needs quite the work. This PR is only here to remove an odd behavior while I think of a better way of doing thing.

mlange05:
Fair enough, In that case let's merge this and file an issue for true generalisation of the sparse-point interface. I'm updating/merging this now and will file the issue.

devito/grid.py
```diff
@@ -80,6 +76,10 @@ def __init__(self, shape, extent=None, origin=None, dimensions=None,
         else:
             self.dimensions = dimensions
 
+        self.origin = tuple(Constant(name='o_%s' % dim.name, value=val)
```
mlange05: :+1:

mlange05:
Ok, going in next, once travis is happy.

PR #406, opened 2017-11-08
ir: Data dependence analysis through Scope

FabioLuporini:
This PR builds on top of #401 . 

* It extends the Vector/IterationInstance hierarchy by introducing the ``Access`` and ``TimedAccess`` subclasses
* It introduces the concept of ``Dependence`` between two ``TimedAccess``
* It introduces the concept of ``Scope``, which captures all sort of ``Dependence``s in a group of expressions. A Scope conservatively calculates flow, anti, and output ``Dependences``. "Conservatively" means that if we cannot establish whether there's a dependence between two TimedAccess or not, then a dependence is assumed.

All this stuff are still unused until the next PR, so no new functionalities added to Devito. 

Some tests are added. Many more will be added in the final PR.


FabioLuporini:
Merged.

PR #407, opened 2017-11-11
Clean up the checkpointing and gradient examples

navjotk:
- Refactored the gradient example so the checkpointing example can reuse the code
- Refactored the checkpointing tests so as to enable parametrisation
- Added another test to test_checkpointing that basically runs the checkpointing example


examples/checkpointing/checkpointing_example.py
```diff
@@ -1,141 +1,56 @@
...
 from devito import Function, TimeFunction
 from examples.seismic import Receiver, demo_model, RickerSource
 from examples.seismic.acoustic import ForwardOperator, GradientOperator, smooth10
 from examples.checkpointing.checkpoint import DevitoCheckpoint, CheckpointOperator
+from examples.seismic.acoustic.gradient_example import GradientExample
 from pyrevolve import Revolver
 
+class CheckpointingExample(GradientExample):
+    @cached_property
+    def forward_field(self):
```
FabioLuporini: cached?

examples/checkpointing/checkpointing_example.py
```diff
@@ -1,141 +1,56 @@
...
+        n_checkpoints = None
+        if maxmem is not None:
+            n_checkpoints = int(floor(maxmem*10**6/(cp.size*self.forward_field.data.itemsize)))
 
-    # Test slope of the  tests
-    p1 = np.polyfit(np.log10(H), np.log10(error1), 1)
-    p2 = np.polyfit(np.log10(H), np.log10(error2), 1)
-    assert np.isclose(p1[0], 1.0, rtol=0.1)
-    assert np.isclose(p2[0], 2.0, rtol=0.1)
+        wrap_fw = CheckpointOperator(self.forward_operator, {'u': self.forward_field, 'rec': self.rec, 'm': m0, 'src': self.src, 'dt': self.dt})
```
FabioLuporini: standard kwargs format ``a=b`` rather than ``{'a': b}`` would be better

examples/seismic/acoustic/gradient_example.py
```diff
@@ -1,109 +1,147 @@
...
 from examples.seismic import Receiver, RickerSource, demo_model
 from examples.seismic.acoustic import ForwardOperator, GradientOperator, smooth10
 
+class GradientExample(object):
+    def __init__(self, shape=(50, 50, 50), spacing=(15.0, 15.0, 15.0), tn=500., time_order=2,
+            space_order=4, nbpml=10):
+        self.time_order = time_order
+        self.space_order = space_order
+        self._setup_model_and_acquisition(shape, spacing, nbpml, tn)
+        self._true_data()
```
FabioLuporini: what's going here?

examples/seismic/acoustic/gradient_example.py
```diff
@@ -1,109 +1,147 @@
...
+
+    @cached_property
+    def dt(self):
+        return self.model.critical_dt * (1.73 if self.time_order == 4 else 1.0)
+
+    def _setup_model_and_acquisition(self, shape, spacing, nbpml, tn):
```
FabioLuporini: nitpicking: this could/should be a free function, returning whatever is needed for initialisation

examples/checkpointing/checkpointing_example.py
navjotk: Making sure that the (memory for the) field is allocated only once in the lifetime of the `CheckpointingExample` object.

examples/seismic/acoustic/gradient_example.py
navjotk: Assuming your question is about the method call `self._true_data()`, in a realistic workflow this method would load the reflection data recorded in the real experiment from disk. Since this is artificial, we run a forward simulation inside this method and store the simulated data on the object.

examples/seismic/acoustic/gradient_example.py
navjotk: If you see the previous version (the one in master currently) of this file, it does what (I think) you're looking for. Since there are so many objects that are created during initialisation, having a free function here means having ~ 10 initialised objects being returned and then being passed around for the other method calls. Having them as class properties to reduce the parameter clutter was a big part of this cleanup.

examples/checkpointing/checkpointing_example.py
navjotk: Fixed. Thanks for this one.

tests/test_checkpointing.py
```diff
@@ -1,32 +1,37 @@
...
-    wrap_fw = CheckpointOperator(fw, {'u': u, 'rec': rec_s, 'm': m0, 'src': src, 'dt': dt}
-                                 )
-    wrap_rev = CheckpointOperator(gradop, {'u': u, 'v': v, 'm': m0, 'rec': rec_g,
-                                           'grad': grad, 'dt': dt})
-    wrp = Revolver(cp, wrap_fw, wrap_rev, None, nt-time_order)
-    fw.apply(u=u, rec=rec_s, m=m0, src=src, dt=dt)
-    u_temp = np.copy(u.data)
-    rec_temp = np.copy(rec_s.data)
-    u.data[:] = 0
+@pytest.mark.parametrize('space_order', [4])
```
FabioLuporini: did you mean do parametrise this?

tests/test_checkpointing.py
```diff
@@ -42,3 +47,16 @@ def test_acoustic_save_and_nosave(shape=(50, 50), spacing=(15.0, 15.0), tn=500.,
     last_time_step = (last_time_step) % (time_order + 1)
     assert(np.allclose(u.data[last_time_step, :, :], field_last_time_step))
     assert(np.allclose(rec.data, rec_bk))
+
+
+@skipif_yask
+@pytest.mark.parametrize('space_order', [4])
```
FabioLuporini: same

tests/test_checkpointing.py
navjotk: Although I added it as a parameter, checkpointing is completely orthogonal to space order, so I left it at a single value that can easily be extended if/when someone sees the need. The interesting parametrisation here would be between different equations that access different offsets in time. That's still a couple of PRs away.

PR #410, opened 2017-11-13
Staggered grid allocation

mlange05:
This PR adds a new `staggered` keyword to `Function` and `TimeFunction` objects that adjusts the data shape and the casts in the generated code according to staggering offsets. That means that, for example, a 2D `Function` staggered in `y` (`staggered=(0, 1)` stores DoFs on the side-edges and allocates a shape of `nx, ny-1`.

Note: **This does not affect any symbolics for staggered grids or derivatives. I'm still playing with that!**

### DIFF
--- a/devito/function.py
+++ b/devito/function.py
@@ -301,6 +301,16 @@ class Function(TensorFunction):
         return sum([second_derivative(first * weight, dim=d, order=order)
                     for d in self.space_dimensions])
 
+    @property
+    def symbolic_shape(self):
+        """
+        Return the symbolic shape of the object. This is simply the
+        appropriate combination of symbolic dimension sizes shifted
+        according to the ``staggered`` mask.
+        """
+        return tuple(i.symbolic_size - s for i, s in
+                     zip(self.indices, self.staggered))
+
 
 class TimeFunction(Function):
     """
--- a/devito/function.py
+++ b/devito/function.py
@@ -319,6 +319,7 @@ class TimeFunction(Function):
     :param name: Name of the resulting :class:`sympy.Function` symbol
     :param grid: :class:`Grid` object from which to infer the data shape
                  and :class:`Dimension` indices.
+    :param staggered: (Optional) tuple containing staggering offsets.
     :param dtype: (Optional) data type of the buffered data
     :param save: Save the intermediate results to the data buffer. Defaults
                  to `False`, indicating the use of alternating buffers.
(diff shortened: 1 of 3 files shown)


devito/function.py
```diff
@@ -96,6 +96,7 @@ class Function(TensorFunction):
     :param shape: (Optional) shape of the associated data for this symbol.
     :param dimensions: (Optional) symbolic dimensions that define the
                        data layout and function indices of this symbol.
+    :param staggere: (Optional) tuple containing staggering offsets.
```

devito/function.py
```diff
@@ -136,66 +148,73 @@ def __init__(self, *args, **kwargs):
...
-                                    order=self.space_order)
-                    setattr(self.__class__, 'd%s2%s2' % (dim.name, dim2.name),
-                            property(dx2y2, 'Return the symbolic expression for '
-                                     'the second cross derivative wrt. the '
-                                     '%s and %s dimensions' %
-                                     (dim.name, dim2.name)))
+            # Dynamically add derivative short-cuts
+            self._initialize_derivatives()
+
+
```
FabioLuporini: probably flake8 is gonna bark (two blank lines)

mlange05:
Yes, thanks, and should be fixed now (ran out of battery before I could run flake8 :wink:).

devito/function.py
```diff
@@ -108,6 +109,15 @@ class Function(TensorFunction):
...
+    .. note::
+
+       The tuple :param staggered: contains a ``1`` in each dimension
+       entry that should be staggered, and ``0`` otherwise. For example,
+       ``staggered=(1, 0, 0)`` entails discretization on horizontal edges,
+       ``staggered=(0, 0, 1)`` entails discretization on vertical edges,
+       ``staggered=(0, 1, 1)`` entails discretization side facets and
+       ``staggered=(1, 1, 1)`` entails discretization on cells.
     """
```
mloubout: CAn we name these (as a Dict or something) to make it more user friendly?

mloubout: Also how does this extend to 2D?

devito/function.py
```diff
@@ -229,7 +247,7 @@ def shape_data(self):
         """
         Full allocated shape of the data associated with this :class:`Function`.
         """
-        return self.shape_domain
+        return tuple(i - s for i, s in zip(self.shape_domain, self.staggered))
```
mloubout: so indices are `i-1` not 1/2?

mlange05: No, this has nothing to do with indices; these are data shapes. In every dimension you discretize on a facet/edge/cell, you need one less data point.

devito/function.py
mlange05: For 2D, simply supply a 2-tuple for `Function` and a 3-tuple for `TimeFunction` (stagger in time). I thought about the naming conventions, and didn't find any obvious ones. But in a future PR we should definitely add this. This PR is pretty much bare-metal to enable experimentation.

devito/function.py
mloubout: My bad, hidden line, saw this with `_indices`. So I guess this PR is shape only, not staggered FD yet?

mlange05: Correct. The symbolics will get their own (set of) PR(s).

FabioLuporini:
Right, go ahead and merge for me

mlange05:
Actually, please wait. There is a sanity check I should add to keep people from shooting themselves in the foot.

mlange05:
Ok, sanity check added. Good to go once green.

FabioLuporini:
Merged.

PR #411, opened 2017-11-13
Drop partial ordering for loop scheduling

FabioLuporini:


FabioLuporini:
Merged as approval given after offline discussion

PR #415, opened 2017-11-15
Fix seismic weblinks

mlange05:
They have not been updated in the latest tutorial push - so this PR fixes that. It also enables the auto-docs for the `yask` sub-package.


FabioLuporini:
yes

PR #421, opened 2017-11-20
Preliminaries before finalising new data dependence analysis engine

FabioLuporini:
Just reorganising stuff around. This should be a simple green Travis, as no new functionalities are actually added. I'm only pulling this out of the final PR for a cleaner/simpler diff


FabioLuporini:
Thanks, merging as Travis is now green...

PR #422, opened 2017-11-20
ir: Introduce new data dependence analysis engine

FabioLuporini:
~*** This should go in after #421, which will make the diff much cleaner ***~
EDIT: ready for review

This PR exploits the previously merged #401 and #406 PRs to perform advanced data dependence analysis in Devito. Some details about this had already been provided in the descriptions of these two PRs.

There are three major things going on here:
* Reimplementation of cluster fusion, based on vector distance analysis. This is what will fix loop construction for staggered, @mlange05 . So this new piece of code **is** actually fixing bugs that weren't exposed with the currently available use cases. This happens in ``ir/clusters/algorithms.py``.
* Reimplementation of Iteration analysis; that is, the engine that detects, for example, parallel and sequential loops. This happens in ``ir/iet/analysis.py``. This is again based on vector distance analysis, which **vastly** improves the quality of the implementation, besides fixing (subtle) bugs not exposed by the current use cases.
* Introduce a significant number of tests checking loop scheduling. In particular, look at ``tests/test_operator.py::test_consistency_anti_dependences``

### DIFF
--- a/devito/ir/clusters/algorithms.py
+++ b/devito/ir/clusters/algorithms.py
@@ -1,55 +1,62 @@
-from collections import OrderedDict, namedtuple
+from collections import OrderedDict
 
-from devito.ir.support.basic import Scope
-from devito.ir.clusters.cluster import Cluster
+from devito.ir.support import Scope
+from devito.ir.clusters.cluster import PartialCluster, ClusterGroup
 from devito.ir.dfg import TemporariesGraph
-from devito.ir.support import Stencil
 from devito.symbolics import xreplace_indices
 from devito.types import Scalar
 from devito.tools import flatten
 
-__all__ = ['clusterize', 'merge']
+__all__ = ['clusterize', 'groupby']
 
 
-def merge(clusters):
+def groupby(clusters):
     """
-    Given an ordered collection of :class:`Cluster` objects, return a
-    (potentially) smaller sequence in which clusters with identical stencil
-    have been merged into a single :class:`Cluster`, as long as data
-    dependences are honored.
+    Given an ordered collection of :class:`PartialCluster` objects, return a
+    (potentially) smaller sequence in which dependence-free PartialClusters with
+    identical stencil have been squashed into a single PartialCluster.
     """
+    clusters = clusters.unfreeze()
 
-    def _merge(source, sink):
-        temporaries = OrderedDict(source.trace.items())
-        temporaries.update(OrderedDict([(k, v) for k, v in sink.trace.items()
-                                        if k not in temporaries]))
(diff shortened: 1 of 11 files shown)


devito/ir/clusters/algorithms.py
```diff
@@ -1,75 +1,182 @@
...
 
-def merge(clusters):
+def groupby(clusters):
     """
-    Given an ordered collection of :class:`Cluster` objects, return a
-    (potentially) smaller sequence in which clusters with identical stencil
-    have been merged into a single :class:`Cluster`.
+    Given an ordered collection of :class:`PartialCluster` objects, return a
+    (potentially) smaller sequence in which dependence-free PartialClusters with
+    identical stencil have been squashed into a single PartialCluster.
```
mlange05: :class:`PartialCluster` for pretty-docs.

devito/ir/clusters/algorithms.py
```diff
@@ -1,75 +1,182 @@
...
+            # Check all data dependences relevant for cluster fusion
+            scope = Scope(exprs=candidate.exprs + c.exprs)
+            anti = scope.d_anti.carried() - scope.d_anti.increment
+            flow = scope.d_flow - (scope.d_flow.inplace() + scope.d_flow.increment)
+            funcs = [i.function for i in anti]
+            if candidate.stencil == c.stencil and\
+                    all(is_local(i, candidate, c, clusters) for i in funcs):
+                # /c/ will be fused into /candidate/. All fusion-induced anti
+                # dependences are eliminated through so called "index bumping and
+                # array contraction", which transforms array accesses into scalars
```
mlange05: Out of interest, is there a book or reference for all the technical compiler lingo that our backends now use?

devito/ir/clusters/algorithms.py
FabioLuporini: fixing in the PR based on top of this (#423 ), thanks

devito/ir/clusters/algorithms.py
FabioLuporini: this is pretty much standard compiler jargon, you can find it in lecture notes (see for example: http://www.cs.cmu.edu/afs/cs/academic/class/15745-s06/web/handouts/15.pdf, http://www.cs.utah.edu/~mhall/cc/Utah-L1.pdf, ...) or in any textbook, such as the "Dragon Book " (https://en.wikipedia.org/wiki/Compilers:_Principles,_Techniques,_and_Tools)

I think/hope I haven't invented anything new except for things that are special to us

FabioLuporini:
Sweet, merging...

PR #423, opened 2017-11-20
Timeslot avoidance optimisation

FabioLuporini:
This should go in after #422 

Use data dependence analysis to reuse buffered timeslots whenever possible, thus reducing the working set. In practice, this depends on things such as the chosen timestepping scheme

Trivial example.
Before:
```
u[t+1, ..] = f(u[t, ...], u[t-1, ...])
```
With this PR, in some cases :
```
u[t+1, ..] = f(u[t, ...], u[t+1, ...])
```

### DIFF
--- a/devito/ir/iet/visitors.py
+++ b/devito/ir/iet/visitors.py
@@ -18,7 +18,7 @@ from devito.dimension import LoweredDimension
 from devito.exceptions import VisitorException
 from devito.ir.iet.nodes import Iteration, Node, UnboundedIndex
 from devito.types import Scalar
-from devito.tools import as_tuple, filter_ordered, filter_sorted, flatten, ctypes_to_C
+from devito.tools import as_tuple, filter_sorted, flatten, ctypes_to_C
 
 
 __all__ = ['FindNodes', 'FindSections', 'FindSymbols', 'MapExpressions',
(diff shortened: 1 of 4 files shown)


FabioLuporini:
@navjotk any suggestion on how to fix ``test_checkpointing`` here? we briefly discusses this offline, and probably PR #424 is related to this. I'm hoping someone can take a look at this

navjotk:
Yes, indeed it is. In a way, #424 introduces a tighter, simpler test to make sure the forward and reverse indices are aligned throughout a checkpointed forward-reverse pass. Try to use that to figure out why the other checkpointing tests failed?

FabioLuporini:
I wonder if you can rather take a look at it? because I have a feeling it'll be much simpler for you than for me

FabioLuporini:
The latest issue with this is that I currently cannot predict whether all of the array entries computed through modulo-buffered iteration will be needed by a subsequent Operator. Because if that's the case, I shouldn't trigger the optimisation. On the other hand, if needed is only the very last array entry written by an Operator through modulo-buffered iteration, then we're in business as long as we have a mechanism to retrieve "the last written-to entry" (issue #69 ).

@mlange05 @navjotk :
1: do you understand the issue above?
2: comments?

navjotk:
Ah I see you now. Yes, I agree that we could figure out a `save=1`, `save=2`, `save=3` option instead of `save=None`.

FabioLuporini:
good question. Perhaps we could simply add an Operator option indicating that "only the very last entry of each written arrays are necessary". Then this option would be a necessary condition for the TAO; it's not sufficient because data dependence analysis would still be necessary under the hood

FabioLuporini:
Closing because this is not gonna see the light any time soon

PR #424, opened 2017-11-21
Test index alignment for checkpointing

navjotk:
This PR primarily adds a test for correct index alignment in `test_checkpointing.py`. 

It also: 
- ~adds a check for missing argument values~
- ~adds a test for the above~

### DIFF
--- a/tests/test_checkpointing.py
+++ b/tests/test_checkpointing.py
@@ -13,6 +13,9 @@ from devito import Grid, TimeFunction, Operator, Backward, Function, Eq
 @pytest.mark.parametrize('time_order', [2])
 @pytest.mark.parametrize('shape', [(70, 80), (50, 50, 50)])
 def test_forward_with_breaks(shape, time_order, space_order):
+    """ Test running forward in one go and "with breaks"
+    and ensure they produce the same result
+    """
     spacing = tuple([15.0 for _ in shape])
     tn = 500.
     example = CheckpointingExample(shape, spacing, tn, time_order, space_order)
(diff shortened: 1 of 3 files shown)


FabioLuporini:
looks good to me

navjotk:
@FabioLuporini Can you take a look at the travis failure here? Seems to be failing on flake errors in code that I didn't even write - seems like a version issue for flake.

FabioLuporini:
maybe it's pulling a newer flake release, which added E722 and E741? can you update your flake locally and try running it again and see what happens? If you can reproduce the issue locally, feel free to add E722 and E741 to ``setup.cfg``

tests/test_checkpointing.py
```diff
@@ -60,3 +61,64 @@ def test_checkpointed_gradient_test(shape, time_order, space_order):
     m0, dm = example.initial_estimate()
     gradient, rec_data = example.gradient(m0)
     example.verify(m0, gradient, rec_data, dm)
+
+
+@skipif_yask
+def test_index_alignment(const):
```
FabioLuporini: docstring

mlange05:
It looks like this test should be split into at least 3 or 4 independent ones. Is there a reason they are all crammed into the same test function?

navjotk:
I think it's quite a simple test which has multiple asserts on the way, to increase coverage. Is there a reason you think this should be split?

FabioLuporini:
> Is there a reason you think this should be split?

It makes debugging much simpler. If it's feasible to split it, that'd be useful I think

navjotk:
As I said, it's a simple test. I don't see how it can be split.

FabioLuporini:
OK, I'm convinced and happy, this is good to go for me. @navjotk feel free to merge

PR #426, opened 2017-11-21
Bug in arguments

navjotk:
When `save=True`, passed time values need to respect the shape of the time dimension in memory. That doesn't seem to be happening. 

I added here a test to highlight the issue.

### DIFF
--- a/tests/test_operator.py
+++ b/tests/test_operator.py
@@ -471,6 +471,17 @@ class TestArguments(object):
         assert(op_arguments[time.start_name] == 0)
         assert(op_arguments[time.end_name] == nt - 2)
 
+    def test_argument_time_save_greater(self, const):
+        nt = 10
+        grid = Grid(shape=(3, 5))
+        time = grid.time_dim
+        t = grid.stepping_dim
+        x, y = grid.dimensions
+        u = TimeFunction(name='u', grid=grid, save=True, time_dim=nt)
+        fwd_eqn = Eq(u.indexed[time+1, x, y], u.indexed[time, x, y] + 1.*const)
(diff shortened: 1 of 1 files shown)


FabioLuporini:
can you reformulate it to trigger an xfail?

navjotk:
Why is it better to mark it as xfail?

FabioLuporini:
if you're trying to highlight the fact that we should support something, but this doesn't actually work, then one should write the test as if it were supposed to pass and then mark it as an xfail; this allows other people to immediately understand that "this is an expected feature/something that should work, but we're not there yet"

more details here: https://docs.pytest.org/en/latest/skipping.html

tests/test_operator.py
```diff
@@ -471,6 +471,17 @@ def test_argument_derivation_order(self, nt=100):
...
+    def test_argument_time_save_greater(self, const):
+        nt = 10
+        grid = Grid(shape=(3, 5))
+        time = grid.time_dim
+        t = grid.stepping_dim
+        x, y = grid.dimensions
+        u = TimeFunction(name='u', grid=grid, save=True, time_dim=nt)
+        fwd_eqn = Eq(u.indexed[time+1, x, y], u.indexed[time, x, y] + 1.*const)
+        fwd_op = Operator(fwd_eqn)
+        fwd_op(time=nt+10, constant=1)
```
mlange05: Can I ask what is supposed to be happening here? From my reading, the user is forcing a memory violation here. SO how would we handle this?

navjotk: By detecting this and throwing an exception.

navjotk:
This needs to be rethought in light of the latest arguments implementation.

PR #427, opened 2017-11-21
Operator bug

navjotk:
This was originally part of #424 but I pulled it out because it seems to trigger an unrelated bug.

This PR passes all tests without OpenMP, but fails to compile with OpenMP.

~I believe this might be related to [this bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=35158)~

### DIFF
--- a/devito/operator.py
+++ b/devito/operator.py
@@ -203,6 +203,11 @@ class Operator(Callable):
 
         arguments = self._default_args()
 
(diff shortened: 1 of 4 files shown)


FabioLuporini:
I'm not sure I understand what's going on here and why is this related to openmp

also, was it on purpose to have both ``const`` and ``constant`` in the test?

PR #428, opened 2017-11-22
autotuning: Fixup after dep analysis restructuring

FabioLuporini:


FabioLuporini:
@navjotk can you have a quick look at this? If we merge this, we can run stuff for the GMD paper

devito/core/autotuning.py
```diff
@@ -30,23 +30,23 @@ def autotune(operator, arguments, tunable):
     iterations = FindNodes(Iteration).visit(operator.body)
     dim_mapper = {i.dim.name: i.dim for i in iterations}
 
-    # Shrink the iteration space of sequential dimensions so that auto-tuner
-    # runs take a negligible amount of time
-    sequentials = [i for i in iterations if i.is_Sequential]
-    if len(sequentials) == 0:
+    # Shrink the iteration space of time-stepping dimension so that auto-tuner
+    # runs will finish quickly
+    steppers = [i for i in iterations if i.dim.is_Time or i.dim.is_Stepping]
```
navjotk: Is the `is_Stepping` required? `is_Time` should be `True` for `SteppingDimensions` as well since they "inherit" (not in the literal OOP sense) these values from the parent dimension which, in our case, is always a time dimension.

FabioLuporini: but it's not, not sure why. @mlange05 ?

mlange05: > is_Time should be True for SteppingDimensions as well since they "inherit" ...

In general this statement is false, since a user can create arbitrary stepping dimensions that are not related to `time`. In practice I'm not aware of any such case in the test base, because we mostly use `t`, for which the above holds. 

> but it's not, not sure why.

Example?

FabioLuporini: > Example?

oh I simply meant that ``SteppingDimension`` is not a subclass of ``TimeDimension``

FabioLuporini:
Merged

devito/core/autotuning.py
navjotk: Let me clarify:
> oh I simply meant that SteppingDimension is not a subclass of TimeDimension

As I mention above, `SteppingDimension` is not a subclass of `TimeDimension` and that's fine. I was referring to [this](https://github.com/opesci/devito/blob/master/devito/dimension.py#L122) piece of code that copies the `is_Time`/`is_Space` identifiers from the parent dimension. Why is `is_Stepping` required here, when, `is_Time` should be `True` for both `t` and `time`?

> since a user can create arbitrary stepping dimensions 

If I understand the logic in this line of code that I highlighted above in autotuning, it is looking for `time` and `t`, and hence checking for `is_Time` is more semantically correct than checking for `is_Stepping`.

PR #429, opened 2017-11-23
Silence checkpointing

mlange05:
Adds a decorator to temporarily change log levels and uses it to silence checkpointing output. Also updates `flake8` rules.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,7 +84,7 @@ script:
       python examples/seismic/acoustic/acoustic_example.py --full;
       python examples/seismic/acoustic/acoustic_example.py --constant --full;
       python examples/seismic/acoustic/gradient_example.py;
-      DEVITO_LOGGING=WARNING python examples/checkpointing/checkpointing_example.py;
+      python examples/checkpointing/checkpointing_example.py;
     fi
 
   # Test tutorial notebooks for the website using nbval
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ env:
   global:
   - ENCRYPTION_LABEL: "1909213e8aea"
   - COMMIT_AUTHOR_EMAIL: "pandolfovince@gmail.com"
-  - DEVITO_LOGGING: "DEBUG"
+  - DEVITO_LOGGING: "INFO"
 
 matrix:
   include:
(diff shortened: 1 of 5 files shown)


navjotk:
Shouldn't be required if raising the log-level to WARNING, no?

devito/tools.py
```diff
@@ -307,3 +307,22 @@ def infer_cpu():
     if platform not in configuration._accepted['platform']:
         platform = default_platform
     return isa, platform
+
+
+class silencio(object):
```
FabioLuporini: in ``tools.py`` or ``logger.py`` ?

.travis.yml
```diff
@@ -8,7 +8,7 @@ env:
   global:
   - ENCRYPTION_LABEL: "1909213e8aea"
   - COMMIT_AUTHOR_EMAIL: "pandolfovince@gmail.com"
-  - DEVITO_LOGGING: "DEBUG"
+  - DEVITO_LOGGING: "INFO"
```
navjotk: INFO or WARNING?

FabioLuporini:
In fact, thinking again about it, why not just stick to WARNING everywhere ? the ``silencio`` util may still be useful though for application code.

mloubout:
>the silencio util may still be useful though for application code.

isn't what "ERROR" does?

FabioLuporini:
> isn't what "ERROR" does?

I'm not sure I understand what you mean by that. If you're using any of the Devito high level wrappers, or you simply want to silence Devito in a particular region, then you can use this utility

.travis.yml
mlange05: I strongly believe it should be `INFO`, since that mirrors the default user experience.

FabioLuporini:
> I strongly believe it should be INFO, since that mirrors the default user experience.

OK, I really don't have strong thoughts in merit, so I'm happy to switch to INFO. I think the major problems in the end were:
- useless DSE/DLE messages (gone time ago)
- excessive pollution from various tests (fixed through ``silencio``)

navjotk:
The next biggest source of log clutter after this are the compiler warnings.

FabioLuporini:
All good, merging...

FabioLuporini:
Merged.

PR #431, opened 2017-11-24
add destructors for Function, TimeFunction and SparseFunction

philippwitte:
This seems to solve the memory leak issue for me.


mloubout:
Travis only fails because of the `test_caching` one, not sure what it is supposed to test to fix.

FabioLuporini:
oh, and regardless, probably you also want to call the superclass destructor

FabioLuporini:
Closing as the leakage was fixed by #433

PR #433, opened 2017-11-24
IET: Drop `Analysis` cache, to fix memory leak

mlange05:
It's a class-level cache of sympy expressions, that otherwise keeps our functions alive.

### DIFF
--- a/devito/ir/iet/analysis.py
+++ b/devito/ir/iet/analysis.py
@@ -18,17 +18,6 @@ __all__ = ['analyze_iterations']
 
 class Analysis(object):
 
-    _cache = {}
-
-    def __new__(cls, iet):
-        if iet in Analysis._cache:
-            return Analysis._cache[iet]
-        else:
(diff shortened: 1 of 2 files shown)


mlange05:
This is still missing a test. I'll add it soon.

mlange05:
Test added. Should be gtg once travis is happy.

mloubout:
Does ir cover the part @philippwitte PR was aiming to fix, which is even if the cache is cleaned, the `Cmemory` isn't.

FabioLuporini:
looks good to me

mlange05:
> Does ir cover the part @philippwitte PR was aiming to fix, which is even if the cache is cleaned, the Cmemory isn't.

It does. The problem was that the cache wasn't really cleared, because there were still legitimate references to `u` dangling around, so the entire `u` was never freed.

mlange05:
Ok, going in....

PR #441, opened 2017-11-28
Benchmark makeover

mlange05:
This merge re-vamps the CLI argument configuration of the `benchmark.py` script using the vastly superior and more pythonic `click` package. :snake: The primary intention here is that the four execution modes are now also available as plain old Python functions, as well as CLI commands, so that we may use almost the same interface from cmd as from an IPython notebook. The secondary motivation is that this is just soooo much cleaner than manual argument hackery.

Please note that this should bring no real functionality changes.


environment.yml
```diff
@@ -20,6 +20,7 @@ dependencies:
   - distributed
   - scikit-image
   - cython
+  - click
```
FabioLuporini: how quickly is the ``click`` API evolving? would it make sense to stick to a specific version?

mlange05: Not certain, but I think the authors are very aware of the impact of API changes; so I think it's low risk. In general I would stick to "latest release" until things break, but that IMHO.

mlange05:
> Can we reuse some of this stuff for the arguments parsing in examples/seismic/{acoustic,tti} ?

:+1: We definitely should, but in another PR, I think.

mlange05:
That being said, `click` has one caveat in that it needs input tuples to be of a defined size, since multi-options and multi-value options become ambiguous due to it's improved parsing. We, however, rely on the `shape`/`dimension` tuple size to infer the problem dimension in some tests, ie. 2D or 3D. I haven't found a clean way around this yet (believe me I tried... :frowning: ).

environment.yml
FabioLuporini: no, OK, that's fine

mlange05:
OK, I pushed a small set of fixes and docstring improvements, inclduing the separation of the CLI wrappers from the argument methods. Should be good to go now, subject to travis, of course. More fixes and tweaks might be required, but I'll leave these for another PR.

FabioLuporini:
still looking good

mlange05:
Hey ho, lets' go...

PR #443, opened 2017-11-28
Update TimeFunction API to generalise save

navjotk:
With this PR, we change the `TimeFunction` API so that instead of using `save=True` in combination with `time_dim=nt`, we just use `save=nt` (and `save=None` for the `save=False` case). All tests and examples have been updated to use the new API which looks cleaner. 

`time_dim` is now changed as a means of setting the symbolic `Dimension` object used to represent the `TimeFunction`'s time dimension internally. This is currently being debated in slack.

Also included are two unrelated bug-fixes from @FabioLuporini :
1. `repr(Operator)` now works
2. Loop scheduling bug when `time` and `t` are used in the same expression - we now generate the correct loop nest. Test included for this.

### DIFF
--- a/devito/function.py
+++ b/devito/function.py
@@ -2,6 +2,7 @@ import numpy as np
 import sympy
 from collections import OrderedDict
 from functools import partial
+from psutil import virtual_memory
 
 from devito.parameters import configuration
 from devito.logger import debug, error, warning
@@ -374,7 +375,11 @@ class TimeFunction(Function):
                 if not isinstance(self.save, int):
                     raise ValueError("save must be an int indicating the number of " +
                                      "timesteps to be saved (is %s)" % type(self.save))
-                
+                available_mem = virtual_memory().available
+
+                if np.dtype(self.dtype).itemsize * self.save > available_mem:
+                    warning("Trying to allocate more memory for symbol %s " % self.name +
+                            "than available on physical device, this will start swapping")
                 self.time_size = self.save
             else:
                 self.time_size = self.time_order + 1
--- a/devito/function.py
+++ b/devito/function.py
@@ -7,7 +7,7 @@ from devito.parameters import configuration
 from devito.logger import debug, error, warning
 from devito.memory import CMemory, first_touch
 from devito.cgen_utils import INT, FLOAT
-from devito.dimension import Dimension
+from devito.dimension import Dimension, TimeDimension
 from devito.arguments import ConstantArgProvider, TensorFunctionArgProvider
 from devito.types import SymbolicFunction, AbstractSymbol
 from devito.finite_difference import (centered, cross_derivative,
@@ -326,9 +326,11 @@ class TimeFunction(Function):
     :param staggered: (Optional) tuple containing staggering offsets.
     :param dtype: (Optional) data type of the buffered data
     :param save: Save the intermediate results to the data buffer. Defaults
-                 to `False`, indicating the use of alternating buffers.
-    :param time_dim: Size of the time dimension that dictates the leading
-                     dimension of the data buffer if :param save: is True.
+                 to `None`, indicating the use of alternating buffers. If
+                 intermediate results are required, the value of save must
+                 be set to the required size of the time dimension.
+    :param time_dim: The :class:`Dimension` object to use to represent time in this
+                     symbol. Defaults to the time dimension provided by the :class:`Grid`.
     :param time_order: Order of the time discretization which affects the
                        final size of the leading time dimension of the
                        data buffer.
@@ -362,22 +364,18 @@ class TimeFunction(Function):
         if not self._cached():
             super(TimeFunction, self).__init__(*args, **kwargs)
             self.time_dim = kwargs.get('time_dim', None)
+            if self.time_dim is not None and not isinstance(self.time_dim, TimeDimension):
+                raise ValueError("time_dim must be a TimeDimension, not %s" %
+                                 type(self.time_dim))
             self.time_order = kwargs.get('time_order', 1)
-            self.save = kwargs.get('save', False)
+            self.save = kwargs.get('save', None)
 
-            if not self.save:
-                if self.time_dim is not None:
-                    warning('Explicit time dimension size (time_dim) found for '
-                            'TimeFunction symbol %s, despite \nusing a stepping time '
-                            'dimension (save=False). This value will be ignored!'
-                            % self.name)
-                self.time_dim = self.time_order + 1
-                self.indices[0].modulo = self.time_dim
+            if self.save is not None:
+                assert(isinstance(self.save, int))
+                self.time_size = self.save
             else:
-                if self.time_dim is None:
-                    error('Time dimension (time_dim) is required'
-                          'to save intermediate data with save=True')
-                    raise ValueError("Unknown time dimensions")
+                self.time_size = self.time_order + 1
+                self.indices[0].modulo = self.time_size
 
     @property
     def shape_data(self):
@@ -385,7 +383,7 @@ class TimeFunction(Function):
         Full allocated shape of the data associated with this :class:`TimeFunction`.
         """
         if self.save:
-            tsize = self.time_dim - self.staggered[0]
+            tsize = self.time_size - self.staggered[0]
         else:
             tsize = self.time_order + 1
         shape_domain = tuple(i - s for i, s in zip(self.shape_domain,
@@ -406,12 +404,20 @@ class TimeFunction(Function):
         """
         save = kwargs.get('save', None)
         grid = kwargs.get('grid', None)
+        time_dim = kwargs.get('time_dim', None)
+
+        if time_dim is not None:
+            assert(isinstance(time_dim, TimeDimension))
 
         if grid is None:
             error('TimeFunction objects require a grid parameter.')
             raise ValueError('No grid provided for TimeFunction.')
 
-        tidx = grid.time_dim if save else grid.stepping_dim
+        if time_dim is not None:
+            tidx = time_dim
+        else:
+            tidx = grid.time_dim if save else grid.stepping_dim
+
         _indices = Function._indices(**kwargs)
         return tuple([tidx] + list(_indices))
 
--- a/devito/function.py
+++ b/devito/function.py
@@ -371,7 +371,10 @@ class TimeFunction(Function):
             self.save = kwargs.get('save', None)
 
             if self.save is not None:
-                assert(isinstance(self.save, int))
+                if not isinstance(self.save, int):
+                    raise ValueError("save must be an int indicating the number of " +
+                                     "timesteps to be saved (is %s)" % type(self.save))
+                
                 self.time_size = self.save
             else:
                 self.time_size = self.time_order + 1
--- a/devito/function.py
+++ b/devito/function.py
@@ -327,7 +327,7 @@ class TimeFunction(Function):
     :param staggered: (Optional) tuple containing staggering offsets.
     :param dtype: (Optional) data type of the buffered data
     :param save: Save the intermediate results to the data buffer. Defaults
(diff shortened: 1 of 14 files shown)


devito/ir/iet/nodes.py
```diff
@@ -465,7 +465,8 @@ def __init__(self, name, body, retval, parameters=None, prefix=('static', 'inlin
         self.parameters = as_tuple(args)
 
     def __repr__(self):
-        parameters = ",".join([c.dtype_to_ctype(i.dtype) for i in self.parameters])
+        parameters = ",".join(['void*' if i.is_PtrArgument else c.dtype_to_ctype(i.dtype)
```
mlange05: Presumably, `dtype_to_ctype` is missing pointer types completely? Should we wrap it then in a separate utility to add pointer types, instead of relying on individual special-casing when it is used?

devito/function.py
```diff
@@ -326,9 +327,11 @@ class TimeFunction(Function):
     :param staggered: (Optional) tuple containing staggering offsets.
     :param dtype: (Optional) data type of the buffered data
     :param save: Save the intermediate results to the data buffer. Defaults
-                 to `False`, indicating the use of alternating buffers.
-    :param time_dim: Size of the time dimension that dictates the leading
-                     dimension of the data buffer if :param save: is True.
+                 to `None`, indicating the use of alternating buffers. If
```
mlange05: Nitpicking: Should use double ` around None to get teletype typesetting in sphinx (assuming that's the goal).

tests/test_operator.py
```diff
@@ -845,7 +865,7 @@ def test_explicit_run(self):
         time_dim = 6
         grid = Grid(shape=(11, 11))
         a = TimeFunction(name='a', grid=grid, time_order=1,
-                         time_dim=time_dim, save=True)
+                         save=time_dim)
```
mlange05: Undo indent?

tests/test_timestepping.py
```diff
@@ -15,28 +15,28 @@ def grid(shape=(11, 11)):
 def a(grid):
     """Forward time data object, unrolled (save=True)"""
     return TimeFunction(name='a', grid=grid, time_order=1,
-                        time_dim=6, save=True)
+                        save=6)
```
mlange05: Undo indent? (same for fixtures `b` and `d` below)

tests/test_save.py
```diff
@@ -27,14 +27,14 @@ def run_simulation(save=False, dx=0.01, dy=0.01, a=0.5, timesteps=100):
...
-        time_order=1, space_order=2, save=save
+        name='u', grid=grid, save=timesteps, initializer=initializer,
+        time_order=1, space_order=2
     )
 
     eqn = Eq(u.dt, a * (u.dx2 + u.dy2))
     stencil = solve(eqn, u.forward)[0]
     op = Operator(Eq(u.forward, stencil), time_axis=Forward)
-    op.apply(time=timesteps, dt=dt)
+    op.apply(time=timesteps-1, dt=dt)
```
mlange05: This is odd. :thinking: What changed to cause this, and why do you change it?

devito/function.py
```diff
@@ -406,12 +412,20 @@ def _indices(cls, **kwargs):
         """
         save = kwargs.get('save', None)
         grid = kwargs.get('grid', None)
+        time_dim = kwargs.get('time_dim', None)
+
+        if time_dim is not None:
+            assert(isinstance(time_dim, TimeDimension))
```
mlange05: Nitpicking: Could combine with below conditional to save lines and make things easier accessible.

devito/function.py
```diff
@@ -362,30 +365,33 @@ def __init__(self, *args, **kwargs):
...
-                            'TimeFunction symbol %s, despite \nusing a stepping time '
-                            'dimension (save=False). This value will be ignored!'
-                            % self.name)
-                self.time_dim = self.time_order + 1
-                self.indices[0].modulo = self.time_dim
+            self.save = kwargs.get('save', None)
+
+            if self.save is not None:
+                if not isinstance(self.save, int):
+                    raise ValueError("save must be an int indicating the number of " +
```
mloubout: Can't we instead fall back to full time size instead?

devito/function.py
```diff
@@ -362,30 +365,33 @@ def __init__(self, *args, **kwargs):
...
+                                     "timesteps to be saved (is %s)" % type(self.save))
+                available_mem = virtual_memory().available
+
+                if np.dtype(self.dtype).itemsize * self.save > available_mem:
+                    warning("Trying to allocate more memory for symbol %s " % self.name +
```
mloubout: nice

devito/function.py
mloubout: Isn't this already caught by the constructor?

FabioLuporini: > Isn't this already caught by the constructor?

indeed

devito/function.py
```diff
@@ -406,12 +412,20 @@ def _indices(cls, **kwargs):
...
+
+        if time_dim is not None:
+            assert(isinstance(time_dim, TimeDimension))
 
         if grid is None:
             error('TimeFunction objects require a grid parameter.')
             raise ValueError('No grid provided for TimeFunction.')
 
-        tidx = grid.time_dim if save else grid.stepping_dim
+        if time_dim is not None:
```
FabioLuporini: AKA:
```
if time_dim is None:
    time_dim = grid.time_dim if save else grid.stepping_dim
```
Then use ``time_dim`` below

devito/function.py
```diff
@@ -362,30 +365,33 @@ def __init__(self, *args, **kwargs):
...
+
+            if self.save is not None:
+                if not isinstance(self.save, int):
+                    raise ValueError("save must be an int indicating the number of " +
+                                     "timesteps to be saved (is %s)" % type(self.save))
+                available_mem = virtual_memory().available
+
+                if np.dtype(self.dtype).itemsize * self.save > available_mem:
+                    warning("Trying to allocate more memory for symbol %s " % self.name +
+                            "than available on physical device, this will start swapping")
```
FabioLuporini: crazy thought that we probably do not wanna do, but maybe it's worth talking about it... 
How about we run ``clean_cache()`` here? so that "this might start swapping", rather than "this will start swapping". Might help with outer loops allocating new wavefields at each iteration, whose references are kept by the sympy cache.

If we did that, we should perform this check regardless of ``self.save``

devito/function.py
```diff
@@ -362,30 +365,33 @@ def __init__(self, *args, **kwargs):
...
+                self.time_size = self.save
             else:
-                if self.time_dim is None:
-                    error('Time dimension (time_dim) is required'
-                          'to save intermediate data with save=True')
-                    raise ValueError("Unknown time dimensions")
+                self.time_size = self.time_order + 1
+                self.indices[0].modulo = self.time_size
```
FabioLuporini: Is this changing the state of a ``Dimension``? If so, this is probably super dangerous. I know that's what we had before too, I just never noticed it.

devito/function.py
navjotk: Actually, the first line of the constructor calls the `super` constructor which [calls](https://github.com/opesci/devito/blob/5fd034fa2c4a7254ff058b6ff634b1e9f64dfdcc/devito/function.py#L141) this so `_indices()` is called before our own constructor and this was thus required. The check in the constructor is actually a second-level check, just in case.

tests/test_save.py
navjotk: I'll be honest, I don't understand why this was required, but the test was failing without it. Maybe I'll spend more time.

devito/function.py
navjotk: This is the "full time size". Think of this as what `time_dim` was doing before - I am asking for the size of the time dimension here.

devito/function.py
navjotk: Oh this is not even the biggest problematic case of changing state on "global" dimension objects - the time reversal in the Operator is more dangerous than this. Outside the scope of this PR though.

devito/function.py
navjotk: `clean_cache()` is potentially a heavy operation invoking the GC - not something I'll be comfortable doing without making it clear to the user (who might be multiple application-abstraction layers above devito). Plus, it is required to cover for a "bug" in devito.

navjotk:
> can we rename save as nb_snap or nb_save or something like that, naming is a bit odd to me (can ignore)

I don't understand why `nb_snap` or `nb_save` is better? It makes less sense to me than `save` does. 

> We need at least some docstring saying how the subsampling is done (uniform, interpolation, nearest time step, ...)

When I introduce subsampling, I will keep that in mind. This is just clearing up the API before introducing subsampling.

tests/test_save.py
mlange05: Yes, it would be good to understand why this is necessary. What happens without it?

devito/ir/iet/nodes.py
navjotk: Can we consider this for a future PR?

tests/test_save.py
navjotk: The bug highlighted in https://github.com/opesci/devito/pull/426 is triggered, i.e. it seems that we are trying to run for more timesteps than the size of the time dimension of the function with `save=True`.

devito/function.py
navjotk: Fixed

devito/function.py
navjotk: > Could combine with below conditional to save lines

Fixed

tests/test_operator.py
navjotk: Fixed

devito/function.py
navjotk: Fixed

tests/test_timestepping.py
navjotk: Fixed

devito/ir/iet/nodes.py
FabioLuporini: yes

devito/function.py
FabioLuporini: Agreed. One last comment: the change of state in ``Operator`` can probably be easily avoided (we can just have an instance-level flag and add a parameter to ``Iteration`` to allow describing things such as the traversal direction etc. On the other hand, I'm unsure how I would fix this.

tests/test_save.py
navjotk: Ok, to be clear. The symptom of the bug in #426 is a segfault on my macbook. When I tried running this test on master, it also segfaulted and making this change made the segfault go away. 
My hypothesis now is that this test was only run on machines with more relaxed memory-access policies and that this test was always "broken" in that sense. 
Here's my best-guess for reasons: prior to the introduction of checkpointing, calling `Operator.apply(nt)` would run the operator for `nt-time_order` timesteps so allocating `nt` elements in `TimeFunction`s along the time dimension would work fine with no illegal memory accesses. During the checkpointing work, I changed the behaviour so `Operator.apply(nt)` runs `nt` timesteps to make things more transparent outside the API. I think the cleanest way to handle this would be to allocate `save + time_order` elements in the time dimension but I'm wary of allocating more memory than the user expects, especially since we will typically be running in memory-constrained environments.

FabioLuporini: > I'm wary of allocating more memory than the user expect

I agree this is not on the table. Good spot for the rest.

mlange05: I find this answer wholly unsatisfying. #426 forces an explicit overrun in the time dimension, while this case merely tries to use the same size as used for allocation. If this is deemed "wrong" and a bug, we probably have a long list of "bugs" in the test base, since this is pretty much default behaviour, no? Simply glossing over that fact with a simple `-1` sets a dangerous precedent, especially if we are chasing sporadic bugs that may well be due to memory corruption.

PR #444, opened 2017-11-29
Introduce Doctr for docs deployment

### LINKED ISSUE #436: More reliable docs deployment
We currently get occasional `master` breakages due to erroneous docstrings, which are only caught once we deploy to master. The reason is that the current script build/deployment script for auto-documentation is only run on `master` to avoid race conditions. To fix this I think we need to separate the docs build from the deployment to github pages.

I believe the best way forward to handling this more effectively is to use the following package, instead of the hand-crafted script:
https://github.com/drdoctr/doctr

FabioLuporini:
Should fix master breakages due to our deployment mechanisms.
Travis seems green, but this **definitely** needs an in-depth review

fixes #436


.travis.yml
```diff
@@ -90,7 +88,15 @@ script:
...
-  # Docs generation and deployment
+  # Docs generation
   - sphinx-apidoc -f -o docs/ examples
   - sphinx-apidoc -f -o docs/ devito devito/yask/*
-  - if [[ $DEVITO_ARCH == 'gcc-5' ]]; then ./docs/deploy.sh; fi
+  # Docs deployment only once from a single build
+  - if [[ $DEVITO_ARCH == 'gcc-5' ]]; then
+      set -e;
+      pip install doctr;
+      cd docs;
```
mlange05: Should we move the `cd docs; make html; cd ..` out of the conditional clause and run it on all builders? This might enable us to capture YASK-related sphinx issues and add the backedn docs back again.

.travis.yml
```diff
@@ -90,7 +88,15 @@ script:
...
+      make html;
+      cd ..;
+      doctr deploy . --built-docs docs --key-path .deploy_key.enc;
```
mlange05: Does `doctr` know not to deploy from branches/PRs?

FabioLuporini: it automatically deploys to ``gh-pages`` if that's what you're asking?

.travis.yml
FabioLuporini: yes

.travis.yml
mlange05: I'm asking if every travis build run, including branches and PRs, is executing a push to the branch (dangerous and racy) or if `doctr` knows only to deploy if the branch we're on is `master`. I'm assuming it does, I'd just like to know that someone double-checked this.

.travis.yml
FabioLuporini: done

.travis.yml
FabioLuporini: Oh, I see. so I had a look at ``doctr deploy --help``. Amongst the things it says:
```
...
  --branch-whitelist [BRANCH [BRANCH ...]]
                        Branches to deploy from. Pass no arguments to not
                        build on any branch (typically used in conjunction
                        with --build-tags). Note that you can deploy from
                        every branch with --no-require-master.
  --no-require-master   Allow docs to be pushed from a branch other than
                        master
  --require-master      Inverse of "--no-require-master"
...
```
I'm not sure how to interpret this, as I cannot see what's the default option. Should I add ``--require-master`` ?

I couldn't find a definitive answer here either: https://github.com/drdoctr/doctr

I also went looking for what SymPy does: https://github.com/sympy/sympy/blob/master/.travis.yml#L285
they deploy conditionally on whether they're testing sphinx, but I cannot see any master-specific instruction

mlange05: Sounds to me like it does the right thing by default, so I'm happy to try this as is. Good to go from me.

FabioLuporini:
OK, let's see...

FabioLuporini:
Merged.

PR #446, opened 2017-11-30
Allocate extra buffers in the time dimension

navjotk:
Temporary workaround as agreed on slack


devito/function.py
```diff
@@ -376,11 +376,11 @@ def __init__(self, *args, **kwargs):
                     raise ValueError("save must be an int indicating the number of " +
                                      "timesteps to be saved (is %s)" % type(self.save))
                 available_mem = virtual_memory().available
-
-                if np.dtype(self.dtype).itemsize * self.save > available_mem:
+                time_size = self.save + self.time_order + 1
```
FabioLuporini: can you add a big TODO here pls

PR #447, opened 2017-12-01
Fix offsets for explicitly passed arguments

mlange05:
**Problem:** The current implementation enforces a different indexing scheme for explicitly passed dimension arguments to the one used throughout the rest of the codebase (allocation, codegen, etc.) by adding stencil offsets to the provided values. Since we currently do not capture user-induced buffer overruns (see #426), this can lead to user-induced buffer overruns that cause sporadic failures in the test base if explicitly passed arguments do not account for this magic behaviour.

**This merge reverts this behaviour and enforces a consistent index addressing scheme throughout the code base.** It accordingly adjusts to the _currently(!) correct_ usage of explicitly provided values for the time dimension in all failing test cases (mostly YASK and checkpointing), and reverts the previous and unsuccessful attempts at fixing the sporadic failures due to the incurred buffer overruns.

**TLDR; More details on indexing schemes:**
Devito's current scheme is, for lack of a better name, "allocation-based". This means we allocate exactly the data shape provided by the user and all dimension indexing is done in terms of absolute values relative to this allocated shape. Since stencils usually have a "read-from" radius that is greater than the "write-to" region, this means that we address the desired iteration space for each loop/dimension as an absolute value that _includes the halo or buffer region_. For example, with a first order timestepping scheme that reads from `t` and writes to `t+1`, passing `t=2` means compute up to time index 2, not "compute two timesteps", and will compute a single timestep `t[1, :, :]` from the initial time index `t[0, :, :]`.

The alternative approach is what we currently call "domain-based". This is the approach taken by YASK and we are planning to switch to this scheme soon. It effectively defines the "domain" of each dimension as the "write-to" region of the data and indexes all access with respect to the first "written" data point, not the first allocated (read) data point. This approach may be more intuitive but requires a significant effort to implement from the current state of play to track and normalise dimensions accesses from data objects with different "read" radiuses. A separate issue will be filed to give more detail.

**Final note: Whichever scheme we may prefer, it is extremely important that this scheme is honoured throughout the entire codebase! Switching to a new scheme thus needs to be done wholesale, not in parts or for individual components, to avoid wasting more time on chasing sporadic phantom bugs!**

### DIFF
--- a/devito/arguments.py
+++ b/devito/arguments.py
@@ -335,17 +335,15 @@ def infer_dimension_values_tuple(value, rtargs, offsets=None):
         default to 0.
     """
     size_arg, start_arg, end_arg = rtargs
-    start_offset = 0 if offsets is None else offsets.get(start_arg.name, 0)
-    end_offset = 0 if offsets is None else offsets.get(end_arg.name, 0)
     if not isinstance(value, tuple):
         # scalar
-        value = (value, start_arg.default_value + start_offset, value + end_offset)
+        value = (value, start_arg.default_value, value)
     else:
         if len(value) == 2:
             # 2-tuple
             # Assume we've been passed a (start, end) tuple
             start, end = value
-            value = (end, start + start_offset, end + end_offset)
+            value = (end, start, end)
         elif len(value) != 3:
             raise InvalidArgument("Expected either a scalar value or a tuple(2/3)")
     return value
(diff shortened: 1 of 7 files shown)


FabioLuporini:
So I went through this a few times already and I cannot find anything odd; it makes sense. So I'd say good to go.

FabioLuporini:
Merging...

FabioLuporini:
Merged.

PR #448, opened 2017-12-04
Misc fixes and improvements

FabioLuporini:
Yet another consolidation PR

- Better, up-to-date docs
- Fix output reports (reported iteration counts might be a small quantity (O(1)) off)
- Add logic iteration access for modulo buffered dimension (see Slack)
- Refactor the ``codepy.Compiler`` hierarchy and (finally!) get rid of all annoying gcc-induced warnings. Travis' trace will now be much cleaner. 
- If the specific compiler version supports OpenMP 4.0, we now unconditionally add the appropriate ``-openmp`` flag so that ``#pragma simd omp`` are reconognised and vectorization properly triggered
- Drop support for Intel MIC (untested => broken; unused => useless; broken + useless => drop)

### DIFF
--- a/devito/__init__.py
+++ b/devito/__init__.py
@@ -56,13 +56,13 @@ configuration.add('backend', 'core', list(backends_registry),
 
 # Set the Instruction Set Architecture (ISA)
 ISAs = [None, 'cpp', 'avx', 'avx2', 'avx512', 'knc']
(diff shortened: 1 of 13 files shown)


FabioLuporini:
This needs some more work, closing for now

PR #450, opened 2017-12-06
Misc: cleanup, fixes, improvements, docs

FabioLuporini:
Yet another consolidation PR

- Improved (up-to-date) docs
- Fix output reports (reported iteration counts might be a small quantity (O(1)) off)
- Refactor the ``codepy.Compiler`` hierarchy and (finally!) get rid of all annoying gcc-induced warnings. Travis' trace will now be much cleaner. 
- If the specific compiler version supports OpenMP 4.0, we now unconditionally add the appropriate ``-openmp`` flag so that ``#pragma simd omp`` are reconognised and vectorization properly triggered
- Drop support for Intel MIC (untested => broken; unused => useless; broken + useless => drop)
- Fix a subtle bug by adding a specialised ``_hashable_content`` to ``Dimensions``
- Minor fixes in YASK

### DIFF
--- a/devito/function.py
+++ b/devito/function.py
@@ -47,6 +47,12 @@ class Constant(AbstractSymbol, ConstantArgProvider):
 
     """
     Symbol representing constant values in symbolic equations.
+
+    .. note::
+
+        The parameters must always be given as keyword arguments, since
+        SymPy uses ``*args`` to (re-)create the dimension arguments of the
+        symbolic function.
     """
 
     is_Constant = True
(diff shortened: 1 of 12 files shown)


devito/function.py
```diff
@@ -47,6 +47,12 @@ class Constant(AbstractSymbol, ConstantArgProvider):
 
     """
     Symbol representing constant values in symbolic equations.
+
+    .. note::
+
+        The parameters must always be given as keyword arguments, since
+        SymPy uses `*args` to (re-)create the dimension arguments of the
```
mlange05: If memory serves, double ` is needed to produce correct teletype in the docs. Also below...

FabioLuporini: fixed

FabioLuporini:
OK, I also pushed a couple of fixes that were making Travis rightfully angry. Will merge once Travis is green.

FabioLuporini:
OK, merging...

FabioLuporini:
Merged.

Issue #416, opened 2017-11-15
Argument derivation ignores value updates in Constant 


mlange05:
The following test demonstrates two examples of deriving the value of a `Constant` object during `Operator` invocation. The second one breaks, due to the value update happening after the operator was created.

```
    def test_derive_constant_value(self):
        """                                                                                                                                                                                
        Ensure that values for :class:`Constant` symbols are derived correctly.                                                                                                            
        """
        grid = Grid(shape=(5, 6))
        f = Function(name='f', grid=grid)
        a = Constant(name='a', value=3.)
        Operator(Eq(f, a))()
        assert np.allclose(f.data, 3.)

        g = Function(name='g', grid=grid)
        b = Constant(name='b')
        op = Operator(Eq(g, b))
        b.data = 4.
        op()
        assert np.allclose(g.data, 4.)
```

FabioLuporini:
fixed by #481

mloubout:
@fablup closing?

FabioLuporini:
yes.

PR #453, opened 2017-12-12
Arguments rewrite

navjotk:
Inviting review on this even though it doesn't pass all the tests yet. 

Edit: This PR now passes all tests


devito/ir/iet/visitors.py
```diff
@@ -233,6 +135,9 @@ def _args_decl(self, args):
         """Convert an iterable of :class:`Argument` into cgen format."""
         ret = []
         for i in args:
+            if not hasattr(i, "is_ScalarArgument"):
```
FabioLuporini: why not ``if i.is_ScalarArgument`` ?

devito/ir/iet/visitors.py
```diff
@@ -233,6 +135,9 @@ def _args_decl(self, args):
...
+                from devito.arguments import ArgumentVisitor
```
FabioLuporini: do we have a cyclic dependence if we move this to the top?

devito/ir/iet/visitors.py
```diff
@@ -246,6 +151,9 @@ def _args_cast(self, args):
         """Build cgen type casts for an iterable of :class:`Argument`."""
         ret = []
         for i in args:
+            if not hasattr(i, "is_ScalarArgument"):
```
FabioLuporini: same as above

devito/ir/iet/visitors.py
```diff
@@ -233,6 +135,9 @@ def _args_decl(self, args):
         """Convert an iterable of :class:`Argument` into cgen format."""
         ret = []
         for i in args:
+            if not hasattr(i, "is_ScalarArgument"):
+                from devito.arguments import ArgumentVisitor
+                i = ArgumentVisitor().visit(i)
```
FabioLuporini: overwriting the iteration variable !!

devito/ir/iet/visitors.py
```diff
@@ -246,6 +151,9 @@ def _args_cast(self, args):
         """Build cgen type casts for an iterable of :class:`Argument`."""
         ret = []
         for i in args:
+            if not hasattr(i, "is_ScalarArgument"):
+                from devito.arguments import ArgumentVisitor
+                i = ArgumentVisitor().visit(i)
```
FabioLuporini: overwriting the iteration variable !!

devito/ir/support/stencil.py
```diff
@@ -267,4 +267,15 @@ def __setitem__(self, key, val):
         super(Stencil, self).__setitem__(entry.dim, entry.ofs)
 
 
+def retrieve_offsets(stencils):
```
FabioLuporini: Can/should we make this a class-level ``Stencil`` method, such as ``union`` for example?

devito/operator.py
```diff
@@ -86,13 +87,12 @@ def __init__(self, expressions, **kwargs):
...
         self.input, self.output, self.dimensions = retrieve_symbols(expressions)
         stencils = make_stencils(expressions)
-        self.offsets = {d.end_name: v for d, v in retrieve_offsets(stencils).items()}
 
         # Set the direction of time acoording to the given TimeAxis
         for time in [d for d in self.dimensions if d.is_Time]:
             if not time.is_Stepping:
                 time.reverse = time_axis == Backward
-
+
```
FabioLuporini: probably flake8 will bark

devito/operator.py
```diff
@@ -430,7 +344,7 @@ def _profile_sections(self, nodes, parameters):
         """Introduce C-level profiling nodes within the Iteration/Expression tree."""
         nodes, profiler = create_profile('timers', nodes)
         self._globals.append(profiler.cdef)
-        parameters.append(Object(profiler.name, profiler.dtype, profiler.new))
+        parameters.append(Object(profiler.name, profiler.dtype, profiler.new()))
```
FabioLuporini: is this unpicking what I did ?

devito/tools.py
```diff
@@ -381,3 +382,104 @@ def sweep(parameters, keys=None):
                     for v in sweep_values]
     for vals in product(*sweep_values):
         yield dict(zip(keys, vals))
+
+
+class GenericVisitor(object):
+
+    """
+    A generic visitor.
```
FabioLuporini: ... for what kind of data structure?

devito/arguments.py
```diff
@@ -1,68 +1,72 @@
...
+from devito.logger import debug, error
+from devito.tools import filter_ordered, flatten, GenericVisitor
+from devito.function import CompositeFunction, SymbolicFunction
+from devito.dimension import Dimension
+from devito.ir.support.stencil import retrieve_offsets
+
+
+""" This module contains a set of classes and functions to deal with runtime arguments
+to Operators. It represents the arguments and their relationships as a DAG (N, E) where
+every node (N) is represented by an object of class Parameter and every edge is an object
```
FabioLuporini: docstring rendering: :class:`Parameter` 

Other instances should be fixed in this file. We don't need to catch them all of course, but the more the better

devito/arguments.py
```diff
@@ -1,68 +1,72 @@
...
+of class Dependency. 
+The various class hierarchies are explained here:
+Parameter:
```
FabioLuporini: pending docstring?

devito/arguments.py
```diff
@@ -1,68 +1,72 @@
...
-    """ Abstract base class for any object that represents a run time argument for
-        generated kernels.
+class Parameter(object):
+    """ Abstract base class for any object that represents a node in the dependency
+        graph. It may or may not represent a runtime argument. 
     """
-
-    __metaclass__ = abc.ABCMeta
-
+    is_Argument = False
```
FabioLuporini: Should we call the class ``Parameter`` or ``Argument``?

I think we should come up with a convention and fix this throughout the whole code base.

If there's a distinction between them, it should be made clear

devito/arguments.py
```diff
@@ -1,68 +1,72 @@
...
     is_ScalarArgument = False
     is_TensorArgument = False
     is_PtrArgument = False
-
-    def __init__(self, name, provider, default_value=None):
+    
+    __metaclass__ = abc.ABCMeta
```
FabioLuporini: leave space or flake8 will bark

devito/arguments.py
```diff
@@ -1,68 +1,72 @@
...
+    def __init__(self, name, dependencies):
```
FabioLuporini: I think it's important to provide a docstring here, especially for ``dependencies``

FabioLuporini: can we set ``dtype`` in the superclass here?

devito/arguments.py
```diff
@@ -158,198 +107,350 @@ class PtrArgument(Argument):
...
+    def __init__(self, stencils, parameters, dle_arguments):
+        self.stencils = stencils
+        self.parameters = parameters
+        self.dle_arguments = dle_arguments
+        self.argument_mapper = self._build_argument_mapper(parameters)
+        self.arguments = filter_ordered([x for x in self.argument_mapper if isinstance(x, Argument)], key=lambda x: x.name)
+        self.dims = [x for x in self.argument_mapper if isinstance(x, DimensionParameter)]
+        self.offsets = {d.end_name: v for d, v in retrieve_offsets(stencils).items()}
+
+    def handle(self, **kwargs):
```
FabioLuporini: docstring for the most important method

devito/arguments.py
```diff
@@ -158,198 +107,350 @@ class PtrArgument(Argument):
...
-
-class TensorFunctionArgProvider(ArgumentProvider):
-
-    """ Class used to decorate Symbolic Data objects with behaviour required for runtime
-        arguments.
+    def visit_SymbolicFunction(self, o):
+        return TensorArgument(o)
+    
+    def visit_DimensionParameter(self, o):
+        dependency = Dependency("gets_value_from", o)
```
FabioLuporini: can we wrap these strings within a class? Have a look at ``IterationProperty`` and the various instances ``SEQUENTIAL, PARALLEL, ...`` for an example of what I mean

devito/arguments.py
```diff
@@ -158,198 +107,350 @@ class PtrArgument(Argument):
...
+    def visit_function(self, o, param=None):
+        return o(self.consumer, self.known_values, param)
+
+    def visit_Object(self, o, param=None):
+        return o.value
+
+    def visit_object(self, o, param=None):
+        return o
+
+    def visit_DimensionParameter(self, o, param=None):
```
FabioLuporini: this is a mess to understand...

devito/arguments.py
```diff
@@ -158,198 +107,350 @@ class PtrArgument(Argument):
...
+    def _build_argument_mapper(self, parameters):
+        # Pass through SymbolicFunction
+        symbolic_functions = [x for x in parameters if isinstance(x, SymbolicFunction)]
+        dimension_dependency_mapper = dict()
+        tensor_arguments = []
+        for f in symbolic_functions:
+            argument = ArgumentVisitor().visit(f)
+            tensor_arguments.append(argument)
+            for i, d in enumerate(f.indices):
+                if d not in dimension_dependency_mapper:
```
FabioLuporini: the pythonic way of doing this sort of things is:
``dimension_dependency_mapper.setdefault(d, []).append(...)``

can we call this just ``mapper``?

devito/arguments.py
```diff
@@ -158,198 +107,350 @@ class PtrArgument(Argument):
...
+                    dimension_dependency_mapper[d] = []
+                dimension_dependency_mapper[d].append(Dependency("gets_value_from", argument, param=i))
+
+        for arg in self.dle_arguments:
+            d = arg.argument
+            if d not in dimension_dependency_mapper:
```
FabioLuporini: same

devito/arguments.py
```diff
@@ -158,198 +107,350 @@ class PtrArgument(Argument):
...
+                    dimension_dependency_mapper[d] = []
+            dimension_dependency_mapper[d].append(Dependency("gets_value_from", derive_dle_argument_value, param=arg))
+            
+        # Record dependencies in Dimensions
+        dimension_parameter_mapper = {}
+        for dim, deps in dimension_dependency_mapper.items():
+            dimension_parameter_mapper[dim] = DimensionParameter(dim, deps)
+
+        # Dimensions that are in parameters but not directly referenced in the expressions
+        for dim in [x for x in parameters if isinstance(x, Dimension) and x not in dimension_dependency_mapper.keys()]:
```
FabioLuporini: don't need ``.keys()``

devito/arguments.py
```diff
@@ -158,198 +107,350 @@ class PtrArgument(Argument):
...
+                try:
+                    dle_arguments[i.argument.name] = i.value(dim_size)
+                except TypeError:
+                    dle_arguments[i.argument.name] = i.value
+                    autotune = False
+            else:
+                dle_arguments[i.argument.name] = dim_size
+        return dle_arguments, autotune
+
+    def _derive_values(self, kwargs):
```
FabioLuporini: I could use a docstring here

devito/arguments.py
```diff
@@ -158,198 +107,350 @@ class PtrArgument(Argument):
...
+            raise InvalidArgument("Unknown arguments passed: " + ", ".join(kwargs.keys()))
+
+        # Derive values for other arguments
+        for i in self.arguments:
+            if values[i] is None:
+                known_values = OrderedDict(chain(values.items(), dimension_values.items()))
+                provided_values = [get_value(i, x, known_values) for x in i.gets_value_from]
+                assert(len(provided_values) > 0)
+        
+                if len(provided_values) == 1:
```
FabioLuporini: inline? or what happens if you just apply ``reduce``? there must be a way of dealing with length-one sequences

devito/arguments.py
```diff
@@ -158,198 +107,350 @@ class PtrArgument(Argument):
...
+                else:
+                    values[i] = reduce(provided_values)
+        
+        # Second pass to evaluate any Unevaluated dependencies from the first pass
+        for k, v in values.items():
+            if isinstance(v, UnevaluatedDependency):
+                values[k] = v.evaluate(values)
+        return values
+
+    def _verify(self, values):
```
FabioLuporini: ``all([]) --> True`` so I think this can be turned into a single line (a single return statement)

devito/arguments.py
```diff
@@ -158,198 +107,350 @@ class PtrArgument(Argument):
...
+
+        for dim in [x for x in parameters if isinstance(x, Dimension) and x.is_Stepping]:
+            dimension_parameter_mapper[dim.parent].dependencies.append(Dependency(Dependency.GETS_VALUE_FROM, dimension_parameter_mapper[dim]))
+
+        dimension_parameters = list(dimension_parameter_mapper.values())
+        
+        # Pass Dimensions
+        scalar_arguments = []
+        for dimension_parameter in dimension_parameters:
+            scalar_arguments += ArgumentVisitor().visit(dimension_parameter)
```
FabioLuporini: ``scalar_arguments = flatten(ArgumentVisitor().visit(x)) for x in dimension_parameters)``

devito/arguments.py
```diff
@@ -158,198 +107,350 @@ class PtrArgument(Argument):
...
+
+        for arg in self.dle_arguments:
+            d = arg.argument
+            if d not in dimension_dependency_mapper:
+                    dimension_dependency_mapper[d] = []
+            dimension_dependency_mapper[d].append(Dependency("gets_value_from", derive_dle_argument_value, param=arg))
+            
+        # Record dependencies in Dimensions
+        dimension_parameter_mapper = {}
+        for dim, deps in dimension_dependency_mapper.items():
```
FabioLuporini: not pythonic.
An alternative is:
``mapper = {k: DimensionParameter(k, v) for k, v in dimension_dependency_mapper.items()}``

devito/tools.py
navjotk: Since this works for Iteration-Expression trees as well as Argument trees, I don't know how to describe it. Help?

devito/operator.py
navjotk: Indeed. This will have to go before it is merged. TODO.

devito/operator.py
navjotk: You can safely assume that I will fix any flake issues before a merge since travis won't let me merge without it anyway. It would be more helpful if you could focus on things travis won't catch for now.

devito/ir/support/stencil.py
navjotk: I don't have an opinion on that either way.

devito/ir/iet/visitors.py
navjotk: Good catch. I had missed that.

devito/ir/iet/visitors.py
navjotk: It does indeed. This snippet can probably not go in master in its current form anyway as this is a very hard dependence on the arguments module - seems very wrong. Help me figure out a way to avoid doing this at all?

devito/ir/iet/visitors.py
navjotk: Because if it is not a `ScalarArgument` object, it might be an object from the `types.py` hierarchy - in which case it won't even have that flag.

devito/ir/iet/visitors.py
```diff
@@ -27,105 +27,7 @@
...
-            :param o: The :class:`Node` to visit.
-            :param args: Optional arguments to pass to the visit methods.
-            :param kwargs: Optional keyword arguments to pass to the visit methods.
-        """
-        meth = self.lookup_method(o)
-        return meth(o, *args, **kwargs)
-
-    def visit_object(self, o, **kwargs):
-        return self.default_retval()
+class Visitor(GenericVisitor):
```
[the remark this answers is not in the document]
navjotk: AFAIU those methods assume that the visitor is operating on an Iteration-Expression tree - which is not true for the visitors in the arguments module. Since the arguments visitors aren't using these methods anyway (and maybe they should - that might be causing some of the complexity in the visitors), I don't see why.

devito/dimension.py
```diff
@@ -31,6 +30,11 @@ def __new__(cls, name, **kwargs):
 
     def __str__(self):
         return self.name
+    
+    @property
+    def dtype(self):
+        #TODO: Do dimensions really need a dtype?
```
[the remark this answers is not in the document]
navjotk: Something in the DLE breaks without it. See https://github.com/opesci/devito/pull/453/commits/0db90be38fd88adb303f8a8a7ff067b7e8eb3362

devito/arguments.py
navjotk: I know. I'm trying to figure out a way to simplify this.

devito/arguments.py
navjotk: Sure.

devito/arguments.py
navjotk: [Explicit is better than implicit](https://www.python.org/dev/peps/pep-0020/)?

devito/arguments.py
navjotk: There are multiple mappers in the namespace at this point so the name is being specific about what it maps, dimensions to their dependencies.

devito/arguments.py
navjotk: `Parameter` objects need not have a `dtype`. `Argument` objects do. Difference explained in comment above.

devito/arguments.py
navjotk: Yes, sorry for posting this in a WIP state. Updating.

devito/arguments.py
navjotk: There is a difference. Not all entities that depend on arguments or that arguments depend on are runtime arguments themselves. The most common example are `Dimension`s. While `Dimension`s appear in the dependency tree - the symbolic `Dimension` object for `time`, for example carries the only relationship between the array being worked on (example `u=TimeData(..., save=True)`) and the runtime argument called `time_e` (which is an argument passed to the C function at execution time). `time` itself is never passed to the C function but without it we wouldn't be able to build a relationship between the allocated memory for `u` and the maximum number of allowed timesteps. All nodes in the dependency tree are required to be of type `Parameter`, but that doesn't make them runtime arguments. Runtime arguments are of type `Argument`. In this example, `time` would appear in the dependency tree as a `Parameter` object while `time_e` would be of type `Argument`. 
This is partially explained in the docstring but I will elaborate using this comment there as well.

FabioLuporini: OK, makes sense and thanks for clarifying this in the docs. 

(nitpicking, maybe): I wonder whether using ``RuntimeArgument``, instead of ``Argument`` , would help? The distinction would then become immediately clear

devito/arguments.py
FabioLuporini: Right; how about in ``Argument``'s constructor then?

devito/arguments.py
FabioLuporini: Maybe an ASCII diagram of the class hierarchy would help

devito/arguments.py
```diff
@@ -75,78 +79,23 @@ class ScalarArgument(Argument):
...
-                # or this is an enforced value
-                if enforce or not self._frozen:
-                    # If we're frozen, ignore, else use this value directly
-                    self._value = value
-                    if enforce:
-                        # If we were forced to use this value, make sure we don't change
-                        # it in the same invocation
-                        self._frozen = True
-        return self._value is not None
+    def __init__(self, name, dependencies, dtype=np.int32):
```
FabioLuporini: for consistency, it'd help to have a (default) ``provider`` for ``ScalarArgument`` too. Then you could move some code into the superclass constructor (``Argument``) and have a property returning the ``dtype`` directly from the provider. Not sure if this is doable

devito/arguments.py
```diff
@@ -158,198 +107,350 @@ class PtrArgument(Argument):
...
+    def _dle_arguments(self, dim_sizes):
+        # Add user-provided block sizes, if any
+        dle_arguments = OrderedDict()
+        autotune = True
+        for i in self.dle_arguments:
+            dim_size = dim_sizes.get(i.original_dim.name, None)
+            if dim_size is None:
+                error('Unable to derive size of dimension %s from defaults. '
+                      'Please provide an explicit value.' % i.original_dim.name)
+                raise InvalidArgument('Unknown dimension size')
```
FabioLuporini: Question: this is something I was seeing in ``function.py`` too. On (user) errors, we have both ``error`` and ``raise``. Why do we need both? I think you can just stick the error message within ``raise``

devito/arguments.py
```diff
@@ -158,198 +107,350 @@ class PtrArgument(Argument):
...
+                dle_arguments[i.argument.name] = dim_size
+        return dle_arguments, autotune
+
+    def _derive_values(self, kwargs):
+        # Use kwargs
+        values = OrderedDict()
+        dimension_values = OrderedDict()
+        
+        for i in self.arguments:
+            values[i] = get_value(i, kwargs.pop(i.name, None), values)
```
FabioLuporini: ``values = OrderedDict([(i, get_value(i, ...) for i in self.arguments)])``

devito/arguments.py
```diff
@@ -158,198 +107,350 @@ class PtrArgument(Argument):
...
+
+        for i in self.dims:
+            dimension_values[i] = kwargs.pop(i.name, None)
```
FabioLuporini: same as before (not pythonic this way)

devito/arguments.py
FabioLuporini: OK so ``reduce`` expects at least two arguments, here you're providing one, how can it work?

And it should just work if you do:
``values[i] = reduce(function_u_want_to_apply, provided_values)``

As for the if-then-else question: I agree it's a matter of taste; what I do is that when they are (semantically) trivial, I try to write them as a single line. WHen I need to emphasise something (and perhaps I feel the need to add a comment), then I split it over multiple lines

devito/arguments.py
```diff
@@ -158,198 +107,350 @@ class PtrArgument(Argument):
...
+        
+                if len(provided_values) == 1:
+                    values[i] = provided_values[0]
+                else:
+                    values[i] = reduce(provided_values)
+        
+        # Second pass to evaluate any Unevaluated dependencies from the first pass
+        for k, v in values.items():
+            if isinstance(v, UnevaluatedDependency):
+                values[k] = v.evaluate(values)
```
FabioLuporini: not so pythonic, but most importantly changing a dict in place through ``items`` which is not doable, I seem to remember, in P3 (correct me if I'm wrong?)

potential fix:
``values = OrderedDict([(k, v.evaluate(values) if isinstance(v, UnevaluatedDependency) else v) for k, v in values.items()])``

Alternative: ``for k, v list(values.items()): ...``

devito/arguments.py
```diff
@@ -158,198 +107,350 @@ class PtrArgument(Argument):
...
-        value = (value, start_arg.default_value, value)
+            arg_str.append('(%s, value=%s, type=%s)' % (k, str(v), type(v)))
+    print("Passing Arguments: " + ", ".join(arg_str))
+
+
+def find_argument_by_name(name, haystack):
+    filtered = [(k, v) for k, v in haystack.items() if k.name == name]
+    assert(len(filtered) < 2)
+    if len(filtered) == 1:
+        return filtered[0][1]
```
FabioLuporini: you could perhaps just store ``v`` in ``filtered``, so you won't need ``[1]`` here?

devito/dimension.py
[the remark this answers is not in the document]
FabioLuporini: @mlange05 do you think it's feasible to have ``Dimension``s inherit from ``devito.Symbol``, rather then ``sympy.Symbol``? Our symbols carry a dtype already. I'm asking because I seem to remember you were the one playing with this stuff not so long ago?

devito/ir/iet/visitors.py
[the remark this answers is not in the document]
FabioLuporini: oh yes you're right, forget this

devito/ir/iet/visitors.py
```diff
@@ -233,6 +135,9 @@ def _args_decl(self, args):
         """Convert an iterable of :class:`Argument` into cgen format."""
```
FabioLuporini: random question: is this docstring now obsolete , in the sense that it's not just ``Argument``?

devito/ir/iet/visitors.py
FabioLuporini: As you know, I'm not a big fan at all of this kind of ``hasattr(...)`` approach, which seems a bit hacky to me. If considered pythonic and I'm actually missing something, I'm all ears!

So I don't know what to suggest ATM, but it feels very weird that "something might not be an actual Argument``, **but**:
* other objects in other hierarchies have an analogous flag (``is_ScalarArgument``)
* below, we're using an ``ArgumentVisitor`` and this can handle stuff which are non-Arguments

what kind of objects from ``types.py`` you may encounter here?

devito/ir/iet/visitors.py
FabioLuporini: we might have to split code generation into multiple steps, such as
* collection of arguments
* generation of body (CGen)
* generation of arguments
* etc ?

Not sure if this actually makes sense

devito/ir/support/stencil.py
FabioLuporini: Then yes please

devito/tools.py
FabioLuporini: how about

"A generic visitor for tree data structures."

devito/arguments.py
navjotk: Fixed

devito/arguments.py
navjotk: Done

devito/arguments.py
navjotk: Done

devito/arguments.py
navjotk: Done

navjotk: Done

devito/arguments.py
navjotk: Fixed

devito/arguments.py
navjotk: Fixed

devito/arguments.py
navjotk: Fixed

devito/arguments.py
navjotk: Fixed

devito/arguments.py
navjotk: `provider`s don't make sense for `ScalarArgument`s.

devito/arguments.py
navjotk: This is how the code in master currently is. I just moved it from `operator.py` to `arguments.py`

devito/arguments.py
navjotk: I believe this is working. I don't see what's wrong. The line that changes the dict does not access it using items, it does it using `values[k]` so it should just work, and I believe it does.

FabioLuporini: weird... 
anyway, if the dictionary isn't huge, a more pyhtonic way might help:

``values = {k: v.evaluate(values) if isinstance(v, UnevaluatedDependency) else v for k, v in values.items()}``

devito/arguments.py
```diff
@@ -158,198 +141,350 @@ class PtrArgument(Argument):
...
+            dimension_dependency_mapper.setdefault(d, []).append(Dependency("gets_value_from", derive_dle_argument_value, param=arg))
+            
+        # Record dependencies in Dimensions
+        dimension_parameter_mapper = {k: DimensionParameter(k, v) for k, v in dimension_dependency_mapper.items()}
+
+        # Dimensions that are in parameters but not directly referenced in the expressions
+        for dim in [x for x in parameters if isinstance(x, Dimension) and x not in dimension_dependency_mapper]:
+            dimension_parameter_mapper[dim] = DimensionParameter(dim, [])
+
+        for dim in [x for x in parameters if isinstance(x, Dimension) and x.is_Stepping]:
```
FabioLuporini: here and above, can you also use the ``dimensions`` property?

devito/arguments.py
```diff
@@ -158,198 +141,350 @@ class PtrArgument(Argument):
...
+    def evaluate(self, known_values):
+        return self.evaluator(self.consumer, known_values, self.extra_param)
 
-    """ Class used to decorate Objects with behaviour required for runtime arguments.
-    """
 
-    @cached_property
-    def rtargs(self):
-        return (PtrArgument(self.name, self),)
+def runtime_arguments(parameters):
```
FabioLuporini: Is this ever used?

devito/ir/iet/visitors.py
navjotk: No. It continues to be accurate.

devito/arguments.py
navjotk: https://github.com/opesci/devito/pull/453/files#diff-1526813b3ca372b26a915777ac6e7f4fR251

navjotk:
Tests passing. I believe I have addressed everything.

devito/dimension.py
```diff
@@ -32,6 +31,11 @@ def __new__(cls, name, **kwargs):
     def __str__(self):
         return self.name
 
+    @property
+    def dtype(self):
+        # TODO: Do dimensions really need a dtype?
```
FabioLuporini: This should be figured out -- it feels quite an important detail. Should this rather inherit from ``devito.Symbol`` ?

devito/dimension.py
```diff
@@ -113,7 +113,6 @@ class SteppingDimension(Dimension):
     Dimension symbol that defines the stepping direction of an
     :class:`Operator` and implies modulo buffered iteration. This is most
     commonly use to represent a timestepping dimension.
-
```
FabioLuporini: why?

devito/dle/backends/basic.py
```diff
@@ -107,7 +107,7 @@ def _create_elemental_functions(self, nodes, state):
             args.extend([(i.name, Scalar(name=i.name, dtype=i.dtype)) for i in required])
 
             call, params = zip(*args)
-            handle = flatten([p.rtargs for p in params])
+            handle = flatten(params)
```
FabioLuporini: the ``handle`` variable can at this point be dropped and its rvalue inlined at 117

devito/tools.py
FabioLuporini: reminder

examples/checkpointing/checkpoint.py
```diff
@@ -13,7 +13,7 @@ class CheckpointOperator(Operator):
                     automatically include these cached arguments in the call to the
                     underlying devito.Operator.apply().
     """
-    t_arg_names = {'t_start': 't_s', 't_end': 't_e'}
+    t_arg_names = {'t_start': 'time_s', 't_end': 'time_e'}
```
FabioLuporini: this hardcoding is bad, although it was this way in the previous PR too...

devito/arguments.py
```diff
@@ -1,152 +1,133 @@
...
+from devito.dimension import Dimension
+from devito.ir.support.stencil import retrieve_offsets
+
+
+""" This module contains a set of classes and functions to deal with runtime arguments
+to Operators. It represents the arguments and their relationships as a DAG (N, E) where
+every node (N) is represented by an object of :class:`Parameter` and every edge is an
+object of class :class:`Dependency`.
+
+The various class hierarchies are explained here:
```
FabioLuporini: It's a single class hierarchy, I guess?

devito/arguments.py
```diff
@@ -1,152 +1,133 @@
...
+                --------------------
+                |         |        |
+        ScalarArgument    |        |
+                    TensorArgument |
+                                PtrArgument
+
+ArgumentEngine: The main external API for this module. It encapsulates all the argument
+                derivation and verification logic.
+
+ArgumentVisitor: Visits objects of devito's data types to return appropriate objects
```
FabioLuporini: "Visits objects of devito's data types ... "

you mean ``:class:`function.Basic` `` ?

devito/arguments.py
```diff
@@ -1,152 +1,133 @@
...
-            if self._value.is_SymbolicFunction:
-                return self._value._data_buffer
-            elif self._value.is_Constant:
-                return self._value.data
-            else:
-                raise InvalidArgument("Unexpected data object %s" % type(self._value))
-        except AttributeError:
-            return self._value
+    def gets_value_from(self):
+        return [x for x in self.dependencies if x.dependency_type == "gets_value_from"]
```
FabioLuporini: ``gets_value_from`` is still a magic string

devito/arguments.py
```diff
@@ -1,152 +1,133 @@
...
 
     @property
-    def dtype(self):
-        return self.provider.dtype
+    def verified_by(self):
+        return [x for x in self.dependencies if x.dependency_type == "verified_by"]
```
FabioLuporini: same is ``verified_by``

devito/arguments.py
```diff
@@ -1,152 +1,133 @@
...
 
-    def reset(self):
-        self._value = self.default_value
+    def __repr__(self):
+        return self.name + ", Depends: " + str(self.dependencies)
```
FabioLuporini: This is an ancient way of constructing strings; should use ``"%s Depends: %s"``

devito/arguments.py
```diff
@@ -1,152 +1,133 @@
...
+        generated kernels. It is necessarily a node in the dependency graph.
+    """
+    is_Argument = True
+
+    def __init__(self, name, dependencies, dtype=np.int32):
+        super(Argument, self).__init__(name, dependencies)
+        self.dtype = dtype
 
 
 class ScalarArgument(Argument):
```
FabioLuporini: So these still don't have providers?

devito/arguments.py
```diff
@@ -1,152 +1,133 @@
...
-
-        verify = verify and all(d.verify(v) for d, v in
-                                zip(self.provider.indices, value.shape))
-        if verify:
-            self._value = value
-
-        return self._value is not None and verify
+    def __init__(self, provider, dependencies=[]):
+        super(TensorArgument, self).__init__(provider.name,
+                                             dependencies + [Dependency("gets_value_from",
```
FabioLuporini: magic string

devito/arguments.py
```diff
@@ -1,152 +1,133 @@
...
-
-        verify = len(self.provider.shape) == len(value.shape)
-
-        verify = verify and all(d.verify(v) for d, v in
-                                zip(self.provider.indices, value.shape))
-        if verify:
-            self._value = value
-
-        return self._value is not None and verify
+    def __init__(self, provider, dependencies=[]):
```
FabioLuporini: Use ``None``, not ``[]``

http://docs.python-guide.org/en/latest/writing/gotchas/

please check this everywhere

devito/arguments.py
```diff
@@ -158,198 +139,381 @@ class PtrArgument(Argument):
...
+class ArgumentEngine(object):
+    """ Class that encapsulates the argument derivation and verification subsystem
     """
+    def __init__(self, stencils, parameters, dle_arguments):
+        self.stencils = stencils
+        self.parameters = parameters
+        self.dle_arguments = dle_arguments
+        self.argument_mapper = self._build_argument_mapper(parameters)
+        self.arguments = filter_ordered([x for x in self.argument_mapper
+                                         if isinstance(x, Argument)],
```
FabioLuporini: not sure if I had asked already. Why not just ``x.is_Argument`` ?

devito/arguments.py
```diff
@@ -158,198 +139,381 @@ class PtrArgument(Argument):
...
+        kwargs = self._offset_adjust(kwargs)
+
+        kwargs = self._extract_children_of_composites(kwargs)
+
+        values = self._derive_values(kwargs)
+
+        # The following is only being done to update the autotune flag. The actual value
+        # derivation for the dle arguments has moved inside the above _derive_values
+        # method.
+        # TODO: Refactor so this is not required
```
FabioLuporini: Refactor now, I strongly encourage

devito/arguments.py
```diff
@@ -158,198 +139,381 @@ class PtrArgument(Argument):
...
-    """ Abstract base class for any object that, post code-generation, might resolve
-        resolve to runtime arguments. We assume that one source object (e.g. Dimension,
-        SymbolicData) might provide multiple runtime arguments.
+class ArgumentEngine(object):
+    """ Class that encapsulates the argument derivation and verification subsystem
     """
+    def __init__(self, stencils, parameters, dle_arguments):
+        self.stencils = stencils
+        self.parameters = parameters
+        self.dle_arguments = dle_arguments
```
FabioLuporini: I hope we can find a way of making the arguments engine unaware of the DLE. These are just arguments in the end...

devito/arguments.py
```diff
@@ -158,198 +139,381 @@ class PtrArgument(Argument):
...
-class ArgumentProvider(object):
-
-    """ Abstract base class for any object that, post code-generation, might resolve
-        resolve to runtime arguments. We assume that one source object (e.g. Dimension,
-        SymbolicData) might provide multiple runtime arguments.
+class ArgumentEngine(object):
+    """ Class that encapsulates the argument derivation and verification subsystem
     """
+    def __init__(self, stencils, parameters, dle_arguments):
+        self.stencils = stencils
```
FabioLuporini: unnecessary

devito/arguments.py
FabioLuporini: And I don't think we need to store them as instance attributes. They seem to be processed at construction time, and then they also become unnecessary. I may be wrong, I just tried following the logical flow

devito/arguments.py
```diff
@@ -158,198 +139,381 @@ class PtrArgument(Argument):
...
+    def _build_argument_mapper(self, parameters):
+        # Pass through SymbolicFunction
+        symbolic_functions = [x for x in parameters if isinstance(x, SymbolicFunction)]
+        dim_dep_mapper = OrderedDict()  # Mapper for dependencies between dimensions
+        tensor_arguments = []
+        for f in symbolic_functions:
+            argument = ArgumentVisitor().visit(f)
+            tensor_arguments.append(argument)
+            for i, d in enumerate(f.indices):
+                dim_dep_mapper.setdefault(d, []).append(Dependency("gets_value_from",
```
FabioLuporini: magic strings again

devito/arguments.py
```diff
@@ -158,198 +139,381 @@ class PtrArgument(Argument):
...
+                                                                   argument, param=i))
+
+        for arg in self.dle_arguments:
+            d = arg.argument
```
FabioLuporini: is this really necessary?

devito/arguments.py
```diff
@@ -158,198 +139,381 @@ class PtrArgument(Argument):
...
+        """ Populate values for all the arguments. The values provided in kwargs will
+            be used wherever provided. The values for the rest of the arguments will be
+            derived from the ones provided. The default values for the tensors come from
+            the data property of the symbols used in the Operator.
         """
-        raise NotImplemented()
+        values = OrderedDict()
+
+        for i in self.arguments:
+            values[i] = get_value(i, kwargs.pop(i.name, None), values)
```
FabioLuporini: not pythonic

devito/arguments.py
FabioLuporini: reminder

devito/arguments.py
```diff
@@ -158,198 +139,381 @@ class PtrArgument(Argument):
...
     """
+    def __init__(self, stencils, parameters, dle_arguments):
+        self.stencils = stencils
+        self.parameters = parameters
+        self.dle_arguments = dle_arguments
+        self.argument_mapper = self._build_argument_mapper(parameters)
+        self.arguments = filter_ordered([x for x in self.argument_mapper
+                                         if isinstance(x, Argument)],
+                                        key=lambda x: x.name)
+        self.dims = [x for x in self.argument_mapper if isinstance(x, DimensionParameter)]
```
FabioLuporini: Having both ``dims`` and ``dimensions`` (a property) in the same class is really confusing

devito/arguments.py
```diff
@@ -158,198 +139,381 @@ class PtrArgument(Argument):
...
+                    will store the dependency object on itself. The obj referred to here
+                    is the target node of the dependency.
+        :param param: An optional parameter that might be required to evaluate the
+                      relationship being defined by this Dependency. e.g. when a Dimension
+                      derives its value from a SymbolicFunction's shape, this param
+                      carries the index of a dimension in the SymbolicFunction's shape.
     """
+    GETS_VALUE_FROM = "gets_value_from"
+    VERIFIED_BY = "verified_by"
+    _types = ["gets_value_from", "verified_by"]
```
FabioLuporini: _types = [GETS_VALUE_FROM, VERIFIED_BY]

devito/arguments.py
```diff
@@ -158,198 +139,381 @@ class PtrArgument(Argument):
...
-    def rtargs(self):
-        return (ScalarArgument(self.name, self, self.dtype),)
+    def __init__(self, dependency_type, obj, param=None):
+        assert(dependency_type in self._types)
+        self.dependency_type = dependency_type
+        self.obj = obj
+        self.param = param
 
+    def __repr__(self):
+        return "(" + self.dependency_type + ":" + str(self.obj) + ")"
```
FabioLuporini: string concatenation is the past

devito/arguments.py
```diff
@@ -158,198 +139,381 @@ class PtrArgument(Argument):
...
-        resolve to runtime arguments. We assume that one source object (e.g. Dimension,
-        SymbolicData) might provide multiple runtime arguments.
+class ArgumentEngine(object):
+    """ Class that encapsulates the argument derivation and verification subsystem
     """
+    def __init__(self, stencils, parameters, dle_arguments):
+        self.stencils = stencils
+        self.parameters = parameters
+        self.dle_arguments = dle_arguments
+        self.argument_mapper = self._build_argument_mapper(parameters)
```
FabioLuporini: it's unnecessary to store it as instance attribute; never used outside of ``__init__``

devito/arguments.py
```diff
@@ -158,198 +139,381 @@ class PtrArgument(Argument):
...
+        if callable(o.value):
+            return o.value()
+        else:
+            return o.value
 
+    def visit_object(self, o, param=None):
+        return o
 
-class ScalarArgProvider(ArgumentProvider):
+    def visit_DimensionParameter(self, o, param=None):
```
FabioLuporini: A docstring here would help

devito/arguments.py
navjotk: No it is not. `Dependency` does not inherit from `Parameter`

devito/arguments.py
navjotk: No. Does not make sense to me for `ScalarArgument`s to have providers.

FabioLuporini: why?

devito/arguments.py
navjotk: Fixed

devito/arguments.py
navjotk: TensorArguments come from objects that are symbols but have data attached to them - that's their provider. ScalarArguments don't have any such "provider". Perhaps you could explain why you think they do.

devito/arguments.py
navjotk: Fixed

devito/arguments.py
navjotk: Fixed

devito/arguments.py
navjotk: Fixed

devito/arguments.py
navjotk: Nested list comprehension? I think I prefer the 4 line approach that is more readable than a nested list comprehension.

devito/arguments.py
navjotk: Fixed

devito/arguments.py
navjotk: https://github.com/opesci/devito/blob/f9ac081b8f829afc3fde0485f2ed01d2769f46df/devito/arguments.py#L262

devito/arguments.py
navjotk: Fixed

devito/arguments.py
navjotk: Fixed

devito/arguments.py
navjotk: Fixed

devito/arguments.py
navjotk: Let's leave it for the next PR

devito/arguments.py
navjotk: Fixed

devito/arguments.py
navjotk: Please elaborate

devito/arguments.py
navjotk: Please elaborate

devito/arguments.py
navjotk: Fixed

devito/arguments.py
navjotk: Fixed

devito/arguments.py
navjotk: Fixed

devito/dimension.py
navjotk: Again, largely unrelated to the point of this PR. Doesn't HAVE to be answered before this goes in.

examples/checkpointing/checkpoint.py
navjotk: Containing this hardcoding is the reason this wrapper exists.

devito/dimension.py
navjotk: Unnecessary space

devito/dle/backends/basic.py
navjotk: Fixed

devito/arguments.py
```diff
@@ -158,198 +139,381 @@ class PtrArgument(Argument):
...
+                          for d in self.dimensions])
+        dle_arguments, dle_autotune = self._dle_arguments(dim_sizes)
+
+        assert(self._verify(values))
+
+        arguments = OrderedDict([(k.name, v) for k, v in values.items()])
+
+        return arguments, user_autotune and dle_autotune
+
+    def _offset_adjust(self, kwargs):
```
FabioLuporini: ``return {k: v + self.offsets.get(k, 0) for k, v in kwargs.items()}``

FabioLuporini: so this can probably be inlined

devito/arguments.py
FabioLuporini: why do you need the temporary variable ``d``, used in only one place? how does it help

devito/arguments.py
```diff
@@ -158,198 +139,381 @@ class PtrArgument(Argument):
...
+                                                               param=arg))
+
+        # Record dependencies in Dimensions
+        dim_param_mapper = OrderedDict([(k, DimensionParameter(k, v)) for k, v in
+                                        dim_dep_mapper.items()])
+
+        # Dimensions that are in parameters but not directly referenced in the expressions
+        more = dict([(x, DimensionParameter(x, [])) for x in self.dimensions if x not in
+                     dim_dep_mapper])
+        dim_param_mapper.update(more)
```
FabioLuporini: probably don't need the temporary variable ``more``

devito/arguments.py
```diff
@@ -158,198 +139,381 @@ class PtrArgument(Argument):
...
+            dim_param_mapper[dim.parent].dependencies.append(
+                Dependency(Dependency.GETS_VALUE_FROM, dim_param_mapper[dim]))
+
+        dimension_parameters = list(dim_param_mapper.values())
+
+        scalar_arguments = flatten([ArgumentVisitor().visit(x)
+                                    for x in dimension_parameters])
+
+        other_arguments = [ArgumentVisitor().visit(x) for x in parameters
+                           if x not in tensor_arguments + scalar_arguments]
```
FabioLuporini: With an ``OrderedDict`` (an ``OrderedSet``, TBH) you wouldn't need these extra checks, and you wouldn't need four different variables (``tensor_arguments``, ``dimension_parameters``, ...)

devito/arguments.py
FabioLuporini: ``values = OrderedDict([(i, get_value(i, kwargs.pop(i.name, None), values)) for i in self.arguments])``

devito/ir/support/stencil.py
navjotk: Fixed

devito/arguments.py
navjotk: Pushed now.

devito/arguments.py
navjotk: In the current design, the `ValueVisitor` is deriving the values so a provider isn't necessary. We can discuss it for the future possibly.

devito/arguments.py
navjotk: Fixed

devito/arguments.py
navjotk: Done

devito/dle/backends/basic.py
navjotk: Look again

devito/arguments.py
navjotk: I prefer to treat `tensor_arguments`, `dimension_parameters` etc separately. I don't see why that is wrong.

devito/arguments.py
navjotk: Fixed

devito/arguments.py
navjotk: I think a temporary variable isn't a very bad price to pay for increased readability.

devito/ir/support/stencil.py
```diff
@@ -106,6 +106,17 @@ def union(cls, *dicts):
...
 
+    @classmethod
+    def retrieve_offsets(cls, stencils):
+        """
+        Return a mapper from :class:`Dimension`s to the min/max integer offsets
+        within ``stencils``.
+        """
+        offs = Stencil.union(*stencils)
+        mapper = {d: v for d, v in offs.diameter.items()}
+        mapper.update({d.parent: v for d, v in mapper.items() if d.is_Stepping})
```
FabioLuporini: oh, no! this is too bad! 
``d.is_Stepping`` is a concept that shouldn't live here. It's completely unrelated.
Unfortunately, I only noticed it now: I suggest we move this whole routine to the only place where it's used.

devito/arguments.py
```diff
@@ -1,152 +1,139 @@
...
-    """ Abstract base class for any object that represents a run time argument for
-        generated kernels.
+    :param name: Name of the parameter
+    :param dependencies: A list of :class:`Dependency` objects that represent all the
+                         incoming edges into this node of the graph.
     """
-
-    __metaclass__ = abc.ABCMeta
-
+    is_Dimension_Parameter = False
```
FabioLuporini: As a convention, the '_' only lives between "is" and the rest

devito/arguments.py
```diff
@@ -1,152 +1,139 @@
...
-
-        verify = verify and all(d.verify(v) for d, v in
-                                zip(self.provider.indices, value.shape))
-        if verify:
-            self._value = value
-
-        return self._value is not None and verify
+    def __init__(self, provider, dependencies=None):
+        if dependencies is None:
+            dependencies = []
```
FabioLuporini: pythonic way:
``dependences = dependences or []``

devito/arguments.py
```diff
@@ -158,198 +145,398 @@ class PtrArgument(Argument):
...
+        dim_param_mapper = OrderedDict([(k, DimensionParameter(k, v)) for k, v in
+                                        dim_dep_mapper.items()])
+
+        # Dimensions that are in parameters but not directly referenced in the expressions
+        more = dict([(x, DimensionParameter(x, [])) for x in self.dimensions if x not in
+                     dim_dep_mapper])
+        dim_param_mapper.update(more)
+
+        for dim in [x for x in self.dimensions if x.is_Stepping]:
+            dim_param_mapper[dim.parent].dependencies +=\
```
FabioLuporini: ``.append`` for consistency (+= creates a new list; in Devito, we try to use ``append`` all over the place when possible, rather than +=)

devito/arguments.py
```diff
@@ -158,198 +145,398 @@ class PtrArgument(Argument):
...
+        return dle_arguments, autotune
+
+    def _derive_values(self, kwargs):
+        """ Populate values for all the arguments. The values provided in kwargs will
+            be used wherever provided. The values for the rest of the arguments will be
+            derived from the ones provided. The default values for the tensors come from
+            the data property of the symbols used in the Operator.
         """
-        raise NotImplemented()
+        values = OrderedDict()
```
FabioLuporini: drop

devito/arguments.py
```diff
@@ -158,198 +145,398 @@ class PtrArgument(Argument):
...
+        # Pass through SymbolicFunction
+        symbolic_functions = [x for x in parameters if isinstance(x, SymbolicFunction)]
+        dim_dep_mapper = OrderedDict()  # Mapper for dependencies between dimensions
+        tensor_arguments = []
+        for f in symbolic_functions:
+            argument = ArgumentVisitor().visit(f)
+            tensor_arguments.append(argument)
+            for i, d in enumerate(f.indices):
+                dim_dep_mapper.setdefault(d,
+                                          []).append(ValueDependency(argument, param=i))
```
FabioLuporini: What I meant with my previous comment was that the nicer decomposition of this into two stmts (which maximises readability) is:
`
v = dim_dep_mapper.setdefault(d, [])
v.append(ValueDependency(argument, param=i))
`

Same a few lines below

devito/arguments.py
FabioLuporini: not seeing it...

devito/arguments.py
```diff
@@ -158,198 +145,398 @@ class PtrArgument(Argument):
...
-        # Derived dimensions could be linked through constraints
-        # At this point, a constraint needs to be added that enforces
-        # dim_e - dim_s < SOME_MAX
-        # Also need a default constraint that dim_e > dim_s (or vice-versa)
-        verify = verify and all([a.verify(v, enforce=enforce) for a, v in
-                                 zip(self.rtargs, value)])
-        return verify
+    def visit_Scalar(self, o):
+        arg = ScalarArgument(o.name, o, dtype=o.dtype)
+        arg.provider = o
```
FabioLuporini: Nope. You're creating a ``ScalarArgument`` and then, externally, you're dynamically adding an attribute to it. Use the constructor instead.

devito/arguments.py
```diff
@@ -158,198 +145,398 @@ class PtrArgument(Argument):
...
+        return arg
 
+    def visit_Constant(self, o):
+        # TODO: Add option for delayed query of default value
+        dependency = ValueDependency(o)
+        arg = ScalarArgument(o.name, [dependency], dtype=o.dtype)
+        arg.provider = o
```
FabioLuporini: same story

devito/arguments.py
```diff
@@ -158,198 +145,398 @@ class PtrArgument(Argument):
...
-                                 zip(self.rtargs, value)])
-        return verify
+    def visit_Scalar(self, o):
+        arg = ScalarArgument(o.name, o, dtype=o.dtype)
+        arg.provider = o
+        return arg
 
+    def visit_Constant(self, o):
+        # TODO: Add option for delayed query of default value
+        dependency = ValueDependency(o)
```
FabioLuporini: you can inline this line into the following one

devito/arguments.py
```diff
@@ -158,198 +145,398 @@ class PtrArgument(Argument):
...
 
-    @cached_property
-    def rtargs(self):
-        return (TensorArgument(self.name, self),)
+    def visit_object(self, o, param=None):
+        return o
 
+    def visit_DimensionParameter(self, o, param=None):
+        """ Called when some Argument's value is being derived and the dependency tree
+            was followed till a DimensionParameter object. The Argument for which we are
```
FabioLuporini: :class:`DimensionParameter`

devito/arguments.py
```diff
@@ -158,198 +145,398 @@ class PtrArgument(Argument):
...
+        else:
+            provided_values = [get_value(o, x, self.known_values)
+                               for x in o.gets_value_from]
+
+        if len(provided_values) > 0:
+            if not all(x is not None for x in provided_values):
+                unknown_args = [x.obj for x in o.gets_value_from
+                                if get_value(o, x, self.known_values) is None]
+
+                def late_evaluate_dim_size(consumer, known_values, partial_values):
```
FabioLuporini: when/why is this used/needed?

devito/arguments.py
```diff
@@ -158,198 +145,398 @@ class PtrArgument(Argument):
...
+               between any pair of nodes.
+        :param obj: Dependencies have a source node and a target node. The source node
+                    will store the dependency object on itself. The obj referred to here
+                    is the target node of the dependency.
+        :param param: An optional parameter that might be required to evaluate the
+                      relationship being defined by this Dependency. e.g. when a Dimension
+                      derives its value from a SymbolicFunction's shape, this param
+                      carries the index of a dimension in the SymbolicFunction's shape.
     """
+    GETS_VALUE_FROM = "gets_value_from"
```
FabioLuporini: ooohhh, but there's a hierarchy here! So you probably don't need ``GETS_VALUE_FROM, VERIFIED_BY, _types, ...``, and you can also drop ``self.dependency_type``. This information is already encoded in the type of the subclass. 
The simplest thing to do is to introduce the usual ``is_VerifiedBy/is_GetsValueFrom`` (better name? ``is_GettingValue``?). This way, the subclasses become empty, apart from the presence of ``is_X = True``

devito/arguments.py
```diff
@@ -158,198 +145,398 @@ class PtrArgument(Argument):
...
+            if values[i] is None:
+                known_values = OrderedDict(chain(values.items(),
+                                                 dimension_values.items()))
+                provided_values = [get_value(i, x, known_values)
+                                   for x in i.gets_value_from]
+                assert(len(provided_values) > 0)
+                values[i] = reduce(max, provided_values)
+
+        # Second pass to evaluate any Unevaluated dependencies from the first pass
+        for k, v in values.items():
```
FabioLuporini: could be made more pythonic

devito/arguments.py
```diff
@@ -158,198 +145,398 @@ class PtrArgument(Argument):
...
-    """ Abstract base class for any object that, post code-generation, might resolve
-        resolve to runtime arguments. We assume that one source object (e.g. Dimension,
-        SymbolicData) might provide multiple runtime arguments.
+class ArgumentEngine(object):
+    """ Class that encapsulates the argument derivation and verification subsystem
     """
+    def __init__(self, stencils, parameters, dle_arguments):
+        self.parameters = parameters
+        self.dle_arguments = dle_arguments
+        argument_mapper = self._build_argument_mapper(parameters)
```
FabioLuporini: I don't like the suffix ``_mapper``, as this seems to be a list. Is this a list?

devito/arguments.py
navjotk: I disagree: https://stackoverflow.com/questions/3901144/difference-between-if-x-and-if-x-is-not-none

devito/arguments.py
navjotk: Tried it. It won't work. 0 is not a good default value because not all arguments support the `+` operation.

devito/arguments.py
navjotk: Won't work without it. The list comprehension requires the array to exist beforehand.

devito/arguments.py
navjotk: Yes, because according to my design, `ScalarArgument`s don't need a provider but in `yask/operator.py/arguments` your code assumes that they have `provider`s. I'm only adding the provider for compatibility with your code since I want to limit the amount of change this PR brings. In a future PR I will clean up your code so it doesn't need a `provider` on `ScalarArgument`s.

devito/arguments.py
navjotk: Same answer

devito/arguments.py
navjotk: If an argument depends on the size of an autotuned dimension, the argument's value can not be determined at the time this derivation routine is run. This lazy evaluation routine is called to derive the value of the arguments once the size of the dimension has been derived by the autotuner.

devito/arguments.py
navjotk: Fixed

devito/arguments.py
navjotk: Fixed

devito/arguments.py
navjotk: Fixed

devito/arguments.py
navjotk: Done

devito/arguments.py
navjotk: Fixed

devito/ir/support/stencil.py
navjotk: Fixed

devito/arguments.py
navjotk: Done

devito/arguments.py
navjotk: I can't see how. Note that I'm modifying a dict in place, not creating one.

devito/arguments.py
navjotk: Fixed

devito/arguments.py
FabioLuporini: dependencies can be either ``None`` or something that evaluates to True, hence the pythonic way is preferable. I'm 99% sure this is common practice when using ``None`` as default argument value (instead of [], for example)

devito/arguments.py
FabioLuporini: still, I encourage to make it pythonic:
``return {k: (v + self.offsets[k]) if k in self.offsets else v for k, v in kwargs.items()}``

devito/arguments.py
FabioLuporini: sure, that was nitpicking on my side. However, let's remember this possibility *if* the performance of the arguments engine turns out to be poor, as I suspect an OrderedDict might make things much faster (as opposed to creating and concatenating lists, and using if-then-else).

devito/arguments.py
FabioLuporini: Do I get it right: you want a reference to the same ``values`` to be passed to ``get_value``? even though this is an empty dict?

devito/arguments.py
FabioLuporini: No. This is seriously a bad hack. I accept postponing all other refactorings, but this should really really go.

devito/arguments.py
navjotk: It is unclear to me why you would want me to change a tighter test to one that is more prone to unintentional effects.

FabioLuporini: it was nitpicking, we can ignore this

devito/arguments.py
FabioLuporini: fine

devito/arguments.py
navjotk: It is not empty even after a single iteration of the loop

devito/arguments.py
navjotk: Premature optimisation?

navjotk:
> I also still don't understand the reluctance towards "all parameters have providers". It seems to work pretty smoothly?

It was working smoothly without it as well, except for that one line in the yask specific argument handling that I wasn't willing to touch in this PR to minimise the footprint of this PR. If we can do without providers on all parameters, that would be a simpler and hence better solution no?


> There's still some clean up to do, as the TODOs point out. Is this the next step? or is this stuff for the longer term?

Maybe not the immediate next step but soon. 

> For example, the handling of the dle_arguments.

I am aware that can be simplified further but requires a small change in the DLE classes regarding autotuning, which also I left for the future in order to keep this PR small. 

> I also assume that you measure the Python overhead of this new solution and it's basically negligible? ie, from calling op.apply() to the actual execution of the cfunction, the overhead is negligible?

I will measure it and post it here. I assume you mean in comparison to master? In absolute terms I won't be surprised if it is not negligible but it should be the same or less than in master. 

> Does this PR fix #416 , or any other standing issues?

It doesn't yet, but brings us closer. This is a good way along to fixing most arguments issues though, in my opinion.

devito/arguments.py
```diff
@@ -158,198 +143,398 @@ class PtrArgument(Argument):
...
-        # Derived dimensions could be linked through constraints
-        # At this point, a constraint needs to be added that enforces
-        # dim_e - dim_s < SOME_MAX
-        # Also need a default constraint that dim_e > dim_s (or vice-versa)
-        verify = verify and all([a.verify(v, enforce=enforce) for a, v in
-                                 zip(self.rtargs, value)])
-        return verify
+    def visit_Scalar(self, o):
+        arg = ScalarArgument(o.name, o, [], dtype=o.dtype)
+        return arg
```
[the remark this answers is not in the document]
navjotk: Fixed

devito/arguments.py
```diff
@@ -158,198 +143,398 @@ class PtrArgument(Argument):
...
 
+    def visit_Constant(self, o):
+        # TODO: Add option for delayed query of default value
+        arg = ScalarArgument(o.name, o, [ValueDependency(o)], dtype=o.dtype)
```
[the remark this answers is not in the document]
navjotk: Fixed

FabioLuporini:
I see you changed the a few things in the autotuner. Could you check that ``acoustic_example`` and/or ``tti_example`` still take up about the same time as on ``master`` when ran with ``DEVITO_AUTOTUNING=aggressive`` ? When using ``DEVITO_LOGGING=DEBUG``, you should see lots of AT prints.

Then I think we're ready to merge

devito/arguments.py
```diff
@@ -158,198 +143,396 @@ class PtrArgument(Argument):
...
+    @property
+    def dimensions(self):
+        return [x for x in self.parameters if isinstance(x, Dimension)]
 
-class DimensionArgProvider(ArgumentProvider):
 
-    """ This class is used to decorate the Dimension class with behaviour required
-        to handle runtime arguments. All properties/methods defined here are available
-        in any Dimension object.
+class ArgumentVisitor(GenericVisitor):
```
FabioLuporini: I actually got one final question: why is this a tree visitor if you never call ``self.visit(child/children)`` ?

navjotk:
Timings for the following code:
```
from examples.seismic.acoustic.acoustic_example import acoustic_setup
solver = acoustic_setup()
solver.op_fwd()
%timeit solver.op_fwd().arguments()
```
Master:
633 µs ± 34.9 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

This branch:
52.7 ms ± 2.95 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

devito/arguments.py
navjotk: So it is not a tree visitor, I guess. I'll change the docstring on the parent class. Is that all?

FabioLuporini:
> Could you check that acoustic_example and/or tti_example still take up about the same time as on master when ran with DEVITO_AUTOTUNING=aggressive ? When using DEVITO_LOGGING=DEBUG, you should see lots of AT prints.

any news on this?

> Master:
> 633 µs ± 34.9 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
> 
> This branch:
> 52.7 ms ± 2.95 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

The slow down is really, really high... is it worth investigating why? Overall, it appears to be negligible though... If TTI doesn't get incredibly slow, we might accept this. @mlange05 ?

devito/arguments.py
FabioLuporini: > So it is not a tree visitor, I guess. I'll change the docstring on the parent class. Is that all?

Is this equivalent to:
```
for i
  if i is ... do()
  elif i is ... do()
  elif i is ... do()
  else: 
```
 If there is no concept of visit (ie, there is never a tree, or the input is always leaf), then using a Visitor here is really misleading. At least, that's the way I conceived those Visitors.

navjotk:
> Could you check that acoustic_example and/or tti_example still take up about the same time as on master when ran with DEVITO_AUTOTUNING=aggressive ? When using DEVITO_LOGGING=DEBUG, you should see lots of AT prints.

I don't see `DEVITO_AUTOTUNING` in our configuration [dict](https://github.com/opesci/devito/blob/master/devito/parameters.py#L95). Not sure if this even exists in master?

navjotk:
Running the examples takes about the same time though. 
Using:
```
from examples.seismic.acoustic.acoustic_example import run as acoustic_run
from examples.seismic.tti.tti_example import run as tti_run
%timeit acoustic_run()
%timeit tti_run()
```
On master:
Acoustic: 10 s ± 457 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
tti: 8.89 s ± 395 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

This branch:
acoustic: 9.51 s ± 291 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
tti: 9.19 s ± 64.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

The difference seems to be within the margin of measurement error in both cases.

devito/arguments.py
navjotk: The code you suggest is dirtier than the clean separation offered in Visitors. Visitors are not only used for trees but also to encapsulate an operation being performed differently on classes of different types. From the [python docs](http://python-3-patterns-idioms-test.readthedocs.io/en/latest/Visitor.html): 
> The visitor pattern allows you to extend the interface of the primary type by creating a separate class hierarchy of type Visitor to virtualize the operations performed upon the primary type.

FabioLuporini:
DEVITO_AUTOTUNING is a perfectly legal Devito option. Have you tried using it? 
Its implementation is here: https://github.com/opesci/devito/blob/master/devito/core/__init__.py

Can you try running the examples above with AT, ie passing ``-a``? Do they take as long as in master? I'm interested in the number of AT prints and how long each AT run takes (DEVITO_LOGGING=DEBUG will help)

devito/arguments.py
FabioLuporini: I was **not** suggesting, I was only asking to understand.

navjotk:
I think it is very bad design to have configuration options spread across the code base.

navjotk:
This fails: 
```
from devito import configuration
configuration['autotuning'] = 'aggressive'
```

devito/tools.py
navjotk: Do you think it might be a better idea to go back to the original docstring?

FabioLuporini:
> I think it is very bad design to have configuration options spread across the code base.

I disagree. That `autotuning` is a core-specific parameter. 

Also, potentially bad design choices should be detected at PR review time



> This fails:
> 
> from devito import configuration
> configuration['autotuning'] = 'aggressive'

This will not:
configuration.core['autotuning'] = 'aggressive'

devito/tools.py
FabioLuporini: yeah, go for it

navjotk:
> That autotuning is a core-specific parameter.

Doesn't explain why it is in `devito/__init__.py`. Not the first place one would expect configuration options to be. There should either be a backend specific dict next to the main one, or this should be in a backend specific configuration module. 

> Also, potentially bad design choices should be detected at PR review time

Maybe the reviewer should have applied stricter review standards. Or let the PR be open long enough for someone else to take a look. Like in this PR, for example.

FabioLuporini:
> Doesn't explain why it is in devito/__init__.py. Not the first place one would expect configuration options to be. There should either be a backend specific dict next to the main one, or this should be in a backend specific configuration module.

No. It's not there, it's in ``devito/core/__init__.py``. And I think you haven't completely understood how the configuration mechanism works. For example:

> There should either be a backend specific dict next to the main one, or this should be in a backend specific configuration module.

That's how it is.

> Yes I agree that the review standards applied to that PR don't match up to those being applied to this PR. Maybe if it remained open for longer, other people might have had a chance to take a look.

No, the team is small and PR reviewing has to be relatively quick. If it weren't quick, there would have been periods in which I would have ended up stacking up tons of PRs. BTW, most PRs have been open for several days; only the very small ones for 1-2.

navjotk:
> I think you haven't completely understood how the configuration mechanism works

Maybe that's a sign of missing and required documentation.

> configuration.core['autotuning'] = 'aggressive'

This doesn't seem to change anything in the execution of the code I posted above. No autotuning is being performed.

navjotk:
What needs to be done here before merge?

FabioLuporini:
As I said, it'd be very useful to verify the correct functioning of autotuning after your changes to ``autotuning.py``. This is probably the last standing item

navjotk:
As I said above, the code:
```
from examples.seismic.acoustic.acoustic_example import run as acoustic_run
from examples.seismic.tti.tti_example import run as tti_run
from devito import configuration
configuration.core['autotuning'] = 'aggressive'
configuration['log_level'] = 'debug'
%timeit acoustic_run()
%timeit tti_run()
```
does not show any log messages related to autotuning, whether on master or this branch.

FabioLuporini:
have you considered passing ``autotune=True``, or running from the command line with ``-a`` ?

navjotk:
The change that I had made in the autotuner was actually causing other issues that made `tti_example.py` segfault with autotuning enabled so I reverted to using the backdoor arguments that you were using. I guess no performance evaluation necessary now.

FabioLuporini:
> I guess no performance evaluation necessary now.

I don't want a plain new performance evaluation, I only want to be 100% certain that AT is taking up as long as in ``master`` (not more, not less). The fact that you have caught a bug is orthogonal to what I had asked. Please, also consider adding ``python tti_example.py -a`` to the test suite.

navjotk:
There are no changes to the autotuner any more. I don't see why the performance should be different.

FabioLuporini:
> There are no changes to the autotuner any more. I don't see why the performance should be different.

At the top of the files, I seem to remember, you have changed some names. Naming is fundamental in the AT to be sure we execute for a small (predefined) number of timesteps

navjotk:
Names haven't changed. The names of the dimension properties where those names are available have changed. There are mechanisms in place to catch if a wrong name was passed anyway.

FabioLuporini:
We are bypassing apply by calling ``cfunction` directly. When are these checks supposed to happen? 

What I'm asking is a 5 minute thing: you go on master and run with full logs to see how long it roughly takes; then you do the same on your branch.

navjotk:
> We are bypassing op.apply by calling op.cfunction directly. When are these checks supposed to happen?

The `at_arguments` `OrderedDict` contains the arguments already processed by the engine once. If the name of an argument was wrong, it would end up adding a new element to this dict, hence changing the number of arguments passed to the kernel, which would raise an error from `ctype`.

navjotk:
>  to be sure we execute for a small (predefined) number of timesteps

There is a [test](https://github.com/opesci/devito/blob/master/tests/test_autotuner.py#L66) to specifically check what you're talking about.

> What I'm asking is a 5 minute thing

There is an infinite space of measurements that can be made that are irrelevant to this PR but take 5 minutes or less. Since my question is about the relevance to this PR, I think the time it takes to measure is orthogonal to my question.

FabioLuporini:
OK, I'll do that check myself, hopefully tomorrow, worst case on monday, then we'll merge. Nothing else gets merged meanwhile.

navjotk:
I'll do it if you can offer me a scientific reason for why it is relevant here.

FabioLuporini:
nothing scientific: just my experience that it's extremely easy to screw up things in the AT, which is at the same time quite challenging to capture through tests. It happend many times. 
For analogous reasons, we were about to merge an AT-induced segfault....

navjotk:
> For analogous reasons, we were about to merge an AT-induced segfault

There was a change earlier which warranted testing. The change has been reverted now. 

Since `%timeit` is not an option without this running inside a jupyter notebook (where the autotuning option was not working), I used  `time python examples/seismic/tti/tti_example.py -a`:
```
Branch:
real	0m29.572s
user	0m29.396s
sys	0m0.169s

real	0m30.703s
user	0m30.460s
sys	0m0.243s

real	0m29.351s
user	0m29.183s
sys	0m0.172s

real	0m30.044s
user	0m29.885s
sys	0m0.157s

real	0m29.774s
user	0m29.617s
sys	0m0.156s



Master:

real	0m31.141s
user	0m30.907s
sys	0m0.229s

real	0m29.439s
user	0m29.273s
sys	0m0.169s

real	0m29.383s
user	0m29.197s
sys	0m0.188s

real	0m29.322s
user	0m29.173s
sys	0m0.153s

real	0m29.142s
user	0m29.005s
sys	0m0.140s
```

PR #454, opened 2017-12-12
Added slack link to README

navjotk:


mlange05:
Would this make more sense closer to the top? Who reads READMEs until the end anyway? :wink:

FabioLuporini:
> Would this make more sense closer to the top? Who reads READMEs until the end anyway?

Isn't it almost at the top?

navjotk:
It was changed and rebased in response to the comment

mlange05:
Cool, thanks. GTG

PR #455, opened 2017-12-12
CustomCompiler: Small hotfix that correctly derives CC and LD from env

mlange05:


FabioLuporini:
yes, let's merge this as soon as TRAVIS is green (although it shouldn't affect travis at all)

FabioLuporini:
Merged.

PR #457, opened 2017-12-14
Arguments: Adding more tests to cover defaulting and override behaviour

mlange05:
Previous argument tests were primarily focused on testing overrides for the time dimension. This set adds a wider spread across dimensions and explicitly tests dimension sub-ranges separately from size overrides.

The last one in particular seems broken, or at least follows a counter-intuitive convention. I would argue that the notation `op(x=4, y=5)` would intuitively means to traverse the bottom-left sub-domain of shape `(4, 5)`, but what happens in practice is that the variables `x_size` and `y_size`, which are used for data casts in the kernel, are set to the provided values. At the very least we should error if these ranges do not agree with the data provided; better yet we could alias `x` and `y` to `x_e` and `y_e` in the above example to create the described behaviour. Regardless of this, I've marked the test as `xfail` for now.

Similarly, aliasing between `t` and `time` does not work yet, which is left as a `TODO` in one of the tests.

Also, even with those new tests, I don't thing we're covering the argument API particularly well. In anybody has more suggestions as to which specific argument-API features or corner cases should be added, please let me know.


tests/test_operator.py
```diff
@@ -399,40 +519,6 @@ def test_override_composite_data(self):
         arg_name = src1.name + "_coords"
         assert(np.array_equal(args[arg_name], np.asarray((new_coords,))))
 
-    def test_start_end_2d(self):
```
navjotk: Why are these tests being removed?

FabioLuporini:
@mlange05 I haven't checked what's stopping this, but can we bring it to a mergeable state?

mlange05:
@FabioLuporini There were still some flaws and local changes that I found during follow-up work and wanted to fix first. Should be good to go now, although I would like to note that there are probably a few more scenarios we should be testing.

tests/test_operator.py
mlange05: Except for minor notational differences, I believe these are the same a `test_override_function_subranges`. I don't think 2D vs. 3D is significant enough to warrant separate tests (2D is a trivial subset of 3D), and the new test uses the `Grid` interface and space dimensions more naturally.

mlange05: Ping @navjotk

mlange05:
Ok, going in next...

PR #459, opened 2017-12-15
domain, halo, padding

FabioLuporini:
This builds the foundation for the domain allocation switch

I think commit messages and edits are quite self-explanatory.

Having simplified ``yask.Data`` also makes me extremely happy

### DIFF
--- a/devito/function.py
+++ b/devito/function.py
@@ -374,7 +374,10 @@ class Function(TensorFunction):
         Shape of the domain plus the read-only stencil boundary associated
         with this :class:`Function`.
         """
-        return tuple(j + i + k for i, (j, k) in zip(self.shape_domain, self._halo))
+        # TODO: for the domain-allocation switch, this needs to return the shape
+        # of the data including the halo region, ie:
+        # `tuple(j + i + k for i, (j, k) in zip(self.shape_domain, self._halo))`
+        raise NotImplementedError
 
     @property
     def shape_allocated(self):
@@ -383,7 +386,10 @@ class Function(TensorFunction):
         It includes the domain and halo regions, as well as any additional
         padding outside of the halo.
         """
-        return tuple(j + i + k for i, (j, k) in zip(self.shape_with_halo, self._padding))
+        # TODO: for the domain-allocation switch, this needs to return the shape
+        # of the data including the halo and padding regions, ie:
+        # `tuple(j + i + k for i, (j, k) in zip(self.shape_with_halo, self._padding))`
+        raise NotImplementedError
 
     @property
     def space_dimensions(self):
@@ -435,7 +441,9 @@ class Function(TensorFunction):
 
         Elements are stored in row-major format.
         """
-        return self._data
+        # TODO: for the domain-allocation switch, this needs to return all
+        # allocated data values, i.e. self._data
+        raise NotImplementedError
 
     @property
     def laplace(self):
--- a/devito/function.py
+++ b/devito/function.py
@@ -280,7 +280,7 @@ class Function(TensorFunction):
         def wrapper(self):
             if self._data is None:
                 debug("Allocating memory for %s (%s)" % (self.name, self.shape))
-                self._data = Data(self.shape, self.indices, self.space_order, self.dtype)
+                self._data = Data(self.shape, self.indices, self.dtype)
                 if self._first_touch:
                     first_touch(self)
                 else:
@@ -301,7 +301,7 @@ class Function(TensorFunction):
     def _offset_halo(self):
         """
         The number of grid points between the first allocated element
-        (possibly in the halo/padding region) and the first domain element,
+        (possibly in the halo/padding region) and the first halo element,
         for each dimension.
         """
         return self._padding
--- a/devito/function.py
+++ b/devito/function.py
@@ -258,6 +258,20 @@ class Function(TensorFunction):
             dimensions = grid.dimensions
         return dimensions
 
+    def _allocate_memory(self):
+        """Allocate memory in terms of numpy ndarrays."""
+        debug("Allocating memory for %s (%s)" % (self.name, str(self.shape)))
+        self._data_object = Data(self.shape, self.indices, self.space_order, self.dtype)
+        if self._first_touch:
+            first_touch(self)
+        else:
+            self.data.fill(0)
+
+    def initialize(self):
+        """Apply the data initilisation function, if it is not None."""
+        if self.initializer is not None:
+            self.initializer(self.data)
+
     @property
     def shape(self):
         """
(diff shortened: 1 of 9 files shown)


mloubout:
Ok, so I may a missed part of the discussion.

1 -  Yes this is definitely the way to go, cutting out the domain as we currently do is bad (specially for  high orders)

2- I however find it odd and a bit risky to do it in `function` as it is more defined by the stencil range. For some advanced stencils that require to call directly the FD functions instead of the shortcuts, the `halo` may not match the `function` initialization


Just putting it there as it is related to the PR, if I missed something just ignore it.

FabioLuporini:
> I however find it odd and a bit risky to do it in function as it is more defined by the stencil range. For some advanced stencils that require to call directly the FD functions instead of the shortcuts, the halo may not match the function initialization.

We don't know the stencil range until the (first) Operator has been created. So we assume the halo starts off as a function of the space order (symmetric stencil), but if we need more grid points on one or both sides we have three options, I think:

- add a further ``Function`` parameter called for example ``padding``, to let user specify "we need more points on the left/right/both"
- delay memory allocation until the first Operator creation, then set the halo from there via a callback, for example
- reallocate memory each time a new Operator is created and this needs to access more halo points than available

each solution has pros and cons, but this PR doesn't tackle this problem yet. Setting the ``_halo`` based on the ``space_order`` it's just the simplest thing I could do to start with; this also allow all yask tests to pass

> 3 - Is the hallo taken into account in SparseFunction

``SparseFunction`` belongs to the same hierarchy as ``Function``, so it will get a halo, but we don't currently specify its value. In fact, it's very difficult to automatically infer a halo for a ``SparseFunction``. One possibility is to go again through a ``padding`` parameter...

mloubout:
For 1 and 2 OK, for 3, just wanna be sure grid origin position is handled accordingly to halo.

devito/function.py
```diff
@@ -259,43 +275,137 @@ def _indices(cls, **kwargs):
...
+        for each dimension.
         """
-        return tuple(i - s for i, s in zip(self.shape_domain, self.staggered))
+        return tuple(np.add(self._halo, self._padding))
+
+    @property
+    def _offset_halo(self):
+        """
+        The number of grid points between the first allocated element
+        (possibly in the halo/padding region) and the first domain element,
```
mlange05: If I understand this correctly, this should read "the first halo element", right?

devito/function.py
```diff
@@ -259,43 +275,137 @@ def _indices(cls, **kwargs):
             dimensions = grid.dimensions
         return dimensions
 
+    def _allocate_memory(func):
+        """Allocate memory as a :class:`Data`."""
+        def wrapper(self):
```
mlange05: Interesting approach. I would have gone for making `self._data` a `cached_property`, but I guess this works too.

devito/function.py
```diff
@@ -400,22 +521,17 @@ def __init__(self, *args, **kwargs):
                 self.time_size = self.time_order + 1
                 self.indices[0].modulo = self.time_size
 
+            self._halo = (0,) + self._halo
```
mlange05: Is the time-halo not the `time_order` as for all other dimensions?

devito/function.py
FabioLuporini: Oh yes, that works too, but then you cannot initialise ``self._data`` to ``None`` in the constructor (right?) 

I agree a ``cached_property`` would work, but perhaps this makes it more explicit the fact that we are allocating on demand

mlange05: Sure, sounds plausible.

devito/function.py
FabioLuporini: uops, yes, __doc__ needs fixing; I'll do that in #462 to avoid another Travis round

devito/function.py
FabioLuporini: not sure; if that were the case, would you not end up writing over a subset of the halo region? I might be wrong...

mlange05: Ahhh ok, this actually shows a deeper problem! The current setup assumes centered differences when it computes the "halo", which is not true for time (after rearranging the primary equation, we usually have a left-sided halo (`stencil=(2, 0)`). The problem here is that we now do not allocate enough entries (we're missing the two "initial condition" read only entries on the left) - or require the user to account for the fact that we magically drop the halo for the time dimension. So this is actually quite a show-stopper.... :thinking:

FabioLuporini: > we now do not allocate

what do you mean by ``now`` ? The ``_halo`` value is computed, but it's basically unused in this PR. I was just laying the foundation for future work

devito/function.py
```diff
@@ -139,31 +146,40 @@ def __init__(self, *args, **kwargs):
...
 
-            self.space_order = kwargs.get('space_order', 1)
             self.initializer = kwargs.get('initializer', None)
             if self.initializer is not None:
                 assert(callable(self.initializer))
             self._first_touch = kwargs.get('first_touch', configuration['first_touch'])
-            self._data_object = None
+            self._data = None
+
+            self._halo = tuple(self.space_order for i in range(self.dim))
```
FabioLuporini: in fact, this should have been halved, @mlange05

devito/function.py
```diff
@@ -100,12 +102,27 @@ class Function(TensorFunction):
...
+    :param padding: (Optional) allocate extra grid points at a space dimension
+                    boundary. These may be used for non-symmetric stencils
+                    or simply to enforce data alignment. Defaults to 0.
+                    In alternative to an integer, an iterable, indicating
+                    the padding in each dimension, may be passed; in this
+                    case, an error is raised if such iterable has fewer entries
+                    then the number of space dimensions.
     :param dtype: (Optional) data type of the buffered data.
-    :param space_order: Discretisation order for space derivatives
+    :param space_order: Discretisation order for space derivatives. By default,
```
FabioLuporini: worth another look @mlange05

FabioLuporini:
@mlange05 can you take a final look? this *should* now include all comments from before Christmas about the padding related stuff and the 3-tuple notation for ``space_order`` so that one can distinguish between left and right halo/padding for asymmetric stencils

devito/function.py
mlange05: :+1:

devito/function.py
```diff
@@ -259,43 +291,151 @@ def _indices(cls, **kwargs):
...
+
+        Elements are stored in row-major format.
+
+        .. note::
+
+            Alias to ``self.data``.
+        """
+        # TODO: for the domain-allocation switch, this needs to be turned
+        # into a view of the domain region
+        return self._data
```
mlange05: This seems incongruent; it looks here like the switch has already happened, although `self.data_allocated` points to the same object. In contrast, the `self.shape_domain` and `self.shape_allocated` do not return the same values, so one of the two cannot fit the returned `._data` object (missing test?).

Should we maybe error out on all "domain" properties (shape/data) and default to "allocated" throughout until we switch completely?

FabioLuporini: oh yes, I only "anticipated" the implementation of ``shape_allocated`` etc assuming nobody will use them until the switch. I can definitely make everything more congruent, I think

FabioLuporini: so here's my final approach: ATM, "everything is domain". Trying to access halo/padding properties will raise a ``NotImplementedError``. The infrastructure for the switch is there though (suitably commented out and with useful docstrings)

mlange05: Great, thanks. :+1:

mlange05:
All addressed. GTG from me.

FabioLuporini:
Thanks. Will merge once Travis is green again.

FabioLuporini:
Merged.

PR #461, opened 2017-12-18
Memoization fix and misc

FabioLuporini:
Our memoizer caches at the class level; for the seismic wrappers, instead, it'd be better to cache at the instance level. Otherwise, these are unusable from ``benchmark.py`` as we would start leaking memory. This PR mainly introduces a new memoizer suitable for instance-level caching. It also ships a number of minor fixes


FabioLuporini:
Same story, @mloubout : I'd like you to check the FWI notebooks with this branch.

PR #462, opened 2017-12-20
Some consolidation

FabioLuporini:
* Our memoizer caches at the class level; for the seismic wrappers, instead, it'd be better to cache at the instance level. Otherwise, these are unusable from benchmark.py as we would start leaking memory. This PR introduces a new memoizer suitable for instance-level caching. 
* Better compilation infrastructure
* ``benchmark.py`` tweaks; in particular, better memory clean up, use preemptive AT in yask mode
* Allow per-backend specialisation of caching clean up
* Fix up potential memory leak in YASK

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,10 +71,6 @@ before_script:
     fi
 
 script:
-  - cat /proc/cpuinfo
-  - lscpu
-  - gcc -march=native -Q --help=target
-  - gcc -march=native -Q --help=target | grep march
   - flake8 --builtins=ArgumentError .
   - py.test -vs tests/
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,6 +71,10 @@ before_script:
     fi
 
 script:
+  - cat /proc/cpuinfo
+  - lscpu
+  - gcc -march=native -Q --help=target
+  - gcc -march=native -Q --help=target | grep march
   - flake8 --builtins=ArgumentError .
   - py.test -vs tests/
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ matrix:
       env: DEVITO_ARCH=gcc-4.9 DEVITO_OPENMP=1 OMP_NUM_THREADS=2
     - os: linux
       python: "3.6"
-      env: DEVITO_ARCH=gcc-4.9 DEVITO_OPENMP=0 DEVITO_BACKEND=yask
+      env: DEVITO_ARCH=gcc-7 DEVITO_OPENMP=0 DEVITO_BACKEND=yask
   allow_failures:
     - os: linux
       python: "2.7"
@@ -33,12 +33,14 @@ matrix:
 addons:
   apt:
     sources:
-      - ubuntu-toolchain-r-test     # For gcc 4.9 and 5
+      - ubuntu-toolchain-r-test     # For gcc 4.9, 5 and 7
     packages:
       - gcc-4.9
       - g++-4.9
       - gcc-5
       - g++-5
+      - gcc-7
+      - g++-7
 
 before_install:
   # Setup anaconda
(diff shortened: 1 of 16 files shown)


.travis.yml
```diff
@@ -21,7 +21,7 @@ matrix:
       env: DEVITO_ARCH=gcc-4.9 DEVITO_OPENMP=1 OMP_NUM_THREADS=2
     - os: linux
       python: "3.6"
-      env: DEVITO_ARCH=gcc-4.9 DEVITO_OPENMP=0 DEVITO_BACKEND=yask
+      env: DEVITO_ARCH=gcc-7 DEVITO_OPENMP=0 DEVITO_BACKEND=yask
```
mlange05: Should we update the above `gcc-5` builder to `gcc-7` too?

devito/compiler.py
```diff
@@ -257,7 +271,8 @@ def make(loc, args):
                 log.write(" ".join(command))
                 log.write("\n\n")
                 try:
-                    subprocess.check_call(command, stderr=err, stdout=log)
+                    #subprocess.check_call(command, stderr=err, stdout=log)
```
mlange05: Debug leftover?

devito/yask/__init__.py
```diff
@@ -91,21 +88,12 @@ def switch_cpu(develop_mode):
...
-                  'skylake': 'skx',
-                  'knc': 'knc', 'knl': 'knl'}
-        if platform in mapper.values():
-            configuration['platform'] = platform
-        elif platform in mapper:
-            configuration['platform'] = mapper[platform]
-        else:
-            exit("Unknown platform `%s` to run in optimized mode" % platform)
+        configuration['platform'] = os.environ.get('DEVITO_PLATFORM', platform)
+        print('AAAAAAAAAAAAAAAAa', isa, platform)
```
mlange05: Debug leftover?

devito/yask/__init__.py
```diff
@@ -91,21 +88,12 @@ def switch_cpu(develop_mode):
...
     else:
-        configuration['isa'], configuration['platform'] = 'cpp', 'intel64'
+        configuration['isa'] = 'cpp'
+        configuration['platform'] = 'intel64'
+        print('BBBBBBBBBBBBBBBBBB')
```
mlange05: Debug leftover?

devito/compiler.py
FabioLuporini: I'm experiencing a weird, YASK-related (ie, not involving devito at all) failure on Travis only. Those are just there because I had to hand an output msg to Chuck to investigate. Will drop them all and perform a final clean up pass before merging

.travis.yml
FabioLuporini: so the only reason I switched to 7 is that I was worried that Travis, or any future CI platform we'll use, would give us a skylake (or newer) architecture; the YASK backend runs some tests with vectorisation, and the architecture autodetection mechanism, based on gcc, won't work if gcc is too old and the architecture is too new. Look at ``infer_cpu`` for more info

Now, to answer your question: I don't think it's too bad to test with a number of gcc versions. It's true it's not consistent, but still it gives us wider coverage. Debatable, though

.travis.yml
```diff
@@ -65,11 +67,14 @@ before_script:
...
-      cd yask; git checkout develop;
-      make compiler && make compiler-api; pip install -e .; cd ../devito;
+      cd yask; make compiler && make compiler-api; pip install -e .; cd ../devito;
     fi
 
 script:
+  - cat /proc/cpuinfo
+  - lscpu
+  - gcc -march=native -Q --help=target
+  - gcc -march=native -Q --help=target | grep march
```
FabioLuporini: @mlange05 I'm also dropping this stuff

FabioLuporini:
> I don't fully understand the difference in caching decorators (yet)

in the implementation or in the spec?

.travis.yml
mlange05: Good. I was about to ask....

mlange05:
Implementation. I've not read the external docs in detail yet, though...

FabioLuporini:
OK, this goes in first.

FabioLuporini:
Merged.

Issue #460, opened 2017-12-15
`inspection.count` slow; Large number of dictionary keys from complex stencil.


gamdow:
The following script generates a stencil with a very large number of operations. The resulting opearator takes approximately 45mins to generate;

https://github.com/gamdow/devito/blob/master/slow-count-performance/DevitoRK.py

According to the following profile data and call graph, the majority (61%) of time in spent in the `inspection.count` function;

https://github.com/gamdow/devito/blob/master/slow-count-performance/myLog.profile
https://github.com/gamdow/devito/blob/master/slow-count-performance/callingGraph.dot (view with xdot)

This is reasonable given the number of dictionary keys, but could use a optimization pass.

![callgraph](https://user-images.githubusercontent.com/22518029/34050870-5ea7da54-e1b4-11e7-97c7-632b6d2f4056.png)

gamdow:
I've had some success optimising `count` using a Counter dictionary (see above pull request). Tried adding caching, but the number of repeated calls with the same arguments is minimal (~5%). Though, obviously could be useful for repeated invocations of the code generation with similar stencils. I think this issue should be approached from a higher level, or failing that, some kind of maximum recursion depth thing should be added?

`indexify` was a good candidate for caching, with 1000s of repeated function calls on the same object, but I got lost in the symbol class heirachy trying to identify when a symbol object had been reused (its indexing had changed?) and it's cache needed clearing. Maybe something for you guys to consider.

FabioLuporini:
> some kind of maximum recursion depth thing should be added?

This is actually very interesting and I'd be curious to know how much could we lose if we did that. I might try something at some point

How big does the dictionary get in the pathological scenario above? How  deep is the recursion level?

FabioLuporini:
Thinking about it a bit more, I'm now surprised that we're spending that much time within ``__eq__``. If I remember correctly, when inserting into a dict you call ``__eq__`` to compare keys only when their hash is identical. So there are two things here:
- is this expected -- ie, do we really have that many hits?
- if these are "real" hits, we could think of using a probabilistic data structure that, unlike dict, only uses ``__hash__``, but not ``__eq__``. Clearly I'm now talking about a relaxed version of ``count()``, for which we tolerate mistakes in the returned count. As long as the probability of a mistake is low, we would be OK, since such a ``count(relaxed=True)`` would only be used as part of a cost model for common sub-expressions elimination (thus not affecting correctness).

FabioLuporini:
[...]

PR #466, opened 2018-01-08
Optimisation of count

gamdow:
Around 33% speed up over for-loop for large stencils (#460).


devito/symbolics/inspection.py
```diff
@@ -7,18 +7,14 @@
...
 from devito.tools import flatten
 
-
 def count(exprs, query):
     """
     Return a mapper ``{(k, v)}`` where ``k`` is a sub-expression in ``exprs``
     matching ``query`` and ``v`` is the number of its occurrences.
     """
-    mapper = OrderedDict()
+    mapper = {}
```
FabioLuporini: Random guess.
Should this be a ``Counter`` too -- thus explaining why some tests are failing? The failures are due to capturing fewer common sub-expressions, which results in higher flop counts than what the tests are expecting. 

It's like with this solution, we get different ``hit``s here:
https://github.com/opesci/devito/blob/master/devito/dse/manipulation.py#L110
A ``hit`` is a candidate common sub-expression that gets replaced by a temporary

gamdow:
You were right about the mapper needing to be a Counter.

FabioLuporini:
Perhaps not an issue if you return a ``Counter``, rather than a ``dict``: if the primary concern was speed, returning a new ``dict`` will cause lots of key insertions, thus increasing the routine time cost? just a random thought.

Are the tests now passing?

Thanks

FabioLuporini:
@gamdow just a flake8 issue (two consecutive blank lines at ``inspection.py:14``) and then this is probably good to go.

What's the ultimate speed improvement after switching to ``Counter``?

FabioLuporini:
@gamdow I've updated the branch to latest ``master``, still keen to fix the remaining ``flake8`` issues? After that, we can land this

gamdow:
What's CodeCov? I've never come across it before and the website is just a list of meaningless slogans.

FabioLuporini:
@gamdow all good, thanks

PR #467, opened 2018-01-09
Fix conditional examples execution

mlange05:
Turns out my previous laziness and lack of understanding/patience for `bash` was causing individually executed examples to fail silently. This hotfix fixes this oversight and the missing import for the `foreign` backend that brought this to light.

Many apologies for the inconvenience caused - I guess `bash` and I will never be friends. :frowning:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,16 +75,14 @@ script:
   - py.test -vs --cov devito tests/
 
   # Additional seismic operator tests
-  - if [[ $DEVITO_BACKEND != 'yask' ]]; then
-      DEVITO_BACKEND=foreign py.test -vs tests/test_operator.py -k TestForeign;
-      python examples/seismic/benchmark.py test -P tti -so 4 -a -d 20 20 20 -n 5;
-      python examples/seismic/benchmark.py test -P acoustic -a;
-      python examples/seismic/acoustic/acoustic_example.py --full;
-      python examples/seismic/acoustic/acoustic_example.py --constant --full;
-      python examples/seismic/acoustic/gradient_example.py;
-      python examples/seismic/tti/tti_example.py -a;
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -75,15 +75,13 @@ script:
...
-  - if [[ $DEVITO_BACKEND != 'yask' ]]; then
-      DEVITO_BACKEND=foreign py.test -vs tests/test_operator.py -k TestForeign;
-      python examples/seismic/benchmark.py test -P tti -so 4 -a -d 20 20 20 -n 5;
-      python examples/seismic/benchmark.py test -P acoustic -a;
-      python examples/seismic/acoustic/acoustic_example.py --full;
-      python examples/seismic/acoustic/acoustic_example.py --constant --full;
-      python examples/seismic/acoustic/gradient_example.py;
-      python examples/checkpointing/checkpointing_example.py;
-    fi
+  - if [[ $DEVITO_BACKEND != 'yask' ]]; then DEVITO_BACKEND=foreign py.test -vs tests/test_operator.py -k TestForeign; fi
```
FabioLuporini: I understand you're not a bash friend (me neither), but can you put all these statements within the same scope? Otherwise this is a mess

mlange05: I'm not sure I understand this right, but wasn't the scoping the reason for the silent failures?  If that is not the type of scoping you mean, can you please provide an example?

Also, I'm not sure I agree this is a mess, since explicit example runs should be used sparingly at best. As such, an individual invocation per example doesn't seem too bad to me (personal opinion of course).

FabioLuporini: please, forget my BS comment

mlange05:
Ok, merging...

PR #468, opened 2018-01-10
Cleanup, fixes, and additions before domain-allocation switch (PART 1)

FabioLuporini:
Misc improvements and fixes. Worth noting:

* dropping lots of unused code (DLE's padding and its dependencies)
* Vastly better ``types`` hierarchy and refreshed documentation
* Better organisation of the IR 
* Add ``Interval`` and ``IterationSpace``, to be used in a subsequent PR. Tests attached

@mloubout this will fail on ``test_save.py``. I'll need your hotfix PR

### DIFF
--- a/devito/function.py
+++ b/devito/function.py
@@ -179,22 +179,22 @@ class Function(TensorFunction):
             space_order = kwargs.get('space_order', 1)
             if isinstance(space_order, int):
                 self.space_order = space_order
-                self._halo = tuple((ceil(space_order/2),)*2 for i in range(self.dim))
+                self._halo = tuple((ceil(space_order/2),)*2 for i in range(self.ndim))
             elif isinstance(space_order, tuple) and len(space_order) == 3:
                 space_order, left_points, right_points = space_order
                 self.space_order = space_order
-                self._halo = tuple((left_points, right_points) for i in range(self.dim))
+                self._halo = tuple((left_points, right_points) for i in range(self.ndim))
             else:
                 raise ValueError("'space_order' must be int or 3-tuple of ints")
 
             padding = kwargs.get('padding', 0)
             if isinstance(padding, int):
-                self._padding = tuple((padding,)*2 for i in range(self.dim))
-            elif isinstance(padding, tuple) and len(padding) == self.dim:
+                self._padding = tuple((padding,)*2 for i in range(self.ndim))
+            elif isinstance(padding, tuple) and len(padding) == self.ndim:
                 self._padding = tuple((i,)*2 if isinstance(i, int) else i
                                       for i in padding)
             else:
-                raise ValueError("'padding' must be int or %d-tuple of ints" % self.dim)
+                raise ValueError("'padding' must be int or %d-tuple of ints" % self.ndim)
 
             # Dynamically add derivative short-cuts
             self._initialize_derivatives()
@@ -435,7 +435,7 @@ class Function(TensorFunction):
         raise NotImplementedError
 
     @property
-    def dim(self):
+    def ndim(self):
         """Return the number of spatial dimensions."""
         return len(self._grid_shape_domain)
 
@@ -467,7 +467,7 @@ class Function(TensorFunction):
         """
         derivs = tuple('d%s2' % d.name for d in self.space_dimensions)
 
-        return sum([getattr(self, d) for d in derivs[:self.dim]])
+        return sum([getattr(self, d) for d in derivs[:self.ndim]])
 
     def laplace2(self, weight=1):
         """
--- a/devito/function.py
+++ b/devito/function.py
@@ -11,7 +11,7 @@ from devito.logger import debug, error, warning
 from devito.data import Data, first_touch
 from devito.cgen_utils import INT, FLOAT
 from devito.dimension import Dimension, TimeDimension
-from devito.types import SymbolicFunction, AbstractSymbol
+from devito.types import SymbolicFunction, AbstractCachedSymbol
 from devito.finite_difference import (centered, cross_derivative,
                                       first_derivative, left, right,
                                       second_derivative, generic_derivative,
@@ -44,7 +44,7 @@ Forward = TimeAxis('Forward')
 Backward = TimeAxis('Backward')
 
 
-class Constant(AbstractSymbol):
+class Constant(AbstractCachedSymbol):
 
     """
     Symbol representing constant values in symbolic equations.
@@ -710,7 +710,7 @@ class SparseFunction(CompositeFunction):
                 raise ValueError('No grid provided for SparseFunction.')
 
             # Allocate and copy coordinate data
-            d = Dimension('d')
+            d = Dimension(name='d')
             self.coordinates = Function(name='%s_coords' % self.name,
                                         dimensions=[self.indices[-1], d],
                                         shape=(self.npoint, self.grid.dim))
@@ -735,7 +735,8 @@ class SparseFunction(CompositeFunction):
         dimensions = kwargs.get('dimensions', None)
         grid = kwargs.get('grid', None)
         nt = kwargs.get('nt', 0)
-        indices = [grid.time_dim, Dimension('p')] if nt > 0 else [Dimension('p')]
+        dim = Dimension(name='p')
+        indices = [grid.time_dim, dim] if nt > 0 else [dim]
         return dimensions or indices
 
     @property
--- a/devito/function.py
+++ b/devito/function.py
@@ -341,11 +341,6 @@ class Function(TensorFunction):
             self.initializer(self.data)
 
     @property
-    def dim(self):
-        """Return the number of spatial dimensions."""
-        return len(self._grid_shape_domain)
-
-    @property
     def shape(self):
         """
         Shape of the domain associated with this :class:`Function`.
@@ -389,11 +384,6 @@ class Function(TensorFunction):
         # of the data including the halo and padding regions, ie:
         # `tuple(j + i + k for i, (j, k) in zip(self.shape_with_halo, self._padding))`
         raise NotImplementedError
-
-    @property
-    def space_dimensions(self):
-        """Tuple of index dimensions that define physical space."""
-        return tuple(d for d in self.indices if d.is_Space)
 
     @property
     def data(self):
@@ -445,6 +435,31 @@ class Function(TensorFunction):
         raise NotImplementedError
 
     @property
+    def dim(self):
+        """Return the number of spatial dimensions."""
+        return len(self._grid_shape_domain)
+
+    @property
+    def dimensions(self):
+        """Tuple of :class:`Dimension`s representing the function indices."""
+        return self.indices
+
+    @property
+    def space_dimensions(self):
+        """Tuple of :class:`Dimension`s that define physical space."""
+        return tuple(d for d in self.indices if d.is_Space)
+
+    @property
+    def symbolic_shape(self):
+        """
+        Return the symbolic shape of the object. This is simply the
+        appropriate combination of symbolic dimension sizes shifted
+        according to the ``staggered`` mask.
+        """
+        return tuple(i.symbolic_size - s for i, s in
+                     zip(self.indices, self.staggered))
+
+    @property
     def laplace(self):
         """
         Generates a symbolic expression for the Laplacian, the second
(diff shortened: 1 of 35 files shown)


FabioLuporini:
Note: a major reason you lots of additions and deletions is the refreshed ``__doc__``s !

devito/function.py
```diff
@@ -335,16 +331,6 @@ def _extent_padding_left(self):
         """
         return self._offset_halo
 
-    def initialize(self):
```
navjotk: I don't think this should be removed

FabioLuporini: why? 
unused and basically useless => drop

examples/seismic/benchmark.py
```diff
@@ -273,15 +269,22 @@ def plot(problem, **kwargs):
...
-            run = (key["dse"], key["dle"])
-            label = "<%s,%s>" % run
-            oi_loc = 0.05 if len(str(key["space_order"])) == 1 else 0.06
-            oi_annotate = {'s': 'SO=%s' % key["space_order"],
-                           'size': 4, 'xy': (oi_value, oi_loc)} if run[0] else None
+        for k, v in gflopss.items():
+            so = dict(k)['space_order']
+
+            oi_value = oi[k]
+            time_value = time[k]
```
navjotk: Couldn't these two just be inlined in the function call later? It would be more concise that way.

tests/test_derivatives.py
```diff
@@ -1,44 +1,65 @@
...
+from conftest import skipif_yask
 from sympy import Derivative, simplify
 
 from devito import Grid, Function, TimeFunction
 
 
 @pytest.fixture
-def shape(xdim=20, ydim=30):
-    return (xdim, ydim)
+def shape(xdim=20, ydim=20, zdim=20):
```
navjotk: Reason for changing `ydim`? There are merits to having non square/cube domains in tests.

tests/test_derivatives.py
```diff
@@ -1,44 +1,65 @@
...
 @skipif_yask
-@pytest.mark.parametrize('SymbolType, dimension', [
+@pytest.mark.parametrize('SymbolType, dim', [
     (Function, x), (Function, y),
     (TimeFunction, x), (TimeFunction, y), (TimeFunction, t),
 ])
-def test_stencil_derivative(grid, shape, SymbolType, dimension):
+def test_stencil_derivative(grid, shape, SymbolType, dim):
     """Test symbolic behaviour when expanding stencil derivatives"""
+    i = dim(grid)  # Known pytest issue: cannot promptly use fixtures with parametrize
```
navjotk: What is the issue? Maybe leave a link in comments?

tests/test_operator.py
```diff
@@ -926,7 +926,7 @@ def test_equations_mixed_densedata_timedata(self, shape, dimensions):
         """
         grid = Grid(shape=shape, dimensions=dimensions, time_dimension=time)
         a = TimeFunction(name='a', grid=grid, time_order=2, space_order=2)
-        p_aux = Dimension(name='p_aux', size=10)
+        p_aux = Dimension(name='p_aux')
```
navjotk: Do we have a way of erroring out on extra arguments to the constructors? I suspect this was found by eyeballing the code but the presence of a `size` argument was silently ignored during execution?

tests/test_save.py
```diff
@@ -16,20 +16,14 @@ def initial(dx=0.01, dy=0.01):
...
     dx2, dy2 = dx**2, dy**2
     dt = dx2 * dy2 / (2 * a * (dx2 + dy2))
 
     grid = Grid(shape=(nx, ny))
-    u = TimeFunction(
-        name='u', grid=grid, save=timesteps, initializer=initializer,
-        time_order=1, space_order=2
-    )
+    u = TimeFunction(name='u', grid=grid, save=timesteps, time_order=1, space_order=2)
+    u.data[0, :] = initial()
```
navjotk: Repeating: Initializers are good. Especially if you're reading in from a file.

devito/function.py
navjotk: It was being used in tests that you modify in this same PR. As I mention, they are quite useful when you're using an external file reader to read in the data. They are also useful in ensuring that the memory isn't allocated too early in the workflow only to specify initialisation. With initializers we can delay memory allocation as much as we want. IIRC Yask has a similar concept.

FabioLuporini: I didn't get the comment about delayed memory allocation (memory is lazily allocated already), but fine, I can revert it

devito/function.py
```diff
@@ -444,6 +425,31 @@ def data_allocated(self):
         # allocated data values, i.e. self._data
         raise NotImplementedError
 
+    @property
+    def dim(self):
```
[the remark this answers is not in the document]
FabioLuporini: right, I'll use ``ndim`` (like numpy)

devito/function.py
```diff
@@ -730,7 +726,8 @@ def _indices(cls, **kwargs):
         dimensions = kwargs.get('dimensions', None)
         grid = kwargs.get('grid', None)
         nt = kwargs.get('nt', 0)
-        indices = [grid.time_dim, Dimension('p')] if nt > 0 else [Dimension('p')]
+        dim = Dimension(name='p')
+        indices = [grid.time_dim, dim] if nt > 0 else [dim]
```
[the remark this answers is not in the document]
FabioLuporini: I haven't changed the semantics of that line, only added a temporary variable to make it fit in the 90 characters limit

devito/symbolics/inspection.py
```diff
@@ -110,3 +114,32 @@ def access(symbol):
...
+    for i, constraint in enumerate(list(constraints)):
+        normalized = []
+        for j in constraint:
+            found = [d for d in j.free_symbols if isinstance(d, Dimension)]
+            normalized.extend([d for d in found if d not in normalized])
+        constraints[i] = normalized
+    ordering = partial_order(constraints)
+
+    # Add any leftover free dimensions (not an Indexed' index)
+    dimensions = [i for i in expr.free_symbols if isinstance(i, Dimension)]
```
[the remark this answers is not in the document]
FabioLuporini: I think at this stage some symbols could be actual ``sympy.Symbol``. I'll check

devito/function.py
navjotk: Lazy allocation is pointless if initialisation triggers memory allocation. With the initializer in place, memory allocation is deferred further until just before the operator is ready to execute.

tests/test_derivatives.py
FabioLuporini: OK

tests/test_derivatives.py
FabioLuporini: the comment already explains it: you can't use fixtures within ``parametrize``

tests/test_operator.py
FabioLuporini: > I suspect this was found by eyeballing the code but the presence of a size argument was silently ignored during execution?

right

> Do we have a way of erroring out on extra arguments to the constructors

no, but I agree it would be useful to come up with something

tests/test_save.py
FabioLuporini: OK, I'll push them back

tests/test_derivatives.py
navjotk: A link to the issue tracker on pytest's website would still be useful.

devito/ir/support/domain.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def _construct(self, intervals):
+        return Space(intervals)
+
+    @property
+    def empty(self):
+        return len(self.intervals) == 0
+
+    @classmethod
+    def intersection_update(self, *spaces):
```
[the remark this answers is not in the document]
FabioLuporini: 👍

devito/symbolics/inspection.py
```diff
@@ -110,3 +114,32 @@ def access(symbol):
...
+    in which they are encountered when visiting ``expr``.
+
+    :param expr: The :class:`sympy.Eq` from which the :class:`Dimension`s are
+                 extracted. They can appear both as array indices or as free
+                 symbols.
+    :param key: A callable used as key to enforce a final ordering.
+    """
+    # Get the Indexed dimensions, in appearance order
+    indexeds = retrieve_indexed(expr, mode='all')
+    constraints = [tuple(i.indices) for i in indexeds]
```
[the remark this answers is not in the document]
FabioLuporini: 👍

examples/seismic/benchmark.py
FabioLuporini: 👍

devito/function.py
FabioLuporini: > Lazy allocation is pointless if initialisation triggers memory allocation.

Assuming that by "initialisation" you mean instantiation of the object, then memory is not allocated straight away; it is allocated the first time one tries to access ``.data``. 

That's why I was droppining the initialiser:
```
f = Function(... initialier=...)
```
is identical to
```
f = Function(...)
f.data[...] = ...
```

FabioLuporini: > memory allocation is deferred further until just before the operator is ready to execute.

Redundant comment: memory allocation is not strictly related to an Operator per se, but rather to the first access to ``.data``. The fact that you mention an Operator is because clearly an Operator has to access ``.data``

navjotk: No. By initialisation, I mean initialisation, i.e. I instantiate by `u = TimeData()`, and I initialise with `u.data[:] = ...`. With the initializer, I can provide the initialisation _routine_ without calling it, so it can be called later in the workflow, hence making memory allocation lazier - this way we hold memory for less time.

FabioLuporini: oh yes, in that sense I agree. As I said somewhere else, I'll put ``initialiser`` back to its place

devito/grid.py
```diff
@@ -83,11 +83,11 @@ def __init__(self, shape, extent=None, origin=None, dimensions=None,
...
         # Store or create default symbols for time and stepping dimensions
         if time_dimension is None:
-            self.time_dim = TimeDimension('time', spacing=Constant(name='dt'))
-            self.stepping_dim = SteppingDimension('t', parent=self.time_dim)
+            self.time_dim = TimeDimension(name='time', spacing=Constant(name='dt'))
+            self.stepping_dim = SteppingDimension(name='t', parent=self.time_dim)
         else:
             self.time_dim = time_dimension
-            self.stepping_dim = SteppingDimension('%s_s' % time_dimension.name,
+            self.stepping_dim = SteppingDimension(name='%s_s' % time_dimension.name,
```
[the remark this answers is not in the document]
FabioLuporini: OK. Bear in mind, however, that ``grid.py`` needs more clean up in general (not in this PR).

I'm raising an exception, rather than an assert, as grids are supposed to be created by users.

devito/ir/support/domain.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+class Interval(AbstractInterval):
+
+    """
+    Interval(dim, lower, upper)
+
+    Create an :class:`Interval` of extent: ::
+
+        dim.extent + abs(upper - lower)
```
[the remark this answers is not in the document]
FabioLuporini: TBH, I'm not sure what else I could write here

examples/seismic/benchmark.py
FabioLuporini: Oh, now I remember. The reason I prefer to make it explicit is that the suffix ``_value`` helps me remembering that those are the (x,y) **values** in the plot. ``oi[k]`` and ``time[k]`` aren't very informative

tests/test_derivatives.py
FabioLuporini: ok

FabioLuporini:
Pushed a number of fixes addressing @navjotk 's comments

devito/symbolics/inspection.py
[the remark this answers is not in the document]
navjotk: Update here?

navjotk:
GTG

devito/symbolics/inspection.py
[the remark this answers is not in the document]
FabioLuporini: haven't tried; will try in #469 (clean up PART 2)

FabioLuporini:
Merging...

FabioLuporini:
Merged.

PR #469, opened 2018-01-11
Rethink loop generation before domain-allocation switch (PART 2)

FabioLuporini:
~The diff is huge, but mainly because this PR needs #468 to go in first. I'm filing it just to show progress is being made.~

EDIT: Ready for review

OK, so, this took a while because I basically had to rethink (simplify) and cleanup the process that lowers equations down to loops. Main features of the PR are

- ``ir.Eq`` is now a first-class citizen in the compiler; it represents a "lowered equation". It is attached an ``IterationSpace``
- ``clusterize`` now groups ``ir.Eq``s and doesn't need stencils anymore; the required information is carried by the equations themselves, through ``IterationSpace``.
- ``Stencil`` is now less powerful, as ``IterationSpace`` does (in a much better and cleaner way) most of the job. 
- an ``IterationSpace`` carries ``sub_iterators``. In essence, a sub_iterator is a mapper from "main" to "derived" dimensions used within the iteration space. 
- ``ResolveIterationVariable`` has been dropped. The used-to-be ``_schedule_expressions``, now moved to the free function ``iet_make`` within ``ir/iet``, does all the job. This is vastly better then before, as now we *construct the loops as soon as we have the necessary information, rather than building them once and then applying a Transformer visitor (``ResolveIterationVariable``) to them*. In particular, we exploit the ``sub_iterator`` information carried by the clusters' iteration spaces. Eventually, this makes adding/supporting further derived dimensions much simpler (@navjotk ), as now we don't need stupid dimension mappers all over the place anymore

### DIFF
--- a/devito/dle/backends/advanced.py
+++ b/devito/dle/backends/advanced.py
@@ -158,9 +158,9 @@ class DevitoRewriter(BasicRewriter):
                 dim = blocked.setdefault(i, Dimension(name=name))
                 block_size = dim.symbolic_size
                 iter_size = i.dim.symbolic_extent
-                start = i.limits[0] - i.offsets[0]
-                finish = i.dim.symbolic_end - i.offsets[1]
-                innersize = iter_size - (-i.offsets[0] + i.offsets[1])
+                start = i.limits[0] + i.offsets[0]
+                finish = i.dim.symbolic_end + i.offsets[1]
+                innersize = iter_size + (-i.offsets[0] + i.offsets[1])
                 finish = finish - (innersize % block_size)
                 inter_block = Iteration([], dim, [start, finish, block_size],
                                         properties=PARALLEL)
@@ -177,7 +177,7 @@ class DevitoRewriter(BasicRewriter):
                 # This will be used for remainder loops, executed when any
                 # dimension size is not a multiple of the block size.
                 start = inter_block.limits[1]
-                finish = i.dim.symbolic_end - i.offsets[1]
+                finish = i.dim.symbolic_end + i.offsets[1]
                 remainder = i._rebuild([], limits=[start, finish, 1], offsets=None)
                 remainders.append(remainder)
 
--- a/devito/ir/equations/equation.py
+++ b/devito/ir/equations/equation.py
@@ -1,0 +1,66 @@
+from collections import OrderedDict
+
+import sympy
+
+from devito.ir.support import Interval, IterationSpace, Stencil
+from devito.symbolics import dimension_sort, indexify
+
+__all__ = ['Eq']
+
+
+class Eq(sympy.Eq):
+
+    """
+    A new SymPy equation with an associated iteration space.
+
+    All :class:`Function` objects within ``expr`` get indexified and thus turned
+    into objects of type :class:`types.Indexed`.
+
+    An iteration space is an object of type :class:`Space`. It represents the
+    data points accessed by the equation along each :class:`Dimension`. The
+    :class:`Dimension`s are extracted directly from the equation.
+    """
+
+    def __new__(cls, input_expr, subs=None):
+        # Sanity check
+        assert isinstance(input_expr, sympy.Eq)
+
+        # Indexification
+        expr = indexify(input_expr)
+
+        # Apply caller-provided substitution
+        if subs is not None:
+            expr = expr.xreplace(subs)
+
+        expr = super(Eq, cls).__new__(cls, expr.lhs, expr.rhs, evaluate=False)
+        expr.is_Increment = getattr(input_expr, 'is_Increment', False)
+
+        # Get the accessed data points
+        stencil = Stencil(expr)
+
+        # Well-defined dimension ordering
+        ordering = dimension_sort(expr, key=lambda i: not i.is_Time)
+
+        # Split actual Intervals (the data spaces) from the "derived" iterators,
+        # to build an IterationSpace
+        iterators = OrderedDict()
+        for i in ordering:
+            if i.is_Derived:
+                iterators.setdefault(i.parent, []).append(stencil.entry(i))
+            else:
+                iterators.setdefault(i, [])
+        intervals = []
+        for k, v in iterators.items():
+            offs = set.union(set(stencil.get(k)), *[i.ofs for i in v])
+            intervals.append(Interval(k, min(offs), max(offs)).negate())
+        expr.ispace = IterationSpace(intervals, iterators)
+
+        return expr
+
+    @property
+    def is_Scalar(self):
+        return self.lhs.is_Symbol
+
+    @property
+    def is_Tensor(self):
+        return self.lhs.is_Indexed
--- a/devito/ir/equations/equation.py
+++ b/devito/ir/equations/equation.py
@@ -1,14 +1,14 @@
 from collections import OrderedDict
 
-import sympy
+from sympy import Eq
 
 from devito.ir.support import Interval, IterationSpace, Stencil
 from devito.symbolics import dimension_sort, indexify
 
-__all__ = ['Eq']
+__all__ = ['LoweredEq']
 
 
-class Eq(sympy.Eq):
+class LoweredEq(Eq):
 
     """
     A new SymPy equation with an associated iteration space.
@@ -23,7 +23,8 @@ class Eq(sympy.Eq):
 
     def __new__(cls, input_expr, subs=None):
         # Sanity check
-        assert isinstance(input_expr, sympy.Eq)
+        assert type(input_expr) != LoweredEq
+        assert isinstance(input_expr, Eq)
 
         # Indexification
         expr = indexify(input_expr)
@@ -32,7 +33,7 @@ class Eq(sympy.Eq):
         if subs is not None:
             expr = expr.xreplace(subs)
 
-        expr = super(Eq, cls).__new__(cls, expr.lhs, expr.rhs, evaluate=False)
+        expr = super(LoweredEq, cls).__new__(cls, expr.lhs, expr.rhs, evaluate=False)
         expr.is_Increment = getattr(input_expr, 'is_Increment', False)
 
         # Get the accessed data points
--- a/devito/ir/iet/scheduler.py
+++ b/devito/ir/iet/scheduler.py
@@ -1,0 +1,158 @@
+from collections import OrderedDict
+
+import numpy as np
+
+from devito.cgen_utils import Allocator
+from devito.dimension import LoweredDimension
+from devito.ir.iet import (Expression, LocalExpression, Element, Iteration, List,
+                           UnboundedIndex, MetaCall, MapExpressions, Transformer,
+                           NestedTransformer, SubstituteExpression, iet_analyze,
+                           compose_nodes, filter_iterations, retrieve_iteration_tree)
+from devito.tools import filter_ordered, flatten
+from devito.types import Scalar
+
+__all__ = ['iet_build', 'iet_insert_C_decls']
+
+
+def iet_build(clusters, dtype):
+    """
+    Create an Iteration/Expression tree (IET) given an iterable of :class:`Cluster`s.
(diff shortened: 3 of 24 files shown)


devito/ir/iet/scheduler.py
```diff
@@ -0,0 +1,158 @@
...
+    for tree in retrieve_iteration_tree(iet):
+        uindices = flatten(i.uindices for i in tree)
+        subs.update({i.expr: LoweredDimension(name=i.index.name, origin=i.expr)
+                     for i in uindices})
+    iet = SubstituteExpression(subs).visit(iet)
+
+    return iet
+
+
+def iet_make(clusters, dtype):
```
FabioLuporini: note for reviewers: this is an improved version of the former ``_schedule_expressions``

devito/ir/iet/scheduler.py
```diff
@@ -0,0 +1,158 @@
...
+            index = 0
+            for i0, i1 in zip(intervals, list(schedule)):
+                if i0 != i1 or i0.dim in clusters.atomics[cluster]:
+                    break
+                root = schedule[i1]
+                index += 1
+            needed = intervals[index:]
+
+            # Build Iterations, including any necessary unbounded index
+            iters = []
```
FabioLuporini: note for reviewers (follows): In particular, this block of code has been pulled from ``ResolveIterationVariable``, refactored, and inserted here

devito/ir/iet/scheduler.py
```diff
@@ -0,0 +1,158 @@
...
+                for k, v in list(schedule.items()):
+                    schedule[k] = transformer.rebuilt.get(v, v)
+        else:
+            # No Iterations are needed
+            processed.extend([Expression(e, dtype) for e in cluster.exprs])
+
+    return List(body=processed)
+
+
+def iet_insert_C_decls(iet, func_table):
```
FabioLuporini: note for reviewers: this is the former ``_insert_declarations``; nothing more, nothing less

devito/dle/backends/advanced.py
```diff
@@ -158,9 +158,9 @@ def _loop_blocking(self, nodes, state):
                 dim = blocked.setdefault(i, Dimension(name=name))
                 block_size = dim.symbolic_size
                 iter_size = i.dim.symbolic_extent
-                start = i.limits[0] - i.offsets[0]
-                finish = i.dim.symbolic_end - i.offsets[1]
-                innersize = iter_size - (-i.offsets[0] + i.offsets[1])
+                start = i.limits[0] + i.offsets[0]
+                finish = i.dim.symbolic_end + i.offsets[1]
+                innersize = iter_size + (-i.offsets[0] + i.offsets[1])
```
navjotk: So this PR already includes the domain switch?

FabioLuporini: > So this PR already includes the domain switch?

no (see description and see my answer to @mloubout below); it does stuff necessary/useful for the switch, but we're still far from it

> Wait, why does + becomes - ?

This PR fixes a stupid hack regarding ``Iteration`` offsets; see https://github.com/opesci/devito/pull/469/files#diff-60ab6d9da10e52ca6ffa3e60cec1dc11L279
In essence, previously we were treating ``offsets`` as relative to the data space, rather than the iteration space. Now we pass in the appropriate information, so we don't have to negate it all over the place anymore.
Example:
previously we were passing ``offsets=(-3, 3)``; now we pass ``offsets=(3, 3)``

devito/ir/equations/equation.py
```diff
@@ -0,0 +1,66 @@
...
+            expr = expr.xreplace(subs)
+
+        expr = super(Eq, cls).__new__(cls, expr.lhs, expr.rhs, evaluate=False)
+        expr.is_Increment = getattr(input_expr, 'is_Increment', False)
+
+        # Get the accessed data points
+        stencil = Stencil(expr)
+
+        # Well-defined dimension ordering
+        ordering = dimension_sort(expr, key=lambda i: not i.is_Time)
```
mlange05: Why does time get special treatment here? If we had a custom dimension `k` and a `TimeFunction` `f(k, t, x, y)`, would the implied ordering be override by this line?

devito/ir/iet/scheduler.py
```diff
@@ -0,0 +1,158 @@
...
+    # Data dependency analysis. Properties are attached directly to nodes
+    iet = iet_analyze(iet)
+
+    # Substitute derived dimensions (e.g., t -> t0, t + 1 -> t1)
+    # This is postponed up to this point to ease /iet_analyze/'s life
+    subs = {}
+    for tree in retrieve_iteration_tree(iet):
+        uindices = flatten(i.uindices for i in tree)
+        subs.update({i.expr: LoweredDimension(name=i.index.name, origin=i.expr)
+                     for i in uindices})
```
mlange05: An aside comment that needs not be addressed in this PR, but has bugged me in another context: The role of `LoweredDimension` and `UnboundIndex` have to be re-though. The problem is that they effectively express the same thing at two different levels (pure expressions and IET), but will not symbolically match each other, making it awkward to find corresponding pairs. Food for thought...

devito/ir/equations/equation.py
```diff
@@ -0,0 +1,66 @@
...
+    into objects of type :class:`types.Indexed`.
+
+    An iteration space is an object of type :class:`Space`. It represents the
+    data points accessed by the equation along each :class:`Dimension`. The
+    :class:`Dimension`s are extracted directly from the equation.
+    """
+
+    def __new__(cls, input_expr, subs=None):
+        # Sanity check
+        assert isinstance(input_expr, sympy.Eq)
```
mlange05: Should we not also accept our own `Eq` type here? It is highly likely that we might import `Eq` via `from devito import *` in examples and tests.

FabioLuporini: No, at least the way I conceived it. Our own Eq is a lower level concept, performing quite a few low level actions. The ``ir`` modules are not imported from ``devito.__init__``, so it should never be the case that one gets that specific ``Eq`` in scope. Right?

Just to be sure and to make things clearer, I could change the name, such as ``LoweredEq`` or something along these lines.

devito/ir/equations/equation.py
FabioLuporini: yes; this is what we have in ``master`` too. This sounds bad, I agree, but it allows us to handle a few corner cases in which the time loop ended up inside an outer dimension, while it was conceptually expected to be the outermost iteration dimension. I already spoke to @mloubout about this, and for the time being we agreed to leave it that way. We can definitely think of a better solution, but this would be non trivial (I tried, already). This PR already improves a very large number of things, so I guess that at least this "issue", which has been around for a while, is for another time.

devito/ir/iet/scheduler.py
FabioLuporini: Agreed. In fact, I tried to kill off ``LoweredDimension`` in this PR, but I failed. I think I'm getting close though

FabioLuporini:
@mlange05 let me know if my answers are satisfactory for the time being

devito/ir/equations/equation.py
mlange05: Yes, so a name change might be best then, as this is in no way obvious atm. My main worry was what happens if we do `LoweredEq(input_expr=LoweredEq(some expression))`, which we need to assert against (I take it it `isinstance(LoweredEq, Eq)`, right?

devito/ir/equations/equation.py
mlange05: Ok, but we really need that as an explicit test case somewhere in `test_operator.py` or equivalent, since again this is in no way obvious.

devito/ir/equations/equation.py
FabioLuporini: I'll change the name.
The assert would still work; however, there's no point in... "lowering an already lowered expression"
Said that, I've already extended this particular piece of code in my current local branch

devito/ir/equations/equation.py
FabioLuporini: I'm always willing to add tests, but how would you write such a test? Note that this is kinda unrelated to this PR, since this is the same behavior as in master; however, I'm happy to add it here.

FabioLuporini: I mean, if you have an equation like this:
```
a(f, x, y) = b(t, x, y)*a(f, x, y)
```
where ``f`` is a "misc" dimension (YASK jargon), what tells you to generate:
```
for t
  for f
    for x
      for y
```
rather than

```
for f
  for t
    for x
      for y
```
?
surely ``y`` comes after ``x`` and both come after ``t/f``, but there's no way to say/infer/... whether ``t`` should appear after or before ``f``.
You may say "well, I see ``f`` first, so it should go first"; but in general we want the time loop to be the outermost. I can also come up with more pathological scenarios...

mlange05: Ok, this is probably stuff for a separate issue, but your example makes no sense:
* The RHS shares `x, y` but then multiplies `t * f`, essentially creating an outer product/tensor.
* The LHS then only knows one of the two dimensions of this tensor, thus implying a reduction in `t` (following our own convention about implied reductions).
* As a corollary, having `t` as the second-level loop is not only correct, but sensible, as it's performing a reduction in `t`_for all_ `f`.

> in general we want the time loop to be the outermost.

In fact, I see no justification in this example as to why we even want `t` outermost, afaict. The semantics will definitely not be harmed by `f` being outermost.

Most likely this example is too artificial to express what you want. I'm also not trying to stop this PR, but basically, whether in`master` or not, this "special treatment" could looks like a dirty hack we want to get rid of, unless someone can provide some sound justification for it that is documented somewhere.

devito/ir/equations/equation.py
mlange05: > there's no point in... "lowering an already lowered expression"

Exactly, and the assert won't catch such foolishness either. Hence my comment about both classes having to be very distinct, and not sharing a name. Ideally we would even catch if someone attempted this, but that's not strictly necessary, I guess.

mlange05:
> @mlange05 let me know if my answers are satisfactory for the time being

I'd like the name change `Eq`->`LoweredEq`, as the assert won;t catch the obvious misuse. The rest is nitpicking and can be discussed on other issues/PRs.

devito/ir/equations/equation.py
FabioLuporini: > his "special treatment" could looks like a dirty hack
I have no objections to calling this a hack; in fact, I probably agree with you. 

You can make the reduction explicit in my example if you prefer:
```
a(f, x, y) = a(f, x, y) + b(t, x, y)*a(f, x, y)
```
and now again you can detect ("read", rather than "detect", if the information is carried by the ``Eq``) that there's a reduction and set up a rule for dimension ordering in case of reductions, OR basically you're just saying the same exact thing that I said -- there is no unique way of scheduling the ``f`` and ``t`` loops because both orders are correct ! 

What I didn't say, as hidden in between the lines, is that preferring the time dimension is outermost is probably, *performance-wise*, the easy thing to do in such a case

mlange05: >  there is no unique way of scheduling the f and t loops because both orders are correct !

Yes, from the pure API semantics that's exactly right. So how do we decide that `t` is indeed special enough to be outermost? Well, if it's a `SteppingDimension` it would make sense to say "be as outermost as semantically permitted", as processing a single buffer-iteration is probably fastest. Maybe solving this becomes as easy as switching `is_Time` -> `is_Stepping`, and leaving a detailed comment (or as I said, test)?

In any case, advanced nitpicking, so ...

FabioLuporini:
@mlange05 ``ir.Eq`` is now ``ir.LoweredEq``.

I'll think about the dimension ordering some more, but perhaps this shouldn't be a stopper for the PR

FabioLuporini:
I guess everything has been addressed, so merging...

FabioLuporini:
Merged.

PR #470, opened 2018-01-11
test save fix

mloubout:
Test_save hotfix for @FabioLuporini 

- save setup properly in `TimeFunction`

- No `initializer`

This PR only fix the test, not the `initilizer` overall removall


FabioLuporini:
in fact, @navjotk just convinced me that leaving the initializer is a good idea..

FabioLuporini:
Merged.

PR #473, opened 2018-01-18
Time order

mloubout:
So this PR removes the `time_order` from the seismic examples. The reason is, everything is 2nd order in time and letting it open may lead to user trying weird ones and complaining. It should also remove compiler warnings for 4th order due to unused time indices.
To summarize:

- TTI has no time_order anymore
- acoustic has now two modes `OT4` and `OT2` for 2nd and 4th order in time, `time_order` is fixed to 2 everywhere.

segfault on my laptop so probably needs some debug/test fix/flake8/ but should be done in a day or two, open for people input.


mloubout:
> but I don't think replacing time_order with OT changes much in acoustic; what am I missing?

Two things 
1- 4th order is not 4th order FD 
2- Want to make sure people don;t put whatever `time_order` thinking if it's an input it can be changed

Not sure what you mean by only TTI, `time_order` is fixed to 2 everywhere now

FabioLuporini:
> Two things
> 1- 4th order is not 4th order FD
> 2- Want to make sure people don;t put whatever time_order thinking if it's an input it can be changed

1- sure
2- ooh, OK, I see what you mean now

mlange05:
Looks good and makes sense to me. So, I'd say GTG.

On a side-note, I notice that we could go further with this and write out all the operators explicitly, so that `kernel` gets resolved by the `WaveSolver` picking the appropriate class (eg. `AcousticForwardOT2`, `AcousticGradientOT2`, etc). This would probably result in some redundancy in the code, but have the advantage of all the pieces sitting side-by-side, instead of having to chase things through `iso_stencil` while the `dt` adjustment lives as a conditional in another part of the file. I personally would find this easier to read, but this largely comes down to taste, I guess. In any case, probably not in this PR - just food for thought.

mloubout:
On hold, DSE issue

FabioLuporini:
you mean #476?

examples/seismic/acoustic/acoustic_example.py
```diff
@@ -49,17 +49,17 @@ def acoustic_setup(shape=(50, 50, 50), spacing=(15.0, 15.0, 15.0),
 
     # Create solver object to provide relevant operators
     solver = AcousticWaveSolver(model, source=src, receiver=rec,
-                                time_order=time_order,
+                                kernel=kernel,
```
FabioLuporini: can we avoid one line here? It's OK in a next PR

mloubout: Probaly, let e update it so

FabioLuporini:
Merged.

PR #474, opened 2018-01-19
symbolics: Fix xreplace_constrained when providing pure symbols

FabioLuporini:
@mloubout this at least fixes the TTI compilation issue (shifted mode)
I cannot reproduce the segfault in aggressive mode though, and I tried really hard
anyway, let's start taking a look at this


FabioLuporini:
also shipping a tweak to the DSE cost model that should make tti faster in advanced mode w.r.t. master

mloubout:
GTG, fixes the dse issue.

FabioLuporini:
Merged.

PR #478, opened 2018-01-24
Travis separate compilers

mlange05:
This PR unrolls the apt-package specs in the travis file for each builder in the build matrix. This means we now only install one compiler version per builder, rather than all of them, dropping the apt-install time from ~10min to about 1-2min.


FabioLuporini:
excellent, GTG

mlange05:
Ok, merging.

PR #479, opened 2018-01-26
HOTFIX origin

mloubout:
Fixes handling of non-zero origin that was done wrong (my bad)

### DIFF
--- a/tests/test_interpolation.py
+++ b/tests/test_interpolation.py
@@ -4,7 +4,8 @@ from conftest import skipif_yask
 
 from devito.cgen_utils import FLOAT
 from devito import Grid, Operator, Function, SparseFunction
-
+from examples.seismic import demo_model, RickerSource, Receiver
+from examples.seismic.acoustic import AcousticWaveSolver
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


FabioLuporini:
fine, I think

tests/test_interpolation.py
```diff
@@ -147,3 +149,53 @@ def test_adjoint_inject_interpolate(shape, coords,
     term1 = np.dot(p2.data.reshape(-1), p.data.reshape(-1))
     term2 = np.dot(c.data.reshape(-1), a.data.reshape(-1))
     assert np.isclose((term1-term2) / term1, 0., atol=1.e-6)
+
+
+@skipif_yask
+@pytest.mark.parametrize('shape', [(50, 50, 50)])
+def test_position(shape):
```
FabioLuporini: it looks like this test is using lots of duplicated code? can we reuse some? as in, it's the usual acoustic problem setup, with a shifted origin...

mloubout: This one is a bit by itself in the test and I find it more readable to see the src/rec position shift rather than dig back and find what the original position/origin was

Issue #387, opened 2017-10-24
Improve code generation


FabioLuporini:
There's currently an explosion in the number of parameters passed to the generated code from Python. This obfuscates the readability of the generated code itself. Instead of passing each parameter, one by one, we could rather use C structs and then generated stuff like "x.start, x.end, ..."

FabioLuporini:
This is probably stupid. Closing.

PR #481, opened 2018-01-26
IET-based parameters and alternative argument "engine"

### LINKED ISSUE #465: Wrong derivation of runtime parameters
The property `Operator.parameters`, which defines the list of arguments that needs to be supplied to or derived by the operator at runtime, includes various `Dimension`-related variables that are not actually required by the generated kernel code. As a result, the auto-derivation and the "Arguments Engine" are overly complex and introduce a lot of potentially redundant complexity to rectify this "wrong spec". I believe a much more prudent approach would be to correct the `op.parameters` derivation based on the final IR, instead of trying to estimate the required variables from the list of `Dimension` objects as we do currently.

As an example, the simple kernel generated by the following snippet requires only the "start" and "end" values for the time loop (`t_s`/`t_e` and `time_e`/`time_s` following PR #453):
```
from devito import *
grid = Grid(shape=(5, 6))
op = Operator(Eq(f, f + 3))
```

However, in both `master` and #453 `op.parameters` contains all of `[t_s, t_e, t_size, time_s, time_e, time_size]`, thus forcing some pretty complex information flow during the auto-derivation, despite the variables never being used.

#### Potential solution:
Instead of adding all variables `dim_s`, `dim_e`, `dim_size` for each `Dimension` `dim`, I believe we should derive the individual variables in `op.parameters` from the final IR AST (`op.nodes`). For this to work we need to ensure that all dimension-related variables are symbols that are reachable using the symbolic search routines (integer-based `Constants` maybe?). This is tricky, however, as neither the variable casts that require the `dim_size` variables, nor the loop headers that need `dim_s`/`dim_e` are currently being traversed by things like the `FindSymbols` visitor. 

Using this approach, the perceived/pseudo-cyclic dependency between derived and parent `Dimensions` (eg. `t` and `time`) could then be avoided; either by ignoring `DerivedDimension` values and always deferring to the parent, or by auto-detecting if a symbol is defined within a particular kernel, or needs to be defined externally via arguments (more complex, but maybe "cleaner").

In addition, I believe an IR-based derivation of `parameters` would also help integrate the DLE and AT-specific arguments more neatly, as the variables introduced by `BlockingDimension`s now just become additional variables that are inserted into the IR/AST by the DLE picked up automatically.

#### Future improvements
If the described idiosyncrasy were removed, I believe large swathes of the "argument engine" code, including the generic visitor pattern introduced in #453, become either redundant or can at least be simplified significantly, because the rules for argument defaulting and user-overrides then become fairly linear and straightforward (objectively). I have a branch to roughly demonstrate this (`mlange-arguments`) and/or I can raise a separate issue with a (again objectively) "neater" design concept if this is of use to anyone. However, I regard the problem described above as a separate "bug" and I would argue that continuing to work around it instead of fixing it will become excessively laborious and detrimental to code quality (personal opinion, of course).

mlange05:
#### Reviewer note:
 Apologies for the somewhat large PR, but unfortunately certain assumptions in the current `parameters` implementation entail that both, parameters and arguments, have to be changed at the same time. There are also still a few minor "hacks" and shortcuts that we might want to iron out during or after thorough review, so I'm submitting this now.

### Description:
This PR has two primary objectives: 
* Change the derivation of required operator/function inputs (`.parameters`) to be based on the actual IR/IET of the function. This entails we now only require `parameters` that are actually used by the function/operator, instead of the conservative over-estimate that we previously used.
* Replace the current `ArgumentEngine` with an (objectively)  much cleaner symbol-based mechanism for deriving argument values. The new mechanism uses two stages: default value derivation (`argument_defaults`) and runtime value overrides based on user input (`argument_values`), both of which are defined by the symbol classes themselves. This makes the definition of our "derivation rules" for arguments more accessible and allows things like aliasing to be implemented much more easily. 
_Sidenote: The new mechanism encapsulates the same shallow tree/forest-like traversal pattern as the previous "engine" but in a more natural way, as the intrinsic relationships (eg. `Composite->Function->Dimension->Scalar`) are already defined by each symbol (eg. a `Function` knows it's `Dimension`s (`indices`) and each `Dimension` knows its `start_symbolic` or its `parent`)._

### TLDR; Detailed feature/change list
**Part 1: IET-based parameter derivation**
* Use `Scalar` types for iteration bound variables and add them to the `free-symbols` finder utility.
* Add a new symbol finder mode `defines` that collects symbols locally defined symbols by an IET node.
* Utility to derive `parameters` as the set of `functions/symbolics` + `free-symbols` - `defines`.
* Add explicit `ArrayCast` and `PointerCast` nodes to the tree to pick up the correct `dim_size` variables required by casts.
* Use new `derive_parameters` utility to derive argument/parameter list for elemental functions.
_Note: This part is probably still a bit wonky and can be simplified further, either as part of the review discussion or in a follow-up PR._

**Part 2: Symbol-based argument derivation**
* No more standalone `ArgumentEngine`
* `parameters` are defined as a list of symbolic objects and arguments are defined as a `name->value` map. There are no more dedicated `xxxArgument` types. 
* Each `Function/Dimension` type now provides `argument_default` and `argument_values(**kwargs)` to derive default values and user-defined overrides.
* The defaulting mechanism utilises a modified `MultiDict` to collect candidate values from multiple sources (eg. multiple `Function` objects may provide values for `Dimension` sizes). This `ArgumentDict` then performs a candidate reduction (`unique()`) during retrieval to ensure the value is uniquely defined and throws an error if competing values are present.
_Note: This could be extended with additional reduction rules if we need such "tie breaker" mechanisms, but the current code base does not actually require it._
* `SteppingDimension`: Derived dimension variables do not show up in `parameters` anymore, as they are now defined locally. This means each `DerivedDimension` object effectively adds their candidate values to the `.parent` during default and value derivation, a.k.a "derived dims alias their parents".
* For value overrides, `op(dim=<value>, ...)` now aliases to `op(dim_e=<value>, ...)` and contributes candidate values to the respective map entry.
* User-defined values for `CompositeFunction` or `Function` types now can easily use `new.default_values(...)` to re-derive default values like dimension sizes during the override stage.

#### Final notes:
* Argument derivation should probably also include a third stage `argument_validation` that checks the default + override values against actual data shapes to pick up inconsistencies. Since the PR is already huge and I'm kind of out of time, I've left this for a follow-up PR (not necessarily by me).
* ~Travis is still failing for one last set of tests all related to `example_gradient.py`. As the regular gradient test is performing fine, I'm assuming this is only a minor inconsistency somewhere in the test and would like to go ahead with the review for now, as it is quite comprehensive.~ Fixed now.

fixes #465 #416 #387


devito/types.py
```diff
@@ -165,6 +165,22 @@ def function(self):
     def indexify(self):
         return self
 
+    def argument_defaults(self):
```
FabioLuporini: why not in ``Basic``, or in a Mixin?

devito/types.py
```diff
@@ -345,6 +361,22 @@ def indexify(self, indices=None):
         indices = [a.subs(subs) for a in self.args]
         return Indexed(self.indexed, *indices)
 
+    def argument_defaults(self):
```
FabioLuporini: if in ``Basic``, as I wrote above, we can drop some redundant code

tests/test_autotuner.py
```diff
@@ -101,7 +101,7 @@ def test_timesteps_per_at_run():
         stencil = Eq(outfield.indexed[t + to, x, y, z],
                      outfield.indexify() + infield.indexify()*3.0)
         op = Operator(stencil, dle=('blocking', {'blockalways': True}))
-        op(infield=infield, outfield=outfield, autotune=True)
+        op(infield=infield, outfield=outfield, t=2, autotune=True)
```
FabioLuporini: ?

devito/ir/iet/nodes.py
```diff
@@ -156,6 +157,24 @@ def __init__(self, name, params=None):
     def __repr__(self):
         return "Call::\n\t%s(...)" % self.name
 
+    @property
+    def functions(self):
```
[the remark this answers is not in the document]
mlange05: :+1:

devito/ir/iet/nodes.py
```diff
@@ -156,6 +157,24 @@ def __init__(self, name, params=None):
...
+        """Return all :class:`Symbol` objects used by this :class:`Call`."""
+        return tuple(p for p in self.params if isinstance(p, types.AbstractFunction))
+
+    @property
+    def free_symbols(self):
+        """Return all :class:`Symbol` objects used by this :class:`Call`."""
+        free = tuple(set(flatten(p.free_symbols for p in self.params)))
+        # HACK: Filter dimensions to avoid them on popping onto outer parameters
```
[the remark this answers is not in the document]
mlange05: Ok, so the example arises when elementals encapsulate the `x` and `y` loops. In this case, `x` and `y` still appear in `functions` due to being part of the symbol `u(t, x, y)`, but the `Iteration` that `defines` them is in an elemental. Now, they are undefined to the outer `Operator` and thus appear on the argument list, unless explicitly taken out. As stated, it's a bit of a hack...

Also :+1: on the `cached_property`.

devito/ir/iet/nodes.py
```diff
@@ -297,6 +328,14 @@ def __repr__(self):
             index += '[%s]' % ','.join(ccode(i.index) for i in self.uindices)
         return "<%sIteration %s; %s>" % (properties, index, self.limits)
 
+    @property
+    def defines(self):
+        """
+        Return any symbols an :class:`Iteration` may define.
+        """
+        dim = self.dim.parent if self.dim.is_Stepping else self.dim
```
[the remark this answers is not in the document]
mlange05: :+1:

devito/ir/iet/nodes.py
```diff
@@ -419,6 +458,23 @@ def children(self):
         """Return the traversable children."""
         return (self.nodes,)
 
+    @property
+    def functions(self):
+        """Return all :class:`Function` objects used by this :class:`Iteration`"""
```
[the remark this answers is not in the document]
mlange05: :+1:

devito/ir/iet/utils.py
```diff
@@ -119,3 +119,21 @@ def compose_nodes(nodes, retrieve=False):
         return body, tree
     else:
         return body
+
+
+def derive_parameters(nodes):
```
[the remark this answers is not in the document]
mlange05: I believe it already does, no?

devito/dle/backends/basic.py
```diff
@@ -68,43 +64,41 @@ def _create_elemental_functions(self, nodes, state):
...
+            defines = [s.name for s in FindSymbols('defines').visit(free)]
+            casts = [ArrayCast(f) for f in f_symbols if f.name not in defines]
+            free = (List(body=casts), free)
+
+            for i in derive_parameters(free):
+                if i.name in defined_args:
+                    args.append((defined_args[i.name], i))
                 elif i.is_Array:
-                    args.append(("(%s*)%s" % (c.dtype_to_ctype(i.dtype), i.name), i))
+                    args.append((i, i))
```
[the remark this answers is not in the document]
mlange05: :+1:

devito/ir/iet/visitors.py
```diff
@@ -132,34 +131,59 @@ def _args_decl(self, args):
         """Convert an iterable of :class:`Argument` into cgen format."""
         ret = []
         for i in args:
-            if i.is_ScalarArgument:
+            if i.is_Object:
+                ret.append(c.Value('void', '*_%s' % i.name))
+            elif i.is_Scalar:
```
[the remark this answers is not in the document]
mlange05: :+1

devito/operator.py
```diff
@@ -85,7 +86,8 @@ def __init__(self, expressions, **kwargs):
                 time.reverse = time_axis == Backward
 
         # Parameters of the Operator (Dimensions necessary for data casts)
-        parameters = self.input + self.dimensions
+        parameter_dims = [d for d in self.dimensions if not d.is_Stepping]
```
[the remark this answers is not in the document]
mlange05: This code is actually dead now, as the actual arguments are overwritten later. I'll remove the dead code.

devito/ir/iet/nodes.py
```diff
@@ -156,6 +157,24 @@ def __init__(self, name, params=None):
     def __repr__(self):
         return "Call::\n\t%s(...)" % self.name
 
+    @property
+    def functions(self):
+        """Return all :class:`Symbol` objects used by this :class:`Call`."""
+        return tuple(p for p in self.params if isinstance(p, types.AbstractFunction))
```
FabioLuporini: ``... if p.is_AbstractFunction`` ?

devito/ir/iet/nodes.py
```diff
@@ -156,6 +157,24 @@ def __init__(self, name, params=None):
...
+
+    @property
+    def free_symbols(self):
+        """Return all :class:`Symbol` objects used by this :class:`Call`."""
```
FabioLuporini: same docstring as ``def functions`` ?

devito/ir/iet/nodes.py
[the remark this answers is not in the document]
FabioLuporini: and I guess you tried fixing this issue within `_create_elemental_functions`, without success, right?

anyway, I read your description 10 times and I'm still unsure what's actually going on. Is this the situation:

```
void elemental (...) {
  ...
}

void kernel (... int x, int y ...) {
   ...
   elemental ( ... )
}
```
? or what?

devito/operator.py
```diff
@@ -117,24 +119,66 @@ def __init__(self, expressions, **kwargs):
...
+        casts = [ArrayCast(f) for f in functions if f.is_Tensor and f._mem_external]
+        profiler = Object(self.profiler.name, self.profiler.dtype, self.profiler.new)
+        casts.append(PointerCast(profiler))
+        nodes = (List(body=casts), nodes)
 
-        parameters = self.argument_engine.arguments
+        # Derive parameters as symbols not defined in the kernel itself
+        parameters = derive_parameters(nodes)
+
+        # Filter all internally-allocated temporary `Array` types
```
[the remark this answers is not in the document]
mlange05: Well, not really as all the DLE transformations have to happen for the IET to be in its final shape and the whole point of this is to use the actual final IET and not lug around assumptions from a previous IR state.

devito/ir/iet/utils.py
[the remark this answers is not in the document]
FabioLuporini: right, I guess I was looking at something else

devito/ir/iet/nodes.py
mlange05: Nope, the DLE insert actual expressions into the call signature, `self.params` is not guaranteed to be internal symbols only.

devito/ir/iet/nodes.py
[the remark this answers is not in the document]
mlange05: Yes, exactly. Nothing in the main `Operator` actually defines the symbols (which makes sense), but the presence of the `Function` symbol brings `x`/`y` into `arguments.

devito/operator.py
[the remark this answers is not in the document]
FabioLuporini: excellent

devito/ir/iet/nodes.py
FabioLuporini: OK

devito/operator.py
```diff
@@ -117,24 +119,66 @@ def __init__(self, expressions, **kwargs):
         # Introduce the required symbol declarations
         nodes = iet_insert_C_decls(dle_state.nodes, self.func_table)
 
-        # Initialise ArgumentEngine
-        self.argument_engine = ArgumentEngine(clusters.ispace, parameters,
-                                              self.dle_arguments)
+        # Insert data and pointer casts for array parameters and profiling structs
+        functions = FindSymbols('symbolics').visit(nodes)
```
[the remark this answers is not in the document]
mlange05: Yes, it would seem so. I also note that `self.input` does not seem to be use anywhere, but should probably be let for another day.

devito/operator.py
```diff
@@ -117,24 +119,66 @@ def __init__(self, expressions, **kwargs):
...
+        casts = [ArrayCast(f) for f in functions if f.is_Tensor and f._mem_external]
+        profiler = Object(self.profiler.name, self.profiler.dtype, self.profiler.new)
+        casts.append(PointerCast(profiler))
+        nodes = (List(body=casts), nodes)
```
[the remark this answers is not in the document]
mlange05: Hmm, I actually tried generalising this and moving it to a utility function, because the same thing needs to be done for newly created elementals. However, this is not as easy as it looks, due to the funky way the profiler `Object`s are handled. Since elementals don't need profiling, I think it would be great to simply pick the `Object` off the `TimedList` IET nodes, but that soon became messy. I'll give this another go, but maybe in a separate PR?

devito/operator.py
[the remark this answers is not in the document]
mlange05: But parameter derivation is about more than just function collection - it requires knowledge of who defines (all types of) variables. This means, all things that adjust variable definitions, like elementals or `insert_C_decls` have to have happened before we can safely derive the final `parameter`s. 

What this "hack" is really doing is effectively "predicting" what the `Allocator` will do, because `Allocator` only inserts raw CGen format for local temporary allocations, which cannot be read from the IET state. In general, I don't like this approach of "predicting what another part of the code will do", since this is exactly what caused the inaccurate over-estimation of parameters in the first place (and thus got us into this mess). So instead, I think "the right thing to do" here is to adjust the `Allocator` to insert local array allocations into the IET, allowing us to actually use a fixed state for derivation instead of a "previous state + assumptions".

devito/operator.py
```diff
@@ -117,24 +119,66 @@ def __init__(self, expressions, **kwargs):
...
+            default_args.update(p.argument_defaults())
+        arguments = {k: default_args.reduce(k) for k in default_args}
+
+        # Next, we insert user-provided overrides
+        for p in self.input + self.dimensions:
+            arguments.update(p.argument_values(**kwargs))
+
+        # Derive additional values for DLE arguments
+        # TODO: This is not pretty, but it works for now. Ideally, the
+        # DLE arguments would be massaged into the IET so as to comply
```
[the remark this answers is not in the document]
mlange05: See my comment about IET state and `Allocator` above. I've tried this, but it turned out to be riddled with subtleties.

devito/ir/iet/visitors.py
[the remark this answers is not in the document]
mlange05: Actually no, doesn't work. Turns out objects that are `is_Scalar` do not have the `.is_Lowered` attribute, as they are not `Dimension` types. This then causes errors, so I'll have to leave this one as it is.

devito/operator.py
```diff
@@ -117,24 +119,66 @@ def __init__(self, expressions, **kwargs):
...
-        """ Process any apply-time arguments passed to apply and derive values for
-            any remaining arguments
         """
+        Process runtime arguments passed to ``.apply()` and derive
+        default values for any remaining arguments.
+        """
+        # First, derive all default values from parameters and reduce
+        # TODO: Should be self.parameters, but that converts to
+        # TensorArgument type, etc...
+        # Also, this part can probably be a cached_property
```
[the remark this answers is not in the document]
mlange05: Looks like this "just works". Nice. :smile:

devito/operator.py
```diff
@@ -171,9 +215,11 @@ def cfunction(self):
             # Associate a C type to each argument for runtime type check
             argtypes = []
             for i in self.parameters:
-                if i.is_ScalarArgument:
+                if i.is_Object:
```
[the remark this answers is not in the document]
mlange05: Since `Object`s don't adhere to the internal hierarchy they don't have `is_Scalar` or `is_Tensor`, so we need to catch them first to avoid errors.

devito/types.py
mlange05: Sure, why not. :+1:

devito/types.py
mlange05: :+1:

tests/test_autotuner.py
mlange05: Ahhh, this is a subtle one. In the previous scheme, a stepping dimension would provide its buffer size as the default runtime loop bound. Since this makes no sense from a modellers perspective, this rule got dropped, meaning every time-dependente operator now needs to either use `save=some_value` or define the number of timesteps on invocation. This test does not (because it doesn't care), but this now gets picked up as a missing value for `time`, hence we need to provide one.

devito/ir/iet/nodes.py
mlange05: :+1:

mlange05:
Ok, I believe I have addressed all/most of @FabioLuporini s comments.

I've also fixed the `gradient_example`-related test failures. Turns out this "second implementation" of our base test uses a lot of weird symbol-name aliasing (eg. `op(v=TimeFunction(name='u', ...), ...)` or `op(rec=SparseFunction(name='rec_g', ...), ...)`, which stretches the aliasing mechanism way beyond what the internal tests do. This probably means we need more internal testing for these extreme cases. Following that, the main builders should be green.

The last item is now the YASK builder, and I'm on it. @ggorman @navjotk Now would be a good time to get another review or two in... :wink:

devito/ir/iet/nodes.py
```diff
@@ -419,6 +486,29 @@ def children(self):
         """Return the traversable children."""
         return (self.nodes,)
 
+    @property
+    def functions(self):
```
FabioLuporini: I'm adding this comment here, but it applies to lots of places. I think all these methods, ``functions, free_symbols, ...`` should return tuples, not lists

FabioLuporini: In fact, after skimming through the changes, it seems like there are some inconsistencies ?

devito/ir/iet/visitors.py
```diff
@@ -132,34 +131,59 @@ def _args_decl(self, args):
...
+            try:
+                if i.is_Object:
+                    ret.append('*_%s' % i.name)
+                elif i.is_Array:
+                    ret.append("(%s*)%s" % (c.dtype_to_ctype(i.dtype), i.name))
+                elif i.is_Scalar:
+                    ret.append(i.name)
+                elif i.is_TensorFunction:
+                    ret.append('%s_vec' % i.name)
+            except:
```
FabioLuporini: Something better than just except?

devito/operator.py
```diff
@@ -96,10 +95,10 @@ def __init__(self, expressions, **kwargs):
         nodes = iet_build(clusters, self.dtype)
 
         # Introduce C-level profiling infrastructure
-        nodes, self.profiler = self._profile_sections(nodes, parameters)
+        nodes, self.profiler = self._profile_sections(nodes)
 
         # Translate into backend-specific representation (e.g., GPU, Yask)
-        nodes = self._specialize(nodes, parameters)
+        nodes = self._specialize(nodes)
```
FabioLuporini: I'm pretty sure that the YASK backend modifies ``parameters``  in place, but I cannot find any changes to the YASK backend?

devito/operator.py
```diff
@@ -109,32 +108,81 @@ def __init__(self, expressions, **kwargs):
...
         self._includes.extend(list(dle_state.includes))
 
         # Introduce the required symbol declarations
         nodes = iet_insert_C_decls(dle_state.nodes, self.func_table)
 
-        # Initialise ArgumentEngine
-        self.argument_engine = ArgumentEngine(clusters.ispace, parameters,
-                                              self.dle_arguments)
+        # Insert data and pointer casts for array parameters and profiling structs
+        casts = [ArrayCast(f) for f in self.input if f.is_Tensor and f._mem_external]
```
FabioLuporini: I'm re-adding a comment that GitHub dropped after the latest changes: this should better go in ``ir/iet/scheduler`` as a ``iet_insert_casts`` free function

devito/operator.py
```diff
@@ -109,32 +108,81 @@ def __init__(self, expressions, **kwargs):
...
+        profiler = Object(self.profiler.name, self.profiler.dtype, self.profiler.new)
+        casts.append(PointerCast(profiler))
+        nodes = (List(body=casts), nodes)
+
+        # Derive parameters as symbols not defined in the kernel itself
+        parameters = derive_parameters(nodes)
 
-        parameters = self.argument_engine.arguments
+        # Filter all internally-allocated temporary `Array` types
+        # TODO: Huge hack, but to fix it, allocations need to be part of IET
```
FabioLuporini: OK, I've thought about this some more. Perhaps it would be better to let ``derive_parameters`` take an optional parameter that, if set to True, drops all local arrays? It would kinda make sense to me

devito/operator.py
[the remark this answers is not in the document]
FabioLuporini: OK

devito/operator.py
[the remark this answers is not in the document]
FabioLuporini: really? how is that? Doesn't ``Object`` inherit from ``Basic`` ?

tests/test_autotuner.py
FabioLuporini: OK

devito/operator.py
```diff
@@ -109,32 +108,81 @@ def __init__(self, expressions, **kwargs):
...
+                if arg.value is None:
+                    arguments[dim.symbolic_size.name] = osize
+                elif isinstance(arg.value, int):
+                    arguments[dim.symbolic_size.name] = arg.value
+                else:
+                    arguments[dim.symbolic_size.name] = arg.value(osize)
 
-        arguments, autotune = self.argument_engine.handle(**kwargs)
+        # Add in the profiler argument
+        arguments[self.profiler.name] = self.profiler.new()
```
FabioLuporini: Should we move this to ``_argument_defaults()`` ?

devito/ir/iet/visitors.py
```diff
@@ -132,34 +131,59 @@ def _args_decl(self, args):
         """Convert an iterable of :class:`Argument` into cgen format."""
```
FabioLuporini: Should update the docstring?

tests/test_autotuner.py
FabioLuporini: > Ahhh, this is a subtle one. [...] hence we need to provide one.

I had to apply the same fix to another couple of tests

FabioLuporini:
I think I've fixed the YASK backend tests. I also pleased flake8 and fixed a couple of leftover trivial bugs. Let's see what Travis has to say now...

devito/operator.py
```diff
@@ -117,24 +119,66 @@ def __init__(self, expressions, **kwargs):
...
 
+        # Execute autotuning and adjust arguments accordingly
+        autotune = kwargs.pop('autotune', False)
         if autotune:
-            arguments = self._autotune(arguments)
+            # AT assumes and ordered dict, so let's feed it one
+            at_args = OrderedDict([(p.name, arguments[p.name]) for p in self.parameters])
+            arguments = self._autotune(at_args)
```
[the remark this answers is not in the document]
FabioLuporini: *looks* OK

tests/test_operator.py
```diff
@@ -588,10 +584,25 @@ def test_argument_derivation_order(self, nt=100):
 
         # Providing a scalar argument explicitly should override the automatically\
         # inferred
-        op_arguments = op.arguments(b=b1, time=nt - 1, time_e=nt - 2)
+        op_arguments = op.arguments(b=b1, time_e=nt - 2)
```
navjotk: Why has this been changed?

devito/ir/iet/visitors.py
navjotk: This

devito/ir/iet/visitors.py
navjotk: Should this be addressed?

devito/ir/iet/nodes.py
navjotk: This probably needs to be addressed

tests/test_operator.py
FabioLuporini: So Michael changed it; I **think** it's because now ``time`` aliases to ``time_e``

devito/ir/iet/visitors.py
FabioLuporini: I don't know how

FabioLuporini: I'm guessing ``AttributeError``. Let's see...

devito/ir/iet/visitors.py
FabioLuporini: yes

devito/ir/iet/nodes.py
FabioLuporini: yes

FabioLuporini:
@navjotk addressed your comments. I'll merge once formally approved.

navjotk:
GTG

FabioLuporini:
Merged.

PR #482, opened 2018-01-26
Initial implementation of sub-regions

FabioLuporini:
This PR allows users to specify whether an equation should be run over the entire domain (default choice) or its interior region only (ie, boundary excluded)

Two main features are shipped:
* introduction of ``DOMAIN`` and ``INTERIOR``, two objects of type``Region``, exposed to the user. With them, users will be able to restrain individual equations to a certain region of the computational domain. For example:
- ``Eq(<my_expression>, region=DOMAIN)``
- ``Eq(<my_expression>, region=INTERIOR)``
- ``Eq(<my_expression>) ==  Eq(<my_expression>, region=DOMAIN)``
The latter one ensures backwards compatibility

2- introduction of ``SubDimension`` type; this is **not** exposed to the user. A ``SubDimension`` is used by the compiler to:
- generate a proper loop nest structure
- build up the expected default values for the runtime arguments.

A new test is introduced:
```
...
eqs = [Eq(u.forward, u + 1),
       Eq(u.forward, u.forward + 2, region=INTERIOR)]
op = Operator(eqs, dse='noop', dle='noop')
...
```
this builds the following loop nest structure (pseudocode):
```
for t
  for x = x_s to x_e
    for y
      for z
        u[t+1,x,y,z] = u[t,x,y,z] + 1
  for xi = xi_s to xi_e
    for yi
      for zi
        u[t+1,xi,yi,zi] = u[t+1,xi,yi,zi] + 2
```
At runtime,  ``xi_s`` and ``xi_e`` will be such that:
```
xi_s == x_s + 1
xi_e == x_e - 1
```

As you can see, no offsets are introduced in the loop headers

EDIT: requires #481 (so you can temporarily abstain from comment, as the diff is huge)


devito/function.py
```diff
@@ -710,12 +735,16 @@ def __init__(self, *args, **kwargs):
             d = Dimension(name='d')
             self.coordinates = Function(name='%s_coords' % self.name,
                                         dimensions=[self.indices[-1], d],
-                                        shape=(self.npoint, self.grid.dim))
+                                        shape=(self.npoint, self.grid.dim),
+                                        space_order=0)
```
mloubout: I wouldn't add that yet and leave it to default, I am still looking at extending FD to SparseFunctions. This seem like a hacky assumption for the compiler.

devito/function.py
```diff
@@ -710,12 +735,16 @@ def __init__(self, *args, **kwargs):
...
             self._children.append(self.coordinates)
             coordinates = kwargs.get('coordinates', None)
             if coordinates is not None:
                 self.coordinates.data[:] = coordinates[:]
 
+            # A SparseFunction has no halo region
+            self._halo = tuple((0, 0) for i in range(self.ndim))
```
mloubout: Same as above.

devito/ir/equations/equation.py
```diff
@@ -2,66 +2,110 @@
...
         if subs is not None:
             expr = expr.xreplace(subs)
 
-        expr = super(LoweredEq, cls).__new__(cls, expr.lhs, expr.rhs, evaluate=False)
-        expr.is_Increment = getattr(input_expr, 'is_Increment', False)
+        # Well-defined dimension ordering
+        ordering = dimension_sort(expr, key=lambda i: not i.is_Time)
+
+        # Introduce space sub-dimensions if need to
+        region = getattr(input_expr, '_region', DOMAIN)
```
mloubout: I know this is for backward compatibility, but I would put `INTERIOR` as default as it makes more mathematical sense.

devito/ir/support/space.py
```diff
@@ -223,6 +241,17 @@ def drop(self, d):
...
+    def __getitem__(self, dim):
+        if not self.is_well_defined:
+            raise ValueError("Cannot fetch Interval from ill defined Space")
+        for i in self.intervals:
+            if i.dim == dim:
+                return i
+
+
+class DataSpace(Space):
+    pass
```
mloubout: What is this class for if empty?

tests/test_operator.py
```diff
@@ -980,6 +980,36 @@ def test_equations_mixed_timedim_stepdim(self):
...
+
+        u = TimeFunction(name='u', grid=grid)  # noqa
+        eqs = [Eq(u.forward, u + 1),
+               Eq(u.forward, u.forward + 2, region=INTERIOR)]
+
+        op = Operator(eqs, dse='noop', dle='noop')
+        trees = retrieve_iteration_tree(op)
+        assert len(trees) == 2
+
+        op.apply()
```
mloubout: SO how many time steps does it run, only updates t=1 so one?

tests/test_operator.py
```diff
@@ -980,6 +980,36 @@ def test_equations_mixed_timedim_stepdim(self):
...
+        Tests regions work properly in terms of code generation and runtime
+        argument derivation.
+        """
+        grid = Grid(shape=(4, 4, 4))
+        x, y, z = grid.dimensions
+        t = grid.stepping_dim  # noqa
+
+        u = TimeFunction(name='u', grid=grid)  # noqa
+        eqs = [Eq(u.forward, u + 1),
+               Eq(u.forward, u.forward + 2, region=INTERIOR)]
```
mloubout: Weird place but looking into it, how easy would it be to create a 'BOUNDARY' region that would be 'DOMAIN' - 'INTERIOR'

devito/function.py
FabioLuporini: I thought ``coordinates`` for a ``SparseFunction`` could only be a ``space_order=0`` field. There's no hidden hack in here, so I can drop it you prefer. I remember I added it because in the upcoming domain-allocation switch it was fixing something, but I don't remember what that was. So this is a cherry-picked commit that it felt appropriate to have it in master ASAP. 

Can we infer the coordinates space order from the ``SparseFunction`` initialisation ?

mloubout: My bad, added the comment at the wrong place, coordinates are indeed always zero order, the SparseFunction may not.

devito/function.py
FabioLuporini: Oh, so that you can have high order sparse functions?

devito/ir/support/space.py
FabioLuporini: So that code becomes self-explanatory when used

devito/function.py
mloubout: So that you can take derivatives of sparse function. Still trying to find a nice way without overhead to do it.

devito/ir/support/space.py
mloubout: Not sure I understand, but trust you on this.

tests/test_operator.py
FabioLuporini: well, doable, but isn't it more natural to have equations with fixed array indices to update the boundary ?

tests/test_operator.py
FabioLuporini: it's exactly the same behaviour as ``master`` ATM.

What I wrote on Slack was for the next PR

devito/function.py
FabioLuporini: yep, that's what I meant. OK

devito/function.py
FabioLuporini: 👍

devito/ir/equations/equation.py
FabioLuporini: does it? @ggorman suggested to have ``DOMAIN`` as default. Anyway, this would be for another day, as changing it to ``INTERIOR`` would require changing all our examples/tests, which assume DOMAIN as default...

tests/test_operator.py
mloubout: So this one is tricky, indices are tricky to manage for boundary for non 1D problems. This is not for this PR, but I think a nice way to have "automatic" indexing of the boundary would help a lot for 2D/3D.

devito/ir/support/space.py
FabioLuporini: Basically, using appropriate names at the right place helps people unfamiliar with the code understanding what's going on. Here I thought using ``DataSpace(...)`` instead of ``Space(...)`` was better in terms of self-documented code. Just a minor thing though.

devito/ir/equations/equation.py
mloubout: I suppose this is not the point of this PR yes. Would be good to keep in mind still (can assign to me).

tests/test_operator.py
FabioLuporini: because you have an explosion in the number of equations?

mloubout: Yes, and also the iteration space becomes quite tricky to write by hand for non specialist.

FabioLuporini: Lets design BOUNDARY together next month

FabioLuporini:
@navjotk and I agreed that this is the next PR in the line for merge.  (@mlange05 )

FabioLuporini:
Updated to the new argument derivation strategy. Now requires #481 to land.

devito/operator.py
```diff
@@ -127,11 +127,14 @@ def __init__(self, expressions, **kwargs):
...
         """
-        Derive all default values from parameters and ensure uniqueness..
+        Derive all default values from parameters and ensure uniqueness.
         """
         default_args = ArgumentMap()
         for p in self.input:
             default_args.update(p.argument_defaults())
+        for p in self.dimensions:
+            if p.is_Sub:
+                default_args.update(p.argument_defaults(default_args))
```
navjotk: Was this not covered in the arguments PR?

FabioLuporini: no because ``SubDimension`` is a new type introduced in this PR. I think I have a way of improving this bit of code, but I'll need another PR for that

PR #483, opened 2018-01-31
Better SparseFunction

FabioLuporini:
Distinguish between ``SparseFunction`` and ``SparseTimeFunction``. Significant use of inheritance to avoid redundant code. 

~Should go in after #481 and #482 ~
A requirement for the domain-allocation switch, as I need to see the ``time_order`` value in all time functions (including the sparse ones, indeed)

~EDIT: this PR is open to debate of course~

EDIT: turns out I had to do deeper clean up and changes. I think the function hierarchy now looks way better. One caveat is that now modulo buffered iteration works in a slightly different way -- better in my view.
If you say that a time function has for example ``time_order=3`` **and** your equations are actually usually a number of slots ``N`` such that ``N < 3 + 1``, then only those ``N`` slots will be iterated over. Look at the small changes that I made to the arguments tests in ``test_operator.py`` for practical examples.

### DIFF
--- a/devito/function.py
+++ b/devito/function.py
@@ -1,6 +1,7 @@
 from collections import OrderedDict, namedtuple
 from functools import partial
 from math import ceil
+import abc
 
 import sympy
 import numpy as np
@@ -101,8 +102,13 @@ class Constant(AbstractCachedSymbol):
 class TensorFunction(SymbolicFunction):
 
     """
-    Utility class to encapsulate all symbolic :class:`Function` types
-    that represent tensor (array) data.
+    Utility class to encapsulate all symbolic types that represent
+    tensor (array) data.
+
+    .. note::
+
+        Users should not instantiate this class. Use :class:`Function` or
+        :class:`SparseFunction` (or their subclasses) instead.
     """
 
     is_TensorFunction = True
@@ -145,6 +151,62 @@ class TensorFunction(SymbolicFunction):
         return wrapper
 
     @property
+    def _offset_domain(self):
+        """
+        The number of grid points between the first (last) allocated element
+        (possibly in the halo/padding region) and the first (last) domain element,
+        for each dimension.
+        """
+        left = tuple(np.add(self._extent_halo.left, self._extent_padding.left))
+        right = tuple(np.add(self._extent_halo.right, self._extent_padding.right))
+
+        Offset = namedtuple('Offset', 'left right')
+        offsets = tuple(Offset(i, j) for i, j in np.add(self._halo, self._padding))
+
+        return EnrichedTuple(*offsets, left=left, right=right)
+
+    @property
+    def _offset_halo(self):
+        """
+        The number of grid points between the first (last) allocated element
+        (possibly in the halo/padding region) and the first (last) halo element,
+        for each dimension.
+        """
+        left = self._extent_padding.left
+        right = self._extent_padding.right
+
+        Offset = namedtuple('Offset', 'left right')
+        offsets = tuple(Offset(i, j) for i, j in self._padding)
+
+        return EnrichedTuple(*offsets, left=left, right=right)
+
+    @property
+    def _extent_halo(self):
+        """
+        The number of grid points in the halo region.
+        """
+        left = tuple(zip(*self._halo))[0]
+        right = tuple(zip(*self._halo))[1]
+
+        Extent = namedtuple('Extent', 'left right')
+        extents = tuple(Extent(i, j) for i, j in self._halo)
+
+        return EnrichedTuple(*extents, left=left, right=right)
+
+    @property
+    def _extent_padding(self):
+        """
+        The number of grid points in the padding region.
+        """
+        left = tuple(zip(*self._padding))[0]
+        right = tuple(zip(*self._padding))[1]
+
+        Extent = namedtuple('Extent', 'left right')
+        extents = tuple(Extent(i, j) for i, j in self._padding)
+
+        return EnrichedTuple(*extents, left=left, right=right)
+
+    @property
     def shape(self):
         """
         Shape of the domain associated with this :class:`TensorFunction`.
@@ -152,6 +214,42 @@ class TensorFunction(SymbolicFunction):
         stencil update.
         """
         return self.shape_domain
+
+    @abc.abstractproperty
+    def shape_domain(self):
+        """
+        Shape of the domain associated with this :class:`TensorFunction`.
+        The domain constitutes the area of the data written to in a
+        stencil update.
+
+        .. note::
+
+            Alias to ``self.shape``.
+        """
+        return
+
+    @property
+    def shape_with_halo(self):
+        """
+        Shape of the domain plus the read-only stencil boundary associated
+        with this :class:`Function`.
+        """
+        # TODO: for the domain-allocation switch, this needs to return the shape
+        # of the data including the halo region, ie:
+        # `tuple(j + i + k for i, (j, k) in zip(self.shape_domain, self._halo))`
+        raise NotImplementedError
+
+    @property
+    def shape_allocated(self):
+        """
+        Shape of the allocated data associated with this :class:`Function`.
+        It includes the domain and halo regions, as well as any additional
+        padding outside of the halo.
+        """
+        # TODO: for the domain-allocation switch, this needs to return the shape
+        # of the data including the halo and padding regions, ie:
+        # `tuple(j + i + k for i, (j, k) in zip(self.shape_with_halo, self._padding))`
+        raise NotImplementedError
 
     @property
     def data(self):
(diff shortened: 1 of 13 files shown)


devito/function.py
```diff
@@ -636,48 +649,24 @@ class TimeFunction(Function):
...
-                self.time_order = time_order
-                self._halo = ((left_points, right_points),) + self._halo
-            else:
-                raise ValueError("'space_order' must be int or 3-tuple of ints")
+            super(TimeFunction, self).__init__(**kwargs)
 
-            self._padding = (kwargs.get('time_padding', (0, 0)),) + self._padding
+            # Check we won't allocate too much memory for the system
+            available_mem = virtual_memory().available
+            if np.dtype(self.dtype).itemsize * self.size > available_mem:
```
FabioLuporini: note for reviewers: using ``self.size`` instead of ``self.save`` should be intended as a bug fix

devito/function.py
```diff
@@ -690,17 +679,34 @@ def _indices(cls, **kwargs):
...
-        if self.save:
-            tsize = self.time_size - self.staggered[0]
+    @classmethod
+    def __shape_setup__(cls, **kwargs):
+        grid = kwargs.get('grid')
+        save = kwargs.get('save') or None  # Force to None if 0/False/None/...
+        shape = kwargs.get('shape')
+        time_order = kwargs.get('time_order', 1)
+
+        if grid is None:
```
FabioLuporini: note for reviewers: better input checks here

devito/function.py
```diff
@@ -816,61 +778,51 @@ class SparseFunction(CompositeFunction):
...
+            self.npoint = npoint
+
+            # Grid must be provided
+            grid = kwargs.get('grid')
+            if kwargs.get('grid') is None:
+                raise ValueError('SparseFunction objects require a grid parameter')
+            self.grid = grid
+
+            self.dtype = kwargs.get('dtype', self.grid.dtype)
+            self.space_order = kwargs.get('space_order', 1)
```
mloubout: Would default to 0, `space_order` not really defined for now for Sparse Functions.

devito/function.py
```diff
@@ -816,61 +778,51 @@ class SparseFunction(CompositeFunction):
...
-        Full allocated shape of the data associated with this
-        :class:`SparseFunction`.
+        Return the default dimension indices for a given data shape.
         """
-        return (self.nt, self.npoint) if self.nt > 0 else (self.npoint, )
+        dimensions = kwargs.get('dimensions')
+        if dimensions is not None:
+            return dimensions
+        else:
+            return (Dimension(name='p'),)
```
mloubout: Fixed it in my branch, but this makes the same default dimension for all sparse functions which is risky.

tests/test_operator.py
```diff
@@ -467,26 +467,26 @@ def test_override_timefunction_data(self):
...
+        op(t=1, a=a2)
+        assert (a2.data[0] == 6.).all()
 
         # Override with user-allocated numpy data
         a3 = np.zeros_like(a.data)
         a3[:] = 4.
-        op(t=2, a=a3)
-        assert (a3[:] == 7.).all()
+        op(t=1, a=a3)
+        assert (a3[0] == 7.).all()
```
mloubout: Why did these ones change (`t`, indices `mask`)?

tests/test_operator.py
```diff
@@ -467,26 +467,26 @@ def test_override_timefunction_data(self):
 
         # Run with default value
         a.data[:] = 1.
-        op(t=2)
-        assert (a.data[:] == 4.).all()
+        op(t=1)
+        assert (a.data[0] == 4.).all()
```
mloubout: Maybe would add
`assert (a.data[1] == 3.).all()`
as only `t=0` should be updated

tests/test_operator.py
```diff
@@ -467,26 +467,26 @@ def test_override_timefunction_data(self):
...
 
         # Override with symbol (different name)
         a1 = TimeFunction(name='a1', grid=grid)
         a1.data[:] = 2.
-        op(t=2, a=a1)
-        assert (a1.data[:] == 5.).all()
+        op(t=1, a=a1)
+        assert (a1.data[0] == 5.).all()
```
mloubout: same

tests/test_operator.py
```diff
@@ -467,26 +467,26 @@ def test_override_timefunction_data(self):
...
 
         # Override with symbol (same name as original)
         a2 = TimeFunction(name='a', grid=grid)
         a2.data[:] = 3.
-        op(t=2, a=a2)
-        assert (a2.data[:] == 6.).all()
+        op(t=1, a=a2)
+        assert (a2.data[0] == 6.).all()
```
mloubout: same

devito/function.py
FabioLuporini: 👍

devito/function.py
FabioLuporini: flaw in the test suite then? I'll wait for your PR to fix it

tests/test_operator.py
FabioLuporini: 👍

tests/test_operator.py
FabioLuporini: perhaps unnecessary, I might have changed it only for homogeneity with the previous lines. I'll check that

devito/function.py
mloubout: My PR is on top of yours, Ill fix it in following (only a problem when start taking FD that calls `new` without `dimension` I think)

tests/test_operator.py
FabioLuporini: OK so if you leave ``t=2``, then you get:
- ``a3[0] == 10``
- ``a3[1] == 4``
this is because of what I explained in the PR description. IOW, there is no stencil in time for this equation, so the buffer slot 0 is the only one used (ie, updates occur in ``a3[0]`` only). If you do ``t=2``, you bring it from 4 to 10; with ``t=1`` (as here) we go from 4 to 7. ``a3[1]`` stays at 4. I'll assert that

mloubout: The stencil is `u(t) = u(t) + 3`  with everything initialized at 4.
`t=1` => `u[0] =7` the rest still at 4
`t=2` =>  `u[0] = u[1] = 7` the rest still at 4

and so on. If you get
`a3[0] == 10`
something is not right.

FabioLuporini: there are only two buffer slots, and ``u[1]`` is never incremented. So I really don't understand what's the issue here. My assert checks against 7 after one timestep, so...

mloubout: The generated code implements `u(t) = u(t-1) + 3`, not `u(t) = u(t) +3`. 

If `u` is initialized with `3` everywhere, every time, then no matter how many time steps you run (including `t=2`, then `u` can only be equal to 7.

mloubout: Actually, thinking about it, this stencil cannot be implemented without solving the full time history.

FabioLuporini: > The generated code implements u(t) = u(t-1) + 3, not u(t) = u(t) +3

??

isn't the Operator:
```
op = Operator(Eq(a, a + 3), dle=None)
```

mloubout: it is but the C code is `u[t1] = u[t1] + 3` (ot t0 don;t rememeber with `t1=time%1`.

So at the second time step, `u[t1]` contains u(t=0) and does `u[t1 (t=1)] = u[t1 (t=0)] + 3`.

FabioLuporini: > The generated code implements u(t) = u(t-1) + 3, not u(t) = u(t) +3.

I checked again. Here's the generated code:
```
int Kernel(float *restrict a_vec, const int time_e, const int time_s, void *_timers, const int x_e, const int x_s, const int x_size, const int y_e, const int y_s, const int y_size, const int z_e, const int z_s, const int z_size)
{
  float (*restrict a)[x_size][y_size][z_size] __attribute__((aligned(64))) = (float (*)[x_size][y_size][z_size]) a_vec;
  struct profile *timers = (struct profile*) _timers;
  for (int time = time_s, t0 = (time)%(1); time < time_e; time += 1, t0 = (time)%(1))
  {
    struct timeval start_section_0, end_section_0;
    gettimeofday(&start_section_0, NULL);
    for (int x = x_s; x < x_e; x += 1)
    {
      for (int y = y_s; y < y_e; y += 1)
      {
        for (int z = z_s; z < z_e; z += 1)
        {
          a[t0][x][y][z] = a[t0][x][y][z] + 3;
        }
      }
    }
    gettimeofday(&end_section_0, NULL);
    timers->section_0 += (double)(end_section_0.tv_sec-start_section_0.tv_sec)+(double)(end_section_0.tv_usec-start_section_0.tv_usec)/1000000;
  }
  return 0;
}
```

I'm not sure why you keep saying the generated code does not implement ``u[t] = u[t] + 3``. It does. ``t_s`` starts at 0 and ramps up to 1, given that I imposed ``op.apply(t=1, ...``). So ``a3[0]`` increases by 3, while ``a3[1]`` stays at 4.

mloubout:
GTG

FabioLuporini:
Merged.

PR #484, opened 2018-02-02
safer fd

mloubout:
Slightly safer FD shortcuts generation and laplacian.

PS: needs some cleanup but just to know if fine like that.


devito/function.py
```diff
@@ -223,6 +223,7 @@ def _initialize_derivatives(self):
         """
         Dynamically create notational shortcuts for space derivatives.
         """
+        derivatives = ()
```
FabioLuporini: always use a ``list``, not ``tuple``, when you have full control and you're going to change its content. The difference is:
- if you do ``list.append`` then you're modifying the list "in place"
- if you use ``+=`` (no matter whether it's a tuple, or list, or a string ...), then you're constructing a **new** object. 

Although this is practically irrelevant here, it's useful to keep this distinction in mind because in some contexts it really does make a difference

devito/function.py
```diff
@@ -273,15 +278,22 @@ def _initialize_derivatives(self):
...
+                                    order=self.space_order)
+                    setattr(self.__class__, 'd%s2%s2' % (dim.name, dim2.name),
+                            property(dx2y2, 'Return the symbolic expression for '
+                                     'the second cross derivative wrt. the '
+                                     '%s and %s dimensions' %
+                                     (dim.name, dim2.name)))
+                    derivatives += ('d%s2%s2' % (dim.name, dim2.name), )
+        if len(derivatives) > 0:
+            debug("Spatial derivatives shortcuts generated for " + str(self) +
+                  " are: \n %s" % list(derivatives))
```
FabioLuporini: I would keep everything on a single line (ie no ``\n``) to minimise noise

devito/function.py
```diff
@@ -273,15 +278,22 @@ def _initialize_derivatives(self):
...
+        else:
+            debug("No spatial derivatives shortcuts generated for " + str(self))
```
FabioLuporini: Perhaps these 4 lines can be turned into:
``debug("Spatial derivatives supported by %s are: %s" % (self, list(derivatives)))``
the empty list simply means no spatial derivatives were generated

devito/function.py
```diff
@@ -231,40 +232,44 @@ def _initialize_derivatives(self):
                     property(dx, 'Return the symbolic expression for '
                              'the centered first derivative wrt. '
                              'the %s dimension' % dim.name))
-
+            derivatives += ('d%s' % dim.name, )
             # First derivative, left
             dxl = partial(first_derivative, order=self.space_order,
```
FabioLuporini: I'm concerned these checks still don't capture all possible derivatives that **cannot** be taken on the given ``Function``.

For **example**, think about the way we're constructing ``dxl`` and ``dxr``: we use the ``space_order`` of the function. The ``space_order`` is also used to build up the ``halo`` region. So here what's happening (to sum up) is that the number of left/right points required for ``dxl/dxr`` increases faster than the number of available points on the left/right  (see how ``_halo`` is computed in the ``Function`` constructor). 

So we should probably use ``_halo``, not ``space_order``. However, this checks would complicate this method, perhaps too much. Alternatively, we could change how ``first_derivative`` and ``generic_derivative`` behave: we could pass them a different parameter -- let's call it ``S`` -- representing "the number of grid points available on the left and on the right of a generic point x0" (basically the ``_halo``); then these function would return **only** the derivatives that can be taken given ``S``.

Does it make sense?

devito/function.py
```diff
@@ -506,20 +518,28 @@ def laplace(self):
         Generates a symbolic expression for the Laplacian, the second
         derivative wrt. all spatial dimensions.
         """
-        derivs = tuple('d%s2' % d.name for d in self.space_dimensions)
+        if self.space_order < 2:
```
FabioLuporini: same?

devito/function.py
FabioLuporini: I keep thinking about this because it causes a bug in the domain-allocation switch branch. So perhaps what's wrong is the way we're constructing ``dxl/dxr``. We should be using ``halo``, not ``space_order``, to generate "the possible maximum order left/right derivative given the available neighboring points".

FabioLuporini: one last update:
if you cherry-pick this commit: https://github.com/opesci/devito/commit/77b900feb1f21b91ed122dd8068798deab02e6e5
you can get the number of left/right points available along a ``Function`` dimension via ``self._extent_halo[x]`` (where ``x`` is a ``Dimension`` and of course it can be any function indices). To be more precise, if you do ``self._extent_halo[x].left`` or ``self._extent_halo[x].right`` you actually get the left/right values.

mloubout: I have to disagree. The FD is defined by its `space_order`. I understand that the compiler would prefer it not it too, but we cannot start to generate stencils based on compiler or domain propeties.
Fd part shouldn't know anything about domain or halo ever, it generates the symbolci  fd scheme the user requires from the spatial order. If we start hacking into FD and change it in the middle of nowhere, this is gonna be problematic.

Devito is
- First generate a stencil
- Second generate code and run

If we start choosing what we run before we generate the stencil this will be a nightmare.

devito/function.py
mloubout: ok

devito/function.py
mloubout: What I had originaly but looked weird, can put it back.

devito/function.py
mloubout: WIll change it to list

devito/function.py
```diff
@@ -438,6 +438,7 @@ def __init__(self, *args, **kwargs):
             self._padding = padding
 
             # Dynamically add derivative short-cuts
+            self.derivatives = self.derivatives if hasattr(self, 'derivatives') else []
```
FabioLuporini: I really don't like this. The presence of ``hasattr`` is a symptom we're doing this in the wrong place (do we need an extra class?). Why was this necessary?

FabioLuporini: btw, why do we even need this instance attribute ``derivatives``? when do we need it?

devito/function.py
```diff
@@ -452,40 +453,44 @@ def _initialize_derivatives(self):
                     property(dx, 'Return the symbolic expression for '
                              'the centered first derivative wrt. '
                              'the %s dimension' % dim.name))
-
+            self.derivatives += ["d%s" % dim.name]
```
FabioLuporini: I suggest to leave a blank line as before to neatly separate different code blocks

devito/function.py
```diff
@@ -631,6 +647,9 @@ class TimeFunction(Function):
 
     def __init__(self, *args, **kwargs):
         if not self._cached():
+            time_order = kwargs.get('time_order', 1)
+            self.derivatives = ["dt"] if time_order > 0 else []
+            self.derivatives += ["dt2"] if time_order > 1 else []
```
FabioLuporini: again: why do we even need this instance attribute ``derivatives``? when do we need it?

devito/function.py
```diff
@@ -709,7 +728,10 @@ def backward(self):
     def dt(self):
         """Symbol for the first derivative wrt the time dimension"""
         _t = self.indices[0]
-        if self.time_order == 1:
+        if self.time_order == 0:
+            warning("Time oder is 0, no time derivatives available, returning 0")
```

devito/function.py
```diff
@@ -709,7 +728,10 @@ def backward(self):
...
+            return 0
+        elif self.time_order == 1:
             # This hack is needed for the first-order diffusion test
             indices = [_t, _t + _t.spacing]
         else:
```
FabioLuporini: can we not merge this ``else`` with the ``time_order == 1`` case?

devito/function.py
```diff
@@ -721,6 +743,10 @@ def dt(self):
     @property
     def dt2(self):
         """Symbol for the second derivative wrt the t dimension"""
+        if self.time_order < 2:
+            warning("Time oder smaller than 2, no second time derivatives available," +
```

devito/function.py
```diff
@@ -721,6 +743,10 @@ def dt(self):
...
+                    " returning 0")
+            return 0
```
FabioLuporini: either use ``else`` here, or remove ``else`` from the other methods, for consistency

devito/function.py
mloubout: So that people can access easily the list of available shortcuts.

devito/function.py
mloubout: ok

devito/equation.py
```diff
@@ -2,7 +2,22 @@
...
 import sympy
 
-__all__ = ['Eq', 'Inc', 'DOMAIN', 'INTERIOR']
+__all__ = ['Eq', 'solve', 'Inc', 'DOMAIN', 'INTERIOR']
+
+
+def solve(*args, **kwargs):
+
+    """
+    Devito `sympy.slove` that bypasses operations simplifications and
```
FabioLuporini: typo.

devito/equation.py
```diff
@@ -2,7 +2,22 @@
...
+    rational values evaluations
+
+    Skiping  simplifcations as it is really slow for complex equations and done
```
FabioLuporini: typo

devito/equation.py
```diff
@@ -2,7 +2,22 @@
...
+    again by DSE for optimizations.
+    Skiping rational to avoid floating point indices
```
FabioLuporini: typo

devito/function.py
```diff
@@ -536,26 +551,45 @@ def _halo_indices(self):
         """Return the function indices for which a halo region is defined."""
         return self.indices
 
+    @property
+    def derivatives(self):
+        """Return the list of available derivatives shortcuts"""
+        return self._derivatives + ["div", "laplace", "laplace2"]
```
FabioLuporini: Why don't we put div, Laplace and laplace2 in ``initialize_derivatives`` as well?

mloubout: `laplace2` is a function, not a property, I guess the other two can.

mloubout:
Better land #483 first than I'll fix this for all types

devito/function.py
mloubout: Looks like we (maybe just I) can't. These div, laplace need to access `self.dx`, .... and it seems to break everything if done inside `initialize_derivatives` (just try `print(self.dx)` after setting it up )

devito/equation.py
FabioLuporini: reminder

devito/equation.py
FabioLuporini: reminder

mloubout:
>I still don't understand why is this PR changing the jupyter notebooks

Sympy `solve` to Devito `solve`.

mloubout:
Will reopen and do nicer way once #493 and #494 are in,

Issue #485, opened 2018-02-03
Memory allocation


mloubout:
The new `master` duplicates all memory allocations introduced by  #481.

```
Allocating memory for m ((130, 130, 130))
Allocating memory for damp ((130, 130, 130))
Allocating memory for src ((83, 1))
Allocating memory for src_coords ((1, 3))
Allocating memory for nrec_coords ((50, 3))
Applying Forward
Allocating memory for rec_coords ((50, 3))
DSE: extract_time_invariants [flops: 63, elapsed: 0.00] >>
     eliminate_inter_stencil_redundancies [flops: 63, elapsed: 0.00] >>
     eliminate_intra_stencil_redundancies [flops: 54, elapsed: 0.04] >>
     factorize [flops: 39, elapsed: 0.03] >>
     finalize [flops: 43, elapsed: 0.00]
     [Total elapsed: 0.07 s]
DLE: avoid_denormals [elapsed: 0.00] >>
     loop_fission [elapsed: 0.01] >>
     loop_blocking [elapsed: 0.07] >>
     simdize [elapsed: 0.14] >>
     ompize [elapsed: 0.02] >>
     create_elemental_functions [elapsed: 0.07] >>
     minimize_remainders [elapsed: 0.01]
     [Total elapsed: 0.32 s]
Allocating memory for rec ((83, 50))
Allocating memory for rec_coords ((50, 3))
Allocating memory for src ((83, 1))
Allocating memory for src_coords ((1, 3))
Allocating memory for u ((3, 130, 130, 130))
Allocating memory for rec ((83, 50))
Allocating memory for u ((3, 130, 130, 130))
GNUCompiler: compiled /var/folders/mx/qs0dn9rx7zn6dz2zvwv7tkk00000gn/T/devito-_yy9ksh8/e19f5d91a33049d44e68b0a917f6d762332176cb.c [0.45 s]
=========================================================================================
Section section_1<81,1> with OI=1.05 computed in 0.000 s [0.10 GFlops/s]
Section section_2<81,50> with OI=2.12 computed in 0.004 s [0.10 GFlops/s]
Section main<81,124,124,124> with OI=2.28 computed in 0.824 s [8.06 GFlops/s, 0.19 GPts/s]
=========================================================================================
```

mloubout:
Fixed by #486, closing.

PR #486, opened 2018-02-05
switch input/default order

mloubout:
This should fix Issue #485 by switching the oder of argument processing.

~`Test_checkpoint` is broken I have no idea why and gonna take me a bi to fix navigating through its architecture~ Fixed

### DIFF
--- a/devito/function.py
+++ b/devito/function.py
@@ -521,14 +521,17 @@ class Function(TensorFunction):
         return sum([second_derivative(first * weight, dim=d, order=order)
                     for d in self.space_dimensions])
 
-    def argument_defaults(self, alias=None):
+    def argument_defaults(self, alias=None, data=True):
         """
         Returns a map of default argument values defined by this symbol.
 
         :param alias: (Optional) name under which to store values.
         """
         key = alias or self.name
-        args = ArgumentMap({key: self._data_buffer})
+        if data:
+            args = ArgumentMap({key: self._data_buffer})
+        else:
+            args = ArgumentMap()
 
         # Collect default dimension arguments from all indices
         for i, s, o in zip(self.indices, self.shape, self.staggered):
--- a/devito/function.py
+++ b/devito/function.py
@@ -79,7 +79,7 @@ class Constant(AbstractCachedSymbol):
     def base(self):
         return self
 
-    def argument_defaults(self):
+    def argument_defaults(self, **kwargs):
         """
         Returns a map of default argument values defined by this symbol.
         """
@@ -760,13 +760,13 @@ class CompositeFunction(Function):
     def children(self):
         return self._children
 
-    def argument_defaults(self, alias=None):
+    def argument_defaults(self, alias=None, data=True):
         """
         Returns a map of default argument values defined by this symbol.
 
         :param alias: (Optional) name under which to store values.
         """
-        args = super(CompositeFunction, self).argument_defaults(alias=alias)
+        args = super(CompositeFunction, self).argument_defaults(alias=alias, data=data)
         for child in self.children:
             args.update(child.argument_defaults())
         return args
--- a/devito/function.py
+++ b/devito/function.py
@@ -824,7 +824,7 @@ class SparseFunction(TensorFunction):
         if dimensions is not None:
             return dimensions
         else:
-            return (Dimension(name='p'),)
+            return (Dimension(name='p_%s' % kwargs["name"]),)
 
     @classmethod
     def __shape_setup__(cls, **kwargs):
@@ -1030,16 +1030,20 @@ class SparseFunction(TensorFunction):
         :param alias: (Optional) name under which to store values.
         """
         # Take a copy of the replacement before super pops it from kwargs
+
         new = kwargs.get(self.name)
-        values = super(SparseFunction, self).argument_values(alias=alias, **kwargs)
+        key = alias or self.name
+
+        values = super(SparseFunction, self).argument_values(alias=key, **kwargs)
+
         if new is not None and isinstance(new, SparseFunction):
             # If we've been replaced with a SparseFunction,
             # we need to re-derive defaults and values...
-            values.update(new.argument_defaults(alias=alias).reduce_all())
+            values.update(new.argument_defaults(alias=key).reduce_all())
             values.update(new.coordinates.argument_defaults(alias=self.coordinates.name))
         else:
             # ..., but if not, we simply need to recurse over children.
-            values.update(self.coordinates.argument_values(alias=alias, **kwargs))
+            values.update(self.coordinates.argument_values(alias=key, **kwargs))
         return values
 
 
@@ -1088,7 +1092,7 @@ class SparseTimeFunction(SparseFunction):
         if dimensions is not None:
             return dimensions
         else:
-            return (kwargs['grid'].time_dim, Dimension(name='p'))
+            return (kwargs['grid'].time_dim, Dimension(name='p_%s' % kwargs["name"]))
 
     @classmethod
     def __shape_setup__(cls, **kwargs):
--- a/devito/function.py
+++ b/devito/function.py
@@ -349,13 +349,6 @@ class TensorFunction(SymbolicFunction):
         # Add value override for own data if it is provided
         if self.name in kwargs:
             new = kwargs.pop(self.name)
-            if isinstance(new, TimeFunction):
-                if new.save != self.save:
-                    raise TypeError("Operator generated for save=%s only," %
-                                    self.save +
-                                    " %s has to be created as " % new.name +
-                                    "TimeFunction(name=%s, grid=grid, save=%s)" %
-                                    (new.name, self.save))
             if isinstance(new, TensorFunction):
                 # Set new values and re-derive defaults
                 values[key] = new._data_buffer
@@ -672,7 +665,7 @@ class TimeFunction(Function):
 
             self.time_dim = kwargs.get('time_dim')
             self.time_order = kwargs.get('time_order', 1)
-            self.save = kwargs.get('save', None)
+            self.save = kwargs.get('save', None) is not None
             if not isinstance(self.time_order, int):
                 raise ValueError("'time_order' must be int")
 
@@ -759,6 +752,26 @@ class TimeFunction(Function):
         indt = [(_t + i * _t.spacing) for i in range(-width_t, width_t + 1)]
 
         return self.diff(_t, _t).as_finite_difference(indt)
+
+    def argument_values(self, alias=None, **kwargs):
+        """
+        Returns a map of argument values after evaluating user input.
+
+        :param kwargs: Dictionary of user-provided argument overrides.
+        :param alias: (Optional) name under which to store values.
+        """
+        saves = {True: 'int', False: 'None'}
+        # Check if data has the right dimension
+        if self.name in kwargs:
+            new = kwargs.get(self.name)
+            if ~isinstance(new, TimeFunction):
+                pass
+            elif new.save != self.save:
+                raise TypeError("Incorrect value encounterd, save should be %s" %
+                                saves[self.save])
+
+        values = super(TimeFunction, self).argument_values(alias=alias, **kwargs)
+        return values
 
 
 class SparseFunction(TensorFunction):
--- a/devito/function.py
+++ b/devito/function.py
@@ -533,6 +533,7 @@ class Function(TensorFunction):
         # Collect default dimension arguments from all indices
         for i, s, o in zip(self.indices, self.shape, self.staggered):
             args.update(i.argument_defaults(size=s+o))
+
         return args
 
     def argument_values(self, alias=None, **kwargs):
--- a/devito/function.py
+++ b/devito/function.py
@@ -662,7 +662,7 @@ class TimeFunction(Function):
 
             self.time_dim = kwargs.get('time_dim')
             self.time_order = kwargs.get('time_order', 1)
-            self.save = kwargs.get('save', None) is not None
+            self.save = type(kwargs.get('save', None) or None)
             if not isinstance(self.time_order, int):
                 raise ValueError("'time_order' must be int")
 
@@ -755,13 +755,12 @@ class TimeFunction(Function):
         :param kwargs: Dictionary of user-provided argument overrides.
         :param alias: (Optional) name under which to store values.
         """
-        saves = {True: 'int', False: 'None'}
         # Check if data has the right dimension
         if self.name in kwargs:
             new = kwargs.get(self.name)
             if isinstance(new, TimeFunction) and new.save != self.save:
-                raise TypeError("Incorrect value encounterd, save should be %s" %
-                                saves[self.save])
+                raise TypeError("Incorrect value encountered, save should be %s" %
+                                self.save)
 
         values = super(TimeFunction, self).argument_values(alias=alias, **kwargs)
         return values
--- a/devito/function.py
+++ b/devito/function.py
@@ -349,6 +349,13 @@ class TensorFunction(SymbolicFunction):
         # Add value override for own data if it is provided
         if self.name in kwargs:
             new = kwargs.pop(self.name)
+            if isinstance(new, TimeFunction):
+                if new.save != self.save:
+                    raise TypeError("Operator generated for save=%s only," %
+                                    self.save +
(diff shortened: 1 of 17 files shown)


devito/operator.py
```diff
@@ -124,31 +123,32 @@ def __init__(self, expressions, **kwargs):
...
         """
         Process runtime arguments passed to ``.apply()` and derive
         default values for any remaining arguments.
         """
-        # First, derive all default values from parameters
-        arguments = self._argument_defaults.copy()
-
-        # Next, we insert user-provided overrides
+        arguments = {}
+        # # First, we insert user-provided override
```
FabioLuporini: Extra #
overrides

devito/operator.py
```diff
@@ -124,31 +123,32 @@ def __init__(self, expressions, **kwargs):
         # Finish instantiation
         super(Operator, self).__init__(self.name, nodes, 'int', parameters, ())
 
-    @cached_property
-    def _argument_defaults(self):
+    def _argument_defaults(self, arguments):
```
FabioLuporini: So this method is useless at this point. I think you can take the code and put it directly into ``def arguments``.

devito/operator.py
```diff
@@ -124,31 +123,32 @@ def __init__(self, expressions, **kwargs):
...
         """
         default_args = ArgumentMap()
         for p in self.input:
-            default_args.update(p.argument_defaults())
+            if p.name not in arguments:
+                default_args.update(p.argument_defaults())
         for p in self.dimensions:
-            if p.is_Sub:
-                default_args.update(p.argument_defaults(default_args))
+            if p.name not in arguments:
```
FabioLuporini: maybe unnecessary adding a guard for dimensions

mloubout:
Just need to fix the `_argumet_default` for Yask, not sure what to do with it.

devito/operator.py
```diff
@@ -124,30 +123,29 @@ def __init__(self, expressions, **kwargs):
...
         for p in self.input + self.dimensions:
             arguments.update(p.argument_values(**kwargs))
+        # Second, derive all remaining default values from parameters
+        default_args = ArgumentMap()
+        for p in self.input:
+            if p.name not in arguments:
+                default_args.update(p.argument_defaults())
+            else:
+                default_args.update(p.argument_defaults(data=False))
+
```
FabioLuporini: nitpicking: don't leave a blank row here because the following block of code still "belongs" to the comment "Second, derive all remaining .."

devito/operator.py
```diff
@@ -124,30 +123,33 @@ def __init__(self, expressions, **kwargs):
...
-            default_args.update(p.argument_defaults())
+            if p.name not in arguments:
+                default_args.update(p.argument_defaults())
+            else:
+                default_args.update(p.argument_defaults(data=False))
         for p in self.dimensions:
-            if p.is_Sub:
-                default_args.update(p.argument_defaults(default_args))
-        return {k: default_args.reduce(k) for k in default_args}
+            if p.name not in arguments:
```
FabioLuporini: probably don't need this ``if``; if you do, you can still join it with the nested one

devito/operator.py
```diff
@@ -124,30 +123,33 @@ def __init__(self, expressions, **kwargs):
...
         """
         Derive all default values from parameters and ensure uniqueness.
         """
         default_args = ArgumentMap()
         for p in self.input:
-            default_args.update(p.argument_defaults())
+            if p.name not in arguments:
+                default_args.update(p.argument_defaults())
+            else:
+                default_args.update(p.argument_defaults(data=False))
```
FabioLuporini: I don't like this at all, though I think I see why you had to do it -- to hit this line:
https://github.com/opesci/devito/pull/486/files#diff-392d49fabd2c76220fb6a2042dc1f460R537

FabioLuporini:
I think a much better way of doing this is using ``partial``, https://docs.python.org/3/library/functools.html#functools.partial
like we're doing for derivative shortcuts
so, to sum up, what I think would be best to do is:
- we reinstate the original ordering: defaults first, overrides then
- however, defaults now return, where necessary, partial objects; so for example instead of ``_data_buffer`` we return ``partial(_data_buffer)``, or something along these lines
- an override, if present, will then kill this partial object
- finally, after ``argument_values``, we do a final pass over the populated arguments dictionary and call each partial so that data gets allocated

this way we don't have to change much code. Thoughts?

mloubout:
>this way we don't have to change much code. Thoughts?

Not sure I understand or know how to do this, but fine with me if better and fixes the allocation overhead.

devito/function.py
```diff
@@ -348,14 +349,26 @@ def argument_values(self, alias=None, **kwargs):
         # Add value override for own data if it is provided
         if self.name in kwargs:
             new = kwargs.pop(self.name)
+            if isinstance(new, TimeFunction):
+                if new.save != self.save:
+                    raise TypeError("Operator generated for save=%s only," %
+                                    self.save +
+                                    " %s has to be created as " % new.name +
+                                    "TimeFunction(name=%s, grid=grid, save=%s)" %
+                                    (new.name, self.save))
```
mloubout: Note for review. THis is actually fully necessary (as well as the change in `test_operator` as calling an operator created with/without save with a different `TimeFunction` will lead to segafaults (calling a `save=..` with a `save=None` one) or fill the history if using a buffer for `save=nt`

mloubout:
@fablup I think this should be green like that.

The only issue (not sure related to that PR) is that alias detection is broken for CompositeFunction that do not catch the dimension names properly at argument passing. Only reason it was previously working was the complete pass through default first

devito/function.py
```diff
@@ -348,14 +349,26 @@ def argument_values(self, alias=None, **kwargs):
         # Add value override for own data if it is provided
         if self.name in kwargs:
             new = kwargs.pop(self.name)
+            if isinstance(new, TimeFunction):
+                if new.save != self.save:
+                    raise TypeError("Operator generated for save=%s only," %
+                                    self.save +
```
FabioLuporini: Use template strings, not ``+`` to concatenate them (much clearer; also, this convention is loosely followed throughout Devito)

devito/function.py
```diff
@@ -1024,16 +1038,20 @@ def argument_values(self, alias=None, **kwargs):
         :param alias: (Optional) name under which to store values.
         """
         # Take a copy of the replacement before super pops it from kwargs
+
```
FabioLuporini: blank line?

devito/function.py
FabioLuporini: are you saying that if I have a ``save=None`` with say 3 buffer slots, and then I try to replace such a ``TimeFunction`` with another ``TimeFunction`` that this time has ``save=3``, we end up with a segfault?

**IF not** , then probably we still don't need ``save``. We could just use ``shape[0]``

devito/function.py
```diff
@@ -348,14 +349,26 @@ def argument_values(self, alias=None, **kwargs):
         # Add value override for own data if it is provided
         if self.name in kwargs:
             new = kwargs.pop(self.name)
+            if isinstance(new, TimeFunction):
```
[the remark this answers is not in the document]
FabioLuporini: IOW, a ``TensorFunction`` does NOT know what a ``TimeFunction`` is !

mloubout:
>are you saying that if I have a save=None with say 3 buffer slots, and then I try to replace such a TimeFunction with another TimeFunction that this time has save=3, we end up with a segfault?

Wether or not it does is not relevant, the `TimeFunction` has to match  the problem the operator solves.

mloubout:
All green, but now times out :(

FabioLuporini:
> but now times out :(

Travis is a pain today

mloubout:
Ok, back to green

FabioLuporini:
> I think I got lost -- how does this now fix the double allocation issue?

reminder @mloubout

mloubout:
>how does this now fix the double allocation issue?

Now, instead of allocating everything then replacing the defaults by the kwargs, it first go through the kwargs then allocate the remaining missing ones.

devito/function.py
```diff
@@ -659,6 +665,7 @@ def __init__(self, *args, **kwargs):
 
             self.time_dim = kwargs.get('time_dim')
             self.time_order = kwargs.get('time_order', 1)
+            self.save = kwargs.get('save', None) is not None
```
FabioLuporini: what's going on here?

devito/function.py
```diff
@@ -744,6 +751,26 @@ def dt2(self):
...
+        Returns a map of argument values after evaluating user input.
+
+        :param kwargs: Dictionary of user-provided argument overrides.
+        :param alias: (Optional) name under which to store values.
+        """
+        saves = {True: 'int', False: 'None'}
+        # Check if data has the right dimension
+        if self.name in kwargs:
+            new = kwargs.get(self.name)
+            if ~isinstance(new, TimeFunction):
```
FabioLuporini: "not" is the Devito way, not "~"

devito/function.py
mloubout: Flag to compare kwargs to default without forcing same time size.

devito/function.py
mloubout: Oh sorry, old habit will fix.

FabioLuporini: squash this and the following condition as
``if isinstance(new, TimeFunction) and new.save != self.save:``

tests/test_tti.py
```diff
@@ -82,14 +82,14 @@ def ricker_source(t, f0):
                                        time_order=2, space_order=space_order)
 
     # Create new wavefield object restart forward computation
-    u = TimeFunction(name='u', grid=model.grid, save=False,
+    u = TimeFunction(name='u', grid=model.grid, save=None,
```
FabioLuporini: shouldn't ``save=None`` be the default, hence useless here?

examples/seismic/acoustic/wavesolver.py
```diff
@@ -193,10 +193,10 @@ def born(self, dmin, src=None, rec=None, u=None, U=None, m=None, **kwargs):
 
         # Create the forward wavefields u and U if not provided
         if u is None:
-            u = TimeFunction(name='u', grid=self.model.grid, save=False,
+            u = TimeFunction(name='u', grid=self.model.grid, save=None,
```
FabioLuporini: useless to specify ``save=None`` ?

devito/operator.py
FabioLuporini: join the IFs pls

devito/function.py
```diff
@@ -355,7 +356,12 @@ def argument_values(self, alias=None, **kwargs):
             else:
                 # We've been provided a pure-data replacement (array)
                 values[key] = new
-                # TODO: Re-derive defaults from shape of array
+                if len(new.shape) != len(self.indices):
```
FabioLuporini: Here's my last concern. Shouldn't this whole block ``if-then-else`` be promoted one level up? ie shouldn't it be nested under ``if self.name in kwargs``, rathen than under the first ``else``?

the check on the length of the shape should per performed always, correct? btw, instead of ``len(self.indices)``, you should use ``self.ndim`` (defined in ``types.py``)

devito/function.py
```diff
@@ -355,7 +356,12 @@ def argument_values(self, alias=None, **kwargs):
...
+                    raise ValueError("Array shape %s does not match" % (new.shape, ) +
+                                     "dimensions %s" % (self.indices, ))
+                else:
+                    for i, s, o in zip(self.indices, new.shape, self.staggered):
```
FabioLuporini: Why is this for loop different than that at line 367 ? here we also take into account ``shape`` and ``staggered``, while below only the indices. I don't know why; but in any case, at least a comment is necessary

FabioLuporini: also, if we go through this else, it feels like we're doing the argument evaluation twice? (one here, one at line 367)

mloubout: >Why is this for loop different than that at line 367 ? 

`**kwargs` only contains the `Function` object, not the indices, so it doesn't update the same. I however should add the check to only update unset indices. Let me check if I can just remove 367

devito/function.py
mloubout: True, should be checked one level up

devito/operator.py
```diff
@@ -124,31 +123,32 @@ def __init__(self, expressions, **kwargs):
         # Finish instantiation
         super(Operator, self).__init__(self.name, nodes, 'int', parameters, ())
 
-    @cached_property
```
FabioLuporini: Since this cannot be a ``cached_property`` anymore, we now incur a performance penalty at every operator invocation. Does this have an evident impact on the user experience?

mloubout: I think this is fairly ok, we incur a really small penalty compared to the actual run for non-tiny model. I also think it's better to have a few seconds extra deriving arguments rather than duplicating memory.

FabioLuporini: "a few seconds" is a **HUGE** amount of time from the developer perspective! can you quantify the penalty? is it really seconds or just fractions of seconds? I guess you really meant fractions of seconds (ie imperceivable).

mloubout: It's like a few seconds for the full TRAVIS, so almost nothing individually.

FabioLuporini: oh, OK then. I was talking in terms of individual operators.

mloubout: Hum, checking on `examples/seismic/acoustic/acoustic_example.py` it actually takes the exact same time  (on my machine). You losse on `cached_property` but you gain on skipping duplicate allocations

devito/function.py
```diff
@@ -744,6 +748,24 @@ def dt2(self):
...
+        :param kwargs: Dictionary of user-provided argument overrides.
+        :param alias: (Optional) name under which to store values.
+        """
+        saves = {True: 'int', False: 'None'}
+        # Check if data has the right dimension
+        if self.name in kwargs:
+            new = kwargs.get(self.name)
+            if isinstance(new, TimeFunction) and new.save != self.save:
+                raise TypeError("Incorrect value encounterd, save should be %s" %
+                                saves[self.save])
```
FabioLuporini: ``save should be %s" % type(self.save))``

ie, no need for ``saves``

devito/function.py
```diff
@@ -348,18 +349,20 @@ def argument_values(self, alias=None, **kwargs):
...
             if isinstance(new, TensorFunction):
                 # Set new values and re-derive defaults
                 values[key] = new._data_buffer
                 values.update(new.argument_defaults(alias=key).reduce_all())
             else:
                 # We've been provided a pure-data replacement (array)
                 values[key] = new
-                # TODO: Re-derive defaults from shape of array
+                # Add value overrides for all associated dimensions
+                for i, s, o in zip(self.indices, new.shape, self.staggered):
```
FabioLuporini: Shouldn't this go directly under the first ``if self.name in kwargs``?

mloubout: No, this one is only necessary if an array is passed as `new. argument_defaults` will take care of indices and shape for a `TensorFunction`.

devito/function.py
mloubout: `self.save` is True or False, type(self.save) would return `<class 'bool'>` both cases instead of `int, None`

mloubout: done

devito/function.py
```diff
@@ -744,6 +748,23 @@ def dt2(self):
...
+        """
+        Returns a map of argument values after evaluating user input.
+
+        :param kwargs: Dictionary of user-provided argument overrides.
+        :param alias: (Optional) name under which to store values.
+        """
+        # Check if data has the right dimension
+        if self.name in kwargs:
+            new = kwargs.get(self.name)
+            if isinstance(new, TimeFunction) and new.save != self.save:
```
FabioLuporini: Note for the future: we should error out if ``new`` is not a TimeFunction (with a different error msg)

devito/function.py
```diff
@@ -659,6 +662,7 @@ def __init__(self, *args, **kwargs):
 
             self.time_dim = kwargs.get('time_dim')
             self.time_order = kwargs.get('time_order', 1)
+            self.save = type(kwargs.get('save', None) or None)
```
FabioLuporini: Note for the future: no need the ``, None)`` when you do ``.get`` on a dict. It's the default return value anyway.

FabioLuporini:
Thanks, merging...

FabioLuporini:
Merged.

PR #487, opened 2018-02-07
dse: extend pow_to_mul to handle negative integer powers

tjb900:
When C code is written, sympy Pow operations remaining in the expressions are converted to pow().  This is a double-valued function, and I have seen it inadvertently cause whole expressions to be evaluated as doubles rather than as floats.

pow_to_mul already converts most such powers into chains of multiplications, and this PR extends that to negative powers, which often appear in 2nd-derivative operators, e.g. as ``pow(h_x,-2)``.  It also adds some sanity checks/limits.

Please comment, in particular on how I should handle the (possibly nonexistent) case of exp=0.0

### DIFF
--- a/devito/symbolics/manipulation.py
+++ b/devito/symbolics/manipulation.py
@@ -3,7 +3,7 @@ from collections import Iterable, OrderedDict, namedtuple
 import sympy
 from sympy import Number, Indexed, Symbol, LM, LC
 
-from devito.symbolics.extended_sympy import Add, Mul, Eq, FrozenExpr
+from devito.symbolics.extended_sympy import Add, Mul, Pow, Eq, FrozenExpr
 from devito.symbolics.search import retrieve_indexed, retrieve_functions
 from devito.dimension import Dimension
 from devito.tools import as_tuple, flatten
@@ -26,6 +26,9 @@ def freeze_expression(expr):
     elif expr.is_Mul:
         rebuilt_args = [freeze_expression(e) for e in expr.args]
         return Mul(*rebuilt_args, evaluate=False)
+    elif expr.is_Pow:
+        rebuilt_args = [freeze_expression(e) for e in expr.args]
+        return Pow(*rebuilt_args, evaluate=False)
     elif expr.is_Equality:
         rebuilt_args = [freeze_expression(e) for e in expr.args]
         if isinstance(expr, FrozenExpr):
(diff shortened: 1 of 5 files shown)


tjb900:
Tests failing because of https://github.com/sympy/sympy/issues/14160

FabioLuporini:
@tmbgreaves Did #498 fix the issue?

FabioLuporini:
> @tmbgreaves Did #498 fix the issue?

reminder @tjb900

tjb900:
I was wondering if you meant me.  No, I don't think #498 would have fixed this issue.

However, the required sympy fixes look like they will land soon.  I guess once there is a conda'd release of sympy with the fixes, then we can land this (assuming it passes, obvs)?

FabioLuporini:
👍 let's leave it open for the time being, then

ggorman:
Any update on this?

tjb900:
I think this one will have to remain parked for a while.  It's dependent on this PR in sympy to be merged, AND a sympy release containing that.

https://github.com/sympy/sympy/pull/14207

Ordinarily, I wouldn't be holding my breath for a sympy release (the last third-digit release was 9 months ago), but then there is this which was only edited 2 weeks ago:

https://github.com/sympy/sympy/wiki/Release-Notes-for-1.2

Let's hope that PR makes it in 🤞 .

devito/symbolics/manipulation.py
```diff
@@ -160,11 +163,15 @@ def pow_to_mul(expr):
...
+        if exp > 10 or exp < -10 or int(exp) != exp or exp == 0:
+            # Large and non-integer powers remain untouched
             return expr
-        else:
+        elif exp > 0:
             return sympy.Mul(*[base]*exp, evaluate=False)
+        else:
+            # sympy represents 1/x as Pow(x,-1)
+            posexpr = sympy.Mul(*[base]*(-exp), evaluate=False)
+            return sympy.Pow(posexpr, -1, evaluate=False)
```
mloubout: Am I beeing stupid or is it the same as `expr` input

tjb900: It should convert negative integer powers into the combo of multiplication and reciprocal

tjb900: To be fair, it does nothing for exp==-1, so we could catch that in the untouched clause higher up.

environment.yml
```diff
@@ -5,7 +5,7 @@ channels:
 dependencies:
   - python>=3.6
   - numpy==1.14
-  - sympy==1.1
+  - sympy==1.2
```
FabioLuporini: I'm merging this, but don't we rather want >= ? not sure why it's still == in the first place

PR #488, opened 2018-02-10
Embryonic support for subsampling

FabioLuporini:
Tests in ``test_dimension.py`` show how time subsampling can be triggered. I'm not claiming this is incredibly robust, but at least it works in basic cases. Also, the design seems OK to me, so this should be at least a good start.

@tjb900 this also includes the constant argument fix (with a test) and your subsampling test with shifted dimension

EDIT: introduces a new (tiny) package dependency: ``frozendict``

### DIFF
--- a/devito/cgen_utils.py
+++ b/devito/cgen_utils.py
@@ -69,6 +69,12 @@ class CodePrinter(C99CodePrinter):
         C99CodePrinter.__init__(self, settings)
         self.known_functions.update(self.custom_functions)
 
+    def _print_Eq(self, expr):
+        if getattr(expr, '_conditional', False):
+            return "%s == %s" % (self._print(expr.lhs), self._print(expr.rhs))
+        else:
+            return super(CodePrinter, self)._print_Eq(expr)
+
     def _print_Indexed(self, expr):
         """Print field as C style multidimensional array
 
@@ -150,6 +156,9 @@ class CodePrinter(C99CodePrinter):
     def _print_ListInitializer(self, expr):
         return "{%s}" % ', '.join([self._print(i) for i in expr.params])
 
+    def _print_IntDiv(self, expr):
+        return str(expr)
+
 
 def ccode(expr, **settings):
     """Generate C++ code from an expression calling CodePrinter class
@@ -158,23 +167,7 @@ def ccode(expr, **settings):
     :param settings: A dictionary of settings for code printing
     :returns: The resulting code as a string. If it fails, then it returns the expr
     """
-    if isinstance(expr, Eq):
-        return ccode_eq(expr)
-    try:
-        return CodePrinter(settings).doprint(expr, None)
-    except:
-        return expr
-
-
-def ccode_eq(eq, **settings):
-    """Generate C++ assignment from an equation assigning RHS to LHS
-
-    :param eq: The equation
-    :param settings: A dictionary of settings for code printing
-    :returns: The resulting code as a string
-    """
-    return CodePrinter(settings).doprint(eq.lhs, None) \
-        + ' = ' + CodePrinter(settings).doprint(eq.rhs, None)
+    return CodePrinter(settings).doprint(expr, None)
 
 
 blankline = c.Line("")
--- a/devito/cgen_utils.py
+++ b/devito/cgen_utils.py
@@ -69,11 +69,8 @@ class CodePrinter(C99CodePrinter):
         C99CodePrinter.__init__(self, settings)
         self.known_functions.update(self.custom_functions)
 
-    def _print_Eq(self, expr):
-        if getattr(expr, '_conditional', False):
-            return "%s == %s" % (self._print(expr.lhs), self._print(expr.rhs))
-        else:
-            return super(CodePrinter, self)._print_Eq(expr)
+    def _print_CondEq(self, expr):
+        return "%s == %s" % (self._print(expr.lhs), self._print(expr.rhs))
 
     def _print_Indexed(self, expr):
         """Print field as C style multidimensional array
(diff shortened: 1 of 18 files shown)


devito/symbolics/extended_sympy.py
```diff
@@ -77,6 +88,29 @@ class Add(sympy.Add, FrozenExpr):
     pass
 
 
+class IntDiv(sympy.Expr):
+
+    """
+    A support type for integer division. Should only be used by the compiler
+    for code generation purporses (i.e., not for symbolic manipulation).
```
navjotk: typo

devito/dimension.py
```diff
@@ -252,8 +257,44 @@ def argument_defaults(self, parent_defaults):
...
+    is_NonlinearDerived = True
+    is_Conditional = True
+
+    """
+    Dimension symbol representing a sub-region of a given ``parent`` Dimension.
+    Unlike a :class:`SubDimension`, a ConditionalDimension does not represent
+    a contiguous region. The iterations touched by a ConditionalDimension
+    are expressible in two different ways: ::
+
+        * ``factor``: an integer indicating the size of the increment.
```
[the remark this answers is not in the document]
FabioLuporini: Agreed. As the title says, this PR brings in the basic compilation technology. Let's do it in the next subsampling PR.

tests/test_dimension.py
```diff
@@ -0,0 +1,61 @@
...
+    usave = TimeFunction(name='usave', grid=grid, save=2, time_dim=time_subsampled)
+    assert(time_subsampled in usave.indices)
+
+    t_sub_shift = Constant(name='t_sub_shift', dtype=np.int32)
+
+    eqns = [Eq(u.forward, u + 1.), Eq(u2.forward, u2 + 1.),
+            Eq(usave.subs(time_subsampled, time_subsampled - t_sub_shift), u)]
+    op = Operator(eqns)
+
+    op.apply(time_s=10, time_e=nt, t_sub_shift=3)
```
mloubout: wait, doesn't this create negative indices?

first `time_subsampled` is `4/4 - 3 = -2`?

tests/test_ir.py
```diff
@@ -408,3 +409,19 @@ def test_intervals_subtract():
     assert ix2.subtract(ix) == ix
     assert ix.subtract(ix2) == Interval(x, -2, 2)
     assert ix3.subtract(ix) == ix2
+
+
+@skipif_yask
+def test_iet_conditional(fc):
+    then_body = Expression(Eq(fc[x, y], fc[x, y] + 1))
+    else_body = Expression(Eq(fc[x, y], fc[x, y] + 2))
+    conditional = Conditional(x < 3, then_body, else_body)
```
mloubout: This is nice. Is this currently only at lower level, or can it be generalized as a "stencil" at operator creation?

tests/test_operator.py
```diff
@@ -603,6 +603,27 @@ def test_derive_constant_value(self):
...
+
+        arbdim = Dimension('arb')
+        u = TimeFunction(name='u', grid=grid, save=None, time_order=2, space_order=0)
+        snap = Function(name='snap', dimensions=(arbdim, x, y), shape=(5, nx, ny),
+                        space_order=0)
+
+        save_t = Constant(name='save_t', dtype=np.int32)
+        save_slot = Constant(name='save_slot', dtype=np.int32)
+
+        expr = Eq(snap.subs(arbdim, save_slot), u.subs(grid.stepping_dim, save_t))
```
mloubout: So we don;t need `indexed` anymore for single indexing?

FabioLuporini: @tjb900 's test, we can improve it next

devito/cgen_utils.py
```diff
@@ -69,6 +69,12 @@ def __init__(self, settings={}):
         C99CodePrinter.__init__(self, settings)
         self.known_functions.update(self.custom_functions)
 
+    def _print_Eq(self, expr):
+        if getattr(expr, '_conditional', False):
```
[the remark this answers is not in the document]
FabioLuporini: done

devito/symbolics/extended_sympy.py
FabioLuporini: done

tests/test_dimension.py
FabioLuporini: right... @tjb900 ?

navjotk: Focussing, for a second, on why this wasn't caught during testing, this sort of a problem could be caught by the arguments "engine" right?

tjb900: There shouldn't be OOB access in this case because time_s=10, so the first matching index is 12, i.e. 12/4 - 3.

The subtraction is important, from my perspective, to be able to cross-correlate small ranges of time with each call - i.e. for checkpointing.

FabioLuporini: right, dumb me!

FabioLuporini:
I think I have addressed all comments, and unless you have other concerns I'd land this @navjotk @mloubout

tests/test_dimension.py
mloubout: Oh ok didn't see that. I personaly would rather remove this one from the test base to avoid confusion, or at least put some comments above `op.apply()` like "starting at tim=10, time_sub - t_shift is in range".

FabioLuporini: > I personaly would rather remove this one from the test base to avoid confusion

I disagree

But I'll leave a comment as you suggested (in the next PR at this point, which is imminent)

FabioLuporini:
Approved, all tests passing and all comments addressed. Merging...

FabioLuporini:
Merged.

PR #489, opened 2018-02-14
Drop time_axis and reverse flag hacks

### LINKED ISSUE #281: Reverse flag for time

FabioLuporini:
The compiler is extended with the capability of detecting "iteration directions" via analysis of input expressions. This means we can now infer whether time should propagate in the forward or backward directions based on the sole analysis of the input. 

This is yet another building block for the domain allocation switch. Knowing the direction of each dimension will allow to produce more sensible defaults.

fixes #281 

As soon as this branch lands, we will have dropped probably the **last** of the ugly hacks in Devito (the global reverse flag)!


devito/ir/support/utils.py
```diff
@@ -0,0 +1,193 @@
...
+                try:
+                    if w.distance(r, d) > 0:
+                        mapper[d].add(Forward)
+                        break
+                    elif w.distance(r, d) < 0:
+                        mapper[d].add(Backward)
+                        break
+                    else:
+                        mapper[d].add(Any)
+                except TypeError:
```
mloubout: I probably understand this `if-then-else` wrong, but it looks like, for a given dimension,  it will choose the first direction it finds and stop (break), not checking the remaining satisfies it. Is it risky?

devito/ir/support/utils.py
```diff
@@ -0,0 +1,193 @@
...
+                    # Nothing can be deduced
```
mloubout: Maybe put a default to `forward` instead, as it is the "default" for what people usually want to do, with maybe a warning message.

devito/ir/support/utils.py
```diff
@@ -0,0 +1,193 @@
...
+    The iteration direction is chosen so that the information "naturally flows"
+    from an iteration to another (i.e., to generate "flow" or "read-after-write"
+    dependencies).
+
+    In the case of a clash (e.g., both Forward and Backward should be used
+    for a given dimension in order to have a flow dependence), the function
+    ``key : D -> I`` is used to pick one value.
+    """
+    mapper = detect_flow_directions(exprs)
+    return force_directions(mapper, key)
```
mloubout: I am not sure why we need this function, seems like `force_directions`  alone is enough

devito/equation.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+
+    def __init__(self, name):
+        self._name = name
+
+    def __repr__(self):
+        return self._name
+
+    def __eq__(self, other):
+        return isinstance(other, Region) and self._name == other._name
```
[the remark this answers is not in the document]
FabioLuporini: not an issue as I've changed the hierarchy; now CondEq is a plain SymPy object as it doesn't inherit anymore for our own Eqs.

devito/function.py
```diff
@@ -6,44 +6,23 @@
...
-    def __eq__(self, other):
-        return self._axis == other._axis
-
-    def __repr__(self):
-        return {-1: 'Backward', 1: 'Forward'}[self._axis]
-
-
-Forward = TimeAxis('Forward')
-Backward = TimeAxis('Backward')
+           'SparseTimeFunction']
```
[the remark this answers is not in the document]
FabioLuporini: Well, we kinda still have these objects, they're in ``ir/support/space.py`` under the name of ``IterationDirection``. It's just that we're not exposing to the user.

devito/ir/support/utils.py
FabioLuporini: Good question. So, here we're calculating the so called vector distance between **two** array accesses. In such a case, if you **have to** go forward/backward along a given axis ``i``, then you can go in **any** directions along any of the ``i + k, k > 0`` axes (ie, the inner dimensions are parallel). However, the actual direction of these ``i + k, k > 0`` axes may be dictated by other array accesses. So, short answer to your question: no, it's not risky.

devito/ir/support/utils.py
FabioLuporini: No, ``Any`` is fine. Here we are just detecting the possible directions; we are not committing to anything yet.

FabioLuporini: Much later on in the compiler, an ``Any`` direction is by default translated to ``++``

devito/ir/support/utils.py
FabioLuporini: ``detect`` doesn't make a choice; ``force`` makes a choice. ``detect`` may be computationally expensive, so it's handy to keep them separated (see ``clusters/algorithms.py``)

FabioLuporini:
Merging...

FabioLuporini:
Merged.

PR #491, opened 2018-02-15
IntDiv hotfix

### LINKED ISSUE #490: `IntDiv` return empty args
The `IntDiv` extension breaks in some cases dse as `IntDiv().args = []`

mloubout:
Small fix on the new `IntDiv`

- set as `is_Atom` to avoid any evalutaion or `.args` access and print as is in the C code
- Added a test for `ConditionalDimension` used as stencil value instead of index only
- Fix the `assert` in test dimension that were not checking anything.

Should fix #490

### DIFF
--- a/devito/symbolics/extended_sympy.py
+++ b/devito/symbolics/extended_sympy.py
@@ -103,6 +103,7 @@ class IntDiv(sympy.Expr):
     denominator (e.g., ``a*3.args -> (a, 1/3)), which ends up generating
     "weird" C code.
     """
+    is_Atom = True
 
     def __new__(cls, lhs, rhs, params=None):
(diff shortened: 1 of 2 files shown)


FabioLuporini:
GTG (merging it first thing tmr morning). I only have a curiosity: you put is_Atom -- would it work if you created the object as ``obj = sympy.Expr.__new__(cls, evaluate=False)`` instead of ``obj = sympy.Expr.__new__(cls)``. As I said, just a curiosity...

mloubout:
I am not sure for the `evaluate=false`, I think it would still end up in the wrong part of dse as it needs to be caught as a "leave as it is".

FabioLuporini:
OK, merging now then

PR #494, opened 2018-02-20
Aliasing hotfix

mloubout:
This small PR fixes  the aliasing for SparseFunction that was only working on its name and coordinate name.

SparseFunction have a `p%name` dimension that needs to be properly addressed too and is done here.


devito/function.py
```diff
@@ -1010,8 +1010,17 @@ def argument_defaults(self, alias=None):
 
         :param alias: (Optional) name under which to store values.
         """
-        args = super(SparseFunction, self).argument_defaults(alias=alias)
-        args.update(self.coordinates.argument_defaults())
+        key_name = self.name if alias is None else alias.name or self.name
```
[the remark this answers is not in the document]
mloubout: Alias is an object or None, which is why I need this line.

devito/function.py
```diff
@@ -1010,8 +1010,17 @@ def argument_defaults(self, alias=None):
...
+        key_indices = self.indices if alias is None else alias.indices
+        key_coords = self.coordinates.name if alias is None else alias.coordinates.name
+
+        args = ArgumentMap({key_name: self._data_buffer})
```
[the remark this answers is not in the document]
mloubout: `super` is what was part of the problem creating a bunch of wrong arguments as it only takes the name but does not know about name dependent indices.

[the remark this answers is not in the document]
FabioLuporini: not reusing superclass code is a symptom that something is wrong so I'm not fully convinced by this yet

devito/function.py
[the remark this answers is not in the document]
FabioLuporini: > Alias is an object or None, which is why I need this line

so you're saying that this doesn't work:

```
key = alias or self

args = ArgumentMap({key.name: self._data_buffer})
...
```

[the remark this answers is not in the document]
mloubout: if `alias` is None, `alias.name` does not exist.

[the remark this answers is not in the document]
FabioLuporini: I just tried the two new tests with this code:
```
        key = alias or self

        args = ArgumentMap({key.name: self._data_buffer})

        for i, s, o, k in zip(self.indices, self.shape, self.staggered, key.indices):
            args.update(i.argument_defaults(size=s+o, alias=k))

        args.update(self.coordinates.argument_defaults(alias=key.coordinates.name))
        return args
```

and they both seem to pass. So either the tests are not capturing something, or there is a misunderstanding here.

The notation
``key = alias or self``
means that ``key`` gets assigned ``self`` if alias is None/False/0/...

devito/function.py
```diff
@@ -1024,18 +1033,17 @@ def argument_values(self, alias=None, **kwargs):
         # Take a copy of the replacement before super pops it from kwargs
 
         new = kwargs.get(self.name)
-        key = alias or self.name
-
-        values = super(SparseFunction, self).argument_values(alias=key, **kwargs)
+        key = alias or self
 
+        # values = super(SparseFunction, self).argument_values(alias=key, **kwargs)
```
FabioLuporini: leftover

devito/function.py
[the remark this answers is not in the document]
mloubout: Nevermind, that is indeed better, was thinking this wring.

devito/function.py
```diff
@@ -1010,8 +1010,15 @@ def argument_defaults(self, alias=None):
...
         :param alias: (Optional) name under which to store values.
         """
-        args = super(SparseFunction, self).argument_defaults(alias=alias)
-        args.update(self.coordinates.argument_defaults())
+        key = alias or self
+
+        args = ArgumentMap({key.name: self._data_buffer})
+
+        # Collect default dimension arguments from all indices
+        for i, s, o, k in zip(self.indices, self.shape, self.staggered, key.indices):
```
FabioLuporini: I'm again confused here: what happens if you just do:
```
args.update(k.argument_defaults(size=s+o)
```

*If* this worked, then I *think* we could move the fix to the superclass (as I don't understand why is this fix necessary here, but not in the superclass, ie for all tensor functions)

mloubout: Because unlike normal Functions that only have indices derived from a grid object that always names its dimensions the same, Sparse functions indices are specific to each one (`p_%s`%name`) so without dimension aliasing `key.indices`, the argument are not correct.

FabioLuporini: I don't understand yet. I tried the proposed change above, and all of the new tests still pass.

mloubout: `TestArguments.test_override_sparse_data_default_dim` doesn't:

```    
        # Check all argument are present
        for p in self.parameters:
            if p.name not in arguments:
>               raise ValueError("No value found for parameter %s" % p.name)
E               ValueError: No value found for parameter p_src1_e
```
as without dimension name aliasing, all end up in `p_src_2_*`

FabioLuporini: > Because unlike normal Functions that only have indices derived from a grid object that always names its dimensions the same

I disagree with this though. For example, when injecting a source, ``u`` gets indexed indirectly. Some other times, one may use subdimensions to index into a Function; users may also start using extra indices for whatever dumb reason. So, I'm not sure why should this fix stay in `SparseFunction`. Shouldn't it go to `TensorFunction`?

mloubout: This is a good point. This will require some more in depth thinking, as it will depend whether or not the Function is constructed from a Grid or from dimensions.

FabioLuporini: Do you suggest to hold for that? 
What's wrong in moving this into the superclass, even though it's not super bullet proof? it'd be at least better than "hiding" it in the subclass...

mloubout: In the super class it would end up with some weird if/else as Function need `self.grid.indices` as aliases while `Sparse` needs `self.indice` which would need to check the type of input.

FabioLuporini: just add a new private property in the superclass and override it in the subclass. Please, use the following format for the name: ``_arg_<name>``, for example ``_arg_aliases``, because that's what I'm going to use anyway as of #507

devito/function.py
```diff
@@ -306,12 +306,12 @@ def argument_defaults(self, alias=None):
...
         """
-        key = alias or self.name
-        args = ArgumentMap({key: self._data_buffer})
+        key = alias or self
+        args = ArgumentMap({key.name: self._data_buffer})
 
         # Collect default dimension arguments from all indices
-        for i, s, o in zip(self.indices, self.shape, self.staggered):
-            args.update(i.argument_defaults(size=s+o))
+        for i, s, o, k in zip(self.indices, self.shape, self.staggered, key.indices):
```
FabioLuporini: so it works this way? you said it wasn't going to be OK this way...

mloubout: Looks like it does for now

FabioLuporini:
Merged.

PR #502, opened 2018-02-24
Move interpolation to its own module

navjotk:
This PR pulls the specifics of the interpolation method out into its own class called `LinearInterpolator` in `interpolators.py`. 

Also adds a currently experimental `LanczosInterpolator`. This is a WIP not currently passing the adjoint test and needs additional testing anyway.

### DIFF
--- a/devito/function.py
+++ b/devito/function.py
@@ -20,6 +20,7 @@ from devito.parameters import configuration
 from devito.symbolics import indexify, retrieve_indexed
 from devito.types import SymbolicFunction, AbstractCachedSymbol
 from devito.tools import EnrichedTuple
+from devito.interpolators import LinearInterpolator
 
 __all__ = ['Constant', 'Function', 'TimeFunction', 'SparseFunction',
            'SparseTimeFunction']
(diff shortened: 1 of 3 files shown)


devito/function.py
```diff
@@ -807,6 +808,9 @@ def __init__(self, *args, **kwargs):
             # Padding region
             self._padding = tuple((0, 0) for i in range(self.ndim))
 
+            self.interpolator = LinearInterpolator(self.point_symbols,
```
FabioLuporini: I'm trying to progressively switch to private attributes (starting with _) whenever possible

devito/interpolators.py
```diff
@@ -0,0 +1,106 @@
...
+class LanczosInterpolator(GenericInterpolator):
+    def __init__(self, point_symbols, dimensions, a):
+        super(LanczosInterpolator, self).__init__(point_symbols, dimensions, a)
+
+    @property
+    def coefficients(self):
+        x, a = sympy.symbols("x a")
+        f = sympy.sinc(x)*sympy.sinc(x/a)
+        vectors = []
+        for dim in self.point_symbols:
```
tjb900: It looks like self.point_symbols is always (px,py,pz) even for a 2D grid.  This causes this code to produce incorrect output when generating coefficients for 2D grids.

The real problem is probably that len(self.point_symbols) != ndim, though.

mloubout:
>This is a WIP not currently passing the adjoint test and needs additional testing anyway.

The can it wait for PR where it is tested and working.

navjotk:
This approach limits the kind of interpolators that can be written so we've agreed to scrap it.

Issue #503, opened 2018-03-01
detect_flow_direction fails when subdimensions are involved


tjb900:
Hit this when trying to hack up a boundary condition implementation.  It looks like the flow direction detection gets flummoxed by the presence of SubDimensions.

These boundary conditions have a dependency not only on the previous timestep, but on the current time step as well.

i.e., for the left boundary, ``u(t+1,x) = a*u(t,x) + b*u(t,x+1) + c*u(t+1,x+1)``

This implies that the x variable must be traversed in reverse order for this phase of the computation - something that devito already handles well.

However, it appears that it gets thrown when x is a subdimension instead of a full dimension.

For example, in the attached file, `test_flowdir_domain` works, but `test_flowdir_interior` does not.

[test_loop_order.py.txt](https://github.com/opesci/devito/files/1770089/test_loop_order.py.txt)

PR #504, opened 2018-03-01
fix detection of flow direction along subdimensions

tjb900:
Proposed fix for #503 

No idea if this is the proper way to do it, though...

### DIFF
--- a/devito/ir/support/utils.py
+++ b/devito/ir/support/utils.py
@@ -69,9 +69,10 @@ def detect_flow_directions(exprs):
                 continue
             dimensions = [d for d in w.aindices if d is not None]
             for d in dimensions:
-                rootd = d
-                while rootd.is_Sub:
-                    rootd = rootd.parent
+                if d.is_Sub:
+                    rootd = d.parent
+                else:
+                    rootd = d
 
                 try:
                     if w.distance(r, rootd) > 0:
--- a/devito/ir/support/utils.py
+++ b/devito/ir/support/utils.py
@@ -69,11 +69,15 @@ def detect_flow_directions(exprs):
                 continue
             dimensions = [d for d in w.aindices if d is not None]
             for d in dimensions:
+                rootd = d
+                while rootd.is_Sub:
+                    rootd = rootd.parent
+
                 try:
-                    if w.distance(r, d) > 0:
+                    if w.distance(r, rootd) > 0:
                         mapper[d].add(Forward)
                         break
-                    elif w.distance(r, d) < 0:
+                    elif w.distance(r, rootd) < 0:
                         mapper[d].add(Backward)
                         break
                     else:
--- a/devito/ir/support/utils.py
+++ b/devito/ir/support/utils.py
@@ -70,15 +70,15 @@ def detect_flow_directions(exprs):
             dimensions = [d for d in w.aindices if d is not None]
             for d in dimensions:
                 if d.is_Sub:
-                    rootd = d.parent
+                    testdim = d.parent
                 else:
-                    rootd = d
+                    testdim = d
 
                 try:
-                    if w.distance(r, rootd) > 0:
+                    if w.distance(r, testdim) > 0:
                         mapper[d].add(Forward)
                         break
-                    elif w.distance(r, rootd) < 0:
+                    elif w.distance(r, testdim) < 0:
                         mapper[d].add(Backward)
                         break
                     else:
--- a/tests/test_operator.py
+++ b/tests/test_operator.py
@@ -1022,6 +1022,46 @@ class TestLoopScheduler(object):
         assert trees[0][-1].nodes[0].write == u1
         assert trees[0][-1].nodes[1].write == u2
 
+    def test_flowdir_domain(self):
+        #
+        # Stencil uses values at new timestep as well as those at previous ones
+        # This forces an evaluation order onto x.
+        # Weights are:
+        #
+        #        x=0     x=1     x=2     x=3
+        # t=n     2    ---3
+        #         v   /
+        # t=n+1   o--+----4
+        #
+        # Flow dependency should traverse x in the negative direction
+        #
+        #        x=2     x=3     x=4     x=5      x=6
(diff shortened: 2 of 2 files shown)


devito/ir/support/utils.py
```diff
@@ -69,11 +69,15 @@ def detect_flow_directions(exprs):
                 continue
             dimensions = [d for d in w.aindices if d is not None]
             for d in dimensions:
+                rootd = d
+                while rootd.is_Sub:
```
FabioLuporini: it is safe to assume that subdimensions cannot be nested, otherwise we'll need a more general mechanism then this to retrieve the parent

so here you can just filter the dimensions list (before the ``for d in dimensions`` loop) as
```
dimensions = [d.parent if d.is_Sub else d for d in dimensions]
```
or alternatively stick an ``if d.is_Sub`` here

tests/test_operator.py
```diff
@@ -1074,3 +1074,77 @@ def test_explicit_run(self):
         args['a'] = array
         op.cfunction(*list(args.values()))
         assert all(np.allclose(args['a'][i], i) for i in range(time_dim))
+
+
+@skipif_yask
+class TestLoopOrder(object):
```
FabioLuporini: Tests for loop iteration direction should go in ``class TestLoopScheduler``

Tests for correctness when iterating over sub-regions should go in ``class TestRegions``

devito/ir/support/utils.py
tjb900: Random thought: is Access.distance actually the right place to put this check?

tjb900:
Something went wrong with Travis, but I don't think it was my change.  Not sure how to (or if I even have permissions to) poke it to retry.

mloubout:
Yeah Travis is sketchy these days, restarted it.

devito/ir/support/utils.py
```diff
@@ -69,11 +69,16 @@ def detect_flow_directions(exprs):
...
+                else:
+                    rootd = d
+
                 try:
-                    if w.distance(r, d) > 0:
+                    if w.distance(r, rootd) > 0:
                         mapper[d].add(Forward)
                         break
-                    elif w.distance(r, d) < 0:
+                    elif w.distance(r, rootd) < 0:
```
FabioLuporini: wait, don't we need ``rootd`` in place of ``d`` a few lines below as well?

tjb900: Where were you thinking?  I confess I don't have a great handle on what everything in here does, just yet, so my instinct was to keep the changes to a minimum.  For this bug at least, it really just seemed to be the distance() call that was failing.

FabioLuporini: oh, OK, sorry, I should have provided more info. I meant lines 82, 85, 88

nitpicking: for the convention used throughout Devito, can we use ``dim`` rather than ``rootd``?

FabioLuporini: and in fact, now that I see it, line 91 too:
```
for d in dimensions[dimensions.index(dim) + 1:]:
```

FabioLuporini: nope, this won't work because ``dim`` isn't in ``dimensions``. I think we have to it this way:
```
dimensions = [d for d in w.aindices if d is not None]
dimensions = [d.parent if d.is_Sub else d for d in dimensions]
...
```
then you can leave all the rest as in ``master``

tests/test_operator.py
```diff
@@ -1022,6 +1022,46 @@ def test_equations_mixed_timedim_stepdim(self):
         assert trees[0][-1].nodes[0].write == u1
         assert trees[0][-1].nodes[1].write == u2
 
+    def test_flowdir_domain(self):
+        #
```
FabioLuporini: can you turn this into a proper docstring, ie with ``"""..."""``

tests/test_operator.py
```diff
@@ -1052,6 +1092,48 @@ def test_domain_vs_interior(self):
         assert np.all(u.data[1, :, :, -1] == 1)
         assert np.all(u.data[1, 1:3, 1:3, 1:3] == 3)
 
+    def test_flowdir_interior(self):
+        #
```
FabioLuporini: same: docstring

tests/test_operator.py
```diff
@@ -1052,6 +1092,48 @@ def test_domain_vs_interior(self):
...
+        u.data[0, 5, 5] = 1.0
+        op.apply(time_e=2)
+        assert u.data[1, 5, 5] == 2
+        assert u.data[1, 4, 5] == 11
+        assert u.data[1, 3, 5] == 44
+        assert u.data[1, 2, 5] == 4*44
+        assert u.data[1, 1, 5] == 4*4*44
+
+        # This point isn't updated because of the INTERIOR selection
+        # assert u.data[1, 0, 5] == 4*4*4*44
```
FabioLuporini: so you may assert it's still at its original value?

devito/ir/support/utils.py
tjb900: With just that change, it doesn't get the direction right anymore.  The subdimension doesn't get the direction assigned to it - and ends up being caught by this:

```
    # Add in any encountered Dimension
    mapper.update({d: {Any} for d in flatten(i.aindices for i in reads + writes)
                   if d is not None and d not in mapper})
```

tjb900:
I think these latest Travis failures are an update to pytest.

https://github.com/computationalmodelling/nbval/commit/4f2c983b51f6a5fb2a94d5e1344b96cd1872d351#diff-806fb60ab15974e9a2fe3495b18ac7d9

ggorman:
@tjb900 @FabioLuporini - Why was this approved but never merged?
Rebase and merge?

FabioLuporini:
@ggorman this has been stalled by the discussion in #509 : 
https://github.com/opesci/devito/pull/509/files#diff-2e8d97600bfec637499d101d3117a521R483

probably we should attempt the approach that I discussed in there before landing either #504 or #509

FabioLuporini:
Closing as the fix is gonna be shipped by #523

Issue #425, opened 2017-11-21
Inconsistent prescribed iteration limits with stencil offsets


mlange05:
Assume the following setup code:
```
grid = Grid(shape=(11, 11))
x, y = grid.dimensions
a = Function(name='a', grid=grid)
a.data[:] = 1.
A = a.indexed
```

Now, if I run an operator with no stencil offsets, I get what I'd expect:
```
Operator(Eq(a, a + a))(x_s=3, x_e=7)
assert (a.data[:3, :] == 1.).all()
assert (a.data[3:7, :] > 1.).all()
assert (a.data[7:, :] == 1.).all()
```
If I add an operator with a stencil offset in `x` though, I get a positive shift (`+1`) in iteration space of _both_, `x_s` and `x_e`:
```
a.data[:] = 1.
Operator(Eq(a, a + (A[x-1, y] + A[x+1, y]) / 2.))(x_s=3, x_e=7)
print(a.data)
```
yields
```
[[ 1.     1.     1.     1.     1.     1.     1.     1.     1.     1.     1.   ]
 [ 1.     1.     1.     1.     1.     1.     1.     1.     1.     1.     1.   ]
 [ 1.     1.     1.     1.     1.     1.     1.     1.     1.     1.     1.   ]
 [ 1.     1.     1.     1.     1.     1.     1.     1.     1.     1.     1.   ]
 [ 2.     2.     2.     2.     2.     2.     2.     2.     2.     2.     2.   ]
 [ 2.5    2.5    2.5    2.5    2.5    2.5    2.5    2.5    2.5    2.5    2.5  ]
 [ 2.75   2.75   2.75   2.75   2.75   2.75   2.75   2.75   2.75   2.75 2.75 ]
 [ 2.875  2.875  2.875  2.875  2.875  2.875  2.875  2.875  2.875  2.875 2.875]
 [ 1.     1.     1.     1.     1.     1.     1.     1.     1.     1.     1.   ]
 [ 1.     1.     1.     1.     1.     1.     1.     1.     1.     1.     1.   ]
 [ 1.     1.     1.     1.     1.     1.     1.     1.     1.     1.     1.   ]]
```
This final result is shifted in `x`, which seems non-sensical. I would either expect the iteration space to be the exact same as before (`x[3:7]`) or to contract due to the stencil offset, ie. `x[4:6]`, but I see no justification for a shifted iteration space `x[4:8]`. 

[...]

PR #507, opened 2018-03-03
Domain-allocation switch

### LINKED ISSUE #449: Domain-based index addressing
As discussed in various other contexts before, Devito currently enforces an index addressing scheme based on absolute value with respect to the allocated data of a `Function`. Such a (for lack of a better name) _allocation-based addressing scheme_ entails that the user needs to know exactly what the radius of the stencil is that he is going to use and allocate enough data for an `Operator` to work on, as well as needing to manually calculate offsets when supplying explicit dimension indices on `Operator` invocation.

To improve on this we should switch to a _domain-based scheme_, where instead grid indices are addressed relative to the "write-to" part of the data, called the "domain". To illustrate, consider a single dimension with a "domain size" of 100 elements/data points, and a 1-point boundary or stencil radius:
```
 boundary             domain                     boundary
  | -- | ---------------------------------------- | -- |
   read                write                       read

Data indices:
  0    1               51     71                 101  102
User indices
       0               50     70                 100
```

Now, I believe there are three sections of the code that need changing to implement this, as well as several API changes required to make this work neatly:

**1. Data allocation and Python/numpy access** 
Considering that the most common way to define the size of the _padding_ regions (the read-only part of the data at the boundary) is through the `Function`-specific `space/time_order` parameter, I would argue that the `shape` taken by the grid should be interpreted as the "domain shape". That is to say that all functions sharing a `Grid` object agree on the same domain size, but may have different padding sizes.

The user data in `f.data` should most likely be trimmed to only present the "domain part" of the data, although the halo/padding/boundary part of the data also needs to be accessible to users. For this reason, I'm proposing to introduce `f.data_domain` and `f.data_allocated`, where `f.data` defaults to `f.data_domain` (naming, as always debatable).  

**2. Runtime argument derivation** 

Since the _domain-based_ approach indexes relative to the "write" part of the data and loop, `x_start`/`x_end` variables can be expressed/expected with respect to the domain and should thus not include any offsets. However, the associated `x_size` needs to be treated carefully, since it is used for the data cassts in the C kernels. It thus needs to be in terms of "allocation", rather than "domain" and we should discuss what the conventions should be.

**3. Loop bounds in the C kernel** 

This is the trickiest part, as we will need to be able to handle data with different padding/boundary sizes in a single loop. As an example, consider the following boundary sizes: 
`a: 0, b: 1, c: 4` and a stencil `a[x] = a[x] + (b[x-1]+b[x+1])/2 + (c[x-4]+c[+4])/2`

This looks like this in memory:
```
a:  | ------------ |
b:  | - | ------------ | - |
c:  | ---- | ------------ | ---- |
```
My intuition on this is that the main loop counter for `x` in this case should be in terms of the _domain_ ("write region"), so that arguments and loop boundaries all keep their natural meaning. How we "translate" or "normalize" the respective boundaries in the generated code is to be discussed.

FabioLuporini:
Objective
=======
The essence of this huge PR is to switch to a domain-centric system. No matter what discretisation is used,
if the grid domain is ``(X, Y, Z)``, then exactly ``(X, Y, Z)`` grid points will be written to by an operator. 

More details available here: #449 

API change
=========
This PR also switches from a start/end semantics of arguments and loop bounds, to a min/max one (ie, extremes included). From a code generation point of view, this means that *all* loops, except (potentially) the DSE-induced ones, will have the form:
```
for x = x_m; x <= x_M; x++
```
where `x_m` means x_min and `x_M` means x_max.
All tests have been changed accordingly.
I found this API change making it simpler to reason about forward and adjoint operators.

Tests
====
Besides adjusting/tweaking the existing tests to the new API, new tests are added (mostly in ``test_operator.py``.

TODO
=====
* Fixes to checkpointing tests (@navjotk )
* Beta-testing ( @navjotk @mloubout @tjb900 )
* Beta-testing via JUDI (@philippwitte )
* IPython notebooks updates/checks (who does this?)
* With this PR, Devito does many more checks than before, and we not get this warning: ``Data type float64 of runtime value `dt` does not match the Constant data type <class 'numpy.float32'>``.  

fixes #449 #425


ggorman:
@FabioLuporini, and @navjotk, @mloubout, @tjb900, @philippwitte - any progress on the todo list above? What is still outstanding to land this?

mloubout:
>Beta-testing via JUDI (@philippwitte )

Produces Nan currently with this branch, I need to find out from where

FabioLuporini:
@ggorman some PRs have made master over the last couple of weeks, so this branch needs to be rebased and conflicts be fixed. Will do on Monday. 

Also, the checkpointing tests are still failing; from the discussion in #development on Slack , I get that @navjotk should be able to fix them as soon as we provide an appropriate shortcut for ``op.apply`` to iterate for *exactly* N iteartions. For example, doing:
```
op.apply(time_n=N)
```
one expects to (citing from Slack):
> with `time_n=N`, the start would be the implicitly derived `time_m`, then iterating up to `time_m+N-1` (included)

the semantics of ``*_m`` and ``*_M`` (* being any arbitrary dimension) are min/max .

I also need someone (gonna update the TODO now) to look into all of the IPython notebooks

examples/cfd/01_convection.ipynb
```diff
@@ -348,19 +351,19 @@
...
      "text": [
-      "CustomCompiler: compiled /var/folders/xr/858m074j16nd3kcz4dd58_f00000gn/T/devito-5i1151cw/99828d970bd77277b20e4c775f82fd0cecee0f39.c [0.13 s]\n",
+      "CustomCompiler: compiled /var/folders/qr/4hc1nw7d3hj55jbd1z7kkrvh0000gn/T/devito-89_5b_7l/1957c22c54d3d7bfc6d9c309adc532acf2d80e64.c [0.66 s]\n",
       "=========================================================================================\n",
-      "Section section_1<100,80> with OI=0.00 computed in 0.000 s [0.00 GFlops/s]\n",
-      "Section section_2<100,80> with OI=0.00 computed in 0.000 s [0.00 GFlops/s]\n",
-      "Section main<100,80,80> with OI=1.45 computed in 0.001 s [7.25 GFlops/s, 0.60 GPts/s]\n",
+      "Section section_1<101,81> with OI=0.00 computed in 0.000 s [0.00 GFlops/s]\n",
+      "Section section_2<101,81> with OI=0.00 computed in 0.000 s [0.00 GFlops/s]\n",
+      "Section main<101,81,81> with OI=1.50 computed in 0.001 s [11.11 GFlops/s, 0.93 GPts/s]\n",
```
navjotk: The number of iterations has changed here. So has the OI. Neither of those should be expected.

examples/cfd/02_convection_nonlinear.ipynb
```diff
@@ -263,27 +255,25 @@
...
      "text": [
-      "CustomCompiler: compiled /tmp/devito-pshw6um6/017c17815b4ef24496c80e316d98b03ae6fc7408.c [1.15 s]\n",
+      "CustomCompiler: compiled /var/folders/qr/4hc1nw7d3hj55jbd1z7kkrvh0000gn/T/devito-8741qqqk/2f1a291c50a8ee234a1861ccb3c37b1903e971cc.c [0.48 s]\n",
       "=========================================================================================\n",
-      "Section section_1<80,101> with OI=0.00 computed in 0.000 s [0.00 GFlops/s]\n",
-      "Section section_2<80,101> with OI=0.00 computed in 0.000 s [0.00 GFlops/s]\n",
-      "Section main<80,101,101> with OI=1.73 computed in 0.002 s [11.05 GFlops/s, 0.39 GPts/s]\n",
+      "Section section_1<81,101> with OI=0.00 computed in 0.000 s [0.00 GFlops/s]\n",
+      "Section section_2<81,101> with OI=0.00 computed in 0.000 s [0.00 GFlops/s]\n",
+      "Section main<81,99,99> with OI=1.75 computed in 0.003 s [6.53 GFlops/s, 0.23 GPts/s]\n",
```
navjotk: Again. I don't understand why the number of iterations or the OI should change.

examples/cfd/04_burgers.ipynb
```diff
@@ -201,53 +195,25 @@
...
-      "     [Total elapsed: 0.16 s]\n",
-      "CustomCompiler: compiled /tmp/devito-1000/1645e03592558a800a6ea8efe72867472be621c5.c [0.25 s]\n",
+      "CustomCompiler: compiled /var/folders/qr/4hc1nw7d3hj55jbd1z7kkrvh0000gn/T/devito-y2kr1ngp/9beba14a9cc90a026c15251528633e01066ebb49.c [0.64 s]\n",
       "=========================================================================================\n",
-      "Section section_1<120,39> with OI=0.00 computed in 0.000 s [Perf: 0.00 GFlops/s]\n",
-      "Section section_2<120,39> with OI=0.00 computed in 0.000 s [Perf: 0.00 GFlops/s]\n",
-      "Section main<120,39,39> with OI=3.81 computed in 0.002 s [Perf: 6.60 GFlops/s]\n",
+      "Section section_1<121,41> with OI=0.00 computed in 0.000 s [0.00 GFlops/s]\n",
+      "Section section_2<121,41> with OI=0.00 computed in 0.000 s [0.00 GFlops/s]\n",
+      "Section main<121,41,41> with OI=4.25 computed in 0.005 s [2.96 GFlops/s, 0.04 GPts/s]\n",
```
navjotk: Another place where iteration spaces and OI have changed.

examples/seismic/acoustic/gradient_example.py
```diff
@@ -54,8 +56,9 @@ def _setup_model_and_acquisition(self, shape, spacing, nbpml, tn):
         self.grad = Function(name="grad", grid=model.grid)
 
     def initial_estimate(self):
-        m0 = smooth10(self.model.m.data, self.model.shape_domain)
-        dm = np.float32(self.model.m.data - m0)
+        m0 = Function(name='m0', grid=self.model.m.grid, space_order=self.space_order)
+        m0.data[:] = smooth10(self.model.m.data, self.model.shape_domain)
```
navjotk: Can use an initializer here.

examples/seismic/acoustic/operators.py
```diff
@@ -189,7 +186,7 @@ def BornOperator(model, source, receiver, space_order=4,
                      time_order=2, space_order=space_order)
     U = TimeFunction(name="U", grid=model.grid, save=None,
                      time_order=2, space_order=space_order)
-    dm = Function(name="dm", grid=model.grid)
+    dm = Function(name="dm", grid=model.grid, space_order=0)
```
navjotk: What's the default space order?

examples/seismic/tutorials/01_modelling.ipynb
```diff
@@ -418,26 +388,24 @@
...
      "text": [
-      "CustomCompiler: compiled /tmp/devito-1000/e1f3caddd102ca43788ca2d36e3b857d86fe4bdd.c [0.31 s]\n",
+      "CustomCompiler: compiled /var/folders/qr/4hc1nw7d3hj55jbd1z7kkrvh0000gn/T/devito-088e8vk1/72a8ff90096e050e38867c2ec004eaf34979ddc8.c [1.41 s]\n",
       "=========================================================================================\n",
-      "Section section_1<594,1> with OI=0.73 computed in 0.000 s [Perf: 0.34 GFlops/s]\n",
-      "Section section_2<594,101> with OI=1.37 computed in 0.001 s [Perf: 1.45 GFlops/s]\n",
-      "Section main<594,119,119> with OI=1.81 computed in 0.022 s [Perf: 11.62 GFlops/s]\n",
+      "Section section_1<595,1> with OI=0.84 computed in 0.000 s [0.49 GFlops/s]\n",
+      "Section section_2<595,101> with OI=1.62 computed in 0.001 s [2.54 GFlops/s]\n",
+      "Section main<595,121,121> with OI=1.88 computed in 0.016 s [16.04 GFlops/s, 0.53 GPts/s]\n",
```
navjotk: Same problem as before

examples/seismic/tutorials/03_fwi.ipynb
```diff
@@ -519,14 +519,14 @@
...
-      "Objective value is 2008.388293 at iteration 7\n",
-      "Objective value is 800.837208 at iteration 8\n"
+      "Objective value is 56771.423590 at iteration 1\n",
+      "Objective value is 39699.872335 at iteration 2\n",
+      "Objective value is 26133.256710 at iteration 3\n",
+      "Objective value is 16177.508945 at iteration 4\n",
+      "Objective value is 9276.513601 at iteration 5\n",
+      "Objective value is 4787.393206 at iteration 6\n",
+      "Objective value is 2017.148281 at iteration 7\n",
+      "Objective value is 805.578966 at iteration 8\n"
```
navjotk: Curiosity @mloubout : What is the source of the randomness in these values?

tests/test_dse.py
```diff
@@ -101,28 +101,6 @@ def tti_nodse():
     return v, rec
 
 
-@skipif_yask
```
navjotk: Has this test been replaced by something else?

tests/test_gradient.py
```diff
@@ -37,8 +37,10 @@ def test_gradientFWI(shape, kernel, space_order):
     wave = setup(shape=shape, spacing=spacing, dtype=np.float64,
                  kernel=kernel, space_order=space_order,
                  nbpml=10+space_order/2)
-    m0 = smooth10(wave.model.m.data, wave.model.shape_domain)
-    dm = np.float32(wave.model.m.data - m0)
+
+    m0 = Function(name='m0', grid=wave.model.m.grid, space_order=space_order)
+    m0.data[:] = smooth10(wave.model.m.data, wave.model.m.shape_domain)
+    dm = np.float32(wave.model.m.data - m0.data)
```
navjotk: If `dtype` is configurable, shouldn't you be using the `dtype` variable?

tests/test_yask.py
```diff
@@ -442,8 +449,8 @@ def test_acoustic_w_src_w_rec(self, model, eqn, m, damp, u, src, rec):
 
         # The expected norms have been computed "by hand" looking at the output
         # of test_adjointA's forward operator w/o using the YASK backend.
-        exp_u = 152.76
-        exp_rec = 212.00
+        exp_u = 156.47
+        exp_rec = 211.81
```
navjotk: Why have the norms changed?

devito/function.py
```diff
@@ -330,50 +284,67 @@ def argument_values(self, alias=None, **kwargs):
...
                 values[key.name] = new._data_buffer
-                values.update(new.argument_defaults(alias=key).reduce_all())
+                values.update(new._arg_defaults(alias=key).reduce_all())
             else:
                 # We've been provided a pure-data replacement (array)
                 values[key.name] = new
                 # Add value overrides for all associated dimensions
                 for i, s, o in zip(self.indices, new.shape, self.staggered):
-                    values.update(i.argument_defaults(size=s+o))
+                    values.update(i._arg_defaults(size=s+o-sum(self._offset_domain[i])))
```
mloubout: Why is `self._offset_domain[i]` necessary here but not in `_args_defaults`?

devito/function.py
```diff
@@ -330,50 +284,67 @@ def argument_values(self, alias=None, **kwargs):
...
+                        ``space_order`` points are available on both sides of
+                        a generic point of interest, including those on the grid
+                        border. Sometimes, fewer points may be necessary; in
+                        other cases, depending on the PDE being approximated,
+                        more points may be necessary. In such cases, one
+                        can pass a 3-tuple ``(o, lp, rp)`` instead of a single
+                        integer representing the discretization order. Here,
+                        ``o`` is the discretization order, while ``lp`` and ``rp``
+                        indicate how many points are expected on left (``lp``)
+                        and right (``rp``) of a point of interest.
```
mloubout: - I'm not sure I like that as part of `space_order'.
- `space_order` as a tuple would makes sense for a different order in each of the dimensions (e.g 2nd in x and 4th in z).

devito/operator.py
```diff
@@ -114,63 +119,58 @@ def __init__(self, expressions, **kwargs):
...
+        args = args.reduce_all()
 
-        # Second, derive all remaining default values from parameters
-        arguments.update(self._argument_defaults(arguments))
+        # Handle dimensions (first adjust data-carriers-induced defaults, then overrides)
+        for p in self.dimensions:
+            args.update(p._arg_infers(args, self._dspace[p], **kwargs))
+        for p in self.dimensions:
+            args.update(p._arg_values(**kwargs))
+
```
mloubout: Would switch, same as `input`, first fill `kwargs` then complete with missing remaining ones.

devito/types.py
```diff
@@ -397,56 +418,86 @@ def size(self):
...
-    :param shape: The shape of the tensor
-    :param dimensions: The symbolic dimensions of the tensor.
+    @property
+    def _mask_with_halo(self):
+        """A mask to access the domain+halo region of the allocated data."""
+        return tuple(slice(i, -j) if j != 0 else slice(i, None)
+                     for i, j in self._offset_halo)
+
+
+class Array(AbstractCachedFunction):
```
mloubout: Not a big fan of the name, would prefer something like `FunctionValue` o

examples/checkpointing/checkpoint.py
```diff
@@ -13,22 +13,28 @@ class CheckpointOperator(Operator):
...
     """
-    t_arg_names = {'t_start': 'time_s', 't_end': 'time_e'}
+    t_arg_names = {'t_start': 'time_m', 't_end': 'time_M'}
 
     def __init__(self, op, **kwargs):
         self.op = op
-        self.time_order = kwargs.pop('time_order', 1)
         self.args = kwargs
+        t = {self.t_arg_names['t_end']: 0}
+        op_default_args = self.op.arguments(**t)
```
mloubout: Doesn't argument allocate all defaults?, that seems a bit overkill to get just one value.

examples/checkpointing/checkpoint.py
```diff
@@ -13,22 +13,28 @@ class CheckpointOperator(Operator):
...
+        self.start_offset = op_default_args[self.t_arg_names['t_start']]
```
mloubout: So this contains the time order now?

examples/checkpointing/checkpoint.py
```diff
@@ -13,22 +13,28 @@ class CheckpointOperator(Operator):
                     automatically include these cached arguments in the call to the
                     underlying devito.Operator.apply().
     """
-    t_arg_names = {'t_start': 'time_s', 't_end': 'time_e'}
+    t_arg_names = {'t_start': 'time_m', 't_end': 'time_M'}
```
mloubout: Docstring to `init`/`prepare`

examples/seismic/model.py
```diff
@@ -37,6 +37,8 @@ def demo_model(preset, **kwargs):
                     filepath. Requires the ``opesci/data`` repository
                     to be available on your machine.
     """
+    space_order = kwargs.pop('space_order', 1)
```
mloubout: Would default to 2,  same as in `wavesolver`, I

examples/seismic/model.py
```diff
@@ -263,27 +260,40 @@ def demo_model(preset, **kwargs):
...
-        if ndim == 2:
-            damp[i, :] += val/spacing[0]
-            damp[-(i + 1), :] += val/spacing[0]
-            damp[:, i] += val/spacing[1]
-            damp[:, -(i + 1)] += val/spacing[1]
+        if damp.ndim == 2:
+            damp.data_with_halo[i, :] += val/spacing[0]
+            damp.data_with_halo[-(i + 1), :] += val/spacing[0]
+            damp.data_with_halo[:, i] += val/spacing[1]
+            damp.data_with_halo[:, -(i + 1)] += val/spacing[1]
```
mloubout: Can't we assign directly to `data` that has already the offset?

examples/seismic/model.py
```diff
@@ -263,27 +260,40 @@ def demo_model(preset, **kwargs):
...
-            damp[:, :, -(i + 1)] += val/spacing[2]
+            damp.data_with_halo[i, :, :] += val/spacing[0]
+            damp.data_with_halo[-(i + 1), :, :] += val/spacing[0]
+            damp.data_with_halo[:, i, :] += val/spacing[1]
+            damp.data_with_halo[:, -(i + 1), :] += val/spacing[1]
+            damp.data_with_halo[:, :, i] += val/spacing[2]
+            damp.data_with_halo[:, :, -(i + 1)] += val/spacing[2]
+
+
+def initialize_function(function, data, nbpml):
```
mloubout: pad_function.

FabioLuporini:
@navjotk  @ggorman @mloubout 

> We also need verification that:
> a) the profiler output is semantically the same as before
> b) Compilation performance is not majorly affected
> c) Runtime performance is same as before

for points b) and c), a fairly detailed comparison is available here : https://gist.github.com/FabioLuporini/6da1522e487a830b1c3da28baedc0df2

for point a), I ran some checks and they look good to me. This is also something I've kept an eye throughout the entire development of this branch. However, in the TODO at the top of this PR, I asked for some beta-testing, and this is one of the things I expect you guys to do. Nothing crazy complicated of course, just some random checks to confirm that what I've seen is correct. 

I'll reply to each of your points, but, in short, if you see different iteration spaces between here and ``master`` in the profile outputs, it's probably because now the test/example runs on a slightly bigger grid (we don't have offsets in the loop bounds anymore)

devito/dimension.py
```diff
@@ -313,39 +382,48 @@ def symbolic_end(self):
...
 
-    def argument_defaults(self, size=None, alias=None):
+    def _arg_defaults(self, start=None, size=None, **kwargs):
         """
         Returns a map of default argument values defined by this symbol.
 
-        :param size: Optional, known size as provided by data-carrying symbols
+        :param start: Optional, known starting point as provided by
+                      data-carrying symbols.
+        :param size: Optional, known size as provided by data-carrying symbols.
```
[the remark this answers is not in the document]
FabioLuporini: No. In fact, ``t_M`` it's not supported even in current master. I actually wanted to drop this ``t_m/t_M`` thing, to force users use ``time_m/time_M``, as ``t`` is a derived dimension and it makes no sense to specify a min/max point

I haven't dropped it yet because I'd have had to go again through the entire test suite and make changes accordingly

devito/dimension.py
```diff
@@ -95,36 +101,89 @@ def base(self):
...
+                      data-carrying symbols.
+        :param size: (Optional) known size as provided by data-carrying symbols.
+        :param alias: (Optional) name under which to store values.
         """
         dim = alias or self
+        return {dim.min_name: start or 0, dim.max_name: size, dim.size_name: size}
+
+    def _arg_infers(self, args, interval, **kwargs):
+        """
+        Returns a map of "better" default argument values, reading this symbol's
```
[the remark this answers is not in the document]
FabioLuporini: :+1:

devito/dimension.py
```diff
@@ -95,36 +101,89 @@ def base(self):
...
+        if interval.is_Defined and args[self.min_name] + interval.lower < 0:
+            raise InvalidArgument("OOB detected due to %s=%d" % (self.min_name,
+                                                                 args[self.min_name]))
+
+        if self.max_name not in args:
+            raise InvalidArgument("No runtime value for %s" % self.max_name)
+        if interval.is_Defined and args[self.max_name] + interval.upper >= size:
+            raise InvalidArgument("OOB detected due to %s=%d" % (self.max_name,
+                                                                 args[self.max_name]))
+
```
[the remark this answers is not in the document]
FabioLuporini: :+1:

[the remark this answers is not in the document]
FabioLuporini: However, if I do that, some of the checkpointing tests don't like it:

[gw12] FAILED test_checkpointing.py::test_forward_with_breaks[shape0-OT2-4] 
[gw1] FAILED test_checkpointing.py::test_checkpointed_gradient_test[shape0-OT2-4] 
[gw13] FAILED test_checkpointing.py::test_forward_with_breaks[shape1-OT2-4] 
[gw2] FAILED test_checkpointing.py::test_checkpointed_gradient_test[shape1-OT2-4] 

I simply added these two lines:

```
if args[self.max_name] < args[self.min_name]:
    raise InvalidArgument("Illegal max=%s < min=%s" % (args[self.max_name], args[self.min_name]))
```

devito/dse/aliases.py
```diff
@@ -239,17 +239,17 @@ class Alias(object):
     is tracked.
     """
 
-    def __init__(self, alias, dimensions, aliased=None, distances=None,
+    def __init__(self, alias, indices, aliased=None, distances=None,
                  ghost_offsets=None):
         self.alias = alias
-        self.dimensions = dimensions
+        self.dimensions = tuple(i.parent if i.is_Derived else i for i in indices)
```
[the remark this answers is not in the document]
FabioLuporini: yes, I screwed names; in fact, these should be called ``dimensions``, while sometimes (in this file) I used ``indices`` instead of ``dimensions``; fixing.

devito/dse/aliases.py
```diff
@@ -260,10 +260,22 @@ def anti_stencil(self):
...
         return handle
 
+    @property
+    def distance_map(self):
+        return [tuple(zip(self.dimensions, i)) for i in self.distances]
+
+    @property
+    def diameter(self):
+        return OrderedDict((d, (min(i), max(i)))
+                           for d, i in zip(self.dimensions, zip(*self.distances)))
```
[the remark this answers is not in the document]
FabioLuporini: :+1:

devito/dse/aliases.py
```diff
@@ -260,10 +260,22 @@ def anti_stencil(self):
...
+
+    @property
+    def relaxed_diameter(self):
+        return OrderedDict((k, (min(v), max(v))) for k, v in self.anti_stencil.items())
```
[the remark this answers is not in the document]
FabioLuporini: :+1:

devito/dse/aliases.py
```diff
@@ -260,10 +260,22 @@ def anti_stencil(self):
...
+
     @property
     def with_distance(self):
-        return [(i, OrderedDict(zip(self.dimensions, j)))
-                for i, j in zip(self.aliased, self.distances)]
+        return tuple(zip(self.aliased, self.distance_map))
```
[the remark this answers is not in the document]
FabioLuporini: :+1:

devito/function.py
```diff
@@ -113,72 +128,26 @@ def _allocate_memory(func):
...
-                if self._first_touch:
+                debug("Allocating memory for %s%s" % (self.name, self.shape_allocated))
+                self._data = Data(self.shape_allocated, self.indices, self.dtype)
+                if self.initializer is not None:
+                    try:
+                        self.initializer(self._data)
+                    except ValueError:
+                        # Perhaps user only wants to initialise the physical domain
+                        self.initializer(self._data[self._mask_domain])
+                elif self._first_touch:
```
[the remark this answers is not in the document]
FabioLuporini: first touch is useless if data has been touched by a process already

devito/function.py
```diff
@@ -113,72 +128,26 @@ def _allocate_memory(func):
...
-        """
-        left = tuple(zip(*self._padding))[0]
-        right = tuple(zip(*self._padding))[1]
-
-        Extent = namedtuple('Extent', 'left right')
-        extents = tuple(Extent(i, j) for i, j in self._padding)
-
-        return EnrichedTuple(*extents, left=left, right=right)
+    def _data_buffer(self):
+        """Reference to the actual data. This is *not* a view of the data.
```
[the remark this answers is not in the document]
FabioLuporini: It's explained after the full stop (btw, this docstring is the same as in master)

devito/function.py
```diff
@@ -330,50 +284,67 @@ def argument_values(self, alias=None, **kwargs):
...
     :param grid: :class:`Grid` object from which to infer the data shape
                  and :class:`Dimension` indices.
+    :param space_order: Discretisation order for space derivatives. By default,
+                        ``space_order`` points are available on both sides of
+                        a generic point of interest, including those on the grid
+                        border. Sometimes, fewer points may be necessary; in
+                        other cases, depending on the PDE being approximated,
+                        more points may be necessary. In such cases, one
+                        can pass a 3-tuple ``(o, lp, rp)`` instead of a single
+                        integer representing the discretization order. Here,
```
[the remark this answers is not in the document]
FabioLuporini: same as before? but again, this docstring is the same as in master, I just had to move it

devito/ir/clusters/cluster.py
```diff
@@ -18,14 +19,17 @@ class PartialCluster(object):
...
+    :param ispace: An object of type :class:`DataSpace`, which represents the
+                   data space (i.e., data items accessed) of the cluster.
     :param atomics: (Optional) non-sharable :class:`Dimension`s in ``ispace``.
     :param guards: (Optional) iterable of conditions, provided as SymPy expressions,
                    under which ``exprs`` are evaluated.
     """
 
-    def __init__(self, exprs, ispace, atomics=None, guards=None):
-        self._exprs = list(ClusterizedEq(i, ispace) for i in exprs)
+    def __init__(self, exprs, ispace, dspace, atomics=None, guards=None):
```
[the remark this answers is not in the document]
FabioLuporini: oh, it's a typo, I had spelled it ``ispace``. Fixing

devito/ir/equations/equation.py
```diff
@@ -23,49 +24,70 @@ def is_Scalar(self):
...
 
     def __new__(cls, *args, **kwargs):
-        # Parse input
         if len(args) == 1:
-            input_expr = args[0]
-            assert type(input_expr) != LoweredEq
-            assert isinstance(input_expr, Eq)
+            # origin: LoweredEq(expr)
+            expr = input_expr = args[0]
+            assert type(expr) != LoweredEq
```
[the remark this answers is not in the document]
FabioLuporini: I think it's just leftover from on-going development. Switching to ``isinstance`` and combining with the next line

devito/ir/equations/equation.py
```diff
@@ -23,49 +24,70 @@ def is_Scalar(self):
...
             expr.is_Increment = stamp.is_Increment
-            expr.ispace = stamp.ispace
+            expr._ispace, expr._dspace = stamp.ispace, stamp.dspace
+            expr.reads, expr.writes = stamp.reads, stamp.writes
+            return expr
+        elif len(args) == 5:
+            # origin: LoweredEq(expr, ispace, space)
+            input_expr, ispace, dspace, reads, writes = args
+            assert isinstance(ispace, IterationSpace)
+            assert isinstance(dspace, DataSpace)
```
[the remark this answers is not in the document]
FabioLuporini: :+1:

devito/ir/equations/equation.py
```diff
@@ -102,33 +136,44 @@ def func(self, *args):
...
-                expr.ispace = maybe_ispace
-            else:
-                expr = Eq.__new__(cls, *args, evaluate=False)
-                expr.ispace = kwargs['ispace']
-                expr.is_Increment = kwargs.get('is_Increment', False)
+        if len(args) == 3:
+            # origin: ClusterizedEq(expr, ispace, dspace)
+            input_expr, ispace, dspace = args
+            assert isinstance(ispace, IterationSpace)
+            assert isinstance(dspace, DataSpace)
```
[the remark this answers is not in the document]
FabioLuporini: :+1:

devito/ir/iet/nodes.py
```diff
@@ -496,8 +503,8 @@ class Callable(Node):
     :param body: A :class:`Node` or an iterable of :class:`Node` objects representing
                  the body of the function.
     :param retval: The type of the value returned by the function.
```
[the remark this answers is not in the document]
FabioLuporini: :+1:

devito/ir/support/space.py
```diff
@@ -262,11 +294,102 @@ def __hash__(self):
...
+    def intervals(self):
+        return self._intervals
+
+    @property
+    def size(self):
+        return len(self.intervals)
+
+    @property
+    def empty(self):
+        return self.size == 0
```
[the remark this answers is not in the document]
FabioLuporini: correct

devito/operator.py
```diff
@@ -69,8 +69,12 @@ def __init__(self, expressions, **kwargs):
         # References to local or external routines
         self.func_table = OrderedDict()
 
-        # Expression lowering and analysis
-        expressions = [LoweredEq(e, subs=subs) for e in expressions]
+        # Expression lowering: indexification, substitution rules, enrichment
+        expressions = [indexify(i) for i in expressions]
+        expressions = [i.xreplace(subs) for i in expressions]
+        expressions = self._specialize_exprs(expressions)
```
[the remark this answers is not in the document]
FabioLuporini: switched enrichment to specialization

devito/operator.py
```diff
@@ -88,17 +93,17 @@ def __init__(self, expressions, **kwargs):
...
         dle_state = transform(nodes, *set_dle_mode(dle))
 
         # Update the Operator state based on the DLE
-        self.dle_arguments = dle_state.arguments
+        self.dle_args = dle_state.arguments
         self.dle_flags = dle_state.flags
         self.func_table.update(OrderedDict([(i.name, MetaCall(i, True))
                                             for i in dle_state.elemental_functions]))
-        self.dimensions.extend([i.argument for i in self.dle_arguments
+        self.dimensions.extend([i.argument for i in self.dle_args
```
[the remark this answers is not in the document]
FabioLuporini: maybe, but not part of this PR (same logic as in ``master``)

devito/operator.py
```diff
@@ -114,63 +119,61 @@ def __init__(self, expressions, **kwargs):
...
+        # Sanity check
+        for p in self.input:
+            p._arg_check(args, self._dspace[p])
 
         # Derive additional values for DLE arguments
         # TODO: This is not pretty, but it works for now. Ideally, the
         # DLE arguments would be massaged into the IET so as to comply
         # with the rest of the argument derivation procedure.
-        for arg in self.dle_arguments:
+        for arg in self.dle_args:
```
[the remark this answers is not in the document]
FabioLuporini: hopefully, but not part of this PR (same logic as in ``master``)

I hate this thing of the DLE arguments treated specially. Someone should fix it at some point

devito/dimension.py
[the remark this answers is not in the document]
navjotk: Not what I'm talking about. If all `TimeFunction`s in an `Operator` are with `save=False`, the `Operator` does not have a default value for `time_M` or `t_M`.

examples/cfd/01_convection.ipynb
FabioLuporini: in fact, they should (read my general comment below). We haven't changed ``shape`` in ``Grid(shape=...)``, so no offsets in the loop bounds means we get exactly 81 here, no matter what the space order is.

devito/function.py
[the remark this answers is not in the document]
navjotk: So it should be done first.

examples/cfd/02_convection_nonlinear.ipynb
FabioLuporini: same as before

examples/cfd/04_burgers.ipynb
FabioLuporini: same as before

examples/seismic/acoustic/gradient_example.py
FabioLuporini: just replicated ``master`` behaviour

devito/function.py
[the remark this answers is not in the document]
navjotk: The sentence after the fullstop explains what it is *not*. Doesn't explain what it is.

examples/seismic/tutorials/01_modelling.ipynb
FabioLuporini: same as before

devito/ir/support/space.py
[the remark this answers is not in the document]
navjotk: Explain that in a docstring.

tests/test_dse.py
FabioLuporini: That test is simply unmaintainable with the changes shipped by this branch. However, it tests a subsets of the things already tested by the other dse/tti tests, so we're not decreasing coverage.

tests/test_gradient.py
FabioLuporini: maybe, just replicated ``master``. What do you suggest I change here, and into what? don't know why the write of this test explicitly set ``dm`` to ``np.float32``. I didn't want to alter this, just to be safe...

examples/checkpointing/checkpoint.py
navjotk: It's not time order because I see some order 2 kernels starting from 1.

examples/checkpointing/checkpoint.py
navjotk: What do you mean?

tests/test_yask.py
FabioLuporini: running on a grid that is slightly larger than before (same story)

devito/function.py
FabioLuporini: agreed, but this is exactly what we have in ``master``. Let's change it another day (volunteers welcome)

devito/operator.py
FabioLuporini: why? why should I bother fetching defaults if I'm not gonna use them?

FabioLuporini: OK, hang on, maybe I didn't understand the comment. ``infers`` must go before ``_arg_values``, because the way I implemented the logic here is that ``infers`` works on defautls

devito/types.py
FabioLuporini: u mean ``Array``? it's been in ``master`` for ages. Just changed its place

examples/checkpointing/checkpoint.py
FabioLuporini: @navjotk

examples/checkpointing/checkpoint.py
FabioLuporini: @navjotk

examples/seismic/model.py
FabioLuporini: maybe? I'm open to contributions here... I just tried to emulate what's in ``master``.

examples/seismic/model.py
FabioLuporini: it's actually doing more than just padding; it's really initializing the Function data. I can switch it if you really prefer, though. I don't have strong opinions in merit.

examples/checkpointing/checkpoint.py
FabioLuporini: Agreed, so eventually @navjotk opted for this change (we discussed it on Slack in #development) but probably we can do better. Open to contributions

examples/seismic/model.py
FabioLuporini: why?

navjotk:
I ask for point a) specifically because the OI has changed in this branch for the same kernel. 
About the performance comparison, the conclusion is that compilation is 10% slower on this branch?

devito/dimension.py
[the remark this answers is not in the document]
navjotk: This check should be kept though. The Revolve wrapper needs to be changed so it doesn't send in invalid arguments.

examples/cfd/01_convection.ipynb
navjotk: What about OI though?

tests/test_gradient.py
navjotk: There is a variable in scope here, called `dtype`, that contains the `dtype`.

devito/dimension.py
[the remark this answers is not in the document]
FabioLuporini: I pasted it so that you can write a patch, add it, commit and push (ie, to avoid getting a red Travis)

devito/dimension.py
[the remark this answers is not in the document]
FabioLuporini: which is ``master``'s current behavior, correct?

devito/function.py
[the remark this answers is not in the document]
FabioLuporini: so we touch data twice?

examples/checkpointing/checkpoint.py
navjotk: So I need a reliable way of telling an `Operator`, "run your 1st timestep" or any nth timestep like that. Since time indices are now completely unpredictable (some operators run from 0 to `n-1`, some from 1 to `n` etc.), I was looking for a way to query this behaviour from the `Operator`. This is the way I was suggested for doing this. I do agree it is overkill and I raised this point on Slack at the time.

devito/function.py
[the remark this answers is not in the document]
FabioLuporini: improving, but it's the same as in master

examples/checkpointing/checkpoint.py
navjotk: ???

examples/cfd/01_convection.ipynb
FabioLuporini: +points => +flops, +accesses => different numerator, different denominator => different OI

FabioLuporini: (according to the way the OI is currently computed, see ``profiling.py``)

tests/test_gradient.py
FabioLuporini: > There is a variable in scope here

i cannot see it

examples/checkpointing/checkpoint.py
FabioLuporini: tagged you before I saw your comment (I hadn't refreshed the window) -- ignore it

examples/checkpointing/checkpoint.py
FabioLuporini: again, ignore my tag

FabioLuporini:
> About the performance comparison, the conclusion is that compilation is 10% slower on this branch?

Same runtime as before, and yes, 10% slower Operator compilation time

examples/seismic/model.py
mloubout: Because the default order for the seismic stencils is `2`,  o should be the model for consistency.

devito/operator.py
mloubout: I mean

```
for p in self.dimensions:
            args.update(p._arg_values(**kwargs))
for p in self.dimensions and not in args:
            args.update(p._arg_infers(args, self._dspace[p], **kwargs))
```

FabioLuporini: No; dimension overrides are fixed values, while inferred ones are meant to be derived from defaults; this made me realize I should probably find a more sensible name for this method as, I agree, it might sound confusing

FabioLuporini: I cannot think of a better name. Any ideas?

mloubout: I'm not sure I understand what you mean.
- overrides => runtime value
- infered => default

It's just number so it doesn't really matter, just was wandering if order in which is done can be consotent with the order for Function
- inputs, then default for the rest

examples/seismic/model.py
FabioLuporini: oooh we're in ``seismic/model.py`` here, right...

devito/operator.py
FabioLuporini: right, and no there's a logic behind it, which is why I said the name "infer" may be misleading. With ``infer`` you add/subtract points to avoid OOB accesses; but you don't want to add/subtract points to overrides. An override is fixed, it's given by the user and that's it. So we run ``infer``, which adds/subtracts points to the values generated by processing ``Function``s (``time_m, time_M, x_m, ...``) and then, if we have overrides, we simply replace them with such overrides

devito/function.py
FabioLuporini: ah, good question ! 
so this ``else`` branch handles the case in which a numpy array is passed in. When we generate the dimension values, however, we should do that w.r.t. the actual domain, not w.r.t. the domain+halo+padding; hence, with ``-sum(...)`` we cut away the region between the last point of the domain and the end of the whole array

FabioLuporini: IOW, in ``_args_default`` you don't need it because for ``self`` you know its shape (which, I recall, is generally smaller than ``shape_with_halo``, which in turn is generally smaller than ``shape_allocated`` -- the latter being the actual shape of the whole array)

devito/dimension.py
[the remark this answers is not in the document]
navjotk: Why go to such lengths to avoid a red travis? You make the change - and if it causes a red travis we change it again to make it green again.

devito/dimension.py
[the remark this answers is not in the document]
navjotk: Do you understand that I am asking for one more line in some docstring somewhere?

devito/function.py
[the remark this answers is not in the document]
navjotk: Instead, you are suggesting not doing first touch even though the user asked for it?

examples/cfd/01_convection.ipynb
navjotk: So our OI calculation is dependent on the grid size?

devito/operator.py
navjotk: You don't need to derive defaults for something that will be overridden with a user-provided value in the next step.

examples/checkpointing/checkpoint.py
mloubout: Maybe we should split `arg_dims` and `arg_function` then to avoid memory overhead and get default loop indices/values?

examples/seismic/model.py
mloubout: I can change it, I have the `dtype` fix too in `grid`

devito/dimension.py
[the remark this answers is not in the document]
FabioLuporini: I think I understand now

devito/operator.py
FabioLuporini: what's the alternative? guarding with ``ifs`` ?

examples/seismic/model.py
mloubout: Tried it, seems to work the right way with `dam.data` and `range(nbpml)`

devito/dimension.py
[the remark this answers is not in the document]
FabioLuporini: Because I didn't know whether you would have time to fix it today, and I don't wanna see 20-day old PRs changing Travis state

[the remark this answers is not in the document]
navjotk: The PR is incorrect without this check in it. Travis should indicate that. I believe that is its purpose.

devito/operator.py
navjotk: Citing @mloubout from above: 
> first fill kwargs then complete with missing remaining ones.

FabioLuporini: I'm not sure how to explain this after what I wrote above. Feel free to try and switch it, it won't work unless with non-trivial changes (ie adding guards).

devito/dimension.py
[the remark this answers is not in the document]
FabioLuporini: > The PR is incorrect without this check in it

so is master?

Anyway, pushed it.

devito/dimension.py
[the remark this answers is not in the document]
FabioLuporini: Added a proper docstring to ``op.apply`` including, among the various things, what you pointed out here.

devito/function.py
```diff
@@ -1006,37 +985,36 @@ def inject(self, field, expr, offset=0, u_t=None, p_t=None):
...
         key = alias or self
-        args = super(SparseFunction, self).argument_defaults(alias=alias)
-        args.update(self.coordinates.argument_defaults(alias=key.coordinates))
 
+        args = super(SparseFunction, self)._arg_defaults(alias=alias)
+        args.update(self.coordinates._arg_defaults(alias=key.coordinates))
         return args
 
-    def argument_values(self, alias=None, **kwargs):
+    def _arg_values(self, alias=None, **kwargs):
```
mloubout: This function should not be necessary, but I can just redo my branch once this is merged,.

navjotk:
So the error was not caused by what I was expecting - I was looking in the wrong place. The problem is still with getting default argumentsI've pushed two more commits here that bypass the arguments check to make sure the tests pass. Now that you know how to make the tests pass, feel free to play around with the code as you see fit.

devito/operator.py
```diff
@@ -114,63 +119,60 @@ def __init__(self, expressions, **kwargs):
...
                 else:
-                    arguments[dim.symbolic_size.name] = arg.value(osize)
+                    args[dim.symbolic_size.name] = arg.value(osize)
 
         # Add in the profiler argument
-        arguments[self.profiler.name] = self.profiler.new()
+        args[self.profiler.name] = self.profiler.new()
+
+        # Add in any backend-specific argument
+        args.update(kwargs.get('backend', {}))
```
navjotk: You should always be using `kwargs.pop` to access `kwargs` and then ensure that it is empty in the end.

devito/dse/aliases.py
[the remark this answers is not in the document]
navjotk: I still don't see a docstring here.

FabioLuporini:
merged

Issue #508, opened 2018-03-06
openmp parallelism not produced under some circumstances involving subdims


tjb900:
It looks like something is going astray in the detection of parallelism with the combination of `region=INTERIOR` and complex dependencies (as seen in e.g. Higdon boundaries).

This works:

```
from devito import (Grid, Operator, TimeFunction, configuration, Eq,
                    INTERIOR, DOMAIN)

configuration["openmp"] = True

grid = Grid(shape=(20,20))
x, y = grid.dimensions
t = grid.time_dim

u = TimeFunction(name='u', grid=grid, save=10, time_order=1)

steps = []
steps.append(Eq(u.indexed[t+1, x, y],
                u.indexed[t+1, x-1, y] + u.indexed[t, x, y], region=DOMAIN))
op = Operator(steps)

print(op)
```

producing:

```
    struct timeval start_section_0, end_section_0;
    gettimeofday(&start_section_0, NULL);
    for (int x = x_s + 1; x < x_e; x += 1)
    {
      #pragma omp parallel
      {
        #pragma omp for schedule(static)
        for (int y = y_s; y < y_e; y += 1)
        {
          u[time + 1][x][y] = u[time][x][y] + u[time + 1][x - 1][y];
        }
      }
    }
```

(though note that it would be presumably be faster with the x and y loops inverted?)

But when we use `region=INTERIOR`, the parallelism disappears:

```
    for (int xi = xi_s + 1; xi < xi_e; xi += 1)
    {
      for (int yi = yi_s; yi < yi_e; yi += 1)
      {
        u[time + 1][xi][yi] = u[time][xi][yi] + u[time + 1][xi - 1][yi];
      }
    }
```

FabioLuporini:
> (though note that it would be presumably be faster with the x and y loops inverted?)

I think the compiler here is acting smartly; it detects a dependence in the ``x`` dimension due to ``u[time + 1][x][y]`` (LHS) and ``u[time + 1][x - 1][y]``, so it picks ``y``, which is a dependence-free loop

> But when we use region=INTERIOR, the parallelism disappears:

this one is odd; investigating

FabioLuporini:
> this one is odd; investigating

Apparently this is getting fixed by your PRs

tjb900:
> I think the compiler here is acting smartly; it detects a dependence in the x dimension due to u[time + 1][x][y] (LHS) and u[time + 1][x - 1][y], so it picks y, which is a dependence-free loop

[...]

PR #509, opened 2018-03-06
dependence tests to use function index (i.e. parent) dimension

tjb900:
proposed fix for #508 

I think this might implicitly assume that functions are never indexed by subdimensions - which I guess seems reasonable...  probably?

There is still more work to do:

* tests
* fixing this case, which could be parallel over `xi` but isn't.

```
    for (int xi = xi_s + 1; xi < xi_e - 1; xi += 1)
    {
      for (int yl = yl_s; yl < yl_e - 1; yl += 1)
      {
        u[time + 1][xi][yl] = a*u[time + 1][xi][yl + 1] + b*u[time][xi][yl + 1] + c*u[time][xi][yl];
      }
      for (int yr = yr_s + 1; yr < yr_e; yr += 1)
      {
        u[time + 1][xi][yr] = a*u[time + 1][xi][yr - 1] + b*u[time][xi][yr - 1] + c*u[time][xi][yr];
      }
    }
```

Interestingly, I think this one legitimately fails to meet either of the criteria:

```
             # (d_1, ..., d_{i-1}) > 0, OR
             # (d_1, ..., d_i) = 0
```

However, I can't see how it's unsafe.

### DIFF
--- a/devito/ir/iet/analysis.py
+++ b/devito/ir/iet/analysis.py
@@ -80,8 +80,9 @@ def mark_parallel(analysis):
             # (d_1, ..., d_i) = 0
             is_sequential = False
             for dep in analysis.scopes[i].d_all:
-                if not ((dims[:-1] and any(dep.is_carried(d) for d in dims[:-1])) or
-                        all(dep.is_independent(d) for d in dims)):
+                test0 = dims[:-1] and any(dep.is_carried(d) for d in dims[:-1])
+                test1 = all(dep.is_independent(d) for d in dims)
+                if not (test0 or test1):
                     is_sequential = True
                     break
             properties[i] = SEQUENTIAL if is_sequential else PARALLEL
--- a/devito/ir/support/basic.py
+++ b/devito/ir/support/basic.py
@@ -477,7 +477,11 @@ class Dependence(object):
     def is_carried(self, dim=None):
         """Return True if a dimension-carried dependence, False otherwise."""
         try:
-            return (self.distance > 0) if dim is None else self.cause == dim
+            if dim is None:
+                return self.distance > 0
+            else:
+                testdim = dim.parent if dim.is_Sub else dim
+                return self.cause == testdim
         except TypeError:
             # Conservatively assume this is a carried dependence
             return True
@@ -488,7 +492,8 @@ class Dependence(object):
             if dim is None or self.source.is_irregular or self.sink.is_irregular:
                 return self.distance == 0
             else:
-                return self.cause != dim
+                testdim = dim.parent if dim.is_Sub else dim
+                return self.cause != testdim
         except TypeError:
             # Conservatively assume this is not dimension-independent
             return False


mloubout:
> functions are never indexed by subdimensions 

what do you mean by that?

tjb900:
I mean that we wouldn't want to have a function which is allocated over only part of the domain?  Although I can think of possible exceptions (extra time history for boundary regions, for instance).

The `Dependence` has a member `findices`, which I thought was (or could be) the `Dimension`s associated with the `Function`.  For access where the loop variable is over a `SubDimension`, I think the change I made forces the comparison to always be between members of the `findices` and the parent dimension?  So this wouldn't work if `findices` ever contained a `SubDimension`.

mloubout:
Isn't it a bit risky if there is already an exception.

devito/ir/iet/analysis.py
```diff
@@ -80,8 +80,9 @@ def mark_parallel(analysis):
             # (d_1, ..., d_i) = 0
             is_sequential = False
             for dep in analysis.scopes[i].d_all:
-                if not ((dims[:-1] and any(dep.is_carried(d) for d in dims[:-1])) or
-                        all(dep.is_independent(d) for d in dims)):
+                test0 = dims[:-1] and any(dep.is_carried(d) for d in dims[:-1])
```
FabioLuporini: this is aesthetics, right?

tjb900: Oh, yeah.  The brackets were breaking my brain for a while.  I didn't mean to actually commit it, but then did accidentally and thought it might not be a bad idea after all.

Happy to drop or put in a separate PR or whatever.

devito/ir/support/basic.py
```diff
@@ -477,7 +477,11 @@ def is_increment(self):
     def is_carried(self, dim=None):
         """Return True if a dimension-carried dependence, False otherwise."""
         try:
-            return (self.distance > 0) if dim is None else self.cause == dim
+            if dim is None:
+                return self.distance > 0
+            else:
+                testdim = dim.parent if dim.is_Sub else dim
```
FabioLuporini: I'm slightly concerned about the proliferation of ``is_Sub`` in the code base.

So here's my proposal to mitigate this issue, which may also impact #504 .

We add a new method to ``Dimension``, let's call it (better names welcome) ``_bound_dimensions``. Then we do the following:

```
class Dimension(...):
    ...
    @property
    def _bound_dimensions(self):
        return {self}

...
class DerivedDimension(...):
    ...
    @property
    def _bound_dimensions(self):
        return {self, self.parent}
```

Now we can change ``is_carried`` (and, likewise, ``is_independent`` a few lines below) as:

```
try:
    if dim is None:
        return self.distance > 0
    else:
        return any(i == self.cause for i in self._bound_dimensions)
except TypeError:
    ...
```

Note that this way we're checking *both* the parent and the derived dimensions, which, in general, **should** be the right thing to do (I hope I'm not missing anything).

We could exploit this new method in #504 too, when calculating the distance (we could now check both the distance from the parent and from self). 

A benefit of this approach would be simpler portability if we start adding other types of weird dimensions

Thoughts? This is really open for discussion

devito/ir/iet/analysis.py
FabioLuporini: it's a very good idea, it was horrible before. Let's keep it !

tjb900:
Yeah, that's the case I was thinking of too...

Not currently supported in any case...  but I think Fabio's suggestion above fixes this future concern?

FabioLuporini:
> but I think Fabio's suggestion above fixes this future concern?
I think it goes in that direction, at least, yeah

devito/ir/support/basic.py
FabioLuporini: what do u think about the above @tjb900 @mloubout ?

tjb900: Sorry!  I've been meaning to get back to you on this, but I want to understand it a bit more and code up what you suggested both here and in #504 first.

I, too, find the creeping growth of is_Sub concerning - many of these exceptions feel like things that could also apply in other cases.

ggorman:
What remains to be done to land this?

FabioLuporini:
Closing as the fix is gonna be shipped by #523

PR #511, opened 2018-03-07
Fix CFD notebook by marking all cells as executed

FabioLuporini:
Required after latest nbval update


FabioLuporini:
Closing as unnecessary after #510

PR #512, opened 2018-03-07
Fix increments

### LINKED ISSUE #506: dependent writes to single variable collapsed to final one rather than sequentialised
Came across this while working on a source injection example:

```
from devito import Grid, Function, Operator, Inc
import numpy as np

grid = Grid(shape=(10,10))
u = Function(name='u', grid=grid, space_order=0)

op = Operator([Inc(u,u+1.0), Inc(u,u+2.0)])
u.data[:] = 0.0
op.apply()

# Array actually contains all 2s
assert np.all(u.data[:] == 3)
```

I don't have a use case for this - all the real injection examples work.  However the above result seems to indicate something dangerous is happening.  This is especially true considering how similar the above example is to the idiom we typically use to perform injection in the wave equation case:

[Eq(u.forward, blah blah), Inc(u.forward, u.forward+source terms)]

In all the cases so far we might just be getting saved by the fact that we generate a separate loop structure for the source injection?

Is this an ambiguity in the DSL?  I had assumed that the sequence of expressions passed to an operator is considered to be ordered, but perhaps this is a misunderstanding on my part?

FabioLuporini:
fixes #506 

@tjb900 

~Depends on #511~


FabioLuporini:
@ggorman thanks
so Travis was green until this morning; then I rebased the branch on top of latest master and now again Travis is simply running out of time, taking ages to run ...

yes, a test has been added (originally produced by @tjb900 )

PR #514, opened 2018-03-08
Non-FD examples

FabioLuporini:
Some trivial examples showing how to express basic linear algebra operations

### DIFF
--- a/examples/misc/linalg.py
+++ b/examples/misc/linalg.py
@@ -3,7 +3,8 @@ import click
 from devito import Eq, Operator, Function, dimensions, info
 from devito.tools import as_tuple
 
-__all__ = ['mat_vec', 'transpose', 'transpose_mat_vec', 'mat_mat', 'mat_mat_sum']
+__all__ = ['mat_vec', 'transpose', 'transpose_mat_vec', 'mat_mat', 'mat_mat_sum',
+           'chain_contractions']
 
 
 @click.group(chain=True)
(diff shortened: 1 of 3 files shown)


examples/misc/linalg.py
```diff
@@ -0,0 +1,134 @@
...
+    """``Ax = b``."""
+    op = Operator(Eq(b, A*x), dle=optimize)
+    op.apply()
+    info('Executed `Ax = b`')
+
+
+def transpose(A, optimize):
+    """``A -> A^T``."""
+    i, j = A.indices
+    op = Operator(Eq(A, A.indexed[j, i]), dle=optimize)
```
mloubout: Don't think this works. In place transpose of a matrix are really tricky.
A[i, j] = A[j,.i]

so when reaches A[j, i] can't replace by A[i, j] anymore.

Can be tested with `np.dot(mat_vec(A,x), y) == np.dot(mat_vec(A^T,y),x)`

examples/misc/linalg.py
```diff
@@ -0,0 +1,134 @@
...
+    op.apply()
+    info('Executed `A -> A^T`')
+
+
+def transpose_mat_vec(A, x, b, optimize):
+    """``A -> A^T, A^Tx = b``."""
+    i, j = A.indices
+    op = Operator([Eq(A, A.indexed[j, i]), Eq(b, A*x)], dle=optimize)
```
mloubout: Same as above.
This one would work simply as `Eq(b, A.indexed[j, i]*x)`

examples/misc/linalg.py
FabioLuporini: dumb me

examples/misc/linalg.py
FabioLuporini: 👍

FabioLuporini:
@mloubout can you take a final look? I addressed your comments (dumb me) and added one final example

mloubout:
Looks good. Maybe for the future add some tests and run the examples on non-zeros.

FabioLuporini:
Merged.

PR #515, opened 2018-03-09
Interpolation with precomputed coefficients

navjotk:
This PR:
- Adds (brings back) a `FixedDimension` class, used when the size of a dimension is known at the time of creation
- Adds a `PrecomputedSparseFunction` class that can do interpolation of sparse points using precomputed coefficients - this can be used to implement arbitrary interpolation schemes. 
- Adds a test for the new `PrecomputedSparseFunction` class by precomputing coefficients using a bilinear scheme

~This branch currently does not pass tests due to a DSE bug where symbols used as indices for `Indexed` types are not found by the visitors. @FabioLuporini could you look into this please. An alternate branch `interpolation2` has a few hacks in place to work around the bug and passes this test.~

### DIFF
--- a/devito/dimension.py
+++ b/devito/dimension.py
@@ -240,9 +240,19 @@ class DefaultDimension(Dimension):
         newobj.default_value = kwargs.get('default_value', None)
         return newobj
 
-    def argument_defaults(self, size=None):
-        value = self.default_value
-        return {self.start_name: 0, self.end_name: value, self.size_name: value}
+    def _arg_defaults(self, start=None, size=None, alias=None):
+        """
+        Returns a map of default argument values defined by this dimension.
+
+        :param start: (Optional) known starting point as provided by
+                      data-carrying symbols.
+        :param size: (Optional) known size as provided by data-carrying symbols.
+        :param alias: (Optional) name under which to store values.
+        """
+        dim = alias or self
+        size = size or dim.default_value
+        return {dim.min_name: start or 0, dim.max_name: size, dim.size_name: size}
+    
 
 
 class DerivedDimension(Dimension):
--- a/devito/dimension.py
+++ b/devito/dimension.py
@@ -6,8 +6,8 @@ from devito.exceptions import InvalidArgument
 from devito.logger import warning
 from devito.types import AbstractSymbol, Scalar, Symbol
 
-__all__ = ['Dimension', 'SpaceDimension', 'TimeDimension', 'SteppingDimension',
-           'SubDimension', 'ConditionalDimension', 'dimensions']
+__all__ = ['Dimension', 'SpaceDimension', 'TimeDimension', 'DefaultDimension',
+           'SteppingDimension', 'SubDimension', 'ConditionalDimension']
 
 
 class Dimension(AbstractSymbol):
@@ -16,6 +16,7 @@ class Dimension(AbstractSymbol):
     is_Space = False
     is_Time = False
 
+    is_Default = False
     is_Derived = False
     is_NonlinearDerived = False
     is_Sub = False
@@ -228,6 +229,20 @@ class TimeDimension(Dimension):
     :param name: Name of the dimension symbol.
     :param spacing: Optional, symbol for the spacing along this dimension.
     """
+
+
+class DefaultDimension(Dimension):
+    is_Default = True
+
+    def __new__(cls, name, **kwargs):
+        newobj = sympy.Symbol.__new__(cls, name)
+        newobj._spacing = kwargs.get('spacing', Scalar(name='h_%s' % name))
+        newobj.default_value = kwargs.get('default_value', None)
+        return newobj
+
+    def argument_defaults(self, size=None):
+        value = self.default_value
+        return {self.start_name: 0, self.end_name: value, self.size_name: value}
 
 
 class DerivedDimension(Dimension):
@@ -480,7 +495,6 @@ class LoweredDimension(Dimension):
 def dimensions(names):
     """
     Shortcut for: ::
-
         dimensions('i j k') -> [Dimension('i'), Dimension('j'), Dimension('k')]
     """
     assert type(names) == str
--- a/devito/dimension.py
+++ b/devito/dimension.py
@@ -252,7 +252,6 @@ class DefaultDimension(Dimension):
         dim = alias or self
         size = size or dim.default_value
         return {dim.min_name: start or 0, dim.max_name: size, dim.size_name: size}
-    
 
 
 class DerivedDimension(Dimension):
--- a/devito/dimension.py
+++ b/devito/dimension.py
@@ -246,12 +246,12 @@ class DefaultDimension(Dimension):
     def __new__(cls, name, **kwargs):
         newobj = sympy.Symbol.__new__(cls, name)
         newobj._spacing = kwargs.get('spacing', Scalar(name='h_%s' % name))
-        newobj.default_value = kwargs.get('default_value', None)
+        newobj._default_value = kwargs.get('default_value', None)
         return newobj
 
     def _arg_defaults(self, start=None, size=None, alias=None):
         dim = alias or self
-        size = size or dim.default_value
+        size = size or dim._default_value
         return {dim.min_name: start or 0, dim.max_name: size, dim.size_name: size}
 
 
--- a/devito/dimension.py
+++ b/devito/dimension.py
@@ -121,7 +121,9 @@ class Dimension(AbstractSymbol):
         :param alias: (Optional) name under which to store values.
         """
         dim = alias or self
-        return {dim.min_name: start or 0, dim.max_name: size, dim.size_name: size}
+        values = {dim.min_name: start or 0, dim.max_name: size, dim.size_name: size}
+        values = {k: values[k] for k in values if values[k] is not None}
+        return values
 
     def _arg_infers(self, args, interval, **kwargs):
         """
@@ -137,12 +139,19 @@ class Dimension(AbstractSymbol):
         :param kwargs: Dictionary of user-provided argument overrides.
         """
         infs = {}
+        defaults = self._arg_defaults()
 
         if not interval:
             return infs
 
+        if self.min_name not in args and self.min_name in defaults:
+            args[self.min_name] = defaults[self.min_name]
+
         if self.min_name in args:
             infs[self.min_name] = args[self.min_name] - min(interval.lower, 0)
+
+        if self.max_name not in args and self.max_name in defaults:
+            args[self.max_name] = defaults[self.max_name]
 
         if self.max_name in args:
             infs[self.max_name] = args[self.max_name] - (1 + max(interval.upper, 0))
--- a/devito/dimension.py
+++ b/devito/dimension.py
@@ -241,14 +241,6 @@ class DefaultDimension(Dimension):
         return newobj
 
     def _arg_defaults(self, start=None, size=None, alias=None):
-        """
-        Returns a map of default argument values defined by this dimension.
-
-        :param start: (Optional) known starting point as provided by
-                      data-carrying symbols.
-        :param size: (Optional) known size as provided by data-carrying symbols.
-        :param alias: (Optional) name under which to store values.
-        """
         dim = alias or self
         size = size or dim.default_value
         return {dim.min_name: start or 0, dim.max_name: size, dim.size_name: size}
@@ -504,6 +496,7 @@ class LoweredDimension(Dimension):
 def dimensions(names):
     """
     Shortcut for: ::
+
         dimensions('i j k') -> [Dimension('i'), Dimension('j'), Dimension('k')]
     """
     assert type(names) == str
--- a/devito/dimension.py
+++ b/devito/dimension.py
@@ -7,7 +7,7 @@ from devito.logger import warning
 from devito.types import AbstractSymbol, Scalar, Symbol
 
 __all__ = ['Dimension', 'SpaceDimension', 'TimeDimension', 'DefaultDimension',
-           'SteppingDimension', 'SubDimension', 'ConditionalDimension']
+           'SteppingDimension', 'SubDimension', 'ConditionalDimension', 'dimensions']
 
 
 class Dimension(AbstractSymbol):
--- a/devito/function.py
+++ b/devito/function.py
@@ -773,7 +773,7 @@ class AbstractSparseFunction(TensorFunction):
             self._padding = tuple((0, 0) for i in range(self.ndim))
 
             # Symbols that are encapsulated within this symbol (e.g. coordinates)
-            self.child_symbols = []
+            self.child_functions = []
 
     @classmethod
     def __indices_setup__(cls, **kwargs):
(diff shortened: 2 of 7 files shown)


devito/dimension.py
```diff
@@ -155,6 +156,21 @@ class TimeDimension(Dimension):
     """
 
 
+class FixedDimension(Dimension):
+    is_Default = True
```
FabioLuporini: broken

devito/dimension.py
```diff
@@ -155,6 +156,21 @@ class TimeDimension(Dimension):
...
+
+    def __new__(cls, name, **kwargs):
+        newobj = sympy.Symbol.__new__(cls, name)
+        newobj._reverse = kwargs.get('reverse', False)
```
FabioLuporini: things changed in Devito

devito/dimension.py
```diff
@@ -155,6 +156,21 @@ class TimeDimension(Dimension):
...
+        newobj._spacing = kwargs.get('spacing', Scalar(name='h_%s' % name))
+        newobj.default_value = kwargs.get('default_value', None)
```
FabioLuporini: if it's fixed, "default" is misleading

devito/function.py
```diff
@@ -1089,3 +1089,178 @@ def __indices_setup__(cls, **kwargs):
     @classmethod
     def __shape_setup__(cls, **kwargs):
         return kwargs.get('shape', (kwargs.get('nt'), kwargs.get('npoint'),))
+
+
+class PrecomputedSparseFunction(TensorFunction):
```
FabioLuporini: I think the two ``SparseFunction`` classes should share some code

tests/test_interpolation.py
```diff
@@ -1,9 +1,12 @@
 import numpy as np
 import pytest
 from conftest import skipif_yask
+from math import sin
+from math import floor
```
FabioLuporini: from math import sin, floor

FabioLuporini:
> @FabioLuporini could you look into this please.

Sure, welcome to my stack of TODOs :-)

devito/dimension.py
navjotk: Default is a better name than fixed.

FabioLuporini: makes me think the value can actually change. So it wouldn't be "fixed"

mloubout: Why not just value?

navjotk: We can always provide a runtime value to `Operator.apply` and the new value would be used, so it isn't fixed. Which is also why it makes sense to call the constructor argument `default_value` since it's the value to be used when one is not provided at runtime.

devito/dimension.py
```diff
@@ -369,13 +384,3 @@ def origin(self):
 
     def _hashable_content(self):
         return Symbol._hashable_content(self) + (self.origin,)
-
-
-def dimensions(names):
```
FabioLuporini: why?

navjotk: Mistake

navjotk:
A normal Dimension is associated with a `Function` or `SparseFunction` object and uses its shape to determine its default sizes. In this case, we have a dimension that is not associated with any `Function`/`SparseFunction` object, but its size is known at the time it is being created so it can be provided as a default value. 

I don't mind making it fixed. I think this added flexibility (of changing the size of the dimension - and hence the interpolator - at runtime) is good.

ggorman:
What remains to land this?

FabioLuporini:
> What remains to land this?

none of my comments have been addressed so far (feel free to have a look) and there are two bugs that might be fixed using the machinery in #517 ; a hacky fix was proposed by Navjot in a separate branch (``interpolation2``)

tests/test_interpolation.py
```diff
@@ -50,6 +52,48 @@ def custom_points(grid, ranges, npoints, name='points'):
...
+    for i, point in enumerate(points):
+        for d in range(grid.dim):
+            coefficients[i, d, 0] = ((gridpoints[i][d] + 1)*grid.spacing[d] -
+                                     point[d])/grid.spacing[d]
+            coefficients[i, d, 1] = (point[d]-gridpoints[i][d]*grid.spacing[d])\
+                / grid.spacing[d]
+    return gridpoints, coefficients
+
+
+def test_precomputed_interpolation():
```
FabioLuporini: docstring

devito/function.py
```diff
@@ -815,6 +781,88 @@ def __indices_setup__(cls, **kwargs):
...
+
+        :param alias: (Optional) name under which to store values.
+        """
+        key = alias or self
+        args = super(AbstractSparseFunction, self).argument_defaults(alias=alias)
+        for child in self.children:
+            args.update(child.argument_defaults(alias=getattr(key, child.name).name))
+        return args
+
+    def argument_values(self, alias=None, **kwargs):
```
mloubout: Why is this needed back?

devito/function.py
```diff
@@ -737,39 +737,11 @@ def _arg_values(self, alias=None, **kwargs):
...
-
-    .. note::
-
-        The parameters must always be given as keyword arguments, since
-        SymPy uses `*args` to (re-)create the dimension arguments of the
-        symbolic function.
-    """
-
-    is_SparseFunction = True
+class AbstractSparseFunction(TensorFunction):
```
mloubout: docstring?

devito/function.py
```diff
@@ -1074,3 +1122,107 @@ def __indices_setup__(cls, **kwargs):
...
+            assert(gridpoints_data is not None)
+            gridpoints.data[:] = gridpoints_data[:]
+            self.gridpoints = gridpoints
+
+            coefficients = Function(name="%s_coefficients" % self.name,
+                                    dimensions=(self.indices[-1], Dimension(name='d'),
+                                                Dimension(name='i')),
+                                    shape=(self.npoint, self.grid.dim, self.r),
+                                    space_order=0)
+            coefficients_data = kwargs.get('coefficients', None)
```
mloubout: So for gridpoints/coefficients, wouldn't it be more consistent to keep `coordinate` and have an `interpolator` as input that computes gridpoints and coefficients from the coordinates.
```
sf = PrecomputedSparseFunction(name='s', grid=grid, r=r, npoint=len(points),
                                                   coordinates=coordinate, inteprolator=precompute_linear_interpolation)
```

devito/function.py
```diff
@@ -1074,3 +1122,107 @@ def __indices_setup__(cls, **kwargs):
...
+            t = self.grid.stepping_dim
+            expr = expr.subs(t, u_t).subs(time, u_t)
+
+        coefficients = self.coefficients.indexed
+        gridpoints = self.gridpoints.indexed
+        p, _, _ = self.coefficients.indices
+        dim_subs = []
+        coeffs = []
+        for i, d in enumerate(self.grid.dimensions):
+            rd = DefaultDimension(name="r%s" % d.name, default_value=self.r)
```
mloubout: Dow we really need a DInmension for that. Unless go to like 100 points in each direction, can just write down the r equations.

devito/function.py
```diff
@@ -1074,3 +1122,107 @@ def __indices_setup__(cls, **kwargs):
...
+            expr = expr.subs(self.indices[0], p_t)
+
+        gridpoints = self.gridpoints.indexed
+        coefficients = self.coefficients.indexed
+
+        p, _ = self.gridpoints.indices
+        dim_subs = []
+        coeffs = []
+        for i, d in enumerate(self.grid.dimensions):
+            rd = DefaultDimension(name="r%s" % d.name, default_value=self.r)
```
mloubout: same

tests/test_interpolation.py
```diff
@@ -50,6 +52,48 @@ def custom_points(grid, ranges, npoints, name='points'):
     return points
 
 
+def precompute_linear_interpolation(points, grid, origin):
```
mloubout: docstring?

tests/test_interpolation.py
```diff
@@ -50,6 +52,48 @@ def custom_points(grid, ranges, npoints, name='points'):
...
+
+    gridpoints, coefficients = precompute_linear_interpolation(points, grid, origin)
+
+    sf = PrecomputedSparseFunction(name='s', grid=grid, r=r, npoint=len(points),
+                                   gridpoints=gridpoints, coefficients=coefficients)
+    eqn = sf.interpolate(m)
+    op = Operator(eqn)
+    op()
+    expected_values = [sin(point[0]) + sin(point[1]) for point in points]
+    assert(all(np.isclose(sf.data, expected_values, rtol=1e-6)))
```
mloubout: adjoint test for this?

devito/function.py
```diff
@@ -815,6 +781,66 @@ def __indices_setup__(cls, **kwargs):
...
+    def _arg_defaults(self, alias=None):
+        """
+        Returns a map of default argument values defined by this symbol.
+
+        :param alias: (Optional) name under which to store values.
+        """
+        key = alias or self
+        args = super(AbstractSparseFunction, self)._arg_defaults(alias=alias)
+        for child in self.child_functions:
+            args.update(child._arg_defaults(alias=getattr(key, child.name).name))
```
mloubout: `alias` should be `gettattr` not just the name.

devito/dimension.py
```diff
@@ -226,6 +227,30 @@ class TimeDimension(Dimension):
...
+    is_Default = True
+
+    def __new__(cls, name, **kwargs):
+        newobj = sympy.Symbol.__new__(cls, name)
+        newobj._spacing = kwargs.get('spacing', Scalar(name='h_%s' % name))
+        newobj.default_value = kwargs.get('default_value', None)
+        return newobj
+
+    def _arg_defaults(self, start=None, size=None, alias=None):
+        """
```
FabioLuporini: drop docstring, as if identical to the superclass', then it should just be inherited. I'm now trying to follow this approach to avoid proliferation of redundant docstrings

devito/dimension.py
```diff
@@ -472,7 +497,6 @@ def _hashable_content(self):
 def dimensions(names):
     """
     Shortcut for: ::
-
```
FabioLuporini: why? It might screw sphinx

devito/function.py
```diff
@@ -785,21 +757,15 @@ def __init__(self, *args, **kwargs):
...
-            self.coordinates = coordinates
-
             # Halo region
             self._halo = tuple((0, 0) for i in range(self.ndim))
 
             # Padding region
             self._padding = tuple((0, 0) for i in range(self.ndim))
 
+            # Symbols that are encapsulated within this symbol (e.g. coordinates)
+            self.child_functions = []
```
FabioLuporini: private

devito/function.py
navjotk: You also need grid spacing and origin to map from coordinates to gridpoints. Can't stay consistent if we have to add those values too.

devito/function.py
navjotk: `Dimension`s are the only way to structure loops in devito. With a loop, the generated code is independent of the interpolator used. 
This is a WIP anyway. I don't think one way is better than the other. The best would be to pass the size information to the compiler and let the compiler decide whether it wants to unroll the loop.

devito/function.py
navjotk: Fixed

devito/dimension.py
navjotk: Fixed

devito/dimension.py
navjotk: Fixed

devito/function.py
navjotk: Fixed

tests/test_interpolation.py
navjotk: Done

tests/test_interpolation.py
navjotk: Done

tests/test_interpolation.py
navjotk: WIP

devito/function.py
mloubout: Why make this one the only Function that does not take a grid in?

navjotk: It [does](https://github.com/opesci/devito/pull/515/files#diff-f3329261db6f5e1efaecf750bcfc52e9R95)?

mloubout: If it follow every other function, you have a Grid, so you have spacing and origin

navjotk: The grid only has *symbols* for spacing and origin which are replaced by values only at apply time. To do any precomputation, I need those *values* before apply time. Not possible with the grid object.

mloubout: > I need those values before apply time.

You shouldn't , the `subs` is supposed to take care of it. Hard codingh values that may have nothing to do with the one that will be used at runtime is super risky.

mloubout: And Grid always have spacing/origin values

navjotk: Combining 
> You also need grid spacing and origin to map from coordinates to gridpoints.
with the fact that gridpoints need to be **pre**computed means I need spacing and origin before apply time. I don't know how else to explain this.
Hence, it's best to keep this mess outside of our classes and in user space - which is where this PR keeps it.

mloubout: I find this really hacky and error prone.

navjotk: This was raised in a meeting and implemented after agreement with you, @ggorman and @FabioLuporini .

navjotk: Forcing the user to do the precomputation is better than silently using the origin and spacing values used to intialise the grid - since these might be changed at apply time, following your own logic set out in:
>Hard codingh values that may have nothing to do with the one that will be used at runtime is super risky.

devito/dimension.py
```diff
@@ -226,6 +227,21 @@ class TimeDimension(Dimension):
     """
 
 
+class DefaultDimension(Dimension):
+    is_Default = True
+
+    def __new__(cls, name, **kwargs):
+        newobj = sympy.Symbol.__new__(cls, name)
+        newobj._spacing = kwargs.get('spacing', Scalar(name='h_%s' % name))
+        newobj.default_value = kwargs.get('default_value', None)
```
FabioLuporini: should this be private?

devito/function.py
```diff
@@ -815,6 +786,68 @@ def __indices_setup__(cls, **kwargs):
...
+    def _arg_defaults(self, alias=None):
+        """
+        Returns a map of default argument values defined by this symbol.
+
+        :param alias: (Optional) name under which to store values.
+        """
+        key = alias or self
+        args = super(AbstractSparseFunction, self)._arg_defaults(alias=alias)
+        for child_name in self._child_functions:
+            child = getattr(self, child_name)
```
FabioLuporini: why don't you just put the actual functions in ``_child_functions`` rather than their names?

navjotk: If I have a receiver symbol called `rec`, `self.coordinates.name` on it would give you `rec_coordinates`. If it is replaced by a symbol called `rec2`, the only way of accessing the coordinates `Function` on it is by doing `getattr(rec2, 'coordinates')` since (assuming `child = self.coordinates`) `getattr(rec2, child.name)` won't be found (the new symbol's coordinates function is automatically named `rec2_coordinates`.

navjotk: Which also means this should probably be `child = getattr(key, child_name)`. Changing that.

navjotk: This line: `args.update(child._arg_defaults(alias=getattr(key, child_name)))` will not work when storing the actual functions, only with the name, since the name is the only consistent way of accessing children across different `SparseFunction` instances.

devito/function.py
mloubout: Makes sense. Can we add some warnings at setup/runtime saying the setup is only valid for a specific grid?

This all PR is gonna be really handy for sparse-fd btw, so not trying to stop it by anyway, just making sure we don't making too different than the rest of SparseFunctions

devito/function.py
mloubout: This is really odd, `child._arg_defaults` should pick up the name itself and all `arg_defaults` take an object as alias, not a name.

FabioLuporini: sorry, I didn't see `_child_function` was a class attribute. Now it does make sense

devito/function.py
```diff
@@ -815,6 +786,68 @@ def __indices_setup__(cls, **kwargs):
...
+        if not self._cached():
+            super(SparseFunction, self).__init__(*args, **kwargs)
+
+            # Set up coordinates of sparse points
+            coordinates = Function(name='%s_coords' % self.name,
+                                   dimensions=(self.indices[-1], Dimension(name='d')),
+                                   shape=(self.npoint, self.grid.dim), space_order=0)
+            coordinate_data = kwargs.get('coordinates')
+            if coordinate_data is not None:
+                coordinates.data[:] = coordinate_data[:]
```
FabioLuporini: random question unrelated to this PR: do we really want to copy data, or would we rather want a ``SparseFunction`` to take ownership of what it's been given?

devito/function.py
```diff
@@ -1074,3 +1090,108 @@ def __indices_setup__(cls, **kwargs):
     @classmethod
     def __shape_setup__(cls, **kwargs):
         return kwargs.get('shape', (kwargs.get('nt'), kwargs.get('npoint'),))
+
+
+class PrecomputedSparseFunction(AbstractSparseFunction):
```
FabioLuporini: we need a good docstring here

devito/function.py
```diff
@@ -1074,3 +1090,108 @@ def __indices_setup__(cls, **kwargs):
...
+    is_PrecomputedSparseFunction = True
```
FabioLuporini: we need ``is_PrecomputedSparseFunction = False`` in ``class Basic``, within ``types.py``

devito/function.py
navjotk: Sure.

devito/function.py
navjotk: I don't follow what you mean. Care to elaborate?

devito/function.py
navjotk: @mloubout there is no difference between what this line is doing and what is being done [here](https://github.com/opesci/devito/blob/master/devito/function.py#L1023), except here, it works for any arbitrarily named child attributes, not just `self.coordinates`. So, for a `PrecomputedSparseFunction`, this will execute:
`args.update(self.gridpoints._arg_defaults(alias=key.gridpoints)`. I believe I am reproducing the same behaviour as master, but in a more general way.

devito/dimension.py
navjotk: Sure

devito/function.py
navjotk: Done

devito/function.py
navjotk: Done

PR #516, opened 2018-03-14
Fix weakref pickling error in tutorials.

ggorman:
Move clear_cache import inside of worker function. This avoids a weakref pickling error when using Dask.


FabioLuporini:
uncontroversial, thanks

PR #518, opened 2018-03-17
function: process data for array input for sparse functions

mloubout:
Leftover fix for `SparseFunction`

Only the coordinates were updated an input array ignored for `umpy` array input for sparse function at operator call (`op(rec=np.array)`)


devito/function.py
```diff
@@ -1027,13 +1027,17 @@ def argument_values(self, alias=None, **kwargs):
 
         new = kwargs.get(self.name)
         key = alias or self
+        values = {}
```
FabioLuporini: ? can be dropped, see my comment below

devito/function.py
```diff
@@ -1027,13 +1027,17 @@ def argument_values(self, alias=None, **kwargs):
...
 
         if new is not None and isinstance(new, SparseFunction):
             # If we've been replaced with a SparseFunction,
             # we need to re-derive defaults and values...
             values = new.argument_defaults(alias=key).reduce_all()
+            values.update(new.coordinates.argument_values(alias=key.coordinates,
+                                                          **kwargs))
         else:
             # ..., but if not, we simply need to recurse over children.
+            values[key.name] = key._data_buffer if new is None else new
```
FabioLuporini:
```
values = {key.name: key._data_buffer if new is None else new}
values.update(...)
```
this way the ``if`` and ``else`` branches are homogeneous

mloubout: Tried it, failed `test_interpolation`

ggorman:
Is this now good to merge?

mloubout:
If this turns green, I think so

devito/function.py
```diff
@@ -1032,9 +1032,18 @@ def argument_values(self, alias=None, **kwargs):
...
-            values = self.coordinates.argument_values(alias=key, **kwargs)
+            # Set the data to the input if it is an array
+            values = {key.name: key._data_buffer if new is None else new}
+            # Process indices
+            shape = self.shape if new is None else new.shape
+            for i, s, o in zip(self.indices, shape, self.staggered):
+                values.update(i.argument_defaults(size=s+o))
+            # Take default coordinates values
+            defaults = key.coordinates.argument_defaults(alias=key.coordinates)
+            values.update(defaults.reduce_all())
```
mloubout: Split in two lines for flake8

mloubout:
This PR is quite urgent so I'd say let's merge it. Bit there is a lot of inconsistencies and weird things in function arguments that need to be fixed.

ggorman:
It also seems like you have left quite a few debugging print statements behind you.

mloubout:
>It also seems like you have left quite a few debugging print statements behind you.

Where?

mloubout:
@FabioLuporini should be updated according to your comments.

mloubout:
>What were you missing before?

Some `self` instead of `key` leftover were screwing stuff around.

devito/function.py
```diff
@@ -1025,18 +1022,20 @@ def argument_values(self, alias=None, **kwargs):
...
-            values = new.argument_defaults(alias=key).reduce_all()
-        else:
-            # ..., but if not, we simply need to recurse over children.
-            values = self.coordinates.argument_values(alias=key, **kwargs)
+        # Add value override for own data if it is provided
+        if self.name in kwargs:
+            # Update function value
+            values = super(SparseFunction, key).argument_values(alias=key, **kwargs)
+            # Get coordinates if input is SparseFunction else take self
+            new = kwargs.get(self.name, self)
```
FabioLuporini: ``self.name`` is surely in ``kwargs`` at this point, right?

devito/function.py
```diff
@@ -1025,18 +1022,20 @@ def argument_values(self, alias=None, **kwargs):
...
+            new = new if hasattr(new, 'coordinates') else self
```
FabioLuporini: I don't understand why you need to treat the ``else``; if the input were not a ``SparseFunction``, then ``self`` is already handled by ``super``.

I also think you should do ``if isinstance(.., SparseFunction)``, rather than using ``hasattr`` (rule of thumb: stay away from it as much as possible)

mloubout:
@FabioLuporini looks like just removing everything (almost) works fine, I guess the problem was some weird dependencies and duplicates screwing up stuff.

devito/function.py
```diff
@@ -335,8 +333,7 @@ def argument_values(self, alias=None, **kwargs):
                                  "dimensions %s" % (self.indices, ))
             if isinstance(new, TensorFunction):
                 # Set new values and re-derive defaults
-                values[key.name] = new._data_buffer
-                values.update(new.argument_defaults(alias=key).reduce_all())
+                values = new.argument_defaults(alias=key).reduce_all()
```
FabioLuporini: smart

mloubout:
@FabioLuporini is this still necessary or fixed in you're PR.

FabioLuporini:
> @FabioLuporini is this still necessary or fixed in you're PR.

TBH, I don't know, I haven't integrated your fix in my PR, though I admit I had to change some of the affected lines while I was developing that branch. I'm merging #507 nowish, so we can test it afterwards.

FabioLuporini:
merged

PR #519, opened 2018-03-19
504+509+final tweaks

### LINKED ISSUE #503: detect_flow_direction fails when subdimensions are involved

FabioLuporini:
@tjb900 this PR exploits your contribution in #504 and #509 , adds in my suggestion about avoiding the ``if is_Sub ...`` pollution, and also ships the tests that you're suggesting in #508 

Note that I've simply cherry-picked your commits, so you get the deserved credits :-)

If everybody's happy (including Travis), we can merge this instead of #504 and #509 )

fixes #503 #508

### DIFF
--- a/devito/dimension.py
+++ b/devito/dimension.py
@@ -95,6 +95,10 @@ class Dimension(AbstractSymbol):
     def _hashable_content(self):
         return super(Dimension, self)._hashable_content() + (self.spacing,)
 
+    @property
+    def _defines(self):
+        return {self}
+
     def argument_defaults(self, size=None, alias=None):
         """
         Returns a map of default argument values defined by this symbol.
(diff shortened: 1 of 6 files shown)


FabioLuporini:
Reopening once #507 will have landed

PR #522, opened 2018-03-25
Resample

ggorman:
Adds support for resampling PointSource data in time.


FabioLuporini:
@ggorman please rebase it on top of latest master so that we can have a look

PR #523, opened 2018-03-26
Stronger SubDimensions

### LINKED ISSUE #503: detect_flow_direction fails when subdimensions are involved

### LINKED ISSUE #508: openmp parallelism not produced under some circumstances involving subdims

### LINKED ISSUE #513: Automatic detection of flow with mixed save/nosave time functions
So a small bug that turned out trying to save one of the fields in `born`, 

```
from devito import *
grid = Grid((10, 10))
u1 = TimeFunction(name="u1", grid=grid, save=10, time_order=1)
u2 = TimeFunction(name="u2", grid=grid, time_order=2)
u1 = TimeFunction(name="u1", grid=grid, save=10, time_order=2)
op = Operator([Eq(u1.forward, u1 + 2.0 - u1.backward), Eq(u2.forward, u2 + 2*u2.backward -  u1.dt2)])
```

Fails to detect the flow direction

```
  File "<stdin>", line 1, in <module>
  File "/nethome/mlouboutin3/Dropbox/London/CodeGen/devito/devito/backends.py", line 75, in __call__
    return t(*args, **kwargs)
  File "/nethome/mlouboutin3/Dropbox/London/CodeGen/devito/devito/core/operator.py", line 54, in __new__
    obj.__init__(*args, **kwargs)
  File "/nethome/mlouboutin3/Dropbox/London/CodeGen/devito/devito/operator.py", line 82, in __init__
    clusters = clusterize(expressions)
  File "/nethome/mlouboutin3/Dropbox/London/CodeGen/devito/devito/ir/clusters/algorithms.py", line 232, in clusterize
    flowmap = detect_flow_directions(g)
  File "/nethome/mlouboutin3/Dropbox/London/CodeGen/devito/devito/ir/support/utils.py", line 98, in detect_flow_directions
    for k, v in mapper.items() if k.is_Stepping)
AssertionError
```

### LINKED ISSUE #521: extra occurrence of no parallelism using subdims
Very similar to #508 , but with the flow dependence on `y` instead of `x`.

```
from devito import (Grid, Operator, TimeFunction, configuration, Eq,
                    INTERIOR, DOMAIN)

configuration["openmp"] = True

grid = Grid(shape=(20,20))
x, y = grid.dimensions
t = grid.time_dim

u = TimeFunction(name='u', grid=grid, save=10, time_order=1)

reg = INTERIOR

steps = []
steps.append(Eq(u.indexed[t+1, x, y],
                u.indexed[t+1, x, y-1] + u.indexed[t, x, y], region=reg))
op = Operator(steps)
```

Loops are parallel over `x` when `reg=DOMAIN`, but not with `reg=INTERIOR`.

Not fixed by #509 or #519.

FabioLuporini:
This PR does a minor cleanup and, most importantly, fixes a number of issues associated with SubDimensions

fixes #503
fixes #508
fixes #513
fixes #521 

Tests for each of these issues are shipped

@mloubout @tjb900 please double check that the issues are actually gone with this PR

### DIFF
--- a/devito/dimension.py
+++ b/devito/dimension.py
@@ -102,6 +102,10 @@ class Dimension(AbstractSymbol):
         return super(Dimension, self)._hashable_content() + (self.spacing,)
 
     @property
+    def _defines(self):
+        return {self}
+
+    @property
     def _arg_names(self):
         """Return a tuple of argument names introduced by this dimension."""
         return (self.name, self.size_name, self.ext_name, self.max_name, self.min_name)
@@ -257,6 +261,10 @@ class DerivedDimension(Dimension):
 
     def _hashable_content(self):
         return (self.parent._hashable_content(),)
+
+    @property
+    def _defines(self):
+        return {self} | self.parent._defines
 
     @property
     def _arg_names(self):
--- a/devito/ir/equations/equation.py
+++ b/devito/ir/equations/equation.py
@@ -135,7 +135,7 @@ class ClusterizedEq(Eq, IREq, FrozenExpr):
 
     """
     ClusterizedEq(expr, ispace, dspace)
-    ClusterizedEq(lhs, rhs, ispace=IterationSpace, dspace=DataSpace)
+    ClusterizedEq(lhs, rhs, stamp=ClusterizedEq)
 
     A SymPy equation with associated :class:`IterationSpace` and
     :class:`DataSpace`.
@@ -154,23 +154,24 @@ class ClusterizedEq(Eq, IREq, FrozenExpr):
     """
 
     def __new__(cls, *args, **kwargs):
-        if len(args) == 3:
+        if len(args) == 2:
+            # origin: ClusterizedEq(lhs, rhs, stamp=...)
+            stamp = kwargs.pop('stamp')
+            assert isinstance(stamp, ClusterizedEq)
+            expr = Eq.__new__(cls, *args, evaluate=False)
+            expr.is_Increment = stamp.is_Increment
+            expr._ispace, expr._dspace = stamp.ispace, stamp.dspace
+        elif len(args) == 3:
             # origin: ClusterizedEq(expr, ispace, dspace)
             input_expr, ispace, dspace = args
             assert isinstance(ispace, IterationSpace) and isinstance(dspace, DataSpace)
             expr = Eq.__new__(cls, *input_expr.args, evaluate=False)
             expr.is_Increment = input_expr.is_Increment
-            expr._dspace = dspace
-            expr._ispace = ispace
+            expr._ispace, expr._dspace = ispace, dspace
         else:
-            # origin: ClusterizedEq(lhs, rhs, ispace=..., dspace=...)
-            assert len(args) == 2
-            expr = Eq.__new__(cls, *args, evaluate=False)
-            expr.is_Increment = kwargs.get('is_Increment', False)
-            expr._dspace = kwargs['dspace']
-            expr._ispace = kwargs['ispace']
+            raise ValueError("Cannot construct ClusterizedEq from args=%s "
+                             "and kwargs=%s" % (str(args), str(kwargs)))
         return expr
 
     def func(self, *args, **kwargs):
-        return super(ClusterizedEq, self).func(*args, dspace=self._dspace,
-                                               ispace=self._ispace)
+        return super(ClusterizedEq, self).func(*args, stamp=self)
--- a/devito/ir/support/basic.py
+++ b/devito/ir/support/basic.py
@@ -252,9 +252,16 @@ class IterationInstance(Vector):
     def is_irregular(self):
         return not self.is_regular
 
-    def distance(self, other, findex=None):
-        """Compute vector distance from ``self`` to ``other``. If ``findex`` is
-        supplied, compute the vector distance up to and including ``findex``."""
+    def distance(self, other, findex=None, view=None):
+        """Compute the distance from ``self`` to ``other``.
+
+        :param other: The :class:`IterationInstance` from which the distance
+                      is computed.
+        :param findex: (Optional) if supplied, compute the distance only up to
+                       and including ``findex`` (defaults to None).
+        :param view: (Optional) an iterable of ``findices`` (defaults to None); if
+                     supplied, project the distance along these dimensions.
+        """
         if not isinstance(other, IterationInstance):
             raise TypeError("Cannot compute distance from obj of type %s", type(other))
         if self.findices != other.findices:
@@ -266,7 +273,12 @@ class IterationInstance(Vector):
                 raise TypeError("Cannot compute distance as `findex` not in `findices`")
         else:
             limit = self.rank
-        return super(IterationInstance, self).distance(other)[:limit]
+        distance = super(IterationInstance, self).distance(other)[:limit]
+        if view is None:
+            return distance
+        else:
+            proj = [i for i, d in zip(distance, self.findices) if d in as_tuple(view)]
+            return Vector(*proj)
 
     def section(self, findices):
         """Return a view of ``self`` in which the slots corresponding to the
--- a/devito/ir/support/basic.py
+++ b/devito/ir/support/basic.py
@@ -480,8 +480,7 @@ class Dependence(object):
             if dim is None:
                 return self.distance > 0
             else:
-                testdim = dim.parent if dim.is_Sub else dim
-                return self.cause == testdim
+                return any(i == self.cause for i in dim._defines)
         except TypeError:
             # Conservatively assume this is a carried dependence
             return True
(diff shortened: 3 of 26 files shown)


tjb900:
So I am just running this through some of the failures I had when submitting the initial issues.  I believe it fixes the issues I submitted, but I had pared some of the issues down to the simplest possible failing case.

This one, which I had originally written as the test case for #521, now throws an error:

```
from devito import (Grid, Operator, TimeFunction, configuration, Eq,
                    INTERIOR, DOMAIN, SubDimension)

configuration["openmp"] = True

grid = Grid(shape=(20,20))
x, y = grid.dimensions
t = grid.time_dim

u = TimeFunction(name='u', grid=grid, save=10, time_order=1)

reg = INTERIOR

steps = []
steps.append(Eq(u.indexed[t+1, x, y],
                u.indexed[t+1, x, y-1] + u.indexed[t, x, y], region=reg))
steps.append(Eq(u.indexed[t+1, x, y],
                u.indexed[t+1, x, y+1] + u.indexed[t, x, y], region=reg))
op = Operator(steps)
```
produces
```
ValueError: Cannot merge `IterationSpace`s with incompatible directions
```

It's worth noting here that it works if I replace `y` with the SubDimension for the left and right, as I would in a real case.  Therefore, this is not a practical example.

However, I had believed until now that this was nonetheless legitimate.  It was my understanding that a sequence of equations passed to an operator would produce an equivalent result as if they had been executed in the sequence provided.  So I think this may still be a bug: these two might not be able to be merged, but that just means they should be run in sequence.

devito/ir/support/utils.py
```diff
@@ -127,16 +127,17 @@ def detect_flow_directions(exprs):
             if not dimensions:
                 continue
             for d in dimensions:
-                if d.is_Sub:
-                    testdim = d.parent
-                else:
-                    testdim = d
-
+                distance = None
+                for i in d._defines:
```
tjb900: This loop is sensitive to order of iteration over a set.  That scares me a little.  Is it theoretically possible for `w.distance(r,i)` to be valid for more than one `i`?

FabioLuporini:
> So I think this may still be a bug: these two might not be able to be merged, but that just means they should be run in sequence.

absolutely, it's a stupid check that I added, but I should actually drop it! Will fix it tmr

tests/test_operator.py
```diff
@@ -1186,35 +1186,78 @@ def test_equations_mixed_timedim_stepdim(self):
...
-        assert np.all(u.data[1, :, :, -1] == 1)
-        assert np.all(u.data[1, 1:3, 1:3, 1:3] == 3)
+        # Deliberately inject into u1, rather than u1.forward, to create a WAR
+        eqn1 = Eq(u1.forward, u1 + 2.0 - u1.backward)
+        eqn2 = sf1.inject(u1, expr=sf1)
+        eqn3 = Eq(u2.forward, u2 + 2*u2.backward - u1.dt2)
+        eqn4 = sf2.interpolate(u2)
+
+        op = Operator([eqn1] + eqn2 + [eqn3] + eqn4)
+        trees = retrieve_iteration_tree(op)
```
mloubout:
```
        u1 = TimeFunction(name="u1", grid=grid, save=10, time_order=2)
        u2 = TimeFunction(name="u2", grid=grid, time_order=2)
        const = Function(name="const", grid=grid, space_order=2)
        sf1 = SparseFunction(name='sf1', grid=grid, npoint=1, ntime=10)
        sf2 = SparseFunction(name='sf2', grid=grid, npoint=1, ntime=10)

        # Deliberately inject into u1, rather than u1.forward, to create a WAR
        from sympy import cos
        eqn1 = Eq(u1.forward, u1 + 2.0 - u1.backward + cos(const))
        eqn2 = sf1.inject(u1.forward, expr=sf1)
        eqn3 = Eq(u2.forward, u2 + 2*u2.backward - u1.dt2 + cos(const))
        eqn4 = sf2.interpolate(u2)

        op = Operator([eqn1] + eqn2 + [eqn3] + eqn4)
```

This one is slightly more complicated, and can be part of some other PR but creates two issues:
- time invariants are redeclared with same name for different clusters
- Time loop is interupted by the second time invariant and restarted

FabioLuporini: I suspect we can produce a smaller failing example without using ``SparseFunction``. Will spend some time on this later this afternoon

FabioLuporini:
@tjb900 @mloubout can you take a final look? I should have addressed your latest issues. Hopefully now we even have a green Travis...

devito/ir/equations/equation.py
```diff
@@ -154,23 +154,24 @@ class ClusterizedEq(Eq, IREq, FrozenExpr):
     """
 
     def __new__(cls, *args, **kwargs):
-        if len(args) == 3:
+        if len(args) == 2:
```
tjb900: Readability: overloading different meanings based on the number of arguments.  maybe two `classmethod`s could be preferable here?  e.g. `from_lhs_rhs_stamp` `from_expr_ispace_dspace`?

devito/ir/support/basic.py
```diff
@@ -266,7 +280,12 @@ def distance(self, other, findex=None):
                 raise TypeError("Cannot compute distance as `findex` not in `findices`")
         else:
             limit = self.rank
-        return super(IterationInstance, self).distance(other)[:limit]
+        distance = super(IterationInstance, self).distance(other)[:limit]
+        if view is None:
+            return distance
+        else:
+            proj = [i for i, d in zip(distance, self.findices) if d in as_tuple(view)]
```
tjb900: This caught my eye; I had to read it twice to realise it was right.  Having `i` for "distance" and `d` for "index" I think is confusing?  (I know it's 'd' because it's a dimension, but maybe 'dist' instead of 'i')?

devito/dimension.py
```diff
@@ -101,6 +101,10 @@ def base(self):
     def _hashable_content(self):
         return super(Dimension, self)._hashable_content() + (self.spacing,)
 
+    @property
```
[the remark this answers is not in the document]
FabioLuporini: so you can have an equation in which only a derived dimension appears; however, this derived dimension is actually "defining" two objects, itself and its parent (main) dimension. That was the intention, but I agree we should probably find a better name...

devito/ir/equations/equation.py
FabioLuporini: but you would still need to use ``if-then-else`` to invoke one or the other? or what am I missing?

devito/ir/support/basic.py
FabioLuporini: oooooohhh yes! that's quite horrible, so I'll definitely swap them :)

devito/ir/equations/equation.py
tjb900: Well, I meant that you'd call one of the classmethods to create an object, rather than using __new__ (or perhaps __new__ is for one of the patterns only).  But I guess that's not great either.  I don't know.

FabioLuporini: ah! well no that's not possible because `__new__` is called via SymPy too, not just internally by Devito (in fact, I think it's never called explicitly in Devito)

devito/ir/support/basic.py
FabioLuporini: done

FabioLuporini:
Merged.

PR #525, opened 2018-03-27
Added support for resampling time series data in PointSource.

ggorman:


examples/seismic/source.py
```diff
@@ -49,12 +57,47 @@ def __new__(cls, name, grid, ntime=None, npoint=None, data=None,
...
+            obj.t0 = time[0]
+            obj.dt = time[1]-time[0]
+
         return obj
 
     def __init__(self, *args, **kwargs):
         if not self._cached():
             super(PointSource, self).__init__(*args, **kwargs)
 
+    def resample(self, dt, window=None):
```
mloubout: @property with setter instead of new obj?

examples/seismic/source.py
```diff
@@ -49,12 +57,47 @@ def __new__(cls, name, grid, ntime=None, npoint=None, data=None,
...
+        # Create resampled data.
+        data = np.zeros((nt, self.data.shape[1]))
+        for i in range(self.data.shape[1]):
+            data[:, i] = signal.resample(self.data[:, i], nt, window=window)
+
+        # Return new object
+        return PointSource(self.name, self.grid, t0=self.t0, dt=dt, data=data,
+                           coordinates=self.coordinates.data)
+
+    def time(self):
```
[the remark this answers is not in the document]
ggorman: Good point - this has been done now in most cases if you want to take a look at the next commit.

examples/seismic/source.py
```diff
@@ -49,12 +57,47 @@ def __new__(cls, name, grid, ntime=None, npoint=None, data=None,
...
+
+        dt = (tn-t0)/nt
+
+        # Create resampled data.
+        data = np.zeros((nt, self.data.shape[1]))
+        for i in range(self.data.shape[1]):
+            data[:, i] = signal.resample(self.data[:, i], nt, window=window)
+
+        # Return new object
+        return PointSource(self.name, self.grid, t0=self.t0, dt=dt, data=data,
```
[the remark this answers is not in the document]
ggorman: I think we don't want to destroy the original data as it might be the original data read from disk for example.

examples/seismic/source.py
ggorman: Same as previous comment.

examples/seismic/source.py
```diff
@@ -49,12 +57,47 @@ def __new__(cls, name, grid, ntime=None, npoint=None, data=None,
...
+                           coordinates=self.coordinates.data)
+
+    def time(self):
+        return np.linspace(self.t0, self.t0+self.data.shape[0]*self.dt,
```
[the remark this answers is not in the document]
ggorman: Replaced code - reduced API to reduce chance of error.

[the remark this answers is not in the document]
mloubout: No I mean if there is `nt` time step and the time step is `dt`, the last time is `(nt-1)*dt` not `nt*dt`

examples/seismic/source.py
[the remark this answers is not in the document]
mloubout: So there is two case
- from disk = > do whatever you want you can read it back
- not from disk => that's where you wand correct forward/adjoint resampling to afford to d o everything in place without duplicating memory and create more ref to track.

I guess it's a detail and doesn't need to be fix in here, but worth considering.

examples/seismic/acoustic/acoustic_example.py
```diff
@@ -43,7 +43,7 @@ def acoustic_setup(shape=(50, 50, 50), spacing=(15.0, 15.0, 15.0),
     src.coordinates.data[0, -1] = model.origin[-1] + 2 * spacing[-1]
 
     # Define receiver geometry (spread across x, just below surface)
-    rec = Receiver(name='rec', grid=model.grid, ntime=nt, npoint=nrec)
+    rec = Receiver(name='rec', grid=model.grid, time=time, npoint=nrec)
```
FabioLuporini: ``grid`` has a reference to time, namely ``grid.time_dim`` or something along these lines. Can you rather use that?

FabioLuporini: same story below

examples/seismic/source.py
```diff
@@ -8,53 +10,100 @@
...
-                       representing the number of points in this source
+                       representing the number of points in this source.
 
     Note, either the dimensions `ntime` and `npoint` or the fully
     initialised `data` array need to be provided.
     """
 
-    def __new__(cls, name, grid, ntime=None, npoint=None, data=None,
-                coordinates=None, **kwargs):
+    def __new__(self, name, grid, time_range=None, time=None, npoint=None,
```
FabioLuporini: the convention is that ``self`` is used as a reference to the object; ``cls`` is the class type. ``new`` should return an instance of ``cls``. So here you should probably stick to ``cls``, not ``self``

examples/seismic/source.py
```diff
@@ -8,53 +10,100 @@
...
             obj.data[:] = data
+
         return obj
 
     def __init__(self, *args, **kwargs):
         if not self._cached():
             super(PointSource, self).__init__(*args, **kwargs)
 
+    @property
+    def t_start(self):
```
FabioLuporini: is this expected to be called from the outside/user? if not, make it private

examples/seismic/source.py
```diff
@@ -8,53 +10,100 @@
...
-        if data is None:
-            if ntime is None:
-                error('Either data or ntime are required to'
-                      'initialise source/receiver objects')
+
+        if time_range is None and time is None:
+            error("Either time_range or time must be defined.")
+
+        if time_range is None:
+            self._time_range = (time[0], time[1]-time[0], time.size)
```
FabioLuporini: I suggest to use a ``namedtuple`` for this

examples/seismic/source.py
```diff
@@ -71,17 +120,15 @@ class WaveletSource(PointSource):
     :param time: Discretized values of time in ms
     """
 
-    def __new__(cls, *args, **kwargs):
-        time = kwargs.get('time')
+    def __new__(self, *args, **kwargs):
```
FabioLuporini: ``cls``, not ``self``

examples/seismic/source.py
```diff
@@ -71,17 +120,15 @@ class WaveletSource(PointSource):
...
         npoint = kwargs.get('npoint', 1)
-        kwargs['ntime'] = len(time)
         kwargs['npoint'] = npoint
-        obj = PointSource.__new__(cls, *args, **kwargs)
+        obj = PointSource.__new__(self, *args, **kwargs)
 
-        obj.time = time
         obj.f0 = kwargs.get('f0')
+        _time = obj.time
```
FabioLuporini: unnecessary?

examples/seismic/source.py
```diff
@@ -8,53 +10,100 @@
...
+    @property
+    def ntime(self):
+        return self._time_range[2]
+
+    @property
+    def dt(self):
+        return self._time_range[1]
+
+    @property
+    def time(self):
```
FabioLuporini: are you using ``time`` for two different things? usually, ``time`` refers to the time ``Dimension``

examples/seismic/acoustic/acoustic_example.py
ggorman: Isn't the problem that grid doesn't always have a time_dim?

examples/seismic/source.py
ggorman: It can be called outside in cases such as when we need to clone the receiver to store the data misfit. I think there is an argument for making it all private and adding a member function to do the cloning. Thoughts?

FabioLuporini: agreed. 

Probably unrelated to this PR: Is cloning an operation that we expect to be needed on other kind of objects? if so, we should provide a superclass capturing this "clonable" property

examples/seismic/source.py
ggorman: Good idea - added to followup commit.

examples/seismic/acoustic/acoustic_example.py
FabioLuporini: ok if you need a real value that's fine, but then calling it ``time`` is confusing because this term is also used everywhere to refer to the time ``:class:`Dimension```

examples/seismic/source.py
ggorman: done

examples/seismic/source.py
ggorman: Yes - this invokes linspace and the following loop where its used could be large.

examples/seismic/source.py
ggorman: Ah - I see the confusion. I need to rename this to time_values perhaps.

examples/seismic/source.py
FabioLuporini: perhaps a symptom that you should/could use a `cached_property` for time

examples/seismic/acoustic/acoustic_example.py
ggorman: Renamed to time_values.

examples/seismic/source.py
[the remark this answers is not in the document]
ggorman: I propose treating this as a feature request for another PR - we can add an argument to enable rewriting in-place. However, my feeling is still that the default behaviour should be not to modify the original object.

examples/seismic/source.py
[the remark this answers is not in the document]
ggorman: It depends on whether you think we should be using:
np.linspace(..., endpoint=False)
or
np.linspace(..., endpoint=True)
I'm advocating endpoint=False because this results in fewer -1's flying about the place.

examples/seismic/source.py
ggorman: Done

examples/seismic/source.py
ggorman: For that moment I made most of these private as you suggested.

examples/seismic/source.py
ggorman: done

examples/seismic/source.py
ggorman: Good point.

examples/seismic/source.py
```diff
@@ -1,60 +1,114 @@
...
+
         return obj
 
     def __init__(self, *args, **kwargs):
         if not self._cached():
             super(PointSource, self).__init__(*args, **kwargs)
 
+    @property
+    def _t_start(self):
+        return self._time_range[0]
```
FabioLuporini: you can exploit the ``namedtuple`` now: you can use its fields rather than explicit indexing

examples/seismic/source.py
```diff
@@ -1,60 +1,114 @@
...
+
+    @property
+    def _t_end(self):
+        return self._time_range[1]*self._time_range[2]
```
FabioLuporini: same, and again below

examples/seismic/source.py
```diff
@@ -1,60 +1,114 @@
...
+        if time_range is None and time_values is None:
+            error("Either time_range or time_values must be defined.")
+
+        Time =  namedtuple('Time', ['t0', 'dt', 'nt'])
+        if time_range is None:
+            cls._time_range = Time(time_values[0], time_values[1]-time_values[0], time_values.size)
         else:
-            ntime = ntime or data.shape[0]
+            cls._time_range = Time(*time_range)
+        nt = cls._time_range[2]
```
FabioLuporini: exploit the ``namedtuple``

examples/seismic/source.py
```diff
@@ -1,60 +1,114 @@
...
-            if ntime is None:
-                error('Either data or ntime are required to'
-                      'initialise source/receiver objects')
+
+        if time_range is None and time_values is None:
+            error("Either time_range or time_values must be defined.")
+
+        Time =  namedtuple('Time', ['t0', 'dt', 'nt'])
+        if time_range is None:
+            cls._time_range = Time(time_values[0], time_values[1]-time_values[0], time_values.size)
```
FabioLuporini: I think the clean thing to do here is to create `time_range` as a local variable, then, after ``obj = SparseTimeFunction.__ ...`` you can do ``obj._time_range = time_range``. 

This is, incidentally, the way one subclasses SymPy and NumPy classes

examples/seismic/source.py
```diff
@@ -1,60 +1,114 @@
...
+    @property
+    def time_range(self):
+        return self._time_range
+
+    def resample(self, dt, window=None):
+        # The new number of sample points and roundoff adjusted dt
+        nt = int(round((self._t_end-self._t_start)/dt))
+        dt = (self._t_end-self._t_start)/nt
+
+        if self.data is None:
```
FabioLuporini: I wonder whether you're actually triggering data allocation by calling ``self.data``. Perhaps you want ``self._data``. Even better, we could add an ``_is_allocated`` property to the root class for all Function objects, namely ``TensorFunction`` in ``function.py``. Obviously, ``_is_allocated`` would still have to check ``self._data``

examples/seismic/source.py
```diff
@@ -1,60 +1,115 @@
...
+            error("resample called but there is no data.")
+
+        # Create resampled data.
+        npoint = self.coordinates.shape[0]
+        data = np.zeros((nt, npoint))
+        for i in range(npoint):
+            data[:, i] = signal.resample(self.data[:, i], nt, window=window)
```
mloubout: Maybe replicate the `window` docstring` for information and options.

mloubout:
>GG: I am assuming that this a preprocessing step

This not necessarily just a preprocessing step. While you need to bring the field data to the computational time-step to compute the gradient, most of the data processing/QC/... is done at field data time-step, so the data generated by the forward modeling needs to be brought back to a coarser time step in most of the cases.

mloubout:
Considering 2, there  is also `signal.decimate` that does not care about rational but only for down sampling.
I'm fine with it overall, didn't you have a test previously?

examples/seismic/source.py
```diff
@@ -1,60 +1,106 @@
...
+
+class TimeTicks:
+    def __init__(self, start=None, step=None, num=None):
+        try:
+            self.start = start
+            self.stop = start + step*num
+            self.step = step
+            self.num = num
+        except:
+            error("error in time range parameters")
```
mloubout: Wouldn't it be simpler to just make input non keyword to fore it as input?

examples/seismic/source.py
```diff
@@ -1,60 +1,106 @@
...
+
+    def __str__(self):
+        return "TimeTicks(%g, %g, %g, %g)" % (self.start, self.stop, self.step, self.num)
+
+    @cached_property
+    def time_values(self):
+        return np.linspace(self.start, self.stop, self.num, endpoint=False)
```
mloubout: Not sure I like skiping endpoint, as explained in overall comment.

ggorman:
> Not a big fan of tht TimeTicks name, would call it TimeSerie or TimeAxis.

I don't like TimeSeries as I think it is easy to get this confused with time series data. Renamed it to TimeAxis in 04d10c8f; I agree it makes more sense.

examples/seismic/source.py
ggorman: Agreed. Committed b99883be

ggorman:
> I think it would be better to have it TimeTicks(t0, T, dt) to avoid any miscalculation of nt all over the place and only pass in time values (start, end, step) and let the function decide how many time steps it is.

Let me explain my rational here. TimeAxis could in principle be defined using any three of four parameters; *start*, *stop*, *step*, *num*. Obviously the first three are real and the forth is integer. I wanted to chose 3 parameters whereby the user gets *exactly* what they asked for.

Assuming that *start* is a must as it defines the time origin as is by far the most common case. Then our options are:
1. *start*, *step*, *num*
2. *start*, *stop*, *num*
3. *start*, *stop*, *step*
Option 2 was discounted as this would then dictate the value of step/dt rather than the stability requirements of the model. 
Option 3 was discounted because in general *step*/dt does not divide evenly into the range (stop-start), meaning that TimeAxis needs to decide whether it is more appropriate to convert (stop-start)/step to an integer via floor, ceil or round. And the object returned then has a different *end* than the input. 
As you suggested the user is surely going to make inconsistent mistakes when calculating nt themselves but this will ultimately only result in them calculating one more or less time steps than expected. When size does matter it is also much easier to detect an error. However, a mistake in calculating dt is much more detrimental and more difficult to detect (might just be a tiny phase shift...).

Does this argument stack up?

mloubout:
Concerning 3, as this PR introduce time resampling, adding an extra time step doesn't create any problem, `ceil` will always work. What I like about 3, is that it can be just reset with a new CFL and everything folows without any user recomputation.
SO yes, 1 or 3 works, if people agree on it and is documented/explained, I'm fine either way.

examples/seismic/acoustic/operators.py
```diff
@@ -101,9 +101,9 @@ def AdjointOperator(model, source, receiver, space_order=4,
 
     v = TimeFunction(name='v', grid=model.grid, save=None,
                      time_order=2, space_order=space_order)
-    srca = PointSource(name='srca', grid=model.grid, ntime=source.nt,
+    srca = PointSource(name='srca', grid=model.grid, time_range=source.time_range,
```
FabioLuporini: nitpicking, potentially for another PR: I see a recurrent pattern here. You could probably pass the ``PointData`` directly to the ``PointSource`` (or ``Receiver``), rather than its fields, and then fetch ``time_range, npoint, ...`` once and for all internally

examples/seismic/source.py
FabioLuporini: why would you get an exception here? and we should do more than an error message (perhaps by raising a ``ValueError`` exception?

examples/seismic/source.py
```diff
@@ -1,60 +1,105 @@
...
-    :param name: Name of the symbol representing this source
+    :param name: Name of the symbol representing this source.
     :param grid: :class:`Grid` object defining the computational domain.
-    :param coordinates: Point coordinates for this source
-    :param data: (Optional) Data values to initialise point data
-    :param ntime: (Optional) Number of timesteps for which to allocate data
-    :param npoint: (Optional) Number of sparse points represented by this source
-    :param time_order: (Optional) Time discretization order (defaults to 2)
+    :param coordinates: Point coordinates for this source.
+    :param time_range: TimeAxis(start, step, num) object.
```
FabioLuporini: :class:`TimeAxis`(...) is what Sphinx likes

examples/seismic/source.py
```diff
@@ -1,60 +1,105 @@
...
         time_order = kwargs.get('time_order', 2)
         npoint = npoint or coordinates.shape[0]
-        if data is None:
-            if ntime is None:
-                error('Either data or ntime are required to'
-                      'initialise source/receiver objects')
-        else:
-            ntime = ntime or data.shape[0]
+
+        cls._time_range = TimeAxis(time_range.start, time_range.step, time_range.num)
```
FabioLuporini: are you just rebuilding the ``time_range``? if so, perhaps it's worth adding a copy/rebuild method to ``TimeAxis`` itself. Also, u shouldn't probably attach this to ``cls``, but rather to ``obj`` directly

tests/test_resample.py
```diff
@@ -0,0 +1,35 @@
+import numpy as np
+
+from examples.seismic import TimeAxis, RickerSource, demo_model
+
+
+def test_resample():
```
FabioLuporini: docstring

FabioLuporini: in the long term, should we expect more tests in this ``test_resample.py`` file?

ggorman:
> Concerning 3, as this PR introduce time resampling, adding an extra time step doesn't create any problem, ceil will always work. What I like about 3, is that it can be just reset with a new CFL and everything folows without any user recomputation.
> SO yes, 1 or 3 works, if people agree on it and is documented/explained, I'm fine either way.

I've added stop as an optional keyword argument in d58288a9. Also added documentation explaining its use.

examples/seismic/acoustic/operators.py
ggorman: I don't follow - can you elaborate?

ggorman:
> why would you get an exception here? and we should do more than an error message (perhaps by raising a ValueError exception?

Rewritten in d58288a9

examples/seismic/source.py
ggorman: Fixed in 3cd5c73a

tests/test_resample.py
ggorman: Fixed in d58288a

ggorman: yes

examples/seismic/source.py
ggorman: fixed in e2024c50

examples/seismic/acoustic/operators.py
FabioLuporini: multiple ``source``'s fields are used and passed to this ``PointSource``. So you could pass directly ``source``, rather than its field. In such a case, the ``PointSource``'s constructor would have to check what it's been passed (if we tolerate both options -- either a single ``PointData`` OR its fields separately). If we do that, all of the call sites become slightly smaller. Also, this scales well if, in the future, we plan to add additional fields to ``PointData`` that will also be needed by ``PointSource`` (or ``Receiver`` or whatever)

But as I said, this is just nitpicking -- you can ignore this.

To make it short, it's a stupid thing only to pass one single parameter rather than N (N=2 in this case)

ggorman: Understood

examples/seismic/source.py
```diff
@@ -1,60 +1,116 @@
...
+    :param start: Start of time axis.
+    :param step: Time step. This should be set to None if stop is set.
+    :param: num: Number of times steps.
+    :stop: (Optional) End time.
+    """
+    def __init__(self, start, step, num, stop=None):
+        if stop is not None:
+            if step is not None:
+                raise ValueError("step should be set to None if stop is set")
+            step = (stop-start)/num
```
mloubout: `num-1`, num step is `num-1` intervals.

examples/seismic/source.py
```diff
@@ -1,60 +1,116 @@
...
+
+        self.start = start
+        self.stop = start + step*num
```
mloubout: `num-1`

examples/seismic/source.py
```diff
@@ -1,60 +1,116 @@
...
+        self.step = step
+        self.num = num
+
+    def __str__(self):
+        return "TimeAxis: start=%g, stop=%g, step=%g, num=%g" % \
+               (self.start, self.stop, self.step, self.num)
+
+    @cached_property
+    def time_values(self):
+        return np.linspace(self.start, self.stop, self.num, endpoint=False)
```
mloubout: I still would put `endpoint=True` to be consistent with the previous computations of `start/end/step`

tests/test_resample.py
```diff
@@ -0,0 +1,35 @@
...
+
+    time_range_f = TimeAxis(t0, None, 10*nt, stop=time_range.stop)
+    src_b = RickerSource(name='src_b', grid=model.grid, f0=f0, time_range=time_range_f)
+
+    src_c = src_b.resample(dt=model.critical_dt)
+
+    assert src_a.data.shape == src_c.data.shape
+
+    res = src_a.data - src_c.data
+    assert np.linalg.norm(res) < 0.005
```
mloubout: Relative error are better usually for comparison:
`np.isclose( np.linalg.norm(res)/np.linalg.norm(src_a.data), 1.0, rtol=1e-6)`

.travis.yml
```diff
@@ -109,8 +109,8 @@ script:
   - if [[ $DEVITO_BACKEND != 'yask' ]]; then python examples/checkpointing/checkpointing_example.py; fi
 
   # Test tutorial notebooks for the website using nbval
-  - if [[ $DEVITO_BACKEND != 'yask' ]]; then py.test -vs --nbval examples/seismic/tutorials; fi
-  - if [[ $DEVITO_BACKEND != 'yask' ]]; then py.test -vs --nbval examples/cfd; fi
+  - if [[ $DEVITO_BACKEND != 'yask' ]]; then DEVITO_OPENMP=1 OMP_NUM_THREADS=4 py.test -vs --nbval examples/cfd; fi
```
FabioLuporini: I don't think we should enforce OPENMP in **all** builds by exporting DEVITO_OPENMP locally

examples/seismic/source.py
ggorman: Hmm - no. Violates 'single source of truth' philosophy. scipy.signal already has excellent documentation. Readers should go to the source documentation rather than relying on us cutting and pasting.

mloubout: >Readers should go to the source documentation rather than relying on us cutting and pasting.

That is a good point and I agree, but it does still need to be linked or referenced as all people see in the doc is the function signature without knowing it is the input for `resample`.

examples/seismic/source.py
```diff
@@ -1,60 +1,116 @@
...
+        return self._time_range.time_values
+
+    @property
+    def time_range(self):
+        return self._time_range
+
+    def resample(self, dt, window=None):
+        # The new number of sample points and roundoff adjusted dt
+        nt = int(round((self._time_range.stop-self._time_range.start)/dt))
+        dt = (self._time_range.stop-self._time_range.start)/nt
```
mloubout: `/nt-1`, 'dt` is the size of the interval. for `nt` points there is `nt-1` intervals.

.travis.yml
ggorman: Was only trying to get around travis issues. Revered in e5d17f24

examples/seismic/source.py
```diff
@@ -1,60 +1,144 @@
...
+    def __init__(self, start=None, step=None, num=None, stop=None):
+        try:
+            if start is None:
+                start = step*(1 - num) + stop
+            elif step is None:
+                step = (stop - start)/(num - 1)
+            elif num is None:
+                num = int(np.ceil((stop - start + step)/step))
+                stop = step*(num - 1) + start
+            elif stop is None:
```
mloubout: Shouldn't these be all seperate `if`. In the case `start` is None, the rest is never checked

examples/seismic/source.py
```diff
@@ -1,60 +1,144 @@
...
     import matplotlib.pyplot as plt
 except:
     plt = None
 
-__all__ = ['PointSource', 'Receiver', 'Shot', 'RickerSource', 'GaborSource']
+__all__ = ['PointSource', 'Receiver', 'Shot', 'WaveletSource',
+           'RickerSource', 'GaborSource', 'TimeAxis']
+
+
+class TimeAxis:
```
FabioLuporini: new style classes:
```
class TimeAxis(object):
```

tests/test_yask.py
```diff
@@ -452,8 +453,9 @@ def test_acoustic_w_src_w_rec(self, model, eqn, m, damp, u, src, rec):
 
         # The expected norms have been computed "by hand" looking at the output
         # of test_adjointA's forward operator w/o using the YASK backend.
-        exp_u = 153.25
-        exp_rec = 212.03
+        exp_u = 159.94
```
mloubout: No idea why these values changed in this PR but it did.

FabioLuporini: anything to do with resampling?

mloubout: Nothing is resampled here. Something changed somewhere but can't find out what

mloubout:
Damn, TRAVIS back to time out

mloubout:
I prefer the previous case that don't change start/end

examples/seismic/source.py
```diff
@@ -1,60 +1,153 @@
...
 
         # Create the underlying SparseTimeFunction object
         obj = SparseTimeFunction.__new__(cls, name=name, grid=grid,
                                          dimensions=[grid.time_dim, p_dim],
-                                         npoint=npoint, nt=ntime, time_order=time_order,
+                                         npoint=npoint, nt=time_range.num,
+                                         time_order=time_order,
                                          coordinates=coordinates, **kwargs)
 
+        obj._time_range = copy(time_range)
```
FabioLuporini: nitpicking: I'd prefer a ``_reconstruct`` method, because that way you have full control on what is being copied and what not. It'd also be closer to what I do throughout Devito. See for example https://github.com/opesci/devito/blob/master/devito/ir/support/space.py#L124

examples/seismic/source.py
```diff
@@ -1,60 +1,153 @@
...
 
+    @cached_property
+    def time_values(self):
+        return self._time_range.time_values
+
+    @property
+    def time_range(self):
+        return self._time_range
+
+    def resample(self, dt, rtol=1e-5):
```
FabioLuporini: do we need a comment here, or can you reference some existing documentation with a link?

tests/test_yask.py
FabioLuporini: that's bad. We should find out

examples/seismic/source.py
ggorman: fixed in ca4d5203

examples/seismic/source.py
```diff
@@ -1,60 +1,165 @@
...
 
+    @cached_property
+    def time_values(self):
+        return self._time_range.time_values
+
+    @property
+    def time_range(self):
+        return self._time_range
+
+    def resample(self, dt=None, num=None, rtol=1e-5):
```
FabioLuporini: I'm merging, but this deserves a docstring in the next PR

tests/test_resample.py
```diff
@@ -0,0 +1,51 @@
+import numpy as np
+from examples.seismic import TimeAxis, RickerSource, demo_model
+
+
+def test_resample():
```
FabioLuporini: same, docstring. What does this test do

tests/test_yask.py
```diff
@@ -385,20 +384,24 @@ def eqn(self, m, damp, u, kernel):
         return iso_stencil(u, m, t.spacing, damp, kernel)
 
     @pytest.fixture
-    def src(self, model, time_params, dtype):
-        time_values = np.linspace(*time_params)  # Discretized time axis
+    def src(self, model, dtype):
```
FabioLuporini: you can pass fixtures as method arguments

tests/test_yask.py
```diff
@@ -385,20 +384,24 @@ def eqn(self, m, damp, u, kernel):
...
         src.coordinates.data[0, -1] = 30.
         return src
 
     @pytest.fixture
-    def rec(self, model, time_params, src, dtype):
+    def rec(self, model, src, dtype):
         nrec = 130  # Number of receivers
-        t0, tn, nt = time_params
-        rec = Receiver(name='rec', grid=model.grid, ntime=nt, npoint=nrec, dtype=dtype)
+        t0, tn, dt = self.time_params(model)
```
FabioLuporini: same, just like we did for ``model, src, dtype``

tests/test_yask.py
FabioLuporini: might it be because the "number of timesteps" calculation has changed (in TimeAxis)?

Anyway, I get we are calculating the right thing, so that's fine

FabioLuporini:
Merged.

PR #527, opened 2018-03-30
Enabled parallel plugin for py.test.

ggorman:


FabioLuporini:
I doubt we get exclusive access to these 16-core nodes, hence this might actually degrade performance...

PR #528, opened 2018-04-05
better arguments processing

FabioLuporini:
I think this PR makes understanding and changing the arguments processing **way** easier than in current master

### DIFF
--- a/devito/function.py
+++ b/devito/function.py
@@ -63,28 +63,32 @@ class Constant(AbstractCachedSymbol):
         """Return a tuple of argument names introduced by this symbol."""
         return (self.name,)
 
-    def _arg_defaults(self):
+    def _arg_defaults(self, alias=None):
         """
         Returns a map of default argument values defined by this symbol.
         """
-        return {self.name: self.data}
+        key = alias or self
+        return {key.name: self.data}
 
     def _arg_values(self, **kwargs):
         """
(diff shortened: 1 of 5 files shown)


devito/function.py
```diff
@@ -293,33 +297,35 @@ def _arg_defaults(self, alias=None):
...
-
-        # Add value override for own data if it is provided
+        # Add value override for own data if it is provided, otherwise
+        # use defaults
         if self.name in kwargs:
             new = kwargs.pop(self.name)
             if isinstance(new, TensorFunction):
                 # Set new values and re-derive defaults
-                values.update(new._arg_defaults(alias=key).reduce_all())
+                values = new._arg_defaults(alias=self).reduce_all()
```
mloubout: `alias=self`? shouldn't it stay `key`?

devito/function.py
```diff
@@ -293,33 +297,35 @@ def _arg_defaults(self, alias=None):
...
             else:
                 # We've been provided a pure-data replacement (array)
-                values[key.name] = new
+                values = {self.name: new}
```
mloubout: Same, no more key?

devito/function.py
mloubout: ok

FabioLuporini:
Closing and opening a new one to see if jenkins disappears

PR #529, opened 2018-04-05
Small updates to docs

navjotk:
Change opesci.org to devitoproject.org throughout. Change `PERFORMANCE.md` to reflect the fact that 320 is a devito bug, not a GCC bug.


FabioLuporini:
Merged.

PR #530, opened 2018-04-05
Parallel loops with local reductions

FabioLuporini:
This PR generates ``#pragma atomic`` over associative and commutative increments so that loops with local reductions can be safely parallelised. 

For seismic, the typical use case is injection loops

EDIT: this is based on #528 ...

### DIFF
--- a/devito/cgen_utils.py
+++ b/devito/cgen_utils.py
@@ -1,5 +1,6 @@
 from collections import OrderedDict
 
+import numpy as np
 import cgen as c
 from mpmath.libmp import prec_to_dps, to_str
 from sympy import Function
@@ -58,7 +59,7 @@ class Allocator(object):
 
 class CodePrinter(C99CodePrinter):
 
-    custom_functions = {'INT': '(int)', 'FLOAT': '(float)'}
+    custom_functions = {'INT': '(int)', 'FLOAT': '(float)', 'DOUBLE': '(double)'}
 
     """Decorator for sympy.printing.ccode.CCodePrinter.
 
@@ -171,3 +172,6 @@ printmark = lambda i: c.Line('printf("Here: %s\\n"); fflush(stdout);' % i)
 printvar = lambda i: c.Statement('printf("%s=%%s\\n", %s); fflush(stdout);' % (i, i))
 INT = Function('INT')
 FLOAT = Function('FLOAT')
+DOUBLE = Function('DOUBLE')
+
+cast_mapper = {np.float32: FLOAT, float: DOUBLE, np.float64: DOUBLE}
--- a/devito/dle/backends/utils.py
+++ b/devito/dle/backends/utils.py
@@ -12,7 +12,8 @@ omplang = {
     'par-region': lambda i: c.Pragma('omp parallel %s' % i),
     'par-for': c.Pragma('omp parallel for schedule(static)'),
     'simd-for': c.Pragma('omp simd'),
-    'simd-for-aligned': lambda i, j: c.Pragma('omp simd aligned(%s:%d)' % (i, j))
+    'simd-for-aligned': lambda i, j: c.Pragma('omp simd aligned(%s:%d)' % (i, j)),
+    'atomic': c.Pragma('omp atomic update')
 }
 
 """
--- a/devito/ir/clusters/cluster.py
+++ b/devito/ir/clusters/cluster.py
@@ -1,3 +1,4 @@
+import numpy as np
 from cached_property import cached_property
 from frozendict import frozendict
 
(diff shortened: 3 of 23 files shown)


devito/cgen_utils.py
```diff
@@ -171,3 +172,6 @@ def ccode(expr, **settings):
 printvar = lambda i: c.Statement('printf("%s=%%s\\n", %s); fflush(stdout);' % (i, i))
 INT = Function('INT')
 FLOAT = Function('FLOAT')
+DOUBLE = Function('DOUBLE')
+
+cast_mapper = {np.float32: FLOAT, float: DOUBLE, np.float64: DOUBLE}
```
mloubout: Now that I see that, would it be a good time to drop the hard coded ` +F` for constants?

mloubout:
Looks like TRAVIS is back to happy with adjoint tests. I'll have a go at it.

FabioLuporini:
> Looks like TRAVIS is back to happy with adjoint tests. I'll have a go at it.

I fixed it. now generating proper data type, data casts etc...so no loss in precision !

devito/cgen_utils.py
FabioLuporini: yes, however I **think** that if you're doing ``float*double``, then the ``float`` gets automatically promoted to ``double``. But I should double check this.

devito/dle/backends/utils.py
```diff
@@ -12,7 +12,8 @@
     'par-region': lambda i: c.Pragma('omp parallel %s' % i),
     'par-for': c.Pragma('omp parallel for schedule(static)'),
     'simd-for': c.Pragma('omp simd'),
-    'simd-for-aligned': lambda i, j: c.Pragma('omp simd aligned(%s:%d)' % (i, j))
+    'simd-for-aligned': lambda i, j: c.Pragma('omp simd aligned(%s:%d)' % (i, j)),
```
mloubout: Is this now consistent with the `__attribute__((aligned`?

devito/ir/clusters/cluster.py
```diff
@@ -69,6 +70,26 @@ def unknown(self):
...
+    @property
+    def dtype(self):
+        """Return the arithmetic data type of this Cluster. If the Cluster is
+        performing floating point arithmetic, then any equation performing
+        integer arithmetic is ignored, assuming that they are only carrying
+        out array index calculations. If two equations are performing floating
+        point calculations with mixed precision, return the data type with
+        highest precision."""
+        dtypes = {i.dtype for i in self.exprs}
+        fdtypes = {i for i in dtypes if np.issubdtype(i, np.float)}
```
mloubout: `FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.

This warning pollutes the log quite a lot.

devito/ir/clusters/cluster.py
```diff
@@ -170,3 +191,28 @@ def finalize(self):
...
+
+    @property
+    def dtype(self):
+        """Return the arithmetic data type of this ClusterGroup. If at least one
+        Cluster is performing floating point arithmetic, then any Cluster performing
+        integer arithmetic is ignored. If two Clusters are performing floating
+        point calculations with different precision, return the data type with
+        highest precision."""
+        dtypes = {i.dtype for i in self}
+        fdtypes = {i for i in dtypes if np.issubdtype(i, np.float)}
```
mloubout: same

tests/test_dle.py
```diff
@@ -413,49 +399,6 @@ def test_loops_ompized(fa, fb, fc, fd, t0, t1, t2, t3, exprs, expected, iters):
...
-      }
-      for (int k = 0; k <= 7; k += 1)
-      {
-        b[i] = a[i] + pow(b[i], 2) + 3;
-      }
-    }
-  }""" in str(handle.nodes)
-    Rewriter.thresholds['min_fission'], Rewriter.thresholds['max_fission'] = old
-
-
```
mloubout: Any reason for removing tests?

devito/dle/backends/utils.py
FabioLuporini: they are different things. With ``__attribute__``, you're **enforcing** data alignment; with ``omp simd aligned`` you're just telling the compiler that "all datasets in this loop are aligned". While the former are necessary, the latter may actually be not, as long as the compiler is able to retrieve the information by itself (which is probably our case, but just to be sure, we put the ``aligned`` anyway)

devito/ir/clusters/cluster.py
FabioLuporini: this warning must disappear. Can you reproduce it on your laptop/workstation? I cannot on my machines...

tests/test_dle.py
FabioLuporini: In this PR, I've dropped obsolete functionalities, among which a DLE pass called ``loop fission``. Here, I'm just dropping its tests

devito/ir/clusters/cluster.py
FabioLuporini: right, this was due to NumPy 14, ,while I had 13.3 on my machine. Now it should be fixed

devito/ir/clusters/cluster.py
FabioLuporini: should be fixed as above

FabioLuporini:
Merged.

PR #535, opened 2018-04-10
dle: check loop is perfectly nested before adding pragma omp collapse

tjb900:
Sometimes the compiler would generate code such as:

```
    #pragma omp parallel
    {
      #pragma omp for collapse(2) schedule(static)
      {
        for (int y = y_M; y >= y_m; y -= 1)
        {
          #pragma ivdep
          #pragma omp simd
          {
            u[time + 1][x + 1][y + 1][z + 1] = u[time][x][y + 1][z + 1] + u[time][x + 1][y][z + 1] + u[time][x + 1][y + 1][z + 1] + 2*u[time][x + 1][y + 1][z + 2] + u[time][x + 1][y + 2][z + 1] + u[time][x + 2][y + 1][z + 1] - u[time - 1][x + 1][y + 1][z + 1];
          }
        }
        for (int y_left = y_left_M; y_left >= y_left_m; y_left -= 1)
        {
          {
            u[time + 1][x + 1][y_left + 1][z + 1] = u[time][x + 1][y_left + 1][z + 1] + u[time][x + 1][y_left + 2][z + 1] + u[time + 1][x + 1][y_left + 2][z + 1];
          }
        }
    ...
  ```

to which Intel responds:

```
Error: parallel loops with collapse must be perfectly nested
```

This appears to fix it.  It still needs a test to be added.

### CODE REVIEW

**Comment by tjb900** [comment]
@FabioLuporini I've just added a test based on the other `ompize` test, but I have not yet figured out how to rig up the test expressions to create an imperfect loop structure which should not be collapsed.

**Comment by FabioLuporini** [comment]
How about expressions that share the same outer loops, but have different inner dimensions, as in (pseudocode)
```
a = Function(x, y)
b = Function(x, z)
c = Function(x, w)
eqs = [Eq(a, a + 1), Eq(b, b + 1), Eq(c, c+1)]
```
?

FabioLuporini:
also note that #539 is moving some DLE code around, and your ``IsPerfectIteration`` fix should be moved here:
https://github.com/opesci/devito/pull/539/files#diff-b984e6bbf4e175e14badc6680e489948R45

It's a trivial change. I'm just saying it in case we merge the other PR first (but, TBH, I don't have a preference for the order!)

tjb900:
Yes I just saw that :-) will update here accordingly once #539 lands.

PR #536, opened 2018-04-10
pip: should install all deps

mloubout:
Update `setup.py` for pip, packages are now directly obtained from `requirements.txt` to avoid duplication and missing entries. 

Don't know if it actually work yet


mloubout:
The pypi requirements should be good.
For the url ones, no idea, still trying to figure out how to test it correctly.

FabioLuporini:
@mloubout let me know if this works as expected

mloubout:
This is getting on my nerves

mloubout:
So should I remove PyRevolve and just add ti to the README?

FabioLuporini:
> So should I remove PyRevolve and just add ti to the README?

I think @navjotk said he was going to make PyRevolve pip-installable. If it's just a matter of days, I think it's worth waiting

mloubout:
No idea why TRAVIS not starting, but should be working correctly now.

README.md
```diff
@@ -38,6 +38,11 @@ conda env create -f environment.yml
 source activate devito
 pip install -e .
 ```
+or alternatively install Devito only (without tests and examples) using `pip`:
```
FabioLuporini: I don't wanna claim we support this, though "we're trying our best". Conda should be **the** way. So I'd drop this whole paragraph

requirements.txt
```diff
@@ -2,9 +2,9 @@ numpy>=1.11
 sympy==1.1
 scipy
 matplotlib
-pytest
+pytest-runner
```
FabioLuporini: ?

.travis.yml
```diff
@@ -95,35 +99,40 @@ before_script:
 
 script:
   - flake8 --builtins=ArgumentError .
-  - py.test -vs --cov devito tests/
+  - if [ "$INSTALL_TYPE" == 'setup' ]; then
+      - python setup.py test
```
FabioLuporini: what should happen here?

setup.py
```diff
@@ -1,8 +1,18 @@
...
+
+from setuptools import setup, find_packages
+
+with open('requirements.txt') as f:
+    required = f.read().splitlines()
+
+reqs = []
+links = []
+for ir in required:
+    if ir[0:3] == 'git':
```
FabioLuporini: nitpicking: you can probably use `startswith`

requirements.txt
mloubout: For `python setup.py test` needs standalone pytest

.travis.yml
mloubout: run the tests

FabioLuporini: so are the tests installed?

.travis.yml
```diff
@@ -95,35 +100,41 @@ before_script:
...
-  - if [[ $DEVITO_BACKEND != 'yask' ]]; then DEVITO_BACKEND=foreign py.test -vs tests/test_operator.py -k TestForeign; fi
-  - if [[ $DEVITO_BACKEND != 'yask' ]]; then python examples/seismic/benchmark.py test -P tti -so 4 -a -d 20 20 20 -n 5; fi
-  - if [[ $DEVITO_BACKEND != 'yask' ]]; then python examples/seismic/benchmark.py test -P acoustic -a; fi
-  - if [[ $DEVITO_BACKEND != 'yask' ]]; then python examples/seismic/acoustic/acoustic_example.py --full; fi
-  - if [[ $DEVITO_BACKEND != 'yask' ]]; then python examples/seismic/acoustic/acoustic_example.py --constant --full; fi
-  - if [[ $DEVITO_BACKEND != 'yask' ]]; then python examples/seismic/acoustic/gradient_example.py; fi
-  - if [[ $DEVITO_BACKEND != 'yask' ]]; then python examples/misc/linalg.py mat-vec mat-mat-sum transpose-mat-vec; fi
-  - if [[ $DEVITO_BACKEND != 'yask' ]]; then python examples/seismic/tti/tti_example.py -a; fi
-  - if [[ $DEVITO_BACKEND != 'yask' ]]; then python examples/checkpointing/checkpointing_example.py; fi
+    # Additional seismic operator tests
```
FabioLuporini: indentation

README.md
FabioLuporini: I'm still against this

requirements.txt
```diff
@@ -2,9 +2,9 @@ numpy>=1.11
...
 flake8>=2.1.0
-parso==0.1.0
+jedi
```
FabioLuporini: ???

mloubout: `jedi` gets install by something else and creates conflicts with `parso==0.1.0` as it already installs it with a different version

FabioLuporini: I didn't get it sorry. We're dropping ``parso`` here correct? who's using ``jedi``? if you don't install it, what does it fail?

mloubout: >who's using jedi?
It's a requirement for one of the other requirements, not sure which

README.md
mloubout: I don't agree, and I don't think that's what we agreed on Monday.

requirements.txt
mloubout: `error: parso 0.1.0 is installed but parso>=0.2.0 is required by {'jedi'}`

Still no idea where this `jedi` is coming from but crashing the install.

README.md
mloubout: You are assuming that people should only ever use Devito as a standalone. We have a python module that people may want to integrate with different work. 
The same way we don't have to create a *virtualenv* for every requirement we have, people should be able to install devito as a requirement in their own project and therefore Devito has to be pip installable.
We have a pythin package, and no matter how we use it, if we want it to survive in the Python enviromnent it needs to be pip installable like all are requirements are.

FabioLuporini: I think you're not getting my point here. You're absolutely right, that's why I supported this PR since the very beginning, you probably remember that. But here I'm talking about **our** installation instructions, not how some other packages pretend to use or install Devito (that's their job).

mloubout: Fine with me then. This is now tested and requirement automatically reads from the `requirement.txt` so we should be fine as of now. Removed the extra line from the README.

requirements.txt
FabioLuporini: ahhhhhh! Then what I don't understand is why you dropped ``parso==0.1.0``? I'm concerned we are installing ``jedi`` only to indirectly get ``parso``. Ie. we're not using jedi at all, but only one of its dependencies

mloubout: I'm still trying to figure out where this `jedi` is coming from. These kind of dependencies never show up with `requirement.txt` as it installs all requires separately while `pip` creates an `easy_install` that checks all cross-dependnecies

FabioLuporini: if you put back ``parso==0.1.0`` and remove ``jedi``, does it work?

mloubout:
This should be GTG yes

PR #539, opened 2018-04-11
Give YASK more freedom to drive the DLE

### LINKED ISSUE #540: receiver interpolation moved to parallel time loop
On current master (04304cd4)

```
from devito import (Grid, Operator, TimeFunction, configuration, Eq,
                    ConditionalDimension, SparseTimeFunction, Constant)

configuration["openmp"] = True

grid = Grid(shape=(20, 20, 20), extent=(19, 19, 19))
x, y, z = grid.dimensions
time = grid.time_dim
t = grid.stepping_dim
tsave = ConditionalDimension(name='tsave', parent=time, factor=2)

u = TimeFunction(name='u', grid=grid, save=None, time_order=2)
usave = TimeFunction(name='usave', grid=grid, time_dim=tsave,
                     time_order=0, space_order=0)

r = SparseTimeFunction(name='r', grid=grid, npoint=1, nt=1000)

steps = []

# save of snapshot
steps.append(Eq(usave, u))

# standard laplace-like - not sure how much it's necessary
steps.append(Eq(u[t+1, x, y, z], u[t, x, y, z]
                - u[t-1, x, y, z]
                + u[t, x-1, y, z] + u[t, x+1, y, z]
                + u[t, x, y-1, z] + u[t, x, y+1, z]
                + u[t, x, y, z-1] + u[t, x, y, z+1]))

steps.extend(r.interpolate(expr=u))

op = Operator(steps)

print(op.ccode)

u.data[:] = 0.0
u.data[0, 10, 10, 10] = 1.0

r.coordinates.data[0, :] = 10

op.apply(time_m=0, time_M=0)

# not sure about these - my main assertion is that we should get to here :-)
assert np.sum(u.data[0, :, :, :]) == 1.0
assert np.sum(u.data[1, :, :, :]) == 7.0                                                                                                                                                                                                                                                                                           assert np.all(usave.data[0, :, :, :] == u.data[0, :, :, :])
assert r.data[0, 0] == 1.0
```

Generates error:

```
2318b36a7ec402b021a2be17e275d0fc18aeba9a.c(130): error: parallel loop increment operator must be one of ++, --, +=, or -=
      for (int time = time_m, t0 = (time)%(1); time <= time_M; time += 1, t0 = (time)%(1))
```

As discussed, the big error is not the loop increment syntax, but the fact that the entire receiver sampling loop is separate from the main time stepping (i.e. there is more than one time loop).

Thanks again!

FabioLuporini:
Gonna be useful soon. In essence, in the YASK backend we can now specialize individual DLE passes or provide entirely new Rewriters.

Also:

fixes #540

### DIFF
--- a/devito/dle/backends/advanced.py
+++ b/devito/dle/backends/advanced.py
@@ -7,30 +7,30 @@ from itertools import combinations
 
 import cgen
 import numpy as np
-import psutil
 
 from devito.cgen_utils import ccode
 from devito.dimension import Dimension
 from devito.dle import fold_blockable_tree, unfold_blocked_tree
-from devito.dle.backends import (BasicRewriter, BlockingArg, dle_pass, omplang,
+from devito.dle.backends import (BasicRewriter, BlockingArg, Ompizer, dle_pass,
                                  simdinfo, get_simd_flag, get_simd_items)
 from devito.exceptions import DLEException
-from devito.ir.iet import (Block, Expression, Iteration, List,
-                           PARALLEL, ELEMENTAL, REMAINDER, tagger,
-                           FindNodes, FindSymbols, IsPerfectIteration, Transformer,
-                           compose_nodes, retrieve_iteration_tree, filter_iterations)
+from devito.ir.iet import (Iteration, List, PARALLEL, ELEMENTAL, REMAINDER, tagger,
+                           FindSymbols, IsPerfectIteration, Transformer, compose_nodes,
+                           retrieve_iteration_tree)
 from devito.logger import dle_warning
 from devito.tools import as_tuple
 
 
 class AdvancedRewriter(BasicRewriter):
 
+    _parallelizer = Ompizer
+
     def _pipeline(self, state):
         self._avoid_denormals(state)
         self._loop_blocking(state)
         self._simdize(state)
         if self.params['openmp'] is True:
-            self._ompize(state)
+            self._parallelize(state)
         self._create_elemental_functions(state)
         self._minimize_remainders(state)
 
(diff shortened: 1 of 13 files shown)


devito/dle/backends/advanced.py
```diff
@@ -196,82 +196,25 @@ def _simdize(self, nodes, state):
...
+                    simd = as_tuple(Ompizer.lang['simd-for'])
                 mapper[i] = i._rebuild(pragmas=i.pragmas + ignore_deps + simd)
 
         processed = Transformer(mapper).visit(nodes)
 
         return processed, {}
 
     @dle_pass
-    def _ompize(self, nodes, state):
+    def _parallelize(self, iet, state):
```
FabioLuporini: note for reviewers: nothing really changed below, I've just moved (and split over multiple methods) this code to ``parallelizer.py``. This gives YASK (and future backends) the possibility of specializing parallelization rules and strategies

devito/ir/support/utils.py
```diff
@@ -209,12 +209,15 @@ def group_expressions(exprs):
...
         if e1 in mapper:
+            # Optimization: we know already that a group for `e1` has been found
             continue
         found = False
-        for e2 in exprs[i+1:]:
-            if Scope([e1, e2]).has_dep:
+        for e2 in exprs:
+            if e1 is e2:
+                continue
+            elif Scope([e1, e2]).has_dep:
```
navjotk: Seems like a symmetric relationship to me. If yes, isn't the old way of starting from `i+1` better?

FabioLuporini: that's what I thought originally (transitive in particular), but no because I'm updating ``mapper`` ASAP, so you may have
A -> B so both A and B are in mapper
then if you also have C and A -> C, you need to check A again

FabioLuporini:
Merged.

PR #541, opened 2018-04-16
handle case where ConditionalDimension has no action on one branch

tjb900:
Fixes this error, encountered in some cases when using `ConditionalDimension`s:

```
ValueError: Unsupported Cluster [mixed integer arithmetic ?]
```

### DIFF
--- a/devito/ir/clusters/algorithms.py
+++ b/devito/ir/clusters/algorithms.py
@@ -97,6 +97,13 @@ def guard(clusters):
             banned = flatten(v for k, v in mapper.items() if k not in dims)
             exprs = [e.xreplace({i: IntDiv(i.parent, i.factor) for i in mapper})
                      for e in c.exprs if e not in banned]
+
+            if not exprs:
+                # This prevents us from generating an empty cluster
+                # when there are no active expressions over the same iteration
+                # variables when the conditional is false
+                continue
+
             guards = [(i.parent, conditions[i]) for i in dims]
             guards.extend([(i.parent, negated[i]) for i in ndims])
             cluster = PartialCluster(exprs, c.ispace, c.dspace, c.atomics, dict(guards))
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -215,3 +215,16 @@ def EVAL(exprs, *args):
     for i in as_tuple(exprs):
         processed.append(eval(i, globals(), scope))
     return processed[0] if isinstance(exprs, str) else processed
+
+
+def configuration_override(key, value):
+    def dec(f):
+        def wrapper(*args, **kwargs):
+            oldvalue = configuration[key]
+            configuration[key] = value
+            f(*args, **kwargs)
+            configuration[key] = oldvalue
+
+        return wrapper
+
+    return dec
--- a/tests/test_dimension.py
+++ b/tests/test_dimension.py
@@ -1,7 +1,7 @@
 import numpy as np
 
 import pytest
-from conftest import skipif_yask
+from conftest import skipif_yask, configuration_override
 
 from devito import (ConditionalDimension, Grid, TimeFunction, Eq, Operator, Constant,  # noqa
                     SubDimension, DOMAIN, INTERIOR)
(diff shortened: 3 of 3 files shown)


devito/ir/iet/visitors.py
```diff
@@ -261,11 +259,10 @@ def visit_Iteration(self, o):
             loop_inc = '%s += %s' % (o.index, o.limits[2])
 
         # Append unbounded indices, if any
-        if o.uindices:
-            uinit = ['%s = %s' % (i.index, ccode(i.start)) for i in o.uindices]
-            loop_init = c.Line(', '.join([loop_init] + uinit))
-            ustep = ['%s = %s' % (i.index, ccode(i.step)) for i in o.uindices]
-            loop_inc = c.Line(', '.join([loop_inc] + ustep))
+        uindex_init = [c.Initializer(c.Value('int', i.index), ccode(i.step))
+            for i in o.uindices]
```
FabioLuporini: perhaps flake8 will complain? probably ``for i in o.uindices`` should start under `c` in `c.Initializer...`

devito/ir/clusters/algorithms.py
```diff
@@ -97,6 +97,10 @@ def guard(clusters):
             banned = flatten(v for k, v in mapper.items() if k not in dims)
             exprs = [e.xreplace({i: IntDiv(i.parent, i.factor) for i in mapper})
                      for e in c.exprs if e not in banned]
+
+            if not exprs:
```
FabioLuporini: Is it worth adding a comment of when would this happen?

devito/ir/iet/visitors.py
tjb900: Yeah, it already did :-)  I was just fixing that one up....

tjb900:
Sorry, but I'm going to have to leave this one for a while.  There are two cases which are not yet handled:

1. collapsed omp loops, where the unbound variable declarations have to move inside the innermost loop rather than between the two
2. cases such as in `tests/test_dse.py::test_tti_rewrite_speculative`, where the unbound loop variable does not satisfy uindex.start == uindex.step

To elaborate on 2:  in the case of SteppingDimensions, the unbound variable initialiser and step are the same - e.g. `t1 = (time+1)%3`.  This is what the above change assumes, and now that I understand the code a little more I see that it's obviously not true in general.

My current thinking:

* add a pass somewhere earlier (not at the C code generation stage) which attempts to remove unbound variables from an Iteration, instead moving them inside the body and assigning them as an expression only in terms of the loop variables.
* later, when figuring out the parallelism, refuse to Ompize loops which have remaining unbound variables

FabioLuporini:
There were two main reasons I decided to bound ``uindices`` to ``Iteration``:
- It makes very clear the relationship between the unbound variable and the true dimension (i.e., the main Iteration variable)
- It preserves perfectness of loop nests

But is this PR trying to fix two different things? Or are the two related to each other?

Another thing we could do is attacking this from a different angle. We might force ``SEQUENTIAL`` Iterations during analysis if an UnboundedIndex is found. Here for example: 

https://github.com/opesci/devito/blob/master/devito/ir/iet/analysis.py#L72

I don't know, however, if this is gonna prevent ompization of loops that are currently (ie, in master) ompized.

tjb900:
> But is this PR trying to fix two different things? Or are the two related to each other?

It is trying to fix two different things - because the change described by the title broke tests by producing loops that openmp didn't like.  Now that it's apparent the fix is not trivial, I should split it into a different PR.

> There were two main reasons I decided to bound uindices to Iteration:

I agree with both of these.  But we shouldn't be afraid of generating C code that's ugly, provided it's correct and performs better, right?  Isn't that almost the point of devito? ;-)

FabioLuporini:
> I agree with both of these. But we shouldn't be afraid of generating C code that's ugly, provided it's correct and performs better, right? Isn't that almost the point of devito? ;-)

TBH, what you were generating looked even nicer than what we are generating now 😄 
But yes, definitely agreed!

FabioLuporini:
Travis good, you're happy, I'm happy. So I'm gonna merge this !

FabioLuporini:
Merged.

PR #546, opened 2018-04-18
Added pickling support for class Model.

ggorman:


examples/seismic/model.py
```diff
@@ -461,3 +462,33 @@ def vp(self, vp):
...
             self.m.data = 1 / vp**2
+
+    def __getstate__(self):
+        state = dict(self.__dict__)
+        state.pop('grid')
+        state.pop('damp')
+        state.pop('m')
+        state.pop('_vp')
+        dim = len(self.shape)
+        if dim == 1:
```
mloubout: Now that I think of it, I don't think this one is supported anywhere so far. Good reminder.

examples/seismic/model.py
```diff
@@ -461,3 +462,33 @@ def vp(self, vp):
...
+            vp = np.sqrt(1. / self.m.data[self.nbpml:-self.nbpml])
+        elif dim == 2:
+            vp = np.sqrt(1. / self.m.data[self.nbpml:-self.nbpml,
+                                          self.nbpml:-self.nbpml])
+        elif dim == 3:
+            vp = np.sqrt(1. / self.m.data[self.nbpml:-self.nbpml,
+                                          self.nbpml:-self.nbpml,
+                                          self.nbpml:-self.nbpml])
+        else:
+            raise ValueError("The programmer was an idiot...")
```
mloubout: Haha.
Question, why do you need it without PML?

examples/seismic/model.py
```diff
@@ -330,14 +334,11 @@ def __init__(self, origin, spacing, shape, space_order, vp, nbpml=20,
         self.grid = Grid(extent=extent, shape=shape_pml,
                          origin=origin, dtype=dtype)
 
-        # Create square slowness of the wave as symbol `m`
-        if isinstance(vp, np.ndarray):
+        if vp is None:
             self.m = Function(name="m", grid=self.grid, space_order=self.space_order)
+            self.m.data[:] = m
```
mloubout: `initialize_function(self.m, m, self.nbpml)`

examples/seismic/model.py
```diff
@@ -461,3 +462,33 @@ def vp(self, vp):
             initialize_function(self.m, 1 / (self.vp * self.vp), self.nbpml)
         else:
             self.m.data = 1 / vp**2
+
+    def __getstate__(self):
+        state = dict(self.__dict__)
+        state.pop('grid')
```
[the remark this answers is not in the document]
ggorman: When using pop I explicitly want to remove this key:value pair from the state object. This surgery is required to prevent more information being dumped or communicated via pickling that necessary.

examples/seismic/model.py
ggorman: That fucks things up because it adds an additional nbpml padding that I don't want.

mloubout: This is at the level of generic model initialization. I f ones decide to provide `m` instead of `vp`, this has to be done. 
This is also inconsitent with the `vp` otpion that does add the abc layer.

examples/seismic/model.py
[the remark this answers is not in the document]
FabioLuporini: is that what they claim to do in the documentation that you're referring to? Having a method called ``*get*`` that actually pops stuff? 

I think it's worth adding a comment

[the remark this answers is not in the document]
FabioLuporini: I'm dumb, or perhaps I was drunk. I thought you're calling pop on self. Don't know why. Ignore all this please.

examples/seismic/model.py
```diff
@@ -329,14 +334,16 @@ def __init__(self, origin, spacing, shape, space_order, vp, nbpml=20,
...
-        if isinstance(vp, np.ndarray):
-            self.m = Function(name="m", grid=self.grid, space_order=space_order)
+        if vp is None:
+            self.m = Function(name="m", grid=self.grid, space_order=self.space_order)
+            try:
+                # Assume m has not yet been padded.
+                initialize_function(self.m, m, self.nbpml)
+            except:
+                # In this case m has already been padded (eg unpickling)
+                self.m.data[:] = m
```
mloubout: Maybe make the try/except part of  `initialize_function` to avoid duplicates. Will also work for TTI parameters then.

examples/seismic/model.py
```diff
@@ -437,58 +438,63 @@ def critical_dt(self):
...
         return self.dtype(coeff * np.min(self.spacing) / (self.scale*np.max(self.vp)))
 
+    def get_vp(self, exclude_pml=True):
+        """Returns the velocity model
+           :param exclude_pml : excludes the pml region if true.
+        """
+        if exclude_pml:
+            dim = len(self.shape)
+            if dim == 1:
+                return np.sqrt(1. / self.m.data[self.nbpml:-self.nbpml])
```
FabioLuporini: I'm 99.9% sure there's a way of doing this whole block ``if dim == 1 ... elif ... else `` with a single numpy line, ie, slicing parametrically in the number of dimensions

examples/seismic/model.py
FabioLuporini: 👍

devito/data.py
```diff
@@ -122,6 +126,24 @@ def reset(self):
         """
         self[:] = 0.0
 
+    def __reduce_ex__(self, proto):
+        """ Pickling support."""
+        return type(self), self.__getnewargs__(), self.__getstate__()
+
+    def __getnewargs__(self):
+        return (self.shape, self.dimensions, self.dtype)
```
FabioLuporini: so pickling requires **all** objects used to initialise a given object? 

I wonder whether we could rather provide (not sure if it's a pickling convention or I'm just inventing it now) a new method in each Pickable class, namely ``__new_frompickling__``, to be called by ``__new__`` depending on the arguments in input to ``__new__`` itself (``__new__``). Just a random thought. In general, it feels weird that we have to keep the entire input only for pickling...

tests/test_pickle.py
```diff
@@ -0,0 +1,74 @@
+import numpy as np
+from examples.seismic import demo_model
+from examples.seismic.source import TimeAxis, RickerSource
+from devito.dimension import TimeDimension, SteppingDimension
+from devito.base import Constant
+
+import cloudpickle as pickle
+
+
+def test_pickle():
```
FabioLuporini: docstring

PR #547, opened 2018-04-18
WIP/RFC: use symbolic representations for lower/upper bounds of subdims

tjb900:
This allows the creation of SubDimensions whose lower bound is not referenced against the lower bound of the parent, and likewise for the upper bound.  This restriction precluded the creation of SubDimensions for left and right boundary regions, since (e.g.) neither bound of a left boundary region should depend on the upper bound of the parent.

It provides SubDimension class methods for creating SubDimensions at the left, middle and right of another dimension, respectively.

Soon:
 * demos/tests of the new classmethods
 * more expressive possibilities for Regions (e.g. products of SubDimensions)


devito/dimension.py
```diff
@@ -287,32 +287,71 @@ class SubDimension(DerivedDimension):
...
+        if lower_symbolic is None:
+            lower_symbolic = parent.symbolic_start + lower
+        if upper_symbolic is None:
+            upper_symbolic = parent.symbolic_end + upper
+
+        newobj._interval = sympy.Interval(lower_symbolic, upper_symbolic)
         return newobj
 
+    @classmethod
+    def left(cls, name, parent, thickness):
```
FabioLuporini: cool

devito/dimension.py
```diff
@@ -287,32 +287,71 @@ class SubDimension(DerivedDimension):
...
-        newobj._upper = upper
+        assert (lower is None) ^ (lower_symbolic is None)
+        assert (upper is None) ^ (upper_symbolic is None)
+
+        if lower_symbolic is None:
+            lower_symbolic = parent.symbolic_start + lower
+        if upper_symbolic is None:
+            upper_symbolic = parent.symbolic_end + upper
+
+        newobj._interval = sympy.Interval(lower_symbolic, upper_symbolic)
```
FabioLuporini: feck, I didn't know ``sympy.Interval`` existed. Otherwise, probably, I wouldn't have implemented ``ir/support/space.py:Interval``...

devito/dimension.py
```diff
@@ -287,32 +287,71 @@ class SubDimension(DerivedDimension):
...
     def _arg_values(self, args, interval, **kwargs):
         values = {}
-        values[self.min_name] = args[self.parent.min_name] + self.lower
-        values[self.max_name] = args[self.parent.max_name] + self.upper
-        values[self.size_name] = args[self.parent.size_name] - (self.lower + self.upper)
+
+        # This works for expressions involving devito.types.Scalar
+        # and sympy types, since _subs is overridden in
+        # devito.types.Scalar
+        evaluated_interval = self._interval.subs(args)
```
FabioLuporini: could you add a test case that triggers this ``subs``? I'm curious

devito/dimension.py
```diff
@@ -287,32 +287,71 @@ class SubDimension(DerivedDimension):
...
     """
 
-    def __new__(cls, name, parent, lower, upper, **kwargs):
+    def __new__(cls, name, parent, lower=None, upper=None,
+                lower_symbolic=None,
+                upper_symbolic=None, **kwargs):
         newobj = DerivedDimension.__new__(cls, name, parent, **kwargs)
-        newobj._lower = lower
-        newobj._upper = upper
+        assert (lower is None) ^ (lower_symbolic is None)
```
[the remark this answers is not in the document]
tjb900: My concern is that an integer is a valid expression, and might be desirable sometimes - as an absolute value, not as a shift vs the lower/upper bound of the parent?

I guess we could break backwards compatibility - not sure how many people are using this yet.

devito/dimension.py
tjb900: Will do.  It's covered already in test_dimension.py, but this exact functionality is not singled-out.

devito/dimension.py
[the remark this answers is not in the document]
FabioLuporini: I don't know of anyone using ``master`` working with SubDimensions. I think we're in a lucky position here

devito/profiling.py
```diff
@@ -143,16 +143,16 @@ def summary(self, arguments, dtype):
             time = max(getattr(arguments[self.name]._obj, profile.name), 10e-7)
 
             # Flops
-            itershape = [i.extent(finish=arguments[dims[i].max_name],
-                                  start=arguments[dims[i].min_name]) for i in itspace]
+            itershape = [i.extent().subs(arguments) for i in itspace]
```
FabioLuporini: great

devito/dimension.py
```diff
@@ -287,33 +287,50 @@ class SubDimension(DerivedDimension):
...
-        newobj._lower = lower
-        newobj._upper = upper
+        newobj._interval = sympy.Interval(lower, upper)
         return newobj
 
+    @classmethod
+    def left(cls, name, parent, thickness):
+        return cls(name, parent,
+                   lower=parent.symbolic_start,
+                   upper=parent.symbolic_start+thickness-1)
```
FabioLuporini: this is taking -1, but ``middle`` is not. Is this consistent/expected?

tjb900: The bounds are inclusive now, right?  So (with square brackets denoting closed set, not python notation):
left = [0,thickness-1]  (e.g. 0,1,2,3)
middle = [thickness, end-thickness] (e.g 4,5,6)
right = [end-thickness+1, end] (e.g. 7,8,9,10)

FabioLuporini: fine, right

FabioLuporini:
Happy to land this, @tjb900 . Just one last question: have you checked that the output summary produced by Devito is still like in ``master``? I'm only asking based on my experience -- we had issues in the past (broken reports) when touching this delicate piece of code (``datashape, itershape`` etc). Your changes look sane though !

tjb900:
I was thinking about this, too.  I'll verify shortly and let you know.

tjb900:
Looks good for the example from #yask yesterday; also matches between master and this branch on a couple of random tests.

FabioLuporini:
good. Does it also match on critical tests, such as ``acoustic/acoustic_example.py`` and/or ``tti/tti_example.py``?

tjb900:
TTI Before/After:

```
=========================================================================================
Section section_0<132,132,132> with OI=0.17 computed in 0.073 s [0.13 GFlops/s]
Section section_2<208,1> with OI=6.43 computed in 0.000 s [0.17 GFlops/s]
Section section_3<208,101> with OI=7.81 computed in 0.003 s [0.89 GFlops/s]
Section main<208,130,130,130> with OI=5.42 computed in 36.212 s [3.28 GFlops/s, 0.01 GPts/s]
=========================================================================================
```

```
=========================================================================================
Section section_0<132,132,132> with OI=0.17 computed in 0.073 s [0.13 GFlops/s]
Section section_2<208,1> with OI=6.43 computed in 0.000 s [0.15 GFlops/s]
Section section_3<208,101> with OI=7.81 computed in 0.003 s [0.77 GFlops/s]
Section main<208,130,130,130> with OI=5.42 computed in 35.048 s [3.39 GFlops/s, 0.01 GPts/s]
=========================================================================================
```

Good for acoustic too.

FabioLuporini:
@tjb900 excellent, thanks for bearing patience :)

Merging...

FabioLuporini:
Merged.

PR #548, opened 2018-04-18
Non indexed conditionals

mloubout:
This is built on top of #541.

This is a bit ugly and brute force so probably a better way to do it, but shows the idea. Mainly add support for conditional dimensions when not used at indices. Added a test for example.


tests/test_dimension.py
```diff
@@ -120,6 +120,29 @@ class TestConditionalDimension(object):
...
+
+        u = TimeFunction(name='u', grid=grid)
+        assert(grid.stepping_dim in u.indices)
+
+        u2 = Function(name='energy', grid=grid)
+
+        factor = 4
+        time_subsampled = ConditionalDimension('t_sub', parent=time, factor=factor)
+
+        eqns = [Eq(u.forward, u + 1), Eq(u2, u2 + u*u*time_subsampled)]
```
tjb900: So this is actually going to scale the u^2 by (time/4), right? - which is probably not your intention, based on the names of the variables?  To prevent that would require constructing an expression which always evaluated to 1, but which sympy would not simplify away - that seems kind of shaky.

I think if we want to use ConditionalDimensions this way - and some people probably do - we might need to use something akin to region= to express that requirement?

mloubout: This is a dummy test just to check it works as it should. I actually use it in a different way for our SLIM stuff.

mloubout: >use something akin to region= to express that requirement?
There is more than a few conditional cases (straight if conditoons/ time-regions/...) that need to be figured out yeah, this is just a step to make the currently supported case more robust.

tests/test_dimension.py
```diff
@@ -120,6 +120,29 @@ class TestConditionalDimension(object):
...
+        op = Operator(eqns)
+        op.apply(t_M=nt-2)
+        assert np.all(np.allclose(u.data[(nt-1) % 3], nt-1))
+        # expected result is 1024
```
FabioLuporini: expected result is 1024 but then you check for 1600?

mloubout:
Done in #551 in a baetter way, closing.

PR #552, opened 2018-04-25
Examples cleanup and extenions

mloubout:
A set of cleanup and extension of examples:

 - type check of expression for the ccode printer (`F` only for float32 now)
 - Any number of dimensions supported by the seismic examples (acoustic) with the `-nd` argument, `-nd 2` is default.
 - Checkpointing setup in seismic cleaned up. It is now just an option for the gradient (`--checkpointing` for `acoustic_example`)
`gradient_example` and `checkpointing_example` are gone, the tests are still there but reshuffled.
- more robust physical parameters processing for TTI, any combination of epsilon/delta/theta/phi is now possible.
- Slightly more stable time resampling for time traces


examples/seismic/model.py
```diff
@@ -386,6 +389,18 @@ def __init__(self, origin, spacing, shape, space_order, vp, nbpml=20,
         else:
             self.phi = 0
 
+    def physical_params(self, m, **kwargs):
+        """
+        Return all set physical parameters
+        """
+        all_phys = [self.m, self.epsilon, self.delta, self.theta, self.phi]
```
FabioLuporini: Here's what I'd rather do. 

You could define the _names_ of the functions representing physical parameters at the class level. Ie,
```
class Model(object):
    _physical_parameters = ('m', 'epsilon', ...)
```

Finally, you can change the implementation as:
```
def defined_parameters(self, **kwargs):
    known = [getattr(self, i) for i in Model._physical_parameters]
    return {i.name: kwargs.get(i.name, i) for i in known if i is not None}
```

However, I'm struggling to understand why we need this contrived logic...

devito/cgen_utils.py
```diff
@@ -81,6 +83,7 @@ def _print_Indexed(self, expr):
 
         :returns: The resulting string
         """
+        self.dtype = expr.dtype
```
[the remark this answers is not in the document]
mloubout: this one is useless shouldn't be here.

examples/seismic/model.py
```diff
@@ -386,6 +389,18 @@ def __init__(self, origin, spacing, shape, space_order, vp, nbpml=20,
         else:
             self.phi = 0
 
+    def physical_params(self, m, **kwargs):
```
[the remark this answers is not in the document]
mloubout: m has to be always here.

examples/seismic/model.py
mloubout: We need this because in practice, we dont alwyas have all the TTI parameters and current setup forces all or nothing.

FabioLuporini: > and current setup forces all or nothing.
so that's what we should fix

devito/cgen_utils.py
```diff
@@ -65,7 +66,8 @@ class CodePrinter(C99CodePrinter):
 
     :param settings: A dictionary containing relevant settings
     """
-    def __init__(self, settings={}):
+    def __init__(self, dtype=np.float32, settings={}):
+        self.dtype = dtype
```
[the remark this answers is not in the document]
mloubout: I didn't touch that, that's how it was.

examples/seismic/model.py
mloubout: And this what does. This merly check what's in the model and what should be ignore for the operator call.

examples/seismic/model.py
```diff
@@ -269,7 +269,10 @@ def damp_boundary(damp, nbpml, spacing):
     for i in range(nbpml):
         pos = np.abs((nbpml - i + 1) / float(nbpml))
         val = dampcoeff * (pos - np.sin(2*np.pi*pos)/(2*np.pi))
-        if damp.ndim == 2:
+        if damp.ndim == 1:
```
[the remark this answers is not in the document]
mloubout: I have no idea how.

examples/seismic/acoustic/acoustic_example.py
```diff
@@ -39,12 +41,13 @@ def acoustic_setup(shape=(50, 50, 50), spacing=(15.0, 15.0, 15.0),
     # Define source geometry (center of domain, just below surface)
     src = RickerSource(name='src', grid=model.grid, f0=0.01, time_range=time_range)
     src.coordinates.data[0, :] = np.array(model.domain_size) * .5
-    src.coordinates.data[0, -1] = model.origin[-1] + 2 * spacing[-1]
-
+    if len(shape) > 1:
```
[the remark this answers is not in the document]
mloubout: Yes, if not `src.coordinates.data[0,0]` is overwriten by `src.coordinates.data[0, -1]`

devito/function.py
```diff
@@ -920,11 +929,13 @@ def coefficients(self):
     @property
     def point_symbols(self):
         """Symbol for coordinate value in each dimension of the point"""
-        return sympy.symbols('px, py, pz')
+        return tuple(sympy.symbols('p%s' % d) for d in self.grid.dimensions)
```
FabioLuporini: I suggest we use our own ``Symbol`` from ``types.py``

devito/cgen_utils.py
```diff
@@ -66,6 +67,7 @@ class CodePrinter(C99CodePrinter):
     :param settings: A dictionary containing relevant settings
     """
     def __init__(self, settings={}):
+        self.dtype = settings.pop('dtype', np.float32)
```
FabioLuporini: get, not pop !

mloubout: I don't think `dtype` is a setting for `C99CodePrinter`

examples/seismic/model.py
```diff
@@ -319,6 +317,7 @@ class Model(object):
     """
     def __init__(self, origin, spacing, shape, space_order, vp, nbpml=20,
                  dtype=np.float32, epsilon=None, delta=None, theta=None, phi=None):
+        self._physical_parameters = ('m', 'epsilon', 'delta', 'theta', 'phi')
```
FabioLuporini: This shuld be a class-level field

devito/function.py
mloubout: These are not Devito object, pure sympy symbols fro substitutions. Our own Symbol are not recognized by `subs`.

examples/seismic/model.py
```diff
@@ -126,6 +126,34 @@ def demo_model(preset, **kwargs):
                      dtype=dtype, spacing=spacing, nbpml=nbpml, epsilon=epsilon,
                      delta=delta, theta=theta, phi=phi, **kwargs)
 
+    elif preset.lower() in ['layers-tti-noazimuth', 'twolayer-tti-noazimuth',
```
FabioLuporini: what's this new one?

examples/seismic/model.py
```diff
@@ -266,21 +294,19 @@ def damp_boundary(damp, nbpml, spacing):
...
-            damp.data[:, i] += val/spacing[1]
-            damp.data[:, -(i + 1)] += val/spacing[1]
-        else:
-            damp.data[i, :, :] += val/spacing[0]
-            damp.data[-(i + 1), :, :] += val/spacing[0]
-            damp.data[:, i, :] += val/spacing[1]
-            damp.data[:, -(i + 1), :] += val/spacing[1]
-            damp.data[:, :, i] += val/spacing[2]
-            damp.data[:, :, -(i + 1)] += val/spacing[2]
+    for i in range(damp.ndim):
```
FabioLuporini: cool

examples/seismic/model.py
```diff
@@ -377,14 +405,22 @@ def __init__(self, origin, spacing, shape, space_order, vp, nbpml=20,
         else:
             self.theta = 0
 
-        if phi is not None:
+        if phi is not None and self.grid.dim>2:
```
FabioLuporini: I can see flake8 complaining here

examples/seismic/model.py
```diff
@@ -377,14 +405,22 @@ def __init__(self, origin, spacing, shape, space_order, vp, nbpml=20,
...
         else:
-            self.phi = 0
+            self.phi = None
+
+    def physical_params(self, **kwargs):
+        """
+        Return all set physical parameters
+        """
+        known = [getattr(self, i) for i in self._physical_parameters]
+        return {i.name: kwargs.get(i.name, i) or i for i in known
```
FabioLuporini: is ``or i`` really necessary here?

examples/seismic/model.py
```diff
@@ -377,14 +405,22 @@ def __init__(self, origin, spacing, shape, space_order, vp, nbpml=20,
...
+                if isinstance(i, Function)}
```
FabioLuporini: It took me a moment to understand why you're checking if it's a Function. I guess it's much more obvious if you check if it's not None

devito/function.py
```diff
@@ -920,11 +929,13 @@ def coefficients(self):
...
 
     @property
     def point_increments(self):
         """Index increments in each dimension for each point symbol"""
+        if self.grid.dim == 1:
```
FabioLuporini: We should use this: https://docs.python.org/2/library/itertools.html#itertools.product

examples/seismic/acoustic/acoustic_example.py
```diff
@@ -15,7 +15,9 @@ def smooth10(vel, shape):
     nz = shape[-1]
 
     for a in range(5, nz-6):
-        if len(shape) == 2:
+        if len(shape) == 1:
```
FabioLuporini: probably there's a way of not specialising this based on the shape dim

examples/seismic/model.py
FabioLuporini: do we really need the ``self.grid.dim > 2`` check?

examples/seismic/model.py
mloubout: Yes it is. If `i` is in `kwargs` as `i=None` you would end up with `i.name:None` that you don't want.

examples/seismic/model.py
mloubout: No, want to check it is a function as some of these parameters can be constant values and should not be passed as argument to `apply`.

tests/test_adjointJ.py
```diff
@@ -60,10 +62,10 @@ def test_acousticJ(shape, space_order):
     # Adjoint test: Verify <Ax,y> matches  <x, A^Ty> closely
     term1 = np.dot(im.data.reshape(-1), dm.reshape(-1))
     term2 = linalg.norm(du.data)**2
-    info('<Ax,y>: %f, <x, A^Ty>: %f, difference: %12.12f, ratio: %f'
-         % (term1, term2, term1 - term2, term1 / term2))
+    info('<Ax,y>: %f, <x, A^Ty>: %f, relative error: %4.4e, ratio: %e'
+         % (term1, term2, 1 - term1/term2, term1 / term2))
```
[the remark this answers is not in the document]
mloubout: Same as `adjointA` should check relative error not absolute.

tests/test_adjointA.py
```diff
@@ -66,6 +69,6 @@ def test_acoustic(mkey, shape, kernel, space_order, nbpml):
     # Adjoint test: Verify <Ax,y> matches  <x, A^Ty> closely
     term1 = np.dot(srca.data.reshape(-1), solver.source.data)
     term2 = linalg.norm(rec.data) ** 2
-    info('<Ax,y>: %f, <x, A^Ty>: %f, difference: %12.12f, ratio: %f'
+    info('<Ax,y>: %f, <x, A^Ty>: %f, difference: %4.4e, ratio: %f'
```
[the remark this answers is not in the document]
mloubout: more readable like that

examples/seismic/model.py
mloubout: no azimuth angle. Only rotation in `xz` plane not in y.

examples/seismic/acoustic/gradient_example.py
```diff
@@ -146,13 +146,13 @@ def initializer(data):
...
-            error2[i] = np.absolute(.5*linalg.norm(d - self.rec_t.data)**2 - F0 - H[i]
-                                    * G)
+            error2[i] = np.absolute(.5*linalg.norm(d - self.rec_t.data)**2 - F0 -
+                                    H[i] * G)
 
         # Test slope of the  tests
         p1 = np.polyfit(np.log10(H), np.log10(error1), 1)
         p2 = np.polyfit(np.log10(H), np.log10(error2), 1)
-
+        print(p1, p2)
```
ggorman: remove print?

ggorman:
I am happy with the changes you made to resample; in so far as it's not exactly what we want but no worse than the previous solution but the code is much cleaner (at some point someone is going to have to implement a proper sinc interpolator for space/time).

Can you perhaps file these changes as a separate PR as it's not really tied to the rest of this PR?

examples/seismic/model.py
mloubout: @FabioLuporini looks like `yask` really doesn't like slice indexing.

mloubout: fixed

devito/function.py
```diff
@@ -868,71 +868,45 @@ def coefficients(self):
...
+        # 1, x1, y1, z1, x1*y1, ...
+        indices = list(powerset(indices1))
+        indices[0] = (1,)
+        point_sym = list(powerset(self.point_symbols))
+        point_sym[0] = (1,)
+        # 1, px. py, pz, px*py, ...
+        A = []
+        ref_A = [np.prod(ind) for ind in indices]
+        # Create the matrix with the same increment order as the point increment
+        for i in self.point_increments:
```
FabioLuporini: can you turn ``point_increments`` into a private variable, namely ``_point_increments``

devito/function.py
```diff
@@ -868,71 +868,45 @@ def coefficients(self):
```
FabioLuporini: I can't claim I understand what's going on in this new implementation, but it looks cool

devito/function.py
```diff
@@ -868,71 +868,45 @@ def coefficients(self):
...
         """Index increments in each dimension for each point symbol"""
-        if self.grid.dim == 2:
-            return ((0, 0), (0, 1), (1, 0), (1, 1))
-        elif self.grid.dim == 3:
-            return ((0, 0, 0), (0, 1, 0), (1, 0, 0), (0, 0, 1),
-                    (1, 1, 0), (0, 1, 1), (1, 0, 1), (1, 1, 1))
-        else:
-            raise NotImplementedError('Point increments not defined '
-                                      'for %d dimensions.' % self.grid.dim)
+        return tuple(product(range(2), repeat=self.grid.dim))
```
FabioLuporini: sweet

devito/function.py
FabioLuporini: EDIT: ah, now I see, it's a property -- fine, it should be a private property, I think...

examples/seismic/acoustic/acoustic_example.py
```diff
@@ -11,14 +11,11 @@
...
-    out = np.ones(shape, dtype=vel.dtype)
+    out = np.copy(vel)
     nz = shape[-1]
 
     for a in range(5, nz-6):
-        if len(shape) == 2:
-            out[:, a] = np.sum(vel[:, a - 5:a + 5], axis=1) / 10
-        else:
-            out[:, :, a] = np.sum(vel[:, :, a - 5:a + 5], axis=2) / 10
+            out[..., a] = np.sum(vel[..., a - 5:a + 5], axis=len(shape)-1) / 10
```
FabioLuporini: cool

examples/seismic/acoustic/acoustic_example.py
```diff
@@ -69,9 +67,9 @@ def run(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
     if constant:
         # With  a new m as Constant
         m0 = Constant(name="m", value=.25, dtype=np.float32)
-        solver.forward(save=full_run, m=m0)
+        solver.forward(save=full_run and not checkpointing, m=m0)
```
FabioLuporini: wait, why? I don't link this combined condition here

examples/seismic/acoustic/gradient_example.py
```diff
@@ -1,169 +0,0 @@
-import numpy as np
```
FabioLuporini: I'm always happy to purge the codebase, you know, but perhaps it's better to inform whoever has maintained this

anyway, as I said, I'm glad to drop redundant code

examples/seismic/acoustic/wavesolver.py
```diff
@@ -167,8 +169,25 @@ def gradient(self, rec, u, v=None, grad=None, m=None, **kwargs):
         if m is None:
             m = m or self.model.m
 
-        summary = self.op_grad().apply(rec=rec, grad=grad, v=v, u=u, m=m,
-                                       dt=kwargs.pop('dt', self.dt), **kwargs)
+        if checkpointing:
+            set_log_level('ERROR')
```
FabioLuporini: why? perhaps what we should do is to change the verbosity of the checkpointer

examples/seismic/model.py
```diff
@@ -377,14 +405,22 @@ def __init__(self, origin, spacing, shape, space_order, vp, nbpml=20,
         else:
             self.theta = 0
 
-        if phi is not None:
+        if phi is not None and self.grid.dim > 2:
```
FabioLuporini: should we raise an error/warning if we get down to this place but ``self.grid.dim`` is <= 2?

tests/test_checkpointing.py
```diff
@@ -164,57 +169,6 @@ def test_acoustic_save_and_nosave(shape=(50, 50), spacing=(15.0, 15.0), tn=500.,
     assert(np.allclose(rec.data, rec_bk))
 
 
-@silencio(log_level='WARNING')
-@skipif_yask
-@pytest.mark.parametrize('space_order', [4])
-@pytest.mark.parametrize('kernel', ['OT2'])
-@pytest.mark.parametrize('shape', [(70, 80), (50, 50, 50)])
-def test_checkpointed_vs_not_checkpointed(shape, kernel, space_order):
```
FabioLuporini: why dropping tests?

FabioLuporini: Is this now covered in ``test_gradient`` ?

examples/seismic/model.py
FabioLuporini: but can ``i`` be ``None`` if you're testing for ``isinstance(i, Function)`` ?

examples/seismic/tti/operators.py
```diff
@@ -390,7 +390,7 @@ def ForwardOperator(model, source, receiver, space_order=4,
     ang1 = sin(theta)
     ang2 = 0
     ang3 = 0
-    if len(model.shape) == 3:
+    if len(model.shape) == 3 and phi is not None:
```
FabioLuporini: should we rather raise an error otherwise? as in, if a 1D/2D problem and ``phi != None``...

tests/test_gradient.py
```diff
@@ -100,6 +154,7 @@ def test_gradientJ(shape, kernel, space_order):
     :param space_order: order of the spacial discretization scheme
     :return: assertion that the Taylor properties are satisfied
     """
+    clear_cache()
```
FabioLuporini: it's not clear why we need this

examples/seismic/acoustic/acoustic_example.py
mloubout: If you use checkpointing, you do not need to store the whole time history, so save should only be True without checkpointing.

examples/seismic/acoustic/gradient_example.py
mloubout: All the example was just duplicates of `wavesolver` and `acoustic_example` that was really a nightmare to maintain and delayed quite a few PRs in the passed.

examples/seismic/acoustic/wavesolver.py
mloubout: Yeah, checkpointing prints a huge load of summary, probably not the best way to silence it.

examples/seismic/model.py
mloubout: Probably not necessary no that I am thinking of it. I probably only need to modify the `_physical_parameter`.

tests/test_checkpointing.py
mloubout: Yes it is

examples/seismic/tti/operators.py
mloubout: No 1D for TTI, tilted doesn't mean anything in 1D, can be done that way too will look at it

tests/test_gradient.py
mloubout: TRAVIS seems to not clear the cache between test which was making `test_adjointJ` crash. Added it there too as we save the full wavefield for safety

devito/cgen_utils.py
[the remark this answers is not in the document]
ggorman: Its not clear to me that this point was resolved.

@FabioLuporini - did you mean:
def __init__(self, settings=None):
    self.dtype = getattr(settings, 'dtype', np.float32)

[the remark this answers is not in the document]
mloubout: So `settings` breaks CodePrinter with ` dtype`, so to avoid `pop` this now a different keyword argument, but it is resolved.

examples/seismic/acoustic/acoustic_example.py
FabioLuporini: yes, sorry there was a TYPO in my original comment -- link --> like

what I meant was that this deserves to be hoisted to a separate line a be properly commented

examples/seismic/acoustic/gradient_example.py
FabioLuporini: I understand, that's why I basically said "I agree with what you're doing". Nevertheless, it is still important to inform the author of this code about what's gonna happen.

examples/seismic/acoustic/wavesolver.py
```diff
@@ -153,7 +155,7 @@ def gradient(self, rec, u, v=None, grad=None, m=None, **kwargs):
 
         :returns: Gradient field and performance summary
         """
-
+        dt = kwargs.pop('dt', self.dt)
```
FabioLuporini: On a side note, I still don't understand why some fields may be **either** passed as method arguments **or** retrieved from the instance attributes.

examples/seismic/acoustic/wavesolver.py
```diff
@@ -167,8 +169,25 @@ def gradient(self, rec, u, v=None, grad=None, m=None, **kwargs):
...
             m = m or self.model.m
 
-        summary = self.op_grad().apply(rec=rec, grad=grad, v=v, u=u, m=m,
-                                       dt=kwargs.pop('dt', self.dt), **kwargs)
+        if checkpointing:
+            set_log_level('ERROR')
+            u = TimeFunction(name='u', grid=self.model.grid,
+                             time_order=2, space_order=self.space_order)
+            cp = DevitoCheckpoint([u])
+            n_checkpoints = None
```
FabioLuporini: what is this variable?

examples/seismic/model.py
FabioLuporini: remind me (even offline if you want) what's the difference with vti?

examples/seismic/model.py
FabioLuporini: reminder: we can drop this, then?

examples/seismic/model.py
FabioLuporini: reminder

examples/seismic/model.py
mloubout: Not quite, still needs to raise a warning and I think this is the best place to do

examples/seismic/acoustic/wavesolver.py
mloubout: Number of checkpoints to save, None lets PyRevolve decide

examples/seismic/acoustic/wavesolver.py
mloubout: Not sure what you mean there

examples/seismic/model.py
mloubout: `kwargs` will always contain all the physical parameter but some may be `None`, so the `or i` is required to get back the default physical parameter.

examples/seismic/tti/operators.py
mloubout: Warning thrown in model, but this is still required to avoid `cos(None)` for a 3D without `phi`

examples/seismic/model.py
```diff
@@ -347,6 +375,7 @@ def __init__(self, origin, spacing, shape, space_order, vp, nbpml=20,
 
         if epsilon is not None:
             if isinstance(epsilon, np.ndarray):
+                self._physical_parameters += ('epsilon',)
```
FabioLuporini: No! This might create issues if we instantiate multiple, different Models. For the way ``_physical_parameters`` is used, we shouldn't let objects to modify this class attribute

tests/test_gradient.py
FabioLuporini: I was convinced I had left a comment about this, but I cannot find it

``clear_cache`` should only appear at the end of those tests allocating lots of memory, and should be properly documented.

Even better: here you're calling ``clear_cache`` in two different tests (this and the one above). So I suggest to put these tests within the same class, and add the special ``pytest`` method [setup_method](https://docs.pytest.org/en/2.7.3/xunit_setup.html) with ``clear_cache`` in it (and a suitable comment).

examples/seismic/model.py
mloubout: I don't follow. This is the initialization of a model, and this attached to this model its physical representation. I don't see what issue it could cause.

tests/test_gradient.py
FabioLuporini: at least put it at the end, otherwise the last test execution will leave around some memory

examples/seismic/model.py
mloubout: It shouldn't which is why I originaly wrote it the other way. `_physical_parameters` is a class property that define what type of model you are working with and should not be shared between two models.

tests/test_gradient.py
mloubout: Isn't it gonna be ignored as still in the scope of the function?

FabioLuporini: I mean at the of each test, rather than at the beginning

examples/seismic/model.py
FabioLuporini: ok, now sorted out

tests/test_adjoints.py
```diff
@@ -0,0 +1,169 @@
...
+        ('layers', (60, 70), 'OT2', 12, 10), ('layers', (60, 70), 'OT4', 4, 10),
+        ('layers', (60, 70), 'OT4', 8, 10), ('layers', (60, 70), 'OT4', 12, 10),
+        # 3D tests with varying time and space orders
+        ('layers', (60, 70, 80), 'OT2', 4, 10), ('layers', (60, 70, 80), 'OT2', 8, 10),
+        ('layers', (60, 70, 80), 'OT2', 12, 10), ('layers', (60, 70, 80), 'OT4', 4, 10),
+        ('layers', (60, 70, 80), 'OT4', 8, 10), ('layers', (60, 70, 80), 'OT4', 12, 10),
+        # Constant model in 2D and 3D
+        ('constant', (60, 70), 'OT2', 8, 14), ('constant', (60, 70, 80), 'OT2', 8, 14),
+    ])
+    def test_adjoint_F(self, mkey, shape, kernel, space_order, nbpml):
```
FabioLuporini: can you add docstrings to these tests?

FabioLuporini:
@mloubout there are now a few leftover conflicts after merging 553

tests/test_derivatives.py
```diff
@@ -112,3 +112,51 @@ def test_second_derivatives_space(grid, derivative, dim, order):
...
     assert(expr == s_expr)  # Exact equailty
+
+
+@skipif_yask
+@pytest.mark.parametrize('space_order', [2, 4, 6, 8, 10, 12, 14, 16, 18, 20])
+# Only test x and t as y and z are the same as x
+@pytest.mark.parametrize('derivative', ['dx', 'dxl', 'dxr', 'dx2'])
+def test_fd_space(derivative, space_order):
+    """
+    This test compare the discrete finite-difference scheme against polynomials
```
FabioLuporini: typo (compares)

tests/test_derivatives.py
```diff
@@ -112,3 +112,51 @@ def test_second_derivatives_space(grid, derivative, dim, order):
...
+    For a given order p, the fiunite difference scheme should
```
FabioLuporini: typo

tests/test_gradient.py
```diff
@@ -3,145 +3,204 @@
...
+        # 'clear_cache()' to release any remaining memory from the previous tests or
+        # previous instances (different parametrizations) of these tests
+        clear_cache()
+
+    @pytest.mark.parametrize('space_order', [4])
+    @pytest.mark.parametrize('kernel', ['OT2'])
+    @pytest.mark.parametrize('shape', [(70, 80)])
+    def test_gradient_checkpointing(self, shape, kernel, space_order):
+        """
+        This test ensure that the FWI gradient computed with devito
```
FabioLuporini: typo (ensure)

FabioLuporini: devito -> "an :class`Operator`"

tests/test_gradient.py
```diff
@@ -3,145 +3,204 @@
...
+        assert np.allclose(gradient.data, gradient2.data)
+
+    @skipif_yask
+    @pytest.mark.parametrize('space_order', [4])
+    @pytest.mark.parametrize('kernel', ['OT2'])
+    @pytest.mark.parametrize('shape', [(70, 80)])
+    @pytest.mark.parametrize('checkpointing', [True, False])
+    def test_gradientFWI(self, shape, kernel, space_order, checkpointing):
+        """
+        This test ensure that the FWI gradient computed with devito
```
FabioLuporini: same typos as before

tests/test_gradient.py
```diff
@@ -3,145 +3,204 @@
...
+        info('2nd order error, Phi(m0+dm)-Phi(m0) - <J(m0)^T \delta d, dm>: %s' % (p2))
+        assert np.isclose(p1[0], 1.0, rtol=0.1)
+        assert np.isclose(p2[0], 2.0, rtol=0.1)
+
+    @pytest.mark.parametrize('space_order', [4])
+    @pytest.mark.parametrize('kernel', ['OT2'])
+    @pytest.mark.parametrize('shape', [(70, 80)])
+    def test_gradientJ(self, shape, kernel, space_order):
+        """
+        This test ensure that the Jacobian computed with devito
```
FabioLuporini: same typos as before

FabioLuporini:
closing as reopened as #583

PR #553, opened 2018-04-25
Spatial conditionals

### LINKED ISSUE #563: test_derivatives.py and test_finite_difference.py should be merged

mloubout:
Adds support fr spatial conditionals

- extends arguments checks as in the spatial case, the size needs to be passed as an argument
- remove unnecessary modulo for non stepping dimension (breaks compilation for multi conditionals)
- test

fixes #563

### DIFF
--- a/devito/dimension.py
+++ b/devito/dimension.py
@@ -351,9 +351,14 @@ class ConditionalDimension(DerivedDimension):
                          All iterations for which the expression evaluates to
                          True are part of the ``SubDimension`` region.
 
-    ConditioanlDimension need runtime arguments. The generated C code will require
+    ConditioanlDimension needs runtime arguments. The generated C code will require
     the size of the dimension to initialize the arrays as e.g:
-    float (*restrict u2)[xsub_size + 1 + 1][ysub_size + 1 + 1]
+    .. code-block:: python
+        x = grid.dimension[0]
+        x1 = ConditionalDimension(name='x1', parent=x, factor=2)
+        u1 = TimeFunction(name='u1', dimensions=(x1,), size=grid.shape[0]/factor)
+        # The generated code will look like
+        float (*restrict u1)[x1_size + 1] = ...
     """
 
     def __new__(cls, name, parent, **kwargs):
--- a/devito/dimension.py
+++ b/devito/dimension.py
@@ -350,6 +350,10 @@ class ConditionalDimension(DerivedDimension):
         * ``condition``: an arbitrary SymPy expression depending on ``parent``.
                          All iterations for which the expression evaluates to
                          True are part of the ``SubDimension`` region.
+
+    ConditioanlDimension need runtime arguments. The generated C code will require
+    the size of the dimension to initialize the arrays as e.g:
+    float (*restrict u2)[xsub_size + 1 + 1][ysub_size + 1 + 1]
     """
 
     def __new__(cls, name, parent, **kwargs):
--- a/devito/dimension.py
+++ b/devito/dimension.py
@@ -373,20 +373,6 @@ class ConditionalDimension(DerivedDimension):
     def _hashable_content(self):
         return (self.parent._hashable_content(), self.factor, self.condition)
 
-    def _arg_defaults(self, **kwargs):
-        """
-        A :class:`ConditionalDimension` provides no arguments, so this
-        method returns an empty dict.
-        """
-        return {}
-
-    def _arg_values(self, *args, **kwargs):
-        """
-        A :class:`ConditionalDimension` provides no arguments, so there are
-        no argument values to be derived.
-        """
-        return {}
-
     def _arg_check(self, *args):
         """
         A :class:`ConditionalDimension` provides no arguments, so there are
--- a/devito/dimension.py
+++ b/devito/dimension.py
@@ -359,6 +359,10 @@ class ConditionalDimension(DerivedDimension):
         return newobj
 
     @property
+    def spacing(self):
+        return self.factor * self.parent.spacing
+
+    @property
     def factor(self):
         return self._factor
 
--- a/devito/dimension.py
+++ b/devito/dimension.py
@@ -351,14 +351,16 @@ class ConditionalDimension(DerivedDimension):
                          All iterations for which the expression evaluates to
                          True are part of the ``SubDimension`` region.
 
-    ConditioanlDimension needs runtime arguments. The generated C code will require
+    ConditionalDimension needs runtime arguments. The generated C code will require
     the size of the dimension to initialize the arrays as e.g:
-    .. code-block:: python
-        x = grid.dimension[0]
-        x1 = ConditionalDimension(name='x1', parent=x, factor=2)
-        u1 = TimeFunction(name='u1', dimensions=(x1,), size=grid.shape[0]/factor)
-        # The generated code will look like
-        float (*restrict u1)[x1_size + 1] = ...
+
+        .. code-block:: python
+           x = grid.dimension[0]
+           x1 = ConditionalDimension(name='x1', parent=x, factor=2)
+           u1 = TimeFunction(name='u1', dimensions=(x1,), size=grid.shape[0]/factor)
+           # The generated code will look like
+           float (*restrict u1)[x1_size + 1] =
+
     """
 
     def __new__(cls, name, parent, **kwargs):
--- a/devito/function.py
+++ b/devito/function.py
@@ -453,64 +453,66 @@ class Function(TensorFunction):
         Dynamically create notational shortcuts for space derivatives.
         """
         for dim in self.space_dimensions:
+            name = dim.parent.name if dim.is_Derived else dim.name
             # First derivative, centred
             dx = partial(first_derivative, order=self.space_order,
                          dim=dim, side=centered)
-            setattr(self.__class__, 'd%s' % dim.name,
+            setattr(self.__class__, 'd%s' % name,
                     property(dx, 'Return the symbolic expression for '
                              'the centered first derivative wrt. '
-                             'the %s dimension' % dim.name))
+                             'the %s dimension' % name))
 
             # First derivative, left
             dxl = partial(first_derivative, order=self.space_order,
                           dim=dim, side=left)
-            setattr(self.__class__, 'd%sl' % dim.name,
+            setattr(self.__class__, 'd%sl' % name,
                     property(dxl, 'Return the symbolic expression for '
                              'the left-sided first derivative wrt. '
-                             'the %s dimension' % dim.name))
+                             'the %s dimension' % name))
 
             # First derivative, right
             dxr = partial(first_derivative, order=self.space_order,
                           dim=dim, side=right)
-            setattr(self.__class__, 'd%sr' % dim.name,
+            setattr(self.__class__, 'd%sr' % name,
                     property(dxr, 'Return the symbolic expression for '
                              'the right-sided first derivative wrt. '
-                             'the %s dimension' % dim.name))
+                             'the %s dimension' % name))
 
             # Second derivative
             dx2 = partial(generic_derivative, deriv_order=2, dim=dim,
                           fd_order=int(self.space_order / 2))
-            setattr(self.__class__, 'd%s2' % dim.name,
+            setattr(self.__class__, 'd%s2' % name,
                     property(dx2, 'Return the symbolic expression for '
                              'the second derivative wrt. the '
-                             '%s dimension' % dim.name))
+                             '%s dimension' % name))
 
             # Fourth derivative
             dx4 = partial(generic_derivative, deriv_order=4, dim=dim,
                           fd_order=max(int(self.space_order / 2), 2))
-            setattr(self.__class__, 'd%s4' % dim.name,
+            setattr(self.__class__, 'd%s4' % name,
                     property(dx4, 'Return the symbolic expression for '
                              'the fourth derivative wrt. the '
-                             '%s dimension' % dim.name))
+                             '%s dimension' % name))
 
             for dim2 in self.space_dimensions:
+                name2 = dim2.parent.name if dim2.is_Derived else dim2.name
                 # First cross derivative
                 dxy = partial(cross_derivative, order=self.space_order,
                               dims=(dim, dim2))
-                setattr(self.__class__, 'd%s%s' % (dim.name, dim2.name),
+                setattr(self.__class__, 'd%s%s' % (name, name2),
                         property(dxy, 'Return the symbolic expression for '
                                  'the first cross derivative wrt. the '
                                  '%s and %s dimensions' %
-                                 (dim.name, dim2.name)))
+                                 (name, name2)))
 
                 # Second cross derivative
                 dx2y2 = partial(second_cross_derivative, dims=(dim, dim2),
                                 order=self.space_order)
-                setattr(self.__class__, 'd%s2%s2' % (dim.name, dim2.name),
+                setattr(self.__class__, 'd%s2%s2' % (dim.name, name2),
                         property(dx2y2, 'Return the symbolic expression for '
                                  'the second cross derivative wrt. the '
                                  '%s and %s dimensions' %
-                                 (dim.name, dim2.name)))
+                                 (name, name2)))
 
     @classmethod
     def __indices_setup__(cls, **kwargs):
--- a/devito/ir/clusters/algorithms.py
+++ b/devito/ir/clusters/algorithms.py
@@ -212,7 +212,7 @@ def bump_and_contract(targets, source, sink):
             processed.append(e.func(e.lhs, e.rhs.xreplace(mapper)))
         else:
             for i in sink.tensors[function]:
-                scalarized = Scalar(name='s%d' % len(mapper)).indexify()
+                scalarized = Scalar(name='s%s' % i.function.name).indexify()
                 mapper[i] = scalarized
 
                 # Index bumping
--- a/devito/ir/clusters/algorithms.py
+++ b/devito/ir/clusters/algorithms.py
@@ -212,7 +212,7 @@ def bump_and_contract(targets, source, sink):
             processed.append(e.func(e.lhs, e.rhs.xreplace(mapper)))
         else:
             for i in sink.tensors[function]:
-                scalarized = Scalar(name='s%s' % i.function.name).indexify()
+                scalarized = Scalar(name='s%s%d' % (i.function.name, len(mapper))).indexify()
                 mapper[i] = scalarized
 
                 # Index bumping
--- a/devito/ir/clusters/algorithms.py
+++ b/devito/ir/clusters/algorithms.py
@@ -207,8 +207,8 @@ def bump_and_contract(targets, source, sink):
             processed.append(e.func(e.lhs, e.rhs.xreplace(mapper)))
         else:
             for i in sink.tensors[function]:
-                scalarized = Scalar(name='s%s%d' % (i.function.name, len(mapper))).indexify()
-                mapper[i] = scalarized
+                scalar = Scalar(name='s%s%d' % (i.function.name, len(mapper))).indexify()
+                mapper[i] = scalar
 
                 # Index bumping
                 assert len(function.indices) == len(e.lhs.indices) == len(i.indices)
@@ -216,7 +216,7 @@ def bump_and_contract(targets, source, sink):
                             zip(function.indices, e.lhs.indices, i.indices)}
 
                 # Array contraction
-                handle = e.func(scalarized, e.rhs.xreplace(mapper))
+                handle = e.func(scalar, e.rhs.xreplace(mapper))
                 handle = xreplace_indices(handle, shifting)
                 processed.append(handle)
     source.exprs = processed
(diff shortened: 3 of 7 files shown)


devito/dimension.py
```diff
@@ -388,6 +388,11 @@ def _arg_check(self, *args):
         """
         return
 
+    @property
+    def _arg_names(self):
```
FabioLuporini: can you reuse something from superclass through ``super``?

is it worth moving (part of) this method to the superclass?

devito/dimension.py
```diff
@@ -372,14 +372,14 @@ def _arg_defaults(self, **kwargs):
         A :class:`ConditionalDimension` provides no arguments, so this
         method returns an empty dict.
         """
-        return {}
+        return {self.size_name: kwargs.get('size')}
```
FabioLuporini: I have no idea why you have to do this

devito/dimension.py
```diff
@@ -372,14 +372,14 @@ def _arg_defaults(self, **kwargs):
...
 
     def _arg_values(self, *args, **kwargs):
         """
         A :class:`ConditionalDimension` provides no arguments, so there are
         no argument values to be derived.
         """
-        return {}
+        return kwargs
```
FabioLuporini: I have no idea why you have to do this

FabioLuporini: I think a second pair of eyes would be useful here @tjb900

devito/ir/iet/scheduler.py
```diff
@@ -122,10 +122,11 @@ def iet_make(clusters):
         for j, offs in ispace.sub_iterators.get(k.dim, []):
             modulo = len(offs)
             for n, o in enumerate(filter_ordered(offs)):
-                name = "%s%d" % (j.name, n)
-                vname = Scalar(name=name, dtype=np.int32)
-                value = (k.dim + o) % modulo
-                uindices.append(UnboundedIndex(vname, value, value, j, j + o))
+                if j.is_Stepping:
```
FabioLuporini: Why? would the test that you added fail without this?

FabioLuporini: I don't think I like the use of ``is_Stepping`` here

tests/test_dimension.py
```diff
@@ -306,3 +325,7 @@ def test_no_index(self):
         # with u[t] = t
         # v = 16 * 1 + 64 * 2 + 144 * 3 + 256 * 4 = 1600
         assert np.all(np.allclose(v.data, 1600))
+
+
+if __name__ == "__main__":
```
FabioLuporini: drop

tests/test_dimension.py
```diff
@@ -170,6 +170,25 @@ def test_basic(self):
         assert np.all([np.allclose(usave.data[i], i*factor)
                       for i in range((nt+factor-1)//factor)])
 
+    def test_spacial_subsampling(self):
```
FabioLuporini: docstring

FabioLuporini: perhaps showing what code we aim to generate would be useful, because I can't decipher it from the code below (or at least, I'm not totally sure I'm getting it...)

mloubout: it's saving `u` every two grid points in both direction.

devito/dimension.py
mloubout: Brute force making work left over to clean up.

devito/dimension.py
mloubout: Because you need to input the size of the subdim

devito/ir/iet/scheduler.py
mloubout: This modulo instanciations are useless for all non-stepping dimensions.

mloubout: without it:
```
devito @ eas-ce-21203 [mlouboutin3]: python tests/test_dimension.py 
/tmp/devito-xb903407/60f8101685795ba12d376ffc42a1298fd38f5883.c: In function ‘Kernel’:
/tmp/devito-xb903407/60f8101685795ba12d376ffc42a1298fd38f5883.c:29:42: error: expected iteration declaration or initialization before ‘x’
       for (int x = x_m, xsub0 = (x)%(1); x <= x_M; x += 1, xsub0 = (x)%(1))
```

tests/test_dimension.py
```diff
@@ -170,6 +170,57 @@ def test_basic(self):
...
+        functions.
+        """
+        nt = 19
+        grid = Grid(shape=(12, 12))
+        time = grid.time_dim
+
+        u = TimeFunction(name='u', grid=grid, save=nt)
+        assert(grid.time_dim in u.indices)
+
+        # Creates subsampled spatial dimensions and accordine grid
```
FabioLuporini: typo

tests/test_dimension.py
```diff
@@ -170,6 +170,57 @@ def test_basic(self):
...
+        dims = tuple([ConditionalDimension(d.name+'sub', parent=d, factor=2)
+                      for d in u.grid.dimensions])
+        grid2 = Grid((6, 6), dimensions=dims)
+        u2 = TimeFunction(name='u2', grid=grid2, save=nt, space_order=2)
+        assert(time in u2.indices)
+        print(u2.dx)
```
FabioLuporini: leftover -- though I have the impression you're still in a debugging phase!

tests/test_dimension.py
```diff
@@ -170,6 +170,57 @@ def test_basic(self):
...
+        u2 = TimeFunction(name='u2', grid=grid2, save=nt)
+        assert(time in u2.indices)
+
+        eqns = [Eq(u.forward, u + 1.), Eq(u2, u)]
+        op = Operator(eqns)
+        op.apply(time_M=nt-2)
+        # Verify that u2[x,y]= u[2*x, 2*y]
+        assert np.allclose(u.data[:-1, 0:-1:2, 0:-1:2], u2.data[:-1, :, :])
+
+    def test_subsampled_fd(self):
```
FabioLuporini: this test should probably go to ``test_finite_differences.py`` or ``test_derivatives.py``, I don't remember

mloubout: Not sure, this one is a weird test can can go with FD/dimension so not sure where it is best to put.

tests/test_dimension.py
mloubout: Yes still debbuging to try to remove the `if is_Stepping`. Trying to catch why all these `x-sub` end up in that part of the scheduler as modulo indices

tests/test_dimension.py
FabioLuporini: OK, I think we should split this test into two tests. The first one should be about the symbolic interface only, and could go to one of the two files above (btw, should we merge them?)

devito/ir/iet/scheduler.py
FabioLuporini: I think it's because of this:
```
class ConditionalDimension(DerivedDimension):

    is_NonlinearDerived = True
```

tests/test_dimension.py
mloubout: >btw, should we merge them?

Probably yes

devito/ir/iet/scheduler.py
FabioLuporini: which in turn causes this:
```
def build_intervals(stencil):
    """
    ...
    """
    iterators = OrderedDict()
    for i in stencil.dimensions:
        if i.is_NonlinearDerived:
            iterators.setdefault(i.parent, []).append(stencil.entry(i))
```

FabioLuporini: **maybe** dropping the ``is_NonlinearDerived`` flag makes it just work. WARNING: this is just a random thought and I might be completely wrong!

mloubout: Just tried, goes crazy. It takes  the SubDim as separate and writes loop for instead of if

devito/ir/support/space.py
```diff
@@ -492,7 +492,12 @@ def project(self, cond):
         else:
             func = lambda i: i in cond
         intervals = [i for i in self.intervals if func(i.dim)]
-        sub_iterators = {k: v for k, v in self.sub_iterators.items() if func(k)}
+        sub_iterators = {}
+        for k, v in self.sub_iterators.items():
+            if func(k) and len(v) == 0:
+                sub_iterators.update({k: v})
+            elif func(k) and len(v) > 0:
+                sub_iterators.update({k: [] if v[0].dim.is_Conditional else v})
```
mloubout: @FabioLuporini this seems to fix the problem for conditionals and modulo. Not sure if it is the right place to do it though.

mloubout:
So now `test_adjointJ.py` is broken and no idea why, runs perfectly fine on my three systems

tests/test_derivatives.py
```diff
@@ -112,3 +113,79 @@ def test_second_derivatives_space(grid, derivative, dim, order):
...
+    for i in range(nt):
+        for j in range(u2.data_allocated.shape[2]):
+            u2.data_allocated[i, :, j] = np.arange(u2.data_allocated.shape[2])
+
+    eqns = [Eq(u.forward, u + 1.), Eq(u2.forward, u2.dx)]
+    op = Operator(eqns, dse="advanced")
+    op.apply(time_M=nt-2)
+    # Verify that u2[x,y]= u[2*x, 2*y]
+    print(u2.data[:, 3, 3])
+    assert np.all(np.allclose(u.data[-1], nt-1))
```
tjb900: just `np.allclose` is enough here

tests/test_dimension.py
```diff
@@ -22,8 +22,8 @@ def test_domain_vs_interior(self):
         t = grid.stepping_dim  # noqa
 
         u = TimeFunction(name='u', grid=grid)  # noqa
-        eqs = [Eq(u.forward, u + 1),
-               Eq(u.forward, u.forward + 2, region=INTERIOR)]
+        eqs = [Eq(u.forward, u + 1.),
+               Eq(u.forward, u.forward + 2., region=INTERIOR)]
```
tjb900: Out of curiosity, what did this fix?  If int vs float in this context matters, I need to be more careful in other places, too.

tests/test_dimension.py
```diff
@@ -170,6 +170,31 @@ def test_basic(self):
...
+        dims = tuple([ConditionalDimension(d.name+'sub', parent=d, factor=2)
+                      for d in u.grid.dimensions])
+        grid2 = Grid((6, 6), dimensions=dims)
+        u2 = TimeFunction(name='u2', grid=grid2, save=nt)
+
+        eqns = [Eq(u.forward, u + 1.), Eq(u2, u)]
+        op = Operator(eqns)
+        op.apply(time_M=nt-2)
+        # Verify that u2[x,y]= u[2*x, 2*y]
+        assert np.all(np.allclose(u.data[nt-1], nt-1))
```
tjb900: same as above - no need for the wrapping `np.all`

devito/ir/clusters/algorithms.py
```diff
@@ -106,9 +106,36 @@ def guard(clusters):
...
+                                                 c.atomics, guards)
+                        processed.append(cluster)
+                        break
+            else:
+                # Add first cluster no matter what
+                cluster = PartialCluster(exprs, c.ispace, c.dspace, c.atomics, guards)
+                processed.append(cluster)
+
+    processed = ClusterGroup(c for c in processed if len(c.exprs) > 0)
+    print([(c.exprs, c.guards)for c in processed if len(c.guards) > 0])
```
tjb900: probably don't want this to land

tests/test_dimension.py
mloubout: It doesn't change anything, some debug leftover due to types issues at some point.

devito/ir/clusters/algorithms.py
mloubout: Indeed, thought I removed it, my bad

devito/dimension.py
```diff
@@ -367,20 +371,6 @@ def condition(self):
...
         return (self.parent._hashable_content(), self.factor, self.condition)
 
-    def _arg_defaults(self, **kwargs):
-        """
-        A :class:`ConditionalDimension` provides no arguments, so this
-        method returns an empty dict.
-        """
-        return {}
-
-    def _arg_values(self, *args, **kwargs):
```
FabioLuporini: Can you update the docstring of ConditionalDimension explaining why it might need runtime arguments?

devito/function.py
```diff
@@ -453,64 +453,66 @@ def _initialize_derivatives(self):
         Dynamically create notational shortcuts for space derivatives.
         """
         for dim in self.space_dimensions:
+            name = dim.parent.name if dim.is_Derived else dim.name
```
FabioLuporini: are we sure this is what we want? @tjb900 ?

tests/test_derivatives.py
```diff
@@ -112,3 +113,78 @@ def test_second_derivatives_space(grid, derivative, dim, order):
...
+    u2 = TimeFunction(name='u2', grid=grid2, save=nt, space_order=1)
+    for i in range(nt):
+        for j in range(u2.data_allocated.shape[2]):
+            u2.data_allocated[i, :, j] = np.arange(u2.data_allocated.shape[2])
+
+    eqns = [Eq(u.forward, u + 1.), Eq(u2.forward, u2.dx)]
+    op = Operator(eqns, dse="advanced")
+    op.apply(time_M=nt-2)
+    # Verify that u2[x,y]= u[2*x, 2*y]
+    print(u2.data[:, 3, 3])
```
FabioLuporini: leftover

devito/ir/clusters/algorithms.py
```diff
@@ -106,8 +106,35 @@ def guard(clusters):
...
+                for i in range(len(processed)):
+                    pr = processed[i]
+                    common_expr = pr.exprs == exprs
+                    common_guards = any(a == b for a, b in zip(pr.guards, guards))
+                    # Check common expression and common coditions
+                    if common_expr and pr.guards == guards:
+                        continue
+                    elif common_expr and common_guards:
+                        for dim in keys:
+                            # Merge conditons if both Eq and Ne for a given expression
```
FabioLuporini: can you add a comment with an example of what you're trying to achieve?

devito/function.py
mloubout: I prefer to keep the shortcut the same `u.dx, ...` for both parent and child as the Dimension is technically the same just with a different spacing. It is also more natural for user I feel to always call '.dx; rather than trying to figure out what the name would be

devito/dimension.py
```diff
@@ -348,6 +348,10 @@ class ConditionalDimension(DerivedDimension):
         * ``condition``: an arbitrary SymPy expression depending on ``parent``.
                          All iterations for which the expression evaluates to
                          True are part of the ``SubDimension`` region.
+
+    ConditioanlDimension need runtime arguments. The generated C code will require
```
FabioLuporini: typo: needs

devito/dimension.py
```diff
@@ -348,6 +348,10 @@ class ConditionalDimension(DerivedDimension):
...
+    the size of the dimension to initialize the arrays as e.g:
+    float (*restrict u2)[xsub_size + 1 + 1][ysub_size + 1 + 1]
```
FabioLuporini: for better docs, use

```
.. code-block::
    
    ...
```

there are a few examples throughout the code base (u can use ``git grep`` to find them)

devito/ir/clusters/algorithms.py
```diff
@@ -106,8 +106,43 @@ def guard(clusters):
 
             guards = [(i.parent, conditions[i]) for i in dims]
             guards.extend([(i.parent, negated[i]) for i in ndims])
-            cluster = PartialCluster(exprs, c.ispace, c.dspace, c.atomics, dict(guards))
-            processed.append(cluster)
+            guards = dict(guards)
+
+            # Merge of conditonals if both conditions. In case of multiple Conditionals
+            # we can have a case where we have:
```
FabioLuporini: The following block of code is quite messy. I need to take a look at it. Next week.

devito/ir/support/space.py
```diff
@@ -523,7 +523,14 @@ def project(self, cond):
         else:
             func = lambda i: i in cond
         intervals = [i for i in self.intervals if func(i.dim)]
-        sub_iterators = {k: v for k, v in self.sub_iterators.items() if func(k)}
+        sub_iterators = {}
+        for k, v in self.sub_iterators.items():
```
FabioLuporini: what was the issue here?

devito/ir/clusters/algorithms.py
```diff
@@ -213,16 +248,16 @@ def bump_and_contract(targets, source, sink):
             processed.append(e.func(e.lhs, e.rhs.xreplace(mapper)))
         else:
             for i in sink.tensors[function]:
-                scalarized = Scalar(name='s%d' % len(mapper)).indexify()
-                mapper[i] = scalarized
+                scalar = Scalar(name='s%s%d' % (i.function.name, len(mapper))).indexify()
```
FabioLuporini: what was the issue here?

devito/ir/support/space.py
mloubout: Some non-modulo dimension where ending up in the for loop signature crashing openMP simd

devito/ir/clusters/algorithms.py
mloubout: In some (very odd) cases, there was some duplicates temporaries so the compiler was complaining about redeclaration.

devito/dimension.py
```diff
@@ -348,6 +348,15 @@ class ConditionalDimension(DerivedDimension):
         * ``condition``: an arbitrary SymPy expression depending on ``parent``.
                          All iterations for which the expression evaluates to
                          True are part of the ``SubDimension`` region.
+
+    ConditioanlDimension needs runtime arguments. The generated C code will require
```
FabioLuporini: typo

devito/dimension.py
```diff
@@ -348,6 +348,15 @@ class ConditionalDimension(DerivedDimension):
...
+    the size of the dimension to initialize the arrays as e.g:
+    .. code-block:: python
```
FabioLuporini: if you don't leave blank lines and indent it properly, I'm not sure this is gonna be rendered properly (I might be wrong). For example:

```
ConditionalDimension needs runtime arguments. The generated C code will require
the size of the dimension to initialize the arrays as, for example:

    .. code-block:: python

        <code goes here>
```

Can you check that?

devito/ir/clusters/algorithms.py
FabioLuporini: do we have a test covering this? but yes, I can see why this might happen. This is a symptom that this routine should better be moved to the DSE... (of course, this is on me and for another day)

devito/ir/support/space.py
FabioLuporini: OK, reminder: I need to look into this.

What's a test causing the issue w/o this fix?

tests/test_adjointJ.py
```diff
@@ -12,11 +13,12 @@
 @pytest.mark.parametrize('space_order', [4, 8, 12])
 @pytest.mark.parametrize('shape', [(60, 70), (40, 50, 30)])
 def test_acousticJ(shape, space_order):
+    clear_cache()
```
FabioLuporini: I think I wrote it somewhere but I cannot find it. ``clear_cache`` should be called only at the very end of those tests that are going to allocate lots of memory, not at the beginning of some (more or less) random tests

devito/ir/clusters/algorithms.py
mloubout: I'm not sure, the kill of some branch does not depend only on the dimension but on the expression too that the poewrset just ignores.

devito/ir/clusters/algorithms.py
```diff
@@ -80,21 +80,26 @@ def guard(clusters):
...
         for e in c.exprs:
+            dimc = ()
             for k, v in e.ispace.sub_iterators.items():
                 for i in v:
                     if i.dim.is_Conditional:
-                        mapper.setdefault(i.dim, []).append(e)
-
+                        dimc += (i.dim,)
+            if len(dimc) > 0:
+                mapper.setdefault(dimc, []).append(e)
```
mloubout: @FabioLuporini I think this way is probably nicer. Mainly the mapper now looks like
{(xsub, ysub, tsub): expr}

This way the the guard is driven by the expression rather than the dimensions and take cares of all the associated conditions at once.

devito/ir/clusters/algorithms.py
```diff
@@ -103,11 +108,20 @@ def guard(clusters):
                 # variables when the conditional is false
                 continue
 
-            guards = [(i.parent, conditions[i]) for i in dims]
-            guards.extend([(i.parent, negated[i]) for i in ndims])
-            cluster = PartialCluster(exprs, c.ispace, c.dspace, c.atomics, dict(guards))
+            if len(dims) > 0:
```
mloubout: The first case case `dims=()` correspond to the unconditional expressions that don't need guards.

FabioLuporini:
as I wrote offline, the ``fablup/multi-cond`` branch contains lots of improvements, within two commits. These commits would have never been possible w/o the work here, so thanks again. I suggest we incorporate them. All tests are already passing.

After that, we must rebase and clean up the history, because in its current form it's too messy. Then, I'm happy to land this.

FabioLuporini:
Merged.

PR #555, opened 2018-04-26
Travis setup

mloubout:
Attempt to go back to line by line TRAVIS to catch all errors.

- currently setup with failing adjoint test to verify it does what it should

- Yes the comments are removed. TRAVIS refuses to recognize the yml file as valid with the comments in.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,16 +79,15 @@ before_install:
   - conda info -a
 
 install:
-  # Install devito with either pip or conda
+  # Install devito with conda
   - if [[ $INSTALL_TYPE == 'conda' ]]; then
       conda env create -q -f environment.yml python=$TRAVIS_PYTHON_VERSION;
       source activate devito;
       pip install -e .;
       conda list;
     fi
-  - if [[ $INSTALL_TYPE == 'pip_setup' ]]; then
-      python setup.py install;
-    fi
+  # Install devito with pip
+  - if [[ $INSTALL_TYPE == 'pip_setup' ]]; then python setup.py install; fi
 
 before_script:
   - echo -e "Host github.com\n\tStrictHostKeyChecking no\n" >> ~/.ssh/config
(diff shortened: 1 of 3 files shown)


mloubout:
Looks like we're back on track. Log looks as before my mess up.

mloubout:
@fablup so this is finally working as expected.

It is failing as designed too to verify TRAVIS works correctly. If you're happy with it I'll put back the correct test and should be goood to go.

.travis.yml
```diff
@@ -29,7 +29,7 @@ matrix:
           packages:
             - gcc-5
             - g++-5
-      env: DEVITO_ARCH=gcc-5 DEVITO_OPENMP=0 INSTALL_TYPE=conda
+      env: DEVITO_ARCH=gcc-5 DEVITO_OPENMP=0 SEISMIC=examples
```
FabioLuporini: can you leave INSTALL_TYPE=conda ? it'd make things easier to understand

.travis.yml
```diff
@@ -99,36 +99,58 @@ before_script:
...
+  - if [[ $SEISMIC == 'examples' ]]; then
+      DEVITO_BACKEND=foreign py.test -vs tests/test_operator.py -k TestForeign;
+    fi
+  - if [[ $SEISMIC == 'examples' ]]; then
+      python examples/seismic/benchmark.py test -P tti -so 4 -a -d 20 20 20 -n 5;
+    fi
+  - if [[ $SEISMIC == 'examples' ]]; then
+      python examples/seismic/benchmark.py test -P acoustic -a;
+    fi
+  - if [[ $SEISMIC == 'examples' ]]; then
```
FabioLuporini: do the examples fail if the ``if-then-else`` is not inline?

.travis.yml
```diff
@@ -99,36 +99,58 @@ before_script:
...
+  # Run tests with pip
+  - if [[ $INSTALL_TYPE == 'pip_setup' ]]; then python setup.py test; fi
+  # Run test with py.tests
+  - if [[ $INSTALL_TYPE != 'pip_setup' ]]; then py.test -vs --cov devito tests/; fi
+  # Run remaining specialized examples and tests
+  # Additional seismic operator tests
+  - if [[ $SEISMIC == 'examples' ]]; then
+      DEVITO_BACKEND=foreign py.test -vs tests/test_operator.py -k TestForeign;
+    fi
+  - if [[ $SEISMIC == 'examples' ]]; then
```
FabioLuporini: I prefer inlined ``if-then-else`` in this file, when possible

tmbgreaves:
Fairly confident I can tidy that up in Jenkins, yes. Having spent the last few days neatening a Jenkinsfile for OK I have a bunch of ideas to simplify/neaten the devito Jenkinsfile.

FabioLuporini:
> Fairly confident I can tidy that up in Jenkins, yes. Having spent the last few days neatening a Jenkinsfile for OK I have a bunch of ideas to simplify/neaten the devito Jenkinsfile.

Awesome!

.travis.yml
mloubout: Yes

.travis.yml
mloubout: no it's fine onliner too

mloubout:
>The proliferation of if-then-else really starts scaring me.

Should technically only need one but we want to run the examples each individually

> SEISMIC

true, will rename, it `RUN_EXAMPLES` fine?

FabioLuporini:
> true, will rename, it RUN_EXAMPLES fine?

OK. However, if ``RUN_EXAMPLES == True`` **if and only if** ``INSTALL_TYPE == conda``, then I wonder why we don't just use ``INSTALL_TYPE`` ?

mloubout:
>if and only if INSTALL_TYPE == conda

No also `False` for YASK backend

FabioLuporini:
I have got no idea why Codecov is complaining that much

mloubout:
I have honestly no idea. Can't figure out why it has such a huge difference.

FabioLuporini:
after quite in-depth inspection, I still can't spot anything odd with `codecov`. Merging

FabioLuporini:
Merged.

PR #560, opened 2018-05-02
ir: ScheduleTree level

FabioLuporini:
The ultimate goal is adding BC support to the YASK backend. This preliminary PR improves the IR to make life easier when it comes to do transformations in the YASK backend.

The main addition of this PR is the ScheduleTree (``stree``), an intermediate level of representation between that of ``Clusters`` and ``IET``. When lowering from a sequence of Clusters to an IET, two main things are required:
1- the logic to determine what iteration spaces can be shared by different clusters;
2- the construction of the IET itself

previously, these two passes were intertwined in the same routine (``iet_make``), at  the same level of abstraction. This PR decouples them; an ``stree`` now takes care of point 1 above, by "unwinding"  the sequence of clusters into a minimal, mutable tree data structure. ``iet_make`` now only traverses the ``stree`` to emit a new, "fatter", immutable tree, namely IET.

This pass allowed me to drastically simplify the profiling engine, which is now vastly more clear (to understand) and robust than it used to be. I compared its output to that of ``master`` on a set of notable examples, but I'd like some of you (@navjotk, @tjb900, @ggorman, @mloubout  ) to do the same. Let me emphasise that all these changes are crucial for the development of the YASK backend.


devito/profiling.py
```diff
@@ -1,121 +1,87 @@
...
     """
+    sections = FindNodes(Section).visit(iet)
+
+    # Construct the Profiler
     profiler = Profiler(name)
+    for section in sections:
+        # All ExpressionBundles within `section`
+        bundles = FindNodes(ExpressionBundle).visit(section)
 
-    trees = retrieve_iteration_tree(node)
```
FabioLuporini: this whole piece of garbage is going away

devito/profiling.py
```diff
@@ -126,48 +92,40 @@ def new(self):
...
-            traffic = float(profile.memory*dataspace*dtype().itemsize)
+            traffic = float(data.traffic.subs(arguments)*dtype().itemsize)
+
+            # Runtime itershapes
+            itershapes = [tuple(i.subs(arguments) for i in j) for j in data.itershapes]
 
             # Derived metrics
-            oi = flops/traffic
+            gflops = float(ops)/10**9
             gflopss = gflops/time
```
tjb900: Since we're here, I should comment that I sometimes see crashes with these `time`s equal to zero.
Could we do `gflops/time if time > 0 else 0` or similar?

devito/profiling.py
```diff
@@ -126,48 +92,40 @@ def new(self):
...
+            gpoints = float(points)/10**9
             gpointss = gpoints/time
+            oi = ops/traffic
```
tjb900: Likewise here for `traffic`

devito/operator.py
```diff
@@ -359,17 +364,21 @@ def _profile_output(self, args):
         summary = self.profiler.summary(args, self._dtype)
         with bar():
             for k, v in summary.items():
-                name = '%s<%s>' % (k, ','.join('%d' % i for i in v.itershape))
-                gpointss = ", %.2f GPts/s" % v.gpointss if k == 'main' else ''
-                info("Section %s with OI=%.2f computed in %.3f s [%.2f GFlops/s%s]" %
+                itershapes = [",".join(str(i) for i in its) for its in v.itershapes]
+                if len(itershapes) > 1:
+                    name = "%s<%s>" % (k, ",".join("<%s>" % i for i in itershapes))
```
tjb900: I love that it lists all the itershapes now.  That makes it much clearer what's included in each one (in the case of BCs).

devito/profiling.py
FabioLuporini: yes!

FabioLuporini: hang on, that's weird because of this line above:
```
max(getattr(arguments[self.name]._obj, section.name), 10e-7)
```
so, ``time`` should never be 0 😮

tjb900: In...deed.  I can't remember exactly what it was.  Maybe it was only `traffic`.

FabioLuporini:
Merged.

PR #561, opened 2018-05-02
Fix and improve dimension_sort

### LINKED ISSUE #559: dimension_sort behaving badly
I think there are perhaps several issues here

#### 1. `partial_order` hangs when given some (impossible) ordering problems.

e.g.

```
from devito.tools import partial_order
partial_order([(1,2,3),(3,4,2)])
<hangs>
```

#### 2. Side issue, but relevant for creating a test: `dimension_sort` is nondeterministic.

This is due to the fact that `free_symbols` is not ordered in the following (`devito/symbolics/inspection.py`):

```
                constraint.extend([d for d in ai.free_symbols
                                   if isinstance(d, Dimension) and d not in constraint])
```

#### 3. `dimension_sort` can sometimes generate such impossible ordering problems.

Due to (1), this is catastrophic.  Due to (2), it's difficult to write a useful CI test for this.

Running the test below several times will illustrate the problem.

```
def test_dimension_sort():
    grid = Grid(shape=(10, 10))
    p = Dimension('p')
    r = Dimension('r')
    d = Dimension('d')
    time = grid.time_dim
    x, y = grid.dimensions

    a = Function(name='a', dimensions=(time, p), shape=(10, 1))
    b = Function(name='b', dimensions=(time, x, y), shape=(10, 10, 10))
    c = Function(name='c', shape=(1, 2),
                 dimensions=(p, d), dtype=np.int32)
    f = Function(name='f', dimensions=(p, r), shape=(1, 1))

    expr = Eq(a[time, p], b[time, c[p, 0]+r, c[p, 1]] * f[p, r])

    op = Operator(expr)
```

To save you adding the logging, the above results in one of the two calls below to `partial_order`.  The second one is one of the impossible hanging problems:

```
partial_order([(time, p), (p, r), (time, p, r)])
partial_order([(time, p), (p, r), (time, r, p)])
```

I think it's generating constraints which are too strict.  The lookup into `b` is what generates the `(time,p,r)` or `(time,r,p)` order constraint.  But I don't believe that part of the expression `(c[p,0]+r, c[p,1])` implies any necessary or desirable ordering between `r` and `p`.

FabioLuporini:
@tjb900 I'm sorry you've been hit by the bugs in #559 -- these are due to archaic code that I should have changed a long time ago. Anyway, this PR hopefully fixes all of the issues that you pointed out, and also improves the code quality in general.

Hopefully this will make Travis happy. Should you encounter analogous issues in the future, please append the failing tests either here:
``test_ir.py::test_dimension_sort``
or
``test_tools.py::test_toposort``

I also took the opportunity to relocate some code

fixes #559

### DIFF
--- a/tests/test_ir.py
+++ b/tests/test_ir.py
@@ -293,11 +293,6 @@ def test_dependences_eq(expr, expected, ti0, ti1, fa):
     (['Eq(ti0[x,y,z], ti1[x,y,z])',
       'Eq(ti3[x,y,z], ti0[y+1,y,y])'],
      ['ti0,flow,x', 'ti0,anti,x']),
-    # Data indices don't match iteration indices, so conservatively assume
-    # all sort of deps
-    (['Eq(ti0[x,y,z], ti1[x,y,z])',
-      'Eq(ti3[x,y,z], ti0[x,y,x])'],
-     ['ti0,flow,z', 'ti0,anti,z']),
 ])
 def test_dependences_scope(exprs, expected, ti0, ti1, ti3, fa):
     """
--- a/tests/test_ir.py
+++ b/tests/test_ir.py
@@ -2,8 +2,11 @@ import pytest
 
 from conftest import EVAL, time, x, y, z, skipif_yask  # noqa
 
-from devito import Eq, Grid, TimeFunction  # noqa
+import numpy as np
+
+from devito import Eq, Grid, Function, TimeFunction, Dimension  # noqa
 from devito.ir.equations import DummyEq, LoweredEq
+from devito.ir.equations.algorithms import dimension_sort
 from devito.ir.iet.nodes import Conditional, Expression
 from devito.ir.support.basic import IterationInstance, TimedAccess, Scope
 from devito.ir.support.space import (NullInterval, Interval, IntervalGroup,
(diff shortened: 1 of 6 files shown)


tests/test_ir.py
```diff
@@ -290,11 +293,6 @@ def test_dependences_eq(expr, expected, ti0, ti1, fa):
     (['Eq(ti0[x,y,z], ti1[x,y,z])',
       'Eq(ti3[x,y,z], ti0[y+1,y,y])'],
      ['ti0,flow,x', 'ti0,anti,x']),
-    # Data indices don't match iteration indices, so conservatively assume
```
FabioLuporini: Note: I'm dropping this simply because it was conceptually broken. The new implementation spotted the flaw.

tests/test_tools.py
```diff
@@ -3,16 +3,20 @@
...
-from devito.tools import partial_order
+from devito.tools import toposort
 
 
 @skipif_yask
 @pytest.mark.parametrize('elements, expected', [
     ([[a, b, c], [c, d, e]], [a, b, c, d, e]),
     ([[e, d, c], [c, b, a]], [e, d, c, b, a]),
-    ([[a, b, c], [b, d, e]], [a, b, c, d, e]),
+    ([[a, b, c], [b, d, e]], [a, b, d, c, e]),
```
FabioLuporini: Note: both orderings would be fine, but this version of ``toposort`` returns ``d, c``, rather than ``c, d``. It is deterministic, and that's the only thing that really matters at this point

FabioLuporini:
Merged.

PR #564, opened 2018-05-03
CI: lower tests verbosity [RFC]

FabioLuporini:
Let's see how it looks.


mloubout:
Looks good, outputs the same as `pip` tests. Can you make one fail to see if it shows which/why it fails correctly?

FabioLuporini:
> Looks good, outputs the same as pip tests. Can you make one fail to see if it shows which/why it fails correctly?

Yes, that was going to be my next step. BTW, we have a hanging dask notebook....... (see #development)

navjotk:
I see the examples still show their timings output in travis logs. I believe that should be the first thing to be removed as that is of less interest on travis than the regular console output on tests.

FabioLuporini:
Regardless, are you OK with what this PR does?

I agree we can/should also shut off the examples.

PR #565, opened 2018-05-03
Changed install source for dask, distributed and tornado to git.

ggorman:


FabioLuporini:
is running off their ``master`` safe? I remember lots of trouble when we were following this approach with PETSc in Firedrake. Especially if these projects are particularly active... what do you think? I guess you wanna do this to get all of the latest bug fixes, correct? The risk is that we get sudden breakages of our own ``master``. 

Also, don't forget that these three packages should belong to a layer on top of that of Devito, so having them as compulsory dependencies feels weird. I think we've been there already, but since we're now also adding ``tornado``...

If you feel that this "is just gonna be OK", then fine. 

At the end of the day, the way I see it, there are three solutions:

- We fork these three projects, just like we do with YASK: https://github.com/opesci/yask 
This way, we would update our forked masters only when we now that the upstream ones are actually making all of our tests passing. The downside is that this require some effort on our side.
- We make these three packages optionals (don't know what's the best way of doing that)
- We just say "fine", we run this "risk", as this is only going to affect our Dask tests, so it shouldn't be a big deal for users

FabioLuporini:
Also, can you squash your commits into a single one?
```
git checkout master
git pull
git checkout deadlock
git rebase -i master
<now you mark with ``s`` all commits except the root one, save and go on>
git push -f origin deadlock
```

tjb900:
There's also this: https://stackoverflow.com/questions/13685920/install-specific-git-commit-with-pip

We've touched on this before, I think.  There's a reasonably strong case for having predictable versions of dependencies used by travis.  Otherwise it breaks suddenly and we assume it's something we did, when in fact it's because some other package got upgraded.

The user-facing install should probably not do that, however.  Nominating exact version requirements is far more likely to cause unpleasant version interactions if their environments include packages not in devito's list.

PR #567, opened 2018-05-04
allow max == min-1 for 0-sized loops

tjb900:
Since the change to inclusive bounds, `d_s==d_e` is no longer a zero-iteration loop.  The ability to have these is very handy, e.g. for disabling a receiver or source loop without needing to recompile the Operator.

This patch pretty much does what it says in the title - allows the specific case of `d_M==d_m-1`, to express this intention.  Open to suggestions for something more elegant.

### DIFF
--- a/devito/dimension.py
+++ b/devito/dimension.py
@@ -188,7 +188,8 @@ class Dimension(AbstractSymbol):
             raise InvalidArgument("OOB detected due to %s=%d" % (self.max_name,
                                                                  args[self.max_name]))
 
-        if args[self.max_name] < args[self.min_name]:
+        # Allow the specific case of max=min-1, which disables the loop
+        if args[self.max_name] < args[self.min_name]-1:
             raise InvalidArgument("Illegal max=%s < min=%s"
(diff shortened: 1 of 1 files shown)


FabioLuporini:
Looping in @navjotk

FabioLuporini:
@tjb900 how about we let ``op.apply`` accept special symbols implying "no iterations should be performed along a given dimension"? 

Example:
```
from devito import ..., O

op = Operator(...)
op.apply(..., time=O.Bottom)
```

Where ``O`` should be a bag of special symbols with special meaning to be used at operator application time. Clearly, we should find a better name than ``O``, but probably something super short. ``Bottom`` is also debatable. But anyway, thoughts?

tjb900:
Yeah, that sounds good to me - and certainly conveys the user's intention more clearly.  I definitely agree that `Bottom` is probably not the right word.  `NoIterations`?

Actually...  another idea...  you still allow the `d_n=N` syntax in lieu of `_m` and `_M`, right?  So we could just use `d_n=0`?

FabioLuporini:
Merged.

PR #575, opened 2018-05-15
RFC: ALLOC_GUARD to raise segfaults rather than causing silent corruption

tjb900:
I was motivated to add this after wasting time tracking down a mistake which turned out to be a stupid one (using `u.forward` on a `TimeFunction` with `time_order=0`), and getting errors like this:

```
Fatal Python error: GC object already tracked
```

or this:

```
Fatal Python error: XXX block stack underflow

Current thread 0x00007f4d9882f740 (most recent call first):
Aborted
```

or various others, all with useless tracebacks under gdb, since the memory had been corrupted long before the crash.

It adds a new allocator, the `GuardAllocator`, which inserts a 1MB (by default) padding both before and after every allocation, and sets `mprotect` to cause a segfault if the padding area is accessed.  This causes more crashes, but at least they crash where they should, and can be debugged.

It sets this to be the default allocator if `DEVITO_DEVELOP=1`.

### DIFF
--- a/devito/data.py
+++ b/devito/data.py
@@ -54,15 +54,15 @@ class MemoryAllocator(object):
         :param shape: Shape of the array to allocate.
         :param dtype: Numpy datatype to allocate.
 
-        :returns (pointer, c_pointer): The first element of the tuple is the reference
-                                       that can be used to access the data as a ctypes
-                                       object. The second element is the low-level
-                                       reference that is needed only for the call to free.
+        :returns (pointer, free_handle): The first element of the tuple is the reference
+                                         that can be used to access the data as a ctypes
+                                         object. The second element is an opaque object
+                                         that is needed only for the call to free.
         """
         size = int(reduce(mul, shape))
         ctype = numpy_to_ctypes(dtype)
 
-        c_pointer = self._alloc_C_libcall(size, ctype)
+        c_pointer, free_args = self._alloc_C_libcall(size, ctype)
         if c_pointer is None:
             raise RuntimeError("Unable to allocate %d elements in memory", str(size))
 
@@ -70,12 +70,14 @@ class MemoryAllocator(object):
                                                                   shape=shape))
         pointer = np.ctypeslib.as_array(c_pointer, shape=shape)
 
-        return (pointer, c_pointer)
+        return (pointer, free_args)
 
     @abc.abstractmethod
     def _alloc_C_libcall(self, size, ctype):
         """
         Perform the actual memory allocation by calling a C function.
+        Should return a 2-tuple (c_pointer, free_args), where the free args
+        are what is handed back to free() later to deallocate.
 
         .. note::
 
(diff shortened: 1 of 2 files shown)


devito/data.py
```diff
@@ -122,11 +126,91 @@ def _alloc_C_libcall(self, size, ctype):
...
+
+        # generate pointers to the left padding, the user data, and the right pad
+        padleft_pointer = c_pointer
+        c_pointer = ctypes.c_void_p(c_pointer.value + self.padding)
+        padright_pointer = ctypes.c_void_p(c_pointer.value + npages_user * pagesize)
+
+        # and set the permissions on the pad memory to 0 (no access)
+        # if these fail, don't worry about failing the entire allocation
+        c_padsize = ctypes.c_ulong(self.padding)
+        print("protect", padleft_pointer, padright_pointer)
```
FabioLuporini: leftover, or use logger (perhaps print out only in high-verbosity mode)

devito/data.py
```diff
@@ -122,11 +126,91 @@ def _alloc_C_libcall(self, size, ctype):
...
+    immediately if an out-of-bounds access occurs.
+
+    Further, the remainder region of the last page (which cannot be
+    protected), is poisoned with NaNs.
+    """
+
+    is_Posix = True
+
+    @classmethod
+    def initialize(cls):
```
FabioLuporini: If you inherit directly from ``PosixAllocator`` you won't need this class method, nor the ``is_Posix`` flag. Not sure if there are other implications though

devito/data.py
```diff
@@ -122,11 +126,91 @@ def _alloc_C_libcall(self, size, ctype):
...
+        handle = find_library('c')
+        if handle is not None:
+            cls.lib = ctypes.CDLL(handle)
+
+    def __init__(self, padding=1048576):
```
FabioLuporini: what's this magic number?

devito/data.py
```diff
@@ -273,18 +358,18 @@ class Data(np.ndarray):
 
     def __new__(cls, shape, dimensions, dtype, allocator=ALLOC_FLAT):
         assert len(shape) == len(dimensions)
-        ndarray, c_pointer = allocator.alloc(shape, dtype)
+        ndarray, free_args = allocator.alloc(shape, dtype)
         obj = np.asarray(ndarray).view(cls)
         obj._allocator = allocator
-        obj._c_pointer = c_pointer
+        obj._free_args = free_args
```
FabioLuporini: I'd call this ``_memfree_args`` (or whatever, as long as it's got a reference to "memory"), because ``_free_args`` may be misinterpreted

devito/data.py
tjb900: oops

devito/data.py
tjb900: 1MB

devito/data.py
tjb900: ah yes, good plan.  Will do that.

FabioLuporini:
Merged.

PR #578, opened 2018-05-15
yask: Initial support for advanced boundary conditions

FabioLuporini:
@tjb900 finally filing this PR.


devito/yask/transformer.py
```diff
@@ -0,0 +1,178 @@
...
+
+    # Add flow dependences to the offloaded equations
+    # TODO: This can be improved by spotting supergroups ?
+    for to, frm in zip(processed, processed[1:]):
+        yc_soln.add_flow_dependency(frm, to)
+
+    return mapper
+
+
+def handle(expr, yc_soln, mapper):
```
FabioLuporini: note for reviewers: this function is nothing new, it's just the formerly called ``sympy2yask``

devito/yask/wrappers.py
```diff
@@ -52,7 +52,7 @@ def __init__(self, name, yc_soln, local_grids=None):
         try:
             compiler = configuration.yask['compiler']
             opt_level = 1 if configuration['develop-mode'] else 3
-            make(namespace['path'], ['-j3', 'YK_CXX=%s' % compiler.cc,
+            make(namespace['path'], ['-j', 'YK_CXX=%s' % compiler.cc,
```
FabioLuporini: @chuckyount note: finally dropping the ``j3``

devito/dimension.py
```diff
@@ -276,6 +276,26 @@ def _defines(self):
     def _arg_names(self):
         return self.parent._arg_names
 
+    def _arg_defaults(self, **kwargs):
```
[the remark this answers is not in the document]
mloubout: Yeah this may be problematic and in conflict with #553

devito/ir/iet/analysis.py
```diff
@@ -70,6 +71,12 @@ def mark_parallel(analysis):
         for depth, i in enumerate(tree):
             if i in properties:
                 continue
+
+            if i.uindices:
```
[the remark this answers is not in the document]
mloubout: I think this does something else, mainly droping omp if the loop looks bad. This is a good thing, but the change in #553 only forces the compiler to only write out the only required indices in the loop (and drop all the %1 of unused indices).
So I guess this would be on top of my changes but still requires #553

devito/dimension.py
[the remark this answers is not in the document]
FabioLuporini: @mloubout OK no worries, I reverted to whatever we had before so this shouldn't create conflicts with what you've changed

FabioLuporini:
any further thoughts @mloubout @tjb900 ?

mloubout:
GTG to me

tests/test_dimension.py
```diff
@@ -62,6 +82,44 @@ def test_subdim_middle(self):
...
+
+        u = TimeFunction(name='u', save=None, grid=grid, space_order=0, time_order=1)
+
+        xleft = SubDimension.left(name='xleft', parent=x, thickness=thickness)
+        xi = SubDimension.middle(name='xi', parent=x,
+                                 thickness_left=thickness, thickness_right=thickness)
+        xright = SubDimension.right(name='xright', parent=x, thickness=thickness)
+
+        yi = SubDimension.middle(name='yi', parent=y,
+                                 thickness_left=4, thickness_right=4)
```
tjb900: the assertions below use `thickness` rather than 4 for this dimension as well - doesn't really matter but should be consistent.

devito/yask/transformer.py
```diff
@@ -0,0 +1,178 @@
...
+        for i in tree:
+            if not i.dim.is_Sub:
+                continue
+
+            ydim = nfac.new_domain_index(i.dim.parent.name)
+
+            try:
+                # Are the extremes statically known ?
+                lower = nfac.new_const_number_node(int(i.dim.symbolic_ofs_lower))
+                upper = nfac.new_const_number_node(int(i.dim.symbolic_ofs_upper))
```
tjb900: This obviously works, but it _looks_ like it might ignore some possible cases.
YASK allows you to construct any expression of the form FIRST+integer or LAST+integer, right?

So there are three cases, I imagine:

1. SubDimension goes from x_m+3 to x_M-3 (handled correctly)
2. SubDimension goes from x_m to x_m+3 (left boundary - I think not handled by this code since symbolic_ofs_upper is not an integer).  This could be expressed in the YASK API, though, as far as I can see.
3. SubDimension goes from x_M-3 to x_M (same situation for the right boundary - symbolic_ofs_lower not an integer)

But it passes the tests, so I'm not sure what I'm missing here.

FabioLuporini: >  I think not handled by this code since symbolic_ofs_upper is not an integer)

but then you fall back to the ``except`` below, don't you? where ``i.extent()`` is read... This should cover cases 2 and 3.
What am I missing?

tjb900: The `except:` part only applies to sequential loops, though.  So that needlessly excludes boundary conditions or other regions where some parallelism is possible - i.e. those without the intra-timestep dependency like the one that we are using.

Something like:

```
start_expressible_in_yask = (
    (i.dim.symbolic_start - i.dim.parent.symbolic_start).is_Number
    or (i.dim.symbolic_start - i.dim.parent.symbolic_end).is_Number)
end_expressible_in_yask = (
    (i.dim.symbolic_end - i.dim.parent.symbolic_start).is_Number
    or (i.dim.symbolic_end - i.dim.parent.symbolic_end).is_Number)
subdim_statically_known = start_expressible_in_yask and end_expressible_in_yask
```

Now that I think about it, is it also possible that you might need to check for is_Parallel in the top part?

devito/yask/transformer.py
```diff
@@ -0,0 +1,178 @@
...
+                if i.is_Sequential:
+                    extent = i.extent()
+
+                    if i.direction is Backward:
+                        _range = reversed(range(extent))
+                        node = nfac.new_first_domain_index(ydim)
+                    elif i.direction is Forward:
+                        _range = range(-extent + 1, 1)
+                        node = nfac.new_last_domain_index(ydim)
+                    else:
```
tjb900: Is this assuming that Forward means "right boundary" and Backward means "left boundary"?  That works for us, but I'm not sure it's a safe assumption, in general.

devito/yask/transformer.py
```diff
@@ -0,0 +1,178 @@
...
+                        assert False
+
+                    unwinded = []
+                    for e, v in conditions:
+                        for r in _range:
+                            expr = nfac.new_add_node(node, nfac.new_const_number_node(r))
+                            condition = nfac.new_equals_node(ydim, expr)
+                            unwinded.append((e, v + [nfac.new_equals_node(ydim, expr)]))
```
tjb900: Since you're (maybe) changing this anyway, and I think I now understand what this code means, I suspect you want `unwound` here.

FabioLuporini: uops :+1:

devito/yask/transformer.py
FabioLuporini: uhm, I think I'm confused and have to think about this !

FabioLuporini: do you suggest to enrich the test suite with some further tests? because if you spotted this, assuming it's actually a bug/issue, then we will have to cover it anyway. 
Of course, no rush: at this point, we can leave this open until next week. But if you could produce one or two MFEs, that'd be great!

tjb900: Just added some examples to `test_yask.py`.  

I think the logic needs to be something like:

* if the loop is parallel AND bounds can be expressed as a yask condition, then
   - build the yask condition, and apply it
* otherwise, if the loop is sequential, and the number of iterations is known at compile time, then:
   - unwind into a series of yask regions, each with its own condition, and with dependencies to enforce the order
* otherwise, don't offload to yask

Is that sensible?

devito/yask/transformer.py
tjb900: See `test_yask.py::TestOperatorSimple::test_subdimleft_notparallel` for a contrived example.

devito/yask/transformer.py
FabioLuporini: > Just added some examples to test_yask.py.

These seem very relevant for ``core`` too, so I hope you don't mind if I move them to the more generic ``test_dimension.py``? Anyway, these look great !

> I think the logic needs to be something like [...] Is that sensible?

yeah... I mean, there are things of the current implementation that I can't remember why I did them this way 😮 as in, the (flow)[https://github.com/opesci/devito/blob/yask-regions/devito/yask/transformer.py#L41] kinda matches what you described, but there are several things that I don't quite understand myself ! Hopefully I'll be able to elucidate soon

tjb900: Also note that one of the tests: `test_yask.py::TestOperatorSimple::test_subdimleft_parallel` passes, but it hasn't offloaded to YASK as it should.  I couldn't immediately see an example of how to test that the offload worked.

FabioLuporini: a hacky way is to check for the presence of the string `run_solution` within `str(op)`

FabioLuporini: > So that needlessly excludes boundary conditions or other regions where some parallelism is possible - i.e. those without the intra-timestep dependency like the one that we are using.

This should be covered in the ``try`` part. Did  you see what ``i.dim.symbolic_ofs_lower/upper`` return?

tjb900: I can't see how, though...?

```
 def symbolic_ofs_upper(self):
        return self._interval.right - self.parent.symbolic_end
```

For a `SubDimension.left`, we might have `interval.right = x_m + 3`.  So the above would return `x_m + 3 - x_M`, which ought to fail to cast to `int`...?

FabioLuporini:
merged.

PR #586, opened 2018-05-23
yask: Add support for misc dimensions

### LINKED ISSUE #577: misc dimensions support for yask backend
Apparently the YASK backend assumes that all dimensions are either space or time.  This leads to errors e.g. in cases where a `Function` was specified via `dimensions` and `shape`, rather than `grid`.

Poorly named reproducer pushed to branch `yask_misc_dims`.  I don't feel like this should be a yask-specific test, so I didn't want to put into test_yask.py.  I put it in its own class since most others are `skipif_yask` (but maybe that's worth revisiting...?).

Core:

```
pytest -v tests/test_operator.py::TestsForCoreAndYask
tests/test_operator.py::TestsForCoreAndYask::test_misc_dim PASSED        
```

Yask:

```
DEVITO_BACKEND=yask pytest -v tests/test_operator.py::TestsForCoreAndYask
...
tests/test_operator.py:1284: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
devito/yask/function.py:112: in data
    return self.data_domain
/node/bud32/envs/devito_contrib/lib/python3.6/site-packages/cached_property.py:26: in __get__
devito/yask/function.py:56: in wrapper
    context = contexts.fetch(self.grid, self.dtype)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = ContextManager(), grid = None, dtype = <class 'numpy.float32'>

    def fetch(self, grid, dtype):
        """
            Fetch the :class:`YaskContext` in ``self`` uniquely identified by
            ``grid`` and ``dtype``. Create a new (empty) :class:`YaskContext` on miss.
            """
        # A unique key for this context.
>       key = (configuration['isa'], dtype, grid.dimensions,
               grid.time_dim, grid.stepping_dim)
E       AttributeError: 'NoneType' object has no attribute 'dimensions'

devito/yask/wrappers.py:353: AttributeError
```

FabioLuporini:
the title says everything

~this now depends on #588~
~this now depends on #590~

fixes #577


devito/ir/support/basic.py
```diff
@@ -284,6 +284,13 @@ def is_irregular(self):
     def is_scalar(self):
         return self.rank == 0
 
+    def affine_if_present(self, findices):
+        """
+        Return True if the provided ``findices`` appear in ``self`` and are
+        affine, False otherwise.
+        """
```
tjb900: Docstring is a bit misleading, if I'm understanding the code correctly?

"Returns False if findices appear in self and are _not_ affine, True otherwise".

devito/yask/function.py
```diff
@@ -148,3 +167,35 @@ def _arg_defaults(self, alias=None):
...
+        # grids
+        indices[cls._time_position] = kwargs['grid'].time_dim
+        return tuple(indices)
+
+    def _arg_defaults(self, alias=None):
+        args = super(TimeFunction, self)._arg_defaults(alias=alias)
+        # This is a little hack: a TimeFunction originally meant to be accessed
+        # via modulo buffered iteration should never impose anything on the time
+        # dimension
+        if type(self.save) is not int:
```
tjb900: I'm not sure that this is a very elegant condition...  e.g. if someone accidentally computes save=(8/2), this will fail and confuse them.

Should we just use the truthiness of `save`, instead?  (and also complain bitterly if it's truthy and not an integer...  but that's presumably elsewhere).

devito/yask/transformer.py
```diff
@@ -130,49 +135,54 @@ def nary2binary(args, op):
         if function.is_Constant:
             if function not in mapper:
                 mapper[function] = yc_soln.new_grid(function.name, [])
-            return mapper[function].new_relative_grid_point([])
-        elif not function.is_Dimension:
+            return mapper[function].new_grid_point([])
+        elif function.is_Dimension:
```
tjb900: Great.  This seems much nicer than what you've deleted lower down!

devito/yask/transformer.py
```diff
@@ -130,49 +135,54 @@ def nary2binary(args, op):
...
-        # Detect offset from dimension. E.g., in `[x+3,y+4]`, detect `[3,4]`
+        # Convert the Indexed into a YASK grid access
         indices = []
-        for i, j in zip(expr.indices, function.indices):
+        for i in expr.indices:
             if isinstance(i, LoweredDimension):
-                access = i.origin
+                indices.append(make_yask_ast(i.origin, yc_soln, mapper))
+            elif i.is_integer:
+                # A YASK misc dimension causes this
```
tjb900: Technically any dimension that we access explicitly by index, but I agree that those are most likely to be misc.

devito/yask/transformer.py
```diff
@@ -130,49 +135,54 @@ def nary2binary(args, op):
...
+                indices.append(make_yask_ast(i, yc_soln, mapper))
             else:
-                # SubDimension require this
+                # We must always use the parent ("main") dimension when creating
+                # YASK expressions
                 af = split_affine(i)
                 dim = af.var.parent if af.var.is_Derived else af.var
-                access = dim + af.shift
-            indices.append(int(access - j))
-        return mapper[function].new_relative_grid_point(indices)
```
tjb900: I just thought of another case that might need handling here, if it doesn't do it already (I have to run shortly but can test later).

The BCs often use the velocity value just inside the boundary, so they will have an expression like.

`u[t+1, xleft, y, z] = some_function_of(vel[x_m+3, y, z], u[t, ....], u[t-1, ....])`

These expressions involving `x_m` and `x_M` may need handling here just like in the loop case in the previous PR.  (in fact, if they are handled here you might even just be able to call `make_yask_ast` from that code.

FabioLuporini: right, but this opens another question. How do we propagate the runtime value ``x_m`` to YASK? (@chuckyount ) perhaps it's worth discussing this on Slack

FabioLuporini: dumb me, just use the usual ``FIRST_INDEX(x)`` or whatever is called, I guess?

devito/yask/function.py
FabioLuporini: Just to be clear. ``self.save`` is a type. In particular, its value can be either `int` or `NoneType`; see here:
https://github.com/opesci/devito/blob/master/devito/function.py#L660

Hence, you basically just spotted a bug here: I shouldn't be using ``type(self.save)``, but rather ``self.save``. I'm fixing this !

devito/yask/transformer.py
FabioLuporini: The PR is already quite big, and there are others coming along. I suggest to leave this for another day (until we have a test)

devito/yask/transformer.py
FabioLuporini: right, gonna tweak this comment

devito/yask/function.py
tjb900: Eek.  Ok!

FabioLuporini:
Merged.

PR #588, opened 2018-05-27
Fix Dimension caching

FabioLuporini:
Critical fix for caching of Dimension object

The test that I added would fail w/o this fix. Quite embarrassing.


FabioLuporini:
> Is this a problem we need to look at for the Function too or are we fine on that side?

should be OK as for Functions we rely on our own caching

FabioLuporini:
Merged.

PR #589, opened 2018-05-31
A general rewrite of Jenkins config

tmbgreaves:
This rewrites the Jenkins config as purely declarative pipeline, doing all the non-system work on the local
Jenkins source through Jenkinsfile as opposed to within the docker build.


Dockerfile.jenkins
```diff
@@ -38,72 +38,13 @@ RUN if [ -f /usr/bin/gcov-dump-$gccvers ] ; then \
...
-RUN bash miniconda.sh -b -p /usr/local/miniconda
-ENV PATH /usr/local/miniconda/bin:$PATH
+    
+# Set up conda
+RUN wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O /tmp/miniconda.sh
+RUN bash /tmp/miniconda.sh -b -p /usr/local/miniconda
+ENV PATH=/usr/local/miniconda/bin:$PATH
 RUN conda config --set always_yes yes --set changeps1 no
 RUN conda update -q conda
-RUN ln -s /usr/local/miniconda/etc/profile.d/conda.sh /etc/profile.d/conda.sh
```
FabioLuporini: I'm trusting you here

FabioLuporini: as in, for all these deletions here

Jenkinsfile
```diff
@@ -1,67 +1,145 @@
...
+                     environment { 
+                         HOME="${WORKSPACE}"
+                         DEVITO_BACKEND="yask"
+                         DEVITO_OPENMP="0"
+                         YC_CXX="g++-7"
+                     }
+                     steps {
+                         cleanWorkspace()
+                         condaInstallDevito()
+                         installYask()
```
FabioLuporini: I have this feeling that with Jenkins we could just have a separate YASK build ?

Jenkinsfile
```diff
@@ -1,67 +1,145 @@
...
-                sh "py.test -vs --nbval examples/cfd"
-            }
-            // Now run coverage
-            sh "codecov"
-            // Now build documentation
-            sh "sphinx-apidoc -f -o docs/ examples/"
-            sh "sphinx-apidoc -f -o docs/ devito/ devito/yask/*"
-            sh "make -C docs/ html"
-        }
+def cleanWorkspace() {
```
FabioLuporini: dumb question: why do we need ``cleanWorkspace``? don't we get a new, clean VM for each build?

Jenkinsfile
```diff
@@ -1,67 +1,145 @@
...
+    sh 'python setup.py test'
+}
+
+def runCondaTests() {
+    sh 'source activate devito ; py.test --cov devito tests/'
+}
+
+def runExamples () {
+    sh 'source activate devito ; python examples/seismic/benchmark.py test -P tti -so 4 -a -d 20 20 20 -n 5'
+    sh 'source activate devito ; python examples/seismic/benchmark.py test -P acoustic -a'
```
FabioLuporini: have you attempted random failures here, to be 100% sure they are correctly caught ?

Jenkinsfile
```diff
@@ -1,67 +1,145 @@
...
+    sh 'source activate devito ; python examples/seismic/acoustic/acoustic_example.py --full --checkpointing'
+    sh 'source activate devito ; python examples/seismic/acoustic/acoustic_example.py --constant --full'
+    sh 'source activate devito ; python examples/misc/linalg.py mat-vec mat-mat-sum transpose-mat-vec'
+    sh 'source activate devito ; python examples/seismic/tti/tti_example.py -a'
+    sh 'source activate devito ; python examples/seismic/tti/tti_example.py -a --noazimuth'
+    sh 'source activate devito ; py.test --nbval examples/cfd'
+    sh 'source activate devito ; py.test --nbval examples/seismic/tutorials/0[1-3]*'
+}
+
+def runCodecov() {
```
FabioLuporini: this one is never called?

Jenkinsfile
```diff
@@ -1,67 +1,145 @@
...
+    sh 'codecov'
+}
+
+def buildDocs() {
```
FabioLuporini: this one is never called?

FabioLuporini:
@mloubout another pair of eyes would be helpful here

Dockerfile.jenkins
tmbgreaves: The deletions are kind of spurious in the context of this PR - it's in the nature of a 'complete rewrite' as opposed to a modification, but diff does its best to generate a set of changes leaving some of the initial version intact ... :-)

Jenkinsfile
tmbgreaves: Could you elaborate on that a bit? As in, would you like to have a gcc7 build without yask and then another one with yask? That's very easy indeed.

Jenkinsfile
tmbgreaves: Frustratingly, this seems to be a cost of agent persistence on azure. I've set the agents to spin up and then remain active for (I think...) two hours so if there's a continuous stream of builds going on it avoids having to wait for the spin-up time on subsequent builds. However, Jenkins seems not to clean up the working source on the agent, at least in terms of removing extra files which are put in on top of the git checkout, ie. the conda environment install which in this case happens inside the source directory as it's the only place guaranteed to be writeable by the jenkins user on the agent. I think this is probably a bug in Jenkins which I should report, but as a fix for now I put in the cleanup stage. Even git clean -f -d isn't enough as that doesn't touch the yask repo which is nested inside the devito checkout...

Jenkinsfile
tmbgreaves: No, good thought. I'll branch and do that.

Jenkinsfile
tmbgreaves: Good catch - fixed in c411d49

Jenkinsfile
tmbgreaves: Good catch, fixed in c411d49

tmbgreaves:
@FabioLuporini - please could you say a bit more about the docker install? Am I interpreting correctly that this would be a build which installed devito into a docker image suitable for pushing as a self-contained devito environment to dockerhub? That would be very possible - though I'd suggest it be a separate branch and PR so Jenkins can get going testing with these changes at first.

tmbgreaves:
On the subject of future work on this, there are two areas *not* addressed by this change, one being improving performance of the Jenkins run, the other being a reimplementation of doctr for Jenkins. The former I guess we play with post-merge, the latter I think probably deserves its own PR as I guess it will be a matter of doing some Jenkins credential-passing and some kind of Jenkinsfile-derived merge of the built docs with the gh-pages repo.

Jenkinsfile
FabioLuporini: no sorry, my bad, I was confused by something else :)

Jenkinsfile
FabioLuporini: ok, that makes sense, thanks

FabioLuporini:
> @FabioLuporini - please could you say a bit more about the docker install? Am I interpreting correctly that this would be a build which installed devito into a docker image suitable for pushing as a self-contained devito environment to dockerhub? That would be very possible - though I'd suggest it be a separate branch and PR so Jenkins can get going testing with these changes at first.

yes, and yes. Happy to have it in a future PR.

> On the subject of future work on this, there are two areas not addressed by this change, one being improving performance of the Jenkins run, the other being a reimplementation of doctr for Jenkins. The former I guess we play with post-merge, the latter I think probably deserves its own PR as I guess it will be a matter of doing some Jenkins credential-passing and some kind of Jenkinsfile-derived merge of the built docs with the gh-pages repo.

OK. So, for the time being and until we have doctr working again, what do you think of keeping both Jenkins and Travis running in parallel?

I also suggest to wait for a second pair of eyes (@mloubout ) to go through this PR , then we can merge it.

Thanks again for the hard work. Really appreciated.

tmbgreaves:
I think it's sensible to run them in parallel anyway for a while to make sure they at least basically agree.

In terms of the doctr methods - am I seeing correctly that what it's doing is checking out gh-pages, merging in the built docs directory, and pushing back any updated files? I think that with the github token already in Jenkins it should be fairly straightforward to replicate that. Famous last words....

Jenkinsfile
tmbgreaves: I've successfully got it to fail in flake8 and when in the example step.

PR #591, opened 2018-06-01
Split tools.py into separate files under tools/

FabioLuporini:
Besides improving the organisation of the various support functions and classes, this PR removes an annoying dependence between tools and ``configuration``.

### DIFF
--- a/devito/tools/utils.py
+++ b/devito/tools/utils.py
@@ -4,7 +4,7 @@ import inspect
 import ctypes
 from collections import Callable, Iterable, OrderedDict, Hashable, Mapping
 from decorator import decorator
-from functools import partial, wraps, reduce
+from functools import partial, reduce
 from itertools import chain, combinations, product, zip_longest
 from operator import attrgetter, mul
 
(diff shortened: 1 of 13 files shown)


devito/tools/utils.py
```diff
@@ -0,0 +1,245 @@
...
+        # the format c_X_p, where X is the C typename, for instance `int` or `float`.
+        return ctype.__name__[2:-2]
+    else:
+        raise TypeError('Unrecognised %s during converstion to C type' % str(ctype))
+
+
+def ctypes_pointer(name):
+    """Create a ctypes type representing a C pointer to a custom data type ``name``."""
+    return type("c_%s_p" % name, (ctypes.c_void_p,), {})
+
```
mloubout: shouldn't these three be in `types` or something?

FabioLuporini: I **think** this is a lower level thing

mloubout: Yes it is, just seem a bit out of place there. I'll leave it to you to decide. GTG otherwise

FabioLuporini:
Merged.

PR #593, opened 2018-06-06
move time substitution to SparseTimeFunction and time shifts test

mloubout:
Moved time substitution for inject/interpolate to `SparseTimeFunction`


devito/function.py
```diff
@@ -1078,6 +1064,62 @@ def __indices_setup__(cls, **kwargs):
...
+        """
+        expr = indexify(expr)
+
+        # Apply optional time symbol substitutions to expr
+        if u_t is not None:
+            time = self.grid.time_dim
+            t = self.grid.stepping_dim
+            expr = expr.subs(t, u_t).subs(time, u_t)
+
+        variables = list(retrieve_indexed(expr))
```
FabioLuporini: we should find a way of calling the superclass method

tests/test_interpolation.py
```diff
@@ -20,6 +20,16 @@ def a(shape=(11, 11)):
     return a
 
 
+@pytest.fixture
+def at(shape=(11, 11)):
```
FabioLuporini: at?

tests/test_interpolation.py
```diff
@@ -281,4 +365,4 @@ def test_position(shape):
 
 
 if __name__ == "__main__":
-    test_interpolate_custom((11, 11), [(.05, .9), (.01, .8)])
+    test_inject_time_shift((11, 11), [(.05, .95), (.45, .45)], 1.)
```
FabioLuporini: unnecessary?

tests/test_interpolation.py
mloubout: time a

devito/function.py
mloubout: It is done for inject. The problem with this one is the `cummulative` that need `p_t` and once the expression is created in `Super` there is no "easy" way to recover the part of the rhs that needs the substitution

tests/test_interpolation.py
mloubout: I like the `__main__` :) but can remove it

devito/function.py
mloubout: Maybe, let me try it.

tests/test_interpolation.py
mloubout: Oh yeah, usually set to the last failing one :D I'll remove it

mloubout:
DO NOT MERGE.

The test pass, but the generated code is invalid for `u_t` not None as the actual index in the C code in incorrect. Looking into it.

devito/function.py
```diff
@@ -966,26 +966,18 @@ def interpolate(self, expr, offset=0, u_t=None, p_t=None, cummulative=False):
...
         # List of indirection indices for all adjacent grid points
         subs, idx_subs = self._interpolation_indices(variables, offset)
         # Substitute coordinate base symbols into the coefficients
         rhs = sum([expr.subs(vsub) * b.subs(subs)
                    for b, vsub in zip(self.coefficients, idx_subs)])
-        # Apply optional time symbol substitutions to lhs of assignment
-        lhs = self if p_t is None else self.subs(self.indices[0], p_t)
 
-        rhs = rhs + lhs if cummulative is True else rhs
+        rhs = rhs + self.subs(self_subs) if cummulative is True else rhs
```
FabioLuporini: if your LHS has lots of sympy objects in it, this is gonna be a bit slower than in master.

perhaps do 
```
lhs = self.subs(self_subs)
```
then use it where needed?

BTW, nitpicking: I'd call it ``lhs_subs``, as tbh "self" is the function, not the indexed object that you are substituting into

tests/test_dimension.py
```diff
@@ -369,7 +369,7 @@ def test_subdimleft_notparallel(self):
         assert all(i.is_Affine and i.is_Sequential for i in iterations if i.dim == xl)
         assert all(i.is_Affine and i.is_Parallel for i in iterations if i.dim == yi)
 
-        op.apply(time_m=0, time_M=0)
+        op.apply(time_m=1, time_M=1)
```
FabioLuporini: what happened here?

devito/ir/iet/scheduler.py
```diff
@@ -65,7 +65,11 @@ def iet_make(stree):
                     # Apart from SteppingDimension, no other type of Dimension
                     # requires generation of uindices
                     continue
-                modulo = len(offs)
+                # Get the modul span. First at 0 (no offset) if not present then
+                # get full offset
+                offs.update({0})
```
FabioLuporini: this might become unnecessary in #594 

look at this line: https://github.com/opesci/devito/pull/594/files#diff-18b194953e1480c7b66bc0641f9c6b9eR82

the extent of the modulo iteration is now dictated by the size of the TimeFunction buffer, not by the stencil indices.

devito/function.py
mloubout: >if your LHS has lots of sympy objects in it, this is gonna be a bit slower than in master.

LHS is always just `rec` or `src` so should be fine

tests/test_dimension.py
mloubout: Modulo change. Mainly the stencil is `u[t+1,...] = u[t+1]``` with 0th time_order so first `time` to be updated is `t+1` for `t=0` so `1`.

devito/ir/iet/scheduler.py
mloubout: Not sure, this litteraly just look at the maximum offsets around the current point, which may be bigger than the `time_order` if for example you take some weird time derivative

FabioLuporini: I think the modulo should be dictated by the extent of the time buffer, not by the offsets used within a particular Operator. Otherwise, you might even generate out-of-bound accesses (and thus segfaults).

tests/test_interpolation.py
```diff
@@ -185,6 +228,46 @@ def test_inject(shape, coords, result, npoints=19):
     assert np.allclose(a.data[indices], result, rtol=1.e-5)
 
 
+@skipif_yask
+@pytest.mark.parametrize('shape, coords, result', [
+    ((11, 11), [(.05, .95), (.45, .45)], 1.),
+    ((11, 11, 11), [(.05, .95), (.45, .45), (.45, .45)], 0.5)
+])
+def test_inject_time_shift(shape, coords, result, npoints=19):
+    """Test generic point interpolation testing the x-coordinate of an
```
FabioLuporini: injection?

examples/seismic/acoustic/acoustic_example.py
```diff
@@ -11,11 +11,14 @@
 def smooth10(vel, shape):
     if np.isscalar(vel):
         return .9 * vel * np.ones(shape, dtype=np.float32)
-    out = np.copy(vel)
+    out = np.zeros(vel.shape)
+    out[:] = vel[:]
     nz = shape[-1]
 
+    full_dims = [slice(0, d) for d in shape[:-1]]
```
FabioLuporini: could you put a comment here explaining what you're trying to achieve?

FabioLuporini:
Merged.

PR #594, opened 2018-06-07
Enhancements for stepping dimension

FabioLuporini:
This PR introduces

- an API to control the size of the time buffer when using a stepping (modulo) dimension. Refer to ``function.Buffer`` and the related test in ``test_save.py`` for more info. This API enables users to implement the timeslot-avoidance optimisation (TAO)
- when running with ``dle=speculative``, the DLE will emit a message if TAO is applicable. The idea is to switch this to ``dle=advanced`` after some beta-testing
- support for multiple timesteppers (ie, multiple ``TimeFunctions`` with different time buffer sizes) in the same Operator
- misc improvements basically everywhere
- several tests for all of these new features

This PR would benefit from some careful beta-testing


devito/equation.py
```diff
@@ -64,3 +64,24 @@ def __eq__(self, other):
 Represent the physical interior domain of the PDE; that is, PDE boundaries are
 not included.
 """
+
+
+def solve(eq, target, **kwargs):
```
FabioLuporini: @mloubout : a first step towards our own ``solve``

devito/ir/iet/analysis.py
```diff
@@ -145,41 +144,54 @@ def mark_vectorizable(analysis):
 def mark_wrappable(analysis):
```
FabioLuporini: note for reviewers: this is the function that determines whether TAO is applicable or not

devito/dimension.py
```diff
@@ -533,28 +563,147 @@ def _arg_values(self, *args, **kwargs):
...
+    def symbolic_start(self):
+        return self._start
+
+    @property
+    def symbolic_incr(self):
+        return self + self.offset
+
+    @property
+    def _properties(self):
+        return (self._start, self._offset)
```
mloubout: Just a comment, but we start to have a looot of different Dimensions type that may be a bit confusing for new users. Not sure if we can merge some.

devito/equation.py
mloubout: Cool great

devito/equation.py
```diff
@@ -64,3 +64,24 @@ def __eq__(self, other):
...
+    :param eq: The :class:`sympy.Eq` to be rearranged.
+    :param target: The symbol w.r.t. which the equation is rearranged.
+    :param kwargs: (Optional) Symbolic optimizations applied while rearranging
+                   the equation. For more information. refer to
+                   ``sympy.solve.__doc__``.
+    """
+    # Enforce certain parameters to values that are known to guarantee a quick
+    # turnaround time
+    kwargs['rational'] = False
+    kwargs['simplify'] = False
```
mloubout: Maybe add in the comment why these two?
rational => not float indices
simplify => very slow for complex equation

devito/function.py
```diff
@@ -590,10 +589,19 @@ class TimeFunction(Function):
...
+                 shape (3, 100), then, in an :class:`Operator`, ``t`` will
+                 assume the values ``1, 2, 0, 1, 2, 0, 1, ...`` (note that the
+                 very first value depends on the stencil equation in which
+                 ``u`` is written.). The default size of the time buffer when
+                 ``save=None`` is ``time_order + 1``.  To specify a different
+                 size for the time buffer, one should use the syntax
+                 ``save=Buffer(mysize)``. Alternatively, if all of the
+                 intermediate results are required (or, simply, to forbid the
+                 usage of an alternating buffer), an explicit value for ``save``
+                 (i.e., an integer) must be provided.
```
mloubout: corresponding to the number of time step

devito/function.py
```diff
@@ -740,6 +747,26 @@ def dt2(self):
...
+    def _time_buffering_default(self):
+        return self._time_buffering and self.save != Buffer
+
+    def _arg_check(self, args, intervals):
+        super(TimeFunction, self)._arg_check(args, intervals)
+        key = args[self.name]
+        if self._time_buffering and self._time_size != key.shape[self._time_position]:
+            raise InvalidArgument("Expected `time_size=%d` for runtime "
+                                  "value `%s`, found `%d` instead"
+                                  % (self._time_size, self.name, key._time_size))
```
mloubout: Not for this PR, but these multiline error message do not process the `%` arguments correctly in a lot of cases

devito/dimension.py
FabioLuporini: we cannot and we shouldn't merge them. 

However, it's not all so bad. These ``UnboundedDimensions`` are not meant to be used by the user directly. They are rather created internally by Devito. So it shouldn't be a big deal. 

Other dimensions could be hidden as well (think about the discussion on ConditionalDimension the other day on Slack)

Not sure about SubDimensions

devito/function.py
FabioLuporini: I didn't know that but yes, we should definitely fix this

tjb900: Actually, I just hit a problem with this anyway:
`key._time_size` => `key.shape[self._time_position]`?

FabioLuporini: isn't it the same thing?

https://github.com/opesci/devito/blob/timeslot-avoidance/devito/function.py#L751

tjb900: It just doesn't match the comparison immediately above it, is all.
I don't know what the underlying error was in my case - it failed saying `key._time_size` was invalid before it could raise that exception.

tjb900: `key` was a `Data`, which appears to have been put there by `TensorFunction._arg_defaults`
and `Data` has no `_time_size`.

FabioLuporini:
addressed all comments except for the save/buffer thing. More info in the comments above and on slack

devito/function.py
```diff
@@ -1223,3 +1260,11 @@ def inject(self, field, expr, offset=0, u_t=None, p_t=None):
         rhs = prod(coeffs) * expr
         field = field.subs(dim_subs)
         return [Eq(field, field + rhs.subs(dim_subs))]
+
+
+# Additional Function-related APIs
+
+class Buffer(Tag):
+
```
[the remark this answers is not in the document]
FabioLuporini: I agree that if we decided to keep this mechanism, the name would need to change

FabioLuporini:
Merged.

Issue #596, opened 2018-06-12
collapse being applied to imperfect loop structure


tjb900:
```
from unittest.mock import patch
@patch("devito.dle.backends.parallelizer.Ompizer.COLLAPSE", 1)
def test_conditional_collapse():
    from devito import (Grid, TimeFunction, ConditionalDimension, Operator,
                        Eq, Inc, Function)

    grid = Grid(shape=(20, 20, 20))

    time_sub = ConditionalDimension(name='time_sub',
                                    parent=grid.time_dim, factor=2)

    u = TimeFunction(name='u', grid=grid, save=10, time_dim=time_sub,
                     space_order=0)
    i = Function(name='i', grid=grid, space_order=0)

    step = Inc(i, u)

    op = Operator([step])

    print(op)
```

yields, with `DEVITO_OPENMP=1`, something which makes C compilers sad:

```
  #pragma omp parallel
  {
    #pragma omp for collapse(3) schedule(static)
    for (int time = time_m, time_sub0 = (time)%(1); time <= time_M; time += 1, time_sub0 = (time)%(1))
    {
      if ((time)%(2) == 0)
      {
        struct timeval start_section0, end_section0;
        gettimeofday(&start_section0, NULL);
        for (int x = x_m; x <= x_M; x += 1)
        {
          for (int y = y_m; y <= y_M; y += 1)
          {
```

FabioLuporini:
probably we just miss one or two handlers here:
https://github.com/opesci/devito/blob/master/devito/ir/iet/visitors.py#L531

e.g., the `Conditional` one

I'll take a look ASAP !

tjb900:
Already on it!  PR to come shortly.

PR #597, opened 2018-06-12
Iterations containing conditionals should not be perfect

tjb900:
Proposed fix for #596 

I can't help but feel there should be a more general solution to this, but this can at least get the ball rolling.

### DIFF
--- a/devito/ir/iet/visitors.py
+++ b/devito/ir/iet/visitors.py
@@ -545,6 +545,9 @@ class IsPerfectIteration(Visitor):
             return False
         return all(self.visit(i, found=found, **kwargs) for i in o.children)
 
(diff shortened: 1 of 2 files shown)


FabioLuporini:
@tjb900 I just had my go here: https://github.com/opesci/devito/compare/conditional_perfect_iter#diff-1526813b3ca372b26a915777ac6e7f4fR548

tjb900:
Ah yeah, I hadn't grokked the logic fully.  If I understand it correctly, your change allows conditionals inside the innermost loop?

I realised that while I keep making these branches on my fork just to keep the main repo clean, I hadn't given you permission to push amendments.  You should have it now.

PR #598, opened 2018-06-12
Changed =='s to >='s in requirements.txt. Only specify distributed in…

ggorman:
… requirements; distributed pulls in the approperiate versions of dask and tornado.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -5,20 +5,20 @@ channels:
 dependencies:
   - python>=3.6
   - numpy>=1.11
-  - sympy==1.1
+  - sympy>=1.1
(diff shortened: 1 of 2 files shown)


tjb900:
Sorry, I take that back - maybe you don't want to change environment.yml after all.  If people are creating an isolated environment for devito (and devito only), then I guess the == don't matter.

I'll leave it just to see what travis says, then revert.

FabioLuporini:
> I believe this is good to go once the tests pass.

guys go ahead and merge if you see this green before I come back here

mloubout:
merged

PR #599, opened 2018-06-12
Fablup/attack of the pickle

ggorman:


tests/test_pickle.py
```diff
@@ -0,0 +1,73 @@
+import numpy as np
+from examples.seismic import demo_model
+from examples.seismic.source import TimeAxis, RickerSource
+from devito.dimension import TimeDimension, SteppingDimension
+from devito.base import Constant
+
+import cloudpickle as pickle
+
+
+def test_pickle():
```
FabioLuporini: this needs to be split into multiple smaller tests

docstrings should be added

tests/test_pickle.py
```diff
@@ -0,0 +1,73 @@
...
+
+    ricker = RickerSource(name='src', grid=model.grid, f0=f0, time_range=time_range)
+
+    pkl_ricker = pickle.dumps(ricker)
+    new_ricker = pickle.loads(pkl_ricker)
+    assert np.isclose(np.linalg.norm(ricker.data.flatten()),
+                      np.linalg.norm(new_ricker.data.flatten()))
+
+
+if __name__ == "__main__":
```
FabioLuporini: useless

FabioLuporini:
we also need to figure out how to retrieve the base pointer of `Function._data` upon unpickling

one possibility emerged during offline discussion is to provide a placeholder object to let the receiver fetch the data directly from disk ()

FabioLuporini:
one last thing: this might require further coding to work with YASK

devito/function.py
```diff
@@ -566,6 +566,9 @@ def laplace2(self, weight=1):
         return sum([second_derivative(first * weight, dim=d, order=order)
                     for d in self.space_dimensions])
 
+    # Pickling support
+    _pickle_kwargs = TensorFunction._pickle_kwargs + ['grid']
+
```
tjb900: Presumably `shape` and `dimensions` too?  What about space order etc?

devito/tools/abc.py
```diff
@@ -84,3 +84,38 @@ def _signature_items(self):
...
+
+    _pickle_args = []
+    """The positional arguments that need to be passed to __new__ upon unpickling."""
+
+    _pickle_kwargs = []
+    """The keyword arguments that need to be passed to __new__ upon unpickling."""
+
+    def __reduce_ex__(self, proto):
+        return object.__reduce_ex__(self, proto)
+
```
tjb900: Unless we are expecting to need to add functionality (or debugging), `__reduce_ex__ = object.__reduce_ex__` ?

devito/function.py
```diff
@@ -566,6 +566,9 @@ def laplace2(self, weight=1):
...
 
 class TimeFunction(Function):
     """
```
tjb900: Will probably need to add `time_order`, `time_dim`, etc here also, I guess?

FabioLuporini: think so. @ggorman ?

FabioLuporini:
Merged.

PR #600, opened 2018-06-13
ensure correct types are passed to numa_free

tjb900:
Without this, we were passing wacky addresses and sizes to numa_free - which blindly unmaps virtual memory areas, or does nothing, depending on the phase of the moon.


FabioLuporini:
merged

PR #601, opened 2018-06-13
benchmark.py fixes and misc improvements

FabioLuporini:
now we have a unified interface for autotuning across ``core`` and ``yask`` (instead of two separate env vars DEVITO_AUTOTUNING and DEVITO_YASK_AUTOTUNING, which was just horrible)

@tjb900 I'd need you to check that AT is still working as expected with ``yask``. Not urgent though. 
EDIT: running after a call to ``mode_performance()`` should suffice


devito/__init__.py
```diff
@@ -109,3 +126,26 @@ def _switch_cpu(develop_mode):
 
 # Expose a mechanism to clean up the symbol caches (SymPy's, Devito's)
 clear_cache = CacheManager().clear  # noqa
+
+
+# Helper functions to switch on/off optimisation levels
+def mode_develop():
```
FabioLuporini: note for reviewers: these three functions have just been moved from parameters.py

FabioLuporini:
@tjb900 reminder: need to check that YASK AT still works as expected with this branch

devito/__init__.py
```diff
@@ -23,25 +25,40 @@
...
+# Autotuning setup
+AT_LEVELs = ['off', 'basic', 'aggressive']
+AT_MODEs = ['preemptive', 'runtime']
+at_default_mode = {'core': 'preemptive', 'yask': 'runtime'}
+at_setup = namedtuple('at_setup', 'level mode')
+at_accepted = AT_LEVELs + [list(i) for i in product(AT_LEVELs, AT_MODEs)]
+def _at_callback(val):  # noqa
+    if isinstance(val, str):
+        level, mode = val, at_default_mode[configuration['backend']]
+    else:
```
tjb900: Just a random thought - are the callbacks evaluated in a deterministic order?  Is `configuration['backend']` necessarily set to its proper value at this point?  What if it is changed?

FabioLuporini: uops, I had forgotten about this. 

Is this what you mean: https://github.com/opesci/devito/pull/601/files#diff-2d5ca91610163c3bffccbb775c230dc1R124 
?

FabioLuporini:
Merged.

PR #554, opened 2018-04-26
add test for time-dependent precomputed interpolation

tjb900:
It appears that `PrecomputedSparseFunction`, despite the docstring, does not support time-dependent functions.

I'm raising this as a PR instead of an issue just to get the ball rolling.

PR #603, opened 2018-06-20
Introduce SparseTimeFunction

navjotk:
This PR replaces #554, which will now be closed. 

It contains:
1. the test from #554 , which was failing there. 
2. A new class `PrecomputedSparseTimeFunction` which is the precomputed counterpart of `SparseTimeFunction`. 
3. With the introduction of the above class, the test now passes.

### DIFF
--- a/devito/__init__.py
+++ b/devito/__init__.py
@@ -10,8 +10,7 @@ from devito.data import *  # noqa
 from devito.dimension import *  # noqa
 from devito.equation import *  # noqa
 from devito.finite_difference import *  # noqa
-from devito.function import (PrecomputedSparseFunction, # noqa
-                             PrecomputedSparseTimeFunction, Buffer) # noqa
+from devito.function import Buffer # noqa
 from devito.logger import error, warning, info, set_log_level, silencio  # noqa
 from devito.parameters import *  # noqa
 from devito.tools import *  # noqa
--- a/devito/__init__.py
+++ b/devito/__init__.py
@@ -10,7 +10,8 @@ from devito.data import *  # noqa
 from devito.dimension import *  # noqa
 from devito.equation import *  # noqa
 from devito.finite_difference import *  # noqa
-from devito.function import Buffer  # noqa
+from devito.function import (PrecomputedSparseFunction, # noqa
+                             PrecomputedSparseTimeFunction, Buffer) # noqa
 from devito.logger import error, warning, info, set_log_level, silencio  # noqa
 from devito.parameters import *  # noqa
 from devito.tools import *  # noqa
--- a/devito/base.py
+++ b/devito/base.py
@@ -6,6 +6,7 @@ from sympy.core.compatibility import with_metaclass
 
 from devito.backends import _BackendSelector
 from devito.grid import Grid
+from devito.function import PrecomputedSparseTimeFunction
 import devito.types as types
 import devito.function as function
 import devito.operator as operator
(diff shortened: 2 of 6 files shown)


mloubout:
Are the coefficient time dependent too, like interpolating over both time and space or is it only a normal `SparseTimeFunction` with precomputed spatial coefficients?

navjotk:
The latter

devito/__init__.py
```diff
@@ -10,7 +10,8 @@
 from devito.dimension import *  # noqa
 from devito.equation import *  # noqa
 from devito.finite_difference import *  # noqa
-from devito.function import Buffer  # noqa
+from devito.function import (PrecomputedSparseFunction, # noqa
+                             PrecomputedSparseTimeFunction, Buffer) # noqa
```
FabioLuporini: this is not gonna be supported by YASK. We should never import from ``devito.function``, but rather from ``devito`` -- the backend mechanism does the rest. 

The new Function types must rather be added here: https://github.com/opesci/devito/blob/master/devito/base.py

tests/test_interpolation.py
```diff
@@ -93,6 +94,39 @@ def init(data):
     assert(all(np.isclose(sf.data, expected_values, rtol=1e-6)))
 
 
+@skipif_yask
```
FabioLuporini: @tjb900 will need to use these classes with YASK. Can we actually verify that they work as expected?

devito/__init__.py
navjotk: At this point, I'm curious what purpose this serves: https://github.com/opesci/devito/blob/master/devito/function.py#L23

tests/test_interpolation.py
navjotk: Uncommenting this, the test fails with the yask backend but that could be for a million different reasons. The [error messages](https://gist.github.com/navjotk/b2754585e734249c89d1e30a6a623406) are all cryptic so they're of no use. 

That's not the point of this PR anyway. There was a failing test submitted. This PR makes it pass.

devito/__init__.py
FabioLuporini: perhaps they're unnecessary

devito/base.py
```diff
@@ -39,6 +40,16 @@ class SparseTimeFunction(with_metaclass(_BackendSelector, function.SparseTimeFun
     pass
 
 
+class PrecomputedSparseFunction(with_metaclass(_BackendSelector,
+                                               function.PrecomputedSparseFunction)):
+    pass
+
+
+class PrecomputedSparseTimeFunction(with_metaclass(_BackendSelector,
+                                                   PrecomputedSparseTimeFunction)):
```
FabioLuporini: just function.PrecomputedSparseTimeFunction no? like all the others . Or is there a problem with that?

tests/test_interpolation.py
FabioLuporini: sure, no problem. Pity we can't use ``Precomputed*`` in YASK, though.

devito/base.py
navjotk: Flake cries about line length.

navjotk:
@FabioLuporini I added the classes to `base.py` as you asked. It's failing on something related to the backend voodoo. Can you take a look at the jenkins logs?

FabioLuporini:
> Can you take a look at the jenkins logs?

Sure. Pushed a fix.

FabioLuporini:
the ASCII diagram was wrong -- I pushed another commit that fixes it

FabioLuporini:
Merged.

PR #604, opened 2018-06-20
Operator caching

FabioLuporini:
An implementation of Operator caching, based on codepy's machinery, supported by both `core` and `yask`. The implementation is (supposed to be) race-free (ie, it should work across MPI processes, or parallel  sessions, or even different users sharing the same machine in parallel).

This PR also changes the way generated files (.c, .so) are stored within ``tmp`` (they now live within the same sub-directory, which doesn't change across consecutive runs)

I also slightly reorganised things within ``compiler.py``


mloubout:
Looks nice overall, and pretty handy for like TTI that do take a while to compile for gradients and stuff. As you said, need some testing to make sure for example that the lib name is unique in the correct way

FabioLuporini:
> As you said, need some testing to make sure for example that the lib name is unique in the correct way

But I'm not sure how to specifically test this -- as in, the fact that the test suite runs smooth is an indication that no name clashes occurred. In fact, the creation of `.{c,so}` is now even stricter than in master, as the alphanumeric string is not just a result of stringifying ``ccode``, but also includes the ``configuration`` dictionary.

devito/compiler.py
```diff
@@ -123,7 +123,15 @@ def __init__(self, **kwargs):
...
-        self.lib_ext = 'so'
+
+        if platform.system() == "Linux":
+            self.so_ext = '.so'
+        elif platform.system() == "Darwin":
+            self.so_ext = '.dylib'
+        elif platform.system() == "Windows":
+            self.so_ext = '.dll'
+        else:
+            raise NotImplementedError("Unsupported platform %s" % platform)
```
navjotk: This segment could be moved to a utility function in `tools/os_helper.py`

FabioLuporini:
Merged.

PR #605, opened 2018-06-20
Add test using gcc8 in Jenkins

tmbgreaves:
This PR:
- Updates Jenkinsfile to add a test using gcc8, cloned from the gcc5 section
- Updates devito/compiler.py to make '8' a valid suffix for gcc when specifying gcc version


tmbgreaves:
There shouldn't be any change - Jenkins should fire up five rather than four nodes, and the bottleneck remains gcc7+yask.

PR #609, opened 2018-07-02
Basic support for operator pickling

### LINKED ISSUE #186: Caching of generated code

FabioLuporini:
We'll need to break this through more aggressive testing and on-field experimentation (eg on clusters).

The good news is that both `core` and `yask` work, at least with fairly simple operators.

fixes #186


devito/yask/operator.py
```diff
@@ -186,6 +185,54 @@ def _compile(self):
...
+                f.write(upk_pyfile.decode())
+        pysofile = Path(namespace['yask-pylib'], '_%s.so' % state['yk-soln-name'])
+        if not pysofile.is_file():
+            with open(pysofile, 'wb') as f:
+                f.write(upk_pysofile)
+
+        # Finally reconstruct the YASK solution -- no code generation or jit
+        # will happen at this point, as all necessary files have been restored
+        local_grids = [i for i in self.parameters if i.is_Array]
+        context = contexts.fetch(self.input[0].grid, self._dtype)  # FIXME
```
FabioLuporini: hack, to be fixed another day I guess

FabioLuporini: fixed in #612

PR #613, opened 2018-07-06
Python-level MPI

FabioLuporini:
This PR, based on top of #612 (so don't panic when you see the `diff`, it's actually much smaller), adds MPI to our `Function`s. This is just Python-level MPI -- whenever a user accesses `f.data`, `data` is marked dirty and upon the next access, a halo exchange will be performed. 

New accessors are introduced, called `data_ro_*` -- these are read-only accessors, which can be used to avoid triggering (Python-level) halo exchanges

Tests show the correct behavior in a 2x2 cartesian grid of processes. Writing this type of tests is quite tedious, so contributions are welcome


devito/distributed.py
```diff
@@ -0,0 +1,115 @@
...
+        """
+        Return the mapper ``proc -> direction``; ``proc`` is the rank of a
+        neighboring process, while ``direction`` tells whether ``proc`` is
+        logically at right (value=1) or left (value=-1) of ``self``.
+        """
+        shifts = {d: self._comm.Shift(i, 1) for i, d in enumerate(self.dimensions)}
+        ret = {}
+        for d, (src, dest) in shifts.items():
+            ret[d] = {}
+            ret[d][LEFT] = src if src != -1 else None
```
mloubout: `src>-1`, can be lot more negative
All test pass with this change

devito/function.py
```diff
@@ -299,6 +341,46 @@ def symbolic_shape(self):
         return tuple(sympy.Add(i, -j, evaluate=False)
                      for i, j in zip(symbolic_shape, self.staggered))
 
+    def _halo_exchange(self):
+        """Perform the halo exchange with the neighboring processes."""
```
mloubout: neighboring processes or sub-domains (semantics can ignore)

devito/function.py
```diff
@@ -256,7 +264,15 @@ def data_with_halo(self):
...
 
         Elements are stored in row-major format.
+
+        .. note::
+
+            With this accessor you are claiming that you will modify
+            the values you get back. If you only need to look at the
+            values, use :meth:`data_ro_with_halo` instead.
         """
+        self._is_halo_dirty = True
```
mloubout: "dirty"?

devito/function.py
```diff
@@ -266,16 +282,42 @@ def data_allocated(self):
...
+        self._halo_exchange()
         return self._data
 
     @property
-    def halo(self):
-        return self._halo
+    @_allocate_memory
+    def data_ro_domain(self):
```
mloubout: why do we need read only?

devito/distributed.py
FabioLuporini: already dropped this, as you'll see in the following PR

devito/function.py
FabioLuporini: there are lots of flags attached to a Function, this one is about the halo, so I think we should keep "halo_dirty". 

Oh, maybe your question question was about the word "dirty" -- why "dirty". It's standard MPI halo exchange jargon: when you write to data, an update (through a halo exchange) is required so that the other processes have accesses to the most recent snapshot. Data that needs updating is commonly referred to as "dirty". Incidentally, this goes beyond MPI. It's for example used for cache lines too, to flag those that need to be committed back to DRAM.

devito/function.py
FabioLuporini: dropping this in the following PR. What I actually meant was "access to data without triggering a halo update". So you can safely ignore these methods for now

mloubout:
Merging

PR #617, opened 2018-07-18
grid as input for new model

mloubout:
Small extra accepted input for model to be able to create two models (v and v0 in `adjoint_J`) with the same grid.


examples/seismic/model.py
```diff
@@ -345,17 +345,19 @@ class Model(object):
...
         shape_pml = np.array(shape) + 2 * self.nbpml
         # Physical extent is calculated per cell, so shape - 1
         extent = tuple(np.array(spacing) * (shape_pml - 1))
-        self.grid = Grid(extent=extent, shape=shape_pml,
-                         origin=origin, dtype=dtype)
-
+        self.grid = kwargs.get('grid', Grid(extent=extent, shape=shape_pml,
+                                            origin=origin, dtype=dtype))
+        assert (self.grid.extent == extent)
+        assert (self.grid.shape == shape_pml).all()
```
FabioLuporini: is ``.all`` necessary?

mloubout: yep it is `shape_pml` is an array

examples/seismic/model.py
```diff
@@ -345,17 +345,23 @@ class Model(object):
...
         self.origin = tuple([dtype(o) for o in origin])
 
         shape_pml = np.array(shape) + 2 * self.nbpml
         # Physical extent is calculated per cell, so shape - 1
         extent = tuple(np.array(spacing) * (shape_pml - 1))
-        self.grid = Grid(extent=extent, shape=shape_pml,
-                         origin=origin, dtype=dtype)
-
+        # Check for input grid
+        self.grid = kwargs.get('grid', None)
```
FabioLuporini: the pythonic way is:
```
self.grid = kwargs.get('grid', Grid(extent=extent, shape=shape_pml, origin=origin, dtype=dtype))
```

mloubout: No, because some if `grid=None` in `kwargs` then `self.grid` is None

PR #620, opened 2018-07-25
Initial support for C-level MPI

FabioLuporini:
On top of a bunch of minor fixes (the most notable one being we are dropping the pointer casts hack in the generated code), this PR adds support for C-level MPI 

For an idea of what kind of code we can generate now, you could try running ``test_mpi::test_simple_operator``.
Of course, you'll need mpi4py and mpi on your system

unless DEVITO_MPI is set, the generated code should look more or less identical to current `master`. 

I don't need to remark that this is still WIP, and many things will have to be improved/fixed in the future


devito/dimension.py
```diff
@@ -517,45 +521,29 @@ class SteppingDimension(DerivedDimension):
...
     def symbolic_end(self):
-        """
-        The symbol defining the iteration end for this dimension.
-
-        note ::
-
-        Internally we always define symbolic iteration ranges in terms
-        of the parent variable.
-        """
         return self.parent.symbolic_end
```
mloubout: docstring gone?

devito/dse/backends/advanced.py
```diff
@@ -142,9 +142,8 @@ def _eliminate_inter_stencil_redundancies(self, cluster, template, **kwargs):
 
             # Build a symbolic function for /alias/
             intervals = ispace.intervals
-            shape = tuple(i.symbolic_size for i in indices)
             halo = [(abs(intervals[i].lower), abs(intervals[i].upper)) for i in indices]
-            function = Array(name=template(), shape=shape, dimensions=indices, halo=halo)
+            function = Array(name=template(), dimensions=indices, halo=halo)
```
mloubout: why i shape not necessary anymore?

devito/mpi/distributed.py
```diff
@@ -90,26 +97,47 @@ def dimensions(self):
...
+        """
+        A :class:`Object` wrapping an MPI communicator.
 
-class RankRelativePosition(Tag):
-    pass
+        Extracted from: ::
 
+            https://github.com/mpi4py/mpi4py/blob/master/demo/wrap-ctypes/helloworld.py
+        """
+        from devito.types import CompositeObject
```
mloubout: Import here?

devito/mpi/distributed.py
```diff
@@ -90,26 +97,47 @@ def dimensions(self):
...
+        obj = CompositeObject('comm', 'MPI_Comm', ptype, [])
+        comm_ptr = MPI._addressof(self._comm)
+        comm_val = obj.dtype.from_address(comm_ptr)
+        obj.value = comm_val
+        return obj
+
+    @cached_property
+    def _C_neighbours(self):
+        """A ctypes Struct to access the neighborhood of a given rank."""
+        from devito.types import CompositeObject
```
mloubout: same

devito/mpi/utils.py
```diff
@@ -0,0 +1,147 @@
...
+                        if rsize > 0:
+                            self._mapper.setdefault(f, []).append((d, RIGHT, rsize))
+                    else:
+                        # Found a serial dimension, we need to determine where,
+                        # along this dimension, the halo will have to be placed
+                        fixed = self._fixed.setdefault(f, OrderedDict())
+                        shift = int(any(d in dep.cause for dep in need_halo[f]))
+                        # Examples:
+                        # u[t+1, x] = f(u[t, x])   => shift == 1
+                        # u[t-1, x] = f(u[t, x])   => shift == 1
```
mloubout: -1?

devito/mpi/utils.py
```diff
@@ -0,0 +1,147 @@
...
+                        # u[t+1, x] = f(u[t+1, x]) => shift == 0
+                        # In the first and second cases, the x-halo should be inserted
+                        # at `t`, while in the last case it should be inserted at `t+1`
+                        if i.ispace.directions[r] is Forward:
+                            last = v[r].upper - shift
+                        else:
+                            last = v[r].lower + shift
```
mloubout: would put `+shift` everywhere with +-1 shifts personally

devito/symbolics/extended_sympy.py
```diff
@@ -58,6 +59,10 @@ class CondEq(sympy.Eq, FrozenExpr):
     def __new__(cls, *args, **kwargs):
         return sympy.Eq.__new__(cls, *args, evaluate=False)
 
+    @property
+    def canonical(self):
```
mloubout: what's this one?

devito/symbolics/extended_sympy.py
```diff
@@ -156,6 +163,49 @@ def base(self):
...
+    def field(self):
+        return self.function
+
+    @property
+    def composite(self):
+        return self.pointer
+
+    __repr__ = __str__
+
+
```
mloubout: Does these two really belong here?

devito/dimension.py
FabioLuporini: it's identical to the one in the superclass, so it gets inherited too

devito/dse/backends/advanced.py
FabioLuporini: an Array is a purely symbolic thing (e.g., it has associated **no** user data), so it's shape also is purely symbolic. And the symbolic shape can completely be inferred from the ``indices`` and the ``halo``. So shape was meaningless and redundant at the same time here

devito/mpi/distributed.py
FabioLuporini: there's a cyclic dependency that I haven't fixed yet, which appears if I move this to the top of the file. But this code is evolving rapidly, so..

devito/mpi/distributed.py
FabioLuporini: same

devito/mpi/utils.py
FabioLuporini: this is changed anyway by my next PR

devito/symbolics/extended_sympy.py
FabioLuporini: stuff that SymPy needs due to some weird internal mechanics

devito/symbolics/extended_sympy.py
FabioLuporini: wdym?

mloubout: FunctionFromPointer isn't a sympy utility, this would be more in function or utils ather than sympy extras no?

FabioLuporini: it is, it inherits from SymPy types

devito/mpi/utils.py
FabioLuporini: EDIT: in fact, I completely rewrote this piece of code (for the sake of clarity), so even better

FabioLuporini:
Merged.

PR #622, opened 2018-07-30
minor pickling updates

tjb900:
0. I've changed one of the pickling tests to use a non-default value for `save`, to illustrate the failure I am fixing.
1. To fix that now-broken test, `TimeFunction.save` is now pickled.  To facilitate that pickling, it is now saved in the instance as the original parameter, not the type.
2. The Pickable superclass is modified to be pickle instance attributes which have leading underscores, but pass them back to the constructor during unpickling _without_ the underscore.  This is to handle the case of e.g. Constant._value.
3. Constant now pickles its dtype and _value attributes.


FabioLuporini:
looks good to me.

@mloubout do you have any remarks about the ``self.save`` type->value flip?

mloubout:
>do you have any remarks about the self.save type->value flip?

I would rather keep it the way it was, the `type` only says whether or not the wavefield is `time` or `t` but saving the number of time_steps will force the creation of a new operator for every new number of time steps.

tjb900:
> I would rather keep it the way it was, the type only says whether or not the wavefield is time or t but saving the number of time_steps will force the creation of a new operator for every new number of time steps.

Oh, right!  So what determines which attributes of a Function are included in that hash?  It looks like this is what `Signer.__signature_items__` is designed to be for, but it isn't really used anywhere, yet...?

mloubout:
Actually `arg_check` was changed accordingly and correctly to only check the type instead of the value, so the beahavior should be the same as before just implemented a different way, there is a test with `time` overwrite (pretty sure) so if Green we should be good to go

FabioLuporini:
Merged.

PR #628, opened 2018-08-08
Interface enhancement

mloubout:
This PR does a few things:

-  Adds `.dx, .dy,....` shortcuts to staggered Functions
- reorganize FD to initialize Function shortcuts
- Add expression support for shortcuts (e.g `(u.dx +v).dy` is now possible)
- 3D elastic modeling


devito/finite_differences/finite_difference.py
```diff
@@ -239,59 +245,118 @@ def staggered_diff(f, dim, order, stagger=centered, theta=0, phi=0):
...
+    diff = dim.spacing
+    idx = list(set([(dim + int(i+.5+off)*diff)
+                    for i in range(-int(fd_order / 2), int(fd_order / 2))]))
+    if int(fd_order / 2) == 1:
+        idx = [dim - diff, dim]
+    deriv = f.diff(*(tuple(dim for _ in range(deriv_order))))
+    return deriv.as_finite_difference(idx, x0=dim + off*dim.spacing).evalf(_PRECISION)
+
+
+def initialize_derivatives(self):
```
FabioLuporini: self->function

devito/finite_differences/finite_difference.py
```diff
@@ -239,59 +245,118 @@ def staggered_diff(f, dim, order, stagger=centered, theta=0, phi=0):
...
+                      fd_order=max(int(self.space_order / 2), 2), stagg=side)
+        setattr(self.__class__, 'd%s4' % name,
+                property(dx4, 'Return the symbolic expression for '
+                         'the fourth derivative wrt. the '
+                         '%s dimension' % name))
+
+        # Cross derivatives, not supported in staggered grid
+        if any(self.staggered) is None:
+            for dim2 in self.space_dimensions:
+                name2 = dim2.parent.name if dim2.is_Derived else dim2.name
```
FabioLuporini: AKA:
name2 = dim2.root.name

devito/finite_differences/operations.py
```diff
@@ -0,0 +1,164 @@
...
+    sp_dims = ()
+    so = 100
+    to = 100
+    is_time = False
+    stagg = (0, 0, 0, 0)
+    # Filter expressions to get space and time Order
+    # The space order is the minimum space order
+    # The time order is the minimum time_order
+    for i in func:
+        if isinstance(i, devito.Function):
```
FabioLuporini: i.is_Function

devito/finite_differences/operations.py
```diff
@@ -0,0 +1,164 @@
...
+            sp_dims += i.space_dimensions
+            so = min(so, i.space_order)
+            stagg = i.staggered
+        if isinstance(i, devito.TimeFunction):
```
FabioLuporini: i.is_TimeFunction

devito/finite_differences/operations.py
```diff
@@ -0,0 +1,164 @@
+import sympy
+
+import devito
+from devito.symbolics.search import retrieve_functions
+from devito.symbolics.extended_sympy import FrozenExpr
+
+__all__ = ['Mul', 'Add', 'Pow']
+
+
+def fd_parameters(obj):
```
FabioLuporini: this looks like the constructor of a class, rather than a free function ...

devito/finite_differences/operations.py
```diff
@@ -0,0 +1,164 @@
...
+    is_Mul = True
+
+    def __new__(cls, *args, **kwargs):
+        return sympy.Mul.__new__(cls, *args)
+
+    def __init__(self, *args, **kwargs):
+        fd_parameters(self)
+        devito.finite_differences.finite_difference.initialize_derivatives(self)
+
+    def __add__(self, other):
```
FabioLuporini: are these __add__ etc really necessary? anyway, if so, you could put all these methods in the superclass I was mentioning earlier

examples/seismic/acoustic/operators.py
```diff
@@ -1,4 +1,4 @@
-from sympy import solve, Symbol
+from sympy import Symbol
 
 from devito import Eq, Operator, Function, TimeFunction, Inc
```
FabioLuporini: from devito import solve, ...

tests/test_derivatives.py
```diff
@@ -165,16 +165,19 @@ def test_fd_space(derivative, space_order):
...
+def test_fd_space_staggered(space_order, stagger, deriv_order):
     """
     This test compares the discrete finite-difference scheme against polynomials
     For a given order p, the finite difference scheme should
     be exact for polynomials of order p
     :param derivative: name of the derivative to be tested
     :param space_order: space order of the finite difference stencil
     """
+    if space_order == 2 and deriv_order == 2:
+        return True
```
FabioLuporini: ?

devito/finite_differences/operations.py
mloubout: This would fail, the only reason this weird loop is that in some case we end up with like `cos(theta)` which doesn't have a `is_Function`

devito/finite_differences/operations.py
mloubout: same

devito/finite_differences/operations.py
mloubout: Yes they are for FD. Maily `u.dx` becomes an `Add` object and needs to know how to`_add__` itself to erform `u.dx.dy` for example. Same for Mul (FD coeff) and Pow (1/hx)

examples/seismic/acoustic/operators.py
mloubout: fixed

tests/test_derivatives.py
mloubout: Stagerred FD only has 2 points for this case wich is not enough for 2nd order FD (needs at least 3)

devito/finite_differences/finite_difference.py
```diff
@@ -1,15 +1,17 @@
...
+from sympy import finite_diff_weights
 
-from devito.logger import error
+from devito.logger import error, debug
+from devito.finite_differences.operations import Add
 
 __all__ = ['first_derivative', 'second_derivative', 'cross_derivative',
            'generic_derivative', 'second_cross_derivative',
-           'left', 'right', 'centered', 'staggered_diff']
+           'left', 'right', 'centered', 'staggered_diff',
```
FabioLuporini: do we actually need ``left/right/center`` outside of this module?

devito/finite_differences/finite_difference.py
```diff
@@ -147,9 +149,9 @@ def cross_derivative(*args, **kwargs):
         for j in range(0, len(ind2r)):
             var1 = [a.subs({dims[0]: ind1r[i], dims[1]: ind2r[j]}) for a in args]
             var2 = [a.subs({dims[0]: ind1l[i], dims[1]: ind2l[j]}) for a in args]
-            deriv += (.5 * c11[i] * c12[j] * reduce(mul, var1, 1) +
-                      .5 * c21[-(j+1)] * c22[-(i+1)] * reduce(mul, var2, 1))
-
+            aux = (.5 * c11[i] * c12[j] * reduce(mul, var1, 1) +
+                   .5 * c21[-(j+1)] * c22[-(i+1)] * reduce(mul, var2, 1))
+            deriv = aux
```
FabioLuporini: why do u need an extra assignment?

devito/finite_differences/finite_difference.py
```diff
@@ -209,10 +211,16 @@ def generic_derivative(function, deriv_order, dim, fd_order):
     :param fd_order: Order of the coefficient discretization and thus
                      the width of the resulting stencil expression.
     """
+    indices = [(dim + i * dim.spacing) for i in range(-fd_order//2, fd_order//2 + 1)]
+    if fd_order == 1:
+        indices = [dim, dim + dim.spacing]
+    c = finite_diff_weights(deriv_order, indices, dim)[-1][-1]
+    deriv = Add(0)
+    for i in range(0, len(indices)):
+            var = [function.subs({dim: indices[i]})]
```
FabioLuporini: extra spaces

devito/finite_differences/finite_difference.py
```diff
@@ -174,7 +176,7 @@ def first_derivative(*args, **kwargs):
     order = int(kwargs.get('order', 1))
     matvec = kwargs.get('matvec', direct)
     side = kwargs.get('side', centered).adjoint(matvec)
-    deriv = 0
+    deriv = Add(0)
```
FabioLuporini: Add(0) makes no sense to me. We could have our own ``Zero`` or ``Number`` (sympy-inherited) class , alternatively

devito/finite_differences/finite_difference.py
```diff
@@ -209,10 +211,16 @@ def generic_derivative(function, deriv_order, dim, fd_order):
     :param fd_order: Order of the coefficient discretization and thus
                      the width of the resulting stencil expression.
     """
+    indices = [(dim + i * dim.spacing) for i in range(-fd_order//2, fd_order//2 + 1)]
+    if fd_order == 1:
+        indices = [dim, dim + dim.spacing]
+    c = finite_diff_weights(deriv_order, indices, dim)[-1][-1]
+    deriv = Add(0)
```
FabioLuporini: same as before

devito/finite_differences/finite_difference.py
```diff
@@ -224,10 +232,10 @@ def second_cross_derivative(function, dims, order):
     :param order: Discretisation order of the stencil to create.
     """
     first = second_derivative(function, dim=dims[0], width=order)
-    return second_derivative(first, dim=dims[1], order=order).evalf(_PRECISION)
+    return Add(second_derivative(first, dim=dims[1], order=order).evalf(_PRECISION))
```
FabioLuporini: feels like it's rather ``second_derivative`` that should already return an ``Add``

devito/finite_differences/finite_difference.py
```diff
@@ -224,10 +232,10 @@ def second_cross_derivative(function, dims, order):
...
 
 
-def staggered_diff(f, dim, order, stagger=centered, theta=0, phi=0):
+def staggered_diff(f, deriv_order, dim, fd_order, stagger=centered):
```
FabioLuporini: for consistency, ``dim`` should probably appear before ``deriv_order``

devito/finite_differences/finite_difference.py
```diff
@@ -239,59 +247,120 @@ def staggered_diff(f, dim, order, stagger=centered, theta=0, phi=0):
...
+    diff = dim.spacing
+    idx = list(set([(dim + int(i+.5+off)*diff)
+                    for i in range(-int(fd_order / 2), int(fd_order / 2))]))
+    if int(fd_order / 2) == 1:
+        idx = [dim - diff, dim]
+    deriv = f.diff(*(tuple(dim for _ in range(deriv_order))))
+    return deriv.as_finite_difference(idx, x0=dim + off*dim.spacing).evalf(_PRECISION)
+
+
+def initialize_derivatives(self):
```
FabioLuporini: and more importantly, I suggest to introduce a Mixin class, such as ``Differentiable(object)``  (name open to debate). Then, our Function/TimeFunction/... classes should inherit from ``Differentiable`` too, and should/could explicitly call ``Differentiable.initialize_derivatives`` at the end of their ``__init__``

devito/finite_differences/finite_difference.py
```diff
@@ -239,59 +247,120 @@ def staggered_diff(f, dim, order, stagger=centered, theta=0, phi=0):
...
+    """
+    Dynamically create notational shortcuts for space derivatives.
+    """
+    start_space = 0
+    if self.is_TimeFunction:
```
FabioLuporini: let's think about this: what makes ``time`` so special here that we have to do ``if is_TimeFunction``... can we generalise this?

FabioLuporini: OK, after some more thinking... how about we do the following instead. 

We keep ``initialize_derivatives`` a free function (ie, we don't put it in a mixin class). But now we let it accept two parameters, a ``dimension`` and an (approximation) ``order``. Isn't this the only piece of information that we actually need to create the shortcuts? Then, ``initialize_derivatives`` would either return a bag of functions/shortcuts that the caller adds to itself, or we could also pass the object (i.e., a Function) to which we want the newly-created shortcuts to be attached .

If we do that, we don't need to distinguish anymore between time and space, I think, which would be invaluable in terms of code organisation and clarity

Within ``Function/TimeFunction.__init__``, we could then have something along the lines of
```
for d in dimensions:
    initialize_derivatives(d, d_order, self)
```
or, equivalently:
```
for d in dimensions:
    for shortcut in generate_derivative_shortcuts(d, d_order):
         setattr(self, shortcut.__name__, shortcut)
```
or something like that...

thoughts? Am I off the track?

devito/finite_differences/finite_difference.py
```diff
@@ -239,59 +247,120 @@ def staggered_diff(f, dim, order, stagger=centered, theta=0, phi=0):
...
+    for (dim, s) in zip(self.space_dimensions, self.staggered[start_space:]):
+        deriv_function = staggered_diff if s is not None else generic_derivative
+        if s == 0:
+            side = left
+        elif s == 1:
+            side = right
         else:
-            return 0
+            side = centered
+        name = dim.parent.name if dim.is_Derived else dim.name
```
FabioLuporini: name = dim.root.name

devito/finite_differences/operations.py
```diff
@@ -0,0 +1,173 @@
...
+from devito.symbolics.search import retrieve_functions
+from devito.symbolics.extended_sympy import FrozenExpr
+
+__all__ = ['Mul', 'Add', 'Pow']
+
+
+def fd_parameters(obj):
+    """
+    Process a set of Mul, Add, Pow expression and outputs
+    the space_order, time_order of the whole expression.
```
FabioLuporini: again, let's think about this properly: do we actually need to distinguish between ``space`` and ``time``? and why?

devito/finite_differences/operations.py
```diff
@@ -0,0 +1,173 @@
...
+        devito.finite_differences.finite_difference.initialize_derivatives(self)
+
+    def __add__(self, other):
+        return Add(self, other)
+
+    def __iadd__(self, other):
+        return Add(self, other)
+
+    def __radd__(self, other):
+        return Add(self, other)
```
FabioLuporini: shouldn't you rather return ``Add(other, self)``?

devito/finite_differences/operations.py
```diff
@@ -0,0 +1,173 @@
...
+        return sympy.Pow.__new__(cls, *args, **kwargs)
+
+    def __init__(self, *args, **kwargs):
+        fd_parameters(self)
+        devito.finite_differences.finite_difference.initialize_derivatives(self)
+
+    def __add__(self, other):
+        return Add(self, other)
+
+    def __iadd__(self, other):
```
FabioLuporini: are you sure that this is typically what ``iadd`` does?

devito/finite_differences/operations.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+class Pow(sympy.Mul, FrozenExpr):
+    """A customized version of :class:`sympy.Add` representing a sum of
+    symbolic object."""
+    def __new__(cls, *args, **kwargs):
+        return sympy.Pow.__new__(cls, *args, **kwargs)
+
+    def __init__(self, *args, **kwargs):
+        fd_parameters(self)
```
FabioLuporini: exactly -- ``fd_parameters`` should become the constructor of a superclass from which our new ``Pow/Add/Mul/...`` inherit

devito/finite_differences/operations.py
```diff
@@ -0,0 +1,173 @@
...
+        return Mul(self, other)
+
+    def __imul__(self, other):
+        return Mul(self, other)
+
+    def __rmul__(self, other):
+        return Mul(self, other)
+
+    @property
+    def _time_size(self):
```
FabioLuporini: I don't like time-specific concept in an operation object

devito/finite_differences/operations.py
FabioLuporini: no I meant the fact that they are identical across Pow/Add/Mul, so they should be moved to a common superclass once and for all

devito/function.py
```diff
@@ -135,7 +132,7 @@ def __init__(self, *args, **kwargs):
             super(TensorFunction, self).__init__(*args, **kwargs)
 
             # Staggered mask
-            self._staggered = kwargs.get('staggered', tuple(0 for _ in self.indices))
+            self._staggered = kwargs.get('staggered', tuple(None for _ in self.indices))
```
FabioLuporini: I'm really struggling to understand why was this change necessary, which in turn requires a proliferation of
```
staggered = tuple([s if s is not None else 0 for s in self.staggered])
```
throughout the class, that I don't really like

devito/function.py
```diff
@@ -669,8 +625,7 @@ def laplace(self):
         derivative wrt. all spatial dimensions.
         """
         derivs = tuple('d%s2' % d.name for d in self.space_dimensions)
-
-        return sum([getattr(self, d) for d in derivs[:self.ndim]])
+        return Add(sum([getattr(self, d) for d in derivs[:self.ndim]]))
```
FabioLuporini: If it's a true Add, perhaps the ``sum`` isn't even necessary anymore?

devito/ir/support/utils.py
```diff
@@ -77,7 +77,7 @@ def build_iterators(mapper):
     iterators = OrderedDict()
     for k, v in mapper.items():
         for d, offs in v.items():
-            if d.is_Stepping:
+            if d.is_Stepping and k is not None:
```
FabioLuporini: not sure why ``k`` could be None ( I don't remember), but even then, I wonder whether we should rather have a
```
if k is None:
    continue
```
right below the ``for k, v ...`` loop?

devito/symbolics/extended_sympy.py
```diff
@@ -312,6 +304,7 @@ def eval_bhaskara_sin(expr):
 
 
 def eval_taylor_sin(expr):
+    from devito.finite_differences import Mul
```
FabioLuporini: to avoid all these local imports, I wonder whether we should leave Add/Mul/ etc in this module, but have them inherit from our own ``finite_difference.py``'s Add/Mul, rather than sympy's. I don't see why the Add/Mul/.. in ``finite_difference.py`` should inherit from ``FrozenExpr`` either

examples/seismic/acoustic/operators.py
```diff
@@ -44,7 +44,7 @@ def iso_stencil(field, m, s, damp, kernel, **kwargs):
     # Add dampening field according to the propagation direction
     eq += damp * field.dt if kwargs.get('forward', True) else -damp * field.dt
     # Solve the symbolic equation for the field to be updated
-    eq_time = solve(eq, next, rational=False, simplify=False)[0]
+    eq_time = solve(eq, next, rational=False, simplify=False)
```
FabioLuporini: do we still need rational/simplify?

examples/seismic/elastic/operators.py
```diff
@@ -1,23 +1,80 @@
...
 
-def ForwardOperator(model, source, receiver, space_order=4,
-                    save=False, **kwargs):
+def stress_fields(model, save, space_order):
     """
-    Constructor method for the forward modelling operator in an elastic media
+    Create the TimeFunction objects for the stress fields in the elastic formulation
+    """
+    if model.grid.dim == 2:
+        stagg_xx = stagg_zz = (0, 0, 0)
```
FabioLuporini: having to provide these 0-1 masks is very low level... can we not rather provide a tuple of dimensions, representing the staggered dimensions of the Function? then internally we translate this information into 0-1 bits. 
if we do that, perhaps to create these TimeFunctions we can getaway with a simple loop, rather than "if 1D do this, elif 2D do this, elif 3D do this, elif 4D do this, ..."

examples/seismic/elastic/operators.py
```diff
@@ -1,23 +1,80 @@
...
-    :param save: Saving flag, True saves all time steps, False only the three buffered
-                 indices (last three time steps)
+    return txx, tyy, tzz, txy, txz, tyz
+
+
+def particle_velocity_fields(model, save, space_order):
+    """
+    Create the particle velocity fields
+    """
+    if model.grid.dim == 2:
```
FabioLuporini: same as before

examples/seismic/elastic/operators.py
```diff
@@ -1,23 +1,80 @@
...
+
+
+def elastic_2d(model, space_order, save, source, receiver):
+    """
+    2D elastic wave equation FD kernel
     """
     vp, vs, rho, damp = model.vp, model.vs, model.rho, model.damp
     s = model.grid.stepping_dim.spacing
-    x, z = model.grid.dimensions
+    # x, z = model.grid.dimensions
```
FabioLuporini: leftover

examples/seismic/elastic/operators.py
```diff
@@ -26,56 +83,111 @@ def ForwardOperator(model, source, receiver, space_order=4,
...
+    if model.grid.dim == 2:
+        rec_expr = vx.dx + vz.dy
+    else:
+        rec_expr = vx.dx + vy.dy + vz.dz
+    rec_term2 = rec2.interpolate(expr=rec_expr, offset=model.nbpml)
+
+    return src_expr + rec_term1 + rec_term2
+
+
+def ForwardOperator(model, source, receiver, space_order=4,
```
FabioLuporini: remind me why we don't have a class like we have instead for acoustic and tti?

examples/seismic/elastic/wavesolver.py
```diff
@@ -72,28 +69,26 @@ def forward(self, src=None, rec1=None, rec2=None, vp=None, vs=None, rho=None,
...
+            kwargs['txy'] = txy
+            kwargs['tyz'] = tyz
         # Pick m from model unless explicitly provided
         vp = vp or self.model.vp
         vs = vs or self.model.vs
         rho = rho or self.model.rho
         # Execute operator and return wavefield and receiver data
-        summary = self.op_fwd(save).apply(src=src, rec1=rec1, vx=vx, vz=vz, txx=txx,
-                                          tzz=tzz, txz=txz, vp=vp, vs=vs, rho=rho,
+        self.op_fwd(save)._compile()
```
FabioLuporini: leftover

devito/finite_differences/operations.py
mloubout: Oh right, would be smarter yes

mloubout:
>we should get rid of the staggering 0-1 masks, and raise this concept to something more human-comprehensible

Definitely, on my todo list, the current way is efficient but quite confusing for normal users,

>we should also add tests for Add/Mul shortcuts


Yes, will look into it

>eventually (another PR, another day) a notebook showing how beautiful it has become to take derivatives would be quite handy

Yep would be nice

devito/finite_differences/finite_difference.py
mloubout: `first_derivative` is exported as a general purpose FD function and needs it so I'd say yes

devito/finite_differences/finite_difference.py
mloubout: typo, debug reminder

devito/finite_differences/finite_difference.py
mloubout: Need to initialize it as an `devito.Add` object or falls back to sympy one

devito/finite_differences/operations.py
mloubout: Probably not actually just another dimension so should be able to generate fd same way as space

devito/finite_differences/operations.py
mloubout: Nope, gotta check

devito/finite_differences/operations.py
mloubout: This is mainly needed for the compiler as now it gets a devito.Add/Mul/Pow object

devito/function.py
mloubout: `sum` is from Python not sympy, stilltrying to figure out where to overload it

devito/ir/support/utils.py
mloubout: `k` could be None due to the new ~Add,..` than have empty stuff in it in some cases

devito/symbolics/extended_sympy.py
mloubout: >I wonder whether we should leave Add/Mul/ etc in this module
Wouldn't it be a bit odd and confusing?

> I don't see why the Add/Mul/.. in finite_difference.py should inherit from FrozenExpr either

The way it was before

examples/seismic/acoustic/operators.py
mloubout: yes

examples/seismic/elastic/operators.py
mloubout: Exact same setup as acoustic and tti

devito/finite_differences/operations.py
FabioLuporini: then we have a bug in `retrieve_functions`, as I seem to remember it is only expected to fetch our own Function/TimeFunction.

Random guess. It's possible the bug is due to ``cos/sin/...`` being SymPy object on which ``is_Function == True`` (as these are indeed..."true" callable functions!). If that's the case, maybe we could fix ``retrieve_functions`` by looking for devito.Functions (``isinstance(...)``), rather than resorting to ``is_Function``.

devito/finite_differences/operations.py
FabioLuporini: where/how/when? can you give me an MFE w/o this property? we should really try our best to drop ``time`` from this place !

devito/function.py
FabioLuporini: ``Add`` should be able to take as input an arbitrary number of symbols/expressions (``*args``) and sum them up internally

devito/ir/support/utils.py
FabioLuporini: the keys in ``mapper`` are expected to be pure ``Function/TimeFunction/...``. We should have no ``Add/...`` at this stage. If we have them, that should be fixed.

examples/seismic/acoustic/operators.py
FabioLuporini: but aren't they automatically applied by our own ``solve``?

examples/seismic/elastic/operators.py
FabioLuporini: I thought we have a class in the other examples, in which ``ForwardOperator`` acted as a method. If it's a free function, then we should rather call it ``forward_operator`` (Python -- and not just Python tbh -- convention)

tests/test_derivatives.py
FabioLuporini: but it does when ``deriv_order == 1``?

tests/test_tti.py
```diff
@@ -124,9 +122,9 @@ def test_tti_staggered(shape):
...
 
-    u_staggered1 = u1.data[last, :] + v1.data[last, :]
-    u_staggered2 = u2.data[last, :] + v2.data[last, :]
-
-    res = np.linalg.norm(u_staggered1.reshape(-1) - u_staggered2.reshape(-1))
-    log("DSE/DLE introduces error %2.4e in %d dimensions" % (res, len(shape)))
-    assert np.isclose(res, 0.0, atol=1e-8)
+    res1 = np.linalg.norm(u1.data.reshape(-1) - u2.data.reshape(-1))
+    res2 = np.linalg.norm(v1.data.reshape(-1) - v2.data.reshape(-1))
+    log("DSE/DLE introduces error %2.4e, %2.4e in %d dimensions" %
```
FabioLuporini: since we're changing this, can we drop "DSE/DLE"? we shouldn't assume anything about the cause of the numerical error ...

mloubout: THe DSE/DLE is not for numerical error, this is the test to make sure dependency with openmp are processed correctly.

devito/finite_differences/operations.py
mloubout: Ok I'll look at it

tests/test_derivatives.py
mloubout: Yes `deriv_order == 1` is fine with 2 points

devito/finite_differences/operations.py
mloubout: Pobably

examples/seismic/acoustic/operators.py
mloubout: My bad you're right

examples/seismic/acoustic/operators.py
```diff
@@ -182,12 +181,14 @@ def BornOperator(model, source, receiver, space_order=4,
...
     # Create wavefields and a dm field
-    u = TimeFunction(name="u", grid=model.grid, save=None,
+    u = TimeFunction(name='u', grid=model.grid,
+                     save=source.nt if save else None,
                      time_order=2, space_order=space_order)
     U = TimeFunction(name="U", grid=model.grid, save=None,
                      time_order=2, space_order=space_order)
     dm = Function(name="dm", grid=model.grid, space_order=0)
 
+    print((u+U).dt)
```
navjotk: Is this a leftover?

devito/function.py
```diff
@@ -1110,7 +1040,7 @@ def coordinate_symbols(self):
     def coordinate_indices(self):
         """Symbol for each grid index according to the coordinates"""
         indices = self.grid.dimensions
-        return tuple([INT(sympy.Function('floor')((c - o) / i.spacing))
+        return tuple([INT(FLOOR((c - o) / i.spacing))
```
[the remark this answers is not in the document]
mloubout: No, may need to call each independently in some cases so prefer to keep it seperate.

devito/cgen_utils.py
```diff
@@ -174,6 +179,13 @@ def _print_IntDiv(self, expr):
     def _print_Byref(self, expr):
         return "&%s" % expr.name
 
+    def _print_Pow(self, expr):
+        base, exp = expr.as_base_exp()
+        if exp < 0:
+            return '1.0/' + self._print(pow_to_mul(base))
```
FabioLuporini: No. ``pow_to_mul`` should be changed to handle negative exponents and turn the Pow into something else, for example into an `IntDiv`

devito/dle/backends/basic.py
```diff
@@ -99,7 +100,6 @@ def _create_elemental_functions(self, nodes, state):
 
             call, params = zip(*args)
             name = "f_%d" % root.tag
-
```
FabioLuporini: nitpicking: there is a logic behind the blank spaces

devito/equation.py
```diff
@@ -80,8 +80,16 @@ def solve(eq, target, **kwargs):
...
     # Enforce certain parameters to values that are known to guarantee a quick
     # turnaround time
     kwargs['rational'] = False  # Avoid float indices
     kwargs['simplify'] = False  # Do not attempt premature optimisation
-    return sympy.solve(eq, target, **kwargs)[0]
+    # Reset to eqn = 0 if provided an Eq object
+    if isinstance(eq, Eq):
+        eq = eq.lhs - eq.rhs
+
+    if isinstance(eq, Differentiable):
```
FabioLuporini: do you actually need this ``if``? even though ``Differentiable``'s ``func`` is defined? that's weird

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,236 @@
+import sympy
+import numpy as np
+
+from devito.symbolics.search import retrieve_functions
+from devito.symbolics.extended_sympy import FrozenExpr
```
FabioLuporini: I still don't understand why we need this to be a FrozenExpr. I think it shouldn't be. Subclassing ``sympy.Expr`` should be enough.

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,236 @@
...
+__all__ = ['Differentiable']
+
+
+class Differentiable(FrozenExpr):
+    """
+    This class represents Devito differentiable objects such as
+    sum of functions, product of function or FD approximation and
+    provides FD shortcuts for such expressions
+    """
+    _op_priority = 100.0
```
FabioLuporini: probably something used by sympy. Needs a comment

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,236 @@
...
+    is_Function = True
```
FabioLuporini: is it a Function?

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,236 @@
...
+    is_Indexed = False
+
+    def __new__(cls, *args, **kwargs):
+        return sympy.Expr.__new__(cls, *args)
+
+    def __init__(self, expr, **kwargs):
+        from devito.finite_differences.finite_difference import generate_fd_functions
```
FabioLuporini: we don't need this to be local, I think

devito/equation.py
```diff
@@ -80,8 +80,16 @@ def solve(eq, target, **kwargs):
                    the equation. For more information. refer to
                    ``sympy.solve.__doc__``.
     """
+    from devito.finite_differences.differentiable import Differentiable
```
FabioLuporini: why local import? cyclic dependence?

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,236 @@
...
+        self.expr = expr.expr if isinstance(expr, Differentiable) else expr
+        self.indices = self._indices()
+        # Generate FD shortcuts for expression or copy from input
+        if isinstance(expr, Differentiable):
+            self.fd = expr.fd
+            # Recover the list of possible FD shortcuts
```
FabioLuporini: wrong comment?

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,236 @@
...
+            self.dtype = expr.dtype
+            self.space_order = expr.space_order
+            self.time_order = expr.time_order
+            self.staggered = expr.staggered
+        else:
+            # Recover the list of possible FD shortcuts
```
FabioLuporini: You mean "determine"?

FabioLuporini: move ``self.fd =...`` right below the comment?

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,236 @@
...
+            return Differentiable(sympy.Add(*[self.expr, other.expr], evaluate=False),
+                                  fd=self.fd)
+        else:
+            return Differentiable(sympy.Add(*[self.expr, other], evaluate=False),
+                                  fd=self.fd)
+
+    __radd__ = __add__
+
+    def __iadd__(self, other):
+        if isinstance(other, Differentiable):
```
FabioLuporini: instead of this painful to read cascade of
``if isinstance(other, Differentiable) ...``
I suggest to use a decorator like here:
https://github.com/opesci/devito/blob/master/devito/ir/support/basic.py#L63
in the decorator you can "normalize" ``other`` so that the method (__iadd__, __add__, and all others...) always gets ``expr``.
This little change should halve the size of this file :grimacing:

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,236 @@
...
+            return self.expr.args == other.args
+        else:
+            return sympy.simplify(self.expr) == other
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+    def func(self, *args, **kwargs):
+        kwargs.pop("evaluate", None)
+        if hasattr(self.expr, 'base'):
```
FabioLuporini: I think we should be more precise here. I don't like using ``hasattr``, it makes it feel like an hack

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,236 @@
...
+            return Differentiable(self.expr.func(*args, **kwargs), fd=self.fd)
+
+    def __str__(self):
+        return self.expr.__str__()
+
+    __repr__ = __str__
+
+    @property
+    def args(self):
+        if hasattr(self.expr, 'base'):
```
FabioLuporini: same

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,236 @@
...
+            return (self.expr,)
+        return self.expr.args
+
+    @property
+    def is_TimeFunction(self):
```
FabioLuporini: shouldn't we call it ``is_TimeExpr`` rather than ``is_TimeFunction`` ?

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,236 @@
...
+        """
+        Check wether or not time is an index
+        """
+        return any(i.is_Time for i in self.indices)
+
+    def _space_order(self):
+        """
+        Infer space_order from expression
```
FabioLuporini: then change method name to something else, such as ``_infer_space_order``

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,236 @@
...
+        """
+        func = list(retrieve_functions(self.expr))
+        order = 100
+        for i in func:
+            order = min(order, getattr(i, 'space_order', order))
+
+        return order
+
+    def _time_order(self):
```
FabioLuporini: same as before

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,236 @@
...
+        """
+        Check wether or not time is an index
+        """
+        return any(i.is_Time for i in self.indices)
+
+    def _space_order(self):
+        """
+        Infer space_order from expression
+        """
+        func = list(retrieve_functions(self.expr))
```
FabioLuporini: no need to cast to list?

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,236 @@
...
+        order = 100
```
FabioLuporini: don't like magic numbers. You know how to handle them.

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,236 @@
...
+        for i in func:
+            order = min(order, getattr(i, 'time_order', order))
+
+        return order
+
+    def _dtype(self):
```
FabioLuporini: _infer_dtype

FabioLuporini: also, the implementation of this method should be revisited. The Functions have a well-defined dtype, why assuming it be either float or double? how about complexes or float16? 
is this any useful: https://github.com/opesci/devito/blob/master/devito/ir/clusters/cluster.py#L87

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,236 @@
...
+        """
+        func = list(retrieve_functions(self.expr))
+        is_double = False
+        for i in func:
+            dtype_i = getattr(i, 'dtype', np.float32)
+            is_double = dtype_i == np.float64 or is_double
+
+        return np.float64 if is_double else np.float32
+
+    def _indices(self):
```
FabioLuporini: _infer_indices

devito/finite_differences/finite_difference.py
```diff
@@ -0,0 +1,401 @@
...
+    """
+    first = staggered_diff(expr, deriv_order=deriv_order[0],
+                           fd_order=fd_order[0], dim=dims[0], stagger=stagger[0])
+    return staggered_diff(first, deriv_order=deriv_order[1],
+                          fd_order=fd_order[1], dim=dims[1], stagger=stagger[1])
+
+
+def generate_fd_functions(function):
+    """
+    Create all possible FD functions from the indices,
```
FabioLuporini: rephrase this docstring?

devito/finite_differences/finite_difference.py
```diff
@@ -0,0 +1,401 @@
...
+    side = dict()
+    for (d, s) in zip(dimensions, function.staggered):
+        if s == 0:
+            side[d] = left
+        elif s == 1:
+            side[d] = right
+        else:
+            side[d] = centered
+
+    derivatives = ()
```
FabioLuporini: list

devito/finite_differences/finite_difference.py
```diff
@@ -0,0 +1,401 @@
...
+    done = ()
```
FabioLuporini: list

devito/finite_differences/finite_difference.py
FabioLuporini: edit: lists are mutable, tuples are immutable.
a list is the right fit here, because you're populating it

devito/finite_differences/finite_difference.py
```diff
@@ -0,0 +1,401 @@
...
+
+    for d in dimensions:
+        # Dimension is treated, remove from list
+        done += (d,)
+        other_dims = tuple(i for i in dimensions if i not in done)
+        # Dimension name and corresponding FD order
+        dim_order = time_fd_order if d.is_Time else space_fd_order
+        name = 't' if d.is_Time else d.root.name
```
FabioLuporini: d.name, not `'t'`

devito/finite_differences/finite_difference.py
```diff
@@ -0,0 +1,401 @@
...
+                            fd_order=dim_order, stagger=side[d])
+            name_fd = 'd%s%d' % (name, o) if o > 1 else 'd%s' % name
+            desciption = 'derivative of order %d w.r.t dimension %s' % (o, d)
+
+            derivatives += ((deriv, name_fd, desciption),)
+            # Cross derivatives with the other dimension
+            # Skip already done dimensions a dxdy is the same as dydx
+            for d2 in other_dims:
+                dim_order2 = time_fd_order if d2.is_Time else space_fd_order
+                name2 = 't' if d2.is_Time else d2.root.name
```
FabioLuporini: same

FabioLuporini: I think what you're after here is:
``d.name if d2.is_Stepping else d.root.name``
(same before)

devito/finite_differences/finite_difference.py
```diff
@@ -0,0 +1,401 @@
...
+            derivatives += ((deriv, name_fd, desciption), )
+            # right
+            deriv = partial(first_derivative, order=o,
+                            dim=d, side=right)
+            name_fd = 'd%sr' % name
+            desciption = 'left first order derivative w.r.t dimension %s' % d
+            derivatives += ((deriv, name_fd, desciption), )
+    else:
+        for d in dimensions:
+            name = 't' if d.is_Time else d.root.name
```
FabioLuporini: 't'

devito/finite_differences/finite_difference.py
```diff
@@ -0,0 +1,401 @@
...
+            deriv = partial(first_derivative, order=o,
+                            dim=d, side=left)
+            name_fd = 'd%sl' % name
+            desciption = 'left first order derivative w.r.t dimension %s' % d
+            derivatives += ((deriv, name_fd, desciption), )
+            # right
+            deriv = partial(first_derivative, order=o,
+                            dim=d, side=right)
+            name_fd = 'd%sr' % name
+            desciption = 'left first order derivative w.r.t dimension %s' % d
```
FabioLuporini: right, not left

devito/finite_differences/finite_difference.py
```diff
@@ -0,0 +1,401 @@
...
+                    desciption += 'w.r.t dimension (%s, %s) ' % (d, d2)
+                    derivatives += ((deriv, name_fd2, desciption), )
+
+    # add non-conventional, non-centered first-order FDs
+    if not is_staggered:
+        for d in dimensions:
+            name = 't' if d.is_Time else d.root.name
+            o = time_fd_order if d.is_Time else space_fd_order
+            # left
+            deriv = partial(first_derivative, order=o,
```
FabioLuporini: single line?

devito/function.py
FabioLuporini: reminder

devito/function.py
```diff
@@ -456,6 +456,56 @@ def _arg_check(self, args, intervals):
     # Pickling support
     _pickle_kwargs = AbstractCachedFunction._pickle_kwargs + ['staggered']
 
+    def __add__(self, other):
```
FabioLuporini: maybe we don't need any of these methods if we make TensorFunction subclass Differentiable? because that's what we probably want right?

devito/function.py
```diff
@@ -669,19 +653,27 @@ def laplace(self):
         derivative wrt. all spatial dimensions.
         """
         derivs = tuple('d%s2' % d.name for d in self.space_dimensions)
-
         return sum([getattr(self, d) for d in derivs[:self.ndim]])
 
     def laplace2(self, weight=1):
```
FabioLuporini: this method, at this point, shouldn't belong to here

devito/function.py
```diff
@@ -669,19 +653,27 @@ def laplace(self):
...
-        first = sum([second_derivative(self, dim=d, order=order)
-                     for d in self.space_dimensions])
-        return sum([second_derivative(first * weight, dim=d, order=order)
-                    for d in self.space_dimensions])
+        derivs = tuple('d%s2' % d.name for d in self.space_dimensions)
+        return sum([getattr(self.laplace * weight, d) for d in derivs[:self.ndim]])
+
+    def getdiff(self, name):
+        """
+        Returns finite dofference aproximation 'name' if depends on that dimension
```
FabioLuporini: many typos

devito/function.py
```diff
@@ -669,19 +653,27 @@ def laplace(self):
...
+        or 0
+        """
+        if name in self.derivatives:
+            return getattr(self, name)
+        else:
+            warning("FD shortcut %s not found for Function %s " % (name, self) +
```
FabioLuporini: warning or raise an error?

devito/function.py
```diff
@@ -669,19 +653,27 @@ def laplace(self):
...
         """
-        order = self.space_order/2
-        first = sum([second_derivative(self, dim=d, order=order)
-                     for d in self.space_dimensions])
-        return sum([second_derivative(first * weight, dim=d, order=order)
-                    for d in self.space_dimensions])
+        derivs = tuple('d%s2' % d.name for d in self.space_dimensions)
+        return sum([getattr(self.laplace * weight, d) for d in derivs[:self.ndim]])
+
+    def getdiff(self, name):
```
FabioLuporini: maybe you actually want to override ``__getattr__``?
see for example here: https://github.com/opesci/devito/blob/master/devito/yask/data.py#L216

devito/function.py
```diff
@@ -1192,7 +1162,7 @@ def inject(self, field, expr, offset=0):
 
         # Substitute coordinate base symbols into the coefficients
         return [Inc(field.subs(vsub),
-                    field.subs(vsub) + expr.subs(subs).subs(vsub) * b.subs(subs))
+                    Add(field.subs(vsub) + expr.subs(subs).subs(vsub) * b.subs(subs)))
```
FabioLuporini: the question is: why do we need an ``Add`` here? shouldn't it be a Differentiable already?

devito/symbolics/inspection.py
```diff
@@ -1,6 +1,5 @@
-from collections import Counter
-
 from sympy import cos, sin
+from collections import Counter
```
FabioLuporini: there was a logic.

Counter is available in the standard Python lib; sympy is not.

I suggest to revert to what we had. This is how imports are organised in Devito

devito/symbolics/manipulation.py
```diff
@@ -108,13 +108,13 @@ def run(expr):
...
                     rebuilt = expr.func(*(other + replaced + unreplaced), evaluate=False)
                     return rebuilt, False
+
             return expr.func(*other, evaluate=False), False
 
     # Process the provided expressions
     for expr in as_tuple(exprs):
         assert expr.is_Equality
         root = expr.rhs
-
```
FabioLuporini: again

devito/symbolics/manipulation.py
```diff
@@ -157,16 +156,21 @@ def xreplace_indices(exprs, mapper, key=None, only_rhs=False):
...
-            return expr
+            out = expr
         else:
-            return sympy.Mul(*[base]*exp, evaluate=False)
+            out = Mul(*[base]*exp, evaluate=False)
     else:
-        return expr.func(*[pow_to_mul(i) for i in expr.args], evaluate=False)
+        out = expr.func(*[pow_to_mul(i) for i in expr.args], evaluate=False)
+
+    if isinstance(expr, Differentiable):
```
FabioLuporini: hack. At this point, ``expr`` should already be either a Differentiable or a pure expr.

devito/symbolics/manipulation.py
```diff
@@ -213,13 +217,15 @@ def indexify(expr):
     :class:`AbstractFunction` objects have been converted into :class:`Indexed`
     objects.
     """
+    expr = expr.expr if isinstance(expr, Differentiable) else expr
```
FabioLuporini: hack

devito/symbolics/search.py
```diff
@@ -94,10 +94,12 @@ def search(expr, query, mode='unique', visit='dfs', deep=False):
     """
     Interface to Search.
     """
+    from devito.finite_differences.differentiable import Differentiable
```
FabioLuporini: hack

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,236 @@
...
+    _op_priority = 100.0
+    is_Function = True
+    is_Indexed = False
+
+    def __new__(cls, *args, **kwargs):
+        return sympy.Expr.__new__(cls, *args)
+
+    def __init__(self, expr, **kwargs):
+        from devito.finite_differences.finite_difference import generate_fd_functions
+        self.expr = expr.expr if isinstance(expr, Differentiable) else expr
```
FabioLuporini: I think this is the source of most of the issues that I commented below (in fact, this one that you're reading is the very last comment that I wrote in this buck). Why do we need ``expr`` to be stored as a field? what's the symbolic difference between ``self``  and ``expr`` at this point?

devito/finite_differences/differentiable.py
mloubout: Will do, higher priority than sympy (sympy is 10) to force `radd,rmul,...`

devito/finite_differences/differentiable.py
mloubout: It's technically not, there for dse debug mostly not entirely sure it's needed

devito/finite_differences/differentiable.py
mloubout: cyclic dependency without it

devito/finite_differences/differentiable.py
mloubout: No mean recover from input expression

devito/finite_differences/differentiable.py
mloubout: Will look at it

devito/finite_differences/differentiable.py
mloubout: It kinda is, for some reason some `Indexed` object end up here and haven't figured out from where yet, should be here

devito/finite_differences/differentiable.py
mloubout: same

devito/finite_differences/differentiable.py
mloubout: I think called it `is_TimeFunction` for dse reasons

devito/finite_differences/differentiable.py
mloubout: Needs robustness added, smae for staggered that is incorrect for now. Will look at it

devito/finite_differences/finite_difference.py
mloubout: oki

devito/finite_differences/finite_difference.py
mloubout: name is `time` for time dimension and `dtime` would be weird

devito/function.py
mloubout: That's what I want yes, just haven't gotten a way to get the `hasattr/getattr` and commen stuff (`dtype, spcae_order,...)` to work properly yet

devito/function.py
mloubout: I tried, but some `sympy.Expr` properties are needed, and I ended up with circular infite calls (why the unused `get_diff`)

devito/function.py
mloubout: warning, assuming if don't exist doesn't depend on it so returns 0

devito/function.py
mloubout: Leftover from early stuff, shouldn't be needed

devito/symbolics/inspection.py
mloubout: Debug cos/sin reminders

devito/finite_differences/differentiable.py
FabioLuporini: Why do we need the custom xreplace? 

so, here's my take on this. Differentiable is only useful in user-land. Once we're compiling, we could even safely get rid of it. Correct me if I'm wrong.

mloubout: Yes, it's only for symbolic FD/manipulation, not required for the compiler

mloubout:
Multiple changes, unreadable, closing and will reopen new cleaner one

PR #631, opened 2018-08-24
compat with sympy 1.2 and requirements

### LINKED ISSUE #615: sympy 1.12 currently breaks devito

mloubout:
Update custom `Indexed` and `.dot` deprecation for symyp 1.2 

fixes #615

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 numpy==1.14
-sympy==1.2
+sympy>=1.1
 scipy
 mpi4py
 matplotlib
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 numpy==1.14
-sympy==1.1
+sympy==1.2
 scipy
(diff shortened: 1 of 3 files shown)


mloubout:
Freaking new version of `pip` is screwing stuff up

requirements.txt
```diff
@@ -1,5 +1,5 @@
 numpy==1.14
-sympy==1.1
```
FabioLuporini: the ``==`` was actually a hack given that 1.12 wasn't working. Can we rather have ``>= 1.1``? or does turning ``dot`` into ``*`` prevent this? I see travis is still having, so you can easily check this with another round of CI

mloubout: Not sure ,lemme try

mloubout: Looks like is fine with `>=`

FabioLuporini:
uncontroversial. Thanks for this, Mathias.

PR #629, opened 2018-08-09
MPI next

FabioLuporini:
A bunch of MPI-related features:

- introduce routines for global-to-local index translation
- enable iteration over grid sub-regions (ie, via x_m/x_M)
- add support for ``SubDimension``
- add support for subsampling in time (only minor changes needed)
- add preliminary support for ``Function(grid, dimensions, shape)`` (ie grid+dimensions)
- augment the test coverage (all of the above cases tested, plus some more extra MPI tests)

~EDIT, WARNING: huge diff as it depends on #627~

PR #633, opened 2018-08-31
MPI preliminaries for examples

FabioLuporini:
... and various changes to logging

Big diff as based on ``mpi-next`` (#629 )


devito/compiler.py
```diff
@@ -12,7 +12,7 @@
 from codepy.toolchain import GCCToolchain
 
 from devito.exceptions import CompilationError
-from devito.logger import log, warning
+from devito.logger import debug, warning
```
ggorman: Why the change from 'log' to 'debug'?

examples/seismic/acoustic/acoustic_example.py
```diff
@@ -2,34 +2,21 @@
...
 from devito.logger import info
-from devito import Constant
+from devito import Constant, Eq, Function, Operator
 from examples.seismic.acoustic import AcousticWaveSolver
 from examples.seismic import demo_model, TimeAxis, RickerSource, Receiver
 
 
-# Velocity models
-def smooth10(vel, shape):
+def smooth(dest, src):
```
ggorman: Calling this 'src' is *very* misleading. Call it 'kernel'.

ggorman: Basically its hard to find anything to like about this function - to be fair the original was also also far from ideal. Why has the radius of the smoothing kernel been dropped? Why are we not using a standard Gaussian smoothing filter (ie applying principle of least astonishment)?

devito/yask/operator.py
```diff
@@ -153,14 +153,12 @@ def apply(self, **kwargs):
...
                 toshare[v] = v.data
 
-        log("Running YASK Operator through Devito...")
         for i in self.yk_solns.values():
             i.pre_apply(toshare)
         arg_values = [args[p.name] for p in self.parameters]
         self.cfunction(*arg_values)
         for i in self.yk_solns.values():
             i.post_apply()
-        log("YASK Operator successfully run!")
```
mloubout: Why removed?

examples/seismic/acoustic/acoustic_example.py
```diff
@@ -70,8 +57,11 @@ def run(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
                             space_order=space_order, kernel=kernel,
                             constant=constant, **kwargs)
 
-    initial_vp = smooth10(solver.model.m.data, solver.model.shape_domain)
-    dm = np.float32(initial_vp**2 - solver.model.m.data)
+    # Smooth velocity
+    initial_vp = Function(name='ivp', grid=solver.model.grid, space_order=space_order)
```
mloubout: would still name it `v0`, `ivp` seems weird (nitpicking)

examples/seismic/acoustic/acoustic_example.py
mloubout: `src` is confusing, we already have actual source called like that.

examples/seismic/model.py
```diff
@@ -388,8 +388,44 @@ def initialize_function(function, data, nbpml):
...
+
+            lpad = o.left - glb_range.left if lslice == 0 else 0
+            rpad = o.right - (rslice - glb_range.right) if rslice == glb_shape[d] else 0
+
+            data_slices.append((lslice, rslice))
+            pad_widths.append((lpad, rpad))
+        except KeyError:
+            # Not a distributed dimension
+            data_slices.append((None, None))
+            pad_widths.append((0, 0))
```
mloubout: Comments on what this does?

devito/compiler.py
FabioLuporini: to avoid pollution to ``stdout``. I think the average user doesn't need to see a jit-compilation related string.

examples/seismic/model.py
```diff
@@ -443,11 +479,6 @@ def dtype(self):
         """
         return self.grid.dtype
 
-    @property
-    def shape_domain(self):
-        """Computational shape of the model domain, with PML layers"""
-        return tuple(d + 2*self.nbpml for d in self.shape)
-
```
mloubout: why is this dropped?

examples/seismic/acoustic/acoustic_example.py
FabioLuporini: the maximum radius based on the amount of halo available is automatically picked by ``src.avg``. Otherwise you (user) can pick your own, as long as you know what you're doing. 

As you said, I simply translated the original function into the Devito language, and generalised it a little bit. Nothing more nothing less

devito/yask/operator.py
FabioLuporini: too much stdout pollution

examples/seismic/acoustic/acoustic_example.py
FabioLuporini: it was called ``initial_vp``, happy to change it to v0 !

examples/seismic/model.py
FabioLuporini: Yes, I can elaborate with a better comment. Basically, it takes the entire array, as there's nothing to be distribtued along the given dimension

examples/seismic/model.py
FabioLuporini: now unused

examples/seismic/acoustic/acoustic_example.py
FabioLuporini: done

FabioLuporini:
Merged.

PR #641, opened 2018-09-07
MPI: initial support for SparseFunction and related operators

FabioLuporini:
The title is self-explanatory

At this stage I think it's more important to state what's **missing**:
* support for SparseFunction injection/interpolation when these operators have a stencil, implying that a same sparse point may need to be communicated to more than one MPI rank
* use cached_property for _dist_* methdos, as they are expensive. The cached value needs to be invalidated upon an access to ``coordinates.data``
* tests of the above features

Once we have all of the above, in theory we can run some fairly advanced examples (wave equations)

### DIFF
--- a/devito/function.py
+++ b/devito/function.py
@@ -155,7 +155,14 @@ class TensorFunction(AbstractCachedFunction, ArgProvider):
                 # Allocate memory and initialize it. Note that we do *not* hold
                 # a reference to the user-provided buffer
                 self._initializer = None
-                self.data_allocated[:] = initializer
+                if len(initializer) > 0:
+                    self.data_allocated[:] = initializer
+                else:
+                    # This is a corner case -- we might get here, for example, when
+                    # running with MPI and some processes get 0-size arrays after
+                    # domain decomposition. We touch the data anyway to avoid the
+                    # case ``self._data is None``
+                    self.data
             else:
                 raise ValueError("`initializer` must be callable or buffer, not %s"
                                  % type(initializer))
@@ -1375,9 +1382,6 @@ class SparseFunction(AbstractSparseFunction):
                                                 dtype=self.dtype, dimensions=dimensions,
                                                 shape=(self.npoint, self.grid.dim),
                                                 space_order=0, initializer=coordinates)
-                if self.npoint == 0:
-                    # Make sure self._data is not None
-                    self._coordinates.data
 
     @property
     def coordinates(self):
--- a/devito/function.py
+++ b/devito/function.py
@@ -1093,7 +1093,7 @@ class AbstractSparseFunction(TensorFunction):
         """Return a tuple of argument names introduced by this function."""
         return tuple([self.name] + [x for x in self._child_functions])
 
-    def is_owned(self, point):
+    def _is_owned(self, point):
         """Return True if ``point`` is in self's local domain, False otherwise."""
         point = as_tuple(point)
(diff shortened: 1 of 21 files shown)


devito/function.py
```diff
@@ -1033,15 +1042,43 @@ def _arg_check(self, args, intervals):
...
     """
     An abstract class to define behaviours common to any kind of sparse
     functions, whether using precomputed coefficients or computing them
     on the fly. This is an internal class only and should never be
     instantiated.
     """
-    # Symbols that are encapsulated within this symbol (e.g. coordinates)
-    _child_functions = []
+
+    _sparse_position = -1
```
mloubout: Bit hacky no?

devito/function.py
```diff
@@ -1182,29 +1365,34 @@ class SparseFunction(AbstractSparseFunction):
...
+                                                shape=(self.npoint, self.grid.dim),
+                                                space_order=0, initializer=coordinates)
 
     @property
     def coordinates(self):
         return self._coordinates
 
     @property
-    def coefficients(self):
+    def coordinate_data(self):
```
mloubout: coordinates

devito/function.py
```diff
@@ -1295,7 +1483,17 @@ def _interpolation_indices(self, variables, offset=0):
...
+
+    @property
+    def gridpoints(self):
+        if self.coordinates._data is None:
+            raise ValueError("No coordinates attached to this SparseFunction")
+        ret = []
+        for coords in self.coordinates.data:
+            ret.append(tuple(int(sympy.floor((c - o.data)/i.spacing.data)) for c, o, i in
+                             zip(coords, self.grid.origin, self.grid.dimensions)))
+        return ret
```
mloubout: When is this called? I guess at apply if tests passes.

devito/function.py
```diff
@@ -477,8 +479,8 @@ def _arg_defaults(self, alias=None):
         args = ReducerMap({key.name: self._data_buffer})
 
         # Collect default dimension arguments from all indices
-        for i, s, o, k in zip(self.indices, self.shape, self.staggered, key.indices):
-            args.update(i._arg_defaults(start=0, size=s+o, alias=k))
+        for i, s, o in zip(key.indices, self.shape, self.staggered):
+            args.update(i._arg_defaults(start=0, size=s+o))
```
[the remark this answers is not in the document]
FabioLuporini: I found out it was useless, so I removed it. This is because in the current implementation we work out aliases in `_arg_values`, and in turn `_arg_values` calls `_arg_defaults` either on `self` or `alias`. Explanation a bit contrived, but in practice it's much simpler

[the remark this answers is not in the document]
mloubout: Ok

devito/function.py
```diff
@@ -1033,15 +1035,43 @@ def _arg_check(self, args, intervals):
     _pickle_kwargs = Function._pickle_kwargs + ['time_order', 'save']
 
 
+class SubFunction(Function):
```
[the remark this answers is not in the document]
mloubout: Do we want it to be a `Function` or `TensorFunction`.

devito/function.py
mloubout: Yes I know, just saying hard_coding index position seems a bit dangerous, for time too, don't have a solution just food for thoughts

devito/function.py
mloubout: yes

PR #642, opened 2018-09-07
Interface2

mloubout:
Reopening interface enhancement, WIP but mostly there

- '(u + v.dx).dy' and things like that possible
- staggered no is defined as `staggered = x` for example, not more `(0,1,0)` and `None` corresponds to cartesian grids. Need to add check to make sure all fields are either staggered or non-staggered
- FD shortcuts are no only created if called to avoid hundreds if not thousands of calls to `setattr` during FD manipulation

- Some misc


examples/seismic/elastic/operators.py
```diff
@@ -26,56 +87,111 @@ def ForwardOperator(model, source, receiver, space_order=4,
...
+    src_expr = src_xx + src_zz
+    if model.grid.dim == 3:
+        src_yy = src.inject(field=tyy.forward, expr=src * s, offset=model.nbpml)
+        src_expr += src_yy
 
     # Create interpolation expression for receivers
-    rec_term1 = rec1.interpolate(expr=txx, offset=model.nbpml)
-    rec_term2 = rec2.interpolate(expr=tzz, offset=model.nbpml)
+    rec_term1 = rec1.interpolate(expr=tzz, offset=model.nbpml)
+    if model.grid.dim == 2:
```
FabioLuporini: this makes me think that we could to introduce the concept of ``VectorFunction``, so that we just carry ``v`` around. Then automatically ``v = [vx]``  in 1D, ``v = [vx, vy]`` in 2D, etc., which means that ``rec_expr`` (and many, many other operations typical of problem setup!) don't need to be specialised depending on whether it's 1/2/3D... it'd just be ``rec_expr = v.sum``

Of course this is just food for thought, no need to do anything in this PR which is already quite big

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def _merge_fd_properties(self, other):
+        merged = {}
+        merged["space_order"] = np.min([getattr(self, 'space_order', 100) or 100,
+                                        getattr(other, 'space_order', 100)])
+        merged["time_order"] = np.min([getattr(self, 'time_order', 100) or 100,
+                                       getattr(other, 'time_order', 100)])
+        merged["indices"] = tuple(set(self.indices + getattr(other, 'indices', ())))
+        merged["fd"] = dict(getattr(self, 'fd', {}), **getattr(other, 'fd', {}))
+        merged["staggered"] = self.staggered
```
FabioLuporini: should we not check that they are actually compatible? 
that's why I was wondering about the use of the word "merge" .

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,182 @@
...
+        return self._grid
+
+    @property
+    def dtype(self):
+        return self._dtype
+
+    def __getattr__(self, name):
+        if name == 'fd' or name == '_fd':
+            raise AttributeError()
+        if name in self.fd.keys():
```
FabioLuporini: reminder: no need for keys()

devito/grid.py
```diff
@@ -201,3 +201,28 @@ def __setstate__(self, state):
...
+
+def staggered(function):
+    """
+    For a given function and it's staggered property `staggered`
+    returns a tuple of 0 or 1 for each dimension:
+    0 if not staggered in the dimension
+    1 if staggered in the dimension
+    """
+    s = (lambda x: x if type(x) is tuple else (x,))(function.staggered)
+    if function.staggered in ['node', None]:
```
FabioLuporini: 'node' and 'cell' should be defined through ``Tag``. ``git grep Tag`` for some examples.

devito/function.py
```diff
@@ -538,7 +532,7 @@ def _arg_check(self, args, intervals):
         ['dtype', 'grid', 'staggered', 'initializer']
 
 
-class Function(TensorFunction):
+class Function(TensorFunction, Differentiable):
```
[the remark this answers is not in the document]
mloubout: doesn't call the `Differential` `super` so not sure

devito/function.py
```diff
@@ -596,18 +590,20 @@ class Function(TensorFunction):
 
     .. note::
 
-       The tuple :param staggered: contains a ``1`` in each dimension
-       entry that should be staggered, and ``0`` otherwise. For example,
-       ``staggered=(1, 0, 0)`` entails discretization on horizontal edges,
-       ``staggered=(0, 0, 1)`` entails discretization on vertical edges,
-       ``staggered=(0, 1, 1)`` entails discretization side facets and
-       ``staggered=(1, 1, 1)`` entails discretization on cells.
+       The dimension or String :param staggered: defines the dimension in which the
```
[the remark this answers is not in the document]
mloubout: `Dimension`

devito/function.py
```diff
@@ -596,18 +590,20 @@ class Function(TensorFunction):
...
+       ``staggered=(x, y)`` entails discretization xy facets and
+       ``staggered='node'`` entails discretization on node
+       ``staggerd='cell'`` entails discretization on cell.
     """
 
     is_Function = True
 
     def __init__(self, *args, **kwargs):
         if not self._cached():
+            self._kwargs = kwargs
```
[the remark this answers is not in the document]
mloubout: gone, leftover

devito/operator.py
```diff
@@ -71,7 +72,7 @@ def __init__(self, expressions, **kwargs):
         self._func_table = OrderedDict()
 
         # Expression lowering: indexification, substitution rules, specialization
-        expressions = [indexify(i) for i in expressions]
+        expressions = [indexify(to_expr(i)) for i in expressions]
```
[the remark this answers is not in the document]
mloubout: No need for `Differential` anymore in the compiler, so get sympy expr only at this point

devito/symbolics/extended_sympy.py
```diff
@@ -109,6 +109,29 @@ def __str__(self):
     __repr__ = __str__
 
 
+class ExprDiv(sympy.Expr):
```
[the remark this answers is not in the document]
mloubout: `IntDiv` uses `Integer` to enforce integer indices for subsampling, `ExprDiv` is general expression and `Integer(Expr)` does not exist

examples/seismic/elastic/operators.py
mloubout: >rec_expr = v.sum

Not thecase in 3D, only `vx+vz` `vy` is ignored. 

>`VectorFunction`

That would be great but requires (estimates :D) same size PR by itself

examples/seismic/elastic/wavesolver.py
```diff
@@ -72,28 +69,25 @@ def forward(self, src=None, rec1=None, rec2=None, vp=None, vs=None, rho=None,
...
+            kwargs['txy'] = txy
+            kwargs['tyz'] = tyz
         # Pick m from model unless explicitly provided
         vp = vp or self.model.vp
         vs = vs or self.model.vs
         rho = rho or self.model.rho
         # Execute operator and return wavefield and receiver data
-        summary = self.op_fwd(save).apply(src=src, rec1=rec1, vx=vx, vz=vz, txx=txx,
-                                          tzz=tzz, txz=txz, vp=vp, vs=vs, rho=rho,
+        summary = self.op_fwd(save).apply(src=src, rec1=rec1, vp=vp, vs=vs, rho=rho,
```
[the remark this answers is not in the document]
mloubout: `if 2d/3d` for the old way but can change

examples/seismic/utils.py
```diff
@@ -0,0 +1,29 @@
+import numpy as np
+
+
+__all__ = ['smooth10']
+
+
+# Velocity models
+def smooth10(vel, shape):
```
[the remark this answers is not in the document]
mloubout: yep

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def __init__(self, expr, **kwargs):
+        # Recover the list of possible FD shortcuts
+        self._dtype = kwargs.get('dtype')
+        self._space_order = kwargs.get('space_order')
+        self._time_order = kwargs.get('time_order', 0)
+        self._indices = kwargs.get('indices', ())
+        self._staggered = kwargs.get('staggered')
+        self._grid = kwargs.get('grid')
+        # Generate FD shortcuts for expression or copy from input
```
FabioLuporini: obsolete comment?

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,182 @@
...
+            assert len(args) == 1
+            expr = args[0]
+            new_obj = sympy.Expr.__new__(cls, expr)
+            # Initialization
+            new_obj.__init__(expr, **kwargs)
+            return new_obj
+        return sympy.Expr.__new__(cls, *args)
+
+    def __init__(self, expr, **kwargs):
+        # Recover the list of possible FD shortcuts
```
FabioLuporini: obsolete comment?

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,182 @@
...
+        self._dtype = kwargs.get('dtype')
+        self._space_order = kwargs.get('space_order')
+        self._time_order = kwargs.get('time_order', 0)
+        self._indices = kwargs.get('indices', ())
+        self._staggered = kwargs.get('staggered')
+        self._grid = kwargs.get('grid')
+        # Generate FD shortcuts for expression or copy from input
+        self._fd = kwargs.get('fd', {})
+        # Save kwargs
```
FabioLuporini: obsolete comment?

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,155 @@
...
+        self._expr = expr
+
+    def __getattr__(self, name):
+        if name == 'fd':
+            raise AttributeError()
+        if name in self.fd.keys():
+            return self.fd[name][0](self)
+        return self.__getattribute__(name)
+
+    def xreplace(self, rule):
```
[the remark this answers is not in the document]
FabioLuporini: then get _expr and call the superclass' xreplace via super?

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,182 @@
...
+        merged["staggered"] = self.staggered
+        merged["dtype"] = self.dtype
+        return merged
+
+    def __add__(self, other):
+        return Differentiable(sympy.Add(*[getattr(self, '_expr', self),
+                                          getattr(other, '_expr', other)]),
+                              **self._merge_fd_properties(other))
+
+    __iadd__ = __add__
```
FabioLuporini: I'm not sure __iadd__ and __radd__ should be the same as __add__. Is it what happens in sympy typically? __iadd__ is += 

same for sub, mul etc

devito/symbolics/extended_sympy.py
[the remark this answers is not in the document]
FabioLuporini: ah, right, I was confused by the docstring which says "a support type for integer division..."

examples/seismic/elastic/wavesolver.py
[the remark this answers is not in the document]
FabioLuporini: can you not just provide None for the unused Functions? does the Operator complain?

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,182 @@
...
+    sum of functions, product of function or FD approximation and
+    provides FD shortcuts for such expressions
+    """
+    _op_priority = 100.0
+
+    def __new__(cls, *args, **kwargs):
+        if cls == Differentiable:
+            assert len(args) == 1
+            expr = args[0]
+            new_obj = sympy.Expr.__new__(cls, expr)
```
FabioLuporini: out of curiosity: what happens if you rather do:
```sympy.Expr.__new__(sympy.Expr, expr)```
and then
```new_obj._space_order = ...
new_obj._dtype = ...
...
return new_obj
```
do we get "true" SymPy objects but with additional attributes? or do you actually get exceptions when trying to add new fields?

FabioLuporini: most likely an exception . Scratch this.

devito/finite_differences/differentiable.py
mloubout: Sympy expressions are immutable, and operation on it should return a new one.

examples/seismic/elastic/wavesolver.py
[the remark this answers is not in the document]
mloubout: Yes it does, `Operator` complains for any kwargs not in `arguments` no matter the value assigned to it.

devito/finite_differences/differentiable.py
FabioLuporini: it appears that += actually has an effect:
```
In [1]: from sympy import *
...
In [5]: a=Symbol('a')
In [6]: b=Symbol('b')
In [7]: a + b
Out[7]: a + b
In [8]: a += b
In [9]: a
Out[9]: a + b
```
note that the content of ``a`` has actually changed

as for __radd__, __rmul__, __rdiv__ ... we should swap the operands. See for example here: https://github.com/sympy/sympy/blob/master/sympy/core/expr.py#L124
not sure, however, what are the practical implications of this in SymPy

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def __rtruediv__(self, other):
+        return Differentiable(sympy.Mul(*[getattr(other, '_expr', other),
+                                          getattr(self, '_expr', self)**(-1)]),
+                              **self._merge_fd_properties(other))
+
+    __floordiv__ = __truediv__
+    __rdiv__ = __rtruediv__
+    __div__ = __truediv__
+    __rfloordiv__ = __rtruediv__
```
FabioLuporini: Note sure where to leave this comment... anyway...
we're missing __mod__, __int__, and __float__ at least

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,189 @@
...
+    provides FD shortcuts for such expressions
+    """
+    # Set the operator priority higher than Sympy (10.0) to force the overloaded
+    # operators to be used
+    _op_priority = 100.0
+
+    def __new__(cls, *args, **kwargs):
+        if cls == Differentiable:
+            assert len(args) == 1
+            expr = args[0]
```
FabioLuporini: I have an idea, I think. You can check here whether:
```
if expr.is_Function
```
in which case, you just
```
    return expr
```
this should avoid the proliferation of ``if self.is_Function`` in the various methods

devito/grid.py
```diff
@@ -201,3 +201,28 @@ def __setstate__(self, state):
         for k, v in state.items():
             setattr(self, k, v)
         self._distributor = Distributor(self.shape, self.dimensions)
+
+
+def staggered(function):
```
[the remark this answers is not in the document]
FabioLuporini: I suggest we move this function into ``TensorFunction``, and turn it into a (overridable, should we need to do something different for SparseFunctions...) ``__staggered_setup__`` method.

devito/grid.py
```diff
@@ -201,3 +201,28 @@ def __setstate__(self, state):
...
+    returns a tuple of 0 or 1 for each dimension:
+    0 if not staggered in the dimension
+    1 if staggered in the dimension
+    """
+    s = (lambda x: x if type(x) is tuple else (x,))(function.staggered)
+    if function.staggered in ['node', None]:
+        return tuple(0 for _ in function.indices)
+    if function.staggered == 'cell':
+        return tuple(1 for _ in function.indices)
+    staggered = ()
```
FabioLuporini: BTW, this should be a list...

devito/grid.py
```diff
@@ -201,3 +201,28 @@ def __setstate__(self, state):
...
+    s_neg = tuple([-ss for ss in s])
+    for d in function.indices:
+        if d in s:
+            staggered += (1,)
+        elif d in s_neg:
+            staggered += (-1,)
+        else:
+            staggered += (0,)
+
+    return staggered
```
FabioLuporini: ...and casted to tuple here

devito/operator.py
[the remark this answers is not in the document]
FabioLuporini: would things break without this ``to_expr`` ?

devito/symbolics/extended_sympy.py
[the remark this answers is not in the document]
FabioLuporini: meaning that we should update the docstring

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,189 @@
...
+            return sympy.Number(1)
+
+    def __neg__(self):
+        return Differentiable(sympy.Mul(*[getattr(self, '_expr', self), -1]),
+                              **self._merge_fd_properties(None))
+
+    def __str__(self):
+        if self.is_Function:
+            return super(Differentiable, self).__str__()
+        return to_expr(self._expr).__str__()
```
FabioLuporini: perhaps we don't need to wrap it with ``to_expr``

devito/types.py
```diff
@@ -606,7 +606,9 @@ def indexify(self, indices=None):
         if indices is not None:
             return Indexed(self.indexed, *indices)
 
-        subs = dict([(i.spacing, 1) for i in self.indices])
+        # Only replace spacing -> 1 if used as index
```
FabioLuporini: why this change?

devito/finite_differences/differentiable.py
mloubout: Hum odd, I can rewrite `iadd,imul,..` to just change `_expr` in place and keep same object, fine with me.

for the `r` ones, additions are commutative so doesn't really matter but can change it.

devito/finite_differences/differentiable.py
mloubout: probably, was getting some `Differentiable` in the str at some point so had to put it.

devito/grid.py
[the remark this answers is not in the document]
mloubout: Will think if not too intricated.

devito/operator.py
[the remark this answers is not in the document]
mloubout: Yes, was doing werid stuff without it

devito/types.py
mloubout: I "fixed" `SparseFunction` tthat was forcing `indexify` in inject/interpolate to make this function happy.

devito/function.py
```diff
@@ -46,7 +43,7 @@ class Constant(AbstractCachedSymbol):
 
     def __init__(self, *args, **kwargs):
         if not self._cached():
-            self.dtype = kwargs.get('dtype', np.float32)
+            self._dtype = kwargs.get('dtype', np.float32)
```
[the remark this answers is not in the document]
mloubout: `AbstractCachedSymbol` doesn't have a dtype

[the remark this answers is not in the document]
FabioLuporini: But types.Symbol and types.Scalar do. We should then factor this into ``AbstractCachedSymbol``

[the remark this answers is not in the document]
mloubout: We can't, `Scalar` defaults to `float` while `Symbol` defaults to `int`, would be a huge diggin to track everywhere it's used to make it work and this PR already big enough.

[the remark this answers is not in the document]
FabioLuporini: no worries, I'm touching those classes in another PR, I'll find a way later on

devito/finite_differences/differentiable.py
```diff
@@ -0,0 +1,237 @@
...
+        return Differentiable(sympy.Mul(*[getattr(self, '_expr', self),
+                                          getattr(other, '_expr', other)]),
+                              **self._merge_fd_properties(other))
+
+    def __imul__(self, other):
+        self._expr = sympy.Mul(*[getattr(self, '_expr', self),
+                                 getattr(other, '_expr', other)])
+        return self
+
+    __rmul__ = __mul__
```
FabioLuporini: (maybe nitpicking, maybe not): floating point operations are not commutative

devito/types.py
```diff
@@ -369,6 +370,11 @@ def __init__(self, *args, **kwargs):
 
     __hash__ = Cached.__hash__
 
+    @classmethod
+    def __staggered_setup__(cls, obj, **kwargs):
+        """Extract the object indices from ``kwargs``."""
```
FabioLuporini: update pasted docstring

devito/function.py
```diff
@@ -193,6 +187,40 @@ def __dtype_setup__(cls, **kwargs):
...
+                staggered += [1]
+            elif d in s_neg:
+                staggered += [-1]
+            else:
+                staggered += [0]
+
+        return tuple(staggered)
+
+    @property
+    def staggered(self):
```
FabioLuporini: this should go in ``AbstractCachedFunction``

devito/function.py
```diff
@@ -193,6 +187,40 @@ def __dtype_setup__(cls, **kwargs):
...
+    @classmethod
+    def __staggered_setup__(cls, func, **kwargs):
+        """
+        For a given function and it's staggered property `staggered`
+        returns a tuple of 0 or 1 for each dimension:
+        0 if not staggered in the dimension
+        1 if staggered in the dimension
+        """
+        stagger = kwargs.get('staggered')
+        s = (lambda x: x if type(x) is tuple else (x,))(stagger)
```
FabioLuporini: perhaps ``as_tuple`` helps here?

devito/function.py
```diff
@@ -193,6 +187,40 @@ def __dtype_setup__(cls, **kwargs):
...
+        if stagger is None:
+            func.is_Staggered = False
+            return tuple(0 for _ in func.indices)
+        func.is_Staggered = True
+        if stagger == 'node':
```
FabioLuporini: is this just a special case of the loop below?

devito/function.py
```diff
@@ -193,6 +187,40 @@ def __dtype_setup__(cls, **kwargs):
...
+            return tuple(0 for _ in func.indices)
+        if stagger == 'cell':
```
FabioLuporini: is this just a special case of the loop below?

devito/types.py
```diff
@@ -348,6 +348,7 @@ def __new__(cls, *args, **kwargs):
             newobj._indices = indices
             newobj._shape = cls.__shape_setup__(**kwargs)
             newobj._dtype = cls.__dtype_setup__(**kwargs)
+            newobj._staggered = cls.__staggered_setup__(newobj, **kwargs)
```
FabioLuporini: don't pass ``newobj``, just have ``__staggered_setup__`` return what you want to store in ``_staggered``

devito/types.py
```diff
@@ -369,6 +370,11 @@ def __init__(self, *args, **kwargs):
 
     __hash__ = Cached.__hash__
 
+    @classmethod
+    def __staggered_setup__(cls, obj, **kwargs):
```
FabioLuporini: from the previous comment, no need for ``obj`` here

devito/function.py
```diff
@@ -193,6 +187,40 @@ def __dtype_setup__(cls, **kwargs):
         else:
             return np.float32
 
+    @classmethod
+    def __staggered_setup__(cls, func, **kwargs):
```
FabioLuporini: no need for ``func`` here. See comments below in ``types.py``.

devito/function.py
```diff
@@ -193,6 +187,40 @@ def __dtype_setup__(cls, **kwargs):
...
+        returns a tuple of 0 or 1 for each dimension:
+        0 if not staggered in the dimension
+        1 if staggered in the dimension
+        """
+        stagger = kwargs.get('staggered')
+        s = (lambda x: x if type(x) is tuple else (x,))(stagger)
+        if stagger is None:
+            func.is_Staggered = False
+            return tuple(0 for _ in func.indices)
+        func.is_Staggered = True
```
FabioLuporini: no need to set anything here. ``is_Staggered`` is better off as a property such as ``return any(i == 1 for i in self.staggered``)

devito/finite_differences/differentiable.py
mloubout: It's on symbolic  expressions  so should be

devito/types.py
mloubout: Necessary, needs to access `obj.indices` in the function

FabioLuporini: ah, so don't make it a classmethod. For example, look at __halo_setup__

tests/test_ir.py
```diff
@@ -524,7 +524,7 @@ def test_iteration_parallelism_2d(self, exprs, atomic, parallel):
         for i, e in enumerate(list(exprs)):
             exprs[i] = eval(e)
 
-        op = Operator(exprs, dle='advanced')
+        op = Operator(exprs, dle='openmp')
```
FabioLuporini: wait, what?

devito/equation.py
```diff
@@ -89,4 +91,7 @@ def solve(eq, target, **kwargs):
     # turnaround time
     kwargs['rational'] = False  # Avoid float indices
     kwargs['simplify'] = False  # Do not attempt premature optimisation
+    if hasattr(eq, 'lhs'):
```
FabioLuporini: what? recently I've seen this ``hasattr`` too often.

can you not just ``if eq.is_Equality`` ? and btw, what else should/could it be?

tests/test_ir.py
mloubout: with `advanced` cache blocking dims appear which make it fail, 

># Use 'openmp' here instead of 'advanced' to disable loop blocking
line 584 for 3D, should be same here

devito/equation.py
mloubout: >and btw, what else should/could it be?
it can be a single expression as solve defaults to `Eq(expr, 0)` if called as `solve(expr)` in which case `expr` does not have `is_Equality` attribute

tests/test_ir.py
FabioLuporini: then it's a bug to be fixed

devito/equation.py
FabioLuporini: why not? all sympy expressions, even symbols, have ``is_Equality``

tests/test_ir.py
mloubout: It;s not a bug, that's what's needed for that test

FabioLuporini: ahhhhhh sorry now I see what you mean!

devito/finite_differences/differentiable.py
mloubout: To avoid duplicates of `add,mul,...` makes evrything lots easier and less messy to have `Function` being differentiable

devito/finite_differences/finite_difference.py
```diff
@@ -0,0 +1,401 @@
...
+                            fd_order=dim_order, stagger=centered)
+            name_fd = 'd%sc' % name
+            desciption = 'centered derivative staggered w.r.t dimension %s' % d
+
+            derivatives[name_fd] = (deriv, desciption)
+
+    return derivatives
+
+
+def initialize_derivatives(function):
```
FabioLuporini: Reminder: we should call generate_fd_shortcuts from within Function.__init__

devito/grid.py
```diff
@@ -49,21 +49,20 @@ class Grid(ArgProvider):
 
        .. code-block:: python
 
-          x ^
-            |
-            |           origin + extent
+                      x
```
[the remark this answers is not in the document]
mloubout: Nope still x horizontal and y vertical, only look weird when plotting because of row major python

devito/equation.py
```diff
@@ -27,6 +27,9 @@ def xreplace(self, rules):
         return self.func(self.lhs.xreplace(rules), self.rhs.xreplace(rules),
                          region=self._region)
 
+    def __str__(self):
```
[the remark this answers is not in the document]
mloubout: Leftover from Differentiable not printing nicelly inside Eq, should be dropable now

devito/finite_differences/finite_difference.py
```diff
@@ -0,0 +1,401 @@
...
+            deriv = partial(first_derivative, order=space_fd_order,
+                            dim=d, side=left)
+            name_fd = 'd%sl' % name
+            desciption = 'left first order derivative w.r.t dimension %s' % d
+            derivatives[name_fd] = (deriv, desciption)
+            # right
+            deriv = partial(first_derivative, order=space_fd_order,
+                            dim=d, side=right)
+            name_fd = 'd%sr' % name
+            desciption = 'left first order derivative w.r.t dimension %s' % d
```
mloubout: I thought I fixed that, commits mix up must have brought it back

PR #644, opened 2018-09-13
Added mpich installation command to Dockerfile but tests now failing

RajatRasal:


docker/Dockerfile
```diff
@@ -1,5 +1,8 @@
 FROM python:3.6
 
+RUN apt-get update && apt-get install -y \ 
+    mpich
```
mloubout: https://github.com/opesci/devito/blob/162abb6b318e77eaa4e8f719047327c45782056f/Dockerfile.jenkins#L43

RajatRasal: When I was trying to build the docker image using docker-compose, there was no MPI present. By doing this I was able to create the image, but the tests are failing.

mloubout: `libmpich-dev ` missing

PR #645, opened 2018-09-18
Guarded injection/interpolation and misc improvements

FabioLuporini:
The main feature is that now injection (indirect) increments and interpolation (indirect) reads are "guarded" with suitable if-then-else constructs. These are transparently generated by the compiler leveraging ``ConditionalDimension``.

A large set of fixes and improvement are also shipped, which touch many bits of the compilation pipeline. Among these:

* More powerful ``ConditionalDimension`` (it now supports arbitrary conditions through the ``condition`` argument.
* Simpler and more powerful clusterization (/ir/clusters/algorithms.py)
* More general Cluster grouping (/ir/clusters/algorithms.py)
* IntervalGroup's Intervals are now partially, instead of totally, ordered. This makes IterationSpace.merge much more powerful (it can now merge iteration spaces that couldn't be merged previously)
* ...

All these changes were somehow needed to make it possible using ``ConditionalDimension``s within non-fd loops, such as those generated by SparseFunctions (eg, injection/interpolation loops)


devito/ir/clusters/algorithms.py
```diff
@@ -222,33 +224,27 @@ def bump_and_contract(targets, source, sink):
...
-                queue.extend([i for i in g if i not in queue])
-
-        # Wrap each tensor expression in a PartialCluster
-        for k, v in mapper.items():
-            if k.is_Tensor:
-                scalars = [i for i in g[:g.index(k)] if i.is_Scalar]
-                intervals, sub_iterators, _ = k.ispace.args
-                ispace = IterationSpace(intervals, sub_iterators, v)
-                clusters.append(PartialCluster(scalars + [k], ispace, k.dspace))
+    for group in group_expressions(exprs):
```
FabioLuporini: note for reviewers: this is so much nicers and simpler to understand than it used to...
note for @tjb900 : this is were I was hoping to have gained some speed. TBC

devito/ir/clusters/graph.py
```diff
@@ -4,40 +4,29 @@
...
+from devito.ir.equations import ClusterizedEq
+from devito.symbolics import (as_symbol, retrieve_indexed, retrieve_terminals,
+                              convert_to_SSA, q_indirect, q_timedimension)
 from devito.tools import DefaultOrderedDict, flatten, filter_ordered
 
 __all__ = ['FlowGraph']
 
 
-class Node(Eq):
+class Node(ClusterizedEq):
```
FabioLuporini: note for reviewers: So much nicer now, as it leverages the existing IR Equation hierarchy

devito/ir/equations/equation.py
```diff
@@ -110,32 +112,27 @@ def __new__(cls, *args, **kwargs):
                              "and kwargs=%s" % (str(args), str(kwargs)))
 
         # Well-defined dimension ordering
-        ordering = dimension_sort(expr, key=lambda i: not i.is_Time)
-
-        # Introduce space sub-dimensions if need to
```
FabioLuporini: note for reviewers: moved to operator.py because the substitution has to be applied globally to all equations, rather than locally. Otherwise, we might create redundant yet symbolically identical SubDimensions

devito/ir/equations/equation.py
```diff
@@ -110,32 +112,27 @@ def __new__(cls, *args, **kwargs):
                              "and kwargs=%s" % (str(args), str(kwargs)))
 
         # Well-defined dimension ordering
-        ordering = dimension_sort(expr, key=lambda i: not i.is_Time)
```
FabioLuporini: note for reviewers: Dropping this ``lambda i: not i.is_Time`` is another good thing of this PR. If it wasn't a hack, it was very close to it.

devito/ir/iet/scheduler.py
```diff
@@ -99,7 +99,7 @@ def iet_lower_dimensions(iet):
     # Lower ConditionalDimensions
     cdims = [d for d in FindSymbols('free-symbols').visit(iet)
              if isinstance(d, ConditionalDimension)]
-    mapper = {d: IntDiv(d.parent, d.factor) for d in cdims}
+    mapper = {d: IntDiv(d.index, d.factor) for d in cdims}
```
FabioLuporini: @tjb900 : this is where the ConditionalDimension is lowered to an array index

devito/operator.py
```diff
@@ -404,7 +434,7 @@ def apply(self, **kwargs):
         self.cfunction(*arg_values)
 
         # Post-process runtime arguments
-        args = self._postprocess_arguments(args)
+        self._postprocess_arguments(args)
```
FabioLuporini: note for reviewers: This essentially makes sure that SparseFunctions do a gather upon returning from C-land. It only has an effect when running with MPI

mloubout: From my test, this fails with user provided arguments as it only gatheres in `self.output` not in `apply(rec=myrec)` for example

devito/operator.py
```diff
@@ -181,8 +182,8 @@ def _postprocess_arguments(self, args):
         """
         Process runtime arguments upon returning from ``.apply()``.
         """
-        finalized = {p.name: p._arg_apply(args[p.name]) for p in self.output}
-        return {k: finalized.get(k, v) for k, v in args.items()}
+        for p in self.output:
+            p._arg_apply(args[p.name])
```
mloubout: See below, need to check if `p` in `kwargs`

devito/operator.py
FabioLuporini: discussed offline with @mloubout -- this is gonna be addressed in `mpi-sparse-duplication` (branch of the next imminent PR)

FabioLuporini:
all good after offline discussion (leftovers addressed in a new imminent PR), merging

PR #646, opened 2018-09-19
compiler: Make clang behave similarly to gcc

FabioLuporini:
This should shut off clang when unused variables are encountered (yes, we still occasionally generate some)

@mloubout can you check on your Mac that this doesn't break jit in any stupid way


mloubout:
Seems to be fine. GTG

PR #647, opened 2018-09-21
DSL enhancements

FabioLuporini:
Reopening from #642 

- '(u + v.dx).dy' and things like that possible
- staggered now is defined as `staggered = x` for example, not more `(0,1,0)` and `None` corresponds to cartesian grids
- FD shortcuts are now created on demand
- Some misc

Most credits to @mloubout


devito/cgen_utils.py
```diff
@@ -169,11 +173,17 @@ def _print_ListInitializer(self, expr):
         return "{%s}" % ', '.join([self._print(i) for i in expr.params])
 
     def _print_IntDiv(self, expr):
-        return str(expr)
+        return expr.__str__()
```
FabioLuporini: droppable?

devito/function.py
```diff
@@ -1479,7 +1406,7 @@ def _interpolation_indices(self, variables, offset=0):
         for i, idx in enumerate(index_matrix):
             ind_subs = dict([(dim, ind) for dim, ind in zip(self.grid.dimensions, idx)])
             v_subs = [(v, v.subs(ind_subs))
-                      for v in variables if not v.base.function.is_SparseFunction]
+                      for v in variables if not isinstance(v, SparseFunction)]
```
FabioLuporini: why was this necessary?

devito/function.py
```diff
@@ -1655,6 +1580,10 @@ class SparseTimeFunction(AbstractSparseTimeFunction, SparseFunction):
 
     is_SparseTimeFunction = True
 
+    @property
+    def fd_indices(self):
```
FabioLuporini: we can drop this, unused

devito/function.py
mloubout: Because `interp/inject` was using `indexify` to make it "easier", now done on functions.

devito/cgen_utils.py
mloubout: yep

mloubout:
TTI is messed up, produces really bad results.

mloubout:
Toook a while, but here it is.

PR #648, opened 2018-09-25
MPI: support for scattering and gathering of sparse data values

FabioLuporini:
+ tests

### DIFF
--- a/devito/function.py
+++ b/devito/function.py
@@ -1242,8 +1242,13 @@ class AbstractSparseFunction(TensorFunction):
 
         return values
 
-    def _arg_apply(self, data):
-        self._dist_gather(data)
+    def _arg_apply(self, data, alias=None):
+        key = alias or self
+        if isinstance(key, AbstractSparseFunction):
+            key._dist_gather(data)
+        elif self.grid.distributor.nprocs > 1:
+            raise NotImplementedError("Don't know how to gather data from an "
+                                      "object of type `%s`" % type(key))
 
     # Pickling support
     _pickle_kwargs = TensorFunction._pickle_kwargs + ['npoint', 'space_order']
--- a/devito/function.py
+++ b/devito/function.py
@@ -1454,8 +1454,8 @@ class SparseFunction(AbstractSparseFunction, Differentiable):
             mapper = {}
             for j, d in zip(idx, self.grid.dimensions):
                 p = points[j]
-                lb = sympy.And(p > d.symbolic_start - self._radius, evaluate=False)
-                ub = sympy.And(p < d.symbolic_end + self._radius, evaluate=False)
+                lb = sympy.And(p >= d.symbolic_start - self._radius, evaluate=False)
+                ub = sympy.And(p <= d.symbolic_end + self._radius, evaluate=False)
                 condition = sympy.And(lb, ub, evaluate=False)
                 mapper[d] = ConditionalDimension(p.name, self._sparse_dim,
                                                  condition=condition, indirect=True)
--- a/devito/function.py
+++ b/devito/function.py
@@ -18,8 +18,8 @@ from devito.symbolics import indexify, retrieve_functions
 from devito.finite_differences import Differentiable, generate_fd_shortcuts
 from devito.types import (AbstractCachedFunction, AbstractCachedSymbol, Symbol, Scalar,
                           OWNED, HALO, LEFT, RIGHT)
-from devito.tools import (EnrichedTuple, Tag, ReducerMap, ArgProvider, as_mapper,
-                          as_tuple, flatten, is_integer, prod, powerset, filter_ordered,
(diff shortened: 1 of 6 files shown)


devito/function.py
```diff
@@ -1052,20 +1055,47 @@ def _is_owned(self, point):
...
-        belong to the same MPI rank.
+        Return a mask to index into ``self.data``, which creates a new
+        data array that logically contains N consecutive groups of sparse
+        data values, where N is the number of MPI ranks. The i-th group
+        contains the sparse data values accessible by the i-th MPI rank.
+        Thus, sparse data values along the boundary of two or more MPI
+        ranks are duplicated.
         """
         dmap = self._dist_datamap
         mask = np.array(flatten(dmap[i] for i in sorted(dmap)), dtype=int)
```
mloubout: Only `dmap[self.rank]` should be scattered, the `alltoall` will take care of the receive part

devito/function.py
```diff
@@ -1074,13 +1104,27 @@ def _dist_mask(self):
...
+
+    @property
+    def _dist_gather_mask(self):
+        """
+        Return a mask to index into the ``data`` received upon returning
+        from ``self._dist_alltoall``. This mask creates a new data array
+        in which duplicate sparse data values have been discarded. The
+        resulting data array can thus be used to populate ``self.data``.
+        """
+        ret = list(self._dist_scatter_mask)
```
mloubout: the received part is not what sent, it should be the complement of dist so 
`mask = np.array(flatten(dmap[i] for i in sorted(dmap) if i != self.rank), dtype=int)`

devito/function.py
```diff
@@ -1203,8 +1242,13 @@ def _arg_values(self, **kwargs):
...
         return values
 
-    def _arg_apply(self, data):
-        self._dist_gather(data)
+    def _arg_apply(self, data, alias=None):
+        key = alias if alias is not None else self
+        if isinstance(key, AbstractSparseFunction):
+            key._dist_gather(data)
+        elif self.grid.distributor.nprocs > 1:
+            raise NotImplementedError("Don't know how to gather data from an "
```
mloubout: I wouldn't throw an error, only a waring, everything ran fine, just saying input was not changed and is to it's original value

devito/function.py
```diff
@@ -1432,8 +1476,8 @@ def _interpolation_indices(self, variables, offset=0):
             mapper = {}
             for j, d in zip(idx, self.grid.dimensions):
                 p = points[j]
-                lb = sympy.And(p > d.symbolic_start - self._radius, evaluate=False)
-                ub = sympy.And(p < d.symbolic_end + self._radius, evaluate=False)
+                lb = sympy.And(p >= d.symbolic_start - self._radius, evaluate=False)
+                ub = sympy.And(p <= d.symbolic_end + self._radius, evaluate=False)
```
mloubout: why?

devito/function.py
```diff
@@ -1545,7 +1589,7 @@ def _dist_scatter(self, data=None):
         mpitype = MPI._typedict[np.dtype(self.dtype).char]
 
         # Pack (reordered) data values so that they can be sent out via an Alltoallv
-        data = data[self._dist_mask]
+        data = data[self._dist_scatter_mask]
```
mloubout: FYI, this actually *** the `c-contiguous` for some reason

devito/function.py
```diff
@@ -1554,7 +1598,7 @@ def _dist_scatter(self, data=None):
         data = scattered
 
         # Pack (reordered) coordinates so that they can be sent out via an Alltoallv
-        coords = self.coordinates.data[self._dist_subfunc_mask]
+        coords = self.coordinates.data[self._dist_subfunc_scatter_mask]
         # Send out the sparse point coordinates
         _, scount, sdisp, rshape, rcount, rdisp = self._dist_subfunc_alltoall
         scattered = np.empty(shape=rshape, dtype=self.coordinates.dtype)
```
mloubout: `coords/data = scattered` that is incorrect. Owned data is `data` no matter to how many process it is sent.

devito/function.py
```diff
@@ -1583,11 +1627,12 @@ def _dist_gather(self, data):
         data = gathered
 
         # Insert back into `self.data` based on the original (expected) data layout
-        self.data[:] = data[self._dist_mask]
+        self.data[:] = data[self._dist_gather_mask]
```
mloubout: `gathered = np.empty(shape=sshape, dtype=self.dtype)`
gahtered is what is received, should be `rshape`

FabioLuporini:
about your comments -- we both know this is a WIP PR. If something is already fixed in the next PR, I'd ignore it here

FabioLuporini:
and in particular, all things that you rightly point out... we need a test that exposes them so that I can easily see what's going on !

devito/function.py
FabioLuporini: because otherwise I don't access values in the last halo entry, which are instead available

PR #650, opened 2018-09-26
explicitly test for the most common cases in q_affine

tjb900:
This is some juicy low-hanging fruit that showed up in profiling.  For me, this speeds up the python component of Operator creation by ~20%.

Happy to pretty it up a bit and/or remove the debug info about the slow path - this is really just to start a discussion.


FabioLuporini:
Merged, thanks.

PR #651, opened 2018-09-26
More performance improvements

tjb900:
As suggested by @FabioLuporini


devito/ir/support/basic.py
```diff
@@ -457,6 +468,9 @@ def __eq__(self, other):
             isinstance(other, TimedAccess) and\
             self.directions == other.directions
 
+    def __hash__(self):
```
FabioLuporini: @tjb900 do we actually need to override __hash__ 4 times? is this due to multiple inheritance?

FabioLuporini: discussed offline, all good

PR #652, opened 2018-09-27
Indexify-hotfix

mloubout:
Small PR to fix `indexify` for Sub/Conditional dimensions

Added two test for FD in combination with Sub/Conditional dimensions.

### DIFF
--- a/devito/types.py
+++ b/devito/types.py
@@ -622,9 +622,18 @@ class AbstractCachedFunction(AbstractFunction, Cached):
         """Create a :class:`sympy.Indexed` object from the current object."""
         if indices is not None:
             return Indexed(self.indexed, *indices)
-        # Only replace spacing -> 1 if used as index
-        subs = dict([(i.spacing, 1) for i in self.indices if
-                     any(i in a.args for a in self.args)])
+
+        # Get spacing symbols for replacement
+        spacings = [i.spacing for i in self.indices]
+
+        # Only keep the ones used as indices.
+        spacings = [s for i, s in enumerate(spacings)
+                    if s.free_symbols.intersection(self.args[i].free_symbols)]
+
+        # Substitution for each index
+        subs = dict([(s, 1) for s in spacings])
+
+        # Indices after substitutions
         indices = [a.subs(subs) for a in self.args]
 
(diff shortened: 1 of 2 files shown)


devito/types.py
```diff
@@ -622,9 +622,18 @@ def indexify(self, indices=None):
...
-        # Only replace spacing -> 1 if used as index
-        subs = dict([(i.spacing, 1) for i in self.indices if
-                     any(i in a.args for a in self.args)])
+
+        # Get spacing symbols for replacement
+        spacings = [i.spacing for i in self.indices]
+
+        # Only keep the ones used as indices.
+        spacings = [s for i, s in enumerate(spacings)
+                    if tuple(s.free_symbols)[0] in self.args[i].free_symbols]
```
FabioLuporini: why u taking the first entry in ``tuple(s.free_symbols)``?
isn't ``tuple(s.free_symbols)[0] == s`` ?

tests/test_dimension.py
```diff
@@ -379,6 +379,29 @@ def test_subdimleft_notparallel(self):
         assert np.all(u.data[0, thickness:] == 0)
         assert np.all(u.data[0, :, thickness+12:] == 0)
 
+    def test_subdim_fd(self):
+        """
+        Test that the FD shortcuts are handled correctly qith subdimensions
```
FabioLuporini: typo: with
subdimensions -> SubDimensions

tests/test_dimension.py
```diff
@@ -437,6 +460,33 @@ def test_spacial_subsampling(self):
         # Verify that u2[x,y]= u[2*x, 2*y]
         assert np.allclose(u.data[:-1, 0:-1:2, 0:-1:2], u2.data[:-1, :, :])
 
+    def test_subsampled_fd(self):
+        """
+        Test that the FD shortcuts are handled correctly qith subdimensions
```
FabioLuporini: same as before

devito/types.py
mloubout: `s.free_symbols` is a set

FabioLuporini: ah, then use intersection

PR #655, opened 2018-10-01
more SubDomains and examples tweaks

FabioLuporini:
@mloubout I still don't like that the setup time of our examples has now increased due to the revisited damp_boundary. I know it's a one time cost, but it slows down debugging... and probably testing as well


devito/grid.py
```diff
@@ -107,11 +116,51 @@ def __repr__(self):
...
+
+    @property
+    def stepping_dim(self):
+        """Stepping dimension associated with this Grid."""
+        return self._stepping_dim
+
+    @property
+    def subdomains(self):
+        """The :class:`SubDomain`s defined in this Grid."""
+        return {i.name: i for i in self._subdomains}
```
mloubout: Why `i.name` instead of `i`

devito/grid.py
```diff
@@ -209,3 +258,118 @@ def __setstate__(self, state):
...
+    def __eq__(self, other):
+        if not isinstance(other, SubDomain):
+            return False
+        return self.name == other.name and self.dimensions == other.dimensions
+
+    def __hash__(self):
+        return hash((self.name, self.dimensions))
+
+    def __str__(self):
+        return "SubDomain[%s]" % self.name
```
mloubout: Maybe more explicit to put the subdims in `str` rather than name

devito/grid.py
```diff
@@ -209,3 +258,118 @@ def __setstate__(self, state):
...
+class Interior(SubDomain):
+
+    """
+    The interior of the computational domain (i.e., boundaries are excluded).
+    """
+
+    name = 'interior'
+
+    def define(self, dimensions):
+        return {d: ('middle', 1, 1) for d in dimensions}
```
mloubout: what's these hard coded `1,1`

devito/grid.py
FabioLuporini: because I need a dictionary (not a list) from strings to the corresponding objects so that I can easily index into it, even in user code (eg, I'm using it in examples/model.py to retrieve a particular subdomain)

devito/grid.py
FabioLuporini: I don't know what else to attach -- what else would you like to see apart from the name? The dimensions?

devito/grid.py
FabioLuporini: this is for the Interior subdomain, which is one of the preexisting subdomains shipped by Devito. And the interior subdomain cuts away the extremes, so 1 on the left and 1 on the right side

devito/grid.py
mloubout: Ok makes sense

devito/grid.py
mloubout: I would do like 
`"SubDomain %s[%s]"  % (self.name, self.dimensions)`
Like `SubDomain interior[x, y]`

devito/grid.py
mloubout: Why isn't default `0`. Default case shouldn't be the domain is the Grid without any halo?

devito/grid.py
```diff
@@ -107,11 +116,51 @@ def __repr__(self):
...
+        """Stepping dimension associated with this Grid."""
+        return self._stepping_dim
+
+    @property
+    def subdomains(self):
+        """The :class:`SubDomain`s defined in this Grid."""
+        return {i.name: i for i in self._subdomains}
+
+    @property
+    def interior(self):
```
tjb900: This is kind of the same as @mloubout 's comment further down about how thick the default should be.  I'd be quite surprised if anything other than the tests used Eq(region=INTERIOR) in its current form - it doesn't work for >2nd order spatial stencils, and you don't need it if you only want zeros in the boundary anyway.  So that doesn't leave a lot of useful scenarios.

Since this PR is (rightfully) breaking compatibility by changing the syntax for `Eq` away from `region=INTERIOR`, should we take the opportunity to retire it completely, or change it to something more generic rather than carry it around?  And the relevant tests could create a subdomain themselves, rather than using this property?

devito/grid.py
```diff
@@ -68,35 +68,44 @@ class Grid(ArgProvider):
...
-            self.dimensions = tuple(SpaceDimension(name=n, spacing=s)
-                                    for n, s in zip(dim_names, dim_spacing))
+            self._dimensions = tuple(SpaceDimension(name=n, spacing=s)
+                                     for n, s in zip(dim_names, dim_spacing))
         else:
-            self.dimensions = dimensions
+            self._dimensions = dimensions
+
+        # Initialize SubDomains
+        subdomains = tuple(i for i in (Domain(), Interior()) + as_tuple(subdomains))
```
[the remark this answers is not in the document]
tjb900: Also see other comment about how I'm not sure creating the interior subdomain in every case is worth it.

examples/seismic/source.py
```diff
@@ -93,9 +93,11 @@ def __new__(cls, **kwargs):
         time_order = kwargs.pop('time_order', 2)
         p_dim = kwargs.pop('dimension', Dimension(name='p_%s' % name))
 
-        # Either `npoint` or `coordinates` must be provided
         coordinates = kwargs.pop('coordinates', None)
-        npoint = kwargs.pop('npoint', None) or coordinates.shape[0]
+        # Either `npoint` or `coordinates` must be provided
+        npoint = kwargs.pop('npoint', None)
+        if npoint is None:
+            npoint = coordinates.shape[0]
```
tjb900: Let's take the opportunity to raise a good exception here?  Like "npoint or coordinates must be provided"?  The current error will be "object of type NoneType doesn't have attribute shape" or something similar.

devito/grid.py
FabioLuporini: that's for interior, and since ``domain = boundary + interior``, with the two 1s we are stripping off the boundary at left and right

devito/grid.py
```diff
@@ -1,49 +1,49 @@
...
 from devito.mpi import Distributor
 from devito.parameters import configuration
-from devito.tools import ArgProvider, ReducerMap
+from devito.tools import ArgProvider, ReducerMap, as_tuple
 
 from sympy import prod
 import numpy as np
 
-__all__ = ['Grid']
+__all__ = ['SubDomain']
```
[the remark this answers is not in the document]
FabioLuporini: No, that was a leftover. `Grid`  is a backend-specific concept, so conceptually:
* you shouldn't actually get it  when you do ``from devito.grid import *``
* you should rather ``from devito.core import Grid`` or ``from devito.yask import Grid``

Things are structured so that when users write ``from devito import Grid, ...`` the right type of Grid is automatically fetched

not sure if it's clear?

FabioLuporini:
is this a green light? I'm really sad that the setup time of our examples has now increased

PR #656, opened 2018-10-04
generate time derivatives for anything with time_order

tjb900:
trivial change to generate time derivative shortcuts on SparseTimeFunction-like things

### DIFF
--- a/devito/finite_differences/finite_difference.py
+++ b/devito/finite_differences/finite_difference.py
@@ -317,7 +317,10 @@ def generate_fd_shortcuts(function):
     """
     dimensions = function.indices
     space_fd_order = function.space_order
-    time_fd_order = function.time_order if function.is_TimeFunction else 0
+    try:
+        time_fd_order = function.time_order
+    except AttributeError:
+        time_fd_order = 0
 
     if function.is_Staggered:
         deriv_function = staggered_diff
--- a/devito/function.py
+++ b/devito/function.py
@@ -1026,6 +1026,9 @@ class AbstractSparseFunction(TensorFunction):
 
             self._space_order = kwargs.get('space_order', 0)
 
+            # Dynamically add derivative short-cuts
+            self._fd = generate_fd_shortcuts(self)
+
     @classmethod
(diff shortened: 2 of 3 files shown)


mloubout:
Nice, was trying to go all the way and add space ones too but WIP obviously.

Can add a test for it?

tjb900:
> Nice, was trying to go all the way and add space ones too but WIP obviously.
> 
> Can add a test for it?

Ah, yes.  Good point :)  Will do.

devito/finite_differences/finite_difference.py
```diff
@@ -317,7 +317,10 @@ def generate_fd_shortcuts(function):
     """
     dimensions = function.indices
     space_fd_order = function.space_order
-    time_fd_order = function.time_order if function.is_TimeFunction else 0
+    try:
```
FabioLuporini: just a reminder: we really need a mixin "Time<Something>" class that does/provides the following:
* set `is_Time` flag to True
* property for time_order
* property for time_dim
* class attribute `_time_position`
* ...

This should be inherited by:
* TimeFunction
*SparseTimeFunction

tjb900:
> I thought with the new tests the objective was to check the symbolic derivatives, but no derivatives are taken on the SparseFunctions?

One of them does.  I was looking for the right place to put them, and couldn't find any tests which verified the results of using `SparseTimeFunction`, so that's why I added more.  I think though that the right place might have been `test_interpolation.py`.

tjb900:
I'm going to have to leave this one for a while - I need to think more about where to call `generate_fd_shortcuts`.

If we are calling the `super()` constructor at the start of __init__, it cannot be done in there, because it relies on the `time_order` etc being set, and that happens later.  And the call to `super.__init__` cannot be systematically moved to the end, either, because e.g. `PrecomputedSparseFunction`'s `__init__` relies on `self.npoints`, which is set in `AbstractSparseFunction.__init__`.

FabioLuporini:
@tjb900 
closed as the same functionality has landed via #742

PR #657, opened 2018-10-04
RFC - Prepare fewer arguments

FabioLuporini:
Opening a PR only to leave some comments

### DIFF
--- a/devito/operator.py
+++ b/devito/operator.py
@@ -110,6 +110,12 @@ class Operator(Callable):
 
         # Derive parameters as symbols not defined in the kernel itself
         parameters = self._build_parameters(iet)
+        
+        # By default operator argument list is not cached
+        self.cache_args = False
+        
+        # By default performance will be profiled
+        self.profiling = True
 
         # Finish instantiation
         super(Operator, self).__init__(self.name, iet, 'int', parameters, ())
@@ -437,7 +443,8 @@ class OperatorRunnable(Operator):
         self._postprocess_arguments(args, **kwargs)
 
         # Output summary of performance achieved
-        return self._profile_output(args)
+        if self.profiling:
+            return self._profile_output(args)
 
     def _profile_output(self, args):
         """Return a performance summary of the profiled sections."""
--- a/devito/operator.py
+++ b/devito/operator.py
@@ -116,6 +116,9 @@ class Operator(Callable):
         
         # By default performance will be profiled
         self.profiling = True
+        
+        # Cached args
+        self.cached_args = None
 
         # Finish instantiation
         super(Operator, self).__init__(self.name, iet, 'int', parameters, ())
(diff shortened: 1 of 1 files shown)


FabioLuporini:
### About Profiling
Look [here](https://github.com/opesci/devito/blob/master/devito/profiling.py#L239). You can add a new profiling level, for example "minimal" (feel free to find a better name). So now you have 'minimal', 'basic', 'advisor', in increasing order of overhead they introduce. We could even a fourth 'no-op' profiler which does nothing, but that's for another day. OK, so you introduce this "minimal" profiler; this needs to correspond to a new MinimalProfiler class, which overrides the Profiler's ``summary`` method so that ``summary = PerformanceSummary()`` is filled in with runtime only, and perhaps dummy values for all other things. This way, no changes to Operator are required. If the user sets DEVITO_PROFILING=minimal (in fact, we should/could make it the default), then under-the-hood MinimalProfiler will be picked and no time will be wasted on computing ``ops``, ``points``, ``overhead``...which is where most of the overhead is coming from

Even better, I'd slightly refactor the class hierarchy in ``profiling.py``. The actual base class, Profiler, should  be the MinimalProfiler class I'm talking about above; this, in turn, should be inherited to get the current Profiler class, which corresponds to ['basic'](https://github.com/opesci/devito/blob/master/devito/profiling.py#L240) (to clarifiy, I'm saying this IOW a few lines below)

yes, names are quite messy and dumb. At the end of the day, we should/could rather have:
* 'basic' profiler (which is what I called 'minimal') above
* 'advanced' profiler (the one that is currently called 'basic', ie the one that computes ops, points, etc)
* 'advisor' profiler

`advanced` ('AdvancedProfiler' class?) inherits from `basic` ('Profiler' class), `advisor` ('AdvisorProfiler' class) inherits from `advanced` ('AdvancedProfiler' class)

questions for you, @rhodrin :
1) does this make sense to you (if not, please insist)
2) do you see the value in doing it this way?

FabioLuporini:
### About argument caching

What you wrote in Operator made me realise that what we actually need is memoization: do you know about memoization? Googling "python memoization" shows this for example: https://stackoverflow.com/questions/1988804/what-is-memoization-and-how-can-i-use-it-in-python
in fact, memoization has nothing to with Python per se, but just to give it some concrete context...
Anyway, it turns out that we do support memoization in Devito: see [here](https://github.com/opesci/devito/blob/master/devito/tools/memoization.py#L45). If the method does not take in input any argument, then we can actually use a ``cached_property`` -- at the end of the day, we'll get the same effect.

I also just thought that there are some subtleties in what we want to achieve here. Caching/memoization would be fine whenever the user calls ``op.apply`` by supplying different dimension arguments (eg, `time_m`, `time_M`, ...), and perhaps ``Constant`` objects, **only**. But it the user supplies a different `Function` (e.g., ``op.apply(..., u=u0)``) then all arguments should be reprocessed for type checking (we need to be sure that ``u0`` is defined on the same ``Grid/Dimensions`` as all other Operator (non-supplied) ``Function``s). And I suspect this is quite a recurring situation...

So, I'm now tempted to address the overheads induced by ``_prepare_arguments`` by instead start caching (cached_property, memoization) the output of those functions/methods, called within ``_prepare_arguments``, that, as your profiling graph shows, are taking up a significant amount of time. 
Here's some simple things that we could do:

* Turn [this](https://github.com/opesci/devito/blob/master/devito/ir/support/space.py#L216) into a cached_property. 
* Same [here](https://github.com/opesci/devito/blob/master/devito/ir/support/space.py#L228)
* **Try** using `is` instead of `==` [here](https://github.com/opesci/devito/blob/master/devito/ir/support/space.py#L297). `is` won't call `__eq__`, which seems to take some time. I've emphasised "try" because I hope this change won't break anything (it really shouldn't !!!)
* Turn [this](https://github.com/opesci/devito/blob/master/devito/mpi/distributed.py#L152) into a cached_property
* Add the `@memoize_meth` decorator to all `_arg_defaults` in `function.py`
* Turn `size_name`, `min_name`, `max_name` in `dimension.py` into cached_properties. Surprisingly, they seem to take up 1% of the runtime each.
* **Try** turning [this](https://github.com/opesci/devito/blob/master/devito/grid.py#L181) into `return any(dim is d for d in self.distributor.dimensions)` -- using `is` instead of `in` will allow us to skip SymPy's __eq__ comparison which is insanely expensive. Again, I'm emphasising "try" only because we should double check this doesn't break anything -- but I'm pretty sure it doesn't

This is a start -- we could try them , profile the code again, and see how much improvement we get. 

My suggestion is to apply, test, and time these changes incrementally, so that we produce a neat breakdown of the performance improvements. It's a bit tedious then doing it all at once, but eventualyl it's gonna pay off, I think

Do you have any thoughts or questions?

rhodrin:
Thanks for the detailed and helpful comments Fabio.

Regarding the first two questions:

1. That makes sense yes, but I'll need to take a better look at the Profiler class to understand all the details.
2. Yes, I certainly see the value in doing it this way. I think adding a DEVITO_PROFILING=NONE option would also be useful, but as you say, that can be left for the future.

Regarding the caching:

I've not heard of memoization but will look into it. The rest sounds good, I'll work through and profile those changes.

Will get back to you when I have results/further questions etc.

rhodrin:
Above issues have been dealt with in two new separate braches:
1) refactor_profiling
2) cache_args

PR #668, opened 2018-10-13
colorbar fix in tutorials

yuriyi:
Apparently, there are issues with axes divider in matplotlib 3.0.x so a colorbar is being plotted over the plt image instead of next to it. a simple plt.colorbar() is suggested instead to fix the issue

### DIFF
--- a/examples/seismic/plotting.py
+++ b/examples/seismic/plotting.py
@@ -3,7 +3,6 @@ try:
     import matplotlib as mpl
     import matplotlib.pyplot as plt
     from matplotlib import cm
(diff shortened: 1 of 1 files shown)


yuriyi:
agree, but the divider does not seem to work properly in the new version of matplotlib. and tutorial examples are not affected by how the colorbar is called

mloubout:
A fix just to make one example fix is dangerous, gonna break at some point for someone else

FabioLuporini:
so what do we do here ?

mloubout:
Think should close this, not a fix just a hacky way round in my opinion.

PR #669, opened 2018-10-14
Functions reorganization

mloubout:
Reorganize all functions related in its own folder, partial preparation for vector valued objects.


devito/functions/__init__.py
```diff
@@ -0,0 +1,6 @@
+from .constants import *  # noqa
```
FabioLuporini: I'd call it "constant", not "constants"

devito/functions/__init__.py
```diff
@@ -0,0 +1,6 @@
...
+from .dimension import *  # noqa
+from .dense import * # noqa
+from .sparse import *  # noqa
+from .types import *  # noqa
```
FabioLuporini: types misleading ad this point, I'd call it basic (just like SymPy does)

devito/functions/__init__.py
```diff
@@ -0,0 +1,6 @@
+from .constants import *  # noqa
+from .dimension import *  # noqa
```
FabioLuporini: if we put dimension here, and it does make sense to me, I'm not sure calling the subdirectory "functions" makes a lot of sense

devito/functions/__init__.py
mloubout: Well for now, but may have constant tensors, ... at some points so won't be just this one.

devito/functions/__init__.py
mloubout: ok

devito/dle/backends/advanced.py
```diff
@@ -9,7 +9,7 @@
 import numpy as np
 
 from devito.cgen_utils import ccode
-from devito.dimension import Dimension
+from devito.functions.dimension import Dimension
```
FabioLuporini: no need for .dimension

devito/dle/backends/basic.py
```diff
@@ -5,14 +5,14 @@
 import cgen as c
 import numpy as np
 
-from devito.dimension import IncrDimension
+from devito.functions.dimension import IncrDimension
```
FabioLuporini: same

devito/dle/backends/basic.py
```diff
@@ -5,14 +5,14 @@
...
 from devito.dle.backends import AbstractRewriter, dle_pass, complang_ALL
 from devito.ir.iet import (Denormals, Call, Callable, List, ArrayCast,
                            Transformer, FindSymbols, retrieve_iteration_tree,
                            filter_iterations, derive_parameters)
 from devito.symbolics import as_symbol
 from devito.tools import flatten
-from devito.types import Scalar
+from devito.functions.basic import Scalar
```
FabioLuporini: no need for .basic

stopping here... this applies everywhere

devito/ir/clusters/graph.py
```diff
@@ -3,7 +3,7 @@
 
 from cached_property import cached_property
 
-from devito.dimension import Dimension
+from devito.functions.dimension import Dimension
```
FabioLuporini: for example here again we can drop .dimension

devito/__init__.py
```diff
@@ -9,11 +9,11 @@
 
 from devito.base import *  # noqa
 from devito.data import *  # noqa
-from devito.dimension import *  # noqa
+from devito.functions.dimension import *  # noqa
```
FabioLuporini: nitpicking: Move down next to the other functions import

devito/base.py
```diff
@@ -6,8 +6,10 @@
 
 from devito.backends import _BackendSelector
 from devito.grid import Grid
-import devito.types as types
-import devito.function as function
+import devito.functions.basic as types
+import devito.functions.dense as function
+import devito.functions.sparse as sparse
```
FabioLuporini: can you not just ``import devito.function as function`` and simply just that, instead of ``function, sparse, constant``?

devito/core/__init__.py
```diff
@@ -20,7 +20,7 @@
 init_dle(modes)
 
 # The following used by backends.backendSelector
-from devito.function import *  # noqa
+from devito.functions import *  # noqa
```
FabioLuporini: nitpicking: move down? imports should be grouped according to a logic

devito/functions/basic.py
```diff
@@ -66,7 +66,7 @@ class Basic(object):
     is_Dimension = False
     is_Constant = False
     # Tensor symbolic objects created by the user
-    is_TensorFunction = False
+    is_GridedFunction = False
```
FabioLuporini: GridFunction, rather than GridedFunction?

devito/functions/dense.py
```diff
@@ -0,0 +1,903 @@
...
+__all__ = ['Function', 'TimeFunction']
+
+
+class GridedFunction(AbstractCachedFunction, ArgProvider):
+
+    """
+    Utility class to encapsulate all symbolic types that represent
+    tensor (array) data.
+
+    .. note::
```
FabioLuporini: occasion to switch to numpydoc

devito/functions/vector.py
```diff
@@ -0,0 +1,96 @@
+from devito.functions.dense import Function, GridedFunction
+
+__all__ = ["VectorFunction"]
+
+
+class VectorFunction(GridedFunction):
```
FabioLuporini: occasion to switch to numpydoc

devito/functions/basic.py
mloubout: `DiscreteFunction`?

FabioLuporini: the all necessarily are Discrete...

The reason I like Grid/Grided is that a TensorFunction, unlike it's superclasses, is aware of the concept of Grid. Now, Grid sounds better than Grided.

Said that, I do not really dislike Discrete. So if you don't like GridFunction, feel free to use DiscreteFunction instead.

PR #670, opened 2018-10-14
Fix dask and distributed versions

navjotk:
Fix versions of dask and distributed to get consistent behaviour

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -18,8 +18,8 @@ dependencies:
   - psutil>=5.1.0
   - sphinx
   - sphinx_rtd_theme
-  - dask
-  - distributed>=1.21.8
+  - dask==0.18.1
+  - distributed==1.22.0
   - scikit-image
   - cython
   - click


environment.yml
```diff
@@ -18,8 +18,8 @@ dependencies:
   - psutil>=5.1.0
   - sphinx
   - sphinx_rtd_theme
-  - dask
-  - distributed>=1.21.8
+  - dask==0.18.1
+  - distributed==1.22.0
```
FabioLuporini: does it break with >= ?

navjotk: We want to be able to reproduce the behaviour of `helm install stable/dask`, and that, as you might see [here](https://github.com/helm/charts/tree/master/stable/dask) , installs these specific versions. The Docker image doesn't seem to start the bokeh server on later versions of dask, and this version of dask is locked into this version of distributed AFAIU.

FabioLuporini: OK

PR #673, opened 2018-10-15
Caching some properties for efficiency. See issue #657.

rhodrin:
Currently `_prepare_arguments` (located in `operator.py`) is fairly time consuming. Upon repeated calls to `op.apply()` many arguments are being 're-prepared' when there is no need. This begins to address this problem by caching certain properties e.t.c.

Note that this is a first step and more work is required to further reduce the overheads of repeated operator calls.

### DIFF
--- a/devito/ir/support/space.py
+++ b/devito/ir/support/space.py
@@ -2,7 +2,6 @@ import abc
 from collections import OrderedDict
 from functools import reduce
 from operator import mul
-
 from cached_property import cached_property
 
 from frozendict import frozendict
--- a/devito/ir/support/space.py
+++ b/devito/ir/support/space.py
@@ -3,6 +3,8 @@ from collections import OrderedDict
 from functools import reduce
 from operator import mul
 
+from cached_property import cached_property
+
 from frozendict import frozendict
 
 from devito.tools import PartialOrderTuple, as_tuple, filter_ordered, toposort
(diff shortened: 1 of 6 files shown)


FabioLuporini:
some flake8 complaints, but apart from that this is GTG. How much is the improvement?

devito/ir/support/space.py
```diff
@@ -3,6 +3,8 @@
 from functools import reduce
 from operator import mul
 
+from cached_property import cached_property
+
```
FabioLuporini: nitpicking: no need for a blank line here.

The rule by which we organise imports is:
- first group are the imports from the Python library
- blank line
- second group are the important from third-party libraries
- blank line
- third group are the devito imports

rhodrin: Thanks for the explanation. Space removed.

Assuming the new 'basic' profiling is enabled, in a basic problem on a 21x21 grid with 10,000 calls to op.apply() these changes being the time spent in `_prepare_arguments` from ~70% down to 40%.

FabioLuporini:
there are some PEP8 issues to be fixed here (see flake8 report on travis/jenkins), and then we're GTG

rhodrin:
Hmm, yeah.
But this problem only seems to occur in yask based setups? I've been trying to reproduce a similar bug via a local test but am failing to do so...

FabioLuporini:
needs rebasing on ``master`` before merging, but otherwise, assuming green CI, this looks GTG to me

FabioLuporini:
> **if** I got it right (and it's a big if), one possible solution is to turn **all** `pop`s into `get`s in **all** `_arg_values` in devito, including those in `dimension.py` and those in `yask/`.
> To be even more rigid to avoid future messups, we could cast the return value of `_arg_defaults` to a `frozendict` (a dictionary that cannot be modified (by the callees)).

this thing I wrote time ago was BS, as we're indeed popping, but only from kwargs, not from the _arg_defaults' return value.

Merging...

Issue #658, opened 2018-10-09
Failure to recompile upon (certain) parameter changes


rhodrin:
After running a devito model with a certain parameter set (i.e. `DEVITO_AUTOTUNING=off python myfile.py`), re-running the model with a modified parameter (i.e. `DEVITO_AUTOTUNING=aggressive python myfile.py`) results in an OSError:

```python
  File "/data/devito/devito/operator.py", line 435, in apply
    self.cfunction(*arg_values)
  File "/data/devito/devito/operator.py", line 233, in cfunction
    self._lib = load(self._soname)
  File "/data/devito/devito/compiler.py", line 317, in load
    return npct.load_library(str(get_jit_dir().joinpath(soname)), '.')
  File "/home/rhodri/anaconda3/envs/devito/lib/python3.6/site-packages/numpy/ctypeslib.py", line 157, in load_library
    raise OSError("no file with expected extension")
```
With certain parameter changes the variable `self._soname` is modified and hence `self._lib = load(self._soname)` (line 233 of operator.py) points at a new shared object. However, this shared object does not get created resulting in the OSError. This error occurs when any parameter modification results in a `self._soname` different from the previous compilation. 

This object is not created since the variable `recompiled`, computed in the `jit_compile` method of compiler.py, is returned as `False` upon such parameter changes. Hence, compiling and running then modifying any parameter that results in a new `self._soname` and re-running results in this error. 

Clearing the cache and then recompiling (which should always result in `recompiled=True`) is a workaround to this issue.

Modifying the function `compile_from_string()` such that it returns values of `True` following the modification of any parameter resulting in a new `self._soname` would seem an appropriate fix?

rhodrin:
Oh, so I see `compile_from_string()` is not actually a Devito function which potentially complicates things a little?

rhodrin:
I've now created a branch [fix_recompile](https://github.com/opesci/devito/tree/fix_recompile). This, along with a CodePy patch seem the cleanest way of solving this issue.

My suggested change is modifying, in [jit.py](https://github.com/inducer/codepy/blob/45dac0819f73f40c125eae2dbb435226d73b7ce3/codepy/jit.py#L294),

```python

    def calculate_hex_checksum():
        try:
            import hashlib
            checksum = hashlib.md5()
        except ImportError:
           ...
```

to

```python

    def calculate_hex_checksum():
        try:
            import hashlib
            checksum = hashlib.sha224(name.encode('utf-8'))
        except ImportError:
           ...
[...]

PR #674, opened 2018-10-15
Refactor profiling

rhodrin:
Introduces a new basic (default) profiling option. The current 'basic' profiling has been renamed 'advanced'. The new basic option saves a lot of time in code which requires repeated calls to operator.

Note that this iteration has a modification to `operator.py` (see from line 448) to allow the 'dummy list' to be provided as an empty list. There is possible a different (better?) way of doing this? But providing a non-empty list with 'dummy' values does not seem like a nice solution.


devito/operator.py
```diff
@@ -447,8 +447,10 @@ def _profile_output(self, args):
             itershapes = [",".join(str(i) for i in its) for its in v.itershapes]
             if len(itershapes) > 1:
                 name = "%s<%s>" % (k, ",".join("<%s>" % i for i in itershapes))
-            else:
+            elif len(itershapes) > 0:
```
FabioLuporini: nitpicking: this can be == 0

rhodrin: Do you mean == 1 by the way?

FabioLuporini: yeah 😅

FabioLuporini:
Ah, some tests are failing because they need the operation count (`ops`). For those, and only for those, can you set the profile level back to the original level? you have several options to do this, probably the cleanest one is to write a decorator similar in spirit to `silencio` (but for profiling) and stick it on top of the failing tests

rhodrin:
Applying an appropriate fix results in error due to #658. 

Ideally that issue should be resolved prior to these changes being implemented but there are some work around options if we decide not to fix that first. Thoughts?

FabioLuporini:
> Applying an appropriate fix results in error due to #658.

yes, agreed. I just re-read the issue. I'm not sure what the codepy patch is exactly supposed to do, but I think we should go ahead and propose it on codepy's github, and see how it goes. Said that, I'm ok with merging the changes in #675 first. I just merged master into #675. If green, we can proceed and merge it; we can revisit the 'profiling' thing mentioned in there afterwards.

Thoughts?

rhodrin:
Sounds like a plan.
I believe with that in place this branch should go green also (after I update it).

FabioLuporini:
nice, merged

PR #675, opened 2018-10-15
Remove parameters from soname that do not affect the c level code

rhodrin:
This addresses #658 for environment variables that do not affect c level code. To address the issue other vars either a codepy patch or additional changes to the code will be required. See issue for details and possible codepy patch.

Possibly some more environment variables can be added to the exclude list? Also, some vars only affect the c-level code on certain settings - do we at this time want to try and address that?


devito/parameters.py
```diff
@@ -94,8 +94,10 @@ def name(self):
         return self._name
 
     def _signature_items(self):
-        items = sorted(it for it in self.items()
-                       if it[0] not in ['log_level', 'first_touch'])
+        # Note: we are discarding some vars that do not affect the c level
+        # code in order to avoid recompiling when such vars are modified
+        discard = ['profiling', 'autotuning', 'log_level', 'first_touch']
```
FabioLuporini: now that I think about it, 'profiling' actually affects the C code as Advisor does something different than basic/advanced

rhodrin: Hmm, ok. I'll look at this in more detail tomorrow then. Thanks.

PR #677, opened 2018-10-27
MPI: enable global indexing for distributed Data

FabioLuporini:
The ultimate goal is to enable users to access and manipulate MPI-distributed Data as if they were using a standard NumPy array.  This PR brings us one step closer to this objective.

### Features
* Enable basic indexing into MPI-distributed Data.  Now, with ``u.data[0]`` the rank that owns the global index 0 returns the actual value in ``u.data[0]``; all other ranks return None

* Enable slice-based indexing into MPI-distributed data. Now with ``u.data[41:75]`` the ranks that own some elements in the range ``[41, 75)`` return a Data containing the owned values (and only those); all other ranks return a 0-shape Data (eg, ``Data([], shape=(0, ...))``

* The previous indexing styles can be arbitrarily mixed just like with any pure NumPy array. So the following, for example, is supported
```
u.data[4, 31:53, 2]
u.data[4:, :, :21]
u.data[4, ..., 25]
```

* The nice thing is that _the distribution is propagated and remapped onto the views_. 
```
view = u.data[1:3, 1:3]
view[:]  = 1.
```
This is a **true** view, and the values in u.data are modified (as expected) individually by each rank

* pure NumPy arrays can be assigned to MPI-distributed Data. Consider:
```
a = np.ones(shape=(4, 4))
u.data[:] = a
```
if ``u.data`` is distributed over four ranks organised in a 2x2 grid, i.e., each rank having access to a (2, 2) region, then assigning ``u.data[:] = a`` causes the global (replicated) ``a`` to be split into (2, 2) chunks, which are in turn assigned to the rank-local (2, 2) data regions.  Eventually, this feature will be generalised to the case in which ``a`` is memory-mapped (maybe it already works out-of-the-box, haven't tried yet).

* ``u.data_with_halo`` now returns the **true** halo, ie the one which typically appears only on boundary ranks. The "inner" ghost regions, required for the halo exchanges, are masked away. There's still is a (currently private) data accessor that also includes the inner ghost regions, called ``_data_with_inhalo``.

* Element-wise operations (adds, muls, etc.) should all work

### TODO in the next PR
* Support for reduction operations, such as norms. Probably we will abstract them at the Function level (``u.norm``, rather than ``u.data.norm``); we might also forbid them at the Data level by raising an exception (at least if used with MPI)

### Tests
There are **lots** of new tests that you can use to understand how all this works.

### Plea for help
* Start implementing reduction operations, such as norms
* Write a jupyter notebook illustrating how this works

### Comments on the PR
~There are lots of commits, but in practice I've seen bigger PRs in the past. The reason for the many commits is that I had to rework stuff over and over again.~
EDIT: I was kidding. It's actually quite big

I haven't tried the seismic examples yet, so I assume they are still not working at this stage. But who knows..
And at this stage it should only be due to small missing features, not to bugs or lack of major components.

I moved the Data-related stuff to a separate folder, ``/data``


examples/seismic/model.py
```diff
@@ -399,43 +399,8 @@ def initialize_function(function, data, nbpml, pad_mode='edge'):
...
-            data_slices.append((lslice, rslice))
-            pad_widths.append((lpad, rpad))
-        except KeyError:
-            # Not a distributed dimension, so we make sure to get the entire array
-            data_slices.append((None, None))
-            pad_widths.append((0, 0))
-    data = data[[slice(*i) for i in data_slices]]
-    data = np.pad(data, pad_widths, pad_mode)
-
     function.data_with_halo[:] = data
```
mloubout: Why is this gone? Does it really do the same?

tests/test_data.py
```diff
@@ -2,198 +2,537 @@
...
+            assert False
+        except IndexError:
+            pass
+        try:
+            v.data[nt]
+            assert False
+        except IndexError:
+            pass
+
+    def test_logic_indexing(self):
```
mloubout: Not sure about the name. Aren't logic indices `v_mod[v_mod > 0]` for example?

tests/test_mpi.py
```diff
@@ -147,29 +169,30 @@ def test_halo_exchange_bilateral_asymmetric(self):
             0 0 0 0 0 0 0     0 0 0 0 0 0 0
         """
         grid = Grid(shape=(12, 12))
-        f = Function(name='f', grid=grid, space_order=(1, 2, 1))
+        x, y = grid.dimensions
 
-        distributor = grid.distributor
-        f.data[:] = distributor.myrank + 1
+        f = Function(name='f', grid=grid, space_order=(1, 2, 1))
```
mloubout: remind me what tuple space_order is?

examples/seismic/model.py
FabioLuporini: yes. The objective of this PR was hiding MPI almost completely away from user code. This is a symptom that this is indeed happening

tests/test_mpi.py
FabioLuporini: (space_order, halo_left_extent, halo_right_extent)

tests/test_data.py
FabioLuporini: it's the same name as before, I just put the pre-existing tests inside a class 

logic indexing is becase if your time buffer is of size 2, then you can access v[1], but also v[3], v[5], v[7], v[9], ...and they would all retrieve the same entry

PR #686, opened 2018-11-06
Fix Static Single Assignment (SSA) translation

FabioLuporini:
fixes + bunch of tests

I'm also dropping the now useless `group_expressions` routine, which might even improve the compilation speed. Like this.

### DIFF
--- a/devito/function.py
+++ b/devito/function.py
@@ -14,7 +14,7 @@ from devito.exceptions import InvalidArgument
 from devito.logger import debug, warning
 from devito.mpi import MPI, SparseDistributor
 from devito.parameters import configuration
-from devito.symbolics import indexify, retrieve_functions
+from devito.symbolics import indexify, retrieve_function_carriers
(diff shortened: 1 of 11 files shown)


FabioLuporini:
> Nothing important to note, but why is 'makeit_ssa' a better name than 'convert_to_SSA' ? Or at least to be 'make_it_ssa'. (?) . It is my first code review for Devito, so your comprehension is needed if I am missing something.

oh, nothing really important, I just thought that it's not really a conversion, as the word conversion makes me think you're changing an object into something else, while here we're actually creating a new list of equations in SSA form (in Devito, we basically never change state, we always recreate/rebuild). BTW, I **think** convert_into_SSA would have been more correct than convert_to_SSA. But as I said, nitpicking :)

rhodrin:
The changes and additions look fine to me (although admittedly I don't appreciate all that is going on in this PR).

FabioLuporini:
Merged.

PR #687, opened 2018-11-06
MPI: support for running isotropic acoustic

FabioLuporini:
This PR adds:

* some built-in Operators to compute norms/inner products/... of Functions when over MPI. These are meant to be used in place of NumPy linalg functions when MPI is used.
* bunch of tweaks to Data indexing and manipulation over MPI
* cleanup and minor fixes
* lots of new MPI-related tests

### DIFF
--- a/devito/builtins.py
+++ b/devito/builtins.py
@@ -1,0 +1,48 @@
+"""
+Built-in :class:`Operator`s provided by Devito.
+"""
+
+import devito as dv
+
+
+def assign(f, v=0):
+    """
+    Assign a value to a :class:`Function`.
+
+    Parameters
+    ----------
+    f : Function
+        The left-hand side of the assignment.
+    v : scalar, optional
+        The right-hand side of the assignment.
+    """
+    dv.Operator(dv.Eq(f, v), name='assign')()
+
+
+def smooth(f, g, axis=None):
+    """
+    Smooth a :class:`Function` through simple moving average.
+
+    Parameters
+    ----------
+    f : Function
+        The left-hand side of the smoothing kernel, that is the smoothed Function.
+    g : Function
+        The right-hand side of the smoothing kernel, that is the Function being smoothed.
+    axis : Dimension or list of Dimensions, optional
+        The :class:`Dimension` along which the smoothing operation is performed.
+        Defaults to ``f``'s innermost Dimension.
+
+    Notes
+    -----
+    More info about simple moving average available at: ::
+
+        https://en.wikipedia.org/wiki/Moving_average#Simple_moving_average
+    """
+    if g.is_Constant:
+        # Return a scaled version of the input if it's a Constant
+        f.data[:] = .9 * g.data
+    else:
+        if axis is None:
+            axis = g.dimensions[-1]
+        dv.Operator(dv.Eq(f, g.avg(dims=axis)), name='smoother')()
--- a/devito/builtins.py
+++ b/devito/builtins.py
@@ -2,6 +2,8 @@
 Built-in :class:`Operator`s provided by Devito.
 """
 
+from sympy import Abs, sqrt
+
 import devito as dv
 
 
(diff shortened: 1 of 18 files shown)


devito/builtins.py
```diff
@@ -0,0 +1,193 @@
...
+    Assign a value to a :class:`Function`.
+
+    Parameters
+    ----------
+    f : Function
+        The left-hand side of the assignment.
+    v : scalar, optional
+        The right-hand side of the assignment.
+    """
+    dv.Operator(dv.Eq(f, v), name='assign')()
```
mloubout: We may want to generate a standard library for something that simple at some point to avoid the jit part in some `lib` directory for example.
For future though probably don't need in this PR.

devito/builtins.py
```diff
@@ -0,0 +1,193 @@
...
+    """
+    Compute the norm of a :class:`Function`.
+
+    Parameters
+    ----------
+    f : Function
+        Input Function.
+    order : int, optional
+        The order of the norm. Defaults to 2.
+    """
```
mloubout: `if norm ==2 return pow(inner(f, f), 1/2)` no?

devito/function.py
```diff
@@ -169,21 +170,21 @@ def _allocate_memory(func):
         def wrapper(self):
             if self._data is None:
                 debug("Allocating memory for %s%s" % (self.name, self.shape_allocated))
-                self._data = Data(self.shape_allocated, self.indices, self.dtype,
-                                  self._decomposition, self._allocator)
+                self._data = Data(self.shape_allocated, self.dtype,
+                                  modulo=self._mask_modulo, allocator=self._allocator)
                 if self._first_touch:
-                    first_touch(self)
+                    assign(self, 0)
```
mloubout: Does this actually fix the first_touch issues we had? Don't rememberwhat the issue was

examples/seismic/source.py
```diff
@@ -117,10 +117,6 @@ def __new__(cls, **kwargs):
 
         return obj
 
-    def __init__(self, *args, **kwargs):
-        if not self._cached():
-            super(PointSource, self).__init__(*args, **kwargs)
-
```
mloubout: Why?

tests/test_mpi.py
```diff
@@ -1055,5 +1152,7 @@ class TestIsotropicAcoustic(object):
     # TestSparseFunction().test_ownership(((1., 1.), (1., 3.), (3., 1.), (3., 3.)))
     # TestSparseFunction().test_local_indices([(0.5, 0.5), (1.5, 2.5), (1.5, 1.5), (2.5, 1.5)], [[0.], [1.], [2.], [3.]])  # noqa
     # TestSparseFunction().test_scatter_gather()
-    TestOperatorAdvanced().test_nontrivial_operator()
+    # TestOperatorAdvanced().test_nontrivial_operator()
```
mloubout: Debug remainder?

devito/builtins.py
FabioLuporini: not sure how doable that is, as the Operator C code really depends on ``f``

devito/builtins.py
FabioLuporini: sure, but not sure what's the advantage of specialising it?

we could however add a test that checks this

tests/test_mpi.py
FabioLuporini: yeah I'll comment it out

I'll leave the ``if __main__`` thing because with ``tmpi`` that's what I use for debugging

examples/seismic/source.py
FabioLuporini: this is a no-op constructor basically, why should I leave it? it just calls the superclass constructor, which is gonna get called anyway w/o these lines

devito/grid.py
```diff
@@ -187,9 +187,9 @@ def spacing_map(self):
...
 
     @property
-    def origin_domain(self):
+    def origin_offset(self):
         """
-        Origin of the local (per-process) physical domain.
+        Offset of the local (per-process) origin from the domain origin.
         """
         grid_origin = [min(i) for i in self.distributor.glb_numb]
         assert len(grid_origin) == len(self.spacing)
```
mloubout: `tuple(i*h for i, h in zip(grid_origin, self.spacing))` only correct for 0 origin

replace by:

`tuple(o.data + i*h for o, i, h in zip(self.origin, grid_origin, self.spacing))`

FabioLuporini: no, this is supposed to be the **offset** from whatever the domain origin is
look where it's used: https://github.com/opesci/devito/blob/mpi-acoustic/devito/function.py#L1869
it's there only to translate coordinates

indeed, I tried running test_mpi::test_adjoint_F with the change you propose here, and it fails

FabioLuporini:
Rebased and added a number of small fixes

devito/grid.py
mloubout: um ok.

mloubout:
Can we just add a warning that examples will only work if src/rec provided at runtime. Except that GTG

PR #690, opened 2018-11-08
Patch data

rhodrin:
Fix for an `AttributeError` being encountered by i.e. non-numpy arrays in `Data(...)__setitem__(...)`.


devito/data/data.py
```diff
@@ -197,6 +197,11 @@ def __setitem__(self, glb_idx, val):
             else:
                 # `val` is replicated`, `self` is replicated -> plain ndarray.__setitem__
                 pass
+        elif isinstance(val, Iterable):
+            if self._is_mpi_distributed:
+                raise NotImplementedError("With MPI data can only be set "
+                                          "via scalars or numpy arrays")
+                pass
```
FabioLuporini: I suppose no need for `pass`

FabioLuporini:
thanks, this looks GTG once CI is green

devito/data/data.py
FabioLuporini: I removed it. Will merge this once green again

PR #637, opened 2018-09-05
OPS backend

btrb:
initial setup for creating OPS backend

PR #692, opened 2018-11-08
devito without mpi

mloubout:
Fix so that Devito still runs asbefore if MPI is not installed.


.travis.yml
```diff
@@ -39,7 +39,17 @@ matrix:
           packages:
             - gcc-4.9
             - g++-4.9
-      env: DEVITO_ARCH=gcc-4.9 DEVITO_OPENMP=1 OMP_NUM_THREADS=2 RUN_EXAMPLES=True INSTALL_TYPE=conda
+      env: DEVITO_ARCH=gcc-4.9 DEVITO_OPENMP=1 OMP_NUM_THREADS=2 RUN_EXAMPLES=True INSTALL_TYPE=conda MPI_INSTALL=1
```
FabioLuporini: why adding a whole new build?

Jenkinsfile
```diff
@@ -70,11 +111,13 @@ pipeline {
                          HOME="${WORKSPACE}"
                          DEVITO_BACKEND="yask"
                          DEVITO_OPENMP="0"
+						 INSTALL_MPI="1"
```
FabioLuporini: indentation ignored here and below

devito/mpi/distributed.py
```diff
@@ -19,9 +19,17 @@
...
-from mpi4py import MPI  # noqa
+try:
+    import mpi4py
+    mpi4py.rc(initialize=False, finalize=False)
+    from mpi4py import MPI  # noqa
+except ImportError:
+    class MPI(object):
+        COMM_NULL = None
+
+        def Is_initialized():
```
FabioLuporini: nitpicking: this should take cls in input and be marked as a @classmethod

mpi_requirements.txt
```diff
@@ -0,0 +1 @@
+mpi4py
```
FabioLuporini: I really think we should generalise this and rather call it ``requirements-optional.txt`` or something along these lines. This is because we have several non-compulsory packages that we may want to mark as optional

setup.py
```diff
@@ -14,6 +15,9 @@
     else:
         reqs += [ir]
 
+if os.environ.get('INSTALL_MPI') == '1':
```
FabioLuporini: you probably meant MPI_INSTALL ...

but this is probably not the way we should do it; I don't think ``setup.py`` should depend on an environment variable. Perhaps there's a way (an cmd line option?) to tell it to **also** pick the packages in ``requirements-optional.txt``

tests/conftest.py
```diff
@@ -17,10 +17,18 @@
 from devito.ir.iet import Iteration
 from devito.tools import as_tuple
 
+try:
```
FabioLuporini: don't we simply want ``from devito import MPI`` ?

tests/conftest.py
```diff
@@ -17,10 +17,18 @@
...
+    from mpi4py import MPI  # noqa
+    no_mpi = False
+except ImportError:
+    no_mpi = True
+    pass
 
 skipif_yask = pytest.mark.skipif(configuration['backend'] == 'yask',
                                  reason="YASK testing is currently restricted")
 
+skipif_mpi = pytest.mark.skipif(no_mpi, reason="mpi not installed")
```
FabioLuporini: instead of adding this, I'd prefer to leverage the more general mechanism in #637 : https://github.com/opesci/devito/pull/637/files#diff-484462fced51d1a06b1d93b4a44dd535R21

we could rename if ``skipif`` and allow it to digest ``yask``, ``ops``, ``nompi`` -- with "nompi", it does the check that if there's no MPI available, it skips the test

tests/conftest.py
FabioLuporini: in fact, we could rather add a function "mpi_available()" to ``devito/mpi/__init__.py`` that returns true/false depending on whether mpi4py/MPI is available on the system. Otherwise, we would have to repeat this trick in all module needing MPI...

mloubout: No we wouldn't, MPI exists even without mpi4py, and why `import MPI` would work. This one is just checking if mpi4py is installed but that's the only place like it except inside `mpi`

setup.py
mloubout: I know this only work for travis for now I'm tryin to figure out what to do

mpi_requirements.txt
mloubout: Optional requirements should be split into individual purpose in my opninion, installing all optional if you only want mpi for example is overkill. Moving everything to an ptional file would result in the same thing we have currently

.travis.yml
mloubout: To check it actually works if mpi is not installed

setup.py
```diff
@@ -14,6 +15,9 @@
     else:
         reqs += [ir]
 
+if os.system('which mpicc') == '0':
```
FabioLuporini: I don't like this.

How about this: https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies

FabioLuporini: same thing, but makes a long story short: https://stackoverflow.com/questions/10572603/specifying-optional-dependencies-in-pypi-python-setup-py

mpi_requirements.txt
FabioLuporini: in a big project you can have 20 optional packages, would you then keep ~20 different files, one for each of them? 

> installing all optional if you only want mpi for example is overkill.
splitting into compulsory+optional is a very good thing already; we cannot clearly support all possible installation combinations. 
* On CI, it would make sense to either install all optionals or none;
* If users don't want a specific subset of packages, then it's up to them. For example, instead of using conda, they could use the setup.py based installation and only install a specific subset of the desired features (see below for my other comment about setup.py)

https://stackoverflow.com/questions/3664478/optional-dependencies-in-a-pip-requirements-file

.travis.yml
mloubout: No extra build anymore

mpi_requirements.txt
mloubout: I mean sure, but this PR is not about this, this is only about making sure we don't break everyone's code forcing MPI.

tests/conftest.py
mloubout: Sure but not the point of this PR

tests/conftest.py
```diff
@@ -279,19 +287,18 @@ def wrapper(*args, **kwargs):
...
     """Run a test in parallel.
 
     :parameter item: The test item to run.
     """
+    # Support to run MPI tests
+    # This is partly extracted from:
+    # `https://github.com/firedrakeproject/firedrake/blob/master/tests/conftest.py`
+
+    mpi_exec = 'mpiexec'
+    mpi_distro = sniff_mpi_distro(mpi_exec)
```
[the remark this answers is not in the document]
mloubout: Breaks if mpi not installed why I moved it inside that is a parallel only section

Jenkinsfile
```diff
@@ -52,10 +54,12 @@ pipeline {
                      environment {
                          HOME="${WORKSPACE}"
                          DEVITO_OPENMP=0
+                         INSTALL_MPI=1
                      }
                      steps {
                          cleanWorkspace()
                          condaInstallDevito()
+						 condaInstallMPI()
```
FabioLuporini: screwed indentation? same below

mpi_requirements.txt
FabioLuporini: doesn't matter, we should do it properly

tests/conftest.py
FabioLuporini: at least let's call it ``skipif_nompi``...

Jenkinsfile
```diff
@@ -88,10 +94,12 @@ pipeline {
...
```
FabioLuporini: indentation

requirements-optional.txt
```diff
@@ -0,0 +1,2 @@
+mpi4py
```
FabioLuporini: @ggorman @mloubout probably dask/distributed should end up in this file too, and we should make sure that the relative examples won't run/won't fail if these packages aren't available

FabioLuporini: Obviously unrelated to this PR

setup.py
```diff
@@ -14,6 +18,17 @@
     else:
         reqs += [ir]
 
+
+opt_reqs = []
+opt_links = []
+for ir in optionals:
+    if ir[0:3] == 'git':
+        opt_links += [ir + '#egg=' + ir.split('/')[-1] + '-0']
```
FabioLuporini: it doesn't really matter in the setup file, but this is **not** the format we use in Devito to build/concatenate strings

tests/conftest.py
```diff
@@ -17,10 +17,16 @@
 from devito.ir.iet import Iteration
 from devito.tools import as_tuple
 
+try:
+    from mpi4py import MPI  # noqa
+except ImportError:
+    pass
```
FabioLuporini: MPI = None

.travis.yml
```diff
@@ -87,18 +87,21 @@ before_install:
...
       conda env create -q -f environment.yml python=$TRAVIS_PYTHON_VERSION;
       source activate devito;
-      pip install -e .;
+      if [[ $MPI_INSTALL == '1' ]]; then pip install -e .[extras]; else pip install -e .; fi
       conda list;
     fi
+
   # Install devito with pip
-  - if [[ $INSTALL_TYPE == 'pip_setup' ]]; then python setup.py install; fi
+  - if [[ $INSTALL_TYPE == 'pip_setup' ]]; then pip install setup.py install; fi
```
mloubout: Not to self, replace back to python

setup.py
mloubout: Thanks, what was supposed to be used my bad

FabioLuporini:
just leftover flake8 issues?

.travis.yml
```diff
@@ -87,7 +87,9 @@ before_install:
   - conda update -q conda
   # Useful for debugging any issues with conda
   - conda info -a
-  - sudo apt-get install -y -q mpich libmpich-dev
+  - if [[ "$MPI_INSTALL" == '1' ]]; then
+      sudo apt-get install -y -q mpich libmpich-dev;
```
FabioLuporini: BTW, now that I look at this -- we should **not** use sudo here when we have a conda environment

.travis.yml
```diff
@@ -97,8 +99,13 @@ install:
       pip install -e .;
       conda list;
     fi
+  - if [[ "$MPI_INSTALL" == '1' ]]; then
```
FabioLuporini: I suppose that this `if` can be merged with the one at line 108

FabioLuporini:
@tmbgreaves note that this PR is reworking CI so that at least one build attempts to install Devito and run the test suite without MPI . In terms of changes to Jenkinsfile and .travis.yml, it's nothing big, but we should replicate this in #714 , which I believe should land after this PR

tmbgreaves:
> @tmbgreaves note that this PR is reworking CI 

Thanks - implemented in Pipelines

devito/mpi/distributed.py
```diff
@@ -19,9 +19,24 @@
...
+    from mpi4py import MPI  # noqa
+except ImportError:
+    class MPI(object):
+        COMM_NULL = None
+
+        @classmethod
+        def Is_initialized(cls):
+            return False
+
+        def _sizeof(comm):
```
FabioLuporini: this should be a classmethod, while the signature should be `_sizeof(cls, comm)`

devito/mpi/distributed.py
```diff
@@ -19,9 +19,24 @@
...
+            return None
```
FabioLuporini: better to return 0

mloubout: I prefer None, size of undefined shouldn't be zero, but that's just me

devito/mpi/distributed.py
mloubout: I digged into 'mpi4py` and they define `_sizeof` as `_sizeof(arg)` not as a class method

FabioLuporini:
Merged.

PR #694, opened 2018-11-09
Revamp autotuning

FabioLuporini:
* 'runtime' mode for 'core' so that we don't waste time
* more tests
* enhanced ``op.apply`` interface (now can pass basic/aggressive/... directly as ``op.apply(autotune=aggressive)`` instead of setting them globally via ``configuration``)

see tests for more info

### DIFF
--- a/devito/__init__.py
+++ b/devito/__init__.py
@@ -62,11 +62,7 @@ def _at_callback(val):  # noqa
         level, mode = val
     if level == 'off':
         level = False
-    if configuration['backend'] == 'core' and mode == 'runtime':
-        warning("Unsupported auto-tuning mode `runtime` with backend `core`")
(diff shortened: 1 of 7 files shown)


mloubout:
So  if I understand that correctly, autontuning is now done on the fly over the first nblock time stpes?

FabioLuporini:
only when setting the runtime mode. The number of time steps spent autotuning depend on several factors (numbers of combinations attempted, ie basic vs aggressive mode, ``options[at-squeezer]``, and other stuff as well)

mloubout:
Ok cool, GTG for me then.

FabioLuporini:
Note: this now depends on #707 , and clearly I want #707 to go in first

PR #698, opened 2018-11-12
Revamp src rec

mloubout:
Rework the source and receiver setup in our seismic example. The current setup is quite intricate and relies too much on copying properties from other src/rec at object creations which breaks a lot of examples with MPI for example.
In this PR:
 - Create a `Geometry` class that contains all the acquisition (src/rec position, time, ..) information so that  it can be used to initialize source and receivers. This class does not use any MPI or low level Devito but only store "outside" information to carry around
- Some cahced properties
- All examples updated accordingly

The new Class kinda reflect how it would work in practice with a segy file metadata to initialize object rather than devito src/rec properties 

On top of #687


devito/function.py
```diff
@@ -1189,7 +1189,7 @@ def __shape_setup__(cls, **kwargs):
             shape = (glb_npoint[grid.distributor.myrank],)
         return shape
 
-    @property
+    @cached_property
```
FabioLuporini: as discussed already, we cannot afford these yet (not unless we ``del`` them each time the coordinates are written to)

examples/seismic/acoustic/acoustic_example.py
```diff
@@ -4,47 +4,43 @@
 from devito.logger import info
 from devito import Constant, Function, smooth
 from examples.seismic.acoustic import AcousticWaveSolver
-from examples.seismic import demo_model, TimeAxis, RickerSource, Receiver
+from examples.seismic import demo_model, Acquisition_geometry
```
FabioLuporini: I prefer Geometry

In general, never use '_' in class names

examples/seismic/acoustic/wavesolver.py
```diff
@@ -46,29 +44,29 @@ def __init__(self, model, source, receiver, kernel='OT2', space_order=2, **kwarg
     @memoized_meth
     def op_fwd(self, save=None):
         """Cached operator for forward runs with buffered wavefield"""
-        return ForwardOperator(self.model, save=save, source=self.source,
-                               receiver=self.receiver, kernel=self.kernel,
+        return ForwardOperator(self.model, save=save, geometry=self.geometry,
+                               kernel=self.kernel,
```
FabioLuporini: regroup arguments , here and below

examples/seismic/elastic/operators.py
```diff
@@ -170,7 +170,7 @@ def src_rec(vx, vy, vz, txx, tyy, tzz, model, source, receiver):
     return src_expr + rec_term1 + rec_term2
 
 
-def ForwardOperator(model, source, receiver, space_order=4,
+def ForwardOperator(model, geometry, space_order=4,
```
FabioLuporini: single line

examples/seismic/elastic/operators.py
```diff
@@ -183,8 +183,8 @@ def ForwardOperator(model, source, receiver, space_order=4,
                  indices (last three time steps)
     """
 
-    pde = kernels[model.grid.dim](model, space_order, source.nt if save else None,
-                                  source, receiver)
+    pde = kernels[model.grid.dim](model, space_order, geometry.nt if save else None,
```
FabioLuporini: I guess it doesn't fit on a single line?

examples/seismic/utils.py
```diff
@@ -1,10 +1,118 @@
...
+            return PointSource(name='src', grid=self.grid,
+                               time_range=self.time_axis, npoint=self.nsrc,
+                               coordinates=self.src_positions)
+        else:
+            return SRC_TYPES[self.src_type](name='src', grid=self.grid, f0=self.f0,
+                                            time_range=self.time_axis, npoitn=self.nsrc,
+                                            coordinates=self.src_positions)
+
+
+SRC_TYPES = {'Wavelet': WaveletSource, 'Ricker': RickerSource, 'Gabor': GaborSource}
```
FabioLuporini: no need to be capital letters . I'd call it (following Devito's convention) ``source_mapper``

examples/seismic/utils.py
```diff
@@ -1,10 +1,118 @@
...
+    - receiver positions and number
+    - source positions and number
+
+    In practice this would only point to a segy file with the
+    necessary information
+    """
+
+    def __init__(self, model, rec_pos, src_pos, t0, tn, **kwargs):
+        """
+        In practice woyuld be __init__(segyfile) and all below parameters
```
FabioLuporini: would (fixable in another PR)

tests/test_adjoint.py
```diff
@@ -48,43 +48,24 @@ def test_adjoint_F(self, mkey, shape, kernel, space_order, nbpml):
...
-        # Define receiver geometry (same as source, but spread across x)
-        rec = Receiver(name='rec', grid=model.grid, time_range=time_range, npoint=nrec)
-        rec.coordinates.data[:, 0] = np.linspace(0., model.domain_size[0], num=nrec)
-        rec.coordinates.data[:, 1:] = src.coordinates.data[0, 1:]
-
-        # Create solver object to provide relevant operators
-        solver = AcousticWaveSolver(model, source=src, receiver=rec,
-                                    kernel=kernel, space_order=space_order)
+        # Create solver from preset
+        solver = acoustic_setup(shape=shape, spacing=[15. for _ in shape], kernel=kernel,
```
FabioLuporini: nice

btrb:
Travis is green for PR689. Jenks is taking a loong while thou..

examples/seismic/utils.py
```diff
@@ -1,10 +1,117 @@
...
+    def dtype(self):
+        return self.grid.dtype
+
+    @property
+    def rec(self):
+        return Receiver(name='rec', grid=self.grid,
+                        time_range=self.time_axis, npoint=self.nrec,
+                        coordinates=self.rec_positions)
+
+    @property
```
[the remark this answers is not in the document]
mloubout: If positions change need a new one, with cahced would always return the initial one

examples/seismic/utils.py
```diff
@@ -1,10 +1,117 @@
...
+        if self.src_type is None:
+            return PointSource(name='src', grid=self.grid,
+                               time_range=self.time_axis, npoint=self.nsrc,
+                               coordinates=self.src_positions)
+        else:
+            return sources[self.src_type](name='src', grid=self.grid, f0=self.f0,
+                                          time_range=self.time_axis, npoitn=self.nsrc,
+                                          coordinates=self.src_positions)
+
+    _pickle_args = ['model', 'rec_positions', 'src_positions', 't0', 'tn']
```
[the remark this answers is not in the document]
mloubout: Yes, pretty sure this doesn't work, was just tryin to see how to get the dask tuto to run

mloubout:
Gonna remove the pickling for now, not sure we want it there anyway.

FabioLuporini:
> not sure we want it there anyway.

TBD with @ggorman

tests/test_mpi.py
```diff
@@ -1094,7 +1093,7 @@ class TestIsotropicAcoustic(object):
     """
 
     @pytest.mark.parametrize('shape,kernel,space_order,nbpml,save,Eu,Erec,Ev,Esrca', [
-        ((60, ), 'OT2', 4, 10, False, 976.825, 9372.604, 18851836.075, 47002871.882),
+        ((60, ), 'OT2', 4, 10, False, 385.853, 12937.250, 63818503.321, 101159204.362),
```
FabioLuporini: why are these changing, but not the ones below?

mloubout: Because the previous setup (don't know when changed) was messing up 1D only because the if dim>1 was removed. Once back 1D needs new values

FabioLuporini:
Nice, merging...

FabioLuporini:
Merged.

PR #699, opened 2018-11-12
Docker: Add env variables inside Docker environment, and update README

navjotk:
Added environment variables inside the docker entry point script to make it easy to get a shell with devito using docker. 

Added instructions in README to get such a shell


FabioLuporini:
Merged.

PR #704, opened 2018-11-13
RFC - Tutorial remakes

georgebisbas:
Comment for review @FabioLuporini @mloubout 

- Added batches for showing wave propagation-evolution in Burgers

Please ignore commits of
@georgebisbas
Merge pull request #(xxx) from opesci/master 
They were before meeting rebase in my fork...


examples/cfd/04_burgers.ipynb
```diff
@@ -76,12 +81,12 @@
     "init_hat(field=u, dx=dx, dy=dy, value=2.)\n",
     "init_hat(field=v, dx=dx, dy=dy, value=2.)\n",
     "\n",
-    "plot_field(u)"
+    "plot_field(u) # Finction for plotting field"
```
georgebisbas: F**u**nction

examples/cfd/04_burgers.ipynb
```diff
@@ -39,20 +39,25 @@
     "%matplotlib inline\n",
     "\n",
     "# Some variable declarations\n",
-    "nx = 41\n",
-    "ny = 41\n",
-    "nt = 120\n",
+    "nx = 61  # Grid size on x axis\n",
+    "ny = 61  # Grid size on y axis\n",
```
mloubout: Number of grid points. Grid size is confusing some people use it for spacing.

examples/cfd/04_burgers.ipynb
```diff
@@ -39,20 +39,25 @@
...
     "c = 1\n",
     "dx = 2. / (nx - 1)\n",
     "dy = 2. / (ny - 1)\n",
     "sigma = .0009\n",
     "nu = 0.01\n",
-    "dt = sigma * dx * dy / nu"
+    "dt = sigma * dx * dy / nu\n",
+    "\n",
+    "batches = 5 # Batches of timesteps\n",
+    "# Total number of timesteps is `batches*nt`\n",
```
mloubout: I personally would prefer `nt = batches*batch_size`

examples/cfd/04_burgers.ipynb
```diff
@@ -252,9 +257,9 @@
     "op = Operator([update_u, update_v] + bc_u + bc_v)\n",
     "\n",
     "# Execute the operator for a number of timesteps\n",
-    "op(time=nt, dt=dt, a=nu)\n",
-    "\n",
-    "plot_field(u.data[0])"
+    "for iters in range(batches):\n",
+    "    op(time=nt, dt=dt, a=nu)\n",
```
mloubout: I would add `time_m=iters*batch_size, time_M=(iters+1)*batches -1` as this has a lot of assumtions from 1st order in time but would be different for different orders.

examples/cfd/04_burgers.ipynb
```diff
@@ -39,20 +39,25 @@
     "%matplotlib inline\n",
     "\n",
     "# Some variable declarations\n",
-    "nx = 41\n",
-    "ny = 41\n",
-    "nt = 120\n",
+    "nx = 61  # Grid size on x axis\n",
+    "ny = 61  # Grid size on y axis\n",
+    "nt = 640 # Number of timesteps\n",
```
mloubout: So propagate 25 times longer right?

mloubout: what do you mean 5.6. nt is 5 times bigger and 5 batches of it so that's 25 times more

georgebisbas: It is 5* ( 640/120 ) = 5 *  5.333 = 26.666 . Both of us confused I guess..

examples/cfd/04_burgers.ipynb
georgebisbas: Mathias I do not understand well what you mean, can you add here 2-3 lines of code?

mloubout: so mainly:
`iters=0` runs from `t=0` to `t=nt`
`iters=1` runs from `t=nt` to `t=nt + nt`
`iters=2` runs from `t=2*nt` to `t=2*nt + nt`
...

Which doesn't really show here

georgebisbas: You said earlier that `nt = batches*batch_size` so I guess you mean:
`batches=0` runs from `t=0` to `t=batch_size`
`batches=1` runs from `t=batch_size` to `t=batch_size + batch_size`
`batches=2` runs from `t=2*batch_size` to `t=2*batch_size + batch_size`

Up to `batches*batch_size`

mloubout: yes

georgebisbas:
@mloubout if I am not mistaken I should have done your proposed fixes...

examples/cfd/04_burgers.ipynb
```diff
@@ -39,9 +39,15 @@
     "%matplotlib inline\n",
     "\n",
     "# Some variable declarations\n",
-    "nx = 41\n",
-    "ny = 41\n",
-    "nt = 120\n",
+    "nx = 61  # Number of grid points on x axis\n",
+    "ny = 61  # Number of grid points on y axis\n",
+    "\n",
+    "batches = 5 # Batches of timesteps, increase number of batches to expand evolution in time\n",
```
FabioLuporini: increase it

examples/cfd/04_burgers.ipynb
```diff
@@ -39,9 +39,15 @@
...
+    "# A figure of the wave state will be produced for each batch.\n",
+    "# Increase number of batches to expand evolution in time\n",
```
FabioLuporini: redundant?

examples/cfd/04_burgers.ipynb
```diff
@@ -76,12 +82,12 @@
     "init_hat(field=u, dx=dx, dy=dy, value=2.)\n",
     "init_hat(field=v, dx=dx, dy=dy, value=2.)\n",
     "\n",
-    "plot_field(u)"
+    "plot_field(u) # Function used to plot field\n"
```
FabioLuporini: useless comment, at this point ? I mean, this is the fourth tutorial so they must have seen/used it already?

examples/cfd/04_burgers.ipynb
georgebisbas: Need to resolve multiple output problem

examples/cfd/04_burgers.ipynb
georgebisbas: Yes...

examples/cfd/04_burgers.ipynb
```diff
@@ -252,9 +261,10 @@
     "op = Operator([update_u, update_v] + bc_u + bc_v)\n",
     "\n",
     "# Execute the operator for a number of timesteps\n",
-    "op(time=nt, dt=dt, a=nu)\n",
-    "\n",
-    "plot_field(u.data[0])"
+    "for batch_no in range(batches+1):\n",
+    "    op(time=batch_size, dt=dt, a=nu)\n",
```
mloubout: I still don't really like this as it will run `t=0,....,batch_size` at each call. SO it works fine because the wavefield is not saved but it would only recompute the same part of the time history over and over with `save=nt` instead of continuing from previous stop.

georgebisbas: Is there a way to take a `snapshot` at some point of the procedure?
What do you suggest? Should we maybe only show the last batch?

georgebisbas:
Commit d11ae37 is just for testing -nbval problems

georgebisbas:
Tests are failing at
 ```py.test --nbval examples/cfd```
Because we produce multiple images it is reasonable due to this:
https://nbval.readthedocs.io/en/latest/
The py.test program doesn’t usually collect notebooks for testing; by passing the --nbval flag at the command line, the IPython Notebook Validation plugin will collect and test notebook cells, comparing their outputs with those saved in the file.

examples/cfd/04_burgers.ipynb
georgebisbas: @FabioLuporini @mloubout any thoughts about this?

FabioLuporini: wdym by "procedure" ?

georgebisbas: The operator runtime.
If we execute for all timesteps we are not able to have a snap for every 100 timesteps for example.

FabioLuporini: I guess this is one of the reasons we keep both batches and timesteps? One could lower the timestep, while suitably increasing the number of batches, to get more frequent snapshots?

georgebisbas: This is true but if we are fine with what mloubout pointed 4 posts before then OK.

FabioLuporini: ah, I see. Perhaps @mloubout is suggesting that we should rather use ``time_m=... , time_M=...`` instead of ``time=batch_size`` ? this way it would work with save=nt too I suppose? do u see what I mean?

georgebisbas: I tried to understand what to use from here [TimeFunction Devito Documentation](https://www.devitoproject.org/devito/devito.html?highlight=timefunction#devito.base.TimeFunction)

but I don't quite get why to use save='nt' as it doesn't seem to work...

georgebisbas: So, I took into account Listing 6 of Devito architecture paper in order to save the state of our wave every `batch_size` timesteps and print this state as an image. My problem is that I cannot find proper documentation for that. Can anyone point me to an example that takes advantage of Listing 6? Additionally, will my method do any additional operations compared to what @mloubout proposes?

rhodrin:
Some quick comments (still digesting the meat of the notebooks):
- What about renaming the diffusion examples Example 3, part A: Suitable title and Example 3, part B: Suitable title.
- 3b has some repetition. It should be made clear that this is a continuation of part A and some repetition removed.
- Clearly stating when ν is a constant and when v=v(x,y) would help.
- In part A is there a reason we're not comparing with an analytical solution and would it help to just show some norm of the difference between the 'user' scheme and the devito scheme just to 'convince' the user? Same for part B. 
Will comment more soon.

georgebisbas:
>     * What about renaming the diffusion examples Example 3, part A: Suitable title and Example 3, part B: Suitable title.
**Fixed**
>     * 3b has some repetition. It should be made clear that this is a continuation of part A and some repetition removed.
I think it shoud not be a continuation rather than something autonomous.
>     * Clearly stating when ν is a constant and when v=v(x,y) would help.
**Fixed**
>     * In part A is there a reason we're not comparing with an analytical solution and would it help to just show some norm of the difference between the 'user' scheme and the devito scheme just to 'convince' the user? Same for part B.
In fact this is a general policy followed in the whole cfd examples before me. If you think that somethink like this should be added, no problem for me.

rhodrin:
In example 3 (part A) the final two figures look like copies of the preceding two? The only difference seems to be:
`init_hat(field=u.data[0], dx=dx, dy=dy, value=1.)` vs `init_hat(field=u2.data[0], dx=dx, dy=dy, value=2.)` but the boundary conditions haven't been updated to reflect this.

georgebisbas:
**Fixed** Thanks for the comments. It should be `value=1.` 
Neumann case should be mentioned in another tutorial and not in this particular one. Don't forget that it should be just modified without getting derailed.
Neumann tutorial will bw addressed to this: [Issue 606](https://github.com/opesci/devito/issues/606).
The change I make just aim to add a better optical/fancy tutorial for 3a and non-uniform for 3b.

FabioLuporini:
closing as reopened as #735

PR #707, opened 2018-11-14
pickling fixes

FabioLuporini:
A bunch of pickling-related fixes plus tests


devito/function.py
```diff
@@ -256,9 +256,12 @@ def staggered(self):
     @cached_property
     def shape(self):
         """
-        Shape of the domain associated with this :class:`TensorFunction`.
-        The domain constitutes the area of the data written to in a
-        stencil update.
+        Shape of the domain region. The domain constitutes the area of the
+        data written to by an :class:`Operator`.
```
mloubout: Is it really the write only region or the read/write one?

devito/function.py
```diff
@@ -812,14 +840,14 @@ def __indices_setup__(cls, **kwargs):
     def __shape_setup__(cls, **kwargs):
         grid = kwargs.get('grid')
         dimensions = kwargs.get('dimensions')
-        shape = kwargs.get('shape')
+        shape = kwargs.get('shape', kwargs.get('shape_global'))
```
mloubout: that look odd to me a `get` inside a `get`

devito/function.py
```diff
@@ -913,7 +941,7 @@ def avg(self, p=None, dims=None):
 
     # Pickling support
     _pickle_kwargs = TensorFunction._pickle_kwargs +\
-        ['space_order', 'shape', 'dimensions']
+        ['space_order', 'shape_global', 'dimensions']
```
mloubout: Only pickle local shape?

FabioLuporini: the global shape should be pickled -- the local shape is inferred individually by each MPI rank from the global shape and from the decomposition. This change fixes exactly this (pickle global, not local)

devito/function.py
FabioLuporini: nesting ``kwargs.get`` essentially means ``A or B or C or ... or None``. It's fine

devito/function.py
FabioLuporini: write; normally, when you got a stencil, you read from the halo region as well

FabioLuporini:
> can we add back the coordinate fix in functions pickling too?
Remind me what that was ? I suspect what we did in Anaheim was a temporary hack?

FabioLuporini:
> As this is a pickling fix, can we add back the coordinate fix in functions pickling too?

now PointSource, and more generally SparseFunctions, do pickle their coordinates. Tests added.

FabioLuporini:
> As this is a pickling fix, can we add back the coordinate fix in functions pickling too?

since this has been addressed, I'm landing the PR. Merging...

FabioLuporini:
Merged.

Issue #709, opened 2018-11-15
one-sided time stencils using space order instead of time order?


tjb900:
Hi!  Hopefully this one isn't too tricky.  On current master 7da13ecbee.

Cheers,
Tim

```
In [1]: import devito

In [2]: grid = devito.Grid(shape=(10,10,10))

In [3]: u = devito.TimeFunction(name='u', grid=grid, space_order=8, time_order=2)

In [4]: u.dt                                                                                                                                                                                                  
Out[4]: -0.5*u(t - dt, x, y, z)/dt + 0.5*u(t + dt, x, y, z)/dt

In [5]: u.dtl                                                                                                                                                                                                 
Out[5]: 0.45*u(t, x, y, z)/dt - 0.00357142857*u(t - 5*dt, x, y, z)/dt + 0.0357142857*u(t - 4*dt, x, y, z)/dt - 0.166666667*u(t - 3*dt, x, y, z)/dt + 0.5*u(t - 2*dt, x, y, z)/dt - 1.25*u(t - dt, x, y, z)/dt + 0.5*u(t + dt, x, y, z)/dt - 0.0714285714*u(t + 2*dt, x, y, z)/dt + 0.00595238095*u(t + 3*dt, x, y, z)/dt

This generates OOB access by the time stencil in the 'dtl' (and 'dtr' case), which later manifests as a segfault.  The size of these seems to be determined by the space order rather than the time order?

FabioLuporini:
@mloubout ?

mloubout:
Fixed in #711 

Upwind/downwind not implemented yet though, only shifted by 1-2 pioints left or right centered FD, if have a name idea for upwind/downwind I'll add it to the PR too

FabioLuporini:
fixed, closing

PR #711, opened 2018-11-15
order fix for left/right

mloubout:
Fixes shifted order for time derivatives left and right #709 

Reminder:
- these are not upwind/downwind FD for slightly shifted by 1 grid point non-centered FD


FabioLuporini:
@tjb900

PR #714, opened 2018-11-16
Set up CI with Azure Pipelines

tmbgreaves:
Close to final configuration - comments appreciated from any interested developers.


tmbgreaves:
Experimental branch, NOT yet ready for merging

azure-pipelines.docker.yml
```diff
@@ -0,0 +1,140 @@
...
+        echo "Setting DEVITO_OPENMP=${DEVITO_OPENMP} and OMP_NUM_THREADS=${OMP_NUM_THREADS}"
+      fi
+
+      if [ $(testMethod) == 'pip' ] ; then
+        export PYTHONPATH=${PYTHONPATH}:/devito/lib/python3.6/site-packages
+        python setup.py test ; 
+      elif [ $(testMethod) == 'conda' ] ; then
+        source activate devito ;
+        py.test --cov devito tests/ ;
+        python examples/seismic/benchmark.py test -P tti -so 4 -a -d 20 20 20 -n 5 ;
```
FabioLuporini: when the devito backend is either `yask` or `ops`, this line up until line 133 (included) shouldn't actually be run/tested

azure-pipelines.docker.yml
```diff
@@ -0,0 +1,140 @@
...
+        python examples/seismic/acoustic/acoustic_example.py --constant --full ;
+        python examples/misc/linalg.py mat-vec mat-mat-sum transpose-mat-vec ;
+        python examples/seismic/tti/tti_example.py -a ;
+        python examples/seismic/tti/tti_example.py -a --noazimuth ; 
+        python examples/seismic/elastic/elastic_example.py ; 
+        py.test --nbval examples/cfd ; 
+        py.test --nbval examples/seismic/tutorials/0[1-3]* ; 
+        py.test --nbval examples/compiler ;
+        codecov ;
+        sphinx-apidoc -f -o docs/ examples ;
```
FabioLuporini: is it OK to build the docs on all builds?

azure-pipelines.docker.yml
```diff
@@ -0,0 +1,140 @@
...
+      ops-conda-gcc7:
+        containerName: devito-conda-gcc7
+        testMethod: 'conda'
+        devitoBackend: 'OPS'
+      conda-gcc8: 
+        containerName: devito-conda-gcc8
+        testMethod: 'conda'
+
+  container: $[ variables['containerName'] ]
+  timeoutInMinutes: 180
```
FabioLuporini: I suppose we can lower it to 120 ?

FabioLuporini:
ah, one more thing, I guess we want both `branch` and `pr-merge`

azure-pipelines.docker.yml
tmbgreaves: Yes, that's easy to add in.

azure-pipelines.docker.yml
tmbgreaves: Thanks - I'll update accordingly.

azure-pipelines.docker.yml
tmbgreaves: Yes, I think the default 60 is in the right ballpark for the expected runtime on the free Pipelines nodes, so 120 should be very safe.

tmbgreaves:
It's entirely independent of the Jenkinsfile but is intended to do everything the Jenkinsfile does, hopefully in a nicer way if I've got it right :-) The gcc version is set when the container is built, passed as an argument to the docker build command and then set inside the docker file. update-alternatives is used to direct symlinks (ie, 'gcc') to the required version.

I'm afraid I'm not at all expert on codecov so would appreciate input into what I can do to best make use of it.

I'll drop the Jenkinsfile with this commit.

I'm seeing quite a lot of problems with the Azure<->dockerhub connection, which I think is completely avoidable by using Azure container registries instead of dockerhub. This has the disadvantage of making the pipelines containers harder to get hold of, though I think that's a much lower priority than CI reliability.

tmbgreaves:
I *think* it's doing both branch and pr-merge but isn't posting both as checks, just updating the commit 'dot'. I'm not immediately sure how to get it to post the branch as a check along the lines of the Travis report.

FabioLuporini:
> I'm afraid I'm not at all expert on codecov so would appreciate input into what I can do to best make use of it.

oh, it's just a tool that tells you the percentage of the codebase covered by your testing infrastructure. So it makes sure that with a PR you're not accidentally lowering the coverage by eg running fewer tests or leaving too much dead code around

Dockerfile.pipelines
```diff
@@ -18,9 +21,21 @@ RUN apt-get -y dist-upgrade
 RUN apt-get -y install wget bzip2 git make
 
 # Default gcc version to install; can be overridden in Jenkinsfile
```
FabioLuporini: probably you mean something else instead of Jenkinsfile?

Dockerfile.pipelines
```diff
@@ -40,7 +55,10 @@ RUN if [ -f /usr/bin/gcov-tool-$gccvers ] ; then \
     fi
 
 # Install MPI
-RUN apt-get install -y -q mpich libmpich-dev
+RUN if [ $useOpenMP != "true" ] ; then \
```
FabioLuporini: it seems that here we install MPI in all builds that won't use OpenMP, while [here](https://github.com/opesci/devito/blob/2de68cd5da086c1c0b888b6aca80586881527cfc/.travis.yml) the situation seems slightly different. At least the OpenMP build installs/uses MPI

Dockerfile.pipelines
```diff
@@ -50,3 +68,34 @@ RUN conda update -q conda
...
+
+RUN if [ $installWithPip == "true" ] ; then \ 
+      python setup.py install --user ; \
+    else \
+      conda env create -q -f environment.yml ; \
+      source activate devito ; \
+      echo $PATH ; \
+      pip install -e . ; \
+      pip install pytest-xdist ; \
+      if [ $useOpenMP != "true" ] ; then \
```
FabioLuporini: the price to pay is introducing an extra flag, but this should actually be something like "if useMPI: then ..."

azure-pipelines.sh
```diff
@@ -0,0 +1,42 @@
...
+  echo "Setting DEVITO_OPENMP=${DEVITO_OPENMP} and OMP_NUM_THREADS=${OMP_NUM_THREADS}"
+fi
+
+if [ "x${testWithPip}" == 'xtrue' ] ; then
+  python setup.py test ; 
+else
+  source activate devito ;
+  py.test --cov devito tests/ ;
+  if [ ! -v DEVITO_BACKEND ] ; then 
+    python examples/seismic/benchmark.py test -P tti -so 4 -a -d 20 20 20 -n 5 ;
```
FabioLuporini: all of my bash ignorance, again, so apologies in advance for the dumb question:

are you sure that **if** there's error in the first `python examples/...`, then it is actually reported? 

[here](https://github.com/opesci/devito/blob/master/.travis.yml#L119) we had to wrap each `python examples/...` within a separate ``if``

Dockerfile.pipelines
tmbgreaves: Ah, yes, I'd missed that the gcc5 build has neither - good spot. I'll fix that :-)

azure-pipelines.sh
tmbgreaves: The 'set -e' takes care of that, though it occurs to me on thinking about this again that this may not be the behaviour you want. The options are:

- Use 'set -e' so the script exits with error on the first error it encounters, or..
- Use 'set -o pipefail' so the script continues on error but exits with error if any command in the script returns an error

I guess you want the second of those behaviours ... ?

Dockerfile.pipelines
tmbgreaves: Oops, yes :-)

Dockerfile.pipelines
tmbgreaves: Good thought - done.

Dockerfile.pipelines
```diff
@@ -0,0 +1,117 @@
...
+ARG DEVITO_OPENMP=0
+ENV DEVITO_OPENMP=$DEVITO_OPENMP
+ENV OMP_NUM_THREADS=2
+
+# Run examples?
+ARG RUN_EXAMPLES='false'
+ENV RUN_EXAMPLES=$RUN_EXAMPLES
+
+# Ensure a non-interactive matplotlib backend
+ENV MPLBACKEND=agg
```
FabioLuporini: not sure if actually still any useful?

Dockerfile.pipelines
```diff
@@ -0,0 +1,117 @@
...
+      pip install -e . ; \
+      if [ $DEVITO_MPI == '1' ] ; then \
+        pip install -r requirements-optional.txt ; \
+      fi ; \
+      pip install pytest-xdist ; \
+      if [ $DEVITO_BACKEND == "yask" ] ; then \
+        conda install swig ; \ 
+        git clone https://github.com/opesci/yask.git /tmp/yask ; \
+        pushd /tmp/yask ; \ 
+          make compiler-api ; \
```
FabioLuporini: this can probably be turned into `make -j compiler-api` to make it faster, since we can exploit all cores on the node

azure-pipelines.py
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python
+
+import os
+import subprocess
+
+err=[]
+
+if os.environ.get('testWithPip') == 'true':
+    err.append(subprocess.run("python setup.py test", shell=True).returncode)
```
FabioLuporini: I think this whole file gets a bit less verbose if we move ``err.append(subprocess.run(CMD).retcode)`` to a separate function taking as input the parameter CMD

Dockerfile.pipelines
tmbgreaves: Gerard commented that this would be a useful backstop in case we have problems along similar lines in future (this is far from the first time I've hit matplotlib backend issues...) and I'm inclined to agree with him unless there are concrete reasons to take it out :-)

Dockerfile.pipelines
tmbgreaves: Thanks - done.

azure-pipelines.py
tmbgreaves: Thanks! This is the kind of thing that I as a sysadmin rather than programmer tend to miss :-)

Implemented, I think the correct way but please educate me if I'm writing bizarre python...

tmbgreaves:
Confirming here - failing tests return a trace, and failing flake8 returns errors too. Unfortunately the default previews on the main build page are not correct, but opening full logs gives the real error output.

PR #715, opened 2018-11-17
DLE refactoring + misc cleanup

FabioLuporini:
This PR drops quite a lot of ancient, obsolete, horrible code, including the hacky "DLEArgument". 

Autotuning is refactored, and now looks much nicer and simpler to follow. It also introduces support for autotuning the number of threads in an OMP region.

depends on #694

### DIFF
--- a/devito/core/autotuning.py
+++ b/devito/core/autotuning.py
@@ -1,12 +1,11 @@
 from collections import OrderedDict
-from itertools import combinations
-from functools import reduce
-from operator import mul
+from itertools import chain, combinations, product
 import resource
+import psutil
 
-from devito.dle import BlockDimension
-from devito.ir import Backward, Iteration, FindNodes, FindSymbols
-from devito.logger import perf, warning
+from devito.dle import BlockDimension, NThreads
+from devito.ir import Backward, Iteration, FindNodes
+from devito.logger import perf, warning as _warning
 from devito.parameters import configuration
 from devito.symbolics import evaluate
 from devito.tools import filter_ordered
@@ -38,9 +37,14 @@ def autotune(operator, args, level, mode):
     if key not in accepted:
         raise ValueError("The accepted `(level, mode)` combinations are `%s`; "
                          "provided `%s` instead" % (accepted, key))
-    if not any(isinstance(i, BlockDimension) for i in operator.dimensions):
+
+    # Tunable objects
+    blockable = [i for i in operator.dimensions if isinstance(i, BlockDimension)]
+    nthreads = [i for i in operator.input if isinstance(i, NThreads)]
+
+    if len(nthreads + blockable) == 0:
         # Nothing to tune for
-        return args
+        return args, {}
 
     # We get passed all the arguments, but the cfunction only requires a subset
     at_args = OrderedDict([(p.name, args[p.name]) for p in operator.parameters])
@@ -52,13 +56,8 @@ def autotune(operator, args, level, mode):
             if k in output:
                 at_args[k] = v.copy()
 
-    iterations = FindNodes(Iteration).visit(operator.body)
-
-    # Detect tunable arguments
-    tunable = {i.dim.step.name: i for i in iterations
-               if isinstance(i.dim, BlockDimension) and not i.is_Remainder}
-
     # Shrink the time dimension's iteration range for quick autotuning
+    iterations = FindNodes(Iteration).visit(operator.body)
     steppers = [i for i in iterations if i.dim.is_Time]
     if len(steppers) == 0:
         stepper = None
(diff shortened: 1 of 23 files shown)


mloubout:
That's pretty big :D Will look at it

devito/data/data.py
```diff
@@ -312,7 +312,7 @@ def index_dist_to_repl(idx, decomposition):
     Convert a distributed array index a replicated array index.
```
georgebisbas: Convert a distributed array index **to(?)** a replicated array index.

tests/test_autotuner.py
```diff
@@ -1,16 +1,10 @@
...
-except ImportError:
-    # Python3 compatibility
-    from io import StringIO
 
 import pytest
 import numpy as np
-from devito import Grid, Function, TimeFunction, Eq, Operator, configuration, silencio
-from devito.logger import logger, logging
+from devito import (Grid, Function, TimeFunction, Eq, Operator, configuration,
+                    silencio, ruido)
```
georgebisbas: silencio, ruido ?

tests/test_autotuner.py
```diff
@@ -98,12 +73,10 @@ def test_timesteps_per_at_run():
...
     op(infield=infield, outfield=outfield, autotune=True)
-    out = [i for i in buffer.getvalue().split('\n') if 'took' in i]
-    assert len(out) == 4
-    assert all('in 1 timesteps' in i for i in out)
-    buffer.truncate(0)
+    assert op._state['autotuning'][-1]['runs'] == 4
+    assert op._state['autotuning'][-1]['tpr'] == 1
 
     # TimeFunction with increasing time order; increasing the time order
     # shouldn't affect how many iterations the autotuner is gonna run
```
georgebisbas: *going to* instead of gonna should be more formal ;-)

tests/test_autotuner.py
```diff
@@ -114,42 +87,77 @@ def test_timesteps_per_at_run():
...
     grid = Grid(shape=(64, 64, 64))
     f = TimeFunction(name='f', grid=grid)
 
-    op = Operator(Eq(f.forward, f + 1))
-    op.apply(time=100, autotune=('basic', 'runtime'))
+    op = Operator(Eq(f.forward, f + 1.), dle=('advanced', {'openmp': False}))
+    summary = op.apply(time=100, autotune=('basic', 'runtime'))
 
-    # AT is expected to have executed 35 timesteps
+    # At is expected to have attempted 6 block shapes
```
georgebisbas: **AT** instead of At

tests/test_autotuner.py
```diff
@@ -114,42 +87,77 @@ def test_timesteps_per_at_run():
...
     """Test autotuning in non-destructive mode."""
     grid = Grid(shape=(64, 64, 64))
     f = TimeFunction(name='f', grid=grid)
 
-    op = Operator(Eq(f.backward, f + 1))
-    op.apply(time=101, autotune=('basic', 'runtime'))
+    op = Operator(Eq(f.backward, f + 1.), dle=('advanced', {'openmp': False}))
+    summary = op.apply(time=101, autotune=('basic', 'runtime'))
+
+    # At is expected to have attempted 6 block shapes
```
georgebisbas: AT instead of At

devito/core/autotuning.py
```diff
@@ -159,34 +129,38 @@ def autotune(operator, args, level, mode):
...
-            warning("AutoTuner: too few time iterations; giving up")
+            warning("too few time iterations; giving up")
             return False
 
     return stepper.extent(start=at_args[dim.min_name], finish=at_args[dim.max_name])
 
 
-def check_time_bounds(stepper, at_args, args):
-    if stepper is None:
+def check_time_bounds(stepper, at_args, args, mode):
```
tjb900: So when testing using runtime autotuning, I'm getting exceptions thrown from this function.

This is because the number of timesteps needed for the autotuning is larger than the number of timesteps I asked to run, which can easily happen in aggressive mode.  But since I have no control over or knowledge of the former, I would argue that this shouldn't cause an exception.

Perhaps the desired behaviour looks like one of the below:

 - "destructive" autotuning could rerun the same time range over and over, thus never going OOB.  I understand that this is different to 'runtime' , which (as I understand?) is kind of a 'tune while doing real work' mode, similar to what Chuck added to YASK.

or

 - it should terminate and choose the best result so far when it cannot try all the combinations

and/or

 - it should remember the results so far, and continue where it left off on the next call to apply() where autotuning is still enabled

devito/core/autotuning.py
```diff
@@ -222,38 +196,68 @@ def finalize_time_bounds(stepper, at_args, args, mode):
...
+            for i in range(ncombs):
+                for j in combinations(dict(bs), i+1):
+                    handle.append(tuple((b, v*2 if b in j else v) for b, v in bs))
+        ret.extend(handle)
+
+    # Drop unnecessary attempts:
+    # 1) Block shapes exceeding the iteration space extent
+    ret = [i for i in ret if all(dict(i)[k] <= v for k, v in max_bs)]
+    # 2) Redundant block shapes
+    ret = filter_ordered(ret)
```
tjb900: Can I make another request here?  Exclude block shapes where it would lead to having extremely few blocks to run (< the CPU count / 4 or something like that).  These almost-single-threaded autotuning steps are what seem to take the vast majority of the autotune time.

devito/core/autotuning.py
```diff
@@ -51,11 +56,8 @@ def autotune(operator, args, level, mode):
             if k in output:
                 at_args[k] = v.copy()
 
+    # Shrink the time dimension's iteration range for quick autotuning
```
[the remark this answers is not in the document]
FabioLuporini: yes, look at ``options['squeezer']`` at the bottom of the file

devito/core/autotuning.py
```diff
@@ -64,92 +66,60 @@ def autotune(operator, args, level, mode):
...
-    # ... Defaults (basic mode)
-    blocksizes = [OrderedDict([(i, v) for i in mapper]) for v in options['at_blocksize']]
-    # ... Always try the entire iteration space (degenerate block)
-    itershape = [mapper[i].iteration.symbolic_extent.subs(args) for i in mapper]
-    blocksizes.append(OrderedDict([(i, mapper[i].iteration.extent(0, j-1))
-                      for i, j in zip(mapper, itershape)]))
-    # ... More attempts if auto-tuning in aggressive mode
-    if level == 'aggressive':
-        blocksizes = more_heuristic_attempts(blocksizes)
+        warning("couldn't understand loop structure; giving up")
```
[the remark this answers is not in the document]
FabioLuporini: agreed. I'm changing this

devito/core/autotuning.py
FabioLuporini: > "destructive" autotuning could rerun the same time range over and over, thus never going OOB

that should be the current behaviour

>  I understand that this is different to 'runtime' , which (as I understand?) is kind of a 'tune while doing real work' mode, similar to what Chuck added to YASK.

correct

> it should terminate and choose the best result so far when it cannot try all the combinations

you're right, this is what I'm gonna do. Exception --> warning , and return the best block shape up until that point

devito/data/data.py
FabioLuporini: fixing, thanks

tests/test_autotuner.py
FabioLuporini: they are wrappers to locally change the logger and profile levels. I think it's just Spanish ? not sure why

tests/test_autotuner.py
FabioLuporini: code comments don't need to be formal.
API docstrings.... yeah

tests/test_autotuner.py
FabioLuporini: thx

tests/test_autotuner.py
FabioLuporini: thx

tests/test_autotuner.py
georgebisbas: We can replace them in English if possible....

devito/core/autotuning.py
```diff
@@ -222,38 +207,87 @@ def finalize_time_bounds(stepper, at_args, args, mode):
...
-        if i not in unique:
-            unique.append(i)
+def generate_nthreads(nthreads, args, level):
+    ret = [((i.name, args[i.name]),) for i in nthreads]
 
-    return unique
+    # On the KNL, also try running with a different number of hyperthreads
+    if level == 'aggressive' and configuration['platform'] == 'knl':
+        ret.extend([((i.name, psutil.cpu_count()),) for i in nthreads])
+        ret.extend([((i.name, psutil.cpu_count(logical=False)),) for i in nthreads])
```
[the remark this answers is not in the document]
FabioLuporini: oh, OK, just fixed this.

devito/core/autotuning.py
tjb900: Oh, I didn't even try preemptive because AFAICT it allocates copies of all the buffers?

FabioLuporini: I think I finally got what you mean, sorry dumb me. I'll produce a fix later/tomorrow.

FabioLuporini: done

PR #716, opened 2018-11-20
fix an issue with 0-sized arrays in NumaAllocator

tjb900:
This just works around a numpy error when `Data` objects have zero size.


FabioLuporini:
merged.

PR #717, opened 2018-11-20
replace isinstance(...,int) with is_integer(...)

tjb900:
This tripped me up a lot in the early days.  It's fairly easy to end up with a numpy integer instead of an int, and then unexpected things happen.

### DIFF
--- a/devito/ir/support/space.py
+++ b/devito/ir/support/space.py
@@ -6,7 +6,8 @@ from cached_property import cached_property
 
 from frozendict import frozendict
 
-from devito.tools import PartialOrderTuple, as_tuple, filter_ordered, toposort
+from devito.tools import PartialOrderTuple, as_tuple, filter_ordered, toposort, is_integer
+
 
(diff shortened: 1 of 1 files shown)


tjb900:
> strictly necessary
Probably not...  I can't actually remember the failing cases, but it made me crazy enough at the time to write the patch :)  I was also not sure, as I submitted this, if it might not be safer to to use `is_integer` for the class-membership tests (so that it behaves the way people expect) but still cast to `int` afterwards so we don't just move the problems elsewhere?

FabioLuporini:
some comments: 
* I can't imagine how can we have something which is not a pure int in function.py
* the fix in operator.py will be dropped as part of #715 , which entirely drops that horrible piece of code
* the fixes in space.py are reasonable . We might enforce int in there, yeah...

FabioLuporini:
(sorry, I accidentally closed the PR for a split second)

tjb900:
Roger, will cut it back to just `space.py` and then can go in after #715

FabioLuporini:
merged.

PR #722, opened 2018-11-26
OPS expressions

vmickus:
Changed location of find_offloadable_trees.
Created simple test for ops expressions.


devito/ir/iet/utils.py
```diff
@@ -137,3 +141,30 @@ def derive_parameters(nodes, drop_locals=False):
         parameters = [p for p in parameters if not isinstance(p, LocalObject)]
 
     return parameters
+
+
+def find_offloadable_trees(iet):
+    """
+    Return the trees within ``iet`` that can be computed by YASK.
```
FabioLuporini: this docstring needs to be updated

devito/ops/node_factory.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+
+    def __init__(self):
+        self.ops_access = {}
+
+    def new_symbol(self, name):
+        """
+            Creates a new sympy Symbol object with the given name.
+
+            :param name: Name of the symbol to be created.
```
FabioLuporini: We're switching everything to ``numpydoc``. You can google for it, or take a look at the [reference](https://numpydoc.readthedocs.io/en/latest/format.html) while some examples are available [here](https://github.com/opesci/devito/blob/master/devito/data/decomposition.py)

devito/ops/node_factory.py
```diff
@@ -0,0 +1,125 @@
+from sympy import Eq, Add, Mul, Pow, Integer, Float, Rational
+from devito.types import Symbol
+
+from devito.ops.utils import namespace
+
+
+class OPSNodeFactory():
+    """
+        Class responsible to generate ops expression for building the OPS ast.
```
FabioLuporini: OPS capital

devito/ops/node_factory.py
```diff
@@ -0,0 +1,125 @@
...
+
+        disp = getDimensionsDisplacement(dimensions[1:])
+
+        grid_id = '%s%s' % (name, dimensions[0])
+
+        # If the grid was alredy created, then use the same Access,
+        # but we should look for different displacements.
+        if grid_id in self.ops_access:
+            ops_acc = self.ops_access[grid_id]
+            symbol = Symbol(name='%s[%s(%s)]' %
```
FabioLuporini: No, putting the square brackets directly here is wrong. You should use an Indexed. Can you see why?

devito/ops/operator.py
```diff
@@ -1,12 +1,63 @@
...
     """
-    A special :class:`OperatorCore` to JIT-compile and run operators through OPS.
+        A special :class:`OperatorCore` to JIT-compile and run operators through OPS.
     """
 
+    _default_includes = OperatorRunnable._default_includes + \
+        ['ops_seq.h', 'ops_lib_cpp.h']
+
     def _specialize_iet(self, iet, **kwargs):
-        raise NotImplementedError
```
FabioLuporini: for this PR, this can stay ``NotImplementedError``. As I said, this should be a relatively small, self-contained PR

devito/ops/transformer.py
```diff
@@ -0,0 +1,109 @@
...
+
+from devito.ops.node_factory import OPSNodeFactory
+from devito.ops.utils import namespace
+
+
+def opsit(trees):
+    """
+    Populate the tree with OPS instructions.
+
+    :param trees: A sequence of offloadable :class: `IterationTree`s, in which the
```
FabioLuporini: numpydoc

devito/ops/transformer.py
```diff
@@ -0,0 +1,109 @@
...
+
+from devito.ir.iet import Callable
+from devito.ir.iet.nodes import Expression, ClusterizedEq
+from devito.ir.iet.visitors import FindNodes
+
+from devito.ops.node_factory import OPSNodeFactory
+from devito.ops.utils import namespace
+
+
+def opsit(trees):
```
FabioLuporini: we don't need this function in this PR

devito/ops/transformer.py
```diff
@@ -0,0 +1,109 @@
...
+                                             for i in expr.args])
+    elif expr.is_Indexed:
+        dimensions = [make_ops_ast(i, nfops, mapper, arguments)
+                      for i in expr.indices]
+
+        grid_access, grid_name = nfops.new_grid(expr.name, dimensions)
+        arguments.append(Array(name=grid_name, dimensions=[], dtype=expr.dtype))
+        return grid_access
+    else:
+        print(expr)
```
FabioLuporini: leftover

devito/ops/transformer.py
```diff
@@ -0,0 +1,109 @@
...
+        raise NotImplementedError("Missing handler in Devito-OPS translation")
+
+
+def create_new_ops_kernel(count, expr, arguments):
```
FabioLuporini: unneeded in this PR

tests/test_ops.py
```diff
@@ -1,2 +1,59 @@
...
+import pytest
+
+from devito import Grid, TimeFunction, Operator, Eq, configuration
+from devito.ir.iet import FindNodes, Expression, printAST
+
+pytestmark = pytest.mark.skipif(configuration['backend'] != 'ops',
+                                reason="'ops' wasn't selected as backend on startup")
+
+
+class TestOpsSyntax(object):
```
FabioLuporini: OPS

FabioLuporini: no need in this PR

tests/test_ops.py
```diff
@@ -1,2 +1,59 @@
...
+        # Checks ops_exit declaration
+        assert 'ops_exit();' in str(op)
+
+
+class TestOpsExpression(object):
+    """
+        Tests OPS generated expressions.
+        Some basic tests for the generated expression with ops.
+    """
+    @pytest.mark.parametrize('expr, expected', [
```
FabioLuporini: as I wrote on Slack, I'd like to see many more test cases. You can/should also compare to stringyfied sympy expressions, not to IET objects

devito/ops/node_factory.py
vmickus: I should have asked about this. Its not a problem, i can configure my editor to use numpydoc as docstring.

devito/ir/iet/utils.py
vmickus: Sure, my bad.

vmickus: """ 
    Return the trees within ``iet`` that can be considered offloadable.

    A tree is "offloadable" if it is embedded in a time stepping loop
    *and* all of the grids accessed by the enclosed equations are homogeneous
    (i.e., same dimensions and data type).
    
    Parameters
    ----------
    iet : list of :class:`Node`
        Original tree in which will be searched for the offloadable tree.
    
    Returns
    -------
    list of :class:`Node`
        A piece of the tree that contains a time stepping loop and all the 
        grid accessed are homogeneous.
    """

This is what i'll change it for.

FabioLuporini: nitpicking:

I'm also dropping all "Return" ad the beginning of a docstring

How about something complete fresh, such as:
```
Find offloadable trees. A tree is offloadable when all of the array accesses are constant/affine functions of the Iteration variables and the Iteration bounds are fixed (but possibly symbolic).

Parameters
----------------
iet : :class:`Node`
    The searched tree

Returns
-----------
list of :class:`Node`
    Each item in the list is the root of an offloadable tree
```

vmickus: Agree, we have a `Returns` title so we could drop it in the method description.

That sounds simpler, i like it.

devito/ops/node_factory.py
vmickus: Actually i didn't understand.

```
            symbol = Symbol(name='%s[%s(%s)]' %
```
This will generate something like: `u[OPS_ACC0(0,0)]`. As the square brackets will always be in this place, i don't see your point. Could you explain me?

FabioLuporini: yes! So, if you generate strings, you're dropping information on the floor, so any later compiler passes (perhaps none at this stage, but only at this stage!) won't be able to identify arrays/indexeds in an expression anymore

vmickus: Ahhhhh. Got it now! 

let me try this `Indexed` .

devito/ops/node_factory.py
vmickus: `OpsNodeFactory`, i'll fix this.

devito/ir/iet/utils.py
```diff
@@ -137,3 +141,38 @@ def derive_parameters(nodes, drop_locals=False):
...
+        Each item in the list is the root of an offloadable tree
+    """
+    offloadable = OrderedDict()
+    roots = [i for i in FindNodes(Iteration).visit(iet) if i.dim.is_Time]
+    for root in roots:
+        sections = FindNodes(Section).visit(root)
+        for section in sections:
+            for tree in retrieve_iteration_tree(section):
+                if not all(i.is_Affine for i in tree):
+                    # Non-affine array accesses unsupported by YASK
```
FabioLuporini: 'YASK' leftover

devito/ops/node_factory.py
```diff
@@ -0,0 +1,214 @@
+from sympy import Eq, Add, Mul, Pow, Integer, Float, Rational
+from devito import Dimension
+from devito.symbolics import Macro
+from devito.types import Symbol, Indexed, Array
+
+from devito.ops.utils import namespace
+
+
+class OpsNodeFactory():
```
FabioLuporini: OPS

devito/ops/node_factory.py
```diff
@@ -0,0 +1,214 @@
...
+    """
+    Generates ops nodes for building an OPS expression.
+
+    Examples
+    --------
+    >>> a = OpsNodeFactory()
+    >>> b = a.new_symbol('symbol_name')
```
FabioLuporini: instead of assigning it to ``b``, you should add another line displaying the returned value, as typically done in the Examples section.

devito/ops/node_factory.py
```diff
@@ -0,0 +1,214 @@
...
+
+        Parameters
+        ----------
+        name : str
+            Name of the symbol to be created.
+
+        Returns
+        -------
+        :class:`Symbol` 
+            Object :class:`Symbol` with the given name.
```
FabioLuporini: No need for "Object" -- neither here, nor in the docstrings below

devito/ops/node_factory.py
```diff
@@ -0,0 +1,214 @@
...
+    >>> a = OpsNodeFactory()
+    >>> b = a.new_symbol('symbol_name')
+    """
+
+    def __init__(self):
+        self.ops_grids = {}
+
+    def new_symbol(self, name):
+        """
+        Creates a new sympy :class:`Symbol` object with the given name.
```
FabioLuporini: nitpicking: SymPy, not sympy

FabioLuporini: here and below

devito/ops/node_factory.py
```diff
@@ -0,0 +1,214 @@
...
+            The division denominator.
+
+        Returns
+        -------
+        :class:`Mul`
+            Object of :class:`Mul` representing the division.
+        """
+        return Mul(num, Pow(den, Integer(-1)))
+
+    def new_grid(self, devito_indexed, dimensions):
```
FabioLuporini: do not call it "grid". "grid" was YASK terminology, no need to use the same here. ``new_indexed`` is probably fine.

and instead of ``devito_indexed``, just ``indexed``

devito/ops/node_factory.py
```diff
@@ -0,0 +1,214 @@
...
+        """
+        Creates an :class:`Indexed` object using OPS representation.
```
FabioLuporini: no need for "object"

devito/ops/node_factory.py
```diff
@@ -0,0 +1,214 @@
...
+        If the pair grid name and time dimension was alredy created, it will return
```
FabioLuporini: "If the pair grid name and time dimension ..."

this smells of copy-paste. Is this actually true here?

devito/ops/node_factory.py
```diff
@@ -0,0 +1,214 @@
...
+            Indexed object using OPS representation.
+        """
+
+        def getDimensionsDisplacement(dimensions):
+            disp = []
+
+            for dim in dimensions:
+                if dim.is_Add:
+                    lhs, = dim.as_two_terms()
+                    disp.append(str(lhs))
```
FabioLuporini: I'm not sure this works. What if the Add has multiple addends, such as ``a + b + 2 + c + d[4] + ...`` ?

devito/ops/node_factory.py
```diff
@@ -0,0 +1,214 @@
...
+                else:
+                    disp.append('0')
```
FabioLuporini: 0, not '0'

devito/ops/node_factory.py
```diff
@@ -0,0 +1,214 @@
...
+
+        # Gets the dimension displacement.
+        disp = getDimensionsDisplacement(dimensions[1:])
+
+        # Builds the grid identifier.
+        grid_id = '%s%s' % (devito_indexed.name, dimensions[0])
+
+        if grid_id not in self.ops_grids:
+            # Creates the indexed object.
+            # Always will have one dimension that will be generated using the macro
```
FabioLuporini: this description isn't immediately clear

devito/ops/node_factory.py
```diff
@@ -0,0 +1,214 @@
...
+            # OPS_ACC but it will not represent the real dimension used, because we
+            # can have a 3 dimension problem and only one Dimension object.
+            grid = Array(name=grid_id,
+                         dimensions=[Dimension(name=namespace['ops_acc'])],
+                         dtype=devito_indexed.dtype)
+            
+            self.ops_grids[grid_id] = grid
+        else:
+            grid = self.ops_grids[grid_id]
```
FabioLuporini: reminder: no need to call these "grids". These objects are Functions in core devito, right? even though here you cast them into array for code generation purposes

devito/ops/node_factory.py
```diff
@@ -0,0 +1,214 @@
...
+
+        # Defines the Macro used in this grid indice.
+        access_macro = Macro('OPS_ACC%d(%s)' % (len(self.ops_grids), ','.join(disp)))
```
FabioLuporini: I don't like this much, I think we should extend Macro to accept parameters rather than flattening everything out into a string

devito/ops/node_factory.py
```diff
@@ -0,0 +1,214 @@
...
+        # Creates Indexed object representing the grid access.
+        indexed = Indexed(grid.indexed, access_macro)
+
+        return indexed
+
+    def new_equation_node(self, *args):
+        """
+            Creates a new sympy equation with the provided arguments.
+
+            :param *args: arguments to construct the new equation.
```
FabioLuporini: numpydoc

devito/ops/operator.py
```diff
@@ -1,12 +1,63 @@
...
 class Operator(OperatorRunnable):
     """
-    A special :class:`OperatorCore` to JIT-compile and run operators through OPS.
+        A special :class:`OperatorCore` to JIT-compile and run operators through OPS.
     """
 
+    _default_includes = OperatorRunnable._default_includes + \
+        ['ops_seq.h', 'ops_lib_cpp.h']
+
     def _specialize_iet(self, iet, **kwargs):
```
FabioLuporini: reminder: no need to implement specialize_iet in this PR.

devito/ops/operator.py
```diff
@@ -1,12 +1,63 @@
...
 __all__ = ['Operator']
 
 
 class Operator(OperatorRunnable):
     """
-    A special :class:`OperatorCore` to JIT-compile and run operators through OPS.
+        A special :class:`OperatorCore` to JIT-compile and run operators through OPS.
     """
 
+    _default_includes = OperatorRunnable._default_includes + \
```
FabioLuporini: no need for this in this PR

devito/ops/operator.py
```diff
@@ -1,12 +1,63 @@
...
+from devito.ops.utils import namespace
+from devito.ops.transformer import opsit
 
 __all__ = ['Operator']
 
 
 class Operator(OperatorRunnable):
     """
-    A special :class:`OperatorCore` to JIT-compile and run operators through OPS.
+        A special :class:`OperatorCore` to JIT-compile and run operators through OPS.
```
FabioLuporini: u mean OperatorRunnable?

devito/ops/node_factory.py
```diff
@@ -0,0 +1,214 @@
...
+    >>> b = a.new_symbol('symbol_name')
+    """
+
+    def __init__(self):
+        self.ops_grids = {}
+
+    def new_symbol(self, name):
+        """
+        Creates a new sympy :class:`Symbol` object with the given name.
+        Its used to define/use variables.
```
georgebisbas: It is(?) [define/use --> Is used necessary?],

devito/ops/node_factory.py
```diff
@@ -0,0 +1,214 @@
...
+        return Integer(number)
+
+    def new_float_node(self, number):
+        """
+        Creates a new sympy :class:`float` object.
+
+        Parameters
+        ----------
+        number : float
+            Number to initialize an Float node.
```
georgebisbas: **a** Float node

devito/ops/node_factory.py
```diff
@@ -0,0 +1,214 @@
...
+        Returns
+        -------
+        :class:`Integer`
+            Object of :class:`Integer`.
+        """
+        return Integer(number)
+
+    def new_float_node(self, number):
+        """
+        Creates a new sympy :class:`float` object.
```
georgebisbas: SymPy as @FabioLuporini noted earlier as well

devito/ops/node_factory.py
```diff
@@ -0,0 +1,214 @@
...
+        -------
+        :class:`Float`
+            Object of :class:`Float`.
+        """
+        return Float(number)
+
+    def new_rational_node(self, num, den):
+        """
+        Creates a new sympy :class:`Rational` object to represent 
+        a mathematical rational.
```
georgebisbas: I am not sure if the expression _mathematical rational_  can stand alone... Mathematical rational what?
Probably I am missing something but I do not quite get it.

devito/ops/transformer.py
```diff
@@ -0,0 +1,107 @@
...
+    mapper = {}
+    processed = []
+    for tree in trees:
+        # All expressions whithin `tree`
+        expressions = [i.expr for i in FindNodes(Expression).visit(tree.inner)]
+
+        # Attach conditional expression for sub-domains
+        conditions = [(i, []) for i in expressions]
+
+        # Only one node factory for all expression so we can keep track
```
georgebisbas: # Only one node factory**,** for all expression**s**

devito/ops/transformer.py
```diff
@@ -0,0 +1,107 @@
...
+    else:
+        print(expr)
+        raise NotImplementedError("Missing handler in Devito-OPS translation")
+
+
+def create_new_ops_kernel(count, expr, arguments):
+    """
+        Creates a Callable object responsable for defining the ops kernel method.
+
+        :param count: number of ops kernel being created.
```
georgebisbas: kernel**s** ?

devito/ops/node_factory.py
vmickus: I misunderstood your previous comment. 
When you said: `OPS capital` you were referring to my comments, not the class name. But i thought it was the class name and changed it.

devito/ops/node_factory.py
vmickus: You are right. As the method will always create new object there is no point in the "use".

devito/ops/node_factory.py
vmickus: So,in this case,  i leave only `:class:`Symbol` with the given name`

devito/ops/node_factory.py
vmickus: I see your point. what about: "mathematical rational operation" or just "rational operation" ?

georgebisbas: Whatever, its up to you.

devito/ops/node_factory.py
vmickus: new_indexed i want to use in the end, when i create the new Indexed object.
For the Array obj i think grid is a good name, because it does represent a grid.
Will change the devito_indexed to indexed.

devito/ir/iet/utils.py
FabioLuporini: you haven't pushed, as I still see "YASK"

btrb:
T_T

`collecting ... /home/travis/.travis/job_stages: line 104:  7120 Segmentation fault      (core dumped) py.test --cov devito tests/`

devito/ops/operator.py
vmickus: If i do this, my tests will fail and wont be possible to run using DEVITO_BACKEND=ops.

I am not sure how can i remake the tests so i can test only the expression, i mean, i dont know how can i build the IterableTree manually to be used in the make_ops_ast.

vmickus: I do like this

devito/ops/transformer.py
vmickus: I need to this so i can have a minimal test.

tests/test_ops.py
vmickus: Ok

devito/ops/transformer.py
FabioLuporini: no, for the expression conversion test we don't even need an Operator. In fact, think about that: you should be able to simplify the test by not allocating any Operator to instead simply call the conversion routing

btrb:
I don't think so. I'd guess that it looks more like YASK-related..

devito/ops/node_factory.py
vmickus: I do agree that this need to be changed, but i think this should be in another PR.

devito/ops/node_factory.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def new_grid(self, indexed):
+        """
+        Creates an :class:`Indexed` node using OPS representation.
+        If the pair grid name and time dimension was alredy created, it will return
+        the stored value associated with this pair.
+
+        Notes
+        -----
+        Its expected the first item in the list `indexed.indices` to be the time
```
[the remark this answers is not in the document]
vmickus: I didn't know about this, let me try it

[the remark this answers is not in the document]
vmickus: This worked for me `indexed.indices[TimeFunction._time_position]`

devito/ops/transformer.py
```diff
@@ -0,0 +1,80 @@
...
+                                      namespace['ops-kernel-retval'],
+                                      []))
+
+    return new_functions
+
+
+def make_ops_ast(expr, nfops):
+    """
+    Transform an devito expression into an OPS expression.
+    Only the interested nodes are changed.
```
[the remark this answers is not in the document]
georgebisbas: I just saw that.

tests/test_ops.py
```diff
@@ -1,2 +1,60 @@
-""" Empty file for now
-"""
+import pytest
```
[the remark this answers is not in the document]
FabioLuporini: true

devito/ops/utils.py
```diff
@@ -0,0 +1,28 @@
+from collections import OrderedDict
+
+
+# OPS Conventions
+namespace = OrderedDict()
+
+namespace['call-ops_init'] = 'ops_init'
+
+namespace['name-ops_grid'] = 'grid'
```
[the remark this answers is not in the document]
vmickus: True

devito/ops/node_factory.py
```diff
@@ -0,0 +1,74 @@
...
+        indexed : :class:`Indexed`
+            Indexed object using devito representation.
+
+        Returns
+        -------
+        :class:`Indexed`
+            Indexed node using OPS representation.
+        """
+
+        def getDimensionsDisplacement(dimension):
```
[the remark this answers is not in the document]
vmickus: That seems to be what i need.

devito/ops/node_factory.py
```diff
@@ -0,0 +1,74 @@
...
+            # Creates the indexed object.
+            grid = Array(name=grid_id,
+                         dimensions=[Dimension(name=namespace['ops_acc'])],
+                         dtype=indexed.dtype)
+
+            self.ops_grids[grid_id] = grid
+        else:
+            grid = self.ops_grids[grid_id]
+
+        space_dim = [e for i, e in enumerate(
```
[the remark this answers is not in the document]
vmickus: Hmmm this `indexed.function.space_dimensions` is almost what i need, the problem with it is that it only return the dimension itself and don't return the the displacement.

For example:
`x - 1`, it would retrieve `x` only, and when i use the `split_affine` with `indexed.function.space_dimensions` it won't find the shift.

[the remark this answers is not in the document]
FabioLuporini: I was confused by the variable name, then. Can you rather call it `space_indices`

devito/ops/node_factory.py
vmickus: I got your point, and you are right.
I'll use `ops_arg`, because it could the data associated with:  `ops_arg_dat`, `ops_arg_idx`, `ops_arg_gbl` ...

btrb: maybe we should ask istvan's opinion on that?
i'd agree with dats although there seems to be a slight difference between `ops_dats` shat those args actually are

btrb: i just saw your commit, `ops_arg` looks good indeed. 
yet wouldn't hurt to have instavan's feedback on that imo

devito/ops/node_factory.py
```diff
@@ -0,0 +1,63 @@
...
+    symbol_name
+    """
+
+    def __init__(self):
+        self.ops_args = {}
+
+    def new_ops_arg(self, indexed):
+        """
+        Create an :class:`Indexed` node using OPS representation.
+        If the pair ops argument name and time dimension were alredy created,
```
FabioLuporini: "If the pair ``(argname, time_dimension)`` ...
feels like this needs some better wording. It's not entirely clear why you do so, for example (well, it is clear to me, but probably not to the newcomer...)

devito/ops/node_factory.py
```diff
@@ -0,0 +1,63 @@
...
+        indexed : :class:`Indexed`
+            Indexed object using devito representation.
+
+        Returns
+        -------
+        :class:`Indexed`
+            Indexed node using OPS representation.
+        """
+
+        # Build the ops argument identifier.
```
FabioLuporini: (nitpicking) convention used throughout devito: never put a full stop at the end of comments

tests/test_ops.py
```diff
@@ -1,2 +1,61 @@
...
+        ('Eq(v,1)', 'Eq(vt[OPS_ACC0(0,0)], 1)'),
+        ('Eq(v,v.dxl + v.dxr - v.dyr - v.dyl)',
+         'Eq(vt[OPS_ACC0(0,0)], vt[OPS_ACC0(0,-1)]/h_y - vt[OPS_ACC0(0,1)]/h_y'
+            ' - vt[OPS_ACC0(-1,0)]/h_x + vt[OPS_ACC0(1,0)]/h_x)'),
+        ('Eq(v,v**2 - 3*v)',
+         'Eq(vt[OPS_ACC0(0,0)], vt[OPS_ACC0(0,0)]**2 - 3*vt[OPS_ACC0(0,0)])'),
+        ('Eq(v,a*v + b)',
+         'Eq(vt[OPS_ACC0(0,0)], 1.43*vt[OPS_ACC0(0,0)] + 9.87e-7)'),
+        ('Eq(v.dt,1/v + v**-1)',
+         'Eq(-vt[OPS_ACC1(0,0)]/dt + vt + 1[OPS_ACC0(0,0)]/dt, 2/vt[OPS_ACC1(0,0)])'),
```
FabioLuporini: this looks wrong to me: ``1[OPS_ACC0 ... `` ??

tests/test_ops.py
```diff
@@ -1,2 +1,61 @@
...
+        ('Eq(w,c*w**2)',
+         'Eq(wt[OPS_ACC0(0,0,0)], 999999999999999*wt[OPS_ACC0(0,0,0)]**2)'),
+        ('Eq(w,u + v )', 'Eq(wt[OPS_ACC0(0,0,0)], ut[OPS_ACC1(0)] + vt[OPS_ACC2(0,0)])'),
```
FabioLuporini: can you add some further tests with ``u.forward`` and ``v.forward``, so that we get to see ``ut0, ut1, ...`` ?

tests/test_ops.py
FabioLuporini: also ``vt`` without ``[]``...

tests/test_ops.py
vmickus: I tried to added it, but u.forward, u.laplace, u.dx# ( wheren # > 1 ) doesn't seems to get translated with what i have. It seems there is something in Operator that I am not using that replace this u.forward for sympy symbols.

vmickus: My point is that this tests should be done when using `Operator`, but i can replicate the part of Operator that do this.

tests/test_ops.py
vmickus: The problem was that it was creating an `indexed` with the name `vt + 1` and then this happened. Alredy fixed.

devito/ops/node_factory.py
vmickus: Actually i got this idea of name from Istvan code on `OPS`. I went to the OPS source code, and looked how they were calling this variables. Of course we can also confirm with him personally, just to be sure.

FabioLuporini:
closing in favor to #760

PR #729, opened 2018-12-03
Symbolic padding

FabioLuporini:
Padding is a per-Function thing, which may vary:
* across different applications of the same Operator (due to using Functions with different shape and thus requiring different padding so that the innermost size is a multiple of the vector length);
* across different MPI ranks in the same MPI session.

In many circumstances, we must generate symbols, rather than numbers, to represent padding values.
In particular, when using YASK or running with MPI we need:
* symbolic array casts

Further, when running with MPI:
* The MPI halo exchange routines must take into account the presence of padding (through symbolic values)

This PR implements this behavior.


tests/test_data.py
```diff
@@ -162,41 +161,61 @@ def test_logic_indexing(self):
...
+    def test_data_regions_metadata(self):
         """
-        Tests access to domain and halo data.
+        Test correctness of metadata describing extent and offset of the various
+        data regions, such as DOMAIN, HALO, etc.
         """
         grid = Grid(shape=(4, 4, 4))
 
-        # Without padding
+        # Without halo and withouth padding
```
georgebisbas: without

tjb900:
Oh, one more thing...  In the generated C code, there are still a lot of hard-coded references to the halo size, but interestingly not the padding.

e.g. `u[time_s1][xi+8][yi+8][zi+8] = ...`

Is this a problem, or does this mean you have rolled the shifts due to the padding into the address of the array passed down into C?

FabioLuporini:
>    symbolic padding is not used when in non-MPI, non-YASK mode.  I'm not sure why you wouldn't want to do it all the time?

yes, you’re correct, in non-MPI, non-YASK mode, the padding is still explicit. This is because of three reasons, though perhaps not very compelling:
* if I had changed it, I’d have had to update lots of tests, while basically nobody is using this feature, AFAICT
* the generated code gets a bit more ugly
* for educational purposes (eg, our notebooks), seeing padding and halo explicitly is quite helpful.

ironically, when I made this decision, I was *not* thinking that you can run the same operator with different functions having different padding :sunglasses:  though I even probably stated it in the PR description

> If the symbolic padding is not always on, then it's probably a good idea to detect that the user has provided Functions to apply with different padding to that specified at the time of Operator creation, and error. Currently it misbehaves badly in this case, and this will be especially true if the new Function has less padding - that could easily lead to OOB/segfault.

you're right, I'll do something for this

> In MPI mode, the top-level Operator function uses the size from the descriptor, but the nested callables (e.g. f_0) are still passed x_size, y_size, etc, and with the padding hard-baked into the code of the function f_0. Interestingly, scatter_u and gather_u seem fine

OK this is weird and I'm not sure I follow, I'll try to write a test to reproduce this

> The selfdoc does not make clear that padding can be a tuple of 2-tuples, to specify left and right padding. Note: if the main purpose of padding is for memory alignment, I think that the default should be one-sided (i.e. only pad on the right). This tripped me up - it was effectively doubling all the padding values I gave it until I figured this out.

I should clear this out. Again, you're being hit by the fact that nobody has used this feature in a long while. In fact, what we should do is to start introducing padding automatically under the hood. I could enforce one-sided padding in `core`. `yask` may still benefit from two-sided padding probably because of the fancy memory layout. However, YASK controls padding, not the user, so ...

> Oh, one more thing... In the generated C code, there are still a lot of hard-coded references to the halo size, but interestingly not the padding.
> e.g. u[time_s1][xi+8][yi+8][zi+8] = ...
> Is this a problem, or does this mean you have rolled the shifts due to the padding into the address of the array passed down into C?

assuming padding is right-sided, this is a non-issue. In non-MPI, non-YASK mode, any left-padding is added to the halo, so you only see one number. About

> does this mean you have rolled the shifts due to the padding into the address of the array passed down into C?

You can't easily do that with multi-dimensional arrays, I suppose...

FabioLuporini:
@tjb900  I'm gonna enforce **right-padding for the innermost dimension** only in ``core``. Note that with classic row-major storage layout, padding the inner dimensions is actually irrelevant.

tjb900:
> @tjb900 I'm gonna enforce right-padding for the innermost dimension only in core. Note that with classic row-major storage layout, padding the inner dimensions is actually irrelevant.

I'm not actually sure this is true; I'd like to experiment before this feature goes away...  would it be possible to request a delay on that?

FabioLuporini:
> I'm not actually sure this is true; I'd like to experiment before this feature goes away... would it be possible to request a delay on that?

OK

FabioLuporini:
One-sided padding -- is that fine with you?

tjb900:
> One-sided padding -- is that fine with you?
Yeah, one sided is good.  I just think there _might_ be value in controlling it not just for the innermost dimension.

devito/function.py
```diff
@@ -623,6 +658,91 @@ def symbolic_shape(self):
...
+        # MPI-related fields
+        dataobj._obj.npsize = (c_int*self.ndim)(*[i - sum(j) for i, j in
+                                                  zip(data.shape, self._extent_padding)])
+        dataobj._obj.hsize = (c_int*(self.ndim*2))(*flatten(self._extent_halo))
+        dataobj._obj.hofs = (c_int*(self.ndim*2))(*flatten(self._offset_halo))
+        dataobj._obj.oofs = (c_int*(self.ndim*2))(*flatten(self._offset_owned))
+        return dataobj
+
+    def _C_as_ndarray(self, dataobj):
+        """Cast the data carried by a TensorFunction dataobj to an ndarray."""
```
georgebisbas: a ndarray

devito/operator.py
```diff
@@ -148,6 +146,16 @@ def _prepare_arguments(self, **kwargs):
...
             p._arg_check(args, self._dspace[p])
 
+        # Turn arguments into a format suitable for the generated code
+        # E.g., instead of NumPy arrays for Functions, the generated code expects
+        # pointers to ctypes.Struct
+        for p in self.input:
+            try:
+                args.update(kwargs.get(p.name, p)._arg_as_ctype(args, alias=p))
+            except AttributeError:
+                # User-provided flots/ndarray obviously do not have `_arg_as_ctype`
```
georgebisbas: floats?

examples/compiler/03_data_regions.ipynb
```diff
@@ -77,6 +77,9 @@
     "from devito import configuration\n",
     "configuration['openmp'] = 0\n",
     "\n",
+    "# For illustation purposes, we ask Devito to generate explicit array casts.\n",
```
georgebisbas: illustration

devito/function.py
FabioLuporini: no, when it comes to acronyms, the 'n' goes with the sound of the following letter, rather than depending on whether it's a vowel or a consonant . 

a PDE
a DSL
an FEM 
an NDarray

etc

examples/compiler/03_data_regions.ipynb
FabioLuporini: fixed

devito/operator.py
FabioLuporini: fixed

PR #734, opened 2018-12-06
order fix for small order staggered diff

mloubout:
Small fix, was ignoring order=1 for staggered.


FabioLuporini:
we could merge it straight away, but it's a symptom we are lacking tests....

PR #737, opened 2018-12-11
Symbolic coefficients

rhodrin:
Submitting for a running discussion while developing.

Next steps:
1) Add symbolic weights to derivatives other than the 'generic_derivative'.
2) Begin creation of the 'coefficient' class and add grid wide replacement of coefficients at the equation level.


devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,89 @@
...
+        # with user provided coefficients and, if possible, generate
+        # replacement rules
+        for d in Coefficients.data:
+            if isinstance(d, tuple):
+                Coefficients.rules = None
+            else:
+                Coefficients.rules = None
+
+
+    #def _apply_fd_coefficients(self, expressions, args):
```
FabioLuporini: tomorrow should I ignore this?

devito/finite_differences/finite_difference.py
```diff
@@ -277,6 +324,13 @@ def staggered_diff(expr, deriv_order, dim, fd_order, stagger=centered):
     :param stagger: (Optional) shift for the FD, `left`, `right` or `centered`.
     """
 
+    # Check if we need symbolic fd weights (not implemented)
```
FabioLuporini: I've seen this code already (above I think). Perhaps we need a decorator?

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,89 @@
...
+class Coefficients(object):
+    """
+    Devito class for users to define custom finite difference weights.
+    """
+
+    def __init__(self, *args, **kwargs):
+
+        # Need to check args are valid here.    
+
+        Coefficients.data = args
```
FabioLuporini: I don't like this, I have no idea what's going on

devito/finite_differences/finite_difference.py
```diff
@@ -392,3 +447,16 @@ def generate_fd_shortcuts(function):
             derivatives[name_fd] = (deriv, desciption)
 
     return derivatives
+
+def symbolic_weights(function, deriv_order, indices, dim):
+
+    n_weights = len(indices)
+
+    # This 'function' should be centralised
```
FabioLuporini: ?

devito/finite_differences/coefficients.py
rhodrin: Yeah ignore.

devito/finite_differences/finite_difference.py
rhodrin: Yeah, this repetition can and should be avoided.

devito/finite_differences/coefficients.py
rhodrin: Yeah, it's not finished so there's no reason you should.

devito/finite_differences/finite_difference.py
rhodrin: The weight function needs to be defined somewhere more appropriate and so that all routines that need it have access to that particular function and don't need to 're-create' it. But since I'm not sure where it's best to put it yet I've left it there (with a reminder for me to change) it for now.

rhodrin:
Only things worth commenting on at the moment are:
1) Changes to function.py.
2) Changes to finite_difference.py

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,89 @@
...
+
+        # Figure out when symbolic coefficients can be replaced
+        # with user provided coefficients and, if possible, generate
+        # replacement rules
+        for d in Coefficients.data:
```
FabioLuporini: I suppose this loop is WIP?

devito/finite_differences/finite_difference.py
```diff
@@ -81,6 +82,15 @@ def second_derivative(expr, **kwargs):
        results in ``(-2.0*f(x, y)*g(x, y) + 1.0*f(-h + x, y)*g(-h + x, y) +
        1.0*f(h + x, y)*g(h + x, y)) / h**2``.
     """
+    # Check if we need symbolic fd weights
+    functions = expr.as_ordered_factors()
```
FabioLuporini: I suppose here you want the actual {Time}Function objects, while ``as_ordered_factors`` gives you...the factors, which can actually be generic expressions. And I assume you don't care about repetitions, ie if you had ``f(x+h_x, y), f(x, y), f(x, y+h_y)`` retrieving ``f(x, y)`` (xor ``f(x+h_x, y)`` xor ...)) would suffice.

If that's the case, you can:
```
from devito.symbolics import retrieve_functions
functions = retrieve_functions(expr)  # get all {Time,Sparse,...}Function objects

from devito.tools import filter_ordered
functions = filter_ordered(functions, key=lambda i: i.name)
```

Even better: ``retrieve_functions`` takes in input the optional parameter mode. Currently, ``mode`` can be either 'unique' or 'all'. Don't get confused by the word 'unique' here: it simply means that if in your expression ``f(x, y)`` appears twice, then you only fetch it once. But ``f(x, y)`` is still different than ``f(x+h-x, y)``. IOW, here we need uniqueness based on the Function name only, not on name+indices. So, the better way to go is add a new mode.

devito/finite_differences/finite_difference.py
```diff
@@ -89,7 +99,10 @@ def second_derivative(expr, **kwargs):
     ind = [(dim + i * diff) for i in range(-int(order / 2),
                                            int(order / 2) + 1)]
 
-    coeffs = finite_diff_weights(2, ind, dim)[-1][-1]
+    if fd_weight_type is 'symbolic':
+        coeffs = symbolic_weights(f, 2, indices, dim)
```
FabioLuporini: what is ``f`` here? it seems undefined, unless you're referring to the last ``f`` in the loop above, which would clearly be flawed

devito/finite_differences/finite_difference.py
```diff
@@ -198,8 +229,10 @@ def first_derivative(expr, **kwargs):
         ind = [(dim + i * diff) for i in range(-int(order / 2),
                                                int((order + 1) / 2) + 1)]
     # Finite difference weights from Taylor approximation with this positions
-    c = finite_diff_weights(1, ind, dim)
-    c = c[-1][-1]
+    if fd_weight_type is 'symbolic':
+        c = symbolic_weights(f, 1, indices, dim)
```
FabioLuporini: and now that I think about it: I have a mild preference towards having ``finite_diff_weights`` accepting a further parameter "symbolic", rather than having two different functions ``symbolic_weights`` and ``fintie_diff_weights``. Because that way we could avoid a proliferation of if-then-else in all functions computing derivatives

devito/function.py
```diff
@@ -950,6 +950,11 @@ def __init__(self, *args, **kwargs):
             else:
                 raise TypeError("`space_order` must be int or 3-tuple of ints")
 
+            # Symbolic (finite difference) coefficients
+            self._symbolic_coefficients = kwargs.get('coefficients', 'standard')
```
FabioLuporini: nitpicking: I'd call it just "coefficients", as ``_symbolic_coefficients`` is a bit too long for an instance attribute

devito/finite_differences/finite_difference.py
FabioLuporini: oh, I think I see.

Shouldn't it be ``Coefficient('W')`` anyway?

Random thought, maybe BS: perhaps a Function should create a "stamp" Coefficient so that when u do ``function._coefficients`` you get it (instead of a string) ?

Also: should we use ``'W_%s % function.name`` as coefficient name, for uniqueness ?

devito/finite_differences/coefficients.py
rhodrin: Yep, will work on it this morning.

devito/finite_differences/finite_difference.py
rhodrin: Yes, will think a little more, but I believe we just wan't to pull unique functions. So will look into adding a new mode.

devito/finite_differences/finite_difference.py
rhodrin: Sounds good. Since `finite_diff_weights` (as it is currently) is a sympy function we'll need our own wrapper with this addition.

devito/function.py
rhodrin: Yes, will change this - it's not actually a good name considering the arguments it takes. Originally, when I named it in another branch it made more sense (prior to some changes).

rhodrin:
>oh, I think I see.
>
>Shouldn't it be Coefficient('W') anyway?
>
>Random thought, maybe BS: perhaps a Function should create a "stamp" Coefficient so that when u >do function._coefficients you get it (instead of a string) ?
>
>Also: should we use 'W_%s % function.name as coefficient name, for uniqueness ?

I believe the current naming method I'm using allows for the uniqueness we require. We could do it that way but considering the available permutations I personally think the current method is a bit cleaner.

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,66 @@
...
+        
+    def check_args(self, args, kwargs):
+        for arg in args:
+            if isinstance(arg, tuple):
+                assert isinstance(arg[0], int)
+                assert(arg[1].is_Function)
+                assert(arg[2].is_Dimension)
+                assert isinstance(arg[3], np.ndarray)
+            else:
+                raise NotImplementedError
```
mloubout: wouldn't it be nicer and "more readable" to have `__init__(order, function, dimensions, coeffs, **kwargs)` rather than obscure `args` with a bunch of tests with it?

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def rules(self):
+        
+        # FIXME: Move this?
+        def generate_subs(d):
+            
+            deriv_order = d[0]
+            function = d[1]
+            dim = d[2]
+            coeffs = d[-1]
```
mloubout: same as above, I (personally) think it's more readable with explicit function args rather that d[0], d[1]...

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,66 @@
...
+            subs = {}
+            
+            if fd_order == 1:
+                indices = [dim, dim + dim.spacing]
+            else:
+                indices = [(dim + i * dim.spacing) for i in range(-fd_order//2, fd_order//2 + 1)]
+            
+            for j in range(len(coeffs)):
+                subs.update({function.symbol()(indices[j], deriv_order, function, dim): coeffs[j]})
+
```
mloubout: I think this all section should not be here as this is now done in two places, here and in ``finite_differences` which is risky in my opinion. Indices should be obtained from finite_differences in my opinion.

devito/finite_differences/finite_difference.py
```diff
@@ -2,10 +2,11 @@
 
 from functools import partial
 
-from sympy import S, finite_diff_weights
+from sympy import S, finite_diff_weights, Function
```
mloubout: unused?

devito/function.py
```diff
@@ -1028,6 +1033,14 @@ def __padding_setup__(self, **kwargs):
     @property
     def space_order(self):
         return self._space_order
+    
+    @memoized_meth
+    def symbol(self):
```
mloubout: confusing name

devito/function.py
```diff
@@ -950,6 +950,11 @@ def __init__(self, *args, **kwargs):
             else:
                 raise TypeError("`space_order` must be int or 3-tuple of ints")
 
+            # Symbolic (finite difference) coefficients
+            self._coefficients = kwargs.get('coefficients', 'standard')
```
mloubout: Isn't this a bit "cyclic". Need to declare function with symbolic coefficient and need function to declare coefficients?

devito/function.py
```diff
@@ -950,6 +950,11 @@ def __init__(self, *args, **kwargs):
             else:
                 raise TypeError("`space_order` must be int or 3-tuple of ints")
```
btrb: shouldn't it allow me to do something like:

`g = Grid((10,10))
f = TimeFunction(name='f',grid=g, space_order=(2,2))`

?

rhodrin:
>Some comments.
>
>WOuld maybe turn the examples into notebooks to see better what's happening.

Yes, that's the plan. When things are nearly ready the examples will be converted to notebooks explaining everything.

devito/finite_differences/coefficients.py
rhodrin: Depending on how symbolic coefficients are going to be used, the input could be in various formats i.e. grid wide (which will be introduced in this PR) vs. modified at specific nodes vs. grid wide function (future PR's). 

So I'll probably need to iterate this a few times but however it develops we need to keeps it flexible in order that future functionality can be incorporated without too much messing around.

devito/finite_differences/coefficients.py
rhodrin: Yes, agreed - I should have added a FIXME note here as this is something I intended to change.

devito/finite_differences/finite_difference.py
rhodrin: Yes. Was used previously but no longer is.

devito/function.py
rhodrin: Will change.

devito/function.py
rhodrin: The declaration of symbolic in the function allows for the 'potential' use of symbolic coefficients. It maybe that the user wants to use custom coefficients for a certain derivative in a particular dimension or just in a certain equation (using standard coefficients in another equation). Therefore, without making further changes to the manner in which the fd coefficients are implemented, we need something of this form to allow this flexibility. 

Note that if the coefficients function is not declared (but the functions coefficients have been declared symbolic) the symbols will be replaced with the default weights.

devito/function.py
rhodrin: That section of code has not been modified in this branch.
Does the `TimeFunction` declaration you wrote not work in Master?

devito/finite_differences/coefficients.py
mloubout: I understand that, but I still think the `__init__` signature should be explicit enough for people to know what it is like we did for most of the Devito codebase (Function,...).

devito/function.py
mloubout: @btrb no it doesn't. The tuple is` (order, halo_left, halo_right)` as said in the docstring, it's not multiple orders for different  dimensions

FabioLuporini:
@rhodrin before I review this, make sure to have followed the [coding guidelines](https://github.com/opesci/devito/blob/master/CONTRIBUTING.md)
they are not comprehensive of course

devito/finite_differences/coefficients.py
FabioLuporini: yes, I agree, I don't quite like this

random thought, potentially completely out of context: perhaps this is a symptom you need subclasses?

rhodrin:
Notebook needs some updating and the classes in devito/finite_difference/coefficients.py need to be re-named but apart from that it's ready to look at.

rhodrin:
I suggest that the notebook (01_DRP_schemes.ipynb) is moved to 08_DRP_schemes.ipynb  in the tutorials folder?

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,191 @@
...
+        return rules
+
+
+class Coefficient(object):
+    """
+    Prepare custom FD coefficients to pass to Coefficients object.
+
+    Parameters
+    ----------
+    deriv_order : represents the order of the derivative being taken.
```
FabioLuporini: actually, the type should go on the RHS of the colon, while below the description (indented)

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,191 @@
...
+                raise ValueError("Number FD weights provided does not "
+                                 "match the functions space_order")
+        elif dimension.is_Space:
+            if len(coefficients)-1 != function.space_order:
+                raise ValueError("Number FD weights provided does not "
+                                 "match the functions space_order")
+
+        self.is_Coefficient = True
+
+        self.deriv_order = deriv_order
```
FabioLuporini: private attributes (starting with _) + properties is the way to go

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,191 @@
...
+        self.function = function
+        self.dimension = dimension
+        self.coefficients = coefficients
+
+    def check_input(self, deriv_order, function, dimension, coefficients):
+        if not isinstance(deriv_order, int):
+            raise TypeError("Derivative order must be an integer")
+        if not function.is_Function:
```
FabioLuporini: probably better off to use isinstance, otherwise you might get an AttribtueError

devito/tools/utils.py
```diff
@@ -154,6 +154,17 @@ def filter_sorted(elements, key=None):
     return sorted(filter_ordered(elements, key=key), key=key)
 
 
+def filter_sparse(elements):
```
FabioLuporini: no, there should actually be nothing Devito-related in ``tools``. So you can't use things such as ``is_DiscreteFunction``

perhaps you don't need a function for this -- with a list comprehension it should become a one liner

devito/types/dense.py
```diff
@@ -888,6 +889,11 @@ def __init__(self, *args, **kwargs):
             else:
                 raise TypeError("`space_order` must be int or 3-tuple of ints")
 
+            # Symbolic (finite difference) coefficients
+            self.coefficients = kwargs.get('coefficients', 'standard')
```
FabioLuporini: we could use the class Tag for these (``git grep Tag`` to see what it does)

devito/types/grid.py
```diff
@@ -182,6 +182,19 @@ def subdomains(self):
         """The SubDomains defined in this Grid."""
         return {i.name: i for i in self._subdomains}
 
+    def add_subdomains(self, subdomain_dict):
```
FabioLuporini: ``Grid`` should be immutable, so no add-like methods. One should create a Grid and pass the pre-created SubDomains through the constructor

devito/types/grid.py
```diff
@@ -381,7 +394,15 @@ def dimensions(self):
 
     @property
     def dimension_map(self):
-        return {d.parent: d for d in self.dimensions}
+        dim_map = {}
```
FabioLuporini: just switch ``d.parent`` to ``d.root`` and the one linear should work again

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,191 @@
...
+    function : represents the function for which the supplied coefficients
+               will be used.
+    dimension : represents the dimension with respect to which the
+                derivative is being taken.
+    coefficients : represents the set of finite difference coefficients
+                  intended to be used in place of the standard
+                  coefficients (obtained from a Taylor expansion).
+    """
+
+    def __init__(self, deriv_order, function, dimension, coefficients):
```
FabioLuporini: perhaps use the same parameters ordering used in the various FD functions, to make it more consistent?

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,191 @@
...
+        if dimension.is_Time:
+            if len(coefficients)-1 != function.time_order:
+                raise ValueError("Number FD weights provided does not "
+                                 "match the functions space_order")
+        elif dimension.is_Space:
+            if len(coefficients)-1 != function.space_order:
+                raise ValueError("Number FD weights provided does not "
+                                 "match the functions space_order")
+
+        self.is_Coefficient = True
```
FabioLuporini: ? 
as there's no hierarchy, I guess this is quite useless?

use isinstance instead

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,191 @@
...
+
+        self.deriv_order = deriv_order
+        self.function = function
+        self.dimension = dimension
+        self.coefficients = coefficients
+
+    def check_input(self, deriv_order, function, dimension, coefficients):
```
FabioLuporini: no need to be a separate method?

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,191 @@
...
+        if not isinstance(coefficients, np.ndarray):
+            raise NotImplementedError
+        return
+
+
+def default_rules(obj, functions):
+
+    def generate_subs(d):
+
+        deriv_order = d[0]
```
FabioLuporini: I prefer an explicit parameter list

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,191 @@
...
+
+    # Determine which 'rules' are missing
+    sym = get_sym(functions)
+    terms = obj.find(sym)
+    # FIXME: Unnecessary conversions between lists and sets
+    args_present = []
+    for term in terms:
+        args = term.args
+        args_present += [args[1:], ]
+    args_present = list(set(args_present))
```
FabioLuporini: list comprehension probably doable

perhaps something along the lines of
``args_present = filter_ordered(term.args[1:] for term in terms)``

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,191 @@
...
+
+    coeffs = obj.coefficients
+    args_provided = []
+    if coeffs:
+        for coeff in coeffs.data:
+            args_provided += [(coeff.deriv_order, coeff.function, coeff.dimension), ]
```
FabioLuporini: list comprehension again

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,191 @@
...
+
+        coeffs = sympy.finite_diff_weights(deriv_order, indices, x0)[-1][-1]
+
+        for j in range(len(coeffs)):
+            subs.update({function.fd_coeff_symbol()
+                         (indices[j], deriv_order, function, dim): coeffs[j]})
+
+        return subs
+
+    rules = {}
```
FabioLuporini: this is declarated a bit too prematurely , as it's populated further down in the function body

devito/finite_differences/finite_difference.py
```diff
@@ -62,8 +65,25 @@ def wrapper(expr, *args, **kwargs):
...
+def check_symbolic(func):
+    @wraps(func)
+    def wrapper(expr, *args, **kwargs):
+        functions = retrieve_functions(expr)
+        functions = filter_ordered(functions, key=lambda i: i.name)
+        functions = filter_sparse(functions)
+        symbolic_coefficients = any(f.coefficients is 'symbolic' for f in functions)
+        if symbolic_coefficients:
+            expr_dict = expr.as_coefficients_dict()
+            if any(len(expr_dict.keys()) > 1 for item in expr_dict):
```
FabioLuporini: no need for keys()

devito/finite_differences/finite_difference.py
```diff
@@ -62,8 +65,25 @@ def wrapper(expr, *args, **kwargs):
...
+                raise NotImplementedError
```
FabioLuporini: add an error message?

devito/finite_differences/finite_difference.py
```diff
@@ -62,8 +65,25 @@ def wrapper(expr, *args, **kwargs):
...
+        kwargs['symbolic_coefficients'] = symbolic_coefficients
+        return func(expr, *args, **kwargs)
+    return wrapper
+
+
 @check_input
-def first_derivative(expr, dim, fd_order=None, side=centered, matvec=direct):
+@check_symbolic
+def first_derivative(expr, dim, fd_order=None, side=centered, matvec=direct, **kwargs):
```
FabioLuporini: probably better to have an explicit ``symbolic=False`` kwarg, instead of ``**kwargs``, and make sure that ``check_symbolic`` always adds in the information (be it True or False)

tests/test_symbolic_coefficients.py
```diff
@@ -0,0 +1,47 @@
...
+        grid = Grid(shape=(20, 20))
+        u0 = TimeFunction(name='u', grid=grid, time_order=order, space_order=order)
+        u1 = TimeFunction(name='u', grid=grid, time_order=order, space_order=order,
+                          coefficients='symbolic')
+        eq0 = Eq(-u0.dx+u0.dt)
+        eq1 = Eq(u1.dt-u1.dx)
+        assert(eq0.evalf(_PRECISION).__repr__() == eq1.evalf(_PRECISION).__repr__())
+
+    @pytest.mark.parametrize('expected', [
+        ('0.1*u(x) - 0.6*u(x - h_x) + 0.6*u(x + h_x)'),
```
FabioLuporini: I assume you mean to add more tests, since I see this in a ``parametrize``?

devito/types/dense.py
```diff
@@ -1014,6 +1028,28 @@ def avg(self, p=None, dims=None):
     _pickle_kwargs = DiscreteFunction._pickle_kwargs +\
         ['space_order', 'shape_global', 'dimensions']
 
+    def get_stagger(self, dim, order):
```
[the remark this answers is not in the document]
rhodrin: The method gets information about the grid staggering in order to generate the correct indices for the finite difference scheme.

devito/types/dense.py
```diff
@@ -1014,6 +1028,28 @@ def avg(self, p=None, dims=None):
...
+        name = 't' if dim.is_Time else dim.name
+        name_fd = 'd%s%d' % (name, order) if order > 1 else 'd%s' % name
+        item = self._fd.get(name_fd, None)
+        if bool(item):
+            stagger = item[0].keywords.get('stagger', None)
+        else:
+            raise ValueError("Requested derivative unavailable")
+        return stagger
+
+    def get_side(self, dim, order):
```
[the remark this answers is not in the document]
rhodrin: Same as above.

devito/finite_differences/coefficients.py
rhodrin: Changed.

devito/finite_differences/coefficients.py
rhodrin: Yeah, I see. But `isintance(function, Function)` won't work by default and I'm not sure how it should be imported to get it to work?

rhodrin: Using try/except for now(?).

devito/tools/utils.py
rhodrin: Fixed now. In addition to putting the function the the incorrect place I overlooked something when I was initially thinking about this filter.

(The initial 'no' in your first sentence is redundant).

devito/types/grid.py
rhodrin: Great, ta.

devito/types/grid.py
rhodrin: The issue was that `class AbstractModel` creates a grid with some required underlying subdomains which the user isn't directly exposed to. You can then pass a grid to `class Model` but this breaks things as it won't have the correct subdomains. I'll therefore have to these classes a bit.

devito/types/dense.py
```diff
@@ -968,6 +974,14 @@ def space_order(self):
         """The space order."""
         return self._space_order
 
+    @memoized_meth
```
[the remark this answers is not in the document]
rhodrin: Ta, fixed.

devito/finite_differences/coefficients.py
rhodrin: Indeed. There was previously a conditional return statement just below this declaration but when that was no longer needed and removed I forgot to move this declaration.

devito/finite_differences/finite_difference.py
rhodrin: Ta, removed.

devito/finite_differences/finite_difference.py
rhodrin: Added.

devito/finite_differences/coefficients.py
rhodrin: Currently not particularly. But as the sparse function stuff etc is added in the future it may end up being a bit neater this way.

devito/finite_differences/coefficients.py
rhodrin: Probably. I was thinking it may be needed/useful depending on future functionality evolves, but will use `isinstance` for the time being.

devito/finite_differences/coefficients.py
rhodrin: I'm not sure it's needed here since this is a local function? The unpacking would just be moved to the for loop which would make things a bit untidier and not really add any clarity.

devito/finite_differences/coefficients.py
rhodrin: Ta, updated.

devito/finite_differences/coefficients.py
rhodrin: Updated.

devito/types/dense.py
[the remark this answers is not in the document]
rhodrin: I should also note that the method is located here because `self._fd` is a private attribute.

devito/finite_differences/finite_difference.py
rhodrin: Ta, changed.

examples/seismic/tutorials/08_DRP_schemes.ipynb
```diff
@@ -0,0 +1,653 @@
...
+   ],
+   "source": [
+    "print(u.dx)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "By default the 'standard' Taylor series expansion result, where `h_x` represents the $x$-direction grid spacing, is returned. However, there may be instances when a user wishes to use 'non-standard' weights when, for example, implimenting a dispersion-relation-preserving (DRP) scheme. See e.g. \n",
```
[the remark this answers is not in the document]
rhodrin: Thank's, typo fixed.

You mean:
"By default the 'standard' Taylor series expansion result, where h_x represents the 𝑥-direction grid spacing, is returned."
Is so, just wondering which part of it?

examples/seismic/tutorials/08_DRP_schemes.ipynb
```diff
@@ -0,0 +1,653 @@
...
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Custom finite difference coefficients in Devito\n",
+    "\n",
+    "## Introduction\n",
+    "\n",
+    "When taking the numerical derivative of a function in Devito, the default behaviour is for 'standard' finite difference weights (obtained via a Taylor series expansion about the point of differntation) to be applied. Consider the following example for some field $u(\\mathbf{x},t)$, where $\\mathbf{x}=(x,y)$. Let us define a computational domain/grid and differentiate our field with respect to $x$."
```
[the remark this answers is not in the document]
rhodrin: Ta, fixed.

examples/seismic/tutorials/08_DRP_schemes.ipynb
```diff
@@ -0,0 +1,653 @@
...
+    "print(u.dx)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "By default the 'standard' Taylor series expansion result, where `h_x` represents the $x$-direction grid spacing, is returned. However, there may be instances when a user wishes to use 'non-standard' weights when, for example, implimenting a dispersion-relation-preserving (DRP) scheme. See e.g. \n",
+    "\n",
+    "[1] Christopher K.W. Tam, Jay C. Webb (1993). ”Dispersion-Relation-Preserving Finite Difference Schemes for Computational Acoustics.” **J. Comput. Phys.**, 107(2), 262--281. https://doi.org/10.1006/jcph.1993.1142\n",
```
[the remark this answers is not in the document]
rhodrin: I think this would be better but was just 'copying' what had been done in other notebooks for consistency. It would be nice to be able to use latex style referencing and bibliographies but such support for jupyter notebooks seems to require an additional package. Anyway, I'm open to suggestions.

examples/seismic/tutorials/08_DRP_schemes.ipynb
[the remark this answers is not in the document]
btrb: what do you think about:

"By default, the return is a 'standard' Taylor series expansion, where h_x represents the 𝑥-direction grid spacing."

devito/equation.py
```diff
@@ -50,15 +57,33 @@ class Eq(sympy.Eq):
     def __new__(cls, *args, **kwargs):
         kwargs['evaluate'] = False
         subdomain = kwargs.pop('subdomain', None)
+        coefficients = kwargs.pop('coefficients', None)
         obj = sympy.Eq.__new__(cls, *args, **kwargs)
         obj._subdomain = subdomain
+        obj._coefficients = coefficients
+        functions = retrieve_functions(obj)
+        functions = filter_ordered(functions, key=lambda i: i.name)
```
FabioLuporini: I think you can drop this line. By default, ``retrieve_functions`` returns unique items, though you can change it, see [here](https://github.com/opesci/devito/blob/master/devito/symbolics/search.py#L129)

devito/equation.py
```diff
@@ -50,15 +57,33 @@ class Eq(sympy.Eq):
...
+        functions = [f for f in functions if not f.is_SparseFunction]
+        if any(f.coefficients == 'symbolic' for f in functions):
+            # NOTE: As Coefficients.py is expanded we will not want
+            # all rules to be expunged during this procress.
+            rules = default_rules(obj, functions)
+            try:
+                obj = obj.xreplace({**coefficients.rules, **rules})
```
FabioLuporini: what could possibly go wrong here? the ``except`` below needs to be specialised for code self-documentation . 
Also, I'm not sure I follow the logic underpinning this try-except: is it the idea that ``rules`` will eventually have to be applied?

devito/equation.py
```diff
@@ -24,6 +28,9 @@ class Eq(sympy.Eq):
     subdomain : SubDomain, optional
         To restrict the computation of the Eq to a particular sub-region in the
         computational domain.
+    coefficients : Coefficients, optional
```
FabioLuporini: reminder, not necessarily for this PR, but quite important: we need to add examples to the docstring, showing how to use SubDomain as well as Coefficient

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,256 @@
...
+
+from devito.finite_differences import generate_indices
+from devito.tools import filter_ordered
+
+__all__ = ['Coefficient', 'CoefficientRules', 'default_rules']
+
+
+class Coefficient(object):
+    """
+    Prepare custom FD coefficients to pass to CoefficientRules object.
```
FabioLuporini: for generality, I'd drop FD? wdy think?

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,256 @@
...
+
+    Parameters
+    ----------
+    deriv_order : integer
```
FabioLuporini: for homogeneity, just "int"

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,256 @@
...
+        Represents the order of the derivative being taken.
```
FabioLuporini: nitpicking: for homogeneity, you can drop all "Represents" (same in the lines below)

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,256 @@
...
+    def dimension(self):
+        """The dimension to which the coefficients will be applied."""
+        return self._dimension
+
+    @property
+    def coefficients(self):
+        """The set of coefficients."""
+        return self._coefficients
+
+    def check_input(self, deriv_order, function, dimension, coefficients):
```
FabioLuporini: this should be a private method

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,256 @@
...
+            raise NotImplementedError
+        return
+
+
+class CoefficientRules(object):
+    """
+    Devito class to convert Coefficient objects into replacent rules
+    to be applied when constructing a Devito Eq.
+
+    Example
```
FabioLuporini: Examples

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,256 @@
...
+    -------
+    Assume we have created the 'u_x_coeffs' Coefficient object
```
FabioLuporini: OK, this is quite important. I'd explicitly recreate ``u_x_coeffs`` ...

... and I'd add this file to ``test_docstrings.py``, [here](https://github.com/opesci/devito/blob/master/tests/test_docstrings.py#L27) , so that CI automatically tests the correctness of all these example code snippets and we're sure they are never out of date. 

Note: I was suggesting to recreate ``u_x_coeffs`` explicitly otherwise the test will fail complaining it couldn't find ``u_x_coeffs``

That's why in another comment I was suggesting to also show the output of some of these variables , although perhaps as you said in some cases it's not that meaningful

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,256 @@
...
+    Now create a Devito equation and pass to it 'coeffs'
+
+    >>> from devito import Eq
+    >>> eq = Eq(u.dt+u.dx, coefficients=coeffs)
+    Eq(0.1*u(t, x, y) - 0.6*u(t, x - h_x, y) + 0.6*u(t, x + h_x, y)
+    - u(t - dt, x, y)/(2*dt) + u(t + dt, x, y)/(2*dt), 0)
+
+    Notes
+    -----
+    If a function is declared with 'symbolic' coefficients and no
```
FabioLuporini: Function

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,256 @@
...
+        self._rules = self.rules
+
+    @property
+    def data(self):
+        """The Coefficient objects passed."""
+        return self._data
+
+    @cached_property
+    def function_list(self):
+        return list(set([d.function for d in self.data]))
```
FabioLuporini: how about ``filter_ordered(i.function for i in self.data, lambda i: i.name)`` so that we also ensure deterministic ordering across different runs?

also, nitpicking: ``d`` is almost always used throughout the codebase to iterate over ``Dimensions``. For ``Functions`` we typically use ``f`` or the more generic ``i``

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,256 @@
...
+            raise TypeError("Non Coefficient object within input")
+
+        self._data = args
+        self._function_list = self.function_list
+        self._rules = self.rules
+
+    @property
+    def data(self):
+        """The Coefficient objects passed."""
+        return self._data
```
FabioLuporini: for code self-documentation: why not "coefficients", instead of ``data``?

devito/finite_differences/coefficients.py
FabioLuporini: the point is, this kind of unpacking doesn't help with code self-documentation. Here, ``d`` could contain 100 parameters for example.

Another example: in the ``rules`` method above, ``d`` actually contains 4 parameters? as in, it includes ``coeffs`` as well

I'm not sure I got this:
> The unpacking would just be moved to the for loop
 
do you mean the caller? what line?

devito/finite_differences/finite_difference.py
```diff
@@ -62,8 +65,28 @@ def wrapper(expr, *args, **kwargs):
     return wrapper
 
 
+def check_symbolic(func):
+    @wraps(func)
+    def wrapper(expr, *args, **kwargs):
+        functions = retrieve_functions(expr)
+        functions = filter_ordered(functions, key=lambda i: i.name)
```
FabioLuporini: no need for this line, probably

devito/finite_differences/finite_difference.py
```diff
@@ -258,34 +279,19 @@ def generic_derivative(expr, dim, fd_order, deriv_order, stagger=None):
...
-    c = finite_diff_weights(deriv_order, indices, x0)[-1][-1]
+        c = finite_diff_weights(deriv_order, indices, x0)[-1][-1]
 
     deriv = 0
     all_dims = tuple(set((dim, ) +
                      tuple([i for i in expr.indices if i.root == dim])))
     for i in range(0, len(indices)):
-            subs = dict([(d, indices[i].subs({dim: d})) for d in all_dims])
-            deriv += expr.subs(subs) * c[i]
+        subs = dict([(d, indices[i].subs({dim: d})) for d in all_dims])
```
FabioLuporini: probably no need to wrap within a [...] (it might save some time to pass a ``generator`` directly to ``dict``)

devito/finite_differences/finite_difference.py
```diff
@@ -366,3 +372,51 @@ def generate_fd_shortcuts(function):
...
+
+
+def symbolic_weights(function, deriv_order, indices, dim):
+
+    n_weights = len(indices)
+
+    weights = []
+    for j in range(n_weights):
+        weights.append(function.fd_coeff_symbol(indices[j], deriv_order, function, dim))
+    return weights
```
FabioLuporini: one-line list comprehension ?

devito/finite_differences/finite_difference.py
```diff
@@ -366,3 +372,51 @@ def generate_fd_shortcuts(function):
...
+                                               int((order+1)/2)+2-(order % 2))]
+        elif side == left:
+            ind = [(dim-i*diff) for i in range(-int(order/2)+1-(order % 2),
+                                               int((order+1)/2)+2-(order % 2))]
+        else:
+            ind = [(dim+i*diff) for i in range(-int(order/2),
+                                               int((order+1)/2)+1)]
+        x0 = None
+    else:
+        if stagger == left or not func.is_Staggered:
```
FabioLuporini: should we move these lines down within ``if func.is_Staggered`` ?

devito/types/dense.py
```diff
@@ -968,6 +974,14 @@ def space_order(self):
...
+    def fd_coeff_symbol(self):
```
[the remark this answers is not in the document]
FabioLuporini: I'd drop the ``fd`` for the sake of generality

the method should be private

in fact, this could be a class level private static attribute

devito/types/dense.py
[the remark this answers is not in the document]
FabioLuporini: These ``get_stagger`` and ``get_side`` are still around but I'm still not convinced we actually need them. As soon as I have an hour, I'll try to understand if we can get rid of them, or to convince myself they're actually necessary

examples/seismic/model.py
```diff
@@ -430,11 +430,12 @@ def __init__(self, origin, spacing, shape, space_order, nbpml=20,
         # at the correct index
         origin_pml = tuple([dtype(o - s*nbpml) for o, s in zip(origin, spacing)])
         phydomain = PhysicalDomain(self.nbpml)
+        subdomains = subdomains + (phydomain, )
         shape_pml = np.array(shape) + 2 * self.nbpml
         # Physical extent is calculated per cell, so shape - 1
         extent = tuple(np.array(spacing) * (shape_pml - 1))
         self.grid = Grid(extent=extent, shape=shape_pml, origin=origin_pml, dtype=dtype,
-                         subdomains=phydomain)
+                         subdomains=subdomains)
```
FabioLuporini: nice !

devito/equation.py
rhodrin: Ta, dropped.

devito/equation.py
rhodrin: If `coefficients` is `None` (i.e. the user has not supplied a `CoefficentRules` object to this particular equation) then `coefficient.rules` will result in an attribute error. 

> Also, I'm not sure I follow the logic underpinning this try-except: is it the idea that rules will 
> eventually have to be applied?

In essence yes, it's saying try and replace coefficients with supplied rules + any default rules necessary, and if no rules have been supplied replace everything with default rules.

Apologies, what do you mean by 'specialised'? I'll look through other existing try/except instances to try and work that out shortly anyway.

devito/equation.py
rhodrin: I can add an example (within the next couple of days) in a different PR?

devito/finite_differences/coefficients.py
rhodrin: Sure. There are various 'coefficients' being named elsewhere in the code base (i.e. Sparse functions have a `_coefficients` property) so not sure how much we want to distinguish to avoid confusion? What does @mloubout think?

devito/finite_differences/coefficients.py
rhodrin: Fixed.

devito/finite_differences/coefficients.py
rhodrin: Changed.

devito/finite_differences/coefficients.py
rhodrin: Updated + `coefficients.py` added to `test_docstrings.py`.

Quick note: printing u_x_coeffs returns things such as `<devito.finite_differences.coefficients.Coefficient object at 0x7f0fa3f4da20>` which is why the output is not currently being printed (and no output will be printed by default owing to the '=').

devito/finite_differences/coefficients.py
rhodrin: Fixed.

devito/finite_differences/coefficients.py
rhodrin: Sounds good. Updated.

devito/finite_differences/coefficients.py
rhodrin: It was for naming reasons to avoid things such as `coeffs.coefficients`. I've changed it to `coefficients` for now and will have a re-look later to see if the various names need some re-organising.

devito/finite_differences/finite_difference.py
rhodrin: Removed.

devito/finite_differences/finite_difference.py
rhodrin: Redundant brackets removed.

devito/finite_differences/finite_difference.py
rhodrin: Now using list comprehension.

devito/finite_differences/finite_difference.py
rhodrin: Yeah, I see no reason not too. And we can also do away with the `or not func.is_staggered` part.

rhodrin:
Re. the remaining presence of `get_stagger` and `get_side`: I was looking through @mloubout's 'Lazy' branch and it looked like these would change when that branch is added. So wasn't sure if we wanted to spend time 'fixing' them now only to have to change things when that branch is applied?

examples/seismic/model.py
rhodrin: > I'd drop the fd for the sake of generality

Dropped.

> the method should be private

I think you were linking to and older version of the code, this got changed to a `cached_property`. The method needs to be called from `coefficients.py` and `finite_difference.py` but if `coefficients='symbolic'` has not been declared it doesn't need to be created at all. I guess you're saying that accessing the public `cached_property` (or whatever is becomes) should internally call a private method? But does that make a difference?

devito/equation.py
rhodrin: I see what you mean, except updated.

devito/types/dense.py
```diff
@@ -981,6 +987,19 @@ def space_order(self):
         """The space order."""
         return self._space_order
 
+    @property
+    def coefficients(self):
+        """Form of the FD coefficients of the function."""
+        return self._coefficients
+
+    @cached_property
+    def coeff_symbol(self):
```
FabioLuporini: private method

devito/types/dense.py
```diff
@@ -981,6 +987,19 @@ def space_order(self):
         """The space order."""
         return self._space_order
 
+    @property
+    def coefficients(self):
+        """Form of the FD coefficients of the function."""
```
FabioLuporini: drop FD ?

devito/equation.py
```diff
@@ -50,15 +56,32 @@ class Eq(sympy.Eq):
     def __new__(cls, *args, **kwargs):
         kwargs['evaluate'] = False
         subdomain = kwargs.pop('subdomain', None)
+        substitutions = kwargs.pop('coefficients', None)
         obj = sympy.Eq.__new__(cls, *args, **kwargs)
         obj._subdomain = subdomain
+        obj._substitutions = substitutions
+        functions = retrieve_functions(obj)
```
FabioLuporini: If ``substitutions`` is None, then probably we don't need to spend time retrieving Functions. 
So we could:
```
if substituions is not None:
    functions = ...
    ...
```

This could save some time when big expressions are supplied (eg, when using high order derivatives)

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,263 @@
...
+    if not_provided:
+        for i in not_provided:
+            deriv_order = i[0]
+            function = i[1]
+            dim = i[2]
+            rules = {**rules, **generate_subs(deriv_order, function, dim)}
+    return rules
+
+
+def get_sym(functions):
```
FabioLuporini: this routine makes me thing that ``coeff_symbol`` should actually be something unique and global defined somewhere here in ``coefficients.py``, rather than as a per-Function method in ``dense.py``

tests/test_symbolic_coefficients.py
```diff
@@ -0,0 +1,54 @@
...
+from devito.finite_differences import Differentiable
+
+_PRECISION = 9
+
+pytestmark = skipif(['yask', 'ops'])
+
+
+class TestSC(object):
+    """
+    Class for testing symbolic FD coefficients
```
FabioLuporini: nitpicking: drop "FD"
nitpicking: no need to break over two lines

devito/types/dense.py
rhodrin: This property has now been removed and replaced with a function in `coefficients.py`.

devito/types/dense.py
rhodrin: Sure.

devito/equation.py
rhodrin: Ta, changed.

devito/finite_differences/coefficients.py
rhodrin: I like that. This symbol is now retrieved via a function in `coefficients.py`.

tests/test_symbolic_coefficients.py
rhodrin: Changed.

devito/types/dense.py
rhodrin: Actually, after thinking about future functionality, I think it may be better to keep this symbol as a property of the function it is attached to. The construction of the symbol has now been made private.

devito/equation.py
rhodrin: Actually, since a function could be declared with symbolic coefficients but for a specific equation no replacement rules supplied I think this needs to stay.

FabioLuporini: OK. Hopefully this should still return control the user in negligible time

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,263 @@
...
+    args_provided = list(set(args_provided))
+    not_provided = list(set(args_provided).symmetric_difference(set(args_present)))
+
+    rules = {}
+    if not_provided:
+        for i in not_provided:
+            deriv_order = i[0]
+            function = i[1]
+            dim = i[2]
+            rules = {**rules, **generate_subs(deriv_order, function, dim)}
```
FabioLuporini: does it work if you just write

```rules = {**rules, **generate_subs(*i)}```
?

File: devito/finite_differences/coefficients.py [hunk 65]
```diff
@@ -0,0 +1,263 @@
...
+    else:
+        args_provided = []
+
+    # NOTE: Do we want to throw a warning if the same arg has
+    # been provided twice?
+    args_provided = list(set(args_provided))
+    not_provided = list(set(args_provided).symmetric_difference(set(args_present)))
+
+    rules = {}
+    if not_provided:
```
FabioLuporini: this ``if`` should be unnecessary

devito/types/dense.py
```diff
@@ -981,6 +987,24 @@ def space_order(self):
...
+    def _coeff_symbol(self):
+        if self.coefficients == 'symbolic':
+            return sympy.Function('W')
+        else:
+            raise ValueError("Function was not declared with symbolic "
+                             "coefficients.")
+
+    @property
+    def coeff_symbol(self):
+        """The symbol representing the functions coefficients."""
```
FabioLuporini: the Function coefficients ?

but more importantly, why should we expose it publicly ? isn't ``_coeff_symbol`` enough?

devito/finite_differences/finite_difference.py
```diff
@@ -366,3 +363,56 @@ def generate_fd_shortcuts(function):
...
+
+        else:
+            ind = [(dim + i*dim.spacing) for i in range(-order//2, order//2 + 1)]
+            x0 = dim
+            if order < 2:
+                ind = [dim, dim + diff]
+    return ind, x0
+
+
+def form_side(dimensions, function):
```
FabioLuporini: @mloubout @rhodrin 

future work: we can definitely store left/right/centered [here](https://github.com/opesci/devito/blob/master/devito/types/dense.py#L144) (and lines below) instead of 0/1/-1 ... and then just use `.val` each time `.staggered` is used (in a very few places, for example [here](https://github.com/opesci/devito/blob/master/devito/types/dense.py#L205) .

I'm fine to leave this code as is for this PR (it was around before so...), but bear in mind that this translation of information is not only useless but also a bit ugly

devito/finite_differences/coefficients.py
rhodrin: Yes!

devito/finite_differences/coefficients.py
rhodrin: Thanks, removed.

devito/types/dense.py
rhodrin: We need to access this from other places within devito. This is probably my python ignorance but is it not the convention that private properties only be accessed from within the class (or directly related classes) itself?

FabioLuporini: In general, you're right, but this class is exposed to the API and in Devito we give priority to user's visibility. "Should the user see ``coeff_symbol``?" --> no, so make it private

devito/finite_differences/coefficients.py
```diff
@@ -0,0 +1,259 @@
...
+        rules = {**rules, **generate_subs(*i)}
+    return rules
+
+
+def get_sym(functions):
+    for j in range(0, len(functions)):
+        try:
+            sym = functions[j].coeff_symbol
+            return sym
+        except:
```
FabioLuporini: In devito, all `except` should carry the specific Exception that is expected to be captured

devito/finite_differences/finite_difference.py
```diff
@@ -62,8 +64,27 @@ def wrapper(expr, *args, **kwargs):
     return wrapper
 
 
+def check_symbolic(func):
+    @wraps(func)
+    def wrapper(expr, *args, **kwargs):
+        functions = retrieve_functions(expr)
+        functions = [f for f in functions if not f.is_SparseFunction]
+        symbolic_coefficients = any(f.coefficients == 'symbolic' for f in functions)
```
FabioLuporini: I just had an idea. All our symbolic objects, including Functions, are of type Differentiable. Why don't we attach a property to Differentiable which does what we do here (or more?) , so that all this becomes a single line ``if expr.symbolic_coefficients`` . Then, we could exploit the same in ``Eq.__new__`` where it seems we need exactly the same information

devito/finite_differences/coefficients.py
rhodrin: Specific exceptions added.

FabioLuporini: I think you forgot to push the latest changes?

FabioLuporini:
Great, thanks for this massive effort. Merging.

PR #747, opened 2019-01-15
(not for merging) numpy 1.16 for testing with CI

tjb900:
I've not personally had any trouble with the 1.15 series (perhaps they fixed whatever the issue was), so just seeing if we can bring things up to date.


environment.yml
```diff
@@ -4,7 +4,7 @@ channels:
   - conda-forge
 dependencies:
   - python>=3.6
-  - numpy==1.14
+  - numpy==1.16
```
FabioLuporini: >= ?

tjb900: Just for the test!

tjb900:
So I _think_ this appears to have worked, now, on 1.16.

Travis seems to have died on a couple of them for an unknown reason, but no tests failed.
The errors in pipelines seem to be from a lot of `FutureWarning`s that probably deserve fixing also.

I'll push another commit now to test on 1.15

devito/data/data.py
```diff
@@ -179,7 +179,7 @@ def __setitem__(self, glb_idx, val):
                 if NONLOCAL in val_idx:
                     # no-op
                     return
-                val_idx = [i for i in val_idx if i is not PROJECTED]
+                val_idx = tuple([i for i in val_idx if i is not PROJECTED])
```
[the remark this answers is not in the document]
tjb900: These were DeprecationWarnings (or FutureWarnings, I can't remember).  Lists as indices have dual meaning at the moment.  You can extract a subset of indices by passing a list as an index:

e.g. `np.arange(10)[[2,4,6]] == np.array([2, 4, 6])`

but then you also can be used to index a multidimensional array, as we are.

The idea, AFAICT, is that the multidimensional case will have to be tuples in the near future to disambiguate these two.

devito/types/dense.py
```diff
@@ -607,10 +609,9 @@ def _C_make_dataobj(self, data):
     def _C_as_ndarray(self, dataobj):
         """Cast the data carried by a DiscreteFunction dataobj to an ndarray."""
         shape = tuple(dataobj._obj.size[i] for i in range(self.ndim))
-        ndp = np.ctypeslib.ndpointer(dtype=self.dtype, shape=shape)
-        data = cast(dataobj._obj.data, ndp)
-        data = np.ctypeslib.as_array(data, shape=shape)
-        return data
+        ctype_1d = dtype_to_ctype(self.dtype) * int(reduce(mul, shape))
+        buf = cast(dataobj._obj.data, POINTER(ctype_1d)).contents
+        return np.frombuffer(buf, dtype=self.dtype).reshape(shape)
```
eric-wieser: In 1.16 you could also have spelt this `cast(dataobj._obj.data, ndp).contents`

PR #748, opened 2019-01-15
YASK: misc fixes

FabioLuporini:
the AFFINE property detection is now much smarter than it used to be , hence YASK can now offload more IETs (in particular, higdon-like boundary condition loops)

### DIFF
--- a/devito/ir/stree/algorithms.py
+++ b/devito/ir/stree/algorithms.py
@@ -14,7 +14,7 @@ __all__ = ['st_build']
 
 def st_build(clusters):
     """
-    Create a :class:`ScheduleTree` from a :class:`ClusterGroup`.
+    Create a ScheduleTree from a ClusterGroup.
     """
     # ClusterGroup -> Schedule tree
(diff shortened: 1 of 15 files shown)


devito/ir/stree/algorithms.py
```diff
@@ -131,8 +129,7 @@ def __init__(self, node):
             self.nodes = [node]
 
         def is_compatible(self, node):
-            return (self.parent == node.parent
-                    and (self.dim == node.dim or node.dim.is_Sub))
+            return self.parent == node.parent and self.dim.root == node.dim.root
```
mloubout: don't they have aways same root if same parent?

FabioLuporini: no, you could have
```
for t
   for a
      ...
   for b
      ...
```
here the ``a`` and ``b`` subtree have the same parent (the ``t`` subtree), but the dimensions themselves, and therefore their root as well, are completely different

FabioLuporini:
merged.

PR #749, opened 2019-01-15
Lazy evaluation

mloubout:
WIP, just so you see where it's going


devito/finite_differences/utils.py
```diff
@@ -0,0 +1,84 @@
+from functools import partial
+
+from devito.finite_differences.finite_difference import left, right, centered
+from devito.finite_differences.derivative import Diff
+
+def partial_derivative(expr, deriv_order, dims, fd_order, stagger=centered, side=centered):
+    new_obj = Diff(expr, dims, deriv_order=deriv_order, fd_order=fd_order, stagger=centered, side=centered)
```
btrb: maybe 

`new_obj = Diff(expr, deriv_order=deriv_order, dims, fd_order=fd_order, stagger=centered, side=centered)`

just to keep the pattern ?

devito/finite_differences/utils.py
```diff
@@ -0,0 +1,84 @@
...
+    for (d, s) in zip(dimensions, function.staggered):
+        if s == 0:
+            side[d] = left
+        elif s == 1:
+            side[d] = right
+        else:
+            side[d] = centered
+
+    deriv_function = partial_derivative
+    c_deriv_function = partial_derivative
```
btrb: can't `deriv_function` be used instead at line 51?

devito/equation.py
```diff
@@ -168,5 +174,6 @@ def solve(eq, target, **kwargs):
     # turnaround time
     kwargs['rational'] = False  # Avoid float indices
```
btrb: shoudn't what comes from `kwargs` be initialized somewhere before, so we avoid using it like an actual parameter?

examples/seismic/acoustic/acoustic_example.py
```diff
@@ -54,7 +54,7 @@ def run(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
     save = full_run and not checkpointing
     # Define receiver geometry (spread across x, just below surface)
     rec, u, summary = solver.forward(save=save, autotune=autotune)
-
+    from IPython import embed; embed()
```
btrb: !

rhodrin:
From playing around with the branch this stuff looks really cool. Will leave some comments over the course of the day.

rhodrin:
This `if` statement and its contents in `equation.py` can be removed:
`if obj._uses_symbolic_coefficients:`
since exchanging numbers for symbols will not be occurring in the stencil method.

https://github.com/opesci/devito/blob/b565aebf67ebc5f2c7938ea2ea1900c1f58961cc/devito/equation.py#L64

devito/equation.py
```diff
@@ -104,8 +123,10 @@ def _symbolic_functions(self):
 
     def xreplace(self, rules):
         """"""
-        return self.func(self.lhs.xreplace(rules), self.rhs.xreplace(rules),
-                         subdomain=self._subdomain)
+        eq = Eq(self.lhs.xreplace(rules), self.rhs.xreplace(rules),
+                subdomain=self._subdomain)
+        eq._substitutions = self._substitutions
```
rhodrin: With the aforementioned `if` statement removed this line can also be removed since currently there's no need to hold any record of the desired substitutions after they've been carried out.

devito/equation.py
```diff
@@ -139,8 +160,22 @@ class Inc(Eq):
...
+        return Inc(lhs, rhs, evaluate=False, subdomain=self._subdomain,
+                   coefficients=self.substitutions)
+
     __repr__ = __str__
 
+    def xreplace(self, rules):
+        """"""
+        eq = Inc(self.lhs.xreplace(rules), self.rhs.xreplace(rules),
+                 subdomain=self._subdomain)
+        eq._substitutions = self._substitutions
```
rhodrin: I'm not sure if `176` is needed?

devito/finite_differences/finite_difference.py
```diff
@@ -287,78 +290,7 @@ def generic_derivative(expr, dim, fd_order, deriv_order, stagger=None, symbolic=
         subs = dict((d, indices[i].subs({dim: d})) for d in all_dims)
         deriv += expr.subs(subs) * c[i]
 
-    return deriv.evalf(_PRECISION)
-
-
-def generate_fd_shortcuts(function):
```
rhodrin: :)

rhodrin: :)

devito/types/dense.py
```diff
@@ -189,6 +189,10 @@ def grid(self):
     def staggered(self):
         return self._staggered
 
+    @property
+    def stencil(self):
```
rhodrin: Just wondering what the purpose of this property is?

devito/types/dense.py
```diff
@@ -881,27 +885,27 @@ class Function(DiscreteFunction, Differentiable):
...
+    Derivative(g(x, y), x)
     >>> (f + g).dx
-    -(f(x, y) + g(x, y))/h_x + (f(x + h_x, y) + g(x + h_x, y))/h_x
+    Derivative(f(x, y) + g(x, y), x)
 
     First-order derivatives through left/right finite-difference approximations
 
     >>> f.dxl
-    f(x, y)/h_x - f(x - h_x, y)/h_x
+    Derivative(f(x, y), x)
```
rhodrin: Since the output of `f.dxl` and `f.dxr` now look the same if would be nice to add something here illustrating the difference between them.

devito/finite_differences/differentiable.py
```diff
@@ -58,7 +58,20 @@ def is_Staggered(self):
     def _fd(self):
         return dict(ChainMap(*[getattr(i, '_fd', {}) for i in self._args_diff]))
 
-    @cached_property
```
rhodrin: Just wondering why this is no longer cached?

devito/equation.py
mloubout: >to hold any record of the desired substitutions after they've been carried out.

xreplace is not only called by the substitution. For example it is called in `inject` and that may contain FD expression that would loose the substitutions without keeping track of it.

devito/equation.py
mloubout: It is, see comment above

devito/types/dense.py
mloubout: I thought about it but I would rather not over clutter `Derivative` with a bunc of extra things like side/stagger/order/.... 
THis is the default `sympy.Derivative` repr and I think it makes sense

devito/equation.py
rhodrin: Yeah, I see. After `.stencil` is called (and hence we can be sure substitutions have been carried out) it would be nice if this property is expunged but I'll worry about that in the future when it becomes relevant.

devito/types/dense.py
rhodrin: Oh I agree with that, I was just wondering is the docstring should also include i.e. `f.dxl.stencil`.

devito/finite_differences/differentiable.py
FabioLuporini: good point

examples/seismic/acoustic/operators.py
```diff
@@ -10,10 +10,15 @@ def laplacian(field, m, s, kernel):
...
     H = (laplacian + s**2/12 laplacian(1/m*laplacian))
-    :param field: Symbolic TimeFunction object, solution to be computed
-    :param m: square slowness
-    :param s: symbol for the time-step
-    :return: H
+
+    Parameters
+    ----------
+    field : TimeFunction
+        Symbolic TimeFunction object, solution to be computed
```
FabioLuporini: all the descriptions should have a full stop at the end

examples/seismic/tti/tti_example.py
```diff
@@ -57,14 +54,14 @@ def run(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
...
-                        choices=['centered', 'shifted', 'staggered'],
+                        choices=['centered', 'staggered'],
                         help="Choice of finite-difference kernel")
     parser.add_argument("-dse", default="advanced",
                         choices=["noop", "basic", "advanced",
                                  "speculative", "aggressive"],
                         help="Devito symbolic engine (DSE) mode")
     parser.add_argument("-dle", default="advanced",
-                        choices=["noop", "advanced", "speculative"],
+                        choices=["basic", "advanced", "speculative"],
```
FabioLuporini: basic was missing, right, but noop can stay too

examples/seismic/tti/wavesolver.py
```diff
@@ -10,26 +10,31 @@ class AnisotropicWaveSolver(object):
...
+    ----------
+    model : Model
+        Physical model with domain parameters
+    geometry : AcquisitionGeometry
+        Source and receivers geometry
+    space_order : Int
+        Order of the spatial stencil discretisation (default: 2)
 
-    Note: space_order must always be greater than time_order
+    Note: space_order must always be an even number
```
FabioLuporini: 1) numpydoc for notes:
```
Notes
-----
...
```

2) raise a ValueError if that's not honored

tests/test_tti.py
```diff
@@ -113,15 +113,23 @@ def test_tti_staggered(shape):
...
-    rec2, u2, v2, _ = solver_tti2.forward(kernel='staggered')
+    @switchconfig(dse='aggressive', dle='advanced')
+    def run_tti(solver):
+        rec1, u1, v1, _ = solver.forward(kernel='staggered')
+        return rec1, u1, v1
+
+    rec1, u1, v1 = run_tti(solver_tti)
+
+    @switchconfig(dse='aggressive', dle='basic')
+    def run_tti2(solver):
```
FabioLuporini: now that I think about it, if this is just a DSE/DLE test, why not moving it to test_dse or test_dle ? I guess I'm a bit confused

devito/finite_differences/finite_difference.py
```diff
@@ -117,7 +117,7 @@ def first_derivative(expr, dim, fd_order=None, side=centered, matvec=direct,
 
     This is also more easily obtainable via:
 
-    >>> (f*g).dx
+    >>> (f*g).dx.evaluate
```
FabioLuporini: I wouldn't do .evaluate here, and show the Derivative repr instead.

Then, line below, I'd add 

```
The corresponding finite difference approximation can be obtained via

>>> (f*g).dx.evaluate
...
```

I'd add this pattern to all FD functions in this file

devito/finite_differences/finite_difference.py
```diff
@@ -184,7 +186,7 @@ def second_derivative(expr, dim, fd_order, stagger=None, **kwargs):
 
     This is also more easily obtainable via:
 
-    >>> (f*g).dx2
+    >>> (f*g).dx2.evaluate
```
FabioLuporini: for example, here

devito/finite_differences/differentiable.py
FabioLuporini: cached_property

LAZY-TODO.txt
```diff
@@ -0,0 +1,8 @@
+[ ] repr (or not just repr) inconsistencies: (f*g).dx.dx.dx.dy  vs  (f*g).dx.dy
+[ ] lack of tests (lazy -> evaluation -> check expected "stencil")
+[ ] first_derivative, cross_derivative, etc, should return a Derivative object
```
FabioLuporini: this is important.

devito/finite_differences/derivative.py
```diff
@@ -0,0 +1,140 @@
...
+                    orders[dims[i-1]] = d
+                elif d in orders:
+                    orders[d] =+ 1
+                else:
+                    orders[d] = 1
+
+        new_dims = []
+        for k, v in orders.items():
+            new_dims += [k for _ in range(v)]
+        new_dims = tuple(new_dims)
```
FabioLuporini: these three lines can be turned into (the more pythonic)

```
new_dims = tuple(flatten(k*len(v) for k, v in orders.items()))
```
```

File: devito/finite_differences/derivative.py [hunk 19]
```diff
@@ -0,0 +1,140 @@
...
+        ndim = len(as_tuple(dims))
+        norder = len(as_tuple(deriv_order)) if deriv_order is not None else 0
+
+        if deriv_order is not None and ndim != norder:
+            raise ValueError("Different number of Dimensions and derivative orders")
+    
+        if deriv_order is not None:
+            orders = dict((d, o) for d, o in zip(dims, as_tuple(deriv_order)))
+        else:
+            orders = dict()
```
FabioLuporini: I guess you want, 

```from collections import Counter```

```
orders = Counter()
...
```

File: devito/finite_differences/derivative.py [hunk 20]
```diff
@@ -0,0 +1,140 @@
...
+            for i, d in enumerate(dims):
+                if isinstance(d, (tuple, sympy.Tuple)):
+                    orders[d[0]] = d[1]
+                elif int(d) == d:
+                    orders[dims[i-1]] = d
+                elif d in orders:
```
FabioLuporini: then here just

```
else:
    orders[d] += 1
```

devito/finite_differences/derivative.py
```diff
@@ -0,0 +1,140 @@
...
+        if deriv_order is not None and ndim != norder:
+            raise ValueError("Different number of Dimensions and derivative orders")
+    
+        if deriv_order is not None:
+            orders = dict((d, o) for d, o in zip(dims, as_tuple(deriv_order)))
+        else:
+            orders = dict()
+            for i, d in enumerate(dims):
+                if isinstance(d, (tuple, sympy.Tuple)):
+                    orders[d[0]] = d[1]
```
FabioLuporini: However, I'm really confused... this is assuming a 2-element length tuple? 

I think we need lots of examples in the docstring, because this is more than impenetrable

devito/finite_differences/derivative.py
```diff
@@ -0,0 +1,140 @@
...
+    @cached_property
+    def indices(self):
+        return self.expr.indices
+
+    @cached_property
+    def staggered(self):
+        return self.expr.staggered
+
+    @cached_property
+    def _stagger_setup(self):
```
FabioLuporini: I also think it's time this code goes away. Here and in Function as well. As I pointed out in the past, ``left`` ``right`` ``centered`` should be given proper ``val`` (they are a Tag, so they do have a ``val`` field -- let's exploit it!), so that we can instead write something like:

```
if self.is_Staggered:
    side = {...}  # one liner
else:
    side = dict((d, None) for d in self.dims)
```

This is something I pointed out in a previous PR and I'm pretty sure I said : let's fix it in the next one. Now it's time

devito/finite_differences/utils.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def deriv_function(expr, deriv_order, dims, fd_order, side=centered, **kwargs):
+        return Derivative(expr, dims, deriv_order=deriv_order, fd_order=fd_order,
+                          side=side, **kwargs)
+
+    derivatives = dict()
+    done = []
+    for d in dimensions:
+        # Dimension is treated, remove from list
+        done += [d]
```
FabioLuporini: append

devito/finite_differences/utils.py
```diff
@@ -0,0 +1,77 @@
...
+    dimensions = function.indices
+    space_fd_order = function.space_order
+    time_fd_order = function.time_order if (function.is_TimeFunction or
+                                            function.is_SparseTimeFunction) else 0
+
+    def deriv_function(expr, deriv_order, dims, fd_order, side=centered, **kwargs):
+        return Derivative(expr, dims, deriv_order=deriv_order, fd_order=fd_order,
+                          side=side, **kwargs)
+
+    derivatives = dict()
```
FabioLuporini: use ``{}`` , not ``dict()``. It's more pythonic

examples/seismic/acoustic/acoustic_example.py
```diff
@@ -53,24 +50,25 @@ def run(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
     # checkpointing, PyRevolve will take care of the time history
     save = full_run and not checkpointing
     # Define receiver geometry (spread across x, just below surface)
-    rec, u, summary = solver.forward(save=save, autotune=autotune)
+    rec, u, summary = solver.forward(save=save, autotune=autotune, kernel=kernel)
```
FabioLuporini: I still think this is more than odd, but it's entirely possible I'm just wrong. So I'd like someone's input as well. For example, @rhodrin , what do you think? should ``kernel`` be a .forward/.backward argument, or a solver (``acoustic_setup(...)`` ) argument?

Maybe it's the name "acoustic_setup" that confuses me?

devito/finite_differences/finite_difference.py
```diff
@@ -232,12 +234,13 @@ def cross_derivative(expr, dims, fd_order, deriv_order, stagger=None, **kwargs):
 
     This is also more easily obtainable via:
 
-    >>> (f*g).dxdy
+    >>> (f*g).dxdy.evaluate
```
mloubout: now this is funny, this will randomly change as it looks like evaluation order is non deterministic so it will do x then y or y then x depending on the mood

LAZY-TODO.txt
mloubout: As said in my comments I disagree, I think we should keep these functions as finite_differences functions one for backward compatibility and two to be able to play with FD easily

devito/finite_differences/derivative.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+
+    def __new__(cls, expr, *dims, **kwargs):
+        # Check dims, can be a dimensions, multiple dimensions as a tuple
+        # or a tuple of tuple (ie ((x,1),))
+        if len(dims) == 1:
+            if isinstance(dims[0], (tuple, sympy.Tuple)):
+                orders = kwargs.get('deriv_order', dims[0][1])
+                if dims[0][1] != orders:
+                    raise ValueError("Two different value of deriv_order")
```
FabioLuporini: values

devito/finite_differences/derivative.py
```diff
@@ -0,0 +1,157 @@
...
+
+    Once again, this can be access via the differential shortcut
+
+    >> u.dx2
+    """
+
+    def __new__(cls, expr, *dims, **kwargs):
+        # Check dims, can be a dimensions, multiple dimensions as a tuple
+        # or a tuple of tuple (ie ((x,1),))
+        if len(dims) == 1:
```
FabioLuporini: we can most likely avoid this branch, maybe via ``as_tuple`` ?

devito/finite_differences/utils.py
```diff
@@ -0,0 +1,139 @@
...
+
+        else:
+            ind = [(dim + i*dim.spacing) for i in range(-order//2, order//2 + 1)]
+            x0 = dim
+            if order < 2:
+                ind = [dim, dim + diff]
+    return ind, x0
+
+
+def form_side(dimensions, function):
```
FabioLuporini: just noticed this: why should ``dimensions`` be different than ``function.dimensions`` ?

devito/finite_differences/utils.py
```diff
@@ -0,0 +1,139 @@
...
+def generate_fd_shortcuts(function):
+    """Create all legal finite-difference derivatives for the given Function."""
+    dimensions = function.indices
+    space_fd_order = function.space_order
+    time_fd_order = function.time_order if (function.is_TimeFunction or
+                                            function.is_SparseTimeFunction) else 0
+
+    from devito.finite_differences.derivative import Derivative
+
+    def deriv_function(expr, deriv_order, dims, fd_order, side=centered, **kwargs):
```
FabioLuporini: I think this function is useless, you should be able to always use:

```
Derivative(expr, *as_tuple(dims), deriv_order=deriv_order, fd_order=fd_order, side=side, **kwargs)
```

devito/finite_differences/utils.py
```diff
@@ -0,0 +1,139 @@
...
+        else:
+            ind = [(dim + i*dim.spacing) for i in range(-order//2, order//2 + 1)]
+            x0 = dim
+            if order < 2:
+                ind = [dim, dim + diff]
+    return ind, x0
+
+
+def form_side(dimensions, function):
+    side = dict()
```
FabioLuporini: side = {}

devito/finite_differences/derivative.py
```diff
@@ -0,0 +1,162 @@
...
+
+    """
+    An unevaluated Derivative, which carries metadata (Dimensions,
+    derivative order, etc) to be evaluated.
+
+    Parameters
+    ----------
+
+    expr : symbolic expression
+    dims : Dimension or tuple
```
FabioLuporini: this should be updated based on the inline comment below

devito/finite_differences/derivative.py
```diff
@@ -0,0 +1,162 @@
...
+    >> Derivative(u, (x, 2))
+    Derivative(u(x, y), (x, 2))
+
+    Once again, this can be access via the differential shortcut
+
+    >> u.dx2
+    """
+
+    def __new__(cls, expr, *dims, **kwargs):
+        # Check dims, can be a dimensions, multiple dimensions as a tuple
```
FabioLuporini: this comment should be moved up inside the docstring

devito/finite_differences/derivative.py
```diff
@@ -0,0 +1,162 @@
...
+        # or a tuple of tuple (ie ((x,1),))
+        if len(dims) == 1:
```
FabioLuporini: another Q: can you check the case `len(dims) == 0` and `raise ValueError` if that's the case?

devito/finite_differences/derivative.py
```diff
@@ -0,0 +1,162 @@
...
+            if isinstance(dims[0], (tuple, sympy.Tuple)):
```
FabioLuporini: what if one provides a list? or more generically, an `Iterable` ?

I **think** you want:

```
from collections import Iterable  # quite sure it lives in collections, otherwise find out
if isinstance(dims[0], Iterable)
```

devito/finite_differences/derivative.py
```diff
@@ -0,0 +1,162 @@
...
+                orders = kwargs.get('deriv_order', dims[0][1])
```
FabioLuporini: why do you assume that ``dims[0]`` is an iterable again? there might be an error in the user input. I'd rethink this part

devito/finite_differences/derivative.py
```diff
@@ -0,0 +1,162 @@
...
+                new_dims = tuple([dims[0][0]]*dims[0][1])
+            else:
+                orders = kwargs.get('deriv_order', 1)
+                new_dims = tuple([dims[0]]*orders)
+        else:
+            # ie ((x, 2), (y, 3))
+            new_dims = []
+            orders = []
+            for d in dims:
+                if isinstance(d, (tuple, sympy.Tuple)):
```
FabioLuporini: again, Iterable

devito/finite_differences/derivative.py
```diff
@@ -0,0 +1,162 @@
...
+                    new_dims += [d[0] for _ in range(d[1])]
```
FabioLuporini: append/extend -- here and below

devito/finite_differences/derivative.py
```diff
@@ -0,0 +1,162 @@
...
+class Derivative(sympy.Derivative, Differentiable):
+
+    """
+    An unevaluated Derivative, which carries metadata (Dimensions,
+    derivative order, etc) to be evaluated.
+
+    Parameters
+    ----------
+
+    expr : symbolic expression
```
FabioLuporini: expr : expr-like
    explanation

devito/finite_differences/derivative.py
FabioLuporini: so, it should be something like

```
dims : Dimension or Iterable of Dimensions or ...
```

though in fact the actual elements could be 2-tuples?

devito/finite_differences/derivative.py
```diff
@@ -0,0 +1,162 @@
...
+    dims : Dimension or tuple
+        Dimenions wrt which to take the derivative
```
FabioLuporini: Also, note: **all** explanations should terminate with a full stop.

So: 
```
Dimenions w.r.t. which to take the derivative.
```

devito/finite_differences/derivative.py
```diff
@@ -0,0 +1,162 @@
...
+    deriv_order: Int or Tuple
```
FabioLuporini: small tuple

devito/finite_differences/derivative.py
FabioLuporini: in the docstring you can use "tuple" instead of Iterable, but then the code it's better off to support all sort of tuple-like iterables

devito/finite_differences/derivative.py
mloubout: sympy does it already in `new` so no need no

devito/finite_differences/derivative.py
mloubout: No, `Derivative(u, ((x, 2), (y, 2))` is a valid input it's 2nd order in x and y and `dims[0]` is an iterable.

devito/finite_differences/finite_difference.py
FabioLuporini: this is fixed right?

mloubout: yes

devito/finite_differences/coefficients.py
```diff
@@ -130,7 +130,7 @@ class Substitutions(object):
     Now create a Devito equation and pass to it 'subs'
 
     >>> from devito import Eq
-    >>> Eq(u.dt+u.dx, coefficients=subs)
+    >>> Eq(u.dt+u.dx, coefficients=subs).evaluate
```
FabioLuporini: why adding evaluate here?

mloubout: To show the custom coefficient are indeed used

tests/test_symbolic_coefficients.py
```diff
@@ -48,6 +48,8 @@ def test_coefficients(self, expr, sorder, dorder, dim, weights, expected):
         coeffs = Coefficient(order, u, dim, weights)
 
         eq = Eq(eval(expr), coefficients=Substitutions(coeffs))
-
+        print("test")
```
FabioLuporini: reminder for myself: leftovers

devito/types/sparse.py
```diff
@@ -650,7 +650,7 @@ def interpolate(self, expr, offset=0, increment=False, self_subs={}):
         idx_subs, temps = self._interpolation_indices(variables, offset)
 
         # Substitute coordinate base symbols into the interpolation coefficients
-        args = [expr.subs(v_sub) * b.subs(v_sub)
+        args = [expr.xreplace(v_sub) * b.xreplace(v_sub)
```
FabioLuporini: remind me why? I guess you've told me already...but I forgot

mloubout: subs has sooooo many cases and I don't like it.
xreplace is nice and easy and is implemented in `Differentiable`

FabioLuporini:
things that still need discussion:
- I have a feeling that `lazy` is killing `solve` , as `solve` requires the expanded expression. What shall we do?
- `first_derivative`, `generic_derivative` etc are expanding derivatives. This is, in my opinion, quite the opposite of what we want as now suddenly some codes are mixin up unevaluated (via eg `u.dx`) and evaluated (via eg `first_derivative(u, x)`) objects. I remark that we should reconsider this
- we need many more tests in `test_derivatives`, perhaps by adding more cases to the various parametrized tests
- staggering, as said above, is a stopper and needs fixing
- `tools.py` and `utils.py` should/could be merged
- the edits (image-related) to the jupyter notebooks should be unstaged as, AFAICT, these aren't affected by this PR. Correct? Let me know if you're struggling with this, I can do it

devito/finite_differences/utils.py
FabioLuporini: reminder

devito/types/dense.py
FabioLuporini: if we don't change the repr, then maybe we should just drop it...

devito/equation.py
```diff
@@ -188,4 +201,6 @@ def solve(eq, target, **kwargs):
     # turnaround time
     kwargs['rational'] = False  # Avoid float indices
     kwargs['simplify'] = False  # Do not attempt premature optimisation
-    return sympy.solve(eq, target, **kwargs)[0]
+    if hasattr(eq, 'lhs'):
```
FabioLuporini: ???

devito/finite_differences/tools.py
```diff
@@ -0,0 +1,211 @@
...
+
+def generate_fd_shortcuts(function):
+    """Create all legal finite-difference derivatives for the given Function."""
+    dimensions = function.indices
+    s_fd_order = function.space_order
+    t_fd_order = function.time_order if (function.is_TimeFunction or
+                                         function.is_SparseTimeFunction) else 0
+    orders = tuple(t_fd_order if i.is_Time else s_fd_order for i in dimensions)
+
+    from devito.finite_differences.derivative import Derivative
```
FabioLuporini: never local imports unless it's a pathological case

tests/test_derivatives.py
```diff
@@ -63,6 +69,40 @@ def test_stencil_derivative(self, SymbolType, dim):
...
+        (Function, ['dx2dydz'], 3, 'Derivative(u(x, y, z), (x, 2), y, z)'),
+        (Function, ['dx2', 'dy'], 3, 'Derivative(Derivative(u(x, y, z), (x, 2)), y)'),
+        (Function, ['dx2dy', 'dz2'], 3,
+         'Derivative(Derivative(u(x, y, z), (x, 2), y), (z, 2))'),
+        (TimeFunction, ['dx2'], 3, 'Derivative(u(t, x, y, z), (x, 2))'),
+        (TimeFunction, ['dx2dy'], 3, 'Derivative(u(t, x, y, z), (x, 2), y)'),
+        (TimeFunction, ['dx2', 'dy'], 3,
+         'Derivative(Derivative(u(t, x, y, z), (x, 2)), y)')
+    ])
+    def test_unevaluation(self, SymbolType, derivative, dim, expected):
```
FabioLuporini: Can we add a single new case, in which **lots** of derivatives are taken, checking that unevaluation really does the right thing. Such as, ``.dx.dy.dx2.dz.dzdy``

tests/test_derivatives.py
```diff
@@ -63,6 +69,40 @@ def test_stencil_derivative(self, SymbolType, dim):
...
+        # Make sure the FD evaluation executes
+        expr.evaluate
+
+    @pytest.mark.parametrize('expr,expected', [
+        ('u.dx + u.dy', 'Derivative(u, x) + Derivative(u, y)'),
+        ('u.laplace',
+         'Derivative(u, (x, 2)) + Derivative(u, (y, 2)) + Derivative(u, (z, 2))'),
+        ('(u.dx*u.dy).dx', 'Derivative(Derivative(u, x) * Derivative(u, y), x)')
+    ])
+    def test_arithmetic(self, expr, expected):
```
FabioLuporini: Can we add more test cases? 

For example

```
(u.dx + u.dy).dx
((u.dx + u.dy).dx + u.dxdy).dx
(u**4).dx
(u / 4).dx
((u.dx + v.dy).dx * v.dx).dy. # yeah, could do arithmetic with another Function too
```

etc. I think this **really needs to be bullet proof**, otherwise things can get screwed in the most unpredictable ways!

examples/seismic/tti/tti_example.py
FabioLuporini: reminder

devito/finite_differences/tools.py
mloubout: You did that one so gonna assume pathological

devito/equation.py
mloubout: Just turn input into `expr=0` so if it's an equation do diff lhs-rhs. This way don't pass eq to solve. Solve(eq) returns sympy.add and we back to wrong types

tests/test_derivatives.py
mloubout: Sure

devito/finite_differences/coefficients.py
```diff
@@ -130,7 +130,12 @@ class Substitutions(object):
     Now create a Devito equation and pass to it 'subs'
 
     >>> from devito import Eq
-    >>> Eq(u.dt+u.dx, coefficients=subs)
+    >>> eq = Eq(u.dt+u.dx, coefficients=subs)
+
+    When evaluated, the derivatives will use the custom coefficients. We can
```
FabioLuporini: typo: double can

FabioLuporini: (my fault sorry!)

devito/finite_differences/tools.py
FabioLuporini: oki

devito/ir/clusters/algorithms.py
```diff
@@ -98,6 +98,7 @@ def guard(clusters):
                 for d in e.conditionals:
                     condition = guards.setdefault(d.parent, [])
                     condition.append(d.condition or CondEq(d.parent % d.factor, 0))
+
```
FabioLuporini: ? is this just a leftover or is it a logical separation of code?

devito/types/dense.py
FabioLuporini: It's at least worth adding a comment along the lines of

```
Note that the fact that it's a left-derivative isn't captured in the representation. However, upon derivative expansion, this becomes clear

>>> f.dxl.evaluate
....
```

devito/finite_differences/finite_difference.py
mloubout: Well no, this is the `first_derivative` function that always returns the FD. the `.dx` only is in `Derivative`

devito/types/dense.py
mloubout: Done

examples/seismic/tti/wavesolver.py
mloubout: Done

tests/test_tti.py
mloubout: Done

devito/finite_differences/finite_difference.py
mloubout: smae

tests/test_derivatives.py
mloubout: Done

tests/test_derivatives.py
mloubout: Done

devito/equation.py
mloubout: cfd tutorials 5 and 6 fail without this.

FabioLuporini: what I'm trying to say is that ``hasattr`` should be avoided at all costs, when possible. Here it seems possible: all and only ``Eq`` have a ``lhs`` attribute, so this should rather be
```
if isinstance(eq, Expr):  # in fact, change the var name from eq into expr
    eq = eq.lhs - eq.rhs
```

FWIW, I also tried dropping the ``if`` entirely (ie, even without the tweak above) and both the ``cfd`` and all tests in ``test_derivative.py`` pass

Despite this might be seen as nitpicking, it really is not, because it's self-documenting what we are prepared to accept in ``solve`` (which, one day, might become our own (faster) ``solve``)

devito/finite_differences/derivative.py
```diff
@@ -0,0 +1,192 @@
...
+    Derivative(u(x, y), (x, 2))
+    """
+
+    _state = ('expr', 'dims', 'side', 'stagger', 'fd_order', 'transpose')
+
+    def __new__(cls, expr, *dims, **kwargs):
+        if type(expr) == sympy.Derivative:
+            raise ValueError("Cannot nest sympy.Derivative with devito.Derivative")
+        if not isinstance(expr, Differentiable):
+            print(expr, type(expr), dims, kwargs)
```
FabioLuporini: AAAHHHH

devito/finite_differences/derivative.py
```diff
@@ -0,0 +1,192 @@
...
+                    new_dims.extend([d for _ in range(o)])
+                    orders.append(o)
+            new_dims = as_tuple(new_dims)
+            orders = as_tuple(orders)
+
+        # Finite difference orders
+        fd_orders = kwargs.get('fd_order', tuple([expr.space_order]*len(dims)))
+        if len(dims) == 1 and isinstance(fd_orders, Iterable):
+            fd_orders = fd_orders[0]
+        # SymPy expects the list of variable w.r.t. which we differentiate to be a list
```
FabioLuporini: a blank line between this comment and the block above wouldn't harm

devito/finite_differences/derivative.py
```diff
@@ -0,0 +1,192 @@
...
+    def side(self):
+        return self._side
+
+    @property
+    def transpose(self):
+        return self._transpose
+
+    @property
+    def T(self):
+        """Transpose of the Derivative."""
```
FabioLuporini: this is clear to me, but I'm not sure the average user will understand what we exactly mean by "transpose of the derivative" ?

devito/finite_differences/finite_difference.py
```diff
@@ -117,7 +53,7 @@ def first_derivative(expr, dim, fd_order=None, side=centered, matvec=direct,
 
     This is also more easily obtainable via:
 
-    >>> (f*g).dx
+    >>> (f*g).dx.evaluate
     -f(x, y)*g(x, y)/h_x + f(x + h_x, y)*g(x + h_x, y)/h_x
 
     The adjoint mode
```
FabioLuporini: nitpicking: we may shown again the shortcut notation for this via .transpose

devito/finite_differences/tools.py
```diff
@@ -0,0 +1,211 @@
...
+
+    def deriv_function(expr, deriv_order, dims, fd_order, side=centered, **kwargs):
```
FabioLuporini: we can greatly probably rewrite this as:
```
def deriv_function(...):
    return Derivative(expr, *as_tuple(dims), deriv_order=deriv_order, fd_order=fd_order, side=side, **kwargs)
```

devito/finite_differences/tools.py
```diff
@@ -0,0 +1,211 @@
...
+                              side=side, **kwargs)
+        return Derivative(expr, dims, deriv_order=deriv_order, fd_order=fd_order,
+                          side=side, **kwargs)
+
+    side = form_side(dimensions, function)
+    all_combs = dim_with_order(dimensions, orders)
+
+    derivatives = {}
+
+    # All conventiona fd shortcuts
```
FabioLuporini: typo

also, fd should be capital letters

devito/finite_differences/tools.py
```diff
@@ -0,0 +1,211 @@
...
+        fd_orders = tuple(t_fd_order if d.is_Time else s_fd_order for d in fd_dims)
+
+        deriv = partial(deriv_function, deriv_order=d_orders, dims=fd_dims,
+                        fd_order=fd_orders,
+                        stagger=tuple(side[d] for d in fd_dims))
+        name_fd = deriv_name(fd_dims, d_orders)
+        desciption = 'derivative of order %s w.r.t dimension %s' % (d_orders, fd_dims)
+        derivatives[name_fd] = (deriv, desciption)
+
+    # Add non-conventional, non-centered first-order FDs
```
FabioLuporini: that's not complete, staggered is also handled inside

mloubout:
>why have some images changed in the notebooks?

I reran all because output f some cells change because of lazy evaluation.
BTW pretty nice, if you make a cell with like on `u.dx` it prints out  a super pretty latex rendering

devito/equation.py
mloubout: >FWIW, I also tried dropping the if entirely (ie, even without the tweak above) and both the cfd and all tests in test_derivative.py pass

I can guaranty you it does not pass. Can change to isinstance if you want fine with me

FabioLuporini: no I mean I was dropping the ``if``, but I still had the ``eq = eq.lhs - eq.rhs``

yes, turning into ``isinstance(expr, Eq)`` would be preferable 

do you get why it's better this way? basically because it's a stronger (less loose) check, and this is a fundamental piece of API (as you've often rightly pointed out)

devito/finite_differences/tools.py
mloubout: Non-conventional ones yes, but all standard, including staggered, are handled in previous one.

devito/finite_differences/tools.py
mloubout: DOne

devito/finite_differences/tools.py
mloubout: Done, testing

devito/finite_differences/finite_difference.py
mloubout: Done

devito/finite_differences/derivative.py
mloubout: Added more explanation

devito/finite_differences/derivative.py
mloubout: DOne

devito/finite_differences/derivative.py
mloubout: Removed

tests/test_derivatives.py
```diff
@@ -268,3 +321,37 @@ def test_all_shortcuts(self, so):
...
+        g = Function(name='g', grid=grid, space_order=so)
+        g_deriv = Function(name='g_deriv', grid=grid, space_order=so)
+        g.data[:, :] = np.random.rand(50, 50, 50)
+
+        # Check symbolic expressio nare expected oens for the adjoint .T
+        deriv = getattr(f, derivative)
+        expected = adjoint_coeff * getattr(f, adjoint_name).evaluate
+        assert deriv.T.evaluate == expected
+
+        # Compute numerical dervivatives and verify dot test
```
georgebisbas: derivatives as well

devito/equation.py
mloubout: Changed. Yes I get it, it's more robust, was tryign to avoid extra imports but that's stupid Eq already imported, also wondering what will happen if an `Inc` is passed now but for later. Anyway I think dropping `if` is dangerous so I kept it

examples/seismic/tti/tti_example.py
```diff
@@ -63,7 +60,7 @@ def run(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
                         choices=["noop", "basic", "advanced", "aggressive"],
                         help="Devito symbolic engine (DSE) mode")
     parser.add_argument("-dle", default="advanced",
-                        choices=["noop", "advanced", "speculative"],
+                        choices=["noop", "basic", "advanced", "speculative"],
```
georgebisbas: Is basic an option for DLE?

devito/equation.py
FabioLuporini: maybe we should raise a ValueError if an Inc ? anyway...that's fine for now

devito/finite_differences/tools.py
```diff
@@ -0,0 +1,207 @@
...
+
+def dim_with_order(dims, orders):
+    """
+    Create all possible derivative order for each dims
+    for example dim_with_order((x, y), 1) outputs:
+    [(1, 0), (0, 1), (1, 1)]
+    """
+    ndim = len(dims)
+    max_order = np.max(orders)
+    # Get all combinations and remove (0, 0 ,0)
```
georgebisbas: (0, 0, 0)

tests/test_derivatives.py
```diff
@@ -268,3 +321,37 @@ def test_all_shortcuts(self, so):
...
+        clear_cache()
+        grid = Grid(shape=(50, 50, 50))
+        f = Function(name='f', grid=grid, space_order=so)
+        f_deriv = Function(name='f_deriv', grid=grid, space_order=so)
+        f.data[:, :] = np.random.rand(50, 50, 50)
+        g = Function(name='g', grid=grid, space_order=so)
+        g_deriv = Function(name='g_deriv', grid=grid, space_order=so)
+        g.data[:, :] = np.random.rand(50, 50, 50)
+
+        # Check symbolic expressio nare expected oens for the adjoint .T
```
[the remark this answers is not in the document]
mloubout: fixed

examples/seismic/tti/tti_example.py
mloubout: removed indeed is not

tests/test_derivatives.py
mloubout: done

devito/finite_differences/derivative.py
```diff
@@ -0,0 +1,196 @@
...
+
+    @property
+    def transpose(self):
+        return self._transpose
+
+    @property
+    def T(self):
+        """Transpose of the Derivative.
+        FD derivatives can be represented as matrices and have adjoint/transpose.
+        This is really usefull for more advance FD definitions. For example
```
FabioLuporini: Code homogeneity: 
```
"""
Transpose of the Derivative.

FD Derivatives can be ...
```
ie, start on new line; leave blank line after initial description
Also, maybe drop "of" ?
Also, double typo: 'useful' and 'advanced'

Also: "For example, the conventional transpose of the laplacian (Laplacian?) is ..."

mloubout: I would keep the "of". Done for all the rest

FabioLuporini:
Merged

PR #750, opened 2019-01-16
MPI codegen improvements

FabioLuporini:
Better halo exchange routines -- in terms of both implementation in Devito and  quality of the generated code

~depends on #748 , so the diff looks bigger than it actually is~

Note that the branch name is misleading so ignore it


FabioLuporini:
Ready to be reviewed

FabioLuporini:
> Why is codecov reduced?
I don't know what these tiny fluctuations (e.g. 0.03% ) are due to, but it's not the first time I see them, could simply be that I managed to drop some obsolete code, and now overall we cover a (tiny) smaller percentage of code paths?

FabioLuporini:
Merged.

PR #751, opened 2019-01-17
stable dependency versions for CI

tjb900:
Really putting this here to start a discussion.

This PR embeds fixed versions of all packages used for CI into the tree, in `ci_environment.yml` (for conda builds) and `ci_requirements.txt` (for pip).

The idea is that a PR should not fail for any reason other than the changes in tree (obviously this is aspirational given potential instability in the CI systems, but still).  The idea is that we don't e.g. have PRs blocked because numpy decided to release a new version, as has happened in the past.

To facilitate maintenance of this scheme, there also is a script, `update_ci_env.sh`, which builds a temporary environment using the default versions of packages (as obtained from environment.yml, as a user might), and saves that into the `ci_requirements.txt` and `ci_environment.yml`.

The idea is that, if this approach is adopted, we would then have standalone PRs which bring the dependencies we test against up to date.  Those PRs will force us to address any incompatibility issues within them.

Notes:

1) this doesn't correctly handle the MPI builds yet.  We can do that if anyone actually likes this idea.
2) I think it would be good to know that master has become broken due to outside dependencies being updated.  I would argue that this is a good job for a nightly CI run against the latest dependencies, to let us know that either:
  a) we need to address some incompatibilities, or
  b) our environment.yml/requirements.txt need editing to keep end-users away from broken combinations


FabioLuporini:
I think I like this. Do you know if this is common practice in, for example, big open source projects?

Perhaps @tmbgreaves would like to chime in -- any thoughts, Tim?

Apologies for hijacking the conversion:
1) is it possible to move **all** CI-related files to a separate ``ci`` subdirectory? Or are there things that are simply expected to be found in the root directory, while being too complicated to change their default location? Note: I'm not asking to do it in this PR.
2) the presence of the two files ``requirements`` and ``environment``, as opposed to a single file, is killing me. They essentially carry the same information, don't they ?

tmbgreaves:
I like this too. In response to Fabio's first question -- I don't know about Travis, but Pipelines is easy to reconfigure to look in subdirectories for config files. Indeed, when I thought the pipelines setup was going to require a lot of files, I had it running out of a subdirectory, and only changed back when it condensed down to a small number of files again.

tjb900:
The couple of projects I just looked at do not appear to worry about this, no (numpy, dask/distributed).

I think `environment.yml` carries _somewhat_ different information than `requirements.txt`, since conda is capable of installing packages which are not python modules - e.g. python itself, and binary dependencies such as MKL or graphics libraries needed by matplotlib.

For example, `requirements.txt` might list matplotlib, but `pip install matplotlib` may not work on systems where `conda install matplotlib` does.  I think?

That probably doesn't help us reason our way out of this mess though.

navjotk:
I think there is a subtle semantic difference between `environment.yml` and `requirements.txt`. `requirements.txt` lists the mandatory dependencies *of* devito so you need those packages in order to get devito running in the first place. But `environment.yml` should be a superset of that, listing some optional dependencies (for lack of a better word), which are required if you're developing with devito. These can include things required for testing. I think we can use devito without `pytest`, for example, but if you're doing serious development with/for devito, you should have `pytest` involves so that's a candidate for `environment.yml` instead of `requirements.txt`. Another such package is `dask`.

FabioLuporini:
OK, this should be ultimately merged, but there are three things that are probably missing

1) we have to mirror some of the travis changes in the pipelines configuration
2) how we do CI has grown in complexity over time, perhaps we should document the key points in a suitable README, (in which, btw, @tmbgreaves , we could also write the precise instructions to reproduce a pipelines failure locally). Such document should also explain the protocol in case of dependency upgrade. I'm talking about a few lines document, nothing really formal
3) the Devito root directory has become a complete mess, with lots of files belonging to CI. We should clean it up and move as much stuff as possible to a separate ``ci`` folder

To be fair, for this PR, probably the only compulsory changes are 1); 3) are desired; 2) may be left for the future. 

I don't know if and who should do these changes. I could do them, but it would have to wait for less busy times. Not an issue I guess, as we're not in a rush ?

I'd also like some input from @ggorman

FabioLuporini:
> I think there is a subtle semantic difference between `environment.yml` and `requirements.txt`. `requirements.txt` lists the mandatory dependencies _of_ devito so you need those packages in order to get devito running in the first place. But `environment.yml` should be a superset of that, listing some optional dependencies (for lack of a better word), which are required if you're developing with devito. These can include things required for testing. I think we can use devito without `pytest`, for example, but if you're doing serious development with/for devito, you should have `pytest` installed. So that's a candidate for `environment.yml` instead of `requirements.txt`. Another such package is `dask`.

very good points. These should be added to the README document I'm envisioning above

tmbgreaves:
@FabioLuporini let me know when/where you need input on a method for reproducing failures and I'll add it - sounds good to me.

FabioLuporini:
I still think this was a good idea , but it's a pity nobody has the time to clean up and finish it off ( @ggorman )

tmbgreaves:
@ggorman @FabioLuporini I'm heavily tied up with postgrad course infrastructure for probably the next month or so but feel free to stick any relevant bits of this on my task list if they need doing in November/December.

FabioLuporini:
CLosing this because we're gonna overhaul CI anyway soon. Also, two issues remain
- new users installing devito may bump into bugs as they use new package versions that are not tested by CI
- it's risky we wait too much between two upgrades and then we are overwhelmed by compatibility bugs from multiple packages

PR #754, opened 2019-01-22
MPI: add new communication scheme

FabioLuporini:
with DEVITO_MPI=diag one can now switch to a different MPI communication scheme, in which send/recv to/from diagonal neighbours are performed explicitly. 

Still no computation/communication overlap

~depends on #750  , hence the big code diff ATM~


devito/ir/iet/analysis.py
```diff
@@ -122,16 +128,15 @@ def mark_parallel(analysis):
...
-          are unit-strided.
-        """
+def mark_iteration_vectorizable(analysis):
+    """
+    Update the ``analysis`` detecting the VECTOR Iterations within ``analysis.iet``.
+    """
     for tree in analysis.trees:
+        # An Iteration is VECTOR iff:
+        # * it's the innermost in an Iteration tree, AND
+        # * it's got at least an outer PARALLEL Iteration, AND
```
mloubout: Is this really true?

devito/ir/iet/visitors.py
```diff
@@ -372,22 +373,74 @@ def visit_Call(self, o, ret=None, queue=None):
...
     """
 
-    def visit_Call(self, o, ret=None, queue=None):
+    def __init__(self, parent_type=None, child_types=None, mode=None, groupby=False):
+        super(MapNodes, self).__init__()
+        self.parent_type = parent_type or Iteration
+        self.child_types = as_tuple(child_types) or (Call, Expression)
+        assert mode is None or mode == 'immediate'
+        self.mode = mode
+        self.groupby = groupby
```
mloubout: private `_` and properties?

devito/ir/support/basic.py
```diff
@@ -750,14 +761,14 @@ def __init__(self, exprs):
             # reads
             for j in retrieve_terminals(e.rhs):
                 v = self.reads.setdefault(j.function, [])
-                mode = 'R' if not q_inc(e) else 'RI'
+                mode = 'RI' if e.is_Increment and j.function is e.lhs.function else 'R'
```
mloubout: why did change to `not is_inq` to `is_inq`?

devito/__init__.py
```diff
@@ -48,9 +48,9 @@ def _reinit_compiler(val):  # noqa
     # Force re-build the compiler
     configuration['compiler'].__init__(suffix=configuration['compiler'].suffix,
                                        mpi=configuration['mpi'])
-    return bool(val)
+    return bool(val) if isinstance(val, int) else val
 configuration.add('openmp', 0, [0, 1], callback=_reinit_compiler)  # noqa
-configuration.add('mpi', 0, [0, 1], callback=_reinit_compiler)
+configuration.add('mpi', 0, [0, 1, 'basic', 'diag'], callback=_reinit_compiler)
```
mloubout: what are these new two? Is it documented?

devito/ir/iet/analysis.py
FabioLuporini: wdym ?

surely a fully parallel loop is also a vectorizable one. Above is just an heuristic to avoid marking all loops in a nest as VECTOR so that the DLE will only find (at most) one per nest

devito/ir/iet/visitors.py
FabioLuporini: there is no public and no private in the IET visitors. We could add them though...

devito/ir/support/basic.py
FabioLuporini: oh, I also swapped 'R' and 'RI' 

If I remember correctly, just to stay within the 90chars limit

devito/__init__.py
FabioLuporini: undocumented. Basically, currently, when you set it to 1 you get 'basic'. This is heavily work in progress and likely to be changed drastically as soon as computation/communication overlap lands

devito/ir/iet/analysis.py
FabioLuporini: ah! I think it's sufficient, not necessary. 
However, the `iff` refers to marking the Iteration as a VECTOR one, not to whether it's theoretically vectorizable or not (hence the property name VECTOR, as opposed to VECTORIZABLE)

FabioLuporini:
@tjb900 : since this is **not** altering the "basic" (dumb) MPI code generation scheme, I'm landing it. Will think about performance evaluation when we finally have a further MPI scheme supporting computation/communication overlap

PR #755, opened 2019-01-24
rewriting commentaries

btrb:


devito/ir/support/basic.py
```diff
@@ -450,14 +450,14 @@ class TimedAccess(Access):
...
-           | z-3 |           | z+1 |           |  y  |          | z |          |  z  |
+           | x+1 |           |  x  |          |  x  |          | w |          | x+y |
+    obj1 = | y+2 | ,  obj2 = |  4  | , obj3 = |  x  | , obj4 = | y | , obj5 = |  y  |
+           | z-3 |           | z+1 |          |  y  |          | z |          |  z  |
 
     We have that: ::
 
         * obj1 and obj2 are regular;
-        * obj3 is irregular because an findex, ``x``, appears outside of its index
+        * obj3 is irregular because a findex, ``x``, appears outside of its index
```
FabioLuporini: I think "an" was right. The 'n' goes with the sound of the following letter, not the type of the letter itself. `findex` is pronounced `ef-index`, and `e` clearly wants an `n`

PR #759, opened 2019-01-30
Ensure DEVITO_MPI is set during testing

tmbgreaves:


FabioLuporini:
merged, thanks a lot @tmbgreaves

PR #762, opened 2019-02-05
MPI prerequisites: DLE overhaul

### LINKED ISSUE #477: More generic creation of elemental functions
This is more of a wishlist item that is meant to serve as a reminder to clean up some of the internals. The main point is that operator creation and the construction of elemental functions share a whole lot of functionalities, including things like `insert_declarations` and the definition of input parameters (function arguments). Yet, `insert_declarations` (which defines `LocalExpression`s) is performed post-hoc, after elemental functions and their arguments have been built, meaning that the list of variables defined internally by the function is not available during the construction of the elem elemental function. This makes it impossible to determine input parameters generically from the elemental function body (IET sub-tree). 

If this were fixed, `Operator` creation and elemental functions could share more infrastructure, which would also help towards #465.

FabioLuporini:
To ultimately do computation/computation overlap, there's the need to be able to identify and take the loop nests sweeping over the grid and promote them to separate routines, to be called before and after the MPI_Wait. Pseudocode follows

```
MPI_start_async_halo_exchange()
f0(X)
MPI_wait_async_halo_exchange()
f0(Y)
```

This PR brings in the machinery to generate ``f0``. The ``_create_elemental_functions`` routine already available in the DLE was unfortunately unusable for this task, as it was quite messy. Coordination with loop blocking was another issue. So I had to do some serious seasonal cleanup, rewriting ancient code and generalising it the way I needed. This had two desirable effects

1) substantial improvements to the codebase, with lots of obsolete code dropped or rewritten (pls, look at the diff)
2) better code generation. I encourage everyone to run our seismic examples and look at how the generated code differs from master's. The loop blocking code is now nicely promoted to separate routines, which drastically improves readability.

The next PR will build upon this one to finally ship computation/communication overlap in MPI

For this PR, I'd like:

* serious reviewing; everyone
* performance regression; @mloubout and @tjb900 you're my candidates. Take your time.

Also:

fixes #477


FabioLuporini:
EDIT: this PR also ships the DEVITO_JIT_BACKDOOR mode. Which can be really handy.

FabioLuporini:
And finally, one last thing: I've dropped the ``dle/backends`` subdirectory and moved its content into ``dle``.

devito/dle/rewriters.py
```diff
@@ -0,0 +1,535 @@
+import abc
```
FabioLuporini: in this file, only two methods have changes:

* _optimize_halospots
* _loop_blocking

You can safely ignore the rest

devito/dle/parallelizer.py
```diff
@@ -22,7 +22,7 @@ def ncores():
 class NThreads(Constant):
 
     def __new__(cls, **kwargs):
-        return super(NThreads, cls).__new__(cls, name='nthreads', dtype=np.int32,
+        return super(NThreads, cls).__new__(cls, name=kwargs['name'], dtype=np.int32,
```
mloubout: name is always in kwargs?

devito/compiler.py
```diff
@@ -366,9 +367,22 @@ def jit_compile(soname, code, compiler):
...
+        cache_dir.mkdir(parents=True, exist_ok=True)
+    else:
+        # Warning: dropping `code` on the floor in favor to whatever is written
+        # within `src_file`
+        try:
+            with open(src_file, 'r') as f:
+                code = f.read()
+            # Bypass the devito JIT cache
+            cache_dir = mkdtemp()
+        except FileNotFoundError:
```
mloubout: Fall back to deefault jit instead?

devito/__init__.py
```diff
@@ -43,6 +43,17 @@
...
+# Devito generate code for an Operator; then, once the session is over, the
+# generated file is manually modified (e.g., for debugging or for performance
+# experimentation); finally, when re-running the same program, Devito won't
+# overwrite the user-modified files (thus entirely bypassing code generation),
+# and will instead use the custom kernel
+configuration.add('jit-backdoor', 0, [0, 1], lambda i: bool(i), False)
+
+# (Undocumented) escape hatch for cross-compilation
+configuration.add('cross-compile', None)
+
```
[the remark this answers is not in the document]
FabioLuporini: cross-compile was there before

are you rather talking about jit-backdoor? point is, we're not really skipping jit

devito/compiler.py
FabioLuporini: no, I prefer to error out. The user tried to bypass the jitter, but something went south, and this should be notified immediately so that one can figure out what's going on -- running the other code path could actually create confusion !

devito/dle/parallelizer.py
FabioLuporini: yes, it has to be, and this is enforcing that

devito/__init__.py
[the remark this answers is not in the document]
FabioLuporini: I'll see if I can come up with a better name

FabioLuporini:
This branch also needs some performance investigation to be sure there are no regressions when using blocking

FabioLuporini:
Some simple performance evaluation showed that the new code generation is basically as fast as ``master``, even a bit more if we systematically use the openmp ``collapse`` clause. However, before enforcing that, I'll need to run more experiments. But this is a separate matter. I'm proceeding with merging this, as it's already been open for too long

PR #765, opened 2019-02-11
MPI: computation/communication overlap

FabioLuporini:
Not ready for review. 

I'm opening it only to trigger `pipelines`. 

The huge diff is mainly due to  #762

### DIFF
--- a/devito/cgen_utils.py
+++ b/devito/cgen_utils.py
@@ -159,7 +159,7 @@ class CodePrinter(C99CodePrinter):
         return expr.__str__()
 
     def _print_Byref(self, expr):
-        return "&%s" % expr.name
+        return "&%s" % expr.base
 
     def _print_TrigonometricFunction(self, expr):
         func_name = str(expr.func)
--- a/devito/cgen_utils.py
+++ b/devito/cgen_utils.py
@@ -158,8 +158,8 @@ class CodePrinter(C99CodePrinter):
     def _print_IntDiv(self, expr):
         return expr.__str__()
 
-    def _print_Byref(self, expr):
-        return "&%s" % expr.base
+    _print_Byref = _print_IntDiv
+    _print_IndexedPointer = _print_IntDiv
 
     def _print_TrigonometricFunction(self, expr):
         func_name = str(expr.func)
--- a/devito/compiler.py
+++ b/devito/compiler.py
@@ -4,7 +4,6 @@ from os import environ, path
 from time import time
 from distutils import version
 from subprocess import DEVNULL, CalledProcessError, check_output, check_call
-from tempfile import mkdtemp
 import platform
 import warnings
 
@@ -367,10 +366,10 @@ def jit_compile(soname, code, compiler):
     target = str(get_jit_dir().joinpath(soname))
     src_file = "%s.%s" % (target, compiler.src_ext)
 
+    cache_dir = get_codepy_dir().joinpath(soname[:7])
     if configuration['jit-backdoor'] is False:
         # Typically we end up here
         # Make a suite of cache directories based on the soname
-        cache_dir = get_codepy_dir().joinpath(soname[:7])
         cache_dir.mkdir(parents=True, exist_ok=True)
     else:
         # Warning: dropping `code` on the floor in favor to whatever is written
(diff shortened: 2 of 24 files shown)


devito/dle/blocking_utils.py
```diff
@@ -38,7 +39,7 @@ def fold_blockable_tree(node, exclude_innermost=False):
             if any(not is_foldable(j) for j in pairwise_folds):
                 continue
             # Maybe heuristically exclude innermost Iteration
-            if exclude_innermost is True:
+            if blockinner is False:
```
georgebisbas: Does that describe a block that is inner to s-th else or does it describe the inner of a block? Or s-th else?

georgebisbas: Ok, it is for loop-blocking, refers to verb. It is fine.

devito/cgen_utils.py
```diff
@@ -158,8 +158,8 @@ def _print_ListInitializer(self, expr):
     def _print_IntDiv(self, expr):
         return expr.__str__()
 
-    def _print_Byref(self, expr):
-        return "&%s" % expr.name
+    _print_Byref = _print_IntDiv
```
mloubout: really? Looks odd to me

devito/cgen_utils.py
```diff
@@ -158,8 +158,8 @@ def _print_ListInitializer(self, expr):
...
+    _print_IndexedPointer = _print_IntDiv
```
mloubout: same

devito/compiler.py
```diff
@@ -379,7 +378,10 @@ def jit_compile(soname, code, compiler):
             with open(src_file, 'r') as f:
                 code = f.read()
             # Bypass the devito JIT cache
-            cache_dir = mkdtemp()
+            # Note: can't simply use Python's `mkdtemp()` as, with MPI, different
```
mloubout: But isn't it what you want. I would assume that in some case (like in the could) each mpi rank won't have access to a common tmp dir but each to their own and will each need to compile it.

devito/mpi/halo_scheme.py
```diff
@@ -105,10 +106,52 @@ def fmapper(self):
...
+        is (2, 2), while ``v``'s is (4, 4). The situation is depicted below.
+
+        .. code-block:: python
+
+              xx**----------------**xx     u
+            xxxx****------------****xxxx   v
+
+        Where 'x' represents a HALO point, '*' a OWNED point, and '-' a CORE point.
+        Together, '*' and '-' constitute the DOMAIN, which is guaranteed to have
+        identical size across all Functions.
```
mloubout: Is that true? I would think this is only true if all functions are not subsampled with same Dimensions.

devito/tools/data_structures.py
```diff
@@ -188,3 +188,119 @@ def relations(self):
...
         raise NotImplementedError
+
+
+class DAG(object):
+    """
+    A trivial implementation of a directed acyclic graph (DAG).
+
+    Originally extracted from:
+
+        https://github.com/thieman/py-dag/
```
mloubout: Is this tested?

devito/cgen_utils.py
FabioLuporini: yes because ``__str__()`` is exactly what we get to generate

devito/cgen_utils.py
FabioLuporini: same

devito/compiler.py
FabioLuporini: unfortunately not. Think about what's going on here. The user has _manually_ modified some C file, but of course he's done that only on a single rank. All MPI processes should eventually pick up that specific file. So this actually requires some further coordination in the case of multi-node execution. Single node MPI should be fine. 

Also, with MPI, the executable ran by the MPI processes must be *exactly the same*.

devito/tools/data_structures.py
FabioLuporini: well it's tested [here](https://github.com/thieman/py-dag/blob/master/tests/init_tests.py)

Our implementation is basically a subset of the linked one so...

devito/mpi/halo_scheme.py
FabioLuporini: in fact, it's even more subtle than that, as even w/o subsampling you might have one MPI rank in each Dimension having a slightly smaller CORE region due to the number of processes not dividing evenly the grid size

But what happened here is that the docstring is misleading. What's important is that, for each FUnction, the number of 'x' and '*' is the **same on all MPI ranks**, the '-' don't matter. And this is really true. IOW, the halo/owned regions have same size across different ranks.

I'll rephrase this docstring. Thanks for pointing it out

FabioLuporini: I fixed the docstring

FabioLuporini:
Merged.

PR #766, opened 2019-02-12
Symbolic coefficients tweaks

rhodrin:
Some tweaks still required. Want to see if anything breaks.


devito/finite_differences/differentiable.py
```diff
@@ -54,6 +58,14 @@ def is_Staggered(self):
     def _fd(self):
         return dict(ChainMap(*[getattr(i, '_fd', {}) for i in self._args_diff]))
 
+    @cached_property
+    def _symbolic_functions(self):
+        return [i for i in list(self._functions) if i.coefficients == 'symbolic']
+
+    @cached_property
+    def _symbolic_coefficients(self):
```
FabioLuporini: is this supposed to return the actual Coefficients, in the future?

devito/finite_differences/differentiable.py
```diff
@@ -54,6 +58,14 @@ def is_Staggered(self):
     def _fd(self):
         return dict(ChainMap(*[getattr(i, '_fd', {}) for i in self._args_diff]))
 
+    @cached_property
+    def _symbolic_functions(self):
+        return [i for i in list(self._functions) if i.coefficients == 'symbolic']
```
FabioLuporini: no need to cast ``self._functions`` to a list

FabioLuporini: and perhaps you should return set, otherwise the output of this method will change across different python sessions (since it's relying on self._functions, which is unordered)

devito/equation.py
```diff
@@ -82,6 +81,27 @@ def subdomain(self):
     def substitutions(self):
         return self._substitutions
 
+    @cached_property
+    def _symbolic_coefficients(self):
+        return bool(self._symbolic_functions)
+
+    @cached_property
+    def _symbolic_functions(self):
+        try:
```
FabioLuporini: what test fails without the try-except ?
if it's "only" because sometimes we can have ``Indexed``, instead of ``Function``, then we can easily add ``symbolic_functions`` to our own ``Indexed`` [here](https://github.com/opesci/devito/blob/master/devito/types/basic.py#L974)

devito/equation.py
```diff
@@ -60,12 +61,10 @@ def __new__(cls, *args, **kwargs):
         obj = sympy.Eq.__new__(cls, *args, **kwargs)
         obj._subdomain = subdomain
         obj._substitutions = substitutions
-        functions = retrieve_functions(obj)
-        functions = [f for f in functions if not f.is_SparseFunction]
-        if any(f.coefficients == 'symbolic' for f in functions):
+        if obj._symbolic_coefficients:
```
FabioLuporini: I **really** like this change!

FabioLuporini: I think there's a point in ``finite_difference.py`` where we can do the same?

devito/finite_differences/differentiable.py
rhodrin: I'm not sure (was thinking about this), but probably not since they will generally be stored in the `Eq` object.

devito/finite_differences/differentiable.py
rhodrin: Ok, will change, thanks.

devito/equation.py
rhodrin: Yeah, it's due to `Indexed`. In fact the test currently failing is due to and `Indexed` I believe. Will look into this now.

devito/equation.py
rhodrin: Yes, in the `check_symbolic` decorator. Thanks for reminding me, will update.

devito/equation.py
rhodrin: Oh, it's also due to the fact that sometimes the lhs or rhs can be zero.

devito/types/sparse.py
```diff
@@ -1048,17 +1048,20 @@ def __init__(self, *args, **kwargs):
             gridpoints.data[:] = gridpoints_data[:]
             self._gridpoints = gridpoints
 
-            coefficients = SubFunction(name="%s_coefficients" % self.name,
-                                       dimensions=(self.indices[-1], Dimension(name='d'),
-                                                   Dimension(name='i')),
-                                       shape=(self.npoint, self.grid.dim, self.r),
-                                       dtype=self.dtype, space_order=0, parent=self)
-            coefficients_data = kwargs.get('coefficients', None)
+            interpolation_coeffs = SubFunction(name="%s_interpolation_coeffs" % self.name,
```
FabioLuporini: does this get less ugly with ``interp_coeffs``?

devito/finite_differences/differentiable.py
FabioLuporini: then perhaps a "booleanyc" name would be better such as ``_uses_symbolic_coefficients``? 

or perhaps use ``_symbolic_functions`` directly?

devito/types/sparse.py
rhodrin: For developers yes, but pre computed sparse functions still requires a public `interpolation_coeffs` interface - If you think `interp_coeffs` is fine for a public name I'm happy to change.

devito/finite_differences/differentiable.py
rhodrin: I've changed to `_uses_symbolic_coefficients` for now?

devito/finite_differences/coefficients.py
```diff
@@ -249,11 +249,11 @@ def generate_subs(deriv_order, function, dim):
 
 
 def get_sym(functions):
-    for j in range(0, len(functions)):
+    for f in functions:
         try:
-            sym = functions[j].coeff_symbol
+            sym = f._coeff_symbol
```
FabioLuporini: can drop `sym`

devito/finite_differences/differentiable.py
```diff
@@ -54,6 +58,14 @@ def is_Staggered(self):
     def _fd(self):
         return dict(ChainMap(*[getattr(i, '_fd', {}) for i in self._args_diff]))
 
+    @cached_property
+    def _symbolic_functions(self):
+        return frozenset([i for i in self._functions if i.coefficients == 'symbolic'])
```
FabioLuporini: nitpicking (for future reference): the inner [...] are actually useless

FabioLuporini:
Merged.

PR #770, opened 2019-02-26
Better diagnostic messages

FabioLuporini:


devito/operator.py
```diff
@@ -253,10 +253,28 @@ def _prepare_arguments(self, **kwargs):
...
         args = ReducerMap()
-        args.update([p._arg_values(**kwargs) for p in self.input if p.name in kwargs])
-        args.update([p._arg_values() for p in self.input if p.name not in args])
+        for p in overrides:
+            args.update(p._arg_values(**kwargs))
+            try:
+                args = ReducerMap(args.reduce_all())
+            except ValueError:
+                raise ValueError("Override `%s` is incompatible with overrides `%s`" %
+                                 (p, [i for i in overrides if i.name in args]))
```
mloubout: From my experience (may have changed) the % followed by line break fails to process the string expression

mloubout: But if test pass may have been fixed in python

tests/test_operator.py
```diff
@@ -947,6 +947,43 @@ def test_function_dataobj(self, so, to, pad, expected):
...
+        a0 = Function(name='a', grid=grid0)
+        b0 = Function(name='b', grid=grid0)
+        a1 = Function(name='a', grid=grid1)
+        b1 = Function(name='b', grid=grid1)
+
+        op = Operator(Eq(a0, a0 + b0 + 1))
+        op.apply()
+
+        try:
+            op.apply(a=a1, b=b1)
```
mloubout: running it locally, this runs fine, doesn't end up in the except section

FabioLuporini: as expected, otherwise CI would be red ? not sure I follow?

mloubout: >Check it's hitting the right error msg

This is never actually checked as op.apply runs fine

FabioLuporini: ah, I see. Good spot. DO you know why? Otherwise I'll take a look

mloubout: Both have same grid so still fine, `op.apply(a=a1, b=b0)` should do the trick

FabioLuporini: ah yes! can you patch it?

FabioLuporini: or I will tmr morning no probs

FabioLuporini:
thanks @mloubout , I fixed it. Will merge once CI is green again

PR #771, opened 2019-03-06
Checkpointing update

navjotk:
With the [compression PR](https://github.com/opesci/pyrevolve/pull/24) ready on pyrevolve - this becomes necessary because that PR changes the API between devito and pyrevolve. 

The direction of this change is to reduce the amount of data that is checkpointed and results in non-trivial speedups. 

TODO:
 - ~~Add test to check numpy flags are set correctly on `devito.data` objects.~~


devito/data/data.py
```diff
@@ -41,7 +41,7 @@ class Data(np.ndarray):
     def __new__(cls, shape, dtype, decomposition=None, modulo=None, allocator=ALLOC_FLAT):
         assert len(shape) == len(modulo)
         ndarray, memfree_args = allocator.alloc(shape, dtype)
-        obj = np.asarray(ndarray).view(cls)
+        obj = ndarray.view(cls)
```
FabioLuporini: curiosity: is this just aesthetics or is it changing behaviour ?

devito/data/allocators.py
```diff
@@ -77,7 +77,7 @@ def alloc(self, shape, dtype):
         # cast to 1D array of the specified size
         ctype_1d = ctype * size
         buf = ctypes.cast(c_pointer, ctypes.POINTER(ctype_1d)).contents
-        pointer = np.frombuffer(buf, dtype=dtype).reshape(shape)
+        pointer = np.array(c_pointer, dtype=dtype, copy=False, order='C', subok=True)
```
FabioLuporini: @tjb900 since you had changed these lines, I wonder whether you could take a second look here?

what was the issue before ?

navjotk: Ah I see it now. I worked with numpy v1.14 while developing this. Meanwhile, of course, numpy's API changed and probably that's why @tjb900 made this change. I'll have to give this some more thought.

devito/data/data.py
navjotk: Yes and No. 
The allocation procedure in allocator should have already done the appropriate wrapping-as-numpy-array (hence the name of the variable being `ndarray`) - in which case, `np.asarray` is not required. From this point of view it is just aesthetics. 
However, if the allocator has done something "special" (in this case, setting some flags for numpy), this effectively undoes that and redoes the wrapping. From this point of view it is changing behaviour.

navjotk:
From what I understand, the failure on CI is due to numpy version 1.16 and the changed API. I need to go back into this to figure out what broke compatibility from 1.14 to 1.16 (will explore @tjb900 's change in the above line from earlier).

FabioLuporini:
@navjotk 

I see the compression PR has been merged in pyrevolve, so I wonder what's the situation here?

navjotk:
In the new version, pyrevolve uses pyzfp for compression. This is another package I wrote myself - a cython wrapper over the zfp library. Some issues on how pyrevolve should specify the dependency on pyzfp were only revealed on travis here. `pip install pyzfp` works fine, but `pip install pyrevolve`, where pyrevolve internally specifies a dependency on pyzfp actually follows a completely different codepath in pip. This is why the tests on the pyzfp repo and the pyrevolve repo didn't catch the issue. Still working on fixing it.

navjotk:
This now passes on travis but fails on Azure pipelines. Any ideas on where the difference in behaviour could come from? @tmbgreaves

navjotk:
@tmbgreaves @ggorman can we please sort this out in person next week? I'll be available Monday/Tuesday/Wednesday

tmbgreaves:
Sorry -- I got diverted onto binderhub stuff and haven't managed to fix this yet.  I can't do Monday or Tuesday next week but can do Wednesday, and can get in to IC to meet any time 1100-1500 if that's any good? I'll do my best to find some debugging time to bring something useful then.

navjotk:
Works for me. Thanks so much for this!

navjotk:
@tmbgreaves I managed to narrow it down to a single line. Commenting the line:
```
RUN update-alternatives --install /usr/bin/ar ar /usr/bin/gcc-ar-$gccVersion 10
```
makes everything pass.

tmbgreaves:
I guess it must be a bug in the ubuntu-toolchain-r version of ar, then. It's probably fairly reasonable to stick with the system version of ar, but it sounds like a bug worth tracking down and reporting. I'll have a go at it tonight.

tmbgreaves:
Is it still worth me coming in tomorrow or is this a reasonable enough workaround to go with for now and revert when there's a fixed version of ar?

navjotk:
I'm happy with this if you are.

tmbgreaves:
OK, let's leave it at that for now, and I'll stick on my list to try and make a small test case to report back to the developers for the ar breakage.

navjotk:
This is ready for merge @FabioLuporini

environment.yml
```diff
@@ -25,8 +25,10 @@ dependencies:
   - multidict
   - frozendict
   - nb_conda_kernels
+  - pip
```
FabioLuporini: why is this necessary?

environment.yml
```diff
@@ -25,8 +25,10 @@ dependencies:
...
   - pip:
     - "git+https://github.com/inducer/codepy"
-    - pyrevolve==1.0.2
+    - pyzfp>=0.2
```
FabioLuporini: so compression becomes more and more a first class citizen in devito and I'm happy about that, we should probably just advertise it more

examples/checkpointing/checkpoint.py
```diff
@@ -62,27 +63,31 @@ def __init__(self, objects):
...
+    def save(*args):
+        raise RuntimeError("Invalid method called. Did you check your version" +
+                           " of pyrevolve?")
+
+    def load(*args):
+        raise RuntimeError("Invalid method called. Did you check your version" +
+                           " of pyrevolve?")
+
+
+def get_symbol_data(symbol, timestep):
```
FabioLuporini: I think "symbol" should actually be "function" ?

Dockerfile.pipelines
```diff
@@ -62,7 +62,6 @@ RUN apt-get -y install gcc-$gccVersion g++-$gccVersion
 RUN update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-$gccVersion 10
 RUN update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-$gccVersion 10
 RUN update-alternatives --install /usr/bin/gcov gcov /usr/bin/gcov-$gccVersion 10
-RUN update-alternatives --install /usr/bin/ar ar /usr/bin/gcc-ar-$gccVersion 10
```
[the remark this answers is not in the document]
navjotk: Read discussion with Tim Greaves earlier in the thread

environment.yml
navjotk: Newer versions of Conda raise a warning without this. Unrelated to the rest of the PR.

examples/checkpointing/checkpoint.py
navjotk: Why?

FabioLuporini: convention throughout devito. Functions have data attached (Function, TimeFunction, SparseFunction ...). Symbols are more generic . But I guess here you're looking for {...}Functions?

environment.yml
FabioLuporini: oh, OK, thanks

FabioLuporini:
Merged.

PR #773, opened 2019-03-08
MPI: add FULL mode

FabioLuporini:
like overlap2, but with MPI_test to prod comp/comm overlap

all other changes are misc improvements to the compiler


devito/dle/rewriters.py
```diff
@@ -406,6 +405,29 @@ def _shm_parallelize(self, iet):
         """
         return self._shm_parallelizer.make_parallel(iet)
 
+    @dle_pass
+    def _hoist_prodders(self, iet):
```
btrb: might it be a typo for `host`?

devito/ir/iet/analysis.py
```diff
@@ -258,13 +262,36 @@ def mark_halospot_hoistable(analysis):
...
-            # sections of the scope
-            found = [f for f in hs.fmapper if not analysis.scopes[i].d_anti.project(f)]
+
+            found = []
+            scope = analysis.scopes[i]
+            for f, hse in hs.fmapper.items():
+                # The sufficient condition for `f`'s halo-update to be
+                # `hoistable` is that there are no `hs.dimensions`-induced
+                # flow-dependences touching the halo
+                test = True
```
btrb: wouldn't be safer to set it to `False` and do `try / excepts` below?

devito/ir/iet/analysis.py
```diff
@@ -231,20 +231,24 @@ def mark_halospot_useless(analysis):
...
+        # `hs` is USELESS if ...
+
+        # * ANY of its Dimensions turn out to be SEQUENTIAL
+        if any(SEQUENTIAL in analysis.properties[i]
+               for i in iterations if i.dim.root in hs.dimensions):
+            properties[hs] = USELESS
+            continue
+
+        # * ALL reads pertain to an increment expression
+        test = False
```
btrb: is this permanent?

devito/dle/rewriters.py
FabioLuporini: nope, it's actually "hoist" in the sense of lifting

devito/ir/iet/analysis.py
FabioLuporini: wdym ? 
this is like "start assuming it's false; if not, will find out and set it true"

devito/ir/iet/analysis.py
FabioLuporini: not sure I get what you mean here?

devito/ir/iet/nodes.py
```diff
@@ -771,6 +780,34 @@ def exprs(self):
...
+
+    """
+    A Call promoting asynchronous progress, to minimize latency.
+
+    Example use cases:
+
+        * To trigger asynchronous progress in the case of distributed-memory
+          parallelism.
+        * Software prefetching.
+    """
```
navjotk: Is it possible to expand this docstring a bit? e.g. in what way are these asynchronous calls really different from standard calls - in terms of generated code?

FabioLuporini: indeed they are not -- but Prodders may need to be distinguished from Calls as they may be subject of optimisation (e.g., code motion).

Also -- and this is implicit in the description -- a Prodder, when invoked, is guaranteed not to alter state (which ensures optimizations such as code motion are safe)

FabioLuporini:
Merged.

PR #775, opened 2019-03-11
Clustering support for scalar equations

FabioLuporini:


devito/equation.py
```diff
@@ -54,19 +59,19 @@ class Eq(sympy.Eq):
...
-        substitutions = kwargs.pop('coefficients', None)
-        obj = sympy.Eq.__new__(cls, *args, **kwargs)
+        obj = sympy.Eq.__new__(cls, lhs, rhs, **kwargs)
+        obj._implicit_dims = as_tuple(implicit_dims)
         obj._subdomain = subdomain
-        obj._substitutions = substitutions
+        obj._substitutions = coefficients
         if obj._uses_symbolic_coefficients:
             # NOTE: As Coefficients.py is expanded we will not want
             # all rules to be expunged during this procress.
```
georgebisbas: process?

rhodrin:
I'm looking into why this is atm but scalar equations are still not working correctly. Take for example:

```
import numpy as np
from devito import Grid, Dimension
from devito import Function, TimeFunction, Eq, Constant, Operator, solve

n_domains = 10
n = Dimension(name = 'n')

const_x = Constant(name='const_x', dtype=np.int32)
const_y = Constant(name='const_y', dtype=np.int32)
const_x.data = 5
const_y.data = 3

bounds_xm = Function(name='bounds_xm', shape = (n_domains, ),
                     dimensions = (n, ), dtype = np.int32)
bounds_ym = Function(name='bounds_ym', shape = (n_domains, ),
                     dimensions = (n, ), dtype = np.int32)
bounds_xm.data[:] = 0
bounds_ym.data[:] = 0

eq_xm = Eq(bounds_xm[n], const_x)
eq_ym = Eq(bounds_ym[n], const_y)

op = Operator([eq_xm, eq_ym])
op(n_m = 0, n_M = n_domains-1)
```
This runs fine and `bounds_xm`, `bounds_ym` are set as expected.

However:
```
import numpy as np
from devito import Grid, Dimension
from devito import Function, TimeFunction, Eq, Constant, Operator, solve

n_domains = 10
n = Dimension(name = 'n')

const_x = Constant(name='const_x', dtype=np.int32)
const_y = Constant(name='const_y', dtype=np.int32)

bounds_xm = Function(name='bounds_xm', shape = (n_domains, ),
                     dimensions = (n, ), dtype = np.int32)
bounds_ym = Function(name='bounds_ym', shape = (n_domains, ),
                     dimensions = (n, ), dtype = np.int32)

eq_xm = Eq(const_x, bounds_xm[n])
eq_ym = Eq(const_y, bounds_ym[n])

for j in range(0, n_domains):
    bounds_xm.data[j] = j
    bounds_ym.data[j] = j

op = Operator([eq_xm, eq_ym])

op(n_m = 0, n_M = n_domains-1)
```
does not work as expected. A code structure is created such that the n loop is not executed and `const_x`, `const_y` remain 0.

rhodrin:
Ok nice, the following produces the c loop structure we're looking for:
```
import numpy as np
from devito import Grid, Dimension
from devito import Function, TimeFunction, Eq, Constant, Operator, solve
from devito.types import Scalar

n_domains = 10
n = Dimension(name = 'n')

const_x = Scalar(name='const_x', dtype=np.int32, is_const=False)
const_y = Scalar(name='const_y', dtype=np.int32, is_const=False)
bounds_xm = Function(name='bounds_xm', shape=(n_domains, ), dimensions=(n, ))
bounds_ym = Function(name='bounds_ym', shape=(n_domains, ), dimensions=(n, ))
    
grid = Grid(extent=(10, 10), shape=(10, 10))
t = grid.time_dim
x, y = grid.dimensions

f.data[:] = 0.0
eq = Eq(f.dt, 1)

stencil = Eq(f.forward, solve(eq, f.forward), implicit_dims=[t, n, x, y])

eq_xm = Eq(const_x, bounds_xm[n], implicit_dims=[t, n])
eq_ym = Eq(const_y, bounds_ym[n], implicit_dims=[t, n])

op = Operator([eq_xm, eq_ym, stencil])
```
The remaining part is now to get this working with subdomains. I was seeing some strange results earlier but they may improve with some further tweaking.

devito/equation.py
```diff
@@ -23,14 +24,18 @@ class Eq(sympy.Eq):
...
-        The right-hand side.
+    rhs : expr-like, optional
+        The right-hand side. Defaults to 0.
     subdomain : SubDomain, optional
         To restrict the computation of the Eq to a particular sub-region in the
         computational domain.
     coefficients : Substitutions, optional
         Can be used to replace symbolic finite difference weights with user
         defined weights.
+    implicit_dims : Dimension or list of Dimension, optional
```
rhodrin: How 'exposed' do we want `implicit_dims` to actually be? If we want it in the docstring I think we should include a few clear examples (which I could add next week)? 
(Currently it takes a little bit of playing around to actually work out what this functionality does).

FabioLuporini: Agreed. In fact, this might not be exposed to the user level at all. 

If you look at the commit history, you see at some point I had introduced a Step class, subclass of Eq, which was basically an Eq with this additional parameter. Step wouldn't have been exposed in the user API. However, I then quickly realised that then I'd have needed a StepInc class as well -- you can see the pattern here. It was horrible. So the shortcut was to do add it directly to Eq/Inc. 

I can't see how ``implicit_dims`` should be helpful to the average user. So I'm happy to drop it from the docstring ("partial hiding"); any other suggestions would be more than welcome

FabioLuporini:
@rhodrin I think we should aim to land this PR first

devito/equation.py
```diff
@@ -104,8 +113,13 @@ def _symbolic_functions(self):
 
     def xreplace(self, rules):
         """"""
-        return self.func(self.lhs.xreplace(rules), self.rhs.xreplace(rules),
-                         subdomain=self._subdomain)
+        return self.func(self.lhs.xreplace(rules), rhs=self.rhs.xreplace(rules),
+                         implicit_dims=self._implicit_dims, subdomain=self._subdomain)
+
+    def __str__(self):
+        return "%s(%s, %s)" % (self.__class__.__name__, self.lhs, self.rhs)
```
mloubout: Why is this necessary?

FabioLuporini: Oh...not strictly. But I dropped the equivalent in Inc (see below), so that any future subclasses are stringified as 
```
Eq(...)
Inc(...)
NewSubClass(...)
```
if you look at the commit history of this PR, you see that at some point I had a new subclass of Eq called Step, then I removed it. But I might reinstate it in the future

rhodrin:
> @rhodrin I think we should aim to land this PR first . Wdyt ?

First as in before what? But in any case we should try and land this asap.

Also, re. how exposed this should be, I agree that not having it exposed to the user (at least at present) is the way forward.

FabioLuporini:
Merged.

PR #777, opened 2019-03-12
Update requirements

navjotk:
This short PR makes the following changes in the requirements/environment.yml setup:
- ~~Add gcc as a dependency in environment.yml so people following the conda procedure get a C compiler without having to explicitly install one. The last two installation issues could have been avoided with this.~~
- Fix the pyrevolve dependency to a specific version. This is so that we can issue a new release of pyrevolve for the new API - which would enable #771 to be tested on travis without breaking compatibility for people until it is merged.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -26,7 +26,6 @@ dependencies:
   - pytest-cov
   - multidict
   - frozendict
-  - gcc
   - pip:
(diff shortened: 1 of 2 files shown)


FabioLuporini:
Merged.

PR #779, opened 2019-03-12
Minor fixes and speedup

mloubout:
To small main thing:

- fix the `__new__` in sources that was ignoring the cache and therefore failing with time derivative for these sources

- reduces the number of test in test_adjoint to speed up Travis and spread the different space orders over these OT 4 test so that we still test all of it.

### DIFF
--- a/examples/seismic/source.py
+++ b/examples/seismic/source.py
@@ -1,3 +1,5 @@
+import sympy
+
 from scipy import interpolate
 from cached_property import cached_property
 import numpy as np
@@ -7,7 +9,7 @@ except:
     plt = None
 
 from devito.types import Dimension, SparseTimeFunction
-
+from devito.types.basic import _SymbolCache
 
 __all__ = ['PointSource', 'Receiver', 'Shot', 'WaveletSource',
            'RickerSource', 'GaborSource', 'DGaussSource', 'TimeAxis']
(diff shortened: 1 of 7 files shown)


tests/test_adjoint.py
```diff
@@ -28,17 +28,16 @@ def setup_method(self, method):
...
-        ('layers', (60, 70), 'OT2', 12, 10), ('layers', (60, 70), 'OT4', 4, 10),
-        ('layers', (60, 70), 'OT4', 8, 10), ('layers', (60, 70), 'OT4', 12, 10),
+        ('layers', (60, 70), 'OT2', 12, 10), ('layers', (60, 70), 'OT4', 8, 10),
         # 3D tests with varying time and space orders
         ('layers', (60, 70, 80), 'OT2', 4, 10), ('layers', (60, 70, 80), 'OT2', 8, 10),
         ('layers', (60, 70, 80), 'OT2', 12, 10), ('layers', (60, 70, 80), 'OT4', 4, 10),
-        ('layers', (60, 70, 80), 'OT4', 8, 10), ('layers', (60, 70, 80), 'OT4', 12, 10),
         # Constant model in 2D and 3D
         ('constant', (60, 70), 'OT2', 8, 14), ('constant', (60, 70, 80), 'OT2', 8, 14),
+        ('constant', (60, 70), 'OT4', 12, 14), ('constant', (60, 70, 80), 'OT4', 4, 14),
```
mloubout: Note for reviewers:

So yes I removed some but if you look at it I changed the space orders too so that it's all still tested but should remove quite some time from CI as OT4 is like 90% of this test

examples/seismic/source.py
```diff
@@ -103,34 +105,39 @@ class PointSource(SparseTimeFunction):
...
-
-        obj._time_range = time_range._rebuild()
-
-        # If provided, copy initial data into the allocated buffer
-        data = kwargs.get('data')
-        if data is not None:
-            obj.data[:] = data
+    def __new__(cls, *args, **kwargs):
+        options = kwargs.get('options', {})
+        if cls in _SymbolCache:
```
FabioLuporini: probably no need for _SymbolCache. I **think** the (right) way to achieve the same thing is more like [this](https://github.com/opesci/devito/blob/master/devito/types/basic.py#L472)

FabioLuporini: to be more precise, I think we should use __init__, not __new__. This __new__ overriding looks hacky to me

mloubout: No it is needed, this is not `__init__` this `_new_` so it doesn't check the one in `basic`, doesn't work with just init

mloubout: W can't use init, these one have specific inputs and properties that disapear at any manipulation (subs, .dt,..) without the proper new

tests/test_dse.py
```diff
@@ -391,3 +391,23 @@ def test_makeit_ssa(exprs, exp_u, exp_v):
...
+    v = TimeFunction(name='v', grid=grid, time_order=2, space_order=0, save=3)
+
+    # The second equation needs a full loop over x/y for u then
+    # a full one over x.y for v
+    eq = [Eq(u.forward, 2 + grid.time_dim),
+          Eq(v.forward, u.forward.dx + u.forward.dy + 1)]
+    op = Operator(eq, dse='advanced', dle='basic')
+
+    trees = retrieve_iteration_tree(op)
+    assert len(trees) == 2
```
mloubout: @fablup so mainly this test replace the tti one that checks that the loop is split for u and v (u needs to be updated everywhere first then v)
Unfortunately, it looks like the only reason the other test was passing is that the src term was between the two forcing it to be split. This one fails :( I'm looking into it.

examples/seismic/source.py
FabioLuporini: I'm not sure I understand, perhaps I'll try this to see what's actually going on. And we can close an eye for this small PR, not an issue. But this is definitely a type of class that an user could want to write in application code, and we can't expect him/her to be able to know and deal with low-level things such as SymbolCache

tests/test_tti.py
```diff
@@ -87,41 +87,3 @@ def test_tti(shape, space_order, kernel):
 
     log("Difference between acoustic and TTI with all coefficients to 0 %2.4e" % res)
     assert np.isclose(res, 0.0, atol=1e-4)
-
-
-@pytest.mark.parametrize('shape', [(50, 60), (50, 60, 70)])
-def test_tti_staggered(shape):
```
FabioLuporini: why is this being completely dropped ?

tests/test_dse.py
FabioLuporini: I tried it. I think this is just yet another example of "machines beating humans". 
Look at the iteration direction of x and y: they actually go backwards to turn the antidependence from the first to the second equation (IOW, the second equation needing information computed by the first one). So we shouldn't worry about this?

FabioLuporini: if you use ``space_order=2`` in the definition of ``u``, then you **also** get a flow dependence, which will force the compiler to split the loops

examples/seismic/source.py
mloubout: That's a good point

tests/test_dse.py
mloubout: Wow indeed nice catch, that's quite amazing :D

tests/test_tti.py
mloubout: This test was oiriginally there to check the loop are split (which the new test does).
This is like suuuuper expensive to run 3D tti with zero compiler optimization so removed it now that the one thing tested here is tested in test_dse

mloubout:
So speed up is quite nice, except the gcc 4.9 build that seem to always take for ever

FabioLuporini:
but the codecov cary is scary. Is it up to date?

mloubout:
>but the codecov report is scary. Is it up to date?
It's not no it's from first commit

examples/seismic/source.py
navjotk: Just like to add that if we're using `_SymbolCache` outside the devito package, it shouldn't be private(marked with an underscore)

FabioLuporini:
Merged.

examples/seismic/source.py
FabioLuporini: indeed we shouldn't

PR #783, opened 2019-03-20
Performance optimisation improvements

### LINKED ISSUE #143: Enforce thread pinning through `Compiler` class
In order to ensure reproducible performance evaluation on NUMA systems we should enforce thread pinning in our `Compiler` classes for the most common tool chains (GNU and KMP). The challenge I see though is that simply "unrolling" the processor IDs as `[0, .., N]` might not be portable, since processor ID numbering conventions can vary between CPU generations. More generic settings like `compact` or `scatter` (KMP) might need to be explored for and compared to the "hand-unrolled" settings we use at the moment.

FabioLuporini:
This PR adds:

* A first version of nested shared-memory parallelism, still based on OpenMP
* The capability of autotuning in presence of MPI (the MPI ranks tune themselves autonomously in preemptive/destructive mode)

Also it fixes/tweaks:

* Autotuning when there are multiple blocked loop nests in the same operator (now it doesn't tune over the cartesian product of all possible block shapes, but rather it tunes each loop nest individually)
* The use of OpenMP collapsing in some nasty cases

Finally, it ships (as usual) a number of small tweaks throughout the codebase

PR equipped with new tests oc

fixes #143

### DIFF
--- a/devito/dle/parallelizer.py
+++ b/devito/dle/parallelizer.py
@@ -6,13 +6,15 @@ import cgen as c
 import psutil
 from sympy import Function, Or
 
-from devito.ir.iet import (Call, Conditional, Block, Expression, List, Prodder,
-                           FindSymbols, FindNodes, Return, COLLAPSED, Transformer,
-                           IsPerfectIteration, retrieve_iteration_tree, filter_iterations)
+from devito.ir import (DummyEq, Call, Conditional, Block, Expression, List, Prodder,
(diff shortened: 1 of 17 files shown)


devito/dle/parallelizer.py
```diff
@@ -62,12 +79,22 @@ def __init__(self, prodder):
 
 class Ompizer(object):
```
georgebisbas: OMPizer?

FabioLuporini:
Merged.

PR #786, opened 2019-03-22
Extend subdomains2

rhodrin:
Add functionality to easily specify a set of N subdomains. 

Work in progress but comments welcome. Still need to re-base, fix some things and add more tests etc. but the general structure is there.


devito/operator.py
```diff
@@ -188,6 +192,36 @@ def __init__(self, expressions, **kwargs):
...
+                    dims.append(subdomain._implicit_dimension)
+                    implicit_expressions = []
+                    for i in dat:
+                        eq = Eq(i['rhs'], i['lhs'], implicit_dims=dims)
+                        implicit_expressions.append(eq)
+                    for ie in implicit_expressions:
+                        updated_expressions.append(ie)
+                    for d in subdomain.dimensions:
+                        dims.append(d)
+                    e._implicit_dims = as_tuple(dims)
```
FabioLuporini: Changing state is evil; we should reconstruct instead

devito/operator.py
```diff
@@ -188,6 +192,36 @@ def __init__(self, expressions, **kwargs):
...
+                except AttributeError:
+                    dat = None
+                if bool(dat):
+                    dims = [d.root for d in e.free_symbols if isinstance(d, Dimension)]
+                    sub_dims = [d.root for d in subdomain.dimensions]
+                    dims = list(set(dims).symmetric_difference(set(sub_dims)))
+                    dims.append(subdomain._implicit_dimension)
+                    implicit_expressions = []
+                    for i in dat:
+                        eq = Eq(i['rhs'], i['lhs'], implicit_dims=dims)
```
FabioLuporini: I'm not sure I understand the bracket syntax? where does it come from ?

devito/operator.py
```diff
@@ -188,6 +192,36 @@ def __init__(self, expressions, **kwargs):
...
+                        implicit_expressions.append(eq)
+                    for ie in implicit_expressions:
+                        updated_expressions.append(ie)
+                    for d in subdomain.dimensions:
```
FabioLuporini: all these three loops can be made more pythonic with list comprehension

FabioLuporini: btw, I'm confused, isn't ``dims`` already updated a few lines above?

devito/operator.py
```diff
@@ -188,6 +192,36 @@ def __init__(self, expressions, **kwargs):
...
 
+    def _add_implicit(self, expressions):
+        """
+        Create and add to the expression tree any associated implicit
+        expressions.
+        """
+        updated_expressions = []
+        for e in expressions:
+            subdomain = e.subdomain
+            if bool(subdomain):
```
FabioLuporini: just 
```
if e.subdomain:
```

casting to bool is likely to be unnecessary

devito/operator.py
```diff
@@ -188,6 +192,36 @@ def __init__(self, expressions, **kwargs):
...
+                try:
+                    dat = e._subdomain._ie_dat
+                except AttributeError:
+                    dat = None
+                if bool(dat):
```
FabioLuporini: no need for this ``if`` ?

I think this piece of code becomes more natural if you take the current if's body and move it inside the ``try``. Then, in the ``except``, you just put ``pass``

devito/operator.py
```diff
@@ -188,6 +192,36 @@ def __init__(self, expressions, **kwargs):
 
     # Compilation
 
+    def _add_implicit(self, expressions):
+        """
+        Create and add to the expression tree any associated implicit
```
FabioLuporini: nitpicking: it's not an "expression tree" yet, it's just an orderer sequence of expressions.

devito/operator.py
```diff
@@ -154,6 +155,9 @@ def __init__(self, expressions, **kwargs):
         # autotuning reports, etc
         self._state = {}
 
+        # Form and gather any required implicit expressions
```
FabioLuporini: I think it's worth spending a few words about what an implicit expression is and, in particular, why does this take place *before* the _expression lowering_ pass

devito/operator.py
```diff
@@ -188,6 +192,36 @@ def __init__(self, expressions, **kwargs):
 
     # Compilation
 
+    def _add_implicit(self, expressions):
+        """
+        Create and add to the expression tree any associated implicit
+        expressions.
+        """
+        updated_expressions = []
```
FabioLuporini: nitpicking (for code homogeneity): throughout the codebase, I use the word "processed", or sometimes "ret", to collect the items that will be returned by the function/method

devito/operator.py
```diff
@@ -188,6 +192,36 @@ def __init__(self, expressions, **kwargs):
...
+            subdomain = e.subdomain
+            if bool(subdomain):
+                try:
+                    dat = e._subdomain._ie_dat
+                except AttributeError:
+                    dat = None
+                if bool(dat):
+                    dims = [d.root for d in e.free_symbols if isinstance(d, Dimension)]
+                    sub_dims = [d.root for d in subdomain.dimensions]
+                    dims = list(set(dims).symmetric_difference(set(sub_dims)))
```
FabioLuporini: maybe you can use ``filter_ordered`` ? (``git grep filter_ordered`` for example usage) 

``filter_ordered`` will retain the ordering, while this way you're losing it due to casting to set (don't know if it's a potential issue or not)

devito/operator.py
```diff
@@ -188,6 +192,36 @@ def __init__(self, expressions, **kwargs):
...
+                    dims.append(subdomain._implicit_dimension)
+                    implicit_expressions = []
+                    for i in dat:
+                        eq = Eq(i['rhs'], i['lhs'], implicit_dims=dims)
+                        implicit_expressions.append(eq)
+                    for ie in implicit_expressions:
```
FabioLuporini: hang on: why do you need this extra loop? can you not directly populate ``updated_expressions`` in the previous loop, instead of having an additional list such as ``implicit_expressions`` ?

devito/types/grid.py
```diff
@@ -421,3 +425,63 @@ class Interior(SubDomain):
...
+    **kwargs
+        * N : int
+            Number of subdomains.
+        * bounds : tuple
+            Tuple of numpy int32 arrays representing the bounds of
+            each subdomain.
+
+    Examples
+    --------
+    Include full example + example of shorthand notation here.
```
FabioLuporini: remainder

devito/types/grid.py
```diff
@@ -421,3 +425,63 @@ class Interior(SubDomain):
 
     def define(self, dimensions):
         return {d: ('middle', 1, 1) for d in dimensions}
+
+
+class SubDomains(SubDomain):
```
FabioLuporini: I think we need a different name, as this one is too similar to ``SubDomain``. 

off top of my head: ``SubDomainBundle`` ? Really don't know, we can discuss this offline if you want

devito/types/grid.py
```diff
@@ -421,3 +425,63 @@ class Interior(SubDomain):
...
+    @property
+    def n_domains(self):
+        return self._n_domains
+
+    @property
+    def bounds(self):
+        return self._bounds
+
+    @cached_property
+    def _ie_dat(self):
```
FabioLuporini: better name? ``_ie_dat`` does convey much information

devito/types/grid.py
```diff
@@ -421,3 +425,63 @@ class Interior(SubDomain):
...
+        if not len(self._bounds) == 2*len(self.dimensions):
+            raise ValueError("Left and right bounds must be supplied for each dimension")
+        n_domains = self.n_domains
+        i_dim = self._implicit_dimension
+        b_f = {}
+        mMstring = ['_m', '_M']
+        dat = []
+        for j in range(0, len(self._bounds)):
```
FabioLuporini: A short comment about what's going on here won't harm

devito/types/grid.py
```diff
@@ -421,3 +425,63 @@ class Interior(SubDomain):
...
+        return self._bounds
+
+    @cached_property
+    def _ie_dat(self):
+        if not len(self._bounds) == 2*len(self.dimensions):
+            raise ValueError("Left and right bounds must be supplied for each dimension")
+        n_domains = self.n_domains
+        i_dim = self._implicit_dimension
+        b_f = {}
+        mMstring = ['_m', '_M']
```
FabioLuporini: you can/should get these strings via 
 https://github.com/opesci/devito/blob/master/devito/types/dimension.py#L132

so maybe fname below is just ``d.min_name`` ? (don't know if I'm interpreting the code correctly)

tests/test_subdomains.py
```diff
@@ -0,0 +1,60 @@
...
+
+pytestmark = skipif(['yask', 'ops'])
+
+
+class TestSubdomains(object):
+    """
+    Class for testing SubDomains
+    """
+
+    def test_iterate_NDomains(self):
```
FabioLuporini: should we add more tests or do you think this one suffices?

devito/operator.py
rhodrin: Yeah, was thinking that.

devito/operator.py
rhodrin: From `dat` being a `dict`.

devito/operator.py
rhodrin: `dims` is different for the user specified equation and the implicit equations.

devito/operator.py
rhodrin: Ok, thanks.

devito/operator.py
rhodrin: Sure, sounds good.

devito/operator.py
rhodrin: Hmm, not sure I like 'processed' here since that's not what's happening. But will have a think.

devito/operator.py
rhodrin: Ok thanks, will check.

devito/operator.py
rhodrin: Indeed. I think I needed it before for some reason but after some changes it's no longer needed.

devito/types/grid.py
rhodrin: Reminder? Or is this some comment about Brexit?

devito/types/grid.py
rhodrin: Yeah, was going to re-consider this. I think `SubDomains` is a good name but (as you say) is too similar.

devito/types/grid.py
rhodrin: Sure.

devito/types/grid.py
rhodrin: Ta, will check.

tests/test_subdomains.py
rhodrin: No, we need a couple more.

devito/types/grid.py
FabioLuporini: :joy:  reminder :)

devito/operator.py
rhodrin: `filter_ordered` wont work here (unless I'm missing some mode it has) since we want the non-overlapping intersection of the two sets.

Order is not important (for the time being at least).

rhodrin:
@btrb I think your suggestions should be addressed in the latest commit?

rhodrin:
Still need to rebase.

devito/types/grid.py
```diff
@@ -421,3 +425,125 @@ class Interior(SubDomain):
...
+        n_domains = self.n_domains
+        i_dim = self._implicit_dimension
+        b_f = {}
+        dat = []
+        # Organise the data contained in 'bounds' into a form such that the
+        # associated implicit equations can easily be created.
+        for j in range(0, len(self._bounds)):
+            index = floor(j/2)
+            d = self.dimensions[index]
+            if j % 2 == 0:
```
FabioLuporini: I have a feeling this code may benefit from rewriting the loop as:

```
for i, fname in enumerate(self._bounds):
     ...
```

because then you can drop `index = ...`, while you can get the `fname` by adding a private property `fnames` (feel free to pick a better name) returning `{self._bounds[0]: d.min_name, self._bounds[1]: d.max_name}`, which ultimately will allow you to drop the `if ...` , while making it easier to understand what we are trying to achieve

devito/types/grid.py
```diff
@@ -421,3 +425,125 @@ class Interior(SubDomain):
...
+                fname = d.min_name
+            else:
+                fname = d.max_name
+            b_f[fname] = Function(name=fname, shape=(n_domains, ),
+                                  dimensions=(i_dim, ), dtype=np.int32)
+            if isinstance(self._bounds[j], int):
```
FabioLuporini: comment for this if?

devito/types/grid.py
```diff
@@ -421,3 +425,125 @@ class Interior(SubDomain):
...
+                bounds = np.full((n_domains,), self._bounds[j], dtype=np.int32)
+                b_f[fname].data[:] = bounds
+            else:
+                b_f[fname].data[:] = self._bounds[j]
+            dat.append({'rhs': d.thickness[j % 2][0], 'lhs': b_f[fname][i_dim]})
```
FabioLuporini: I suppose you want a ``namedtuple`` here, not a ``dict``

FabioLuporini: Or you could alternatively return an `Eq`, and then in `operator.py` you could rebuild it (`eq.func(*eq.args, ...)`) to append subdomain and implicit_dims

devito/operator.py
```diff
@@ -188,6 +196,27 @@ def __init__(self, expressions, **kwargs):
...
+                    dims = [d.root for d in e.free_symbols if isinstance(d, Dimension)]
+                    sub_dims = [d.root for d in e.subdomain.dimensions]
+                    dims = list(set(dims).symmetric_difference(set(sub_dims)))
+                    dims.append(e.subdomain._implicit_dimension)
+                    implicit_expressions = [Eq(i['rhs'], i['lhs'], implicit_dims=dims)
+                                            for i in dat]
+                    processed.extend(implicit_expressions)
+                    dims.extend(e.subdomain.dimensions)
+                    e = Eq(e.lhs, e.rhs, subdomain=e.subdomain, implicit_dims=dims)
+                except AttributeError:
```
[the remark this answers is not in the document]
rhodrin: When we have a `SubDomain` and not a `SubDomainSet`.

devito/operator.py
```diff
@@ -188,6 +196,27 @@ def __init__(self, expressions, **kwargs):
...
+                    pass
+            processed.append(e)
+        return list(dict.fromkeys(processed))
```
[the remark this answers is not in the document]
rhodrin: We need to remove repeat instances of the same implicit equations.

devito/types/grid.py
rhodrin: I'm not quite sure what you mean by the above. Where are you suggesting that `fnames` is coming from?

devito/operator.py
```diff
@@ -155,6 +156,13 @@ def __init__(self, expressions, **kwargs):
         # autotuning reports, etc
         self._state = {}
 
+        # Form and gather any required implicit expressions.
+        # Implicit expressions are those not explicitly defined by the user
+        # but instead are requisites of some specified functionality. Since they
+        # must be treated in a similar manner to user specified expressions
+        # they are added prior to the creation of the IET.
```
FabioLuporini: "prior to the creation of the IET". In fact the IET is created much, much later. So I'd say "prior to expression lowering"

tests/test_subdomains.py
```diff
@@ -0,0 +1,114 @@
+import numpy as np
+from math import floor
+
+from conftest import skipif
+from devito import Grid, Function, TimeFunction, Eq, solve, Operator, SubDomainSet
+
+pytestmark = skipif(['yask', 'ops'])
+
+
+class TestSubdomains(object):
```
FabioLuporini: I suggest a test with two  SubDomainSets in the same Operator ?

devito/operator.py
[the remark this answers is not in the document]
FabioLuporini: ah! sure sorry my bad

devito/operator.py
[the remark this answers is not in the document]
FabioLuporini: so this might break the case in which we have two SubDomainSets which create the same implicit_equation?

if so, to fix it, we have to make sure that the implicit_dimension created within SubDomainSet isn't cached as all Dimensions (as opposed to Functions) do. For example, look [here](https://github.com/opesci/devito/blob/master/tests/test_symbol_caching.py#L208): ``d0`` is actually ``d1``. And SubDomainSet creates a ``Dimension(name='n')``, so it will be eventually shared by multiple SubDomainSets. We have multiple options here again, but the simplest one is to enforce uniqueness by name

```
class SubDomainSet(object):

    _counter = 0
    """docstring here."""

    ...
    Dimension(name='n%d' % SubDomainSet._counter)
    SubDomainSet._counter += 1
```

I'm not a big fan of this approach because now if you execute tests (or Operators, in general) in different order you get different generated code (eg, n0 in one Operator and n1 in the other, and then swapped), but it's definitely the simplest one. 

Thoughts?

devito/types/grid.py
FabioLuporini: sorry, I thought bounds were provided as a tuple of 2-tuples, ie, ``((bounds_xm, bounds_xM), (bounds_ym, bounds_yM))``, then it would have been easier to have a property ``fnames`` returning a dict which associates a name to each of that bounds, but I see now that actually bounds is a flatten tuple, so never mind, no worries !

devito/operator.py
rhodrin: Thanks. Updated.

tests/test_subdomains.py
rhodrin: Good idea. Will add this morning.

rhodrin:
> sorry, I thought bounds were provided as a tuple of 2-tuples, ie, ((bounds_xm, bounds_xM), 
> (bounds_ym, bounds_yM)), then it would have been easier to have a property fnames returning a 
> dict which associates a name to each of that bounds, but I see now that actually bounds is a flatten 
> tuple, so never mind, no worries !

Ah, I see. I did it this way since this is a common notation of plotting tools etc. We could have users provide input in that fashion you mention above though, what do you think?

devito/operator.py
rhodrin: > wdym ?

Currently, will anyone be using this functionality for Functions not of the form `F(t,x,y,z)`? In this case `(x,y,z)` will always be showing up in the subdomain and therefore we only need to filter the `t`. Order only becomes important when we have a function of the form `F(t1,t2,x,y,z)` and the order of `t1, t2` is important for loop construction. But anyway, I'll modify the code so that it should still work in the latter case.

> can you please remind me why do you have to compute the set difference of dims and subdims? I'm pretty sure 
> you've explained it to me, but I forgot

So in order to get the correct loop structure the implicit dims of the implicit equations needs to be (for say `F(t,xi,yi)`)
needs to be `dims(F)-dims(subdomain)+n` (i.e. `(t,n)`). Then the implicit dims of `F` need to be `(t,n,xi,yi)`. Hence the difference, then the addition of `n` and then later the addition of `subdomain.dimensions`. Does that make sense?

FabioLuporini: yes, thanks, I see what you mean.

devito/operator.py
```diff
@@ -208,6 +216,31 @@ def objects(self):
...
+        processed = []
+        for e in expressions:
+            if e.subdomain:
+                try:
+                    dims = [d.root for d in e.free_symbols if isinstance(d, Dimension)]
+                    sub_dims = [d.root for d in e.subdomain.dimensions]
+                    dims = [d for d in dims if d not in frozenset(sub_dims)]
+                    dims.append(e.subdomain.implicit_dimension)
+                    if e._subdomain._implicit_exprs is None:
+                        e._subdomain._implicit_exprs = \
```
FabioLuporini: One of the things that must be fixed and that I hadn't realised before is that here we are changing an Eq's state, and even worse, we're doing it explicitly from outside an Eq. But Eqs are immutable objects, so this should never happen, otherwise there may be unexpected consequences (eg if used as dictionary keys). 

I'll take care of this next week

FabioLuporini:
merged.

PR #787, opened 2019-03-25
Introduce archinfo to quickly retrieve architectural properties

FabioLuporini:
So far, collection of architectural properties was horribly scattered over multiple modules (some in ``__init__.py``, some in ``dle.utils.py``, ...). Now everything lives under (the more general) ``devito/archinfo.py``.

Also, calling ``cpuinfo`` takes, on my machine, between 1 and 1.5s. The function was memoized, but anyway this was introducing a non-negligible overhead each time one wanted to run a single, quick test. So as a workaround for Linux systems, with this PR we first try to gather the information directly reading from ``/proc/cpuinfo``, which is much faster, while ``cpuinfo`` is used as fallback. 

~This is based on #783 , hence the big diff ATM, otherwise the PR is quite small~


FabioLuporini:
> May need some reorga at some point for GPU support too (or ARM or whatever)

Absolutely. This is intended to be a first step in that direction

FabioLuporini:
merged

PR #788, opened 2019-03-25
Minor performance-related tweaks

FabioLuporini:


devito/core/autotuning.py
```diff
@@ -300,7 +300,7 @@ def generate_nthreads(nthreads, args, level):
 
 
 options = {
-    'squeezer': 4,
+    'squeezer': 2,
```
mloubout: WHy?

devito/core/autotuning.py
```diff
@@ -300,7 +300,7 @@ def generate_nthreads(nthreads, args, level):
...
     'blocksize': sorted({8, 16, 24, 32, 40, 64, 128}),
```
mloubout: 4?

examples/seismic/benchmark.py
```diff
@@ -109,17 +111,32 @@ def run(problem, **kwargs):
...
+    options = {}
+
     time_order = kwargs.pop('time_order')[0]
     space_order = kwargs.pop('space_order')[0]
-    run(space_order=space_order, time_order=time_order, **kwargs)
+    autotune = kwargs.pop('autotune')
+
+    # Should a specific block-shape be used? Useful if one wants to skip
+    # the autotuning pass as a good block-shape is already known
+    block_shape = as_tuple(kwargs.pop('block_shape'))
```
mloubout: sweet

examples/seismic/benchmark.py
```diff
@@ -109,17 +111,32 @@ def run(problem, **kwargs):
...
+    if all(block_shape):
+        if autotune:
+            warning("Skipping autotuning (using explicit block-shape `%s`)"
+                    % str(block_shape))
+            autotune = False
+        # This is quite hacky, but it does the trick
+        for d, bs in zip(['x', 'y', 'z'], block_shape):
```
mloubout: Ouch

tests/test_autotuner.py
```diff
@@ -140,10 +139,10 @@ def test_mode_destructive():
     op = Operator(Eq(f, f + 1.), dle=('advanced', {'openmp': False}))
     op.apply(time=100, autotune=('basic', 'destructive'))
 
-    # AT is expected to have executed 30 timesteps (6 block shapes, 5 timesteps each)
+    # AT is expected to have executed 18 timesteps (6 block shapes, 3 timesteps each)
     # The operator runs for 101 timesteps
-    # So, overall, f.data[0] is incremented 131 times
-    assert np.all(f.data == 131)
+    # So, overall, f.data[0] is incremented 119 times
+    assert np.all(f.data == 119)
```
mloubout: So what's the reason behind reducing the number of shapes tried?

tests/test_dse.py
```diff
@@ -139,12 +139,12 @@ def test_tti_rewrite_aggressive(tti_nodse):
     # * four Arrays are allocated on the heap
     # * two Arrays are allocated on the stack and only appear within an efunc
     arrays = [i for i in FindSymbols().visit(op) if i.is_Array]
-    assert len(arrays) == 4
+    assert len(arrays) == 3
```
mloubout: I'm confused, can't see where this change is coming from.

tests/test_dse.py
```diff
@@ -162,7 +162,7 @@ def test_tti_rewrite_aggressive_wmpi():
 
 @switchconfig(profiling='advanced')
 @pytest.mark.parametrize('kernel,space_order,expected', [
-    ('centered', 8, 168), ('centered', 16, 300)
+    ('centered', 8, 148), ('centered', 16, 264)
```
mloubout: What happened there?

devito/core/autotuning.py
FabioLuporini: quicker auto-tuning, same result at the end on my machine. Tested on a few examples

examples/seismic/benchmark.py
FabioLuporini: this is horrible, but I'd need to break too many abstractions to do it cleanly...

tests/test_autotuner.py
FabioLuporini: it's not the number of shapes tried, this is just due to the number of timesteps per shape . Ie, due to the change to ``at_squeezer`` above

devito/core/autotuning.py
mloubout: Is this user modifiable?

tests/test_dse.py
FabioLuporini: Optimization in TTI. We now only need 3 temporaries for the time invariants fields , instead of 4 . Becuase instead of storing:
* sin(theta)
* cos(theta)
* sin(phi)
* cos(phi)

the DSE now detects
* sin(theta)
* cos(theta)*sin(phi)
* sin(theta)*cos(phi)

 This is quite cute. Go check out the generated code

examples/seismic/benchmark.py
mloubout: Yeah but this won't even run in 2D

tests/test_dse.py
FabioLuporini: a consequence of now having 3 time-invariant temporaries , instaed of 4. Also a ~10% flop reduction, consistent across all space orders

tests/test_autotuner.py
mloubout: I mean since when AT is modifying the ouput? No matter how many block you try you still run `f=f+` 100 times no matter what AT does so should always be `f.data=1` this is super dangerous if not the case.

tests/test_dse.py
mloubout: Nice

devito/core/autotuning.py
FabioLuporini: only manually ATM

examples/seismic/benchmark.py
FabioLuporini: in fact, it will. Unfortunately, the entire benchmark.py won't, probably, crashing at argument parsing time

devito/core/autotuning.py
mloubout: Then I would keep 4, don't want to hard code too restrictive constants (personal can ignore)

tests/test_autotuner.py
FabioLuporini: ah, are you not aware that we have had for quite sometime "runtime autotuning" ? this is simply performing autotuning at runtime for some blockshapes

examples/seismic/benchmark.py
mloubout: That's even scarier :D

tests/test_autotuner.py
mloubout: Yes I know, but runtime AT means it  uses the first 13 steps to do AT then does the rest, but the overall should still be 100 steps, not 113.

FabioLuporini: sorry, this isn't 'runtime' mode, this is 'destructive', which essentially doesn't allocate new data, hence the 119

FabioLuporini:
merged.

PR #790, opened 2019-03-26
Rework benchmarking

FabioLuporini:


benchmarks/user/README.md
```diff
@@ -0,0 +1,233 @@
+# Benchmarking a Devito Operator
+
+## Running benchmarks
+
+`benchmark.py` implements a minimalist framework to evaluate the performance of
+a Devito Operator while varying
+
+* the problem size (e.g., shape of the computational grid)
```
georgebisbas: Trivial :
Maybe size and shape/ size of the shape...
* the problem size (e.g., shape size of the computational grid)
I know it is defined as `shape` but sounds like weird IMHO....

benchmarks/user/README.md
```diff
@@ -0,0 +1,233 @@
...
+* the discretization (e.g., space- and time-order of the input/output fields),
+* the simulation time,
```
georgebisbas: e.g. number of timesteps?

benchmarks/user/README.md
```diff
@@ -0,0 +1,233 @@
...
+## A step back: configuring your machine for reliable benchmarking
+
+If you are tempted to use your laptop to run a benchmark, you may want to
+reconsider: heat and power management may affect the results you get in an
+unpredictable way.
+
+It is important that *both* the Python process running Devito (process*es* if
+running with MPI) and the OpenMP threads spawned while running an Operator are
+pinned to specific CPU cores, to get reliable and determinisic results. There
+are several ways to achieve this.
```
georgebisbas: ways to achieve this **:**

benchmarks/user/README.md
FabioLuporini: in fact no, only the simulation time in milliseconds con be provided, from which the number of timesteps will be derived

mloubout: In millisecond

benchmarks/user/README.md
```diff
@@ -0,0 +1,233 @@
+# Benchmarking a Devito Operator
+
+## Running benchmarks
+
+`benchmark.py` implements a minimalist framework to evaluate the performance of
+a Devito Operator while varying
+
+* the problem size (e.g., shape of the computational grid)
+* the discretization (e.g., space- and time-order of the input/output fields),
```
mloubout: so and to of wavefields

benchmarks/user/README.md
```diff
@@ -0,0 +1,233 @@
...
+optimizations applied by an Operator, how auto-tuning is getting along, and to
+emit more performance metrics.
+
+You can also set `DEVITO_DEBUG_COMPILER=1` to emit the command used to compile
+the generated code.
+
+## Tips
+
+* The 'O3' mode has been found to be particularly beneficial in TTI Operators,
+  by resulting in significant reductions in operation count.
```
mloubout: With intel arch

benchmarks/user/README.md
```diff
@@ -0,0 +1,233 @@
...
+htop
+```
+or to keep the memory footprint under control:
+```
+watch numastat -m
+```
+
+## Known limitations and possible work arounds
+
+ * The DSE `aggressive` mode will not work with backend compilers that are not
```
mloubout: It will work just won't be efficient

benchmarks/user/README.md
FabioLuporini: mmhh, to be fair, I think it's quite clear the way it is

benchmarks/user/README.md
FabioLuporini: right, fixing

benchmarks/user/README.md
FabioLuporini: right, fixing

benchmarks/user/README.md
FabioLuporini: I think I prefer "input/output fields" because one day we may want to use this for something that is not waves

FabioLuporini:
@mloubout @georgebisbas addressed some of your comments

FabioLuporini:
merged.

PR #792, opened 2019-03-28
Rework Platform-related machinery

FabioLuporini:
This PR adds the Platform class hierarchy (CPU64, Intel64, KNL, ... ARM, ..., Power, ...) so that the compiler can make platform-specific choices when it comes to optimization. 

The nice thing now is that if you want to cross-compile (note: I'm talking about the code that devito generates, not the actual machine code) for a certain target architecture, now you can do it with ``DEVITO_PLATFORM=...`` and for example ``knl`` rather than ``hsw``. On KNL, for example, nested parallel loops are collapsed and nested parallelism is enabled.
By default, we stick to the ``cpu64`` mode, which is essentially what we have in master

This PR also drops now uselss environment variables: DEVITO_DLE_OPTIONS and DEVITO_ISA. In particular, the highest ISA now is bound directly to a Platform, so no need to set it up explicitly.

### DIFF
--- a/benchmarks/user/benchmark.py
+++ b/benchmarks/user/benchmark.py
@@ -56,12 +56,11 @@ def option_performance(f):
 
     _preset = {
         # Fixed
-        'O1': {'dse': 'basic', 'dle': 'basic'},
+        'O1': {'dse': 'basic', 'dle': 'advanced'},
         'O2': {'dse': 'advanced', 'dle': 'advanced'},
         'O3': {'dse': 'aggressive', 'dle': 'advanced'},
         # Parametric
         'dse': {'dse': ['basic', 'advanced', 'aggressive'], 'dle': 'advanced'},
-        'dle': {'dse': 'advanced', 'dle': ['basic', 'advanced']}
     }
 
     def from_preset(ctx, param, value):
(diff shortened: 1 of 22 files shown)


benchmarks/user/benchmark.py
```diff
@@ -56,12 +56,11 @@ def option_performance(f):
 
     _preset = {
         # Fixed
-        'O1': {'dse': 'basic', 'dle': 'basic'},
+        'O1': {'dse': 'basic', 'dle': 'advanced'},
```
mloubout: really?

devito/archinfo.py
```diff
@@ -25,37 +22,72 @@ def get_cpu_info():
...
-    isa = 'cpp'
-    for i in reversed(known_isas):
-        if any(j.startswith(i) for j in cpu_info['flags']):
-            # Using `startswith`, rather than `==`, as a flag such as 'avx512'
-            # appears as 'avx512f, avx512cd, ...'
-            isa = i
-            break
-    return isa
+def lscpu():
+    p1 = Popen(['lscpu'], stdout=PIPE, stderr=PIPE)
```
mloubout: `lscpu` doesn't exist on osx

benchmarks/user/benchmark.py
FabioLuporini: yes, I dropped dle basic, it was useless

FabioLuporini:
> lscpu only works on linux system

indeed

> so this gonna break on osx at least

no, or, I mean, this is not the point. That piece of code **attempts** to leverage ``lscpu`` -- we just hope it's available on the underlying system. But if it's not, we proceed silently (see Popen's ``stderr`` gets captured). IOW, the platform autodetection attempts a number of approaches and then if they all fail we still to the fallback ``CPU64``, representing "a generic 64-bit cpu architecture"

FabioLuporini:
all sorted out and green light given offline, so merging

FabioLuporini:
merged.

PR #794, opened 2019-03-28
fast math with gcc

mloubout:
Add fast-math to gcc


FabioLuporini:
maybe we can hack this up by dropping ffast-math from [here](https://github.com/opesci/devito/blob/master/tests/conftest.py) ie at test loading time, clearly adding a big comment warning why we do this

obviously, assuming that all failures are due to numerical precision

mloubout:
Merging, long overdue

PR #798, opened 2019-04-04
Reinstate all seismic tutorials on CI

FabioLuporini:
now attempting to use 2 dask workers instead of 5


ggorman:
I see multiple issue here:
* I thought we agreed that Dask, like MPI, wasn't going to be tested on Travis because nodes get oversubscribed and die randomly.
* master needs merging - package versions in requirements.txt are out of date etc.
* The actual error is 
```ImportError: cannot import name 'PollIOLoop'
```
This is a dependency of tornado which is a dependency of distributed (ie dask). Not clear why it would be failing on travis and not elsewhere.

PR #801, opened 2019-04-08
fix bunch of minor bugs

FabioLuporini:
- Symbol and Scalar pickling
- ConditionalDimension pickling
- SubDimension pickling
- ConditionalDimension usage with symbolic ``factor``

all equipped with tests


FabioLuporini:
merged.

PR #804, opened 2019-04-11
MMS-elastic RFC

georgebisbas:
Code for review
Acoustic is Mathias's


mloubout:
Few comments on the elastic:
Cell 2, wrong weights/parenthesis on sources 
Cell 6, sources missing

There is a loooot of duplicates/commented code so it's not very easy to go through

rhodrin:
Not sure if you're ready for me to look through these notebooks in details yet but a couple of initial questions/comments:
1) Are these meant to be examples or tests?
2) Review the grammar and spelling in the notebooks (if these are going to be examples).

georgebisbas:
> Not sure if you're ready for me to look through these notebooks in detail yet but a couple of initial questions/comments:
> 
> 1. Are these meant to be examples or tests?
> 2. Review the grammar and spelling in the notebooks (if these are going to be examples).

Not ready yet. Currently it should be something in between test and examples. We just need to focus in having the proper convergence. The way this will be pushed to repo will be decided at a later stage. 
Sent from a mobile.

georgebisbas:
> Few comments on the elastic:
> Cell 2, wrong weights/parenthesis on sources
@mloubout is it possible to point me the lines and what you mean by weights/parentheses?
If you have a handy/ example kind of fix for the correct weights, it would be of great convenience.

mloubout:
All the lines like
`wave_vx  = Eq(vx.forward,  vx - dt*ro*(txx.dx + txz.dz) + src_vx) `

are incoorrect, you need to remember that the equation is

`1/rho vx.dt = txx.dx + txz.dz   + src_vx`

so you need to be really carefull all the time about what `rho, dt, v, ...` factors too like in the above equation it should be 

`wave_vx  = Eq(vx.forward,  vx - dt*ro*(txx.dx + txz.dz + src_vx)) `

mloubout:
@georgebisbas we would usually use `solve` to avoid these constants/weoghts misplacement in the equations, it may be extremely costly and slow considering the complexity of `src_vx,...` and the sympy implementation of it. So I would still do it by hand but carefully

FabioLuporini:
> @georgebisbas we would usually use solve to avoid these constants/weoghts misplacement in the equations, it may be extremely costly and slow considering the complexity of src_vx,... and the sympy implementation of it. So I would still do it by hand but carefully

though **maybe** with lazy this wont be an issue anymore ? 😮

mloubout:
>though maybe with lazy this wont be an issue anymore ?

No, the issue here is the massive analytical expression as a source, nothing to do with FD.

FabioLuporini:
we're currently checking against precomputed norms. So closing for now, because nobody has the time to work on it

PR #811, opened 2019-04-23
openmp scheduling fixes

FabioLuporini:
all non-affine loop nests now use ``schedule(static)`` rather than ``schedule(static,1)``. This improve performance of ompized sparse loops

### DIFF
--- a/devito/dle/parallelizer.py
+++ b/devito/dle/parallelizer.py
@@ -76,7 +76,8 @@ class Ompizer(object):
     """
 
     lang = {
-        'for': lambda i: c.Pragma('omp for collapse(%d) schedule(static,1)' % i),
+        'for-static': lambda i: c.Pragma('omp for collapse(%d) schedule(static)' % i),
+        'for-static-1': lambda i: c.Pragma('omp for collapse(%d) schedule(static,1)' % i),
         'par-for': lambda i, j: c.Pragma('omp parallel for collapse(%d) '
                                          'schedule(static,1) num_threads(%d)' % (i, j)),
         'simd-for': c.Pragma('omp simd'),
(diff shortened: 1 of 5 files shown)


devito/dle/parallelizer.py
```diff
@@ -69,14 +71,22 @@ class Ompizer(object):
     than this threshold.
     """
 
-    COLLAPSE = 32
+    COLLAPSE_NCORES = 32
```
FabioLuporini: @tjb900 note I have changed these names

devito/dle/parallelizer.py
```diff
@@ -161,8 +159,10 @@ def _make_partree(self, candidates, omp_pragma=None):
                 if i.is_Vectorizable:
                     break
 
+                # Would there be enough work per parallel iteration?
                 try:
-                    work = reduce(mul, [int(j.dim.symbolic_size) for j in candidates[n:]])
+                    work = prod([int(j.dim.symbolic_size) for j in candidates[n+1:]],
+                                default=Ompizer.COLLAPSE_WORK)
```
tjb900: I don't think you want this default here - the last argument to `reduce` is prepended to the list, which means this will never be less than `Ompizer.COLLAPSE_WORK`.  I think we need to handle the empty list case explicitly.

```
In [4]: reduce(mul, [1,2,3],2)
Out[4]: 12
```

tjb900: Actually, what's the harm in it returning `1` if there are no deeper loops?  Isn't that exactly the truth? (and indeed, we would not want it to include that last non-loop in the collapse).

FabioLuporini:
Merged.

tests/test_dle.py
```diff
@@ -314,6 +314,28 @@ def test_collapsing(self, eq, expected, blocking):
...
+        Affine iterations -> #pragma omp ... schedule(static,1) ...
+        Non-affine iterations -> #pragma omp ... schedule(static) ...
+        """
+        grid = Grid(shape=(11, 11))
+
+        u = TimeFunction(name='u', grid=grid, time_order=2, save=5, space_order=0)
+        sf1 = SparseTimeFunction(name='s', grid=grid, npoint=1, nt=5)
+
+        eqns = [Eq(u.forward, u + 1)]
+        eqns += sf1.interpolate(u)
```
btrb: is this possible only for sparse functions ?

PR #813, opened 2019-04-26
Autotuning pickle fix

tjb900:
This fixes a tricky issue where runtime autotuning generates incorrect results, if the Operator has been pickled/unpickled.  The unpickling process creates a new instance of `IterationDirection`, which is equal to but not identical to `Backward`, and so the `is` comparison fails.  The runtime autotuning then executes the timesteps in the wrong order.

This changes the comparisons to equality, which works correctly.

I've put the "be extra careful" as a separate commit to make it simpler to throw away if you don't want to go that far.

A potentially more elegant solution would somehow create `Forward`, `Backward`, and `Any` as singletons, and somehow override the unpickling to ensure that there is only one instance.

A grep of the rest of the code only shows up one other case where this kind of test is used.  In `devito/types/dense.py` there are a lot of comparisons using `is`, e.g.:

```
        if side is not None:
            idx = idx*2 + (0 if side is LEFT else 1)
```

These might or might not be an issue.

### DIFF
--- a/devito/core/autotuning.py
+++ b/devito/core/autotuning.py
@@ -7,7 +7,7 @@ import psutil
 
 from devito.archinfo import KNL
 from devito.dle import BlockDimension
-from devito.ir import Backward, retrieve_iteration_tree
+from devito.ir import Backward, Forward, retrieve_iteration_tree
 from devito.logger import perf, warning as _warning
 from devito.mpi.distributed import MPI, MPINeighborhood
 from devito.mpi.routines import MPIMsgEnriched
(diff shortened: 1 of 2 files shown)


FabioLuporini:
Uncontroversial.

Is it possible to add a test to ``test_autotuner`` ?

I agree singletons is the way to go; overall, we want as many ``is`` comparisons as possible for efficiency. I'll file an issue

FabioLuporini:
Im tempted to say we should go with the better fix here (ie, make Tag inherit from Enum). I'll add it to my backlog.

FabioLuporini:
closing this for now. I'm gonna see soon if we can make these objects singletons thus overriding the pickling/unpickling behaviour which leads to multiple instances of what should conceptually be the same entity

also, this would lack a test anyway

PR #824, opened 2019-05-02
Data slice fix

rhodrin:
In the current master, slicing function data does not work correctly when the step is negative i.e. something along the lines of `u.data[5::-1]` does not return the expected result.

This problem occurs due to how the `start` and `stop` of a `slice` are set in `convert_index` (found in `devito/data/decomposition.py`).

The purpose of this PR will be to fix the bug stated above and to add some tests checking slicing with negative steps is carried out correctly.

Note that the main purpose PR in its current state is to ensure that the update passes all tests. The manner in which the bug is addressed in `convert_index` can still potentially be tidied up.

### DIFF
--- a/devito/data/data.py
+++ b/devito/data/data.py
@@ -1,8 +1,10 @@
 from collections.abc import Iterable
+from functools import wraps
 
 import numpy as np
 
 from devito.data.allocators import ALLOC_FLAT
+from devito.data.utils import *
 from devito.parameters import configuration
 from devito.tools import Tag, as_tuple, is_integer
 
@@ -26,6 +28,9 @@ class Data(np.ndarray):
         If the i-th entry is True, then the i-th array dimension uses modulo indexing.
     allocator : MemoryAllocator, optional
         Used to allocate memory. Defaults to ``ALLOC_FLAT``.
+    distributor : Distributor, optional
+        The distributor from which the original decomposition was produced. Note that
+        the decomposition Parameter above may be different to distributor.decomposition.
 
     Notes
     -----
@@ -38,7 +43,8 @@ class Data(np.ndarray):
     Data.
     """
 
-    def __new__(cls, shape, dtype, decomposition=None, modulo=None, allocator=ALLOC_FLAT):
+    def __new__(cls, shape, dtype, decomposition=None, modulo=None, allocator=ALLOC_FLAT,
+                distributor=None):
         assert len(shape) == len(modulo)
         ndarray, memfree_args = allocator.alloc(shape, dtype)
         obj = ndarray.view(cls)
@@ -46,6 +52,7 @@ class Data(np.ndarray):
         obj._memfree_args = memfree_args
         obj._decomposition = decomposition or (None,)*len(shape)
         obj._modulo = modulo or (False,)*len(shape)
+        obj._distributor = distributor
 
         # This cannot be a property, as Data objects constructed from this
         # object might not have any `decomposition`, but they would still be
@@ -76,6 +83,7 @@ class Data(np.ndarray):
             # `self` was created through __new__()
             return
 
+        self._distributor = None
         self._index_stash = None
 
         # Views or references created via operations on `obj` do not get an
@@ -91,6 +99,7 @@ class Data(np.ndarray):
         elif obj._index_stash is not None:
             # From `__getitem__`
             self._is_distributed = obj._is_distributed
+            self._distributor = obj._distributor
             glb_idx = obj._normalize_index(obj._index_stash)
             self._modulo = tuple(m for i, m in zip(glb_idx, obj._modulo)
                                  if not is_integer(i))
@@ -105,6 +114,7 @@ class Data(np.ndarray):
             self._decomposition = tuple(decomposition)
         else:
             self._is_distributed = obj._is_distributed
+            self._distributor = obj._distributor
             if self.ndim == obj.ndim:
                 # E.g., from a ufunc, such as `np.add`
                 self._modulo = obj._modulo
@@ -132,6 +142,27 @@ class Data(np.ndarray):
         ret._is_distributed = any(i is not None for i in decomposition)
         return ret
 
+    def _check_idx(func):
+        """Check if __getitem__/__setitem__ may require communication across MPI ranks."""
+        @wraps(func)
+        def wrapper(data, *args, **kwargs):
+            glb_idx = args[0]
+            if len(args) > 1 and isinstance(args[1], Data) \
+                    and args[1]._is_mpi_distributed:
+                comm_type = index_by_index
+            elif data._is_mpi_distributed:
+                for i in as_tuple(glb_idx):
+                    if isinstance(i, slice) and i.step is not None and i.step < 0:
+                        comm_type = index_by_index
+                        break
+                    else:
+                        comm_type = serial
+            else:
+                comm_type = serial
+            kwargs['comm_type'] = comm_type
+            return func(data, *args, **kwargs)
+        return wrapper
+
     @property
     def _is_mpi_distributed(self):
         return self._is_distributed and configuration['mpi']
@@ -139,9 +170,53 @@ class Data(np.ndarray):
     def __repr__(self):
         return super(Data, self._local).__repr__()
 
-    def __getitem__(self, glb_idx):
-        loc_idx = self._convert_index(glb_idx)
-        if loc_idx is NONLOCAL:
+    @_check_idx
+    def __getitem__(self, glb_idx, comm_type):
+        loc_idx = self._index_glb_to_loc(glb_idx)
+        if comm_type is index_by_index:
+            # Retrieve the pertinent local data prior to mpi send/receive operations
+            data_idx = loc_data_idx(loc_idx)
+            local_val = super(Data, self).__getitem__(data_idx)
+
+            comm = self._distributor.comm
+            rank = comm.Get_rank()
+
+            owners, send, global_si, local_si = \
+                mpi_index_maps(loc_idx, local_val.shape, self._distributor.topology,
+                               self._distributor.all_coords, comm)
+
+            retval = Data(local_val.shape, local_val.dtype.type,
+                          decomposition=local_val._decomposition,
+                          modulo=local_val._modulo)
+            it = np.nditer(owners, flags=['refs_ok', 'multi_index'])
+            # Iterate over each element of data
+            while not it.finished:
+                index = it.multi_index
+                if rank == owners[index] and rank == send[index]:
+                    # Current index and destination index are on the same rank
+                    loc_ind = local_si[index]
+                    send_ind = local_si[global_si[index]]
+                    retval.data[send_ind] = local_val.data[loc_ind]
+                elif rank == owners[index]:
+                    # Current index is on this rank and hence need to send
+                    # the data to the appropriate rank
+                    loc_ind = local_si[index]
+                    send_rank = send[index]
+                    send_ind = global_si[index]
+                    send_val = local_val.data[loc_ind]
+                    reqs = comm.isend([send_ind, send_val], dest=send_rank)
+                    reqs.wait()
+                elif rank == send[index]:
+                    # Current rank is required to receive data from this index
+                    recval = comm.irecv(source=owners[index])
+                    local_dat = recval.wait()
+                    loc_ind = local_si[local_dat[0]]
+                    retval.data[loc_ind] = local_dat[1]
+                else:
+                    pass
+                it.iternext()
+            return retval
+        elif loc_idx is NONLOCAL:
             # Caller expects a scalar. However, `glb_idx` doesn't belong to
             # self's data partition, so None is returned
             return None
@@ -151,8 +226,9 @@ class Data(np.ndarray):
             self._index_stash = None
             return retval
 
-    def __setitem__(self, glb_idx, val):
-        loc_idx = self._convert_index(glb_idx)
+    @_check_idx
+    def __setitem__(self, glb_idx, val, comm_type):
+        loc_idx = self._index_glb_to_loc(glb_idx)
         if loc_idx is NONLOCAL:
             # no-op
             return
@@ -164,7 +240,26 @@ class Data(np.ndarray):
             else:
                 super(Data, self).__setitem__(glb_idx, val)
         elif isinstance(val, Data) and val._is_distributed:
-            if self._is_distributed:
+            if comm_type is index_by_index:
+                glb_idx, val = self._process_args(glb_idx, val)
+                val_idx = as_tuple([slice(i.glb_min, i.glb_max+1, 1) for
+                                    i in val._decomposition])
+                idx = self._set_global_idx(val, glb_idx, val_idx)
+                comm = self._distributor.comm
+                nprocs = self._distributor.nprocs
+                # Prepare global lists:
+                data_global = []
+                idx_global = []
+                for j in range(nprocs):
+                    data_global.append(comm.bcast(np.array(val), root=j))
+                    idx_global.append(comm.bcast(idx, root=j))
+                # Set the data:
+                for j in range(nprocs):
+                    skip = any(i is None for i in idx_global[j]) \
+                        or data_global[j].size == 0
+                    if not skip:
+                        self.__setitem__(idx_global[j], data_global[j])
+            elif self._is_distributed:
                 # `val` is decomposed, `self` is decomposed -> local set
                 super(Data, self).__setitem__(glb_idx, val)
             else:
@@ -173,9 +268,10 @@ class Data(np.ndarray):
         elif isinstance(val, np.ndarray):
             if self._is_distributed:
                 # `val` is replicated, `self` is decomposed -> `val` gets decomposed
-                val_idx = self._normalize_index(glb_idx)
+                glb_idx = self._normalize_index(glb_idx)
+                glb_idx, val = self._process_args(glb_idx, val)
                 val_idx = [index_dist_to_repl(i, dec) for i, dec in
-                           zip(val_idx, self._decomposition)]
+                           zip(glb_idx, self._decomposition)]
                 if NONLOCAL in val_idx:
                     # no-op
                     return
@@ -188,6 +284,15 @@ class Data(np.ndarray):
                 # * one of them is 1
                 # Conceptually, below we apply the same rule
                 val_idx = val_idx[len(val_idx)-val.ndim:]
+                processed = []
+                # Handle step size > 1
+                for i, j in zip(glb_idx, val_idx):
+                    if isinstance(i, slice) and i.step is not None and i.step > 1 and \
+                            j.stop > j.start:
+                        processed.append(slice(j.start, j.stop, 1))
+                    else:
+                        processed.append(j)
+                val_idx = as_tuple(processed)
                 val = val[val_idx]
             else:
                 # `val` is replicated`, `self` is replicated -> plain ndarray.__setitem__
@@ -195,8 +300,9 @@ class Data(np.ndarray):
             super(Data, self).__setitem__(glb_idx, val)
         elif isinstance(val, Iterable):
             if self._is_mpi_distributed:
-                raise NotImplementedError("With MPI data can only be set "
-                                          "via scalars or numpy arrays")
+                raise NotImplementedError("With MPI, data can only be set "
+                                          "via scalars, numpy arrays or "
+                                          "other data ")
             super(Data, self).__setitem__(glb_idx, val)
         else:
             raise ValueError("Cannot insert obj of type `%s` into a Data" % type(val))
@@ -210,16 +316,61 @@ class Data(np.ndarray):
             if any(i is Ellipsis for i in idx):
                 # Explicitly replace the Ellipsis
                 items = (slice(None),)*(self.ndim - len(idx) + 1)
-                return idx[:idx.index(Ellipsis)] + items + idx[idx.index(Ellipsis)+1:]
+                items = idx[:idx.index(Ellipsis)] + items + idx[idx.index(Ellipsis)+1:]
             else:
-                return idx + (slice(None),)*(self.ndim - len(idx))
+                items = idx + (slice(None),)*(self.ndim - len(idx))
+            # Normalize slice steps:
+            processed = [slice(i.start, i.stop, 1) if
+                         (isinstance(i, slice) and i.step is None)
+                         else i for i in items]
+            return as_tuple(processed)
 
-    def _convert_index(self, glb_idx):
+    def _process_args(self, idx, val):
+        """If comm_type is parallel we need to first retrieve local unflipped data."""
+        if any(isinstance(i, slice) and i.step is not None and i.step < 0
+               for i in as_tuple(idx)):
+            processed = []
+            transform = []
+            for j, k in zip(idx, self._distributor.glb_shape):
+                if isinstance(j, slice) and j.step is not None and j.step < 0:
+                    if j.start is None:
+                        stop = None
+                    else:
+                        stop = j.start + 1
+                    if j.stop is None and j.start is None:
+                        start = int(np.mod(k-1, -j.step))
+                    elif j.stop is None:
+                        start = int(np.mod(j.start, -j.step))
+                    else:
+                        start = j.stop + 1
+                    processed.append(slice(start, stop, -j.step))
+                    transform.append(slice(None, None, np.sign(j.step)))
+                else:
+                    processed.append(j)
+            if isinstance(val, Data) and len(transform) > 0 and \
+                    len(val._distributor.shape) > len(val.shape):
+                # Rebuild the distributor since the dimension of the slice
+                # is different to that of the original array
+                distributor = \
+                    val._distributor._rebuild(val.shape,
+                                              self._distributor.dimensions,
+                                              self._distributor.comm)
+                new_val = Data(val.shape, val.dtype.type,
+                               decomposition=val._decomposition, modulo=val._modulo,
+                               distributor=distributor)
+                slc = as_tuple([slice(None, None, 1) for j in transform])
+                new_val[slc] = val[slc]
+                return as_tuple(processed), new_val[as_tuple(transform)]
+            else:
+                return as_tuple(processed), val[as_tuple(transform)]
+        else:
+            return idx, val
+
+    def _index_glb_to_loc(self, glb_idx):
         glb_idx = self._normalize_index(glb_idx)
-
         if len(glb_idx) > self.ndim:
             # Maybe user code is trying to add a new axis (see np.newaxis),
-            # so the resulting array will be higher dimensional than `self`,
+            # so the resulting array will be higher dimensional than `self`
             if self._is_mpi_distributed:
                 raise ValueError("Cannot increase dimensionality of MPI-distributed Data")
             else:
@@ -235,7 +386,7 @@ class Data(np.ndarray):
                 # Need to convert the user-provided global indices into local indices.
                 # Obviously this will have no effect if MPI is not used
                 try:
-                    v = index_glb_to_loc(i, dec)
+                    v = convert_index(i, dec, mode='glb_to_loc')
                 except TypeError:
                     if self._is_mpi_distributed:
                         raise NotImplementedError("Unsupported advanced indexing with "
@@ -260,108 +411,54 @@ class Data(np.ndarray):
 
         return loc_idx[0] if len(loc_idx) == 1 else tuple(loc_idx)
 
+    def _set_global_idx(self, val, idx, val_idx):
+        """
+        Compute the global indices to which val (the locally stored data) correspond.
+        """
+        data_loc_idx = as_tuple(val._index_glb_to_loc(val_idx))
+        data_glb_idx = []
+        # Convert integers to slices so that shape dims are preserved
+        if is_integer(as_tuple(idx)[0]):
+            data_glb_idx.append(slice(0, 1, 1))
+        for i, j in zip(data_loc_idx, val._decomposition):
+            if not j.loc_empty:
+                data_glb_idx.append(j.index_loc_to_glb(i))
+            else:
+                data_glb_idx.append(None)
+        mapped_idx = []
+        # Based `data_glb_idx` the indices to which the locally stored data
+        # block correspond can now be computed
+        for i, j, k in zip(data_glb_idx, as_tuple(idx), self._decomposition):
+            if isinstance(j, slice) and j.start is None:
+                norm = 0
+            elif isinstance(j, slice) and j.start is not None:
+                norm = j.start
+            else:
+                norm = j
+            if i is not None:
+                loc_max = k.loc_abs_max
+                if isinstance(j, slice) and j.step is not None:
+                    stop = j.step*i.stop+norm
+                else:
+                    stop = i.stop+norm
+                if stop > loc_max+1:
+                    stop = loc_max+1
+            if i is not None:
+                if isinstance(j, slice) and j.step is not None:
+                    mapped_idx.append(slice(j.step*i.start+norm,
+                                            stop, j.step))
+                else:
+                    mapped_idx.append(slice(i.start+norm, stop, i.step))
+            else:
+                mapped_idx.append(None)
+        return as_tuple(mapped_idx)
+
     def reset(self):
         """Set all Data entries to 0."""
         self[:] = 0.0
 
 
-class Index(Tag):
+class CommType(Tag):
     pass
-NONLOCAL = Index('nonlocal')  # noqa
-PROJECTED = Index('projected')
-
-
-def index_is_basic(idx):
-    if is_integer(idx):
-        return True
-    elif isinstance(idx, (slice, np.ndarray)):
-        return False
-    else:
-        return all(is_integer(i) or (i is NONLOCAL) for i in idx)
-
-
-def index_apply_modulo(idx, modulo):
-    if is_integer(idx):
-        return idx % modulo
-    elif isinstance(idx, slice):
-        if idx.start is None:
-            start = idx.start
-        elif idx.start >= 0:
-            start = idx.start % modulo
-        else:
-            start = -(idx.start % modulo)
-        if idx.stop is None:
-            stop = idx.stop
-        elif idx.stop >= 0:
-            stop = idx.stop % (modulo + 1)
-        else:
-            stop = -(idx.stop % (modulo + 1))
-        return slice(start, stop, idx.step)
-    elif isinstance(idx, (tuple, list)):
-        return [i % modulo for i in idx]
-    elif isinstance(idx, np.ndarray):
-        return idx
-    else:
-        raise ValueError("Cannot apply modulo to index of type `%s`" % type(idx))
-
-
-def index_dist_to_repl(idx, decomposition):
-    """Convert a distributed array index into a replicated array index."""
-    if decomposition is None:
-        return PROJECTED if is_integer(idx) else slice(None)
-
-    # Derive shift value
-    value = idx.start if isinstance(idx, slice) else idx
-    if value is None:
-        value = 0
-    elif not is_integer(value):
-        raise ValueError("Cannot derive shift value from type `%s`" % type(value))
-
-    # Convert into absolute local index
-    idx = decomposition.convert_index(idx, rel=False)
-
-    if is_integer(idx):
-        return PROJECTED
-    elif idx is None:
-        return NONLOCAL
-    elif isinstance(idx, (tuple, list)):
-        return [i - value for i in idx]
-    elif isinstance(idx, np.ndarray):
-        return idx - value
-    elif isinstance(idx, slice):
-        return slice(idx.start - value, idx.stop - value, idx.step)
-    else:
-        raise ValueError("Cannot apply shift to type `%s`" % type(idx))
-
-
-def index_glb_to_loc(idx, decomposition):
-    """Convert a global index into a local index."""
-    if is_integer(idx) or isinstance(idx, slice):
-        return decomposition(idx)
-    elif isinstance(idx, (tuple, list)):
-        return [decomposition(i) for i in idx]
-    elif isinstance(idx, np.ndarray):
-        return np.vectorize(lambda i: decomposition(i))(idx)
-    else:
-        raise ValueError("Cannot convert global index of type `%s` into a local index"
-                         % type(idx))
-
-
-def index_handle_oob(idx):
-    # distributed.convert_index returns None when the index is globally
-    # legal, but out-of-bounds for the calling MPI rank
-    if idx is None:
-        return NONLOCAL
-    elif isinstance(idx, (tuple, list)):
-        return [i for i in idx if i is not None]
-    elif isinstance(idx, np.ndarray):
-        if idx.dtype == np.bool_:
-            # A boolean mask, nothing to do
-            return idx
-        elif idx.ndim == 1:
-            return np.delete(idx, np.where(idx == None))  # noqa
-        else:
-            raise ValueError("Cannot identify OOB accesses when using "
-                             "multidimensional index arrays")
-    else:
-        return idx
+index_by_index = CommType('index_by_index')  # noqa
+serial = CommType('serial')
--- a/devito/data/decomposition.py
+++ b/devito/data/decomposition.py
@@ -4,7 +4,7 @@ import numpy as np
 from cached_property import cached_property
 
 from devito.data.meta import LEFT
-from devito.tools import is_integer
+from devito.tools import is_integer, as_tuple
 
 __all__ = ['Decomposition']
 
@@ -121,11 +121,20 @@ class Decomposition(tuple):
             ret.append(item)
         return 'Decomposition(%s)' % ', '.join(ret)
 
-    def __call__(self, *args, rel=True):
-        """Alias for ``self.convert_index``."""
-        return self.convert_index(*args, rel=rel)
+    def __call__(self, *args, mode='glb_to_loc', rel=True):
+        """
+        Alias for ``self.index_glb_to_loc`` or ``self.index_loc_to_glb``
+        depending on mode.
+        """
+        if mode == 'glb_to_loc':
+            return self.index_glb_to_loc(*args, rel=rel)
+        elif mode == 'loc_to_glb':
+            return self.index_loc_to_glb(*args, rel=rel)
+        else:
+            raise ValueError("Mode not recognised. Available options: "
+                             "'glb_to_loc', 'loc_to_glb'")
 
-    def convert_index(self, *args, rel=True):
+    def index_glb_to_loc(self, *args, rel=True):
         """
         Convert a global index into a relative (default) or absolute local index.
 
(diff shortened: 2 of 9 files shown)


devito/data/data.py
```diff
@@ -210,9 +210,21 @@ def _normalize_index(self, idx):
             if any(i is Ellipsis for i in idx):
                 # Explicitly replace the Ellipsis
                 items = (slice(None),)*(self.ndim - len(idx) + 1)
-                return idx[:idx.index(Ellipsis)] + items + idx[idx.index(Ellipsis)+1:]
+                items = idx[:idx.index(Ellipsis)] + items + idx[idx.index(Ellipsis)+1:]
             else:
-                return idx + (slice(None),)*(self.ndim - len(idx))
+                items = idx + (slice(None),)*(self.ndim - len(idx))
+            # Normalize slice steps:
+            processed = []
```
FabioLuporini: this whole loop looks like is expressible as
```
processed = [slice(i.start, i.stop, 1) (if isinstance(i, slice) and i.step is None) else i
             for i in items]
```

devito/data/decomposition.py
```diff
@@ -241,6 +249,25 @@ def convert_index(self, *args, rel=True):
                 if glb_idx_max is not None and glb_idx_max < 0:
                     glb_idx_max = self.glb_max + glb_idx_max + 1
                 # -> Do the actual conversion
+                # First check for the special case of a slice with negative step
+                if isinstance(glb_idx, slice):
+                    if glb_idx.step < 0:
```
FabioLuporini: you can squash this ``if`` with the upper one?
```
if isinstance(glb_idx, slice) and glb_idx.step < 0:
    ...
```

devito/data/data.py
rhodrin: Updated.

devito/data/decomposition.py
rhodrin: Cool, I just learned about 'short circuit evaluation'.

tests/test_data.py
```diff
@@ -63,6 +63,17 @@ def test_advanced_indexing(self):
         assert np.all(u.data[1, :, :, 0] == 0.)
         assert np.all(u.data[1, :, :, -1] == 0.)
 
+    @skipif('yask')
+    def test_negative_step(self):
+        """Test slicing with a negative step."""
+        grid = Grid(shape=(6, 6, 6))
+        u = TimeFunction(name='u', grid=grid, space_order=1, time_order=1)
```
FabioLuporini: FWIW, space_order=1 and time_order=1 are the default

devito/data/decomposition.py
```diff
@@ -230,9 +230,17 @@ def convert_index(self, *args, rel=True):
...
-                    glb_idx_max = self.glb_max if glb_idx.stop is None else glb_idx.stop-1
-                    retfunc = lambda a, b: slice(a, b + 1, glb_idx.step)
+                    if glb_idx.step >= 0:
+                        glb_idx_min = self.glb_min if glb_idx.start is None \
+                            else glb_idx.start
+                        glb_idx_max = self.glb_max if glb_idx.stop is None \
+                            else glb_idx.stop-1
+                        retfunc = lambda a, b: slice(a, b + 1, glb_idx.step)
+                    else:
+                        glb_idx_max = self.glb_max if glb_idx.start is None \
```
FabioLuporini: are you sure this is not ``self.glb_min if glb_idx.start is None else glb_idx.start`` ?
ie ``self.glb_min`` instead of ``self.glb_max`` ? I haven't thought about this too much, but it seems off that you pick the glb_max if None *and* the slice start (ie the **min**) otherwise

devito/data/decomposition.py
```diff
@@ -241,6 +249,24 @@ def convert_index(self, *args, rel=True):
                 if glb_idx_max is not None and glb_idx_max < 0:
                     glb_idx_max = self.glb_max + glb_idx_max + 1
                 # -> Do the actual conversion
+                # First check for the special case of a slice with negative step
+                if isinstance(glb_idx, slice) and glb_idx.step < 0:
+                    if glb_idx_max is None or glb_idx_max > self.loc_abs_max:
```
FabioLuporini: Now that I look at this more carefully...are these lines, 254~259 (extremes included), identical to lines 276~281?

devito/data/decomposition.py
```diff
@@ -241,6 +249,24 @@ def convert_index(self, *args, rel=True):
...
+                        loc_max = self.loc_abs_max - base
+                    elif glb_idx_max < self.loc_abs_min:
+                        return retfunc(-1, -3)
+                    else:
+                        loc_max = glb_idx_max - base
+                    if glb_idx_min is not None:
```
FabioLuporini: likewise, these look very similar to the block below, except for the -1 ...

devito/data/decomposition.py
rhodrin: If we're stepping backwards `slice.stop` < `slice.start`: hence the flip. Does that make sense to you?

devito/data/decomposition.py
rhodrin: Indeed. I've tidied the code a bit now.

devito/data/decomposition.py
rhodrin: This `if glb_idx_min is not None:` is also different which makes compacting things a bit tricky?

rhodrin:
> I was gonna merge this, but I just noticed there are no new MPI-related tests?

Yeah, trying to figure out how this stuff works atm.

devito/data/decomposition.py
FabioLuporini: ahhhh! sure sorry my bad

devito/data/decomposition.py
FabioLuporini: yep agreed

devito/data/decomposition.py
```diff
@@ -270,6 +286,92 @@ def convert_index(self, *args, rel=True):
         else:
             raise TypeError("Expected 1 or 2 arguments, found %d" % len(args))
 
+    def convert_index_global(self, *args, rel=True):
```
FabioLuporini: I think it's worth renaming the `convert_index*` methods into something like...
```
convert_index_glb_to_loc
```
and 
```
convert_index_loc_to_glb
```

or maybe something else...as long as the symmetry between the two is clear

devito/data/data.py
```diff
@@ -26,6 +27,8 @@ class Data(np.ndarray):
         If the i-th entry is True, then the i-th array dimension uses modulo indexing.
     allocator : MemoryAllocator, optional
         Used to allocate memory. Defaults to ``ALLOC_FLAT``.
+    distributor : Distributor, optional
+        The distributor attached to the grid on which the Data object sits.
```
FabioLuporini: shouldn't talk about "grid" here . There's no concept of grid in ``data``

devito/data/data.py
```diff
@@ -38,14 +41,16 @@ class Data(np.ndarray):
...
+    def __new__(cls, shape, dtype, decomposition=None, modulo=None, allocator=ALLOC_FLAT,
+                distributor=None):
         assert len(shape) == len(modulo)
         ndarray, memfree_args = allocator.alloc(shape, dtype)
         obj = np.asarray(ndarray).view(cls)
         obj._allocator = allocator
         obj._memfree_args = memfree_args
         obj._decomposition = decomposition or (None,)*len(shape)
         obj._modulo = modulo or (False,)*len(shape)
+        obj._distributor = distributor
```
FabioLuporini: reminder: explain with a comment why/when ``decomposition`` is different than ``distributor.decomposition``

devito/data/data.py
```diff
@@ -132,16 +140,76 @@ def _global(self, glb_idx, decomposition):
         ret._is_distributed = any(i is not None for i in decomposition)
         return ret
 
+    def check_slicing(func):
+        """Check if the operation will communication across MPI ranks."""
```
FabioLuporini: will trigger communication?

instead of "operation" you can be more specific -- ``__getitem__`` or ``__setitem__`` right?

devito/data/data.py
```diff
@@ -132,16 +140,76 @@ def _global(self, glb_idx, decomposition):
...
+                self._mpi_index_maps(loc_idx, local_val.shape, comm, rank)
+
+            retval = Data(local_val.shape, local_val.dtype.type,
+                          decomposition=local_val._decomposition,
+                          modulo=local_val._modulo,
+                          distributor=local_val._distributor)
+            it = np.nditer(owners, flags=['refs_ok', 'multi_index'])
+            while not it.finished:
+                index = it.multi_index
+                if rank == owners[index] and rank == send[index]:
```
FabioLuporini: Add comment within each ``if`` ``elif`` ``else`` of this block explaining what is exactly going on?

devito/data/data.py
```diff
@@ -132,16 +140,76 @@ def _global(self, glb_idx, decomposition):
...
     @property
     def _is_mpi_distributed(self):
         return self._is_distributed and configuration['mpi']
 
     def __repr__(self):
         return super(Data, self._local).__repr__()
 
-    def __getitem__(self, glb_idx):
+    @check_slicing
+    def __getitem__(self, glb_idx, mpi_slicing=False):
```
FabioLuporini: instead of ``mpi_slicing`` I'd use ``mode`` (being a string such as ''mpi-slicing", or better an object of type ``Tag``). This would be easier to generalise, I suppose

devito/data/data.py
```diff
@@ -188,15 +278,26 @@ def __setitem__(self, glb_idx, val):
                 # * one of them is 1
                 # Conceptually, below we apply the same rule
                 val_idx = val_idx[len(val_idx)-val.ndim:]
+                processed = []
+                # Handle step size > 1
+                for i, j in zip(glb_idx, val_idx):
+                    if isinstance(i, slice) and i.step is not None and i.step > 1 and \
```
FabioLuporini: I've seen this pattern many times:

``` if isinstance(i, slice) and i.step is not None and i.step > ...```

maybe we should add a free function ``is_step_gt`` (ie, is step greater than) so that you can do

```
if is_step_gt(i, 1) and j.stop > j.start
```

or something along these lines.

Otherwise, four conditions in an ``if`` becomes a bit unreadable

File: devito/data/data.py (hunk 1 above)
Comment (FabioLuporini) [reply devito/data/data.py]: Feels like my previous comment is obsolete and should now be ignored ?

File: devito/data/data.py [hunk 14]
```diff
@@ -210,13 +311,40 @@ def _normalize_index(self, idx):
...
             else:
-                return idx + (slice(None),)*(self.ndim - len(idx))
+                items = idx + (slice(None),)*(self.ndim - len(idx))
+            # Normalize slice steps:
+            processed = [slice(i.start, i.stop, 1) if
+                         (isinstance(i, slice) and i.step is None)
+                         else i for i in items]
+            return as_tuple(processed)
+
+    def _process_args(self, idx, val):
```
FabioLuporini: this could be a free function (self isn't actually used). 

A docstring would help

devito/data/data.py
```diff
@@ -260,6 +388,176 @@ def _convert_index(self, glb_idx):
 
         return loc_idx[0] if len(loc_idx) == 1 else tuple(loc_idx)
 
+    def _set_global_idx(self, val, idx, val_idx):
```
FabioLuporini: this could be a free function (self isn't actually used).

devito/data/data.py
```diff
@@ -260,6 +388,176 @@ def _convert_index(self, glb_idx):
...
+                if isinstance(j, slice) and j.step is not None:
+                    mapped_idx.append(slice(j.step*i.start+norm,
+                                            j.step*i.stop+norm, j.step))
+                else:
+                    mapped_idx.append(slice(i.start+norm, i.stop+norm, i.step))
+            else:
+                mapped_idx.append(None)
+        return as_tuple(mapped_idx)
+
+    def _local_data_idx(self, loc_idx):
```
FabioLuporini: this could be a free function (self isn't actually used).

devito/data/data.py
```diff
@@ -260,6 +388,176 @@ def _convert_index(self, glb_idx):
...
+                    loc_data_idx.append(slice(lmin-i.step, i.start+1, -i.step))
+                else:
+                    loc_data_idx.append(slice(i.stop+1, i.start+1, -i.step))
+            elif is_integer(i):
+                loc_data_idx.append(slice(i, i+1, 1))
+            else:
+                loc_data_idx.append(i)
+        return as_tuple(loc_data_idx)
+
+    def _mpi_index_maps(self, loc_idx, val_shape, comm, rank):
```
FabioLuporini: this could be a free function 

self isn't actually necessary:
* nprocs -> can be obtained from ``comm``
* topology and all_cords can be passed explicitly

devito/data/decomposition.py
```diff
@@ -230,9 +230,21 @@ def convert_index(self, *args, rel=True):
                 elif isinstance(glb_idx, slice):
                     if self.loc_empty:
                         return slice(-1, -3)
-                    glb_idx_min = self.glb_min if glb_idx.start is None else glb_idx.start
-                    glb_idx_max = self.glb_max if glb_idx.stop is None else glb_idx.stop-1
-                    retfunc = lambda a, b: slice(a, b + 1, glb_idx.step)
+                    # NOTE: step should not be None at this point
```
FabioLuporini: this method name should probably be ``convert_glb_to_loc``, while the dual one could be ``convert_loc_to_glb``

devito/data/decomposition.py
rhodrin: I've gone with `_convert_index_loc` and `_convert_index_glb` for now?

devito/data/data.py
rhodrin: Updated.

devito/data/data.py
rhodrin: > will trigger communication?
Yes, this will trigger at least some communication.

devito/data/data.py
```diff
@@ -132,16 +141,82 @@ def _global(self, glb_idx, decomposition):
...
+    @check_slicing
+    def __getitem__(self, glb_idx, slicing):
+        loc_idx = self._convert_index_loc(glb_idx)
+        if slicing is MPI:
+            # Retrieve the pertinent local data prior to mpi send/receive operations
+            loc_data_idx = local_data_idx(loc_idx)
+            local_val = super(Data, self).__getitem__(loc_data_idx)
+
+            comm = self._distributor.comm
+            rank = self._distributor.myrank
```
mloubout: Maybe make these property if used a lot rather than access _distriubutor.somethign all the time and just do self.rank, self.comm,...

devito/data/data.py
```diff
@@ -365,3 +560,151 @@ def index_handle_oob(idx):
...
+        elif is_integer(i):
+            loc_data_idx.append(slice(i, i+1, 1))
+        else:
+            loc_data_idx.append(i)
+    return as_tuple(loc_data_idx)
+
+
+class SLICING(Tag):
+    pass
+MPI = SLICING('mpi')  # noqa
```
mloubout: Can we rename that, we already have MPI in devito so maybe confusing.

like distributed_slice and serial_slice or something

tests/test_data.py
```diff
@@ -63,6 +63,17 @@ def test_advanced_indexing(self):
...
 
+    @skipif('yask')
+    def test_negative_step(self):
+        """Test slicing with a negative step."""
+        grid = Grid(shape=(6, 6, 6))
+        u = TimeFunction(name='u', grid=grid, dtype=np.int32)
+        u.data[:] = 0.
+        dat = np.array([1, 2, 3, 4, 5, 6])
+        u.data[0, :, 0, 0] = dat
+        assert (np.array(u.data[0, 3::-1, 0, 0]) == dat[3::-1]).all()
```
mloubout: Wouldn't the test be more "improved" as 
`u.data[0, 3::-1, 0, 0] = dat[3::-1]`
`assert (np.array(u.data[0, :, 0, 0])  == dat[:]).all()`

devito/data/data.py
```diff
@@ -132,16 +141,82 @@ def _global(self, glb_idx, decomposition):
         ret._is_distributed = any(i is not None for i in decomposition)
         return ret
 
+    def check_slicing(func):
+        """Check if __getitem__/__setitem__ may require communication across MPI ranks."""
```
FabioLuporini: does this PR support the case
```
a[0] = a[20]
```
ie, doing the right thing regardless of whether `0` and `20` belong to the same of different processes?

if yes, then maybe use another name instead of ``check_slicing`` ?

FabioLuporini: also, it should be a private method

devito/data/data.py
FabioLuporini: seconded

devito/data/data.py
```diff
@@ -365,3 +560,151 @@ def index_handle_oob(idx):
...
+    return as_tuple(loc_data_idx)
+
+
+class SLICING(Tag):
+    pass
+MPI = SLICING('mpi')  # noqa
+STANDARD = SLICING('standard')
+
+
+def mpi_index_maps(loc_idx, val_shape, topology, rank_coords, comm):
```
FabioLuporini: this free function, and perhaps many of the other free functions from this file, should probably be moved to a different file

``rank_coords`` for homogeneity should probably just be ``coords``

how is ``val_shape`` different than ``shape``?

devito/data/data.py
```diff
@@ -365,3 +560,151 @@ def index_handle_oob(idx):
...
+    indices = np.zeros(global_size, dtype=tuple)
+    global_si = np.zeros(global_size, dtype=tuple)
+    it = np.nditer(indices, flags=['refs_ok', 'multi_index'])
+    while not it.finished:
+        index = it.multi_index
+        indices[index] = index
+        it.iternext()
+    global_si[:] = indices[transform]
+
+    # create the 'rank' slices
```
FabioLuporini: Create

devito/data/decomposition.py
```diff
@@ -122,10 +122,10 @@ def __repr__(self):
         return 'Decomposition(%s)' % ', '.join(ret)
 
     def __call__(self, *args, rel=True):
-        """Alias for ``self.convert_index``."""
-        return self.convert_index(*args, rel=rel)
+        """Alias for ``self.convert_index_loc``."""
+        return self.convert_index_loc(*args, rel=rel)
 
-    def convert_index(self, *args, rel=True):
+    def convert_index_loc(self, *args, rel=True):
```
FabioLuporini: I honestly prefer to see the "direction" of the conversion reflected in the method name

something like ``convert_glb_to_loc`` or simply ``glb_to_loc`` (and then ``loc_to_glb``)

devito/mpi/distributed.py
```diff
@@ -375,6 +375,9 @@ def _obj_neighborhood(self):
         """
         return MPINeighborhood(self.neighborhood)
 
+    def create_new(self, shape, dimensions, comm):
```
FabioLuporini: should be `_rebuild`

FabioLuporini: also, the parameters should be kwargs such that you reuse self's if not provided

```
def rebuild(self, shape=None, ...):
    return Distributor(shape or self.shape, ...)
```

devito/types/dense.py
```diff
@@ -111,8 +111,14 @@ def _allocate_memory(func):
         def wrapper(self):
             if self._data is None:
                 debug("Allocating memory for %s%s" % (self.name, self.shape_allocated))
+                try:
+                    distributor = self._grid._distributor
+                except AttributeError:
+                    # Data not attached to a grid
```
FabioLuporini: I don't understand this try-catch? what case is it covering?

devito/data/data.py
```diff
@@ -26,6 +27,9 @@ class Data(np.ndarray):
         If the i-th entry is True, then the i-th array dimension uses modulo indexing.
     allocator : MemoryAllocator, optional
         Used to allocate memory. Defaults to ``ALLOC_FLAT``.
+    distributor : Distributor, optional
+        The distributor attached to the original Function from which any subsequent
```
FabioLuporini: in an ideal world, we avoid to talk about "Function" (and any other devito concept) in here

devito/data/data.py
```diff
@@ -210,13 +316,56 @@ def _normalize_index(self, idx):
...
+                    else:
+                        start = j.stop + 1
+                    processed.append(slice(start, stop, -j.step))
+                    transform.append(slice(None, None, np.sign(j.step)))
+                else:
+                    processed.append(j)
+            if isinstance(val, Data) and len(transform) > 0 and \
+                    len(val._distributor.shape) > len(val.shape):
+                distributor = \
+                    val._distributor.create_new(val.shape,
```
FabioLuporini: see my comment near ``def create_new``

devito/data/data.py
rhodrin: Sure, but since `_distributor` can be `None` these properties will need a try, except which is not normally needed since they're currently only accessed when we know `distributor` is not `None`. Can change if you'd like though?

devito/data/data.py
rhodrin: How about `PARALLEL_SLICE` and `SERIAL_SLICE` or just `PARALLEL`/`SERIAL`?

tests/test_data.py
rhodrin: The 'full' length of dim 1 of `u.data` is not being set in your example and hence there's no reason for
`assert (np.array(u.data[0, :, 0, 0]) == dat[:]).all()`
to be true.

The test is meant to test cases when a negative slice is not starting from the end point so I think is fine as is?

devito/data/data.py
rhodrin: Yes, `a[0] = a[20]` is supported.  Hmm, could change to `check_idx`?

devito/data/data.py
rhodrin: `utils.py`?

devito/data/decomposition.py
rhodrin: Currently the direction is 'implied' which I think is fine - but what about `idx_glb_to_loc` and `idx_loc_to_glb`?

rhodrin:
> I think the only thing that might need some thinking is the fact that you're passing both a decomposition and a distributor to Data. It is odd because in theory a Decomposition is attached to a Distributor, so Data should get it from there.

So currently the `decomposition` passed to `Data` is not always the same as that stored in `distributor.decomposition` (since `Data` doesn't always have an associated `distributor`). However, when `distributor` is needed they will be the same. So, I think we need the option to pass both but in the case a `distributor` is being passed we should get `decomposition` from that and throw a warning/error if a `decomposition` is also being passed?

devito/data/data.py
rhodrin: It's the `local_val.shape` which can be different from `self.shape`. The naming is a relic from when this was a class method (since 'shape' could easily be confused with `self.shape`) but since it's now a free function the `val_` can be removed.

devito/types/dense.py
rhodrin: Can't remember why it was there originally but is not needed and has been removed.

devito/data/data.py
```diff
@@ -132,16 +142,82 @@ def _global(self, glb_idx, decomposition):
...
+            elif data._is_mpi_distributed:
+                for i in as_tuple(glb_idx):
+                    if isinstance(i, slice) and i.step is not None and i.step < 0:
+                        slicing = PARALLEL
+                        break
+                    else:
+                        slicing = SERIAL
+            else:
+                slicing = SERIAL
+            kwargs['slicing'] = slicing
```
FabioLuporini: slicing -> 'needs_comm' ?

FabioLuporini: here and everywhere below in getitem/setitem

devito/data/data.py
FabioLuporini: also, I'm not sure I like it be called "_set" which makes one think it's changing internal state?

devito/data/data.py
```diff
@@ -260,108 +410,49 @@ def _convert_index(self, glb_idx):
...
+                mapped_idx.append(None)
+        return as_tuple(mapped_idx)
+
     def reset(self):
         """Set all Data entries to 0."""
         self[:] = 0.0
 
 
-class Index(Tag):
+class SLICING(Tag):
```
FabioLuporini: need to change SLICING probably...?

devito/data/decomposition.py
```diff
@@ -270,6 +316,96 @@ def convert_index(self, *args, rel=True):
         else:
             raise TypeError("Expected 1 or 2 arguments, found %d" % len(args))
 
+    def index_loc_to_glb(self, *args):
+        """
+        Convert a local index into a global index.
```
FabioLuporini: blank line before Parameters

devito/data/decomposition.py
```diff
@@ -270,6 +316,96 @@ def convert_index(self, *args, rel=True):
...
+            There are two possible cases:
+            * int. Given ``I``, a local index, return the corresponding
+              global local.
+            * slice(a, b, c). As above, return the corresponding global slice.
+
+        Raises
+        ------
+        TypeError
+            If the input doesn't adhere to any of the supported format.
+        Examples
```
FabioLuporini: blank line

devito/data/utils.py
```diff
@@ -0,0 +1,259 @@
...
+
+__all__ = ['Index', 'NONLOCAL', 'PROJECTED', 'index_is_basic', 'index_apply_modulo',
+           'index_dist_to_repl', 'convert_index', 'index_handle_oob',
+           'local_data_idx', 'mpi_index_maps']
+
+
+class Index(Tag):
+    pass
+NONLOCAL = Index('nonlocal')  # noqa
+PROJECTED = Index('projected')
```
FabioLuporini: I wonder whether we can use this class for PARALLEL and SERIAL too ?

to be fair, I'm not sure I remember how I was using it, so I might be confusing things and talking BS...

devito/data/utils.py
```diff
@@ -0,0 +1,259 @@
...
+            return np.delete(idx, np.where(idx == None))  # noqa
+        else:
+            raise ValueError("Cannot identify OOB accesses when using "
+                             "multidimensional index arrays")
+    else:
+        return idx
+
+
+def local_data_idx(loc_idx):
+    """Retrieve the unflipped idx corresponding to loc_idx."""
```
FabioLuporini: can you elaborate on what you mean by 'unflipped' here? 

either here or where it's used for the first time as jargon

devito/data/utils.py
```diff
@@ -0,0 +1,259 @@
...
+                loc_data_idx.append(slice(i.stop+1, i.start+1, -i.step))
+        elif is_integer(i):
+            loc_data_idx.append(slice(i, i+1, 1))
+        else:
+            loc_data_idx.append(i)
+    return as_tuple(loc_data_idx)
+
+
+def mpi_index_maps(loc_idx, shape, topology, coords, comm):
+    """Generate the various mpi index maps."""
```
FabioLuporini: what is an index map? I could use a proper example here

devito/data/utils.py
FabioLuporini: do we have a test specifically for this function? or is it just tested indirectly?

devito/data/utils.py
rhodrin: Hmm, would be nice but I don't think we can. Will have a look though.

devito/data/utils.py
rhodrin: Yes, this has a specific test. Will elaborate on 'unflipped'.

devito/data/utils.py
rhodrin: Will elaborate. But basically some structures are created which tell us to which index `(i, j, k, ...)` should go i.e. `(i, j, k, ...) -> (I, J, K, ...)`.

devito/data/data.py
rhodrin: `comm_on`? :p

FabioLuporini: not sure if it's an english joke or ...? but this really got my attention :joy:

devito/data/data.py
rhodrin: `COMM_TYPE`?

devito/data/utils.py
```diff
@@ -0,0 +1,325 @@
...
+            loc_data_idx.append(i)
+    return as_tuple(loc_data_idx)
+
+
+def mpi_index_maps(loc_idx, shape, topology, coords, comm):
+    """
+    Generate various data structures used to determine what MPI communication
+    is required. The function creates the following:
+
+    owners: An array of shape shape where each index signifies the rank on which
```
FabioLuporini: of shape \`\`shape\'\' ?

FabioLuporini: same below

devito/data/utils.py
```diff
@@ -0,0 +1,325 @@
...
+    send: An array of shape shape where each index signifies the rank to which
+    data beloning to that index should be sent.
+
+    global_si: An array of shape shape where each index contains the global index
+    to which that index should be sent.
+
+    local_si: An array of shape shape where each index contains the local index
+    (on the destination rank) to which that index should be sent.
+
+    Consider the following example. An array is given by A = [[  0,  1,  2,  3],
```
FabioLuporini: This could be put under

```

Examples
--------
...
```
?

devito/data/utils.py
```diff
@@ -0,0 +1,325 @@
...
+        return idx
+
+
+def local_data_idx(loc_idx):
+    """
+    Return tuple of slices containing the unflipped idx corresponding to loc_idx.
+    By 'unflipped' we mean that if a slice has a negative step, we wish to retrieve
+    the corresponding indices but not in reverse order. i.e.
+
+    >>> local_data_idx(slice(11, None, -3)) = (slice(2, 12, 3))
```
FabioLuporini: to be put under 


```

Examples
--------
...
```

Then, we should/could test the docstring via ``test_docstrings.py``

devito/data/decomposition.py
```diff
@@ -230,9 +239,18 @@ def convert_index(self, *args, rel=True):
                 elif isinstance(glb_idx, slice):
                     if self.loc_empty:
                         return slice(-1, -3)
-                    glb_idx_min = self.glb_min if glb_idx.start is None else glb_idx.start
-                    glb_idx_max = self.glb_max if glb_idx.stop is None else glb_idx.stop-1
-                    retfunc = lambda a, b: slice(a, b + 1, glb_idx.step)
+                    if glb_idx.step >= 0:
+                        glb_idx_min = self.glb_min if glb_idx.start is None \
```
FabioLuporini: pity we have to break all these lines

devito/data/decomposition.py
```diff
@@ -241,21 +259,49 @@ def convert_index(self, *args, rel=True):
...
                 elif glb_idx_min > self.loc_abs_max:
                     return retfunc(-1, -3)
                 else:
                     loc_min = glb_idx_min - base
-                if glb_idx_max is None or glb_idx_max > self.loc_abs_max:
+                if isinstance(glb_idx, slice) and glb_idx.step is not None \
+                        and glb_idx.step < -1:
+                    if glb_idx_max < self.loc_abs_min:
+                        return retfunc(-1, -3)
+                    elif glb_idx.start is None:
```
FabioLuporini: I could use a comment here? or maybe assign ``np.mod(....)`` to a variable with a name such that I can vaguely grasp what's going on?

devito/data/data.py
FabioLuporini: yep, now it's OK

devito/data/data.py
```diff
@@ -132,16 +142,82 @@ def _global(self, glb_idx, decomposition):
...
+        @wraps(func)
+        def wrapper(data, *args, **kwargs):
+            glb_idx = args[0]
+            if len(args) > 1 and isinstance(args[1], Data) \
+                    and args[1]._is_mpi_distributed:
+                comm_type = PARALLEL
+            elif data._is_mpi_distributed:
+                for i in as_tuple(glb_idx):
+                    if isinstance(i, slice) and i.step is not None and i.step < 0:
+                        comm_type = PARALLEL
```
FabioLuporini: I think I still prefer a boolean "needs_comm" over PARALLEL vs SERIAL, which is not btw what the docstring claims to check?

devito/data/data.py
```diff
@@ -132,16 +142,82 @@ def _global(self, glb_idx, decomposition):
...
+
+            comm = self._distributor.comm
+            rank = comm.Get_rank()
+
+            owners, send, global_si, local_si = \
+                mpi_index_maps(loc_idx, local_val.shape, self._distributor.topology,
+                               self._distributor.all_coords, comm)
+
+            retval = Data(local_val.shape, local_val.dtype.type,
+                          decomposition=local_val._decomposition,
```
FabioLuporini: Could you clarify with a comment why are you passing in this ``local_val._decomposition``  ?

devito/data/data.py
FabioLuporini: reminder

devito/data/data.py
```diff
@@ -260,108 +410,49 @@ def _convert_index(self, glb_idx):
...
+        # Convert integers to slices so that shape dims are preserved
+        if is_integer(as_tuple(idx)[0]):
+            data_global_idx.append(slice(0, 1, 1))
+        for i, j in zip(data_loc_idx, val._decomposition):
+            if not j.loc_empty:
+                data_global_idx.append(j.index_loc_to_glb(i))
+            else:
+                data_global_idx.append(None)
+        mapped_idx = []
+        for i, j, k in zip(data_global_idx, as_tuple(idx), self._decomposition):
```
FabioLuporini: I could use a comment

devito/data/data.py
rhodrin: Docstring says:
```
"""Check if __getitem__/__setitem__ may require communication across MPI ranks."""
```
?

So you want a boolean `needs_comm` and remove the `class COMM_TYPE(Tag)`? If so, this is exactly how the code was previously, albeit with different names.

devito/data/data.py
rhodrin: Numpy `__getitem__` loops though each element. So after the 'parallel' `__getitem__` there'll be a load of serial `__getitem__` calls in which we'll still need to work out the `loc_idx` which needs the decomposition.

devito/data/utils.py
rhodrin: Amended.

devito/data/utils.py
rhodrin: Done.

devito/data/utils.py
rhodrin: Done + added to `test_docstring.py`.

devito/data/decomposition.py
rhodrin: Comments added.

devito/data/data.py
rhodrin: Brief comment added.

devito/data/utils.py
FabioLuporini: missing this line and the one corresponding to ``local_si`` ?

rhodrin: I'm blind. Thanks!

devito/data/data.py
```diff
@@ -210,13 +316,56 @@ def _normalize_index(self, idx):
...
+            else:
+                items = idx + (slice(None),)*(self.ndim - len(idx))
+            # Normalize slice steps:
+            processed = [slice(i.start, i.stop, 1) if
+                         (isinstance(i, slice) and i.step is None)
+                         else i for i in items]
+            return as_tuple(processed)
+
+    def _process_args(self, idx, val):
+        """If comm_type is parallel we need to first retrieve local unflipped data."""
```
FabioLuporini: %s/parallel/index_by_index

devito/data/data.py
```diff
@@ -210,13 +316,56 @@ def _normalize_index(self, idx):
...
+                    else:
+                        start = j.stop + 1
+                    processed.append(slice(start, stop, -j.step))
+                    transform.append(slice(None, None, np.sign(j.step)))
+                else:
+                    processed.append(j)
+            if isinstance(val, Data) and len(transform) > 0 and \
+                    len(val._distributor.shape) > len(val.shape):
+                distributor = \
+                    val._distributor._rebuild(val.shape,
```
FabioLuporini: I could use a comment here -- why is rebuilding necessary?

devito/data/data.py
```diff
@@ -210,13 +316,56 @@ def _normalize_index(self, idx):
...
+                        start = int(np.mod(j.start, -j.step))
+                    else:
+                        start = j.stop + 1
+                    processed.append(slice(start, stop, -j.step))
+                    transform.append(slice(None, None, np.sign(j.step)))
+                else:
+                    processed.append(j)
+            if isinstance(val, Data) and len(transform) > 0 and \
+                    len(val._distributor.shape) > len(val.shape):
+                distributor = \
```
FabioLuporini: nitpicking: don't need the \ ? you can inline the rhs here and split each argument over multiple lines

devito/data/data.py
FabioLuporini: reminder. We should really drop the _set

FabioLuporini: also: I **think** ``val_idx`` should actually be ``glb_idx`` ?

devito/data/data.py
```diff
@@ -260,108 +409,54 @@ def _convert_index(self, glb_idx):
...
+        # Convert integers to slices so that shape dims are preserved
+        if is_integer(as_tuple(idx)[0]):
+            data_global_idx.append(slice(0, 1, 1))
+        for i, j in zip(data_loc_idx, val._decomposition):
+            if not j.loc_empty:
+                data_global_idx.append(j.index_loc_to_glb(i))
+            else:
+                data_global_idx.append(None)
+        mapped_idx = []
+        # Based data_global_idx the indices to which the locally stored data
```
FabioLuporini: nitpicking (for homogeneity with rest of devito): use backtips for variable names in comments, ie \`data_global_idx\`

also, it should be "glb", not "global", again for consistency/homogeneity ?

devito/data/utils.py
```diff
@@ -0,0 +1,330 @@
...
+        elif idx.ndim == 1:
+            return np.delete(idx, np.where(idx == None))  # noqa
+        else:
+            raise ValueError("Cannot identify OOB accesses when using "
+                             "multidimensional index arrays")
+    else:
+        return idx
+
+
+def local_data_idx(loc_idx):
```
FabioLuporini: 'loc' instead of 'local' for consistency?

devito/data/utils.py
```diff
@@ -0,0 +1,330 @@
...
+    Return tuple of slices containing the unflipped idx corresponding to loc_idx.
+    By 'unflipped' we mean that if a slice has a negative step, we wish to retrieve
+    the corresponding indices but not in reverse order.
+
+    Examples
+    --------
+    >>> local_data_idx(slice(11, None, -3))
+    (slice(2, 12, 3),)
+    """
+    loc_data_idx = []
```
FabioLuporini: you can use a generic 'retval' here if this becomes the function name

devito/data/utils.py
```diff
@@ -0,0 +1,330 @@
...
+            loc_data_idx.append(i)
+    return as_tuple(loc_data_idx)
+
+
+def mpi_index_maps(loc_idx, shape, topology, coords, comm):
+    """
+    Generate various data structures used to determine what MPI communication
+    is required. The function creates the following:
+
+    owners: An array of shape ``shape'' where each index signifies the rank on which
```
FabioLuporini: should be backtips on both sides of the word \`\`shape\`\`

devito/data/utils.py
```diff
@@ -0,0 +1,330 @@
...
+    owners = np.zeros(global_size, dtype=np.int32)
+    send = np.zeros(global_size, dtype=np.int32)
+    for i in range(len(n_rank_slice)):
+        if any([j is None for j in n_rank_slice[i]]):
+            continue
+        else:
+            owners[n_rank_slice[i]] = i
+    send[:] = owners[transform]
+
+    # local_indices
```
FabioLuporini: Can you elaborate -- local_indices what? you mean "Create local indices" ?

tests/test_interpolation.py
```diff
@@ -126,8 +126,9 @@ def test_precomputed_interpolation_time():
     op = Operator(eqn)
     op(time_m=0, time_M=4)
 
+    # NOTE: Work out why this change was needed before any merge
```
FabioLuporini: ?

tests/test_data.py
```diff
@@ -188,6 +200,22 @@ def test_indexing_into_sparse(self):
         sf.data[1:-1, 0] = np.arange(8)
         assert np.all(sf.data[1:-1, 0] == np.arange(8))
 
+    @pytest.mark.parametrize('idx, expected', [
+        ('(slice(10, None, -1), slice(11, None, -3))',
+         '(slice(0, 11, 1), slice(2, 12, 3))'),
+        ('(2, 5)', '(slice(2, 3, 1), slice(5, 6, 1))')
+    ])
+    def test_local_data_idx(self, idx, expected):
```
FabioLuporini: This probably needs its own class, as it's a test for a support function, not for an actual Data

tests/test_data.py
```diff
@@ -284,28 +312,48 @@ class TestDecomposition(object):
...
-        assert d.convert_index((1, 6), rel=False) == (5, 6)
-        assert d.convert_index((None, None), rel=False) == (5, 7)
+        assert d.index_glb_to_loc((5, 7), rel=False) == (5, 7)
+        assert d.index_glb_to_loc((5, 9), rel=False) == (5, 7)
+        assert d.index_glb_to_loc((1, 3), rel=False) == (-1, -3)
+        assert d.index_glb_to_loc((1, 6), rel=False) == (5, 6)
+        assert d.index_glb_to_loc((None, None), rel=False) == (5, 7)
+
+        # Convert local indices to global indices
+        assert d.index_loc_to_glb((0, 2)) == (5, 7)
```
FabioLuporini: The loc_to_glb test should be a separate unit test

tests/test_data.py
```diff
@@ -500,6 +548,366 @@ def test_slicing(self):
             assert np.all(u.data[2:, 2:] == myrank)
             assert u.data[:2, 2:].size == u.data[2:, :2].size == u.data[:2, :2].size == 0
 
+        # Test slicing with a negative step
```
FabioLuporini: this is huge. We need small unit tests, or better, as small as reasonably possible

this should then be split over multiple tests

FabioLuporini: I just looked at the entire test :open_mouth:  "over many! multiple tests" :)

devito/data/data.py
rhodrin: Comment added. The dimension of the slice is different to that of the original data and therefore the original distributor has 1+ superfluous dimensions.

devito/data/data.py
rhodrin: I think it is needed no?

devito/data/utils.py
rhodrin: It's constructing `local_si` which has been described in the docsting. Changed comment to be more appropriate.

tests/test_interpolation.py
rhodrin: I remember I needed to change this assert as it started to break. Could have been a python update and the same change was also needed in master?

PR #828, opened 2019-05-09
restore _lib when unpickling Operator

tjb900:
There are really two changes here:

* save `soname` in `__getstate__`, so that the matching logic in `__setstate__` is executed
* add loading of the library as part of `__setstate__`, so that `_compile` doesn't recompile it after unpickling

### DIFF
--- a/devito/operator.py
+++ b/devito/operator.py
@@ -577,7 +577,7 @@ class Operator(Callable):
     def __getstate__(self):
         if self._lib:
             state = dict(self.__dict__)
-            state.pop('_soname')
+            # state.pop('_soname')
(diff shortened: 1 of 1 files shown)


devito/operator.py
```diff
@@ -577,7 +577,7 @@ def __call__(self, **kwargs):
     def __getstate__(self):
         if self._lib:
             state = dict(self.__dict__)
-            state.pop('_soname')
+            # state.pop('_soname')
```
FabioLuporini: reminder for myself: drop this

FabioLuporini:
Merged.

PR #829, opened 2019-05-09
pickle TimeFunction.time_dim

tjb900:
Name says it all, really.  Without this, I think that TimeFunctions declared over non-standard time dims (e.g. ConditionalDimensions) are not unpickled correctly.


FabioLuporini:
uncontroversial

FabioLuporini:
Merged.

PR #830, opened 2019-05-09
knl detection fix

tjb900:
Trivial fixes to KNL flat mode detection.  The "is KNL" doesn't work anymore because there is a second definition of the KNL7210 platform now, which is a distinct object.


FabioLuporini:
uncontroversial

FabioLuporini:
some pep8 violations here (due to uselessly importing KNL)

FabioLuporini:
merged.

PR #832, opened 2019-05-10
Lifting expressions in OPS kernels

vincepandolfo:
Same as #823 but with fixes from review applied

### DIFF
--- a/devito/ir/iet/visitors.py
+++ b/devito/ir/iet/visitors.py
@@ -317,11 +317,6 @@ class CGen(Visitor):
                                                     self._args_decl(i.root.parameters)))
                 efuncs.extend([i.root.ccode, blankline])
 
-        for i in o._callables:
-            esigns.append(c.FunctionDeclaration(c.Value(i.retval, i.name),
-                                                self._args_decl(i.parameters)))
-            efuncs.extend([i.ccode, blankline])
-
         # Header files, extra definitions, ...
         header = [c.Line(i) for i in o._headers]
         includes = [c.Include(i, system=False) for i in o._includes]
--- a/devito/ir/iet/visitors.py
+++ b/devito/ir/iet/visitors.py
@@ -152,7 +152,9 @@ class CGen(Visitor):
         """Generate cgen declarations from an iterable of symbols and expressions."""
         ret = []
         for i in args:
-            if i.is_Tensor:
+            if i.is_Tensor and i.is_OPS:
+                ret.append(c.Value(i._C_typename, i.name))
+            elif i.is_Tensor:
                 ret.append(c.Value('%srestrict' % i._C_typename, i._C_name))
             elif i.is_AbstractObject or i.is_Symbol:
                 ret.append(c.Value(i._C_typename, i._C_name))
@@ -314,6 +316,11 @@ class CGen(Visitor):
                 esigns.append(c.FunctionDeclaration(c.Value(i.root.retval, i.root.name),
                                                     self._args_decl(i.root.parameters)))
                 efuncs.extend([i.root.ccode, blankline])
+
+        for i in o._callables:
+            esigns.append(c.FunctionDeclaration(c.Value(i.retval, i.name),
+                                                self._args_decl(i.parameters)))
+            efuncs.extend([i.ccode, blankline])
 
         # Header files, extra definitions, ...
         header = [c.Line(i) for i in o._headers]
--- a/devito/ir/iet/visitors.py
+++ b/devito/ir/iet/visitors.py
@@ -152,12 +152,10 @@ class CGen(Visitor):
         """Generate cgen declarations from an iterable of symbols and expressions."""
         ret = []
         for i in args:
-            if i.is_Tensor and i.is_OPS:
-                ret.append(c.Value(i._C_typename, i.name))
+            if i.is_AbstractObject or i.is_Symbol:
+                ret.append(c.Value(i._C_typename, i._C_name))
             elif i.is_Tensor:
                 ret.append(c.Value('%srestrict' % i._C_typename, i._C_name))
-            elif i.is_AbstractObject or i.is_Symbol:
-                ret.append(c.Value(i._C_typename, i._C_name))
             else:
                 ret.append(c.Value('void', '*_%s' % i._C_name))
         return ret
--- a/devito/ops/node_factory.py
+++ b/devito/ops/node_factory.py
@@ -18,7 +18,7 @@ class OPSNodeFactory(object):
     def __init__(self):
         self.ops_args = OrderedDict()
 
-    def new_ops_arg(self, indexed):
+    def new_ops_arg(self, indexed, is_Write):
         """
         Create an :class:`Indexed` node using OPS representation.
 
@@ -39,7 +39,8 @@ class OPSNodeFactory(object):
 
         if ops_arg_id not in self.ops_args:
             # Create the indexed object
-            ops_arg = Array(name=ops_arg_id,
+            ops_arg = Array(is_Write,
+                            name=ops_arg_id,
                             dimensions=[Dimension(name=namespace['ops_acc'])],
                             dtype=indexed.dtype)
 
--- a/devito/ops/node_factory.py
+++ b/devito/ops/node_factory.py
@@ -1,9 +1,10 @@
 from collections import OrderedDict
 
 from devito import Dimension, TimeFunction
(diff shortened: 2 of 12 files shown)


devito/ops/types.py
```diff
@@ -0,0 +1,15 @@
+from devito.tools import Signer
+
+import devito.types.basic as basic
+
+__all__ = ['Array']
+
+
+class Array(basic.Array, Signer):
```
FabioLuporini: why does it have to a Signer too?

devito/ops/types.py
```diff
@@ -0,0 +1,15 @@
...
+
+    from_OPS = True
+    is_Symbol = True
```
FabioLuporini: the entire PR looks good to go to me except for this ``is_Symbol`` -- an Array definitely is not a symbol! Why do you need it ? Maybe we can find a solution by introducing a new subclass...

devito/ir/iet/visitors.py
```diff
@@ -152,10 +152,10 @@ def _args_decl(self, args):
         """Generate cgen declarations from an iterable of symbols and expressions."""
         ret = []
         for i in args:
-            if i.is_Tensor:
-                ret.append(c.Value('%srestrict' % i._C_typename, i._C_name))
-            elif i.is_AbstractObject or i.is_Symbol:
+            if i.is_AbstractObject or i.is_Symbol:
```
FabioLuporini: I guess there's a rational behind this flipping...?

vincepandolfo: Same reason `Array` has `is_Symbol` set to `True`.

I want OPS `Array` types to fall in that if clause and not the `i.is_Tensor` case

devito/ops/types.py
vincepandolfo: leftover from old code, will remove

devito/ir/iet/visitors.py
FabioLuporini: are you just trying to avoid the `restrict` ?

vincepandolfo: yes

tests/test_ops.py
```diff
@@ -71,3 +74,29 @@ def test_ast_convertion(self, equation, expected):
...
+
+
+class TestOPSLifting(object):
+
+    @pytest.mark.parametrize('eq, expected', [
+        ('Eq(u,3*a - 4**a)', 'void Kernel0(float * ut00)\n\
+{\n\
+  ut00[OPS_ACC0(0)] = -2.97015324253729F;\n\
+}'),
+        ('Eq(w,u + v)', 'void Kernel0(float * ut00, float * vt00, float * wt00)\n\
```
vmickus: If the `const` modifier is a must for the OPS kernel to all variables that only have read operations, then this test should be `void Kernel0(float * ut00,const  float * vt00, const float * wt00)`. 

Just asked Istvan in OPS channel to be sure, but if its necessary i think it should be added in this PR, because you could check it from the Expression if its on LHS or something like this.

vincepandolfo: Good call. I had assumed it wouldn't make a difference but won't hurt to check

devito/ops/transformer.py
```diff
@@ -1,3 +1,38 @@
...
+    arguments -= set(to_remove)
+
+    callable_kernel = OPSKernel(
+        namespace['ops_kernel'](count),
+        expressions,
+        "void",
+        list(arguments)
+    )
+
+    return callable_kernel, None
```
vmickus: I don't really see the use of returning the `None`. 
I mean if not needed now, just return the first argument. In the future, when the PR for `par_loop_block` is created then this will be added.

devito/ops/transformer.py
```diff
@@ -1,3 +1,38 @@
+from devito.ir.iet import Expression, FindNodes
+from devito.ops.node_factory import OPSNodeFactory
+from devito.ops.nodes import OPSKernel
+from devito.ops.utils import namespace
+
+
+def opsit(trees, count):
```
vmickus: The docstring for this method is important, i think it should be added.

tests/test_ops.py
vmickus: I think it would be great if there is more tests. 

For example:
* Testing an expression in which the same `Function` is used more the one time. This is useful to check if the are no duplicates in the generated arguments. ( this should be working as i have seen in the code, but the test is important).
* Also test the generation of a kernel with more than 1 expression.

@FabioLuporini  could have more ideas here

vmickus: Considering Istvan answer, i think `const` modifier is necessary.

devito/ir/iet/visitors.py
mloubout: The wouldn't it be better to have `restrict` as a backend keyword rather than "cheating" in if?

FabioLuporini: yeah, ``restrict`` should somewhat be moved to the backends (core, ops, etc) but I'm not sure it's in the scope of this PR

devito/ops/transformer.py
```diff
@@ -1,3 +1,38 @@
...
+    for tree in trees:
+        expressions.extend([
+            Expression(make_ops_ast(i.expr, node_factory))
+            for i in FindNodes(Expression).visit(tree.inner)
+        ])
+
+    arguments = set()
+    to_remove = []
+    for exp in expressions:
+        func = [f for f in exp.functions if f.name != "OPS_ACC_size"]
```
FabioLuporini: this looks hacky?

devito/ops/types.py
FabioLuporini: reminder

tests/test_ops.py
vincepandolfo: added

devito/ops/transformer.py
vincepandolfo: do you have a suggestion on how to do this properly?

devito/ops/types.py
FabioLuporini: What I think you should do is:

* call it differently: SymbolicArray ? remind me again (sorry) what C symbols it is supposed to represent?
* Add is_Tensor = False
* unflip the change in ``devito/ir/iet/visitors.py``

Remind me why can you not just inherit from Symbol and change (override) what's required?

devito/ops/transformer.py
FabioLuporini: Add a global tuple of "private names" and check if it's in there?

Alternatively, what is ``f`` when its name is "OPS_ACC_size" ?

devito/ops/node_factory.py
```diff
@@ -38,7 +39,8 @@ def new_ops_arg(self, indexed):
 
         if ops_arg_id not in self.ops_args:
             # Create the indexed object
-            ops_arg = Array(name=ops_arg_id,
+            ops_arg = Array(is_Write,
```
FabioLuporini: If this were a subclass of Scalar, instead of Array (see also my other comment below!), then you could drive the generation of const by simply providing a proper value for the [is_const flag](https://github.com/opesci/devito/blob/master/devito/types/basic.py#L383), which eventually gets used to generate or not the `const` modifier, see [here](https://github.com/opesci/devito/blob/master/devito/types/basic.py#L301)

devito/ir/iet/visitors.py
vincepandolfo: This is irrelevant now with the new OPS API that uses `ACC<T>` objects instead of arrays and accessor macros. I'm working on a new PR that uses that API. This could be closed if there's no objections? @FabioLuporini

FabioLuporini: sure, go ahead, thanks

vmickus: As we talked in the channel with Istvan about this OPS update, he said that the `OPSACC#` will continue to be supported. I don't see a reason to change it as you have done so much with it @vincepandolfo.  IMO.

vincepandolfo:
It would simplify a few things related to accessing arrays (I had to
revisit the expressions a few times to use the accessors properly using
OPS_ACC). Long term I'd prefer the new API and it's a cheap change at this
stage

vmickus:
I also prefer the new syntax as its easier to write the accessors like you said.

I was only concerned with the refactoring work. But I agree =]

FabioLuporini:
I understand from @vincepandolfo 's comments that we can close this -- a new PR supporting the updated OPS syntax will replace this one. So, proceeding...

PR #833, opened 2019-05-14
plot_field(...) enhancements

georgebisbas:
Adding xmin/ymin/zmin functionality

### DIFF
--- a/examples/cfd/tools.py
+++ b/examples/cfd/tools.py
@@ -2,10 +2,10 @@ from mpl_toolkits.mplot3d import Axes3D  # noqa
 
 import numpy as np
 from matplotlib import pyplot, cm
-
+from devito.logger import warning
 
 def plot_field(field, xmin=0., xmax=2., ymin=0., ymax=2., zmin=None, zmax=None,
-        view=None, linewidth=0):
+                view=None, linewidth=0):
     """Utility plotting routine for 2D data
 
     :param field: Numpy array with field data to plot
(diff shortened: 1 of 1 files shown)


examples/cfd/tools.py
```diff
@@ -4,27 +4,27 @@
...
                     linewidth=linewidth, antialiased=False)
 
     # Enforce axis measures and set view if given
-    ax.set_xlim(0., xmax)
-    ax.set_ylim(0., ymax)
+    ax.set_xlim(xmin, xmax)
+    ax.set_ylim(ymin, ymax)
     if zmax is not None:
-        ax.set_zlim(1., zmax)
+        ax.set_zlim(zmin, zmax)
```
mloubout: zmin or 0.

not sure why was starting at 1

georgebisbas: I believe because in all cfd examples the field was initialized to 1, so to avoid seeing an "elevated" field with a non-useful gap. Up to you.
The tutorial plots as they are now look fine, if we change to zmin=0(by default) they would be a bit uglier. In any case it is not like x and y where it is easy for user to know the space extend. I say that we leave it as it always was and maybe some day that tutorials will be revisited we can touch it again.

mloubout: then maybe `zmin or np.min(field)`

georgebisbas: np.min(field) is the best I think...

mloubout: no `zmin` defaults to  `None` so the correct synthax is `zmin or np.min(field)` so that input is not ignored if not `None`

examples/cfd/tools.py
```diff
@@ -4,27 +4,28 @@
 from matplotlib import pyplot, cm
 
 
-def plot_field(field, xmax=2., ymax=2., zmax=None, view=None, linewidth=0):
+def plot_field(field, xmin=0., xmax=2., ymin=0., ymax=2., zmin=None, zmax=None,
+        view=None, linewidth=0):
```
FabioLuporini: pep8 violations: indentation is broken, so CI will complain

examples/cfd/tools.py
```diff
@@ -4,27 +4,28 @@
...
                     linewidth=linewidth, antialiased=False)
 
     # Enforce axis measures and set view if given
-    ax.set_xlim(0., xmax)
-    ax.set_ylim(0., ymax)
-    if zmax is not None:
-        ax.set_zlim(1., zmax)
+    ax.set_xlim(xmin, xmax)
+    ax.set_ylim(ymin, ymax)
+    if zmax is not None or zmin is not None:
```
FabioLuporini: these could have been two separate ``if``, instead of adding an ``or``

FabioLuporini:
there's a small conflict to fix
also, @mloubout, this PR has changed a bit since your approval. Could you take a second look

georgebisbas:
I guess as soon as you have Mathias' approval you can merge...all green now..

FabioLuporini:
merged.

PR #837, opened 2019-05-29
operator: configuration settings are saved and printed in performance logging

jrt54:
The devito configuration dictionary at the time an operator is constructed is now saved to that operator's state, accessible via Operator._state['optimizations']. 

The purpose of this is to allow reproducibility in the case that a user did not modify their Python script (same operator, space_order, etc.) but maybe called a Python script from a terminal or runscript where certain environment variables (e.g. DEVITO_DSE) were set. New printouts include pertinent configuration settings for a given operator.

### DIFF
--- a/devito/operator.py
+++ b/devito/operator.py
@@ -190,6 +190,11 @@ class Operator(Callable):
         iet = self._finalize(iet, parameters)
 
         super(Operator, self).__init__(self.name, iet, 'int', parameters, ())
+        config_state = {}
+        self._state['optimizations'] = config_state
+        for key in configuration.keys():
+            config_val = kwargs.get(key, configuration[key])
+            config_state[key] = config_val
 
     # Read-only fields exposed to the outside world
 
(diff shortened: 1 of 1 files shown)


devito/operator.py
```diff
@@ -190,6 +190,11 @@ def __init__(self, expressions, **kwargs):
         iet = self._finalize(iet, parameters)
 
         super(Operator, self).__init__(self.name, iet, 'int', parameters, ())
+        config_state = {}
```
FabioLuporini: I think we should make this such that it can be specialised by an Operator subclass. That is, we should put this in a method and replace ``self._state = {}`` a few lines above with, for example, ``self._state = self._initialize_state(...)``

then you have
```
def _initialize_state(self, **kwargs):
    # equivalent to what you wrote, but more pythonic and homogeneous with the rest of devito
    return {i: kwargs.get(i, v) for k, v in configuration.items()}
```

jrt54: Submitted a new commit with this kind of change, except the state dictionary stores configuration items as a nested dictionary, so that the keys in the configuration dictionary at construction and the pre-existing keys used in the state dictionary (i.e. 'autotuning') don't interact.

jrt54:
After offline discussion with @FabioLuporini, I think the only critical things to document in the configuration dictionary are 'dse' and 'dle'. 

For brevity, this is now all that's printed and logged. The motivation is to document performance-critical options to an operator kernel that are controlled by environment variables and are not otherwise obvious in the current performance logging.

FabioLuporini:
OK, this PR now looks good to me and can land. A build is failing on CI, but not because of this PR (it's about installation), so we'll investigate the issue and once fixed we'll land this. Many thanks @jrt54 !

Out of curiosity, could you attach how the output log looks like now with the new print? (clearly when running with a suitable logging level)

FabioLuporini:
Thanks, merged!

PR #838, opened 2019-05-30
compiler: change load/save/make/get_jit_dir/get_codepy_dir into methods

tjb900:
This allows Compiler subclasses to customise the locations of the build directories etc, for cases where `tempfile.gettempdir()` returns something that we don't want to use.


FabioLuporini:
Thanks! Merged

PR #844, opened 2019-06-05
Use absolute tolerance in adjoint test

navjotk:
When studying the adjoint test, we realised that `np.isclose` has a default value for `atol` which makes it inappropriate to use directly for numbers much smaller than 1 ( https://docs.scipy.org/doc/numpy/reference/generated/numpy.isclose.html )

To make things behave as expected, we must pass an `atol` value of 0 when using `rtol`.


navjotk:
It's actually the opposite. The test as defined in master right now is very loose - even a "wrong" adjoint can pass this test because `np.isclose()` uses the following equation to determine equality: 
`absolute(a - b) <= (atol + rtol * absolute(b))` .
The default value of `atol` being of the order of `10^-4` means that an `rtol` value of `10^-10` isn't actually testing anything - as long as `a` and `b` are within `10^-4` of each other (from the default value of `atol`), the test will pass regardless of `rtol`. 

This is not a good test right now. 

I am trying to make this test actually work as one would expect.

navjotk:
@mloubout Are you happy to let me change `np.isclose((term1 - term2)/term1, 0.)` to `np.isclose(term1, term2)`?

mloubout:
No, first one makes sense mathematically too so I would prefer to keep it

navjotk:
The line `np.isclose((term1 - term2)/term1, 0., rtol=1.e-10)` is first calculating the relative error between term1 and term2, and then comparing it to 0 using a *relative* tolerance again. AFAIU, this is calculating the relative error of the relative error. 
That doesn't make sense to me. Am I missing something?

tests/test_adjoint.py
```diff
@@ -110,7 +110,7 @@ def test_adjoint_J(self, shape, space_order):
         term2 = linalg.norm(du.data.reshape(-1))**2
         info('<Jx,y>: %f, <x, J^Ty>: %f, difference: %4.4e, ratio: %f'
              % (term1, term2, (term1 - term2)/term1, term1 / term2))
-        assert np.isclose((term1 - term2)/term1, 0., rtol=1.e-10)
+        assert np.isclose((term1 - term2)/term1, 0., rtol=1.e-10, atol=0.)
```
mloubout: Ok so I re-read this with frech brain.

`absolute(a - b) <= (atol + rtol * absolute(b))`

`b` is always `0` so `rtlo` is uselss (0*rtol always 0), but `atol=0` is completley unfeasible never gonna be that exact. test in double precision so should be 

`np.isclose((term1 - term2)/term1, 0., atol=1e-12)`

which test 

`abs(a-b) = abs(a) <= 1e-12)` or whatever `atol` you think is consistent, current one is default which is 1e-8

navjotk: > Ok so I re-read this with frech brain.

I read this as "French brain" :laughing:

The biggest problem I see in doing it this way is if term1 is really close to 0, it is scary to divide by term1.

mloubout: term1 should never be close to zero unless setup weirdly

FabioLuporini:
@mloubout I see you approved this PR -- is it ready to land? because then I saw some further discussion...

mloubout:
No not ready see my recent review

PR #846, opened 2019-06-05
Add DSE rewriter passes

georgebisbas:
[WIP][CFR]
A small PR aiming to introduce template for new passes to the DSE rewriters. Work in progess. 
Probably needs rebasing.


devito/dse/rewriters.py
```diff
@@ -405,8 +418,24 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
+        try:
+            passes = passes.split(',')
+        except AttributeError:
+            # Already in tuple format
+            if not all(i in CustomRewriter.passes_mapper for i in passes):
+                raise DSEException("Unknown passes `%s`" % str(passes))
+        self.passes = passes
+        self.template = template
+        self.profile = profile
+        #super(CustomRewriter, self).__init__(params)
```
georgebisbas: Create superclass

devito/dse/transformer.py
```diff
@@ -49,18 +49,18 @@ def rewrite(clusters, mode='advanced'):
...
+    except KeyError:
+        rewriter = CustomRewriter(mode)
 
     # We use separate rewriters for dense and sparse clusters; sparse clusters have
     # non-affine index functions, thus making it basically impossible, in general,
     # to apply the more advanced DSE passes.
     # Note: the sparse rewriter uses the same template for temporaries as
     # the dense rewriter, thus temporaries are globally unique
-    rewriter = modes[mode]()
     fallback = BasicRewriter(False, rewriter.template)
```
georgebisbas: fix gaps

devito/dse/rewriters.py
```diff
@@ -382,6 +383,7 @@ def _eliminate_inter_stencil_redundancies(self, cluster, template, **kwargs):
 class AggressiveRewriter(AdvancedRewriter):
 
     def _pipeline(self, state):
+        self._new_pass(state)
```
FabioLuporini: what is this?

devito/dse/rewriters.py
```diff
@@ -408,5 +418,22 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
+        try:
+            passes = passes.split(',')
+        except AttributeError:
+            # Already in tuple format
+            if not all(i in CustomRewriter.passes_mapper for i in passes):
+                raise DSEException("Unknown passes `%s`" % str(passes))
+        self.passes = passes
+        self.template = template
+        self.profile = profile
+        # super(CustomRewriter, self).__init__(params)
```
FabioLuporini: why is this commented out?

we should call the superclass constructor and pass in the right values

devito/dse/rewriters.py
```diff
@@ -408,5 +418,22 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
 
 class CustomRewriter(AggressiveRewriter):
 
+    passes_mapper = {
+        'new_pass': AggressiveRewriter._new_pass
```
FabioLuporini: we need the actual passes here, with reasonable names. I suggest you propose them ?

devito/dse/transformer.py
```diff
@@ -49,18 +49,18 @@ def rewrite(clusters, mode='advanced'):
...
-
```
FabioLuporini: no need to drop this line

devito/dse/transformer.py
```diff
@@ -49,18 +49,18 @@ def rewrite(clusters, mode='advanced'):
 
     if mode is None or mode == 'noop':
         return clusters
-    elif mode not in dse_registry:
-        warning("Unknown rewrite mode(s) %s" % mode)
-        return clusters
+
+    try:
```
FabioLuporini: maybe it's better if this whole try-except goes right below the big comment below

devito/dse/rewriters.py
georgebisbas: I need a dummy name as long as I do not have a real-existing pass.

georgebisbas: How should I call it?

devito/dse/rewriters.py
georgebisbas: ok, answered before. Let's find a name, I think it will be temporary as soon as skewing is introduced.

devito/dse/rewriters.py
georgebisbas: I left it in order not to forget to do the superclass as a reminder. Should have added TODO. I forgot it earlier...

devito/dse/transformer.py
georgebisbas: Agree, But looking at the big-comment, is there any need for rewriting it... Is it easy for someone to understand when we deal with a sparse or dense? How/When/Why... It is probably just me that did not yet observe any info about sparse/dense rewriters...

devito/dse/rewriters.py
FabioLuporini: since it's unused and basically a no-op, it shouldn't appear in this PR

devito/dse/rewriters.py
```diff
@@ -393,6 +395,14 @@ def _pipeline(self, state):
         self._factorize(state)
         self._eliminate_intra_stencil_redundancies(state)
 
+    @dse_pass
+    def _new_pass(self, cluster, template, **kwargs):
```
FabioLuporini: same, this can be dropeed

devito/dse/rewriters.py
FabioLuporini: no I think we need names for the current passes, such as `_extract_sum_of_products`, `_extract_time_invariants`, etc

sorry I hadn't explained myself properly before

devito/dse/rewriters.py
FabioLuporini: 👍

devito/dse/transformer.py
FabioLuporini: I mean to move it below? not sure I follow.

The comment below says something, that something is partly about what happens in this try-except

georgebisbas:
I added a dummy extract_sum_of products pass like the already exisitng one. Of course I will not leave it like this, but I want to ask whether I follow what exactly I think you have in mind. Let me know if I should add any more tests for this PR. Thanks.

devito/dse/rewriters.py
```diff
@@ -405,8 +406,37 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
 
         return cluster.rebuild(processed)
 
+    @dse_pass
+    def _extract_sum_of_products2(self, cluster, template, **kwargs):
```
FabioLuporini: I really don't understand why you're introducing this

devito/dse/rewriters.py
```diff
@@ -405,8 +406,37 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
+        costmodel = lambda e: not (q_leaf(e) or q_terminalop(e))
+        processed, _ = xreplace_constrained(cluster.exprs, make, rule, costmodel)
+
+        return cluster.rebuild(processed)
+
 
 class CustomRewriter(AggressiveRewriter):
 
+    passes_mapper = {
+        'extract_sum_of_products2': AggressiveRewriter._extract_sum_of_products2
```
FabioLuporini: Still missing the mapping to all existing passes

devito/dse/rewriters.py
```diff
@@ -405,8 +406,37 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
+        try:
+            passes = passes.split(',')
+        except AttributeError:
+            # Already in tuple format
+            if not all(i in CustomRewriter.passes_mapper for i in passes):
+                raise DSEException("Unknown passes `%s`" % str(passes))
+        self.passes = passes
+        self.template = template
+        self.profile = profile
+        super(CustomRewriter, self).__init__()
```
FabioLuporini: Wrong. This should forward template and profile up the chain.

tests/test_dse.py
```diff
@@ -429,6 +429,14 @@ def test_acoustic_rewrite_basic():
     assert np.allclose(ret1[1].data, ret2[1].data, atol=10e-5)
 
 
+def test_new_extract_pass():
```
FabioLuporini: test_custom_rewriter ?

georgebisbas:
Fabio, I feel a bit confused about this. Correct me for whatever wrong I write. Currently DSE does not have any extra option for a DSE pass. In the future 'skewing' will be an example of such a step. This PR (dse-rewriter-modes) will go first for sure. Do you propose to move one of the existing pass as an extra-single-option > (I think not). Our purpose is to have a pass that will be doing sth trivial. You said earlier: 
>no I think we need names for the current passes, such as _extract_sum_of_products, _extract_time_invariants, etc

I don't get the reason why... .

Please enlighten me what do you want this pass to be doing. It seems to me that your comments here defy our original plans.

FabioLuporini:
The purpose of a "CustomRewriter" (see for example the one in the DLE) is to let users specify an arbitrary sequence of passes. So **all** passes (ie, all methods tagged with `@dse_pass`) need to appear in the mapper. One day, when ready and in an appropriate PR, to this mapper we will also add skewing.

georgebisbas:
> The purpose of a "CustomRewriter" (see for example the one in the DLE) is to let users specify an arbitrary sequence of passes. So **all** passes (ie, all methods tagged with `@dse_pass`) need to appear in the mapper. One day, when ready and in an appropriate PR, to this mapper we will also add skewing.

I believe that now I understood what you mean, check out the latest push when you have time...Thanks!

devito/dse/rewriters.py
```diff
@@ -408,5 +409,26 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
 
 class CustomRewriter(AggressiveRewriter):
 
+    passes_mapper = {
+        'extract_sum_of_products': AggressiveRewriter._extract_sum_of_products,
+        'factorization': AggressiveRewriter._factorize,
+        'eliminate_inter_stencil_redundancies':
```
FabioLuporini: some passes are missing

devito/dse/rewriters.py
```diff
@@ -408,5 +409,26 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
+        try:
+            passes = passes.split(',')
+        except AttributeError:
+            # Already in tuple format
+            if not all(i in CustomRewriter.passes_mapper for i in passes):
+                raise DSEException("Unknown passes `%s`" % str(passes))
+        self.passes = passes
+        self.template = template
+        self.profile = profile
+        super(CustomRewriter, self).__init__()
```
FabioLuporini: as I think I said in a previous message, this is not how you should invoke the superclass constructor. You should also pass the arguments that the superclass is expected to handle, such as `template` and `profile`

devito/dse/rewriters.py
georgebisbas: Added from all Rewriters now. Is that something we want ? I guess yes?

devito/dse/rewriters.py
georgebisbas:
```
super(CustomRewriter, self).__init__(self.profile, self.template)
```
?

devito/dse/rewriters.py
georgebisbas: `super(CustomRewriter, self).__init__(self.profile, self.template)` ?

devito/dse/rewriters.py
```diff
@@ -408,5 +409,32 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
+        try:
+            passes = passes.split(',')
+        except AttributeError:
+            # Already in tuple format
+            if not all(i in CustomRewriter.passes_mapper for i in passes):
+                raise DSEException("Unknown passes `%s`" % str(passes))
+        self.passes = passes
+        self.template = template
+        self.profile = profile
+        super(CustomRewriter, self).__init__(self.profile, self.template)
```
FabioLuporini: still conceptually wrong

devito/dse/rewriters.py
FabioLuporini: not yet

devito/dse/rewriters.py
FabioLuporini: I think we need super short names for these passes. I'll try listing them, in order, but please feel free to improve them

```
extract_sop
factorize
gcse  # Generalized Common Sub-expressions Elimination  (this is new terminology I'm introducing)
cse  # Common Sub-expressions Elimination 
extract_invariants
extract_indices
opt_trascendentals
```

devito/dse/rewriters.py
```diff
@@ -431,9 +428,7 @@ def __init__(self, passes, template=None, profile=True):
             if not all(i in CustomRewriter.passes_mapper for i in passes):
                 raise DSEException("Unknown passes `%s`" % str(passes))
         self.passes = passes
-        self.template = template
-        self.profile = profile
-        super(CustomRewriter, self).__init__(self.profile, self.template)
+        super(CustomRewriter, self).__init__(profile, template)
```
FabioLuporini: now yes

FabioLuporini:
OK, now this is good to go. 

Thanks!

devito/dse/rewriters.py
georgebisbas: I admit being quite frivolous in this PR. ~

FabioLuporini:
Thanks! Merged.

devito/dse/rewriters.py
```diff
@@ -411,10 +411,16 @@ class CustomRewriter(AggressiveRewriter):
 
     passes_mapper = {
         'extract_sum_of_products': AggressiveRewriter._extract_sum_of_products,
-        'factorization': AggressiveRewriter._factorize,
+        'factorize': AggressiveRewriter._factorize,
         'eliminate_inter_stencil_redundancies':
-            AggressiveRewriter._eliminate_inter_stencil_redundancies
-
+            AggressiveRewriter._eliminate_inter_stencil_redundancies,
+        'eliminate_intra_stencil_redundancies':
```
btrb: why the difference between `inter` and `intra`? couldn't it be the same in both?

PR #848, opened 2019-06-07
Update README.md

georgebisbas:
Add banner, MPI + compiler info, binder under consideration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,3 @@
-![alt text](https://www.devitoproject.org/images/banner-devito.jpg)
-
 # Devito: Fast Finite Difference Computation from Symbolic Specification
 
 ![Build Status](https://travis-ci.org/opesci/devito.svg?branch=master)
@@ -72,10 +70,9 @@ expressions. Examples of how to define operators are provided:
   directory and run `jupyter notebook`.
 * A set of tutorial notebooks for seismic inversion examples is available in
   `examples/seismic/tutorials`.
-* A set of tutorial notebooks concerning the compiler of Devito can be found in
+* A set of tutorial notebooks concerning the Devito compiler can be found in
   `examples/compiler`.
-* Devito with MPI can be explored in
-  `examples/MPI`.
+* Devito with MPI can be explored in `examples/MPI`.
 * Example implementations of acoustic forward, adjoint, gradient and born
   operators for use in full-waveform inversion (FWI) methods can be found in
   `examples/seismic/acoustic`.
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ expressions. Examples of how to define operators are provided:
   `examples/seismic/tutorials`.
 * A set of tutorial notebooks concerning the Devito compiler can be found in
   `examples/compiler`.
-* Devito with MPI can be explored in `examples/MPI`.
+* Devito with MPI can be explored in `examples/mpi`.
 * Example implementations of acoustic forward, adjoint, gradient and born
   operators for use in full-waveform inversion (FWI) methods can be found in
   `examples/seismic/acoustic`.
--- a/README.md
+++ b/README.md
@@ -1,7 +1,10 @@
+![alt text](https://www.devitoproject.org/images/banner-devito.jpg)
+
 # Devito: Fast Finite Difference Computation from Symbolic Specification
 
 ![Build Status](https://travis-ci.org/opesci/devito.svg?branch=master)
 ![Code Coverage](https://codecov.io/gh/opesci/devito/branch/master/graph/badge.svg)
+[![Binder](http://devito.rocks/badge_logo.svg)](http://devito.rocks/v2/gh/opesci/devito/binder)
 
 [Devito](http://www.devitoproject.org) is a software to
 implement optimised finite difference (FD) computation from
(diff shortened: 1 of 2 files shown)


georgebisbas:
@FabioLuporini the issue with MPI notebook is happening again...

FabioLuporini:
@georgebisbas because it's still missing the #NBVAL_IGNORE_OUTPUT ?_?
Look at Cell 4 !

README.md
```diff
@@ -1,7 +1,10 @@
+![alt text](https://www.devitoproject.org/images/banner-devito.jpg)
```
FabioLuporini: Actually, I'm not sure we want the banner here ?

README.md
```diff
@@ -1,7 +1,10 @@
...
+
 # Devito: Fast Finite Difference Computation from Symbolic Specification
 
 ![Build Status](https://travis-ci.org/opesci/devito.svg?branch=master)
 ![Code Coverage](https://codecov.io/gh/opesci/devito/branch/master/graph/badge.svg)
+[![Binder](http://devito.rocks/badge_logo.svg)](http://devito.rocks/v2/gh/opesci/devito/binder)
```
FabioLuporini: Perfect!

README.md
```diff
@@ -69,6 +72,10 @@ expressions. Examples of how to define operators are provided:
   directory and run `jupyter notebook`.
 * A set of tutorial notebooks for seismic inversion examples is available in
   `examples/seismic/tutorials`.
+* A set of tutorial notebooks concerning the compiler of Devito can be found in
```
FabioLuporini: concerning the Devito compiler

README.md
```diff
@@ -69,6 +72,10 @@ expressions. Examples of how to define operators are provided:
...
+  `examples/compiler`.
+* Devito with MPI can be explored in
+  `examples/MPI`.
```
FabioLuporini: no need to break line?

FabioLuporini:
Many thanks for this 🙏 

Can you inject the NBVAL thing as per comment above into this PR? so that we fix it once and for all

README.md
georgebisbas: Whatever you want, np...

README.md
georgebisbas: Fixed

README.md
georgebisbas: Fixed

README.md
FabioLuporini: I'd say let's drop it

examples/mpi/overview.ipynb
```diff
@@ -128,6 +129,7 @@
    "outputs": [],
    "source": [
     "%%px\n",
+    "#NBVAL_IGNORE_OUTPUT\n",
```
FabioLuporini: why do you need this one here? did you get another error in this cell?

georgebisbas: No, I did not, but the same happened before. cell 3 failed ->added nbval there -> cell 4 failed -> I will remove it probably...Just want to see how this thing works.. I will not be surprised if this error is raised again in other cells... Give it some time to be tested and I will fix/remove redundant things...

FabioLuporini: I think cells 3 and 4 should be the only ones needing the NBVAL_IGNORE_OUTPUT tag because that's where we've seen the error appear. Maybe due to changing `configuration` ? for reasons I can't really explain. It's however important not to add the NBVAL tag randomly in the other cells too . If the error starts appearing in other cells, I'm afraid we'll have to investigate deeper what's going on. Hopefully this won't be the case

georgebisbas: Yes totally agree, I will investigate next hours...

README.md
georgebisbas: dropped

examples/mpi/overview.ipynb
georgebisbas: Tests pass right now in pipelines.

FabioLuporini: but nonetheless, you should drop this

FabioLuporini:
@georgebisbas can you inject the NBVAL tag for the failing MPI cell into this PR?

georgebisbas:
> @georgebisbas can you inject the NBVAL tag for the failing MPI cell into this PR?
pushed

georgebisbas:
> Why? Was it failing?
Unfortunately yes.

README.md
mloubout: I would still put a comment that it doesn't always work/is experimental (failed to build every time I tried)

examples/mpi/overview.ipynb
```diff
@@ -1465,6 +1467,7 @@
    ],
    "source": [
     "%%px\n",
+    "#NBVAL_IGNORE_OUTPUT\n",
```
mloubout: I (personally) find it very dangerous to add the flag to everything just to make CI happy as it was not necessary before.

georgebisbas: I agree with both comments. Just talked with Fabio about targeting to explore this issue.

FabioLuporini: @mloubout is completely right. We need to find out what's going on. Who can we ask ? maybe the ipyparallel guys?

FabioLuporini:
I think I have to withdraw my previous suggestions. The MPI-notebook issue is more serious than I thought, so we have to treat it separately in a different branch/PR. We should pull of the NBVAL tags from this PR. Apologies

georgebisbas:
> I think I have to withdraw my previous suggestions. The MPI-notebook issue is more serious than I thought, so we have to treat it separately in a different branch/PR. We should pull of the NBVAL tags from this PR. Apologies

NP, I will remove them now..

README.md
```diff
@@ -69,6 +69,9 @@ expressions. Examples of how to define operators are provided:
   directory and run `jupyter notebook`.
 * A set of tutorial notebooks for seismic inversion examples is available in
   `examples/seismic/tutorials`.
+* A set of tutorial notebooks concerning the Devito compiler can be found in
+  `examples/compiler`.
+* Devito with MPI can be explored in `examples/MPI`.
```
FabioLuporini: this is wrong. 'MPI' should be 'mpi'

FabioLuporini:
Merged, thanks

PR #850, opened 2019-06-11
examples: Hack-fix source.show

FabioLuporini:


FabioLuporini:
actually, @rhodrin , this should be doable in a better way once #824 is in?

FabioLuporini:
closing as this, hopefully, will become unnecessary once #824 is in (@rhodrin )

PR #852, opened 2019-06-11
mpi: block and force grouping of output in notebook

mloubout:
Try to fix random failure of mpi notebook


georgebisbas:
Probably not the way to get rid of the error as I can see.

FabioLuporini:
Merged.

PR #853, opened 2019-06-15
FWI Tutorial: Fixed gradient scaling, step size consistent, and computation of objective value

LukasMosser:
I made a few minor fixes for the FWI tutorial:

- The step size was inconsistent in the notebook: changed to 0.005
- The objective function was computed as 0.5*norm((true-pred)**2) and is now fixed to 0.5*norm(true-pred)**2
- The gradient scaling could run into issues when only taking the max(gradient) e.g. when the gradient is strictly negative, the largest value could be 0 and hence you'd divide by zero.I changed this to scaling by abs(gradient).max()


mloubout:
>The step size was inconsistent in the notebook: changed to 0.005

Inconsistent with what? 

>The objective function was computed as 0.5norm((true-pred)**2) and is now fixed to 0.5norm(true-pred)**2

Good catch

>The gradient scaling could run into issues when only taking the max(gradient) e.g. when the gradient is strictly negative, the largest value could be 0 and hence you'd divide by zero.I changed this to scaling by abs(gradient).max()

I have never seen a stricltly negative FWI gradient but sure no problem with that.

LukasMosser:
> Inconsistent with what?

There are two cells where the gradient computation was performed as an example.
The first showed just what the gradient would look like in a single FWI iteration. That line had 0.05.
The second instance was within the actual ```fwi_gradient``` method, where a learning rate of 0.005 was used. I just thought it's less confusing if they are consistent with each other, so that new users can relate to what the numerical values mean i.e. that they are the same in this case. Otherwise, the question could arise "Why is it 0.05 here and 0.005 there?"

FabioLuporini:
Many thanks, @LukasMosser !

Will merge this soon

FabioLuporini:
Thanks! Merged

PR #855, opened 2019-06-17
Fix nbpml model marmousi

fpicetti:
Dear mantainers,

I've spot a couple of errors related to Marmousi preset in seismic examples and tutorials.

1) In `examples/seismic/tutorials/02-rtm.ipynb` when selecting the preset to be loaded the name should be _marmousi2d-isotropic_ instead of _marmousi2d_ only.
2) in the same cell there is a function `create_model` that is referred to `demo_model` in `examples/seismic/model.py`. Selecting _marmousi2d-isotropic_ will raise an error: `UnboundLocalError: local variable 'nbpml' referenced before assignment`.

The patch I propose implement the mod 1) as above and 2) by adding a `nbpml = kwargs.pop('nbpml', 20)` as done for _marmousi2d-tti_ (line 276 of the master file).

With these fixes, the aforementioned notebook works.


FabioLuporini:
Merged, thanks

PR #856, opened 2019-06-18
Introduce skewing pass

georgebisbas:
This PR introduces a skewing pass as part of implementing time-tiling.
Data accesses remain the same after skewing.

Known issues/to fix in parallel with reviewing are commented in code.

### DIFF
--- a/devito/dse/rewriters.py
+++ b/devito/dse/rewriters.py
@@ -22,8 +22,7 @@ from devito.tools import flatten, generator
 from devito.types import Array, Scalar
 from devito.types import TimeDimension, SpaceDimension
 
-__all__ = ['BasicRewriter', 'AdvancedRewriter', 'AggressiveRewriter',
-           'SkewingRewriter', 'CustomRewriter']
+__all__ = ['BasicRewriter', 'AdvancedRewriter', 'AggressiveRewriter', 'CustomRewriter']
 
 
 class State(object):
@@ -373,56 +372,31 @@ class AggressiveRewriter(AdvancedRewriter):
 
         return cluster.rebuild(processed)
 
-
-class SkewingRewriter(AggressiveRewriter):
-
-    def _pipeline(self, state):
-        self._skewing(state)
-        self._extract_sum_of_products(state)
-        self._extract_time_invariants(state, with_cse=False)
-        self._eliminate_inter_stencil_redundancies(state)
-
-        self._extract_sum_of_products(state)
-        self._eliminate_inter_stencil_redundancies(state)
-        self._extract_sum_of_products(state)
-
-        self._factorize(state)
-        self._eliminate_intra_stencil_redundancies(state)
-
     @dse_pass
     def _skewing(self, cluster, template, **kwargs):
         """
-        Function to perform only the basic skewing in order to have valid data
+        A pass to perform only the basic skewing in order to have valid data
         dependences.
         """
         skew_factor = configuration['skew_factor']
         skew_dim, mapper, int_mapper = None, {}, []
         total_int = {}
         intervals, sub_iterators, directions = cluster.ispace.args
-        cnt = 0
+
         for dim in cluster.ispace.dimensions:
             if isinstance(dim, SpaceDimension):
                 mapper[dim] = dim + skew_factor*skew_dim
                 int_mapper.append(IntervalGroup([Interval(dim, -skew_factor*skew_dim,
                                                 -skew_factor*skew_dim)]))
-                cnt = cnt + 1
             elif dim.is_Time:
                 if isinstance(dim, TimeDimension):
                     int_mapper.append(IntervalGroup([Interval(dim, 0, 0)]))
-                    cnt = cnt + 1
                     skew_dim = dim
-                elif isinstance(dim.parent, TimeDimension):
-                    int_mapper.append(IntervalGroup([Interval(dim, 0, 0)]))
-                    cnt = cnt + 1
+
         if skew_dim is None:
             return cluster
 
-        if cnt == 5:
-            total_int = IntervalGroup.generate('union', int_mapper[0], int_mapper[1],
-                                               int_mapper[2], int_mapper[3])
-        elif cnt in {3, 4}:
-            total_int = IntervalGroup.generate('union', int_mapper[0], int_mapper[1],
-                                               int_mapper[2])
+        total_int = IntervalGroup.generate('union', *int_mapper)
 
         ispace = IterationSpace(total_int, sub_iterators, directions)
 
@@ -438,6 +412,7 @@ class CustomRewriter(AggressiveRewriter):
         'factorize': AggressiveRewriter._factorize,
         'gcse': AggressiveRewriter._eliminate_inter_stencil_redundancies,
         'cse': AggressiveRewriter._eliminate_intra_stencil_redundancies,
+        'skewing': AggressiveRewriter._skewing,
         'extract_invariants': AdvancedRewriter._extract_time_invariants,
         'extract_increments': BasicRewriter._extract_increments,
         'opt_transcedentals': BasicRewriter._optimize_trigonometry
--- a/devito/dse/rewriters.py
+++ b/devito/dse/rewriters.py
@@ -379,13 +379,28 @@ class AggressiveRewriter(AdvancedRewriter):
     def _skewing(self, cluster, template, **kwargs):
         """
         Skew the accesses along the time Dimension.
+        Example:
+        Transform
+        for i = 2, n-1
+            for j = 2, m-1
+                a[i,j] = (a[a-1,j] + a[i,j-1] + a[i+1,j] + a[i,j+1]) / 4
+            end for
+        end for
+        
+        to
+
+        for i = 2, n-1
+            for j = 2+i, m-1+i
+                a[i,j-i] = (a[a-1,j-i] + a[i,j-1-i] + a[i+1,j-i] + a[i,j+1-i]) / 4
+            end for
+        end for
         """
 
         skew_factor = self.DEFAULT_SKEW_FACTOR
 
         _, sub_iterators, directions = cluster.ispace.args
 
-        skew_dim, mapper, intervals = None, {}, []
+        #skew_dim, mapper, intervals = None, {}, []
 
         skew_dims = {i.dim for i in cluster.ispace.intervals if i.dim.is_Time}
         try:
@@ -396,24 +411,26 @@ class AggressiveRewriter(AdvancedRewriter):
         if len(skew_dims) > 0:
             raise ValueError("More than 1 time dimensions. Aborting tt...")
 
+        mapper, intervals = {}, []
+
         # Initializing a default time_dim index position in loop
         index = 0
-        if skew_dim is not None:
-            for i in cluster.ispace.intervals:
-                if i.dim.is_Time:
-                    intervals.append(Interval(i.dim, 0, 0))
-                    skew_dim = i.dim
-                    index = intervals.index(Interval(i.dim, 0, 0))
-                elif index < cluster.ispace.intervals.index(i):
-                    mapper[i.dim] = i.dim + skew_factor*skew_dim
-                    intervals.append(Interval(i.dim, -skew_factor*skew_dim,
-                                              -skew_factor*skew_dim))
-                else:
-                    intervals.append(i)
+        #if skew_dim is not None:
+        for i in cluster.ispace.intervals:
+            if i.dim.is_Time:
+                intervals.append(Interval(i.dim, 0, 0))
+                skew_dim = i.dim
+                index = intervals.index(Interval(i.dim, 0, 0))
+            elif index < cluster.ispace.intervals.index(i):
+                mapper[i.dim] = i.dim + skew_factor*skew_dim
+                intervals.append(Interval(i.dim, -skew_factor*skew_dim,
+                                           -skew_factor*skew_dim))
+            else:
+                intervals.append(i)
 
-            processed = xreplace_indices(cluster.exprs, mapper)
-            ispace = IterationSpace(intervals, sub_iterators, directions)
-            cluster = Cluster(processed, ispace, cluster.dspace, guards=cluster.guards)
+        processed = xreplace_indices(cluster.exprs, mapper)
+        ispace = IterationSpace(intervals, sub_iterators, directions)
+        cluster = Cluster(processed, ispace, cluster.dspace, guards=cluster.guards)
 
         return cluster
 
--- a/devito/dse/rewriters.py
+++ b/devito/dse/rewriters.py
@@ -13,14 +13,12 @@ from devito.dse.manipulation import (common_subexprs_elimination, collect_nested
                                      compact_temporaries)
 from devito.exceptions import DSEException
 from devito.logger import dse_warning as warning
-from devito.parameters import configuration
 from devito.symbolics import (bhaskara_cos, bhaskara_sin, estimate_cost, freeze,
                               iq_timeinvariant, pow_to_mul, retrieve_indexed,
                               q_affine, q_leaf, q_scalar, q_sum_of_product,
                               q_terminalop, xreplace_constrained, xreplace_indices)
 from devito.tools import flatten, generator
 from devito.types import Array, Scalar
-from devito.types import TimeDimension, SpaceDimension
 
 __all__ = ['BasicRewriter', 'AdvancedRewriter', 'AggressiveRewriter', 'CustomRewriter']
 
@@ -399,7 +397,7 @@ class AggressiveRewriter(AdvancedRewriter):
             else:
                 mapper[i.dim] = i.dim + skew_factor*skew_dim
                 intervals.append(Interval(i.dim, -skew_factor*skew_dim,
-                                         -skew_factor*skew_dim))
+                                          -skew_factor*skew_dim))
 
         ispace = IterationSpace(intervals, sub_iterators, directions)
         cluster = Cluster(cluster.exprs, ispace, cluster.dspace, cluster.atomics,
--- a/devito/dse/rewriters.py
+++ b/devito/dse/rewriters.py
@@ -348,8 +348,6 @@ class AdvancedRewriter(BasicRewriter):
 
 class AggressiveRewriter(AdvancedRewriter):
 
-
-
     def _pipeline(self, state):
         self._extract_sum_of_products(state)
         self._extract_time_invariants(state, with_cse=False)
@@ -394,7 +392,6 @@ class AggressiveRewriter(AdvancedRewriter):
 
         skew_dim, mapper, intervals = None, {}, []
 
-
         for i in cluster.ispace.intervals:
             if i.dim.is_Time:
                 intervals.append(Interval(i.dim, 0, 0))
@@ -402,7 +399,7 @@ class AggressiveRewriter(AdvancedRewriter):
             else:
                 mapper[i.dim] = i.dim + skew_factor*skew_dim
                 intervals.append(Interval(i.dim, -skew_factor*skew_dim,
-                                                -skew_factor*skew_dim))
+                                          -skew_factor*skew_dim))
 
         # TODO Can I avoid 2 rebuilds?
         processed = xreplace_indices(cluster.exprs, mapper)
--- a/devito/dse/rewriters.py
+++ b/devito/dse/rewriters.py
@@ -348,6 +348,11 @@ class AdvancedRewriter(BasicRewriter):
 
 class AggressiveRewriter(AdvancedRewriter):
 
+    MAX_SKEW_FACTOR = 8
+    """
+    The maximum skew factor applied for the skewing pass.
+    """
+
     def _pipeline(self, state):
         self._extract_sum_of_products(state)
         self._extract_time_invariants(state, with_cse=False)
@@ -378,17 +383,12 @@ class AggressiveRewriter(AdvancedRewriter):
         Skew the accesses along the time Dimension.
         """
 
-        MAX_SKEW_FACTOR = 8
-        """
-        The maximum skew factor applied for the skewing pass.
-        """
+        skew_factor = 3  # TOFIX
 
-        skew_factor = 2  # TOFIX
+        if skew_factor not in range(0, self.MAX_SKEW_FACTOR):
+            raise ValueError("skew_factor must be less than %d" % self.MAX_SKEW_FACTOR)
 
-        intervals, sub_iterators, directions = cluster.ispace.args
-
-        if skew_factor not in range(0, MAX_SKEW_FACTOR):
-            raise ValueError("skew_factor must be less than %d" % MAX_SKEW_FACTOR)
+        _, sub_iterators, directions = cluster.ispace.args
 
         skew_dim, mapper, intervals = None, {}, []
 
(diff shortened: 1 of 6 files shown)


devito/dse/transformer.py
```diff
@@ -16,6 +16,8 @@
 }
 """The DSE transformation modes."""
 
+MAX_SKEW_FACTOR = 8
+configuration.add('skew_factor', 0, range(MAX_SKEW_FACTOR))
```
georgebisbas: I tried to move it, but did not succeed yet. Where do you think should this live in?

devito/dse/rewriters.py
```diff
@@ -406,6 +408,47 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
+                    int_mapper.append(IntervalGroup([Interval(dim, 0, 0)]))
+                    cnt = cnt + 1
+                    skew_dim = dim
+                elif isinstance(dim.parent, TimeDimension):
+                    int_mapper.append(IntervalGroup([Interval(dim, 0, 0)]))
+                    cnt = cnt + 1
+        if skew_dim is None:
+            return cluster
+
+        if cnt == 5:
```
georgebisbas: I do not like this as I have it here. How can I parametrize somehow the int_mapper[] to pass them like all together
without having 4-5 arguments in a row?

devito/symbolics/manipulation.py
```diff
@@ -222,6 +222,8 @@ def as_symbol(expr):
         return Symbol(expr)
     elif isinstance(expr, Dimension):
         return Symbol(expr.name)
+    elif (isinstance(expr.args[0], Number) and isinstance(expr.args[1], Dimension)):
```
georgebisbas: @FabioLuporini what about having that now instead of Mul? Do you think it is better now?

devito/dse/rewriters.py
```diff
@@ -406,6 +408,47 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
         return cluster.rebuild(processed)
 
+    @dse_pass
+    def _skewing(self, cluster, template, **kwargs):
+        """
+        A pass to perform only the basic skewing in order to have valid data
+        dependences.
+        """
+        skew_factor = configuration['skew_factor']
+        skew_dim, mapper, int_mapper = None, {}, []
```
FabioLuporini: Exercise: rethink this entire piece of code and make it:
- more general (see also comments below)
- more robust
- **simpler** (e.g., here you're using two dicts, one list, one integer counter, and some other stuff...to do something which is very basic) ...
- ... and consequently easier to read

use as model the other algorithms in devito. 

This is a coding exercise

recall: let **simplicity** guide you. If it's not simple, there's a good chance it's flawed/wrong/broken/ etc.

devito/dse/rewriters.py
georgebisbas: fixed

devito/dse/rewriters.py
```diff
@@ -406,6 +408,38 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
 
         return cluster.rebuild(processed)
 
+    @dse_pass
+    def _skewing(self, cluster, template, **kwargs):
+        """
+        A pass to perform only the basic skewing in order to have valid data
```
FabioLuporini:
```
"""
Skew the accesses along the time Dimension.
"""
```

devito/dse/rewriters.py
```diff
@@ -406,6 +408,38 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
+            if isinstance(dim, SpaceDimension):
+                mapper[dim] = dim + skew_factor*skew_dim
+                int_mapper.append(IntervalGroup([Interval(dim, -skew_factor*skew_dim,
+                                                -skew_factor*skew_dim)]))
+            elif dim.is_Time:
+                if isinstance(dim, TimeDimension):
+                    int_mapper.append(IntervalGroup([Interval(dim, 0, 0)]))
+                    skew_dim = dim
+
+        if skew_dim is None:
```
FabioLuporini: this check should somehow happen at the beginning of this method, not so late. Then you won't need to carry along ``skew_dim``

devito/dse/rewriters.py
```diff
@@ -406,6 +408,38 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
 
+    @dse_pass
+    def _skewing(self, cluster, template, **kwargs):
+        """
+        A pass to perform only the basic skewing in order to have valid data
+        dependences.
+        """
+        skew_factor = configuration['skew_factor']
+        skew_dim, mapper, int_mapper = None, {}, []
+        total_int = {}
```
FabioLuporini: useless definition

devito/dse/rewriters.py
```diff
@@ -406,6 +408,38 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
+        intervals, sub_iterators, directions = cluster.ispace.args
+
+        for dim in cluster.ispace.dimensions:
+            if isinstance(dim, SpaceDimension):
+                mapper[dim] = dim + skew_factor*skew_dim
+                int_mapper.append(IntervalGroup([Interval(dim, -skew_factor*skew_dim,
+                                                -skew_factor*skew_dim)]))
+            elif dim.is_Time:
+                if isinstance(dim, TimeDimension):
```
FabioLuporini: maybe useless?

devito/dse/rewriters.py
```diff
@@ -406,6 +408,38 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
+                    skew_dim = dim
+
+        if skew_dim is None:
+            return cluster
+
+        total_int = IntervalGroup.generate('union', *int_mapper)
+
+        ispace = IterationSpace(total_int, sub_iterators, directions)
+
+        cluster._ispace = ispace
```
FabioLuporini: leftover, I hope/presume

devito/dse/rewriters.py
```diff
@@ -406,6 +408,38 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
+                                                -skew_factor*skew_dim)]))
+            elif dim.is_Time:
+                if isinstance(dim, TimeDimension):
+                    int_mapper.append(IntervalGroup([Interval(dim, 0, 0)]))
+                    skew_dim = dim
+
+        if skew_dim is None:
+            return cluster
+
+        total_int = IntervalGroup.generate('union', *int_mapper)
```
FabioLuporini: this variable, for homogeneity with the rest of the codebase, should be called just ``intervals``

devito/dse/rewriters.py
```diff
@@ -406,6 +408,38 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
+        """
+        skew_factor = configuration['skew_factor']
+        skew_dim, mapper, int_mapper = None, {}, []
+        total_int = {}
+        intervals, sub_iterators, directions = cluster.ispace.args
+
+        for dim in cluster.ispace.dimensions:
+            if isinstance(dim, SpaceDimension):
+                mapper[dim] = dim + skew_factor*skew_dim
+                int_mapper.append(IntervalGroup([Interval(dim, -skew_factor*skew_dim,
```
FabioLuporini: You don't need an IntervalGroup at this point.

Also, ``int_mapper`` should actually be called ``intervals``

devito/dse/rewriters.py
```diff
@@ -406,6 +408,38 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
+                                                -skew_factor*skew_dim)]))
+            elif dim.is_Time:
+                if isinstance(dim, TimeDimension):
+                    int_mapper.append(IntervalGroup([Interval(dim, 0, 0)]))
```
FabioLuporini: You don't need an IntervalGroup here

devito/dse/rewriters.py
```diff
@@ -406,6 +408,38 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
+        """
+        A pass to perform only the basic skewing in order to have valid data
+        dependences.
+        """
+        skew_factor = configuration['skew_factor']
+        skew_dim, mapper, int_mapper = None, {}, []
+        total_int = {}
+        intervals, sub_iterators, directions = cluster.ispace.args
+
+        for dim in cluster.ispace.dimensions:
```
FabioLuporini: I think you want to iterate over ``cluster.ispace.intervals`` (or even ``cluster.itintervals``)

then you can get the ``dim`` via ``i.dim``

devito/dse/rewriters.py
```diff
@@ -406,6 +408,38 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
+            if isinstance(dim, SpaceDimension):
```
FabioLuporini: the body of this loop, conceptually, should be
```
if i.dim.is_Time:
    ...
else:
    ...
```
ie, no need to be specifically a SpaceDimension. THe only special one is Time, and to be fair, I'd love to find a way such that it's not so special... but perhaps better for another day

devito/dse/rewriters.py
georgebisbas: Let me think

devito/dse/rewriters.py
```diff
@@ -406,6 +408,47 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
 
         return cluster.rebuild(processed)
 
+    @dse_pass
+    def _skewing(self, cluster, template, **kwargs):
+        """
+        A pass to perform only the basic skewing in order to have valid data
+        dependences.
+        """
+        skew_factor = configuration['skew_factor']
```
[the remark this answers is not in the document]
georgebisbas: in class State(object): ?

devito/dse/rewriters.py
```diff
@@ -406,6 +408,40 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
 
         return cluster.rebuild(processed)
 
+    @dse_pass
+    def _skewing(self, cluster, template, **kwargs):
+        """
+        Skew the accesses along the time Dimension.
+        """
+
+        MAX_SKEW_FACTOR = 8
```
FabioLuporini: No, not here

devito/dse/rewriters.py
```diff
@@ -406,6 +408,40 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
+        """
+        The maximum skew factor applied for the skewing pass.
```
FabioLuporini: Btw, why should there be a maximum skew factor? 
do you mean that in certain cases, for certain numbers of time iterations and values of skew factors, there might be runtime segfaults due to illegal accesses?

devito/dse/rewriters.py
```diff
@@ -406,6 +408,40 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
+                intervals.append(Interval(i.dim, 0, 0))
+                skew_dim = i.dim
+            else:
+                mapper[i.dim] = i.dim + skew_factor*skew_dim
+                intervals.append(Interval(i.dim, -skew_factor*skew_dim,
+                                          -skew_factor*skew_dim))
+
+        # TODO Can I avoid 2 rebuilds?
+        processed = xreplace_indices(cluster.exprs, mapper)
+        cluster.rebuild(processed)
```
FabioLuporini: What is the purpose of this rebuild? 
The effect of this line is 0. A rebuild returns something, which you're dropping on the floor, so... it's like lines 439-440 are not there

devito/dse/rewriters.py
```diff
@@ -406,6 +408,40 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
+
+        processed2 = xreplace_indices(cluster.exprs, IntervalGroup(intervals))
```
FabioLuporini: The second argument to ``xreplace_indices`` should **not** be an IntervalGroup.
Even though an exception isn't raised (that's an issue in ``xreplace_indices``), this is conceptually wrong

devito/dse/rewriters.py
FabioLuporini: I also think this docstring would benefit from a simple example -- given a certain input expression, show the effect of skewing

devito/dse/rewriters.py
FabioLuporini: Given the TODO comment and what you're doing here, my feeling is that immutability vs mutability isn't completely clear yet. 

With immutability, we *rebuild* objects, instead of changing their state. Rebuilding means creating a complete new object (in which potentially only some attributes have changed)

devito/symbolics/manipulation.py
```diff
@@ -226,6 +226,8 @@ def as_symbol(expr):
         return expr
     elif isinstance(expr, Indexed):
         return expr.base.label
+    elif isinstance(expr, Expr):
```
FabioLuporini: This should be dropped

tests/test_dse.py
```diff
@@ -43,6 +43,25 @@ def test_scheduling_after_rewrite():
...
+def test_skew_vs_advanced(nx, ny):
+    """Trivial testing for DSE skewing"""
+    nx = 10
+    ny = 10
+    timesteps = 1
+    grid = Grid(shape=(nx, ny))
+    u_skew = TimeFunction(name='u_skew', grid=grid)
+    u = TimeFunction(name='u', grid=grid)
+
+    eq_skew = Eq(u_skew.forward, u_skew.dx + u_skew.dy + u_skew.dy.dy)
```
FabioLuporini: reminder: we've discussed about how to simplify this code

devito/dse/rewriters.py
georgebisbas: > What is the purpose of this rebuild?
> The effect of this line is 0. A rebuild returns something, which you're dropping on the floor, so... it's like lines 439-440 are not there

This makes sense. My mistake.

devito/dse/rewriters.py
georgebisbas: Yes.

devito/dse/rewriters.py
georgebisbas: moved to right position

tests/test_dse.py
georgebisbas: I did it as presented in #time-tiling thread in Slack.

devito/dse/rewriters.py
```diff
@@ -406,6 +411,36 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
+            else:
+                mapper[i.dim] = i.dim + skew_factor*skew_dim
+                intervals.append(Interval(i.dim, -skew_factor*skew_dim,
+                                          -skew_factor*skew_dim))
+
+        ispace = IterationSpace(intervals, sub_iterators, directions)
+        cluster = Cluster(cluster.exprs, ispace, cluster.dspace, cluster.atomics,
+                          cluster.guards)
+        processed = xreplace_indices(cluster.exprs, mapper)
+        return cluster.rebuild(processed)
```
FabioLuporini: This way (including the lines above) still doesn't make much sense, conceptually

devito/symbolics/manipulation.py
georgebisbas: REMINDER TO MYSELF

devito/dse/rewriters.py
```diff
@@ -382,6 +382,11 @@ def _eliminate_inter_stencil_redundancies(self, cluster, template, **kwargs):
 
 class AggressiveRewriter(AdvancedRewriter):
 
+    DEFAULT_SKEW_FACTOR = 8
+    """
+    The maximum skew factor applied for the skewing pass.
```
FabioLuporini: wrong docstring

devito/dse/rewriters.py
```diff
@@ -406,6 +411,32 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
+        Skew the accesses along the time Dimension.
+        """
+
+        skew_factor = self.DEFAULT_SKEW_FACTOR
+
+        _, sub_iterators, directions = cluster.ispace.args
+
+        skew_dim, mapper, intervals = None, {}, []
+
+        for i in cluster.ispace.intervals:
```
FabioLuporini: what would happen here if the time dimension appeared **after** even a single non-time dimension?  you would multiply (line 431) skew_factor times None, which is clearly buggy

I know that basically always you see time, x, y ... in this order, but this is actually not strictly necessary the case

devito/dse/rewriters.py
```diff
@@ -406,6 +411,32 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
+                intervals.append(Interval(i.dim, 0, 0))
+                skew_dim = i.dim
+            else:
+                mapper[i.dim] = i.dim + skew_factor*skew_dim
+                intervals.append(Interval(i.dim, -skew_factor*skew_dim,
+                                          -skew_factor*skew_dim))
+
+        processed = xreplace_indices(cluster.exprs, mapper)
+        ispace = IterationSpace(intervals, sub_iterators, directions)
+        cluster = Cluster(processed, ispace, cluster.dspace, cluster.guards)
```
FabioLuporini: as I said, this is supposed to be ``guards=cluster.guards ``

devito/dse/rewriters.py
georgebisbas: Yes, I know what you mean, I should add a typical test with just `Functions`, right?

devito/dse/rewriters.py
georgebisbas: I had changed already, thanks, forgot to push :-(.
Had it commited locally.

devito/dse/rewriters.py
georgebisbas: Fixed.

devito/dse/rewriters.py
```diff
@@ -406,6 +411,36 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
 
         return cluster.rebuild(processed)
 
+    @dse_pass
+    def _skewing(self, cluster, template, **kwargs):
+        """
+        Skew the accesses along the time Dimension.
```
FabioLuporini: Add an example?

devito/dse/rewriters.py
FabioLuporini: no, it's not (just) about adding tests, I'm saying that this is broken if someone writes (for whatever reason -- and there are reasons) this piece of code:

```
grid = Grid(...)
x, y, z = grid.dimensions
t = grid.stepping_dim
n = Dimension(name='n')
f = Function(name='f', dimensions=(n, t, x, y, z), shape=(...))
...
```

georgebisbas: Can you please check again, now that I pushed some changes that were not here due to my HOME-pc rebase?
I think you may have missed this change.

georgebisbas: Also may you provide me with a typical operator for such a problem?

FabioLuporini: following up from the code snippet above

```
...
op = Operator(Eq(f[n, t+1, x, y, z], f[n, t, x, y, z] + 1))
...
```

some "true" tests in `test_subdomains.py`

there are others but I can't now easily pinpoint one

devito/dse/rewriters.py
```diff
@@ -406,6 +411,39 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
+        skew_factor = self.DEFAULT_SKEW_FACTOR
+
+        _, sub_iterators, directions = cluster.ispace.args
+
+        skew_dim, mapper, intervals = None, {}, []
+
+        # Check if time dim exists
+        for i in cluster.ispace.intervals:
+            if i.dim.is_Time:
+                skew_dim = i.dim
```
FabioLuporini: I think it's best if you also handle weird/crazy user input , like if there were 2 time dimensions. Pseudocode below

```
skew_dim = {i.dim for i in cluster.ispace.intervals if i.dim.is_Time}
try:
    skew_dim = skew_dim.pop()
except KeyError:
    # No time dimensions -> nothing to do
    return ...
if len(skew_dim) > 0:
    raise ValueError("suitable error msg")
```

devito/dse/rewriters.py
```diff
@@ -406,6 +411,39 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
+
+        if skew_dim is not None:
```
FabioLuporini: if you use something along the lines of the code snippet above, you won't need this ``if`` anymore

tests/test_dse.py
```diff
@@ -44,6 +44,46 @@ def test_scheduling_after_rewrite():
...
+    assert np.all(u_skew.data[0] == u.data[0])
+
+@pytest.mark.parametrize("nx, nt, ny, nz", [(5, 20, 6, 10), (4, 30, 16, 10),
+                                            (20, 20, 20, 10), (100, 100, 100, 10)])
+def test_complex_skew_cases_I(nx, nt, ny, nz):
+    """Trivial testing for DSE skewing"""
+    nx = nx
+    ny = ny
+    nz = nz
+    nt = nt
```
btrb: are those necessary ?

tests/test_dse.py
```diff
@@ -322,7 +362,46 @@ def test_contracted_alias_shape_after_blocking():
     # Leads to 2D aliases
     eqn = Eq(u.forward, ((u[t, x, y, z] + u[t, x, y+1, z+1])*3*f +
                          (u[t, x, y+2, z+2] + u[t, x, y+3, z+3])*3*f + 1))
-    op0 = Operator(eqn, dse='noop', dle=('advanced', {'openmp': True}))
+    op0 = Operator(eqn, dse='basic', dle=('advanced', {'openmp': True}))
+    op1 = Operator(eqn, dse='aggressive', dle=('advanced', {'openmp': True}))
+
+    y0_blk_size = op1.parameters[9]
+    z_size = op1.parameters[-1]
```
btrb: isn't it fragile ?

devito/dse/rewriters.py
```diff
@@ -369,6 +375,48 @@ def _extract_sum_of_products(self, cluster, template, **kwargs):
...
+    def _skewing(self, cluster, template, **kwargs):
+        """
+        Skew the accesses along the time Dimension.
+        """
+
+        skew_factor = self.DEFAULT_SKEW_FACTOR
+
+        _, sub_iterators, directions = cluster.ispace.args
+
+        skew_dim, mapper, intervals = None, {}, []
```
btrb: if i may suggest, i think `skew_dim` doesn't need to be initialized

also, maybe `mapper` and `intervals`' initialization could come "logically before" it's utilization, like before lines 399

tests/test_dse.py
georgebisbas: This is copied from the already existing test. This will be probably removed however.

tests/test_dse.py
georgebisbas: No they are not. Removed.

devito/dse/rewriters.py
georgebisbas: Should that be in pseudocode?

devito/dse/rewriters.py
georgebisbas: Got it. Thanks.

FabioLuporini:
As I was discussing with @navjotk , this PR should probably be closed as we realised that `skewing` is a clustering-level transformation, rather than a DSE-level transformation. To clarify:

* A DSE pass works "one Cluster at a time", which gets transformed in complete isolation.
* A clustering pass works (recursively) on groups of Clusters sharing an iteration space.

My making skewing a clustering pass we would catch all Clusters sharing (for example) the time dimension and we would be able to transform all of them coherently and in a single go.

Thoughts? Questions?

georgebisbas:
> Thoughts? Questions?
Yes we can close it for the moment, I can continue working on it on the background.

PR #863, opened 2019-07-01
Default FD order

mloubout:
Quick fix of default FD order that was always space instead of time for time dimensions. Will make a test.


devito/finite_differences/derivative.py
```diff
@@ -115,7 +115,9 @@ def __new__(cls, expr, *dims, **kwargs):
             orders = as_tuple(orders)
 
         # Finite difference orders
-        fd_orders = kwargs.get('fd_order', tuple([expr.space_order]*len(dims)))
+        fd_orders = kwargs.get('fd_order', tuple([expr.time_order if
```
mloubout: Note for reviewer, the `getattr` instead of `.` is necessary as in some case a dummy `_X0` dimension is the input (during solve for example) that does not have a `is_TIme` property.

FabioLuporini: I think I'd prefer a try-except (AttributeError),then.

Said that, not sure I understand why the default fd order should be different depending on whether it's a TimeFunction or not?

mloubout: It's not different it's just picking up the order of the expression, and if derivative wrt to time you need time_order not space_order.

>I think I'd prefer a try-except (AttributeError),then.

Not sure I do will think about it

FabioLuporini: you use getattr only if the attribute you're trying to access is a **variable** (or in general if you don't know statically what you will access) , but in this it's a constant (is_Time).

also, about this: 

> if derivative wrt to time you need time_order not space_order.

why do you say "need time_order" ? wouldn't a convention be better (eg, always default fd_order=space_order)? perhaps I'm the one missing something here

mloubout: The convention is you take the fd order of the dimension, if it's time it's time_order, which is what this fixes, I am really confused of what the issue is here. This not about wether it's a TImeFunction or a Function, it's about if it `.dt` or `.dx`.

FabioLuporini: > This not about wether it's a TImeFunction or a Function, it's about if it .dt or .dx
ah, OK, this wasn't clear :))) my bad

FabioLuporini: perhaps BS but...if you do

```
[ ... for d in new_dims]
```

instead of 
```
[... for d in dims]
```

then you shouldn't need the getattr, nor the try/except right....? these would be guaranteed to be Dimensions

devito/finite_differences/derivative.py
```diff
@@ -192,6 +193,6 @@ def evaluate(self):
                                    matvec=self.transpose, stagger=self.stagger)
         else:
             res = generic_derivative(expr, *self.dims, self.fd_order,
-                                     self.deriv_order, stagger=self.stagger,
+                                     self.deriv_order, stagger=self.stagger[0],
```
FabioLuporini: why this change?

mloubout: Single dimension so stagger is (stagger[0],) and need to pass it not as tuple to generic_derivative. Ignores the stagger setup without it

devito/finite_differences/derivative.py
mloubout: Not it's not guaranteed, new_dims is just a rearranged filtered version of the input wrt but it's still a Dummy if input is a Dummy.

FabioLuporini: what is a Dummy?

devito/finite_differences/derivative.py
FabioLuporini: what do you mean "by single dimension", and how do you know?

it feels really weird that you're passing ``*self.dims`` (ie all dimensions), but only one stagger value

devito/types/sparse.py
```diff
@@ -607,11 +607,11 @@ def _interpolation_indices(self, variables, offset=0):
                 ub = sympy.And(p <= d.symbolic_max + self._radius, evaluate=False)
                 condition = sympy.And(lb, ub, evaluate=False)
                 mapper[d] = ConditionalDimension(p.name, self._sparse_dim,
+                                                 spacing=d.spacing,
                                                  condition=condition, indirect=True)
 
             # Track Indexed substitutions
-            idx_subs.append(OrderedDict([(v, v.subs(mapper)) for v in variables
-                                         if v.function is not self]))
+            idx_subs.append(mapper)
```
FabioLuporini: lolz, really? what the hell did I do when I wrote it?

devito/types/sparse.py
```diff
@@ -607,11 +607,11 @@ def _interpolation_indices(self, variables, offset=0):
                 ub = sympy.And(p <= d.symbolic_max + self._radius, evaluate=False)
                 condition = sympy.And(lb, ub, evaluate=False)
                 mapper[d] = ConditionalDimension(p.name, self._sparse_dim,
+                                                 spacing=d.spacing,
```
FabioLuporini: compactable in two lines

devito/types/dimension.py
```diff
@@ -676,22 +676,24 @@ class ConditionalDimension(DerivedDimension):
...
         newobj._indirect = indirect
+        newobj._spacing = spacing or parent.spacing
         return newobj
 
     __xnew_cached_ = staticmethod(cacheit(__new_stage2__))
 
     @property
     def spacing(self):
-        return self.factor * self.parent.spacing
+        return self.factor * self._spacing
```
FabioLuporini: much more clear if you return directly ``self._spacing`` here, and change ``__new_stage2__`` such that ``newobj._spacing = factor*(spacing or parent.spacing)``

tests/test_operator.py
```diff
@@ -317,6 +317,29 @@ def test_sparsefunction_interp(self):
         # Exactly in the middle of 4 points, only 1 nonzero is 4
         assert sf1.data[0] == pytest.approx(1.0)
 
+    def test_sparsefunction_interp_dx(self):
+        """
+        Test interpolation of a SparseFunction from a Function
```
FabioLuporini: shouldn't these tests (this and those above) go into test_interpolation ?

devito/types/dimension.py
```diff
@@ -676,22 +676,24 @@ class ConditionalDimension(DerivedDimension):
     is_NonlinearDerived = True
     is_Conditional = True
 
-    def __new__(cls, name, parent, factor=None, condition=None, indirect=False):
+    def __new__(cls, name, parent, factor=None, condition=None, indirect=False,
+                spacing=None):
```
FabioLuporini: the docstring needs to be updated. You can probably use the ``spacing`` definition from other docstrings in this file

devito/finite_differences/derivative.py
mloubout: It's a sympy type, it's kinda like Symbol but unique ie `Dummy("x") == Dummy("x")` is False. Mainly sympy uses it as tmp variable in subs/solve/...

devito/types/dimension.py
FabioLuporini: though you also have to say optional -- defaults to parent's

devito/finite_differences/derivative.py
mloubout: stagger is a kwargs so can't pass *stagger.

>and how do you know?
because line above treats `len(dim)>1` this case is only for single dimension.

devito/types/sparse.py
mloubout: Yeah the dict was `{v(x,y,z) : v(idx1, idx2)}` which isn't great for expression, this is pn;y the indices dict that way.

tests/test_operator.py
mloubout: Not sure ,lots of sparse  tests here too never know where to put what

FabioLuporini: I'd say test_interpolation. Docstring may say more about why we want to test in particular interpolation from ``u.dx`` (instead of ``u``)

devito/types/sparse.py
FabioLuporini:
```
this is pn;y the indices dict that way.
```
ehm...sorry? :D

devito/finite_differences/derivative.py
FabioLuporini: you see, in its current form, the code doesn't make clear why we're using getattr. With a try-except, you'd be able to explain the Dummy thing as a comment in the except branch

devito/finite_differences/tools.py
```diff
@@ -161,6 +161,22 @@ def symbolic_weights(function, deriv_order, indices, dim):
...
 
+    # If staggered finited difference
+    if func.is_Staggered:
+        if stagger == left:
+            off = -.5
+        elif stagger == right:
+            off = .5
+        else:
+            off = 0
+        ind = list(set([(dim + int(i+.5+off) * dim.spacing)
```
FabioLuporini: no need for square brackets

devito/finite_differences/derivative.py
FabioLuporini: ah ok... I was expecting ``self.dims[0]``... that ``*self.dims`` confused me at first

devito/finite_differences/derivative.py
mloubout: But doesn't really fit nicely as need an extra loop to have a try except for each dimension then go back to kwargs.... that would be pretty ugly

PR #864, opened 2019-07-02
Separable

mloubout:
Introduce function separable in their dimensions.

WIP but working. Left to do

-  ~setup tests, not sure where and what thinking about it~ ADDED
-  ~more cases? Like mix +/*~ Not separable anymore so not gonna be added
-  ~Smarter FD for sparable functions~ For future, but may be useful
-  ~yask~ Fixed
-  ~data (like data access and write too)~ done


devito/types/dense.py
```diff
@@ -1311,3 +1311,35 @@ def parent(self):
...
+        name = kwargs.get('name')
+        grid = kwargs.get('grid')
+        new_obj = np.prod([Function(name=name+'_'+d.name, dimensions=(d,),
+                                    shape=(d.shape,), space_order=kwargs.get('space_order', 1))
+                           for d in grid.dimensions])
+        return new_obj
+        
+class SeparableTimeFunction(Differentiable):
+    """
+    A function separable in its dimensions, ie f(x,y,z) = f(x)*f(y)*f(z)
```
btrb: might take `t` into account at some point in the description maybe?

devito/types/dense.py
```diff
@@ -1334,22 +1355,62 @@ def __new__(self, *args, **kwargs):
...
+            func_list += [Function(name=name+'_'+names(nonseparated),
+                                   dimensions=nonseparated,
                                    shape=nonsep_shape, space_order=so)]
         return sep_op(func_list)
 
 
 class SeparableTimeFunction(Differentiable):
     """
-    A function separable in its dimensions, ie f(x,y,z) = f(x)*f(y)*f(z)
+    A function separable in its dimensions, ie f(t, x,y,z) = f(t)f(x)*f(y)*f(z)
```
btrb: typo ?

FabioLuporini:
what's the ultimate goal of this?

devito/types/dense.py
mloubout: yes

devito/types/dense.py
mloubout: yes

mloubout:
>what's the ultimate goal of this?

To use separable function :D

Like for example if you want a time-space dependent source q(t,x,y,z) but you know it's q(t)*q(x,yz) it is a bit overkill to create a full TimeFunction that uses (nt-1)*nx*ny*nz more memory. It can't really be used as an unknown though unless you specify which part is fixed, like I know q(t) in which case you can solve for it. Pretty useful for like custom point sources with "area of effect" but single time signature for example and can bypass inject. May actually be usefull for george too for time tiling

FabioLuporini:
one thing I don't understand is that ``__new__`` returns the sum/prof of plain Functions, rather than an actual "SeparableFunction" object. Right?  **If** yes (big if), then this looks flawed to me? otherwise, how could you implement eg

> * Smarter FD for spearable functions

 aside from this, I'm not sure we should introduce a new class for SeparableFunctions? why aren't these just new parameters of DiscreteFunction? don't get me wrong, maybe this is fine, maybe there's a rationale...just trying to assess the various alternatives. Also, this would imply that SparseFunctions can't be separable ?
Even weirder is the fact that you're inheriting from ``Differentiable`` instead of from one of the intermediate Function classes

mloubout:
>returns the sum/prof of plain Functions, rather than an actual "SeparableFunction" object. Right? If yes (big if), then this looks flawed to me? 

Ye it does, not sure what you mean by flawed there .

>why aren't these just new parameters of DiscreteFunction?
Because 99% of what's in Discrete is useless here. You don't want to have to put a bunch of if/cases everywhere for the shape, data size, allocation, ...... in case it is separable. Separable functions are jsut product/sums (with in future extra dimensions like `\sum_r(f(x,r)g(r,z))` for example).

> Also, this would imply that SparseFunctions can't be separable ?
Np Sparse funtions are too just didn't get there yet, I actually think it's a smarter way to implement sparse function (at least for source). Do interpolation once to getthe spatial distribution src_pos(x,y,z) and then just pout back in stencil `u = u + src(t)*src_pos(x,y,z)`.

>Even weirder is the fact that you're inheriting from Differentiable instead of from one of the intermediate Function classes

That's current state, yes but this is because for some reason `Mul(cls, *func_list)` errors with some weird property missing so I had to put Differentiable for now

FabioLuporini:
> Like for example if you want a time-space dependent source q(t,x,y,z) but you know it's q(t)**q(x,y,z) it is a bit overkill to create a full TimeFunction that uses (nt-1)*nx*ny*nz more memory

you mean "SparseTimeFunction" here right? in which case, it's not such a huge waste...is it?

However, maybe I see why you want to return multiple Functions from ``__new__``... 

OK, I convinced myself that I have to think about this carefully

note: if you instead return an actuale SeparableFunction object from ``__new__``, then things will probably break up soon

mloubout:
>you mean "SparseTimeFunction" here right? in which case, it's not such a huge waste...is it?

No I mean TimeFunction, like a full time-space source wavefield.

devito/types/dense.py
FabioLuporini: fix it?

devito/types/dense.py
```diff
@@ -1311,3 +1312,198 @@ def parent(self):
...
+
+    Examples
+    --------
+
+    Creation
+
+    >>> from devito import Grid, TimeFunction
+    >>> grid = Grid(shape=(4, 4, 5))
+    >>> x, y, z = grid.dimensions
+    >>> a = SeparableFunction(name='a', grid=grid)
```
FabioLuporini: No need to assign it to ``a``. Same for the lines below. Same for the SeparableTimeFunction docstring

devito/types/dense.py
```diff
@@ -1311,3 +1312,198 @@ def parent(self):
         return self._parent
 
     _pickle_kwargs = Function._pickle_kwargs + ['parent']
+
+
+class SeparableFunction(Differentiable):
+    """
+    A function separable in its dimensions, ie f(x,y,z) = f(x)*f(y)*f(z)
```
FabioLuporini: I'm not gonna repeat the usual things over and over again but... **keywords should always be capitalized**

function -> Function
dimension -> Dimension

and yes, we also do plurals:

functions -> Functions

etc

FabioLuporini: ie -> i.e.

devito/types/dense.py
```diff
@@ -1311,3 +1312,198 @@ def parent(self):
...
+
+    Takes the same parameters as Function and
```
FabioLuporini: Always first person
```
Take the same ...
```

devito/types/dense.py
```diff
@@ -1311,3 +1312,198 @@ def parent(self):
...
+
+    Parameters
+    ----------
+        sep: 'sum' or 'prod'
```
FabioLuporini: de-indent

the type is str

values (sum, prod) should be discussed in the description

devito/types/dense.py
```diff
@@ -1311,3 +1312,198 @@ def parent(self):
...
+            type of separation as a product or sum of function
```
FabioLuporini: Capital letter

FabioLuporini: Functions

devito/types/dense.py
```diff
@@ -1311,3 +1312,198 @@ def parent(self):
...
+        separated: tuple or Dimension
```
FabioLuporini: Dimension or tuple of Dimensions

devito/types/dense.py
```diff
@@ -1311,3 +1312,198 @@ def parent(self):
...
+            Dimension in which the function is separable
```
FabioLuporini: Dimensions
Function

full stop

mloubout:
>q.space_order picks Differentiable.space_order which is 100 by default

That one is actually because of your change I didn't realize, space_order=0 doesn't work because 0 or 100 is 100. Removed `or 100` that actually is not needed

devito/types/dense.py
```diff
@@ -1311,3 +1313,199 @@ def parent(self):
...
+    >>> SeparableTimeFunction(name='a', grid=grid, space_order=8, time_order=2)
+    a_t(t)*a_x(x)*a_y(y)*a_z(z)
+    >>> SeparableTimeFunction(name='b', grid=grid, sep='sum')
+    b_t(t) + b_x(x) + b_y(y) + b_z(z)
+    >>> SeparableTimeFunction(name='c', grid=grid, sep='sum', separated=t)
+    c_t(t) + c_xyz(x, y, z)
+    >>> SeparableTimeFunction(name='d', grid=grid, separated=(t, x))
+    d_t(t)*d_x(x)*d_yz(y, z)
+    """
+    def __new__(cls, *args, **kwargs):
```
FabioLuporini: I also believe that much of the code in this ``__new__`` can be performed by the superclass via ``super``. We should really aim to achieve this, as currently there's too much code duplication

devito/types/separable.py
```diff
@@ -0,0 +1,192 @@
...
+        self._sep_op = spearation_op[kwargs.get('sep', 'prod')]
+
+        # Conventional Function inputs
+        self._grid = kwargs.get('grid')
+        self._name = kwargs.get('name')
+        self._space_order = kwargs.get('space_order', 1)
+
+        # Initialize subfunctions
+        # Separable dimensions
+        sep = self.separated_dimensions(**kwargs)
```
FabioLuporini: for convention with the rest of the code, since this is a hook to be specialized by subclasses, I'd call it ``__setup_separated__``

devito/types/separable.py
```diff
@@ -0,0 +1,192 @@
...
+    >>> SeparableFunction(name='a', grid=grid)
+    a_x(x)*a_y(y)*a_z(z)
+    >>> SeparableFunction(name='b', grid=grid, sep='sum')
+    b_x(x) + b_y(y) + b_z(z)
+    >>> SeparableFunction(name='c', grid=grid, sep='sum', separated=x)
+    c_x(x) + c_yz(y, z)
+    """
+    def __init__(self, *args, **kwargs):
+        # Is it sum or product separable
+        self._sep_op = spearation_op[kwargs.get('sep', 'prod')]
```
FabioLuporini: just ``operator`` or ``op`` ?

"_sep" is implicit

devito/types/separable.py
```diff
@@ -0,0 +1,192 @@
...
+        self._space_order = kwargs.get('space_order', 1)
+
+        # Initialize subfunctions
+        # Separable dimensions
+        sep = self.separated_dimensions(**kwargs)
+        nonseparated = tuple(d for d in self.grid.dimensions if d not in sep)
+        dims = tuple(d for d in sep) + (nonseparated,)
+        self._dims = dims
+
+        func_list = [self.function(d, **kwargs) for d in sep]
```
FabioLuporini: same story, ``__setup_funclist__``.

devito/types/separable.py
```diff
@@ -0,0 +1,192 @@
...
+
+class SeparableFunction(Differentiable, AbstractCachedSymbol):
+    """
+    A Function separable in its Dimensions, ie f(x,y,z) = f(x)*f(y)*f(z)
+
+    Take the same parameters as a Function plus
+
+    Parameters
+    ----------
+    sep: str
```
FabioLuporini: a space is expected between the variable name and the :

devito/types/separable.py
```diff
@@ -0,0 +1,192 @@
...
+        Type of separation as a product or sum of Function, 'sum' or 'prod'.
+    separated: Dimension or tuple of Dimensions
```
FabioLuporini: a space is expected between the variable name and the :

devito/types/separable.py
FabioLuporini: also, ...sum of FunctionS (plural)

FabioLuporini: how about calling it "op" or "operator" ?

also, probably "separated" should come first

devito/types/separable.py
```diff
@@ -0,0 +1,192 @@
...
+        dims = tuple(d for d in sep) + (nonseparated,)
+        self._dims = dims
+
+        func_list = [self.function(d, **kwargs) for d in sep]
+        if len(nonseparated) > 0:
+            func_list.append(self.function(nonseparated, **kwargs))
+
+        expr = self._sep_op(*func_list)
+
+        dims = tuple(d for d in sep) + (nonseparated,)
```
FabioLuporini: tuple(sep) ?

devito/types/separable.py
```diff
@@ -0,0 +1,192 @@
...
+    def __getitem__(self, dim):
+        return self.subfunctions[dim]
+
+    @property
+    def data(self):
+        return tuple(f.data for f in self.subfunctions.value)
+
+    def subs(self, *args, **kwargs):
+        new = super(SeparableFunction, self).subs(*args, **kwargs)
+        new._expr = self.expr.subs(*args, **kwargs)
```
FabioLuporini: changing state, so this is wrong

I don't see why we need to overwrite subs or xreplace. If things are done properly, we should just be able to seamlessly leverage sympy's xreplace

maybe the issue is that ``expr`` isn't part of the ``args``  of the SeparableFunction ?

otherwise [it should just work](https://github.com/sympy/sympy/blob/master/sympy/core/basic.py#L1201)

EDIT -- I think I see what you're trying to work around. ``expr`` is constructed by ``__init__`` from some other arguments... so the solution is that either these other arguments are made part of ``args``, or (this is worse, though) we apply subs/xreplace to them, and then we return ``SeparableFunction(**new_kwargs)``

devito/types/separable.py
```diff
@@ -0,0 +1,192 @@
...
+        return new
+
+    def xreplace(self, rules):
```
FabioLuporini: same considerations of ``subs`` apply here

devito/types/separable.py
```diff
@@ -0,0 +1,192 @@
...
+
+class SeparableTimeFunction(SeparableFunction):
+    """
+    A TimeFunction separable in its Dimensions, ie f(t,x,y,z) = f(t)*f(x)*f(y)*f(z)
+
+    Take the same parameters as a TimeFunction plus
+
+    Parameters
+    ----------
+    sep: str
```
FabioLuporini: space

devito/types/separable.py
```diff
@@ -0,0 +1,192 @@
...
+        Type of separation as a product or sum of Function, 'sum' or 'prod'.
+    separated: Dimension or tuple of Dimensions
```
FabioLuporini: space

devito/types/separable.py
```diff
@@ -0,0 +1,192 @@
...
+        self._subfunctions = {d: f for d, f in zip(dims, func_list)}
+        self._dims = dims
+        self._expr = expr
+        self._fd = self.expr._fd
+
+    def function(self, dimensions, **kwargs):
+        dims = as_tuple(dimensions)
+        func_type = Function
+        names = "".join(d.name for d in dims)
+        shape = tuple(self.grid.dimension_map[d].loc for d in dims)
```
FabioLuporini: so, currently, SeparableFunctions won't work if there's a generic dimensions that doesn't appear in ``grid`` , correct?

devito/types/separable.py
```diff
@@ -0,0 +1,192 @@
...
+        names = "".join(d.name for d in dims)
+        save = kwargs.get('save', None)
+        shape = tuple(save or self.time_order+1 if d.is_Time
+                      else self.grid.dimension_map[d].loc for d in dims)
+        kwargs["name"] = '%s_%s' % (self.name, names)
+        func = func_type(dimensions=dims, shape=shape, **kwargs)
+        return func
+
+
+spearation_op = {'sum': Add, 'prod': Mul}
```
FabioLuporini: this should be a (private) class attribute of SeparableFunction

devito/types/separable.py
mloubout: >a space is expected between the variable name and the :

wait really, how french... :D

>also, probably "separated" should come first

I don't know, "separated" is the less-likely to be used, general case is don't specify dim only if separable or not and if + or *

>how about calling it "op" or "operator" ?

Can find a better name yes

devito/types/separable.py
mloubout: But isn't our convention that `__setup_separated__` would be a class method whil this is just an attribute

devito/types/separable.py
mloubout: same

devito/types/separable.py
mloubout: Yes good catch, need to add that case

devito/types/separable.py
FabioLuporini: > wait really, how french... :D

just numpydoc convention 😎 

> I don't know, "separated" is the less-likely to be used, general case is don't specify dim only if separable or not and if + or *

oki

> Can find a better name yes

I think ``op`` or ``operator`` is preferable

devito/types/separable.py
FabioLuporini: that's how it started off, yes... then I had to extend it to instance methods too. See for example ``__halo_setup__`` and ``__padding_setup__`` -- these just can't be class methods. Same for ``__setup_staggered__`` I seem to remember.

Basically, the idea is that we use the convention ``__setup_X__`` to highlight (class, instance) methods that are expected to be specialized by subclasses.

devito/types/basic.py
```diff
@@ -463,6 +463,61 @@ class AbstractFunction(sympy.Function, Basic, Pickable):
     is_AbstractFunction = True
 
 
+class AbstraceCachedAnonymousFunction(AbstractFunction, Cached, Evaluable):
```
FabioLuporini: why can you not just use AbstractCachedFunction, from which Array ( a data-free tensor symbol) inherits?

devito/types/separable.py
```diff
@@ -0,0 +1,221 @@
...
+        names = "".join(d.name for d in dims)
+        save = kwargs.get('save', None)
+        shape = tuple(save or self.time_order+1 if d.is_Time
+                      else self.grid.dimension_map[d].loc for d in dims)
+        kwargs["name"] = '%s_%s' % (self.name, names)
+        func = func_type(dimensions=dims, shape=shape, **kwargs)
+        return func
+
+    @classmethod
+    def __indices_setup__(cls, **kwargs):
```
FabioLuporini: this is a copy-paste, and therefore it's not acceptable

devito/types/separable.py
```diff
@@ -0,0 +1,221 @@
...
+
+    def __getitem__(self, dim):
+        return self.subfunctions[dim]
+
+    @property
+    def data(self):
+        return tuple(f.data for f in self.subfunctions.value)
+
+    @classmethod
+    def __indices_setup__(cls, **kwargs):
```
FabioLuporini: this is a copy-paste, and therefore it's not acceptable

devito/finite_differences/derivative.py
```diff
@@ -207,6 +206,30 @@ def evaluate(self):
             res = generic_derivative(expr, *self.dims, self.fd_order,
                                      self.deriv_order, stagger=self.stagger[0],
                                      matvec=self.transpose)
-        for e in self._eval_at:
-            res = res.xreplace(e)
         return res
+
+
+class Derivative_at(Derivative):
```
FabioLuporini: Never use _ in class names

I **think** this can be made general (``At`` or ``Eval``), ie not specifically related to Derivative

mloubout:
>Why don't we just return a product/sum of two Functions (or TimeFunctions, or whatever) ; IOW, how is it useful to have a first-class SeparableFunction object? I thought it would be useful if you wanna do smarter FD, but otherwise...?

Well no, it's not just for FD it's also extremely useful memory-wise for example, as if you have a separable instead of conventional function you need `nt+nx` instead of `nt * nx` for example. This is one of the main reason it is its own class as it needs data read/write capabilities that you don't really have with just a product without asking the user to play with `.args` and such

devito/types/basic.py
mloubout: There is a lot of stuff (like halo/shape/indices) setups that are unnecessary in AbstractCachedFunction that make it nasty.

devito/types/separable.py
mloubout: This has to go here for debug, same for other one

devito/finite_differences/derivative.py
mloubout: Sure

devito/finite_differences/derivative.py
```diff
@@ -207,6 +206,36 @@ def evaluate(self):
...
+    is used to delay evaluation of subs and xreplace of Derivative and is
+    particularly useful for SparseFunctions injection and interpolations
+    that involve derivatives.
+    
+    """
+    def __new__(cls, expr, variables, point, **kwargs):
+        variables = as_tuple(variables)
+        point = as_tuple(point)
+
+        obj = Derivative.__new__(cls, expr.expr, *expr.dims, deriv_order=expr.deriv_order,
```
FabioLuporini: I think this makes no sense. ``Eval`` should wrap ``Derivative`` (or any another evaluable object), *not become* a Derivative itself.

Also, the fact that you're adding attributes that you don't have when you create a Derivative (e.g., ``_rule``) is a hack. Instead, if Eval is a first-class citizen, it's absolutely fine to keep ``_rule`` around

devito/types/separable.py
```diff
@@ -0,0 +1,195 @@
...
+
+    @property
+    def grid(self):
+        return self._grid
+
+    @property
+    def space_order(self):
+        return self._space_order
+
+    def __setup_dimensions__(self, **kwargs):
```
FabioLuporini: ``__setup_separated`` ?

``dimensions`` is an alias of ``indices`` (attribute-wise), so I wouldn't use it in a method name (we already have ``__setup_indices__`` right?

devito/types/basic.py
FabioLuporini: override and make them no-op?

FabioLuporini:
> Well no, it's not just for FD it's also extremely useful memory-wise for example, as if you have a separable instead of conventional function you need nt+nx instead of nt * nx for example. This is one of the main reason it is its own class as it needs data read/write capabilities that you don't really have with just a product without asking the user to play with .args and such

don't get me wrong, I obviously understand the part about memory efficiency :) and yep, sorry, I was being stupid with my previous comment. Scratch that!

devito/types/basic.py
FabioLuporini: I think we should avoid a new Abstract class at all costs (btw, typo: Abstrace...)

devito/types/separable.py
```diff
@@ -0,0 +1,195 @@
...
+    @property
+    def evaluate(self):
+        return self.expr.evaluate
+
+    @property
+    def expr(self):
+        return self.args[0]
+
+    @property
+    def dims(self):
```
FabioLuporini: this doesn't have any sense...users expect ``dimensions`` and ``indices``, not ``dims``. This is a result of not subclassing the right thing

FabioLuporini:
My feeling is that, simply, this will never work unless we subclass from ``DiscreteFunction`` -- things will start breaking as soon as we add in staggered, mpi, and more. There is also a high probability the API will become inconsistent (as I pointed out in a comment)

If you subclass DiscreteFunction, you have all the hooks you need at your disposal to treat data specially. 

If ``q`` is your SeparableFunction, what do you want ``q.shape`` to return ? 

My feeling is that the API of SeparableFunction should be exactly the API of DiscreteFunction -- with specialization (or raising exceptions) if/where necessary

mloubout:
> things will start breaking as soon as we add in staggered, mpi, and more

If it works for Function this won't break anything, that's the whole point each subfunction takes care of itself this doesn't break anything.

>If you subclass DiscreteFunction, you have all the hooks you need at your disposal to specialize the construction of a SeparableFunction. I don't see where the issue is, based on the current implementation

This is the opposite, this would require to litteraly rewrite every single attribute because they don't behave the same way which makes it kinda useless.

>If q is your SeparableFunction, what do you want q.shape to return ? I assume the cumulative shape

Same as most of property here, shape of each subfunction

>My feeling is that the API of SeparableFunction should be exactly the API of DiscreteFunction -- with specialization (or raising exceptions) if/where necessary

The if/where is always, I really disagree with that. I should be able to avoid an extra abstrat object and use an existing one but  I really don't like it being a DisrecteFunction, it's just gonna make everythign super messy

devito/types/separable.py
```diff
@@ -0,0 +1,195 @@
...
+    b_x(x) + b_y(y) + b_z(z)
+    >>> SeparableFunction(name='c', grid=grid, op='sum', separated=x)
+    c_x(x) + c_yz(y, z)
+    """
+    _spearation_op = {'sum': Add, 'prod': Mul}
+
+    def __init__(self, *args, **kwargs):
+        if not self._cached():
+            # Is it sum or product separable
+            self._sep_op = self._spearation_op[kwargs.get('op', 'prod')]
```
FabioLuporini: ``'prod'`` or ``Mul`` ?

devito/types/separable.py
```diff
@@ -0,0 +1,195 @@
...
+
+            # Conventional Function inputs
+            self._grid = kwargs.get('grid')
+            self._space_order = kwargs.get('space_order', 1)
+
+            # Initialize subfunctions
+            # Separable dimensions
+            sep_d, nonsep_d = self.__setup_dimensions__(**kwargs)
+            self._dims = sep_d + (nonsep_d,)
```
FabioLuporini: to be renamed

devito/types/separable.py
```diff
@@ -0,0 +1,195 @@
...
+
+            func_list = self.__setup_functions__(sep_d, nonsep_d, **kwargs)
+
+            expr = self._sep_op(*func_list)
+
+            self._subfunctions = {d: f for d, f in zip(self.dims, func_list)}
```
FabioLuporini: Note that the ``_sub_functions`` in SparseFunction is used in a slightly (subtly) different way. 
I don't like that we alias the same name (_subfunctions/_sub_functions) to do different things in different function objects.

So...other name? ``_sep_functions`` ?

devito/types/separable.py
```diff
@@ -0,0 +1,195 @@
...
+            # Is it sum or product separable
+            self._sep_op = self._spearation_op[kwargs.get('op', 'prod')]
+
+            # Conventional Function inputs
+            self._grid = kwargs.get('grid')
+            self._space_order = kwargs.get('space_order', 1)
+
+            # Initialize subfunctions
+            # Separable dimensions
+            sep_d, nonsep_d = self.__setup_dimensions__(**kwargs)
```
FabioLuporini: argh...just noticed two very important things:

1) 'setup' goes after ! it should have been ``__dimensions_setup__`` ...
2) it's actually ``__indices_setup__`` -- this is how it's called in all other classes of the hierarchy

devito/types/separable.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def __setup_functions__(self, sep_d, nonsep_d, **kwargs):
+
+        func_list = [self.function(d, **kwargs) for d in sep_d]
+        if len(nonsep_d) > 0:
+            func_list.append(self.function(nonsep_d, **kwargs))
+
+        return func_list
+
+    def function(self, dimensions, **kwargs):
```
FabioLuporini: this should be private

also, let's change it into something such that it's easier to grasp what's going on. ``_make_sep_function`` ?

devito/types/separable.py
```diff
@@ -0,0 +1,195 @@
...
+from devito.tools import as_tuple
+from devito.types.basic import AbstraceCachedAnonymousFunction
+from devito.types.dense import Function, TimeFunction
+
+__all__ = ['SeparableFunction', 'SeparableTimeFunction']
+
+
+class SeparableFunction(Differentiable, AbstraceCachedAnonymousFunction):
+    """
+    A Function separable in its Dimensions, ie f(x,y,z) = f(x)*f(y)*f(z)
```
georgebisbas: Is this always the case?
You may also have f(x,y,z) = f(x)*f(y,z) right?
```
  >>> SeparableFunction(name='c', grid=grid, op='sum', separated=x)
    c_x(x) + c_yz(y, z)
```

devito/types/separable.py
```diff
@@ -0,0 +1,195 @@
...
+        return tuple(f.data for f in self.subfunctions.value)
+
+    @property
+    def args(self):
+        return self._args
+
+
+class SeparableTimeFunction(SeparableFunction):
+    """
+    A TimeFunction separable in its Dimensions, ie f(t,x,y,z) = f(t)*f(x)*f(y)*f(z)
```
georgebisbas: same case with before...

tests/test_operator.py
```diff
@@ -390,6 +390,85 @@ def test_sparsetimefunction_inject_dt(self):
...
+        # Rec as separable function
+        # q(x,y) is the spatial weight ie .inject(3)
+        # and only needs the coordinates
+        sf2 = SparseFunction(name='s', grid=grid, npoint=1)
+        sf2.coordinates.data[0, :] = (0.45, 0.45)
+        sf2.data[:] = 1.
+        eq_src = sf2.inject(q[(x, y)], expr=sf2)
+        # Operator. The expression is then
+        # - First get the weights sf.inject
+        # - Secnd the stencil is simply u += q[time]*q[(x,y)]
```
georgebisbas: Secondly*

tests/test_operator.py
```diff
@@ -390,6 +390,85 @@ def test_sparsetimefunction_inject_dt(self):
...
+        op2 = Operator(eq_src + [Inc(q[t], u*q[(x, y)])])
+        op2(time_m=1, time_M=3)
+        # Both result should be the same
```
georgebisbas: results*

PR #865, opened 2019-07-02
miniconda

mloubout:
back to latest version, version that my hard coding was skiping is not there anymore so we shouldn't need that old version anymore

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,9 +17,9 @@ matrix:
           sources:
             - ubuntu-toolchain-r-test     # For gcc 4.9, 5 and 7
           packages:
-            - gcc-7
-            - g++-7
-      env: DEVITO_ARCH=gcc-7 DEVITO_OPENMP=0 INSTALL_TYPE=pip_setup RUN_EXAMPLES=False MPI_INSTALL=1
+            - gcc-4.9
+            - g++-4.9
+      env: DEVITO_ARCH=gcc-4.9 DEVITO_OPENMP=0 INSTALL_TYPE=pip_setup RUN_EXAMPLES=False MPI_INSTALL=1
     - python: 3.7
       dist: xenial
       os: linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,18 @@ env:
 
 matrix:
   include:
-    - os: linux
-      python: "3.6"
+    - python: 3.5
+      os: linux
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test     # For gcc 4.9, 5 and 7
+          packages:
+            - gcc-4.9
+            - g++-4.9
+      env: DEVITO_ARCH=gcc-4.9 DEVITO_OPENMP=0 INSTALL_TYPE=pip_setup RUN_EXAMPLES=False MPI_INSTALL=0
+    - python: 3.6
+      os: linux
       addons:
         apt:
           sources:
@@ -20,8 +30,8 @@ matrix:
             - gcc-4.9
             - g++-4.9
       env: DEVITO_ARCH=gcc-4.9 DEVITO_OPENMP=0 INSTALL_TYPE=pip_setup RUN_EXAMPLES=False MPI_INSTALL=1
-    - os: linux
-      python: "3.7"
+    - python: 3.7
+      os: linux
       addons:
         apt:
           sources:
@@ -30,8 +40,8 @@ matrix:
             - gcc-5
             - g++-5
       env: DEVITO_ARCH=gcc-5 DEVITO_OPENMP=0 RUN_EXAMPLES=False INSTALL_TYPE=conda MPI_INSTALL=0
-    - os: linux
-      python: "3.7"
+    - python: 3.7
+      os: linux
       addons:
         apt:
           sources:
@@ -40,8 +50,8 @@ matrix:
             - gcc-8
             - g++-8
       env: DEVITO_ARCH=gcc-8 DEVITO_OPENMP=1 OMP_NUM_THREADS=2 RUN_EXAMPLES=True INSTALL_TYPE=conda MPI_INSTALL=1 MPI_RESTRAIN=1
-    - os: linux
-      python: "3.7"
+    - python: 3.7
+      os: linux
       addons:
         apt:
           sources:
@@ -50,8 +60,8 @@ matrix:
             - gcc-7
             - g++-7
       env: DEVITO_ARCH=gcc-7 DEVITO_OPENMP=0 DEVITO_BACKEND=yask YC_CXX=g++-7 INSTALL_TYPE=conda RUN_EXAMPLES=False MPI_INSTALL=1
-    - os: linux
-      python: "3.7"
+    - python: 3.7
+      os: linux
       addons:
         apt:
           sources:
@@ -60,13 +70,6 @@ matrix:
             - gcc-7
             - g++-7
       env: DEVITO_ARCH=gcc-7 DEVITO_OPENMP=0 DEVITO_BACKEND=ops YC_CXX=g++-7 INSTALL_TYPE=conda RUN_EXAMPLES=False MPI_INSTALL=0
-  allow_failures:
-    - os: linux
-      python: "2.7"
-      env: DEVITO_ARCH=gcc-4.9 DEVITO_OPENMP=0
-    - os: osx
-      python: "2.7"
-      env: DEVITO_ARCH=clang DEVITO_OPENMP=0
 
 addons:
   apt:
@@ -74,20 +77,19 @@ addons:
       - ubuntu-toolchain-r-test     # For gcc 4.9, 5 and 7
 
 before_install:
-  # Setup anaconda
-  - if [[ $INSTALL_TYPE == 'conda' ]]; then
-	  wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh;
-	  bash miniconda.sh -b -p $HOME/miniconda;
-	  export PATH="$HOME/miniconda/bin:$PATH";
-	  hash -r;
-	  conda config --set always_yes yes --set changeps1 no;
-	  conda update -q conda;
-	  # Useful for debugging any issues with conda
-	  conda info -a;
-	fi
   - if [[ "$MPI_INSTALL" == '1' ]]; then
       sudo apt-get install -y -q mpich libmpich-dev;
       export MPICH_CC=$DEVITO_ARCH;
+    fi
+  # Setup anaconda
+  - if [[ $INSTALL_TYPE == 'conda' ]]; then
+      wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh;
+      bash miniconda.sh -b -p $HOME/miniconda;
+      export PATH="$HOME/miniconda/bin:$PATH";
+      hash -r;
+      conda config --set always_yes yes --set changeps1 no;
+      conda update -q conda;
+      conda info -a;
     fi
 
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ matrix:
             - gcc-4.9
             - g++-4.9
       env: DEVITO_ARCH=gcc-4.9 DEVITO_OPENMP=0 INSTALL_TYPE=pip_setup RUN_EXAMPLES=False MPI_INSTALL=1
-    - python: 3.7
+    - python: 3.6
       os: linux
       addons:
         apt:
@@ -40,7 +40,7 @@ matrix:
             - gcc-5
             - g++-5
       env: DEVITO_ARCH=gcc-5 DEVITO_OPENMP=0 RUN_EXAMPLES=False INSTALL_TYPE=conda MPI_INSTALL=0
-    - python: 3.7
+    - python: 3.6
       os: linux
       addons:
         apt:
@@ -50,7 +50,7 @@ matrix:
             - gcc-8
             - g++-8
       env: DEVITO_ARCH=gcc-8 DEVITO_OPENMP=1 OMP_NUM_THREADS=2 RUN_EXAMPLES=True INSTALL_TYPE=conda MPI_INSTALL=1 MPI_RESTRAIN=1
-    - python: 3.7
+    - python: 3.6
       os: linux
       addons:
         apt:
@@ -60,7 +60,7 @@ matrix:
             - gcc-7
             - g++-7
       env: DEVITO_ARCH=gcc-7 DEVITO_OPENMP=0 DEVITO_BACKEND=yask YC_CXX=g++-7 INSTALL_TYPE=conda RUN_EXAMPLES=False MPI_INSTALL=1
-    - python: 3.7
+    - python: 3.6
       os: linux
       addons:
         apt:
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,20 +17,9 @@ matrix:
           sources:
             - ubuntu-toolchain-r-test     # For gcc 4.9, 5 and 7
           packages:
-            - gcc-4.9
-            - g++-4.9
-      env: DEVITO_ARCH=gcc-4.9 DEVITO_OPENMP=0 INSTALL_TYPE=pip_setup RUN_EXAMPLES=False MPI_INSTALL=1
-    - python: 3.7
-      dist: xenial
-      os: linux
-      addons:
-        apt:
-          sources:
-            - ubuntu-toolchain-r-test     # For gcc 4.9, 5 and 7
-          packages:
-            - gcc-4.9
-            - g++-4.9
-      env: DEVITO_ARCH=gcc-4.9 DEVITO_OPENMP=0 INSTALL_TYPE=pip_setup RUN_EXAMPLES=False MPI_INSTALL=1
+            - gcc-7
+            - g++-7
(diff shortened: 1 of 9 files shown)


Dockerfile.pipelines
```diff
@@ -93,10 +98,10 @@ RUN chown -R devito /home/devito
 USER devito
 WORKDIR /home/devito
 
-RUN if [ $installWithPip == "true" ] ; then \ 
-      python setup.py install --user ; \
+RUN if [ $installWithPip == "true" ] ; then \
+      python3 setup.py install --user ; \
       if [ $MPI_INSTALL == '1' ] ; then \
-        pip install --user -e .[extras] ; \
+        pip3 install --user -e .[extras] ; \
```
tmbgreaves:
```suggestion
        python3 -m pip install --user -e .[extras] ; \
```

This seems to be the way the world is going..

tmbgreaves:
Versions on Travis using pip setup appear to be falling over with a pytest version issue -- looking for >3.6 but finding version 3.3. I guess 3.3 is the version preinstalled in the Travis environment and needs upgrading?

tmbgreaves:
The package problem on the pip installer is coming from our base image -- we're building on Ubuntu Xenial which only has python 3.5, and we're requesting 3.6. The fix would be to either regress to 3.5, or to bump the base image up to Bionic which has 3.6 and 3.7 available. Moving to bionic sounds worthwhile anyway, as Xenial is pretty old now, but would need a few more changes to the Dockerfile (which I can easily do if that would be useful). Let me know which way you'd prefer to go on this :-)

.travis.yml
```diff
@@ -60,37 +70,27 @@ matrix:
...
-  - conda info -a
   - if [[ "$MPI_INSTALL" == '1' ]]; then
       sudo apt-get install -y -q mpich libmpich-dev;
       export MPICH_CC=$DEVITO_ARCH;
     fi
+  # Setup anaconda
+  - if [[ $INSTALL_TYPE == 'conda' ]]; then
+      wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh;
+      bash miniconda.sh -b -p $HOME/miniconda;
+      export PATH="$HOME/miniconda/bin:$PATH";
```
tmbgreaves:
```suggestion
      export PATH="/usr/local/miniconda/bin:$PATH";
```
Path fix for different install location

Dockerfile.pipelines
```diff
@@ -78,11 +78,16 @@ RUN if [ "$MPI_INSTALL" == "1" ] ; then \
     fi
 
 # Set up conda
-RUN wget --no-verbose https://repo.continuum.io/miniconda/Miniconda3-4.5.4-Linux-x86_64.sh -O /tmp/miniconda.sh
-RUN bash /tmp/miniconda.sh -b -p /usr/local/miniconda
+RUN if [ $installWithPip == "true" ] ; then \ 
+      apt-get install -y python3.6-dev python3-pip ; \
+    else \
+      wget --no-verbose https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O /tmp/miniconda.sh ; \
+      bash /tmp/miniconda.sh -b -p /usr/local/miniconda ; \
```
tmbgreaves:
```suggestion
      bash /tmp/miniconda.sh -b -p /usr/local/miniconda ; \
      export PATH=/usr/local/miniconda/bin:$PATH
```
We need an extra bit of path setting here so the following conda commands work. It should be safe to leave the ENV command in line 90 as if there's no conda installed it won't have any effect.

mloubout: Yeah these lines are giving me trouble so far no luck having it to run. Like in travis I'm just trying to skip the conda installation to make sure we don't get lucky with pip getting so extras from conda.

mloubout:
>directly calling pip to calling 'python3 -m pip
Yeah it's usually more stable and safe

>bump the base image up to Bionic
I'm fine with that

mloubout:
"Beginning with Matplotlib 3.1, Python 3.6 or above is required."

Gotta have separate requirements now....

tmbgreaves:
Testing with bionic, we hit the problem that ubuntu-toolchain-r only builds gcc9, so we'd be limited to 7 and 9 to test on bionic. Does that change the decision to move?

mloubout:
Other are tested on travis s I'd be fine with it. @ggorman @FabioLuporini what do you think?

Also I kinda want to drop python 3.5 because some (lots?) dependencies don't support it anyway.

tmbgreaves:
OK! I've pushed some changes as suggestions :-)

Dockerfile.pipelines
```diff
@@ -93,10 +76,10 @@ RUN chown -R devito /home/devito
 USER devito
 WORKDIR /home/devito
 
-RUN if [ $installWithPip == "true" ] ; then \ 
-      python setup.py install --user ; \
+RUN if [ $installWithPip == "true" ] ; then \
+      python3 setup.py install --user ; \
```
FabioLuporini: why isn't ``python`` the default ``python3``?

mloubout: default is not the 3.6 which is called python3

mloubout:
SO mainly all green except:
- waiting on inducer to burn new realease of codepy
- One failure with conda and openmp8 for some reason I don't understand

.travis.yml
```diff
@@ -20,8 +20,20 @@ matrix:
             - gcc-4.9
             - g++-4.9
       env: DEVITO_ARCH=gcc-4.9 DEVITO_OPENMP=0 INSTALL_TYPE=pip_setup RUN_EXAMPLES=False MPI_INSTALL=1
-    - os: linux
-      python: "3.6"
+    - python: 3.7
```
FabioLuporini: is this a new build??

.travis.yml
```diff
@@ -60,37 +75,27 @@ matrix:
             - gcc-7
             - g++-7
       env: DEVITO_ARCH=gcc-7 DEVITO_OPENMP=0 DEVITO_BACKEND=ops YC_CXX=g++-7 INSTALL_TYPE=conda RUN_EXAMPLES=False MPI_INSTALL=0
-  allow_failures:
```
FabioLuporini: what was this doing before? I have absolutely no idea

.travis.yml
```diff
@@ -30,8 +42,9 @@ matrix:
             - gcc-5
             - g++-5
       env: DEVITO_ARCH=gcc-5 DEVITO_OPENMP=0 RUN_EXAMPLES=False INSTALL_TYPE=conda MPI_INSTALL=0
-    - os: linux
-      python: "3.6"
+    - python: 3.7
+      dist: xenial
```
FabioLuporini: why fixing xenial?

.travis.yml
```diff
@@ -60,37 +75,27 @@ matrix:
...
-      env: DEVITO_ARCH=clang DEVITO_OPENMP=0
 
 addons:
   apt:
     sources:
       - ubuntu-toolchain-r-test     # For gcc 4.9, 5 and 7
 
 before_install:
-  # Setup anaconda
-  - if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
```
FabioLuporini: here you're just dropping the hack documented in the issues page right? ie pulling an old miniconda version for reasons I now don't remember

.travis.yml
```diff
@@ -60,37 +75,27 @@ matrix:
...
-  - hash -r
-  - conda config --set always_yes yes --set changeps1 no
-  - conda update -q conda
-  # Useful for debugging any issues with conda
-  - conda info -a
   - if [[ "$MPI_INSTALL" == '1' ]]; then
       sudo apt-get install -y -q mpich libmpich-dev;
       export MPICH_CC=$DEVITO_ARCH;
     fi
+  # Setup anaconda
```
FabioLuporini: just wondering why has this been moved after the ``if MPI_INSTALL`` block?

Dockerfile.pipelines
```diff
@@ -1,15 +1,8 @@
-FROM ubuntu:xenial
+FROM ubuntu:bionic
```
FabioLuporini: why this change?

Dockerfile.pipelines
```diff
@@ -1,15 +1,8 @@
...
 
 # This DockerFile is looked after by
 MAINTAINER Tim Greaves <tim.greaves@imperial.ac.uk>
 
-# Add the ubuntu-toolchain-r test ppa
```
FabioLuporini: is this a consequence of not using xenial?

Dockerfile.pipelines
```diff
@@ -25,11 +18,9 @@ RUN apt-get -y dist-upgrade
 # Needed for the conda and devito installs later
 RUN apt-get -y install wget bzip2 git make
 
-# Default gcc version to install
-ARG gccVersion=8
-ENV DEVITO_ARCH=gcc-$gccVersion
-# Set Yask variables in case they are needed
-ENV YC_CXX=g++-$gccVersion
+# Set environment
```
FabioLuporini: why?

Dockerfile.pipelines
```diff
@@ -25,11 +18,9 @@ RUN apt-get -y dist-upgrade
 # Needed for the conda and devito installs later
 RUN apt-get -y install wget bzip2 git make
 
-# Default gcc version to install
-ARG gccVersion=8
-ENV DEVITO_ARCH=gcc-$gccVersion
```
FabioLuporini: I'm concerned by this change. This is not parametric anymore? why should it be the case?

setup.py
```diff
@@ -19,12 +19,21 @@
 
 opt_reqs = []
 opt_links = []
+dict_opt_reqs = {}
```
FabioLuporini: call it ``extras_require``?

.travis.yml
mloubout: 3.7 only supported on xenial dist, so has to be put there.

.travis.yml
mloubout: With was allowing failure on [ython 2.7 which is not tested pr supported so was useless

.travis.yml
mloubout: Yes, now two builds, one for 3.6 and one for 3.7

.travis.yml
mloubout: Yes. A previous version had issues with numpy so had to be skipped. Newest seem to have fixed it so we can get back to latest, and dropped untested 2.7

.travis.yml
mloubout: No special reason, was a tab issue in the yaml was moving stuff around to find it, I can change it back doesn't really matter what order

Dockerfile.pipelines
mloubout: @tmbgreaves commented above about it, necessary for python versions

setup.py
mloubout: Can change name yes.

Dockerfile.pipelines
mloubout: Not sure about these changes, was necessary for bionic.

.travis.yml
FabioLuporini: I don't think a whole new build is a good idea, as this will increase the turnaround time by at least another hour (maybe two). Can we not use 3.6 in one of the existing builds?

.travis.yml
FabioLuporini: OK

.travis.yml
FabioLuporini: it would only make the diff clearer, but otherwise no significant reason (had to compare the lines one by one)

Dockerfile.pipelines
FabioLuporini: I think we then have to step back -- we absolutely need to test multiple gcc versions, in particular the newest ones

mloubout: We still do, it's just that now gcc-8 comes default so no need to add extra install, so it's the exact same setup as before just "easier" to install with defaults in bionic

tmbgreaves: I think I need to regress some changes here, actually -- I was assuming (in error) something special about toolchain-r which wasn't provided by system packages.

Dockerfile.pipelines
tmbgreaves: Yes - toolchain-r drops all except 9 when you move to xenial. But you do get 9 - is this useful to test against?

tmbgreaves: Dropped this change - see thoughts below on gcc versioning.

tmbgreaves:
Thoughts on gcc:

Versions available from bionic are 5, 6, 7, and 8, plus 9 from toolchain-r. I've provisionally added a 9 builder -- change as required. 4 is no longer available on bionic so I've moved the pip builder to 8 -- again change as required.

tmbgreaves:
Turns out the gcc installs from bionic aren't as straightforward as expected -- hacking on it now.

tmbgreaves:
It appears that adding in ubuntu-toolchain-r breaks installs of the system compiler packages, so gcc 9 from there might be better to leave for now. If 9 would be useful, we could add in the toolchain-r ppa bit as a conditional on gcc 9 being requested.

tmbgreaves:
OK -- I think latest commit has fixed things and this now has working 5, 7, and 8 from system packages.

.travis.yml
mloubout: Well no, this test python 3.6 and 3.7 with pip then there is the conda builds. Can try to merge one of the conda with one of the two pip but will reduce conda testing. I'm not sure about turnaround time as they supposed to run toghether so don't see why it would be slower

mloubout: Removed one, added one to pipelines instead as it is faster

.travis.yml
```diff
@@ -134,6 +128,8 @@ script:
   - if [[ $RUN_EXAMPLES == 'True' ]]; then python examples/misc/linalg.py mat-vec mat-mat-sum transpose-mat-vec; fi
   - if [[ $RUN_EXAMPLES == 'True' ]]; then python examples/seismic/tti/tti_example.py -a; fi
   - if [[ $RUN_EXAMPLES == 'True' ]]; then python examples/seismic/tti/tti_example.py -a --noazimuth; fi
+  - if [[ $RUN_EXAMPLES == 'True' ]]; then python examples/cfd/example_diffusion.py; fi
```
FabioLuporini: why do we have to run it once with ``python`` and once with ``pytest``?

azure-pipelines.yml
```diff
@@ -12,48 +12,62 @@ jobs:
     vmImage: 'Ubuntu 16.04'
   strategy:
     matrix:
-      pip-gcc4:
-        gccVersion: '4.9'
+      pip-gcc8-py36:
```
FabioLuporini: this is called gcc8, but below it says gccVersions=6

I'm not sure I understand why we want to add a new build

.travis.yml
```diff
@@ -10,18 +10,19 @@ env:
...
         apt:
           sources:
             - ubuntu-toolchain-r-test     # For gcc 4.9, 5 and 7
           packages:
-            - gcc-4.9
-            - g++-4.9
-      env: DEVITO_ARCH=gcc-4.9 DEVITO_OPENMP=0 INSTALL_TYPE=pip_setup RUN_EXAMPLES=False MPI_INSTALL=1
-    - os: linux
-      python: "3.6"
+            - gcc-7
```
FabioLuporini: why do you wanna drop 4.9? it's still (unfortunately) one of the most used gcc versions, so it should be tested right?

what we may want is testing gcc9 ...

tmbgreaves: 4.9 - Lack of a build for bionic, there only appears to be a stub set of packages on the system set, and none on ubuntu-toolchain-r, but I haven't looked ultra-carefully.

9 - I think this should work with some bash conditionals wrapped round the gcc install to only turn on the ubuntu-toolchain-r repo when installing 9.

I'm a bit tied up with things this week but will do my best to fit in having a look at it if nobody beats me to it.

azure-pipelines.yml
mloubout: We want to test both python 3.7 and 3.6 so that's two buildm then the conda ones with the examples and stuff

.travis.yml
mloubout: There is a test in it that used to be run in a weird way I just moved it here because it needs example install for matplotlib

azure-pipelines.yml
mloubout: Can have up to 10 at same time on pipeline so it' fine I think

tmbgreaves:
A thought -- do we even need to touch system python here or could we do everything using conda? Ie, if we don't care about system python, it might be feasible to stick with xenial for the purposes of getting this PR through, and deal with the xenial->bionic transition separately. I think moving off xenial is a good thing to do at some point soon given we're now a year past the following LTS (bionic) release.

mloubout:
Well system python is used to test the pip way. The only issue left currently is blosc failing to compile with gcc-9 and everything else should be fine. I added a line to force-link the installed gcc version and python so this part is fixed.

azure-pipelines.yml
```diff
@@ -12,48 +12,54 @@ jobs:
     vmImage: 'Ubuntu 16.04'
   strategy:
     matrix:
-      pip-gcc4:
-        gccVersion: '4.9'
+      pip-gcc6-py36:
+        gccVersion: '6'
         DEVITO_OPENMP: '0'
-        MPI_INSTALL: '1'
+        MPI_INSTALL: '0'
```
FabioLuporini: this should be 1?

PR #868, opened 2019-07-10
MPI fixes 

FabioLuporini:
fixes for:

* comp/comm overlap in presence of subdomains/subdimensions
* coupled equations case (some halo exchanges were being dropped on the floor due to using dicts)

to achieve this, this PR re-engineers some of the IR data structures representing IterationSpaces. In particular, Intervals now accepts generic sympy expressions in their offsets


devito/dle/blocking_utils.py
```diff
@@ -1,18 +1,114 @@
...
+                body.append(List(body=call))
+
+            mapper[root] = List(body=body)
+
+            # Next blockable nest, use different (unique) variable/function names
+            self.nblocked += 1
+
+        iet = Transformer(mapper).visit(iet)
+
+        # Force-unfold if some folded Iterations haven't been blocked in the end
```
georgebisbas: Could you elaborate more on this docstring?

devito/ir/equations/equation.py
```diff
@@ -122,21 +122,34 @@ def __new__(cls, *args, **kwargs):
...
-
-        # The data space is relative to the computational domain. Note that we
-        # are deliberately dropping the intervals ordering (by turning `intervals`
-        # into a list), as this is irrelevant (even more: dangerous) for data spaces
-        intervals = [i if i.dim in oobs else i.zero() for i in intervals]
-        intervals += [Interval(i, 0, 0) for i in ordering
-                      if i not in ispace.dimensions + conditionals]
-        parts = {k: IntervalGroup(build_intervals(v)) for k, v in mapper.items() if k}
-        dspace = DataSpace(intervals, parts)
+        iintervals = []
```
georgebisbas: I am not sure if I like iintervals varname... just saying...

devito/ir/stree/algorithms.py
```diff
@@ -51,7 +51,7 @@ def st_schedule(clusters):
 
         # The reused sub-trees might acquire some new sub-iterators
         for i in pointers[:index]:
-            mapper[i].ispace = IterationSpace.merge(mapper[i].ispace,
+            mapper[i].ispace = IterationSpace.union(mapper[i].ispace,
                                                     c.ispace.project([i.dim]))
         # Later sub-trees, instead, will not be used anymore
```
georgebisbas: What about later sub-trees? Do you mean following? I do not quite get it?
Does it mean next trees or do you refer to a later part of code?

devito/ir/support/space.py
```diff
@@ -121,25 +112,24 @@ class Interval(AbstractInterval):
...
-        dim.size + abs(upper - lower)
+        (dim.extreme_max - dim.extreme_min + 1) + (upper - lower)
     """
 
     is_Defined = True
 
     def __init__(self, dim, lower, upper, stamp=0):
-        assert is_integer(lower)
-        assert is_integer(upper)
+        assert is_integer(lower) or isinstance(lower, Expr)
```
georgebisbas: :-)

devito/ir/support/basic.py
```diff
@@ -621,6 +319,63 @@ def distance(self, other, findex=None):
...
+        except AttributeError:
+            if d.is_NonlinearDerived:
+                # We should only end up here with subsampled Dimensions
+                m, M = self.intervals[d.root].offsets
+            else:
+                # A constant (integer) is actually used to index into `findex`
+                assert is_integer(self[findex])
+                m, M = 0, 0
+
+        # If `m + (self[d] - d) < self.function._size_halo[d].left`, then `self`
```
georgebisbas: Maybe ( m+() ) ? Should you also use `findex` instead of `d`?

devito/ir/support/basic.py
```diff
@@ -621,6 +319,63 @@ def distance(self, other, findex=None):
...
+        except TypeError:
+            # Two reasons we might end up here:
+            # * `d` is a constant integer
+            # * `m` is a symbol (e.g., a SubDimension-induced offset)
+            #   TODO: we could exploit the properties attached to `m` (if any), such
+            #         as `nonnegative` etc, to do something smarter than just
+            #         assuming, conservatively, `touch_halo_left = True`
+            touch_halo_left = True
+
+        # If `M + (self[d] - d) > self.function._size_halo[d].left`, then
```
georgebisbas: same

devito/ir/support/vector.py
```diff
@@ -0,0 +1,355 @@
...
+    functions of the same variables, with identical coefficient.
+    If the Vector is instantiated passing the keyword argument ``smart = True``,
+    some manipulation will be attempted to infer the truth value of a non-trivial
+    symbolic relation. This increases the cost of the comparison, while potentially
+    being ineffective, so use it judiciously. By default, ``smart = False``.
+
+    Raises
+    ------
+    TypeError
+        If two Vectors cannot be compared, e.g. due to uncomparable symbolic entries.
```
georgebisbas: incomparable?

devito/ir/support/vector.py
```diff
@@ -0,0 +1,355 @@
...
+    1) Vector-scalar comparison
+    If a comparison between a vector and a non-vector is attempted, then the
+    non-vector is promoted to a vector; if this is not possible, an exception
+    is raised. This is handy because it turns a vector-scalar comparison into
+    a vector-vector comparison with the scalar broadcasted to all vector entries.
+    For example: ::
+
+        (3, 4, 5) > 4 => (3, 4, 5) > (4, 4, 4) => False
+
+    2) Comparing Vector entries when these are SymPy expression
```
georgebisbas: expressions?

devito/ir/support/vector.py
```diff
@@ -0,0 +1,355 @@
...
+# Utility functions
+
+def vmin(*vectors):
+    """
+    Retrieve the minimum out of an iterable of Vectors.
+
+    Raises
+    ------
+    TypeError
+        If there are two non-comparable Vectors.
```
georgebisbas: incomparable or in any case to be the same...

devito/mpi/routines.py
```diff
@@ -765,9 +745,15 @@ def _make_remainder(self, hs, key, callcompute, region):
 
         dim = Dimension(name='i')
         regioni = IndexedPointer(region, dim)
```
georgebisbas: region_i?

devito/dle/blocking_utils.py
```diff
@@ -1,18 +1,114 @@
...
+
+class Blocker(object):
+
+    def __init__(self, blockinner, blockalways):
+        self.blockinner = bool(blockinner)
+        self.blockalways = bool(blockalways)
+
+        self.nblocked = 0
+
+    def make_blocking(self, iet):
```
FabioLuporini: note for reviewers: 99% of this has just been lifted from rewriters into here -- nothing really new aside from ``self.nblocked``

devito/dle/rewriters.py
```diff
@@ -263,84 +267,7 @@ def _loop_blocking(self, iet):
         """
         Apply loop blocking to PARALLEL Iteration trees.
         """
-        blockinner = bool(self.params.get('blockinner'))
```
FabioLuporini: note for reviewers: this was basically moved into blocking_utils.py

devito/ir/support/basic.py
```diff
@@ -1,260 +1,14 @@
...
 from devito.symbolics import retrieve_terminals, q_monoaffine
 from devito.tools import (EnrichedTuple, Tag, as_tuple, is_integer,
                           filter_sorted, flatten, memoized_meth)
 from devito.types import Dimension
 
-__all__ = ['Vector', 'LabeledVector', 'IterationInstance', 'Access',
-           'TimedAccess', 'Scope']
-
-
-class Vector(tuple):
```
FabioLuporini: note for reviewers: this entire deleted block was actually simply moved to vector.py

devito/ir/support/space.py
FabioLuporini: yep :) as promised. Try to understand what I did to make it work "the right way". It's not trivial

devito/ir/support/space.py
```diff
@@ -154,32 +144,27 @@ def relaxed(self):
...
-    def limits(self):
+    def offsets(self):
         return (self.lower, self.upper)
 
     def intersection(self, o):
-        if self.overlap(o):
-            return Interval(self.dim, max(self.lower, o.lower), min(self.upper, o.upper),
-                            self.stamp)
+        if self.is_compatible(o):
+            svl, svu = Vector(self.lower, smart=True), Vector(self.upper, smart=True)
```
FabioLuporini: @georgebisbas : note I'm using Vector here (understanding why could be a nice exercise). You may start from the tests in test_ir.py

devito/ir/stree/algorithms.py
FabioLuporini: I think I actually meant "nested"

devito/ir/support/basic.py
FabioLuporini: No I think it's clear this way. The inner parentheses is just to emphasize the relationship between the access function (``self[d]``, e.g. "x + 3") and the dimension (e.g., ``x``)

tests/test_mpi.py
```diff
@@ -1230,6 +1230,56 @@ def test_nontrivial_operator(self):
...
+        glb_pos_map = v.grid.distributor.glb_pos_map
+        if LEFT in glb_pos_map[x] and LEFT in glb_pos_map[y]:
+            assert np.all(v.data_ro_domain[1] == [[0, 4], [0, 4]])
+        elif LEFT in glb_pos_map[x] and RIGHT in glb_pos_map[y]:
+            assert np.all(v.data_ro_domain[1] == [[4, 0], [4, 0]])
+        elif RIGHT in glb_pos_map[x] and LEFT in glb_pos_map[y]:
+            assert np.all(v.data_ro_domain[1] == [[0, 4], [0, 4]])
+        elif RIGHT in glb_pos_map[x] and RIGHT in glb_pos_map[y]:
+            assert np.all(v.data_ro_domain[1] == [[4, 0], [4, 0]])
+
```
btrb: just to make sure it'd double checked. no typos on the 0 / 4 orders on the tests above right?

devito/ir/support/vector.py
FabioLuporini: yes!

tests/test_mpi.py
FabioLuporini: should be alright

FabioLuporini: as confirmed by the last assert of the test

devito/core/autotuning.py
```diff
@@ -60,7 +60,7 @@ def autotune(operator, args, level, mode):
     if mode in ['preemptive', 'destructive']:
         for p in operator.parameters:
             if isinstance(p, MPINeighborhood):
-                at_args.update(MPINeighborhood(p.fields)._arg_values())
+                at_args.update(MPINeighborhood(p.neighborhood)._arg_values())
```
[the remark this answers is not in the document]
FabioLuporini: I agree, but this is not related to this PR, so I'll leave it for another today (also since this PR is quite urgent)

devito/dle/blocking_utils.py
```diff
@@ -1,18 +1,114 @@
...
+            iterations = filter_iterations(tree, lambda i: i.is_Parallel)
+            if not self.blockinner:
+                iterations = iterations[:-1]
+            if len(iterations) <= 1:
+                continue
+            root = iterations[0]
+            if not self.blockalways:
+                # Heuristically bypass loop blocking if we think `tree`
+                # won't be computationally expensive. This will help with code
+                # size/redability, JIT time, and auto-tuning time
```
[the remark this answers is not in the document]
FabioLuporini: fixed

devito/dle/blocking_utils.py
FabioLuporini: I tried, but I can't find an easy way. Will do another day

devito/ir/equations/equation.py
FabioLuporini: I can't afford longer variable names otherwise I have to break to many lines and it gets horrible

But you made a fair point: I added comments clarifying what they represent

devito/ir/stree/algorithms.py
FabioLuporini: fixed

devito/ir/support/vector.py
FabioLuporini: fixed, thanks

devito/ir/support/vector.py
FabioLuporini: fixed

devito/ir/support/vector.py
```diff
@@ -0,0 +1,355 @@
...
+
+
+def vmax(*vectors):
+    """
+    Retrieve the maximum out of an iterable of Vectors.
+
+    Raises
+    ------
+    TypeError
+        If there are two non-comparable Vectors.
```
FabioLuporini: fixed

devito/mpi/halo_scheme.py
```diff
@@ -108,88 +147,204 @@ def fmapper(self):
     @cached_property
     def omapper(self):
```
FabioLuporini: @rhodrin you may wanna read this ``omapper`` docstring

devito/ir/stree/tree.py
```diff
@@ -54,7 +54,11 @@ def dim(self):
 
     @property
     def limits(self):
-        return self.interval.limits
+        return (self.dim.extreme_min, self.dim.extreme_max, 1)
```
rhodrin: Where is the `1` coming from here?

FabioLuporini: it's the step size. Dimensions (and SubDimensions) imply a step size of 1 (ie, ++ or --)

some subclasses (e.g., ModuloDimension) behave differently

devito/ir/support/basic.py
```diff
@@ -270,14 +24,39 @@ class IterationInstance(LabeledVector):
...
-          is non-affine, then it may not be possible to detect its aindex;
-          in such a case, None is used as placeholder.
+          There is one aindex for each non-constant affine index function. If
+          the index function is non-affine, then it may not be possible to detect
+          its aindex; in such a case, None is used as placeholder.
         * ``findices``: the Dimensions describing what *data* space point
           are accessed.
+
+    An IterationInstance may be regular or irregular. It is regular if and only
+    if *all* index functions are affine in their respective findex.  The
```
rhodrin: `findex`, type?

devito/ir/support/basic.py
```diff
@@ -270,14 +24,39 @@ class IterationInstance(LabeledVector):
...
+        x, y, z : findices
+        w : a generic Dimension
+
+           | x+1 |           |  x  |          |  x  |          | w |          | x+y |
+    obj1 = | y+2 | ,  obj2 = |  4  | , obj3 = |  x  | , obj4 = | y | , obj5 = |  y  |
+           | z-3 |           | z+1 |          |  y  |          | z |          |  z  |
+
+    We have that:
+
+        * obj1 and obj2 are regular;
```
rhodrin: So the 'regularity' is just determined topologically? i.e. if my given `findices` are `z, x, y` then
```     
         | a_1*z+b_1 |
obj =    | a_2*x+b_2 | 
         | a_3*y+b_3 |
```
would be regular?

devito/ir/support/space.py
```diff
@@ -121,25 +112,24 @@ class Interval(AbstractInterval):
 
     Create an Interval of size:
 
-        dim.size + abs(upper - lower)
+        (dim.extreme_max - dim.extreme_min + 1) + (upper - lower)
```
rhodrin: Just wondering why `size` is no longer used since the new codes looks like what `size` should be doing? Also was the old `abs` superfluous since we always have `upper > lower` or did something else change?

devito/ir/support/vector.py
```diff
@@ -0,0 +1,355 @@
...
+from devito.tools import as_tuple, is_integer, memoized_meth
+from devito.types import Dimension
+
+__all__ = ['Vector', 'LabeledVector', 'vmin', 'vmax']
+
+
+class Vector(tuple):
+
+    """
+    A representation of an object in Z^n.
```
rhodrin: Why `Z`? (I'm assuming it stands for the integers).

devito/mpi/halo_scheme.py
```diff
@@ -39,54 +40,53 @@ class HaloScheme(object):
...
-    The dict ``loc_indices`` tells how to access/insert the halo along the
-    keyed Function's non-halo indices. For example, consider the Function
-    ``u(t, x, y)``. Assume ``x`` and ``y`` require a halo exchange. The
-    question is: once the halo exchange is performed, at what offset in ``t``
-    should it be placed? should it be at ``u(0, ...)`` or ``u(1, ...)`` or even
-    ``u(t-1, ...)``? ``loc_indices`` has as many entries as non-halo
-    dimensions, and each entry provides symbolic information about how to
-    access the corresponding non-halo dimension. Thus, in this example
-    ``loc_indices`` could be, for instance, ``{t: 0}`` or ``{t: t-1}``.
+    ``halos`` is a list of 2-tuples ``(Dimension, DataSide)``. This is metadata
```
rhodrin: That looks like a tuple of 2-tuples?

devito/mpi/halo_scheme.py
```diff
@@ -39,54 +40,53 @@ class HaloScheme(object):
...
                                                   frozenset(halos))
-
-        # A HaloScheme is immutable, so let's make it hashable
         self._mapper = frozendict(self._mapper)
 
+        # Track the IterationSpace offsets induced by SubDomains/SubDimensions.
+        # These should be honored in the derivation of the `omapper`
+        self._honored = {}
+        for i in ispace.intervals:
+            if not i.dim._defines & set(self.dimensions):
```
rhodrin: How come we need `set` here?

devito/types/dimension.py
```diff
@@ -520,19 +532,27 @@ def middle(cls, name, parent, thickness_left, thickness_right, local=False):
                    thickness=((lst, thickness_left), (rst, thickness_right)),
                    local=local)
 
-    @property
+    @cached_property
```
rhodrin: Caching these is not going to cause problems when we're updating them via implicit equations right?

tests/test_ir.py
```diff
@@ -37,16 +56,52 @@ def ii_literal(self, fa, fc):
...
-        fwd_directions = {x: Forward, y: Forward}
-        mixed_directions = {x: Backward, y: Forward}
-        tcxy_w0 = TimedAccess(fc[x, y], 'W', 0, fwd_directions)
-        tcxy_r0 = TimedAccess(fc[x, y], 'R', 0, fwd_directions)
-        tcx1y1_r1 = TimedAccess(fc[x + 1, y + 1], 'R', 1, fwd_directions)
-        tcx1y_r1 = TimedAccess(fc[x + 1, y], 'R', 1, fwd_directions)
-        rev_tcxy_w0 = TimedAccess(fc[x, y], 'W', 0, mixed_directions)
-        rev_tcx1y1_r1 = TimedAccess(fc[x + 1, y + 1], 'R', 1, mixed_directions)
+        intervals = [Interval(x, 0, 0), Interval(y, 0, 0)]
+        fwd_ispace = IterationSpace(intervals, directions={x: Forward, y: Forward})
```
rhodrin: `bwd_ispace` test?

devito/ir/support/space.py
FabioLuporini: so the ``abs`` was wrong

``dim.extreme_max - dim.extreme_min + 1`` is more correct than ``size`` (and also reflects the implementation). If we say "size", it would be like including ``upper`` and ``lower``, which instead are treated separately (they are the ``offsets`` of the Interval)

devito/ir/support/vector.py
FabioLuporini: Correct, integers

It's Z (and not R, for example) because certain operations that this class provides are guaranteed to work (or are well-defined) only over integers 

BTW, note that this class has been in master for ages, I only moved it here from a different file

devito/mpi/halo_scheme.py
FabioLuporini: yes...unless I see other major (coding-related) issues, I'll fix docstrings in the next (upcoming) PR

devito/mpi/halo_scheme.py
FabioLuporini: well ``self.dimensions `` returns tuple, and intersection (``&``) is only defined over sets

devito/types/dimension.py
FabioLuporini: it must not. These objects are immutable

tests/test_ir.py
FabioLuporini: ah, it's actually included in mixed_ispace ...there's both a Backward and a Forward

devito/ir/support/vector.py
rhodrin: Oki.

> BTW, note that this class has been in master for ages, I only moved it here from a different file

Yeah, I realize that, I'd just not been through it before.

devito/ir/support/basic.py
FabioLuporini: yes

devito/mpi/routines.py
FabioLuporini: yes absolutely. Thought I had fixed it. Missed it actually. Will fix shortly

devito/ir/support/vector.py
```diff
@@ -1,6 +1,6 @@
 from collections import OrderedDict
 
-from sympy import Basic
+from sympy import Basic, true
```
btrb: i know it has been merged already. but just for learning: what's the importance of this? wouldn't it be considered an overkill?

FabioLuporini: sympy booleans are "more powerful" than python booleans when it comes to comparing sympy expressions. Look for where it's used and try replace it with ``True`` , you'll see things start breaking (why?)

btrb: i see. 
i would guess that's because something like 
'2 + 2' == '4'
could be evaluated as 'sympy.true'
but would not be considered 'True'?

PR #870, opened 2019-07-15
Tolerance

mloubout:
truncate rand value to avoid numerical issues

there is one other `rand` in `test_dependency_bugs.py` not sure if should be changed too


FabioLuporini:
merged.

PR #872, opened 2019-07-17
Auto ops calls

btrb:
This PR is for enabling OPS calls to be inserted automatically within the host code.


btrb:
sorry, accidentally mixed the remotes

PR #876, opened 2019-07-21
Using the new OPS API for expression translation

vincepandolfo:
Defines `OpsAccessible` and `OpsAccess` types representing new-style accesses

### DIFF
--- a/devito/ops/transformer.py
+++ b/devito/ops/transformer.py
@@ -1,4 +1,4 @@
-def make_ops_ast(expr, nfops):
+def make_ops_ast(expr, nfops, is_write=False):
     """
     Transform a devito expression into an OPS expression.
     Only the interested nodes are rebuilt.
@@ -19,6 +19,12 @@ def make_ops_ast(expr, nfops):
     if expr.is_Symbol or expr.is_Number:
         return expr
     elif expr.is_Indexed:
-        return nfops.new_ops_arg(expr)
+        return nfops.new_ops_arg(expr, is_write)
+    elif expr.is_Equality:
+        res = expr.func(
+            make_ops_ast(expr.lhs, nfops, True),
+            make_ops_ast(expr.rhs, nfops)
+        )
+        return res
     else:
         return expr.func(*[make_ops_ast(i, nfops) for i in expr.args])
--- a/devito/ops/types.py
+++ b/devito/ops/types.py
@@ -1,0 +1,22 @@
+import devito.types.basic as basic
+
+
+class OpsAccessible(basic.AbstractObject):
+
+    def __init__(self, symbol, access_indexes, read_only):
+        self.symbol = symbol
+        self.access_indexes = [str(idx) for idx in access_indexes]
+        self.read_only = read_only
+        super().__init__(self._C_name, symbol.dtype)
+
+    @property
+    def _C_name(self):
+        return '%s(%s)' % (self.symbol.name, ",".join(self.access_indexes))
+
+    @property
+    def _C_typename(self):
+        return '%sACC<%s>' % ('const ' if self.read_only else '', self.symbol._C_typename)
+
+    @property
+    def _C_typedata(self):
+        return 'ACC<%s>' % self.symbol._C_typename
--- a/devito/ops/types.py
+++ b/devito/ops/types.py
@@ -1,22 +1,90 @@
+import sympy
+
 import devito.types.basic as basic
 
+from devito.tools import dtype_to_cstr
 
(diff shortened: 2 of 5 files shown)


devito/ops/transformer.py
```diff
@@ -1,4 +1,4 @@
-def make_ops_ast(expr, nfops):
+def make_ops_ast(expr, nfops, is_write=False):
     """
     Transform a devito expression into an OPS expression.
```
btrb: Devito, Expression?

devito/ops/types.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+class OpsAccessible(basic.Symbol):
+    """
+    OPS accessible symbol
+
+    Parameters
+    ----------
+    name : str
+        Name of the symbol.
```
btrb: Symbol?

devito/ops/types.py
```diff
@@ -0,0 +1,90 @@
+import sympy
+
+import devito.types.basic as basic
+
+from devito.tools import dtype_to_cstr
+
+
+class OpsAccessible(basic.Symbol):
+    """
+    OPS accessible symbol
```
btrb: Symbol?

btrb: nvrmd, its not a Devito object right

devito/ops/utils.py
```diff
@@ -5,7 +5,6 @@
 namespace = OrderedDict()
 
 # OPS Kernel strings.
-namespace['ops_acc'] = 'OPS_ACC'
```
btrb: can we not drop this yet?

tests/test_ops.py
```diff
@@ -17,31 +17,30 @@
...
-        ('Eq(u.forward,u+1)', 'Eq(ut1[OPS_ACC0(0)], ut0[OPS_ACC1(0)] + 1)'),
+         'Eq(wt0(0, 0, 0), ut0(0) + vt0(0, 0))'),
+        ('Eq(u.forward,u+1)', 'Eq(ut1(0), ut0(0) + 1)'),
         ('Eq(u, u.laplace)',
-         'Eq(ut0[OPS_ACC0(0)], ut0[OPS_ACC0(-1)]/h_x**2 - '
-            '2.0*ut0[OPS_ACC0(0)]/h_x**2 + ut0[OPS_ACC0(1)]/h_x**2)'),
+         'Eq(ut0(0), ut0(-1)/h_x**2 - '
+            '2.0*ut0(0)/h_x**2 + ut0(1)/h_x**2)'),
     ])
     def test_ast_convertion(self, equation, expected):
```
btrb: maybe it will fork here if we will cover both old and new OPS syntax
@FabioLuporini should we cover both styles or aim to drop the old one? 
i'm asking this because i believe it could drastically impact our current efforts

btrb: @vmickus fyi

devito/ops/utils.py
vincepandolfo: why?

btrb: if we'll cover the old syntax as well, we will have to add it again 
anyways, i understand it's currently not being used. 
so maybe it's ok to add it again when we push new PR

tests/test_ops.py
vincepandolfo: I'm almost done with the next PR (if I'm quick enough I might be able to add it on top of this) were the expressions are put in a callable and placed in the header and the parameters extracted. This should make using this very easy

devito/ops/transformer.py
FabioLuporini: I think 'expression' is ok because it's loosely referring to a fairly generic expression

devito/ops/types.py
```diff
@@ -0,0 +1,90 @@
...
+
+class OpsAccess(basic.Basic):
+    """
+    OPS access
+
+    Parameters
+    ----------
+    base : OpsAccessible
+        Symbol to access
+    indexes: [int]
```
FabioLuporini: nitpicking: we use "indices" throughout devito

devito/ops/utils.py
FabioLuporini: why would you want to support the old syntax if the new one is just better? 
having to support two things increases the maintenance burden. Would it be that difficult to update whatever you have to the new syntax?

tests/test_ops.py
FabioLuporini: I'm happy to drop the old one, as long as this doesn't cause excessive pain to you guys @btrb .

vincepandolfo:
closing in favours of #877

PR #877, opened 2019-07-22
OPS translation using new API and OPS Kernel generation

vincepandolfo:
supersedes #876.

Uses the new API to translate expressions and generates callables for OPS kernels

@FabioLuporini

### DIFF
--- a/devito/ops/node_factory.py
+++ b/devito/ops/node_factory.py
@@ -34,7 +34,7 @@ class OPSNodeFactory(object):
 
         # Build the OPS arg identifier
         time_index = split_affine(indexed.indices[TimeFunction._time_position])
-        ops_arg_id = ('%s%s%s' % (indexed.name, time_index.var, time_index.shift)
+        ops_arg_id = ('%s%s' % (indexed.name, time_index.var)
                       if indexed.function.is_TimeFunction else indexed.name)
 
         if ops_arg_id not in self.ops_args:
--- a/devito/ops/node_factory.py
+++ b/devito/ops/node_factory.py
@@ -1,9 +1,9 @@
 from collections import OrderedDict
 
-from devito import TimeFunction
+from devito import Constant, TimeFunction
 from devito.types.dimension import SpaceDimension
 from devito.symbolics import split_affine
-from devito.ops.types import OpsAccessible
+from devito.ops.types import OpsAccess, OpsAccessible
 
 
 class OPSNodeFactory(object):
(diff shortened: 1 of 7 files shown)


devito/symbolics/printer.py
```diff
@@ -127,6 +127,9 @@ def _print_TrigonometricFunction(self, expr):
             func_name += 'f'
         return func_name + '(' + self._print(*expr.args) + ')'
 
+    def _print_Basic(self, expr):
```
FabioLuporini: this is needed for...?

vincepandolfo: `_print_Basic` defaults to `_print_not_supported` that appends `// Not supported in C` everywhere

vincepandolfo:
@vmickus @btrb any comments?

devito/ops/node_factory.py
```diff
@@ -35,28 +34,32 @@ def new_ops_arg(self, indexed):
...
-                                                ','.join(str(split_affine(i).shift)
-                                                         for i in space_indices)))
+    def new_ops_gbl(self, c):
+        if c in self.ops_args:
+            return self.ops_args[c]
 
-        # Create Indexed object representing the OPS arg access
-        new_indexed = Indexed(ops_arg.indexed, access_macro)
+        new_c = Constant(name='*%s' % c.name, dtype=c.dtype)
+        self.ops_args[c] = new_c
```
btrb: fuse lines?

devito/ops/operator.py
```diff
@@ -14,17 +15,19 @@ class OperatorOPS(Operator):
...
-
         warning("The OPS backend is still work-in-progress")
 
         ops_init = Call(namespace['ops_init'], [0, 0, 2])
         ops_exit = Call(namespace['ops_exit'])
 
         dims = []
         for n, (section, trees) in enumerate(find_affine_trees(iet).items()):
+            ops_kernel = opsit(trees, n)
+            self._ops_kernels.append(ops_kernel)
```
btrb: fuse?

vincepandolfo: This will be changed in the future as `opsit` will return more than the kernel, so joining these lines is not needed

devito/ops/node_factory.py
vincepandolfo: Doing it this way saves a lookup in the dictionary

btrb:
i'm still trying to understand how `return OpsAccess(symbol_to_access, space_indices)` works. 
it looks very beautiful!

devito/ops/types.py
```diff
@@ -0,0 +1,106 @@
+import sympy
+
+import devito.types.basic as basic
+
+from devito.tools import dtype_to_cstr
+
+
+class OpsAccessible(basic.Symbol):
```
vmickus: @btrb here
maybe this name leads to confusion.
what about something like `OpsKernelArg` or `OpsKernelParam`

(also, i´d vote for capital OPS like `OPSLalala` instead of `OpsLalala`)

devito/ops/types.py
```diff
@@ -0,0 +1,106 @@
...
+
+class OpsAccess(basic.Basic, sympy.Basic):
+    """
+    OPS access
+
+    Parameters
+    ----------
+    base : OpsAccessible
+        Symbol to access
+    indices: [int]
```
btrb: list of int?

devito/ops/types.py
vincepandolfo: I'm not particularly fond of either suggestion. Maybe `OpsAccessBase` in line with `Indexed`

devito/ops/types.py
vincepandolfo: Yes

FabioLuporini:
LGTM, hence merging.

@vincepandolfo @vmickus this means that the discussion on class names is postponed to the next PR or slack

PR #878, opened 2019-07-22
Including ops_partition("") call

maelso:
This commit is including the ops_partition("") call into code generated by Devito using OPS backend.
The class Literal was created in the codebase.

### DIFF
--- a/devito/symbolics/extended_sympy.py
+++ b/devito/symbolics/extended_sympy.py
@@ -12,7 +12,7 @@ from devito.tools import Pickable, as_tuple
 
 __all__ = ['FrozenExpr', 'Eq', 'CondEq', 'CondNe', 'Mul', 'Add', 'Pow', 'IntDiv',
            'FunctionFromPointer', 'FieldFromPointer', 'FieldFromComposite',
-           'ListInitializer', 'Byref', 'IndexedPointer', 'Macro', 'taylor_sin',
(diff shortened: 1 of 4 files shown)


vincepandolfo:
LGTM

devito/symbolics/extended_sympy.py
```diff
@@ -357,6 +357,14 @@ class Macro(sympy.Symbol):
     pass
 
 
+class Literal(sympy.Symbol):
```
mloubout: Why is this needed? Can't you use the `Symbol` class in devito.types?

vincepandolfo: Because of the check in `utils.py` I believe.

mloubout: I get that the check needs an extra, but I don't see why it needs to create a new (mostly copy) of an already existing class. just use `devito.symbol` and same in the check in utils.

vincepandolfo: You can't use Symbol in that check though, it would break using symbols as arguments

maelso: Symbol works, but I could break other locations. Here I am creating a type for this specific case.

maelso:
I can't understand why the travis is breaking.

FabioLuporini:
yeah, that's really odd... I'll restart it, no idea what's going on. pipelines seems fine at least

FabioLuporini:
Merged.

PR #879, opened 2019-07-23
bench: fix multiple logging with MPI

georgebisbas:
This PR aims to drive MPI bench-logging output to be generated only from one MPI process.

### DIFF
--- a/benchmarks/user/benchmark.py
+++ b/benchmarks/user/benchmark.py
@@ -336,8 +336,8 @@ def get_ob_bench(problem, resultsdir, parameters):
             devito_params['at'] = params['autotune']
 
             if configuration['openmp']:
-                devito_params['nt'] = os.environ.get(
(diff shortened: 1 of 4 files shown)


FabioLuporini:
uncontroversial

FabioLuporini:
is this ready to land? or shall we wait for a ``test_benchmarking`` (not your fault this is missing -- definitely something we should have done ages ago)

georgebisbas:
> is this ready to land? or shall we wait for a `test_benchmarking` (not your fault this is missing -- definitely something we should have done ages ago)

I suggest that as long as tests are missing, we use it temporarily for our purposes from branch and when tests are ready we can land

FabioLuporini:
Im tempted to say that if you tried it, and it seems to do the right thing, I'd fix the remaining flake8 issue and merge it as is. We need to get rid of as many branches as possible at this stage

FabioLuporini:
Merged.

PR #880, opened 2019-07-23
Allowing calls to be nested

vincepandolfo:
eg: `foo(bar(baz()))`

This will be used when nesting `ops_arg` calls within `ops_par_loop`

### DIFF
--- a/devito/ir/iet/nodes.py
+++ b/devito/ir/iet/nodes.py
@@ -201,8 +201,9 @@ class Call(Simple, Node):
 
     is_Call = True
 
-    def __init__(self, name, arguments=None):
+    def __init__(self, name, arguments=None, statement=None):
         self.name = name
+        self.statement = statement if statement is not None else True
         self.arguments = as_tuple(arguments)
 
     def __repr__(self):
@@ -227,6 +228,10 @@ class Call(Simple, Node):
     @property
     def defines(self):
         return ()
+
+    @property
+    def children(self):
+        return tuple(i for i in self.arguments if isinstance(i, Call))
 
 
 class Expression(Simple, Node):
--- a/devito/ir/iet/nodes.py
+++ b/devito/ir/iet/nodes.py
@@ -203,7 +203,7 @@ class Call(Simple, Node):
 
     def __init__(self, name, arguments=None, statement=None):
         self.name = name
-        self.statement = statement if statement is not None else True
+        self.statement = statement or True
         self.arguments = as_tuple(arguments)
 
     def __repr__(self):
--- a/devito/ir/iet/nodes.py
+++ b/devito/ir/iet/nodes.py
@@ -201,8 +201,6 @@ class Call(Simple, Node):
 
     is_Call = True
 
-    _traversable = ['arguments']
-
     def __init__(self, name, arguments=None):
         self.name = name
         self.arguments = as_tuple(arguments)
@@ -213,6 +211,10 @@ class Call(Simple, Node):
     @property
     def functions(self):
         return tuple(i for i in self.arguments if isinstance(i, AbstractFunction))
+
+    @property
+    def children(self):
+        return tuple(i for i in self.arguments if isinstance(i, Call))
 
     @cached_property
     def free_symbols(self):
--- a/devito/ir/iet/nodes.py
+++ b/devito/ir/iet/nodes.py
@@ -201,9 +201,10 @@ class Call(Simple, Node):
 
     is_Call = True
 
-    def __init__(self, name, arguments=None, statement=None):
+    _traversable = ['arguments']
+
+    def __init__(self, name, arguments=None):
         self.name = name
-        self.statement = statement or True
         self.arguments = as_tuple(arguments)
 
     def __repr__(self):
@@ -228,10 +229,6 @@ class Call(Simple, Node):
     @property
     def defines(self):
         return ()
-
-    @property
-    def children(self):
-        return tuple(i for i in self.arguments if isinstance(i, Call))
 
 
 class Expression(Simple, Node):
--- a/devito/ir/iet/visitors.py
+++ b/devito/ir/iet/visitors.py
@@ -232,9 +232,7 @@ class CGen(Visitor):
     def visit_Call(self, o):
         arguments = self._args_call(o.arguments)
         code = '%s(%s)' % (o.name, ','.join(arguments))
-        if o.statement:
-            return c.Statement(code)
-        return c.Line(code)
+        return c.Statement(code)
 
     def visit_Conditional(self, o):
         then_body = c.Block(self._visit(o.then_body))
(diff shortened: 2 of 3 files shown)


devito/ir/iet/nodes.py
```diff
@@ -201,8 +201,9 @@ class Call(Simple, Node):
 
     is_Call = True
 
-    def __init__(self, name, arguments=None):
+    def __init__(self, name, arguments=None, statement=None):
         self.name = name
+        self.statement = statement if statement is not None else True
```
mloubout: statement or True

devito/ir/iet/visitors.py
```diff
@@ -166,7 +166,10 @@ def _args_call(self, args):
         ret = []
         for i in args:
             try:
-                if i.is_LocalObject:
+                if isinstance(i, Call):
+                    res = self.visit(i)
+                    ret.append(res.text)
```
mloubout: can oneline

devito/ir/iet/visitors.py
```diff
@@ -524,8 +530,12 @@ def visit_Iteration(self, o):
     def visit_Expression(self, o):
         return filter_sorted([f for f in self.rule(o)], key=attrgetter('name'))
 
+    def visit_Call(self, o):
+        symbols = self._visit(o.children)
+        symbols += [f for f in self.rule(o)]
```
mloubout: symbols.append(...)

devito/ir/iet/nodes.py
vincepandolfo: I knew there was a better way. Cheers

devito/ir/iet/visitors.py
vincepandolfo: That wouldn't work. The alternative would be `extend` but it's equivalent

devito/ir/iet/nodes.py
```diff
@@ -228,6 +229,10 @@ def free_symbols(self):
     def defines(self):
         return ()
 
+    @property
+    def children(self):
```
FabioLuporini: this is a bit hacky as it's bypassing a class attribute?

devito/ir/iet/nodes.py
```diff
@@ -201,8 +201,9 @@ class Call(Simple, Node):
 
     is_Call = True
 
-    def __init__(self, name, arguments=None):
+    def __init__(self, name, arguments=None, statement=None):
         self.name = name
+        self.statement = statement or True
```
FabioLuporini: ?

devito/ir/iet/visitors.py
```diff
@@ -166,7 +166,9 @@ def _args_call(self, args):
         ret = []
         for i in args:
             try:
-                if i.is_LocalObject:
+                if isinstance(i, Call):
+                    ret.append(self.visit(i).text)
```
FabioLuporini: .text ?

devito/ir/iet/nodes.py
FabioLuporini: ah, ok, "is_statement" (for homogeneity)

devito/ir/iet/nodes.py
vincepandolfo: What would be a better way?

devito/ir/iet/visitors.py
vincepandolfo: It's a field of the cgen object. Perhaps it would be easier to use `str`

FabioLuporini: ah. But why not just ``self.visit(i)`` then?

devito/ir/iet/nodes.py
FabioLuporini: don't know...off top of my head (pseudocode)

```
class Call(...):
    ...
    children = ('nested',)

    @cached_property
     def nested(self):
           return tuple(i for i in self.arguments if isinstance(i, Call))
```
?

tests/test_ir.py
```diff
@@ -793,3 +794,32 @@ def test_dimension_sort(self, expr, expected):
...
+    @pytest.mark.parametrize('mode,expected', [
+        ('free-symbols', '["f", "x"]'),
+        ('symbolics', '["f"]')
+    ])
+    def test_find_symbols_nested(self, mode, expected):
+        grid = Grid(shape=(4, 4, 4))
+        call = Call('foo', [
+            Call('bar', [
+                Symbol(name='x'),
+                Call('baz', [Function(name='f', grid=grid)])
```
btrb: bar?

btrb: oh nevermind, cross that.

devito/ir/iet/visitors.py
vincepandolfo: That does not return a string

vincepandolfo: Hmm, so here's the thing:
if using the `text` field, the semicolon is not appended even if it is a statement. This would make the `is_statement` field useless. Using `str` the cgen element is rendered properly and therefore the `is_statement` becomes necessary. What do you reckon would be better?

devito/ir/iet/nodes.py
vincepandolfo: the problem with that is that `_traversable` needs to have `__init__` arguments (so I can't use a `nested` property, but I can use `arguments`). `arguments` seems to work though. (although I can't find any logic to skip non-Node objects?)

FabioLuporini: sorry, my bad, that's what I meant! ``_traversable``, not ``children``

what error do you get with ``_traversable`` in place of ``children``?

devito/ir/iet/visitors.py
FabioLuporini: I **think** you have to use ``is_statement``?  Same reason I thought we should have used ``self.visit(i)`` -- how can this work (ie, with ``.text``) if we have, say, triply nested Calls? Maybe I'm missing something...would triply nested Calls work the current way?

Another possibility is that we move part of the logic inside ``visit_Call``

vincepandolfo: It works just fine using `.text` with however many nested calls (we are using `self.visit(i)` first after all). My only reservation is that it is a cgen internal field

FabioLuporini: oh ignore me...ignore me...I'm stupid

I think Im happy with whatever makes you happier. I don't really have a strong opinion here

devito/ir/iet/nodes.py
vincepandolfo: with `_traversable = ['arguments']`, none.
with `_traversable = ['nested'`]` (with nested as per your code) I get 
```
def _rebuild(self, *args, **kwargs):
        """Reconstruct self."""
        handle = self._args.copy()  # Original constructor arguments
        argnames = [i for i in self._traversable if i not in kwargs]
        handle.update(OrderedDict([(k, v) for k, v in zip(argnames, args)]))
        handle.update(kwargs)
>       return type(self)(**handle)
E       TypeError: __init__() got an unexpected keyword argument 'nested'
```

vincepandolfo:
Using `_traversable` seems to be breaking tests. I'm reverting to overriding `children`

FabioLuporini:
Merged

PR #881, opened 2019-07-24
more and more mpi fixes

FabioLuporini:
Fix:

* MPI and staggered grids
* MPI and non perfect loop nests:
```
for x  # *Must not* be OpenMP-parallelized
  for y
     ...
  haloupdate()  # *Must not* go before `x`
  for y
    ...
```
* MPI and naming of halo exchange routines (was mistakenly reusing the same name for different routines)

Also:

* Restrict some MPI tests to pipelines

### DIFF
--- a/devito/__init__.py
+++ b/devito/__init__.py
@@ -1,6 +1,9 @@
 from collections import namedtuple
 from itertools import product
 
+# Import the global `configuration` dict
+from devito.parameters import *  # noqa
+
 # API imports
 from devito.base import *  # noqa
 from devito.builtins import *  # noqa
@@ -17,7 +20,6 @@ from devito.compiler import compiler_registry
 from devito.dle import dle_registry
 from devito.dse import dse_registry
 from devito.logger import error, warning, info, logger_registry, set_log_level  # noqa
-from devito.parameters import *  # noqa
 from devito.profiling import profiler_registry
 
 
--- a/devito/__init__.py
+++ b/devito/__init__.py
@@ -10,6 +10,7 @@ from devito.builtins import *  # noqa
 from devito.data.allocators import *  # noqa
 from devito.equation import *  # noqa
 from devito.finite_differences import *  # noqa
+from devito.mpi import MPI  # noqa
 from devito.types import NODE, CELL, Buffer, SubDomain, SubDomainSet  # noqa
 from devito.types.dimension import *  # noqa
 
--- a/devito/builtins.py
+++ b/devito/builtins.py
@@ -92,6 +92,7 @@ class MPIReduction(object):
             self.v = comm.allreduce(np.asarray(self.n.data), self.op)[0]
 
 
+@dv.switchconfig(log_level='ERROR')
 def norm(f, order=2):
     """
     Compute the norm of a Function.
--- a/devito/ir/iet/nodes.py
+++ b/devito/ir/iet/nodes.py
@@ -11,7 +11,7 @@ import cgen as c
 from devito.data import FULL
 from devito.ir.equations import ClusterizedEq
 from devito.ir.iet import (IterationProperty, SEQUENTIAL, PARALLEL, PARALLEL_IF_ATOMIC,
-                           VECTOR, WRAPPABLE, ROUNDABLE, AFFINE, USELESS, OVERLAPPABLE)
+                           VECTOR, WRAPPABLE, ROUNDABLE, AFFINE, OVERLAPPABLE)
 from devito.ir.support import Forward, detect_io
 from devito.symbolics import FunctionFromPointer, as_symbol, ccode
 from devito.tools import (Signer, as_tuple, filter_ordered, filter_sorted, flatten,
@@ -888,8 +888,11 @@ class HaloSpot(Node):
         return ()
 
     @property
-    def is_Useless(self):
-        return USELESS in self.properties
+    def useless(self):
+        for i in self.properties:
+            if i.name == 'useless':
+                return i.val
+        return ()
 
     @property
     def is_Overlappable(self):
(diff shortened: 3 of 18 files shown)


devito/ir/iet/nodes.py
```diff
@@ -888,8 +888,11 @@ def hoistable(self):
         return ()
 
     @property
-    def is_Useless(self):
-        return USELESS in self.properties
+    def useless(self):
```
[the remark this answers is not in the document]
FabioLuporini: the compiler solves a complex problem -- generating halo exchange routines at the optimum depth in the code tree -- through a sequence of _simpler_ subtasks

one such subtaks creates HaloSpots, ie nodes where **potentially** a halo exchange is required. Ie, first the compiler acts conservatively, assuming a halo exchange is require as soon as a stencil expression is encountered (e.g., ``u[t, x-1]``)

later on, another subtask uses data dependence analysis to check that a halo exchange is **actually** required. Whey might it not be required? well, for example if the value of ``u`` hasn't changed since the last halo update. These useless HaloSpots are marked as "USELESS".

A final subtask, in the DLE, searches for all USELESS HaloSpots and drops them

devito/__init__.py
```diff
@@ -1,12 +1,16 @@
...
+# Import the global `configuration` dict
+from devito.parameters import *  # noqa
+
 # API imports
 from devito.base import *  # noqa
 from devito.builtins import *  # noqa
 from devito.data.allocators import *  # noqa
 from devito.equation import *  # noqa
 from devito.finite_differences import *  # noqa
+from devito.mpi import MPI  # noqa
```
mloubout: Wait shouldn't this be a try catch?

devito/builtins.py
```diff
@@ -92,6 +92,7 @@ def __exit__(self, exc_type, exc_value, traceback):
             self.v = comm.allreduce(np.asarray(self.n.data), self.op)[0]
 
 
+@dv.switchconfig(log_level='ERROR')
```
mloubout: Why?

devito/ir/iet/nodes.py
```diff
@@ -888,8 +888,11 @@ def hoistable(self):
...
+        for i in self.properties:
+            if i.name == 'useless':
```
mloubout: Rather than i.name ==, would be better to have a like `.useless` attribute?

devito/types/dense.py
```diff
@@ -167,7 +165,7 @@ def __staggered_setup__(self, **kwargs):
                     mask.append(-1)
                 else:
                     mask.append(0)
-            return tuple(mask)
+            return tuple(mask), True
```
mloubout: I thought this PR was droping the mask?

mloubout: Now that I think of it, I did but that's because I changed the FD accordingly too

devito/types/dense.py
```diff
@@ -1010,7 +994,11 @@ def __halo_setup__(self, **kwargs):
                 halo = (left_points, right_points)
             else:
                 raise TypeError("`space_order` must be int or 3-tuple of ints")
-            return tuple(halo if i.is_Space else (0, 0) for i in self.dimensions)
+            base = [halo if i.is_Space else (0, 0) for i in self.dimensions]
+            # left-/right-staggering require extra points
+            extra = [(-i, 0) if i < 0 else (0, i) for i in self.staggered]
```
mloubout: Wondering if all passes without this, I would think maybe OOB would take care of it and avoid this but I'm not sure.

tests/test_mpi.py
```diff
@@ -1411,8 +1446,8 @@ def test_adjoint_codegen(self, shape, kernel, space_order, nbpml, save):
         op_adj = solver.op_adj()
         adj_calls = FindNodes(Call).visit(op_adj)
 
-        assert len(fwd_calls) == 2
-        assert len(adj_calls) == 2
+        assert len(fwd_calls) == 1
+        assert len(adj_calls) == 1
```
mloubout: Why did this change again? Better halo exchange?

devito/__init__.py
FabioLuporini: no, it's safe, see [here](https://github.com/opesci/devito/blob/master/devito/mpi/distributed.py#L33)

devito/builtins.py
FabioLuporini: because ``norm`` produces the usual Operator log, which is painful (especially if you are in PERF/DEBUG mode) if you have codes with lots of Functions (eg elastic) for which a norm needs to be computed. With visco-elatic, the stdout is basically unreadable w/o this

devito/ir/iet/nodes.py
FabioLuporini: how do you mean? it is improvable, sure, but I don't get how you would do it? this is the ``useless`` attribute?

devito/types/dense.py
FabioLuporini: as you can see from the other changes, what this PR is dropping is the shape resizing based on the staggered mask

I also moved the  initialization
```
self.is_Staggered = ...
```
because attributes should be added to ``self`` only in the constructor, and never in subroutines

devito/types/dense.py
FabioLuporini: you suggested to do it this way, I haven't thought much about it? 
If you want, you can take a look (in another branch?) as you said you're gonna change parts of staggering anyway?

tests/test_mpi.py
FabioLuporini: yes

One less time-invariant halo exchange

devito/builtins.py
mloubout: I get that, but hard coding it here is not great, would change the config in the code that uses it not here

devito/ir/iet/nodes.py
mloubout: ok

devito/builtins.py
FabioLuporini: I don't think there's anything to be concerned about. It will still emit errors if there are errors; I don't think anybody cares about what the DSE/DLE do to the norm operator, and nobody will care if it takes 1 or 2 secs

FabioLuporini:
Merged.

PR #883, opened 2019-07-25
Generate less code with MPI

FabioLuporini:
Reuse MPI halo exchange routines for multiple Functions/halo-exchanges whenever possible

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ matrix:
           packages:
             - gcc-4.9
             - g++-4.9
-      env: DEVITO_ARCH=gcc-4.9 DEVITO_OPENMP=0 INSTALL_TYPE=pip_setup RUN_EXAMPLES=False MPI_INSTALL=1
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -19,7 +19,7 @@ matrix:
           packages:
             - gcc-4.9
             - g++-4.9
-      env: DEVITO_ARCH=gcc-4.9 DEVITO_OPENMP=0 INSTALL_TYPE=pip_setup RUN_EXAMPLES=False MPI_INSTALL=1
+      env: DEVITO_ARCH=gcc-4.9 DEVITO_OPENMP=0 INSTALL_TYPE=pip_setup RUN_EXAMPLES=False MPI_INSTALL=1 MPI_RESTRAIN=1
```
mloubout: What is that, and why only here?

FabioLuporini: oh no, it was actually missing before. It's in the openmp build too, it's been there for ages.
it prevents long running MPI tests from being executed on travis . 

If MPI_RESTRAIN==True **and** there's a True in a tuple parameter like [here](https://github.com/opesci/devito/blob/master/tests/test_mpi.py#L1489), then that test is skipped. The skipping logic is [here](https://github.com/opesci/devito/blob/master/tests/conftest.py#L327)

FabioLuporini:
Merged.

PR #884, opened 2019-07-25
Generate OPS stencils

vincepandolfo:
Introduces a new `ExternalFunctionCall` object to represent calls to external functions (such as OPS API calls) and uses it to initialize `ops_stencil` objects.

eg:
```
  ops_stencil S2D_UT0_5PT;
  ops_stencil S2D_UT1_1PT;
  int s2d_ut0_5pt[10] = {1, 0, -1, 0, 0, 1, 0, -1, 0, 0};
  S2D_UT0_5PT = ops_decl_stencil(2, 5, s2d_ut0_5pt, "S2D_UT0_5PT");
  int s2d_ut1_1pt[2] = {0, 0};
  S2D_UT1_1PT = ops_decl_stencil(2, 1, s2d_ut1_1pt, "S2D_UT1_1PT");
```


devito/ops/transformer.py
```diff
@@ -37,7 +42,44 @@ def opsit(trees, count):
...
+    ops_stencil = OpsStencil(stencil_name.upper())
+
+    return [
+        Expression(ClusterizedEq(Eq(
+            stencil_array,
+            ListInitializer(list(itertools.chain(*accesses)))
+        ))),
+        Expression(ClusterizedEq(Eq(
+            ops_stencil,
+            ExternalFunctionCall("ops_decl_stencil", [
```
maelso: namespace['ops_decl_stencil']

devito/ops/transformer.py
```diff
@@ -37,7 +42,44 @@ def opsit(trees, count):
...
+
+    pre_time_loop = stencil_arrays_initializations
+
+    return pre_time_loop, ops_kernel
+
+
+def to_ops_stencil(param, accesses):
+    dims = len(accesses[0])
+    pts = len(accesses)
+    stencil_name = "s%sd_%s_%spt" % (dims, param.name, pts)
```
maelso: put in ops/utils.py:
namespace['stencil_name'] = lambda dims, name, pts: 's%sd_%s_%spt' % (dims, name, pts)
and replace 
stencil_name = "s%sd_%s_%spt" % (dims, param.name, pts)
to
stencil_name = namespace['stencil_name'](dims, param.name, pts)

devito/ops/transformer.py
```diff
@@ -37,7 +42,44 @@ def opsit(trees, count):
...
+
+    stencil_array = SymbolicArray(
+        name=stencil_name,
+        dimensions=(pts * dims,),
+        dtype=np.int32
+    )
+
+    ops_stencil = OpsStencil(stencil_name.upper())
```
maelso: @vincepandolfo as I said on slack, the idea is to have something generic to use for all declaration.
Thus, I've created OPSDeclaration when I was working on ops_decl_block.

Therefore, I think we should wait for the 885 to be merged with the master. Then, use OPSDeclarator instead of OpsStencil.

This is to avoid having several classes that do the same thing.

vincepandolfo: That doesn't work as each Ops type needs to specify its type

vincepandolfo: Unless we pass the type name in the constructor, but I'm not a fan of that

maelso: ops_stencil = OPSDeclarator(stencil_name.upper(), namespace['ops_stencil_type'])

maelso: LocalObject receives name and type in its constructor. It is already a standard used.

devito/ops/node_factory.py
```diff
@@ -16,6 +16,8 @@ class OPSNodeFactory(object):
 
     def __init__(self):
         self.ops_args = OrderedDict()
+        self.ops_args_accesses = defaultdict(list)
+        self.ops_params = []
```
FabioLuporini: is it ok that a factory class keeps state?

devito/symbolics/extended_sympy.py
```diff
@@ -202,6 +202,52 @@ def base(self):
     __reduce_ex__ = Pickable.__reduce_ex__
 
 
+class ExternalFunctionCall(sympy.Expr, Pickable):
+
+    """
+    Symbolic representation of the C notation ``pointer->function(params)``.
```
FabioLuporini: copy-paste leftover?

devito/types/basic.py
```diff
@@ -938,6 +942,10 @@ def _C_typename(self):
     def _C_ctype(self):
         return self.dtype
 
+    @property
+    def function(self):
```
FabioLuporini: if you add this here, there's a chance we can drop something from subclasses? or maybe not?

devito/symbolics/search.py
```diff
@@ -46,6 +46,8 @@ def __init__(self, query, mode, deep=False):
         self.deep = deep
 
     def _next(self, expr):
+        if isinstance(expr, int):
```
FabioLuporini: the fact that all of a sudden we need this here, is very suspicious. What's going on?

also, you should use `is_integer`, but we should really try our best to avoid changing this piece of code

devito/symbolics/queries.py
```diff
@@ -40,7 +40,7 @@ def q_function(expr):
 
 
 def q_terminal(expr):
-    return expr.is_Symbol or expr.is_Indexed
+    return not isinstance(expr, int) and (expr.is_Symbol or expr.is_Indexed)
```
FabioLuporini: the fact that all of a sudden we need this here, is very suspicious. What's going on?

also, you should use is_integer, but we should really try our best to avoid changing this piece of code

devito/symbolics/printer.py
```diff
@@ -31,6 +32,11 @@ def _print_Function(self, expr):
         # There exist no unknown Functions
         if expr.func.__name__ not in self.known_functions:
             self.known_functions[expr.func.__name__] = expr.func.__name__
+        if isinstance(expr, devito.types.SymbolicArray):
```
FabioLuporini: this is odd...maybe SymbolicArray is just lacking proper ``__repr__`` and ``__str__`` which in turn will be used by the sympy printer

devito/symbolics/extended_sympy.py
```diff
@@ -262,6 +308,8 @@ def __new__(cls, params):
         for p in as_tuple(params):
             if isinstance(p, str):
                 args.append(Symbol(p))
+            elif isinstance(p, int):
+                args.append(Integer(p))
```
FabioLuporini: I assume you've checked already that if we don't cast to Integer sympy complains and raises an exception?

devito/ops/transformer.py
```diff
@@ -37,7 +42,44 @@ def opsit(trees, count):
...
+
+    return pre_time_loop, ops_kernel
+
+
+def to_ops_stencil(param, accesses):
+    dims = len(accesses[0])
+    pts = len(accesses)
+    stencil_name = "s%sd_%s_%spt" % (dims, param.name, pts)
+
+    stencil_array = SymbolicArray(
```
FabioLuporini: remind me what is this for, that is, what kind of C is it supposed to generate?

devito/ops/node_factory.py
vincepandolfo: This greatly simplifies the next steps

devito/types/basic.py
vincepandolfo: No subclass implements it

devito/symbolics/printer.py
vincepandolfo: `_print_Function` doesn't use `__str__` but manually puts stuff in parentheses

devito/ops/transformer.py
vincepandolfo: We want to generate `int array[size] = { x, x};`
If we use a normal Array we get no type (we'd get `array[size] = { x, x };`)

devito/symbolics/extended_sympy.py
vincepandolfo: If we don't cast to `Integer` this constructor complains

vincepandolfo: It doesn't complain but it seemed cleaner to use sympy types everywhere

devito/symbolics/queries.py
vincepandolfo: We need this because we now have `int` arguments in calls

devito/symbolics/extended_sympy.py
FabioLuporini: they should be converted internally to Integer

devito/symbolics/queries.py
FabioLuporini: just use ``Integer(13)`` in the Call arguments. Then it is a sympy expression and it should just work

devito/ops/node_factory.py
FabioLuporini: oki, as long as factories are never used as instance attributes, I'm fine with it

devito/ops/transformer.py
FabioLuporini: what, really? do you have a (minimal) reproducer by any chance? I'd be happy to take a look. This could be relevant for several other things

vincepandolfo: If you take this branch and change this line to use `Array` you can observe it on any example (I've been using `example_diffusion`)

devito/symbolics/extended_sympy.py
vincepandolfo: :+1:

devito/symbolics/queries.py
vincepandolfo: It wasn't the call sorry - it was the array dimension. Using `Integer` there works

devito/symbolics/extended_sympy.py
```diff
@@ -202,6 +202,52 @@ def base(self):
     __reduce_ex__ = Pickable.__reduce_ex__
 
 
+class ExternalFunctionCall(sympy.Expr, Pickable):
```
FabioLuporini: could you add a pickle test next to those for the other sympy objects implemented in this file? you'll see it's just two lines

devito/ops/transformer.py
vincepandolfo: Are we expecting the API to change?

vincepandolfo: And actually, even then, this is only used in one place. There's no need to take it out

maelso: Shouldn't we add all OPS conventions?
like [that](https://github.com/opesci/devito/blob/5008e36a3b2ba268e257cc9fcdc91b48c170ac3b/devito/yask/utils.py#L66)

devito/ir/iet/scheduler.py
```diff
@@ -147,7 +147,7 @@ def iet_insert_decls(iet, external):
     allocator = Allocator()
     for k, v in MapSections().visit(iet).items():
         if k.is_Expression:
-            if k.is_scalar_assign:
+            if k.is_scalar_assign or k.is_tensor_assign:
```
FabioLuporini: is it worth adding a ``is_definition`` property returning precisely this, that is ``self.is_scalar_assign or self.is_tensor_assign`` ?

devito/ops/transformer.py
FabioLuporini: I think it should go in the namespace too.

Although it's not beautiful, it's what we do with all other names (same in YASK), so for homogeneity I  think it's better this way

devito/types/basic.py
```diff
@@ -893,6 +893,10 @@ def update(self, **kwargs):
     _pickle_kwargs = AbstractCachedFunction._pickle_kwargs + ['dimensions', 'scope']
 
 
+class SymbolicArray(Array):
```
FabioLuporini: leftover

devito/ir/iet/utils.py
```diff
@@ -110,7 +110,9 @@ def derive_parameters(nodes, drop_locals=False):
     parameters = tuple(s for s in symbols if s.name not in defines)
 
     # Drop globally-visible objects
-    parameters = [p for p in parameters if not isinstance(p, (Literal, Macro))]
+    parameters = [
+        p for p in parameters if not isinstance(p, (DefaultDimension, Literal, Macro))
```
FabioLuporini: are we sure this is necessary?

devito/ir/iet/scheduler.py
vincepandolfo: `is_scalar_assign` isn't used anywhere else so we can join the two together safely

devito/ir/iet/utils.py
vincepandolfo: Another option would be overriding `free_symbols` in Array to not return `DefaultDimension`s

devito/ops/transformer.py
```diff
@@ -25,19 +33,55 @@ def opsit(trees, count):
...
     ops_kernel = Callable(
-        namespace['ops-kernel'](count),
+        namespace['ops_kernel'](count),
         ops_expressions,
         "void",
         parameters
     )
 
-    return ops_kernel
+    print(ops_kernel)
```
FabioLuporini: leftover

devito/ir/iet/scheduler.py
FabioLuporini: yep, I'd like that

devito/symbolics/printer.py
```diff
@@ -31,6 +32,11 @@ def _print_Function(self, expr):
         # There exist no unknown Functions
         if expr.func.__name__ not in self.known_functions:
             self.known_functions[expr.func.__name__] = expr.func.__name__
+        if isinstance(expr, devito.types.Array):
```
FabioLuporini: is this still necessary?

vincepandolfo: Yes. Otherwise we would get `int array(10) = { ... }`

devito/ir/iet/nodes.py
```diff
@@ -288,14 +288,12 @@ def is_tensor(self):
...
     @property
-    def is_scalar_assign(self):
-        """True if a scalar, non-increment expression."""
-        return self.is_scalar and not self.is_Increment
-
-    @property
-    def is_tensor_assign(self):
-        """True if an array initialization"""
-        return self.is_tensor and isinstance(self.expr.rhs, ListInitializer)
+    def is_assign(self):
```
FabioLuporini: is_definition

definition is the word in C that implies introduction of a symbol **and** allocation of memory (eg on the stack)

FabioLuporini: otherwise, all expressions are actually assignments (well, not all, but most)

devito/ir/iet/nodes.py
```diff
@@ -288,13 +288,16 @@ def is_tensor(self):
         return not self.is_scalar
 
     @property
-    def is_scalar_assign(self):
-        """True if a scalar, non-increment expression."""
-        return self.is_scalar and not self.is_Increment
+    def is_definition(self):
+        """
+        True if it is an assignment, False otherwise
```
georgebisbas: definition?

devito/ops/types.py
```diff
@@ -104,3 +105,13 @@ def as_coeff_Add(self):
         return sympy.S.Zero, self
 
     __repr__ = __str__
+
+
+class OpsStencil(basic.LocalObject):
```
georgebisbas: Will it be better to have them as OPS_stencil?

devito/ir/iet/nodes.py
```diff
@@ -288,13 +288,16 @@ def is_tensor(self):
...
+        """
+        return ((self.is_scalar and not self.is_Increment) or
+                (self.is_tensor and isinstance(self.expr.rhs, ListInitializer)))
```
mloubout: `is_deifinition` is a bit of an odd name to me, why not `is_assign`?

devito/symbolics/extended_sympy.py
mloubout: Why is this an Expr and not a sympy function?

devito/ir/iet/nodes.py
vincepandolfo: See previous discussion with @fablup

devito/ir/iet/nodes.py
vincepandolfo: See previous discussion with @fablup

devito/symbolics/extended_sympy.py
vincepandolfo: It can have an arbitrary number of arguments

mloubout: So does a sympy function. A sympy function is literraly `func.name(*args)` which you are rebuilding in some ways rather than using it

devito/ops/operator.py
```diff
@@ -27,18 +27,21 @@ def _specialize_iet(self, iet, **kwargs):
...
         assert (d == dims[0] for d in dims), \
             "The OPS backend currently assumes that all kernels \
             have the same number of dimensions"
 
-        self._headers.append(namespace['ops-define-dimension'](dims[0]))
+        self._headers.append(namespace['ops_define_dimension'](dims[0]))
         self._includes.append('stdio.h')
 
-        body = [ops_init, ops_partition, iet, ops_exit]
+        body = [ops_init, ops_partition, *pre_time_loop, iet, ops_exit]
```
maelso: ops_partition must be after *pre_time_loop

devito/ops/types.py
```diff
@@ -104,3 +105,13 @@ def as_coeff_Add(self):
...
+
+    def __init__(self, name, *args, **kwargs):
+        super().__init__(name, np.void, *args, **kwargs)
+
+    @property
+    def _C_typename(self):
+        return "ops_stencil"
```
vmickus: This should be in the dictionary `namespace`.

devito/ir/iet/nodes.py
```diff
@@ -288,13 +288,16 @@ def is_tensor(self):
...
 
     @property
     def defines(self):
-        return (self.write,) if self.is_scalar else ()
+        return (self.write,) if self.is_scalar or self.is_definition else ()
```
FabioLuporini: I think the right thing to have here is ``self.is_definition`` , that is drop ``self.is_scalar or``

devito/ir/iet/utils.py
FabioLuporini: I think I'd prefer that?

after all, we do not generate the DefaultDimension right?

devito/ops/types.py
FabioLuporini: No. No underscores in class names

devito/ops/types.py
FabioLuporini: probably yes

devito/symbolics/printer.py
FabioLuporini: maybe this should go under a ``_print_Array`` ?

my feeling is that an ``isinstance`` within a visitor handler is a bit odd

devito/ir/iet/nodes.py
btrb: i'm not sure if it has been decided already, but this sounds a little strange
i'd vote to something like `is_assign` or `is_assignment`

vincepandolfo: @FabioLuporini I lest `self.is_scalar` because it's a different condition than the old `is_scalar_assign` (I wanted to avoid any change in behaviour due to that) but I agree it makes more sense to only have `is_definition`

@btrb a definition is a different thing compared to an assignment. Here we are specifically referring to a definition

devito/symbolics/extended_sympy.py
vincepandolfo: You're right - a simple `sympy.Function` seems to work just fine. I was under the impression it didn't work for some reason :/

devito/symbolics/printer.py
vincepandolfo: A separate `_print_Array` doesn't seem to work

FabioLuporini:
Merged.

PR #885, opened 2019-07-25
Created OPSDeclarator class

maelso:
This is a generic class for OPS declarations.

Example:
Pass `name` and `type` args.
stencil = OPSDeclarator('stencil_n', 'ops_stencil')
block = OPSDeclarator('block', 'ops_block')

### DIFF
--- a/devito/ops/types.py
+++ b/devito/ops/types.py
@@ -109,13 +109,14 @@ class OpsAccess(basic.Basic, sympy.Basic):
 
 class OPSDeclarator(basic.LocalObject):
     """
-    OPS declarator
+    OPS declarator is used for any declarations that need to pass a type defined in OPS.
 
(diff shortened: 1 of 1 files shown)


FabioLuporini:
LGTM. @vincepandolfo ?

vincepandolfo:
@FabioLuporini as mentioned in #884 I would prefer separate objects (`OpsStencil`, `OpsDat`, etc.) instead of this but this would work too

FabioLuporini:
I just thought that the dtype of these objects is statically known and it is known in advance, hence ``dtype`` should actually be a class attribute, just like [here](https://github.com/opesci/devito/blob/master/devito/yask/types.py#L274)

So I not only prefer separate objects, but also class-level dtypes

PR #886, opened 2019-07-25
Created an specific compiler for ops, producing an extra file for the…

Gabrielsvc:
Added the new compiler class, its initializing parameters, how the operator calls it and a test for it.


devito/ops/compiler.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class CompilerOPS(configuration['compiler'].__class__):
+    def __init__(self, *args, **kwargs):
+        super(CompilerOPS, self).__init__(*args, **kwargs)
+
+    def jit_compile_ops(self, soname, code, hcode):
+        target = str(self.get_jit_dir().joinpath(soname))
+        src_file = "%s.%s" % (target, self.src_ext)
+        src_file2 = "%s.h" % (target)
```
FabioLuporini: can we use more representative names than ``src_file`` and ``src_file2`` ? how about ``c_file`` and ``h_file`` ?

devito/ops/compiler.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+__all__ = ['CompilerOPS']
+
+
+class CompilerOPS(configuration['compiler'].__class__):
+    def __init__(self, *args, **kwargs):
+        super(CompilerOPS, self).__init__(*args, **kwargs)
+
+    def jit_compile_ops(self, soname, code, hcode):
```
FabioLuporini: you're not exploiting inheritance, I don't like this

devito/ops/compiler.py
```diff
@@ -0,0 +1,74 @@
...
+        target = str(self.get_jit_dir().joinpath(soname))
+        src_file = "%s.%s" % (target, self.src_ext)
+        src_file2 = "%s.h" % (target)
+        cache_dir = self.get_codepy_dir().joinpath(soname[:7])
```
FabioLuporini: this entire block of code is copy-paste. Not exploiting inheritance -> mistake

this needs fixing

devito/ops/compiler.py
FabioLuporini: ``c_code`` and ``h_code`` ?

devito/ops/compiler.py
```diff
@@ -0,0 +1,74 @@
...
+                                    sleep_delay=sleep_delay)
+            _, _, _, recompiled_kernel = \
+                compile_from_string(self, target, hcode, src_file2,
+                                    cache_dir=cache_dir,
+                                    debug=configuration['debug-compiler'],
+                                    sleep_delay=sleep_delay)
+
+            toc = time()
+
+        if recompiled and recompiled_kernel:
```
FabioLuporini: this is odd... 

I now think that if you have to compile two files, the **caller** should just call ``jit_compile`` twice

devito/ops/operator.py
```diff
@@ -42,3 +44,14 @@ def _specialize_iet(self, iet, **kwargs):
         body = [ops_init, ops_partition, iet, ops_exit]
 
         return List(body=body)
+
+    @property
+    def hcode(self):
```
FabioLuporini: this isn't pythonic

FabioLuporini:
@vincepandolfo can you take a look at this please

devito/ops/__init__.py
```diff
@@ -17,6 +19,7 @@
                   'speculative': PlatformRewriter})
 
 # The following used by backends.backendSelector
+from devito.ops.compiler import CompilerOPS as Compiler # noqa
```
mloubout: I don't really like that, this should be another compiler in `devito.compiler` not a different one here, especially considering there is a lot of copy-paste from devito compiler.

devito/ops/compiler.py
Gabrielsvc: Noted, going to make it clearer

devito/ops/compiler.py
Gabrielsvc: My initial idea was to override the function, but as i utilize an extra parameter i opted for copy-pasting it. Is there a way to overwrite it?

devito/ops/operator.py
Gabrielsvc: noted, going to change it

devito/ops/compiler.py
```diff
@@ -0,0 +1,74 @@
...
+            tic = time()
+            # Spinlock in case of MPI
+            sleep_delay = 0 if configuration['mpi'] else 1
+            _, _, _, recompiled = \
+                compile_from_string(self, target, code, src_file,
+                                    cache_dir=cache_dir,
+                                    debug=configuration['debug-compiler'],
+                                    sleep_delay=sleep_delay)
+            _, _, _, recompiled_kernel = \
+                compile_from_string(self, target, hcode, src_file2,
```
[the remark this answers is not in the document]
Gabrielsvc: From what i understood the files are created as we call the compile_from_string function, my bad. Going to look further into it and try to call only once.

devito/ops/compiler.py
Gabrielsvc: Huh...
Ok going to make that change and see how it works

devito/ops/compiler.py
Gabrielsvc: How so? Could you comment further?

devito/ops/__init__.py
Gabrielsvc: Our original idea was defining everything that would work differently for ops inside the ops folder. I thought that the compilerOPS case would be similar to the operatorOPS, and tried to make them work alike. As we work further on the compiler, there won't be as much copy-paste

devito/ops/compiler.py
```diff
@@ -0,0 +1,74 @@
...
+import warnings
+
+from devito.logger import debug
+from devito.parameters import configuration
+
+
+__all__ = ['CompilerOPS']
+
+
+class CompilerOPS(configuration['compiler'].__class__):
```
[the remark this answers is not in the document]
Gabrielsvc: While testing the simple example the compiler class would subclass from GNUCompiler, i've got this from the Yask Compiler

devito/ops/__init__.py
mloubout: If you look at how Yask is setup for example, the backend selector takes care of the selection but everything is still called the same and the compiler goes in `compiler.py`.  The same way `OperatorOPS` shouldn't exist the backend selector should choose ops depending on DEVITO_BACCKEND but the call should still be `Operator(..)` no matter what  not `OperatorOPS`

devito/ops/compiler.py
FabioLuporini: you're not overriding ``jit_compile`` and/or calling it via ``super``

you're just copy-pasting, which is of course bad

devito/ops/compiler.py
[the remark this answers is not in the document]
vincepandolfo: You don't need to compile the header file by itself. Just write it in a file in the same folder

devito/ops/__init__.py
Gabrielsvc: I'm not quite sure I follow... From what I saw from Yask, it declares its own compiler inside the __init__.py 

https://github.com/opesci/devito/blob/cfb3d380690ea10d05498024f365d51b68cc4a16/devito/yask/__init__.py#L70

devito/ops/compiler.py
```diff
@@ -0,0 +1,17 @@
...
+    def __init__(self, *args, **kwargs):
+        super(CompilerOPS, self).__init__(*args, **kwargs)
+
+    def jit_compile(self, soname, code, hcode):
+        super(CompilerOPS, self).jit_compile(soname, code)
+        target = str(self.get_jit_dir().joinpath(soname))
+        h_file = "%s.h" % (target)
+        kernel_file = open(h_file, "w")
+        for kernel in hcode:
+            kernel_file.write(kernel)
```
vincepandolfo: you're not compiling anything now?

Gabrielsvc: Now i just write the  h. file after the creation of the .c file in the jit-compile function. Tried to keep them near each other. Further on i plan to add the call to the ops translator and so on on this compiler, if that makes sense.

devito/ops/compiler.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+__all__ = ['CompilerOPS']
+
+
+class CompilerOPS(configuration['compiler'].__class__):
+    def __init__(self, *args, **kwargs):
+        super(CompilerOPS, self).__init__(*args, **kwargs)
+
+    def jit_compile(self, soname, code, hcode):
```
FabioLuporini: ccode and hcode

tests/test_ops.py
```diff
@@ -69,4 +69,4 @@ def test_kernel_generation(self, equation, expected):
 
         operator = OperatorOPS(eval(equation))
 
-        assert str(operator._ops_kernels[0]) == expected
+        assert str(operator.hcode) == expected
```
vincepandolfo: where is the `hcode` property defined?

devito/ops/compiler.py
```diff
@@ -0,0 +1,17 @@
...
+        super(CompilerOPS, self).jit_compile(soname, code)
+        target = str(self.get_jit_dir().joinpath(soname))
+        h_file = "%s.h" % (target)
+        kernel_file = open(h_file, "w")
+        for kernel in hcode:
```
vincepandolfo: why is this a loop?

devito/ops/compiler.py
Gabrielsvc: Ops... Changing it, my bad!

devito/ops/compiler.py
Gabrielsvc: It shouldn't be, was using hcode as a list before, but changed it. Pushing the changes!

tests/test_ops.py
Gabrielsvc: Wasn't using it as a property, changed it. Thanks for poiting out, pushing the changes now

devito/ops/__init__.py
mloubout: Duplicate of line 9

devito/ops/compiler.py
```diff
@@ -0,0 +1,16 @@
...
+
+__all__ = ['CompilerOPS']
+
+
+class CompilerOPS(configuration['compiler'].__class__):
+    def __init__(self, *args, **kwargs):
+        super(CompilerOPS, self).__init__(*args, **kwargs)
+
+    def jit_compile(self, soname, code, hcode):
+        super(CompilerOPS, self).jit_compile(soname, code)
```
vincepandolfo: This should not be called here

vincepandolfo:
When making new commits can you write what you did in the commit message?

devito/ops/__init__.py
Gabrielsvc: Noted! going to remove it

devito/ops/compiler.py
Gabrielsvc: My idea was using the already defined  ```jit_compile``` function of the Compiler class and only appending what I need the CompilerOPS to do (writing the .h file). What do you suggest?

vincepandolfo: There's a few problems with this (and where you placed this):

- you need to call the ops translator
- you need to create the header file BEFORE compiling (therefore calling jit_compile from your superclass simply won't work) 
- what platform are we targeting? For example, CUDA would require multiple compilation steps (for host and device code)

My suggestion is to make this the base class of all OPS compilers where all we do is write the code to files and invoke the translator, then (in future PRs) subclass this compiler to create platform specific implementations. 

For some of these we might be able to reuse code from the `Compiler` class but for most we won't.

Gabrielsvc: Done! Now calling ops translator and generating code from it

vincepandolfo:
LGTM

devito/ops/operator.py
FabioLuporini: better _hcode

FabioLuporini:
How are we doing here? @vincepandolfo @Gabrielsvc could you give me some final feedback?

FabioLuporini:
also, this should be rebased on top of latest master

vincepandolfo:
As per my previous comment, LGTM. @FabioLuporini

Gabrielsvc:
I think the base work for the ops compiler is set. Going to rebase to the latest master. I think maybe Travis need some extra configuration since the failed check seems to be related to the travis environment not having an `OPS_INSTALL_PATH` variable

FabioLuporini:
Merged.

PR #887, opened 2019-07-26
bench: autotuning args and README updates

georgebisbas:
Update benchmarking:

To decide:
Should we mention about event='main' for evaluating each section?
Should section0 be  default ?
Should it be an argument?


benchmarks/user/benchmark.py
```diff
@@ -87,7 +87,8 @@ def from_value(ctx, param, value):
         click.option('--dle', callback=from_value,
                      type=click.Choice(['noop'] + configuration._accepted['dle']),
                      help='Devito loop engine (DLE) mode'),
-        click.option('-a', '--autotune', is_flag=True,
+        click.option('-a', '--autotune', default='aggressive',
+                     type=click.Choice(['basic','aggressive']),
```
georgebisbas: Note for reviewers: This is a reminder to take care about --autotune in bench and plotting that is probably obsolete.
Also to take care about the default values of autotuning that are attached to an operator.
The problem is that after a 'bench', user cannot do a 'plot' with same Operator arguments.

benchmarks/user/benchmark.py
```diff
@@ -333,7 +334,7 @@ def param_string(self, params):
             devito_params['to'] = params['time_order']
             devito_params['dse'] = params['dse']
             devito_params['dle'] = params['dle']
-            devito_params['at'] = params['autotune']
+            devito_params['at'] = 'aggressive'
```
georgebisbas: Same note as above.

benchmarks/user/README.md
```diff
@@ -211,6 +211,51 @@ Do not forget to pin processes, especially on NUMA systems; below, we do so with
 ```
 numactl --cpubind=0 --membind=0 python benchmark.py ...
 ```
+
+## Benchmarking performance
```
FabioLuporini: Benchmark output ?

benchmarks/user/README.md
```diff
@@ -211,6 +211,51 @@ Do not forget to pin processes, especially on NUMA systems; below, we do so with
...
+
+In order for a user to benchmark the performance of an operator calculating
```
FabioLuporini: This paragraph is useless since the last bench-related PR landed. See [this line](https://github.com/opesci/devito/blob/master/benchmarks/user/benchmark.py#L397) and [this line](https://github.com/opesci/devito/blob/master/benchmarks/user/benchmark.py#L405)

FabioLuporini: you may replace it with something along the lines of:

_The GFlops/s and GPoints/s performance, Operational Intensity (OI) and execution time are emitted to standard output at the end of each run. Further, when running in ``bench`` mode, a \`.json\` file is produced (see \`python benchmark.py bench --help\` for more info)

benchmarks/user/README.md
```diff
@@ -211,6 +211,51 @@ Do not forget to pin processes, especially on NUMA systems; below, we do so with
...
+ 
+```
+python benchmark.py plot -P acoustic -d 512 512 512 -so 12 --tn 100 --max-bw 12.8
+ --flop-ceil 80 "linpack"
+```
+
+In order to derive --max-bw for your machine we strongly advise that you use
+[STREAM][http://www.cs.virginia.edu/stream/ref.html].
+
 While a benchmark is running, you can have some useful programs running in
```
FabioLuporini: where should this now go?

benchmarks/user/README.md
```diff
@@ -211,6 +211,51 @@ Do not forget to pin processes, especially on NUMA systems; below, we do so with
...
+
+```
+`DEVITO_LOGGING=DEBUG` python benchmark.py bench -P acoustic -d 512 512 512 -so 12 --tn 100
+```
+
+A `.json` file will be created in a folder named `results` except if otherwise specified
+with the `-r` option specifying the results directory.
+
+
+## Plotting a roofline
```
FabioLuporini: Generation of roofline models ?

FabioLuporini: is "plotting" the right word? it's more than a "simple chart/plot" at the end of the day. Maybe I'm being too picky...

benchmarks/user/README.md
```diff
@@ -211,6 +211,51 @@ Do not forget to pin processes, especially on NUMA systems; below, we do so with
...
+
+In order for a user to plot the results obtained from the `bench` in the previous
```
FabioLuporini: I've iterated through what you wrote to try improving it. Lemme know what you think

_To generate a [roofline model](https://people.eecs.berkeley.edu/~kubitron/courses/cs252-S12/handouts/papers/RooflineVyNoYellow.pdf) from the results obtained in \`bench\` mode, one can execute \`benchmark.py\` in \`plot\` mode. For example, the command

\`\`\`
 python benchmark.py plot -P acoustic -d 512 512 512 -so 12 --tn 100 -a --max-bw 12.8
  --flop-ceil 80 linpack
\`\`\`

will generate a roofline model for the results obtained from

\`\`\`
python benchmark.py bench -P acoustic -d 512 512 512 -so 12 --tn 100 -a
\`\`\`

The \`plot\` mode expects the same arguments used in \`bench\` mode plus two additional arguments to generate the roofline:

* `--max-bw FLOAT`
     DRAM bandwidth (GB/s)
 * `--flop-ceil <FLOAT TEXT>`
     CPU machine peak. A 2-tuple (float, str) is expected, representing the performance ceil (GFlops/s) and how the ceil was obtained (ideal peak, linpack, ...), respectively

In addition, points can be annotated with the runtime value, passing the `--point-runtime` argument.

To obtain the DRAM bandwidth of a system, we advise to use [STREAM](http://www.cs.virginia.edu/stream/ref.html)

To obtain the ideal CPU peak, one should instantiate this formula
```
#[cores] · #[avx units] · #[vector lanes] · #[FMA ports] · [ISA base frequency]
```
More details [in this paper](https://arxiv.org/pdf/1807.03032.pdf).

benchmarks/user/README.md
FabioLuporini: Probably below "Example commands" 

You mistakenly pushed it down inside another section

benchmarks/user/benchmark.py
FabioLuporini: why?

FabioLuporini: what does `params['autotune']` contain? I don't remember

georgebisbas: It results to either `True` or `False`, making it non-compatible with `aggressive`,  `basic` or related """OFF"" version...

benchmarks/user/README.md
```diff
@@ -221,6 +222,68 @@ or to keep the memory footprint under control:
...
+
+#[cores] · #[avx units] · #[vector lanes] · #[FMA ports] · [ISA base frequency]
+
+More details in this [paper][https://arxiv.org/pdf/1807.03032.pdf].
+
+
+In order for a user to plot the results obtained from the `bench` in the previous
+paragraph user shoud execute `benchmark.py` in plot mode. It is essential that 
+some more arguments should be provided at this stage.
+The necessary arguments can be also seen through `python benchmark.py plot --help`.
```
FabioLuporini: leftover paragraph to be dropped?

benchmarks/user/benchmark.py
FabioLuporini: maybe then you should use this? `configuration['autotuning'].level`

see [here](https://github.com/opesci/devito/blob/master/benchmarks/user/benchmark.py#L181)

don't know, but hardcoding "aggressive" doesn't look like a fix

benchmarks/user/README.md
georgebisbas: yes

benchmarks/user/benchmark.py
georgebisbas: Tried that on Friday but did not work, will try again.

benchmarks/user/benchmark.py
```diff
@@ -245,17 +250,17 @@ def plot(problem, **kwargs):
         warning("Could not load any results, nothing to plot. Exiting...")
         sys.exit(0)
 
-    gflopss = bench.lookup(params=kwargs, measure="gflopss", event='main')
-    oi = bench.lookup(params=kwargs, measure="oi", event='main')
-    time = bench.lookup(params=kwargs, measure="timings", event='main')
+    gflopss = bench.lookup(params=kwargs, measure="gflopss", event='section0')
+    oi = bench.lookup(params=kwargs, measure="oi", event='section0')
+    time = bench.lookup(params=kwargs, measure="timings", event='section0')
```
mloubout: Well this is not gonna be correct for tti staggered for example that is split into two major loops + extras. I am not sure what should be taken but taking one section by name seems dangerous.

georgebisbas: I agree absolutely with that, may be we can discuss tomorrow.

FabioLuporini: this is hacky, yes, but it can't be worse than not working

also, I **thought** staggered tti, just like all other examples supported by benchmark.py, actually ended up putting all FD stencils in section0

benchmarks/user/benchmark.py
```diff
@@ -87,8 +87,9 @@ def from_value(ctx, param, value):
         click.option('--dle', callback=from_value,
                      type=click.Choice(['noop'] + configuration._accepted['dle']),
                      help='Devito loop engine (DLE) mode'),
-        click.option('-a', '--autotune', is_flag=True,
-                     help='Switch auto tuning on/off')
+        click.option('-a', '--autotune', default='aggressive',
+                     type=click.Choice(['off', 'basic', 'aggressive', 'max']),
```
FabioLuporini: Maybe ``type=configuration['autotuning']._accepted``  ?

benchmarks/user/benchmark.py
georgebisbas: Should we add a comment or a note in a README.md about that?

FabioLuporini: we should aim for a fix...maybe not this PR, but that's what we should do. Maybe a hacky one! It's just a benchmarking script after all.

Maybe we could let users attach a name to Eqs, to be used in place of numbers in the section name?
```
Eq(...., name=X)
```
and then in the generated code
```
/* Begin section X */
...
....
```

though this is probably future work

benchmarks/user/README.md
```diff
@@ -129,10 +129,12 @@ attempt a small batch of block shapes. With `autotune='aggressive'`, the
 auto-tuning phase will take up more time, but it will also evaluate more
 block-shapes. 
 
-When running `python benchmark.py -a ...`, the underlying Operators will
+When running `python benchmark.py ...`, the underlying Operators will
 automatically be run in aggressive mode, that is as
 `op.apply(autotune='aggressive')`.
 
+The autotuning method can be selected with `-a ... `
```
FabioLuporini: :%s/method/mode

benchmarks/user/README.md
```diff
@@ -221,6 +223,61 @@ or to keep the memory footprint under control:
 watch numastat -m
 ```
 
+## Benchmarks' output
```
FabioLuporini: Benchmark output

benchmarks/user/README.md
```diff
@@ -221,6 +223,61 @@ or to keep the memory footprint under control:
...
+
+The GFlops/s and GPoints/s performance, Operational Intensity (OI) and 
+execution time are emitted to standard output at the end of each run. 
+Further, when running in bench mode, a `.json` file is produced
+(see `python benchmark.py bench --help` for more info) in a folder named
+`results` except if otherwise specified with the `-r` option specifying
```
FabioLuporini: Drop "specifying the results directory"

benchmarks/user/README.md
```diff
@@ -221,6 +223,61 @@ or to keep the memory footprint under control:
...
+the results directory.
+
+So the isotropic acoustic wave forward Operator in a `512**3` grid, space order
```
FabioLuporini: How is this example any useful? we have analogous examples in the section above, and this doesn't seem to add anything. We can probably drop this

benchmarks/user/README.md
```diff
@@ -221,6 +223,61 @@ or to keep the memory footprint under control:
...
+`DEVITO_LOGGING=DEBUG` python benchmark.py bench -P acoustic -d 512 512 512 -so 12 --tn 100
+```
+
+## Generating a roofline model
+
+To generate a roofline model from the results obtained in `bench` mode,
+one can execute `benchmark.py` in `plot` mode. For example, the command
+
+```
+python benchmark.py plot -P acoustic -d 512 512 512 -so 12 --tn 100 -a aggressive
```
FabioLuporini: this command is currently split over three lines of the .md file; should be a one liner

benchmarks/user/README.md
```diff
@@ -221,6 +223,61 @@ or to keep the memory footprint under control:
...
+python benchmark.py bench -P acoustic -d 512 512 512 -so 12 --tn 100 -a
+```
+
+The `plot` mode expects the same arguments used in `bench` mode plus 
+two additional arguments to generate the roofline:
+
+*    --max-bw FLOAT
+    DRAM bandwidth (GB/s)
+*    --flop-ceil <FLOAT TEXT>
+    CPU machine peak. A 2-tuple (float, str) is expected,
```
FabioLuporini: text should wrap at 79 chars

benchmarks/user/benchmark.py
```diff
@@ -177,8 +178,13 @@ def cli_bench(problem, **kwargs):
     """
     Complete benchmark with multiple simulation and performance parameters.
     """
+
     mode_benchmark()
-    kwargs['autotune'] = configuration['autotuning'].level
+    if kwargs['autotune'] != configuration['autotuning'].level:
```
FabioLuporini: why is this in `bench`, but not in `run`?

probably this should be a callback when accessing the `autotune` parameter for the first time from click. See [here](https://github.com/opesci/devito/blob/master/benchmarks/user/benchmark.py#L45) for example -- it makes sure that if one runs with `-so 12`, then the argument value is interpreted for consistency as `[12]` (just like `-so 12 16` becomes [12, 16]) instead of `16` (ie, an int instead of a list)

so with a suitable callback you make it return a tuple such as `('aggressive', 'preemptive')` where clearly `'preemptive'` is fixed, while 'aggressive' depends on whatever the user has provided in input

FabioLuporini: so if you do it this way, you automatically get the right and **same** thing in all modes

benchmarks/user/README.md
georgebisbas: fixed

benchmarks/user/README.md
georgebisbas: fixed

benchmarks/user/README.md
georgebisbas: fixed

benchmarks/user/README.md
georgebisbas: It states the need for DEVITO_LOGGING =DEBUG (?)

benchmarks/user/README.md
georgebisbas: FIXED

benchmarks/user/README.md
georgebisbas: i THINK fixed

benchmarks/user/benchmark.py
georgebisbas: I am a bit confused here...let me work it out...

.travis.yml
```diff
@@ -118,16 +118,16 @@ script:
...
+  - if [[ $RUN_EXAMPLES == 'True' ]]; then python benchmarks/user/benchmark.py test -P acoustic ; fi
   - if [[ $RUN_EXAMPLES == 'True' ]]; then python examples/seismic/acoustic/acoustic_example.py --full; fi
   - if [[ $RUN_EXAMPLES == 'True' ]]; then python examples/seismic/acoustic/acoustic_example.py --full --checkpointing; fi
   - if [[ $RUN_EXAMPLES == 'True' ]]; then python examples/seismic/acoustic/acoustic_example.py --constant --full; fi
   - if [[ $RUN_EXAMPLES == 'True' ]]; then python examples/seismic/elastic/elastic_example.py --2d; fi
   - if [[ $RUN_EXAMPLES == 'True' ]]; then python examples/seismic/elastic/elastic_example.py; fi
   - if [[ $RUN_EXAMPLES == 'True' ]]; then python examples/misc/linalg.py mat-vec mat-mat-sum transpose-mat-vec; fi
-  - if [[ $RUN_EXAMPLES == 'True' ]]; then python examples/seismic/tti/tti_example.py -a; fi
-  - if [[ $RUN_EXAMPLES == 'True' ]]; then python examples/seismic/tti/tti_example.py -a --noazimuth; fi
+  - if [[ $RUN_EXAMPLES == 'True' ]]; then python examples/seismic/tti/tti_example.py -a aggressive; fi
```
FabioLuporini: is ``-a aggressive`` even necessary?

benchmarks/user/README.md
```diff
@@ -127,12 +127,14 @@ in the generated code.
 With `autotune=True`, the auto-tuner gets set in `basic` mode, which will only
 attempt a small batch of block shapes. With `autotune='aggressive'`, the
 auto-tuning phase will take up more time, but it will also evaluate more
-block-shapes. 
+block-shapes.
```
FabioLuporini: you can drop the - from block shapes

benchmarks/user/README.md
```diff
@@ -236,6 +238,58 @@ for more information, and in particular `python make-pbs.py generate --help`.
 `make-pbs.py` is especially indicated if interested in running strong scaling
 experiments.
 
+## Benchmark output
+
+The GFlops/s and GPoints/s performance, Operational Intensity (OI) and
+execution time are emitted to standard output at the end of each run.
+Further, when running in bench mode, a `.json` file is produced
```
FabioLuporini: \`bench\`

benchmarks/user/README.md
FabioLuporini: DEBUG is now automatic when running benchmark.py since the last related PR: https://github.com/opesci/devito/blob/master/benchmarks/user/benchmark.py#L408

benchmarks/user/benchmark.py
```diff
@@ -47,6 +47,9 @@ def default_list(ctx, param, value):
                      callback=default_list, help='Time order of the simulation'),
         click.option('-t', '--tn', default=250,
                      help='End time of the simulation in ms'),
+        click.option('-a', '--autotune', default='aggressive',
```
FabioLuporini: now that I think about it, having moved `-a` from `option_performance` to `option_simulation` doesn't make any sense

examples/seismic/tti/tti_example.py
```diff
@@ -47,8 +47,9 @@ def run(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
                         help="Preset to determine the physical problem setup")
     parser.add_argument('--noazimuth', dest='azi', default=False, action='store_true',
                         help="Whether or not to use an azimuth angle")
-    parser.add_argument('-a', '--autotune', default=False, action='store_true',
-                        help="Enable autotuning for block sizes")
+    parser.add_argument('-a', '--autotune', default='off',
+                        choices=['off', 'basic', 'max', 'aggressive'],
```
FabioLuporini: configuration['autotuning']._accepted...

examples/seismic/tti/tti_example.py
```diff
@@ -47,8 +47,9 @@ def run(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
...
+                        help="Choice of autotuning method for block sizes")
```
FabioLuporini: 'Operator auto-tuning mode' ?

examples/seismic/tti/tti_example.py
```diff
@@ -47,8 +47,9 @@ def run(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
                         help="Preset to determine the physical problem setup")
     parser.add_argument('--noazimuth', dest='azi', default=False, action='store_true',
                         help="Whether or not to use an azimuth angle")
-    parser.add_argument('-a', '--autotune', default=False, action='store_true',
-                        help="Enable autotuning for block sizes")
+    parser.add_argument('-a', '--autotune', default='off',
```
FabioLuporini: also, shouldn't this be updated in all examples, including acoustic, elastic, visco-elastic?

benchmarks/user/README.md
```diff
@@ -236,6 +238,51 @@ for more information, and in particular `python make-pbs.py generate --help`.
...
+one can execute `benchmark.py` in `plot` mode. For example, the command
+
+```
+python benchmark.py plot -P acoustic -d 512 512 512 -so 12 --tn 100 -a aggressive --max-bw 12.8 --flop-ceil 80 linpack
+```
+
+will generate a roofline model for the results obtained from
+
+```
+python benchmark.py bench -P acoustic -d 512 512 512 -so 12 --tn 100 -a
```
FabioLuporini: missing "aggressive"

benchmarks/user/benchmark.py
FabioLuporini: should it not be in both? why should `run` behave differently than `bench` when it gets to autotuning?

you have to understand how click callbacks work, I think. And you have an example right above, so it shouldn't be too difficult

georgebisbas: I am not contrary to that. I speak from the view of avoiding duplicate code.

benchmarks/user/README.md
georgebisbas: fixed

benchmarks/user/benchmark.py
georgebisbas: added, currently duplicated

.travis.yml
```diff
@@ -118,8 +118,8 @@ script:
   - if [[ $INSTALL_TYPE == 'conda' ]]; then py.test --durations=20 --cov devito tests/; fi
   # Run remaining specialized examples and tests
   # Additional seismic operator tests
-  - if [[ $RUN_EXAMPLES == 'True' ]]; then python benchmarks/user/benchmark.py test -P tti -so 4 -a -d 20 20 20 -n 5; fi
-  - if [[ $RUN_EXAMPLES == 'True' ]]; then python benchmarks/user/benchmark.py test -P acoustic -a; fi
+  - if [[ $RUN_EXAMPLES == 'True' ]]; then python benchmarks/user/benchmark.py test -P tti -so 4 -d 20 20 20 -n 5; fi
+  - if [[ $RUN_EXAMPLES == 'True' ]]; then python benchmarks/user/benchmark.py test -P acoustic; fi
```
mloubout: why no more autotuning?

benchmarks/user/benchmark.py
```diff
@@ -98,8 +138,12 @@ def from_value(ctx, param, value):
         click.option('--dle', callback=from_value,
                      type=click.Choice(['noop'] + configuration._accepted['dle']),
                      help='Devito loop engine (DLE) mode'),
-        click.option('-a', '--autotune', is_flag=True,
-                     help='Switch auto tuning on/off')
+        click.option('-bs', '--block-shape', nargs=3, type=(int, int, int),
+                     callback=config_blockshape, multiple=True, is_eager=True,
+                     help='Loop-blocking shape, bypass autotuning'),
+        click.option('-a', '--autotune', default='aggressive', callback=config_autotuning,
```
mloubout: default to basic maybe?

benchmarks/user/benchmark.py
```diff
@@ -226,6 +262,10 @@ def bench(problem, **kwargs):
                    'ceil was obtained (ideal peak, linpack, ...)')
 @click.option('--point-runtime', is_flag=True, default=True,
               help='Annotate points with runtime values')
+@click.option('--section', default=None,
+              help='Code section for which the roofline is plotted')
```
mloubout: Is there a way to tell people what this neame look like, like "should be `section_0,1,...` as I am not sure people will know what the names are out of the box

benchmarks/user/benchmark.py
FabioLuporini: mmhh, I think I'm just concerned people trying out devito might forget to set it to advanced. I think I prefer aggressive

benchmarks/user/benchmark.py
FabioLuporini: I think if they use this option, they are expected to look at the generated C code?

.travis.yml
FabioLuporini: oh, it's like before. Autotuning is on by default now

benchmarks/user/benchmark.py
mloubout: I don't know about that, think would still prefer a "naming convention is ...."

benchmarks/user/benchmark.py
mloubout: ok

.travis.yml
mloubout: ok

FabioLuporini:
Merged

PR #889, opened 2019-07-30
Add make-pbs.py script

FabioLuporini:


benchmarks/user/make-pbs.py
```diff
@@ -0,0 +1,74 @@
+import os
```
mloubout: Is this script tested somewhere?

FabioLuporini: no. Use it as a template/support tool -- that's the idea (see also updated README)

FabioLuporini:
merged.

PR #888, opened 2019-07-29
examples: Improve damp setup to lower memory consumption at init time

FabioLuporini:

PR #890, opened 2019-07-31
Import fix

mloubout:
This gets rids of and is preemptive change for future 3.8 support

```
/usr/local/bin/ipython:1: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  #!/usr/local/opt/python/bin/python3.7
```

On top of #888


FabioLuporini:
Merged

PR #891, opened 2019-08-01
Generate ops_block and ops_dats

vincepandolfo:
Like so:
```
  block = ops_decl_block(2, "block");
  int u_dim[2] = {100, 100};
  int u_base[2] = {0, 0};
  int u_d_p[2] = {2, 2};
  int u_d_m[2] = {-2, -2};
  ops_dat u_dat[2] = {ops_decl_dat(block, 1, u_dim, u_base, u_d_m, u_d_p, &(u[0]), "float", "ut0"), ops_decl_dat(block, 1, u_dim, u_base, u_d_m, u_d_p, &(u[1]), "float", "ut1")};
```


devito/ops/operator.py
```diff
@@ -26,22 +30,45 @@ def _specialize_iet(self, iet, **kwargs):
         ops_partition = Call(namespace['ops_partition'], Literal('""'))
         ops_exit = Call(namespace['ops_exit'])
 
+        ops_block = OpsBlock('block')
+
         dims = []
+        to_dat = set()
```
FabioLuporini: watch out, this might lead to non-deterministic code gen. You may wanna use
```
from devito.tools import filter_sorted
```
?

devito/ops/transformer.py
```diff
@@ -82,6 +85,104 @@ def to_ops_stencil(param, accesses):
...
+        d_m_val = tuple(sympify([-(p[1] + h[1]) for p, h in zip(padding, halo)]))
+
+        ops_dat_array = Array(
+            name=namespace['ops_dat_name'](f.name),
+            dimensions=(DefaultDimension(name='dat', default_value=time_dims),),
+            dtype='ops_dat',
+        )
+
+        dat_decls = []
+        for i in range(time_dims):
```
mloubout: `for i in range(time_dims)`
So if there are a thousand time-steps this creates a thousand declarations?

vincepandolfo: We need a declaration for each time stepping variables (`t0`, `t1`, etc.) because the time loop is not offloaded to OPS.

mloubout: `time_dims = f.shape[time_pos]`

That's one per time-steps, so again for 1000 time-steps that's 1000 declaration. The time dimension is not always a buffered dimension, it can be a normal dimension in which case this is gonna break.

tests/test_ops.py
```diff
@@ -124,3 +126,98 @@ def test_to_ops_stencil(self, _accesses):
             Symbol(stencil_name),
             Literal('"%s"' % stencil_name.upper())
         )
+
+    def test_create_ops_dat_time_function(self):
+        grid = Grid(shape=(4))
+
+        u = TimeFunction(name='u', grid=grid, space_order=2)
```
mloubout: The test should work with and without `save` and as pointed out above, with something like `save=200` (which is quite small) this is gonna blow up.

devito/ops/operator.py
```diff
@@ -26,22 +31,47 @@ def _specialize_iet(self, iet, **kwargs):
...
 
             pre_time_loop.extend(pre_loop)
             self._ops_kernels.append(ops_kernel)
-            dims.append(len(trees[n].dimensions))
 
         assert (d == dims[0] for d in dims), \
             "The OPS backend currently assumes that all kernels \
             have the same number of dimensions"
 
+        ops_block_init = Expression(ClusterizedEq(Eq(
```
maelso: What do you think of creating a function `create_ops_block` into transformer.py, as you did for ops_dat?

vincepandolfo: It's two lines of code, I'm not sure it's worth it

FabioLuporini:
merged.

PR #892, opened 2019-08-01
Various fix

mloubout:
Various fix from previous PR

Fixed adjoint FD with a better implementation and better corresponding test.


devito/finite_differences/finite_difference.py
```diff
@@ -66,16 +66,15 @@ def first_derivative(expr, dim, fd_order=None, side=centered, matvec=direct,
 
     >>> g = Function(name='g', grid=grid)
     >>> first_derivative(f*g, dim=x, matvec=transpose)
-    f(x, y)*g(x, y)/h_x - f(x + h_x, y)*g(x + h_x, y)/h_x
+    -f(x, y)*g(x, y)/h_x + f(x - h_x, y)*g(x - h_x, y)/h_x
```
FabioLuporini: ?

mloubout: adjoint of [0, 1, -1] is [-1, 1, 0] for some reason was wrong before but was still and FD approx so wasn't breaking  stuff.

Issue #893, opened 2019-08-01
OPS test is incorrect


mloubout:
https://github.com/opesci/devito/blob/935621d5f020b08f3a229a53632b304a9761c8e0/tests/test_ops.py#L34

This test produces a wrong stencil but test it against it so passes, `.dxl` and `.dxr` have indices `x, x+hx, x+2hx` (or minus sign) the test produces stencil that only has `x, x+2hx`. (index 0, +2 and -2)

This test needs to be fixed asap as this makes me think like the ops backend is processing the stencils incorrectly.

@vincepandolfo and OPS team please look at it as `expected` should contain all of `vt0(0, 1), vt0(0, -1), vt0(1, 0), vt0(-1, 0)`

FabioLuporini:
Im removing myself from this and reassigning to @btrb which subsumes his team , and @vincepandolfo

btrb:
ok, we will have a look now

in fact, just run the test and it returned a Fail in:

```
>       assert result[1].expr.rhs.name == namespace['ops_decl_stencil'].name
E       AttributeError: 'ops_decl_stencil' object has no attribute 'name'

tests/test_ops.py:120: AttributeError
```
which is weird regarded that @vincepandolfo cleverly defined `namespace['ops_decl_stencil']` as a `Function`

btrb:
it seems the correct `space_indices` are being generated in `new_ops_arg` at `OPSNodeFactory` for the case 

`Eq(v,v.dxl + v.dxr - v.dyr - v.dyl)`

```
[0, 0]
[-2, 0]
[0, 2]
[2, 0]
[0, -2]
[1, 0]
[0, -1]
[-1, 0]
[0, 1]
```

so it seems the method is not completely broken

btrb:
the last `expr.expr` from [here](https://github.com/opesci/devito/blob/master/devito/ops/transformer.py#L34) seems to be ok:

`Eq(v[t0, x, y], 0.5*(r0*v[t0, x - 2, y] - r0*v[t0, x + 2, y] - r1*v[t0, x, y - 2] + r1*v[t0, x, y + 2]) + 2.0*(-r0*v[t0, x - 1, y] + r0*v[t0, x + 1, y] + r1*v[t0, x, y - 1] - r1*v[t0, x, y + 1]))`

to the problem must be either in `make_ops_ast` or in `node_factory`

btrb:
[...]

PR #894, opened 2019-08-02
Fixed OPS code gen bug

vincepandolfo:
See #893.

The problem was that `hashable_content` in `OpsAccess` only used `self.base`, meaning that sympy would cache them incorrectly

### DIFF
--- a/example.py
+++ b/example.py
@@ -1,0 +1,6 @@
+from devito import TimeFunction, Grid, Operator, Eq
+
+grid = Grid((4, 4))
+v = TimeFunction(name='v', grid=grid, space_order=2)
+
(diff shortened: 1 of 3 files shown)


example.py
```diff
@@ -0,0 +1,6 @@
+from devito import TimeFunction, Grid, Operator, Eq
```
mloubout: debug file?

mloubout:
The weird thing is, the test was failing on `vector2` branch because it was actually producing the correct stencil, so I don't know what I di that fixed it too because I didn't touch OPS.

example.py
vincepandolfo: yup, careless `git add .`

PR #896, opened 2019-08-02
Notebook small fix

btrb:
(I tried to squeeze 3 commits into one with `git rebase -i 52df` but i'm not sure if that resulted in any effect)


FabioLuporini:
> (I tried to squeeze 3 commits into one with git rebase -i 52df but i'm not sure if that resulted in any effect)

as you can see from the commit list, it did not

FabioLuporini:
Thanks, merged

PR #897, opened 2019-08-05
removing :class: from comments

btrb:
This PR is for removing obsolete `:class:` from commentaries.

### DIFF
--- a/devito/ops/types.py
+++ b/devito/ops/types.py
@@ -64,7 +64,7 @@ class OpsAccess(basic.Basic, sympy.Basic):
     ----------
     base : OpsAccessible
         Symbol to access
-    indices: [int]
+    indices: list of tuples of int
         Indices to access
     """
 
--- a/devito/profiling.py
+++ b/devito/profiling.py
@@ -34,8 +34,8 @@ class Profiler(object):
     def instrument(self, iet):
         """
         Enrich the Iteration/Expression tree ``iet`` adding nodes for C-level
-        performance profiling. In particular, turn all :class:`Section`s within ``iet``
-        into :class:`TimedList`s.
+        performance profiling. In particular, turn all Sections within ``iet``
+        into TimedLists.
         """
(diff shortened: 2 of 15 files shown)


devito/ops/types.py
```diff
@@ -64,7 +64,7 @@ class OpsAccess(basic.Basic, sympy.Basic):
     ----------
     base : OpsAccessible
         Symbol to access
-    indices: [int]
+    indices: list of tuples of int
```
FabioLuporini: list of int 
?

btrb: it's not something like `[1,2,3,4]`
it's more like `[[1,2],[3,4]]` instead

devito/profiling.py
```diff
@@ -100,7 +100,7 @@ class AdvancedProfiler(Profiler):
     # Override basic summary so that arguments other than runtime are computed.
     def summary(self, arguments, dtype):
         """
```
FabioLuporini: Could you **cut** this entire  docstring and paste it in the superclass Profiler.summary (deleting what's currently in there)?

devito/ops/types.py
FabioLuporini: list of 2-tuples of int 
?

btrb: it will depend on the number of dimensions.
it could be `[[1,2,3],[4,5,6]]` if 3D
right, @vicenzo ?

devito/profiling.py
btrb: ok

devito/ops/types.py
vincepandolfo: It's actuallt a list of `sympy.Integers`

FabioLuporini: the PR is 99% OK apart from this minor thing. Since there are too many open PRs, Im gonna merge it as is now. Please fix this in the next OPS PR

btrb: ok

PR #899, opened 2019-08-05
 topology fix

mloubout:
Tiny fix to DAG setup.

Mainly, the `Indexed` have to be added to depency list if the yare lhs of one of the eq of the cluster of equations. Currently was assuming all indexed in rhs are not lhs.

Now 

```
from devito import *
from devito.types import *

from sympy import pprint
grid = Grid((10, 10, 10))
t = TimeFunction(name="t", grid=grid, space_order=4)
r = TimeFunction(name="r", grid=grid, space_order=4)

eq_t = Eq(t.forward, r.forward)
eq_r = Eq(r.forward, 2*r)

op = Operator([eq_t, eq_r])
```

produces

```
            #pragma ivdep
            #pragma omp simd aligned(r,t:32)
            {
              r[t1][x + 4][y + 4][z + 4] = 2*r[t0][x + 4][y + 4][z + 4] + 2;
              t[t1][x + 4][y + 4][z + 4] = r[t1][x + 4][y + 4][z + 4] + t[t0][x + 4][y + 4][z + 4];
            }
```

that has the correct dependency

### DIFF
--- a/devito/dse/manipulation.py
+++ b/devito/dse/manipulation.py
@@ -192,7 +192,11 @@ def topological_sort(exprs):
                 # Eq(f, f + 1)
                 continue
             elif r.is_Indexed and r not in mapper.keys():
-                # Only scalars enforce an ordering
+                # Only scalars or indexed that not lhs of equations
+                continue
+            elif not e.lhs.is_Indexed:
+                # Skip dependency if lhs is a temporary that automatically caries
+                # dependency
                 continue
             else:
                 dag.add_edge(mapper[r], e, force_add=True)
--- a/devito/dse/manipulation.py
+++ b/devito/dse/manipulation.py
@@ -191,7 +191,7 @@ def topological_sort(exprs):
                 # Avoid cyclic dependences, such as
                 # Eq(f, f + 1)
                 continue
-            elif r.is_Indexed:
+            elif r.is_Indexed and r not in mapper.keys():
                 # Only scalars enforce an ordering
                 continue
             else:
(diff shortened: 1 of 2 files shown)


mloubout:
Need some tweaking for corner case will update tmr

FabioLuporini:
as I explained already on slack, this makes no sense. There might a bug lurking somewhere and this fix might even be necessary, but definitely not for the MFE above. Devito honours the semantics inherent in the input equation ordering -- it only reorders the equations for performance when this doesn't impact the computation semantics. You cannot expect Devito to reorder the equations based on I don't know what, and especially in the DSE

mloubout:
> You cannot expect Devito to reorder the equations based on I don't know what

This is *not* I don't know what, this is the dependency rule that defines the DAG and is valid for every variable except, and this one is I don't know what, let's ignore indexed object. If something depends on something else it has to be computed first. This is applied to spatial loop to make sure equation are clusterized according to dependency, it should be organized and sorted properly within a loop too. THe DAG itself does not have a "semantic order" knowledge, only dependency and the only way to have semantically correct equations is to enforce dependency.

devito/dse/manipulation.py
```diff
@@ -178,29 +178,29 @@ def compact_temporaries(temporaries, leaves):
...
-                # Skip dependency if lhs is a temporary that automatically caries
-                # dependency
+            elif r.is_Indexed:
+                # Only scalars enforce an ordering
                 continue
             else:
                 dag.add_edge(mapper[r], e, force_add=True)
 
     processed = dag.topological_sort()
+    # Append indexed eqaution at the  end in the user-provided order
```
btrb: equation?
double spaces after `the`

PR #900, opened 2019-08-06
Ops parallel loop

maelso:
This pull request creates the range used in ops_par_loop:
 - int par_loop_range = {x_m, x_M, y_m, y_M};

and replace
 
`/* Begin section0 */

    {

    #pragma omp simd aligned(damp, u, vp : 32)

    {

        float r3 = dt * dt;
        float r2 = vp[x + 2][y + 2] * vp[x + 2][y + 2];
        float r0 = r2 * damp[x + 1][y + 1];
        float r1 = r2;
        u[t1][x + 2][y + 2] = 1.99999991e-4F * (2.5e+3F * r0 * dt * u[t2][x + 2][y + 2] - 2.0e+2F * r1 * r3 * u[t0][x + 2][y + 2] + 5.0e+1F * (r1 * r3 * u[t0][x + 1][y + 2] + r1 * r3 * u[t0][x + 2][y + 1] + r1 * r3 * u[t0][x + 2][y + 3] + r1 * r3 * u[t0][x + 3][y + 2]) + 1.0e+4F * u[t0][x + 2][y + 2] - 5.0e+3F * u[t2][x + 2][y + 2]) / (5.0e-1F * r0 * dt + 1.0F);
      }
    }
    /* End section0 */`
to 

`/* Begin section0 */
    
    ops_par_loop(OPS_Kernel_0, "OPS_Kernel_0", block, 2, par_loop_range,
                 ops_arg_dat(damp_dat, 1, S2D_DAMP_1PT, "float", OPS_READ),
                 ops_arg_dat(u_dat[t0], 1, S2D_UT0_5PT, "float", OPS_READ),
                 ops_arg_dat(u_dat[t1], 1, S2D_UT1_1PT, "float", OPS_WRITE),
                 ops_arg_dat(u_dat[t2], 1, S2D_UT2_1PT, "float", OPS_READ),
                 ops_arg_dat(vp_dat, 1, S2D_VP_1PT, "float", OPS_READ),
                 ops_arg_gbl(&dt, 1, "float", OPS_READ));
    
/* End section0 */`


FabioLuporini:
probably a few tests would be useful

even something trivial, such as checking that the generated ``ops_arg_dat`` are as expected

aside from that, I think this is fine? @vincepandolfo

devito/ops/transformer.py
```diff
@@ -30,8 +31,23 @@ def opsit(trees, count):
...
+
+    par_loop_array = Array(
+        name=namespace['ops_par_loop_range_name'],
+        dimensions=(DefaultDimension(name='len', default_value=sum(len(x)
+                                                                   for x in it_range)),),
+        dtype=np.int32,
+    )
+
+    par_loop_range = Expression(ClusterizedEq(Eq(
+        Symbol(par_loop_array.name),
```
vincepandolfo: This is wrong. It won't generate the correct LHS.
It has to be the Array object, not a symbol of the same name

devito/ops/transformer.py
```diff
@@ -45,17 +61,61 @@ def opsit(trees, count):
...
-        to_ops_stencil(p, node_factory.ops_args_accesses[p])
+    # Extract all symbols that need to be converted to ops_dat
+    to_dat = set()
+    symbols = set(FindSymbols('symbolics').visit(trees[0].root))
+    symbols -= set(FindSymbols('defines').visit(trees[0].root))
+    to_dat |= symbols
+
+    # To ensure deterministic code generation we order the datasets to
+    # be generated (since a set is an unordered collection)
+    to_dat = filter_sorted(to_dat)
```
vincepandolfo: We can use `parameters` here, no need to use FindSymbols

devito/ops/transformer.py
```diff
@@ -45,17 +61,61 @@ def opsit(trees, count):
...
+                                 [Literal('%s' % name_to_ops_dat[p.name]), 1,
+                                  Literal(name_to_ops_stencil[p.name]),
+                                  Literal('"float"'),
+                                  Literal('%s' % namespace['ops_read']
+                                          if p.read_only else namespace['ops_write'])]))
+
+    par_loop = Call(namespace['ops_par_loop'], [
+        Literal(ops_kernel.name),
+        Literal('"%s"' % ops_kernel.name),
+        Symbol(block.name),
```
vincepandolfo: Might aswell make OpsBlock subclass symbol

devito/ops/transformer.py
```diff
@@ -45,17 +61,61 @@ def opsit(trees, count):
...
+    # be generated (since a set is an unordered collection)
+    to_dat = filter_sorted(to_dat)
+
+    name_to_ops_dat = {}
+    pre_loop = []
+    for f in to_dat:
+        if f.is_Constant:
+            continue
+
+        _ops_dat = create_ops_dat(f, name_to_ops_dat, block)
```
vincepandolfo: Also the ops_dats have already been created in the Operator

devito/ops/transformer.py
maelso: The range is generated correctly. But I think I understand what you mean. It seems valid to have an Array in the espressions tree.
Just replace _Symbol(par_loop_array.name)_ to _par_loop_array_ keep working and solves this.

devito/ops/transformer.py
maelso: They aren't. Stencil, dat and ops_par_loop are correlated, so they are all in the same place.

devito/ops/transformer.py
maelso: > probably a few tests would be useful

I agree

devito/ops/transformer.py
maelso: This was not created here. It was just moved. Because I need the dats to be here.
stencil and dat are used into ops_par_loop call.

maelso: This was done in PR #891 about decl_dat, so this change is beyond the scope of this PR.

maelso: Just using parameters will break, I tried. So to maintain the purpose of this PR I left findSymbols created earlier.

devito/ops/transformer.py
vincepandolfo: The range is, the initialisation isn't. `int range = { x_m, x_M, y_m, y_M}` is wrong. It should be `int range[4] = { x_m, x_M, y_m, y_M}`

devito/ops/transformer.py
vincepandolfo: Check my PR (#901)

devito/ops/transformer.py
vincepandolfo: They are. See https://github.com/opesci/devito/blob/master/devito/ops/operator.py#L55
`name_to_ops_dat` can then be passed into `opsit` to retrieve them. Check #901

devito/ops/transformer.py
vincepandolfo: The reasoning behind creating the `ops_dat` in the `Operator` instead of the `opsit` function is that we can potentially use the same data in multiple loop nests. If we create the `ops_dat` in `opsit` we could end up recreating datasets that are used in multiple nests insteed of reusing them

devito/ops/operator.py
```diff
@@ -33,32 +31,20 @@ def _specialize_iet(self, iet, **kwargs):
...
-        # be generated (since a set is an unordered collection)
-        to_dat = filter_sorted(to_dat)
-
-        name_to_ops_dat = {}
         pre_time_loop = []
-        for f in to_dat:
-            if f.is_Constant:
-                continue
-
-            pre_time_loop.extend(create_ops_dat(f, name_to_ops_dat, ops_block))
```
vincepandolfo: oops, didn't see you removed this. This should stay here as per my other comment

devito/ops/transformer.py
maelso: This is the return: _int par_loop_range[4] = {x_m, x_M, y_m, y_M};_
The value is being set when _the par_loop_array_ is created. See https://github.com/maelso/devito/blob/8c6ecbc27aaf15c88e0ae2b216213219fc31633e/devito/ops/transformer.py#L42

devito/ops/transformer.py
maelso: Check files changed.

maelso: > name_to_ops_dat can then be passed into opsit
Seems like a good idea.

devito/ops/transformer.py
vincepandolfo: Yup it's correct with your latest commit!

devito/ops/utils.py
```diff
@@ -24,3 +24,9 @@
 namespace['ops_dat_d_p'] = lambda i: '%s_d_p' % i
 namespace['ops_dat_d_m'] = lambda i: '%s_d_m' % i
 namespace['ops_dat_name'] = lambda i: '%s_dat' % i
+namespace['ops_par_loop_range_name'] = 'par_loop_range'
+namespace['ops_par_loop'] = 'ops_par_loop'
+namespace['ops_arg_dat'] = 'ops_arg_dat'
+namespace['ops_arg_gbl'] = 'ops_arg_gbl'
+namespace['ops_read'] = 'OPS_READ'
```
vincepandolfo: This could be a `Macro` already so it can be used directly

devito/ops/utils.py
```diff
@@ -24,3 +24,9 @@
 namespace['ops_dat_d_p'] = lambda i: '%s_d_p' % i
 namespace['ops_dat_d_m'] = lambda i: '%s_d_m' % i
 namespace['ops_dat_name'] = lambda i: '%s_dat' % i
+namespace['ops_par_loop_range_name'] = 'par_loop_range'
+namespace['ops_par_loop'] = 'ops_par_loop'
+namespace['ops_arg_dat'] = 'ops_arg_dat'
```
vincepandolfo: This (and the next one) can be safely made into functions (like other ops api calls)

devito/ops/transformer.py
```diff
@@ -45,17 +61,69 @@ def opsit(trees, count):
...
+    if data.is_Constant:
+        return Call(namespace['ops_arg_gbl'],
+                             [Byref(Constant(name=data.name.replace('*', ''))),
+                              1, Literal('"%s"' % data._C_typedata),
+                              Literal(namespace['ops_read'])])
+    else:
+        return Call(namespace['ops_arg_dat'],
+                             [Literal('%s' % name_to_ops_dat[data.name]), 1,
+                              Literal(name_to_ops_stencil[data.name]),
+                              Literal('"float"'),
```
vincepandolfo: This might not be a float

devito/ops/transformer.py
```diff
@@ -45,17 +61,69 @@ def opsit(trees, count):
...
+def create_ops_arg(data, name_to_ops_dat, name_to_ops_stencil):
+    if data.is_Constant:
+        return Call(namespace['ops_arg_gbl'],
+                             [Byref(Constant(name=data.name.replace('*', ''))),
+                              1, Literal('"%s"' % data._C_typedata),
+                              Literal(namespace['ops_read'])])
+    else:
+        return Call(namespace['ops_arg_dat'],
+                             [Literal('%s' % name_to_ops_dat[data.name]), 1,
+                              Literal(name_to_ops_stencil[data.name]),
```
vincepandolfo: Doesn't need to be a Literal

devito/ops/transformer.py
```diff
@@ -67,6 +135,7 @@ def to_ops_stencil(param, accesses):
     )
 
     ops_stencil = OpsStencil(stencil_name.upper())
+    name_to_ops_stencil[param.name] = ops_stencil.name
```
vincepandolfo: You can pass the ops_stencil here, not only the name

devito/ops/transformer.py
maelso: Can not, this is a call

devito/ops/transformer.py
maelso: Changed!

devito/ops/transformer.py
vincepandolfo: You can make this a Function

maelso: I believe a Call is the right thing here.

vincepandolfo: Why?

maelso: As far as I know
- ops_arg_gbl(T *data, int dim, char *type, ops access acc);
- ops_arg_dat(ops dat dat, ops stencil stencil, char *type, ops access acc);
should mapped to a Devito Call, just as we do with ops_par_loop(...), ops_init(...), ops_exit(). 

I know that in some places Call has not been used, but I would like to keep a standard whenever we **call** something from the OPS API.

This is useful for testing, too.

vincepandolfo: Fair enough. I think however it would prefer to use `Functions` when possible and leave `Call`s for when they are necessary (ops_par_loop, ops_init, ops_exit)

maelso: > leave Calls for when they can't be avoided

Why do you want to avoid Call?

vincepandolfo: In general I think it's preferable to use other people's stuff when possible. It's one less thing we need to maintain

PR #902, opened 2019-08-06
viscoelastic.operators: Fix ordering of equations

rhodrin:
In `viscoelastic/operators.py` the ordering of the equations was incorrect, tau and r were flipped. This PR corrects this.


mloubout:
Fine with me, but ccode still has wrong order anyway.

FabioLuporini:
> Fine with me, but ccode still has wrong order anyway.

probably / hopefully due to the DSE bug you found in topological_sort

FabioLuporini:
merged

PR #914, opened 2019-08-27
examples+MPI: Several fixes related to MPI benchmarking and execution of examples

georgebisbas:
This PR makes the following contributions:

- Enables the use of `mpirun` with non constant `vp` that was causing hanging by changing the way minimum is returned from `builtins.py`
- Changes parameters in `model.py` for a more stable `viscoelastic` model.
- Changes the order of the returned values from elastic and `viscoelastic` in order to
be compatible with `benchmark.py`.

### DIFF
--- a/devito/builtins.py
+++ b/devito/builtins.py
@@ -6,8 +6,6 @@ from sympy import Abs, Pow
 import numpy as np
 
 import devito as dv
-from devito.logger import warning
-
 
 __all__ = ['assign', 'smooth', 'norm', 'sumall', 'inner', 'mmin', 'mmax']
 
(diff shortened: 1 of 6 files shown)


devito/builtins.py
```diff
@@ -225,5 +227,6 @@ def mmax(f):
         return np.max(f)
 
     with MPIReduction(f, op=dv.mpi.MPI.MAX) as mr:
+        warning("MPIReduction is used.")
```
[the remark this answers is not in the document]
georgebisbas: Like that?
```
    if not isinstance(f, dv.Function):
        if configuration['mpi']:
              with MPIReduction(f, op=dv.mpi.MPI.MAX) as mr:
                      warning("returning local max of array, use mmx(v::Function) for global maximum")
                      mr.n.data[0] = np.max(f.data_ro_domain).item()
                      return mr.v.item()
        return np.max(f)
```

examples/seismic/viscoelastic/viscoelastic_example.py
```diff
@@ -1,14 +1,14 @@
...
 from devito.logger import info
 from examples.seismic.viscoelastic import ViscoelasticWaveSolver
 from examples.seismic import demo_model, AcquisitionGeometry
 
 
-def viscoelastic_setup(shape=(50, 50), spacing=(15.0, 15.0), tn=500.,
-                       space_order=4, nbpml=10,
-                       constant=True, **kwargs):
+def viscoelastic_setup(shape=(50, 50), spacing=(15.0, 15.0), tn=500., space_order=4,
+                       nbpml=10, constant=True, **kwargs):
```
FabioLuporini: @rhodrin @mloubout note that `constant=True` here, while it's `constant=False` in elastic

devito/builtins.py
```diff
@@ -205,6 +207,9 @@ def mmin(f):
         Input operand.
     """
     if not isinstance(f, dv.Function):
+        if dv.configuration['mpi']:
+            warning("returning local min of array,"
```
FabioLuporini: As I wrote on Slack, I would actually be even more restrictive 

```
if not isinstance(f, dv.Function):
    raise ValueError("Expected Function, not `%s`" % type(f))
...
```

that is, I'd raise a ValueError regardless of MPI

examples/seismic/viscoelastic/viscoelastic_example.py
georgebisbas: We talked about that in slack, @rhodrin  did that in purpose.

FabioLuporini:
Merged.

PR #915, opened 2019-08-29
MPI: flip domain decomposition

FabioLuporini:
In master we decompose z -> y -> x
Now we decompose x -> y -> z

that is, the fastest varying dimension is decomposed last.

so if I have a grid `(16, 16, 16)` and 2 MPI ranks, each rank gets
* with master: `(16, 16, 8)`
* with this PR `(8, 16, 16)`

The larger the fastest (innermost) dimension, the better the performance should be


FabioLuporini:
Merged.

PR #918, opened 2019-09-01
2D Visco benchmark fix

rhodrin:
Fix a typo in the 2d viscoelastic benchmark equations. Update the preset models and viscoelastic_example.py accordingly.

### DIFF
--- a/examples/seismic/model.py
+++ b/examples/seismic/model.py
@@ -62,15 +62,15 @@ def demo_model(preset, **kwargs):
         # A constant single-layer model in a 2D or 3D domain
         # with velocity 2.2 km/s.
         shape = kwargs.pop('shape', (101, 101))
-        spacing = kwargs.pop('spacing', tuple([1. for _ in shape]))
+        spacing = kwargs.pop('spacing', tuple([10. for _ in shape]))
         origin = kwargs.pop('origin', tuple([0. for _ in shape]))
         nbpml = kwargs.pop('nbpml', 10)
         dtype = kwargs.pop('dtype', np.float32)
-        vp = kwargs.pop('vp', 1.6)
-        qp = kwargs.pop('qp', 40.)
-        vs = kwargs.pop('vs', 0.4)
-        qs = kwargs.pop('qs', 30.)
-        rho = 1.3
+        vp = kwargs.pop('vp', 2.2)
+        qp = kwargs.pop('qp', 100.)
+        vs = kwargs.pop('vs', 1.2)
+        qs = kwargs.pop('qs', 70.)
+        rho = 2.
 
         return ModelViscoelastic(space_order=space_order, vp=vp, qp=qp, vs=vs,
                                  qs=qs, rho=rho, origin=origin, shape=shape,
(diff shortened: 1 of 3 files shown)


examples/seismic/model.py
```diff
@@ -168,21 +168,21 @@ def demo_model(preset, **kwargs):
...
-        rho_top = kwargs.pop('rho_top', 1.)
-        vp_bottom = kwargs.pop('vp_bottom', 2.5)
-        qp_bottom = kwargs.pop('qp_bottom', 10000.)
-        vs_bottom = kwargs.pop('vs_bottom', 0. * vp_bottom)
-        qs_bottom = kwargs.pop('qs_bottom', 0.)
-        rho_bottom = kwargs.pop('qs_bottom', 2.5/1.5)
+        ratio = kwargs.pop('ratio', 3)
+        vp_top = kwargs.pop('vp_top', 1.6)
+        qp_top = kwargs.pop('qp_top', 40.)
+        vs_top = kwargs.pop('vs_top', 0.4)
```
mloubout: I would prefer this to stay `0` as one of the challenge of elastic and visco-elastic is ocena bottom that goes from acoustic vs=0 to elastic vs>0 and only work with proper staggering and this would show it well)

examples/seismic/model.py
```diff
@@ -168,21 +168,21 @@ def demo_model(preset, **kwargs):
         # By default, the top part of the domain has 1.5 km/s,
         # and the bottom part of the domain has 2.5 km/s.
         shape = kwargs.pop('shape', (101, 101))
-        spacing = kwargs.pop('spacing', tuple([10. for _ in shape]))
+        spacing = kwargs.pop('spacing', tuple([1. for _ in shape]))
```
mloubout: Any reason why?

examples/seismic/model.py
```diff
@@ -168,21 +168,21 @@ def demo_model(preset, **kwargs):
...
-        qs_top = kwargs.pop('qs_top', 0.)
-        rho_top = kwargs.pop('rho_top', 1.)
-        vp_bottom = kwargs.pop('vp_bottom', 2.5)
-        qp_bottom = kwargs.pop('qp_bottom', 10000.)
-        vs_bottom = kwargs.pop('vs_bottom', 0. * vp_bottom)
-        qs_bottom = kwargs.pop('qs_bottom', 0.)
-        rho_bottom = kwargs.pop('qs_bottom', 2.5/1.5)
+        ratio = kwargs.pop('ratio', 3)
+        vp_top = kwargs.pop('vp_top', 1.6)
+        qp_top = kwargs.pop('qp_top', 40.)
```
mloubout: That's quite the change, I thought Q was supposed to be quite high like previously.

examples/seismic/model.py
```diff
@@ -168,21 +168,21 @@ def demo_model(preset, **kwargs):
...
+        vs_top = kwargs.pop('vs_top', 0.4)
+        qs_top = kwargs.pop('qs_top', 30.)
+        rho_top = kwargs.pop('rho_top', 1.3)
+        vp_bottom = kwargs.pop('vp_bottom', 2.2)
+        qp_bottom = kwargs.pop('qp_bottom', 100.)
```
mloubout: same, seems low and unrealistic

examples/seismic/model.py
rhodrin: So that if a user uses a pre-set model the results don't look 'crap'. Any reason for it not be 1?

examples/seismic/model.py
rhodrin: Due to the error in the 2d equations low Q was running into trouble which is why it was ramped up.

examples/seismic/model.py
rhodrin: A water -> soft rock -> harder rock example is covered in the tutorial. These values are for soft rock -> harder rock.

examples/seismic/model.py
rhodrin: There's nothing unrealistic about these values. If we wish to benchmark with a different physical case that's something that can be considered though.

FabioLuporini:
merged.

PR #923, opened 2019-09-04
Update README.md

georgebisbas:


benchmarks/user/README.md
```diff
@@ -66,7 +66,7 @@ working set). There are multiple scenarios that are worth considering:
   parallel region.  Pinning is typically enforced via environment variables.
 
 Some more information about pinning is available
-[here](https://perf.readthedocs.io/en/latest/system.html).
+[here](https://buildmedia.readthedocs.org/media/pdf/perf/latest/perf.pdf).
```
FabioLuporini: While I agree we should replace a broken link, I wonder whether this is really the best we can link to...the doc of another tool

georgebisbas: Yes of course. It was a very lazy reaction by my side.
What about this: https://hpc-wiki.info/hpc/Binding/Pinning

FabioLuporini: yeah, LGTM

FabioLuporini:
Merged, thanks

PR #925, opened 2019-09-05
CustomCompiler should respect environment variables

navjotk:
It seems strange to find this bug so please correct me if I'm missing something. 

CustomCompiler, as it is in master, effectively seems to ignore the `CC` and other environment variables. Even though it reads these values once but they are overwritten by the hardcoded values in the lines I changed below. 

Changed the lines so the values are always read from the environment.


FabioLuporini:
Merged.

PR #926, opened 2019-09-05
Change from using data to data_allocated in checkpointing

navjotk:
This PR:
1. Changes the the c contiguous test in tests/test_data.py to use data_allocated and to check for a space_order > 0 to really test the condition that is likely to fail. 
2. Changes the checkpoint API with pyrevolve to use data_allocated instead of data. 

Note: ~This PR includes #925 so merge that first.~ 
TODO: Remove class attributes like `CC` etc from `CustomCompiler`

### DIFF
--- a/devito/compiler.py
+++ b/devito/compiler.py
@@ -452,11 +452,6 @@ class CustomCompiler(Compiler):
     flags are read from OMP_LDFLAGS or otherwise default to ``-fopenmp``.
     """
 
-    CC = environ.get('CC', 'gcc')
-    CXX = environ.get('CXX', 'g++')
-    MPICC = environ.get('MPICC', 'mpicc')
-    MPICXX = environ.get('MPICXX', 'mpicxx')
-
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


FabioLuporini:
@navjotk I'm happy to land this, but I wanted to ask a question first: can you remind me why you want/need the `*_allocated ` elements, as opposed to just the domain elements ? I'd have expected one checkpoints only the domain region, not the whole array

navjotk:
The performance hit from dealing with non-contiguous memory is so big (introduces >1 copy), it's better to checkpoint stuff you don't need just to get a contiguous chunk of memory.

FabioLuporini:
Ah! This is a very good point indeed. Would you mind adding a comment? 

nitpicking: would you also mind moving `def size_allocated` right after `def size` ?

FabioLuporini:
I will do those nitpicks. Thanks, merging

FabioLuporini:
> nitpicking: would you also mind moving def size_allocated right after def size ?

sry, hadn't noticed `size_allocated` is in dense.py, while `size` is in basic.py

PR #927, opened 2019-09-05
Add support to ops_dat_fetch_data 

vmickus:
**Add ops_dat_fetch_data call to the OPS backend** 

- This OPS API call is responsible to get data from the GPU and returning to the pointer allocated from Devito in the main memory.   
- For all ops_dat created, it is then created an `ops_dat_fetch_dat` call, that will bring the result from the GPU.   
- Modified `Byref` class to use ccode instead of plain `str`.   
- Created new tests for this OPS feature and modified tests for `ops_dat`.  
- Modified `OpsAccessible` to hold reference of what Devito type it was generated from.
- Modified dict `name_to_ops_dat` to only have values that are `OpsDat` or `Array` of `OpsDat`, instead of also having Indexed.

### DIFF
--- a/devito/ops/node_factory.py
+++ b/devito/ops/node_factory.py
@@ -43,7 +43,9 @@ class OPSNodeFactory(object):
             symbol_to_access = OpsAccessible(
                 ops_arg_id,
                 indexed.dtype,
-                not is_write
+                not is_write,
+                indexed.function,
+                time_index.var
             )
             self.ops_args[ops_arg_id] = symbol_to_access
             self.ops_params.append(symbol_to_access)
--- a/devito/ops/transformer.py
+++ b/devito/ops/transformer.py
@@ -1,7 +1,7 @@
-import numpy as np
 import itertools
+import numpy as np
 
-from sympy import sympify
+from sympy import sympify, Mod
 from sympy.core.numbers import Zero
 
 from devito import Eq
@@ -11,7 +11,7 @@ from devito.ir.iet.visitors import FindNodes
 from devito.ops.node_factory import OPSNodeFactory
 from devito.ops.types import Array, OpsAccessible, OpsDat, OpsStencil
 from devito.ops.utils import namespace
-from devito.symbolics import Byref, ListInitializer, Literal
+from devito.symbolics import Add, Byref, ListInitializer, Literal
 from devito.tools import dtype_to_cstr
 from devito.types import Constant, DefaultDimension, Symbol
 
@@ -35,16 +35,19 @@ def opsit(trees, count, name_to_ops_dat, block, dims):
         expressions.extend(FindNodes(Expression).visit(tree.inner))
 
     for expr in expressions:
-        ops_expressions.append(Expression(make_ops_ast(expr.expr, node_factory)))
+        ops_expressions.append(Expression(
+            make_ops_ast(expr.expr, node_factory)))
 
-    parameters = sorted(node_factory.ops_params, key=lambda i: (i.is_Constant, i.name))
+    parameters = sorted(node_factory.ops_params,
+                        key=lambda i: (i.is_Constant, i.name))
 
     stencil_arrays_initializations = []
     par_to_ops_stencil = {}
 
     for p in parameters:
         if isinstance(p, OpsAccessible):
-            stencil, initialization = to_ops_stencil(p, node_factory.ops_args_accesses[p])
+            stencil, initialization = to_ops_stencil(
+                p, node_factory.ops_args_accesses[p])
 
             par_to_ops_stencil[p] = stencil
             stencil_arrays_initializations.append(initialization)
@@ -138,11 +141,13 @@ def create_ops_dat(f, name_to_ops_dat, block):
         padding = f.padding[:time_pos] + f.padding[time_pos + 1:]
         halo = f.halo[:time_pos] + f.halo[time_pos + 1:]
         d_p_val = tuple(sympify([p[0] + h[0] for p, h in zip(padding, halo)]))
-        d_m_val = tuple(sympify([-(p[1] + h[1]) for p, h in zip(padding, halo)]))
+        d_m_val = tuple(sympify([-(p[1] + h[1])
+                                 for p, h in zip(padding, halo)]))
 
         ops_dat_array = Array(
             name=namespace['ops_dat_name'](f.name),
-            dimensions=(DefaultDimension(name='dat', default_value=time_dims),),
+            dimensions=(DefaultDimension(
+                name='dat', default_value=time_dims),),
             dtype='ops_dat',
             scope='stack'
         )
@@ -150,9 +155,7 @@ def create_ops_dat(f, name_to_ops_dat, block):
         dat_decls = []
         for i in range(time_dims):
             name = '%s%s%s' % (f.name, time_index, i)
-            name_to_ops_dat[name] = ops_dat_array.indexify(
-                [Symbol('%s%s' % (time_index, i))]
-            )
+
             dat_decls.append(namespace['ops_decl_dat'](
                 block,
                 1,
@@ -160,7 +163,7 @@ def create_ops_dat(f, name_to_ops_dat, block):
                 Symbol(base.name),
                 Symbol(d_m.name),
                 Symbol(d_p.name),
-                Byref(f.indexify([i])),
+                f.indexify([i]),
                 Literal('"%s"' % f._C_typedata),
                 Literal('"%s"' % name)
             ))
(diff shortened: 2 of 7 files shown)


tests/test_ops.py
```diff
@@ -256,3 +225,24 @@ def test_create_ops_arg_function(self, read):
...
+         '[\'ops_dat_fetch_data(v_dat[(time_M)%(2)],0,&(v[(time_M)%(2)]));\','
+         '\'ops_dat_fetch_data(v_dat[(time_M + 1)%(2)],0,&(v[(time_M + 1)%(2)]));\','
+         '\'ops_dat_fetch_data(u_dat[(time_M)%(2)],0,&(u[(time_M)%(2)]));\','
+         '\'ops_dat_fetch_data(u_dat[(time_M + 1)%(2)],0,&(u[(time_M + 1)%(2)]));\']'),
+    ])
+    def test_create_fetch_data(self, equation, expected):
+
+        grid = Grid(shape=(4, 4))
+
+        u = TimeFunction(name='u', grid=grid)  # noqa
```
FabioLuporini: May I suggest to try also with:

- more time_orders, ie `TimeFunction(..., time_order=x)`
- equations using time derivatives such as iso acoustic (`Eq(u.forward, u.dt ....)`)

tests/test_ops.py
```diff
@@ -131,60 +131,29 @@ def test_to_ops_stencil(self, _accesses):
...
+         'u[1], "float", "ut1")}\']'),
+        ('Eq(u.forward, u + v.dx)',
+         '[\'ops_dat u_dat[2] = {ops_decl_dat(block, 1, u_dim, u_base, u_d_m, u_d_p, '
+         'u[0], "float", "ut0"), ops_decl_dat(block, 1, u_dim, u_base, u_d_m, u_d_p, '
+         'u[1], "float", "ut1")}\','
+         '\'ops_dat v_dat;\','
+         '\'v_dat = ops_decl_dat(block, 1, v_dim, v_base, v_d_m, v_d_p, '
+         'v[0], "float", "v")\']')
+    ])
+    def test_create_ops_dat(self, equation, expected):
```
FabioLuporini: this test seems much thinner than before?

devito/ops/types.py
```diff
@@ -31,13 +31,16 @@ class OpsAccessible(basic.Symbol):
     """
     is_Scalar = True
 
-    def __new__(cls, name, dtype, read_only, *args, **kwargs):
+    def __new__(cls, name, dtype, read_only, origin, origin_time_index, * args, **kwargs):
         obj = basic.Symbol.__new__(cls, name, dtype, *args, **kwargs)
-        obj.__init__(name, dtype, read_only, *args, **kwargs)
+        obj.__init__(name, dtype, read_only, origin, origin_time_index, *args, **kwargs)
```
FabioLuporini: Im not a big fan of adding this sort of information to a Symbol....it really feels odd. Can't the caller just retain this info, perhaps in a dict?

devito/ops/transformer.py
```diff
@@ -186,7 +195,7 @@ def create_ops_dat(f, name_to_ops_dat, block):
                 Symbol(base.name),
                 Symbol(d_m.name),
                 Symbol(d_p.name),
-                Byref(f.indexify([0])),
+                f.indexify([0]),
```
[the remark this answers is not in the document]
vmickus: The `ops_decl_dat` is documented in OPS API as:
```
ops_dat ops_decl_dat(ops_block block, int dim, int* size, int *base,int *dm, int *dp, T *data, char *type, char *name) 
...
data     ->     input data of typeT
...
```

So, I need a pointer to the data. 
This line represents a `Function` access in Devito and it is translated into an array in C. 

Actually, after i removed the `Byref` I should have dropped the indexify, and just pass the `f`.

I hope that this will be translated into an access to the function (for example if the function was declared as `v = Function(name='v', grid=grid, space_order=2)`, I want this part to just generate `v` in C or, if this is not possible, generate as it was being done `&(v[0])`).

devito/ops/transformer.py
```diff
@@ -169,12 +172,18 @@ def create_ops_dat(f, name_to_ops_dat, block):
...
+        # Inserting the ops_dat array in case of TimeFunction.
+        name_to_ops_dat[f.name] = ops_dat_array
+
     else:
         ops_dat = OpsDat("%s_dat" % f.name)
         name_to_ops_dat[f.name] = ops_dat
 
-        d_p_val = tuple(sympify([p[0] + h[0] for p, h in zip(f.padding, f.halo)]))
-        d_m_val = tuple(sympify([-(p[1] + h[1]) for p, h in zip(f.padding, f.halo)]))
+        d_p_val = tuple(sympify([p[0] + h[0]
```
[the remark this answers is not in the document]
vmickus: Actually I also don't know.
My guess is that the padding and halo attribute are both python `int` at this point, but we need this information to be translated back to `Devito` API so it can use it to generate nodes that will declare the `d_p` and `d_m` variables.

For example, it need to generate something like this in C:
`int v_d_p[2] = {2, 2};`

But @vincepandolfo can explain better than me this part.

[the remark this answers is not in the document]
vmickus: Also, this part shouldn't be changed in my commit. 
It was just the format but the previous format was already good. I think my IDE changed it and I didn't realize.

tests/test_ops.py
vmickus: Sure! I'll put some new tests.

tests/test_ops.py
vmickus: Although it is thinner, it is actually testing more things than before.
The first test in the `@pytest.mark.parametrize` cover the case that was being tested, and I added the second one to test a mix of `TimeFunction` and `Function`.

devito/ops/types.py
vmickus: That's a good idea. 
I try the dictionary approach.

devito/ops/transformer.py
```diff
@@ -138,29 +141,29 @@ def create_ops_dat(f, name_to_ops_dat, block):
         padding = f.padding[:time_pos] + f.padding[time_pos + 1:]
         halo = f.halo[:time_pos] + f.halo[time_pos + 1:]
         d_p_val = tuple(sympify([p[0] + h[0] for p, h in zip(padding, halo)]))
```
FabioLuporini: I **think** you wanna use `_size_halo` and in particular `_size_halo.left` and `_size_halo.right` instead of accessing `halo` and  `padding` directly, with explicit integer indexing (which is somewhat like having magic numbers in the code)

devito/ops/transformer.py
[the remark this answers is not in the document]
FabioLuporini: I explained it in the comment above . Let me know if you got it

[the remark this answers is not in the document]
vincepandolfo: We need `simpify` as otherwise `simpy` will complain later. I do not remember the details, though. Should be easy enough to try it out

devito/ops/transformer.py
vmickus: I managed to replace the `halo` and `padding` variables with the `_size_halo` that you said. 

But, `_size_halo` also bring the time dimension halo (although it is 0 in my tests), and in this part we are only interested in the space dimensions, so the 3 lines above got replace by this:

`d_p_val = sympify(f._size_halo.left[time_pos+1:])` 

I didn't manage to get rid of the `sympify` at this point. When I delete it, I got the following error: 
`AttributeError: 'int' object has no attribute 'is_Indexed'` 

So we need the sympfy to transform the int into Sympy objects again.

devito/ops/types.py
vmickus: I tried creating a dictionary approach... but it didn't feel right.

I think this `OpsAccessible` is much like the `Indexed` class, as it also needs information of what it is accessing (in this case TimeFunction or Function). But, to not store all the data from `TimeFunction`, I'll try to just use the necessary information from them. 

I'll commit and then you can see if this is acceptable.

devito/ops/transformer.py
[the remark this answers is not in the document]
FabioLuporini: Do not use magic numbers 0/1 -- use `_size_halo.left` for example

devito/ops/transformer.py
```diff
@@ -201,8 +207,24 @@ def create_ops_dat(f, name_to_ops_dat, block):
...
+    if f.is_TimeFunction:
+        ops_fetch = [namespace['ops_dat_fetch_data'](
+            name_to_ops_dat[f.name].indexify(
+                [Mod(Add(time_upper_bound, -i), f._time_size)]),
+            Byref(f.indexify([Mod(Add(time_upper_bound, -i), f._time_size)])))
+            for i in range(f._time_size)]
+
+    else:
+        ops_fetch = [namespace['ops_dat_fetch_data'](
+            name_to_ops_dat[f.name], Byref(f.indexify([0])))]
```
FabioLuporini: 0?

devito/ops/node_factory.py
```diff
@@ -45,6 +45,10 @@ def new_ops_arg(self, indexed, is_write):
                 indexed.dtype,
                 not is_write
             )
+
+            symbol_to_access._origin_name = indexed.function.name
```
FabioLuporini: this is wrong. You should never ever change state. The only thing you can do is to supply the arguments at construction time via ``__init__``

What you're doing here breaks immutability

tests/test_ops.py
FabioLuporini: oki

devito/ops/types.py
FabioLuporini: OK, I see

devito/ops/transformer.py
FabioLuporini: > But, _size_halo also bring the time dimension halo (although it is 0 in my tests), and in this part we are only interested in the space dimensions,

does
```
padding = [f._size_padding[d] for d in f.space_dimensions]
```
help?
Analogous for halo.
It should give you what you need while retaining simplicity and clarity

devito/ops/node_factory.py
vmickus: Fixed.

devito/ops/transformer.py
vmickus: As explained, this is an hack. 

I needed to access a pointer to the beginning of an array. But I couldn't generate just `v`, instead, I managed to generate `&(v[0])`.
As you have suggested, I added a comment here to explain this.

PR #929, opened 2019-09-09
Fix handling of mixed anti- and output-dependences during clustering

FabioLuporini:
Also adds tests


FabioLuporini:
@CarlosCuetoM

PR #931, opened 2019-09-09
MPI fixes + smoother builtin

rhodrin:
This PR:
1) Includes some bug fixes associated with MPI data slicing along with the relevant tests.
2) A gaussian smoother builtin for Devito functions that mimics the `scipy.ndimage` `gaussian_filter`.

Note: Some changes and tidying are still required.

### DIFF
--- a/devito/builtins.py
+++ b/devito/builtins.py
@@ -234,6 +234,9 @@ def initialize_function(function, data, nbl, mapper=None, mode='constant',
           [2, 3, 3, 3, 3, 2],
           [2, 2, 2, 2, 2, 2]], dtype=int32)
     """
+    if isinstance(function, dv.TimeFunction):
+        raise NotImplementedError("TimeFunctions are not currently supported.")
+
     slices = tuple([slice(nbl, -nbl) for _ in range(function.grid.dim)])
     if isinstance(data, dv.Function):
         function.data[slices] = data.data[:]
@@ -255,7 +258,7 @@ def initialize_function(function, data, nbl, mapper=None, mode='constant',
         if any(np.array(b) < 0):
             raise ValueError("Function `%s` halo is not sufficiently thick." % function)
 
-    for d in function.dimensions:
+    for d in function.space_dimensions:
         dim_l = dv.SubDimension.left(name='abc_%s_l' % d.name, parent=d,
                                      thickness=nbl)
         dim_r = dv.SubDimension.right(name='abc_%s_r' % d.name, parent=d,
--- a/devito/builtins.py
+++ b/devito/builtins.py
@@ -29,9 +29,29 @@ def assign(f, RHS=0, options=None, name='assign', **kwargs):
 
     Examples
     --------
-    >>>
-    >>>
-    >>>
+    >>> from devito import Grid, Function, assign
+    >>> grid = Grid(shape=(4, 4))
+    >>> f = Function(name='f', grid=grid, dtype=np.int32)
+    >>> g = Function(name='g', grid=grid, dtype=np.int32)
+    >>> h = Function(name='h', grid=grid, dtype=np.int32)
+    >>> functions = [f, g, h]
+    >>> scalars = [1, 2, 3]
+    >>> assign(functions, scalars)
+    >>> f.data
+    Data([[1, 1, 1, 1],
+          [1, 1, 1, 1],
+          [1, 1, 1, 1],
+          [1, 1, 1, 1]], dtype=int32)
+    >>> g.data
+    Data([[2, 2, 2, 2],
+          [2, 2, 2, 2],
+          [2, 2, 2, 2],
+          [2, 2, 2, 2]], dtype=int32)
+    >>> h.data
+    Data([[3, 3, 3, 3],
+          [3, 3, 3, 3],
+          [3, 3, 3, 3],
+          [3, 3, 3, 3]], dtype=int32)
     """
     if not isinstance(f, list):
         f = [f]
@@ -151,13 +171,12 @@ def gaussian_smooth(f, sigma=1, _order=4, mode='reflect'):
     return f
 
 
-def initialize_function(function, data, nbpml, additional_expressions=dict(),
+def initialize_function(function, data, nbl, additional_expressions=dict(),
                         mode='constant', name='padfunc'):
     """
-    # TODO: Generalise some of the below descriptions. Change nbpml.Finish docstring.
     Initialize a `Function` with the given ``data``. ``data``
-    does *not* include the PML layers for the absorbing boundary conditions;
-    these are added via padding by this function.
+    does *not* include the ``nbl`` outer/boundary layers; these are added via padding
+    by this function.
 
     Parameters
     ----------
@@ -165,17 +184,22 @@ def initialize_function(function, data, nbpml, additional_expressions=dict(),
         The initialised object.
     data : ndarray of Function
         The data used for initialisation.
-    nbpml : int
-        Number of PML layers for boundary damping.
+    nbl : int
+        Number of outer layers (such as PML layers for boundary damping).
     additional_expressions : dict, optional
-        Dictionary containing ...
+        Dictionary containing, for each dimension of function, a sub-dictionary
+        containing the following keys:
+        1) 'lhs': List of additional expressions to be added to the LHS expressions list.
+        2) 'rhs': List of additional expressions to be added to the RHS expressions list.
+        3) 'options': Options pertaining to the additional equations that will be
+        constructed.
     mode : str, optional
         The function initialisation mode. 'constant' and 'reflect' are
         accepted.
     name : str, optional
         The name assigned to the operator.
     """
-    slices = tuple([slice(nbpml, -nbpml) for _ in range(function.grid.dim)])
+    slices = tuple([slice(nbl, -nbl) for _ in range(function.grid.dim)])
     if isinstance(data, dv.Function):
         function.data[slices] = data.data[:]
     else:
@@ -184,17 +208,29 @@ def initialize_function(function, data, nbpml, additional_expressions=dict(),
     rhs = []
     options = []
 
+    if mode == 'reflect' and function.grid.distributor.is_parallel:
+        # Check that HALO size is appropriate
+        halo = function.halo
+        local_size = function.grid.distributor.shape
+
+        def buff(i, j):
+            return [(i + k - 2*nbl) for k in j]
+
+        b = [min(l) for l in (w for w in (buff(i, j) for i, j in zip(local_size, halo)))]
+        if any(np.array(b) < 0):
+            raise ValueError("Function's halo is not sufficiently thick.")
+
     for d in function.dimensions:
         dim_l = dv.SubDimension.left(name='abc_%s_l' % d.name, parent=d,
-                                     thickness=nbpml)
+                                     thickness=nbl)
         dim_r = dv.SubDimension.right(name='abc_%s_r' % d.name, parent=d,
-                                      thickness=nbpml)
+                                      thickness=nbl)
         if mode == 'constant':
-            subsl = nbpml
-            subsr = d.symbolic_max - nbpml
+            subsl = nbl
+            subsr = d.symbolic_max - nbl
         elif mode == 'reflect':
-            subsl = 2*nbpml - 1 - dim_l
-            subsr = 2*(d.symbolic_max - nbpml) + 1 - dim_r
+            subsl = 2*nbl - 1 - dim_l
+            subsr = 2*(d.symbolic_max - nbl) + 1 - dim_r
         else:
             raise ValueError("Mode not available")
         lhs.append(function.subs({d: dim_l}))
--- a/devito/builtins.py
+++ b/devito/builtins.py
@@ -11,18 +11,36 @@ __all__ = ['assign', 'smooth', 'gaussian_smooth', 'initialize_function', 'norm',
            'sumall', 'inner', 'mmin', 'mmax']
 
 
-def assign(f, v=0):
+def assign(f, RHS=0, options=None, name='asign', **kwargs):
     """
-    Assign a value to a Function.
+    Assign a list of RHS's to a list of Function's.
 
     Parameters
     ----------
-    f : Function
+    f : Function or list of Function's
         The left-hand side of the assignment.
-    v : scalar, optional
+    RHS : expression or list of expression's, optional
         The right-hand side of the assignment.
+    options : list, optional
+        ...
+    name : str, optional
+        Name of the operator.
     """
-    dv.Operator(dv.Eq(f, v), name='assign')()
+    if not isinstance(f, list):
+        f = list(f)
+    if not isinstance(RHS, list):
+        RHS = [RHS, ]*len(f)
+    eqs = []
+    if bool(options):
+        for i, j, k in zip(f, RHS, options):
+            if k is not None:
+                eqs.append(dv.Eq(i, j, **k))
+            else:
+                eqs.append(dv.Eq(i, j))
+    else:
+        for i, j in zip(f, RHS):
+            eqs.append(dv.Eq(i, j))
+    dv.Operator(eqs, name=name, **kwargs)()
 
 
 def smooth(f, g, axis=None):
@@ -69,36 +87,6 @@ def gaussian_smooth(f, sigma=1, _order=4, mode='reflect'):
         def define(self, dimensions):
             return {d: ('middle', self.lw, self.lw) for d in dimensions}
 
-    def smooth(fo, fi, data):
-        # TODO: Merge part of this with part of initialize_function and assign
-        slices = tuple([slice(lw, -lw) for _ in range(fi.grid.dim)])
-        fi.data[slices] = data
-        eqs = []
-        for d in fi.dimensions:
-            dim_l = dv.SubDimension.left(name='abc_%s_l' % d.name, parent=d,
-                                         thickness=lw)
-            dim_r = dv.SubDimension.right(name='abc_%s_r' % d.name, parent=d,
-                                          thickness=lw)
-            if mode == 'constant':
-                subsl = lw
-                subsr = d.symbolic_max - lw
-            elif mode == 'reflect':
-                subsl = 2*lw - 1 - dim_l
-                subsr = 2*(d.symbolic_max - lw) + 1 - dim_r
-            else:
-                raise ValueError("Mode not available")
-            eqs += [dv.Eq(fi.subs({d: dim_l}), fi.subs({d: subsl}))]
-            eqs += [dv.Eq(fi.subs({d: dim_r}), fi.subs({d: subsr}))]
-
-            rhs = dv.generic_derivative(fi, d, 2*lw, 1)
-            coeffs = dv.Coefficient(1, fi, d, weights)
-            eqs += [dv.Eq(fo, rhs, coefficients=dv.Substitutions(coeffs),
-                          subdomain=grid.subdomains['objective_domain'])]
-
-            eqs += [dv.Eq(fi, fo, subdomain=grid.subdomains['objective_domain'])]
-
-        dv.Operator(eqs)()
-
     def fset(f, g):
         indices = [slice(lw, -lw, 1) for _ in g.grid.dimensions]
         slices = (slice(None, None, 1), )*len(g.grid.dimensions)
@@ -126,17 +114,43 @@ def gaussian_smooth(f, sigma=1, _order=4, mode='reflect'):
     weights = np.exp(-0.5/sigma**2*(np.linspace(-lw, lw, 2*lw+1))**2)
     weights = weights/weights.sum()
 
-    if isinstance(f, dv.Function):
-        smooth(f_o, f_c, f.data[:])
-    else:
-        smooth(f_o, f_c, f)
+    additional_expressions = {}
+    for d in f_c.dimensions:
+        expr = {}
+        lhs = []
+        rhs = []
+        options = []
 
-    fset(f, f_o)
+        lhs.append(f_o)
+        rhs.append(dv.generic_derivative(f_c, d, 2*lw, 1))
+        coeffs = dv.Coefficient(1, f_c, d, weights)
+        options.append({'coefficients': dv.Substitutions(coeffs),
+                        'subdomain': grid.subdomains['objective_domain']})
+        lhs.append(f_c)
+        rhs.append(f_o)
+        options.append({'subdomain': grid.subdomains['objective_domain']})
+
+        expr['lhs'] = lhs
+        expr['rhs'] = rhs
+        expr['options'] = options
+        additional_expressions[d] = expr
+
+    if isinstance(f, dv.Function):
+        initialize_function(f_c, f.data[:], lw,
+                            additional_expressions=additional_expressions,
+                            mode='reflect', name='smooth')
+    else:
+        initialize_function(f_c, f, lw,
+                            additional_expressions=additional_expressions,
+                            mode='reflect', name='smooth')
+
+    fset(f, f_c)
 
     return f
 
 
-def initialize_function(function, data, nbpml, mode='constant'):
+def initialize_function(function, data, nbpml, additional_expressions=dict(),
+                        mode='constant', name='padfunc'):
     """
     Initialize a `Function` with the given ``data``. ``data``
     does *not* include the PML layers for the absorbing boundary conditions;
@@ -156,7 +170,9 @@ def initialize_function(function, data, nbpml, mode='constant'):
     """
     slices = tuple([slice(nbpml, -nbpml) for _ in range(function.grid.dim)])
     function.data[slices] = data
-    eqs = []
+    lhs = []
+    rhs = []
+    options = []
 
     for d in function.dimensions:
         dim_l = dv.SubDimension.left(name='abc_%s_l' % d.name, parent=d,
@@ -171,11 +187,31 @@ def initialize_function(function, data, nbpml, mode='constant'):
             subsr = 2*(d.symbolic_max - nbpml) + 1 - dim_r
         else:
             raise ValueError("Mode not available")
-        eqs += [dv.Eq(function.subs({d: dim_l}), function.subs({d: subsl}))]
-        eqs += [dv.Eq(function.subs({d: dim_r}), function.subs({d: subsr}))]
+        lhs.append(function.subs({d: dim_l}))
+        lhs.append(function.subs({d: dim_r}))
+        rhs.append(function.subs({d: subsl}))
+        rhs.append(function.subs({d: subsr}))
+        options.extend([None, None])
+
+        if bool(additional_expressions):
+            exprs = additional_expressions[d]
+            lhs_extra = exprs['lhs']
+            rhs_extra = exprs['rhs']
+            lhs.extend(lhs_extra)
+            rhs.extend(rhs_extra)
+            if 'options' in exprs.keys():
+                options_extra = exprs['options']
+            else:
+                options_extra = len(lhs_extra)*[None, ]
+            options.extend(options_extra)
+
+    if all(options is None for i in options):
+        options = None
 
     # TODO: Figure out why yask doesn't like it with dse/dle
-    dv.Operator(eqs, name='padfunc', dse='noop', dle='noop')()
+    # NOTE: Need to sort out the "dse='noop', dle='noop'" issue
+    assign(lhs, rhs, options=options, name=name)
+    #assign(lhs, rhs, options=options, name=name, dse='noop', dle='noop')
 
 
 # Reduction-inducing builtins
--- a/devito/builtins.py
+++ b/devito/builtins.py
@@ -6,20 +6,21 @@ from sympy import Abs, Pow
 import numpy as np
 
 import devito as dv
+from devito.tools import as_list
 
 __all__ = ['assign', 'smooth', 'gaussian_smooth', 'initialize_function', 'norm',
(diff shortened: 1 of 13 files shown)


devito/builtins.py
```diff
@@ -53,6 +54,135 @@ def smooth(f, g, axis=None):
...
+            super(DataDomain, self).__init__()
+            self.lw = lw
+
+        def define(self, dimensions):
+            return {d: ('middle', self.lw, self.lw) for d in dimensions}
+
+    def fill_data(g, f, dim, lw, mode):
+
+        sl = slice(lw, -lw, 1)
+        indices = ()
```
FabioLuporini: any time you find yourself adding items to a collections, use list/set (these are mutable) and the relevant methods append/add

tuples are immutable and with `+= ` you're merely creating a new object

the effect is the same, but the latter is conceptually wrong

what you could do, once you're done populating your collection, you could cast it to tuple (or frozenset) before returning to the caller

This applies here and pretty much everywhere below

devito/builtins.py
```diff
@@ -53,6 +54,131 @@ def smooth(f, g, axis=None):
         dv.Operator(dv.Eq(f, g.avg(dims=axis)), name='smoother')()
 
 
+def gaussian_smooth(f, sigma=1, _order=4, mode='reflect'):
+    """
+    Gaussian smooth function.
+    """
+    class DataDomain(dv.SubDomain):
```
FabioLuporini: Is it worth moving this inside Devito?
BTW, I wouldn't call it "DataDomain" -- all subdomains are data domains after all ?

rhodrin: > Is it worth moving this inside Devito?
Not at the moment I don't think - if we're going to have some 'pre-set' subdomain methods it's probably better we introduce them after a little more thought. Or is that not what you meant?

>BTW, I wouldn't call it "DataDomain" -- all subdomains are data domains after all ?
Yeah, sure. I think `ObjectiveDomain` is a better name.

devito/builtins.py
```diff
@@ -53,6 +54,131 @@ def smooth(f, g, axis=None):
...
+        name = 'datadomain'
+
+        def __init__(self, lw):
+            super(DataDomain, self).__init__()
+            self.lw = lw
+
+        def define(self, dimensions):
+            return {d: ('middle', self.lw, self.lw) for d in dimensions}
+
+    def fill_data(g, f, dim, lw, mode):
```
mloubout: isn't this basically `model/initialize_function`? Would a be lot shorter/more readable to reuse it IMO

devito/builtins.py
```diff
@@ -53,6 +54,131 @@ def smooth(f, g, axis=None):
...
+                g.data[tuple(indgl)] = f.data[tuple(indfl)]
+                g.data[tuple(indgr)] = f.data[tuple(indfr)]
+            else:
+                raise NotImplementedError
+        else:
+            raise ValueError("Mode not available")
+
+        return
+
+    def subfloor(f, g):
```
mloubout: Probably don't need this with the right sizes + Operator rather than python fill_data

devito/builtins.py
```diff
@@ -53,6 +54,131 @@ def smooth(f, g, axis=None):
...
+    dims = grid.dimensions
+
+    f_c = dv.Function(name='f_c', grid=grid, space_order=2*lw,
+                      coefficients='symbolic', dtype=np.int32)
+    f_o = dv.Function(name='f_o', grid=grid, coefficients='symbolic', dtype=np.int32)
+
+    weights = np.exp(-0.5/sigma**2*(np.linspace(-lw, lw, 2*lw+1))**2)
+    weights = weights/weights.sum()
+
+    for d in dims:
```
mloubout: Can make it in one go with a 3D kernel from outer product of 1D kernel and avoid the loop as 3D smooth is
W[i,j,k] = w[i]*w[j]*w[k] and need one 3d conv rather than multiple 1D ones with padding.

devito/builtins.py
rhodrin: Will take a look, but I would have thought `model/initialize_function` can't do the `'reflect'` conditions?

devito/builtins.py
rhodrin: Yeah, probably, will test.

devito/builtins.py
mloubout: I didn't put it because didn't need it but it's just a small change in the rhs of the equation. Currently is `Eq(u(d), u(npml)` can just do `Eq(u(d), u(nbpm+d)` or something like that and you got the reflect

devito/builtins.py
rhodrin: Sure, but then the stencil is a square/cube (2d/3d). I think the symbolic coefficients stuff may need an update to get that to work, will double check. (It should of course be done that way, and will be at some point, but I don't think it'll work immediately which is why it's currently like this).

devito/builtins.py
rhodrin: Oki, will check it out.

devito/builtins.py
mloubout: Yes but you can just do `sum w[i]*f.subs(...)` rather than symbolic coeefs and do just all dims at once

devito/builtins.py
FabioLuporini: is it worth promoting initialize_function to a builtin ?

devito/data/utils.py
```diff
@@ -331,3 +309,75 @@ def mpi_index_maps(loc_idx, shape, topology, coords, comm):
         local_si[index] = as_tuple(rnorm_index)
         it.iternext()
     return owners, send, global_si, local_si
+
+
+def flip_idx(idx, decomposition):
```
FabioLuporini: I know this is WIP, but before landing I'd like docstring + examples here (some for the other util below), as these are fairly "delicate" functions

devito/data/utils.py
```diff
@@ -331,3 +309,75 @@ def mpi_index_maps(loc_idx, shape, topology, coords, comm):
...
+                stop = i.stop + j.glb_max + 1
+            else:
+                stop = i.stop
+            processed.append(slice(start, stop, i.step))
+        else:
+            processed.append(i)
+    return as_tuple(processed)
+
+
+def distributed_data_size(shape, coords, topology, nprocs):
```
FabioLuporini: just to be sure: does this name mean "compute_distributed_data_size" ?

devito/data/utils.py
rhodrin: Sure, will add.

devito/data/utils.py
rhodrin: It's to compute the 'cumulative distrusted data size' (could maybe call it cum_dist_dat_size? Anyway, alternate name suggestions are welcome!). Quick example:
Rank 0-3 dat shapes are `(2, 2)` and grid is a square.
This function will return:
```
[[(2, 2), (2, 4)],
 [(4, 2), (4, 4)]]
```

devito/builtins.py
```diff
@@ -53,6 +54,135 @@ def smooth(f, g, axis=None):
...
+    def smooth(fo, fi, data):
+        slices = tuple([slice(lw, -lw) for _ in range(fi.grid.dim)])
+        fi.data[slices] = data
+        eqs = []
+        for d in fi.dimensions:
+            dim_l = dv.SubDimension.left(name='abc_%s_l' % d.name, parent=d,
+                                         thickness=lw)
+            dim_r = dv.SubDimension.right(name='abc_%s_r' % d.name, parent=d,
+                                          thickness=lw)
+            if mode == 'constant':
```
FabioLuporini: aside from the `to_copy` variables, the construction of the `eqs` is pretty much identical in the two cases (constant, reflect). So we can probably trim down the bodies of this if-else ?

devito/builtins.py
```diff
@@ -53,6 +54,135 @@ def smooth(f, g, axis=None):
...
+        smooth(f_o, f_c, f.data[:])
+    else:
+        smooth(f_o, f_c, f)
+
+    fset(f, f_o)
+
+    return f
+
+
+def initialize_function(function, data, nbpml, mode='constant'):
```
FabioLuporini: Maybe this should be squashed with the `assign` builtin so as to create something unique and sufficiently general for all our needs ? otherwise, there's an overlap between these builtins...although their names are totally different.

devito/data/utils.py
FabioLuporini: cool thanks. Remind me: is this docstring tested via CI?

devito/data/utils.py
```diff
@@ -331,3 +309,133 @@ def mpi_index_maps(loc_idx, shape, topology, coords, comm):
...
+def flip_idx(idx, decomposition):
+    """
+    This function serves two purposes:
+    1) To Convert a global index with containing a slice with step < 0 to a 'mirrored'
+       index with all slice steps > 0.
+    2) Normalize indices with slices containing negative start/stops.
+
+    Parameters
+    ----------
+    idx: Tuple of slices/ints/tuples
```
FabioLuporini: numpydoc nitpicking:

```
idx : tuple of slices/ints/tuples
     ...
```

devito/data/utils.py
```diff
@@ -331,3 +309,133 @@ def mpi_index_maps(loc_idx, shape, topology, coords, comm):
...
+        else:
+            processed.append(i)
+    return as_tuple(processed)
+
+
+def distributed_data_size(shape, coords, topology):
+    """
+    Compute the cumulative shape of the distributed data (cshape).
+
+    coords: Array containing the local shape of data to each rank.
```
FabioLuporini: use numpydoc?

```
Parameters
-----------
...
```

devito/data/utils.py
```diff
@@ -331,3 +309,133 @@ def mpi_index_maps(loc_idx, shape, topology, coords, comm):
...
+
+    send: Array of rank coordinates corresponding to coords.
+
+    topology: Topology of the associated distributor.
+
+    Examples
+    --------
+    Given a set of distributed data such that:
```
FabioLuporini: it's pretty much all there already...maybe we can turn this into a testable example, with `>>>` .. ?
if I'm asking too much, ignore me please

devito/types/grid.py
```diff
@@ -260,14 +259,15 @@ def _make_stepping_dim(self, time_dim, name=None):
...
+    @memoized_meth
     def _arg_defaults(self):
         """A map of default argument values defined by this Grid."""
         args = ReducerMap()
 
         for k, v in self.dimension_map.items():
             args.update(k._arg_defaults(_min=0, size=v.loc))
 
-        if configuration['mpi']:
+        if self.distributor.is_parallel:
```
FabioLuporini: I didn't even remember I had added this property, lol

devito/builtins.py
rhodrin: Yes, can do.

devito/data/utils.py
rhodrin: Yeah, still need to fix a few docstrings.

devito/types/grid.py
rhodrin: :) Glad you did!

devito/builtins.py
rhodrin: Yeah, should probably do something along those lines.

devito/data/utils.py
rhodrin: Yes, it is.

devito/data/utils.py
rhodrin: Could do but there's just a lot of things to import and create so it gets a bit messy. I'd prefer to leave as is but can add if you'd really like.

FabioLuporini: no it's OK, I get it, no worries. This is however heavily tested indirectly, so that's fine

devito/types/dense.py
```diff
@@ -734,7 +734,6 @@ def _arg_names(self):
         """Tuple of argument names introduced by this function."""
         return (self.name,)
 
-    @memoized_meth
```
mloubout: why?

rhodrin: With MPI enabled it results in halo updates not always occurring when they need to. But some other stuff has been memoized now though so that the performance when calling an operator repeatedly it not affected too much.

FabioLuporini: Do we have a test for this? Im sure in one year time I’ll try to add it back because Ill have forgotten about this MPI detail. Also a comment “Note: this, unlike all other _arg_defaults, cant be memoized because ...” would help

rhodrin: Good idea. Will add the test + note.

devito/builtins.py
```diff
@@ -53,6 +54,130 @@ def smooth(f, g, axis=None):
         dv.Operator(dv.Eq(f, g.avg(dims=axis)), name='smoother')()
 
 
+def gaussian_smooth(f, sigma=1, _order=4, mode='reflect'):
+    """
+    Gaussian smooth function.
+    """
+    class ObjectiveDomain(dv.SubDomain):
```
FabioLuporini: Remind me again why this can't be (yet?) moved to `grid.py` along with DOMAIN and INTERIOR ?

rhodrin: Hmm, it possibly can now that the 'reflection' bc is being dealt with by an `Operator`. Will check.

FabioLuporini:
just noticed that basically this is a single commit with loads of changes in it 😬

rhodrin:
> just noticed that basically this is a single commit with loads of changes in it
Oh. I've messed up a rebase somewhere. Will take a look.

devito/builtins.py
```diff
@@ -7,21 +7,40 @@
 
 import devito as dv
 
-__all__ = ['assign', 'smooth', 'norm', 'sumall', 'inner', 'mmin', 'mmax']
+__all__ = ['assign', 'smooth', 'gaussian_smooth', 'initialize_function', 'norm',
+           'sumall', 'inner', 'mmin', 'mmax']
 
 
-def assign(f, v=0):
+def assign(f, RHS=0, options=None, name='asign', **kwargs):
```
FabioLuporini: typo (asign)

devito/builtins.py
```diff
@@ -7,21 +7,40 @@
...
     """
-    Assign a value to a Function.
+    Assign a list of RHS's to a list of Function's.
```
FabioLuporini: nitpicking: we don't use apostrophes for plurals such as Functions

devito/builtins.py
```diff
@@ -7,21 +7,40 @@
...
 
     Parameters
     ----------
-    f : Function
+    f : Function or list of Function's
         The left-hand side of the assignment.
-    v : scalar, optional
+    RHS : expression or list of expression's, optional
```
FabioLuporini: expression -> expr-like

devito/builtins.py
```diff
@@ -7,21 +7,40 @@
...
         The right-hand side of the assignment.
+    options : list, optional
+        ...
+    name : str, optional
+        Name of the operator.
     """
-    dv.Operator(dv.Eq(f, v), name='assign')()
+    if not isinstance(f, list):
```
FabioLuporini: does it work with 
```
f = as_tuple(f) 
```
?

devito/builtins.py
FabioLuporini: perhaps it's now worth adding a couple examples?

devito/builtins.py
```diff
@@ -53,6 +72,148 @@ def smooth(f, g, axis=None):
...
+            subsr = 2*(d.symbolic_max - nbpml) + 1 - dim_r
+        else:
+            raise ValueError("Mode not available")
+        lhs.append(function.subs({d: dim_l}))
+        lhs.append(function.subs({d: dim_r}))
+        rhs.append(function.subs({d: subsl}))
+        rhs.append(function.subs({d: subsr}))
+        options.extend([None, None])
+
+        if bool(additional_expressions):
```
FabioLuporini: pythonicity :  no need for bool

devito/builtins.py
```diff
@@ -7,21 +7,40 @@
...
+    name : str, optional
+        Name of the operator.
     """
-    dv.Operator(dv.Eq(f, v), name='assign')()
+    if not isinstance(f, list):
+        f = list(f)
+    if not isinstance(RHS, list):
+        RHS = [RHS, ]*len(f)
+    eqs = []
+    if bool(options):
```
FabioLuporini: pythonicity :  no need for bool

devito/builtins.py
```diff
@@ -53,6 +72,148 @@ def smooth(f, g, axis=None):
...
+        lhs.append(f_c)
+        rhs.append(f_o)
+        options.append({'subdomain': grid.subdomains['objective_domain']})
+
+        expr['lhs'] = lhs
+        expr['rhs'] = rhs
+        expr['options'] = options
+        additional_expressions[d] = expr
+
+    if isinstance(f, dv.Function):
```
FabioLuporini: perhaps this if-else should be inside `initialize-function` ? ie, it's up to `initialize-function` to do the right thing based on whether the second argument is of type numpy.ndarray or Function?

devito/builtins.py
```diff
@@ -53,6 +72,148 @@ def smooth(f, g, axis=None):
...
+        The function initialisation mode. 'constant' and 'reflect' are
+        accepted.
+    """
+    slices = tuple([slice(nbpml, -nbpml) for _ in range(function.grid.dim)])
+    function.data[slices] = data
+    lhs = []
+    rhs = []
+    options = []
+
+    for d in function.dimensions:
```
FabioLuporini: @mloubout 
Note that this won't work with TimeFunctions 

perhaps we should use `space_dimensions` instead of `dimensions` ?

or am I missing something?

devito/builtins.py
rhodrin: Yes, there will be + quite a few tests.

devito/builtins.py
```diff
@@ -7,21 +7,67 @@
...
-def assign(f, v=0):
+def assign(f, RHS=0, options=None, name='assign', **kwargs):
     """
-    Assign a value to a Function.
+    Assign a list of RHSs to a list of Functions.
 
     Parameters
     ----------
-    f : Function
+    f : Function or list of Function's
```
FabioLuporini: nitpicking: Functions

devito/builtins.py
```diff
@@ -7,21 +7,67 @@
 
 import devito as dv
 
-__all__ = ['assign', 'smooth', 'norm', 'sumall', 'inner', 'mmin', 'mmax']
+__all__ = ['assign', 'smooth', 'gaussian_smooth', 'initialize_function', 'norm',
+           'sumall', 'inner', 'mmin', 'mmax']
 
 
-def assign(f, v=0):
+def assign(f, RHS=0, options=None, name='assign', **kwargs):
```
FabioLuporini: 'rhs', not 'RHS': not sure it should be all capital letters

devito/builtins.py
```diff
@@ -53,6 +99,165 @@ def smooth(f, g, axis=None):
...
+    f_c = dv.Function(name='f_c', grid=grid, space_order=2*lw,
+                      coefficients='symbolic', dtype=np.int32)
+    f_o = dv.Function(name='f_o', grid=grid, coefficients='symbolic', dtype=np.int32)
+
+    weights = np.exp(-0.5/sigma**2*(np.linspace(-lw, lw, 2*lw+1))**2)
+    weights = weights/weights.sum()
+
+    additional_expressions = {}
+    for d in f_c.dimensions:
+        expr = {}
```
FabioLuporini: nitpicking: u probably don't need this one (see below)

devito/builtins.py
```diff
@@ -53,6 +99,165 @@ def smooth(f, g, axis=None):
...
+        options.append({'coefficients': dv.Substitutions(coeffs),
+                        'subdomain': grid.subdomains['objective_domain']})
+        lhs.append(f_c)
+        rhs.append(f_o)
+        options.append({'subdomain': grid.subdomains['objective_domain']})
+
+        expr['lhs'] = lhs
+        expr['rhs'] = rhs
+        expr['options'] = options
+        additional_expressions[d] = expr
```
FabioLuporini:
```
additional_expressions[d] = {'lhs': lhs, 'rhs': rhs, 'options': options}
```

devito/builtins.py
```diff
@@ -53,6 +99,165 @@ def smooth(f, g, axis=None):
...
+        # Check that HALO size is appropriate
+        halo = function.halo
+        local_size = function.grid.distributor.shape
+
+        def buff(i, j):
+            return [(i + k - 2*nbl) for k in j]
+
+        b = [min(l) for l in (w for w in (buff(i, j) for i, j in zip(local_size, halo)))]
+        if any(np.array(b) < 0):
+            raise ValueError("Function's halo is not sufficiently thick.")
```
FabioLuporini: Best if we say which Function
```
"Function `%s`  ... " % function
```

devito/builtins.py
```diff
@@ -53,6 +99,165 @@ def smooth(f, g, axis=None):
...
+        1) 'lhs': List of additional expressions to be added to the LHS expressions list.
+        2) 'rhs': List of additional expressions to be added to the RHS expressions list.
+        3) 'options': Options pertaining to the additional equations that will be
+        constructed.
+    mode : str, optional
+        The function initialisation mode. 'constant' and 'reflect' are
+        accepted.
+    name : str, optional
+        The name assigned to the operator.
+    """
```
FabioLuporini: any chance we can add example here, showing the flexibility given by ``additional_expressions`` ?

devito/builtins.py
```diff
@@ -53,6 +99,165 @@ def smooth(f, g, axis=None):
...
+    grid = dv.Grid(shape=shape_padded, subdomains=objective_domain)
+
+    f_c = dv.Function(name='f_c', grid=grid, space_order=2*lw,
+                      coefficients='symbolic', dtype=np.int32)
+    f_o = dv.Function(name='f_o', grid=grid, coefficients='symbolic', dtype=np.int32)
+
+    weights = np.exp(-0.5/sigma**2*(np.linspace(-lw, lw, 2*lw+1))**2)
+    weights = weights/weights.sum()
+
+    additional_expressions = {}
```
FabioLuporini: nitpicking: for this sort of things, throughout devito we use the conventional name "mapper"

devito/builtins.py
```diff
@@ -53,6 +99,165 @@ def smooth(f, g, axis=None):
...
+    else:
+        function.data[slices] = data
+    lhs = []
+    rhs = []
+    options = []
+
+    if mode == 'reflect' and function.grid.distributor.is_parallel:
+        # Check that HALO size is appropriate
+        halo = function.halo
+        local_size = function.grid.distributor.shape
```
FabioLuporini: why not just ``function.shape`` ?

rhodrin:
> I've just added some more nitpicking, but I understand this is pretty much ready to go ?

Yes, should be.
Will address the comments above with a new commit shortly.

devito/builtins.py
```diff
@@ -53,6 +100,212 @@ def smooth(f, g, axis=None):
...
+
+    Examples
+    --------
+    In the following example the `inner` region of a function is set to one plus the
+    value on the boundary.
+
+    >>> import numpy as np
+    >>> from devito import Grid, SubDomain, Function, initialize_function
+
+    Create a subdomain representing the `inner` region:
```
FabioLuporini: this is actually `grid.interior` ?

devito/builtins.py
```diff
@@ -53,6 +100,212 @@ def smooth(f, g, axis=None):
...
+
+    fset(f, f_c)
+
+    return f
+
+
+def initialize_function(function, data, nbl, mapper=dict(), mode='constant',
+                        name='padfunc'):
+    """
+    Initialize a `Function` with the given ``data``. ``data``
```
FabioLuporini: we don't use backtips for types (so \`Function\` -> Function)

devito/builtins.py
```diff
@@ -53,6 +100,212 @@ def smooth(f, g, axis=None):
...
+    does *not* include the ``nbl`` outer/boundary layers; these are added via padding
+    by this function.
+
+    Parameters
+    ----------
+    function : Function
+        The initialised object.
+    data : ndarray of Function
```
FabioLuporini: *or ?

devito/builtins.py
```diff
@@ -53,6 +100,212 @@ def smooth(f, g, axis=None):
...
+    initialize_function(f_c, f.data[:], lw,
+                        mapper=mapper,
+                        mode='reflect', name='smooth')
+
+    fset(f, f_c)
+
+    return f
+
+
+def initialize_function(function, data, nbl, mapper=dict(), mode='constant',
```
FabioLuporini: never use mutable objects as default values . Just google for it, you'll find loads of explanations, for example [here](https://docs.python-guide.org/writing/gotchas/)

devito/builtins.py
```diff
@@ -53,6 +100,212 @@ def smooth(f, g, axis=None):
...
+    Parameters
+    ----------
+    function : Function
+        The initialised object.
+    data : ndarray of Function
+        The data used for initialisation.
+    nbl : int
+        Number of outer layers (such as PML layers for boundary damping).
+    mapper : dict, optional
+        Dictionary containing, for each dimension of function, a sub-dictionary
```
FabioLuporini: here you probably mean \`function\` as you're referring to a specific variable name

devito/builtins.py
```diff
@@ -53,6 +100,212 @@ def smooth(f, g, axis=None):
...
+    >>> grid = Grid(shape=(6, 6), subdomains=inner)
+    >>> x, y = grid.dimensions
+
+    Create the function we wish to set along with the data we wish to use to set it:
+
+    >>> f = Function(name='f', grid=grid, dtype=np.int32)
+    >>> data = np.full((4, 4), 2, dtype=np.int32)
+
+    Create the additional expressions and options required to set the value of
+    the inner region to one greater than the boundary value. Note that the equation
```
FabioLuporini: > Note that the equation is specified on the second grid dimension so that one is added to the inner values after all boundary values have been set.

This behaviour is not very clear to me. What would happen if you had used `x` in place of `y` ?

devito/builtins.py
FabioLuporini: reminder

devito/builtins.py
FabioLuporini: or use the new `as_list` ?

devito/builtins.py
```diff
@@ -53,6 +100,212 @@ def smooth(f, g, axis=None):
...
+        rhs.append(function.subs({d: subsr}))
+        options.extend([None, None])
+
+        if mapper and d in mapper.keys():
+            exprs = mapper[d]
+            lhs_extra = exprs['lhs']
+            rhs_extra = exprs['rhs']
+            lhs.extend(as_list(lhs_extra))
+            rhs.extend(as_list(rhs_extra))
+            if 'options' in exprs.keys() and exprs['options']:
```
FabioLuporini: no need for `.keys()`

devito/builtins.py
```diff
@@ -53,6 +100,212 @@ def smooth(f, g, axis=None):
...
+                options_extra = exprs['options']
+            else:
+                options_extra = len(as_list(lhs_extra))*[None, ]
+            if isinstance(options_extra, list):
```
FabioLuporini: you may be able to use `as_list` again, instead of this if-else ?

devito/tools/utils.py
```diff
@@ -12,13 +12,36 @@
...
            'filter_sorted', 'dtype_to_cstr', 'dtype_to_ctype', 'dtype_to_mpitype',
-           'ctypes_to_cstr', 'ctypes_pointer', 'pprint', 'sweep', 'all_equal']
+           'ctypes_to_cstr', 'ctypes_pointer', 'pprint', 'sweep', 'all_equal', 'as_list']
 
 
 def prod(iterable, initial=1):
     return reduce(mul, iterable, initial)
 
 
+def as_list(item, type=None, length=None):
```
FabioLuporini: how about just
```
return list(as_tuple(item, type=type, length=length))
```

devito/builtins.py
rhodrin: You'd pad in the `y` direction with values of 1 greater than those padded in the `x` direction.

devito/builtins.py
rhodrin: Not in this case since is could be a single dictionary which the current versions of `as_tuple` and `as_list` mess up. But I'm not sure if this behavior was intended or is a bug?

devito/builtins.py
```diff
@@ -53,6 +98,207 @@ def smooth(f, g, axis=None):
...
+        rhs.append(function.subs({d: subsr}))
+        options.extend([None, None])
+
+        if mapper and d in mapper.keys():
+            exprs = mapper[d]
+            lhs_extra = exprs['lhs']
+            rhs_extra = exprs['rhs']
+            lhs.extend(as_list(lhs_extra))
+            rhs.extend(as_list(rhs_extra))
+            if 'options' in exprs and exprs['options']:
```
FabioLuporini: the pythonic way to write these 4 lines would be

options_extra = exprs.get('options', len(as_list(lhs_extra))*[None, ])

devito/builtins.py
```diff
@@ -53,6 +98,207 @@ def smooth(f, g, axis=None):
...
+
+    >>> import numpy as np
+    >>> from devito import Grid, SubDomain, Function, initialize_function
+
+    Create the computational domain:
+
+    >>> grid = Grid(shape=(6, 6))
+    >>> x, y = grid.dimensions
+
+    Create the function we wish to set along with the data to set it:
```
FabioLuporini: Function

FabioLuporini:
merged

PR #934, opened 2019-09-10
Ops/small fix

btrb:
This PR enables correct code gen for `ops_block`.

### DIFF
--- a/devito/ops/types.py
+++ b/devito/ops/types.py
@@ -135,7 +135,7 @@ class OpsBlock(basic.Symbol):
         super().__init__(name, np.void, *args, **kwargs)
 
     @property
-    def _C_typename(self):
+    def _C_typedata(self):
         return namespace['ops_block_type']
 
 
--- a/tests/test_ops.py
+++ b/tests/test_ops.py
@@ -20,7 +20,8 @@ from devito.ops.types import OpsAccessible, OpsDat, OpsStencil, OpsBlock # noqa
 from devito.ops.utils import namespace # noqa
 from devito.symbolics import Byref, Literal, indexify # noqa
 from devito.tools import dtype_to_cstr # noqa
-from devito.types import Constant, Symbol # noqa
+from devito.types import Constant, Symbol, Buffer # noqa
+
 
 
 class TestOPSExpression(object):
@@ -226,7 +227,7 @@ class TestOPSExpression(object):
             Literal('"u"')
         )
 
-        def test_create_ops_arg_constant(self):
+    def test_create_ops_arg_constant(self):
             a = Constant(name='*a')
 
             res = create_ops_arg(a, {}, {})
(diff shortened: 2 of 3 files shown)


tests/test_ops.py
```diff
@@ -226,33 +226,49 @@ def test_create_ops_dat_function(self):
...
+            stencil,
+            Literal('"%s"' % dtype_to_cstr(u.dtype)),
+            namespace['ops_read'] if read else namespace['ops_write']
+        ]
+
+    @pytest.mark.parametrize('equation, expected', [
+        ('Eq(u.forward, u.dt2 + u.dxr - u.dyr - u.dyl)',
+            'ops_block block = ops_decl_block(2, "block");'),
+        ('Eq(u.forward,u+1)',
+            'ops_block block = ops_decl_block(2, "block");')
```
maelso: For greater coverage, you can test 1D, 2D, and 3D blocks.

btrb:
weird, the failings in Travis are completely unrelated to this PR. plus it also fails only for `gcc-7`

tests/test_ops.py
```diff
@@ -20,7 +20,7 @@
 from devito.ops.utils import namespace # noqa
 from devito.symbolics import Byref, Literal, indexify # noqa
 from devito.tools import dtype_to_cstr # noqa
-from devito.types import Constant, Symbol # noqa
+from devito.types import Constant, Symbol, Buffer # noqa
```
vmickus: Just for coherence Devito usually sort the import alphabetically, so here `Buffer` should be ther first one. 
But as there is thi s` # noqa`. I don't know if its necessary.

PS: `indexify` should also change its position but I know it was probably me that did this.

tests/test_ops.py
```diff
@@ -226,33 +226,49 @@ def test_create_ops_dat_function(self):
...
+        ('Eq(u.forward,u+1)',
+            'ops_block block = ops_decl_block(2, "block");')
+    ])
+    def test_create_ops_block(self, equation, expected):
+        """
+        Test if ops_block has been successfully generated
+        """
+        grid_2d = Grid(shape=(4, 4))
+        u = TimeFunction(name='u', grid=grid_2d, time_order=2, save=Buffer(10))  # noqa
+        op = OperatorOPS(eval(equation))
```
vmickus: I am curious why you didn't use just `Operator`. 

I mean, when using Devito we will only specify the backend through the `DEVITO_BACKEND` env. variable. And then, Devito will choose the correct `Operator`  to use.  
So if you test using this command: `DEVITO_BACKEND=ops pytest tests/test_ops.py` it will switch to this Operator automaticaly.

devito/ops/types.py
```diff
@@ -135,7 +135,7 @@ def __init__(self, name, *args, **kwargs):
         super().__init__(name, np.void, *args, **kwargs)
 
     @property
-    def _C_typename(self):
+    def _C_typedata(self):
```
vmickus: I liked that the solution was this simple! =]

tests/test_ops.py
btrb: thank you, fixed

tests/test_ops.py
btrb: actually i just followed the pattern from `test_kernel_generation`.
u're right thou, let me change both

btrb:
all tests seem to pass locally now

PR #935, opened 2019-09-11
Tweak elastic and viscoelastic testing

rhodrin:
Running the 3d elastic example was time consuming and the example was not being 'tested'. This PR removes running (untested) elastic and viscoelastic examples during build testing and now only the 2d simulations with norm tests are run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -121,15 +121,13 @@ script:
   - if [[ $RUN_EXAMPLES == 'True' ]]; then python examples/seismic/acoustic/acoustic_example.py --full; fi
   - if [[ $RUN_EXAMPLES == 'True' ]]; then python examples/seismic/acoustic/acoustic_example.py --full --checkpointing; fi
   - if [[ $RUN_EXAMPLES == 'True' ]]; then python examples/seismic/acoustic/acoustic_example.py --constant --full; fi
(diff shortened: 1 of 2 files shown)


mloubout:
I don't really like that. Yes these are slower but that's not the point. We are running all the other examples I don't see why we should remove these ones or why even bother running any example. The point is we provide examples and we make sure it runs.

rhodrin:
> I don't really like that. Yes these are slower but that's not the point. We are running all the other examples I don't see why we should remove these ones or why even bother running any example. The point is we provide examples and we make sure it runs.

The other examples don't have associated tests. Also, we can't run all elastic and viscoelastic examples examples with all the options a user could provide, so we're not really 'testing' much extra by ensuring that an elastic example doesn't throw a run time error (it could be producing `NaN`'s for 6 minutes).
Maybe we should add a norm test for a small 3d case? But I don't think we should be spending a great deal of time running simulations whose results aren't tested in any sense.

rhodrin:
> Yes these are slower but that's not the point

Also, this point comes back to the discussion we had during the meeting yesterday i.e. Do we want some separate testing suite where longer examples are being run (which I think we should).

FabioLuporini:
Merged.

PR #936, opened 2019-09-12
Feature - add ops fetch

vmickus:
**Add ops_dat_fetch_data call to the OPS backend**    

- This OPS API call is responsible to get data from the GPU and returning to the pointer allocated from Devito in the main memory.
- For all ops_dat created, it is then created an ops_dat_fetch_dat call, that will bring the result from the GPU.
- Modified Byref class to use ccode instead of plain str.
- Created new tests for this OPS feature and modified tests for ops_dat.
- Modified `ops_args` dictionary to hold more information about the `OpsAccessible` creation.
- Modified dict name_to_ops_dat to only have values that are OpsDat or Array of OpsDat, instead of also having Indexed.

**This is basically #927 + rebasing on top of latest master**

### DIFF
--- a/devito/ops/operator.py
+++ b/devito/ops/operator.py
@@ -58,7 +58,8 @@ class OperatorOPS(Operator):
         to_dat = filter_sorted(to_dat)
 
         iteration_tree = retrieve_iteration_tree(iet, mode='normal')[0]
-        time_upper_bound = iteration_tree.dimensions[TimeFunction._time_position].extreme_max
+        time_upper_bound = iteration_tree.dimensions[TimeFunction._time_position]\
+            .extreme_max
 
         name_to_ops_dat = {}
         pre_time_loop = []
--- a/devito/ops/operator.py
+++ b/devito/ops/operator.py
@@ -58,8 +58,9 @@ class OperatorOPS(Operator):
         to_dat = filter_sorted(to_dat)
 
         iteration_tree = retrieve_iteration_tree(iet, mode='normal')[0]
-        time_upper_bound = iteration_tree.dimensions[TimeFunction._time_position]\
-            .extreme_max
+        if iteration_tree:
+            time_upper_bound = iteration_tree.dimensions[TimeFunction._time_position]\
+                .extreme_max
 
         name_to_ops_dat = {}
         pre_time_loop = []
--- a/devito/ops/operator.py
+++ b/devito/ops/operator.py
@@ -8,7 +8,7 @@ from devito.symbolics import Literal
 from devito.tools import filter_sorted
 
 from devito.ops import ops_configuration
-from devito.ops.transformer import create_ops_dat, opsit
+from devito.ops.transformer import create_ops_dat, create_ops_fetch, opsit
 from devito.ops.types import OpsBlock
 from devito.ops.utils import namespace
 
@@ -58,11 +58,15 @@ class OperatorOPS(Operator):
 
         name_to_ops_dat = {}
         pre_time_loop = []
+        after_time_loop = []
         for f in to_dat:
             if f.is_Constant:
                 continue
 
             pre_time_loop.extend(create_ops_dat(f, name_to_ops_dat, ops_block))
+            # To return the result to Devito, it is necessary to copy the data
+            # from the GPU into the CPU memory
+            after_time_loop.append(create_ops_fetch(f, name_to_ops_dat))
 
         # Generate ops kernels for each offloadable iteration tree
         mapper = {}
@@ -74,7 +78,8 @@ class OperatorOPS(Operator):
             pre_time_loop.extend(pre_loop)
             self._ops_kernels.append(ops_kernel)
             mapper[trees[0].root] = ops_par_loop_call
-            mapper.update({i.root: mapper.get(i.root) for i in trees})  # Drop trees
+            mapper.update({i.root: mapper.get(i.root)
+                           for i in trees})  # Drop trees
 
         iet = Transformer(mapper).visit(iet)
 
@@ -85,7 +90,8 @@ class OperatorOPS(Operator):
         self._headers.append(namespace['ops_define_dimension'](dims[0]))
         self._includes.append('stdio.h')
 
-        body = [ops_init, ops_block_init, *pre_time_loop, ops_partition, iet, ops_exit]
+        body = [ops_init, ops_block_init, *pre_time_loop,
+                ops_partition, iet, *after_time_loop, ops_exit]
 
         return List(body=body)
 
--- a/devito/ops/operator.py
+++ b/devito/ops/operator.py
@@ -1,6 +1,7 @@
-from devito import Eq
+from devito import Eq, TimeFunction
 from devito.ir.equations import ClusterizedEq
-from devito.ir.iet import Call, List, Expression, find_affine_trees
+from devito.ir.iet import (Call, Expression, find_affine_trees,
+                           FindNodes, Iteration, List, retrieve_iteration_tree)
 from devito.ir.iet.visitors import FindSymbols, Transformer
 from devito.logger import warning
 from devito.operator import Operator
@@ -9,7 +10,7 @@ from devito.tools import filter_sorted
 
 from devito.ops import ops_configuration
 from devito.ops.transformer import create_ops_dat, create_ops_fetch, opsit
-from devito.ops.types import OpsBlock
+from devito.ops.types import OpsBlock, OpsAccessible
 from devito.ops.utils import namespace
 
 __all__ = ['OperatorOPS']
@@ -65,8 +66,8 @@ class OperatorOPS(Operator):
 
             pre_time_loop.extend(create_ops_dat(f, name_to_ops_dat, ops_block))
             # To return the result to Devito, it is necessary to copy the data
-            # from the GPU into the CPU memory
-            after_time_loop.append(create_ops_fetch(f, name_to_ops_dat))
+            # from the dat object into the CPU memory
+            after_time_loop.extend(create_ops_fetch(f, name_to_ops_dat))
 
         # Generate ops kernels for each offloadable iteration tree
         mapper = {}
--- a/devito/ops/operator.py
+++ b/devito/ops/operator.py
@@ -84,8 +84,7 @@ class OperatorOPS(Operator):
             pre_time_loop.extend(pre_loop)
             self._ops_kernels.append(ops_kernel)
             mapper[trees[0].root] = ops_par_loop_call
-            mapper.update({i.root: mapper.get(i.root)
-                           for i in trees})  # Drop trees
+            mapper.update({i.root: mapper.get(i.root) for i in trees})  # Drop trees
 
         iet = Transformer(mapper).visit(iet)
 
--- a/devito/ops/operator.py
+++ b/devito/ops/operator.py
@@ -30,6 +30,13 @@ class OperatorOPS(Operator):
     def _specialize_iet(self, iet, **kwargs):
         warning("The OPS backend is still work-in-progress")
 
+        # If there is no iteration tree, then there is no loop to be optimized using OPS.
+        iteration_tree = retrieve_iteration_tree(iet, mode='normal')
+        if not len(iteration_tree):
+            return iet
+        time_upper_bound = iteration_tree[0].dimensions[TimeFunction._time_position]\
+            .extreme_max
+
         ops_init = Call(namespace['ops_init'], [0, 0, 2])
         ops_partition = Call(namespace['ops_partition'], Literal('""'))
         ops_exit = Call(namespace['ops_exit'])
@@ -56,11 +63,6 @@ class OperatorOPS(Operator):
         # To ensure deterministic code generation we order the datasets to
         # be generated (since a set is an unordered collection)
         to_dat = filter_sorted(to_dat)
-
-        iteration_tree = retrieve_iteration_tree(iet, mode='normal')[0]
-        if iteration_tree:
-            time_upper_bound = iteration_tree.dimensions[TimeFunction._time_position]\
-                .extreme_max
 
         name_to_ops_dat = {}
         pre_time_loop = []
--- a/devito/ops/operator.py
+++ b/devito/ops/operator.py
@@ -1,7 +1,7 @@
 from devito import Eq, TimeFunction
 from devito.ir.equations import ClusterizedEq
 from devito.ir.iet import (Call, Expression, find_affine_trees,
-                           FindNodes, Iteration, List, retrieve_iteration_tree)
+                           List, retrieve_iteration_tree)
 from devito.ir.iet.visitors import FindSymbols, Transformer
 from devito.logger import warning
 from devito.operator import Operator
@@ -10,7 +10,7 @@ from devito.tools import filter_sorted
 
 from devito.ops import ops_configuration
 from devito.ops.transformer import create_ops_dat, create_ops_fetch, opsit
-from devito.ops.types import OpsBlock, OpsAccessible
+from devito.ops.types import OpsBlock
 from devito.ops.utils import namespace
 
 __all__ = ['OperatorOPS']
@@ -57,6 +57,9 @@ class OperatorOPS(Operator):
         # be generated (since a set is an unordered collection)
         to_dat = filter_sorted(to_dat)
 
+        iteration_tree = retrieve_iteration_tree(iet, mode='normal')[0]
+        time_upper_bound = iteration_tree.dimensions[TimeFunction._time_position].extreme_max
+
         name_to_ops_dat = {}
         pre_time_loop = []
         after_time_loop = []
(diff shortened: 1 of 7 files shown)


tests/test_ops.py
```diff
@@ -225,25 +194,26 @@ def test_create_ops_dat_function(self):
             Literal('"u"')
         )
 
-    def test_create_ops_arg_constant(self):
-        a = Constant(name='*a')
+        def test_create_ops_arg_constant(self):
+            a = Constant(name='*a')
```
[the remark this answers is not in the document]
vmickus: Done.

devito/ops/transformer.py
```diff
@@ -169,12 +166,18 @@ def create_ops_dat(f, name_to_ops_dat, block):
...
+        # Inserting the ops_dat array in case of TimeFunction.
+        name_to_ops_dat[f.name] = ops_dat_array
+
     else:
         ops_dat = OpsDat("%s_dat" % f.name)
         name_to_ops_dat[f.name] = ops_dat
 
-        d_p_val = tuple(sympify([p[0] + h[0] for p, h in zip(f.padding, f.halo)]))
-        d_m_val = tuple(sympify([-(p[1] + h[1]) for p, h in zip(f.padding, f.halo)]))
+        d_p_val = tuple(sympify([p[0] + h[0]
```
FabioLuporini: Thinking about this again... Could you try:

 `d_p_val = tuple(i.left for i in f._size_nodomain)`

or perhaps just:

`d_p_val = f._size_nodoamin.left` 

might work?

about the jargon: nodomain == padding+halo

devito/ops/transformer.py
```diff
@@ -35,16 +35,19 @@ def opsit(trees, count, name_to_ops_dat, block, dims):
         expressions.extend(FindNodes(Expression).visit(tree.inner))
 
     for expr in expressions:
-        ops_expressions.append(Expression(make_ops_ast(expr.expr, node_factory)))
+        ops_expressions.append(Expression(
```
FabioLuporini: ops_expressions.extend([Expression(...) for expr in expressions])

devito/ops/transformer.py
```diff
@@ -150,9 +149,7 @@ def create_ops_dat(f, name_to_ops_dat, block):
         dat_decls = []
         for i in range(time_dims):
             name = '%s%s%s' % (f.name, time_index, i)
-            name_to_ops_dat[name] = ops_dat_array.indexify(
-                [Symbol('%s%s' % (time_index, i))]
-            )
+
```
FabioLuporini: nitpicking: unneeded?

devito/ops/transformer.py
FabioLuporini: reminder

devito/ops/types.py
```diff
@@ -31,7 +31,7 @@ class OpsAccessible(basic.Symbol):
     """
     is_Scalar = True
 
-    def __new__(cls, name, dtype, read_only, *args, **kwargs):
+    def __new__(cls, name, dtype, read_only, * args, **kwargs):
```
FabioLuporini: drop leftover

devito/symbolics/extended_sympy.py
```diff
@@ -303,9 +304,9 @@ def base(self):
 
     def __str__(self):
         if self.base.is_Symbol:
-            return "&%s" % self.base
+            return "&%s" % ccode(self.base)
```
FabioLuporini: rebase ?

devito/ops/transformer.py
vmickus: I was trying to use this, but the problem is that the type of the elements inside the tuple of `f._size_nodoamin.left` are `numpy.int64`. Whereas the types in the previous solution are `sympy.core.numbers.Integer`. 
This difference causes error when using the `ListInitializer` (few steps later):
`ValueError: `params` must be an iterable of Expr or str`

And if I remove the `ListInitializer` call I won't have errors but the code generated is wrong:
``` 
u_d_p(d_p) = (0, 2, 2);
# instead of
int u_d_p[2] = {2, 2};
```

So I think I should stick with the current solution.

devito/ops/transformer.py
vmickus: Fixed, now using python list comprehension. 

Also I'll add this same approach to the loop above this one.

devito/ops/transformer.py
vmickus: Fixed.

devito/symbolics/extended_sympy.py
vmickus: I think I didn't understand. 

I'll do a rebase, but this part of the code is correct, the `ccode` need to be inserted here.

devito/ops/types.py
vmickus: Strange, in my local branch does not have this space.

vmickus: Forget what I said, I was looking in the wrong branch.

Fixed!

devito/ops/transformer.py
vmickus: This was moved to after the loop, line 171.

devito/ops/transformer.py
FabioLuporini: Can u cast them to int(...)

vmickus: I tried casting to `int(...)`, but I get the following error:

```AttributeError: 'int' object has no attribute 'is_Indexed'```

But converting to `sympy.core.numbers.Integer` works!

I just don't know if this is acceptable and if this is better than using `sympfy`, because, in the end, this is what sympy is doing.

File: devito/symbolics/extended_sympy.py (hunk 6 above)
Comment (FabioLuporini) [reply devito/symbolics/extended_sympy.py]: never mind, ignore me: I thought for a moment that this had already been fixed in master and that somewhat you were missing a rebase

File: devito/ops/transformer.py (hunk 2 above)
Comment (FabioLuporini) [reply devito/ops/transformer.py]: anyway, clearly this is a deficiency of `ListInitializer`

Turn [this](https://github.com/opesci/devito/blob/master/devito/symbolics/extended_sympy.py#L265) into `is_integer` and we should be good to go leaving everything else as I suggested in the very first place

Comment (vmickus) [reply devito/ops/transformer.py]: This also is not enough.

So I used the first version, setting `f._size_nodomain.left` as the `d_p_val`.
This property returns a tuple of `numpy.int64`. 

Fixing the `ListInitializer` helped, but the problem is that this is not Sympy object, and later there is pass in

Although now it can identify the `int` inside the `ListInitializer`, it will fail because the type `int` or `np.int64` does not have `is_Indexed` attribute.

The stack trace from the error:
```
The OPS backend is still work-in-progress
  File "04-fun-timefun.py", line 9, in <module>
    op = Operator(Eq(u.forward, u + v.dx))
  File "/home/mickus/Documents/master/projects/devito-vmickus/devito/backends.py", line 73, in __call__
    return t(*args, **kwargs)
  File "/home/mickus/Documents/master/projects/devito-vmickus/devito/ops/operator.py", line 26, in __init__
    super().__init__(*args, **kwargs)
  File "/home/mickus/Documents/master/projects/devito-vmickus/devito/operator.py", line 187, in __init__
    iet = self._specialize_iet(iet, **kwargs)
  File "/home/mickus/Documents/master/projects/devito-vmickus/devito/ops/operator.py", line 70, in _specialize_iet
    pre_time_loop.extend(create_ops_dat(f, name_to_ops_dat, ops_block))
  File "/home/mickus/Documents/master/projects/devito-vmickus/devito/ops/transformer.py", line 196, in create_ops_dat
    res.append(Expression(ClusterizedEq(Eq(d_p, ListInitializer(d_p_val)))))
  File "</home/mickus/miniconda3/envs/devito-vmickus/lib/python3.7/site-packages/decorator.py:decorator-gen-3>", line 2, in __init__
  File "/home/mickus/Documents/master/projects/devito-vmickus/devito/tools/validators.py", line 33, in wrapper
    return f(*args, **kwargs)
  File "/home/mickus/Documents/master/projects/devito-vmickus/devito/ir/iet/nodes.py", line 245, in __init__
    self.__expr_finalize__(expr)
  File "/home/mickus/Documents/master/projects/devito-vmickus/devito/ir/iet/nodes.py", line 250, in __expr_finalize__
    self._reads, _ = detect_io(expr, relax=True)
  File "/home/mickus/Documents/master/projects/devito-vmickus/devito/ir/support/utils.py", line 152, in detect_io
    terminals = flatten(retrieve_terminals(i, deep=True) for i in roots)
  File "/home/mickus/Documents/master/projects/devito-vmickus/devito/tools/utils.py", line 124, in flatten
    for el in l:
  File "/home/mickus/Documents/master/projects/devito-vmickus/devito/ir/support/utils.py", line 152, in <genexpr>
    terminals = flatten(retrieve_terminals(i, deep=True) for i in roots)
  File "/home/mickus/Documents/master/projects/devito-vmickus/devito/symbolics/search.py", line 155, in retrieve_terminals
    return search(expr, q_terminal, mode, 'dfs', deep)
  File "/home/mickus/Documents/master/projects/devito-vmickus/devito/symbolics/search.py", line 114, in search
    return searcher.dfs(expr)
  File "/home/mickus/Documents/master/projects/devito-vmickus/devito/symbolics/search.py", line 67, in dfs
    found.update(self.dfs(a))
  File "/home/mickus/Documents/master/projects/devito-vmickus/devito/symbolics/search.py", line 66, in dfs
    for a in self._next(expr):
  File "/home/mickus/Documents/master/projects/devito-vmickus/devito/symbolics/search.py", line 49, in _next
    if self.deep is True and expr.is_Indexed:
AttributeError: 'numpy.int64' object has no attribute 'is_Indexed'
```

File: devito/ops/operator.py [hunk 7]
```diff
@@ -54,13 +55,22 @@ def _specialize_iet(self, iet, **kwargs):
         # be generated (since a set is an unordered collection)
         to_dat = filter_sorted(to_dat)
 
+        iteration_tree = retrieve_iteration_tree(iet, mode='normal')[0]
```
FabioLuporini: [0] ? 
what if there's none?

devito/ops/operator.py
```diff
@@ -72,7 +82,8 @@ def _specialize_iet(self, iet, **kwargs):
             pre_time_loop.extend(pre_loop)
             self._ops_kernels.append(ops_kernel)
             mapper[trees[0].root] = ops_par_loop_call
-            mapper.update({i.root: mapper.get(i.root) for i in trees})  # Drop trees
+            mapper.update({i.root: mapper.get(i.root)
+                           for i in trees})  # Drop trees
```
FabioLuporini: does it not fit on a single line?

devito/ops/transformer.py
FabioLuporini: can you not convert it to Integer/Number ? In the ListInitializer constructor I mean

so, long story short: there is this `_size_nodomain` method, and we should aim to use it. Accessing `f.padding` or `f.halo` directly doesn't make much sense. In fact, you reminded me that these attributes should be turned into private attributes

devito/ops/transformer.py
```diff
@@ -244,13 +267,23 @@ def create_ops_arg(p, name_to_ops_dat, par_to_ops_stencil):
...
-        return namespace['ops_arg_dat'](
-            name_to_ops_dat[p.name],
-            1,
-            par_to_ops_stencil[p],
-            Literal('"%s"' % dtype_to_cstr(p.dtype)),
-            namespace['ops_read'] if p.read_only else namespace['ops_write']
-        )
+        accessible_info = accessible_origin[p.name]
+
+        if accessible_info.time:
```
FabioLuporini: This if-the-else takes 15 lines, but the difference in the `return` statement is only in one argument... it's a real pity... can we not improve this? Ideally I'd have

```
return namespace['ops_arg_dat'](....., 1, par_to_ops_stencil[p], Literal(...) ...)
```

devito/ops/types.py
FabioLuporini: no it's not fixed

devito/ops/transformer.py
vmickus: Sure, can do.

devito/ops/operator.py
vmickus: Yes, it fits.

devito/ops/operator.py
vmickus: I added a test in case there is none. 

But basically,if there is none, means that there is no loop and then there is no code to be specialized using OPS.

devito/ops/transformer.py
vmickus: Soo.. I end it up needing to add a new if/else...
But i followed your advise and tried to shorten it. 

But the previous if/else that existed I didn't manage to improve, because it was getting messy and confused IMO.

devito/ops/operator.py
FabioLuporini: yeah, what I meant was that instead of adding a test, you could `return iet` and perhaps raise a warning in case there's none. 

Devito can generate codes beyond finite differences, which might not fit the OPS abstraction, so...

devito/symbolics/extended_sympy.py
```diff
@@ -262,8 +263,8 @@ def __new__(cls, params):
         for p in as_tuple(params):
             if isinstance(p, str):
                 args.append(Symbol(p))
-            elif isinstance(p, int):
-                args.append(p)
+            elif isinstance(p, (int, np.integer)):
```
FabioLuporini: you're looking for
```
from devito.tools import is_integer
```

devito/ops/transformer.py
```diff
@@ -169,13 +164,17 @@ def create_ops_dat(f, name_to_ops_dat, block):
...
+        name_to_ops_dat[f.name] = ops_dat_array
+
     else:
         ops_dat = OpsDat("%s_dat" % f.name)
         name_to_ops_dat[f.name] = ops_dat
 
-        d_p_val = tuple(sympify([p[0] + h[0] for p, h in zip(f.padding, f.halo)]))
-        d_m_val = tuple(sympify([-(p[1] + h[1]) for p, h in zip(f.padding, f.halo)]))
-        dim_shape = sympify(f.shape)
+        dim_shape = f.shape
```
FabioLuporini: I now find it much more readable...don't you?

devito/ops/transformer.py
```diff
@@ -134,11 +133,9 @@ def create_ops_dat(f, name_to_ops_dat, block):
         time_index = f.indices[time_pos]
         time_dims = f.shape[time_pos]
 
-        dim_shape = sympify(f.shape[:time_pos] + f.shape[time_pos + 1:])
-        padding = f.padding[:time_pos] + f.padding[time_pos + 1:]
-        halo = f.halo[:time_pos] + f.halo[time_pos + 1:]
-        d_p_val = tuple(sympify([p[0] + h[0] for p, h in zip(padding, halo)]))
-        d_m_val = tuple(sympify([-(p[1] + h[1]) for p, h in zip(padding, halo)]))
+        dim_shape = f.shape[:time_pos] + f.shape[time_pos + 1:]
```
FabioLuporini: now it is readable :)

devito/ops/transformer.py
vmickus: Totally agree. When I was creating this feature it was real trouble understanding this part of the code so I could modify to my needs.

devito/symbolics/extended_sympy.py
vmickus: What happens if p is already an Integer?

The `is_integer` also tests for this case and this is the reason why I didn't use it.

FabioLuporini: In [11]: from sympy import Integer                                                                                                                                                                                 

In [12]: Integer(Integer(2))                                                                                                                                                                                       
Out[12]: 2

devito/ops/operator.py
vmickus: Good idea, I'll do it.

devito/symbolics/extended_sympy.py
vmickus: It still seems a little awkward to create an Integer if it is already an Integer. 

The Integer case will still be added by the else condition here.

But I'm OK with change it to use is_integer if there is no problem in the tests.

FabioLuporini: so the point is in theory we could at some point have other objects that behave "like" integers, but are not any of sympy/numpy/int . And if we use `is_integer`, the idea is that we capture them all. So it's "safer", put it that way

devito/ops/operator.py
vmickus: Fixed.

I also moved this verification to the beginning of the `specialize_iet` because there is no point in continuing the specialization for OPS if there is no iteration tree.

devito/symbolics/extended_sympy.py
vmickus: Fixed.

devito/ops/operator.py
```diff
@@ -29,6 +30,13 @@ def __init__(self, *args, **kwargs):
     def _specialize_iet(self, iet, **kwargs):
         warning("The OPS backend is still work-in-progress")
 
+        # If there is no iteration tree, then there is no loop to be optimized using OPS.
+        iteration_tree = retrieve_iteration_tree(iet, mode='normal')
```
FabioLuporini: for the next PR: for homogeneity of code, we should name this variable "trees"

devito/ops/operator.py
```diff
@@ -29,6 +30,13 @@ def __init__(self, *args, **kwargs):
...
+        if not len(iteration_tree):
+            return iet
+        time_upper_bound = iteration_tree[0].dimensions[TimeFunction._time_position]\
```
FabioLuporini: for the next PR: assuming that `TimeFunction._time_position` corresponds to the first entry in the tree is generally wrong (though often true in practice)

FabioLuporini:
merged.

PR #937, opened 2019-09-17
Clear_cache inside devito

FabioLuporini:
You know what this is about.

~Obviously: **This PR isn't ready to be merged yet**~


navjotk:
Ok this looks like a nice cleanup on the way. What is missing now?

FabioLuporini:
> Ok this looks like a nice cleanup on the way. What is missing now?

Pretty much ready. I think it needs some beta testing, as it's a quite profound change

devito/types/basic.py
```diff
@@ -1116,10 +1127,23 @@ class CacheManager(object):
     data is lost (and thus memory is freed).
     """
 
+    gc_ths = 3*10**8
```
georgebisbas: How is this value chosen?

examples/seismic/tutorials/02_rtm.ipynb
```diff
@@ -470,7 +470,7 @@
    ],
    "source": [
     "# Run imaging loop over shots\n",
-    "from devito import Function, clear_cache\n",
+    "from devito import Function\n",
```
mloubout: So nitpicking, but currently `u0` is still created once at the beginning and passed to the operator for each shot so memory overhead is still tiny without clear_cache the way the loop is implemented. Would be better to remove `u0=..., u0.data.fill` and actually see if creating a new `u` at each shots doesn't blow up memory. Same for FWI.

FabioLuporini:
unfortunately it turns out that calling `clear_cache` at the end of `op.apply` doesn't suffice. I'm closing this PR and will open a new one as soon as it's ready

PR #938, opened 2019-09-17
Created and using compiler to CUDA architecture

maelso:
Moving forward OPS compiler.

Since PR [886](https://github.com/opesci/devito/pull/886) has created the OPS Compiler, in this PR the code is being compiled for NVIDIA GPUs.

### DIFF
--- a/devito/ops/compiler.py
+++ b/devito/ops/compiler.py
@@ -35,7 +35,23 @@ class CompilerOPS(configuration['compiler'].__class__):
             + link
         )
 
-    def prepare_ops(self, soname, ccode, hcode):
+    def jit_compile(self, soname, ccode, hcode):
+        self.__translate_ops(soname, ccode, hcode)
+        self.target = str(self.get_jit_dir().joinpath(soname))
+        self.ops_src = '%s/%s_ops.cpp' % (self.get_jit_dir(), soname)
+        self.cache_dir = self.get_codepy_dir().joinpath(soname[:7])
+
+        # Make a suite of cache directories based on the soname
+        self.cache_dir.mkdir(parents=True, exist_ok=True)
+        try:
+            with open(self.ops_src, 'r') as f:
+                self.code = f.read()
+        except FileNotFoundError:
+            warning("The file %s isn't present" % self.ops_src)
+        else:
+            self.__compile_cuda(soname)
+
+    def __translate_ops(self, soname, ccode, hcode):
         # Creating files
         file_name = str(self.get_jit_dir().joinpath(soname))
         h_file = open("%s.h" % (file_name), "w")
@@ -54,7 +70,7 @@ class CompilerOPS(configuration['compiler'].__class__):
             warning("Couldn't find OPS_INSTALL_PATH \
                 environment variable, please check your OPS installation")
 
-    def cuda_compiler(self, soname):
+    def __compile_cuda(self, soname):
         # CUDA kernel compilation
         cuda_src = '%s/CUDA/%s_kernels.cu' % (self.get_jit_dir(), soname)
         cuda_target = '%s/%s_kernels_cu' % (self.get_jit_dir(), soname)
@@ -95,22 +111,6 @@ class CompilerOPS(configuration['compiler'].__class__):
                 [src_o, cuda_o],
                 debug=configuration['debug-compiler']
             )
-
-    def jit_compile(self, soname):
-        self.target = str(self.get_jit_dir().joinpath(soname))
-        self.ops_src = '%s/%s_ops.cpp' % (self.get_jit_dir(), soname)
-        self.cache_dir = self.get_codepy_dir().joinpath(soname[:7])
-
-        # Typically we end up here
-        # Make a suite of cache directories based on the soname
-        self.cache_dir.mkdir(parents=True, exist_ok=True)
-        try:
-            with open(self.ops_src, 'r') as f:
-                self.code = f.read()
-        except FileNotFoundError:
-            warning("The file %s isn't present" % self.ops_src)
-        else:
-            self.cuda_compiler(soname)
 
 
 class CUDADeviceCompiler(CompilerOPS):
--- a/devito/ops/compiler.py
+++ b/devito/ops/compiler.py
@@ -117,7 +117,6 @@ class CompilerOPS(configuration['compiler'].__class__):
 
 
 class CUDADeviceCompiler(CompilerOPS):
-    CXX = os.environ.get('CC', 'nvcc')
 
     def __init__(self, *args, **kwargs):
         super(CUDADeviceCompiler, self).__init__(*args, **kwargs)
@@ -146,14 +145,11 @@ class CUDADeviceCompiler(CompilerOPS):
         self.include_dirs = include_dirs.split(' ')
 
     def __lookup_cmds__(self):
-        self.CC = 'nvcc'
+        self.CC = os.environ.get('CC', 'nvcc')
+        self.CXX = os.environ.get('CXX', 'nvcc')
 
 
 class CudaHostCompiler(CompilerOPS):
-    CC = os.environ.get('CC', 'gcc')
-    CXX = os.environ.get('CXX', 'g++')
-    MPICC = os.environ.get('MPICC', 'mpicc')
-    MPICXX = os.environ.get('MPICXX', 'mpicxx')
 
     def __init__(self, *args, **kwargs):
         super(CudaHostCompiler, self).__init__(*args, **kwargs)
@@ -172,11 +168,10 @@ class CudaHostCompiler(CompilerOPS):
         self.library_dirs.append(os.path.join(self._cuda_install_path, 'lib64'))
         self.library_dirs.append(os.path.join(self._ops_install_path, 'c', 'lib'))
 
-        self.libraries.append('cudart')
-        self.libraries.append('ops_cuda')
+        self.libraries.extend(['cudart', 'ops_cuda'])
 
     def __lookup_cmds__(self):
-        self.CC = 'gcc'
-        self.CXX = 'g++'
-        self.MPICC = 'mpicc'
-        self.MPICXX = 'mpicxx'
+        self.CC = os.environ.get('CC', 'gcc')
+        self.CXX = os.environ.get('CXX', 'g++')
+        self.MPICC = os.environ.get('MPICC', 'mpicc')
+        self.MPICXX = os.environ.get('MPICXX', 'mpicxx')
--- a/devito/ops/compiler.py
+++ b/devito/ops/compiler.py
@@ -97,9 +97,6 @@ class CompilerOPS(configuration['compiler'].__class__):
             )
 
     def jit_compile(self, soname):
-        from devito import clear_cache
-        clear_cache()
-
         self.target = str(self.get_jit_dir().joinpath(soname))
         self.ops_src = '%s/%s_ops.cpp' % (self.get_jit_dir(), soname)
         self.cache_dir = self.get_codepy_dir().joinpath(soname[:7])
--- a/devito/ops/compiler.py
+++ b/devito/ops/compiler.py
@@ -1,5 +1,8 @@
 import os
 import subprocess
+import warnings
+
+from codepy.jit import compile_from_string
 from devito.logger import warning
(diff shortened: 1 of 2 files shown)


devito/ops/compiler.py
```diff
@@ -30,3 +53,130 @@ def prepare_ops(self, soname, ccode, hcode):
...
+                debug=True
+            )
+
+    def jit_compile(self, soname):
+        from devito import clear_cache
+        clear_cache()
+
+        self.target = str(self.get_jit_dir().joinpath(soname))
+        self.ops_src = '%s/%s_ops.cpp' % (self.get_jit_dir(), soname)
+        self.cache_dir = self.get_codepy_dir().joinpath(soname[:7])
```
btrb: can you remember me why `[:7]` here

maelso: Just a temporary directory. Its name is created using the beginning of the hash.
Equals to devito/compile.py.

devito/ops/compiler.py
```diff
@@ -30,3 +53,130 @@ def prepare_ops(self, soname, ccode, hcode):
...
+            )
+            cuda_host_compiler.link_extension(
+                '%s%s' % (self.target, cuda_host_compiler.so_ext),
+                [src_o, cuda_o],
+                debug=configuration['debug-compiler']
+            )
+
+    def jit_compile(self, soname):
+        from devito import clear_cache
+        clear_cache()
```
FabioLuporini: leftover

devito/ops/operator.py
```diff
@@ -83,7 +85,7 @@ def _specialize_iet(self, iet, **kwargs):
             have the same number of dimensions"
 
         self._headers.append(namespace['ops_define_dimension'](dims[0]))
-        self._includes.append('stdio.h')
+        self._includes += ['stdio.h', 'ops_seq.h']
```
FabioLuporini: extend

FabioLuporini: this was nitpicking, but do you know why is it conceptually more sound?

devito/ops/compiler.py
```diff
@@ -30,3 +53,130 @@ def prepare_ops(self, soname, ccode, hcode):
...
+            with open(self.ops_src, 'r') as f:
+                self.code = f.read()
+        except FileNotFoundError:
+            warning("The file %s isn't present" % self.ops_src)
+        else:
+            self.cuda_compiler(soname)
+
+
+class CUDADeviceCompiler(CompilerOPS):
+    CXX = os.environ.get('CC', 'nvcc')
```
mloubout: CXX should be `os.environ.get('CXX'...)` so think should be CC = ...

devito/ops/compiler.py
```diff
@@ -30,3 +53,130 @@ def prepare_ops(self, soname, ccode, hcode):
...
+        else:
+            raise ValueError("Unknown NVIDIA architecture, select: \
+                Fermi, Kepler, Pascal or Volta")
+
+        self.ldflags = []
+
+        include_dirs = '%s %s/c/include' % (self.get_jit_dir(), self._ops_install_path)
+        self.include_dirs = include_dirs.split(' ')
+
+    def __lookup_cmds__(self):
```
mloubout: to be removed, this overwrites the `environ.get`

devito/ops/compiler.py
```diff
@@ -30,3 +53,130 @@ def prepare_ops(self, soname, ccode, hcode):
...
+        self.include_dirs.append(os.path.join(self._cuda_install_path, 'include'))
+        self.include_dirs.append(os.path.join(self._ops_install_path, 'c', 'include'))
+
+        self.library_dirs.append(os.path.join(self._cuda_install_path, 'lib64'))
+        self.library_dirs.append(os.path.join(self._ops_install_path, 'c', 'lib'))
+
+        self.libraries.append('cudart')
+        self.libraries.append('ops_cuda')
+
+    def __lookup_cmds__(self):
```
mloubout: to be removed, this overwrites the environ.get

devito/ops/operator.py
maelso: I found that INPLACE_ADD fails for numpy array.

devito/ops/compiler.py
maelso: Just removing causes an error.
I'll move `environ.get` to `__lookup_cmds__`.

devito/ops/compiler.py
maelso: Same as above.

devito/ops/compiler.py
maelso: makes sense, but It should be CXX = os.environ.get('CXX', 'nvcc').
CXX is required because we'he `kwargs['cpp'] = True` in the CompilerOPS.

maelso:
> looking good.
> 
> How much of this is inherited from @vincepandolfo 's work?

28%

devito/ops/compiler.py
maelso: @mloubout do you agree with that?

devito/ops/compiler.py
```diff
@@ -30,3 +53,122 @@ def prepare_ops(self, soname, ccode, hcode):
         else:
             warning("Couldn't find OPS_INSTALL_PATH \
                 environment variable, please check your OPS installation")
+
+    def cuda_compiler(self, soname):
```
FabioLuporini: nitpicking: about about calling `jit_cuda` ? `cuda_compiler` makes me think of something else. Otherwise, `compile_cuda`...

FabioLuporini: also, this should probably be a private method? as the actual API is still `jit_compile`

devito/ops/compiler.py
```diff
@@ -30,3 +53,122 @@ def prepare_ops(self, soname, ccode, hcode):
...
+    def __lookup_cmds__(self):
+        self.CC = os.environ.get('CC', 'nvcc')
+        self.CXX = os.environ.get('CXX', 'nvcc')
+
+
+class CudaHostCompiler(CompilerOPS):
+
+    def __init__(self, *args, **kwargs):
+        super(CudaHostCompiler, self).__init__(*args, **kwargs)
+        self.o_ext = '.o'
```
FabioLuporini: when is this used?

devito/ops/compiler.py
```diff
@@ -30,3 +53,122 @@ def prepare_ops(self, soname, ccode, hcode):
...
+            warning("The file %s isn't present" % self.ops_src)
+        else:
+            self.cuda_compiler(soname)
+
+
+class CUDADeviceCompiler(CompilerOPS):
+
+    def __init__(self, *args, **kwargs):
+        super(CUDADeviceCompiler, self).__init__(*args, **kwargs)
+        self.o_ext = '.o'
```
FabioLuporini: when is this used?

devito/ops/operator.py
```diff
@@ -95,7 +97,7 @@ def _specialize_iet(self, iet, **kwargs):
             have the same number of dimensions"
 
         self._headers.append(namespace['ops_define_dimension'](dims[0]))
-        self._includes.append('stdio.h')
+        self._includes.extend(['stdio.h', 'ops_seq.h'])
```
FabioLuporini: `ops_seq.h` ? what is it used for?

devito/ops/operator.py
```diff
@@ -107,5 +109,7 @@ def hcode(self):
         return ''.join(str(kernel) for kernel in self._ops_kernels)
 
     def _compile(self):
+        self._includes.append('%s.h' % self._soname)
         if self._lib is None:
             self._compiler.prepare_ops(self._soname, str(self.ccode), str(self.hcode))
```
FabioLuporini: is `prepare_ops` really necessary? can we not keep everything in  `jit_copmpile` ?

devito/ops/operator.py
maelso: It is responsible for implementing the OPS API calls. It is necessary to have it.

devito/ops/operator.py
maelso: Well noted.  Now we can move this call to the first line of jit_compile, keeping the `def prepare_ops` to have a clean code. 
I think we can rename `prepare_ops`.  Something like `translate_ops`.

devito/ops/compiler.py
maelso: I agree. `compile_cuda` is better than `cuda_compiler`.
About the private - **I'm not sure**, but doesn't that go against Python's philosophy? (In this case)

devito/ops/compiler.py
maelso: In _.../codepy/jit.py_, line 399, in `compile_from_string(...)`
`suffix = toolchain.o_ext`

maelso: Without `self.o_ext`, we've an error:
`AttributeError: 'CUDADeviceCompiler' object has no attribute 'o_ext'`

devito/ops/compiler.py
maelso: Same case as mentioned below.

devito/ops/compiler.py
```diff
@@ -30,3 +53,122 @@ def prepare_ops(self, soname, ccode, hcode):
...
+                [src_o, cuda_o],
+                debug=configuration['debug-compiler']
+            )
+
+    def jit_compile(self, soname):
+        self.target = str(self.get_jit_dir().joinpath(soname))
+        self.ops_src = '%s/%s_ops.cpp' % (self.get_jit_dir(), soname)
+        self.cache_dir = self.get_codepy_dir().joinpath(soname[:7])
+
+        # Typically we end up here
```
FabioLuporini: either we support the jit backdoor, or you drop this comment

devito/ops/compiler.py
FabioLuporini: oki

devito/ops/operator.py
FabioLuporini: ok

devito/ops/operator.py
FabioLuporini: sounds good to me. Just make it a private method, since it's never going to be used by the outside world

devito/ops/compiler.py
FabioLuporini: it's never going to be used by the outside world right? so it should be private

I don't think this has anything to do with python and its philosophy. Maybe you wanna elaborate what you meant?

maelso: Python does not have the concept of private methods because the language has the philosophy: – “_we are all adults consenting here_”. (The phrase commonly used).

Since I do not have the habit of using private things, my goal here is just to align this decision to future similar cases. 

I will change to private.

devito/ops/operator.py
maelso: Ok. I will change to private.

devito/ops/compiler.py
```diff
@@ -9,10 +12,46 @@
...
+        self.cache_dir.mkdir(parents=True, exist_ok=True)
+        try:
+            with open(self.ops_src, 'r') as f:
+                self.code = f.read()
+        except FileNotFoundError:
+            warning("The file %s isn't present" % self.ops_src)
+        else:
+            self.__compile_cuda(soname)
+
+    def __translate_ops(self, soname, ccode, hcode):
```
FabioLuporini: single `_` to be used to denote private

devito/ops/compiler.py
```diff
@@ -30,3 +69,106 @@ def prepare_ops(self, soname, ccode, hcode):
         else:
             warning("Couldn't find OPS_INSTALL_PATH \
                 environment variable, please check your OPS installation")
+
+    def __compile_cuda(self, soname):
```
FabioLuporini: single `_` to be used to denote private

FabioLuporini: hang on, this is weird: why doesn't this method belong to `CUDADeviceCompiler`

FabioLuporini:
@maelso I'm merging this now, but:

- there is still one answered comment, please take a look. I suspect the cuda_compile method should be moved the CUDA subclass

- all commits about the OPS backend should start with "ops: <commit msg>", as per instructions [here](https://github.com/opesci/devito/blob/master/CONTRIBUTING.md). This applies to @btrb @Gabrielsvc @vmickus too

PR #940, opened 2019-09-18
arch: Add _detect_isa method for ARM

jaimesouza:


devito/archinfo.py
```diff
@@ -151,6 +151,7 @@ class Platform(object):
 
     def __init__(self, name, **kwargs):
         self.name = name
+        self.known_isas = []
```
FabioLuporini: this should be a class attribute, not an instance attribute. And it can even be a tuple, instead of a list, since the ISAs are fixed on each architecture.

devito/archinfo.py
```diff
@@ -188,23 +189,26 @@ def simd_items_per_reg(self, dtype):
...
 
     def _detect_isa(self):
-        known_isas = ['cpp', 'sse', 'avx', 'avx2', 'avx512']
-        for i in reversed(known_isas):
-            if any(j.startswith(i) for j in get_cpu_info()['flags']):
-                # Using `startswith`, rather than `==`, as a flag such as 'avx512'
-                # appears as 'avx512f, avx512cd, ...'
-                return i
-        return 'cpp'
+        self.known_isas = ['cpp', 'sse', 'avx', 'avx2', 'avx512']
```
FabioLuporini: of course, the comment above applies here too

devito/archinfo.py
```diff
@@ -187,24 +187,30 @@ def simd_items_per_reg(self, dtype):
...
                 # Using `startswith`, rather than `==`, as a flag such as 'avx512'
                 # appears as 'avx512f, avx512cd, ...'
                 return i
         return 'cpp'
 
 
+class Intel64(Cpu64):
+
+    def _detect_isa(self):
+        self.known_isas = ('cpp', 'sse', 'avx', 'avx2', 'avx512')
```
FabioLuporini: no, this doesn't make any sense conceptually

I don't know the reason you didn't pass continuous integration, I looked at it quickly but definitely it was unrelated to your changes. Can you rollback to the previous commit and push again?

you should also learn and use `git rebase` at this point,  because now we have 6 commits for what is supposed to be one simple change. [Here](https://git-scm.com/docs/git-rebase) is a good starting point

we can talk about it maybe even later

FabioLuporini: I will investigate the continuous investigation failure (@mloubout : actually, it was really weird stuff in `test_symbol_cache`......)

jaimesouza: rollback done

devito/archinfo.py
```diff
@@ -187,24 +187,32 @@ def simd_items_per_reg(self, dtype):
...
                 # appears as 'avx512f, avx512cd, ...'
                 return i
         return 'cpp'
 
 
+class Intel64(Cpu64):
+
+    known_isas = ('cpp', 'sse', 'avx', 'avx2', 'avx512')
+
+    def _detect_isa(self):
```
FabioLuporini: you can drop this. It's gonna automatically pick the one of the superclass

devito/archinfo.py
```diff
@@ -187,24 +187,32 @@ def simd_items_per_reg(self, dtype):
...
+        return super(Intel64, self)._detect_isa()
+
+
 class Arm(Cpu64):
-    pass
+
+    known_isas = ('fp', 'asimd', 'asimdrdm')
+
+    def _detect_isa(self):
```
FabioLuporini: you can drop this. It's gonna automatically pick the one of the superclass

jaimesouza: Oh, that's right. I'm gonna do that.

FabioLuporini:
OK @jaimesouza many thanks now it's perfect. I'll merge it soon !

FabioLuporini:
Merged.

PR #942, opened 2019-09-24
Add mpi_mode in json

georgebisbas:
To add some plotting fixes for benchmark.py

### DIFF
--- a/benchmarks/user/benchmark.py
+++ b/benchmarks/user/benchmark.py
@@ -408,9 +408,14 @@ def get_ob_bench(problem, resultsdir, parameters):
             else:
                 devito_params['nt'] = 1
 
-            devito_params['np'] = MPI.COMM_WORLD.size
-            devito_params['rank'] = MPI.COMM_WORLD.rank
             devito_params['mpi'] = configuration['mpi']
+
+            if configuration['mpi']:
+                devito_params['np'] = MPI.COMM_WORLD.size
+                devito_params['rank'] = MPI.COMM_WORLD.rank
+            else:
+                devito_params['np'] = 1
+                devito_params['rank'] = 0
 
             return '_'.join(['%s[%s]' % (k, v) for k, v in devito_params.items()])
 
--- a/benchmarks/user/benchmark.py
+++ b/benchmarks/user/benchmark.py
@@ -293,7 +293,6 @@ def plot(problem, **kwargs):
     flop_ceils = kwargs.pop('flop_ceil')
     point_runtime = kwargs.pop('point_runtime')
     autotune = kwargs['autotune']
-
     arch = kwargs['arch']
     space_order = "[%s]" % ",".join(str(i) for i in kwargs['space_order'])
     time_order = kwargs['time_order']
@@ -322,8 +321,8 @@ def plot(problem, **kwargs):
              if len(set(dict(j)[i] for j in gflopss)) > 1]
 
     # Filename
-    figname = "%s_dim%s_so%s_to%s_arch[%s]_bkend[%s]_at[%s]pdf" % (
-        problem, shape, space_order, time_order, arch, backend, autotune
+    figname = "%s_arch[%s]_shape%s_so%s_to%s_bkend[%s]_at[%s]" % (
+        problem, arch, shape, space_order, time_order, backend, autotune
     )
 
     # Legend setup. Do not plot a legend if there's no variation in performance
@@ -409,12 +408,9 @@ def get_ob_bench(problem, resultsdir, parameters):
             else:
                 devito_params['nt'] = 1
 
-            if configuration['mpi']:
-                devito_params['np'] = MPI.COMM_WORLD.size
-                devito_params['rank'] = MPI.COMM_WORLD.rank
-            else:
-                devito_params['np'] = 1
-                devito_params['rank'] = 0
+            devito_params['np'] = MPI.COMM_WORLD.size
+            devito_params['rank'] = MPI.COMM_WORLD.rank
+            devito_params['mpi'] = configuration['mpi']
 
             return '_'.join(['%s[%s]' % (k, v) for k, v in devito_params.items()])
 
--- a/benchmarks/user/benchmark.py
+++ b/benchmarks/user/benchmark.py
@@ -321,8 +321,8 @@ def plot(problem, **kwargs):
              if len(set(dict(j)[i] for j in gflopss)) > 1]
 
     # Filename
-    figname = "%s_arch[%s]_shape%s_so%s_to%s_bkend[%s]_at[%s]" % (
-        problem, arch, shape, space_order, time_order, backend, autotune
+    figname = "%s_shape%s_so%s_to%s_arch[%s]_bkend[%s]_at[%s]" % (
+        problem, shape, space_order, time_order, arch, backend, autotune
     )
 
     # Legend setup. Do not plot a legend if there's no variation in performance


benchmarks/user/benchmark.py
```diff
@@ -412,6 +412,7 @@ def param_string(self, params):
             if configuration['mpi']:
                 devito_params['np'] = MPI.COMM_WORLD.size
                 devito_params['rank'] = MPI.COMM_WORLD.rank
+                devito_params['mpi_mode'] = configuration['mpi']
```
FabioLuporini: I suppose we should have a corresponding dummy value in the else branch otherwise things might fail when parsing the filename, depending on whether MPI was used or not

georgebisbas: I thought that if MPI is not used, there is no need to 'pollute' the filename. That is because in fact it is not a standalone parameter but a sub-parameter.

FabioLuporini: but based on what I wrote above -- does it work if you are **not** using MPI ?

also, nitpicking: can we simply call it "mpi" ?

FabioLuporini:
I think you messed up with your branches, I see some skewing stuff here now

benchmarks/user/benchmark.py
georgebisbas: Yes, let's go to mpi. It worked without MPI. I tested my bench scripts and also the plotting. The issue that persists is how to pass to plot functionality the info of the file it needs to parse. It should be done in a different way than it is now. Now you have to do `mpirun -n 2` but two threads plot together the same pic.

georgebisbas:
> I think you messed up with your branches, I see some skewing stuff here now

Checkout out and pushed accidentally, fixed.

FabioLuporini:
> Now you have to do mpirun -n 2 but two threads plot together the same pic.

I don't think there's any point in plotting a roofline when using MPI.

One plots a scalability graph when using MPI, typically.

FabioLuporini:
let me ask again: does this patch really work if you run not just without `mpirun ...`, but rather *without DEVITO_MPI= ... * ?

also: I see you've rebased the commit history, that's good. However, the commit message "Revert ..." doesn't make much sense

georgebisbas:
> > Now you have to do mpirun -n 2 but two threads plot together the same pic.
> 
> I don't think there's any point in plotting a roofline when using MPI.
> 
> One plots a scalability graph when using MPI, typically.

Yes, but we now aim to LinePlotter for scalability, not Roofline.

georgebisbas:
> let me ask again: does this patch really work if you run not just without `mpirun ...`, but rather *without DEVITO_MPI= ... * ?
> 
> also: I see you've rebased the commit history, that's good. However, the commit message "Revert ..." doesn't make much sense


If you dont touch DEVITO_MPI and run python benchmark.py bench ....
we produce: 
`acoustic_arch[unknown]_shape[50,50,50]_nbpml[10]_tn[250]_so[2]_to[2]_dse[advanced]_dle[advanced]_at[aggressive]_nt[1]_np[1]_rank[0]`
If you do `mpirun -n 2 python benchmark.py bench -P ` you get the message `It seems that you're running over MPI with 2 processes, but DEVITO_MPI is unset. Setting `DEVITO_MPI=basic`...
`
you get 2 files: ` acoustic_arch[unknown]_shape[50,50,50]_nbpml[10]_tn[250]_so[2]_to[2]_dse[advanced]_dle[advanced]_at[aggressive]_nt[1]_np[2]_rank[0]_mpi[basic]`
and rank[1] as well

FabioLuporini:
What I think I'm asking is whether the *plotter* is capable of parsing two different filenames -- one with `_mpi[...]` (longer) and the other without ?

FabioLuporini:
yeah probably it's all good as long as one runs the plotter the same way bench was run (ie, both w/ or w/o DEVITO_MPI...). RIght?

georgebisbas:
SO just to sum up, what is needed, that can be probably be done in other PR is opescibench tweaking to parse according to the filename. This need emerges for one reason. RooflinePlotter and LinePlotter inherit from Plotter in general, and the one is needed for weak, while the other for strong scaling. SO the arguments of Roofline should be irrelevant to MPI, wihle this is not true for LinePlotter.

georgebisbas:
@FabioLuporini I was thinking of even removing np and rank from no-mpi-produced jsons. As long as we are going to use them form Rooflines we do not need this info.

FabioLuporini:
I think perhaps the filename template could be fixed once and for all -- ie always with mpimode, mpirank, etc. Then it's up to whatever uses it decide what to do with the extrapolated information (so the RooflinePlotter could/will drop mpimode, mpirank etc on the floor ?)

what would you say?

georgebisbas:
> I think perhaps the filename template could be fixed once and for all -- ie always with mpimode, mpirank, etc. Then it's up to whatever uses it decide what to do with the extrapolated information (so the RooflinePlotter could/will drop mpimode, mpirank etc on the floor ?)
> 
> what would you say?

Well, it could be. I will make these changes and then it could be merged, based on the hypothesis that Roofline does not care for nothing that has (rank>0 && np>1).

benchmarks/user/benchmark.py
```diff
@@ -409,12 +408,9 @@ def param_string(self, params):
             else:
                 devito_params['nt'] = 1
 
-            if configuration['mpi']:
-                devito_params['np'] = MPI.COMM_WORLD.size
-                devito_params['rank'] = MPI.COMM_WORLD.rank
-            else:
-                devito_params['np'] = 1
-                devito_params['rank'] = 0
+            devito_params['np'] = MPI.COMM_WORLD.size
```
FabioLuporini: this fails if one doesn't have mpi4py installed
https://github.com/opesci/devito/blob/master/devito/mpi/distributed.py#L39

georgebisbas: Sure, thanks for testing.

benchmarks/user/benchmark.py
```diff
@@ -322,8 +321,8 @@ def plot(problem, **kwargs):
              if len(set(dict(j)[i] for j in gflopss)) > 1]
 
     # Filename
-    figname = "%s_dim%s_so%s_to%s_arch[%s]_bkend[%s]_at[%s]pdf" % (
-        problem, shape, space_order, time_order, arch, backend, autotune
+    figname = "%s_arch[%s]_shape%s_so%s_to%s_bkend[%s]_at[%s]" % (
```
FabioLuporini: there was a reason ``arch`` was next to ``bkend`` -- they both are "platform information", while ``shape/dim`` (good changing dim->shape), ``so``, ``to`` are about the problem executed

FabioLuporini:
> based on the hypothesis that Roofline does not care for nothing that has (rank>0 && np>1).

wdym by "doesn't care" exactly?

anyway, is this now good to go, basically?

benchmarks/user/benchmark.py
georgebisbas: Good, then we should change the order in produced .json files. In any case I was trying to have json and fignames homogeneous.

georgebisbas:
@btrb has offered to double-check it but I think it is GTG now.

btrb:
so, if i'm properly reproducing what's expected, i'm doing (for the MPI case, let's say) is:

```
DEVITO_MPI=1 DEVITO_OPENMP=1 mpirun -n 2 python benchmarks/user/benchmark.py bench -P acoustic -x 1
DEVITO_MPI=1 DEVITO_OPENMP=1 mpirun -n 2 python benchmarks/user/benchmark.py plot -P acoustic --max-bw 12.8 --flop-ceil 80 linpack
```

which returns the following img in `devito/results`:

acoustic_arch unknown _shape 50,50,50 _so 2 _to 2 _bkend core _at aggressive

i'm not sure if that makes any sense.. please let me know if i should apply any corrections

FabioLuporini:
merged.

PR #946, opened 2019-10-01
Adds one to dimension max-value in create_ops_par_loop

Gabrielsvc:
Since ops iterates a loop until the value is < x_Max and devito iterates
a loop until the value is <= x_Max, we lose a value when going to OPS.
This commit solves this adding one to each dimension max-value, making
OPS iterate the ops_par_loop until the value is < x_Max + 1

### DIFF
--- a/devito/ops/transformer.py
+++ b/devito/ops/transformer.py
@@ -224,6 +224,8 @@ def create_ops_par_loop(trees, ops_kernel, parameters, block, name_to_ops_dat,
     for tree in trees:
         if isinstance(tree, IterationTree):
             for bounds in [it.bounds() for it in tree]:
+                # from IPython import embed
+                # embed()
                 it_range.extend(bounds[:-1]+(bounds[-1]+1,))
 
     range_array = Array(
--- a/devito/ops/transformer.py
+++ b/devito/ops/transformer.py
@@ -224,7 +224,7 @@ def create_ops_par_loop(trees, ops_kernel, parameters, block, name_to_ops_dat,
     for tree in trees:
         if isinstance(tree, IterationTree):
             for bounds in [it.bounds() for it in tree]:
-                it_range.extend(bounds)
+                it_range.extend(bounds[:-1]+(bounds[-1]+1,))
 
     range_array = Array(
         name='%s_range' % ops_kernel.name,
--- a/devito/ops/transformer.py
+++ b/devito/ops/transformer.py
@@ -4,7 +4,7 @@ import numpy as np
 from sympy import Mod
 from sympy.core.numbers import Zero
 
-from devito import Eq
+from devito import Eq, TimeFunction
 from devito.ir.equations import ClusterizedEq
 from devito.ir.iet.nodes import Call, Callable, Expression, IterationTree
 from devito.ir.iet.visitors import FindNodes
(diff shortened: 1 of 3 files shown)


devito/ops/transformer.py
```diff
@@ -224,7 +224,7 @@ def create_ops_par_loop(trees, ops_kernel, parameters, block, name_to_ops_dat,
     for tree in trees:
         if isinstance(tree, IterationTree):
             for bounds in [it.bounds() for it in tree]:
-                it_range.extend(bounds)
+                it_range.extend(bounds[:-1]+(bounds[-1]+1,))
```
btrb: i wonder if there might be a way to access it's upper bound so even if the tuple position eventually changes, we're still referring to the same parameter. 
otherwise, looks good to me

FabioLuporini:
34 commits for a 1-line change + test ?

Gabrielsvc:
Ops, my bad! I merged instead of rebasing, doing the rebase now!

devito/ops/transformer.py
Gabrielsvc: Found a solution! Check it out on the last commit =)

devito/ops/transformer.py
```diff
@@ -221,10 +221,13 @@ def create_ops_fetch(f, name_to_ops_dat, time_upper_bound):
 def create_ops_par_loop(trees, ops_kernel, parameters, block, name_to_ops_dat,
                         accessible_origin, par_to_ops_stencil, dims):
     it_range = []
+    devito_to_ops_indexer = 1
```
FabioLuporini: I think this entire piece of code is incredibly and uselessly complicated...

btw, what do you need the ``it_ranges`` for? can you provide examples?

devito/ops/transformer.py
```diff
@@ -221,10 +221,13 @@ def create_ops_fetch(f, name_to_ops_dat, time_upper_bound):
...
     for tree in trees:
         if isinstance(tree, IterationTree):
-            for bounds in [it.bounds() for it in tree]:
-                it_range.extend(bounds)
+            for it in tree:
+                it_range.extend((it.dimensions[TimeFunction._time_position].extreme_min,
```
FabioLuporini: extending a list with a tuple? 😮

FabioLuporini: why can you not just use ``it.symbolic_max`` etc ? btw , call it ``i``, not ``it``, for homogeneity. We never use `it` for Iterations

devito/ops/transformer.py
Gabrielsvc: In OPS we define an ´OPS_Kernel_range´ variable which contains all the minimal and maximal values in each dimension. We currently are getting those from devito with `it_ranges`. So in a simple example where we add one to each point of a grid in each time step we would have:
`int OPS_Kernel_0_range[4] = {x_m, x_M + 1, y_m, y_M + 1}`

devito/ops/transformer.py
Gabrielsvc: Well, I was really overthinking it! thanks for the feedback! About the tuple, I was using it as the code was but it doesn't really make a difference, switched it to a list :)

FabioLuporini:
@maelso @btrb should I proceed?

maelso:
> @maelso @btrb should I proceed and merge this?

Looks good.

FabioLuporini:
merged

PR #947, opened 2019-10-02
Fixed a bug in the guard function

ArthurGiannotta:
Fixed a bug in the guard function regarding the usage of ConditionalDimension with conditions that cannot be evaluated in python. Also added a test that verifies if this bug is fixed.

### DIFF
--- a/devito/archinfo.py
+++ b/devito/archinfo.py
@@ -187,15 +187,11 @@ class Platform(object):
 
 class Cpu64(Platform):
 
-    def _detect_isa(self):
-        return 'cpp'
-
-
-class Intel64(Cpu64):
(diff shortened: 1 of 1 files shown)


devito/ir/clusters/algorithms.py
```diff
@@ -388,7 +388,11 @@ def guard(clusters):
                 guards = {}
                 for d in e.conditionals:
                     condition = guards.setdefault(d.parent, [])
-                    condition.append(d.condition or CondEq(d.parent % d.factor, 0))
+
```
FabioLuporini: nitpicking: unnecessary blank line -- I don't think the line `condition = ...` should be logically separated from the if-then below

PR #950, opened 2019-10-02
Ops/specializing iet

btrb:
This PR fixes the case in which there isn't affine trees to be converted into OPS.

### DIFF
--- a/devito/ops/operator.py
+++ b/devito/ops/operator.py
@@ -30,7 +30,7 @@ class OperatorOPS(Operator):
     def _specialize_iet(self, iet, **kwargs):
         warning("The OPS backend is still work-in-progress")
 
-        # If there is no iteration trees nor affine trees, 
+        # If there is no iteration trees nor affine trees,
         # then there is no loop to be optimized using OPS.
         iteration_tree = retrieve_iteration_tree(iet, mode='normal')
         if not len(iteration_tree) or not find_affine_trees(iet):
--- a/devito/ops/operator.py
+++ b/devito/ops/operator.py
@@ -30,10 +30,12 @@ class OperatorOPS(Operator):
     def _specialize_iet(self, iet, **kwargs):
         warning("The OPS backend is still work-in-progress")
 
-        # If there is no iteration tree, then there is no loop to be optimized using OPS.
+        # If there is no iteration trees nor affine trees, 
+        # then there is no loop to be optimized using OPS.
         iteration_tree = retrieve_iteration_tree(iet, mode='normal')
-        if not len(iteration_tree):
+        if not len(iteration_tree) or not find_affine_trees(iet):
             return iet
+
         time_upper_bound = iteration_tree[0].dimensions[TimeFunction._time_position]\
             .extreme_max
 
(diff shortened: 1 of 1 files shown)


devito/ops/operator.py
```diff
@@ -30,10 +30,12 @@ def __init__(self, *args, **kwargs):
     def _specialize_iet(self, iet, **kwargs):
         warning("The OPS backend is still work-in-progress")
 
-        # If there is no iteration tree, then there is no loop to be optimized using OPS.
+        # If there is no iteration trees nor affine trees,
+        # then there is no loop to be optimized using OPS.
         iteration_tree = retrieve_iteration_tree(iet, mode='normal')
-        if not len(iteration_tree):
+        if not len(iteration_tree) or not find_affine_trees(iet):
```
FabioLuporini: `find_affine_tree` is already called at line 83

you should hoist it at the top of this method and use the result when needed

btrb: ok, just re-commited considering this comment. thank you

devito/ops/operator.py
```diff
@@ -44,7 +48,7 @@ def _specialize_iet(self, iet, **kwargs):
         # Extract all symbols that need to be converted to ops_dat
         dims = []
         to_dat = set()
-        for section, trees in find_affine_trees(iet).items():
+        for (_, trees) in affine_trees:
```
FabioLuporini: no need for the parentheses here

devito/ops/operator.py
```diff
@@ -78,7 +82,7 @@ def _specialize_iet(self, iet, **kwargs):
 
         # Generate ops kernels for each offloadable iteration tree
         mapper = {}
-        for n, (section, trees) in enumerate(find_affine_trees(iet).items()):
+        for n, (_, trees) in enumerate(affine_trees):
```
FabioLuporini: here needed, instead

devito/ops/operator.py
```diff
@@ -30,10 +30,14 @@ def __init__(self, *args, **kwargs):
...
         warning("The OPS backend is still work-in-progress")
 
-        # If there is no iteration tree, then there is no loop to be optimized using OPS.
         iteration_tree = retrieve_iteration_tree(iet, mode='normal')
-        if not len(iteration_tree):
+        affine_trees = find_affine_trees(iet).items()
+
+        # If there is no iteration trees nor affine trees,
+        # then there is no loop to be optimized using OPS.
+        if not len(iteration_tree) or not len(affine_trees):
```
FabioLuporini: you don't need `len`

FabioLuporini: in fact, I think this condition should just be

```
if not affine_trees: 
     ....
```

devito/ops/operator.py
btrb: ok

devito/ops/operator.py
btrb: ok, fixing

btrb:
closing in favor of [#953](https://github.com/opesci/devito/pull/953)

PR #951, opened 2019-10-02
ops: Avoided unwanted redeclaration when an Array is passed as a Call argument

maelso:
**This implementation is avoiding to create a third type for _scope.**

For the backend OPS, we are creating an Array and passing it as an argument to a Call, generating unnecessary alignment.
```
int OPS_Kernel_0_range[4] = {x_m, x_M, y_m, y_M};
for (int time = time_m, t0 = (time)%(3), t1 = (time + 1)%(3), t2 = (time + 2)%(3); .....)
{
   int OPS_Kernel_0_range[4] __attribute__((aligned(64)));
   ops_par_loop(..., (int *)OPS_Kernel_0_range, ...);
}
```

allocation should be True if the associated data was/is/will be allocated directly from Python.

### DIFF
--- a/devito/ops/types.py
+++ b/devito/ops/types.py
@@ -9,6 +9,14 @@ from devito.ops.utils import namespace
 
 class Array(basic.Array):
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._allocation = kwargs.get('allocation')
+
+    @property
+    def _mem_external(self):
+        return self._allocation == 'external'
+
     @property
     def _C_typedata(self):
         if isinstance(self.dtype, str):
--- a/devito/ops/types.py
+++ b/devito/ops/types.py
@@ -9,14 +9,6 @@ from devito.ops.utils import namespace
 
 class Array(basic.Array):
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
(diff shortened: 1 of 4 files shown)


devito/ops/types.py
```diff
@@ -9,6 +9,14 @@
 
 class Array(basic.Array):
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._allocation = kwargs.get('allocation')
```
FabioLuporini: do you really need an extra attribute? can you not reuse `scope` ?

maelso: In your comment on slack, I understood that I should not use scope:

> I’m not particularly fond of the _scope attribute for reasons that we can discuss another day, so if we managed to find a way not to touch it, I’d prefer

As we have:
```
self._scope = kwargs.get('scope', 'heap')
assert self._scope in ['heap', 'stack']
```
I would need to change scope

maelso: I'll change that `assert` and overwrite _mem_external

FabioLuporini: > In your comment on slack, I understood that I should not use scope

right, but if the alternative is to add a further attribute, then we can just use scope I'd say?

maelso: ahh. Ok! Changed.

FabioLuporini:
closing as superseded by #961

PR #952, opened 2019-10-04
Numpy allocator

navjotk:
Add a new type of allocator backed by an externally allocated numpy array

@btrb

### DIFF
--- a/devito/data/allocators.py
+++ b/devito/data/allocators.py
@@ -149,6 +149,7 @@ class PosixAllocator(MemoryAllocator):
 
 
 class GuardAllocator(PosixAllocator):
+
     """
     Memory allocator based on ``posix`` functions. The allocated memory is
     aligned to page boundaries.  Additionally, it allocates extra memory
@@ -306,7 +307,41 @@ class NumaAllocator(MemoryAllocator):
         return self._node == 'local'
 
 
-class NumpyAllocator(MemoryAllocator):
+class ExternalAllocator(MemoryAllocator):
+
+    """
+    External allocator for definig Function symbols with data loaded by the
+    user, without copying it to Devito's allocated memory.
+
+    Parameters
+    ----------
+    numpy_array : numpy.array
+        Array of data.
+
+    Returns
+    -------
+    pointer, None
+        The first element of the tuple is the reference that can be used to
+        access the data.
+
+    Notes
+    -------
+    * Use ExternalAllocator and pass a reference to the external memory when
+      creating a function. This Function will now use this memory as it's f.data.
+
+    * If the data present in this external memory is valuable, provide a noop
+      initialiser, or else Devito will reset it to 0.
+
+    Example
+    --------
+    >>> shape = (2, 2)
+    >>> space_order = 0
+    >>> numpy_array = np.ones(shape, dtype=np.float32)
+    >>> g = Grid(shape)
+    >>> f = Function(name='f', space_order=space_order, grid=g,
+    ...      allocator=ExternalAllocator(numpy_array), initializer=lambda x: None)
+    """
+
     def __init__(self, numpy_array):
         self.numpy_array = numpy_array
 
--- a/devito/data/allocators.py
+++ b/devito/data/allocators.py
@@ -306,6 +306,21 @@ class NumaAllocator(MemoryAllocator):
         return self._node == 'local'
 
 
+class NumpyAllocator(MemoryAllocator):
+    def __init__(self, numpy_array):
+        self.numpy_array = numpy_array
+
+    def alloc(self, shape, dtype):
+        assert shape == self.numpy_array.shape, \
+            "Provided array has shape %s. Expected %s" %\
+            (str(self.numpy_array.shape), str(shape))
+        assert dtype == self.numpy_array.dtype, \
+            "Provided array has dtype %s. Expected %s" %\
+            (str(self.numpy_array.dtype), str(dtype))
+
+        return (self.numpy_array, None)
+
+
 ALLOC_GUARD = GuardAllocator(1048576)
 ALLOC_FLAT = PosixAllocator()
 ALLOC_KNL_DRAM = NumaAllocator(0)
--- a/devito/data/allocators.py
+++ b/devito/data/allocators.py
@@ -310,36 +310,37 @@ class NumaAllocator(MemoryAllocator):
 class ExternalAllocator(MemoryAllocator):
 
     """
-    External allocator for definig Function symbols with data loaded by the
-    user, without copying it to Devito's allocated memory.
+    An ExternalAllocator is used to assign pre-existing user data to Functions.
+    Thus, Devito does not allocate any memory.
 
     Parameters
     ----------
-    numpy_array : numpy.array
-        Array of data.
-
-    Returns
-    -------
-    pointer, None
-        The first element of the tuple is the reference that can be used to
-        access the data.
+    array : array-like
+        Any object exposing the buffer interface, such as a numpy.ndarray.
 
     Notes
     -------
     * Use ExternalAllocator and pass a reference to the external memory when
-      creating a function. This Function will now use this memory as it's f.data.
+      creating a Function. This Function will now use this memory as its f.data.
 
     * If the data present in this external memory is valuable, provide a noop
       initialiser, or else Devito will reset it to 0.
 
     Example
     --------
+    >>> from devito import Grid, Function
+    >>> from devito.data.allocators import ExternalAllocator
+    >>> import numpy as np
     >>> shape = (2, 2)
-    >>> space_order = 0
     >>> numpy_array = np.ones(shape, dtype=np.float32)
     >>> g = Grid(shape)
-    >>> f = Function(name='f', space_order=space_order, grid=g,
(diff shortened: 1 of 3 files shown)


btrb:
has the first_touch been changed yet ? i noticed you're feeding a new `initializer` parameter. let me reproduce the tests here .
thank you

navjotk:
I don't think that's necessary. The initialisation code doesn't go into the first touch branch unless first touch is set to True, and it is False by default. I utilised the initialiser branch to supply a noop initialiser that bypasses the fill(0.) initialiser.

btrb:
got it !

devito/data/allocators.py
```diff
@@ -306,6 +306,21 @@ def put_local(self):
         return self._node == 'local'
 
 
+class NumpyAllocator(MemoryAllocator):
```
FabioLuporini: docstring? basically, to be used whenever the user wants to hand ownership of some data to devito, right?

tests/test_data.py
```diff
@@ -1231,6 +1232,32 @@ def test_numpy_c_contiguous():
...
 
 
+@skipif(['yask', 'ops'])
+def test_numpy_allocator():
+    shape = (2, 2)
+    space_order = 0
+    numpy_array = np.ones(shape, dtype=np.float32)
+    g = Grid(shape)
+    f = Function(name='f', space_order=space_order, grid=g,
+                 allocator=NumpyAllocator(numpy_array), initializer=lambda x: None)
```
FabioLuporini: I assume that this `initializer = lambda x: None` is here to avoid an extra memset ? so this is basically implicit to whenever NumpyAllocator is used?

devito/data/allocators.py
navjotk: It is the opposite of handing ownership of data to Devito in my opinion? 

If I'm the user, and let's say I used some proprietary file reader that insists on allocating its own memory, I need to be able to define a Devito (Function) symbol that is backed by the data I already have loaded, without copying it to Devito's allocated memory. Does that make sense?

tests/test_data.py
navjotk: It is more than just "extra". If the data in `numpy_array` has meaning - e.g. initial conditions for the PDE I want to solve, `fill(0.)` will erase the data that was valuable. So in this case, the memset is harmful as well as being extra. 

Yes I'm using a noop initialiser here to bypass that memset. 

I don't think anything is implicit, unless I misunderstood what you mean. 

Use the `NumpyAllocator` when you want to use externally allocated memory as the backend for a `devito.Function`. 

If the data in that memory is important, use a noop `initializer` to preserve it, else, Devito's default behaviour will be to overwrite it with 0.

devito/data/allocators.py
FabioLuporini: yep, it does. Though note that without a `clear_cache` the data will stick around even if out of scope in user code, since it's now attached to the NumpyAllocator itself

tests/test_data.py
FabioLuporini: oh, OK, makes sense

navjotk:
@btrb Can I please leave it to you to add a docstring?

I am also having second thoughts on whether `NumpyAllocator` is the best name for this class, since nothing here assumes Numpy. It could be allocated by any means, even a malloc, now that I think about it. Right? Perhaps better named `ExternalAllocator`. 

This could be used for:
- Implementing memcpy in Devito (reference to a year ago when I was struggling to find an efficient library implementation of memcpy)
- Implementing operators in Devito that can even compute one timestep while doing a "copy", hence potentially eliminating redundant copies
- (speculation) The example I used above where the initial conditions are being loaded from a file. 
- (speculation) Another variation of the example above, where the data loaded is not initial conditions (example `m`), but if the file reader insists on allocating it's own memory, using a symbol like this, we could avoid doing a copy. If there is performance to be gained out of it. 
- (speculation) Some weird scenario on a platform like KNL (that's where I've seen it, but this might also be true on GPUs) where one might have specialised malloc-like functions available to allocate on fast memory vs slow memory. In the long term these scenarios should probably use their specialised allocators, but this allows rapid prototyping. 
- In fact, external use of Devito (e.g. from Julia), will benefit from this in most cases.


API:
- Use `NumpyAllocator` and pass a reference to the external memory when creating a function. This Function will now use this memory as it's `f.data`. 
- If the data present in this external memory is valuable, provide a noop initialiser, else Devito will reset it to 0. 

I agree that it's not the most straightforward API, but this is an advanced use case anyway so it's not unfair to expect the user to work a bit :) 

Suggestions?

FabioLuporini:
I think I agree on pretty much everything

The only thing maybe is that I envisage some hardware-specialized (NUMA nodes, KNL, GPU...) allocators to be available directly in devito (there are a few already available). But yeah, everything makes sense.

FabioLuporini:
I think all this information should go in the docstring.

btrb:
ofc @navjotk , thank you.

devito/data/allocators.py
```diff
@@ -306,6 +307,55 @@ def put_local(self):
...
+    """
+    External allocator for definig Function symbols with data loaded by the
+    user, without copying it to Devito's allocated memory.
+
+    Parameters
+    ----------
+    numpy_array : numpy.array
+        Array of data.
+
+    Returns
```
FabioLuporini: A class doesn't return anything

devito/data/allocators.py
```diff
@@ -306,6 +307,55 @@ def put_local(self):
         return self._node == 'local'
 
 
+class ExternalAllocator(MemoryAllocator):
+
+    """
+    External allocator for definig Function symbols with data loaded by the
```
FabioLuporini: typo (definig)

navjotk:
I'm not good at this. How can it be made better?

devito/data/allocators.py
```diff
@@ -306,6 +307,55 @@ def put_local(self):
...
+    user, without copying it to Devito's allocated memory.
```
FabioLuporini: the point is: devito doesn't allocate any memory

So, how about

```
An ExternalAllocator is used to assign pre-existing user data to Functions. Thus, Devito does not allocate any memory. 
```

devito/data/allocators.py
```diff
@@ -306,6 +307,55 @@ def put_local(self):
...
+
+    Parameters
+    ----------
+    numpy_array : numpy.array
```
FabioLuporini: array-like  instead of numpy.array ?

array instead of numpy_array ?

and then in the description, "Any object exposing the buffer interface, such as a numpy.ndarray"

devito/data/allocators.py
FabioLuporini: this should be dropped

devito/data/allocators.py
```diff
@@ -306,6 +307,55 @@ def put_local(self):
...
+    Returns
+    -------
+    pointer, None
+        The first element of the tuple is the reference that can be used to
+        access the data.
+
+    Notes
+    -------
+    * Use ExternalAllocator and pass a reference to the external memory when
+      creating a function. This Function will now use this memory as it's f.data.
```
FabioLuporini: Function

it's -> its

devito/data/allocators.py
```diff
@@ -306,6 +307,55 @@ def put_local(self):
...
+
+    * If the data present in this external memory is valuable, provide a noop
+      initialiser, or else Devito will reset it to 0.
+
+    Example
+    --------
+    >>> shape = (2, 2)
+    >>> space_order = 0
```
FabioLuporini: no need for this extra variable. In fact, no need to specify space_order at all

devito/data/allocators.py
```diff
@@ -306,6 +307,55 @@ def put_local(self):
...
+    >>> numpy_array = np.ones(shape, dtype=np.float32)
+    >>> g = Grid(shape)
+    >>> f = Function(name='f', space_order=space_order, grid=g,
+    ...      allocator=ExternalAllocator(numpy_array), initializer=lambda x: None)
```
FabioLuporini: at this point you could add another couple of lines

```
>>> f.data[0, 1] = 2
>>> numpy_array
[[ 1 2 ]
 [ 1 1 ]]
```

and could also add this file to `test_docstrings.py`

devito/data/allocators.py
btrb: ofc, this makes no sense! fixing

devito/data/allocators.py
btrb: fixed

devito/data/allocators.py
btrb: updated

devito/data/allocators.py
btrb: thank you, fixed

devito/data/allocators.py
btrb: thank you, fixed

devito/data/allocators.py
btrb: ok, updated

devito/data/allocators.py
btrb: oki

FabioLuporini:
Thanks both. Merged.

PR #954, opened 2019-10-04
Optimize MPI computation/communication overlap

FabioLuporini:
Improve the `DEVITO_MPI=full` mode such that:

- one OpenMP thread continuously prods the asynchronous progress engine until all comms have finished
- dynamic scheduling is used in place of static if we are prodding

tests updated and enriched accoringly

EDIT: yes, this makes the code go faster according to my measurements

### DIFF
--- a/devito/dle/parallelizer.py
+++ b/devito/dle/parallelizer.py
@@ -3,9 +3,9 @@ import os
 
 import numpy as np
 import cgen as c
-from sympy import Function, Or
+from sympy import Function, Or, Not
 
-from devito.ir import (Call, Conditional, Block, Expression, List, Prodder,
+from devito.ir import (Conditional, Block, Expression, List, Prodder, While,
                        FindSymbols, FindNodes, Return, COLLAPSED, Transformer,
                        IsPerfectIteration, retrieve_iteration_tree, filter_iterations)
 from devito.symbolics import CondEq
@@ -50,13 +50,21 @@ class ParallelRegion(Block):
         return (self.nthreads,)
 
 
-class SingleThreadProdder(Conditional, Prodder):
+class ThreadedProdder(Conditional, Prodder):
 
     _traversable = []
 
     def __init__(self, prodder):
+        # Atomic-ize any single-thread Prodders in the parallel tree
         condition = CondEq(Function('omp_get_thread_num')(), 0)
-        then_body = Call(prodder.name, prodder.arguments)
+
+        # Prod within a while loop until all communications have completed
+        # In other words, the thread delegated to prodding is entrapped for as long
+        # as it's required
+        prod_until = Not(Function(prodder.name)(*[i.name for i in prodder.arguments]))
+        then_body = List(header=c.Comment('Entrap thread until comms have completed'),
+                         body=While(prod_until))
+
         Conditional.__init__(self, condition, then_body)
         Prodder.__init__(self, prodder.name, prodder.arguments, periodic=prodder.periodic)
 
@@ -118,9 +126,8 @@ class Ompizer(object):
         partree = Transformer(mapper).visit(partree)
         return partree
 
-    def _make_atomic_prodders(self, partree):
-        # Atomic-ize any single-thread Prodders in the parallel tree
-        mapper = {i: SingleThreadProdder(i) for i in FindNodes(Prodder).visit(partree)}
+    def _make_threaded_prodders(self, partree):
+        mapper = {i: ThreadedProdder(i) for i in FindNodes(Prodder).visit(partree)}
         partree = Transformer(mapper).visit(partree)
         return partree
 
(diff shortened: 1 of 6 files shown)


devito/dle/parallelizer.py
```diff
@@ -50,13 +50,21 @@ def functions(self):
         return (self.nthreads,)
 
 
-class SingleThreadProdder(Conditional, Prodder):
+class ThreadedProdder(Conditional, Prodder):
```
mloubout: Never quite sure what the difference between Prob and Prod is so gonna assume you chose the right one ^^

devito/ir/iet/nodes.py
```diff
@@ -521,6 +534,31 @@ def write(self):
         return []
 
 
+class While(Node):
```
mloubout: So this is only a compiler level tool that is not accessible/usable at user level right?

devito/ir/iet/nodes.py
```diff
@@ -521,6 +534,31 @@ def write(self):
...
+
+    """
+    Implement a while-loop.
```
mloubout: why `-`?

devito/ir/iet/nodes.py
```diff
@@ -521,6 +534,31 @@ def write(self):
...
+
+    Parameters
+    ----------
+    condition : sympy.Function or sympy.Relation or bool
```
mloubout: Function as a condition? like what?

devito/ir/iet/nodes.py
```diff
@@ -521,6 +534,31 @@ def write(self):
...
+        The whie-loop body.
+    """
+
+    is_While = True
+
+    _traversable = ['body']
+
+    def __init__(self, condition, body=None):
+        self.condition = condition
+        self.body = as_tuple(body)
```
mloubout: private _?

devito/ir/iet/visitors.py
```diff
@@ -299,6 +312,15 @@ def visit_Iteration(self, o):
 
         return handle
 
+    def visit_While(self, o):
+        condition = ccode(o.condition)
+        if o.body:
+            body = flatten(self._visit(i) for i in o.children)
+            return c.While(condition, body)
+        else:
+            # Hack: cgen doesn't support body-less while-loops, i.e. `while(...);`
```
mloubout: That's cute :D

tests/test_mpi.py
```diff
@@ -820,26 +820,46 @@ def test_poke_progress(self):
...
-        # Make sure `pokempi0` is within the inner Iteration over blocks
-        assert len(tree) == 4
+        # Make sure `pokempi0` is the last node within the inner Iteration over blocks
+        assert len(tree) == 2
         assert len(tree.root.nodes[0].nodes) == 2
-        call = tree.root.nodes[0].nodes[0]
+        call = tree.root.nodes[0].nodes[1]
         assert call.name == 'pokempi0'
         assert call.arguments[0].name == 'msg0'
+        try:
```
mloubout: I don't really like try except in tests

tests/test_mpi.py
```diff
@@ -1531,7 +1551,6 @@ def test_adjoint_F_no_omp(self, shape, kernel, space_order, nbpml, save,
 
 
 if __name__ == "__main__":
-    from devito import configuration
```
mloubout: ?

devito/ir/iet/nodes.py
FabioLuporini: correct

devito/ir/iet/nodes.py
FabioLuporini: because I used 'for-loop' in the `Iteration` docstring. Probably because it just went out that way. I can kill the hyphen

devito/ir/iet/nodes.py
FabioLuporini:
```
while(! poke_comms(...))
```

`poke_comms` returns true/false depending on whether all comms have completed

devito/ir/iet/nodes.py
FabioLuporini: in theory yes, in practice no, because none of the IET classes use private attribute

it's a massive rewrite which we don't wanna do now, and perhaps never?

tests/test_mpi.py
FabioLuporini: now imported at the top, so redundant

tests/test_mpi.py
FabioLuporini: mm agreed, but do you have alternatives?

mloubout:
```
if configuration['openmp']
   ...
else:
  ...
```

or make `isopenmp` a fixture maybe.

devito/ir/iet/nodes.py
mloubout: Oh ok, for later maybe then, was just saying to have consistent private over the codebase but I get it's tidious

tests/test_mpi.py
FabioLuporini: given the assert within try/except (see `assert configuration ...`),  it's practically the same....
... BUT you're right, your way is definitely clearer. I'll change it

devito/ir/iet/nodes.py
mloubout: Just me being picky, people use way too many hyphens for my taste :D

FabioLuporini:
merged

PR #955, opened 2019-10-09
Create OPS call to set data from GPU

vmickus:
- Create call to ops_dat_fetch_data.
- Merged ops_dat_fetch_data and ops_dat_set_data into one method called create_ops_memory_call.
- Access to array is now using nth dimension access.
- Updated tests to this new approach.
- Create class to handle type cast.

### DIFF
--- a/devito/ops/types.py
+++ b/devito/ops/types.py
@@ -1,6 +1,7 @@
 import numpy as np
 import sympy
 
+from devito.symbolics.printer import ccode
 import devito.types.basic as basic
 
 from devito.tools import dtype_to_cstr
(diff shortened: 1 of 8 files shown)


devito/ops/types.py
```diff
@@ -147,3 +147,18 @@ def __init__(self, name, *args, **kwargs):
...
+
+class TypeCast(basic.Object):
+    """
+        A node encapsulating a cast to a raw C type.
+    """
+
+    def __init__(self, name, dtype, cast_to_ctype):
+        self.cast_to_ctype = cast_to_ctype
+        self.dtype = dtype
+        self.name = str(name)
```
maelso: Error during apply():
```
File "devito/types/basic.py", line 974, in _arg_defaults
    if callable(self.value):
AttributeError: 'TypeCast' object has no attribute 'value'
```

vmickus: I think it's time to add a test for a basic apply in our `test_ops.py`.

I'll try to add a simple time marching test in this PR. :crossed_fingers:

maelso: The CI checks will fail with an apply test. How do you intend to manage that?

vmickus: Right, so no testing this for now.

vmickus: Fixed with the last commit

vmickus:
As discussed in the OPS channel, this approach does not handle memory transfer of the halo data. 

So I'll close this PR and work in a new one, which works for the domain and halo regions.

Issue #726, opened 2018-11-29
opescibench no installed but required for benchmark and accuracy notebook


mloubout:
`opescibench` needs to be added to the docker image

FabioLuporini:
shall we add this to ``requirements-optional.txt`` ?

georgebisbas:
Also binded with PR #957

georgebisbas:
Relative to #1085

mloubout:
#1098 should fix that

FabioLuporini:
closing

PR #957, opened 2019-10-10
#Hacktoberfest - testing

georgebisbas:
Add opescibench testing to CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -94,6 +94,13 @@ install:
       pip install -e .;
       conda list;
     fi
+  - if [[ $INSTALL_TYPE == 'conda' ]]; then
+      cd ../;
+      git clone https://github.com/opesci/opescibench.git;
+      cd opescibench;
+      pip install -e .;
+      cd ../devito;
+    fi
   - if [[ "$MPI_INSTALL" == '1' ]]; then
       pip install -r requirements-optional.txt;
     fi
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -127,7 +127,14 @@ script:
   - if [[ $RUN_EXAMPLES == 'True' ]]; then python examples/cfd/example_diffusion.py; fi
   - if [[ $RUN_EXAMPLES == 'True' ]]; then py.test examples/cfd/example_diffusion.py; fi
   - if [[ $RUN_EXAMPLES == 'True' ]]; then py.test examples/seismic/elastic/elastic_example.py; fi
-  - if [[ $RUN_EXAMPLES == 'True' ]]; then py.test examples/seismic/viscoelastic/viscoelastic_example.py; fi
+  - if [[ $RUN_EXAMPLES == 'True' ]]; then py.test examples/seismic/viscoelastic/viscoelastic_example.py;     fi
+  - if [[ $RUN_EXAMPLES == 'True' ]]; then git clone https://github.com/opesci/opescibench.git ../;
```
mloubout: Oww no this doesn't go there this goes above with the other installs

FabioLuporini:
no this is all wrong

this is the way we pip install from github repos : https://github.com/opesci/devito/blob/master/environment.yml#L30

and even then, Im not sure we want to systematically bring in opescibench

perhaps to be added to the optional requirements ?

also, see this: https://github.com/opesci/devito/issues/726

mloubout:
>this is the way we pip install from github repos :

That's for conda, his change is the one pip install requires,  I don't mind having it this way, my only concern is that "optionals" is growing and containing a lot of unrelated packages but not sure how to split all of that clearly.

.travis.yml
```diff
@@ -128,6 +128,17 @@ script:
...
   - if [[ $RUN_EXAMPLES == 'True' ]]; then py.test examples/seismic/elastic/elastic_example.py; fi
   - if [[ $RUN_EXAMPLES == 'True' ]]; then py.test examples/seismic/viscoelastic/viscoelastic_example.py; fi
+
+  - if [[ $RUN_EXAMPLES == 'True' ]]; then python benchmarks/user/benchmark.py test -P tti -so 4 -d 20 20 20 -n 5; fi
+  - if [[ $RUN_EXAMPLES == 'True' ]]; then python benchmarks/user/benchmark.py test -P acoustic; fi
+  - if [[ $RUN_EXAMPLES == 'True' ]]; then python benchmarks/user/benchmark.py bench -P acoustic; fi
+  - if [[ $RUN_EXAMPLES == 'True' ]]; then python benchmarks/user/benchmark.py bench -P elastic; fi
+  - if [[ $RUN_EXAMPLES == 'True' ]]; then python benchmarks/user/benchmark.py bench -P tti; fi
+  - if [[ $RUN_EXAMPLES == 'True' ]]; then python benchmarks/user/benchmark.py bench -P viscoelastic; fi
+  - if [[ $RUN_EXAMPLES == 'True' ]]; then python benchmarks/user/benchmark.py plot -P acoustic --flop-ceil 80 linpack --max-bw 12.8; fi
```
FabioLuporini: these 4 tests only test that the execution doesn't fail in python. They don't test whether an image is produced, etc. So I'm not sure how useful are they

FabioLuporini:
For example, in Firedrake they have a large set of [demos](https://github.com/firedrakeproject/firedrake/tree/master/demos)

and to test them all, they have [this](https://github.com/firedrakeproject/firedrake/blob/master/tests/demos/test_demos_run.py)

the firedrake approach is scalable, automated, and guaranteed to be tested locally (I always forget to test the examples locally); the one in this PR is none of this

.travis.yml
georgebisbas: Well this was my purpose for having a first naive coupling between devito and opescibench.

georgebisbas:
> explicitly (issuing a new command just like we do for the mpi tests) or, better, from python directly (ie, importing the individual modules (bench, run, etc). In the latter case, we won't test the argument

Can you show me an example for that? I am a bit confused..

georgebisbas:
> Besides the fact that this PR is lacking the corresponding changes in pipelines.

That will happen in #960 ?

FabioLuporini:
> That will happen in #960 ?
no

FabioLuporini:
> Can you show me an example for that? I am a bit confused..

an example for ... what exactly? 

I was just suggesting to move everything inside a `test_benchmark.py`, because the current way leads to an explosion of lines as more and more use cases are added (think of adding a further problem next to acoustic/tti/elastic/...)

georgebisbas:
> > Can you show me an example for that? I am a bit confused..
> 
> an example for ... what exactly?
> 
> I was just suggesting to move everything inside a `test_benchmark.py`, because the current way leads to an explosion of lines as more and more use cases are added (think of adding a further problem next to acoustic/tti/elastic/...)

I mean an example of that `(issuing a new command just like we do for the mpi tests)`... I did not quite got it when I checked `test_mpi.py`.

FabioLuporini:
@georgebisbas 

ah, [this](https://github.com/opesci/devito/blob/master/tests/conftest.py#L301) is what I had to write to implement `pytest.mark.parallel`: this marker is something I have written specifically for devito, it doesn't exist in pytest. So, with `pytest.mark.parallel` (as you can appreciate if you look at that link), we end up running a [new python command](https://github.com/opesci/devito/blob/master/tests/conftest.py#L353)

what I meant to say is that if I were to write `test_benchmark.py`, I'd have both standard tests (to check the individual modes/problems) and, just like I do for running mpi tests, explicit runs `python benchmark.py ...` to make sure that the argument parsing works as expected (otherwise, I wouldn't know how to test that...there might be a programmatic way, but I don't know it yet)

georgebisbas:
Closing temporarily, related to PR #960 and issue #726.

PR #958, opened 2019-10-10
Remove PyZFP from environment.yml and requirements.txt, move to optional

navjotk:
I'm sure we're all happy to see it go :)


FabioLuporini:
uncontroversial, merging. Thanks

Issue #817, opened 2019-04-29
Revise install instructions


ggorman:
* Update README.md with revised instillation instructions - docker listed as the first option.
* Add/update troubleshooting section to capture gcc+mac issue, Python version, etc
* ~~Update pipeline tests to test docker on Ubuntu, Mac, Windows.~~ Misguided
* ~~Update Devito instillation documentation under docs to point at README.md (ie enforce single source of truth).~~ Done

navjotk:
The first task on this: "Update pipeline tests to test docker on Ubuntu, Mac, Windows." is on @tmbgreaves

tmbgreaves:
Could you fill me in a bit more on this please? I don't think I entirely understand the question.

Docker ought to be host-independent - ie, the container the tests run in shouldn't have any knowledge of the system the container is running on, and indeed I don't think we get an information about what the host OS is for the Pipelines runs. The container itself as far as I know doesn't have any ability to have Windows or OS X on the inside -- OS X licensing would preclude this as I understand it, and I have no idea how you would implement Windows-in-a-container :-)

ggorman:
Perhaps we are talking past each other. According to https://azure.microsoft.com/en-gb/services/devops/pipelines/ we should be able to "Run in parallel on Linux, macOS and Windows".

navjotk:
> Docker ought to be host-independent - ie, the container the tests run in shouldn't have any knowledge of the system the container is running on

I think the idea is to verify this assumption so that we can s/ought to/is, s/shouldn't/doesn't

tmbgreaves:
> Perhaps we are talking past each other. According to https://azure.microsoft.com/en-gb/services/devops/pipelines/ we should be able to "Run in parallel on Linux, macOS and Windows".

I have no idea how they do that, but will figure it out!

tmbgreaves:
> > Docker ought to be host-independent - ie, the container the tests run in shouldn't have any knowledge of the system the container is running on
> 
> I think the idea is to verify this assumption so that we can s/ought to/is, s/shouldn't/doesn't

If that breaks, Docker is in trouble :-)

tmbgreaves:
Looking into this a bit more, from:

https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml

[...]

PR #959, opened 2019-10-10
Make Docker the Default Install Method #Hacktoberfest

cc-a:
Made Docker the preferred installation method over anaconda in the project README.


mloubout:
why?

georgebisbas:
> why?

Agreed with @ggorman ?

FabioLuporini:
for completeness: I basically won't be able to provide any kind of support should there be any issues related to docker install 

we have pretty much 0 installation issues, I'm not sure I understand all this stress about using Docker over miniconda. What do I gain from this? What are the advantages?

super dumb question: if I'm using docker, and want to try a change to the devito code (e.g., to the notebooks), is it straightforward just as if I were using conda ?

navjotk:
> if I'm using docker, and want to try a change to the devito code (e.g., to the notebooks), is it straightforward just as if I were using conda ?

I was under the impression that we've been through this entire discussion before and agreed to do this. As was @ggorman .

FabioLuporini:
OK no problem

ggorman:
As @navjotk said - we had this whole discussion before and the conclusion was docker should be the first option. Some of the rational:
- it ‘just works’ for those that just want to take devito for a quick test drive. From the OSX broken compiler problem to rouge env variables, building devito doesn’t always go as smoothly as we’d like, as clearly illustrated on Thursday. These user experience install issues are not appearing as GitHub issues but we have seen a lot on slack - and that’s from those that even go to slack (we have plenty of indirect evidence of this) so using number of GitHub issues as the sole indicator is a classic example of selection bias. 
- the docker route *must* always be fully supported because we use it for pipelines/CI and tutorials/training (binderhub). 
- not everyone is a fan of conda. There is a large community of Enthought users in geo. Our first move should not to tell them they have to change python distribution just to install our code.

mloubout:
>the docker route must always be fully supported because we use it for pipelines/CI

These don't use the same docker build though. Pipelines has its own that is not the one from the install instructions so the recommended install is not actually tested while conda and docker (*not docker-compose* which is another all story) are indeed tested. So I would only move towards this PR if it is actually tested and explicitly explained in the readme that this is *docker-compose* not and actual available docker image of devito.

FabioLuporini:
We agreed that this PR is on hold until someone takes responsibility of the issue that @mloubout pointed out above. 

@ggorman how do you want to proceed?

cc-a:
Apologies, was away for a few days. Happy to take this further.

Regarding the docker vs docker-compose I'd make the case for providing instructions for just docker as:
- The volumes mounted in docker-compose.yml are convenient for development work but not needed or necessarily desirable for the end user. 
- The use of `docker-compose up` requires users to edit the yml to use a different port for the notebook server. 
- Removes a dependency

The equivalent docker invocations would be something like:
```
docker build -f docker/Dockerfile -t devito .
docker run -p 8888:8888 --rm -it devito
docker run -p 8888:8888 --rm -it devito /tests
docker run -p 8888:8888 --rm -it devito /bin/bash
```

If you're happy to move forward on that I'll look at adding some testing too.

georgebisbas:
Closing temporarily, related to issue #817

navjotk:
This can be reopened now that the Docker installation method is tested.

README.md
```diff
@@ -53,6 +38,26 @@ docker-compose up devito
...
+image is built. Subsequent calls to docker-compose will be much
+quicker.
+
+Alternatively you can use the Conda package manager for installation of
+the necessary software dependencies. Please
+install either [Anaconda](https://www.continuum.io/downloads) or
+[Miniconda](https://conda.io/miniconda.html) using the instructions
+provided at the download links. You will need the Python 3 version.
+
+To install Devito, including examples, tests and tutorial notebooks,
```
FabioLuporini: this may be dropped

navjotk:
Closing as superceded by #1116

PR #960, opened 2019-10-10
Opescibench on Azure Pipelines #Hacktoberfest

cc-a:
Installation only of opescibench added to Dockerfile.pipelines so far. Preparation for running the benchmarks on Azure Pipelines

### DIFF
--- a/Dockerfile.pipelines
+++ b/Dockerfile.pipelines
@@ -132,10 +132,10 @@ RUN if [ $installWithPip == "true" ] ; then \
       fi ; \
       if [ $RUN_EXAMPLES == "true" ] ; then \
         bash scripts/create_ipyparallel_mpi_profile.sh; \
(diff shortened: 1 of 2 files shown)


Dockerfile.pipelines
```diff
@@ -132,6 +132,9 @@ RUN if [ $installWithPip == "true" ] ; then \
       fi ; \
       if [ $RUN_EXAMPLES == "true" ] ; then \
         bash scripts/create_ipyparallel_mpi_profile.sh; \
+        git clone https://github.com/opesci/opescibench.git /tmp/opescibench ; \
```
mloubout: I know it an extra block but I'd like this in a separate `if` to separate from MPI installs

cc-a:
Was that what you had in mind for the if block? Otherwise added a runStep for execution of the benchmark with the acoustic problem.

georgebisbas:
Closed temporarily, related to issue #726.

PR #961, opened 2019-10-10
Scheduler: Fix redeclaration

maelso:
This solution checks all scopes created on the stack to see if the object has already been declared.

The issue occurs when an Array is passed as an argument to a Call.

This PR is a cleaner alternative to [PR-951](https://github.com/opesci/devito/pull/951).


FabioLuporini:
this is hackish...

it works OK with your use case

but what if you had

```
if 
    <LocalObject X>
    ...
else
    <LocalObject X>
    ...
```
with this patch, X would only be declared once

also, it would have been more pythonic as 
```
if obj in flatten(self.stack.values()): 
    return
```

maelso:
> this is hackish...
> 
> it works OK with your use case
> 
> but what if you had
> 
> ```
> if 
>     <LocalObject X>
>     ...
> else
>     <LocalObject X>
>     ...
> ```
> 
> with this patch, X would only be declared once
> 
> also, it would have been more pythonic as
> 
> ```
> if obj in flatten(self.stack.values()): 
>     return
> ```

I didn't understand this "if/else". There is already an `if obj.is_LocalObject:`


P.S. I can't answer your comments, I think you're selecting "Add single comment" instead of "Start a review> comment"

devito/ir/iet/scheduler.py
```diff
@@ -204,6 +204,9 @@ def push_object_on_stack(self, scope, obj):
         """Define an Array or a composite type (e.g., a struct) on the stack."""
         handle = self.stack.setdefault(scope, OrderedDict())
 
+        if obj in flatten(self.stack.values()):
+            return
+
```
maelso:
```
if 
    <LocalObject X>
    ...
else
    <LocalObject X>
```

> with this patch, X would only be declared once
We have a big challenge!

FabioLuporini: yeah now u get it ? perhaps we should add an MFE to cover this case too, and then think about a proper rearchitecture

FabioLuporini:
> P.S. I can't answer your comments, I think you're selecting "Add single comment" instead of "Start a review> comment"

sorry I didn't know that ! Thanks

devito/ir/iet/scheduler.py
maelso: For now, I think creating an `def push_array_on_stack` will avoid this problem.  

> and then think about a proper rearchitecture

Ok.

maelso:
```
if 
    <Array X>
    ...
else
    <Array X>
```
Can we assume that above case doesn't occur for array?

FabioLuporini: mmhh... I'm fine with the current solution, but I'd also comment the code such that it points out that the above is unsupported 

in an ideal world, with this PR you would also:

- Add an MFE to test_ir.py::TestIETConstruction, marked as [xfail](http://doc.pytest.org/en/latest/skipping.html), that triggers the issue
- File an issue on github that explains the situation, the current implementation, and points to the relevant MFE

maelso: Ok. I'm working on that.

maelso:
Hey @fabio

What is the possibility of merge this PR temporarily ?  These past few weeks I've been busy with some internal tasks and the declarator's rearchitecture is still pending. I ask this because @vitor is currently having difficulty testing his solutions without this fix.

FabioLuporini:
If I understand correctly:

- there is a bug in master that leads to redeclaring the same object in the same scope more than once
- with this PR, Arrays won't be declared twice in the same scope. 
- but LocalObjects may still be declared twice ?

I *might* consider merging this PR but before it needs to be equipped with tests. At least one test covering the bug that this PR fixes, and potentially another (x-failing ?) test(s) for anything that this PR doesn't cover. Probably to be added inside the fairly obsolete class `test_operator.py::TestDeclarator`

> I ask this because @vitor is currently having difficulty testing his solutions without this fix.

I don't think this is an issue at all, one can always work a few steps ahead of master (it happens to me quite frequently too)

FabioLuporini:
Reminder: this needs rebasing and tests before landing

tests/test_ops.py
```diff
@@ -255,11 +255,24 @@ def test_create_ops_block(self, equation, expected):
...
+    ])
+    def test_single_declaration(self, equation, declaration):
+        grid = Grid((5, 5))
+        u = TimeFunction(name='u', grid=grid)  # noqa
+        op = Operator(eval(equation))
+
+        occurrences = [i for i in str(op.ccode).split('\n') if declaration in i]
+
+        assert len(occurrences) == 1
+
```
maelso: I am still looking at how to create a failure case.

devito/ir/iet/scheduler.py
maelso: Hey @FabioLuporini 

The declarations are working fine, as seen at: `test_conditional_declarations`

FabioLuporini: good, but what I meant is that we would need a test for `core` devito, since you're changing `core` devito here

maelso: Ok, my bad.

maelso: `test_conditional_declarations` created into `test_operator.py`
But I keep into `test_ops.py` too, because `test_operator.py` has `skipif(['yask', 'ops'])`, and the Array used in the backend ops comes from `devito/ops/types`.

tests/test_operator.py
```diff
@@ -950,6 +953,37 @@ def test_incomplete_override(self):
 
 class TestDeclarator(object):
 
+    def test_conditional_declarations(self):
```
FabioLuporini: this is test is much more complicated than strictly necessary

why do you need all that ListInitializer stuff?

also, if we have this test, can we not drop the one in OPS? they are identical...

maelso: Just explaining a little more, this test is to ensure that the case below doesn't occur (undeclared array inside else)
```
if (x < 3)
{
  int s2d_name_4pt[8] = {0, 0, 0, 1, 1, 0, 1, 1};
}
else
{
  s2d_name_4pt[8] = {0, 0, 0, 1, 1, 0, 1, 1};
}
```

> why do you need all that ListInitializer stuff?

When the array isn't initialized an **AssertionError** occurs during `ccode` generation.
```
devito/ir/iet/visitors.py:259: in visit_Conditional
    then_body = c.Block(self._visit(o.then_body))
self = <cgen.Block object at 0x7f6dcb2c9450>, contents = (None,)
for item in contents:
   assert isinstance(item, Generable)
```


> also, if we have this test, can we not drop the one in OPS? they are identical...

They are identical! The reason is that the backend OPS overwrites Array in devito.ops.types.Array. It is possible that future changes in ...ops.Array may break something from this fix. Once `test_operator.py` has skipif(['yask', 'ops']), backend ops stay uncovered.

FabioLuporini: I still don't think that 30 almost identical lines are duplicated across two different files

>  It is possible that future changes in ...ops.Array may break something from this fix

if you further change ops.Array, then that's when you should write a new test, I feel

FabioLuporini: nitpicking: this is an advanced test, so it should be put at the bottom of the `TestDeclarator` class, not at the top

tests/test_operator.py
```diff
@@ -950,6 +953,37 @@ def test_incomplete_override(self):
...
+        list_initialize = Expression(ClusterizedEq(Eq(
+            stencil_array,
+            ListInitializer(list(itertools.chain(*accesses)))
+        )))
+
+        iet = Conditional(x < 3, list_initialize, list_initialize)
+
+        parameters = derive_parameters(iet, True)
+        iet = iet_insert_decls(iet, parameters)
+        iet = iet_insert_casts(iet, parameters)
```
FabioLuporini: do we really need this?

tests/test_operator.py
```diff
@@ -950,6 +953,37 @@ def test_incomplete_override(self):
...
 class TestDeclarator(object):
 
+    def test_conditional_declarations(self):
+        accesses = [[0, 0], [0, 1], [1, 0], [1, 1]]
+        dims = len(accesses[0])
+        pts = len(accesses)
+        stencil_name = 's%dd_%s_%dpt' % (dims, 'name', pts)
+        stencil_array = Array(
+            name=stencil_name,
+            dimensions=(DefaultDimension(name='len', default_value=dims * pts),),
```
FabioLuporini: do we really need DefaultDimension ?

tests/test_operator.py
```diff
@@ -950,6 +953,37 @@ def test_incomplete_override(self):
 
 class TestDeclarator(object):
 
+    def test_conditional_declarations(self):
+        accesses = [[0, 0], [0, 1], [1, 0], [1, 1]]
+        dims = len(accesses[0])
+        pts = len(accesses)
+        stencil_name = 's%dd_%s_%dpt' % (dims, 'name', pts)
```
FabioLuporini: Do we really need this complicated name? can we not just use `'a'` ?

tests/test_operator.py
```diff
@@ -950,6 +953,37 @@ def test_incomplete_override(self):
...
+        stencil_array = Array(
+            name=stencil_name,
+            dimensions=(DefaultDimension(name='len', default_value=dims * pts),),
+            dtype=np.int32,
+            scope='stack'
+        )
+        list_initialize = Expression(ClusterizedEq(Eq(
+            stencil_array,
+            ListInitializer(list(itertools.chain(*accesses)))
```
FabioLuporini: why itertools? can it not just be 
```
Expression(...(array, ListInitializer([0, 1])))
```

maelso:
Thanks @FabioLuporini. I'll work on your comments and soon I'll update this PR, including the rebase as well.

tests/test_operator.py
maelso: Insert decls is needed. I'll remove insert casts.

tests/test_operator.py
maelso: > I still don't think that 30 almost identical lines are duplicated across two different files

I agree!

> if you further change ops.Array, then that's when you should write a new test, I feel

I sincerely believe that testing should come before code evolution to ensure project integrity, but as I agree with the previous point and I don't have a solution for code duplicity, so I removed it.

tests/test_operator.py
maelso: Makes sense. Changed.

tests/test_operator.py
maelso: Yeah! Changed, thanks.

tests/test_operator.py
maelso: It is not possible to pass a tuple directly:

```
return super().free_symbols - {d for d in self.dimensions if d.is_Default}
E   AttributeError: 'tuple' object has no attribute 'is_Default' 
```

tests/test_operator.py
```diff
@@ -1118,6 +1120,33 @@ def test_stack_vector_temporaries(self):
   timers->section0 += (double)(end_section0.tv_sec-start_section0.tv_sec)\
 +(double)(end_section0.tv_usec-start_section0.tv_usec)/1000000;""" in str(operator)
 
+    def test_conditional_declarations(self):
+        accesses = [0, 0]
+        stencil_array = Array(
```
FabioLuporini: this WFM:

```
a = Array(name='a', dimensions=(x, y), dtype=np.int32, scope='stack')
```

we don't need DefaultDimension

for homogeneity with the rest of the codebase, the variable name "stencil_array" feels out of place. For dummy variables, we use single letter names, such as `a`, or at worst, `array`

the word "stencil" doesn't mean anything here

FabioLuporini: if/when we drop DefaultDimension, we can then also drop the `accesses` variable

FabioLuporini: `Array(...)` should be one/two lines (if needs wrapping), definitely not six

maelso: > a = Array(name='a', dimensions=(x, y), dtype=np.int32, scope='stack')

The generated code will be:
`int a[x_size] = {0, 0};` instead of `int a[2] = {0, 0};`

Is this correct?

FabioLuporini: probably? well, `a[x_size, y_size]` in my case, but yeah

is this enough as an MFE? I don't feel like you need DefaultDimension for this MFE, but please correct me if I'm missing something

maelso: Yes. Changed!

FabioLuporini:
This one seems OK now, I'll take a final look on Monday

FabioLuporini:
Im merging this, but bear in mind it's just a hack for what was already hacky/messy code

as I said multiple times, this entire piece of code (the declarator) needs a full rewrite

PR #964, opened 2019-10-14
Functions as fd weights

rhodrin:
Small PR to allow custom FD weights to be set using Devito Functions. Relevant test included.

### DIFF
--- a/devito/finite_differences/coefficients.py
+++ b/devito/finite_differences/coefficients.py
@@ -51,13 +51,23 @@ class Coefficient(object):
 
         # Ensure the given set of weights is the correct length
         if dimension.is_Time:
-            if len(weights)-1 != function.time_order:
-                raise ValueError("Number FD weights provided does not "
-                                 "match the functions space_order")
+            if isinstance(weights, np.ndarray):
+                if len(weights)-1 != function.time_order:
+                    raise ValueError("Number FD weights provided does not "
+                                     "match the functions space_order")
+            else:
+                if weights.shape[-1]-1 != function.time_order:
+                    raise ValueError("Number FD weights provided does not "
+                                     "match the functions space_order")
         elif dimension.is_Space:
-            if len(weights)-1 != function.space_order:
-                raise ValueError("Number FD weights provided does not "
-                                 "match the functions space_order")
+            if isinstance(weights, np.ndarray):
+                if len(weights)-1 != function.space_order:
+                    raise ValueError("Number FD weights provided does not "
+                                     "match the functions space_order")
+            else:
+                if weights.shape[-1]-1 != function.space_order:
+                    raise ValueError("Number FD weights provided does not "
+                                     "match the functions space_order")
 
         self._deriv_order = deriv_order
         self._function = function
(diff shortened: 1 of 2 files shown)


devito/finite_differences/coefficients.py
```diff
@@ -50,13 +50,17 @@ def __init__(self, deriv_order, function, dimension, weights):
...
-            if len(weights)-1 != function.time_order:
-                raise ValueError("Number FD weights provided does not "
+            if wl != function.time_order:
+                raise ValueError("Number of FD weights provided does not "
                                  "match the functions space_order")
         elif dimension.is_Space:
-            if len(weights)-1 != function.space_order:
-                raise ValueError("Number FD weights provided does not "
+            if wl != function.space_order:
+                raise ValueError("Number of FD weights provided does not "
```
mloubout: For future PR, but this still asumes the use knows what the indices are with the ordering, because for example for staggered FD in my new PR the indices are not ordered anymore in some cases. Mayeb would be good to give it as a dict `pos: w` or something like that.

devito/finite_differences/coefficients.py
```diff
@@ -87,8 +91,6 @@ def weights(self):
     def _check_input(self, deriv_order, function, dimension, weights):
         if not isinstance(deriv_order, int):
             raise TypeError("Derivative order must be an integer")
-        # NOTE: Can potentially be tidied up following the implementation
-        # of lazy evaluation.
```
mloubout: why removed?

rhodrin: It didn't seem immediately obvious to me that the lazy implementation could help in tidying this up. I'm very possibly missing something though?

devito/finite_differences/coefficients.py
rhodrin: Yeah, good point will add note.

devito/finite_differences/coefficients.py
```diff
@@ -185,9 +191,23 @@ def generate_subs(i):
...
                                            side=None, stagger=stagger)
 
-            for j in range(len(weights)):
-                subs.update({function._coeff_symbol
-                             (indices[j], deriv_order, function, dim): weights[j]})
+            # NOTE: This implementation currently assumes that indices are ordered
+            # according to their position in the FD stencil. This may not be the
+            # case in all schemes and should be changed such that the weights are
+            # passed as a dictionary of the form {pos: w} (or something similar).
+            if isinstance(weights, np.ndarray):
```
FabioLuporini: Not sure if now or next, but probably it's worth having a wrapper class `Weights` that exposes a common interface and performs the necessary checks. Then:

- you wouldn't need any sort of if-then-else
- you manage to handle the {pos: w} thing easily

I wonder what your thoughts are. I'm fine if we decide to leave this for another day, but perhaps we could add a note somewhere (assuming our thoughts are in sync)

rhodrin: Yes, I think a `class weights` sounds like a good option.

If nobody has any strong objections though I'd prefer to leave this for a later date.

FabioLuporini:
merged.

Issue #827, opened 2019-05-07
Cyclic depencency error: Equations with implicit_dims + SubDomainSet


rhodrin:
Take for example the following code:

```
import numpy as np
from math import floor
from devito import (Grid, Function, TimeFunction, Eq, solve, Operator, SubDomainSet,
                    Dimension)

n_domains = 10

class Inner(SubDomainSet):
    name = 'inner'

bounds_xm = np.zeros((n_domains,), dtype=np.int32)
bounds_xM = np.zeros((n_domains,), dtype=np.int32)
bounds_ym = np.zeros((n_domains,), dtype=np.int32)
bounds_yM = np.zeros((n_domains,), dtype=np.int32)

for j in range(0, n_domains):
    bounds_xm[j] = j
    bounds_xM[j] = n_domains-1-j
    bounds_ym[j] = floor(j/2)
    bounds_yM[j] = floor(j/2)

bounds = (bounds_xm, bounds_xM, bounds_ym, bounds_yM)

inner_sd = Inner(N=n_domains, bounds=bounds)

grid = Grid(extent=(10, 10), shape=(10, 10), subdomains=(inner_sd, ))

f = TimeFunction(name='f', grid=grid, dtype=np.int32)
f.data[:] = 0

v = Function(name='v', dimensions=(inner_sd.implicit_dimension, ), shape=(n_domains, ))

for j in range(1, n_domains):
    v.data[j] = v.data[j-1]+1.0
    
idims = (grid.stepping_dim, inner_sd.implicit_dimension, 
         inner_sd.dimensions[0], inner_sd.dimensions[1])

stencil = Eq(f.forward, v*solve(Eq(f.dt, 1), f.forward),
            subdomain=grid.subdomains['inner'], implicit_dims=idims)

op = Operator(stencil, dle='noop')
op(time_m=0, time_M=1, dt=1)
```

In the above we are trying to set `v` to a different value in each subdomain of the `SubDomainSet`. This 'sometimes' works (as expected) but other times throws and error of the form:
```
ValueError: A cyclic dependency exists amongst OrderedDict([(n, {n, time}), (time, {n})])
```

Note that this was previously untested functionality and therefore there's no reason it should have worked 'out of the box', but the above issue needs to be fixed.

FabioLuporini:
I wonder whether this is still an issue

rhodrin:
I checked fairly recently and it still was. Will check again when I get back.

rhodrin:
[...]

PR #965, opened 2019-10-15
Fix for issue #827

rhodrin:
Fixes a cyclic dependency error that sometimes occurred with subdomain sets. See issue #827 for further details.


devito/operator.py
```diff
@@ -236,6 +236,7 @@ def _add_implicit(self, expressions):
                 try:
                     dims = [d.root for d in e.free_symbols if isinstance(d, Dimension)]
                     sub_dims = [d.root for d in e.subdomain.dimensions]
+                    sub_dims.append(e.subdomain.implicit_dimension)
```
mloubout: Is there an MFE this fixes to go as a test?

FabioLuporini: fair point

rhodrin: Yes, a shorter version of #827 would do - although it would only fail ~50% of the time (depending on how `free_symbols` works).

FabioLuporini:
merged.

PR #967, opened 2019-10-16
ops: applying declarations for each generated OPS kernel

maelso:
Currently, the backend OPS isn't calling __finalize_ for generated kernels, this will generate code like:
```
void OPS_Kernel_0(...)
{
  r2 = 1.0/(*h_y**h_y);
  r1 = 1.0/(*h_x**h_x);
  r0 = 1.0/(*dt**dt);
}
```
The correct code must be:
```
void OPS_Kernel_0(...)
{
  float r2 = 1.0/(*h_y**h_y);
  float r1 = 1.0/(*h_x**h_x);
  float r0 = 1.0/(*dt**dt);
}
```

### DIFF
--- a/devito/ops/operator.py
+++ b/devito/ops/operator.py
@@ -1,6 +1,6 @@
 from devito import Eq
 from devito.ir.equations import ClusterizedEq
-from devito.ir.iet import Call, Expression, List, find_affine_trees
+from devito.ir.iet import Call, Expression, List, derive_parameters, find_affine_trees
 from devito.ir.iet.visitors import FindSymbols, Transformer
 from devito.logger import warning
 from devito.operator import Operator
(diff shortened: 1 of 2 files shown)


devito/ops/operator.py
```diff
@@ -110,6 +110,14 @@ def _specialize_iet(self, iet, **kwargs):
     def hcode(self):
         return ''.join(str(kernel) for kernel in self._ops_kernels)
 
+    def _finalize(self, iet, parameters):
+        # Applies _finalize for each generated OPS kernel, which will generate the .h file
+        self._ops_kernels = [super(OperatorOPS, self)._finalize(
```
FabioLuporini: `.extend([...])` instead of `= [ ...]` ?

maelso: No. The kernels must be replaced. Otherwise, we will duplicate kernels, w/ and w/o declarations.

FabioLuporini: I'm merging this, but I'm not a fan of methods -- even though they're private methods -- that replace mutable variables

PR #968, opened 2019-10-17
RFC: revisit symbolic processing

FabioLuporini:
@mloubout after all, I think monkey-patching sympy.Basic is too horrible, even if it works. So I'm here resurrecting one of your former approaches -- rebuilding as devito core types upon sympy.flatten. There are a few differences w.r.t. what you did:

- technically, this is now based on the single dispatch pattern
- only the immediate children are reconstructed
- the sympy assumptions are propagated too

### DIFF
--- a/devito/finite_differences/differentiable.py
+++ b/devito/finite_differences/differentiable.py
@@ -194,7 +194,7 @@ class DifferentiableOp(Differentiable):
 
         # Unfortunately SymPy may build new sympy.core objects (e.g., sympy.Add),
         # so here we have to rebuild them as devito.core objects
-        obj = Rebuild.evaluate(obj)
+        obj = diffify(obj)
 
         return obj
 
@@ -215,20 +215,25 @@ class Mod(sympy.Mod, DifferentiableOp):
     __new__ = DifferentiableOp.__new__
 
 
-class Rebuild(object):
+class diffify(object):
 
     """
     Helper class based on single dispatch to reconstruct all nodes in a sympy
     tree such they are all of type Differentiable.
+
+    Notes
+    -----
+    The name "diffify" stems from SymPy's "simpify", which has an analogous task --
+    converting all arguments into SymPy core objects.
     """
 
-    def evaluate(obj):
-        args = [Rebuild._doit(i) for i in obj.args]
-        obj = Rebuild._doit(obj, args)
+    def __new__(cls, obj):
+        args = [diffify._doit(i) for i in obj.args]
+        obj = diffify._doit(obj, args)
         return obj
 
     def _doit(obj, args=None):
-        cls = Rebuild._cls(obj)
+        cls = diffify._cls(obj)
         if cls is obj.__class__:
             return obj
 
@@ -239,7 +244,7 @@ class Rebuild(object):
         # As facts are deduced, a SymPy core object's ``_assumptions`` data
         # structure is updated. Since here we use ``evaluate=False`` all over
         # the place, we have to explicitly reassign the ``_assumptions``.
-        newobj._assumptions = obj._assumptions
+        newobj._assumptions = obj._assumptions.copy()
 
         return newobj
 
--- a/devito/finite_differences/differentiable.py
+++ b/devito/finite_differences/differentiable.py
@@ -1,4 +1,5 @@
 from collections import ChainMap
+from functools import singledispatch
 
 import sympy
 from sympy.functions.elementary.integers import floor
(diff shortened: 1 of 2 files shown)


devito/finite_differences/differentiable.py
```diff
@@ -186,20 +187,93 @@ def _has(self, pattern):
...
+        obj = diffify._doit(obj, args)
+        return obj
+
+    def _doit(obj, args=None):
+        cls = diffify._cls(obj)
+        if cls is obj.__class__:
+            return obj
+
+        args = args or obj.args
+        newobj = cls(*args, evaluate=False)
```
mloubout: Wouldn't `cls._from_args(*args)` be more efficient as this one still goes through flatten while `_from_args` just pcopy the already process args to new object.

devito/finite_differences/differentiable.py
```diff
@@ -186,20 +187,93 @@ def _has(self, pattern):
...
+
+        # SymPy objects pretend to be immutable, but in reality they are not.
+        # As facts are deduced, a SymPy core object's ``_assumptions`` data
+        # structure is updated. Since here we use ``evaluate=False`` all over
+        # the place, we have to explicitly reassign the ``_assumptions``.
+        newobj._assumptions = obj._assumptions.copy()
```
mloubout: I think this would also go with `_from_args' but need to double check

devito/finite_differences/differentiable.py
```diff
@@ -186,20 +187,93 @@ def _has(self, pattern):
...
+    tree such they are all of type Differentiable.
+
+    Notes
+    -----
+    The name "diffify" stems from SymPy's "simpify", which has an analogous task --
+    converting all arguments into SymPy core objects.
+    """
+
+    def __new__(cls, obj):
+        args = [diffify._doit(i) for i in obj.args]
```
mloubout: would move that inside `_doit`

devito/finite_differences/differentiable.py
```diff
@@ -186,20 +187,93 @@ def _has(self, pattern):
...
+class DifferentiableOp(Differentiable):
 
+    def __new__(cls, *args, **kwargs):
+        obj = cls.__base__.__new__(cls, *args, **kwargs)
 
-class Mul(sympy.Mul, Differentiable):
-    pass
+        # Unfortunately SymPy may build new sympy.core objects (e.g., sympy.Add),
+        # so here we have to rebuild them as devito.core objects
+        obj = diffify(obj)
```
mloubout: I thought you didn't like that.......

devito/finite_differences/differentiable.py
FabioLuporini: ah OK fair point, then I can use that in place of evaluate=False

devito/finite_differences/differentiable.py
FabioLuporini: if it goes from `_from_args` it would be awesome, I'll check that

devito/finite_differences/differentiable.py
FabioLuporini: yes, I think I can do this

devito/finite_differences/differentiable.py
FabioLuporini: this is the point of this PR -- I'm going back to rebuilding in place of monkey-patching. I thought deeply about how to do this, including a few other ways, and with all that in mind I think I can say this is the way to go. It's not a good one, but you know...we do what we can do. 

Unless you're talking about the name... I don't like it, but it makes a parallel with sympy

devito/finite_differences/differentiable.py
FabioLuporini: https://github.com/sympy/sympy/blob/master/sympy/core/operations.py#L70

only is_commutative ? 😮

devito/finite_differences/differentiable.py
FabioLuporini: no, turns out I cannot. I need the first lines of `_doit` to be executed for each `i` (ie for each obj argument), and I don't want to duplicate code

devito/finite_differences/differentiable.py
mloubout: Oh dang inded, I checked `rawargs` and same only the commutative is saved, that's not great

tests/test_dse.py
```diff
@@ -625,7 +625,7 @@ def test_tti_rewrite_aggressive_wmpi():
 
 @switchconfig(profiling='advanced')
 @pytest.mark.parametrize('space_order,expected', [
-    (8, 152), (16, 268)
+    (8, 154), (16, 270)
```
mloubout: how?

FabioLuporini: if you look back at the monkey patching PR, you may see that there we decreased them, while here we are simply restoring the pre-monkey-patching values. I this happens because in the DSE now we are back to where we were -- syntactically identical sub-expressions fail on == comparison because some of them have arguments which are sympy core objects, whereas with monkey patching everything was devito objects

FabioLuporini:
Merged.

PR #969, opened 2019-10-18
dse: Fix handling of nested aliases

FabioLuporini:
@mloubout : hopefully this fixes the explosion in number of hoisted time invariants 

can you check again with the use case you provided me with long time ago?


FabioLuporini:
I want this PR to fix #970 too . Keeping it open for the time being, but it still needs some little work

FabioLuporini:
Plans have changed. It will take a little more to fix #970 (I could fix it easily in fact, but I prefer a much more structured solution). So merging this for the time being,

PR #971, opened 2019-10-21
ops: naming the return of some functions

btrb:
This PR is for renaming some variable names, and also for avoiding the return of unnamed results when declaring OPS Dats and OPS Args.

### DIFF
--- a/devito/ops/operator.py
+++ b/devito/ops/operator.py
@@ -73,7 +73,7 @@ class OperatorOPS(Operator):
             if f.is_Constant:
                 continue
 
-            pre_time_loop.extend(create_ops_dat(f, name_to_ops_dat, ops_block))
+            pre_time_loop.extend(list(create_ops_dat(f, name_to_ops_dat, ops_block)))
             # To return the result to Devito, it is necessary to copy the data
             # from the dat object back to the CPU memory.
             after_time_loop.extend(create_ops_fetch(f,
--- a/devito/ops/transformer.py
+++ b/devito/ops/transformer.py
@@ -10,7 +10,7 @@ from devito.ir.iet.nodes import Call, Callable, Expression, IterationTree
 from devito.ir.iet.visitors import FindNodes
 from devito.ops.node_factory import OPSNodeFactory
 from devito.ops.types import Array, OpsAccessible, OpsDat, OpsStencil
-from devito.ops.utils import namespace
+from devito.ops.utils import namespace, OpsDatDecl, OpsArgDecl
 from devito.symbolics import Add, Byref, ListInitializer, Literal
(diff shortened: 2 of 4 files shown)


devito/ops/operator.py
```diff
@@ -73,7 +73,7 @@ def _specialize_iet(self, iet, **kwargs):
             if f.is_Constant:
                 continue
 
-            pre_time_loop.extend(create_ops_dat(f, name_to_ops_dat, ops_block))
+            pre_time_loop.extend(list(create_ops_dat(f, name_to_ops_dat, ops_block)))
```
vmickus: This is the only part that i didn't like, but for the sake of the rest of  the code, I think its better.

devito/ops/transformer.py
```diff
@@ -191,13 +190,18 @@ def create_ops_dat(f, name_to_ops_dat, block):
...
+    d_m_val = Expression(ClusterizedEq(Eq(d_m, ListInitializer(d_m_val))))
 
-    return res
+    ops_dat = OpsDatDecl(dim_val=dim_val,
+                         base_val=base_val,
+                         d_p_val=d_p_val,
+                         d_m_val=d_m_val,
+                         ops_decl_dat=ops_decl_dat)
+
+    return ops_dat
```
vmickus: What do you about just returning the `OpsDatDecl`? 
The code sure is cleaner now.

FabioLuporini:
I'm quite happy about this, I'll just wait for you guys, @vmickus and @maelso , to formally approve the PR here on GH, and then I'll proceed with merging it

devito/ops/transformer.py
btrb: sure, thought about it as soon as pushed. thanks for pointing

maelso:
Looks good.

devito/ops/transformer.py
```diff
@@ -240,43 +241,56 @@ def create_ops_par_loop(trees, ops_kernel, parameters, block, name_to_ops_dat,
...
     )))
 
+    ops_args = []
+    for p in parameters:
+        ops_arg = create_ops_arg(p,
+                                 accessible_origin,
+                                 name_to_ops_dat,
+                                 par_to_ops_stencil)
+
+        ops_args.append(ops_arg.ops_type(ops_arg.ops_name,
```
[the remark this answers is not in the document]
btrb: indeed i haven't see a pattern like that before in the codebase
it is correct though. the thing here is that `ops_arg.ops_type` is a `[Function](https://github.com/opesci/devito/blob/master/devito/ops/utils.py#L23)`

FabioLuporini:
merged.

PR #974, opened 2019-10-22
bench: Fix hierarchical blocking setup

FabioLuporini:


FabioLuporini:
the entire benchmarking framework is only expected to work for 3d

there might be some scenarios in which it works in 2d too, but we only need to guarantee functionality for 3d

FabioLuporini:
merged.

PR #983, opened 2019-10-25
Fix sympy type 

mloubout:
Basically forces diffify to replace args at `doit` as currently it is ignored so eben if arguments are turned back to our type if the top class is our it ignores the replacement.

### DIFF
--- a/devito/finite_differences/differentiable.py
+++ b/devito/finite_differences/differentiable.py
@@ -194,7 +194,9 @@ class DifferentiableOp(Differentiable):
 
         # Unfortunately SymPy may build new sympy.core objects (e.g., sympy.Add),
         # so here we have to rebuild them as devito.core objects
-        obj = diffify(obj)
+        # Check if already processed and don't diffify again
+        if kwargs.get('diffify', True):
+            obj = diffify(obj)
 
         return obj
 
@@ -234,11 +236,13 @@ class diffify(object):
 
     def _doit(obj, args=None):
         cls = diffify._cls(obj)
-        if cls is obj.__class__:
+        # Only keep object if arfs don't need to be replaced
+        if cls is obj.__class__ and args is None:
             return obj
 
         args = args or obj.args
-        newobj = cls(*args, evaluate=False)
+        # Already processed so don't diffify again
+        newobj = cls(*args, evaluate=False, diffify=False)
 
         # SymPy objects pretend to be immutable, but in reality they are not.
         # As facts are deduced, a SymPy core object's ``_assumptions`` data
--- a/devito/finite_differences/differentiable.py
+++ b/devito/finite_differences/differentiable.py
@@ -194,9 +194,7 @@ class DifferentiableOp(Differentiable):
 
         # Unfortunately SymPy may build new sympy.core objects (e.g., sympy.Add),
         # so here we have to rebuild them as devito.core objects
-        # Check if already processed and don't diffify again
-        if kwargs.get('diffify', True):
-            obj = diffify(obj)
+        obj = diffify(obj)
 
         return obj
 
@@ -212,9 +210,17 @@ class Mul(sympy.Mul, DifferentiableOp):
 class Pow(sympy.Pow, DifferentiableOp):
     __new__ = DifferentiableOp.__new__
 
+    @classmethod
+    def _from_args(cls, args, is_commutative=None):
+        return cls(*args, evaluate=False)
+
 
 class Mod(sympy.Mod, DifferentiableOp):
     __new__ = DifferentiableOp.__new__
+
+    @classmethod
+    def _from_args(cls, args, is_commutative=None):
+        return cls(*args, evaluate=False)
 
 
 class diffify(object):
@@ -236,13 +242,18 @@ class diffify(object):
 
     def _doit(obj, args=None):
         cls = diffify._cls(obj)
-        # Only keep object if arfs don't need to be replaced
-        if cls is obj.__class__ and args is None:
-            return obj
-
         args = args or obj.args
-        # Already processed so don't diffify again
-        newobj = cls(*args, evaluate=False, diffify=False)
+
+        if cls is obj.__class__:
+            # Try to just update the args if possible (Add, Mul)
+            try:
+                return obj._new_rawargs(*args)
+            # Or just return the object (Float, Symbol, ....)
+            except AttributeError:
+                return obj
+
+        # Create object directly from args, avoid any rebuild
+        newobj = cls._from_args(args, is_commutative=obj.is_commutative)
 
         # SymPy objects pretend to be immutable, but in reality they are not.
         # As facts are deduced, a SymPy core object's ``_assumptions`` data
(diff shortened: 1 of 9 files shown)


tests/test_dse.py
```diff
@@ -701,4 +701,4 @@ def test_tti_v2_rewrite_aggressive_opcounts(space_order, expected):
 
 
 if __name__ == "__main__":
-    test_tti_rewrite_aggressive_opcounts(16, 270)
+    test_tti_v2_rewrite_aggressive_opcounts(4, 185)
```
FabioLuporini: leftover?

devito/finite_differences/differentiable.py
```diff
@@ -234,18 +235,19 @@ def __new__(cls, obj):
...
     def _doit(obj, args=None):
         cls = diffify._cls(obj)
+        args = args or obj.args
+
         if cls is obj.__class__:
+            # Try to just update the args if possible (Add, Mul)
+            if getattr(obj, 'is_Add', False):
+                return obj._new_rawargs(*args, is_commutative=obj.is_commutative)
+            # # Or just return the object (Float, Symbol, ....)
+            # except AttributeError:
```
FabioLuporini: leftover?

devito/finite_differences/differentiable.py
```diff
@@ -234,18 +235,19 @@ def __new__(cls, obj):
 
     def _doit(obj, args=None):
         cls = diffify._cls(obj)
+        args = args or obj.args
+
         if cls is obj.__class__:
+            # Try to just update the args if possible (Add, Mul)
+            if getattr(obj, 'is_Add', False):
```
FabioLuporini: the try-except was way neater

avoid `getattr` at all costs when possible

mloubout: This is only for Add and the try/except would catch mul wich creates problems

tests/test_dse.py
mloubout: Only changed which function to debug doenst really.matter

devito/finite_differences/differentiable.py
mloubout: Yes sorry

devito/finite_differences/differentiable.py
FabioLuporini: Two things

1. that's not what the comment above says (Add, Mul)

2. if it were just for Add, it'd smell like a hack

mloubout: Ok so it doesn't break with both, but the dse doesn't like the expression anymore and end up with like double opcount because of it in the tti tests. So it is a hack but not because of sympy but because factorization isn't detected if I put back mul in it

mloubout: So I can put back the try/except which is the one that should be done but I'll have to change the opcounts in way higher values in test_dse (the 1/h_x are not factorized anymore for example)

FabioLuporini: we should fix the factorizer

devito/symbolics/search.py
```diff
@@ -131,6 +131,11 @@ def retrieve_functions(expr, mode='all'):
     return search(expr, q_function, mode, 'dfs')
 
 
+def retrieve_scalars(expr, mode='all'):
+    """Shorthand to retrieve the DiscreteFunctions in ``expr``."""
```
FabioLuporini: revisit docstring

devito/symbolics/queries.py
```diff
@@ -22,10 +22,14 @@
 """
 
 
-def q_scalar(expr):
+def q_number(expr):
```
FabioLuporini: this name is misleading. Can we stick to `q_scalar`? 

as for the new one, maybe we could use `q_symbol`, then rename `retrieve_scalars`  as `retrieve_symbol`, and use `retrieve_symbol` in `collect_temp` ? after all, a Scalar is just a Symbol subclass. I however appreciate that there is some naming mess up

devito/symbolics/manipulation.py
```diff
@@ -274,3 +275,13 @@ def indexify(expr):
         except AttributeError:
             pass
     return expr.xreplace(mapper)
+
+
+def collect_temp(expr):
```
FabioLuporini: this is super small and used in only one place, so it's better to squash it inside the main collect routine

devito/finite_differences/differentiable.py
```diff
@@ -234,17 +235,18 @@ def __new__(cls, obj):
...
 
-        # SymPy objects pretend to be immutable, but in reality they are not.
-        # As facts are deduced, a SymPy core object's ``_assumptions`` data
-        # structure is updated. Since here we use ``evaluate=False`` all over
-        # the place, we have to explicitly reassign the ``_assumptions``.
-        newobj._assumptions = obj._assumptions.copy()
+        if cls is obj.__class__:
+            # Try to just update the args if possible (Add, Mul)
+            try:
+                newobj = obj._new_rawargs(*args, is_commutative=obj.is_commutative)
```
FabioLuporini: so remind me how is `_new_rawargs` "better" or different than `from_args` ? and can you add a comment about why this function and not (for example) from args?

FabioLuporini: also, something's odd here: you create `newobj = ...` but then, line 249, you basically override it? as there's no `return` statement here. What am I missing?

If instead I spotted an error, I suggest to add a test that would have caught this issue

mloubout: >What am I missing?
debugging that weird lone ops error left should be return.

> _new_rawargs "better" or different than from_args ? 
Not much some slight difference on `is_commutative`

devito/symbolics/queries.py
mloubout: Not sure I follow or agree, feels like semantics preferences

FabioLuporini: it's convention.
1) SymPy calls Number objects that are non-symbolic numbers. 
2) Devito calls Scalars objects that are subclass of Symbol
so either we change Scalar all over the place (...) or we tweak this name

mloubout: `q_scalar` check if it is a `Scalar`, why the hell would I name this one like that that makes zero sense

devito/ops/types.py
```diff
@@ -112,6 +112,9 @@ def as_coeff_Mul(self):
     def as_coeff_Add(self):
         return sympy.S.Zero, self
 
+    def as_ordered_factors(self):
+        return (self,)
```
mloubout: Note for review.
This is necessary for `_print_Mul` because they define `_as_coeff_Mul()` above.

devito/symbolics/queries.py
```diff
@@ -23,7 +23,7 @@
 
 
 def q_scalar(expr):
-    return expr.is_Number or expr.is_Symbol
+    return getattr(expr, 'is_Scalar', False)
```
FabioLuporini: One should use `getattr` only when the attribute name is unknown , or perhaps in list comprehension (extreme case).

Otherwise, the pythonic way is

```
try:
    return expr.is_Scalar
except AttributeError:
    return False
```

FabioLuporini:
if you decide not to add a test for `collect_nested`, could you open an issue as a reminder? thanks

apart from the pythonicity nitpick above , this PR now seems GTG. 

EDIT: ah, it needs rebasing

mloubout:
>if you decide not to add a test for collect_nested, could you open an issue as a reminder? thanks

Well not sure how would test it, I'll think about it, just add one step of factorization on the cse found temporaries.

PR #993, opened 2019-11-03
Replace ops fetch memories call

vmickus:
Instead of using the OPS API method `ops_dat_fetch_data` to get data from GPU, we are now using `ops_dat_get_raw_pointer` method. Because the first method do not bring back the data from the halo region.

### DIFF
--- a/devito/ops/transformer.py
+++ b/devito/ops/transformer.py
@@ -1,17 +1,16 @@
 import itertools
 import numpy as np
 
-from sympy import Mod
 from sympy.core.numbers import Zero
 
-from devito import Eq
+from devito import Eq, TimeFunction
 from devito.ir.equations import ClusterizedEq
 from devito.ir.iet.nodes import Call, Callable, Expression, IterationTree
 from devito.ir.iet.visitors import FindNodes
 from devito.ops.node_factory import OPSNodeFactory
-from devito.ops.types import Array, OpsAccessible, OpsDat, OpsStencil
+from devito.ops.types import Array, OpsAccessible, OpsDat, OpsMemSpace, OpsStencil
 from devito.ops.utils import namespace, OpsDatDecl, OpsArgDecl
-from devito.symbolics import Add, Byref, ListInitializer, Literal
+from devito.symbolics import Byref, ListInitializer, Literal, Macro, split_affine
 from devito.tools import dtype_to_cstr
 from devito.types import Constant, DefaultDimension, Symbol
 
@@ -62,7 +61,7 @@ def opsit(trees, count, name_to_ops_dat, block, dims):
 
     pre_time_loop = stencil_arrays_initializations + ops_par_loop_init
 
-    return pre_time_loop, ops_kernel, ops_par_loop_call
+    return pre_time_loop, ops_kernel, ops_par_loop_call, par_to_ops_stencil
 
 
 def to_ops_stencil(param, accesses):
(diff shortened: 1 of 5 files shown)


tests/test_ops.py
```diff
@@ -262,30 +261,16 @@ def test_upper_bound(self, equation, expected):
...
+            '[\'ops_dat_get_raw_pointer(u_dat[t0],0,S2D_UT0_1PT,&memspace);\','
+            '\'ops_dat_get_raw_pointer(u_dat[t1],0,S2D_UT1_1PT,&memspace);\','
+            '\'ops_memspace memspace = OPS_HOST;\']'),
+        ('Eq(v_3d.backward, v_3d.dt2 + b_3d)',
+            '[\'ops_memspace memspace = OPS_HOST;\','
+            '\'ops_dat_get_raw_pointer(v_dat[t5],0,S3D_VT5_1PT,&memspace);\','
+            '\'ops_dat_get_raw_pointer(v_dat[t6],0,S3D_VT6_1PT,&memspace);\','
+            '\'ops_dat_get_raw_pointer(v_dat[t7],0,S3D_VT7_1PT,&memspace);\','
+            '\'ops_dat_get_raw_pointer(b_dat,0,S3D_B_1PT,&memspace);\']')])
+    def test_memory_transfer_call(self, equation, expected):
```
FabioLuporini: this one looks thinner now?

devito/ops/transformer.py
```diff
@@ -202,22 +204,56 @@ def create_ops_dat(f, name_to_ops_dat, block):
...
+    par_to_ops_stencil : dict of {str : :class:`devito.ops.types.OpsStencil`}
+        Link an OpsAccessible name to its OpsStencil variable.
+    accessibles_info : dict of {str : :namedTuple:`devito.ops.types.Accessible_info`}
+        Contains the time variable used.
+    memspace : :class:`devito.ops.types.OpsMemSpace`
+        Specify which memory space should the data be transfered into.
+
+    Returns
+    ------
+    :class:`devito.ir.iet.nodes.Call`
```
FabioLuporini: no need for `:class:` (remember? we've eliminated them all, thanks to @btrb too)

you should remove all of them

vmickus: Sure.

tests/test_ops.py
```diff
@@ -298,7 +283,10 @@ def test_create_fetch_data(self, equation, expected):
         v_3d = TimeFunction(name='v', grid=grid_3d, time_order=2)  # noqa
         x_3d = TimeFunction(name='x', grid=grid_3d, time_order=3)  # noqa
 
-        op = Operator(eval(equation))
+        a_2d = Function(name='a', grid=grid_2d)                     # noqa
+        b_3d = Function(name='b', grid=grid_3d)                     # noqa
 
+        op = Operator(eval(equation))
+        print(op)
```
FabioLuporini: leftover

vmickus: Fixed

tests/test_ops.py
vmickus: Yes it is, I just think that for now there is no point in add more test than this because what will really matter is when we can test using the translation that @btrb is working with

vmickus:
Some issues when I did the rebase. I'll need to look in more detail.

I'll close this PR and then when ready I'll submit it again.

PR #995, opened 2019-11-04
nthreads2 for nonaffine loops

FabioLuporini:
In master:
* the `nthreads0` variable is used for both "FD loops" and "sparse loops"
* the `nthreads1` variable is used for nested "FD loops" (applies to KNL only ATM)

with this PR:
* the `nthreads0` variable is used for "FD loops"
* the `nthreads1` variable is used for nested "FD loops" (applies to KNL only ATM)
* the `nthreads2` variable is used for "sparse loops"

performance-wise nothing changes unless one supplies different (non-default) values for `nthreads0` and `nthreads2` to `op.apply`

### DIFF
--- a/devito/dle/blocking_utils.py
+++ b/devito/dle/blocking_utils.py
@@ -399,6 +399,8 @@ class IterationFold(Iteration):
 
 class BlockDimension(IncrDimension):
 
+    is_PerfKnob = True
+
     @cached_property
     def symbolic_min(self):
         return Scalar(name=self.min_name, dtype=np.int32, is_const=True)
--- a/devito/dle/parallelizer.py
+++ b/devito/dle/parallelizer.py
@@ -23,21 +23,11 @@ def nhyperthreads():
 
 
 class NThreadsMixin(object):
-    pass
-
-
-class NThreads(Constant, NThreadsMixin):
-
-    name = 'nthreads0'
-
-    @classmethod
-    def default_value(cls):
-        return int(os.environ.get('OMP_NUM_THREADS', ncores()))
 
     def __new__(cls, **kwargs):
         name = kwargs.get('name', cls.name)
         value = cls.default_value()
-        obj = super(NThreads, cls).__new__(cls, name=name, dtype=np.int32, value=value)
+        obj = Constant.__new__(cls, name=name, dtype=np.int32, value=value)
         obj.aliases = as_tuple(kwargs.get('aliases')) + (name,)
         return obj
 
@@ -53,24 +43,27 @@ class NThreads(Constant, NThreadsMixin):
         return self._arg_defaults()
 
 
-class NThreadsNested(Constant, NThreadsMixin):
+class NThreads(NThreadsMixin, Constant):
 
-    name = 'nthreads1'
+    name = 'nthreads'
+
+    @classmethod
+    def default_value(cls):
+        return int(os.environ.get('OMP_NUM_THREADS', ncores()))
+
+
+class NThreadsNested(NThreadsMixin, Constant):
+
+    name = 'nthreads_nested'
 
     @classmethod
     def default_value(cls):
         return nhyperthreads()
 
-    def __new__(cls, **kwargs):
-        name = kwargs.get('name', cls.name)
-        value = NThreadsNested.default_value()
-        return super(NThreadsNested, cls).__new__(cls, name=name, dtype=np.int32,
-                                                  value=value)
-
 
 class NThreadsNonaffine(NThreads):
 
-    name = 'nthreads2'
+    name = 'nthreads_nonaffine'
 
 
 class ParallelRegion(Block):
@@ -192,9 +185,9 @@ class Ompizer(object):
             self.key = key
         else:
             self.key = lambda i: i.is_ParallelRelaxed and not i.is_Vectorizable
-        self.nthreads = NThreads(aliases='nthreads')
-        self.nthreads_nested = NThreadsNested()
-        self.nthreads_nonaffine = NThreadsNonaffine()
+        self.nthreads = NThreads(aliases='nthreads0')
+        self.nthreads_nested = NThreadsNested(aliases='nthreads1')
+        self.nthreads_nonaffine = NThreadsNonaffine(aliases='nthreads2')
 
     def _make_atomic_incs(self, partree):
         if not partree.is_ParallelAtomic:
--- a/devito/dle/parallelizer.py
+++ b/devito/dle/parallelizer.py
@@ -24,6 +24,8 @@ def nhyperthreads():
 
 class NThreadsMixin(object):
 
+    is_PerfKnob = True
+
     def __new__(cls, **kwargs):
         name = kwargs.get('name', cls.name)
         value = cls.default_value()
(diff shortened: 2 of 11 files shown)


devito/dle/parallelizer.py
```diff
@@ -61,12 +62,17 @@ def default_value(cls):
...
-        name = kwargs.get('name', NThreadsNested.name)
+        name = kwargs.get('name', cls.name)
         value = NThreadsNested.default_value()
         return super(NThreadsNested, cls).__new__(cls, name=name, dtype=np.int32,
                                                   value=value)
 
 
+class NThreadsNonaffine(NThreads):
+
+    name = 'nthreads2'
```
mloubout: Maybe rename these according to class, like `nthreadAff` and so on?

FabioLuporini: I thought about that, but the generated code looks a bit ugly that way... I'm paranoid, aren't I? 

I thought it's just three names and they'll never change, tied deterministically to specific classes of loops, so it shouldn't be too bad ?

However, I'm willing to reconsider if you feel like this is gonna be a mental burnout for users

mloubout: I don't know with a 3 letter thing like `Aff, Std, ...` I think it's still fairly readable and makes it a bit better for the readability of the C code.

FabioLuporini: OK. I've just finished it. Eventually, I picked `nthreads` (basically "all classic stencil loops"), `nthreads_nested`, `nthreads_nonaffine`.

devito/operator.py
```diff
@@ -602,8 +606,24 @@ def _profile_output(self, args):
...
+                continue
+            try:
+                perf_args[i.name] = args[i.name]
+            except KeyError:
+                # Try with the aliases
+                for a in i._arg_names:
+                    if a in args:
+                        perf_args[a] = args[a]
+                        break
+        perf("Performance arguments:  %s" % perf_args)
```
FabioLuporini: with a low log level, now also emitting the performance-related arguments values when returning from op.apply

E.g.
```
...
Performance arguments:  {'nthreads_nonaffine': 4, 'nthreads': 4, 'x0_blk0_size': 8, 'y0_blk0_size': 8}
```

devito/logger.py
```diff
@@ -104,6 +104,14 @@ def set_log_noperf():
     logger.setLevel(WARNING)
 
 
+def would_emit(level):
```
georgebisbas: This function name seems like "the odd one out" for me, but maybe I am just nitpicking.

tests/test_dle.py
```diff
@@ -337,23 +338,38 @@ def test_iterations_ompized(self, fa, fb, fc, fd, t0, t1, t2, t3,
...
+        assert len(parregions) == 2
+
+        # Check suitable `num_threads` appear in the generated code
         # Not very elegant, but it does the trick
-        assert 'num_threads(nthreads0)' in str(op)
+        assert 'num_threads(nthreads)' in str(parregions[0].header[0])
+        assert 'num_threads(nthreads_nonaffine)' in str(parregions[1].header[0])
 
-        # Check `op` accepts the `nthreads` kwarg
+        # Check `op` accepts the `nthreads*` kwargs
```
georgebisbas: `nthreads` *kwargs ?

tests/test_dle.py
```diff
@@ -432,14 +448,17 @@ def test_basic(self):
 
         # Try again but this time supplying specific values for the num_threads
         u.data[:] = 0.
-        op.apply(t_M=9, nthreads0=1, nthreads1=2)
+        op.apply(t_M=9, nthreads=1, nthreads_nested=2)
         assert np.all(u.data[0] == 10)
-        assert op.arguments(t_M=9, nthreads1=2)['nthreads1'] == 2
+        assert op.arguments(t_M=9, nthreads_nested=2)['nthreads_nested'] == 2
+        # Same as above, but with the alias
+        assert op.arguments(t_M=9, nthreads1=2)['nthreads_nested'] == 2
```
georgebisbas: nthreads1 ?

devito/dle/parallelizer.py
```diff
@@ -190,8 +187,9 @@ def __init__(self, key=None):
             self.key = key
         else:
             self.key = lambda i: i.is_ParallelRelaxed and not i.is_Vectorizable
-        self.nthreads = NThreads()
-        self.nthreads_nested = NThreadsNested()
+        self.nthreads = NThreads(aliases='nthreads0')
+        self.nthreads_nested = NThreadsNested(aliases='nthreads1')
```
georgebisbas: ?

FabioLuporini:
@georgebisbas I changed it this morning. Now it's `nthreads`, `nthreads_nonaffine`, `nthreads_nested`. So, quite clear I'd say

devito/dle/parallelizer.py
FabioLuporini: what's the question ?

tests/test_dle.py
FabioLuporini: oh, that is just an alias

tests/test_dle.py
FabioLuporini: no, there are multiple nthreads, didn't bother writing them

devito/logger.py
FabioLuporini: feels OK to me . it reads quite well "if would_emit(a_certain_log_level) then do something". No?

devito/dle/blocking_utils.py
```diff
@@ -399,6 +399,8 @@ def unfold(self):
 
 class BlockDimension(IncrDimension):
 
+    is_PerfKnob = True
```
mloubout: ?

devito/dle/parallelizer.py
```diff
@@ -23,48 +23,49 @@ def nhyperthreads():
...
-
-class NThreads(Constant, NThreadsMixin):
-
-    name = 'nthreads0'
-    aliases = (name, 'nthreads')
-
-    @classmethod
-    def default_value(cls):
-        return int(os.environ.get('OMP_NUM_THREADS', ncores()))
+    is_PerfKnob = True
```
mloubout: ? same

devito/dle/parallelizer.py
mloubout: But why the aliases and not just keep the default names?

devito/logger.py
mloubout: Feels like a repetition of `isEnableFor`

devito/operator.py
mloubout: `Perf` is stil lbroken I think so not sure this is usefull.

devito/dle/blocking_utils.py
FabioLuporini: types.Basic defines  `is_PerfKnob` and has a comment about that is

FabioLuporini:
@mloubout 

> But why the aliases and not just keep the default names?

the aliases are a plus, mainly for backwards compatibility. Not sure whay you mean by "keep"...the default names is still the main way to set the value at apply time. `op.apply(... nthreads=4, nthreads_nonaffine=4, ...)`

FabioLuporini:
Merged.

PR #997, opened 2019-11-08
ops/compiler: refactoring OPSCompiler

btrb:
- Currently, the OPS Compiler would be called even if there is no affine trees found in the IET. With this PR, if no affine tree is found in the IET, the `_compiler` method should not be overwritten. 

- This PR also refactors OPSCompiler for allowing multiple OPS backend in the future.

- Finally, this PR installs OPS in both azure and travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,7 +63,7 @@ matrix:
           packages:
             - gcc-7
             - g++-7
-      env: DEVITO_ARCH=gcc-7 DEVITO_OPENMP=0 DEVITO_BACKEND=ops YC_CXX=g++-7 INSTALL_TYPE=conda RUN_EXAMPLES=False MPI_INSTALL=0
+      env: DEVITO_ARCH=gcc-7 DEVITO_OPENMP=0 DEVITO_BACKEND=ops YC_CXX=g++-7 INSTALL_TYPE=conda RUN_EXAMPLES=False MPI_INSTALL=0 OPS_COMPILER=gnu OPS_INSTALL_PATH=$HOME/build/opesci/OPS/ops
 
 addons:
   apt:
@@ -108,6 +108,10 @@ before_script:
       conda install swig; cd ../;
       git clone https://github.com/opesci/yask.git;
       cd yask; make compiler-api; pip install -e .; cd ../devito;
+    elif [[ $DEVITO_BACKEND == 'ops' ]]; then
+      cd ../;
+      git clone https://github.com/opesci/OPS.git;
+      cd OPS/ops/c; make; cd ../../../devito;
     fi
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,7 +63,7 @@ matrix:
           packages:
             - gcc-7
             - g++-7
-      env: DEVITO_ARCH=gcc-7 DEVITO_OPENMP=0 DEVITO_BACKEND=ops INSTALL_TYPE=conda RUN_EXAMPLES=False MPI_INSTALL=0 OPS_COMPILER=gnu OPS_INSTALL_PATH=$HOME/build/opesci/OPS/ops
+      env: DEVITO_ARCH=gcc-7 DEVITO_OPENMP=0 DEVITO_BACKEND=ops INSTALL_TYPE=conda RUN_EXAMPLES=False MPI_INSTALL=0 OPS_COMPILER=gnu OPS_INSTALL_PATH=$HOME/opesci/OPS/ops
 
 addons:
   apt:
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,7 +63,7 @@ matrix:
           packages:
             - gcc-7
             - g++-7
-      env: DEVITO_ARCH=gcc-7 DEVITO_OPENMP=0 DEVITO_BACKEND=ops YC_CXX=g++-7 INSTALL_TYPE=conda RUN_EXAMPLES=False MPI_INSTALL=0 OPS_COMPILER=gnu OPS_INSTALL_PATH=$HOME/build/opesci/OPS/ops
+      env: DEVITO_ARCH=gcc-7 DEVITO_OPENMP=0 DEVITO_BACKEND=ops INSTALL_TYPE=conda RUN_EXAMPLES=False MPI_INSTALL=0 OPS_COMPILER=gnu OPS_INSTALL_PATH=$HOME/build/opesci/OPS/ops
 
 addons:
   apt:
--- a/Dockerfile.pipelines
+++ b/Dockerfile.pipelines
@@ -35,6 +35,12 @@ ENV testWithPip=$installWithPip
 ARG DEVITO_BACKEND=none
 ENV DEVITO_BACKEND=$DEVITO_BACKEND
 
+# Set OPS variable in case they are needed
+ARG ops_compiler=gnu
+ARG ops_install_path=/tmp/OPS/ops
+ENV OPS_COMPILER=$ops_compiler
+ENV OPS_INSTALL_PATH=$ops_install_path
+
 # Use OpenMP?
 ARG DEVITO_OPENMP=0
 ENV DEVITO_OPENMP=$DEVITO_OPENMP
@@ -129,6 +135,11 @@ RUN if [ $installWithPip == "true" ] ; then \
           make compiler-api ; \
           pip install -e . ; \
         popd ; \
+      elif [ $DEVITO_BACKEND == "ops" ] ; then \
+        git clone https://github.com/opesci/OPS.git /tmp/OPS ; \
+        cd $OPS_INSTALL_PATH/c ; \
+        make ; \
+        cd $HOME ; \
       fi ; \
       if [ $RUN_EXAMPLES == "true" ] ; then \
         bash scripts/create_ipyparallel_mpi_profile.sh; \
--- a/Dockerfile.pipelines
+++ b/Dockerfile.pipelines
@@ -137,9 +137,9 @@ RUN if [ $installWithPip == "true" ] ; then \
         popd ; \
       elif [ $DEVITO_BACKEND == "ops" ] ; then \
         git clone https://github.com/opesci/OPS.git /tmp/OPS ; \
-        cd $OPS_INSTALL_PATH/c ; \
-        make ; \
-        cd $HOME ; \
+        pushd /tmp/OPS/ops/c ; \
+          make ; \
+        popd ; \
       fi ; \
       if [ $RUN_EXAMPLES == "true" ] ; then \
         bash scripts/create_ipyparallel_mpi_profile.sh; \
--- a/devito/ops/compiler.py
+++ b/devito/ops/compiler.py
@@ -66,7 +66,9 @@ class CompilerOPS(configuration['compiler'].__class__):
         if self._ops_install_path:
             # Calling OPS Translator
             translator = '%s/../ops_translator/c/ops.py' % (self._ops_install_path)
-            subprocess.run([translator, c_file.name], cwd=self.get_jit_dir())
+            translation = subprocess.run([translator, c_file.name], cwd=self.get_jit_dir())
+            if translation.returncode == 1:
+                raise ValueError("OPS Translation Error")
         else:
             warning("Couldn't find OPS_INSTALL_PATH \
                 environment variable, please check your OPS installation")
--- a/devito/ops/compiler.py
+++ b/devito/ops/compiler.py
@@ -7,14 +7,73 @@ from devito.logger import warning
 from devito.parameters import configuration
 
 
-__all__ = ['CompilerOPS']
+__all__ = ['OPSCompiler']
 
 
-class CompilerOPS(configuration['compiler'].__class__):
+class OPSMetaCompiler(configuration['compiler'].__class__):
+
     def __init__(self, *args, **kwargs):
         kwargs['cpp'] = True
+        super(OPSMetaCompiler, self).__init__(*args, **kwargs)
+
+
+class OPSCompiler(OPSMetaCompiler):
+
+    def __init__(self, *args, **kwargs):
+        super(OPSCompiler, self).__init__(*args, **kwargs)
+
+    def jit_compile(self, soname, ccode, hcode):
+
         self._ops_install_path = os.environ.get('OPS_INSTALL_PATH')
-        super(CompilerOPS, self).__init__(*args, **kwargs)
+        if not self._ops_install_path:
+            raise ValueError("Couldn't find OPS_INSTALL_PATH\
+                environment variable, please check your OPS installation")
+
+        self._ops_backend = os.environ.get('OPS_BACKEND')
+        if not self._ops_backend:
+            raise ValueError("Couldn't find OPS_BACKEND\
+                environment variable, current options: `seq` or `cuda`")
+
+        self._translate_ops(soname, ccode, hcode)
+        self.ops_src = '%s/%s_ops.cpp' % (self.get_jit_dir(), soname)
+        self.cache_dir = self.get_jit_dir()
+        self.target = str(self.get_jit_dir().joinpath(soname))
+
+        # Make a suite of cache directories based on the soname
+        self.cache_dir.mkdir(parents=True, exist_ok=True)
+        try:
+            with open(self.ops_src, 'r') as f:
+                self.code = f.read()
+        except FileNotFoundError:
+            warning("Couldn't find file: %s" % self.ops_src)
+
+        if self._ops_backend == 'cuda':
+            OPSCompilerCUDA._compile(self, soname)
+        elif self._ops_backend == 'seq':
+            pass
+
+    def _translate_ops(self, soname, ccode, hcode):
+        # Creating files
+        file_name = str(self.get_jit_dir().joinpath(soname))
+        h_file = open("%s.h" % (file_name), "w")
+        c_file = open("%s.cpp" % (file_name), "w")
+
+        c_file.write(ccode)
+        h_file.write(hcode)
+
+        c_file.close()
+        h_file.close()
+
+        # Calling OPS Translator
+        translator = '%s/../ops_translator/c/ops.py' % (self._ops_install_path)
+        translation = subprocess.run([translator, c_file.name], cwd=self.get_jit_dir())
+        if translation.returncode == 1:
+            raise ValueError("OPS Translation Error")
+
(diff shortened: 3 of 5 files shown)


.travis.yml
```diff
@@ -146,3 +166,7 @@ script:
           pip install doctr;
           doctr deploy . --built-docs docs/_build/html --key-path .deploy_key.enc;
     fi
+  # Test CUDA installation
+  - ls -al /usr/local | grep cuda
+  - export PATH=${CUDA_BIN_PATH}:$PATH
+  - nvcc --version
```
mloubout: these need to be inside a `if [[ $DEVITO_BACKEND == 'ops' ]];`

.travis.yml
```diff
@@ -60,10 +60,14 @@ matrix:
...
             - ubuntu-toolchain-r-test     # For gcc 4.9, 5 and 7
+            - llvm-toolchain-xenial-6.0
           packages:
             - gcc-7
             - g++-7
-      env: DEVITO_ARCH=gcc-7 DEVITO_OPENMP=0 DEVITO_BACKEND=ops YC_CXX=g++-7 INSTALL_TYPE=conda RUN_EXAMPLES=False MPI_INSTALL=0
+            - wget
+            - unzip
+            - tree
+      env: DEVITO_ARCH=gcc-7 DEVITO_OPENMP=0 DEVITO_BACKEND=ops YC_CXX=g++-7 INSTALL_TYPE=conda RUN_EXAMPLES=False MPI_INSTALL=0 CUDA_PATH=/usr/local/cuda CUDA_BIN_PATH=/usr/local/cuda/bin
```
mloubout: Note for myself, this is starting to be way to many env variable there, same for other build, need to rethink it. Part of CI refactoring

.travis.yml
```diff
@@ -60,10 +60,14 @@ matrix:
         apt:
           sources:
             - ubuntu-toolchain-r-test     # For gcc 4.9, 5 and 7
+            - llvm-toolchain-xenial-6.0
           packages:
             - gcc-7
             - g++-7
-      env: DEVITO_ARCH=gcc-7 DEVITO_OPENMP=0 DEVITO_BACKEND=ops YC_CXX=g++-7 INSTALL_TYPE=conda RUN_EXAMPLES=False MPI_INSTALL=0
+            - wget
```
mloubout: Don't need that

.travis.yml
```diff
@@ -60,10 +60,14 @@ matrix:
...
+            - unzip
+            - tree
```
mloubout: would assume these two are default too but not sure

.travis.yml
```diff
@@ -102,6 +113,15 @@ install:
   - if [[ $INSTALL_TYPE == 'pip_setup' ]]; then python setup.py install; fi
   - if [[ $MPI_INSTALL == '1' ]]; then pip install -e .[extras]; fi
 
+  - if [[ $DEVITO_BACKEND == 'ops' ]]; then
+      # Install CUDA
```
mloubout: would do the `export PATH` here rather than in `env`

.travis.yml
btrb: makes sense ! fixing

.travis.yml
btrb: it does indeed

.travis.yml
btrb: ok

.travis.yml
btrb: i can try without those

.travis.yml
btrb: but there should be no OPS_INSTALL_PATH in cases where DEVITO_BACKEND != 'ops'

.travis.yml
```diff
@@ -63,7 +63,7 @@ matrix:
           packages:
             - gcc-7
             - g++-7
-      env: DEVITO_ARCH=gcc-7 DEVITO_OPENMP=0 DEVITO_BACKEND=ops YC_CXX=g++-7 INSTALL_TYPE=conda RUN_EXAMPLES=False MPI_INSTALL=0
+      env: DEVITO_ARCH=gcc-7 DEVITO_OPENMP=0 DEVITO_BACKEND=ops YC_CXX=g++-7 INSTALL_TYPE=conda RUN_EXAMPLES=False MPI_INSTALL=0 OPS_COMPILER=gnu OPS_INSTALL_PATH=$HOME/build/opesci/OPS/ops
```
mloubout: Why is ask `YC_CXX` here?

btrb: have no clue.. found it there. removing !

btrb:
can someone please rerun azure ?

btrb:
> can someone please rerun azure ? i think it should pass

reminder

.travis.yml
```diff
@@ -63,7 +63,7 @@ matrix:
           packages:
             - gcc-7
             - g++-7
-      env: DEVITO_ARCH=gcc-7 DEVITO_OPENMP=0 DEVITO_BACKEND=ops YC_CXX=g++-7 INSTALL_TYPE=conda RUN_EXAMPLES=False MPI_INSTALL=0
+      env: DEVITO_ARCH=gcc-7 DEVITO_OPENMP=0 DEVITO_BACKEND=ops INSTALL_TYPE=conda RUN_EXAMPLES=False MPI_INSTALL=0 OPS_COMPILER=gnu OPS_INSTALL_PATH=$HOME/build/opesci/OPS/ops
```
FabioLuporini: how about
```
OPS_INSTALL_PATH=$HOME/OPS
```
? 

why should it go under `opesci` afterall?

devito/ops/compiler.py
```diff
@@ -7,39 +7,35 @@
...
-            [self.cc]
-            + self.cflags
-            + ld_options
-            + ["-D%s" % define for define in self.defines]
-            + ["-U%s" % undefine for undefine in self.undefines]
-            + ["-I%s" % idir for idir in self.include_dirs]
-            + files
-            + link
-        )
+        self._devito_platform = os.environ.get('DEVITO_PLATFORM')
```
FabioLuporini: I think you can drop this entire block and ... (see next comment below)

devito/ops/compiler.py
```diff
@@ -48,8 +44,11 @@ def jit_compile(self, soname, ccode, hcode):
                 self.code = f.read()
         except FileNotFoundError:
             warning("Couldn't find file: %s" % self.ops_src)
-        else:
-            self._compile_cuda(soname)
+
+        if self._devito_platform == 'nvidiaX':
```
FabioLuporini: ... here, you should just look at `configuration['platform']`. No need for `_devito_platform`

'seq' should be replaced by 'intel64' or something along these lines (see [here](https://github.com/opesci/devito/blob/master/devito/archinfo.py#L249) )

devito/ops/compiler.py
```diff
@@ -63,33 +62,77 @@ def _translate_ops(self, soname, ccode, hcode):
...
             raise ValueError("Couldn't find file: %s" % cuda_src)
 
-        cuda_device_compiler = CUDADeviceCompiler()
-        cuda_host_compiler = CudaHostCompiler()
-
         with warnings.catch_warnings():
             warnings.simplefilter('ignore')
 
             # Spinlock in case of MPI
+            cuda_device_compiler = CUDADeviceCompiler()
```
FabioLuporini: It makes totally no sense to instantiate an object of the subclass in the superclass

this is just voodoo that Python lets you doing, but it simply is a software engineering error

.travis.yml
btrb: i thought that by doing that i could stick with the pattern from the section `before_scrip`
that way, the part for ops could be very similar to the part for yask
otherwise i guess i would need something like:
`cd $HOME/build/opesci/devito;`
after making for ops (see line 114)

devito/ops/compiler.py
btrb: ok, fixing

devito/ops/compiler.py
btrb: so is there a problem in line 45 of `ops/compiler.py` as well?
`OPSCompilerCUDA._compile(self, soname)` is called from `OPSCompiler`

.travis.yml
FabioLuporini: Fine, then it could go under $HOME/opesci , just "next to" yask ?

also, can you drop OPS_COMPILER? 'gnu' should be anyway the default choice, so it's not adding anything I'd say

Dockerfile.pipelines
```diff
@@ -35,6 +35,12 @@ ENV testWithPip=$installWithPip
 ARG DEVITO_BACKEND=none
 ENV DEVITO_BACKEND=$DEVITO_BACKEND
 
+# Set OPS variable in case they are needed
+ARG ops_compiler=gnu
+ARG ops_install_path=/tmp/OPS/ops
+ENV OPS_COMPILER=$ops_compiler
```
FabioLuporini: probably you can drop this, for the same reasons as above? what's its purpose otherwise?

Dockerfile.pipelines
```diff
@@ -129,6 +135,11 @@ RUN if [ $installWithPip == "true" ] ; then \
           make compiler-api ; \
           pip install -e . ; \
         popd ; \
+      elif [ $DEVITO_BACKEND == "ops" ] ; then \
+        git clone https://github.com/opesci/OPS.git /tmp/OPS ; \
```
FabioLuporini: so...you clone under `/tmp`, but you build under `$HOME`. Did I get it right? Not sure why, but fine

FabioLuporini: this seems different than what happens on Travis

devito/ops/compiler.py
```diff
@@ -48,8 +40,11 @@ def jit_compile(self, soname, ccode, hcode):
                 self.code = f.read()
         except FileNotFoundError:
             warning("Couldn't find file: %s" % self.ops_src)
-        else:
-            self._compile_cuda(soname)
+
+        if configuration['platform'] == 'nvidiaX':
```
FabioLuporini: really? shouldn't it be
```
if configuration['platform'] is NVIDIAX
```

like [here](https://github.com/opesci/devito/blob/e54d1c83b9d69ca406c5e9e718349b5d4066bd1b/devito/compiler.py#L370)
(note that the `==` there should and could actually be `is` -- exercise for you, understand why. Not that the `==` comparison is wrong, but the `is` comparison is conceptually better)

devito/ops/compiler.py
FabioLuporini: that is totally fine in object oriented programming (OOP). You should read something about polymorphism in OOP  (@maelso can you help him out please) 

the error I described above is that a superclass shouldn't know anything about what/who the subclasses are, while the converse is fine

devito/ops/operator.py
```diff
@@ -37,7 +36,10 @@ def _specialize_iet(self, iet, **kwargs):
 
         # If there is no affine trees, then there is no loop to be optimized using OPS.
         if not affine_trees:
-            return iet
+            self._should_compile_with_ops = False
```
FabioLuporini: this flag is a hack, I'd guess? If you had a tree with some loops that can be offloaded, and some loops that cannot, what would you do with it?

.travis.yml
btrb: this env var is required by ops` make
it has to be set

Dockerfile.pipelines
btrb: this env var is required by ops` makefile

Dockerfile.pipelines
btrb: yes, indeed. it's a different path than from Travis
i just reproduced the pattern i've found on their respective scripts using yask as a template

in Travis yask is cloned into `$/HOME/build/opesci/yask` and in pipelines it is cloned into `/tmp/yask`

devito/ops/operator.py
btrb: if a tree has at least one loop that can be offloaded, then this will be enough for building a list of kernels to be ops`fied. 
therefore, there will be a `self.hcode`, and thus there will be no problem calling `self._compiler.jit_compile(self._soname, str(self.ccode), str(self.hcode))`.

only if there aren't any loops at all that can be offloaded there will be no `str(self.hcode)` to serve as the 3rd param for ops` `jit_compile`

devito/ops/compiler.py
btrb: thank you, fixed

btrb:
applied requested changes, and rebased to master
i think there are a couple of discussions to be agreed, but i think it's ok to go
CI should be green

devito/ops/compiler.py
btrb: as far as i can see, CUDADeviceCompiler is not a subclass of OPSCompiler, but more like a 'sister' class. i'm not sure if that makes sense.
 
i'm uploading how i understand their relation:
WhatsApp Image 2019-12-11 at 16 21 23

maelso: You moved `cuda_device_compiler = CUDADeviceCompiler()` from **OPSCompilerCUDA** to **OPSCompiler**, that change solved the problem mentioned by Fabio. Now, there is an **association** between OPSCompiler and OPSCompilerCUDA.

devito/ops/compiler.py
```diff
@@ -63,33 +58,28 @@ def _translate_ops(self, soname, ccode, hcode):
...
         # CUDA kernel compilation
         cuda_src = '%s/CUDA/%s_kernels.cu' % (self.get_jit_dir(), soname)
         cuda_target = '%s/%s_kernels_cu' % (self.get_jit_dir(), soname)
 
-        cuda_code = ""
         try:
             with open(cuda_src, 'r') as f:
                 cuda_code = f.read()
         except FileNotFoundError:
             raise ValueError("Couldn't find file: %s" % cuda_src)
```
maelso: It can still get better (for currently or future implementations).
cuda_src, cuda_target, etc. should stay in OPSCompilerCUDA. The same goes for the others (seq_src, openmp_src, etc).

devito/ops/compiler.py
```diff
@@ -107,31 +102,77 @@ def _compile_cuda(self, soname):
...
+            if not os.path.exists(src_o):
+                raise ValueError("Error when compiling for host\
+                                 Couldn't find file: %s" % src_o)
+
+            try:
+                cuda_host_compiler.link_extension(
+                    '%s%s' % (self.target, cuda_host_compiler.so_ext),
+                    [src_o, cuda_o],
+                    debug=configuration['debug-compiler']
+                )
```
maelso: Same as above.

maelso: The `_compile` function shouldn't be in CompilerOPS. Each target architecture must have its implementation.

devito/ops/compiler.py
```diff
@@ -5,41 +5,33 @@
...
-            + ["-D%s" % define for define in self.defines]
-            + ["-U%s" % undefine for undefine in self.undefines]
-            + ["-I%s" % idir for idir in self.include_dirs]
-            + files
-            + link
-        )
+
+class OPSCompiler(OPSMetaCompiler):
+
+    def __init__(self, *args, **kwargs):
```
FabioLuporini: this init is useless. It's just calling the superclass', which happens automatically if you don't override it

devito/ops/compiler.py
```diff
@@ -5,41 +5,33 @@
...
+        super(OPSCompiler, self).__init__(*args, **kwargs)
 
     def jit_compile(self, soname, ccode, hcode):
         self._translate_ops(soname, ccode, hcode)
-        self.target = str(self.get_jit_dir().joinpath(soname))
         self.ops_src = '%s/%s_ops.cpp' % (self.get_jit_dir(), soname)
-        self.cache_dir = self.get_codepy_dir().joinpath(soname[:7])
+        self.cache_dir = self.get_jit_dir()
```
FabioLuporini: it's a bad practice to **create** instance variables in methods that are not the constructor (or any other method called directly by the constructor).

devito/ops/compiler.py
```diff
@@ -48,8 +40,11 @@ def jit_compile(self, soname, ccode, hcode):
                 self.code = f.read()
         except FileNotFoundError:
             warning("Couldn't find file: %s" % self.ops_src)
-        else:
-            self._compile_cuda(soname)
+
+        if configuration['platform'] is NVIDIAX:
+            self._compile(soname)
+        elif configuration['platform'] == 'bdw':
```
FabioLuporini: why is this `elif` necessary?

devito/ops/compiler.py
FabioLuporini: yes, I agree. It's indeed a violation that concepts of the subclass appear in the superclass

devito/ops/compiler.py
```diff
@@ -63,33 +58,28 @@ def _translate_ops(self, soname, ccode, hcode):
...
+        # Calling OPS Translator
+        translator = '%s/../ops_translator/c/ops.py' % (self._ops_install_path)
+        translation = subprocess.run([translator, c_file.name], cwd=self.get_jit_dir())
+        if translation.returncode == 1:
+            raise ValueError("OPS Translation Error")
 
-    def _compile_cuda(self, soname):
+    def _compile(self, soname):
         # CUDA kernel compilation
         cuda_src = '%s/CUDA/%s_kernels.cu' % (self.get_jit_dir(), soname)
```
FabioLuporini: this should all go in the CUDA subclass...

devito/ops/compiler.py
FabioLuporini: I'm not sure this association makes much sense. If you:

* move the cuda-specific stuff from OPSCompiler to OPSCompilerCuda (or any of its subclasses)
* move the compilation-specific stuff to OPSMetaCompiler (to be renamed as OPSCompiler)

wouldn't it become cleaner and simpler? What's your take, @maelso ?

devito/ops/operator.py
FabioLuporini: If you support the _general_ case of an Opertor that contains **both** an affine and a number of non-affine nests, then I don't see why in practice you shouldn't be able to support the two special cases

* only affine nests are present
* only non-affine (ie non-offloadable) nests are present

FabioLuporini: if you run a plain C code through the OPS compiler/code generator, w/o any of the OPS constructs, it should just produce/return the same input code, unmodified

FabioLuporini:
As discussed offline with @btrb , closing for now as nobody will be working on this for some time

PR #1000, opened 2019-11-13
Allow nbl to be zero in initialize_function

rhodrin:
This small PR:
- Allows `nbl` to be zero when calling `initialize_function` hence allowing a `Model` be created with no damping region (as was previously the case).
- If a `Model` is created with no damping region a damping field is now not created.
- A test has been added for when `initialize_function` is called with `nbl=0`.

### DIFF
--- a/devito/builtins.py
+++ b/devito/builtins.py
@@ -277,7 +277,7 @@ def initialize_function(function, data, nbl, mapper=None, mode='constant',
         if isinstance(data, dv.Function):
             function.data[:] = data.data[:]
         else:
-            function.data[:] = data
+            function.data[:] = data[:]
         return
 
     if len(as_tuple(nbl)) == 1 and len(as_tuple(nbl)) < function.ndim:
--- a/devito/builtins.py
+++ b/devito/builtins.py
@@ -273,6 +273,13 @@ def initialize_function(function, data, nbl, mapper=None, mode='constant',
     if isinstance(function, dv.TimeFunction):
         raise NotImplementedError("TimeFunctions are not currently supported.")
 
+    if nbl == 0:
+        if isinstance(data, dv.Function):
+            function.data[:] = data.data[:]
+        else:
+            function.data[:] = data
+        return
+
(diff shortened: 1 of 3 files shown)


examples/seismic/model.py
```diff
@@ -499,9 +499,10 @@ def __init__(self, origin, spacing, shape, space_order, nbl=20,
         self.grid = Grid(extent=extent, shape=shape_pml, origin=origin_pml, dtype=dtype,
                          subdomains=subdomains)
 
-        # Create dampening field as symbol `damp`
-        self.damp = Function(name="damp", grid=self.grid)
-        initialize_damp(self.damp, self.nbl, self.spacing, mask=damp_mask)
+        if self.nbl != 0:
+            # Create dampening field as symbol `damp`
+            self.damp = Function(name="damp", grid=self.grid)
+            initialize_damp(self.damp, self.nbl, self.spacing, mask=damp_mask)
```
mloubout: still needs `else` part
```
else:
    self.damp = 1 if damp_mask else 0
```
so that examples work too with nbl=0 e.g:
`python examples/seismic/acoustic/acoustic_example.py -nbl 0`

devito/builtins.py
```diff
@@ -273,6 +273,13 @@ def initialize_function(function, data, nbl, mapper=None, mode='constant',
     if isinstance(function, dv.TimeFunction):
         raise NotImplementedError("TimeFunctions are not currently supported.")
 
+    if nbl == 0:
+        if isinstance(data, dv.Function):
+            function.data[:] = data.data[:]
+        else:
+            function.data[:] = data
```
mloubout: `= data[:]`? I think without the `:` gets error in some weird case with empty dimensions

rhodrin: I think it's normally fine if `data` is a numpy array but certainly doesn't hurt so will add.

examples/seismic/model.py
rhodrin: Ah yes, thanks.

PR #991, opened 2019-10-31
Fixing division by pointer bug

Gabrielsvc:
When translating to OPS a division by pointer (like `/*dt`) would make the
rest of the code commentary in c++. This was solved by encasing
the pointer by parenthesis.

PR #1002, opened 2019-11-15
ops/node_factory: refactoring new_ops_gbl

btrb:
This PR is an updated version of #991 .

From #991:
When translating to OPS a division by pointer (like /*dt) would make the
rest of the code commentary in c++. This was solved by encasing the 
pointer by parenthesis.

In this PR, requested changes are made upon updated master.

### DIFF
--- a/devito/ops/node_factory.py
+++ b/devito/ops/node_factory.py
@@ -9,9 +9,9 @@ from devito.ops.utils import AccessibleInfo
 
 class OPSNodeFactory(object):
     """
-    Generate OPS nodes for building an OPS expression.
+    Generate OPS nodes for building OPS expressions.
     A new OPS argument is created based on the indexed name, and its time dimension.
-    Such a pair identifies an unique argument within the OPS kernel. The function
+    Such a pair identifies an unique argument within an OPS kernel. The function
     returns the stored argument associated with this pair, if it was already created.
     """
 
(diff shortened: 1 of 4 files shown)


btrb:
in case the changes proposed in this PR are not acceptable, i thought it could be refactored into a class, something like:

```
class OPSNodeGbl(OPSNodeFactory):

    def __init__(self, expr, **kwargs):
        super(OPSNodeGlb, self).__init__(**kwargs)
        self.expr = expr

        if self.expr not in self.ops_args:
            self.ops_args[self.expr] = AccessibleInfo(Constant(name='*%s' % self.expr.name,
                                                               dtype=self.expr.dtype),
                                                      None,
                                                      None)
            self.ops_params.append(self.ops_args[self.expr].accessible)

    def get_accessible(self):
        return Constant(name='(*%s)' % self.expr.name, dtype=self.expr.dtype)
```

to be used in `transformer.py` at `make_ops_ast` like:

```
        if expr.is_Constant:
            return OPSNodeGlb(expr).get_accessible()
```

devito/ops/node_factory.py
```diff
@@ -68,14 +68,25 @@ def new_ops_arg(self, indexed, is_write):
...
+               Constant expression.
 
-        return new_c.accessible
+        Returns
+        -------
+        Constant
+            Constant symbol, as a pointer.
+        """
+        if expr not in self.ops_args:
+            self.ops_args[expr] = AccessibleInfo(Constant(name='*%s' % expr.name,
```
FabioLuporini: hack

btrb: is it so because there is a `*` in the `name` of a `Constant`?

FabioLuporini: yes

you should try to figure how devito manages to generate all those stars

FabioLuporini:
Temporarily closing as the proposed fix is still hacky (https://github.com/opesci/devito/pull/1002/files#diff-4e2d887ad1748ef17a61a5f65431af82R91)

Issue #1007, opened 2019-11-19
Accumulating into data slices doesn't work with MPI


mikeando:
The following small program errors when run with 2 MPI ranks. The error is:

~~~
ValueError: could not broadcast input array from shape (16,31) into shape (0,30,31)
ValueError: could not broadcast input array from shape (16,31) into shape (16,30,31)
~~~

~~~
from devito import Function, Grid, TimeFunction, Eq, Operator, configuration
from mpi4py import MPI
configuration['mpi'] = True

grid = Grid(shape=(31,31,31), comm=MPI.COMM_WORLD)
f = Function(name="f", grid=grid)
f.data[:] = 1

#OK
print("Adding f[1,:,:] to f[0,:,:]")
f.data[1,:,:] = f.data[1,:,:] + f.data[0,:,:]
f.data[1,:,:] += f.data[0,:,:]

#Neither of these is OK
print("Adding f[:,1,:] to f[:,0,:]")
f.data[:,1,:] = f.data[:,1,:] + f.data[:,0,:]
f.data[:,1,:] += f.data[:,0,:]
~~~

rhodrin:
Thanks for the report @mikeando. Looking into this at the moment, should have a fix at some point today.

rhodrin:
This should be fixed on https://github.com/opesci/devito/tree/misc-data-fixes.

FabioLuporini:
closing as fixed by #1008

Issue #1006, opened 2019-11-19
Negative index slicing seems to behave weird on devito.Data


navjotk:
```
grid = Grid(shape=(20,))
u = TimeFunction(name='u', save=None, grid=grid, space_order=0, time_order=1)

print(u.data[1, -2:])
```

produces the error: `IndexError: index 20 is out of bounds for axis 0 with size 2`

but the error does not happen on numpy objects. Also, it only happens when printing.

mloubout:
Looks like `__repr__` is fine but `__str__` is not

rhodrin:
Thanks. Will look into this today.

rhodrin:
So I assume this must have been a problem for a long time then? Basically the 'local indices' don't get passed correctly meaning you get out of bounds errors. This is fixed by adding:
```
    def __str__(self):
        return super(Data, self._local).__str__()
```
I'll include the above in a PR which also fixes #1007.

rhodrin:
Fixed on https://github.com/opesci/devito/tree/misc-data-fixes. Will be PR'ing after adding a few more tests.

FabioLuporini:
closing as fixed by #1008

PR #1008, opened 2019-11-20
Misc data fixes

rhodrin:
This PR addresses issues #1006 and #1007.

### DIFF
--- a/devito/data/data.py
+++ b/devito/data/data.py
@@ -431,12 +431,12 @@ class Data(np.ndarray):
             else:
                 data_glb_idx.append(None)
         mapped_idx = []
-        # Testing block:
+        # Add any integer indices that were not present in `val_idx`.
         if len(as_list(idx)) > len(data_glb_idx):
             for index, value in enumerate(idx):
-                if index > 0 and is_integer(value):
+                if is_integer(value) and index > 0:
                     data_glb_idx.insert(index, value)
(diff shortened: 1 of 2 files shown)


devito/data/data.py
```diff
@@ -170,6 +170,9 @@ def _is_mpi_distributed(self):
     def __repr__(self):
         return super(Data, self._local).__repr__()
 
+    def __str__(self):
+        return super(Data, self._local).__str__()
```
mloubout: nitpicking: `__str__ = __repr__` works to

rhodrin: But this will result in slightly different outputs right? So with `__str__ = __repr__`, `print(f.data[some_slice])` will result in `Data([some_array], dtype=float32)` being printed. Where as with the current code you'd get only `[some_array]` (which mimics the numpy array behavior).

FabioLuporini: Rhodri is right. I was tricked too for a second like Mathias; note that `__repr__` is calling `.__repr__()`, while `__str__`  is calling `.__str__()`

FabioLuporini:
I think this is uncontroversial and should be merged as soon as green

devito/data/data.py
mloubout: Oh right my bad, always bit confused but the difference. GTG

Issue #908, opened 2019-08-08
OPS backend breaking during create_ops_arg 


maelso:
**Traceback** (most recent call last):
  File "./examples/test/forward.py", line 80, in <module>
    subs=model.spacing_map, name='Forward')
  File "/media/cdd4a77d-df70-4379-b548-d27951f6e9c6/Documentos/devito/devito/backends.py", line 73, in __call__
    return t(*args, **kwargs)
  File "/media/cdd4a77d-df70-4379-b548-d27951f6e9c6/Documentos/devito/devito/ops/operator.py", line 25, in __init__
    super().__init__(*args, **kwargs)
  File "/media/cdd4a77d-df70-4379-b548-d27951f6e9c6/Documentos/devito/devito/operator.py", line 185, in __init__
    iet = self._specialize_iet(iet, **kwargs)
  File "/media/cdd4a77d-df70-4379-b548-d27951f6e9c6/Documentos/devito/devito/ops/operator.py", line 69, in _specialize_iet
    trees, n, name_to_ops_dat, ops_block, dims[0]
  File "/media/cdd4a77d-df70-4379-b548-d27951f6e9c6/Documentos/devito/devito/ops/transformer.py", line 61, in opsit
    name_to_ops_dat, par_to_ops_stencil, dims
  File "/media/cdd4a77d-df70-4379-b548-d27951f6e9c6/Documentos/devito/devito/ops/transformer.py", line 231, in create_ops_par_loop
    *[create_ops_arg(p, name_to_ops_dat, par_to_ops_stencil) for p in parameters]
  File "/media/cdd4a77d-df70-4379-b548-d27951f6e9c6/Documentos/devito/devito/ops/transformer.py", line 231, in <listcomp>
    *[create_ops_arg(p, name_to_ops_dat, par_to_ops_stencil) for p in parameters]
  File "/media/cdd4a77d-df70-4379-b548-d27951f6e9c6/Documentos/devito/devito/ops/transformer.py", line 248, in **create_ops_arg
    name_to_ops_dat[p.name]**,
**KeyError: 'utime'**

### **Test file:** 
https://github.com/maelso/devito/blob/local_tests/examples/test/forward.py

vincepandolfo:
As said on slack, I believe this is caused by having `save` enabled in your example, which is not currently supported by the OPS backend

maelso:
I understood.
It's an open issue because the backend breaks.

vincepandolfo:
[...]

PR #1011, opened 2019-11-22
ops: Enabling save parameter

maelso:
The `save` parameter is not currently supported by the OPS backend. [issue #908](https://github.com/opesci/devito/issues/908)

_u = TimeFunction(name='u', grid=grid, **save=nt**)_

**Equation:**
`Eq(u.forward, u + 1)`

**Currently code:**
`ops_par_loop(OPS_Kernel_0,"OPS_Kernel_0",block,2,(int *)OPS_Kernel_0_range,ops_arg_dat(u_dat[time], 1, "float", OPS_WRITE));`
```
void OPS_Kernel_0(ACC<float> & utime)
{
  utime(0, 0) = 1 + utime(0, 0);
}
```

**New Version:**
`ops_par_loop(OPS_Kernel_0,"OPS_Kernel_0",block,2,(int *)OPS_Kernel_0_range,ops_arg_dat(u_dat[time], 1, "float", OPS_READ),ops_arg_dat(u_dat[time + 1], 1, "float", OPS_WRITE));`
```
void OPS_Kernel_0(const ACC<float> & utime0, ACC<float> & utime1)
{
  utime1(0, 0) = 1 + utime0(0, 0);
}
```

### DIFF
--- a/tests/test_ops.py
+++ b/tests/test_ops.py
@@ -152,16 +152,16 @@ class TestOPSExpression(object):
          '\'v_dat = ops_decl_dat(block, 1, v_dim, v_base, v_d_m, v_d_p, '
          '&(v[0]), "float", "v")\']'),
         ('Eq(w1.forward, w1 + 1)',
-         '[\'ops_dat w1_dat[2] = {ops_decl_dat(block, 1, w1_dim, w1_base, w1_d_m, w1_d_p, '
-         '&(w1[0]), "float", "w1time0"), ops_decl_dat(block, 1, w1_dim, w1_base, w1_d_m, w1_d_p, '
-         '&(w1[1]), "float", "w1time1")}\']'),
+         '[\'ops_dat w1_dat[2] = {ops_decl_dat(block, 1, w1_dim, w1_base, w1_d_m, '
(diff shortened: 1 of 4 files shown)


tests/test_ops.py
```diff
@@ -223,7 +239,7 @@ def test_create_ops_arg_function(self, read):
         u = OpsAccessible('u', dtype=np.float32, read_only=read)
         dat = OpsDat('u_dat')
         stencil = OpsStencil('stencil')
-        info = AccessibleInfo(u, None, None)
+        info = AccessibleInfo(u, None, None, None)
```
FabioLuporini: @btrb for later PR... namedtuple can be engineering to use default arguments. Could be handy since many times most arguments are just None

FabioLuporini:
LGTM

mloubout:
So, just t one sure, this still requires to create one `Function` and one function call in the code per time step, right? Still think may not be the best thing, but PR looks fine to me otherwise

FabioLuporini:
merged

tests/test_ops.py
btrb: oki

PR #1012, opened 2019-11-24
Fixups for OpenMP-based GPU offloading

FabioLuporini:
- deterministic code generation
- collapse more loops
- minor refactoring
- more tests

### DIFF
--- a/devito/dle/rewriters.py
+++ b/devito/dle/rewriters.py
@@ -514,34 +514,6 @@ class SpeculativeRewriter(CPU64Rewriter):
         self._minimize_remainders(state)
         self._hoist_prodders(state)
 
-    @dle_pass
-    def _nontemporal_stores(self, nodes):
-        """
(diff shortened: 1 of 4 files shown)


devito/dle/rewriters.py
```diff
@@ -377,12 +380,15 @@ def _simdize(self, iet):
...
-        Add OpenMP pragmas to the Iteration/Expression tree to emit shared-memory
-        parallel code.
+        Add OpenMP pragmas to the Iteration/Expression tree to emit SIMD and
+        shared-memory parallel code.
         """
-        return self._node_parallelizer.make_parallel(iet)
+        if self.params['openmp']:
+            return self._parallelizer_shm.make_parallel(iet)
+        else:
+            return iet, {}
```
mloubout: I think this is what make the test fail as with this (more rigorous) update, no pragmas are generated at all without openmp now. Not sure if you wanna go back to "serial pragmas" or add a "skip" openmp.

Issue #998, opened 2019-11-08
Operators dont get their distributor information from the current grid


mikeando:
At the moment it seems that the MPI information used by the operator are baked into it when it is constructed, rather than taken from its arguments.

There are checks in place that all the arguments are on the same grid - but this may have a different distributor than the operator was built with. 

Currently I can work around this in my examles by doing:

~~~
    m={}
    m['comm'] = u.grid.distributor._obj_comm
    m['nb'] = u.grid.distributor._obj_neighborhood
    op.objects = tuple( m.get(o.name, o) for o in op.objects)
    op.apply(u=u)
~~~

but it feels pretty hacky.

A complete example can be found at 

https://gist.github.com/mikeando/35f661d13b374e857315859cd0d975d6

mikeando:
I tried fixing this by patching operator.py to contain something like

~~~
        # Process Objects (which may need some `args`)
        for o in self.objects:
            args.update(o._arg_values(args, **kwargs))
+
+    if grid is not None:
+        args.update(grid._arg_defaults())
~~~

but was unable to get that to work.

mikeando:
For the MPI + cloudpickle version in that example to run devito/mpi/distributed.py must be patched to create SELF based distribtors rather than WORLD by default. That is done using the following patch.

~~~
diff --git a/devito/mpi/distributed.py b/devito/mpi/distributed.py
index 8801da2f..e607f055 100644
--- a/devito/mpi/distributed.py
+++ b/devito/mpi/distributed.py
@@ -181,7 +181,7 @@ class Distributor(AbstractDistributor):
             if not MPI.Is_initialized():
                 MPI.Init()
 
-            self._input_comm = (input_comm or MPI.COMM_WORLD).Clone()
+            self._input_comm = (input_comm or MPI.COMM_SELF).Clone()
 
             # Make sure the cloned communicator will be freed up upon exit
             def cleanup():
~~~~

rhodrin:
@mikeando,

Thanks for opening this issue. It will be looked into soon.

rhodrin:
@mikeando Just a quick update. I've been looking into this today and the problem is probably to do with `Operator` caching some things that it shouldn't resulting in `comm` and `nb` not being updated 'automatically' as they should be. Pretty sure I know roughly the area where some changes are needed to fix this, just need to figure out the best way of doing so.

rhodrin:
@mikeando When you run the example code you pasted above what run times are you seeing for each of the Operators?
[...]

PR #1016, opened 2019-11-27
Update `comm` and `nb` argument values from grid

rhodrin:
This small PR fixes issue #998. Operator objects `comm` and `nb` now have their argument values correctly updated if a grid with a new distributor is 'passed' to the operator.

Note: Still need to add some comments in the code.

### DIFF
--- a/devito/mpi/distributed.py
+++ b/devito/mpi/distributed.py
@@ -470,6 +470,7 @@ class MPICommObject(Object):
 
     def _arg_values(self, *args, **kwargs):
         grid = kwargs.get('grid', None)
+        # Update `comm` based on object attached to `grid`
         if grid is not None:
             return grid.distributor._obj_comm._arg_defaults()
         else:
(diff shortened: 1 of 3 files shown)


rhodrin:
Will also add a test for this.

FabioLuporini:
this is the place where to add the test: https://github.com/opesci/devito/blob/master/tests/test_operator.py#L965

you may wanna look at the tests above in that class

otherwise, this LGTM

Thanks!

FabioLuporini:
history should be rebased such that it's one commit only (plus one commit for the test optionally)

FabioLuporini:
Merging because I want this fix in ASAP, but:

* Commit history should and could have been better
* You turned `args` into `*args` in the `_arg_values` signature and then in `operator.py` you're passing `args=args` which works around the `args->*args` change. Perhaps some tidying for a future PR

FabioLuporini:
Ah no, sorry: this is still lacking a test. Will wait a little longer then

rhodrin:
Yeah, don't merge yet please. Will tidy up this morning.

p.s. I know the commit history needs to be cleaned up. It was going to be done along with the final commit.

PR #1017, opened 2019-11-28
Replace _ops_kernels by _func_table in OPS    

vmickus:
OPS was using a list to store all the kernels that will go into the .h file, but Devito already has one dictionary for this.

### DIFF
--- a/devito/ops/operator.py
+++ b/devito/ops/operator.py
@@ -112,9 +112,10 @@ class OperatorOPS(Operator):
         return ''.join(str(kernel.root) for kernel in self._func_table.values())
 
     def _finalize(self, iet, parameters):
-
-        self._func_table.update({k: MetaCall(super(OperatorOPS, self)._finalize(v.root, derive_parameters(v.root, True)), v.local) # noqa
-            for k, v in self._func_table.items()})
+        for k, v in self._func_table.items():
+            parameters = derive_parameters(v.root, True)
+            root = super(OperatorOPS, self)._finalize(v.root, parameters)
+            self._func_table[k] = MetaCall(root, v.local)
 
         return super()._finalize(iet, parameters)
 
--- a/devito/ops/operator.py
+++ b/devito/ops/operator.py
@@ -1,6 +1,7 @@
 from devito import Eq
 from devito.ir.equations import ClusterizedEq
-from devito.ir.iet import Call, Expression, List, derive_parameters, find_affine_trees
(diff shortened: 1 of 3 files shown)


FabioLuporini:
uncontroversial, thanks

devito/ops/operator.py
```diff
@@ -108,13 +109,13 @@ def _specialize_iet(self, iet, **kwargs):
 
     @property
     def hcode(self):
-        return ''.join(str(kernel) for kernel in self._ops_kernels)
+        return ''.join(str(kernel.root) for kernel in self._func_table.values())
 
     def _finalize(self, iet, parameters):
         # Applies _finalize for each generated OPS kernel, which will generate the .h file
         self._ops_kernels = [super(OperatorOPS, self)._finalize(
```
FabioLuporini: this PR looks broken, or your testing is broken in a bad way

here, there's still an occurrence of `_ops_kernels`  and nothing is failing

vmickus: Thats because I was using the `func_table` as a temporary holder, and in the end I was still using the `_ops_kernels` list as you have pointed in here.

I have pushed a fix now. It seems to be working, and I generate the code locally and the .h file seems to be ok.

devito/ops/operator.py
```diff
@@ -108,13 +109,12 @@ def _specialize_iet(self, iet, **kwargs):
...
-        return ''.join(str(kernel) for kernel in self._ops_kernels)
+        return ''.join(str(kernel.root) for kernel in self._func_table.values())
 
     def _finalize(self, iet, parameters):
-        # Applies _finalize for each generated OPS kernel, which will generate the .h file
-        self._ops_kernels = [super(OperatorOPS, self)._finalize(
-            kernel, derive_parameters(kernel, True))
-            for kernel in self._ops_kernels]
+
+        self._func_table.update({k: MetaCall(super(OperatorOPS, self)._finalize(v.root, derive_parameters(v.root, True)), v.local) # noqa
```
vmickus: I accept suggestion for this line here... I tried to separate but it got weird and confusing.

FabioLuporini:
```
for k, v in list(self._func_table.items()):
    parameters = derive_parameters(v.root, True)
    root = super(OperatorOPS, self)._finalize(v.root, parameters)
    self._func_table[k] = MetaCall(root, v.local)
```
?

vmickus: I was only thinking about list comprehension, but this seems better.

FabioLuporini:
Thanks, merged.

PR #1022, opened 2019-12-03
Topo fix

mloubout:
Small fi of `_topological_sort` when there is multiple equations and temporaries that needs to stat between equation.

Added test with it


devito/dse/manipulation.py
```diff
@@ -284,14 +284,25 @@ def _topological_sort(exprs):
...
                 continue
             else:
                 dag.add_edge(mapper[r], e, force_add=True)
     processed = dag.topological_sort()
 
     # Append tensor equations at the end in user-provided order
-    processed.extend(tensors)
+    processed_ordered = []
+    for p in processed:
+        # Tensor equation was required for the sorting, add all up to that
```
rhodrin: This comment is unclear to me and could do with some re-wording.

mloubout: Tried to be clearer

devito/dse/manipulation.py
```diff
@@ -284,14 +284,29 @@ def _topological_sort(exprs):
...
+                # Cyclic dependencies stay "as is" ie
+                # [Eq(r1, u + 1), Eq(u, r1 + 2)]
                 continue
             else:
                 dag.add_edge(mapper[r], e, force_add=True)
     processed = dag.topological_sort()
 
     # Append tensor equations at the end in user-provided order
-    processed.extend(tensors)
+    processed_ordered = []
```
FabioLuporini: OK after a very quick look... what I don't understand why is it not enough to add a few more edges to the DAG? 

Also, not sure if useful or not, but note that dag's `topological_sort` takes an optional argument `callback` to steer sorting

I think it's not neat that half of the algorithm is with DAG and half is post-processing. I'm fairly confident we can fix the bug with DAG+toposort.

As I said, I don't have much time these days, but I'll be happy to take a look on monday

tests/test_dse.py
```diff
@@ -20,6 +21,42 @@
 pytestmark = skipif(['yask', 'ops'])
 
 
+@pytest.mark.parametrize('expected', [
```
FabioLuporini: instead of writing out the entire Eqs, what if we had just the LHSs? 

```
@pytest.mark.parametrize('expected', [(r3, r2, d(t,x,y), a(...), ...)])
```

this would make it much simpler to add new cases

I think also this test is nice, but we could add some easier cases before this one. Indeed, the four equations [eq1, eq4] should be part of the parametrization as well

devito/dse/manipulation.py
mloubout: >I'm fairly confident we can fix the bug with DAG+toposort.

I'm not but feel free to try.

tests/test_dse.py
mloubout: I prefer it like that, it's a bit hard to figure out what's going on with just he LHS names.

FabioLuporini: well this makes the test hard to extend... and AFAICT the RHSs do not add any information ?

mloubout: It makes it readable, can't understand/see why that's supposed to be the order without the res, so if this test fail in the future people may think just needs to reorder the list and all good.

mloubout: It makes it readable, can't understand/see why that's supposed to be the order without the RHS, so if this test fail in the future people may think just needs to reorder the list and all good.

devito/dse/manipulation.py
```diff
@@ -284,14 +284,29 @@ def _topological_sort(exprs):
...
+            # up to that one to satisfy order
+            j = tensors.index(p)
+            for t in tensors[:j+1]:
+                if t not in processed_ordered:
+                    processed_ordered.extend([t])
+            tensors = tensors[j+1:]
+        else:
+            # Temporary, nothing to do
+            processed_ordered.extend([p])
+    # Add remainging tensor equations
```
georgebisbas: nitpicking typo

tests/test_dse.py
```diff
@@ -20,6 +21,42 @@
...
+     ' 0.666666667*r3*a[t, x + 1, y] - 0.0833333333*r3*a[t, x + 2, y] + 1)',
+     'Eq(r1, -r2*a[t, x, y])',
+     'Eq(r0, r2*a[t + 1, x, y])',
+     'Eq(b[t + 1, x, y], r0 + r1 + 1)',
+     'Eq(c[t + 1, x, y], r0 + r1 + 2)']])
+def test_topological_sorting(expected):
+    """
+    Test that verifies that:
+    - Temporaries are sorted according to dependencies
+    - Temporaries are not moved before and equation that defines its rhs
```
georgebisbas: Maybe rephrase that? Does not seem that clear at least for me.

FabioLuporini:
I didn't forget about this, just have some patience and I'll get to it

tests/test_dse.py
```diff
@@ -21,6 +22,42 @@
...
+     ' 0.0833333333*r3*a[t, x - 2, y] - 0.666666667*r3*a[t, x - 1, y] +'
+     ' 0.666666667*r3*a[t, x + 1, y] - 0.0833333333*r3*a[t, x + 2, y] + 1)',
+     'Eq(r1, -r2*a[t, x, y])',
+     'Eq(r0, r2*a[t + 1, x, y])',
+     'Eq(b[t + 1, x, y], r0 + r1 + 1)',
+     'Eq(c[t + 1, x, y], r0 + r1 + 2)']])
+def test_topological_sorting(expected):
+    """
+    Test that verifies that:
+    - Temporaries are sorted according to dependencies
```
rhodrin: Nitpicking: '... according to their ...' is a touch cleaner.

tests/test_dse.py
rhodrin: I believe it's just a typo: `and` -> `an` then it makes sense :)

FabioLuporini:
Merged.

PR #1023, opened 2019-12-04
ops: adding RawAccess/RawAccessible

btrb:
This PR is an updated version of #1002 .

From #991:
When translating to OPS a division by pointer (like `/*dt`), it would make the rest of the code to be a commentary according to C++ syntax. 
This was solved by encasing the pointer by parenthesis (like `/(*dt)`).

In this PR, requested changes are made upon updated master.

### DIFF
--- a/devito/ops/node_factory.py
+++ b/devito/ops/node_factory.py
@@ -1,9 +1,9 @@
 from collections import defaultdict, OrderedDict
 
-from devito import Constant, TimeFunction
+from devito import TimeFunction
 from devito.types.dimension import SpaceDimension
 from devito.symbolics import split_affine
-from devito.ops.types import OpsAccess, OpsAccessible
+from devito.ops.types import OpsAccess, OpsAccessible, RawAccess, RawAccessible
 from devito.ops.utils import AccessibleInfo
 
 
@@ -69,15 +69,10 @@ class OPSNodeFactory(object):
 
         return OpsAccess(symbol_to_access, space_indices)
 
-    def new_ops_gbl(self, c):
-        if c in self.ops_args:
-            return self.ops_args[c].accessible
+    def new_ops_gbl(self, expr):
+        if expr not in self.ops_args:
+            param = RawAccessible(expr.name, expr.dtype)
+            self.ops_args[expr] = AccessibleInfo(param, None, None, None)
+            self.ops_params.append(param)
 
-        new_c = AccessibleInfo(Constant(name='*%s' % c.name, dtype=c.dtype),
-                               None,
-                               None,
-                               None)
-        self.ops_args[c] = new_c
-        self.ops_params.append(new_c.accessible)
-
-        return new_c.accessible
+        return RawAccess(expr)
--- a/devito/ops/node_factory.py
+++ b/devito/ops/node_factory.py
@@ -71,7 +71,7 @@ class OPSNodeFactory(object):
 
     def new_ops_gbl(self, expr):
         if expr not in self.ops_args:
-            param = RawAccessible(expr.name, expr.dtype)
+            param = RawAccessible(name=expr.name, dtype=expr.dtype)
             self.ops_args[expr] = AccessibleInfo(param, None, None, None)
             self.ops_params.append(param)
 
--- a/devito/ops/transformer.py
+++ b/devito/ops/transformer.py
@@ -271,7 +271,7 @@ def create_ops_arg(p, accessible_origin, name_to_ops_dat, par_to_ops_stencil):
     elements_per_point = 1
     dtype = Literal('"%s"' % dtype_to_cstr(p.dtype))
 
-    if p.is_Constant:
+    if p.is_Scalar:
         ops_type = namespace['ops_arg_gbl']
         ops_name = Byref(Constant(name=p.name[1:]))
         rw_flag = namespace['ops_read']
--- a/devito/ops/transformer.py
+++ b/devito/ops/transformer.py
@@ -37,7 +37,7 @@ def opsit(trees, count, name_to_ops_dat, block, dims):
                        for expr in expressions]
 
     parameters = sorted(node_factory.ops_params,
-                        key=lambda i: (i.is_Constant, i.name))
+                        key=lambda i: (i.is_Scalar, i.name))
 
     stencil_arrays_initializations = []
     par_to_ops_stencil = {}
--- a/devito/ops/types.py
+++ b/devito/ops/types.py
@@ -51,12 +51,6 @@ class RawAccess(basic.Basic, sympy.Basic):
         # other is not NaN, 0, or 1
         return None
 
-    def as_coeff_Mul(self):
-        return sympy.S.One, self
-
-    def as_coeff_Add(self):
-        return sympy.S.Zero, self
-
 
 class OpsAccessible(basic.Symbol):
     """
--- a/devito/ops/types.py
+++ b/devito/ops/types.py
@@ -3,8 +3,10 @@ import sympy
 
 import devito.types.basic as basic
 
-from devito.tools import dtype_to_cstr
+from devito.tools import dtype_to_cstr, ctypes_to_cstr
 from devito.ops.utils import namespace
+
+from ctypes import c_float
 
 
 class Array(basic.Array):
@@ -15,6 +17,34 @@ class Array(basic.Array):
             return self.dtype
 
         return super()._C_typedata
+
+
+class RawAccessible(basic.Symbol):
+    is_Scalar = True
+
+    def __init_finalize__(self, name, read_only=True, **kwargs):
+        self.read_only = read_only
+        super().__init_finalize__(name, **kwargs)
+
+    @property
+    def _C_typename(self):
+        # TODO: can't change self.dtype from __init_finalize__
+        return '%s%s *' % (
+            'const ' if self.read_only else '',
+            ctypes_to_cstr(c_float)
+        )
+
+
+class RawAccess(basic.Basic, sympy.Basic):
+    def __init__(self, base, *args, **kwargs):
+        self.base = base
+        super().__init__(*args, **kwargs)
+
+    def __str__(self):
+        return '(*%s)' % self.base._C_name
+
+    def __repr__(self):
+        return self.base._C_name
 
 
 class OpsAccessible(basic.Symbol):
@@ -30,7 +60,7 @@ class OpsAccessible(basic.Symbol):
         to ``np.float32``.
     """
 
-    is_Scalar = True
+    is_Scalar = False
 
     def __init_finalize__(self, name, read_only=False, **kwargs):
         self.read_only = read_only
--- a/devito/ops/types.py
+++ b/devito/ops/types.py
@@ -46,6 +46,17 @@ class RawAccess(basic.Basic, sympy.Basic):
     def __repr__(self):
         return self.base._C_name
 
+    def _eval_power(self, other):
+        # subclass to compute self**other for cases when
+        # other is not NaN, 0, or 1
+        return None
+
+    def as_coeff_Mul(self):
+        return sympy.S.One, self
+
+    def as_coeff_Add(self):
+        return sympy.S.Zero, self
+
 
 class OpsAccessible(basic.Symbol):
     """
@@ -146,6 +157,9 @@ class OpsAccess(basic.Basic, sympy.Basic):
         return (self,)
 
     __repr__ = __str__
+
+    def as_base_exp(self):
+        return self.args, sympy.S.One
 
 
 class OpsBlock(basic.Symbol):
--- a/tests/test_ops.py
+++ b/tests/test_ops.py
@@ -30,13 +30,13 @@ class TestOPSExpression(object):
         ('Eq(u,3*a - 4**a)', 'void OPS_Kernel_0(ACC<float> & ut00)\n'
          '{\n  ut00(0) = -2.97015324253729F;\n}'),
         ('Eq(u, u.dxl)',
-         'void OPS_Kernel_0(const float * h_x, ACC<float> & ut00)\n'
+         'void OPS_Kernel_0(ACC<float> & ut00, const float * h_x)\n'
          '{\n  float r0 = 1.0/(*h_x);\n  '
          'ut00(0) = (-2.0F*ut00(-1) + 5.0e-1F*ut00(-2) + 1.5F*ut00(0))*r0;\n}'),
         ('Eq(v,1)', 'void OPS_Kernel_0(ACC<float> & vt00)\n'
          '{\n  vt00(0, 0) = 1;\n}'),
         ('Eq(v,v.dxl + v.dxr - v.dyr - v.dyl)',
-         'void OPS_Kernel_0(const float * h_x, const float * h_y, ACC<float> & vt00)\n'
+         'void OPS_Kernel_0(ACC<float> & vt00, const float * h_x, const float * h_y)\n'
          '{\n  float r1 = 1.0/(*h_y);\n  float r0 = 1.0/(*h_x);\n  '
          'vt00(0, 0) = (5.0e-1F*(-vt00(2, 0) + vt00(-2, 0)) + 2.0F*(-vt00(-1, 0) + '
          'vt00(1, 0)))*r0 + (5.0e-1F*(-vt00(0, -2) + vt00(0, 2)) + '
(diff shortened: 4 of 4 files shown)


btrb:
@FabioLuporini i'm working on fixing the failing test

```
>               obj = b._eval_power(e)
E               AttributeError: 'RawAccess' object has no attribute '_eval_power'
```

but in the meanwhile i'd like your input on weather the proposed solution looks acceptable

devito/ops/node_factory.py
```diff
@@ -69,15 +69,10 @@ def new_ops_arg(self, indexed, is_write):
...
 
-        new_c = AccessibleInfo(Constant(name='*%s' % c.name, dtype=c.dtype),
-                               None,
-                               None,
-                               None)
-        self.ops_args[c] = new_c
-        self.ops_params.append(new_c.accessible)
-
-        return new_c.accessible
+        return RawAccess(expr)
```
mloubout: So basically:
- change `self` state
- return another object that does not depend on it?

That feels odd

devito/ops/transformer.py
```diff
@@ -271,7 +271,7 @@ def create_ops_arg(p, accessible_origin, name_to_ops_dat, par_to_ops_stencil):
     elements_per_point = 1
     dtype = Literal('"%s"' % dtype_to_cstr(p.dtype))
 
-    if p.is_Constant:
+    if p.is_Scalar:
```
mloubout: why?

devito/ops/types.py
```diff
@@ -17,6 +19,45 @@ def _C_typedata(self):
...
+    def _eval_power(self, other):
+        # subclass to compute self**other for cases when
+        # other is not NaN, 0, or 1
+        return None
+
+    def as_coeff_Mul(self):
+        return sympy.S.One, self
+
+    def as_coeff_Add(self):
+        return sympy.S.Zero, self
```
mloubout: *I really want these ones gone.*
`Sympy`  operations should not be backend specific, it is what `Differentiable` and `basic` is for, this makes a lot of things not right and quite messy to fix.

devito/ops/types.py
```diff
@@ -30,7 +71,7 @@ class OpsAccessible(basic.Symbol):
         to ``np.float32``.
     """
 
-    is_Scalar = True
+    is_Scalar = False
```
mloubout: why?

devito/ops/types.py
```diff
@@ -117,6 +158,9 @@ def as_ordered_factors(self):
 
     __repr__ = __str__
 
+    def as_base_exp(self):
+        return self.args, sympy.S.One
```
mloubout: No, this can't be there.

devito/ops/node_factory.py
btrb: i see what you mean.. 

this is one of the two base cases for the recursive method `make_ops_ast` (https://github.com/opesci/devito/blob/master/devito/ops/transformer.py#L296)

this case (`new_ops_gbl`) follows the same pattern from the other case (`new_ops_arg` - https://github.com/opesci/devito/blob/master/devito/ops/node_factory.py#L23)

devito/ops/transformer.py
btrb: an `ops_par_loop` method receives either `ops_arg_dat` (arrays) or `ops_arg_gbl` (numbers)
these types are represented by either a `RawAccessible` or a `OpsAccessible`.
both inherit from basic.Symbol. thus it seems it illogical that those should be marked as a `Constant` (https://github.com/opesci/devito/blob/master/devito/types/basic.py#L168)

devito/ops/types.py
btrb: this flag serves only for differentiating between `ops_arg_gbl` and `ops_arg_dat` in the method `create_ops_arg`
https://github.com/opesci/devito/blob/master/devito/ops/transformer.py#L274

moreover, an `OpsAccessible` represents a small array

devito/ops/types.py
btrb: ok, removing! used that for a fix, but it seems it's not quite required indeed

devito/ops/types.py
btrb: if i remove it, this is what i get with `DEVITO_BACKEND=ops pytest --maxfail=1 tests/test_ops.py`:

```
============================= test session starts ==============================
platform linux -- Python 3.6.9, pytest-5.0.1, py-1.5.3, pluggy-0.12.0
rootdir: /home/lucas/devito, inifile: setup.cfg
plugins: cov-2.5.1, parallel-0.0.9, nbval-0.9.0
collected 30 items

tests/test_ops.py ..........F

=================================== FAILURES ===================================
 TestOPSExpression.test_accesses_extraction[Eq(u, u.dxl)-{ "ut0": [[0], [-1], [-2]] }] 

self = <test_ops.TestOPSExpression object at 0x7fd49b1d86d8>
equation = 'Eq(u, u.dxl)', expected = '{ "ut0": [[0], [-1], [-2]] }'

    @pytest.mark.parametrize('equation, expected', [
        ('Eq(u,3*a - 4**a)', '{ "ut0": [[0]] }'),
        ('Eq(u, u.dxl)', '{ "ut0": [[0], [-1], [-2]] }'),
        ('Eq(u,v+1)', '{ "ut0": [[0]], "vt0": [[0]] }')
    ])
    def test_accesses_extraction(self, equation, expected):
        grid_1d = Grid(shape=(4))
        grid_3d = Grid(shape=(4, 4, 4))
    
        a = 1.43  # noqa
        c = 999999999999999  # noqa
        u = TimeFunction(name='u', grid=grid_1d, space_order=2)  # noqa
        v = TimeFunction(name='v', grid=grid_1d, space_order=2)  # noqa
        w = TimeFunction(name='w', grid=grid_3d, space_order=2)  # noqa
    
        node_factory = OPSNodeFactory()
    
>       make_ops_ast(indexify(eval(equation).evaluate), node_factory)

tests/test_ops.py:111: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
devito/ops/transformer.py:325: in make_ops_ast
    make_ops_ast(expr.rhs, nfops))
devito/ops/transformer.py:327: in make_ops_ast
    return expr.func(*[make_ops_ast(i, nfops) for i in expr.args])
devito/ops/transformer.py:327: in <listcomp>
    return expr.func(*[make_ops_ast(i, nfops) for i in expr.args])
devito/ops/transformer.py:327: in make_ops_ast
    return expr.func(*[make_ops_ast(i, nfops) for i in expr.args])
devito/finite_differences/differentiable.py:257: in __new__
    obj = cls.__base__.__new__(cls, *args, **kwargs)
../.local/lib/python3.6/site-packages/sympy/core/cache.py:94: in wrapper
    retval = cfunc(*args, **kwargs)
../.local/lib/python3.6/site-packages/sympy/core/operations.py:47: in __new__
    c_part, nc_part, order_symbols = cls.flatten(args)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

cls = <class 'devito.finite_differences.differentiable.Mul'>
seq = [0.500000000, RawAccess(h_x)**(-1), ut0(-2)]

    @classmethod
    def flatten(cls, seq):
        """Return commutative, noncommutative and order arguments by
        combining related terms.
    
        Notes
        =====
            * In an expression like ``a*b*c``, python process this through sympy
              as ``Mul(Mul(a, b), c)``. This can have undesirable consequences.
    
              -  Sometimes terms are not combined as one would like:
                 {c.f. https://github.com/sympy/sympy/issues/4596}
    
                >>> from sympy import Mul, sqrt
                >>> from sympy.abc import x, y, z
                >>> 2*(x + 1) # this is the 2-arg Mul behavior
                2*x + 2
                >>> y*(x + 1)*2
                2*y*(x + 1)
                >>> 2*(x + 1)*y # 2-arg result will be obtained first
                y*(2*x + 2)
                >>> Mul(2, x + 1, y) # all 3 args simultaneously processed
                2*y*(x + 1)
                >>> 2*((x + 1)*y) # parentheses can control this behavior
                2*y*(x + 1)
    
                Powers with compound bases may not find a single base to
                combine with unless all arguments are processed at once.
                Post-processing may be necessary in such cases.
                {c.f. https://github.com/sympy/sympy/issues/5728}
    
                >>> a = sqrt(x*sqrt(y))
                >>> a**3
                (x*sqrt(y))**(3/2)
                >>> Mul(a,a,a)
                (x*sqrt(y))**(3/2)
                >>> a*a*a
                x*sqrt(y)*sqrt(x*sqrt(y))
                >>> _.subs(a.base, z).subs(z, a.base)
                (x*sqrt(y))**(3/2)
    
              -  If more than two terms are being multiplied then all the
                 previous terms will be re-processed for each new argument.
                 So if each of ``a``, ``b`` and ``c`` were :class:`Mul`
                 expression, then ``a*b*c`` (or building up the product
                 with ``*=``) will process all the arguments of ``a`` and
                 ``b`` twice: once when ``a*b`` is computed and again when
                 ``c`` is multiplied.
    
                 Using ``Mul(a, b, c)`` will process all arguments once.
    
            * The results of Mul are cached according to arguments, so flatten
              will only be called once for ``Mul(a, b, c)``. If you can
              structure a calculation so the arguments are most likely to be
              repeats then this can save time in computing the answer. For
              example, say you had a Mul, M, that you wished to divide by ``d[i]``
              and multiply by ``n[i]`` and you suspect there are many repeats
              in ``n``. It would be better to compute ``M*n[i]/d[i]`` rather
              than ``M/d[i]*n[i]`` since every time n[i] is a repeat, the
              product, ``M*n[i]`` will be returned without flattening -- the
              cached value will be returned. If you divide by the ``d[i]``
              first (and those are more unique than the ``n[i]``) then that will
              create a new Mul, ``M/d[i]`` the args of which will be traversed
              again when it is multiplied by ``n[i]``.
    
              {c.f. https://github.com/sympy/sympy/issues/5706}
    
              This consideration is moot if the cache is turned off.
    
            NB
            --
              The validity of the above notes depends on the implementation
              details of Mul and flatten which may change at any time. Therefore,
              you should only consider them when your code is highly performance
              sensitive.
    
              Removal of 1 from the sequence is already handled by AssocOp.__new__.
        """
    
        from sympy.calculus.util import AccumBounds
        from sympy.matrices.expressions import MatrixExpr
        rv = None
        if len(seq) == 2:
            a, b = seq
            if b.is_Rational:
                a, b = b, a
                seq = [a, b]
            assert not a is S.One
            if not a.is_zero and a.is_Rational:
                r, b = b.as_coeff_Mul()
                if b.is_Add:
                    if r is not S.One:  # 2-arg hack
                        # leave the Mul as a Mul
                        rv = [cls(a*r, b, evaluate=False)], [], None
                    elif global_distribute[0] and b.is_commutative:
                        r, b = b.as_coeff_Add()
                        bargs = [_keep_coeff(a, bi) for bi in Add.make_args(b)]
                        _addsort(bargs)
                        ar = a*r
                        if ar:
                            bargs.insert(0, ar)
                        bargs = [Add._from_args(bargs)]
                        rv = bargs, [], None
            if rv:
                return rv
    
        # apply associativity, separate commutative part of seq
        c_part = []         # out: commutative factors
        nc_part = []        # out: non-commutative factors
    
        nc_seq = []
    
        coeff = S.One       # standalone term
                            # e.g. 3 * ...
    
        c_powers = []       # (base,exp)      n
                            # e.g. (x,n) for x
    
        num_exp = []        # (num-base, exp)           y
                            # e.g.  (3, y)  for  ... * 3  * ...
    
        neg1e = S.Zero      # exponent on -1 extracted from Number-based Pow and I
    
        pnum_rat = {}       # (num-base, Rat-exp)          1/2
                            # e.g.  (3, 1/2)  for  ... * 3     * ...
    
        order_symbols = None
    
        # --- PART 1 ---
        #
        # "collect powers and coeff":
        #
        # o coeff
        # o c_powers
        # o num_exp
        # o neg1e
        # o pnum_rat
        #
        # NOTE: this is optimized for all-objects-are-commutative case
        for o in seq:
            # O(x)
            if o.is_Order:
                o, order_symbols = o.as_expr_variables(order_symbols)
    
            # Mul([...])
            if o.is_Mul:
                if o.is_commutative:
                    seq.extend(o.args)    # XXX zerocopy?
    
                else:
                    # NCMul can have commutative parts as well
                    for q in o.args:
                        if q.is_commutative:
                            seq.append(q)
                        else:
                            nc_seq.append(q)
    
                    # append non-commutative marker, so we don't forget to
                    # process scheduled non-commutative objects
                    seq.append(NC_Marker)
    
                continue
    
            # 3
            elif o.is_Number:
                if o is S.NaN or coeff is S.ComplexInfinity and o is S.Zero:
                    # we know for sure the result will be nan
                    return [S.NaN], [], None
                elif coeff.is_Number or isinstance(coeff, AccumBounds):  # it could be zoo
                    coeff *= o
                    if coeff is S.NaN:
                        # we know for sure the result will be nan
                        return [S.NaN], [], None
                continue
    
            elif isinstance(o, AccumBounds):
                coeff = o.__mul__(coeff)
                continue
    
            elif o is S.ComplexInfinity:
                if not coeff:
                    # 0 * zoo = NaN
                    return [S.NaN], [], None
                if coeff is S.ComplexInfinity:
                    # zoo * zoo = zoo
                    return [S.ComplexInfinity], [], None
                coeff = S.ComplexInfinity
                continue
    
            elif o is S.ImaginaryUnit:
                neg1e += S.Half
                continue
    
            elif o.is_commutative:
                #      e
                # o = b
                b, e = o.as_base_exp()
    
                #  y
                # 3
                if o.is_Pow:
                    if b.is_Number:
    
                        # get all the factors with numeric base so they can be
                        # combined below, but don't combine negatives unless
                        # the exponent is an integer
                        if e.is_Rational:
                            if e.is_Integer:
                                coeff *= Pow(b, e)  # it is an unevaluated power
                                continue
                            elif e.is_negative:    # also a sign of an unevaluated power
                                seq.append(Pow(b, e))
                                continue
                            elif b.is_negative:
                                neg1e += e
                                b = -b
                            if b is not S.One:
                                pnum_rat.setdefault(b, []).append(e)
                            continue
                        elif b.is_positive or e.is_integer:
                            num_exp.append((b, e))
                            continue
    
                c_powers.append((b, e))
    
            # NON-COMMUTATIVE
            # TODO: Make non-commutative exponents not combine automatically
            else:
                if o is not NC_Marker:
                    nc_seq.append(o)
    
                # process nc_seq (if any)
                while nc_seq:
                    o = nc_seq.pop(0)
                    if not nc_part:
                        nc_part.append(o)
                        continue
    
                    #                             b    c       b+c
                    # try to combine last terms: a  * a   ->  a
                    o1 = nc_part.pop()
                    b1, e1 = o1.as_base_exp()
>                   b2, e2 = o.as_base_exp()
E                   AttributeError: 'OpsAccess' object has no attribute 'as_base_exp'

../.local/lib/python3.6/site-packages/sympy/core/mul.py:339: AttributeError
===================== 1 failed, 10 passed in 1.28 seconds ======================
```

devito/ops/node_factory.py
FabioLuporini: > change self state

@mloubout good spot ! 👏

devito/ops/types.py
```diff
@@ -17,6 +19,39 @@ def _C_typedata(self):
...
+class RawAccessible(basic.Symbol):
+    is_Scalar = True
+
+    def __init_finalize__(self, name, read_only=True, **kwargs):
+        self.read_only = read_only
+        super().__init_finalize__(name, **kwargs)
+
+    @property
+    def _C_typename(self):
+        # TODO: can't change self.dtype from __init_finalize__
```
FabioLuporini: `__dtype_setup__` . Is that what you're looking for?

devito/ops/types.py
```diff
@@ -17,6 +19,39 @@ def _C_typedata(self):
...
+        )
+
+
+class RawAccess(basic.Basic, sympy.Basic):
+    def __init__(self, base, *args, **kwargs):
+        self.base = base
+        super().__init__(*args, **kwargs)
+
+    def __str__(self):
+        return '(*%s)' % self.base._C_name
```
FabioLuporini: I just thought that a nicer way would be to introduce a SymPy subclass along the lines of https://github.com/opesci/devito/blob/master/devito/symbolics/extended_sympy.py#L286

this way you wouldn't need to introduce a new symbolic type. And you would be able to use that new class  to wrap any symbolic object you like, potentially even a whole function call.... `(*a)`, `(**a)` (with a suitable parameter `nindirections`), `(*foo())`, ...

FabioLuporini:
As discussed offline with @btrb , closing for now as nobody will be working on this for some time

PR #1025, opened 2019-12-08
Lazy sources and receivers

navjotk:
- Move interpolation out of sparse functions into its own module.
- Make interpolation behave lazy


devito/interpolators.py
```diff
@@ -0,0 +1,292 @@
...
+
+
+class Interpolation(UnevaluatedSparseOperation):
+    is_Interpolation = True
+
+
+class LinearInterpolator:
+    def __init__(self, grid, obj):
+        self.grid = grid
+        self.obj = obj
```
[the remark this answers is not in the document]
FabioLuporini: and finally, I suspect, but I may be wrong, that `obj` should be given through inject() or interpolate() directly. This way, the interpolators could even be singleton-ized

devito/types/sparse.py
```diff
@@ -474,7 +474,7 @@ class SparseFunction(AbstractSparseFunction):
 
     def __init_finalize__(self, *args, **kwargs):
         super(SparseFunction, self).__init_finalize__(*args, **kwargs)
-
+        self.interpolator = LinearInterpolator(self.grid, self)
```
FabioLuporini: I would add a class attribute for this

```
class SparseFunction(...):

     _default_interpolator = LinearInterpolator

     def __init_finalize__(...):
           ...
           self.interpolator = self._default_interpolator(...)
           # or self.interpolator = kwargs.get('interpolator', self._default_interpolator)()
```

for subclasses and user it will then be possible to specialize what interpolator to use, as long as the interpolator interface is honored (that's why I was suggesting the common abstract class a few comments ago)

devito/types/sparse.py
```diff
@@ -1082,92 +937,17 @@ def __init_finalize__(self, *args, **kwargs):
...
-        # Apply optional time symbol substitutions to lhs of assignment
-        lhs = self.subs(self_subs)
-        rhs = prod(coeffs) * expr.subs(dim_subs)
-
-        return [Eq(lhs, lhs + rhs)]
-
-    def inject(self, field, expr, offset=0):
-        """
-        Generate equations injecting an arbitrary expression into a field.
+        self.interpolator = PrecomputedInterpolator(self, r, gridpoints,
```
FabioLuporini: I think if we do as suggested above, this line becomes unnecessary as executed automatically through `super(...)` while implicitly using the right interpolator type

devito/types/sparse.py
```diff
@@ -1082,92 +937,17 @@ def __init_finalize__(self, *args, **kwargs):
...
-        dim_subs = []
-        coeffs = []
-        for i, d in enumerate(self.grid.dimensions):
-            rd = DefaultDimension(name="r%s" % d.name, default_value=self.r)
-            dim_subs.append((d, INT(rd + self.gridpoints[p, i])))
-            coeffs.append(self.interpolation_coeffs[p, i, rd])
-        rhs = prod(coeffs) * expr
-        field = field.subs(dim_subs)
-        return [Eq(field, field + rhs.subs(dim_subs))]
+    def interpolate(self, *args, **kwargs):
```
FabioLuporini: unnecessary ?

devito/types/sparse.py
```diff
@@ -1082,92 +937,17 @@ def __init_finalize__(self, *args, **kwargs):
...
+        return Interpolation(self.interpolator, args, kwargs)
+
+    def inject(self, *args, **kwargs):
```
FabioLuporini: unnecessary?

devito/types/sparse.py
FabioLuporini: as the same method is in the superclass, I mean

devito/interpolators.py
```diff
@@ -0,0 +1,292 @@
+import sympy
+import numpy as np
+
+from devito.equation import Eq, Inc
+from devito.symbolics import retrieve_function_carriers, indexify, INT
+from devito.tools import Evaluable, powerset, flatten, prod
+from devito.types.dimension import ConditionalDimension, Dimension, DefaultDimension
+from devito.types.basic import Scalar
+from devito.types.dense import SubFunction
+from devito.logger import warning
```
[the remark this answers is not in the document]
navjotk: Fixed

devito/interpolators.py
```diff
@@ -0,0 +1,292 @@
...
+
+
+class UnevaluatedSparseOperation(Evaluable):
+    is_Injection = False
+    is_Interpolation = False
+    subdomain = None
+
+    def __init__(self, interpolator, args, kwargs):
```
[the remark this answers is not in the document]
navjotk: Fixed

devito/interpolators.py
```diff
@@ -0,0 +1,292 @@
...
+            return_value = self.interpolator.interpolate(*self._args, **self.kwargs)
+        elif self.is_Injection:
+            return_value = self.interpolator.inject(*self._args, **self.kwargs)
+        else:
+            raise ValueError
+        assert(all(isinstance(i, Eq) for i in return_value))
+        return return_value
+
+
+class Injection(UnevaluatedSparseOperation):
```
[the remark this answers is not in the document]
navjotk: Fixed

devito/interpolators.py
```diff
@@ -0,0 +1,292 @@
...
+from devito.equation import Eq, Inc
+from devito.symbolics import retrieve_function_carriers, indexify, INT
+from devito.tools import Evaluable, powerset, flatten, prod
+from devito.types.dimension import ConditionalDimension, Dimension, DefaultDimension
+from devito.types.basic import Scalar
+from devito.types.dense import SubFunction
+from devito.logger import warning
+
+
+class UnevaluatedSparseOperation(Evaluable):
```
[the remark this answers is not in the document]
navjotk: Fixed

devito/interpolators.py
```diff
@@ -0,0 +1,292 @@
...
+
+class Injection(UnevaluatedSparseOperation):
+    is_Injection = True
+
+
+class Interpolation(UnevaluatedSparseOperation):
+    is_Interpolation = True
+
+
+class LinearInterpolator:
```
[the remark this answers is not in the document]
navjotk: Fixed

devito/interpolators.py
```diff
@@ -0,0 +1,292 @@
...
+
+        # Substitute coordinate base symbols into the interpolation coefficients
+        eqns = [Inc(field.xreplace(vsub), expr.xreplace(vsub) * b,
+                    implicit_dims=self.obj.dimensions)
+                for b, vsub in zip(self._interpolation_coeffs, idx_subs)]
+
+        return temps + eqns
+
+
+class PrecomputedInterpolator:
```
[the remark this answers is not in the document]
navjotk: Fixed

devito/interpolators.py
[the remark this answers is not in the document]
navjotk: Changed for the first two. 

Last one - things started breaking horribly the two times I attempted to make the Interpolator object carry no state. Gave up on that for now.

devito/types/sparse.py
navjotk: Can be kept as a TODO because the constructors for LinearInterpolator and PrecomputedInterpolator take different arguments for now. Not easy to homogenise yet.

devito/types/sparse.py
navjotk: As discussed above

devito/types/sparse.py
navjotk: Fixed

devito/types/sparse.py
navjotk: Fixed

devito/operator.py
```diff
@@ -131,10 +129,6 @@ class Operator(Callable):
     def __init__(self, expressions, **kwargs):
         expressions = as_tuple(expressions)
 
-        # Input check
-        if any(not isinstance(i, Eq) for i in expressions):
```
FabioLuporini: we don't want to drop this

devito/types/sparse.py
FabioLuporini: I think github isn't showing me this "as discussed above"? could you paste it or write it again please?

devito/types/sparse.py
```diff
@@ -94,6 +94,18 @@ def gridpoints(self):
         """
         raise NotImplementedError
 
+    def interpolate(self, *args, **kwargs):
+        """
+        Implement an interpolation operation from the grid onto the given sparse points
+        """
+        return Interpolation(self.interpolator, *args, **kwargs)
```
FabioLuporini: would the API be neater as `self.interpolator.interpolate(*args, **kwargs)` ? this way, here you don't need to know about `Interpolation` and `Injection`

devito/operator.py
FabioLuporini: perhaps replace `Eq` with `Evaluable` ?

FabioLuporini: and if you think this were a good idea, probably also time to move `Evaluable` from [here](https://github.com/devitocodes/devito/blob/master/devito/tools/abc.py#L145) to `devito/types/evaluable.py` or some other place (most likely still somewhere inside `types.py`

note that after landing #1026 , we now have `Eq` defined in `types.py` too, so it sounds like a nice fit

devito/interpolators.py
```diff
@@ -0,0 +1,354 @@
...
+        Interpolator object that will be used to evaluate the Operation
+    *args, **kwargs
+        The arguments passed to the corresponding method
+    """
+    subdomain = None
+
+    def __init__(self, interpolator, *args, **kwargs):
+        self.interpolator = interpolator
+        self._args = args
+        self._kwargs = kwargs
```
mloubout: why do you need to store args and kwargs?

devito/interpolators.py
```diff
@@ -0,0 +1,354 @@
...
+
+    @property
+    def evaluate(self):
+        return_value = self.interpolator.interpolate(self.expr, *self._args,
+                                                     **self._kwargs)
+        assert(all(isinstance(i, Eq) for i in return_value))
+        return return_value
+
+    def __repr__(self):
+        return "Interpolation(%s)" % repr(self.expr)
```
mloubout: So Interpolation doesn't know what it's interpolated into? (rec)

devito/interpolators.py
```diff
@@ -0,0 +1,354 @@
...
+        try:
+            expr = expr.evaluate
+        except AttributeError:
+            # E.g., a generic SymPy expression or a number
+            pass
+
+        variables = list(retrieve_function_carriers(expr))
+
+        # Need to get origin of the field in case it is staggered
+        # TODO: handle each variable staggereing spearately
```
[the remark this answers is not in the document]
mloubout: Probably yes.

devito/interpolators.py
```diff
@@ -0,0 +1,354 @@
...
+
+        # Substitute coordinate base symbols into the interpolation coefficients
+        eqns = [Inc(field.xreplace(vsub), expr.xreplace(vsub) * b,
+                    implicit_dims=self.sfunction.dimensions)
+                for b, vsub in zip(self._interpolation_coeffs, idx_subs)]
+
+        return temps + eqns
+
+
+class PrecomputedInterpolator(GenericInterpolator):
```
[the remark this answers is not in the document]
mloubout: Is this precomputed interpolator still in use anywhere?

devito/types/sparse.py
navjotk: I am referencing this comment:
> Can be kept as a TODO because the constructors for LinearInterpolator and PrecomputedInterpolator take different arguments for now. Not easy to homogenise yet.

devito/interpolators.py
```diff
@@ -0,0 +1,354 @@
...
+from devito.types.dense import SubFunction
+from devito.types.dimension import ConditionalDimension, Dimension, DefaultDimension
+
+
+class UnevaluatedSparseOperation(Evaluable):
+
+    """
+    Represents an Injection or an Interpolation operation performed on a
+    SparseFunction. Evaluates to a list of Eq objects.
+    objects.
```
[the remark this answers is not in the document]
navjotk: Fixed

devito/interpolators.py
```diff
@@ -0,0 +1,354 @@
...
+
+    Parameters
+    ----------
+    interpolator : Interpolator
+        Interpolator object that will be used to evaluate the Operation
```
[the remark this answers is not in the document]
navjotk: Fixed

devito/interpolators.py
```diff
@@ -0,0 +1,354 @@
...
+    *args, **kwargs
+        The arguments passed to the corresponding method
```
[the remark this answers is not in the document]
navjotk: Fixed

devito/interpolators.py
```diff
@@ -0,0 +1,354 @@
...
+    """
+    subdomain = None
+
+    def __init__(self, interpolator, *args, **kwargs):
+        self.interpolator = interpolator
+        self._args = args
+        self._kwargs = kwargs
+
+    def __add__(self, other):
```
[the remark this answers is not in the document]
navjotk: In the previous API, `SparseFunction.inject()` (or `interpolate`) would return a list of `devito.Eq` objects:
`src_terms = src.inject(...)`
The stencil would be created as:
`stencil = Eq(...)`

and then they would be passed to the `Operator` as:
`op = Operator(src_terms + [stencil] + rec_terms)`

This stuff is to maintain compatibility with that API.

devito/interpolators.py
```diff
@@ -0,0 +1,354 @@
...
+class Injection(UnevaluatedSparseOperation):
+
+    """
+    Represents an Injection operation performed on a SparseFunction.
+    Evaluates to a list of Eq objects.
+    """
+
+    def __init__(self, interpolator, field, expr, *args, **kwargs):
+        self.interpolator = interpolator
+        self.field = field
```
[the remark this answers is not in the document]
navjotk: Then it becomes ambiguous what function is being referred to. Elsewhere we call the sparse function "function". Calling the function being injected into "function" will only create con"function".

devito/interpolators.py
```diff
@@ -0,0 +1,354 @@
...
+    @abstractmethod
+    def inject(self, *args, **kwargs):
+        pass
+
+    @abstractmethod
+    def interpolate(self, *args, **kwargs):
+        pass
+
+
+class LinearInterpolator(GenericInterpolator):
```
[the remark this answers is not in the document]
navjotk: If this is of importance, this should be in the flake8 rules. Fixed anyway.

devito/interpolators.py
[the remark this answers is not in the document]
navjotk: Resolving because unrelated to this PR.

devito/types/sparse.py
navjotk: It was like what you're suggesting until Mathias made me change it, in order to get better __repr__ for the objects. 

In any case, does this have to happen in this PR?

devito/interpolators.py
```diff
@@ -0,0 +1,354 @@
...
+
+    @property
+    def evaluate(self):
+        return_value = self.interpolator.inject(self.field, self.expr, *self._args,
+                                                **self._kwargs)
+        assert(all(isinstance(i, Eq) for i in return_value))
+        return return_value
+
+    def __repr__(self):
+        return "Injection(%s on %s)" % (repr(self.expr), repr(self.field))
```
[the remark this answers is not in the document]
navjotk: On is actually better

devito/interpolators.py
navjotk: If anything was passed to the inject method on the sparse function, it needs to be passed to the interpolator. So it has to be stored.

devito/interpolators.py
navjotk: Right now, directly, it doesn't.

devito/interpolators.py
```diff
@@ -0,0 +1,354 @@
...
+
+
+class Interpolation(UnevaluatedSparseOperation):
+
+    """
+    Represents an Injection operation performed on a SparseFunction.
```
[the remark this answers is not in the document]
navjotk: Fixed

devito/interpolators.py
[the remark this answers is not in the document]
mloubout: No, it's inject in or inject into, not on.

devito/interpolators.py
mloubout: If there is explicit arguments/keyword needed should be explicit, if not this is not needed, specially without doctoring saying why should be in args/kwargs.

devito/types/sparse.py
mloubout: Don;t get how it's related to __repr__.

devito/interpolators.py
navjotk: This bit of code doesn't need to know what those arguments are. So why should it be explicit?This is just forwarding the arguments to where they are actually used. Aren't you making the opposite point of what @FabioLuporini is making here : https://github.com/devitocodes/devito/pull/1025#discussion_r363207849 ?

devito/interpolators.py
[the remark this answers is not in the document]
navjotk: Inject onto is better than inject on, sure. But not in.

devito/types/sparse.py
navjotk: The only reason the classes had to become specific for interpolation and injection was to be able to have the repr specific to interpolation/injection.

devito/interpolators.py
```diff
@@ -0,0 +1,354 @@
...
+    def interpolate(self, *args, **kwargs):
+        pass
+
+
+class LinearInterpolator(GenericInterpolator):
+    def __init__(self, sfunction):
+        self.grid = sfunction.grid
+        self.sfunction = sfunction
+
+    @property
```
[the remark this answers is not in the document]
navjotk: Fixed

devito/interpolators.py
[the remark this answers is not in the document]
navjotk: Not all user code is on this repository or maintained by us so it's not a straightforward question to answer.

devito/operator.py
navjotk: Done the first part - having a check there makes sense. 

I will leave the refactoring about moving `Evaluable` to a later PR.

devito/interpolators.py
mloubout: No I don't think it's opposite or related to his comment. I am talking about readability and maintainability, there is currently no way to know what `init` and therefor Interp/Inject expect as input, only some anonymous args/kwargs, so it need  to either be explicit or add a complete doctoring that says what it expects as an input.

devito/types/sparse.py
mloubout: That's not incompatible, `self.interpolato.interpolate` returns and `Interpolation` object and `.inject` returns and `Injection` object, that's not against separating both.

devito/interpolators.py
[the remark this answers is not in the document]
georgebisbas: Strange @navjotk ? I d say in/into as well.

devito/operator/operator.py
```diff
@@ -20,6 +21,7 @@
 from devito.tools import (DAG, Signer, ReducerMap, as_tuple, flatten, filter_ordered,
                           filter_sorted, split, timed_pass, timed_region)
 from devito.types import Dimension, Eq
+from devito.tools import Evaluable
```
georgebisbas: Can be merged with line 22. There is also a redundant blank line in 9.

devito/interpolators.py
navjotk: I ask you to look at the complete change I made in the previous commit in entirety. This object is created in `Interpolator.inject` and serves to store state to call `Interpolator._inject`. The docstring is present where this object is created and is right next to the code that consumes those arguments. I can add it here too if you want, but I think it is unnecessary. 
I would like to move as much of the logic to the interpolator class as possible so that these classes can get much closer to the `Future` class that they conceptually are. Any ideas for reducing code in the UnevaluatedSparseOperation/Injection/Interpolation classes are welcome.

devito/interpolators.py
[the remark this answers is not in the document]
navjotk: Fixed

devito/interpolators.py
navjotk: Fixed, but perhaps not in the way you wanted. Mark as resolved if you don't mind my fix @mloubout .

devito/interpolators.py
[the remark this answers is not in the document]
navjotk: See #1046

devito/types/sparse.py
navjotk: See #1046

devito/operator.py
navjotk: See #1046

devito/types/sparse.py
navjotk: I believe I have fixed this. Resolve if you confirm.

devito/interpolators.py
mloubout: `args` and `kwargs` are not used anywhere within the class and just dumping everything here in case some subclass need it is not good even for Future

devito/interpolators.py
```diff
@@ -0,0 +1,410 @@
...
+class Injection(UnevaluatedSparseOperation):
+
+    """
+    Represents an Injection operation performed on a SparseFunction.
+    Evaluates to a list of Eq objects.
+    """
+
+    def __init__(self, interpolator, field, expr, *args, **kwargs):
+        self.field = field
+        self.expr = expr
```
mloubout: Not used anywhere, why is it saved here.

devito/interpolators.py
```diff
@@ -0,0 +1,410 @@
...
+        args = tuple([field, expr] + list(args))
```
mloubout: inject doesn't take any extra `args` so why is some extra added here.

devito/interpolators.py
```diff
@@ -0,0 +1,410 @@
...
+        super().__init__(interpolator, *args, **kwargs)
+
+    @property
+    def evaluate(self):
+        return_value = self.interpolator._inject(*self._args, **self._kwargs)
```
mloubout: if keep in `self` what's needed can be called explicitly with `_inject(expr, ...)` not it's go figure out somewhere how it's supposed to be called for future people.

devito/interpolators.py
```diff
@@ -0,0 +1,410 @@
...
+
+    """
+    Represents an Interpolation operation performed on a SparseFunction.
+    Evaluates to a list of Eq objects.
+    """
+
+    def __init__(self, interpolator, expr, *args, **kwargs):
+        self.expr = expr
+        args = tuple([expr] + list(args))
+        super().__init__(interpolator, *args, **kwargs)
```
mloubout: same as Injection

devito/interpolators.py
```diff
@@ -0,0 +1,410 @@
...
+        Interpolator object that will be used to evaluate the Operation.
+    *args, **kwargs
+        The arguments passed to the corresponding method.
+    """
+    subdomain = None
+
+    def __init__(self, interpolator, *args, **kwargs):
+        assert(isinstance(interpolator, GenericInterpolator))
+
+        self.interpolator = interpolator
```
mloubout: This is cyclic, ` self.interpolator` is a`Injection/Interpolation` that is an `UnevaluatedSparseOperation`, that doesn't feel right

devito/interpolators.py
mloubout: >The docstring is present where this object is created a
That's not ok. The doctoring for a class goes with the class, can't expect people to dig to figure out where someone creates an instance of it to figure out  what the input is.

devito/interpolators.py
navjotk: Can you write out a code example of what you are saying here?

devito/interpolators.py
navjotk: To reduce the coupling between this class and the one where inject is defined. If someone updates the definition of inject to add more arguments, they shouldn't need to care about this.

devito/interpolators.py
navjotk: It is used in the repr. This is part of the code I was forced to add because we wanted pretty printing of these classes.

devito/interpolators.py
navjotk: > args and kwargs are not used anywhere within the class and just dumping everything here in case some subclass need it is not good even for Future
At first look, this is a valid concern. However, I've been trying to figure out a way to move `evaluate` to this class - which would make the subclasses much simpler. The question is of how the subclass will modify the behaviour between inject and interpolate. I don't want to write the name of the method as a string in the subclass and I can't think of another way to do it. I'm still hoping to find a way to do it but I'll write this as another TODO in the issue. 

For the medium to long term future, I envision that this entire series of classes could be abstracted out into decorator-based setups - a decorator feels like the right abstraction for this situation, but it would be too much work to go into for this PR. 

> That's not ok. The doctoring for a class goes with the class, can't expect people to dig to figure out where someone creates an instance of it to figure out what the input is.

In my opinion, these classes are not intended for direct use by the end user. The inject/interpolate functions on the Interpolator are. Either both functions should have the same doctstring, or the user-facing functions should have a doctsring and the internal classes could reference that. Do you want me to add a docstring here that says "See LinearInterpolator.inject"?

devito/interpolators.py
navjotk: Also see https://github.com/devitocodes/devito/pull/1025/files/fd3bee45a5e343ac649a6e669b3c4464f430db08#diff-2f2cddfeb7dcd0e1a6ee101e3cfa2948R385

devito/interpolators.py
navjotk: This method is only providing one bit of functionality - it specifies that the `_inject` method needs to be called. Even right now, this method is called through two different signatures by `LinearInterpolator` and `PrecomputedInterpolator`. Why make it more complex than it needs to be by duplicating the signature of the inject method here?

devito/interpolators.py
navjotk: `self.interpolator` is of type `GenericInterpolator`. I specifically assert that in the line above.

devito/interpolators.py
mloubout: >In my opinion, these classes are not intended for direct use by the end user. 

That's not the point, most of the API is documented wether it's internal or user facing so that people can contribute and maintain it properly and not have to dig everywhere and goes what something is doing. This is a class with a constructor, it needs a doctoring, and I don't like pointing to "just look at how we call it there that works but can maybe do otherwise".

FabioLuporini:
despite the lengthy discussion, I think it won't take much time to deal with the issues discussed above . I will have a go next week

devito/interpolators.py
```diff
@@ -0,0 +1,410 @@
...
+            rd = DefaultDimension(name="r%s" % d.name, default_value=self.r)
+            dim_subs.append((d, INT(rd + self.obj.gridpoints[p, i])))
+            coeffs.append(self.obj.interpolation_coeffs[p, i, rd])
+        # Apply optional time symbol substitutions to lhs of assignment
+        lhs = self.obj.subs(self_subs)
+        rhs = prod(coeffs) * expr.subs(dim_subs)
+
+        return [Eq(lhs, lhs + rhs)]
+
+    def inject(self, field, expr, offset=0):
```
FabioLuporini: missing _inject?

devito/interpolators.py
```diff
@@ -0,0 +1,410 @@
...
+        super().__init__(interpolator, *args, **kwargs)
+
+    @property
+    def evaluate(self):
+        return_value = self.interpolator._interpolate(*self._args, **self._kwargs)
+        assert(all(isinstance(i, Eq) for i in return_value))
+        return return_value
+
+    def __repr__(self):
+        return "Interpolation(%s into %s)" % repr(self.expr,
```
FabioLuporini: looks like another bug

rhodrin:
Are there going to be any new tests associated with this PR or are the current tests sufficient?

ggorman:
Agreed actions - @navjotk to:
- pull changes from @FabioLuporini 's branch that resolves outstanding review comments (agreed online/slack).
- rebase on master and merge PR once green.

PR #1026, opened 2019-12-11
Introducing Targets

FabioLuporini:
You see a huge number of edits, but in practice this PR is "just" about refactoring, as almost no new functionalities are added (see below for more details)

The reason you see lots of files changed is that I had to shuffle things around and occasionally split things over new modules. The objective was to introduce the concept of `Target`. A `Target` is a triple `<Platform, mode, language>`, where:

* `Platform` is like in current `master`  (Intel, Device, ...)
* `mode` is `master`'s DLE mode (basic, advanced, ...)
* `language` is a new thing, but it's still unused (it will be in the future). For example, `openmp` is a language, and so are `openacc` and `cuda`.

The idea is that for the same `<Platform, mode>` one may want to have specializations for multiple languages.

You may have guessed that the need for Target stems from code generation for GPUs, but it's easily generalizable to any architecture.

In this rewrite, things have changed as in the figure below:

IMG_1937

In current `master`, the DLE could be specialized on an Platform-basis, but other passes of the compilation pipeline (for example, placing declarations) could not. So, for example, it was not possible to declare that an object lives on the GPU memory, as there was no way to create a GPU-specific pass outside of the DLE. Things, with this PR, are different as shown in the RHS of the image

In this rewrite, I found myself in need to introduce a more structured concept of "compile pass": this was the perfect opportunity to improve logging and provide more information about the devito-compilation time when running with `DEVITO_LOGGING=DEBUG`. You may try it for yourself -- now devito emits information about the hotspot passes, that is the passes that took up most of the compilation time. This is the only new piece of "functionality" added.

Also, observe that:
- `equation.py` was moved to `finite_differences`
- `operator.py` and `profiling.py` were moved to the new `operator` subdirectory
- the `dle` sub-directory has been destroyed and its content redistributed inside the new `targets` sub-directory

The backends (ops, yask) have been updated accordingly

I personally think that despite no new functionalities are added, this is massive improvement over current master, as it makes architecture-specializations vastly easier and neater


devito/core/__init__.py
```diff
@@ -5,31 +5,34 @@
 """
 
 from devito.archinfo import Cpu64, Intel64, Arm, Power, Device
-from devito.dle import (CPU64Rewriter, Intel64Rewriter, ArmRewriter, PowerRewriter,
-                        PlatformRewriter, SpeculativeRewriter, DeviceOffloadingRewriter,
-                        modes)
 from devito.parameters import Parameters, add_sub_configuration
+from devito.targets import (CPU64Target, Intel64Target, ArmTarget, PowerTarget,
+                            CPU64NoopTarget, CustomTarget, DeviceOffloadingTarget,
+                            targets)
```
georgebisbas: Maybe Targets with capital T ?

devito/ir/iet/properties.py
```diff
@@ -30,6 +30,9 @@ def __init__(self, name, val=None):
 VECTOR = IterationProperty('vector-dim')
 """The Iteration can be SIMD-vectorized."""
 
+TILABLE = IterationProperty('tilable')
+"""The Iteration can be tiled (or "loop blocking")."""
```
georgebisbas: nitpicking: or "loop blocking can be applied" (?)

devito/ir/iet/visitors.py
```diff
@@ -678,18 +678,19 @@ class IsPerfectIteration(Visitor):
     def visit_object(self, o, **kwargs):
         return False
 
-    def visit_tuple(self, o, **kwargs):
-        return all(self._visit(i, **kwargs) for i in o)
+    def visit_tuple(self, o, found=False, nomore=False):
```
georgebisbas: nitpicking: TBH nomore is a bit vague. Although I get its use....just seems  'ugly'... ;-)

devito/targets/basic.py
```diff
@@ -0,0 +1,153 @@
...
+from itertools import product
+
+from devito.archinfo import Platform
+from devito.ir.iet import Node
+from devito.logger import warning
+from devito.parameters import configuration
+from devito.targets.common import Graph
+from devito.tools import Singleton
+
+__all__ = ['iet_lower', 'targets', 'targets_registry', 'Target']
```
georgebisbas: I am not sure whether it is good like targets/Target.possible cause of future confusion...just saying...

devito/targets/basic.py
```diff
@@ -0,0 +1,153 @@
...
+
+    """
+    A special mapper for Targets:
+
+        (platform, mode, language) -> target
+
+    where:
+
+        * `platform` is an object of type Platform, that is the architecture
+          the generated code should be specializer for.
```
georgebisbas: typo: specialized

devito/targets/basic.py
```diff
@@ -0,0 +1,153 @@
...
+                        communications, ...).
+    options : dict, optional
+        - ``openmp``: Enable/disable OpenMP. Defaults to `configuration['openmp']`.
+        - ``mpi``: Enable/disable MPI. Defaults to `configuration['mpi']`.
+        - ``blockinner``: Enable/disable blocking of innermost loops. By default,
+                          this is disabled to maximize SIMD vectorization. Pass True
+                          to override this heuristic.
+        - ``blocklevels``: Levels of blocking for hierarchical tiling (blocks,
+                           sub-blocks, sub-sub-blocks, ...). Different Platforms have
+                           different default values.
```
georgebisbas: Add : ...Defaults to...

devito/targets/basic.py
```diff
@@ -0,0 +1,153 @@
...
+        else:
+            params[k] = v
+
+    # Force OpenMP/MPI if parallelism was requested, even though mode is 'noop'
+    if mode is None:
+        mode = 'noop'
+    elif mode == 'noop':
+        mode = tuple(i for i in ['mpi', 'openmp'] if params[i]) or 'noop'
+
+    # Fetch the requested rewriter
```
georgebisbas: Should the 'rewriter' terminology be changed after the Rewriter  classes are not there anymore?
I mean no classes from DLE anymore as a term. Just a thought...

devito/core/__init__.py
FabioLuporini: it's a global variable

devito/ir/iet/visitors.py
FabioLuporini: any suggestion? I just reused what was there before

devito/targets/basic.py
FabioLuporini: I don't follow? could you elaborate? you mean the names?

devito/targets/basic.py
FabioLuporini: yes, you're right

devito/targets/common/blocking.py
```diff
@@ -36,28 +37,12 @@ def make_blocking(self, iet):
         block_dims = []
         for tree in retrieve_iteration_tree(iet):
             # Is the Iteration tree blockable ?
```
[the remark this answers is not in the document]
FabioLuporini: they are interchangeable at this point

devito/targets/basic.py
georgebisbas: Yes, I feel like that so many target*** vaiables may sometime cause confusion.

devito/targets/common/mpi.py
```diff
@@ -0,0 +1,149 @@
...
+    #   <HaloSpot(u)>, in y    ---->   <Iteration x>
+    #   <Iteration y>                    <Iteration y>
+    mapper = {}
+    for i, halo_spots in MapNodes(Iteration, HaloSpot).visit(iet).items():
+        hoistable = [hs for hs in halo_spots if hs.hoistable]
+        if not hoistable:
+            continue
+        elif len(hoistable) > 1:
+            # We should never end up here, but for now we can't prove it formally
+            perf_adv("Found multiple hoistable HaloSpots, skipping optimization")
```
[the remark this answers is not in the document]
FabioLuporini: I think `perf_adv` is OK

devito/core/__init__.py
georgebisbas: ok

devito/targets/basic.py
FabioLuporini: gotcha, thanks

devito/ir/iet/visitors.py
btrb: maybe `final`, `last`, `ultimate`, or `ending` ?

devito/finite_differences/__init__.py
```diff
@@ -1,6 +1,7 @@
...
-from devito.finite_differences.tools import generate_fd_shortcuts  # noqa
-from devito.finite_differences.coefficients import *  # noqa
-from devito.finite_differences.operators import *  # noqa
+from .differentiable import *  # noqa
+from .finite_difference import *  # noqa
+from .derivative import *  # noqa
+from .tools import generate_fd_shortcuts  # noqa
+from .coefficients import *  # noqa
+from .operators import *  # noqa
+from .equation import *  # noqa
```
vmickus: I am curious. Why the relative import? 

Is it because now each target can implement those classes and with a relative import it can be automatically loaded, something like this?

devito/ir/iet/analysis.py
```diff
@@ -2,8 +2,9 @@
 from functools import cmp_to_key
 
 from devito.ir.iet import (Iteration, HaloSpot, SEQUENTIAL, PARALLEL, PARALLEL_IF_ATOMIC,
-                           VECTOR, WRAPPABLE, ROUNDABLE, AFFINE, OVERLAPPABLE, hoistable,
-                           useless, MapNodes, Transformer, retrieve_iteration_tree)
+                           VECTOR, TILABLE, WRAPPABLE, ROUNDABLE, AFFINE, OVERLAPPABLE,
+                           hoistable, useless, MapNodes, Transformer,
```
vmickus: nitpicking: I am not sure about the sequence here.  
Should `hoistable` and `useless` be after `Transformer`?  And  `MapNodes` and `Transformer` closer to `Iteration` and `HaloSpot`?

devito/yask/targets.py
```diff
@@ -40,7 +40,7 @@ def _pipeline(self, graph):
         avoid_denormals(graph)
         loop_wrapping(graph)
         if self.params['openmp']:
-            self.ompizer.make_openmp(graph)
+            self.ompizer.make_openmp_parallel(graph)
```
btrb: did you mean `make_parallel` ? i can't find the definition of `make_openmp_parallel`

devito/yask/targets.py
```diff
@@ -54,5 +54,5 @@ def passes_mapper(self):
         return {
             'denormals': partial(avoid_denormals),
             'wrapping': partial(loop_wrapping),
-            'openmp': partial(self.ompizer.make_openmp),
+            'openmp': partial(self.ompizer.make_openmp_parallel),
```
btrb: the same here

devito/compiler.py
```diff
@@ -225,7 +224,6 @@ def make(self, loc, args):
         logfile = path.join(self.get_jit_dir(), "%s.log" % hash_key)
         errfile = path.join(self.get_jit_dir(), "%s.err" % hash_key)
 
-        tic = time()
```
mloubout: why? Actually useful I find to know how ling it takes to compile

devito/finite_differences/equation.py
```diff
@@ -78,7 +78,8 @@ def subdomain(self):
     def evaluate(self):
         """
         Evaluate the Equation or system of Equations.
-        The rhs of the Equation is evaluated at the indices of the lhs if required.
+
```
mloubout: ?

And really not sure should be in FD

devito/ir/clusters/algorithms.py
```diff
@@ -6,12 +6,13 @@
 from devito.ir.support import Any, Backward, Forward, IterationSpace, Scope
 from devito.ir.clusters.cluster import Cluster, ClusterGroup
 from devito.symbolics import CondEq, xreplace_indices
-from devito.tools import DAG, as_tuple, flatten, filter_ordered, generator
+from devito.tools import DAG, as_tuple, flatten, filter_ordered, generator, timed_pass
```
mloubout: timed_pass? wouldn't `time_passed` be better?

devito/ir/equations/equation.py
```diff
@@ -1,10 +1,10 @@
 import sympy
 
-from devito.equation import Eq
 from devito.ir.equations.algorithms import dimension_sort
 from devito.ir.support import (IterationSpace, DataSpace, Interval, IntervalGroup,
                                Stencil, detect_accesses, detect_oobs, detect_io,
                                build_intervals, build_iterators)
+from devito.finite_differences import Eq
```
mloubout: why is Eq in finite-differences now?

devito/operator/operator.py
```diff
@@ -266,47 +273,63 @@ def _apply_substitutions(self, expressions, subs):
...
-        self._func_table.update({i: MetaCall(None, False) for i in profiler._ext_calls})
-        return iet, profiler
+    @classmethod
+    @timed_pass
+    def _lower_exprs(cls, expressions, **kwargs):
+        """
+        Expression lowering:
+
+            * Form and gather any required implicit expressions;
+            * Evaluation of derivatives;
```
mloubout: Evaluate

devito/operator/operator.py
```diff
@@ -266,47 +273,63 @@ def _apply_substitutions(self, expressions, subs):
...
+            * Flatten vectorial equations;
+            * Indexification;
```
mloubout: Indexify

devito/operator/operator.py
```diff
@@ -266,47 +273,63 @@ def _apply_substitutions(self, expressions, subs):
...
+            * Apply substitution rules;
+            * Specialization (e.g., index shifting from Halo to Domain)
```
mloubout: Specialize

devito/compiler.py
FabioLuporini: `make` is never used except for yask

now timings all go via `timed_pass`

devito/finite_differences/equation.py
FabioLuporini: I can move it to somewhere else if you prefer -- types ?

definitely it needs to get out of the root directory

I think `finite_difference` is okish to be honest... an alternative is `types`. I can move it there if you prefer?

devito/ir/clusters/algorithms.py
FabioLuporini: "pass" stands for "compiler pass". So it's a "timed compiler pass". It's pretty standard compiler jargon

devito/finite_differences/equation.py
mloubout: I'd say in `extended_sympy` personally

FabioLuporini: no, definitely it can't go there. 

`Eq` is an actual type of the DSL, and `extended_sympy` has got nothing to do with it, it's just some sympy refinements.

The DSL types are spread across `finite_differences`, `types`, and `operator`. This separation is quite neat I think, so I'm not debating it. But I agree with you `Eq` can have a better place, only not `extended_sympy`. I'll think about that, but `types` actually might not be such a bad place. Or peraphs the new `operator` directory too... but it's a bit weird, because I'd prefer to keep `Eq` together with the other symbolic objects

devito/operator/operator.py
FabioLuporini: I will fix these

devito/compiler.py
mloubout: ok

devito/finite_differences/equation.py
mloubout: I get your point. I think type is a bit too overcrowded though, maybe we should split it into
- "datatypes" (Function, ...)
-"devito-sympy types" (Eq, ..)

FabioLuporini: Maybe, I don't know, but I'd refrain from creating new directories and shuffling other modules around in this PR . If, instead, it's just a matter of moving a specific module (`equation.py`), then not a problem; however, if it's more than that (and in principle I'd have no objections) I'd wait until the stack of currently open PRs will have landed

devito/finite_differences/__init__.py
FabioLuporini: no it's simply "why not", I think relatives are fine in `__init__.py` ?

tests/test_dle.py
```diff
@@ -505,10 +509,10 @@ def test_basic(self):
 
         assert trees[0][1].pragmas[0].value ==\
             'omp target teams distribute parallel for collapse(3)'
-        assert op.body[0].body[1][0].body[0].header[0].value ==\
+        assert op.body[0].body[0].header[0].value ==\
```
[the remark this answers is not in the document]
FabioLuporini: I've managed to systematically drop an useless extra `List`, AFAIR

devito/targets/basic.py
FabioLuporini: This docstring is changing in one of the follow up PRs, so let's just wait for it I'd say

devito/ir/iet/visitors.py
FabioLuporini: I think `nomore` conveys the idea to be fair

all your suggestions don't make much sense to me

CONTRIBUTING.md
```diff
@@ -5,7 +5,7 @@ We always welcome third-party contributions. And we would love you to become an
 ### Reporting issues
 
 There are several options:
-* Talk to us. You can join our Slack team via this [link](https://devito-slackin.now.sh/). Should you have installation issues, or should you bump into something that appears to be a Devito-related bug, do not hesitate to get in touch. We are always keen to help out.
+* Talk to us. You can join our Slack team via this [link](https://opesci-slackin.now.sh/). Should you have installation issues, or should you bump into something that appears to be a Devito-related bug, do not hesitate to get in touch. We are always keen to help out.
```
mloubout: Need rebase on master to not reintroduce the opesci, same below

devito/operator/operator.py
```diff
@@ -128,100 +127,102 @@ class Operator(Callable):
...
+        with timed_region('op-compile') as r:
+            op = cls._compile(expressions, **kwargs)
+        op._profiler.py_timers.update(r.timings)
+
+        op._emit_compile_profiling()
+
+        return op
+
+    @classmethod
+    def _compile(cls, expressions, **kwargs):
```
mloubout: Not sure an appropriate name, des more than compile, maybe `_build`?

devito/operator/operator.py
```diff
@@ -128,100 +127,102 @@ class Operator(Callable):
...
 
-        # Finalization: introduce declarations, type casts, etc
-        iet = self._finalize(iet, parameters)
+        # Lower IET to a Target-specific IET
+        op, target_state = cls._specialize_iet(op, **kwargs)
 
-        super(Operator, self).__init__(self.name, iet, 'int', parameters, ())
+        # Make it an actual Operator
+        op = Callable.__new__(cls, **op.args)
+        Callable.__init__(op, **op.args)
```
mloubout: Isn't this called at `__new__`?

devito/finite_differences/equation.py
mloubout: I think I prefer it in types then.

devito/targets/basic.py
FabioLuporini: I'm dropping the name "target" in one of the subsequent PRs

devito/targets/basic.py
FabioLuporini: Dropping this in a subsequent PR. Will leave it for now, but it was a fair point

devito/finite_differences/equation.py
FabioLuporini: OK, will do

devito/operator/operator.py
FabioLuporini: done

devito/operator/operator.py
FabioLuporini: done

devito/operator/operator.py
FabioLuporini: done

devito/core/operator.py
```diff
@@ -1,19 +1,20 @@
...
 from devito.operator import Operator
+from devito.targets import NThreads
 
 __all__ = ['OperatorCore']
 
 
 class OperatorCore(Operator):
 
-    def _specialize_exprs(self, expressions):
+    @classmethod
```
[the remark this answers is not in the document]
FabioLuporini: a very long story.... but basically I need `__new__` instead of `__init__`. So everything called from the former `__init__` needs to be turned into a `classmethod`. Note this is also much more conceptually correct than before: now we are incrementally building a Callable/Operator, while previously we were initializing its state (body). There is a subtle yet meaningful difference

devito/operator/operator.py
FabioLuporini: it needs to be called explicitly at this point so that we have a proper signature, which allows us to proceed with the later passes

devito/operator/operator.py
FabioLuporini: fair enough, changing

PR #1027, opened 2019-12-16
Analysis at a higher level of abstraction

FabioLuporini:
**WARNING**: this PR is based on top of #1026 , so it's pointless to look at the diff with master until #1026 will have been merged. 

This PR moves several analysis passes from the IET level to the Cluster level. 

No new functionality is added. This is in preparation for another PR.


devito/operator/profiling.py
```diff
@@ -104,7 +104,7 @@ def timer_on(self, name, comm=None):
             If provided, the global execution time is derived by a single MPI
             rank, with timers started and stopped right after an MPI barrier.
         """
-        if comm is not MPI.COMM_NULL:
+        if comm and comm is not MPI.COMM_NULL:
```
georgebisbas: This should have been spotted earlier on, right?

devito/ir/iet/nodes.py
```diff
@@ -11,9 +11,9 @@
 
 from devito.data import FULL
 from devito.ir.equations import ClusterizedEq
-from devito.ir.iet import (IterationProperty, SEQUENTIAL, PARALLEL, PARALLEL_IF_ATOMIC,
-                           VECTOR, WRAPPABLE, ROUNDABLE, AFFINE, OVERLAPPABLE)
-from devito.ir.support import Forward, detect_io
+from devito.ir.support import (SEQUENTIAL, PARALLEL, PARALLEL_IF_ATOMIC, VECTORIZED,
```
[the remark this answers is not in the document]
FabioLuporini: no

devito/ir/support/properties.py
```diff
@@ -0,0 +1,73 @@
...
+PARALLEL = Property('parallel')
+"""A fully parallel Dimension."""
+
+PARALLEL_IF_ATOMIC = Property('parallel_if_atomic')
+"""A parallel Dimension with local reductions."""
+
+COLLAPSED = lambda i: Property('collapsed', i)
+"""Collapsing Dimensions."""
+
+VECTORIZED = Property('vector-dim')
```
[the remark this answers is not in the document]
FabioLuporini: no, now it's really VECTORIZED

devito/ir/iet/nodes.py
```diff
@@ -366,7 +366,7 @@ class Iteration(Node):
         The for-loop direction. Accepted:
         - ``Forward``: i += stepping (defaults)
         - ``Backward``: i -= stepping
-    properties : IterationProperty or list of IterationProperty, optional
+    properties : Property or list of Property, optional
```
mloubout: I get the idea, but "Property" feels very vague/general, maybe "CompilerProperty" or "PasssProperty" or just keep the previous IterationProperty

devito/ir/support/properties.py
```diff
@@ -0,0 +1,73 @@
+from devito.tools import Tag
+
+
+class Property(Tag):
```
mloubout: As above, don't really like the name.

devito/ir/support/properties.py
```diff
@@ -0,0 +1,73 @@
...
+"""A SIMD-vectorized Dimension."""
+
+TILABLE = Property('tilable')
+"""A fully parallel Dimension that would benefit from tiling (or "blocking")."""
+
+WRAPPABLE = Property('wrappable')
+"""
+A modulo-N Dimension (i.e., cycling over i, i+1, i+2, ..., i+N-1) that could
+safely be turned into a modulo-K Dimension, with K < N. For example:
+u[t+1, ...] = f(u[t, ...], u[t-1, ...]) --> u[t-1, ...] = f(u[t, ...], u[t-1, ...]).
```
mloubout: that's `t-1` on both side of `=` in the second part.

tests/test_autotuner.py
```diff
@@ -96,7 +96,7 @@ def test_mode_destructive():
     grid = Grid(shape=(96, 96, 96))
     f = TimeFunction(name='f', grid=grid, time_order=0)
 
-    op = Operator(Eq(f, f + 1.), dle=('advanced', {'openmp': False}))
+    op = Operator(Eq(f.forward, f + 1.), dle=('advanced', {'openmp': False}))
```
mloubout: why?

tests/test_operator.py
```diff
@@ -498,7 +498,7 @@ def test_default_functions(self):
         grid = Grid(shape=(5, 6, 7))
         f = TimeFunction(name='f', grid=grid)
         g = Function(name='g', grid=grid)
-        op = Operator(Eq(g, g + f), dle=('advanced', {'openmp': False}))
+        op = Operator(Eq(f.forward, g + f), dle=('advanced', {'openmp': False}))
```
mloubout: why?

devito/ir/clusters/cluster.py
```diff
@@ -51,17 +55,32 @@ def from_clusters(cls, *clusters):
...
-    def rebuild(self, exprs):
+    def rebuild(self, *args, **kwargs):
         """
-        Build a new Cluster from a sequence of expressions. All other attributes
-        are inherited from ``self``.
+        Build a new Cluster from the attributes given as keywords. All other
+        attributes are taken from ``self``.
         """
-        return Cluster(exprs, self.ispace, self.dspace, self.guards)
+        # Shortcut for backwards compatibility
```
mloubout: Maybe raise a warning if this is only there for backward compatibility.

devito/ir/iet/nodes.py
georgebisbas: Would vote for IterationProperty

FabioLuporini: It is **not** an IterationProperty (anymore), otherwise I'd have kept the original name

devito/ir/support/properties.py
FabioLuporini: yes, that's the point. A very special case in which you could get away with two timeslots instead of three (here it's t-1 and t, could have been t+1 and t)

tests/test_autotuner.py
FabioLuporini: I think it was just a leftover. I have reinstated the origin version

tests/test_operator.py
FabioLuporini: ``f`` would have worked too. 

But not ``g``. With ``g`` on the LHS, this is a fully parallel loop nest -- you may go in parallel even over time -- and heuristically the compiler doesn't do blocking if there is not at least one SEQUENTIAL Iteration

the test however expects `x0_blk_size`, ie it expects blocking, so...

devito/ir/clusters/cluster.py
FabioLuporini: oh no this is just an internal thing

and in fact, no worries -- this is changing again in one of the subsequent PRs (getting a proper cleanup)

devito/ir/support/properties.py
mloubout: If it's inside `f` it's note the case as it implies it's not just a overwrite but need to know `t-1` everywhere.

PR #1031, opened 2019-12-23
FD extras

mloubout:
Added centred FD shortcuts for so=1, fd_o=2 (`.dxc`,...) that some people may need


devito/finite_differences/derivative.py
```diff
@@ -222,7 +222,7 @@ def evaluate(self):
 
     def _eval_fd(self, expr):
         expr = getattr(expr, 'evaluate', expr)
-        if self.side in [left, right] and self.deriv_order == 1:
+        if self.side in [left, right, centered] and self.deriv_order == 1:
```
FabioLuporini: nitpicking: neater as ``if self.side is not None`` ?

devito/finite_differences/derivative.py
```diff
@@ -135,7 +135,7 @@ def __new__(cls, expr, *dims, **kwargs):
         obj._dims = tuple(OrderedDict.fromkeys(new_dims))
         obj._fd_order = fd_orders
         obj._deriv_order = orders
-        obj._side = kwargs.get("side", centered)
+        obj._side = kwargs.get("side", None)
```
FabioLuporini: `.get` assigns `None` automatically, ie it should be `kwargs.get("side")`

btw, do we actually need to default to `None`?

devito/finite_differences/tools.py
```diff
@@ -128,14 +128,14 @@ def deriv_function(expr, deriv_order, dims, fd_order, side=centered, **kwargs):
     # Add non-conventional, non-centered first-order FDs
     for d, o in zip(dimensions, orders):
         name = 't' if d.is_Time else d.root.name
-        if function.is_Staggered:
+        if function.is_Staggered or o == 2:
```
FabioLuporini: so if `o == 2` **and** the function isn't staggered, we will enter two `if`s -- this one and the one at line 138 ?

devito/finite_differences/derivative.py
mloubout: Yes we need None

devito/finite_differences/derivative.py
mloubout: Clearer like that what the options can be IMO

devito/finite_differences/tools.py
mloubout: yes

FabioLuporini:
we can land this as soon as travis behaves again

FabioLuporini:
Merged.

PR #1033, opened 2019-12-28
opesci purge

ggorman:
Getting rid of all references to opesci.


benchmarks/user/benchmark.py
```diff
@@ -433,12 +433,12 @@ def param_string(self, params):
 
 
 def get_ob_exec(func):
-    """Return a special ``opescibench.Executor`` to execute performance runs."""
+    """Return a special ``devitobench.Executor`` to execute performance runs."""
     try:
-        from opescibench import Executor
+        from devitobench import Executor
```
mloubout: Subfolder/module in devitobench is still called opescibench so this "may" break.

ggorman: Fixed.

FabioLuporini:
Thanks.

what's the output of
```
git grep opesci
```
and
```
git grep OPESCI
```

from the root directory?

PR #1040, opened 2020-01-02
MPI testing

mloubout:
This one is on me, setup of the Acoustic Class in `test_mpi` was pretty stupid and not doing what I thought i made it do, this should save quite the time in CI (~32 less PDE solve with mpi) and make `test_mpi.py/TestAcoustic` about two times faster (eyeballing it).

It is no super pretty but the "hack" is necessary because basically `purest.mark.parallel` starts a new `pytest` session each time and therefore creates a new class an "regenerate" the norms (and because at that point it's inside MPI it would recompute the norms with MPI). So now I just dump it locally (it's tiny so it's fine) so that the reference norms are computed once *in serial mode* instead of like 30 times.

Travis : Saves 2-4 min per MPI build
Pipeline: Saves ~10-15min per MPI build

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -2,7 +2,7 @@ trigger:
 - master
 
 pr:
-- master  
+- master
 
 jobs:
 
@@ -36,14 +36,14 @@ jobs:
         DEVITO_BACKEND: 'core'
         RUN_EXAMPLES: 'true'
         PYTHON_VERSION: '3.7'
-      nompi-conda-gcc5:
+      nompi-conda-gcc5-py38:
         gccVersion: '5'
-        DEVITO_OPENMP: '0'
+        DEVITO_OPENMP: '1'
         MPI_INSTALL: '0'
         installWithPip: 'false'
         DEVITO_BACKEND: 'core'
         RUN_EXAMPLES: 'false'
-        PYTHON_VERSION: '3.7'
+        PYTHON_VERSION: '3.8'
       ops-conda-gcc7:
         gccVersion: '7'
         DEVITO_OPENMP: '0'
--- a/environment.yml
+++ b/environment.yml
@@ -3,7 +3,7 @@ channels:
   - defaults
   - conda-forge
 dependencies:
-  - python>=3.6
+  - python=3
   - numpy>=1.14
   - sympy>=1.4
   - cgen
--- a/tests/test_adjoint.py
+++ b/tests/test_adjoint.py
@@ -1,9 +1,8 @@
 import numpy as np
 import pytest
-from numpy import linalg
 
 from conftest import unit_box, points, skipif
-from devito import Operator
+from devito import Operator, norm
 from devito.logger import info
 from examples.seismic import demo_model, Receiver
 from examples.seismic.acoustic import acoustic_setup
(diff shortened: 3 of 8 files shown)


tests/test_mpi.py
```diff
@@ -12,7 +13,7 @@
 from devito.mpi import MPI
 from examples.seismic.acoustic import acoustic_setup
 
-pytestmark = skipif(['yask', 'ops', 'nompi'])
+pytestmark = skipif(['yask', 'ops', 'nompi'], whole_module=True)
```
FabioLuporini: can you remind me what the `whole_module` is for?

mloubout: This skips the entire file without reading it. Without it,  `pytest` collects and initializes the classes then skip the tests only, so this avoids initialization of TestAcoustic for the non-mpi builds.

azure-pipelines.yml
```diff
@@ -36,14 +36,14 @@ jobs:
         DEVITO_BACKEND: 'core'
         RUN_EXAMPLES: 'true'
         PYTHON_VERSION: '3.7'
-      nompi-conda-gcc5:
+      nompi-conda-gcc5-py38:
         gccVersion: '5'
-        DEVITO_OPENMP: '0'
+        DEVITO_OPENMP: '1'
```
FabioLuporini: why this one?

environment.yml
```diff
@@ -3,7 +3,7 @@ channels:
   - defaults
   - conda-forge
 dependencies:
-  - python>=3.6
+  - python=3
```
FabioLuporini: what's the effect of this ?

tests/test_adjoint.py
```diff
@@ -58,7 +57,7 @@ def test_adjoint_F(self, mkey, shape, kernel, space_order):
 
         # Adjoint test: Verify <Ax,y> matches  <x, A^Ty> closely
         term1 = np.dot(srca.data.reshape(-1), solver.geometry.src.data)
-        term2 = linalg.norm(rec.data.reshape(-1)) ** 2
+        term2 = norm(rec) ** 2
```
FabioLuporini: note that this is causing more and more codegens and compiles, thus it also affects the CI turnaround time

tests/test_installation.py
```diff
@@ -0,0 +1,14 @@
+import platform
+import os
+
+
+def test_python_version():
```
FabioLuporini: The test is OK, but instead of creating a new file, how about merging test_dependencies.py and test_installation.py into a single "more general" file?

azure-pipelines.yml
mloubout: No openmp already tested no reason to not use it

environment.yml
mloubout: None, travis/pipeline replace this line, and for users it just installs any python3, which is 3.7 by default in mini conda

tests/test_adjoint.py
mloubout: Not by much

azure-pipelines.yml
FabioLuporini: makes it slower?

mloubout: That build is ~10min faster tan before

FabioLuporini: thanks to openmp?

mloubout: Probably partially yes, all azure instances have at least 2 CPU

PR #1041, opened 2020-01-06
change DependenceGroup to subclass set instead of list

tjb900:
It feels like it can't be this easy, but will see what CI has to say!


FabioLuporini:
yes, I like this. Let's see what CI has to say

mloubout:
LGTM

PR #1042, opened 2020-01-06
Test cleanup

mloubout:
cleanup conftest
fwi/dask notebook extra runs and assert
mere install/dependency tests into one


tests/test_dle.py
```diff
@@ -270,33 +269,35 @@ class TestNodeParallelism(object):
...
-                                exprs, expected, iters):
-        scope = [fa, fb, fc, fd, t0, t1, t2, t3]
-        node_exprs = [Expression(DummyEq(EVAL(i, *scope))) for i in exprs]
-        iet = iters[6](iters[7](node_exprs))
-
-        parameters = derive_parameters(iet, True)
-        iet = Callable('kernel', iet, 'int', parameters)
-
-        iet = iet_analyze(iet)
+    def test_iterations_ompized(self, exprs, expected):
```
mloubout: Note: Ignore these changes, This is just to pass test, premptive pseudo merge of analysis PR will be ovewriteen

tests/test_ir.py
```diff
@@ -33,14 +49,22 @@ def v_num(self):
...
         vs3 = Vector(s + 3, smart=True)
         return vx, vxy, vx1y, vs3
 
+    @pytest.fixture
+    def fa(self, grid, x):
+        return Array(name='fa', dimensions=(x,), shape=(3,)).indexed
+
+    @pytest.fixture
+    def fc(self, grid, x, y):
+        return Array(name='fc', shape=(3, 5), dimensions=(x, y)).indexed
```
georgebisbas: Fixing argument order for consistency

.travis.yml
```diff
@@ -116,7 +116,7 @@ script:
   - flake8 --builtins=ArgumentError .
   # Run test with py.tests
   - if [[ $INSTALL_TYPE == 'pip_setup' ]]; then python setup.py test; fi
-  - if [[ $INSTALL_TYPE == 'conda' ]]; then py.test --durations=20 --maxfail=5 $COV devito tests/; fi
+  - if [[ $INSTALL_TYPE == 'conda' ]]; then py.test $COV devito tests/; fi
```
FabioLuporini: why? I actually like those two options

tests/test_dse.py
```diff
@@ -57,7 +57,16 @@ def test_scheduling_after_rewrite():
       'Eq(tu, ((ti0*ti1*t0)*tv + (ti0*ti1*tv)*t1))'],
      ['t0 + 2.0', 't0*ti0[x, y, z]*ti1[x, y, z]', 't1*ti0[x, y, z]*ti1[x, y, z]']),
 ])
-def test_yreplace_time_invariants(tu, tv, tw, ti0, ti1, t0, t1, exprs, expected):
+def test_yreplace_time_invariants(exprs, expected):
+    grid = Grid((3, 3, 3))
+    dims = grid.dimensions
+    tu = TimeFunction(name="tu", grid=grid, space_order=4).indexify()
```
FabioLuporini: I'm writing it here, but basically this comment applies to all your edits -- I don't actually think we need the `.indexify()` ? have you tried without? 
When doing this sort of cleanup, I never needed it

tests/conftest.py
```diff
@@ -58,221 +55,6 @@ def skipif(items, whole_module=False):
             return pytest.mark.skip(skipit)
 
 
-# Testing dimensions for space and time
```
FabioLuporini: dropping all this crap -- absolutely fantastic

setup.cfg
```diff
@@ -4,6 +4,7 @@ ignore = F403,E226,E731,W503,F405,E722,E741,W504,W605
 
 [tool:pytest]
 python_files = test_*.py
+addopts = --durations=20 --maxfail=5
```
FabioLuporini: ahhhhhhhhhh ignore my previous comment

.travis.yml
FabioLuporini: OK -- seen the change to .cfg, ignore this !

tests/test_dse.py
mloubout: Object are indexed in the expressions to evaluate so yeah it breaks without it, had to go back to figure out why the test where failing after the changes and put it back.

tests/test_ir.py
mloubout: It's kwargs doesn't really matter, I'll try to remember to change it though.

.travis.yml
mloubout: You'll notice that adding it to .cfg also now prints the slowest cells for the notebooks testing which is nice.

tests/test_dse.py
FabioLuporini: ahh... I think what other tests do is to use parentheses `(x, y, z)` instead of index notation in the expressions

tests/test_operator.py
```diff
@@ -1474,7 +1473,13 @@ def test_equations_emulate_bc(self, t0):
         assert len(trees) == 4
         assert all(id(trees[0][0]) == id(i[0]) for i in trees)
 
-    def test_different_section_nests(self, tu, ti0, t0, t1):
+    def test_different_section_nests(self):
+        grid = Grid((3, 3, 3))
+        tu = TimeFunction(name='tu', grid=grid, space_order=4)
+        t0 = Scalar(name='t0').indexify()
```
FabioLuporini: these indexify should be unnecessary

tests/test_operator.py
```diff
@@ -1457,12 +1453,15 @@ def test_expressions_imperfect_loops(self):
...
+    def test_equations_emulate_bc(self):
         """
         Test that bc-like equations get inserted into the same loop nest
         as the "main" equations.
         """
-        grid = Grid(shape=(3, 3, 3), dimensions=(x, y, z), time_dimension=time)
+        grid = Grid(shape=(3, 3, 3))
+        x, y, z = grid.dimensions
+        time = grid.time_dim
+        t0 = Scalar(name='t0').indexify()
```
FabioLuporini: this indexify should be unncessary

examples/seismic/model.py
```diff
@@ -215,6 +214,13 @@ def __init__(self, origin, spacing, shape, space_order, vp, nbl=20,
         if self.grid.dim > 2:
             self.phi = self._gen_phys_param(phi, 'phi', space_order)
 
+    @property
+    def _max_vp(self):
+        # mmax not supported with yask
+        if configuration['backend'] == 'yask':
```
FabioLuporini: this `if` cannot be here

how does it fail with yask? is there anything we can do about that?

shouldn't this check go inside `mmax` / `mmin` ?

mloubout: no `data_ro_domain` error with task. Basically had to turn `)max_vp` in a property because pickling the model was messing with it with the previous version. I'll move it to max

PR #1048, opened 2020-01-09
Added gh action for flake8

ggorman:
Added github action to perform flake8 using the same arguments as existing travis test.

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -1,0 +1,23 @@
+name: Python application
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements.txt
+    - name: Lint with flake8
+      run: |
+        pip install flake8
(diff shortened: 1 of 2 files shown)


.github/workflows/pythonapp.yml
```diff
@@ -0,0 +1,23 @@
...
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements.txt
+    - name: Lint with flake8
+      run: |
+        pip install flake8
```
georgebisbas: nitpicking: Installing flake8 is redundant?
Already installed before?
This is what I can see from the logs if I am not mistaken.

.github/workflows/pythonapp.yml
```diff
@@ -0,0 +1,23 @@
...
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+    - name: Install dependencies
```
mloubout: Only need flake 8 can skip stalling dependencies.

.github/workflows/pythonapp.yml
```diff
@@ -0,0 +1,23 @@
+name: Python application
```
mloubout: This is a too generic name, we are gonna have multiple actions so the name should be explicit, like "Test flake8:, and the yaml file should be renamed accordingly, like `flake.yml`

Issue #1057, opened 2020-01-09
Add action: deploy docs to website, on release


EdCaunt:

navjotk:
Automate documentation deployment on release. 
Currently done through travis here: https://github.com/devitocodes/devito/blob/master/.travis.yml#L149

Let me know if you need more information to get started

mloubout:
#1061 closing

PR #1061, opened 2020-01-09
Initial documentation action

BlockSprintZIf:


.github/workflows/documentation.yml
```diff
@@ -0,0 +1,19 @@
+name: CI
```
mloubout: documentation

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,19 @@
...
+
+on: [release]
```
mloubout: push to master maybe? I think the doc should be up to date with GitHub version, @FabioLuporini @ggorman ?

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,19 @@
...
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Run documentation script
```
mloubout: Copy paster leftover?

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,19 @@
...
+      run: echo Hello, world Documentation!
+    - name: Run a multi-line script
```
mloubout: Better name would be good

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,19 @@
...
+      run: |
+        echo Add other actions to build,
+        echo test, and deploy your project.
```
mloubout: These two echo are maybe not necessary

navjotk:
Unsurprisingly, we are not the first people to try to use doctr with Github Actions: https://github.com/drdoctr/doctr/issues/351

navjotk:
Addresses https://github.com/devitocodes/devito/issues/1057

.github/workflows/documentation.yml
mloubout: on:
   # Trigger the workflow on push or pull request,
   # but only for the master branch
   push:
     branches:
       - master

mloubout:
the `make html` part is missing

FabioLuporini:
how can we test whether this actually works ?

navjotk:
I'm testing this here: https://github.com/navjotk/devito/pull/1

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,25 @@
...
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+      with:
+        python-version: 3.7
+    - name: Run documentation script
+      run: |
```
mloubout: Would keep the actions syntax. organization:

```
- name: install dependencies
  run: |
    python -m pip install --upgrade pip 
    pip install sphinx doctr

- name: build
  run: make html
  working_directory: docs

- name: Deploy documentation
   run: doctr deploy . --built-docs docs/_build/html --key-path .deploy_key.enc
```

navjotk: I think your comment is outdated

mloubout: It's not, pip and make in same part, and actions uses `working_directory` not `cd`.  Also `actions-gh-pages` doesn't to use doctr so doesn't need to be installed.

mloubout: Sorry this one was outdated yes

PR #1063, opened 2020-01-09
Update local installations to include pip install of requirements

sgibson91:
This PR updates the installation instructions to include `pip install -r requirements.txt` so that they are complete for running the tutorial notebooks.

Another way of achieving this would be including the contents of `requirements.txt` as a `pip` block in `environment.yaml`, making `requirements.txt` redundant. (**cc:** @tallamjr)

Related issue: https://github.com/devitocodes/devito/issues/1065

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ git clone https://github.com/devitocodes/devito.git
 cd devito
 conda env create -f environment.yml
 source activate devito
+pip install -r requirements.txt
 pip install -e .
 ```
 


README.md
```diff
@@ -32,6 +32,7 @@ git clone https://github.com/devitocodes/devito.git
 cd devito
 conda env create -f environment.yml
 source activate devito
+pip install -r requirements.txt
```
mloubout: That is not necessary, the environment.yml contains all the requirements and installs it

sgibson91: Thanks @mloubout we're having a discussion here: https://github.com/devitocodes/devito/issues/1065

navjotk:
It seems to me that this might not be required eventually. What do you think @sgibson91 ?

sgibson91:
@navjotk Agreed :)

Issue #1062, opened 2020-01-09
Dockerfile does not install all the dependencies


sgibson91:
Installation method: Docker
Executed file: examples/cfd/01_convection.ipynb

Error from cell 1:
```
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-1-0bcda1ec0be4> in <module>
----> 1 from examples.cfd import plot_field, init_hat
      2 import numpy as np
      3 get_ipython().run_line_magic('matplotlib', 'inline')
      4 
      5 # Some variable declarations

/app/examples/cfd/__init__.py in <module>
----> 1 from .tools import *  # noqa

/app/examples/cfd/tools.py in <module>
----> 1 from mpl_toolkits.mplot3d import Axes3D  # noqa
      2 
      3 import numpy as np
      4 from matplotlib import pyplot, cm
      5 from devito.logger import warning

ModuleNotFoundError: No module named 'mpl_toolkits'
```

BlockSprintZIf:
Dockerfile fixed to add matplotlib dependencies

navjotk:
Shall close this when the PR is merged. Shout before then if someone disagrees.

FabioLuporini:
AFAICT the corresponding PR has landed -- closing

Issue #1024, opened 2019-12-05
Matplotlib module is not found when using Docker + Jupyter


vmickus:
When using Devito with Docker the matplotlib is not being installed.
So out of the box, the tutorials that uses matplolib throws the following error when trying to import it:

`ModuleNotFoundError: No module named 'matplotlib'`

 Creating an empty cell before the usage of matplotlib and installing matplotlib fix the problem:

`!pip install matplotlib`

mloubout:
>Creating an empty cell before the usage

Fixing the docker image would be a better  and cleaner way. This links to #817 and homogenisation of docker between the instruction and pipelines

navjotk:
Yes indeed. Not necessarily related to the linked issue.

mloubout:
Couldn't;t find the issue that goes with the homogenisation sorry linked wrong one.

vmickus:
@navjotk Does adding the requirements-optional to the docker image install is a good option?

What I have tried is to change the following lines from the `dockerfile` file:

```
diff --git a/docker/Dockerfile b/docker/Dockerfile
index 0f0e4e50..c07e9d21 100644
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -5,11 +5,13 @@ RUN apt-get update && apt-get install -y -q \
     libmpich-dev 
 
 ADD ./requirements.txt /app/requirements.txt
+ADD ./requirements-optional.txt /app/requirements-optional.txt
 
 RUN python3 -m venv /venv && \
     /venv/bin/pip install --no-cache-dir --upgrade pip && \
     /venv/bin/pip install --no-cache-dir jupyter && \
-    /venv/bin/pip install --no-cache-dir -r /app/requirements.txt
+    /venv/bin/pip install --no-cache-dir -r /app/requirements.txt && \
+    /venv/bin/pip install --no-cache-dir -r /app/requirements-optional.txt
```

As I have just started using Docker, I am not sure if this is correct. 
And basically this solution involves installing the optional requirements by default. 

IMO if the notebooks need something from the optional requirements to work, it shouldn't be optional.

mloubout:
You can look at the pipeline dockerfile for optional arguments.

BlockSprintZIf:
Dockerfile fixed to add dependencies as mentioned by @vmickus above.

mloubout:
FIxed in #1067

PR #1067, opened 2020-01-09
Updated Dockerfile to #1024 #1062

BlockSprintZIf:
Bug fixes for bug nos: #1024 #1062

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -5,12 +5,14 @@ RUN apt-get update && apt-get install -y -q \
     libmpich-dev 
 
 ADD ./requirements.txt /app/requirements.txt
+ADD ./requirements-optional.txt /app/requirements-optional.txt
 
 RUN python3 -m venv /venv && \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -5,12 +5,14 @@ RUN apt-get update && apt-get install -y -q \
     libmpich-dev 
 
 ADD ./requirements.txt /app/requirements.txt
+ADD ./requirements-optional.txt /app/requirements-optional.txt
```
mloubout: maybe could be added as an optional build arguments. You can look at `Dockerfile.pipeline` to see how to mange it and add the if statement, could extend to the libmpich above if not needed

ggorman: I think it's better to include all optional dependencies needed to run all tutorials as this is proving to be a big issue at every hackathon and workshop we participate in - even with otherwise experienced users. 'Experts' can always create whatever bespoke Dockerfile they need as by definition they 'know what they are doing'.

navjotk:
This is GTG I think

PR #1068, opened 2020-01-09
Pytest actions

mloubout:
Added actions for purest for:

- core (7 build):
   - python 3.5, 3.6, 3.7 
   - gcc 4.9, 5, 6, 7, 8, 9
   - with and without openmp 

- Yask (one build) python 3.7  gcc 7
- OPS (one build) python 3.7 gcc 7
- MPI (one build) python 3.7 gcc 7

Removed the existing above ones from travis (still in pipelines). Travis now running only conda install with examples

Already tested at https://github.com/mloubout/devito/runs/381529142

For OSX/Windows, should be just changing `runs-on` and tweak install


.coveragerc
```diff
@@ -3,6 +3,8 @@ branch = True
 omit =
     # Autogenerated missed code handles other VCSes.
     devito/_version.py
+concurrency = multiprocessing
+parallel = True
```
mloubout: Note: to avoid coverage errors with mpi

.github/workflows/pytest-core-mpi.yml
```diff
@@ -0,0 +1,51 @@
...
+        sudo apt-get install -y mpich libmpich-dev
+
+    - name: Install dependencies
+      run: |
+        pip install --upgrade pip
+        pip install -e .[extras]
+
+    - name: Test with pytest
+      run: |
+        pytest --cov --cov-config=.coveragerc --cov-report=xml -m parallel tests/
```
mloubout: Note: same for following files. This `cov` options have to be there as if I put it in the `setup.cfg` file for pytest, the coverage is called inside coverage (pytest calls pytest with mpi) and breaks everything.

Dockerfile.pipelines
```diff
@@ -121,16 +121,12 @@ RUN if [ $installWithPip == "true" ] ; then \
...
       fi ; \
-      pip install pytest-xdist ; \
-      if [ $DEVITO_BACKEND == "yask" ] ; then \
-        conda install swig ; \
-        git clone https://github.com/devitocodes/yask.git /tmp/yask ; \
-        pushd /tmp/yask ; \
-          make compiler-api ; \
-          pip install -e . ; \
-        popd ; \
-      fi ; \
```
mloubout: Note: yask moved to actions

README.md
```diff
@@ -1,6 +1,9 @@
 # Devito: Fast Finite Difference Computation from Symbolic Specification
 
-![Build Status](https://travis-ci.org/devitocodes/devito.svg?branch=master)
+![Build Status for the Core backend](https://github.com/devitocodes/devito/workflows/CI-core/badge.svg)
+![Build Status for the OPS backend](https://github.com/devitocodes/devito/workflows/CI-ops/badge.svg)
+![Build Status for the YASK backend](https://github.com/devitocodes/devito/workflows/CI-yask/badge.svg)
+![Build Status with MPI](https://github.com/devitocodes/devito/workflows/CI-mpi/badge.svg)
```
mloubout: Note: maybe a bit much but this way we have badges for each backend + mpi, doesn't take much space

tests/test_checkpointing.py
```diff
@@ -155,8 +155,8 @@ def test_acoustic_save_and_nosave(shape=(50, 50), spacing=(15.0, 15.0), tn=500.,
     """ Run the acoustic example with and without save=True. Make sure the result is the
     same
     """
-    solver = acoustic_setup(shape=shape, spacing=spacing, nbl=nbl, tn=tn,
-                            space_order=space_order, time_order=time_order)
+    solver = acoustic_setup(shape=shape, spacing=spacing, nbl=nbl, dtype=np.float64,
```
mloubout: Note: because of gcc 7.4.0 that has bad roundup errors

.coveragerc
FabioLuporini: what errors?

.github/workflows/pytest-core-mpi.yml
FabioLuporini: ahhhhh gotcha

README.md
FabioLuporini: I suggest however to remove the OPS and YASK ones

I like CI-core and CI-mpi

FabioLuporini: force-push to master once landed is fine

.coveragerc
mloubout: https://github.com/devitocodes/devito/commit/29b555b15eee3784b9d167454929e514978f41d4/checks?check_suite_id=394482751

for example

README.md
mloubout: Can do that, also will change badges so that clicking on it actually send you to the CI not just the image of the badge

PR #1070, opened 2020-01-09
Created pytest core for mac

BlockSprintZIf:
Based on mloubout's test install branch


navjotk:
Depends on 1068 so merge that before this.

mloubout:
yask does not support OS X so remove that one, not sure about ops (@FabioLuporini)

You probably don't need a new yaml file but you can add cases to the existing `pytest-core`

FabioLuporini:
> not sure about ops 

ops is fine for now as there is no actual offloading (yet)

FabioLuporini:
> You probably don't need a new yaml file but you can add cases to the existing pytest-core

agreed

FabioLuporini:
@navjotk @mloubout @ggorman : do we actually need to run tests on the mac build ? isn't it not just enough to check the installation ? perhaps run `test_operator` only ? 

I think running the entire test suite is overkill

mloubout:
>I think running the entire test suite is overkill

I agree, only thing needed on OSX is to make sure it's installed and compiler works with it so a single test should be fine.

PR #1072, opened 2020-01-10
GitHub Action to publish to pypi

navjotk:


mloubout:
LGTM

FabioLuporini:
> We may want to publish 4.0 too by hand too in case

agreed

FabioLuporini:
LGTM.

Dumb question: where/when can we see this on pipy?

navjotk:
First release after merging this should trigger a publish to pypi, so after that?

FabioLuporini:
shall we merge this?

mloubout:
Yes, but I'd rather merge pytest first then just force merge this one (smaller) than the opposite :D

PR #1073, opened 2020-01-10
Publish releases

iamleeg:
An action to build packages automatically, and publish them to pypi. Only runs on pushing a tag, name your tag after the package version e.g. `git tag -a v4.0.1 -m "Release patch fix 4.0.1"; git push --tags`.

Also: you need to create yourself an API key at pypi.org, and paste it into a secret called `pypi_password` in this repo's settings.

I tested this by renaming the package and publishing to test.pypi.org, which published to https://test.pypi.org/project/devito-iamleeg-fork/.

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -10,6 +10,9 @@ jobs:
       uses: actions/setup-python@v1
       with:
         python-version: 3.7
+    - name: Install wheel
+      run: >-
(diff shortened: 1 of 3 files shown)


mloubout:
isn't this a duplicate of #1072 ?

iamleeg:
It wasn’t at the time I worked on it, but yeah.

PR #1076, opened 2020-01-13
Alow ModelElastic and ModelViscoelastic to also receive subdomains

rhodrin:
Fix an oversight that meant `subdomains` couldn't be passed to `ModelElastic` and `ModelViscoelastic`.


FabioLuporini:
Merged.

PR #1079, opened 2020-01-13
Tutos examples

mloubout:


.github/workflows/tutorials.yml
```diff
@@ -0,0 +1,81 @@
...
+           brew install gcc
+           brew install mpich
+        fi
+
+    - name: Install dependencies
+      run: |
+        pip install --upgrade pip
+        pip install -e .[extras]
+
+    - name: CFD notebooks
```
navjotk: I need to be able to call these tests from within the Docker testing workflow. Splitting them like this means this entire thing needs to repeated for the other workflow. Obviously not a good idea.

mloubout: Separating in step and being able to see and maintain the CI is the whole point of actions, I'm not stashing stuff in some script "for docker" or what's the point of all this, just put everything in a single bash script. This is clear, readable and maintainable.

navjotk: Copying and pasting the same workflow to multiple places is a clear violation of the DRY principle. What principle is being violated by putting things in a shell script?

.github/workflows/tutorials.yml
```diff
@@ -0,0 +1,79 @@
...
+          ubuntu-16.04-gcc-py37,
+          macOS-10.14-gcc-py37
+          ]
+
+        include:
+          - name: ubuntu-16.04-gcc-py37
+            os: ubuntu-16.04
+            compiler: gcc-7
+
+          - name: macOS-10.14-gcc-py37
```
tallamjr: Just to note, it was discussed that many users develop on macOS and often use `clang` as their compiler. Perhaps it would be desirable to test `gcc` on `ubuntu` and test `clang` on `macOS`

At the moment only `gcc-7` and `gcc-9` are tested.

mloubout: Will add a clang build

mloubout: Addec clang

.github/workflows/tutorials.yml
```diff
@@ -0,0 +1,105 @@
...
+      run: |
+        if [ "${{ matrix.compiler }}" = "gcc-9" ]; then
+          brew install gcc
+        else
+          sudo xcode-select -s /Applications/Xcode_11.app/Contents/Developer
+        fi
+        # dask error on osx, skip dask tuto
+        echo "::set-env name=SKIP::--deselect=examples/seismic/tutorials/04_dask.ipynb"
+
+    - name: Buiold docker image
```
FabioLuporini: typo

.github/workflows/tutorials.yml
```diff
@@ -0,0 +1,105 @@
...
+    - name: Install compilers for OSX
+      if: runner.os == 'macOS'
+      run: |
+        if [ "${{ matrix.compiler }}" = "gcc-9" ]; then
+          brew install gcc
+        else
+          sudo xcode-select -s /Applications/Xcode_11.app/Contents/Developer
+        fi
+        # dask error on osx, skip dask tuto
+        echo "::set-env name=SKIP::--deselect=examples/seismic/tutorials/04_dask.ipynb"
```
FabioLuporini: I think this should be in the test file, not in the action. 

Could we mark it (pseudocode) `xfail(condition=OS is MacOS, reason=sympy+dask not behaving)`

I'd also say file an issue for this MacOS + sympy + dask bug, plus a reproducer

.github/workflows/pytest-core-nompi.yml
```diff
@@ -36,102 +35,104 @@ jobs:
...
+    - name: Install GCC ${{ matrix.arch }}
+      if: runner.os == 'linux'
+      run : |
+        sudo apt-get install -y ${{ matrix.arch }}
 
-    - name: Set up Python 3.7
-      uses: actions/setup-python@v1
-      with:
-        python-version: 3.7
+    - name: Reduce number of tests for OSX
```
FabioLuporini: Is it worth doing this for some of the other builds? Or would it be too much?

.github/workflows/pytest-core-nompi.yml
```diff
@@ -36,102 +35,104 @@ jobs:
...
 
     - name: Test with pytest
       run: |
-        pytest --cov --cov-config=.coveragerc --cov-report=xml tests/test_operator.py
+        $RUN_CMD pytest -m "not parallel" --cov --cov-config=.coveragerc --cov-report=xml $TESTS
 
     - name: Upload coverage to Codecov
-      uses: codecov/codecov-action@v1.0.5
+      if: matrix.name != 'docker-devito'
+      uses: codecov/codecov-action@v1
```
FabioLuporini: back to 1 from 1.0.5 ?

.github/workflows/pytest-core-nompi.yml
```diff
@@ -36,102 +35,104 @@ jobs:
...
+          arch: "gcc-9"
+          openmp: 1
+
+        - name: py37-osx
+          python-version: 3.7
+          os: macos-latest
+          arch: "osx"
+          openmp: 0
+
+        - name: docker-devito
```
FabioLuporini: Im not sure I like having this build here with the pip install ones. It pollutes the yml with lots of `ifs`. I think it be cleaner as a separate file, thus showing how to setup Docker step by step

.github/workflows/pytest-core-nompi.yml
```diff
@@ -36,102 +35,104 @@ jobs:
...
+          arch: "gcc-8"
+          openmp: 1
+
+        - name: py38-gcc9-openmp
+          python-version: 3.8
+          os: ubuntu-16.04
+          arch: "gcc-9"
+          openmp: 1
+
+        - name: py37-osx
```
FabioLuporini: it's a bit weird that we call this build `-osx` , so we suffix it with the OS name, while the other ones with the backend compiler `gcc8` `gcc9` etc

perhaps the others should be `py37-linux-gcc8-omp` (note: also renaming openmp as omp)

.github/workflows/tutorials.yml
mloubout: That's a slow tutorial to run bit overkill to run it with xfail, can only mark cells not the file with nbval

.github/workflows/pytest-core-nompi.yml
mloubout: Could be done to speed up yes but can't trim too much or not testing enough

.github/workflows/pytest-core-nompi.yml
mloubout: Doesn't seem to matter but can update

.github/workflows/pytest-core-nompi.yml
mloubout: Can make a renaming pass

.github/workflows/pytest-core-nompi.yml
mloubout: Then gonna complain there is repeated code, and I think a bash script is not adapted at all to CI.

FabioLuporini: > and I think a bash script is not adapted at all to CI.

You know I agree about this

So yesterday, as I was saying, I was perhaps too tired to realize two things:

- we're talking about how many repeated lines? those for running the tests, and then what?
- there are a few exceptions allowing for *some little* code duplication. Plots, for example, is one of those. I started thinking (this morning, with a fresh mind) that this may well be yet another of those exceptions. 

Thoughts @mloubout @ggorman @navjotk

.github/workflows/pytest-core-nompi.yml
FabioLuporini: OK

.github/workflows/tutorials.yml
FabioLuporini: So we are conflating two issues here:

- slowness
- macos + dask + sympy bug (which btw needs a reproducer)

Let's ignore slowness for a moment. Can we xfail individual cells? I doubt we can (based on what you just said). But if we could, I'd by far prefer the xfail. Otherwise, we'll have to leave it the way it is now.

.github/workflows/pytest-core-nompi.yml
mloubout: I don't think some duplication is an issue here. This is CI and the more simple and explicit each part of the testing is the easier it is to maintain and debug. Now this way there is less duplication and docker is just another venv  build so erroring/passing is still trackable easily and clearly

.github/workflows/tutorials.yml
mloubout: Can xfail a single cell not sure how nbval handles conditional xfail, gotta look if feasible

mloubout: So far, all iI found is `NBVAL_SKIP` but it's just a flag that doesn't take input so would skip it for all builds. Still digging but skipping the notebook for osx probably the simplest. The whole notebook fails anyway so skipping it isn't really an issue, but yes need to be said somewhere for users that there is an issue. I have no idea where to file that issue though

FabioLuporini: I think we need an MFE?

Remind me -- you cannot reproduce the behaviour on your mac right ?

mloubout: I can, fails too on my mac with same error, MFE sure don't really have time to look into it right now but will try.

mloubout:
Rebased on master, approved by @fablup. Will merge once green

.github/workflows/pytest-core-nompi.yml
mloubout: renamed

.github/workflows/pytest-core-nompi.yml
```diff
@@ -11,127 +11,128 @@ on:
...
-            - name: py37-gcc8-openmp
-              python-version: 3.7
-              gcc-version: 8
-              openmp: 1
-
-            - name: py38-gcc9-openmp
-              python-version: 3.8
-              gcc-version: 9
-              openmp: 1
+        - name: pytest-ubuntu-py36-gcc49-omp
```
georgebisbas: Should the tests be in ascending order of python version? Now it is gcc, but just saying...

.github/workflows/tutorials.yml
ggorman: From slack discussions, the concern is that when there is a failure within a bash script the source of the error is not clear from the Github-action logs. I feel there is still an issue with *not* using bash scripts as we might lose out on some code reuse. Perhaps there needs to be another way of composing actions to avoid this issue, or perhaps there is already a way we have overlooked as we are all Github novices. 

I am going to mark this thread as resolved for now but I expect the to-bash-or-not-to-bash issue is going to arise again and we should continue to reconsider the question as Github-actions matures.

PR #1083, opened 2020-01-14
Codecov 

mloubout:
WIP


mloubout:
@FabioLuporini should be mergeable now. Basically like numpy target to 1% coverage so that it passes but still part of CI, and removed patch (doesn't really add much). Should be around 90% most of the time unless upload failures

FabioLuporini:
merged

PR #1084, opened 2020-01-15
Fix issue 1081

### LINKED ISSUE #1081: aliasing expressions scheduled to different loops

FabioLuporini:
fixes #1081 

basically a performance issue due to too much Clusters fusion -- Clusters with ConditionalDimensions shouldn't be fused with so called dense Clusters, as they won't be optimized by the DSE

This PR fixes this

Issue was introduced by #1026 . Was fixing a thing and I ended up causing this. Anyway, now suitable tests added so we should be golden

### DIFF
--- a/tests/test_dimension.py
+++ b/tests/test_dimension.py
@@ -7,7 +7,7 @@ import pytest
 from conftest import skipif
 from devito import (ConditionalDimension, Grid, Function, TimeFunction, SparseFunction,  # noqa
                     Eq, Operator, Constant, Dimension, SubDimension, switchconfig)
-from devito.ir.iet import Iteration, FindNodes, retrieve_iteration_tree
+from devito.ir.iet import Expression, Iteration, FindNodes, retrieve_iteration_tree
 from devito.types import Array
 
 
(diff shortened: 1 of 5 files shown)


tests/test_dimension.py
```diff
@@ -819,3 +819,79 @@ def test_implicit_dims(self):
...
+        h = Function(name='h', grid=grid)
+
+        ctime = ConditionalDimension(name='ctime', parent=time, condition=time > 4)
+
+        eqns = [Eq(f.forward, f + 1),
+                Eq(h, f + 1),
+                Eq(g, f + 1, implicit_dims=[ctime]),
+                Eq(f.forward, f + 1, implicit_dims=[ctime]),
+                Eq(f.forward, f + 1),
+                Eq(g, f + 1)]
```
mloubout: Not sure I follow, there is duplicate equation on purpose?

FabioLuporini: All it matters is that the RHSs are identical

Semantically this test is stupid.

Practically it's useful to see that guarded and un-guarded Clusters are
1. processed separately
2. optimized as expected

mloubout: Oh ok, it's super confusing to read the equations :D

FabioLuporini: Think of them as statements. 

I agree the example is stupid, but at least it's concise

PR #1080, opened 2020-01-14
Fix issue 1060

FabioLuporini:
fixes #1060 

@georgebisbas can you check that this actually fixes the execution through benchmark.py as well?

PR #1085, opened 2020-01-15
Test DevitoBench

georgebisbas:
WIP-testing
Testing basic execution of cross product for modes and problems on DevitoBench.


requirements.txt
```diff
@@ -21,3 +21,4 @@ frozendict
 anytree>=2.4.3
 pyrevolve>=2.1.3
 distributed>=1.27
+git+https://github.com/devitocodes/devitobench
```
rhodrin: I don't think DevitoBench should be a default requirement.

georgebisbas: It will probably not be. At the moment is just WIP-testing.
Related discussion happened under here: 
https://devitocodes.slack.com/archives/C7JMLMSG0/p1579101476448300

mloubout:
Ok so I like how it's setup but I have two main concerns:
 - It is now just another test but quite costly so makes all the pytest a lot slower
 -  More than `isfile` would check if file is not empty maybe?

So maybe I would create a marker for `bench` or something so that can make a bench only test rather than add all of that to all the tests runs

requirements.txt
FabioLuporini: still WIP?

setup.py
```diff
@@ -9,28 +9,24 @@
     optionals = f.read().splitlines()
 
 reqs = []
-links = []
 for ir in required:
     if ir[0:3] == 'git':
-        links += [ir + '#egg=' + ir.split('/')[-1] + '-0']
-        reqs += [ir.split('/')[-1]]
+        name = ir.split('/')[-1]
+        reqs += ['%s @ %s@master' % (name, ir)]
```
FabioLuporini: @mloubout are all these changes sane?

tests/test_benchmark.py
```diff
@@ -0,0 +1,70 @@
...
+    Test combinations of modes x problems and created .json filename.
+    Testing for both OpenMP enabled or disabled.
+    """
+    if configuration['openmp']:
+        nthreads = configuration['platform'].cores_physical
+    else:
+        nthreads = 1
+
+    command_bench = ['python', 'benchmarks/user/benchmark.py', mode, '-P', problem,
+                     '-d', '40', '40', '40', '--tn', '20', '-x', '1']
```
FabioLuporini: How about making it even smaller/shorter:
```
-d 16 16 16 
--tn 4 
```

tests/test_benchmark.py
```diff
@@ -0,0 +1,70 @@
...
+    try:
+        check_call(command_bench)
+    except CalledProcessError as e:
+        raise CompilationError('Command "%s" return error status %d. '
```
FabioLuporini: why is this needed ?

tests/test_benchmark.py
```diff
@@ -0,0 +1,70 @@
...
+                               'Unable to compile code.\n' %
```
FabioLuporini: and most importantly, what makes you think that's due to an error with compilation?

tests/test_benchmark.py
```diff
@@ -0,0 +1,70 @@
...
+                               (e.cmd, e.returncode))
+    dir_name = 'results/'
+
+    base_filename = problem
```
FabioLuporini: **all** the underscores in the list of variables below should be dropped

then you just do:

```
bench_file = path.join('_'.join(....))
```

tests/test_benchmark.py
```diff
@@ -0,0 +1,70 @@
...
+    to = '_to[2]'
+    dse = '_dse[advanced]'
+    dle = '_dle[advanced]'
+    at = '_at[aggressive]'
+    nt = path.join('_nt[' + str(nthreads) + ']')
+    mpi = '_mpi[False]'
+    np = '_np[1]'
+    rank = '_rank[0]'
+
+    listdir("results/")
```
FabioLuporini: ?

tests/test_benchmark.py
```diff
@@ -0,0 +1,70 @@
...
+    bench_file = path.join(dir_name + base_filename + arch + shape + nbl + tn + so + to
+                           + dse + dle + at + nt + mpi + np + rank + filename_suffix)
+
+    command_plot = ['python', 'benchmarks/user/benchmark.py', 'plot', '-P', problem,
+                    '-d', '40', '40', '40', '--tn', '20', '--max-bw', '12.8',
+                    '--flop-ceil', '80', 'linpack']
+    try:
+        check_call(command_plot)
+    except CalledProcessError as e:
+        raise CompilationError('Command "%s" return error status %d. '
```
FabioLuporini: ?

tests/test_benchmark.py
```diff
@@ -0,0 +1,70 @@
...
+    mpi = '_mpi[False]'
+    np = '_np[1]'
+    rank = '_rank[0]'
+
+    listdir("results/")
+
+    bench_file = path.join(dir_name + base_filename + arch + shape + nbl + tn + so + to
+                           + dse + dle + at + nt + mpi + np + rank + filename_suffix)
+
+    command_plot = ['python', 'benchmarks/user/benchmark.py', 'plot', '-P', problem,
```
FabioLuporini: plotting should be a separate test, if we really want to test it

tests/test_benchmark.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+pytestmark = skipif(['yask', 'ops'])
+
+
+@pytest.mark.parametrize('mode', ['bench'])
+@pytest.mark.parametrize('problem', ['acoustic', 'tti', 'elastic', 'viscoelastic'])
+def test_bench(mode, problem):
+    """
+    Test combinations of modes x problems and created .json filename.
```
FabioLuporini: x -> and

Drop "and created .json filename" (also there was a typo)

tests/test_benchmark.py
```diff
@@ -0,0 +1,70 @@
...
+                    '-d', '40', '40', '40', '--tn', '20', '--max-bw', '12.8',
+                    '--flop-ceil', '80', 'linpack']
+    try:
+        check_call(command_plot)
+    except CalledProcessError as e:
+        raise CompilationError('Command "%s" return error status %d. '
+                               'Unable to compile code.\n' %
+                               (e.cmd, e.returncode))
+
+    bkend = '_bkend[core]'
```
FabioLuporini: drop underscore

tests/test_benchmark.py
```diff
@@ -0,0 +1,70 @@
...
+
+    plot_file = path.join(dir_name + base_filename + shape + so + to + arch + bkend + at)
```
FabioLuporini: same story -- `'_'.join(...)`

setup.py
mloubout: Yes, I made these newer `setuptool` standard, actually works better than before.

requirements.txt
georgebisbas: just a bit more

tests/test_benchmark.py
georgebisbas: Sure

tests/test_benchmark.py
georgebisbas: Sure

tests/test_benchmark.py
georgebisbas: They are just leftovers from a pytest template.

tests/test_benchmark.py
georgebisbas: To check created files locally, will be dropped.

tests/test_benchmark.py
georgebisbas: Can I ensure that two tests are excecuted sequentially? So that pytest doesn't execute plot before bench?

tests/test_benchmark.py
georgebisbas: Same, to be dropped.

tests/test_benchmark.py
georgebisbas: this is probably wrong? It creates this: `a_c_o_u_s_t_i_c_a_r_c_h_[_u_n_k_n_o_w_n_]_s_`

tests/test_benchmark.py
georgebisbas: Same yes

tests/test_benchmark.py
georgebisbas: ok, found the way

FabioLuporini:
Can we update tags and title of this PR? are we close to landing ?

requirements.txt
FabioLuporini: this should be an optional dependence

if you use the mode `run`, instead of `bench`, you shouldn't need `devitobench` at all. That's what I would do, I think

tests/test_benchmark.py
```diff
@@ -0,0 +1,61 @@
...
+    nt = path.join('nt[' + str(nthreads) + ']')
+    mpi = 'mpi[False]'
+    np = 'np[1]'
+    rank = 'rank[0]'
+    bkend = 'bkend[core]'
+
+    bench_corename = path.join('_'.join([base_filename, arch, shape, nbl, tn,
+                               so, to, dse, dle, at, nt, mpi, np, rank]))
+
+    bench_filename = path.join(dir_name + bench_corename + filename_suffix)
```
FabioLuporini: this should be:

```
path.join(dir_name, bench_corename, filename_suffix)
```

Otherwise it's totally useless to use `path.join`

FabioLuporini: ignore the comment above, I think? but why do you need path.join ?

tests/test_benchmark.py
FabioLuporini: fair point. I don't think you can, AFAICT after 1 min of googling

tests/test_benchmark.py
```diff
@@ -0,0 +1,61 @@
...
+    Testing for both OpenMP enabled or disabled.
+    """
+    if configuration['openmp']:
+        nthreads = configuration['platform'].cores_physical
+    else:
+        nthreads = 1
+
+    command_bench = ['python', 'benchmarks/user/benchmark.py', mode, '-P', problem,
+                     '-d', '16', '16', '16', '--tn', '4', '-x', '1']
+    check_call(command_bench)
```
FabioLuporini: did you verify that if you for example supply an unknown flag to the command , the test actually fails ?

For example:
```
['python', 'benchmarks/user/benchmark.py', mode, '-P', problem, ......, '-asdjoias', ...]
```

and, analogously, if there were an error in the execution of the benchmark, the test fails as well?

just to be sure...

tests/test_benchmark.py
```diff
@@ -0,0 +1,61 @@
...
+    arch = 'arch[unknown]'
+    shape = 'shape[16,16,16]'
+    nbl = 'nbl[10]'
+    tn = 'tn[4]'
+    so = 'so[2]'
+    to = 'to[2]'
+    dse = 'dse[advanced]'
+    dle = 'dle[advanced]'
+    at = 'at[aggressive]'
+    nt = path.join('nt[' + str(nthreads) + ']')
```
FabioLuporini: in devito, we do **not** concatenate strings with '+'. We rather use

```
"nt[%d]" % nthreads
```

tests/test_benchmark.py
```diff
@@ -0,0 +1,61 @@
...
+    bench_filename = path.join(dir_name + bench_corename + filename_suffix)
+
+    command_plot = ['python', 'benchmarks/user/benchmark.py', 'plot', '-P', problem,
+                    '-d', '16', '16', '16', '--tn', '4', '--max-bw', '12.8',
+                    '--flop-ceil', '80', 'linpack']
+    check_call(command_plot)
+
+    plot_corename = path.join('_'.join([base_filename, shape, so, to, arch,
+                              bkend, at]))
+    plot_filename = path.join(dir_name + plot_corename)
```
FabioLuporini: no string concatenation with +

tests/test_benchmark.py
```diff
@@ -0,0 +1,61 @@
...
+    Test combinations of modes and problems.
+    Testing for both OpenMP enabled or disabled.
+    """
+    if configuration['openmp']:
+        nthreads = configuration['platform'].cores_physical
+    else:
+        nthreads = 1
+
+    command_bench = ['python', 'benchmarks/user/benchmark.py', mode, '-P', problem,
+                     '-d', '16', '16', '16', '--tn', '4', '-x', '1']
```
FabioLuporini: out of curiosity, how many timesteps does '--tn 4' imply?

tests/test_benchmark.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+pytestmark = skipif(['yask', 'ops'])
+
+
+@pytest.mark.parametrize('mode', ['bench'])
+@pytest.mark.parametrize('problem', ['acoustic', 'tti', 'elastic', 'viscoelastic'])
+def test_bench(mode, problem):
+    """
+    Test combinations of modes and problems.
```
FabioLuporini: "Test the Devito benchmark framework on various combinations of modes and problems."

tests/test_benchmark.py
```diff
@@ -0,0 +1,61 @@
...
+    Testing for both OpenMP enabled or disabled.
```
FabioLuporini: Drop, I'd say irrelevant/obvious.

Also, it should have read "Test both ..."

tests/test_benchmark.py
```diff
@@ -0,0 +1,61 @@
...
+    command_plot = ['python', 'benchmarks/user/benchmark.py', 'plot', '-P', problem,
+                    '-d', '16', '16', '16', '--tn', '4', '--max-bw', '12.8',
+                    '--flop-ceil', '80', 'linpack']
+    check_call(command_plot)
+
+    plot_corename = path.join('_'.join([base_filename, shape, so, to, arch,
+                              bkend, at]))
+    plot_filename = path.join(dir_name + plot_corename)
+
+    assert path.isfile(bench_filename)
```
FabioLuporini: This assert must be moved somewhere around line 50

tests/test_benchmark.py
georgebisbas: Yes, it does fail, I have tested mispellings, wrong args/unkonwn flags, Well this how we realized the error that you fixed in https://github.com/devitocodes/devito/pull/1080

tests/test_benchmark.py
georgebisbas: Hell, sorry , that was very stupid.

tests/test_benchmark.py
georgebisbas: Well, bad leftovers...

tests/test_benchmark.py
georgebisbas: hmm... lets see.

tests/test_benchmark.py
georgebisbas: Moved to "%s%s"

tests/test_benchmark.py
georgebisbas: Moved, fixed.

georgebisbas:
> How long does this test take **overall** ?

90 secs sequentially locally

==================================================== slowest 20 test durations =====================================================
37.90s call     tests/test_benchmark.py::test_bench[tti-bench]
24.52s call     tests/test_benchmark.py::test_bench[viscoelastic-bench]
14.03s call     tests/test_benchmark.py::test_bench[elastic-bench]
8.01s call     tests/test_benchmark.py::test_bench[acoustic-bench]

mloubout:
>90 secs sequentially locally

@FabioLuporini @georgebisbas there is three of it in the 20 slowest in the log for ~2min so I'd say overall adds about 3min to each run, and that's with tn=4, so maybe better to have its own build (like osx only runs one)

mloubout:
Looks like something went wrong with rebase

georgebisbas:
> Looks like something went wrong with rebase

Where can you see this? As I can see it is rebased on latest master.

georgebisbas:
> > Looks like something went wrong with rebase
> 
> Where can you see this? As I can see it is rebased on latest master.

Sorry, I mean what is wrong?

requirements.txt
FabioLuporini: nonetheless, it should be an optional dependence, so be moved to a different file (@mloubout ?)

tests/test_benchmark.py
FabioLuporini: good

tests/test_benchmark.py
```diff
@@ -0,0 +1,60 @@
...
+
+    dir_name = 'results/'
+
+    base_filename = problem
+    filename_suffix = '.json'
+    arch = 'arch[unknown]'
+    shape = 'shape[16,16,16]'
+    nbl = 'nbl[10]'
+    tn = 'tn[4]'
+    so = 'so[2]'
```
FabioLuporini: @mloubout `so[2]` and TTI running? 😮

requirements.txt
mloubout: Still not sure what to do with these optionals requirements

tests/test_benchmark.py
mloubout: Yeah it runs, not recommended for the quality of solution but runs fine

FabioLuporini:
> @FabioLuporini @georgebisbas there is three of it in the 20 slowest in the log for ~2min so I'd say overall adds about 3min to each run, and that's with tn=4, so maybe better to have its own build (like osx only runs one)

but @mloubout , what was the runtime of 
https://github.com/devitocodes/devito/blob/master/azure-pipelines.py#L30

Perhaps what we should consider is the difference? ie how much slow*er* are we now ?

I don't know if I like the idea of moving it do a different action

mloubout:
> what was the runtime

pipeline is running all examples too, but takes about 25-30min

>I don't know if I like the idea of moving it do a different action

I'm not saying that I'm just saying to have one of the pytests build to do just this one.

requirements.txt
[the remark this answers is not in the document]
mloubout: But then need to install MPI because optionals have mpi4py

PR #1087, opened 2020-01-17
tweaks to make operators picklable without cloudpickle

tjb900:
This PR makes a few simple changes which allow `Operator`s (and other devito objects) to be pickled using regular pickling, rather than requiring `cloudpickle`.  This in turn lets it use the C implementation of pickling, which is both faster and (we believe) does not suffer from a Python bug we are running into.

Specifically, it just moves some inner functions and classes to module level, so that pickle can instantiate them without needing to store the state of the enclosing class or instance.  It also harmonises the names given to namedtuples with the name needed for the unpickler to instantiate them.  For `DataSide` this required a tiny bit of rejigging, but everything else is pretty simple.

Thoughts/suggestions welcome obviously.


devito/tools/abc.py
```diff
@@ -90,6 +90,10 @@ def _signature(self):
         return Signer._sign(self._signature_items())
 
 
+def wrapper(cls, args, kwargs):
```
FabioLuporini: can this be a private method inside Pickable ?

tjb900: Maybe a classmethod - I'll try.

tjb900:
We can move the wrapper back as a staticmethod by the looks of it.

Unfortunately, the namedtuple as an inner class isn't possible though:

```
# These two are very similar

class B:
    BB = namedtuple("BB", "field1")

class A:
    class AB:
        pass 
```

and in the second case, python assigns the fully qualified names correctly:

```
In [30]: A.AB.__qualname__                                                                                                                                                                                                               
Out[30]: 'A.AB'

In [40]: A.AB                                                                                                                                                                                                                            
Out[40]: __main__.A.AB
```

In the `namedtuple` case though, since the class is being manufactured by a function rather than defined the "normal way", we need to supply the type name for the new class (since it's an attribute of the class object that's being created) as an argument to the function.  That name is what the class thinks it is called - but in this case it is not actually a valid name in the namespace:

```
In [37]: B.BB.__qualname__                                                                                                                                                                                                               
Out[37]: 'BB'

In [38]: BB                                                                                                                                                                                                                              
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-38-9a4cea89a54a> in <module>
----> 1 BB

NameError: name 'BB' is not defined

and that's going to cause problems when pickling, because the type thinks it's called `mymodule.BB` but `mymodule.BB` doesn't exist.  

So: the name we give to `namedtuple` must be the name that we assign the type to.

Unfortunately (and one could argue that this is a missing feature of `namedtuple`) it's impossible to supply a fully qualified type name (e.g. `B.BB`) to the `namedtuple`, which means that "inner class" namedtuples cannot be pickled.

```
In [32]: namedtuple("X.Y", "field1 field2").__qualname__                                                                                                                                                                                 
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-32-36f455bf9341> in <module>
----> 1 namedtuple("X.Y", "field1 field2").__qualname__

/xxx/lib/python3.7/collections/__init__.py in namedtuple(typename, field_names, rename, defaults, module)
    359             raise TypeError('Type names and field names must be strings')
    360         if not name.isidentifier():
--> 361             raise ValueError('Type names and field names must be valid '
    362                              f'identifiers: {name!r}')
    363         if _iskeyword(name):

ValueError: Type names and field names must be valid identifiers: 'X.Y'
```

File: devito/tools/abc.py [hunk 2]
```diff
@@ -117,6 +117,10 @@ class Pickable(object):
     _pickle_kwargs = []
     """The keyword arguments that need to be passed to __new__ upon unpickling."""
 
+    @staticmethod
+    def wrapper(cls, args, kwargs):
```
FabioLuporini: this should be private (ie prefix with `_`), and perhaps be called `_pickle_wrapper`. Since `Pickable` is inherited by our first-class citizen classes (eg, `Function`), I think this is a fairly useful tweak

FabioLuporini:
@mloubout any idea why we have a hanging codecov action here ?

mloubout:
@FabioLuporini I think the codecov may not run on Forked PR because of the token, can just ignore for now I'll look into it. I'd say just rebase and merge this one

FabioLuporini:
@mloubout 👍

PR #1088, opened 2020-01-17
Subdomains tutorial

rhodrin:
Addition of a new tutorial on subdomains.


rhodrin:
> My 2 cents: While I believe that showing the data in cell-25 after all makes absolute sense, I am not sure whether we need to define all these classes before. We do not want a user to write something similar when needed to work with SubDomains right? Or it works like that? I am not familiar about how you model a problem with subds. Other than that, I think the pics in the end do not help understand something different happening in each subdomain.

Not sure if I'm misunderstanding you or not, but yes, users need to define such classes to use subdomains. The **only** two inbuilt/pre-defined subdomains currently in Devito are those discussed at the beginning of the notebook. And they are discussed so users don't get confused when they look at the subdomains attached to a grid and see two that they didn't create.

examples/userapi/subdomains.ipynb
```diff
@@ -0,0 +1,954 @@
...
+    "    def define(self, dimensions):\n",
+    "        d = dimensions\n",
+    "        return {d: ('middle', 1, 1) for d in dimensions}"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "First, note that in the above we have used the shortand `d = dimensions` so that `d` will be a tuple of N-dimensions and then `{d: ... for d in dimensions}` such that our mapper will be valid for N-dimensional grids. Next we note the inclusion of `('middle', 1, 1)`. For mappers of the form `('middle', N, M)`, the `SubDomain` spans a contiguous region of ``dimension_size - (N + M)`` points starting at ``N`` and finishing at ``dimension_size - M``.\n",
```
mloubout: - ``dimension_size - (N + M)`` points starting at ``N`` and finishing at ``dimension_size - M``

start at `N` and finishing at ``dimension_size - M`` is `dimension_size - M-N+1` not `dimension_size - M -N`

examples/userapi/subdomains.ipynb
```diff
@@ -0,0 +1,954 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from devito import Grid\n",
+    "shape = (10, 10, 10)\n",
+    "grid = Grid(shape=shape, extent=shape)"
```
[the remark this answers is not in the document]
rhodrin: Indeed, but spacing is not used in this example and 10's seemed a little nicer for the visualization. But I'm happy to chance if we'd generally prefer to introduce grids with a 'nice' spacing?

examples/userapi/subdomains.ipynb
rhodrin: Indeed, good catch.

examples/userapi/subdomains.ipynb
```diff
@@ -0,0 +1,954 @@
...
+    "    def define(self, dimensions):\n",
+    "        x, y, z = dimensions\n",
+    "        return {x: x, y: y, z: z}"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In the above snippet, we have defined a class `FullDomain` based on `SubDomain` with the method `define` overridden. This method takes as input a set of Dimensions and produces a mapper from which we can create a subdomain. It it through utilizing this mapper that various types of subdomains can be created. The mapper `{x: x, y: y, z: z}` is simply saying that we wish for the three dimensions our of subdomain be exactly the input dimensions `x, y, z`. Note that for a 2D domain we would need to remove one of the dimensions, i.e. `{x: x, y: y}`, or more generally for an N-dimensional domain one could write `{d: d for d in dimensions}`.\n",
```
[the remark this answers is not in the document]
rhodrin: I think overridden is correct here but I'm not sure?

examples/userapi/subdomains.ipynb
```diff
@@ -0,0 +1,954 @@
...
+    "vp[:,:,int(0.5*shape[2])+1:int(0.5*shape[2])+1+int(4/h)] = 1.6\n",
+    "vs[:,:,int(0.5*shape[2])+1:int(0.5*shape[2])+1+int(4/h)] = 0.4\n",
+    "rho[:,:,int(0.5*shape[2])+1:int(0.5*shape[2])+1+int(4/h)] = 1.3\n",
+    "\n",
+    "vp[:,:,int(0.5*shape[2])+1+int(4/h):] = 2.2\n",
+    "vs[:,:,int(0.5*shape[2])+1+int(4/h):] = 1.2\n",
+    "rho[:,:,int(0.5*shape[2])+1+int(4/h):] = 2.\n",
+    "\n",
+    "origin = (0, 0, 0)\n",
+    "spacing = (h, h, h)"
```
mloubout: `qp = np.zeros(shape)` isn't that for viscoelastic?
Isn't it "easier" to just use a preset?

examples/userapi/subdomains.ipynb
```diff
@@ -0,0 +1,954 @@
...
+    "# Source freq. in MHz (note that the source is defined below):\n",
+    "f0 = 0.12\n",
+    "\n",
+    "# Thorbecke's parameter notation\n",
+    "l = model.lam\n",
+    "mu = model.mu\n",
+    "ro = model.irho\n",
+    "\n",
+    "t0, tn = 0., 30.\n",
+    "dt = 0.9*model.critical_dt\n",
```
mloubout: ?

examples/userapi/subdomains.ipynb
```diff
@@ -0,0 +1,954 @@
...
+   "outputs": [],
+   "source": [
+    "u_v_u = Eq(v.forward, model.damp * (v + s*ro*div(tau)), subdomain = model.grid.subdomains['upper'])\n",
+    "u_t_u = Eq(tau.forward, model.damp * (tau + s * l * diag(div(v.forward))),\n",
+    "           subdomain = model.grid.subdomains['upper'])\n",
+    "\n",
+    "u_v_l = Eq(v.forward, model.damp * (v + s*ro*div(tau)), subdomain = model.grid.subdomains['lower'])\n",
+    "u_t_l = Eq(tau.forward, model.damp * (tau + s * l * diag(div(v.forward))\n",
+    "                                      + s * mu * (grad(v.forward) + grad(v.forward).T)),\n",
+    "           subdomain = model.grid.subdomains['lower'])"
```
mloubout: This isn't really acoustic water and elastic rock no? This does show nicely subdomain but it is still acoustic wave-equation with `mu=0` in the water so does the same as putting `0` in`mu.data`.

Also misleading that the intro talks about `05_staggered_acoustic.ipynb` which I thought would be the equation in the water

examples/userapi/subdomains.ipynb
rhodrin: > `qp = np.zeros(shape)` isn't that for viscoelastic?

Yes, remnants I'd forgotten to delete.
> Isn't it "easier" to just use a preset?

I wanted it to be clear to the user how the subdomains line up with the model (which is a bit hidden when using a preset). But I should tidy this up through setting some constants etc.

examples/userapi/subdomains.ipynb
rhodrin: Needed for numerical stability, meant to mention this in the meeting last week but forgot.

examples/userapi/subdomains.ipynb
rhodrin: Yes, I should clean this and the explanations up a bit (and remove mention of the staggered notebook).

As you mention, with the current setup the model being solved is the same as one not having subdomains since `mu` is zero in the water layer anyway. Originally I wanted to actually solve the acoustic equation as is done in the acoustic notebook and then form `tau` from `p` for the elastic eqn. But doing this 'nicely' wasn't really suitable for this tutorial (and should probably be in a different one?).

examples/userapi/subdomains.ipynb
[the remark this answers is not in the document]
mloubout: It's ok to keep it at 10 I think, it's just these `10/9, 50./49` and such can lead to inconsistent results (not by much just tiny differences) depending on the compiler, so I usually prefer "nic" spacing so odd number of grid points (shape).
Up to you, just sayin

examples/userapi/subdomains.ipynb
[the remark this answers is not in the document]
mloubout: No sure neither

examples/userapi/subdomains.ipynb
mloubout: Wait what is unstable and when? dt should be fine for any model.

examples/userapi/subdomains.ipynb
mloubout: Yes, it is clearer like that then adding interface conditions for the purpose of this tutorial, small text changes/cleanup nd should be gtg

examples/userapi/subdomains.ipynb
rhodrin: If you run with `dt = model.critical_dt` you'll see that the numerics blow up in the denser lower layer (iirc). Will investigate further.

examples/seismic/model.py
```diff
@@ -340,7 +340,7 @@ def critical_dt(self):
         #
         # The CFL condtion is then given by
         # dt < h / (sqrt(ndim) * max(vp)))
-        dt = .95*np.min(self.spacing) / (np.sqrt(3)*self.maxvp)
+        dt = .95*np.min(self.spacing) / (np.sqrt(3.)*self.maxvp)
```
mloubout: Looks liek this was changed by mistake at some point. Should be 
`.85*np.min(self.spacing) / (np.sqrt(self.dim)*self.maxvp)`

examples/userapi/subdomains.ipynb
mloubout: see new comment

examples/seismic/model.py
rhodrin: Ah cool. Will fix. Ta.

examples/userapi/subdomains.ipynb
```diff
@@ -0,0 +1,966 @@
...
+   "metadata": {},
+   "source": [
+    "# Subdomains tutorial"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This tutorial is designed to introduce users to the concept of subomains in Devito and how to utilize them within simulations for a variety of purposes.\n",
```
FabioLuporini: typo: subdomains

this whole file perhaps needs to go through a spell checker?

examples/userapi/subdomains.ipynb
```diff
@@ -0,0 +1,966 @@
...
+    "\n",
+    "We will begin by exploring the subdomains created internally (and by default) when creating a Devito `Grid` and then explore in detail how users can define and utilize their own subdomains.\n",
```
FabioLuporini: I'd drop "Devito" from "Devito Grid"

examples/userapi/subdomains.ipynb
```diff
@@ -0,0 +1,966 @@
...
+   "outputs": [],
+   "source": [
+    "full_domain = FullDomain()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "It is important to note that whilst the `full_domain` object has now been initialised, it has not yet been finalised, i.e. it is not currently associated with any grid. To attach this subdomain to a grid we must pass it to the grid during its creation (the sudomain will then be finalised during creation of the grid):"
```
FabioLuporini: I wonder whether it should "subdomain" or "`SubDomain`". Same for "dimension" vs "`Dimension`" (below) and "grid" vs "`Grid`"

examples/userapi/subdomains.ipynb
```diff
@@ -0,0 +1,966 @@
...
+    "Now, let us create a subdomain consisting of all but the outer grid point in each dimension (similar to 'interior'):"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 14,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "class Innerdomain(SubDomain):\n",
```
FabioLuporini: nitpicking: `InnerDomain` for homogeneity

examples/userapi/subdomains.ipynb
```diff
@@ -0,0 +1,966 @@
...
+    "        return {d: ('middle', 1, 1) for d in dimensions}"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "First, note that in the above we have used the shortand `d = dimensions` so that `d` will be a tuple of N-dimensions and then `{d: ... for d in dimensions}` such that our mapper will be valid for N-dimensional grids. Next we note the inclusion of `('middle', 1, 1)`. For mappers of the form `('middle', N, M)`, the `SubDomain` spans a contiguous region of ``dimension_size - (N + M)`` points starting at (in terms of python indexing) ``N`` and finishing at ``dimension_size - M - 1``.\n",
+    "\n",
+    "The two other options available are `'left'` and `'right'`. For a statement of the form ``d: ('left', N)`` the SubDomain spans a contiguous region of ``N`` points starting at ``d``\\'s left extreme. A statement of the form ``('right', N)`` is analogous to the previous case but starting at the dimensions right extreme instead.\n",
```
FabioLuporini: apparently only one `\` is needed?

examples/userapi/subdomains.ipynb
FabioLuporini: This applies to the whole notebook probably.

examples/userapi/subdomains.ipynb
```diff
@@ -0,0 +1,966 @@
...
+   ],
+   "source": [
+    "f.data"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Seismic wave propagation example with subdomains\n",
```
FabioLuporini: I wonder whether this should be moved to a different notebook somewhere under `examples/seismic` ? clearly stating then that `userapi/subdomains.ipynb` is a prerequisite 

Thoughts?

examples/userapi/subdomains.ipynb
rhodrin: Ta, fixed.

examples/userapi/subdomains.ipynb
rhodrin: Dropped.

examples/userapi/subdomains.ipynb
rhodrin: In most places we're referring to the 'concepts' more than the Devito objects and so I'd prefer to keeps as is. I've changed in a couple of places when it's most appropriate to refer to the object though.

examples/userapi/subdomains.ipynb
rhodrin: Changed.

examples/userapi/subdomains.ipynb
rhodrin: Removed the superfluous `\`'s.

examples/userapi/subdomains.ipynb
rhodrin: I prefer keeping it here since after going through the basics I think it's nice to directly have a more 'practical' example.

FabioLuporini:
Awesome, merged.

Issue #859, opened 2019-06-18
mpi notebook and ipyparallel race conditions


FabioLuporini:
this applies to ``examples/mpi/overview.ipynb`` and possibly to any ipyparallel-based notebook tested via ``nbval``

random failures seem to be happening due to [this issue](https://github.com/ipython/ipyparallel/issues/294) in ipyparallel

we're not totally sure this is the actual issue, but it's more than a weak conjecture 

currently worked around by adding NBVAL_IGNORE to all non-output cells. This seems to make CI happy for now...

FabioLuporini:
horrible, but we might consider using [flaky](https://docs.pytest.org/en/latest/flaky.html) ... better than nothing?

integration with NBVAL is still missing though

PR #1094, opened 2020-01-24
Actions for examples + goodbye pipelines

mloubout:
WIP

Moved leftover pipeline to actions:

- mpi tuto int pytest-mpi
- examples into example action

Example action uses conda, test already use some of examples so no need for extra pip.


.github/workflows/pytest-core-mpi.yml
```diff
@@ -45,6 +45,12 @@ jobs:
       run: |
         pytest --cov --cov-config=.coveragerc --cov-report=xml -m parallel tests/
 
+    - name: Test mpi notebook
+      run : |
+        ipcluster start --profile=mpi -n 4 --daemon
+        py.test --nbval examples/mpi
```
FabioLuporini: I suspect this is gonna fail. See #859

.github/workflows/examples.yml
```diff
@@ -0,0 +1,75 @@
...
+       environment-file: environment.yml
+       python-version: 3.7
+       auto-activate-base: false
+
+    - name: Install dependencies
+      # See setup-miniconda README, necessary to activate environment
+      shell: bash -l {0}
+      run: |
+        pip install -e .
+        pip install matplotlib
```
FabioLuporini: actually necessary?

.github/workflows/examples.yml
```diff
@@ -0,0 +1,75 @@
...
+        python examples/seismic/tti/tti_example.py -a basic
+        python examples/seismic/tti/tti_example.py -a basic --noazimuth
+
+    - name: Seismic elastic examples
+      shell: bash -l {0}
+      run: |
+        py.test examples/seismic/elastic/elastic_example.py
+        py.test examples/seismic/viscoelastic/viscoelastic_example.py
+
+    - name: Linear algebra example
```
FabioLuporini: nitpicking: I'd use "exampleS" (there are quite a few in linalg.py ; also, we may add more in the future)

FabioLuporini: maybe more than "linear algebra"...how about "linear and non-linear operators" (in perspective of extending this? I'm thinking for example about adding convolutions and some more exotic examples/operators)

.github/workflows/examples.yml
mloubout: yes, matplotlib is in optionals but needed by the cfd test/example

.github/workflows/examples.yml
mloubout: >adding convolutions 

That's still linear algebra ;)
Only one example file so can add an s hen there is more or rename it but though was accurate for what's currently in it

mloubout:
>they should probably stay in until #1085 is in ?

Yes that's the plan

.github/workflows/examples.yml
```diff
@@ -0,0 +1,75 @@
...
+       auto-activate-base: false
+
+    - name: Install dependencies
+      # See setup-miniconda README, necessary to activate environment
+      shell: bash -l {0}
+      run: |
+        pip install -e .
+        pip install matplotlib
+
+    - name: Seismic acoustic examples
```
FabioLuporini: Question, for another day...technically speaking, wouldn't it be more correct to call this "isotropic" or "iso-acoustic" ?

.github/workflows/examples.yml
```diff
@@ -0,0 +1,75 @@
...
+    - name: Seismic elastic examples
+      shell: bash -l {0}
+      run: |
+        py.test examples/seismic/elastic/elastic_example.py
+        py.test examples/seismic/viscoelastic/viscoelastic_example.py
+
+    - name: Linear algebra example
+      shell: bash -l {0}
+      run: |
+        python examples/misc/linalg.py mat-vec mat-mat-sum transpose-mat-vec
```
FabioLuporini: why explicit list ?

.github/workflows/examples.yml
mloubout: isotropic-acoustic is more accurate

.github/workflows/examples.yml
mloubout: That's how it was in pipelines

mloubout:
@fablup this is ready to go and doesn't really touch anything so can ust stay there open until #1085 is done I don't mind

FabioLuporini:
@mloubout 

>  this is ready to go

agreed

> so can ust stay there open until #1085 is done I don't mind

yes, let's give it a few more days (@georgebisbas ), it shouldn't take much

environment.yml
```diff
@@ -3,31 +3,6 @@ channels:
...
-  - sphinx
-  - sphinx_rtd_theme
-  - distributed>=1.27
-  - cython
-  - click
-  - codecov
-  - pytest-cov
-  - multidict
-  - frozendict
-  - nb_conda_kernels
```
FabioLuporini: this doesn't appear in requirements.txt ?

the following also are missing (not sure why):
- cython
- sphinx_rtd_theme ( I think necessary for rendering of documentation?)
- sphinx
- parso>=0.1.0 ??

mloubout: Gonna assume old unused dependency because everything works on all os/env and platforms

PR #1103, opened 2020-01-28
[WIP - RFC] make generator scripts for Azure pools/jobs

georgebisbas:
Towards automating our experiment suite for MPI benchmarking on Azure.
- Needs: Better parametrization on arguments
- Needs more testing
- Needs batch-shipyard bridging

### DIFF
--- a/benchmarks/user/make-azure-jobs.py
+++ b/benchmarks/user/make-azure-jobs.py
@@ -1,0 +1,80 @@
+import os
+
+import click
+
+from benchmark import option_simulation
+import devito
+
+
+@click.group()
+def menu():
+    pass
+
+
+@menu.command(name='generate')
+@option_simulation
+@click.option('-id', default='bench', help='Job ID')
+@click.option('-image', default='mloubout/devito-bench:v0.1', help='Devito docker image')
+@click.option('-arch', default='gcc-8', help='Devito architecture')
+@click.option('-nt', default='4', help='OpenMP number of threads')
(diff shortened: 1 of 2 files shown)


mloubout:
Will look at it, but should this be really here rather than Azure-devito?

benchmarks/user/make-azure-jobs.py
```diff
@@ -0,0 +1,80 @@
...
+      OMP_PROC_BIND: 'close'
+      DEVITO_MPI: 1
+      DEVITO_OPENMP: 1
+      OMP_PLACES: 'cores'
+      DEVITO_LOGGING: 'DEBUG'
+      LC_ALL: 'C.UTF-8'
+      LANG: 'C.UTF-8'
+    default_working_dir: container
+    multi_instance:
+      num_instances: pool_current_dedicated
```
mloubout: this can also be fixed to a number, say you start a pool with 8 nodes you can decide to use only half or something
```
multi_instance:
       num_instances: pool_current_dedicated/2
      mpi:
         runtime: mpich
         processes_per_node: 1
         options:
           - --bind-to socket
```

For example will use half the nodes with one worker per node

benchmarks/user/make-pool.py
```diff
@@ -0,0 +1,70 @@
+import os
```
mloubout: duplicate file?

FabioLuporini:
shouldn't these belong to azure-devito? mm

mloubout:
Opened in Azure-devito, closing

PR #1107, opened 2020-01-30
utils: use numpy to filter uniques without key

mloubout:


devito/tools/utils.py
```diff
@@ -156,10 +157,31 @@ def filter_ordered(elements, key=None):
...
+        try:
+            unordered, inds = np.unique(elements, return_index=True)
+            return unordered[np.argsort(inds)].tolist()
+        except:
+            return sorted(list(set(elements)), key=elements.index)
     return [e for e in elements if not (key(e) in seen or seen.add(key(e)))]
 
 
+def filter_ordered_index(elements, key=None):
+    """Filter elements in a list while preserving order.
```
georgebisbas: nitpicking: docstring under quotes

devito/tools/utils.py
```diff
@@ -156,10 +157,31 @@ def filter_ordered(elements, key=None):
...
+    if key is None:
+        _, inds = np.unique(elements, return_index=True)
+        inds.sort()
+        return inds.tolist()
+    lst_e = elements.tolist()
+    return [lst_e.index(e) for e in lst_e if not (key(e) in seen or seen.add(key(e)))]
+
+
 def filter_sorted(elements, key=None):
     """Filter elements in a list and sort them by key. The default key is
```
georgebisbas: same

tests/test_tools.py
```diff
@@ -1,8 +1,10 @@
 from sympy.abc import a, b, c, d, e
 import pytest
+import numpy as np
```
georgebisbas: needs alphabetical order

tests/test_tools.py
```diff
@@ -1,8 +1,10 @@
...
+import time
 
 from conftest import skipif
-from devito.tools import toposort
+from devito.tools import toposort, filter_ordered, filter_ordered_index
```
georgebisbas: needs one line space

devito/tools/utils.py
```diff
@@ -156,10 +157,31 @@ def filter_ordered(elements, key=None):
     """
     seen = set()
     if key is None:
-        key = lambda x: x
+        try:
+            unordered, inds = np.unique(elements, return_index=True)
+            return unordered[np.argsort(inds)].tolist()
+        except:
+            return sorted(list(set(elements)), key=elements.index)
```
FabioLuporini: no, this is wrong. `sorted` may change the input ordering. This would only be OK for `filter_sorted`

devito/tools/utils.py
FabioLuporini: the docstring is identical to that of `filter_ordered`

Im not sure I appreciate the difference between this function and `filter_ordered`, on a semantic point of view?

devito/finite_differences/coefficients.py
```diff
@@ -244,7 +244,7 @@ def generate_subs(deriv_order, function, dim):
     # Determine which 'rules' are missing
     sym = get_sym(functions)
     terms = obj.find(sym)
-    args_present = filter_ordered(term.args[1:] for term in terms)
+    args_present = filter_ordered([term.args[1:] for term in terms])
```
FabioLuporini: I think we should just use `as_tuple` inside `filter_ordered` so that it's fine to pass in generators

devito/tools/utils.py
mloubout: No this is correct because of  `key=elements.index`, that's the only reason sorted is called to revert `set`  to original ordering

FabioLuporini: right. I think I'm an idiot. Ignore me. 

How slower/faster this fallback is?

FabioLuporini: ah, but there's one thing you wanna do: `except ...` Missing the exception type

mloubout: >except ... Missing the exception type
Kinda left it out on purpose because it depends on what type of input there is. Can be sorting errors if non-sortables (relational error), or type errors or ...
Basically it try numpy if you can, if not go back to list defaults, can put a generic one though

mloubout: >How slower/faster this fallback is?

See test, ~ 2 times better on smallish, for the index part get around 10k better for very large shot recs

devito/finite_differences/coefficients.py
mloubout: Probably yeah, on it

devito/tools/utils.py
```diff
@@ -147,22 +149,51 @@ def single_or(l):
...
     return [e for e in elements if not (key(e) in seen or seen.add(key(e)))]
 
 
+def filter_ordered_index(elements, key=None):
```
FabioLuporini: I think this is too highly specialised to exist and be considered a "tool".

Since in the only place where it's currently used we have `key=None`, can you not just

* drop this function
* use a one liner [here](https://github.com/devitocodes/devito/pull/1107/files#diff-5d662e9e9c311233ce7b83c1d18d291aR168)
```
np.unique(elements, return_index=True)[1].sort().tolist()
```

or if you really want it to be a tool, I'd at least call it
```
filter_indices
```
to draw a neat difference w.r.t. `filter_ordered`

mloubout: >to draw a neat difference w.r.t. filter_ordered

Sure. I prefer to keep it as a tool though.

>np.unique(elements, return_index=True)[1].sort().tolist()

`sort` and `np.argsort` are not the same from what I remember.

FabioLuporini: the one liner is just syntactic sugar... I think I put on a single line exactly what you had split over multiple lines? 

I'll check again

mloubout: Nevermind, was thinking about something else, moved there, just needs two lines becasue `sort` returns nothing

PR #1109, opened 2020-01-31
Sparse MPI redundant computation fix

mloubout:
This basically makes `_dist_datamap` and `_support` memoized method based on the gridpoints (coordinates).

This reduce the call to `_dist_datamap` to 1 instead of 8 per operator call per sparse function
For example:
```
Allocating memory for rec(1355, 159600)
Allocating memory for rec(1355, 159600)
Allocating memory for rec(1355, 159600)
Allocating memory for rec(1355, 159601)
```

`args_defaults` (currently) takes 36 minutes on this size (4 MPI ranks), in testing with this
update

### DIFF
--- a/devito/types/sparse.py
+++ b/devito/types/sparse.py
@@ -111,12 +111,8 @@ class AbstractSparseFunction(DiscreteFunction, Differentiable):
         The grid points surrounding each sparse point within the radius of self's
         injection/interpolation operators.
         """
-        return self._build_support(points=self.gridpoints)
-
-    @memoized_meth
-    def _build_support(self, points=None):
         ret = []
-        points = points or self.gridpoints
+        points = self.gridpoints
         for i in points:
             support = [range(max(0, j - self._radius + 1), min(M, j + self._radius + 1))
                        for j, M in zip(i, self.grid.shape)]
@@ -597,7 +593,7 @@ class SparseFunction(AbstractSparseFunction):
             raise ValueError("No coordinates attached to this SparseFunction")
         ret = []
         for coords in self.coordinates.data._local:
-            ret.append(tuple(int(sympy.floor((c - o.data)/i.spacing.data)) for c, o, i in
+            ret.append(tuple(int(np.floor(c - o.data)/i.spacing.data) for c, o, i in
                              zip(coords, self.grid.origin, self.grid.dimensions)))
         return tuple(ret)
 
--- a/devito/types/sparse.py
+++ b/devito/types/sparse.py
@@ -112,8 +112,7 @@ class AbstractSparseFunction(DiscreteFunction, Differentiable):
         injection/interpolation operators.
         """
         ret = []
-        points = self.gridpoints
-        for i in points:
+        for i in self.gridpoints:
             support = [range(max(0, j - self._radius + 1), min(M, j + self._radius + 1))
                        for j, M in zip(i, self.grid.shape)]
             ret.append(tuple(product(*support)))
--- a/devito/types/sparse.py
+++ b/devito/types/sparse.py
@@ -111,8 +111,13 @@ class AbstractSparseFunction(DiscreteFunction, Differentiable):
         The grid points surrounding each sparse point within the radius of self's
         injection/interpolation operators.
         """
+        return self._build_support(points=self.gridpoints)
+
+    @memoized_meth
+    def _build_support(self, points=None):
         ret = []
-        for i in self.gridpoints:
+        points = points or self.gridpoints
+        for i in points:
             support = [range(max(0, j - self._radius + 1), min(M, j + self._radius + 1))
                        for j, M in zip(i, self.grid.shape)]
             ret.append(tuple(product(*support)))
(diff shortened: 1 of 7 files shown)


devito/types/sparse.py
```diff
@@ -587,7 +593,7 @@ def gridpoints(self):
         for coords in self.coordinates.data._local:
             ret.append(tuple(int(sympy.floor((c - o.data)/i.spacing.data)) for c, o, i in
                              zip(coords, self.grid.origin, self.grid.dimensions)))
-        return ret
+        return tuple(ret)
```
mloubout: NOTE: necessary to return a tuple instead of list for hashing in memoizization (how do you spell that!!!)

devito/types/sparse.py
```diff
@@ -107,24 +107,30 @@ def inject(self, *args, **kwargs):
 
     @property
     def _support(self):
-        """
-        The grid points surrounding each sparse point within the radius of self's
-        injection/interpolation operators.
-        """
+        return self._build_support(points=self.gridpoints)
```
georgebisbas: docstring deleted?

mloubout: Indeed didn't notice, putting it back

georgebisbas: Is it back? (just a reminder)

mloubout:
Ok so overall yes, I agree, there needs to be a rework of Sparse in general but I think all these 4 options are just different way to do this temporary patch. 

>I have a feeling we can't accept the solution in this PR as if you have 1M receivers as you were saying, you may leak a lot of memory

Well the coordinates are quite cheap still, even the case of couple M of receivers, it's only a couple Mb of memory so nothing that is gonna crash anything. The data is massive memorywise not the coordinates.

# Options 1

This is overall the best long term option but needs a lot of rethinking and refactoring as the coordinates data is two levels down SparseFunction inside its Subfunction's data. This would need to setup setters/getters to know when and where something changes and probably rethink Subfunction.

So probably the best option long-term but way ahead of this PR.

# Option 2

This one may not be doable as the expensive part isn't really the loop but the `global_to_rank` inside it for each coordinate.
Support itself doesn't need to be memoized as it's quite quick, I only memoized it too for homogeneity, but `dist_map` is the costly one.
May make things a bit better to do it, but overall `dist_map` is called 8 times at prepare args and 8 times at post_process args so 16 times overall when nothing changes.


# Options 3


I think link to what I said for Option 1


# Option 4

Not sure how much we'd gain from the current case


To summarize, yes sparse needs a lot of work, this is a temporary patch that I think is fairly cheap (couple Mb is something we can leave with), while actually fixing something that for large number of rec (even couple 100k) takes (a lot) more than 50% of the total runtime for a 3D TTI modeling that is already quite expensive, and up to over 80% of the time for acoustic, up to hours of pre/post processing of sparse arguments.

devito/types/sparse.py
mloubout: yes it is

FabioLuporini:
@mloubout OK... I think I can live with this

however, could you just tell me which line exactly is responsible for slowing everything down?

is it https://github.com/devitocodes/devito/pull/1109/files#diff-5d662e9e9c311233ce7b83c1d18d291aR121 

?

devito/types/sparse.py
```diff
@@ -111,20 +111,30 @@ def _support(self):
...
+    def _build_dist_datamap(self, support=None):
         """
         Mapper ``M : MPI rank -> required sparse data``.
         """
         ret = {}
-        for i, s in enumerate(self._support):
+        support = support or self._support
+        for i, s in enumerate(support):
             # Sparse point `i` is "required" by the following ranks
             for r in self.grid.distributor.glb_to_rank(s):
```
mloubout: @FabioLuporini that's the bad line overall, `_support` is fast and can be unmemoized if too overkill

FabioLuporini: yeah, should only memoize the killer

also...is the actual culprit somewhere around here?
https://github.com/devitocodes/devito/blob/master/devito/mpi/distributed.py#L258

mloubout: Not sure, wasn't able to find exactly where it slows down.

devito/types/sparse.py
```diff
@@ -585,9 +591,9 @@ def gridpoints(self):
             raise ValueError("No coordinates attached to this SparseFunction")
         ret = []
         for coords in self.coordinates.data._local:
-            ret.append(tuple(int(sympy.floor((c - o.data)/i.spacing.data)) for c, o, i in
+            ret.append(tuple(int(np.floor(c - o.data)/i.spacing.data) for c, o, i in
```
mloubout: @fablup this `sympy.floor` was super bad actually

FabioLuporini: OK. Worth trying it out again w/o memoization?

mloubout: Did, still need it for `dist_map` but removed the mempization on `_support` that is now way faster (most time was spent here)

devito/types/sparse.py
```diff
@@ -112,19 +112,25 @@ def _support(self):
         injection/interpolation operators.
         """
         ret = []
-        for i in self.gridpoints:
+        points = self.gridpoints
```
FabioLuporini: doesn't need to be a separate line?

devito/types/sparse.py
FabioLuporini: question: after changing from sympy.floor to np.floor, what is the next hotspot ?

mloubout: `glb_to_rank`  is where most of the time is spent now

examples/seismic/utils.py
```diff
@@ -1,8 +1,49 @@
...
+
+    rec_coordinates = rec_coords(model)
+
+    geometry = AcquisitionGeometry(model, rec_coordinates, src_coordinates,
+                                   t0=0.0, tn=tn, src_type='Ricker', f0=0.010)
+
+    return geometry
+
+
+def rec_coords(model):
```
FabioLuporini: setup_rec_coords ?

devito/types/sparse.py

examples/seismic/utils.py
mloubout: fixed

PR #1110, opened 2020-02-03
ci: Add summary workflow

FabioLuporini:
Basically so that we can display a unique "build passing" badge in README.md

But generalisable to other actions that we only want to run when all other workflows have succeeded


mloubout:
I don't particularly like that. We have a normal and mpi install (optionals requirements) we should keep these two as separate badges.

mloubout:
Actually the more I think about it the less I like it.

Currently, we have two explicit and clear badges:
- Core , that is green if pass and red if fails and the badge links to CI-core. So if you click on it you end up on the  CI_core build and see the list of builds with fail/pass
- MPI, same but for CI-mpi

Now with this one, we hide this and now have a badge that will always be green whether or not the actual CI build pass or fail and if someone clicks on it it sends to a "we merged a PR" workflow, which basically sends the message "we made up a badge to look good on README but don't actually run CI" to any newcomer that just wanna check it. The badge is not for us, we know where the tests are and how to check if they pass.

I think this maker a lot more harm than good and shoudl be closed. I still don't get what's wrong with the current badges that perfectly reflect that we have a core devito version and an MPI devito version that needs (and should be added to README at some point) its own install with extra requirements.

FabioLuporini:
I totally disagree with pretty much everything

There are no multiple versions. It's just Devito, and everything that works sequentially should/must work in parallel over MPI. It doesn't matter if a requirement (MPI) is installed or not. The key thing to notice here is that this badge becomes green *iff* all other workflows have succeded.

So this summary badge will incorporate documentation, flake, ... everything. Currently, we may have a green test badge with some tutorials failing, and this won't be reflected in our front page

about the redirection link: I think we can tweak it to point to the main actions page.

So I really don't understand your reservations about this

Maybe the issue is that you see it as "two different versions". But they are not

mloubout:
>So this summary badge will incorporate documentation,

It doesn't, that yml file doens't run anything and the badhe will always be green no matter what

>There are no multiple versions. 

Then `requirement-optionals` should be gone and there is only devito,  not extras or  different setup. 

> Currently, we may have a green test badge with some tutorials failing

Because you didn't want a badge for these. 


What is wrong with having explicit badge for specific parts, I really don't get why it "better" to just put everything under a badge that will be green even if absolutely every build fails.

FabioLuporini:
you're wrong, wonder if you actually checked the diff. Maybe my bad I didn't explain it.

https://github.com/devitocodes/devito/pull/1110/files#diff-50febde5cf60d667233f85c9658d5b35R8
+
https://github.com/devitocodes/devito/pull/1110/files#diff-50febde5cf60d667233f85c9658d5b35R15

this workflow fires up iff all we merge into master, which in turn can only happen if all of the other workflows have turned green

mloubout:
>this workflow fires up iff all we merge into master, which in turn can only happen if all of the other workflows have turned green

Yeah, because master build never failed after merge....

FabioLuporini:
> Yeah, because master build never failed after merge....

typically only because travis and pipelines sucked

I think this is vastly better even though the price to pay were to accept a wrong badge in the 0.01 % of the time which may happen over a solar year

I actually find it a very nice way to say "hey, look, **all** is green"

FabioLuporini:
also, if we have a failure upon a push to master:

* either it's not a bug -- in which case it's great to have a green badge
* or it's a bug, and yeah, highest priority to fix it

mloubout:
> typically only because travis and pipelines sucked

Well no, most of the time was because sympy/numpy/conda. or some dependency changed between the  last PR comment and the merge, there is multiple reason the master build could fail and the badge should reflect it

>I actually find it a very nice way to say "hey, look, all is green"

Except that's not what it reflects, that's just a badge that always says "it's green" because it's made to always be, not because the CI actually passes.

mloubout:
>also, if we have a failure upon a push to master: ...

That doesn't justify having a fake badge.

FabioLuporini:
> That doesn't justify having a fake badge.

"fake" is really excessive. It is based on the outcome of all other workflows, so it's all but fake

> Well no, most of the time was because sympy/numpy/conda. or some dependency changed between the last PR comment and the merge, there is multiple reason the master build could fail and the badge should reflect it

and how is that different from what we've had in place since always? a dependence may change thus creating an unexpected failure. The badges will still display green until we merge a PR fixing the issue.

> Except that's not what it reflects, that's just a badge that always says "it's green" because it's made to always be, not because the CI actually passes.

because we don't accept pushes to master until a PR is definitely green. So it should never be anything other than green.

I agree with you about the dependences-changing issue, but unless I'm missing the obvious, this is exactly what we have and always had

Anyway, I can survive without this PR or at least until you're 100% happy, so not a big issue for me to close it until I have a "better" alternative (though I doubt there's one) :D

FabioLuporini:
closing as useless

PR #1111, opened 2020-02-04
Back to smoother advanced DSE/DLE levels

georgebisbas:
Smoother advanced DSE/DLE levels


devito/builtins.py
```diff
@@ -343,7 +343,7 @@ def buff(i, j):
         options = None
 
     # TODO: Figure out why yask doesn't like it with dse/dle
-    assign(lhs, rhs, options=options, name=name, dse='noop', dle='noop')
+    assign(lhs, rhs, options=options, name=name, dse='advanced', dle='advanced')
```
FabioLuporini: no, you should just drop them, because the defaults are advanced already

also, the comment "TODO: ..." above should be dropped

Issue #1065, opened 2020-01-09
Streamline dependency management


tallamjr:
With `enviroment.yml` some `pip` and `conda` dependencies are defined, but there also exists a separate `requirements.txt`. This is effectively redundant as all dependencies can be contained within the `enviroment.yml` file

navjotk:
Firstly, `environment.yml` defines the development environment. It is the packages you need to develop Devito itself. 

`requirements.txt` is the user environment. The minimum packages you need to be able to install and use Devito.

sgibson91:
Follow up question: Why does the quickstart tutorial instruct user to install the development environment?

navjotk:
@sgibson91 raised an interesting question as to why do we direct people to the conda/ `environment.yml` installation route in the Quickstart even though this is the developer environment and not the user environment.

mloubout:
These are two different objects.

 `environment.yml` is a conda specific file to sept a virtual environment with dependencies including non-pip ones such as conda defaults, python versions, ...

`requirement.txt` is the list of *pip* requirements. One could write a wrapper that reads it and feed it to the `environment.yml` but it is not recommended as some of the pip packages are installed via conda rather than pip in the conda environment.

SO there may be a better way to do it, but not a straightforward one.

navjotk:
Ok let's rephrase the above question. What do we lose if we delete `requirements*.txt` ?

mloubout:
>What do we lose if we delete requirements*.txt ?

No pip install

navjotk:
pip install uses `intall_requires` from `setup.py` anyway?

mloubout:
>pip install uses intall_requires from setup.py anyway?

These are from `requirement.txt`, hard coding dependencies in `setup.py` is doable but better suited for small dependency list, a separate requirement file is more maintainable 

>Or do you mean pip install -r requirements.txt

no I mean `pip install git....`

navjotk:
To summarise, we have the following installation routes, which use the associated requirements lists:
- `pip install git....` uses `requirements.txt`
- `pip install -e .`/`pip install devito` uses `setup.py/install_requires`
- conda -f environment.yml` uses `environment.yml`

Now, onto a best practices discussion with @sgibson91 :)

navjotk:
What do we lose we if we remove `environment.yml`?

mloubout:
The conda environment basically.

navjotk:
What do we need that for? i.e. why don't we just drop it?

[...]

PR #1116, opened 2020-02-10
Update install instructions

navjotk:
Addresses https://github.com/devitocodes/devito/issues/1065#issuecomment-573106105

Following up from the linked issue, assuming @mloubout did his part about removing redundancies between `requirements.txt` and `environment.yml` already, this PR does the rest of the things from that issue including:
- Add three separate subsections each for "docker installation"/"pip installation" and "conda installation", each explaining their respective target audience
- Rename `environment.yml` to `environment-dev.yml` to highlight that this is the development environment. 
- Make the quickstart section the one where the user decides which of the three use cases from above they fall into, and link to the appropriate section. 

Hope to end these discussions once and for all with this PR.


README.md
```diff
@@ -19,41 +19,100 @@ are facing issues or just trying it out, join the
...
-The recommended way to install Devito uses the Conda package manager
-for installation of the necessary software dependencies. Please
-install either [Anaconda](https://www.continuum.io/downloads) or
-[Miniconda](https://conda.io/miniconda.html) using the instructions
-provided at the download links. You will need the Python 3 version.
+For the least friction method, see the [Docker](#docker-installation) instructions.
 
-To install Devito, including examples, tests and tutorial notebooks,
-follow these simple passes:
+If you are familiar with python and pip, just run `pip install devito`. Also look
```
mloubout: Maybe keep github url for now as there isn't a release on PyPi yet.

FabioLuporini:
README.md is the first thing that newcomers see, and I don't think we want to put too much explanation about the installation process. Essentially, just the code snippets and a couple of sentences should suffice.

In the same spirit as https://github.com/tensorflow/tensorflow#install , we should/could move the full explanation to https://github.com/devitocodes/devito/blob/master/docs/source/download.rst , which anyway needs to be updated with this PR

BTW, I will refresh README.md before Rice

ggorman:
@FabioLuporini - your comments appear to have been addressed. GTO?

PR #1117, opened 2020-02-11
Add boundary conditions tutorial to userapi

EdCaunt:
First version of a boundary conditions tutorial. Shows how to build a domain with both PMLs and free surface boundary conditions.

Still to do: 
- Add PMLs to bottom of domain (just treat corners as sides for simplicity) <- Done
- Make free surface boundary a true free surface (currently just a zero Dirichlet) <- Done
- Add a figure to illustrate where the boundaries are along with their types <- Done


EdCaunt:
Also fixes an inconsistency in numbering with the new SymPy tutorial. Changed 00-sympy.ipynb to 00_sympy.ipynb so it matches the other tutorials

examples/userapi/04_boundary_conditions.ipynb
```diff
@@ -0,0 +1,471 @@
...
+    "shape = (101, 101)\n",
+    "extent = (2000., 2000.)\n",
+    "PMLS = 10  # Number of PMLs on left and right sides"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Additionally, we will need to specify the dimensions for our `Grid`."
```
FabioLuporini: is this actually needed ? why isn't 
```
x, y = grid.dimensions
```
enough ?

examples/userapi/04_boundary_conditions.ipynb
```diff
@@ -0,0 +1,471 @@
...
+    "                                              value=extent[0]/(shape[0]-1)))\n",
+    "y = SpaceDimension(name='y', spacing=Constant(name='h_y',\n",
+    "                                              value=extent[1]/(shape[1]-1)))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We will need to use subdomains to accomodate the modified equations in the PML regions. As `Grid` objects are immutable, we must define our subdomains beforehand."
```
FabioLuporini: I wouldn't talk about "immutability", this will only confuse people. Just describe the API. In fact, you can probably just drop the sentence "As Grid objects are immutable, ... beforehand."

examples/userapi/04_boundary_conditions.ipynb
```diff
@@ -0,0 +1,471 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "shape = (101, 101)\n",
+    "extent = (2000., 2000.)\n",
+    "PMLS = 10  # Number of PMLs on left and right sides"
```
FabioLuporini: I think for homogeneity with other notebooks, this should be called ``nbpml`` (@rhodrin @mloubout please correct me if I'm wrong)

examples/userapi/04_boundary_conditions.ipynb
```diff
@@ -0,0 +1,471 @@
...
+   "source": [
+    "density = 1.  # 1000kg/m^3\n",
+    "velocity = 4.  # km/s"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We also need a `TimeFunction` object for each of our wavefields. Note that the pressure field is split in this case: a necessity for implementing the PML boundary conditions. As particle velocity is a vector, we will choose a `VectorTimeFunction` object to encapsulate it."
```
FabioLuporini: Why can it not be a VectorTimeFunction ? you should still be able to access the components `ux` and `uy`

FabioLuporini: EDIT: OK, I read the stuff below now. I'm a bit confused I think

examples/userapi/04_boundary_conditions.ipynb
```diff
@@ -0,0 +1,471 @@
...
+   ],
+   "source": [
+    "print(v[0].data)  # Print the data attribute associated with the x component of v"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "You may have also noticed the keyword `staggered` in the arguments when we created these functions. As one might expect, these are used for specifying where derivatives should be evaluated relative to the grid, as required for implementing formulations such as the first-order acoustic wave equation or P-SV elastic. Passing a function `staggered=NODE` specifies that its derivatives should be evaluated at the node. One can also pass `staggered=x` or `staggered=y` to stagger the grid by half a spacing in those respective directions. Additionally, a tuple of dimensions can be passed to stagger in multiple directions (e.g. `staggered=(x, y)`). `VectorTimeFunction` objects have their associated grids staggered by default.\n",
```
FabioLuporini: when you talk about ``staggered``, add a forward link to the staggered grids notebooks?

examples/userapi/04_boundary_conditions.ipynb
mloubout: First order acoustic is `p` TimeFunction (pressure) and `v` VectorTimeFunction (particle velocity), these two ux, uy are a bit odd. Even if you need the two components, it should be a VectorTimeFunction I think.

examples/userapi/04_boundary_conditions.ipynb
```diff
@@ -0,0 +1,471 @@
...
+     },
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "#NBVAL_SKIP\n",
+    "import matplotlib.pyplot as plt\n",
+    "\n",
+    "fig = plt.figure()\n",
+    "plt.imshow(np.swapaxes(ux.data[1] + uy.data[1], 0, 1),\n",
```
mloubout: (ux.data[1] + uy.data[1]).T

examples/userapi/04_boundary_conditions.ipynb
EdCaunt: Where are they?

examples/userapi/04_boundary_conditions.ipynb
EdCaunt: Fixed this in my most recent commit. The field splitting was a crutch in the paper to get out of Fourier space, but it never detailed how they got rid of it, so I left it in in my original version

examples/userapi/04_boundary_conditions.ipynb
EdCaunt: I made this more euphemistic, but left it in in spirit. "Immutibility" isn't mentioned anymore, but I've kept a bit about how subdomains come before grid, as the order is a tad unintuitive for new users otherwise

examples/userapi/04_boundary_conditions.ipynb
```diff
@@ -0,0 +1,455 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from devito import TimeFunction, VectorTimeFunction, NODE\n",
+    "\n",
+    "u = TimeFunction(name='u', grid=grid, time_order=1,\n",
```
mloubout: would call it `p` to match the tutorial

examples/userapi/04_boundary_conditions.ipynb
```diff
@@ -0,0 +1,455 @@
...
+    "                     subdomain=grid.subdomains['left'])\n",
+    "\n",
+    "# Right side\n",
+    "eq_v_damp_right = Eq(v.forward,\n",
+    "                     (1-damp_r)*v - dt*grad(u)/density,\n",
+    "                     subdomain=grid.subdomains['right'])\n",
+    "\n",
+    "eq_u_damp_right = Eq(u.forward,\n",
+    "                    (1-damp_r)*u - dt*velocity**2*density*div(v.forward),\n",
+    "                     subdomain=grid.subdomains['right'])"
```
mloubout: Ok I am a bit confused, are these actual PML. PML usually require auxiliary fields to implement pseudo-convolution over time, this mostly looks like the absorbing layer implemented in the seismic examples but as a sub-domain rather than a global dampening array that is zero inside the physical domain.

examples/userapi/04_boundary_conditions.ipynb
```diff
@@ -0,0 +1,455 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Add our free surface boundary conditions (currently only Drichlet...):"
```
mloubout: That is not a free surface, this is a zero Dirichlet boundary, so would not call it like that. Free surface requires to mirror `vz`

examples/userapi/04_boundary_conditions.ipynb
EdCaunt: The doi is at the top of the notebook. Not sure if the authors have just mislabeled it or if I've made some mistake in the implementation

examples/userapi/04_boundary_conditions.ipynb
EdCaunt: Yeah, this is on my to do list. I figured I'd get it in a working state for review, but I haven't got around to finishing this bit yet

examples/userapi/04_boundary_conditions.ipynb
```diff
@@ -0,0 +1,530 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "time = grid.stepping_dim\n",
+    "\n",
+    "bc_p = [Eq(p[time+1, x, 3], 0.)]  # Set pressure field to zero at boundary\n",
+    "\n",
+    "bc_v = [Eq(v[1][time+1, x, 2], v[1][time+1, x, 4])]  # Mirror y velocity across boundary\n",
+    "bc_v += [Eq(v[1][time+1, x, 1], v[1][time+1, x, 5])]\n",
+    "bc_v += [Eq(v[1][time+1, x, 0], v[1][time+1, x, 6])]"
```
mloubout: THis is basically hard-coded for space_order=6. Would be good to use a SubDim/SubDomain to make it generic and be able to switch space order easily.

see for example https://github.com/devitocodes/devito/blob/8fb68519d477e12df14960264980c3ffb3b526f9/examples/seismic/boundaries.py#L57

examples/userapi/04_boundary_conditions.ipynb
```diff
@@ -0,0 +1,530 @@
...
+   "metadata": {},
+   "source": [
+    "# Boundary conditions tutorial"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This tutorial aims to demonstrate how users can implement various boundary conditions in Devito, building on concepts introduced in previous tutorials. Over the course of this notebook we will go over the implementation of both free surface boundary conditions and perfectly-matched layers (PMLs) in the context of the first-order acoustic wave equation. This tutorial is based on a simplified version of the method outlined in Liu and Tao's 1997 paper (https://doi.org/10.1121/1.419657).\n",
```
mloubout: simplified version of PML isn't really PML. I know I'm beeing picky but saying Devito can do PML and having reflections as strong as the result at the end does not send a great message. People expect near perfect absorption with PML not damping layer quality.

EdCaunt: Could you send me an example of a non-simplified PML? As far as I can tell, my implementation is equivalent to any that I've seen in the literature (excluding CPMLs), and also near-identical to one that an industry guy showed me at OGHPC. The main difference is I'm pretty sure the tapers used are somewhat sub-optimal (because few papers actually explicitly state what tapers they used), which will affect the reflections.

mloubout: The paper below is simplified PML, but here is a simplified version of simplified PML, like most of the f1/f2/f3/f4/f5 from the equations are missing. So yes, most are simplifications in some ways of CPML, and I don't say I have an issue with the implementation or the tutorial, just would have worded it in a more cautious way .

EdCaunt: Should they not just fall out if you print the update equation? f1/2/3/etc are just assorted bits of finite difference stencil right?

mloubout: Inside the domain yes, but in the layer it depends on the omega's so it doesn't.

EdCaunt: Iirc, the omega_etas inside f1/2/3/etc are just badly named damping parameters and don't have anything to do with the frequency

mloubout: Yes they are, but they are still part of the stencil inside the layer, while only f1 is there currently

examples/userapi/04_boundary_conditions.ipynb
mloubout: Also, this is supposed to be opposite sign 
 [Eq(v[1][time+1, x, 1], -v[1][time+1, x, 5])]

mloubout:
Merging, only GPU issue as expected.

PR #1124, opened 2020-02-17
Extend GPU support via openmp offloading

FabioLuporini:


devito/passes/iet/definitions.py
```diff
@@ -131,37 +146,7 @@ def place_definitions(self, iet, **kwargs):
...
-        for decl, alloc, free in storage._on_high_bw_mem:
-            if decl is None:
-                header.append(alloc)
-            else:
-                header.extend([decl, alloc])
-            footer.append(free)
-        if header or footer:
-            body = List(header=header, body=iet.body, footer=footer)
-            iet = iet._rebuild(body=body)
+        iet = self._dump_storage(iet, storage)
```
FabioLuporini: note for reviewers: I simply moved this to a method so that it can be called by subclasses when needed

devito/builtins.py
```diff
@@ -431,8 +435,13 @@ def sumall(f):
...
     p, eqns = f.guard() if f.is_SparseFunction else (f, [])
 
+    s = dv.types.Scalar(name='sum', dtype=f.dtype)
+
     with MPIReduction(f) as mr:
-        op = dv.Operator(eqns + [dv.Inc(mr.n[0], p)], name='sum')
+        op = dv.Operator([dv.Eq(s, 0.0)] +
+                         eqns +
+                         [dv.Inc(s, p), dv.Eq(mr.n[0], s)],
+                         name='sum')
```
mloubout: These two bultins update look very hacky.

FabioLuporini: No it's fine, no devito issues here. it's just to work around what I think to be an openmp 5 (clang) bug

```
#pragma omp ... reduction(+:s)
```

works like a charm

```
#pragma omp ... reduction(+:n[0])
```

does not (ie, with arrays), ie it produces garbage. It works fine without offloading, so...

mloubout: But what do you need `mr` for then? You already have the result in `s` then copy it in `mr.n[0]` then output it, that's a temporary and extra eq that makes it look odd for no specific reason.

devito/core/__init__.py
```diff
@@ -32,8 +33,9 @@
 operator_registry.add(PowerOperator, Power, 'advanced')
 operator_registry.add(CustomOperator, Power, 'custom')
 
-operator_registry.add(CPU64NoopOperator, Device, 'noop')
-operator_registry.add(DeviceOffloadingOperator, Device, 'advanced')
+operator_registry.add(DeviceOpenMPNoopOperator, Device, 'noop')
+operator_registry.add(DeviceOpenMPOperator, Device, 'advanced')
+operator_registry.add(DeviceOpenMPCustomOperator, Device, 'custom')
```
mloubout: Maybe add GPU somewhere in the name?

devito/core/gpu_openmp.py
```diff
@@ -0,0 +1,336 @@
...
+from devito.passes.iet import (DataManager, Storage, Ompizer, ParallelIteration,
+                               ParallelTree, optimize_halospots, mpiize, hoist_prodders,
+                               iet_pass)
+from devito.tools import as_tuple, filter_sorted, generator, timed_pass
+
+__all__ = ['DeviceOpenMPNoopOperator', 'DeviceOpenMPOperator',
+           'DeviceOpenMPCustomOperator']
+
+
+class DeviceParallelIteration(ParallelIteration):
```
mloubout: Overall needs docstring throughout, but understand wasn't/isn't a priority.

devito/core/gpu_openmp.py
```diff
@@ -0,0 +1,336 @@
...
+        # input to Operator
+        kwargs['options'].pop('openmp')
+
+        return super(DeviceOpenMPNoopOperator, cls)._build(*args, **kwargs)
+
+    @classmethod
+    @timed_pass(name='specializing.Clusters')
+    def _specialize_clusters(cls, clusters, **kwargs):
+        # TODO: this is currently identical to CPU64NoopOperator._specialize_clusters,
+        # but it will have to change
```
mloubout: still the case i assume?

devito/ir/clusters/analysis.py
```diff
@@ -131,11 +131,15 @@ def _callback(self, clusters, d, prefix):
...
                 continue
 
-            if not dep.is_increment:
-                return SEQUENTIAL
+            if dep.is_increment:
+                if dep.function in scope.initialized:
+                    # False alarm, the increment is over a locally-defined symbol
+                    pass
+                else:
+                    is_parallel_atomic = True
```
mloubout: `is_parallel_atomic = is_parallel_atomic or True` or only last `dep` required.

devito/ir/support/basic.py
```diff
@@ -588,9 +588,6 @@ def is_indep(self, dim=None):
                 return False
             elif dim is None:
                 return self.distance == 0
-            elif self.source.is_local and self.sink.is_local:
```
mloubout: fixed somewhere else?

devito/builtins.py
FabioLuporini: `mr.[0]` has always been there

it's what you need to propagate a scalar back to the user

you cannot return `s` because it's not passed by reference

tests/test_operator.py
```diff
@@ -103,6 +103,21 @@ def test_multiple_steppers(self, expr, exp_uindices, exp_mods):
         assert(i.indices[i.function._time_position].modulo == exp_mods[i.function.name]
                for i in flatten(retrieve_indexed(i) for i in exprs))
 
+    @skipif('device')
```
mloubout: so what's the issue with these few skipped ones?

devito/ir/clusters/analysis.py
FabioLuporini: ? 

no, I think the current way is correct: as soon as the first `dep` inducing an increment is encountered, then we are inside a parallel-atomic dimension.

IOW, you may have 1000 deps, and 999 of them could be fully parallel, but if only one of them requires an atomic increment, then the whole dimension is parallel-atomic

FabioLuporini: it clearly doesn't matter whether it's the first or last dep

FabioLuporini: I could actually add an optimization that we don't waste time computing `test00` and `test01` and `test1` if we know already that `is_parallel_atomic is True`, because at that point the only worse thing that can happen is that the dimension turns out to be `SEQUENTIAL`

devito/ir/support/basic.py
FabioLuporini: not sure when, possibly months ago through the various refinements to data dependence analysis. This `elif` was probably added a loong time ago when the analysis wasn't as sophisticated as it is now. So I realised that it now was actually causing false positives. So I tried to drop it and everything (all tests) were still working

FabioLuporini: and this is a good thing, especially here in this file -- the lighter these routines are, the better, and not just because of speed, but also because despite they are fairly neat, the complexity they hide is quite scary

tests/test_operator.py
FabioLuporini: oh, nothing relevant. Just that we are `asserting` against a certain IET structure, which clearly is different if we generate omp-offloading code

devito/builtins.py
mloubout: Hum ok, I get your point. Looks like a quite "unnatural" way to do it, so some user may run into issues trying to add a norm into their own Operator. Food for thoughts

devito/core/__init__.py
mloubout: OmpGpu instead of OpenMP? It's same length but maybe bit more explicit?

FabioLuporini: argh, I totally missed this one, sorry. It's green again now, so I'll merge it now and then tweak the name in the next PR. Sorry

PR #1126, opened 2020-02-17
[WIP] improve tensor 

mloubout:


mloubout:
Need work and too nusy with SEG abstract and such, will reopen when ready

PR #1129, opened 2020-02-18
Revamp README.md

FabioLuporini:


README.md
```diff
@@ -1,23 +1,70 @@
-# Devito: Fast Finite Difference Computation from Symbolic Specification
+# Devito: Fast Tensor Computation from Symbolic Specification
```
georgebisbas: Do you really want to completely drop FD?

README.md
```diff
@@ -1,23 +1,70 @@
...
+- [Installation](#installation)
+- [Resources](#resources)
+- [Performance](#performance)
+- [Get in touch](#get-in-touch)
+
+## About Devito
+
+Devito provides a symbolic language to implement matrix-free operators, that is
+operators that do not require the explicit creation of a dense or sparse
+matrix. A typical use case is explicit finite difference (FD) methods for
```
mloubout: Matrix free implies Matrix, and we do not implement linear algebra at a high level. While the operator implements (most of the time) linear operation w.r.t to some of the input, we cannot do `c = A*b` where A, b and c are Devito objects. 
Devito provides a symbolic language for stencil computation, JUDI is a matrix-free operator language.

README.md
```diff
@@ -1,23 +1,70 @@
...
-## Quickstart
+There is virtually no limit to the type of matrix-free operators that one can
+implement with Devito. The Devito compiler will automatically analyze the input,
+detect and apply optimizations (including single- and multi-node parallelism),
+and eventually generate code with arbitrarily complex loop nests, as required
+by the symbolic specification -- clearly, all this is hidden away to the user.
+
+Key features include:
+
+* A tensor language to express FD operators.
```
mloubout: IMO the language is still a functional symbolic language, the compiler is a tensorial compiler.

README.md
```diff
@@ -1,23 +1,70 @@
...
+* Straightforward mechanisms to dynamically adjust the discretizion.
+* Constructs to express sparse operators (e.g., interpolation), classic linear
+  operators (e.g., convolutions), and tensor contractions.
+* Seamless support for adjoint operators.
```
mloubout: Well... if you can implement it by hand.

README.md
mloubout: IF we drop FD I would say stencil, not Tensor.

FabioLuporini: > Do you really want to completely drop FD?

No, it's mentioned all over the place

> IF we drop FD I would say stencil, not Tensor.

I don't mind stencil, but I'd want the others' opinion on this

README.md
FabioLuporini: > JUDI is a matrix-free operator language.

Indeed, Devito is a system "_to implement_ matrix-free operators" -- and indeed JUDI lives on top of Devito or, IOW, Devito is used to implement (forgive my CS reminiscences) the runtime support of JUDI

> we cannot do c = A*b where A, b and c are Devito objects.

are you saying we can't do a matrix-vector product in Devito ? or, more generally, any arbitrary tensor contraction?

Clearly, I agree with you that JUDI is higher level in this regards. So don't get me wrong, I think you're right when you say what I wrote above may benefit from some smoothing

README.md
FabioLuporini: Isn't a `Function` a tensor (since it's a discrete function)? `div` `curl` and all other operators make it a functional language, sure

README.md
FabioLuporini: yeah, I'm not saying automated anywhere 

"seamless" refers to the fact that ... "it's just yet another operator"

README.md
mloubout: >are you saying we can't do a matrix-vector product in Devito ? 

You can, but that doesn't mean it is a matrix-free operator.

 A matrix-free operator *is* a matrix, without the memory, but that behaves in any way the exact same way as a matrix.  And behave as one is the key point because this is what most software/algorithms expect as an input, a matrix.

> Devito is used to implement (forgive my CS reminiscences) the runtime support of JUDI

yes, it is, but the runtime support is not what makes it matrix-free, could have used pure julia or C or anything, what makes it matrix free is the abstraction that makes the operator look and behave like a matrix.

So Devito "can be used" to implement matrix-free frameworks, but does not implement one and I would not mention it at all.

README.md
mloubout: Its data is a tensor, but from a user/interface perspective, it is a pure functional object, you can do any sympy operation (integrals, polynomial, ...) on it. So yes on one side the compiler sees it as a tensor (nd-array) but the language used to define the object is still functional.

README.md
FabioLuporini: I totally agree with all you have said, which is sort of implicit for me. I think I now understand why what I wrote seems so confusing. I'll fix it properly

README.md
FabioLuporini: OK, using "stencil" for now

README.md
FabioLuporini: Now completely reworked

README.md
FabioLuporini: OK, I think I see the confusion. You're right, functional is the right thing to use here

README.md
```diff
@@ -1,23 +1,71 @@
...
 [![Build Status for the Core backend](https://github.com/devitocodes/devito/workflows/CI-core/badge.svg)](https://github.com/devitocodes/devito/actions?query=workflow%3ACI-core)
 [![Build Status with MPI](https://github.com/devitocodes/devito/workflows/CI-mpi/badge.svg)](https://github.com/devitocodes/devito/actions?query=workflow%3ACI-mpi)
 [![Code Coverage](https://codecov.io/gh/devitocodes/devito/branch/master/graph/badge.svg)](https://codecov.io/gh/devitocodes/devito)
+[![Slack Status](https://img.shields.io/badge/chat-on%20slack-%234A154B)](https://opesci-slackin.now.sh)
+
+[Devito](http://www.devitoproject.org) is a Python package to implement
+optimised stencil computation from high-level symbolic problem definitions.
+Devito builds on [SymPy](http://www.sympy.org/en/index.html) and employs
+automated code generation and just-in-time (JIT) compilation to execute
+optimized computational kernels on multiple computer platforms, including
```
georgebisbas: computational kernels **for a variety of** (?) computer platforms

multiple is targeted more on quantity than variety I think.

README.md
```diff
@@ -1,23 +1,71 @@
...
+- [About Devito](#about-devito)
+- [Installation](#installation)
+- [Resources](#resources)
+- [Performance](#performance)
+- [Get in touch](#get-in-touch)
+
+## About Devito
+
+Devito provides a functional language to implement sophisticated operators that
+can be made up of multiple stencil computations, boundary conditions, sparse
```
georgebisbas: may be consisting of ()?

README.md
```diff
@@ -1,23 +1,71 @@
...
-[Devito](http://www.devitoproject.org) is a software to
-implement optimised finite difference (FD) computation from
-high-level symbolic problem definitions. Starting from symbolic
-equations defined in [SymPy](http://www.sympy.org/en/index.html),
-Devito employs automated code generation and just-in-time (JIT)
-compilation to execute FD kernels on multiple computer platforms.
-
-## Get in touch
+An `Operator` generates low-level code from an ordered collection of `Eq` (just
+one in this simple example). This code may also be compiled and executed
```
georgebisbas: *in the above simple example* (?)

README.md
```diff
@@ -1,23 +1,71 @@
...
 
-If you're using Devito, we would like to hear from you. Whether you
-are facing issues or just trying it out, join the
-[conversation](https://opesci-slackin.now.sh).
+```python
+>>> op(t=timesteps)
+```
 
-## Quickstart
+There is virtually no limit to the complexity of an `Operator` -- the Devito
```
georgebisbas: virtually? (WDYM)
Maybe conceptually/theoretically?

README.md
```diff
@@ -1,23 +1,71 @@
...
+compiler will automatically analyze the input, detect and apply optimizations
+(including single- and multi-node parallelism), and eventually generate code
+with suitable loops and expressions.
+
+Key features include:
+
+* A functional language to express finite difference operators.
+* Straightforward mechanisms to dynamically adjust the discretizion.
```
georgebisbas: discretization?

README.md
```diff
@@ -1,23 +1,71 @@
...
+* Constructs to express sparse operators (e.g., interpolation), classic linear
+  operators (e.g., convolutions), and tensor contractions.
+* Seamless support for boundary conditions and adjoint operators.
+* A flexible API to define custom stencils, sub-domains, sub-sampling,
+  staggered grids.
```
georgebisbas: either `and staggered grids` or `staggered grids etc...`

README.md
```diff
@@ -1,23 +1,71 @@
...
+
+* A functional language to express finite difference operators.
+* Straightforward mechanisms to dynamically adjust the discretization.
+* Constructs to express sparse operators (e.g., interpolation), classic linear
+  operators (e.g., convolutions), and tensor contractions.
+* Seamless support for boundary conditions and adjoint operators.
+* A flexible API to define custom stencils, sub-domains, sub-sampling,
+  and staggered grids.
+* Generation of highly optimized parallel code (SIMD vectorization, CPU and GPU
+  parallelism via OpenMP, multi-node parallelism via MPI, blocking, aggressive
```
FabioLuporini: @georgebisbas in the end, I rewrote this to explicitly add OpenMP as you suggested.

georgebisbas: Good.

README.md
```diff
@@ -1,23 +1,71 @@
-# Devito: Fast Finite Difference Computation from Symbolic Specification
+# Devito: Fast Stencil Computation from Symbolic Specification
 
 [![Build Status for the Core backend](https://github.com/devitocodes/devito/workflows/CI-core/badge.svg)](https://github.com/devitocodes/devito/actions?query=workflow%3ACI-core)
 [![Build Status with MPI](https://github.com/devitocodes/devito/workflows/CI-mpi/badge.svg)](https://github.com/devitocodes/devito/actions?query=workflow%3ACI-mpi)
 [![Code Coverage](https://codecov.io/gh/devitocodes/devito/branch/master/graph/badge.svg)](https://codecov.io/gh/devitocodes/devito)
+[![Slack Status](https://img.shields.io/badge/chat-on%20slack-%2336C5F0)](https://opesci-slackin.now.sh)
+
+[Devito](http://www.devitoproject.org) is a Python package to implement
```
rhodrin: >Devito is a Python package to implement optimized stencil computation from high-level symbolic problem definitions.

Some people (e.g. certain applied mathematicians) are not going to understand exactly what you mean by 'stencil computation'. They'd understand finite difference though - don't you think we should mention FD in some form in this opening paragraph also?

README.md
```diff
@@ -1,23 +1,71 @@
...
-## Quickstart
+There is virtually no limit to the complexity of an `Operator` -- the Devito
+compiler will automatically analyze the input, detect and apply optimizations
+(including single- and multi-node parallelism), and eventually generate code
+with suitable loops and expressions.
+
+Key features include:
+
+* A functional language to express finite difference operators.
+* Straightforward mechanisms to dynamically adjust the discretization.
```
rhodrin: What exactly are you referring to here by 'dynamically' adjust the discretization?

README.md
FabioLuporini: yes. The reason I switched to stencil was for the exact opposite reason -- to avoid scaring away the non applied mathematicians.

So "stencil" is the superclass. How about adding:

_Devito is a Python package to implement optimized stencil computation (e.g., finite differences, image processing, machine learning) from high-level symbolic problem definitions._

? or something along the lines. Ideas welcome, and yes, I agree with you we should work in this direction

README.md
FabioLuporini: fair point, I think "dynamically" can be dropped as it doesn't add much. In fact, it may just create confusion.

README.md
rhodrin: Yes, the above sentence looks good to me.

README.md
rhodrin: > In fact, it may just create confusion.

Indeed. I think dropping it is fine.

FabioLuporini: done

README.md
FabioLuporini: done

FabioLuporini:
Merged.

PR #1131, opened 2020-02-19
Add GPU CI

rhodrin:
This PR adds testing for GPU's through deploying an Azure GPU VM (from an image) and running GPU tests on the VM via ssh.

Two additional  GPU tests are also added: 1) A simple operator test. 2) Isotropic acoustic test.

Some things to think about before landing this PR:
- ~~The PSA folder should be renamed and moved. I discussed this with Fabio previously but can't remember our intended solution at this moment: will dig it out and modify appropriately soon.~~
~~- How do we want to deploy VMs? i.e. one per branch or?~~
- The deployed VM is not currently being de-allocated. De-allocating it and redeploying it for every push of course doesn't make sense, but we also probably want add something somewhere to stop the VM if it's idle for a certain amount of time. Will look into a solution for that soon.
~~- Probably forgot to mention something else? (but will add if/when it comes back to me).~~


tests/test_gpu.py
```diff
@@ -0,0 +1,116 @@
...
+
+        time_steps = 1000
+        op.apply(time_M=time_steps)
+
+        assert np.all(np.array(u.data[0, :, :, :]) == time_steps)
+
+    @switchconfig(platform='nvidiaX')
+    def test_iso_ac(self):
+        from examples.seismic import TimeAxis, RickerSource, Receiver
+        from devito import Function, solve, norm
```
FabioLuporini: to be moved to the top level?

PSA/deployVM.ps1
```diff
@@ -0,0 +1,93 @@
+# Adapted from https://github.com/weeyin83/vm-actions
```
FabioLuporini: Shall we move this file to a separate subdirectory ?

either `scripts` or, if it's just for CI purposes (don't remember sry), under `.github/vms` for example?

.github/workflows/pytest-gpu.yml
```diff
@@ -0,0 +1,64 @@
...
+        command: |
+          export PATH=~/programs/install/bin:$PATH
+          export LD_LIBRARY_PATH=~/programs/install/lib:$LD_LIBRARY_PATH
+          source ~/anaconda3/etc/profile.d/conda.sh
+          conda activate devito
+          cd ~/programs/devito
+          git fetch
+          git checkout ${{ steps.extract_branch.outputs.branch }}
+          git pull
+          DEVITO_ARCH=clang DEVITO_PLATFORM=nvidiaX py.test -s -x --cov devito tests/test_gpu.py
```
FabioLuporini: nitpicking 1: for consistency with the other workflows, the env vars should have been defined earlier on around the workflow top

nitpicking 2: I don't think we want `-s` . Also unsure about `-x`. I think we want the same flags as in the other workflows

.github/workflows/pytest-gpu.yml
```diff
@@ -0,0 +1,64 @@
...
+
+    - name: extract branch name
+      shell: bash
+      run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
+      id: extract_branch
+
+    - name: look for ps1 file
+      run: |
+        ls '${{ env.OUTPUT_PATH }}/PSA'
+    - name: deploy VM
```
FabioLuporini: nitpicking: blank spce before this line?

.github/workflows/pytest-gpu.yml
```diff
@@ -0,0 +1,64 @@
...
+
+    - name: set host
+      run: echo ::set-output name=action_host::$(az vm show -d -g ${{ secrets.RESOURCE_GROUP }} -n gpu_ci_vm --query publicIps -o tsv)
+      id: host
+
+    - name: executing remote ssh commands test
+      uses: fifsky/ssh-action@master
+      with:
+        command: |
+          export PATH=~/programs/install/bin:$PATH
```
FabioLuporini: are the PATH and LD_LIBRARY_PATH exports so that `clang` and the cuda libs are visible ? if so, could we add a comment? 'cause I'm sure I will forget one day

.github/workflows/pytest-gpu.yml
FabioLuporini: so, if there's a test failure, then this command `executing remote ssh commands test` fails, which in turn makes the workflow fail. Correct?

rhodrin: Correct. And of course the bash output with the trace will be directly available in the github ui.

rhodrin: 1: These are 'step' environment variables, not 'global' ones so I believe they are defined in the correct location.

2: Will update.

PSA/deployVM.ps1
rhodrin: moved to `.github/azure`.

.github/workflows/pytest-gpu.yml
rhodrin: Added.

.github/workflows/pytest-gpu.yml
rhodrin: Yes. Added a comment.

.github/workflows/pytest-gpu.yml
```diff
@@ -0,0 +1,66 @@
...
+          # Ensure clang and cuda libraries are visible
+          export PATH=~/programs/install/bin:$PATH
+          export LD_LIBRARY_PATH=~/programs/install/lib:$LD_LIBRARY_PATH
+          source ~/anaconda3/etc/profile.d/conda.sh
+          conda activate devito
+          cd ~/programs/devito
+          git fetch
+          git checkout ${{ steps.extract_branch.outputs.branch }}
+          git pull
+          DEVITO_ARCH=clang DEVITO_PLATFORM=nvidiaX py.test --cov devito tests/test_gpu.py
```
mloubout: Two things
- If devito is already installed on the VM (looks like it no pip install), why not just set the env variables and PATH on the vm no make this a bit neater
- `--cov` is not doing anything by itself, need to run `codecov` afterward to upload the report

FabioLuporini:
about

> The deployed VM is not currently being de-allocated. De-allocating it and redeploying it for every push of course doesn't make sense, but we also probably want add something somewhere to stop the VM if it's idle for a certain amount of time. Will look into a solution for that soon.

any idea how much could it cost to keep it always on? 

@ggorman any insights into this ? 

> but we also probably want add something somewhere to stop the VM if it's idle for a certain amount of time.

that's a reasonable idea

rhodrin:
> but we also probably want add something somewhere to stop the VM if it's idle for a certain amount of time.

Looks like there should be a solution in here:
https://docs.microsoft.com/en-us/azure/automation/automation-solution-vm-management

But since this solution is independent of the PR it doesn't need to hold us back.

.github/workflows/pytest-gpu.yml
rhodrin: >If devito is already installed on the VM (looks like it no pip install), why not just set the env variables and PATH on the vm no make this a bit neater

I was getting funny behavior with paths in the `.bashrc` not being available though this action which is why `source ~/anaconda3/etc/profile.d/conda.sh` appears in the script. But I'll look for a solution to tidy things up a bit.

rhodrin: >` --cov` is not doing anything by itself, need to run codecov afterward to upload the report

Indeed, thanks.

.github/workflows/pytest-gpu.yml
```diff
@@ -0,0 +1,67 @@
...
+        command: |
+          #! /bin/bash
+          # Ensure clang and cuda libraries are visible
+          export PATH=~/programs/install/bin:$PATH
+          export LD_LIBRARY_PATH=~/programs/install/lib:$LD_LIBRARY_PATH
+          source ~/anaconda3/etc/profile.d/conda.sh
+          conda activate devito
+          cd ~/programs/devito
+          git fetch
+          git checkout ${{ steps.extract_branch.outputs.branch }}
```
rhodrin: Reminder to myself: This needs to be fixed to avoid potential merge conflicts from git pull.

FabioLuporini:
@rhodrin about

> How do we want to deploy VMs? i.e. one per branch or?

I'd say one per PR-ed branch ?

FabioLuporini:
Merged.

PR #1132, opened 2020-02-21
Fix subdomain subdims

rhodrin:
In current master, if two or more subdomains using, e.g., the 'middle' definition are created on the same grid dimension, 'new' subdimensions are not being defined correctly after the creation of the first subdomain. This PR fixes this bug.

Note: With the current fix, new subdimensions are being defined regardless of whether pre-existing subdimensions could be used. Using pre-existing subdimensions where possible is easy enough but requires a bit more work and thought - my suggestion is that we merge this PR for now and open an issue regarding the potential creation of redundant subdimensions.


FabioLuporini:
I think we can merge this and open an issue for the remaining questions. And here's one further question: am I wrong or now, in the general case, a SubDomain can safely be associated with only one Grid?

rhodrin:
> I think we can merge this and open an issue for the remaining questions. And here's one further question: am I wrong or now, in the general case, a SubDomain can safely be associated with only one Grid?

If we initialize a subdomain via, e.g., `subdom = MySub()` then `subdom` should be able to be passed to several grids right? At least that's the behavior we want no?

FabioLuporini:
But AFAIU the Dimension name is picked via the finalize method, so one cannot and shouldnt re-init through another grid ?

rhodrin:
Sounds good. I'll open and issue after the merge.

> But AFAIU the Dimension name is picked via the finalize method, so one cannot and shouldnt re-init through another grid ?

Oh, indeed, you're right. What a user should do is `subdom0 = MySub()`, `subdom1 = MySub()` etc. if they want to use the same subdomain definition in multiple grids.

PR #1136, opened 2020-02-24
Patch benchmark

FabioLuporini:


benchmarks/user/benchmark.py
```diff
@@ -245,8 +246,13 @@ def run_jit_backdoor(problem, **kwargs):
...
 
-    configuration['jit-backdoor'] = True
-    solver.forward(autotune=autotune)
+    @switchconfig(jit_backdoor=True)
+    def _run_jit_backdoor():
+        return solver.forward(autotune=autotune)
+
+    retval = _run_jit_backdoor()
+
+    return retval
```
mloubout: `return _run_jit_backdoor()` ?

FabioLuporini: fixed. I'm developing an aesthetic preference for splitting the two TBH, and I have no idea why !

PR #1139, opened 2020-02-25
Support for implicit solvers

BritishPiper:


devito/ir/equations/equation.py
```diff
@@ -27,10 +32,18 @@ def is_Scalar(self):
...
+    @property
+    def is_Break(self):
+        return self._is_Break
+
     @property
     def is_Increment(self):
         return self._is_Increment
 
+    @property
+    def is_Reduce(self):
```
FabioLuporini: what's, in practice, the difference between a Reduce and an Increment ?

devito/ir/equations/equation.py
```diff
@@ -154,12 +167,18 @@ def __new__(cls, *args, **kwargs):
         # Finally create the LoweredEq with all metadata attached
         expr = super(LoweredEq, cls).__new__(cls, expr.lhs, expr.rhs, evaluate=False)
 
+        expr._data = []
```
FabioLuporini: this smells like a hack

devito/ir/iet/nodes.py
```diff
@@ -353,6 +356,19 @@ def __init__(self, expr):
         super(Increment, self).__init__(expr, '+')
 
 
+class Break(Expression):
+    """A node representing a break statement"""
+
+    is_Break = True
+
+
+class Reduce(Expression):
```
FabioLuporini: Difference with `class Increment` ?

devito/types/equation.py
```diff
@@ -156,6 +161,58 @@ def __str__(self):
     __repr__ = __str__
 
 
+class Break(Eq):
```
georgebisbas: What is the purpose of this class? Can that be avoided?

devito/ir/equations/equation.py
ArthurGiannotta: The difference is that this can be used to, for example, calculate other useful norms, for example, the Max norm. The Increment only works in additive norms such as the p-norm. I will leave the Reduce class for now and use only the p-norm in my notebooks, even though the Max norm is recommended for the analysis of the Poisson problem.

devito/ir/equations/equation.py
ArthurGiannotta: 'Break' is a type of equation that will have specific properties that must be passed on to the LoweredEq to finally reach the visitors. These properties (namely 'condition' and 'else_expr') are specific to the Break class and I did not find any other way to pass such information further without creating an extra field to store equation specific data. The same is valid for the Reduce class.

devito/types/equation.py
ArthurGiannotta: I also discussed the purpose of the Break equation with Nelson. It will be used to allow halting the execution of an operator while in the middle of its execution. Currently all the loop iterations must be executed before exiting the Operator call. Let's say we want to stop the operator execution if we find out that the energy is getting dissipated or after a residual criteria has been met. Currently, the only way to do that is to execute some iterations, go back to python, check for the condition and then come back to C code. The Break equation introduces a way to implement iterative solvers since it allows checking for a condition and breaking the loop inside C code.

devito/ir/equations/equation.py
ArthurGiannotta: The problem is, as pointed out by Nelson, that this class does not work with MPI. Before discussing that further, I will remove that from this pull request.

devito/ir/iet/nodes.py
ArthurGiannotta: The "Reduce" class will be removed from the pull request.

devito/ir/clusters/algorithms.py
```diff
@@ -236,7 +236,9 @@ def callback(self, clusters, prefix, backlog=None, known_break=None):
 
         # Schedule Clusters over different IterationSpaces if this increases parallelism
         for i in range(1, len(clusters)):
-            if self._break_for_parallelism(scope, candidates, i):
+            # Implicit dimensions should prevent loop splitting
+            if not clusters[i].has_implicit_dims and \
```
rhodrin: Shouldn't implicit dims be handled before we get here so that this `if` statement is not needed?

devito/ir/clusters/cluster.py
```diff
@@ -149,6 +149,10 @@ def functions(self):
     def has_increments(self):
         return any(e.is_Increment for e in self.exprs)
 
+    @cached_property
```
rhodrin: Again, shouldn't implicit dims be handled prior to this point to avoid needing this property?

devito/ir/equations/equation.py
rhodrin: Yeah, I don't like this. @ArthurGiannotta lets talk about this at some stage to see what other implementations are possible.

devito/operator/operator.py
```diff
@@ -168,6 +169,15 @@ def _build(cls, expressions, **kwargs):
         # Python-level (i.e., compile time) and C-level (i.e., run time) performance
         profiler = create_profile('timers')
 
+        # Add missing implicit dimensions in the presence of a Break equation
+        if any(isinstance(i, Break) for i in expressions):
+            for e in expressions:
+                dims = list(e.implicit_dims)
+                for expr in preorder_traversal(e):
```
mloubout: How large is `e`? We tried to get rid of `preorder_traversal` everywhere a while back because it's quite heavy and slow.

devito/types/equation.py
```diff
@@ -81,11 +82,14 @@ def evaluate(self):
 
         The RHS of the Equation is evaluated at the indices of the LHS if required.
         """
-        try:
-            lhs, rhs = self.lhs.evaluate, self.rhs._eval_at(self.lhs).evaluate
-        except AttributeError:
-            lhs, rhs = self._evaluate_args()
-        eq = self.func(lhs, rhs, subdomain=self.subdomain,
+        if self.is_Break:
+            args = self._evaluate_args()
```
mloubout: Not sure why Break would be a property of Eq.

ArthurGiannotta: "Break" is a type of equation, just like "Increment". There is a property "is_Increment" and so there should be a property "is_Break", shouldn't it?

All expressions passed to the "Operator" class must have some properties exclusive to the "Eq" type, such as "lhs" and "lhs.function", which are used in various places. That is why I had to inherit from it when creating the "Break" class.

devito/ir/equations/equation.py
ArthurGiannotta: I will resolve this since the implementation completely changed.

devito/ir/support/utils.py
```diff
@@ -79,7 +79,7 @@ def build_iterators(mapper):
     iterators = OrderedDict()
     for k, v in mapper.items():
         for d, offs in v.items():
-            if d.is_Stepping:
+            if d.is_Stepping and k is not None:
```
ArthurGiannotta: Any code that uses a stepping dimension as either a plain symbol or an implicit_dimension generates this bug. Here is an example:

`from devito import Grid, TimeFunction, Eq, Operator
from numpy import int32

grid = Grid(shape=(50,), extent=(1.,))
u = TimeFunction(name = 'u', grid = grid)

# As an implicit dimension
u_eq = Eq(u.forward, u, implicit_dims = [grid.stepping_dim])

# As a plain symbol
#u_eq = Eq(u.forward, u + grid.stepping_dim)

op = Operator(u_eq)
`

devito/operator/operator.py
ArthurGiannotta: Removed `preorder_traversal` in favor of a proper devito.symbolics function `retrieve_dimensions`.

devito/ir/clusters/algorithms.py
FabioLuporini: correct, this is a hack

devito/ir/clusters/cluster.py
```diff
@@ -35,8 +35,11 @@ class Cluster(object):
     """
 
     def __init__(self, exprs, ispace, dspace, guards=None, properties=None):
-        self._exprs = tuple(ClusterizedEq(i, ispace=ispace, dspace=dspace)
-                            for i in as_tuple(exprs))
+        self._exprs = tuple(
+            ClusterizedBreak(i, ispace=ispace, dspace=dspace) if i.is_Break
```
FabioLuporini: ClusterizedBreak looks like a hack

I wonder whether we're actually getting the abstractions right (@rhodrin )

FabioLuporini: btw, if it's an Eq subtype, why is it even necessary to have a separate type of Cluster?

devito/ir/equations/equation.py
```diff
@@ -227,6 +263,26 @@ def func(self, *args, **kwargs):
     __reduce_ex__ = Pickable.__reduce_ex__
 
 
+class ClusterizedBreak(ClusterizedEq):
```
FabioLuporini: yeah, basically you're doubling the IR, which is something that should be avoided at all costs

devito/ir/support/utils.py
FabioLuporini: could you move this to a separate issue+PR+MFE+patch ?

also, the `if k is None: continue` could be added right below the `for k, v` loop

devito/passes/clusters/misc.py
```diff
@@ -30,8 +30,8 @@ def callback(self, clusters, prefix):
         lifted = []
         processed = []
         for n, c in enumerate(clusters):
-            # Increments prevent lifting
-            if c.has_increments:
+            # Increments and implicit dimensions prevent lifting
+            if c.has_increments or c.has_implicit_dims:
```
FabioLuporini: as Rhodri has pointed out, this smells like a hacky workaround

devito/types/equation.py
FabioLuporini: btw, should this really be called `Break`? 

why can't we just have `condition` has an additional kwarg of `Eq`, thus avoiding the significant amount of addition that this PR is trying to bring in?

FabioLuporini:
my overarching question ATM can be summarized as follows:

why, instead of `class Break`, don't we just add a `condition` kwarg/property to `Eq` ? wouldn't this just vastly simplify the compilation pipeline changes that this PR is trying to bring in?

devito/ir/support/utils.py
ArthurGiannotta: This bug will be moved to a new PR once I am in Brazil.

devito/types/equation.py
ArthurGiannotta: The initial reason behind making a new class is that `Break` is not actually an equation, but an inequation. One thing to notice too is that a Break condition is not a property of a single equation, but is a property of a set of equation that should stop executing whenever a condition is met.

So if implementing using this `condition` property, the actual Break instruction would look something like `brkEq = Eq(condition=residual<epsilon)`, is that OK for you? I will try to rewrite everything using that idea, which should avoid a lot of code indeed.

devito/ir/clusters/algorithms.py
ArthurGiannotta: Shouldn't `implicit_dims` force a loop configuration? If not, then there should be a way to do it. I was having 2 problems which lead me to make this "hacky" commit:

1. If the user inputs `implicit_dims = [time]`, then the equation should be forced inside the time loop, no matter what are the actual dependencies of the equation. I was having problems with the equation not entering the time loop even if the `implicit_dims` argument was set. By debugging it, I found that the `Lift` class was removing the implicit dimension from the clusters, even though it was an implicit dimension which should be a forced dependency.

2. Sometimes two equations that could be grouped together were being separated. This was due to the scheduling over two separated IterationSpaces having increased parallelism. This should not happen if one of the equations is a Break equation, for example. So maybe instead of checking for the presence of the implicit dimensions, maybe we should check for the presence of a break equation.

I do not have enough knowledge about Devito compilation steps to know exactly what could be done to create these sets of "forced" dimensions.

devito/types/equation.py
FabioLuporini: I think this:

> "Break" is a type of equation

is what @mloubout was arguing about

devito/types/equation.py
FabioLuporini: I think we have a number of possibilities and we should explore them all. I'm not sure which one is best yet, and I'd like feedback from the more seasoned developers (@rhodrin @mloubout )

> but is a property of a set of equation that

fair point. 

```
brkEq = Eq(break=residual<epsilon)
```

yes, I think this could work.  Thoughts?

I'm still a bit reluctant about the chosen name. How about `condition` ? 

I also have another question. How is this going to interact with [ConditionalDimension](http://devitocodes.github.io/devito/dimension.html#devito.types.dimension.ConditionalDimension) ? 

* Is this a completely separate abstraction? 
* Both will generate an `if-then` eventually, so can we, and if so, how are we going to, share the common bits of the compilation pipeline? For example, ConditionalDimensions are attached [here](https://github.com/devitocodes/devito/blob/master/devito/ir/equations/equation.py#L159) and later processed and attached to Clusters [here](https://github.com/devitocodes/devito/blob/master/devito/ir/clusters/algorithms.py#L287)

Again, I'm looking for feedback here

devito/ir/clusters/algorithms.py
FabioLuporini: Yes, I see what you mean. 

Can you write a Minimal Example where you have one such Eq being lifted out of the time loop, while you would want to keep it inside ? I'm curious to see the structure of that Eq. I'm not sure whether it's an API issue, a user-code issue, or a compiler issue quite yet. Clearly,  write this example for `master`, ie ignore all the Break story etc -- we don't need to see things wrapped in an `if`, and you may still comment the generated code for clarity (eg, "the `if` should have gone here") if it feels like it'd be useful for us

FabioLuporini: the concepts you wrote above may apply to `ConditionalDimension` (ie to other `if`s ) too. That's why I want to see clear in this story (hence the request above)

Here's one more reason to have `ConditionalDimension` and "`Break`" share as much of the compilation pipeline as possible. (IOW, at the cost of repeating myself, I think we should factor them togheter as quickly as possible as soon as we process them)

devito/ir/clusters/cluster.py
rhodrin: Yeah, not sure, will take a look.

devito/ir/support/utils.py
rhodrin: I'm not sure `grid.stepping_dim` should be used as an implicit dimension. Why would it be used over `grid.time_dim`? 

Also, I don't think it should be put into an equation either? I was going to suggest that `sympy.Mod` should be used instead, but that currently doesn't work and we have to write it as `grid.time_dim - c*floor(grid.time_dim/c)`.

devito/types/equation.py
rhodrin: I should have thought of and mentioned this before, but Sympy has inequality solvers right? Shouldn't we be 'mimicking' that api if possible? I'll look into them as post feedback.

ArthurGiannotta:
Changed basically everything to share the compilation pipeline of the `Break` equation with the `ConditionalDimension`. Now there are two types of conditions (break conditions and normal conditions), which can be changed by the parameter `brk` in `ConditionalDimension`. Break conditions will cause the loop to break whenever the conditions stops being met.

devito/ir/support/utils.py
```diff
@@ -78,6 +78,8 @@ def build_iterators(mapper):
     """
     iterators = OrderedDict()
     for k, v in mapper.items():
+        if k is None:
+            continue
         for d, offs in v.items():
             if d.is_Stepping:
                 sub_iterators = iterators.setdefault(d.parent, set())
```
ArthurGiannotta, line 85: This commit must be moved to a new pull request.

devito/ir/iet/visitors.py
```diff
@@ -265,12 +267,15 @@ def visit_Break(self, o):
         return c.Statement("break")
 
     def visit_Conditional(self, o):
-        then_body = c.Block(self._visit(o.then_body))
-        if o.else_body:
-            else_body = c.Block(self._visit(o.else_body))
-            return c.If(ccode(o.condition), then_body, else_body)
+        if o.condition is o.condition is BooleanTrue():
```
ArthurGiannotta, line 270: This check is done so the C code does not contain conditions like:
```
if (true) {
    ...;
}
```

devito/passes/clusters/misc.py
ArthurGiannotta: This has been moved to a configuration option.

FabioLuporini:
This is stale and now obsolete, unfortunately. Closing. @edsml-zl5621 is currently working on implicit solvers. @BritishPiper : should you ever get back to this, please get in touch. We always welcome collaborations, and your work here was great!

PR #1140, opened 2020-02-26
GPU workflow cleanup

rhodrin:
Change gpu workflow so that it only executes on a PR. Also some tidying and a new method for extracting the current branch name.

### DIFF
--- a/.github/workflows/pytest-gpu.yml
+++ b/.github/workflows/pytest-gpu.yml
@@ -17,7 +17,7 @@ on:
 jobs:
 
   # Deploy VM in Azure
-  DeployVM:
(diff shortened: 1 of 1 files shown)


.github/workflows/pytest-gpu.yml
```diff
@@ -1,36 +1,39 @@
...
+      - master
+  pull_request:
+    branches:
+      - master
 
 jobs:
 
   # Deploy VM in Azure
-  DeployVM:
+  deploy-vm:
```
mloubout: Still not correct this has to be:
```
build:
   name: deploy-vm

PR #1144, opened 2020-02-26
Flexible Operators (now accepting 'compiler' and 'platform')

FabioLuporini:
basically, to allow mixed execution modes in the same app -- so for example one operator runs on the CPU while another one on the GPU


FabioLuporini:
Merged

PR #1149, opened 2020-02-27
[RFC] backdoor: Add a timestamp comment in generated code

georgebisbas:
This PR just adds a timestamp each time a code is edited and recompiled under DEVITO_JIT_BACKDOOR.
May be useful. Not yet sure.
@FabioLuporini maybe the issue of creating a new hash for recompiled code I was talking about for this is not an actual problem.
I thought that adding a comment would change the hash but this is not the case I guess.


FabioLuporini:
it's not about the hash name, it's about some internal `codepy` garbage

devito/compiler.py
```diff
@@ -271,6 +272,7 @@ def jit_compile(self, soname, code):
             try:
                 with open(src_file, 'r') as f:
                     code = f.read()
+                    code = ''.join([code, '/* Backdoor edit at %s*/ \n' % time.ctime()])
```
FabioLuporini: OK so this way it's 100% sure that each time a different string is compiled right? 

LGTM. Have you tested it?

georgebisbas: So yes, this is tested. If anybody else would like to try, please go on.

PR #1153, opened 2020-03-04
Acoustic wavesolver: Specify stencil explicitly

rhodrin:
Remove the use of `solve` from the acoustic wave-solver.

### DIFF
--- a/examples/seismic/acoustic/operators.py
+++ b/examples/seismic/acoustic/operators.py
@@ -56,7 +56,9 @@ def iso_stencil(field, m, s, damp, kernel, **kwargs):
     eq = m * field.dt2 - H - kwargs.get('q', 0)
     # Add dampening field according to the propagation direction
     eq += damp * field.dt if kwargs.get('forward', True) else damp * field.dt.T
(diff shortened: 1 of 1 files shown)


examples/seismic/acoustic/operators.py
```diff
@@ -53,11 +53,12 @@ def iso_stencil(field, m, s, damp, kernel, **kwargs):
     # Define time sep to be updated
     next = field.forward if kwargs.get('forward', True) else field.backward
     # Define PDE
-    eq = m * field.dt2 - H - kwargs.get('q', 0)
-    # Add dampening field according to the propagation direction
-    eq += damp * field.dt if kwargs.get('forward', True) else damp * field.dt.T
-    # Solve the symbolic equation for the field to be updated
-    eq_time = solve(eq, next)
+    if kwargs.get('forward', True):
```
mloubout: - this fits in one line and no if else with a prev like next 
- I'd keep commented the solvebpqrt for reference
- H not needed without solve

mloubout: - q is missing
- 2.0 can go inside parentheses for cleaner expression

PR #1163, opened 2020-03-12
tutorials: Executed DRP notebook

rhodrin:
At some stage the 07_DRP_schemes notebook was somehow replaces with an unexecuted version. This PR fixes that.


rhodrin:
> Good catch. Can we add and assert on norm(u) and norm(u_DPR) for safety?

Yeah, probably a good idea. Will do shortly.

PR #1165, opened 2020-03-13
padfunc fix

mloubout:
add symbolic start to avoid pure integer index


mloubout:
Closing, need thought about Subdimension.

PR #1176, opened 2020-03-19
remove norm at end of test

### LINKED ISSUE #1133: Automatically get rid of norms*.npy
Any way to speed up local testing by avoiding deleting the norms ?
Keeping the norms fails test_mpi.py.

mloubout:
Fixes #1133


mloubout:
Nope, something wrong , `test_mpi` now is craaaazzzy slow, gonna assume the mpi setup that keeps creating/deleting not because it runs a new `py.test ...` for each rank. SO I think we should stick to previous case and maybe have a final cleaup at the end of test suite

Issue #1168, opened 2020-03-16
review/improve GPU CI action


ggorman:
Requesting a review from one of the Azure team so see if there is a better way of doing this.

georgebisbas:
This issue is being resolved in #1178, #1179.

PR #1178, opened 2020-03-20
tweak CI-gpu

rhodrin:
Tweaks to CI-gpu - now uses an updated, cleaner image. Secrets names 'future proofed'.

PR #1179, opened 2020-03-20
Add OpenACC testing to CI  Fixes: #1168

georgebisbas:
This PR adds OpenACC testing to CI.
It is built on top of #1178.

### DIFF
--- a/.github/workflows/pytest-gpu.yml
+++ b/.github/workflows/pytest-gpu.yml
@@ -46,7 +46,7 @@ jobs:
         RESOURCE_GROUP_REGION: uksouth
         VM_SIZE: Standard_NC6_Promo
         SERVER_NAME: gpu_ci_vm_${{ steps.extract_branch.outputs.branch }}
-        IMAGE: gpu_ci_vm-image-1168
+        IMAGE: gpu_ci_vm-image-0104
       run: >
         pwsh -command "& '${{ env.OUTPUT_PATH }}\.github\azure\deployVM.ps1'"
         -servicePrincipal ${{ secrets.SERVICE_PRINCIPAL_APPID }}
--- a/.github/workflows/pytest-gpu.yml
+++ b/.github/workflows/pytest-gpu.yml
@@ -71,11 +71,11 @@ jobs:
         command: |
           #! /bin/bash
           # Ensure clang libraries are visible
-          export PATH=/usr/local/cuda-10.1/bin:$PATH
-          # Ensure CUDA libraries are visible
-          export LD_LIBRARY_PATH=/usr/local/cuda-10.1/lib64:$LD_LIBRARY_PATH
           export PATH=~/install/bin:$PATH
           export LD_LIBRARY_PATH=~/install/lib:$LD_LIBRARY_PATH
+          # Ensure CUDA libraries are visible
+          export PATH=/usr/local/cuda-10.1/bin:$PATH
+          export LD_LIBRARY_PATH=/usr/local/cuda-10.1/lib64:$LD_LIBRARY_PATH
           # Ensure PGI libraries are visible
           export PGI=/opt/pgi
           export PATH=/opt/pgi/linux86-64/19.10/bin:$PATH
@@ -97,7 +97,6 @@ jobs:
           DEVITO_ARCH=clang DEVITO_PLATFORM=nvidiaX py.test tests/test_gpu_openmp.py
           DEVITO_ARCH=clang DEVITO_PLATFORM=nvidiaX py.test examples/seismic/elastic/elastic_example.py
           DEVITO_ARCH=clang DEVITO_PLATFORM=nvidiaX py.test examples/seismic/viscoelastic/viscoelastic_example.py
-          unset DEVITO_PLATFORM;
           DEVITO_ARCH=pgcc DEVITO_LANGUAGE=openacc py.test examples/seismic/acoustic/acoustic_example.py
           DEVITO_ARCH=pgcc DEVITO_LANGUAGE=openacc py.test examples/seismic/elastic/elastic_example.py
         host: ${{ steps.host.outputs.action_host }}
--- a/.github/workflows/pytest-gpu.yml
+++ b/.github/workflows/pytest-gpu.yml
@@ -46,7 +46,7 @@ jobs:
         RESOURCE_GROUP_REGION: uksouth
         VM_SIZE: Standard_NC6_Promo
         SERVER_NAME: gpu_ci_vm_${{ steps.extract_branch.outputs.branch }}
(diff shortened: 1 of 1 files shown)


.github/workflows/pytest-gpu.yml
```diff
@@ -56,24 +56,30 @@ jobs:
...
-          export LD_LIBRARY_PATH=~/programs/install/lib:$LD_LIBRARY_PATH
+          # Ensure clang, PGI and cuda libraries are visible
+          export PATH=/usr/local/cuda-10.1/bin:$PATH
+          export LD_LIBRARY_PATH=/usr/local/cuda-10.1/lib64:$LD_LIBRARY_PATH
+          export PATH=~/install/bin:$PATH
+          export LD_LIBRARY_PATH=~/install/lib:$LD_LIBRARY_PATH
+          export PGI=/opt/pgi
+          export PATH=/opt/pgi/linux86-64/19.10/bin:$PATH
+          export MANPATH=$MANPATH:/opt/pgi/linux86-64/19.10/man
+          export LM_LICENSE_FILE=$LM_LICENSE_FILE:/opt/pgi/license.dat-COMMUNITY-19.10
```
mloubout: This really need to go somwhere on the VM, this is a massive extra here.

georgebisbas: Maybe we should add all the paths in .bashrc of the image?

rhodrin: Yeah. Iirc putting the exports wasn't working with the action previously - some weird interaction with the action. But I'll play around and put them somewhere suitable.

rhodrin:
Note that #1178 has had a couple of further commits that'll need to be merged into this PR.

.github/workflows/pytest-gpu.yml
```diff
@@ -70,11 +70,15 @@ jobs:
       with:
         command: |
           #! /bin/bash
-          # Ensure clang and cuda libraries are visible
+          # Ensure clang, PGI and cuda libraries are visible
```
FabioLuporini: nitpicking: Split the comment. First a comment for clang exports, then a comment for pgcc exports, then a comment for cuda libs exports

.github/workflows/pytest-gpu.yml
```diff
@@ -91,6 +95,9 @@ jobs:
           DEVITO_ARCH=clang DEVITO_PLATFORM=nvidiaX py.test tests/test_gpu_openmp.py
           DEVITO_ARCH=clang DEVITO_PLATFORM=nvidiaX py.test examples/seismic/elastic/elastic_example.py
           DEVITO_ARCH=clang DEVITO_PLATFORM=nvidiaX py.test examples/seismic/viscoelastic/viscoelastic_example.py
+          unset DEVITO_PLATFORM;
```
FabioLuporini: I don't like it

I think the env vars should now be centralised in the action setup, so that you don't have to set/unset anything at this stage

.github/workflows/pytest-gpu.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         RESOURCE_GROUP_REGION: uksouth
         VM_SIZE: Standard_NC6_Promo
         SERVER_NAME: gpu_ci_vm_${{ steps.extract_branch.outputs.branch }}
-        IMAGE: CI-gpu-base-image-01
+        IMAGE: test-pgi-vm-image-20200320182727
```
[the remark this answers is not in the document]
georgebisbas: Changed

.github/workflows/pytest-gpu.yml
georgebisbas: Fixed

.github/workflows/pytest-gpu.yml
georgebisbas: I did a tried before on exporting an image from a VM and it was retained in .bashrc. I will try a bit later again.

.github/workflows/pytest-gpu.yml
georgebisbas: I agree. I will try to fix this together with the general env variables hiding.

.github/workflows/pytest-gpu.yml
mloubout: @FabioLuporini I don't agree. These are compiler and VM specif path that are not workflow related

.github/workflows/pytest-gpu.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         RESOURCE_GROUP_REGION: uksouth
```
georgebisbas: @rhodrin I just noticed this line here: So I would suspect that it could be straightforward if we also save an image in that region.

.github/workflows/pytest-gpu.yml
FabioLuporini: I meant the _action_ I think?

mloubout: But these are still VM/node/.... environment variables/ setups, not devito or pytest parameters.  What if you switch linux distribution, or want to test it on osx? you gonna hve 200 lines of exports/env vars that should not be here.

georgebisbas: Is there any verdict here? Should I try any other approach? Ideas?

.github/workflows/pytest-gpu.yml
```diff
@@ -70,11 +70,17 @@ jobs:
       with:
         command: |
           #! /bin/bash
-          # Ensure clang and cuda libraries are visible
+          # Ensure clang libraries are visible
           export PATH=/usr/local/cuda-10.1/bin:$PATH
+          # Ensure CUDA libraries are visible
           export LD_LIBRARY_PATH=/usr/local/cuda-10.1/lib64:$LD_LIBRARY_PATH
           export PATH=~/install/bin:$PATH
```
FabioLuporini: is this ~/install/bin really for CUDA?

.github/workflows/pytest-gpu.yml
FabioLuporini: I agree with @mloubout

.github/workflows/pytest-gpu.yml
FabioLuporini: what happens if you just drop the `unset` ?

.github/workflows/pytest-gpu.yml
georgebisbas: No. How did that happen, I remember it was fine. Maybe I rebased wrongly at some point.
Will fix now.

.github/workflows/pytest-gpu.yml
rhodrin: Well the image wouldn't be in the required resource group - so that'd have to be set up first.

Issue #1134, opened 2020-02-24
VM's set up by CI not automatically being de-allocated


rhodrin:
Currently the GPU CI creates one VM per pull request. These VM's are not automatically being de-allocated and hence have to be deleted manually.

A solution for this can potentially be found in here:
https://docs.microsoft.com/en-us/azure/automation/automation-solution-vm-management

Also, the GPU CI resource group name should probably be changed at some point.

rhodrin:
An Azure based solution has now been implimented for this.

A monitor has been set up to trigger an alert if a VM is idle and in turn a Web App Function is executed to remove the relevant VM.

Closing.

PR #1187, opened 2020-03-25
Delete GPU VM after build

dciborow:
#1134 

This is not testing but is an example of the recommendation.


rhodrin:
A solution for this has now been implimented directly in Azure through an alert triggering a function (and therefore closing this).

PR #1188, opened 2020-03-26
ci: Upgrade to use codecov-action/v1.0.6

FabioLuporini:


mloubout:
I think coverage is fixed with nbval too so would use the occasion to remove it from requirements and put back coverge of tutos

.github/workflows/tutorials.yml
```diff
@@ -90,20 +90,28 @@ jobs:
 
     - name: Seismic notebooks
       run: |
-        $RUN_CMD py.test --nbval $SKIP examples/seismic/tutorials/
+        $RUN_CMD py.test --nbval --cov . --cov-config=.coveragerc --cov-report=xml:seis_coverage.xml $SKIP examples/seismic/tutorials/
```
FabioLuporini: out of curiosity, what are the extra options for?

mloubout: Same as in `pytest` one, basically config, report format and name

FabioLuporini:
Merged.

Issue #1128, opened 2020-02-18
--help not working in benchmarking


georgebisbas:
benchmark.py run/test/bench/run-jit-backdoor --help is broken:

---To reproduce:
`python benchmark.py run --help`

--Log:
```
python benchmarks/user/benchmark.py run --help
Traceback (most recent call last):
  File "benchmarks/user/benchmark.py", line 507, in <module>
    benchmark()
  File "/home/gb4018/anaconda3/envs/devito/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/gb4018/anaconda3/envs/devito/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/gb4018/anaconda3/envs/devito/lib/python3.7/site-packages/click/core.py", line 1135, in invoke
    sub_ctx = cmd.make_context(cmd_name, args, parent=ctx)
  File "/home/gb4018/anaconda3/envs/devito/lib/python3.7/site-packages/click/core.py", line 641, in make_context
    self.parse_args(ctx, args)
  File "/home/gb4018/anaconda3/envs/devito/lib/python3.7/site-packages/click/core.py", line 940, in parse_args
    value, args = param.handle_parse_result(ctx, opts, args)
  File "/home/gb4018/anaconda3/envs/devito/lib/python3.7/site-packages/click/core.py", line 1477, in handle_parse_result
    self.callback, ctx, self, value)
  File "/home/gb4018/anaconda3/envs/devito/lib/python3.7/site-packages/click/core.py", line 96, in invoke_param_callback
    return callback(ctx, param, value)
  File "/home/gb4018/anaconda3/envs/devito/lib/python3.7/site-packages/click/core.py", line 860, in show_help
    echo(ctx.get_help(), color=ctx.color)
  File "/home/gb4018/anaconda3/envs/devito/lib/python3.7/site-packages/click/core.py", line 516, in get_help
    return self.command.get_help(self)
  File "/home/gb4018/anaconda3/envs/devito/lib/python3.7/site-packages/click/core.py", line 879, in get_help
    self.format_help(ctx, formatter)
  File "/home/gb4018/anaconda3/envs/devito/lib/python3.7/site-packages/click/core.py", line 898, in format_help
    self.format_options(ctx, formatter)
  File "/home/gb4018/anaconda3/envs/devito/lib/python3.7/site-packages/click/core.py", line 919, in format_options
[...]

PR #1171, opened 2020-03-17
Consolidation and cleanup -- revisited

FabioLuporini:
supersedes #1154 

all original comments in #1154 have already been addressed, AFAICT

I'll leave some comments below to highlight the new changes

PR #1190, opened 2020-03-26
Fixes issue #1128

georgebisbas:
This PR fixes issue #1128 related to printing `dle` and `autotuning` options.


georgebisbas:
> Maybe a test for it?

The test should just be a call to `benchmark.py run/test/bench/run-jit-backdoor --help`
Should I add it? 0-cost test in any case.

FabioLuporini:
no that's fine

this is gonna create a minor conflict with #1171 since I'm dropping `dle` in there, I think

so depending on the merge order, someone will have to deal with it. Anyway, really minor observation

PR #1194, opened 2020-03-26
Fix index_glb_to_loc and _size_outhalo

### LINKED ISSUE #1166: Potentially wrong `shape_with_halo` with HALO > DOMAIN
This might cause NaNs or segfaults by essentially breaking the contract between python-land and generated code

Reproducer here: https://github.com/devitocodes/devito/compare/fix-shape-with-halo

For an Operator that produces NaNs, take [this test](https://github.com/devitocodes/devito/blob/consolidate-aliases-3/tests/test_dse.py#L582
) and change the grid such that, in at least one decomposed Dimension, it's smaller than the halo size, for example (3, 3, 3)
it will produce garbage values -- but you have to run it many times until it produces NaNs, thus causing np.close to fail ! or make the test "more reasonable" such that it produces a given expected value against which you can assert deterministically

rhodrin:
This PR fixes the method `index_glb_to_loc` in `decomposition.py` for when an offset + side are received. Some appropriate tests have been added.

The computation of `_size_outhalo` in `dense.py` has also been fixed.

This fixes #1166.

Note: An MPI test for `_size_outhalo` must be added prior to merge. Submitting as it currently to ensure all tests passing.

### DIFF
--- a/devito/data/decomposition.py
+++ b/devito/data/decomposition.py
@@ -144,10 +144,10 @@ class Decomposition(tuple):
             There are three possible cases:
             * int. Given ``I``, a global index, return the corresponding
               relative local index if ``I`` belongs to the local subdomain,
-              None otherwise.
+              ``None`` otherwise.
             * int, DataSide. Given ``O`` and ``S``, respectively a global
               offset and a side, return the relative local offset. This
-              can be 0 if the local subdomain doesn't intersect with the
+              can be ``None`` if the local subdomain doesn't intersect with the
               region defined by the given global offset.
             * (int, int).  Given global ``(min, max)``, return ``(min', max')``
               representing the corresponding relative local min/max. If the
@@ -313,16 +313,20 @@ class Decomposition(tuple):
         elif len(args) == 2:
             # index_glb_to_loc(offset, side)
             if self.loc_empty:
-                return 0
-            rel_ofs, side = args
+                return None
+            abs_ofs, side = args
             if side is LEFT:
-                abs_ofs = self.glb_min + rel_ofs
-                size = self.loc_abs_max - base + 1
-                return min(abs_ofs - base, size) if abs_ofs > base else 0
+                rel_ofs = self.glb_min + abs_ofs - base
+                if abs_ofs >= base and abs_ofs <= top:
+                    return rel_ofs
+                else:
+                    return None
             else:
-                abs_ofs = self.glb_max - rel_ofs
-                size = top - self.loc_abs_min + 1
-                return min(top - abs_ofs, size) if abs_ofs < top else 0
+                rel_ofs = abs_ofs - (self.glb_max - top)
+                if abs_ofs >= self.glb_max - top and abs_ofs <= self.glb_max - base:
+                    return rel_ofs
+                else:
+                    return None
         else:
             raise TypeError("Expected 1 or 2 arguments, found %d" % len(args))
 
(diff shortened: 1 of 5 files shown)


devito/types/dense.py
```diff
@@ -311,10 +311,10 @@ def _size_outhalo(self):
         if self._distributor is None:
             return self._size_inhalo
```
FabioLuporini: Could you add a comment briefly explaining what's going on here ?

devito/data/decomposition.py
```diff
@@ -313,16 +313,20 @@ def index_glb_to_loc(self, *args, rel=True):
...
+                    return None
             else:
-                abs_ofs = self.glb_max - rel_ofs
-                size = top - self.loc_abs_min + 1
-                return min(top - abs_ofs, size) if abs_ofs < top else 0
+                rel_ofs = abs_ofs - (self.glb_max - top)
+                if abs_ofs >= self.glb_max - top and abs_ofs <= self.glb_max - base:
+                    return rel_ofs
+                else:
+                    return None
```
mloubout: Looks fine to me. Small question, does this incur some performance slow down? This is called a lot in sparse with MPI (for each coordinate individually). Doesn't look like it would but  just checking

rhodrin: The MPI build is currently taking ages, will investigate to see if this could be the reason.

mloubout: MPI taking ages for quite a while now, suddently doubled time at some point. Still not sure why, so check with previous build if slower or just normal slow.

rhodrin: Checked against some previous build and they were ~80 mins. This is at 100 and still going...

devito/types/dense.py
```diff
@@ -308,16 +308,29 @@ def shape_global(self):
...
-                 for d, i in zip(self.dimensions, self._size_inhalo.right)]
+        left = [abs(min(i.loc_abs_min-i.glb_min-j, 0)) if i and not i.loc_empty else 0
+                for i, j in zip(self._decomposition, self._size_inhalo.left)]
+        right = [max(i.loc_abs_max+j-i.glb_max, 0) if i and not i.loc_empty else 0
+                 for i, j in zip(self._decomposition, self._size_inhalo.right)]
 
         sizes = tuple(Size(i, j) for i, j in zip(left, right))
 
+        if self._distributor.is_parallel and any(left) > 0 or any(right) > 0:
+            try:
```
FabioLuporini, line 325: why do you need the try-catch ?

devito/types/dense.py
```diff
@@ -308,16 +308,29 @@ def shape_global(self):
...
+                if not self._distributor.boundary_rank:
+                    warning("A space order of %d has been set but the current rank "
```
FabioLuporini, line 327: I think we should include the "halo size" in this msg. One could indeed override the default halo that a given space order gives you through "Function(...., halo=...)".

I also wonder whether (nitpicking) "domain size" sounds better than "spatial extent" ? we use "extent" for physical quantities and "size" for discrete ones

devito/mpi/distributed.py
```diff
@@ -245,6 +245,12 @@ def nprocs(self):
     def topology(self):
         return self._topology
 
+    @cached_property
+    def boundary_rank(self):
```
FabioLuporini, line 249: nitpicking (for homogeneity): in Devito, all introspection properties start with "is_"

devito/types/dense.py
rhodrin: It's for sparse functions.

devito/types/dense.py
rhodrin: Oki, will edit.

PR #1202, opened 2020-03-31
Speed up compilation

FabioLuporini:
A collection of self-contained, orthogonal commits. I could have filed 20 PRs, one per commit, or just one.

On my workstation, the compilation times w.r.t. `master` become
- tti so12 : ~27s -> ~12s
- smoothers : ~5.5s -> ~1s
- elastic so12: ~18s  -> ~8.5s
- viscoelastic so12 : ~41s -> ~14s
- iso-acoustic so12 :  ~4s -> ~2.5s

@tjb900 I'm not sure if you care about this ... tagging you, just in case

@ggorman I've been working on this intermittently for a few days, but there is still work to do

@mloubout @rhodrin if you take a profile with Python's cProfile, you shall see there's a margin of improvement especially in the DSL/SymPy layers... wonder whether/when you could this give a go. Not a top priority of course!

### DIFF
--- a/devito/finite_differences/lazy.py
+++ b/devito/finite_differences/lazy.py
@@ -1,0 +1,44 @@
+__all__ = ['Evaluable']
+
+
+class Evaluable(object):
+
+    """
+    A mixin class for types inherited from SymPy that may carry nested
+    unevaluated arguments.
+
+    This mixin is used to implement lazy evaluation of expressions.
+    """
+
+    @classmethod
+    def _evaluate_maybe_nested(cls, maybe_evaluable):
+        if isinstance(maybe_evaluable, Evaluable):
+            return maybe_evaluable.evaluate
+        try:
+            # Not an Evaluable, but some Evaluables may still be hidden within `args`
+            if maybe_evaluable.args:
+                evaluated = [Evaluable._evaluate_maybe_nested(i)
+                             for i in maybe_evaluable.args]
+                return maybe_evaluable.func(*evaluated)
+            else:
+                return maybe_evaluable
+        except AttributeError:
+            # No `args` to be visited
+            return maybe_evaluable
+
+    @property
+    def args(self):
+        return ()
+
+    @property
+    def func(self):
+        return self.__class__
+
+    def _evaluate_args(self):
+        return [Evaluable._evaluate_maybe_nested(i) for i in self.args]
+
+    @property
+    def evaluate(self):
+        """Return a new object from the evaluation of ``self``."""
+        args = self._evaluate_args()
+        return self.func(*args)
--- a/devito/finite_differences/lazy.py
+++ b/devito/finite_differences/lazy.py
@@ -17,9 +17,13 @@ class Evaluable(object):
         try:
             # Not an Evaluable, but some Evaluables may still be hidden within `args`
             if maybe_evaluable.args:
-                evaluated = [Evaluable._evaluate_maybe_nested(i)
-                             for i in maybe_evaluable.args]
-                return maybe_evaluable.func(*evaluated)
+                args = [Evaluable._evaluate_maybe_nested(i) for i in maybe_evaluable.args]
+                evaluate = not all(i is j for i, j in zip(args, maybe_evaluable.args))
+                try:
+                    return maybe_evaluable.func(*args, evaluate=evaluate)
+                except TypeError:
+                    # Not all objects support the `evaluate` kwarg
+                    return maybe_evaluable.func(*args)
             else:
                 return maybe_evaluable
         except AttributeError:
@@ -41,4 +45,5 @@ class Evaluable(object):
     def evaluate(self):
         """Return a new object from the evaluation of ``self``."""
         args = self._evaluate_args()
-        return self.func(*args)
+        evaluate = not all(i is j for i, j in zip(args, self.args))
+        return self.func(*args, evaluate=evaluate)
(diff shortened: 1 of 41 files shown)


devito/symbolics/manipulation.py
```diff
@@ -203,6 +156,74 @@ def run(expr):
     return built + rebuilt, built
 
 
+def uxreplace(expr, rule):
```
FabioLuporini: @mloubout you may like this one

FabioLuporini:
Forgot to add: depends on the removal of OPS and YASK.

devito/types/tensor.py
```diff
@@ -389,15 +389,15 @@ def forward(self):
         i = int(self.time_order / 2) if self.time_order >= 2 else 1
         _t = self.indices[self._time_position]
 
-        return self.subs(_t, _t + i * _t.spacing)
+        return self.xreplace({_t: _t + i * _t.spacing})
```
mloubout, line 392: These ones probably need subs instead of xreplace too (.forward/backward)

devito/symbolics/manipulation.py
```diff
@@ -203,6 +156,74 @@ def run(expr):
...
+    return expr.func(*args)
+
+
+@_uxreplace_handle.register(sympy.Add)
+def _(expr, args):
+    if all(i.is_commutative for i in args):
+        _addsort(args)  # In-place sorting
+        return expr.func(*args, evaluate=False)
+    else:
+        return expr.func(*args)
```
mloubout, line 208: Why not just `return expr`?

FabioLuporini: by the time I get here we know the `args` have changed, so I must rebuild `expr` with the new args

mloubout: Isn't it rebuilinding everything from scratch. Wouldn't `_raw_from_args` or something be better or a `differentiable.....` to avoid reprocessing already processed args? Just wondering, this whole uxreplace looks pretty good overall.

georgebisbas:
On my Azure workstation, the compilation times w.r.t. master become with Intel Compiler:
(DEVITO_ARCH=icc, DEVITO_OPENMP=1, 10 threads-ht off)
    tti so12 : ~5s -> ~5s
    elastic so12: ~12s -> ~12s
    viscoelastic so12 : ~16s -> ~14s
    iso-acoustic so12 : ~2s -> ~2s

Am I doing sth wrong?

devito/symbolics/manipulation.py
FabioLuporini: it's a subtletly: not all args are commutative (see `if` above) and `raw_fromargs` does **not** handle that AFAIU, so the safe way to go is re-evaluation. I **might** be wrong though

devito/finite_differences/lazy.py
```diff
@@ -0,0 +1,49 @@
+__all__ = ['Evaluable']
+
+
+class Evaluable(object):
```
mloubout, line 4: More for future, would it be good to have a state that knows whether it is already evaluated or not?

devito/finite_differences/tools.py
```diff
@@ -157,6 +158,11 @@ def symbolic_weights(function, deriv_order, indices, dim):
             for j in range(0, len(indices))]
 
 
+@cacheit
```
mloubout, line 161: ? why cache it and why not our memoized?

devito/ir/clusters/queue.py
```diff
@@ -1,13 +1,16 @@
...
+
+__all__ = ['Queue', 'QueueStateful']
 
 
 class Queue(object):
 
     """
-    A special queue to process objects in nested IterationSpaces based on
-    a divide-and-conquer algorithm.
+    A special queue to process Clusters based on a divide-and-conquer algorithm.
```
mloubout, line 13: Queue?

devito/symbolics/manipulation.py
mloubout: It does in some minimal way but probably not much impact on perf
https://github.com/sympy/sympy/blob/d3ae62701d21bb07c941609d0fd0e045e55ff809/sympy/core/operations.py#L116

devito/finite_differences/lazy.py
FabioLuporini: perhaps we should be using cached_properties ? 
not sure how would a state be useful

devito/finite_differences/tools.py
FabioLuporini: good spot, there was an obvious issue with our memoization that I know have forgotten, I should have added a comment...

devito/ir/clusters/queue.py
FabioLuporini: it's not a queue, it's basically a special "implicit-tree visitor". This is just a legacy name. I'll change it to something else once and for all in another PR, but before another minor cleanup will be required, so I'm sticking to Queue for the time being.

It's not "too wrong" in the sense that it processes clusters in FIFO order (like a standard queue), where FIFO is dictated by how the clusters would appear in the generated code

devito/symbolics/manipulation.py
FabioLuporini: Tried in a few examples (visco, tti). Didn't see any improvement basically, probably because we rarely enter that branch ( which is good). However, I ran all tests and it seems to be OK. I pushed the change, so if you feel confident about it, we can proceed this way for me

PR #1203, opened 2020-03-31
Remove YASK and OPS backends

FabioLuporini:
Reason: we do not have the manpower to sustain the implementation and maintenance effort


FabioLuporini:
Merged.

PR #1205, opened 2020-04-01
OpenACC inception

FabioLuporini:


devito/core/gpu_openacc.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+class DeviceAccizer(DeviceOmpizer):
+
+    lang = {
+        'atomic': c.Pragma('acc atomic update'),
+        'map-enter-to': lambda i, j:
+            c.Pragma('acc enter data copyin(%s%s)' % (i, j)),
+        # 'map-enter-alloc': lambda i, j:
+        #    c.Pragma('omp target enter data map(alloc: %s%s)' % (i, j)),
```
mloubout, line 34: Leftover debug?

devito/core/__init__.py
```diff
@@ -32,6 +34,10 @@
 operator_registry.add(DeviceOpenMPOperator, Device, 'advanced')
 operator_registry.add(DeviceOpenMPCustomOperator, Device, 'custom')
 
+operator_registry.add(DeviceOpenACCNoopOperator, Device, 'noop', 'openacc')
+operator_registry.add(DeviceOpenACCOperator, Device, 'advanced', 'openacc')
+operator_registry.add(DeviceOpenACCCustomOperator, Device, 'custom', 'openacc')
+
```
mloubout, line 40: So shouldn't the rest of the registery have a language too like openmp for OMPoffloading and C for the rest, only ths one seems to use language

devito/core/gpu_openacc.py
georgebisbas: As far as I know there is no explicit correspondence bewtween openacc and openmp in this directive

FabioLuporini: yes, more of a memo. Anyway, already dropped in the branch on top of this

devito/core/__init__.py
FabioLuporini: yes. `operator_registry` defaults to `language=C` if not provided. Currently, we still have the issue that OpenMP is somewhat special (it even has its own env var and compilation passes!).

Some cleanup is necessary, but I'd say that's for another day

devito/core/gpu_openacc.py
```diff
@@ -0,0 +1,171 @@
...
+    @classmethod
+    @timed_pass(name='specializing.IET')
+    def _specialize_iet(cls, graph, **kwargs):
+        options = kwargs['options']
+
+        # Distributed-memory parallelism
+        if options['mpi']:
+            mpiize(graph, mode=options['mpi'])
+
+        # GPU parallelism via OpenAC offloading
```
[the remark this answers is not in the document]
FabioLuporini: OK

PR #1212, opened 2020-04-03
Adjoint tti

ofmla:
Adjoint for TTI based on Mathias work https://library.seg.org/doi/abs/10.1190/segam2018-2996274.1 Central FD scheme is used to solve the  TTI  adjoint equation.


examples/seismic/tti/wavesolver.py
```diff
@@ -126,3 +133,78 @@ def forward(self, src=None, rec=None, u=None, v=None, vp=None,
         summary = op.apply(src=src, rec=rec, u=u, v=v,
                            dt=kwargs.pop('dt', self.dt), **kwargs)
         return rec, u, v, summary
+
+    def adjoint(self, rec, srca=None, u=None, v=None, vp=None,
+                epsilon=None, delta=None, theta=None, phi=None,
+                save=False, kernel='centered', **kwargs):
+        """
+        Forward modelling function that creates the necessary
```
georgebisbas, line 141: Description needs to be changed, right?

examples/seismic/tti/operators.py
```diff
@@ -5,31 +5,37 @@
 from devito.finite_differences import centered, first_derivative, transpose
 
 
-def second_order_stencil(model, u, v, H0, Hz):
+def second_order_stencil(m, damp, epsilon, delta, s, u, v, H0, Hz, **kwargs):
```
mloubout, line 8: Not sure why this change, m,damp,... all in model, no need for explicit args.
Also would avoid kwargs and just use `(.., forward=False)` as it is always needed.

examples/seismic/tti/operators.py
```diff
@@ -193,12 +199,22 @@ def kernel_centered_2d(model, u, v, space_order):
...
+    s = model.grid.stepping_dim.spacing
+
+    if kwargs.get('forward', True):
+        Gxx = Gxx_centered_2d(model, u, costheta, sintheta, space_order)
+        Gzz = Gzz_centered_2d(model, v, costheta, sintheta, space_order)
+        return second_order_stencil(m, damp, epsilon, delta, s, u, v, Gxx, Gzz)
+    else:
+        Gxx = Gxx_centered_2d(model, (epsilon*u + delta*v), costheta, sintheta, space_order)
+        Gzz = Gzz_centered_2d(model, (delta*u + v), costheta, sintheta, space_order)
+        return second_order_stencil(m, damp, epsilon, delta, s, u, v, Gxx, Gzz, forward=False)
```
mloubout, line 214: you would avoid the other `if/else` if you compute H0/Hz inside this one and pass it to second_order_stencil:
```
forward: H0 = 1+2epsilon Gxx _ ..., 
adjoint_: H0 = Gxx_...

second_order_stencil(model, H0, Hz)
```

examples/seismic/tti/operators.py
```diff
@@ -193,12 +199,22 @@ def kernel_centered_2d(model, u, v, space_order):
...
 
 
-def kernel_centered_3d(model, u, v, space_order):
+def kernel_centered_3d(model, u, v, space_order, **kwargs):
```
mloubout, line 217: make forwar explicit too, no kargs.

examples/seismic/tti/operators.py
```diff
@@ -399,5 +425,60 @@ def ForwardOperator(model, geometry, space_order=4,
     return Operator(stencils, subs=model.spacing_map, name='ForwardTTI', **kwargs)
 
 
+def AdjointOperator(model, geometry, space_order=4,
+                    save=False, kernel='centered', **kwargs):
```
mloubout, line 429: if staggered not supported kernel should not be there. No save

examples/seismic/tti/operators.py
```diff
@@ -399,5 +425,60 @@ def ForwardOperator(model, geometry, space_order=4,
...
+    """
+    Construct an adjoint modelling operator in an acoustic media.
```
mloubout, line 431: tti not acoustic

examples/seismic/tti/operators.py
```diff
@@ -399,5 +425,60 @@ def ForwardOperator(model, geometry, space_order=4,
...
+    Parameters
+    ----------
+    model : Model
+        Object containing the physical parameters.
+    geometry : AcquisitionGeometry
+        Geometry object that contains the source (SparseTimeFunction) and
+        receivers (SparseTimeFunction) and their position.
+    data : ndarray
+        IShot() object containing the acquisition geometry and field data.
+    time_order : int
```
mloubout, line 442: deperecated.

examples/seismic/tti/operators.py
```diff
@@ -399,5 +425,60 @@ def ForwardOperator(model, geometry, space_order=4,
...
+
+    Parameters
+    ----------
+    model : Model
+        Object containing the physical parameters.
+    geometry : AcquisitionGeometry
+        Geometry object that contains the source (SparseTimeFunction) and
+        receivers (SparseTimeFunction) and their position.
+    data : ndarray
+        IShot() object containing the acquisition geometry and field data.
```
mloubout, line 441: ISHot is from 4 years ago, no data in the function signature.

examples/seismic/tti/operators.py
```diff
@@ -399,5 +425,60 @@ def ForwardOperator(model, geometry, space_order=4,
...
+    time_order : int
+        Time discretization order.
+    space_order : int
+        Space discretization order.
+    """
+
+    dt = model.grid.time_dim.spacing
+    m = model.m
+    time_order = 1 if kernel == 'staggered' else 2
```
mloubout, line 450: No, all staggerd must go and add a warning/error it's nit supported in adjoint.

examples/seismic/tti/operators.py
```diff
@@ -399,5 +425,60 @@ def ForwardOperator(model, geometry, space_order=4,
...
+    else:
+        stagg_u = stagg_v = None
+
+    # Create symbols for forward wavefield, source and receivers
+    u = TimeFunction(name='u', grid=model.grid, staggered=stagg_u,
+                     save=geometry.nt if save else None,
+                     time_order=time_order, space_order=space_order)
+    v = TimeFunction(name='v', grid=model.grid, staggered=stagg_v,
+                     save=geometry.nt if save else None,
+                     time_order=time_order, space_order=space_order)
```
mloubout, line 462: These should not be called u/v to avoid future messy confusion trying to make the gradient. Please follow convention in ref (u/v forward, p/r adjoint)

examples/seismic/tti/wavesolver.py
```diff
@@ -126,3 +133,78 @@ def forward(self, src=None, rec=None, u=None, v=None, vp=None,
         summary = op.apply(src=src, rec=rec, u=u, v=v,
                            dt=kwargs.pop('dt', self.dt), **kwargs)
         return rec, u, v, summary
+
+    def adjoint(self, rec, srca=None, u=None, v=None, vp=None,
+                epsilon=None, delta=None, theta=None, phi=None,
+                save=False, kernel='centered', **kwargs):
```
mloubout, line 139: no kernel, only standard supported.

examples/seismic/tti/wavesolver.py
```diff
@@ -126,3 +133,78 @@ def forward(self, src=None, rec=None, u=None, v=None, vp=None,
...
+            The time-constant velocity.
+        epsilon : Function or float, optional
+            The time-constant first Thomsen parameter.
+        delta : Function or float, optional
+            The time-constant second Thomsen parameter.
+        theta : Function or float, optional
+            The time-constant Dip angle (radians).
+        phi : Function or float, optional
+            The time-constant Azimuth angle (radians).
+        save : int or Buffer
```
mloubout, line 163: no save in adjoint?

examples/seismic/tti/wavesolver.py
```diff
@@ -126,3 +133,78 @@ def forward(self, src=None, rec=None, u=None, v=None, vp=None,
...
+            Option to store the entire (unrolled) wavefield.
+        kernel : str, optional
```
mloubout, line 165: No kernel

examples/seismic/tti/wavesolver.py
```diff
@@ -126,3 +133,78 @@ def forward(self, src=None, rec=None, u=None, v=None, vp=None,
...
+            Type of discretization, centered or shifted.
+
+        Returns
+        -------
+        Receiver, wavefield and performance summary.
+        """
+
+        if kernel == 'staggered':
```
mloubout, line 173: This should go, no adjoint staggered.

examples/seismic/tti/wavesolver.py
```diff
@@ -126,3 +133,78 @@ def forward(self, src=None, rec=None, u=None, v=None, vp=None,
...
+
+        Parameters
+        ----------
+        geometry : AcquisitionGeometry
+            Geometry object that contains the source (SparseTimeFunction) and
+            receivers (SparseTimeFunction) and their position.
+        u : TimeFunction, optional
+            The computed wavefield first component.
+        v : TimeFunction, optional
+            The computed wavefield second component.
```
mloubout, line 152: rename p/r

examples/seismic/tti/wavesolver.py
```diff
@@ -126,3 +133,78 @@ def forward(self, src=None, rec=None, u=None, v=None, vp=None,
...
+            stagg_v = (-dims[0], -dims[1]) if self.model.grid.dim == 3 else (-dims[0])
+        else:
+            time_order = 2
+            stagg_u = stagg_v = None
+        # Source term is read-only, so re-use the default
+        srca = srca or PointSource(name='srca', grid=self.model.grid,
+                                   time_range=self.geometry.time_axis,
+                                   coordinates=self.geometry.src_positions)
+        # Create the forward wavefield if not provided
+        if u is None:
```
mloubout, line 186: rename p/r

examples/seismic/tti/wavesolver.py
```diff
@@ -126,3 +133,78 @@ def forward(self, src=None, rec=None, u=None, v=None, vp=None,
...
+            u = TimeFunction(name='u', grid=self.model.grid, staggered=stagg_u,
+                             save=self.geometry.nt if save else None,
+                             time_order=time_order, space_order=self.space_order)
+        # Create the forward wavefield if not provided
+        if v is None:
+            v = TimeFunction(name='v', grid=self.model.grid, staggered=stagg_v,
+                             save=self.geometry.nt if save else None,
+                             time_order=time_order, space_order=self.space_order)
+
+        if kernel == 'staggered':
```
mloubout, line 196: remove

examples/seismic/tti/operators.py
FabioLuporini: or ``raise NotImplementedError``

examples/seismic/tti/wavesolver.py
ofmla: yes, the dangers of copy-paste practice :P

ofmla:
@mloubout amendments adopted OK. I also add the dot-test to `test\test_adjoint.py`. I wrote a separate test function for the TTI case, I guess  there should be a way of testing both operators Isotropic and anisotropic.

ofmla:
It seems that the dot-test fails for the 3D cases, however I was able to run the `test_adjoint.py` without problems in my workstation. How can I know what was the issue?

ofmla:
> It seems that the dot-test fails for the 3D cases, however I was able to run the `test_adjoint.py` without problems in my workstation. The values for <Ax,y> are huge, so I will take a look.

My workstation --> gcc 4.8.5, python 3.7.6 :heavy_check_mark: 
With newer version of gcc the TTI adjoint test for 3D fails :suspect:

FabioLuporini:
@ofmla you will have to tell us how it fails. It might just be a tolerance issue. Can you paste the whole error trace?

tests/test_adjoint.py
```diff
@@ -59,6 +64,46 @@ def test_adjoint_F(self, mkey, shape, kernel, space_order):
...
         assert np.isclose((term1 - term2)/term1, 0., atol=1.e-12)
 
+    @pytest.mark.parametrize('mkey, shape, space_order', [
+        # 2D tests with varying space orders
+        ('layers', (30, 35), 8), ('layers', (30, 35), 4),
+        # 3D tests with varying space orders
+        ('layers', (30, 35, 40), 8), ('layers', (30, 35, 40), 4),
+        # Constant model in 2D and 3D
+        ('constant', (30, 35), 8), ('constant', (30, 35, 40), 4),
+        ('constant', (30, 35), 8), ('constant', (30, 35, 40), 4),
```
mloubout, line 74: I would avoid to use this model. Putting anisotropy at source location is really bad for this equation so would stick to layers only.

ofmla: Is true, so the `constant-tti` model should be taken out of `demo_model` right?

ofmla:
> @ofmla you will have to tell us how it fails. It might just be a tolerance issue. Can you paste the whole error trace?
> 
> BTW: that compiler really is old... :)

As I said before, the values for  <Ax,y> are huge, the operators run without problems, no segmentation fault problems. I guess the problem is not related to the compiler as I test different versions of gcc in an older version of Devito (31d17f0) installed on a cluster and all tests passed. I think there is some relation with the problem reported by @speglich  at slack (https://devitocodes.slack.com/archives/C1HSXP6KB/p1585856549285900), as I tested the version 12e980c, which seems to be the last version working for his example and it also worked for me (all tests passed). I tried some newer versions i.e, 1a3c3e9 and ccfc9c5 and get errors in the acoustic adjoint tests for `OT4` and 2D TTI adjoin tests (very different values for `<Ax,y>` and `<x, A^Ty>` in some cases or hug e values for `<Ax,y>` ).   What can I do to help?

examples/seismic/tti/wavesolver.py
```diff
@@ -119,10 +129,69 @@ def forward(self, src=None, rec=None, u=None, v=None, vp=None,
...
+        phi : Function or float, optional
+            The time-constant Azimuth angle (radians).
+
+        Returns
+        -------
+        Adjoint source, wavefield and performance summary.
+        """
+
+        time_order = 2
+        stagg_p = stagg_r = None
```
mloubout, line 174: remove

examples/seismic/tti/wavesolver.py
```diff
@@ -119,10 +129,69 @@ def forward(self, src=None, rec=None, u=None, v=None, vp=None,
...
+        # Source term is read-only, so re-use the default
+        srca = srca or PointSource(name='srca', grid=self.model.grid,
+                                   time_range=self.geometry.time_axis,
+                                   coordinates=self.geometry.src_positions)
+        # Create the wavefield if not provided
+        if p is None:
+            p = TimeFunction(name='p', grid=self.model.grid, staggered=stagg_p,
```
mloubout, line 181: remove staggered

tests/test_adjoint.py
```diff
@@ -59,6 +63,43 @@ def test_adjoint_F(self, mkey, shape, kernel, space_order):
...
+                        coordinates=solver.geometry.src_positions)
+
+        # Run forward and adjoint operators
+        rec, _, _, _ = solver.forward(save=False)
+        solver.adjoint(rec=rec, srca=srca)
+
+        # Adjoint test: Verify <Ax,y> matches  <x, A^Ty> closely
+        term1 = np.dot(srca.data.reshape(-1), solver.geometry.src.data)
+        term2 = norm(rec) ** 2
+        info('<Ax,y>: %f, <x, A^Ty>: %f, difference: %4.4e, ratio: %f'
```
mloubout, line 99: Not sure if same above but description is switched, <Ax, y> is `term2` not `term1`

tests/test_adjoint.py
```diff
@@ -59,6 +63,43 @@ def test_adjoint_F(self, mkey, shape, kernel, space_order):
              % (term1, term2, (term1 - term2)/term1, term1 / term2))
         assert np.isclose((term1 - term2)/term1, 0., atol=1.e-12)
 
+    @pytest.mark.parametrize('mkey, shape, space_order', [
+        # 2D tests with varying space orders
+        ('layers', (30, 35), 8), ('layers', (30, 35), 4),
+        # 3D tests with varying space orders
+        ('layers', (30, 35, 40), 8), ('layers', (30, 35, 40), 4),
+    ])
+    def test_adjoint_tti_F(self, mkey, shape, space_order):
```
mloubout, line 72: This can probably be merged with above test with 
`test_adjoint_f(self, mkey, shape, space_order, setup_func)`
and
solver = `setup_func` ....

ofmla:
Requested changes were made @mloubout.

tests/test_adjoint.py
```diff
@@ -49,15 +70,18 @@ def test_adjoint_F(self, mkey, shape, kernel, space_order):
                         coordinates=solver.geometry.src_positions)
 
         # Run forward and adjoint operators
-        rec, _, _ = solver.forward(save=False)
+        if setup_func.__name__ == 'acoustic_setup':
+            rec, _, _ = solver.forward(save=False)
+        else:
+            rec, _, _, _ = solver.forward(save=False)
```
mloubout, line 76: can do `rec = solver.forward(save=False)[0]` as a one line and avoid if/else

tests/test_adjoint.py
```diff
@@ -39,7 +54,13 @@ def test_adjoint_F(self, mkey, shape, kernel, space_order):
         tn = 500.  # Final time
 
         # Create solver from preset
-        solver = acoustic_setup(shape=shape, spacing=[15. for _ in shape], kernel=kernel,
+        if setup_func.__name__ == 'acoustic_setup':
```
mloubout, line 57: does it actually need the if/else? Only difference is `kernel` in the kwargs but I "think" tti would just ignore it no?

ofmla: I think the if/else cannot be avoided as the kwargs passed in `tti_setup` are also passed to the `AnisotropicWaveSolver` function and the `ForwardOperator` has a kernel parameter set by default to centered, so the kernel is a positional parameter in `ForwardOperator`

tests/test_adjoint.py
ofmla: Done

tests/test_adjoint.py
ofmla: I would be passing twice the kernel parameter in ` ForwardOPerator`

ofmla: Sorry, the solution was in my reply ;)

FabioLuporini:
Thanks for this! Merging...

PR #1213, opened 2020-04-06
Make fd shortcut callable

mloubout:
Fd shortcut are now callable to allow change of fd_order/x0/side


devito/finite_differences/derivative.py
```diff
@@ -73,6 +73,14 @@ class Derivative(sympy.Derivative, Differentiable):
 
     >>> u.dx2
     Derivative(u(x, y), (x, 2))
+
+    Derivative object are also callable to change default setup:
```
FabioLuporini, line 77: probably Function and TimeFunction docstrings to be updated too?

devito/finite_differences/derivative.py
```diff
@@ -141,6 +149,24 @@ def __new__(cls, expr, *dims, **kwargs):
         obj._x0 = kwargs.get('x0', None)
         return obj
 
+    def __call__(self, x0=None, fd_order=None, side=None):
```
FabioLuporini, line 152: this will require some engineering to make it **super quick**

devito/finite_differences/derivative.py
```diff
@@ -141,6 +149,24 @@ def __new__(cls, expr, *dims, **kwargs):
...
+        self._fd_order = fd_order or self._fd_order
```
FabioLuporini, line 153: Actually, just noticed this. State can't and shouldn't change. And in fact, i don't see why you need to change it at this point

devito/finite_differences/derivative.py
FabioLuporini: e.g., use try-catch instead of isinstance below

however, this will need some profiling with cProfile I suspect

devito/finite_differences/tools.py
```diff
@@ -169,20 +189,71 @@ def generate_indices(func, dim, order, side=None, x0=None):
...
+    Returns
+    -------
+    Ordered list of indices
+    """
     shift = 0
+    # Min/max offsets for indices
+    o_start = -order//2
+    o_end = order//2 + 1
+    # Shift if x0 is not on the grid
+    offset_c = 0 if sympify(x0).is_Integer else (dim - x0)/dim.spacing
```
FabioLuporini, line 215: probably best with a try-catch because sympify is slow

we need to make sure these routines are as quick as possible

devito/finite_differences/derivative.py
mloubout: Why? This is  a Derivative feature.

devito/finite_differences/derivative.py
mloubout: >this will require some engineering to make it super quick
This is done a single time, so there won't be much of a hit, and it's a single simple change

devito/finite_differences/tools.py
mloubout: Again, will be done once and only once

devito/finite_differences/derivative.py
mloubout: This is changed here because the rest is ignored if x0 is None.

devito/finite_differences/tools.py
mloubout: And this is only done on `x0` that is a Symbol or a Number, not an expression.

devito/finite_differences/derivative.py
FabioLuporini: because I was still sleepy when I wrote it this morning. Ignore me

devito/finite_differences/derivative.py
```diff
@@ -129,17 +167,37 @@ def __new__(cls, expr, *dims, **kwargs):
...
+        except AttributeError:
+            raise TypeError("Multi-dimensional Derivative, input expected as a dict")
+
+        return self._new_from_self(fd_order=_fd_order, x0=new_x0)
+
+    def _new_from_self(self, **kwargs):
+        _kwargs = {'deriv_order': self.deriv_order, 'fd_order': self.fd_order,
+                   'side': self.side, 'transpose': self.transpose, 'subs': self._subs,
+                   'x0': self.x0, 'preprocessed': True}
+        _kwargs.update(**kwargs)
```
FabioLuporini, line 201: more pythonic (no need for extra support variable) as:

```
kwargs = {......, **kwargs}
```

devito/finite_differences/derivative.py
```diff
@@ -129,17 +169,37 @@ def __new__(cls, expr, *dims, **kwargs):
...
+            _side = side or self._side
+            new_x0 = {self.dims[0]: x0} if x0 is not None else self.x0
+            return self._new_from_self(fd_order=_fd_order, side=_side, x0=new_x0)
+
+        if side is not None:
+            raise TypeError("Side only supported for first order single"
+                            "Dimension derivative such as `.dxl` or .dx(side=left)")
+        # Cross derivative
+        new_x0 = self._x0 or {}
+        _fd_order = self.fd_order._getters
```
FabioLuporini, line 186: watch out: this is changing state because you're *updating* (line below) `self.fd_order._getters` in-place

devito/finite_differences/derivative.py
```diff
@@ -129,17 +169,37 @@ def __new__(cls, expr, *dims, **kwargs):
...
+            _fd_order = fd_order or self._fd_order
+            _side = side or self._side
+            new_x0 = {self.dims[0]: x0} if x0 is not None else self.x0
+            return self._new_from_self(fd_order=_fd_order, side=_side, x0=new_x0)
+
+        if side is not None:
+            raise TypeError("Side only supported for first order single"
+                            "Dimension derivative such as `.dxl` or .dx(side=left)")
+        # Cross derivative
+        new_x0 = self._x0 or {}
```
FabioLuporini, line 185: nitpicking: why `new_x0` and not just `_x0` like all other variables?

devito/finite_differences/derivative.py
```diff
@@ -153,35 +213,37 @@ def _xreplace(self, subs):
...
     def x0(self):
         return self._x0
 
-    @property
+    @cached_property
     def fd_order(self):
         return self._fd_order
 
-    @property
+    @cached_property
```
FabioLuporini, line 234: remove all cached_properties when you're just returning a field (ie the property has 0 cost), because they will be of no help (in fact, they might increase access time and storage requirements!)

devito/finite_differences/derivative.py
mloubout: I prefer it split to have the default and the updated

PR #1217, opened 2020-04-07
Basic support for OpenACC + MPI

FabioLuporini:


devito/core/gpu_openacc.py
```diff
@@ -1,13 +1,20 @@
-from functools import partial
+from functools import partial, singledispatch
 
 import cgen as c
+from sympy import Function
 
 from devito.core.gpu_openmp import (DeviceOpenMPNoopOperator, DeviceOpenMPIteration,
```
ggorman, line 6: Looks a little odd to import from gpu_openmp for gpu_openacc. Might want to refine abstraction in the future.

devito/core/gpu_openmp.py
```diff
@@ -76,6 +77,10 @@ def _map_alloc(cls, f):
         return cls.lang['map-enter-alloc'](f.name, ''.join('[0:%s]' % i
                                                            for i in cls._map_data(f)))
 
+    @classmethod
+    def _map_present(cls, f):
+        raise NotImplementedError
```
ggorman, line 82: On slack, you gave a detailed explanation as to why you had to drop `present`. I think it would be useful to include that rational as a comment here and/or add an issue to GitHub as it is surely going to come up again as we continue to refine/optimize OpenACC support and you don't want to remember/re-explain to others (thinking about hackathons etc).

devito/core/gpu_openacc.py
```diff
@@ -52,6 +86,78 @@ class DeviceOpenACCDataManager(DeviceOpenMPDataManager):
...
+        storage._high_bw_mem[obj] = (None, init, free)
+
+
+@iet_pass
+def initialize(iet, **kwargs):
+    """
+    Initialize the OpenACC environment.
+    """
+
+    @singledispatch
```
mloubout, line 112: Why is this needed?

devito/core/gpu_openmp.py
```diff
@@ -163,18 +172,17 @@ def _map_function_on_high_bw_mem(self, obj, storage, read_only=False):
 
     @iet_pass
     def place_ondevice(self, iet, **kwargs):
-        efuncs = kwargs['efuncs']
-
-        storage = Storage()
 
-        if iet.is_ElementalFunction:
-            return iet, {}
+        @singledispatch
```
mloubout, line 176: needed why?

devito/core/gpu_openmp.py
```diff
@@ -185,13 +193,28 @@ def place_ondevice(self, iet, **kwargs):
...
-        for i in filter_sorted(writes):
-            self._map_function_on_high_bw_mem(i, storage)
-        for i in filter_sorted(reads):
-            self._map_function_on_high_bw_mem(i, storage, read_only=True)
+            # Populate `storage`
+            storage = Storage()
+            for i in filter_sorted(writes):
+                self._map_function_on_high_bw_mem(i, storage)
+            for i in filter_sorted(reads):
+                self._map_function_on_high_bw_mem(i, storage, read_only=True)
```
mloubout, line 201: so what happens if something is a write and a read?

devito/core/gpu_openacc.py
FabioLuporini: There is a TODO a few lines below which says:

> TODO: currently inhereting from the OpenMP Operators. Ideally, we should/could
abstract things away so as to have a separate, language-agnostic superclass

devito/core/gpu_openacc.py
FabioLuporini: all IET passes are applied, individually, to all Callables in the IET. singledispatch is the software engineering pattern that allows you to specialize a pass for a specific type of Callable. In this case, it is used to make the pass such that with all non-Operator Callables it is essentially a no-op

`_initialize` is for initialization of the OpenACC runtime and it should be performed only in the Kernel function

devito/core/gpu_openmp.py
FabioLuporini: same reason as above

devito/core/gpu_openmp.py
FabioLuporini: underneath sets are used and only one declaration is added. This happens in pretty much all tests (`u[t+1, x] = u[t, x]` -- `u` is both a write and a read)

devito/core/gpu_openmp.py
FabioLuporini: good point, I'm adding it

devito/core/gpu_openacc.py
FabioLuporini: let me elaborate:

the reason I use the `singledispatch` is to make sure that that specific IET pass has effect only on the main function, while leaving all other Callables untouched

```
int Kernel(...) {
 ...
  haloupdate(...)
  bf0(...)
}
haloupdate(...) {
  ...
}
bf0(...) {
  ...
}
```

I want the pass to be applied only to
```
int Kernel(...) {
 ...
}
```

PR #1221, opened 2020-04-08
Gpu ci fix test

rhodrin:
Not for merge. Testing that fails in gpu ci are correctly reported.


tests/test_gpu_openmp.py
```diff
@@ -257,4 +257,4 @@ def test_iso_ac(self):
 
         op(time=time_range.num-1, dt=dt)
 
-        assert np.isclose(norm(rec), 490.55, atol=1e-2, rtol=0)
+        assert np.isclose(norm(rec), 491.55, atol=1e-2, rtol=0)
```
FabioLuporini, line 260: this is scary

rhodrin: Hah, I guess I should have put a comment above. This was changed to enforce a fail to ensure fails are being reported correctly :p

Issue #1215, opened 2020-04-07
OOB accesses into Devito generated code


speglich:
This issue occurs in the commit: 79f012799aa32c21e327c848dbeb61e20ab001a8

The code attached makes the Devito compiler generate a code with OOB accesses. 

The loop bounds are not computed correctly by Devito. Basically the div(1/rho grad(p) makes the stencil twice as wide as the space_order and the halo isn’t big enough so end up in an OOB that isn’t detected.

This issue is divided in two mainly steps:
 
* Find why OOB wasn't detected and fix it.

[example.txt](https://github.com/devitocodes/devito/files/4441495/example.txt)

FabioLuporini:
I agree about the first step

I'm not sure about the latter step. I don't think we need to fix boundaries in the generated code, but rather implement stencil auto-wrapping. Let's discuss about this once the first issue is dealt with !

Thanks a lot for this

speglich:
Just for documentation: 

@FabioLuporini said:

I investigated the segfault issue (I didn't mean to overstep you @speglich, but given @OscarM's comments here: https://github.com/devitocodes/devito/pull/1212#issuecomment-611201569 we should fix this ASAP because this appears to be much deeper than expected). Here are my findings:

Mathias, Joao: the reason Devito doesn't automatically shrink the x_m/x_M bounds is that detect_oob is called at the beginning of compilation, well before the OOB is generated
the OOB is caused by what formerly we were calling dse=aggressive mode, that is now default when you run an Operator. Plus, the changes I made to one compilation passes (Mathias: CIRE), are causing the OOB. Still investigating exactly why that is happening.

Mathias: we said we were expecting the "+ 20" offsets in the stencils due to the given space order and taken derivatives, why that is not happening still escapes my mind. The OOB (the "+ 20") is a by-product of a compilation pass, which as I said I'm still investigating.

PR #1224, opened 2020-04-09
Fix OOB access caused by CIRE

FabioLuporini:
patch + test

### DIFF
--- a/devito/operator/operator.py
+++ b/devito/operator/operator.py
@@ -635,7 +635,7 @@ class Operator(Callable):
     # Execution
 
     def __call__(self, **kwargs):
-        self.apply(**kwargs)
+        return self.apply(**kwargs)
 
     def apply(self, **kwargs):
         """
--- a/devito/symbolics/manipulation.py
+++ b/devito/symbolics/manipulation.py
@@ -248,7 +248,7 @@ def xreplace_indices(exprs, mapper, key=None, only_rhs=False):
         handle = [i for i in handle if i.base.label in key]
     elif callable(key):
         handle = [i for i in handle if key(i)]
-    mapper = dict(zip(handle, [uxreplace(i, mapper) for i in handle]))
+    mapper = dict(zip(handle, [i.xreplace(mapper) for i in handle]))
     replaced = [uxreplace(i, mapper) for i in as_tuple(exprs)]
     return replaced if isinstance(exprs, Iterable) else replaced[0]
 
--- a/tests/test_dse.py
+++ b/tests/test_dse.py
@@ -5,8 +5,9 @@ from unittest.mock import patch
 from cached_property import cached_property
 
 from conftest import skipif, EVAL  # noqa
-from devito import (Eq, Inc, Constant, Function, TimeFunction, SparseTimeFunction,  # noqa
-                    Dimension, SubDimension, Grid, Operator, switchconfig, configuration)
+from devito import (NODE, Eq, Inc, Constant, Function, TimeFunction, SparseTimeFunction,  # noqa
+                    Dimension, SubDimension, Grid, Operator, norm, grad, div,
+                    switchconfig, configuration)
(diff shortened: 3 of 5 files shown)


devito/operator/operator.py
```diff
@@ -635,7 +635,7 @@ def cfunction(self):
     # Execution
 
     def __call__(self, **kwargs):
-        self.apply(**kwargs)
+        return self.apply(**kwargs)
```
mloubout, line 638: ?

devito/symbolics/manipulation.py
```diff
@@ -248,7 +248,7 @@ def xreplace_indices(exprs, mapper, key=None, only_rhs=False):
         handle = [i for i in handle if i.base.label in key]
     elif callable(key):
         handle = [i for i in handle if key(i)]
-    mapper = dict(zip(handle, [uxreplace(i, mapper) for i in handle]))
+    mapper = dict(zip(handle, [i.xreplace(mapper) for i in handle]))
```
mloubout, line 251: why back to xreplace?

tests/test_dse.py
```diff
@@ -1134,6 +1135,44 @@ def test_full_shape_big_temporaries(self):
...
+        """
+        Test that the aliases are scheduled correctly when extracted from
+        Clusters whose iteration space is lifted (ie, stamp != 0).
+        """
+        so = 8
+        grid = Grid(shape=(6, 6, 6))
+
+        v = TimeFunction(name="v", grid=grid, space_order=so)
+        v.data_with_halo[:] = 1.
+        p = TimeFunction(name="p", grid=grid, space_order=so, staggered=NODE)
```
rhodrin, line 1150: `staggered=NODE` not needed here?

mloubout: It is needed, all funcs are staggered

rhodrin: Why then does the test pass without it?

mloubout: Because it's a dummy equation for the test, but I prefer to have it as it is because for exampel first order acoustic won't work without it.

devito/operator/operator.py
FabioLuporini: `self.apply` returns the summary ! we've had this bug for like forver -- when using the `op()` shortcut we were systematically returning None instead of the Summary dict. I bumped into it accidentally while writing the test

devito/symbolics/manipulation.py
FabioLuporini: `xreplace` to be used every time you need a re-evaluation of the arguments. Here, we need it (my bad missing this in the previous PR) . Without xreplace, you get indices such as `x + 3 + 4` instead of `x + 7` which makes the generated code correct, but ugly. To reproduce the issue on master, try running `tti_example.py` and you'll see it in some of the temporary expressions

FabioLuporini: so yeah this has nothing to do with the PR per se. I noticed it, and fixed it.

tests/test_dse.py
FabioLuporini: I carved the test out of @speglich 's code, I trimmed down all the unnecessary except for that NODE, which might not make a difference, but won't hurt either. So yeah, this was intentional

devito/operator/operator.py
mloubout: Oooooh good catch

devito/symbolics/manipulation.py
mloubout: Ok

speglich:
I tested here and everything is running smoothly well. So this PR closes the issue # 1215. Great job @FabioLuporini. I regret don't be in touch directly with this issue in the correct time. I contacted Oscar to test the same patch in his code.

speglich:
Closing the issue #1215

PR #1228, opened 2020-04-10
OpenACC-MPI testing in python3

georgebisbas:


.github/workflows/pytest-gpu.yml
```diff
@@ -83,14 +82,26 @@ jobs:
...
+          err=0
+          trap 'err=1' ERR
           # OpenMP offloading
-          DEVITO_ARCH=clang DEVITO_PLATFORM=nvidiaX py.test tests/test_gpu_openmp.py
-          DEVITO_ARCH=clang DEVITO_PLATFORM=nvidiaX py.test examples/seismic/elastic/elastic_example.py
-          DEVITO_ARCH=clang DEVITO_PLATFORM=nvidiaX py.test examples/seismic/viscoelastic/viscoelastic_example.py
+          mpicc --version
+          which mpicc
+          mpirun --version
+          mpiexec --version
```
mloubout, line 92: debug leftover?

tests/conftest.py
```diff
@@ -131,10 +131,10 @@ def parallel(item):
             testname = "%s::%s::%s" % (item.fspath, item.cls.__name__, item.name)
         else:
             testname = "%s::%s" % (item.fspath, item.name)
-        args = ["-n", "1", "python", "-m", "pytest", "--runxfail", "-s",
+        args = ["-n", "1", "python3", "-m", "pytest", "--runxfail", "-s",
                 "-q", testname]
         if nprocs > 1:
-            args.extend([":", "-n", "%d" % (nprocs - 1), "python", "-m", "pytest",
+            args.extend([":", "-n", "%d" % (nprocs - 1), "python3", "-m", "pytest",
```
mloubout, line 137: that's risky, as this will break other peopletest.

.github/workflows/pytest-gpu.yml
georgebisbas: Sure yes. Intentionally left for some testing, will remove.

tests/conftest.py
georgebisbas: Wee, I agree with that. But if I may ask, what are these cases where this will break?

georgebisbas: @mloubout would you prefer that now?

mloubout: Well no, that's still hardcoded and bit dangerous. I would use an env var, like `python = os.getenv('PYTHON', 'python')` so that it's trivial to swtich from one to another depending on system setup.

tests/conftest.py
```diff
@@ -125,16 +126,17 @@ def parallel(item):
...
 
+        pyversion = sys.executable
         # Only spew tracebacks on rank 0.
         # Run xfailing tests to ensure that errors are reported to calling process
         if item.cls is not None:
             testname = "%s::%s::%s" % (item.fspath, item.cls.__name__, item.name)
         else:
             testname = "%s::%s" % (item.fspath, item.name)
-        args = ["-n", "1", "python", "-m", "pytest", "--runxfail", "-s",
+        args = ["-n", "1", "%s" % pyversion, "-m", "pytest", "--runxfail", "-s",
```
FabioLuporini, line 136: why

```
"%s" % pyversion
```

and not just

```
pyversion
```

?

georgebisbas: @FabioLuporini I fixed this. I forgot that there from a previous attempt. Thanks. As I can see GPU-test passes. Feel free to merge or review again after it gets green.

rhodrin:
Looks good. Just made a test vm from the new image and everything worked as expected.

rhodrin:
@georgebisbas Oh, and remember to clean up CI-gpu after this goes in.

FabioLuporini:
Merged! thanks

PR #1229, opened 2020-04-10
indexify fix

mloubout:
fix subsampled dim FD.

A little bug was introduce by xreplace in indexify as `(-2*dt).xreplace({2*dt: 1}) = -2*dt` as xreplace doesn't recognise it with - sign, which breaks FD with subsampling. Wasn't caught in test because it's forward FD (only 2*dt, not -2*dt), so added a test for it too.

### DIFF
--- a/devito/operator/operator.py
+++ b/devito/operator/operator.py
@@ -306,7 +306,7 @@ class Operator(Callable):
                 dimension_map = {}
 
             # Handle Functions (typical case)
-            mapper = {f: f.indexify(w_shift=True, extra_subs=dimension_map)
+            mapper = {f: f.indexify(lshift=True, subs=dimension_map)
                       for f in retrieve_functions(expr)}
 
             # Handle Indexeds (from index notation)
--- a/devito/operator/operator.py
+++ b/devito/operator/operator.py
@@ -305,26 +305,9 @@ class Operator(Callable):
             else:
                 dimension_map = {}
 
-            mapper = {}
-
             # Handle Functions (typical case)
-            for f in retrieve_functions(expr):
-                # Get spacing symbols for replacement
-                spacings = [i.spacing for i in f.dimensions]
-
-                # Only keep the ones used as indices
-                spacings = [s for i, s in enumerate(spacings)
-                            if s.free_symbols.intersection(f.args[i].free_symbols)]
-
-                # Substitution for each index
-                subs = {**{s: 1 for s in spacings}, **dimension_map}
-
-                # Introduce shifting to align with the computational domain,
-                # and apply substitutions
-                indices = [(a - i + o).xreplace(subs)
-                           for a, i, o in zip(f.args, f.origin, f._size_nodomain.left)]
-
-                mapper[f] = f.indexed[indices]
+            mapper = {f: f.indexify(w_shift=True, extra_subs=dimension_map)
+                      for f in retrieve_functions(expr)}
 
             # Handle Indexeds (from index notation)
             for i in retrieve_indexed(expr):
--- a/devito/types/basic.py
+++ b/devito/types/basic.py
@@ -919,17 +919,17 @@ class AbstractFunction(sympy.Function, Basic, Cached, Pickable, Evaluable):
         """
         return default_allocator().guaranteed_alignment
 
-    def indexify(self, indices=None, w_shift=False, extra_subs=None):
+    def indexify(self, indices=None, lshift=False, subs=None):
         """Create a types.Indexed from the current object."""
         if indices is not None:
             return Indexed(self.indexed, *indices)
 
         # Substitution for each index (spacing only used in own dimension)
-        extra_subs = extra_subs or {}
-        subs = [{**d._indexify_map, **extra_subs} for d in self.dimensions]
+        subs = subs or {}
+        subs = [{**{d.spacing: 1, -d.spacing: -1}, **subs} for d in self.dimensions]
 
         # Add halo shift
-        shift = self._size_nodomain.left if w_shift else tuple([0]*len(self.dimensions))
+        shift = self._size_nodomain.left if lshift else tuple([0]*len(self.dimensions))
         # Indices after substitutions
         indices = [(a - o + f).xreplace(s) for a, o, f, s in
                    zip(self.args, self.origin, shift, subs)]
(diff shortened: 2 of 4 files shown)


devito/types/basic.py
```diff
@@ -932,7 +932,7 @@ def indexify(self, indices=None):
                     if s.free_symbols.intersection(self.args[i].free_symbols)]
 
         # Substitution for each index
-        subs = {s: 1 for s in spacings}
+        subs = {a*s: a for s in spacings for a in [-1, 1]}
```
FabioLuporini, line 935: this dictionary will be reconstructed zilions of times in non-trivial pdes -- once for each Indexed. And in 3D for a TimeFunction, it's 8 sympy evaluations per Indexed. So I'm again a little concerned about speed. In my timings, `indexify` was indeed taking up some time. Can we build it once and for all ? 

```
@cached_property
def _spacing_subs(self):
    spacings = [i.spacing for i in self.dimensions]
    subs = {a*s: a for s in spacings for a in [-1, 1]}
    return subs
```
?

oc, to be retrieved from the mother function, ie `self.function._spacing_subs`

and then what is left to do in indexify are the filtering (`# Only keep the ones used as indices.`) and the actual substitutions

FabioLuporini: I think any feedback welcome. I appreciate this is a minor and it should just go in and full stop, but many minor incautious patches will eventually kill off the compilation speeds improvements for which I spilled so much blood

mloubout: All of this was actually almost never called because there was some duplicate of it in `operator`.

devito/operator/operator.py
```diff
@@ -305,26 +305,8 @@ def _lower_exprs(cls, expressions, **kwargs):
             else:
                 dimension_map = {}
 
-            mapper = {}
-
             # Handle Functions (typical case)
-            for f in retrieve_functions(expr):
```
mloubout, line 311: Note: These lines were literally copy-pasted from `basic` and bypassing `indexify` in most cases, so fixed indexify and removed.

devito/types/dimension.py
```diff
@@ -742,6 +746,10 @@ def free_symbols(self):
             retval |= self.condition.free_symbols
         return retval
 
+    @cached_property
+    def _indexify_map(self):
+        return {self.spacing: 1, -self.spacing: -1}
```
FabioLuporini, line 751: why is this different than the superclass'?

devito/operator/operator.py
```diff
@@ -305,26 +305,9 @@ def _lower_exprs(cls, expressions, **kwargs):
...
-                # Substitution for each index
-                subs = {**{s: 1 for s in spacings}, **dimension_map}
-
-                # Introduce shifting to align with the computational domain,
-                # and apply substitutions
-                indices = [(a - i + o).xreplace(subs)
-                           for a, i, o in zip(f.args, f.origin, f._size_nodomain.left)]
-
-                mapper[f] = f.indexed[indices]
+            mapper = {f: f.indexify(w_shift=True, extra_subs=dimension_map)
```
FabioLuporini, line 309: The reason this was "duplicated" is that the Operator version was optimized for speed, while `indexify` was not, but I couldn't just drop it for backwards compatibility

FabioLuporini:
@mloubout I tried dropping the `_indexify_map` and replacing it with

```
subs = [{**{d.spacing: 1, -d.spacing: -1}, **extra_subs} for d in self.dimensions]
```

it just runs as fast. Note that there's a subtle difference w.r.t. the first version of this PR, when we had a MUL (`__mul__`), not a SUB (`__sub__`), in the construction of the dict, which was inherently slower (yes, SymPy handles them differently AFAIR).

So, I take my initial suggestion back, and rather advise to stick to the simpler version (no new properties, just compute the dict on the fly with `self.spacing` and `-self.spacing` as per above)

devito/types/dimension.py
mloubout: Super-class is `{self.spacing: 1}`

mloubout:
>So, I take my initial suggestion back, and rather advise to stick to the simpler version (no new properties

I actually like the new property so that dimensions know how to be used as indexes rather than only in Basic. This way know for just the dimension anywhere.

devito/types/basic.py
```diff
@@ -773,7 +773,7 @@ def symbolic_shape(self):
                     for i, j, k in zip(domain, halo, padding))
         return DimensionTuple(*ret, getters=self.dimensions)
 
-    @property
+    @cached_property
```
rhodrin, line 776: Why wasn't this cached previously?

devito/types/basic.py
```diff
@@ -919,25 +919,22 @@ def _data_alignment(self):
         """
         return default_allocator().guaranteed_alignment
 
-    def indexify(self, indices=None):
+    def indexify(self, indices=None, w_shift=False, extra_subs=None):
```
rhodrin, line 922: Nitpicking, but `additional_subs` would be better than `extra_subs`?

devito/types/dimension.py
```diff
@@ -305,6 +305,10 @@ def __new__(cls, *args, **kwargs):
     def __init_finalize__(self, name, spacing=None):
         self._spacing = spacing or Scalar(name='h_%s' % name, is_const=True)
 
+    @cached_property
+    def _indexify_map(self):
```
rhodrin, line 309: Initially, the design of this and the property appearing in `DefaultDimension` seem a bit weird to me. Will take a closer look though.

devito/types/basic.py
mloubout: Not sure why, but don't see why it shouldn't be, probably just not caught

devito/types/basic.py
mloubout: Doesn't fit in a line, why I made it shorter, and stil lthinks it explicit enough

devito/types/dimension.py
mloubout: This property basically say "what should I do when I'm used in an index" which I think beongs here not in a dict built in indexify.

devito/types/basic.py
FabioLuporini: how about just "subs" (and "shift") ?

PR #1232, opened 2020-04-14
Self-hosted CI-mpi

rhodrin:
Update CI-mpi workflow to use our self-hosted runners.


rhodrin:
MPI runs took ages on this machine - will switch back to another machine and add a few more runners. Should be ready later today.

rhodrin:
Temporarily closing for some testing.

PR #1234, opened 2020-04-15
Self-hosted CI-mpi

rhodrin:
Update CI-mpi workflow to use our self-hosted runners.


rhodrin:
> Only 33 mins for MPI?

Yep. Overlocked 4vm skylakes.

PR #1235, opened 2020-04-16
Fix scheduling of CIRE-detected aliasing expressions

FabioLuporini:
This should bring #1212 back to green (@ofmla )

### DIFF
--- a/devito/core/cpu.py
+++ b/devito/core/cpu.py
@@ -4,8 +4,7 @@ from devito.core.operator import OperatorCore
 from devito.exceptions import InvalidOperator
 from devito.ir.clusters import Toposort
 from devito.passes.clusters import (Blocking, Lift, cire, cse, eliminate_arrays,
-                                    extract_increments, factorize, fuse, optimize_pows,
-                                    scalarize)
+                                    extract_increments, factorize, fuse, optimize_pows)
 from devito.passes.iet import (DataManager, Ompizer, avoid_denormals, mpiize,
                                optimize_halospots, loop_wrapping, hoist_prodders,
                                relax_incr_dimensions)
@@ -111,7 +110,6 @@ class CPU64Operator(CPU64NoopOperator):
         # turn may enable further optimizations
         clusters = fuse(clusters)
         clusters = eliminate_arrays(clusters, template)
-        clusters = scalarize(clusters, template)
 
         return clusters
 
@@ -225,7 +223,6 @@ class Intel64FSGOperator(Intel64Operator):
         # turn may enable further optimizations
         clusters = fuse(clusters)
         clusters = eliminate_arrays(clusters, template)
-        clusters = scalarize(clusters, template)
 
         # Blocking to improve data locality
         clusters = Blocking(options).process(clusters)
--- a/devito/finite_differences/differentiable.py
+++ b/devito/finite_differences/differentiable.py
@@ -369,6 +369,31 @@ class diffify(object):
         return obj.__class__
 
 
+def diff2sympy(expr):
+    """
+    Translate a Differentiable expression into a SymPy expression.
+    """
+
+    def _diff2sympy(obj):
+        flag = False
+        args = []
+        for a in obj.args:
+            ax, af = _diff2sympy(a)
+            args.append(ax)
+            flag |= af
+        if isinstance(obj, DifferentiableOp):
+            for cls in obj.__class__.mro()[1:]:
+                if obj.__class__.__name__ == cls.__name__:
+                    return cls(*args, evaluate=False), True
+            assert False
+        elif flag:
+            return obj.func(*args, evaluate=False), True
+        else:
+            return obj, False
+
+    return _diff2sympy(expr)[0]
+
+
 # Make sure `sympy.evalf` knows how to evaluate the inherited classes
 # Without these, `evalf` would rely on a much slower, much more generic, and
 # thus much more time-inefficient fallback routine. This would hit us
--- a/devito/finite_differences/differentiable.py
+++ b/devito/finite_differences/differentiable.py
@@ -254,6 +254,8 @@ class Differentiable(sympy.Expr, Evaluable):
 
 class DifferentiableOp(Differentiable):
 
+    __sympy_class__ = None
+
     def __new__(cls, *args, **kwargs):
         obj = cls.__base__.__new__(cls, *args, **kwargs)
 
@@ -294,18 +296,22 @@ class DifferentiableOp(Differentiable):
 
 
 class Add(DifferentiableOp, sympy.Add):
+    __sympy_class__ = sympy.Add
     __new__ = DifferentiableOp.__new__
 
 
 class Mul(DifferentiableOp, sympy.Mul):
+    __sympy_class__ = sympy.Mul
     __new__ = DifferentiableOp.__new__
 
 
 class Pow(DifferentiableOp, sympy.Pow):
+    __sympy_class__ = sympy.Pow
     __new__ = DifferentiableOp.__new__
 
 
 class Mod(DifferentiableOp, sympy.Mod):
+    __sympy_class__ = sympy.Mod
     __new__ = DifferentiableOp.__new__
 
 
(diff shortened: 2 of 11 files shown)


devito/core/cpu.py
```diff
@@ -111,7 +110,6 @@ def _specialize_clusters(cls, clusters, **kwargs):
         # turn may enable further optimizations
         clusters = fuse(clusters)
         clusters = eliminate_arrays(clusters, template)
-        clusters = scalarize(clusters, template)
```
FabioLuporini, line 114: note for reviewers: this pass now unnecessary, as everything is done naturally inside cire()

devito/ir/equations/equation.py
```diff
@@ -147,8 +148,11 @@ def __new__(cls, *args, **kwargs):
                  for k, v in mapper.items() if k}
         dspace = DataSpace(dintervals, parts)
 
+        # Lower all Differentiable operations into SymPy operations
+        rhs = diff2sympy(expr.rhs)
```
FabioLuporini, line 152: note for reviewers: avoids minor redundant expressions in the generated code, such as `a + b` and `a + b` where the former was of type Differentiable and the latter of type sympy. And even though Differentiable inherits from sympy, it would still be the case that `a + b != a + b` . This was BTW causing a minor op-count discrepancy in one of the tests

devito/passes/clusters/factorization.py
```diff
@@ -64,8 +64,8 @@ def collect_const(expr):
 
     terms = []
     for k, v in inverse_mapper.items():
-        if len(v) == 1:
-            # We can actually evaluate everything to avoid, e.g., (-1)*a
+        if len(v) == 1 and not v[0].is_Add:
```
FabioLuporini, line 67: note for reviewers: minor bug fix orthogonal to the main purpose of the PR

georgebisbas: Out of curiosity. What was the bug in this case?

devito/passes/clusters/aliases.py
```diff
@@ -688,7 +696,7 @@ def get(self, key):
                 return aliaseds
         return []
 
-    def schedule(self, ispace):
+    def iter(self, ispace):
```
georgebisbas, line 699: What about this name change?

devito/passes/clusters/aliases.py
```diff
@@ -433,6 +441,9 @@ def rebuild(cluster, others, aliases, subs):
     return cluster.rebuild(exprs=exprs, ispace=ispace, dspace=dspace)
 
 
+# Utilities
+
+
```
georgebisbas, line 446: spaces?

FabioLuporini: was like that before, I don't mind them to be honest, it denotes a different section of the file

devito/passes/clusters/aliases.py
FabioLuporini: it wasn't really "scheduling" before

devito/passes/clusters/factorization.py
FabioLuporini: breaking a pre-factorized expression

```a(b+c)```

becoming

```ab + ac```

File: devito/finite_differences/differentiable.py [hunk 6]
```diff
@@ -369,6 +369,31 @@ def _(obj):
...
+    """
+
+    def _diff2sympy(obj):
+        flag = False
+        args = []
+        for a in obj.args:
+            ax, af = _diff2sympy(a)
+            args.append(ax)
+            flag |= af
+        if isinstance(obj, DifferentiableOp):
```
mloubout, line 384: Isn't it "easier" to just have a `cls._parent`?

devito/ir/equations/equation.py
mloubout: Not done on lhs?

devito/passes/clusters/aliases.py
```diff
@@ -88,42 +88,50 @@ def cire(cluster, template, mode, options, platform):
...
-        # Collection rule
+        return extractor, model, ignore_collected, selector
+
+    def callbacks_sops(context, n):
+        # The `depth` determines "how big" the extracted sum-of-products will be.
+        # We observe that in typical FD codes:
+        #   add(mul, mul, ...) -> stems from first order derivative
+        #   add(mul(add(mul, mul, ...), ...), ...) -> stems from second order derivative
+        # To catch the former, we would need `depth=1`; for the latter, `depth=3`
+        depth = 2*(n + 1) - 1
```
mloubout, line 104: 2 * n + 1

tests/test_dse.py
```diff
@@ -1171,7 +1171,65 @@ def test_extraction_from_lifted_ispace(self):
...
+        # Equation
+        eqn = [Eq(r.backward, Gzz)]
+
+        op0 = Operator(eqn, subs=grid.spacing_map, opt=('noop', {'openmp': True}))
+        op1 = Operator(eqn, subs=grid.spacing_map, opt=('advanced', {'openmp': True}))
+
+        # Check numerical output
+        op0(time_M=1)
+        exp_norm_r = norm(r)
+        r.data_with_halo[:] = 0.5
```
mloubout, line 1226: so `r` is unchanged after op()?

devito/finite_differences/differentiable.py
FabioLuporini: it's not a direct parent due to multiple inheritance. I could add an extra field... but I wouldn't think it'd make much of a difference

devito/ir/equations/equation.py
FabioLuporini: at this point the lhs is already a `types/...` so it's fine

tests/test_dse.py
FabioLuporini: what do you mean? `op` computes `r`, that's why I'm re-initializing it to its original value, namely 0.5

mloubout: Oh right ok sorry missread that. Shouldn't it run for mre than just one time step to make sure stuff don't blow up?

devito/passes/clusters/aliases.py
```diff
@@ -133,7 +141,7 @@ def extractor(context):
         chosen, others = choose(exprs, aliases, selector)
         if not chosen:
             # Do not waste time
-            break
+            continue
```
mloubout, line 144: what does this change? I get it should be continue and don't want to exit the loop but wondering why wasn't breaking something before.

devito/finite_differences/differentiable.py
mloubout: Hum ok. Was just wondering how expensive it is. I guess it's easier to add extra sympy classes like that.

devito/passes/clusters/aliases.py
FabioLuporini: no it's just that previously there was no point in keep going, while now the logic is different, and if the first round goes bad there's still hope for the second one

IOW, the previous version of the algo was acting constructively -- small temporaries first, then bigger ones. The newer version acts speculatively -- first it search for big ones, then smaller ones. This, for various reasons, makes everything so much simpler, while killing the OOB bug in an extremely natural way

tests/test_dse.py
FabioLuporini: Well I'm sure this MFE is unstable so if we run for more it might indeed blow up. I'm totally confident that running for two timesteps here is more than enough and fits our needs. If something were screwed at this point, it would be screwed in such a way that you'd have a spike in the computed values (due to reading from OOB or un-initialized data)

devito/finite_differences/differentiable.py
FabioLuporini: I think you're right, speed may be a concern. I'm changing this now to avoid the `isinstance`

devito/passes/clusters/aliases.py
mloubout: ok

tests/test_dse.py
mloubout: Sounds abut right 👍

devito/finite_differences/differentiable.py
FabioLuporini: done

PR #1237, opened 2020-04-17
Frozendict to immutabledict

jkwashbourne:
fixes deprecation for frozendict, an abandoned project forked to immutabledict


jkwashbourne:
broken, branched from a branch instead of master

PR #1238, opened 2020-04-17
frozendict --> immutabledict

jkwashbourne:
frozendict deprecated, project abandoned
immutabledict forked from frozendict to replace

### DIFF
--- a/devito/ir/clusters/cluster.py
+++ b/devito/ir/clusters/cluster.py
@@ -2,7 +2,7 @@ from itertools import chain
 
 import numpy as np
 from cached_property import cached_property
-from frozendict import frozendict
+from immutabledict import immutabledict
 
 from devito.ir.equations import ClusterizedEq
 from devito.ir.support import (IterationSpace, DataSpace, Scope, detect_io,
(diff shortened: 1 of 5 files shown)


.github/workflows/pytest-gpu.yml
```diff
@@ -81,6 +81,8 @@ jobs:
               git fetch
               git checkout ${{ steps.extract_branch.outputs.branch }}
           fi
+          # Ensure any new requirements are installed
```
georgebisbas, line 84: This is a bit of a hack as we actually need to tweak the image used for testing. It is good for us to think how we treat these cases as a new image will be needed every time we update reqs.

rhodrin: It's not hack, it's just ensuring the correct requirements are installed if they get changed. Saying it's a hack is a bit like saying 'git pull' is a hack.

rhodrin: At some point yes, but only when an update occurs that requires to modify something fundamental on the image.

rhodrin:
@jkwashbourne This will now need a rebase owing to the changes to CI-gpu.

FabioLuporini:
closing as superseded by #1278 

we agreed (offline) that the way to go is to rather take control of `frozendict` , which is a tiny data structure, rather than keeping it as a dependence

PR #1241, opened 2020-04-20
CI examples update

mloubout:
Small update to CI due to growth and new contribution. All `example_*, *_example` are scanned for tesgt and all run so that yaml doesn't need update for new example (unless want to add example run).


.github/workflows/examples.yml
```diff
@@ -60,13 +64,8 @@ jobs:
...
       run: |
-        py.test examples/seismic/elastic/elastic_example.py
-        py.test examples/seismic/viscoelastic/viscoelastic_example.py
-
-    - name: Seismic viscoacoustic examples
-      shell: bash -l {0}
-      run: |
-        py.test examples/seismic/viscoacoustic/viscoacoustic_example.py
+        python examples/seismic/elastic/elastic_example.py
+        python examples/seismic/viscoelastic/viscoelastic_example.py
```
georgebisbas, line 68: This is a big expensive to execute compared to py.test.

mloubout: This CI runs in ~8min, still way faster than pytest so doesn't really matter

georgebisbas: ok, then. I am fine anyway.

FabioLuporini: are errors caught as expected even w/o pytest ? 

pytest was gonna run `test_...` which has asserts inside

.github/workflows/examples.yml
```diff
@@ -60,13 +64,8 @@ jobs:
     - name: Seismic elastic examples
       shell: bash -l {0}
       run: |
-        py.test examples/seismic/elastic/elastic_example.py
-        py.test examples/seismic/viscoelastic/viscoelastic_example.py
-
-    - name: Seismic viscoacoustic examples
-      shell: bash -l {0}
-      run: |
-        py.test examples/seismic/viscoacoustic/viscoacoustic_example.py
```
FabioLuporini, line 69: Dropping because... ?

.github/workflows/examples.yml
mloubout: All tests are run above......

.github/workflows/examples.yml
mloubout: All tests are run above......

.github/workflows/examples.yml
FabioLuporini: if nothing can fail, why are we running these? for segfaults?

.github/workflows/examples.yml
FabioLuporini: right, I guess I'm an idiot again. So the py.test command searches for tests recursively in all subdirectories in all files named `*_example.py example_*.py`?

mloubout: yep

.github/workflows/examples.yml
mloubout: Yeah these are 3D test is for 2D so making sure 3D doesn't error

PR #1243, opened 2020-04-21
self-hosted gpu runners

rhodrin:
Move gpu testing over to self-hosted runners.

~Currently testing. Not yet for merge.~


.github/workflows/pytest-gpu.yml
```diff
@@ -15,85 +11,31 @@ on:
...
-        -adminPassword ${{ secrets.CI_GPU_VM_ADMIN_PASSWORD }}
-
-    - name: set host
-      run: echo ::set-output name=action_host::$(az vm show -d -g ${{ secrets.CI_GPU_RESOURCE_GROUP }} -n gpu_ci_vm_${{ steps.extract_branch.outputs.branch }} --query publicIps -o tsv)
-      id: host
-
-    - name: execute tests on VM
-      uses: fifsky/ssh-action@master
+        # OpenMP5
+        DEVITO_ARCH=clang DEVITO_PLATFORM=nvidiaX pytest --cov --cov-config=.coveragerc --cov-report=xml tests/test_gpu_openmp.py
```
mloubout, line 30: Now that it's a normal action, can these ENV var be set outside of run? Or as a mtrix or something?

rhodrin: Yes, should be able to achieve something like that.

.github/workflows/pytest-gpu.yml
```diff
@@ -15,85 +11,56 @@ on:
...
+      # Prevent all build to stop if a single one fails
+      fail-fast: false
 
-  # Deploy VM in Azure
-  deploy-vm:
-    runs-on: ubuntu-latest
+      matrix:
+        name: [
+          pytest-gpu-omp,
+          pytest-gpu-omp
```
mloubout, line 30: pytest-gpu-acc

.github/workflows/pytest-gpu.yml
```diff
@@ -15,85 +11,56 @@ on:
...
-              git rev-parse --verify --quiet ${{ steps.extract_branch.outputs.branch }}
-              if [ $? -eq 0 ]; then git branch -D ${{ steps.extract_branch.outputs.branch }}; fi
-              git fetch
-              git checkout ${{ steps.extract_branch.outputs.branch }}
-          fi
-          # Set error trap to catch any failing test
-          err=0
-          trap 'err=1' ERR
-          # OpenMP offloading
-          pgaccelinfo
```
georgebisbas, line 88: can we keep this?

.github/workflows/pytest-gpu.yml
rhodrin: Ta. Fixed.

.github/workflows/pytest-gpu.yml
rhodrin: Yeah, probably should. It's added back in now.

PR #1244, opened 2020-04-22
Fix incorrect Cluster fusion (aka loop fusion)

### LINKED ISSUE #1130: Incorrectly fused loops in generated ccode for leapfrog schemes using subdomains
MFE and generated ccode here (comments in MFE):
https://drive.google.com/open?id=1a-hiKnHZ-cmc-wTev5xq5agi0yuEJLWz

With the single operator, the 'us' field in the example is positively offset in the right subdomain. All fields should be zero.

Looking at the generated ccode, it looks like two sections which should be in their own loops end up in a single loop for some reason (ccode_combined_operator.c, lines 72 and 73). The single lines can be found at ccode_subdomain_split_operator, lines 52 and 70.

Since the fields using subdomains and those without are unrelated, splitting the operator into a "subdomains" and a "no subdomains" operator should not change the inner loops (or the results in the MFE).

If you do so, "subdomains" and "no subdomains" have 4 and 2 innermost loops respectively. With the combined operator, you get 5 innermost loops (3 and 2 loops each respectively). The fields with subdomains have two of their loops fused.

FabioLuporini:
fixes #1130


devito/ir/clusters/algorithms.py
```diff
@@ -35,137 +31,6 @@ def clusterize(exprs):
     return ClusterGroup(clusters)
 
 
-class Toposort(Queue):
```
FabioLuporini, line 38: note for reviewers: topological sorting is now part of fusion (ie, performed inside `fuse()`)

devito/core/cpu.py
```diff
@@ -203,8 +201,7 @@ def _specialize_clusters(cls, clusters, **kwargs):
         template = lambda: "r%d" % counter()
 
         # Toposort+Fusion (the former to expose more fusion opportunities)
-        clusters = Toposort().process(clusters)
-        clusters = fuse(clusters)
+        clusters = fuse(clusters, toposort=True)
```
mloubout, line 204: why not directly `topofuse`?

devito/passes/clusters/misc.py
```diff
@@ -1,10 +1,11 @@
+from collections import Counter
```
mloubout, line 1: can we come up with  a better name than `misc` like manipulation

devito/passes/clusters/misc.py
```diff
@@ -72,30 +73,135 @@ def callback(self, clusters, prefix):
...
+    Fuse Clusters with compatible IterationSpace.
     """
-    key = lambda c: (set(c.itintervals), c.guards)
 
-    processed = []
-    for k, g in groupby(clusters, key=key):
-        maybe_fusible = list(g)
+    def __init__(self, toposort):
+        super(Fusion, self).__init__()
+        self.toposort = toposort
```
mloubout, line 84: private `_`?

mloubout: Hum method already call like that ok

devito/passes/clusters/misc.py
```diff
@@ -72,30 +73,135 @@ def callback(self, clusters, prefix):
         return lifted + processed
 
 
-@timed_pass()
-def fuse(clusters):
+class Fusion(Queue):
```
mloubout, line 76: Is making a class really necessary, seems to be "single use" class

devito/core/cpu.py
FabioLuporini: because `fuse` is called later when there's no need for topo-sorting

devito/passes/clusters/misc.py
FabioLuporini: yes, because it's a subclass of `Queue` which is in practice an implicit-tree-visitor. `Fusion` is performing an implicit-tree visit over a list of Clusters

devito/passes/clusters/misc.py
FabioLuporini: yeah can make private but this is an inside class in the guts of the compiler, so possible sure but IMHO not really necessary

devito/passes/clusters/misc.py
FabioLuporini: we could, but all files in this subdirectory contain "manipulation" routines, so it's a bit odd?

mloubout: True, just find misc bit too vague, food for thought

examples/cfd/01_convection.ipynb
```diff
@@ -472,6 +471,17 @@
       "    /* End section1 */\n",
       "    gettimeofday(&end_section1, NULL);\n",
       "    timers->section1 += (double)(end_section1.tv_sec-start_section1.tv_sec)+(double)(end_section1.tv_usec-start_section1.tv_usec)/1000000;\n",
+      "    struct timeval start_section2, end_section2;\n",
```
georgebisbas, line 474: So here we have a new section for BC?

FabioLuporini: loop fusion acting more conservatively => less fusion => one extra section

FabioLuporini:
Merged.

PR #1247, opened 2020-04-23
FD for composite staggered expr

### LINKED ISSUE #1248: Handle expressions that involve multiple Functions with different staggering
Hi, we found something that could not be working as expected. 

We started from the code below, that have different outputs for the same input. The difference between equations is the integer to float representation of the number 1. 
```python
from devito import Grid, VectorTimeFunction, TimeFunction, NODE, Eq, div, Operator

grid = Grid(shape=(5,5))

v = VectorTimeFunction(name="v", grid=grid, time_order=1, space_order=4)
p1 = TimeFunction(name="p1", grid=grid, time_order=1, space_order=4, staggered=NODE)
p2 = TimeFunction(name="p2", grid=grid, time_order=1, space_order=4, staggered=NODE)

v[0].data[:] = 1.
v[1].data[:] = 1.

eq1 = Eq(p1, div(1*v))
eq2 = Eq(p2, div(1.*v))

op1 = Operator([eq1])
op2 = Operator([eq2])

op1.apply(time_M=0)
op2.apply(time_M=0)

print("---------------------------------------------------------------")
print(">> p1:")
print(p1.data)
print("---------------------------------------------------------------")
print(">> p2:")
print(p2.data)
print("---------------------------------------------------------------")
print(">> difference:")
print(p1.data - p2.data)

```

The first thing that I noticed is that the code generated is different, this difference are in the indices computed. 

The indices computed for integer:
``
(x - 3*h_x/2, x - h_x/2, x + h_x/2, x + 3*h_x/2)
``
The indices computed for float:
``
(x - 2*h_x, x - h_x, x, x + h_x, x + 2*h_x)
``
So this was where I started my investigation. I stopped the execution of each equation in the function **generate_indices_staggered**: (tools.py, line: 236) and noticed that the **func.indices_ref** are different for both equations.

This information threw me to the top of Devito and I understood that this issue is not necessary connected with the div operator but in the evaluation between a scalar and a function. To obtain the answer I developed a new example:
```python
  grid = Grid(shape=(5,5))

  v = VectorTimeFunction(name="v", grid=grid, time_order=1, space_order=4)
  v2 = VectorTimeFunction(name="v2", grid=grid, time_order=1, space_order=4)
  v3= VectorTimeFunction(name="v3", grid=grid, time_order=1, space_order=4)
  p1 = TimeFunction(name="p1", grid=grid, time_order=1, space_order=4, staggered=NODE)
  p2 = TimeFunction(name="p2", grid=grid, time_order=1, space_order=4, staggered=NODE)

  v[0].data[:] = 1.
  v[1].data[:] = 1.

  eq1 = Eq(v2, 1*v)
  eq2 = Eq(v3, 1.*v)

  op = Operator([eq1, eq2])

  op.apply(time_M=0)

```

In this time the difference between results don’t exist but the **indices_ref** are different to both equations. 

Continuing with my investigation I stopped the execution in the function **_eval_scalar_mul** (tensor.py, line: 484), where mat object is computed as:

``
	mat = [a*other for a in self._mat]
``
And other is a scalar. Here I perceived that the mat object computed is different for each equation, the indices_ref are different, I found:

For integer: 
``
indices_ref = (t, x + h_x/2, y)
``
For float:  
``
indices_ref = (t, x, y)
``

So I dived into the the multiplication operation again and I probably found where the difference occurs (differentiable.py, line 255):
```python
class DifferentiableOp(Differentiable):

  __sympy_class__ = None

    obj = cls.__base__.__new__(cls, *args, **kwargs)

    # Unfortunately SymPy may build new sympy.core objects (e.g., sympy.Add),
    # so here we have to rebuild them as devito.core objects
    if kwargs.get('evaluate', True):
      obj = diffify(obj)

    return obj
```

The evaluation by __base__ (If I understood well, in this moment,  base is sympy.core.mul.Mul) create different objects, with different indices_ref for different types of “others”.

I hope that this investigation could contribute in some way for the solution of this issue.

mloubout:
Fixes handling of expression with multiple Function and multiple staggering.

- Add are split so that FD handles each staggering properly
- Mul are "evaluated" at highest priority staggering before FD so that the expression has single indices

Important note. I had to make a choice on the priority levels of each staggering that is 
`param < not staggered, NODE < staggered`

This example from slack now produce same output for bot operators
```
import numpy as np
from devito import *

grid = Grid(shape=(5,5))

v = VectorTimeFunction(name="v", grid=grid, time_order=1, space_order=4)
p1 = TimeFunction(name="p1", grid=grid, time_order=1, space_order=4, staggered=NODE)
p2 = TimeFunction(name="p2", grid=grid, time_order=1, space_order=4, staggered=NODE)

v[0].data[:] = 5.
v[1].data[:] = 5.

den = np.zeros((5,5), dtype=np.float32)
den[:] = 2.

A = Function(name="A", grid=grid, space_order=4, staggred=NODE)

initialize_function(A, 1./den, 0, dtype=np.float32)

av = VectorTimeFunction(name="av", grid=grid, time_order=1, space_order=4)

test =  div(A*v).args[0]._eval_at(p2).evaluate 

eq1 = Eq(av, A*v)
eq2 = Eq(p1, div(av))

op = Operator([eq1, eq2])
op.apply(time_M=0)

eq3 = Eq(p2, div(A*v))
op2 = Operator([eq3])
op2.apply(time_M=0)

print(p1.data)
print(p2.data)
```

fixes #1248


devito/finite_differences/derivative.py
```diff
@@ -267,7 +268,14 @@ def _eval_at(self, func):
         setup where one could have Eq(u(x + h_x/2), v(x).dx)) in which case v(x).dx
         has to be computed at x=x + h_x/2.
         """
-        x0 = dict(zip(func.dimensions, func.indices_ref))
+        # Split the Add case to handle different staggereing
+        x0 = dict(func.indices_ref._getters)
+        if self.expr.is_Add:
```
FabioLuporini, line 273: I have a feeling that anything resorting to `is_Add` or `is_Mul` during evaluation is breakable more or less easily

devito/finite_differences/derivative.py
```diff
@@ -289,6 +297,8 @@ def _eval_fd(self, expr):
         else:
             res = generic_derivative(expr, *self.dims, self.fd_order, self.deriv_order,
                                      matvec=self.transpose, x0=self.x0)
+        if expr.is_Mul:
```
FabioLuporini, line 300: same here, plus increases the number of evaluations... I understand it's for a fix, but isn't there another way?

devito/finite_differences/differentiable.py
```diff
@@ -304,6 +306,45 @@ class Mul(DifferentiableOp, sympy.Mul):
...
+            else:
+                return 1
+
+        func_args = sorted(self._args_diff, key=stagg_prio, reverse=True)
+        new_args = []
+        ref_inds = func_args[0].indices_ref._getters
+
+        for f in self.args:
+            if f not in self._args_diff:
+                new_args += [f]
```
FabioLuporini, line 338: no reason to construct a new list

devito/finite_differences/derivative.py
FabioLuporini: `x*x` isn't a `Mul`, but it's a multiplication
`cos(x)` is neither addition nor multiplication

I mean, I admit I haven't thought much about this, but it doesn't sound very robust after a first glance

mloubout: -`x*x` has no issue has they both have smae indices and fallback to default works perfectly.
- `cos(x)` has zero issue as `cos` doesn't have indices so `cos(x)` and `x` have smae indices and fallback works perfectly

mloubout: >is breakable more or less easily

why? All inputs are expressions (and is checked at creation), if not would break with or without this.

devito/finite_differences/derivative.py
mloubout: no

devito/finite_differences/derivative.py
FabioLuporini: I find it difficult to believe that it boils down to just acting differently based on whether `expr` is an Add or Mul. So, you're associating semantics to the type of operation, do I get it right? 

and if it's neither a Mul nor an Add, then we're 100% sure just returning `self._new_from_self(x0=x0)` is _always_ correct?

devito/finite_differences/derivative.py
FabioLuporini: so if it's a Mul you evaluate later, otherwise you evaluate before... this is IMHO an impenetrable mess without proper explanation

devito/finite_differences/derivative.py
```diff
@@ -277,8 +285,25 @@ def evaluate(self):
...
+        evaluate `f.dx` first.
+        - 2: Evaluate the finite difference for the (new) expression
+        - 3: Evaluate remaining (as `g` may need to be evaluated at a different point)
+        - 4: apply subsititutions.
+
+        """
+        # Evaluate derivatives in the expression
+        was_mul = expr.is_Mul
+        expr = getattr(expr, '_eval_deriv', expr)
+        # Evaluate FD of the new expressiob
```
FabioLuporini, line 306: typo: expression

devito/finite_differences/derivative.py
```diff
@@ -277,8 +285,25 @@ def evaluate(self):
...
     def _eval_fd(self, expr):
-        expr = getattr(expr, 'evaluate', expr)
+        """
+        valuate finite difference approximation of the Derivative.
+        The evaluation goes in four steps:
+        - 1: evaluate derivatives within the expression. For example `f.dx * g` will
+        evaluate `f.dx` first.
+        - 2: Evaluate the finite difference for the (new) expression
+        - 3: Evaluate remaining (as `g` may need to be evaluated at a different point)
+        - 4: apply subsititutions.
```
FabioLuporini, line 300: typo

devito/finite_differences/derivative.py
```diff
@@ -277,8 +285,25 @@ def evaluate(self):
...
+    @property
+    def _eval_deriv(self):
+        return self._eval_fd(self.expr)
+
     def _eval_fd(self, expr):
-        expr = getattr(expr, 'evaluate', expr)
+        """
+        valuate finite difference approximation of the Derivative.
+        The evaluation goes in four steps:
+        - 1: evaluate derivatives within the expression. For example `f.dx * g` will
```
FabioLuporini, line 296: evaluate -> Evaluate

devito/finite_differences/derivative.py
FabioLuporini: also, apply -> Apply

devito/finite_differences/derivative.py
```diff
@@ -277,8 +285,25 @@ def evaluate(self):
...
         return self._eval_fd(self.expr)
 
+    @property
+    def _eval_deriv(self):
+        return self._eval_fd(self.expr)
+
     def _eval_fd(self, expr):
-        expr = getattr(expr, 'evaluate', expr)
+        """
+        valuate finite difference approximation of the Derivative.
```
FabioLuporini, line 294: typo: valuate -> Evaluate

devito/finite_differences/derivative.py
```diff
@@ -277,8 +285,25 @@ def evaluate(self):
         # types of discretizations.
         return self._eval_fd(self.expr)
 
+    @property
+    def _eval_deriv(self):
+        return self._eval_fd(self.expr)
+
     def _eval_fd(self, expr):
-        expr = getattr(expr, 'evaluate', expr)
+        """
```
FabioLuporini, line 293: I like how the new docstring is coming along. Here's a suggestion. I'd remove
```
For example `f.dx * g` will evaluate `f.dx` first
```

at point 1, and I'd add a _complete_ example that shows the evolution of evaluation at each of the four steps. For example:

```
Evaluate the finite difference approximation of the Derivative.

The evaluation consists of four steps:
<the four points here>

Examples
--------
Consider:

f = Function(...)
...

After step 1:
...

After step 2:
...
```

I think we need such a thorough docstring because this method is too crucial and we need to make sure people (including me) understands exactly what's going on and why

devito/finite_differences/derivative.py
```diff
@@ -277,8 +285,25 @@ def evaluate(self):
...
+        The evaluation goes in four steps:
+        - 1: evaluate derivatives within the expression. For example `f.dx * g` will
+        evaluate `f.dx` first.
+        - 2: Evaluate the finite difference for the (new) expression
+        - 3: Evaluate remaining (as `g` may need to be evaluated at a different point)
+        - 4: apply subsititutions.
+
+        """
+        # Evaluate derivatives in the expression
+        was_mul = expr.is_Mul
```
FabioLuporini, line 304: nitpicking: instead of `was_mul`, you could name the output variables `expr1` , `expr2`, ... for a 1-to-1 correspondence to the four steps of the docstring

you may even separate the conceptual blocks of the method by interleaving each of the four steps with a blank line, and prefixing code with a comment along the lines of `# Step 1: ...`

devito/finite_differences/derivative.py
```diff
@@ -277,8 +285,25 @@ def evaluate(self):
...
+
     def _eval_fd(self, expr):
-        expr = getattr(expr, 'evaluate', expr)
+        """
+        valuate finite difference approximation of the Derivative.
+        The evaluation goes in four steps:
+        - 1: evaluate derivatives within the expression. For example `f.dx * g` will
+        evaluate `f.dx` first.
+        - 2: Evaluate the finite difference for the (new) expression
+        - 3: Evaluate remaining (as `g` may need to be evaluated at a different point)
```
FabioLuporini, line 299: for the sake of speed, we need to make sure that no useless xreplaces are applied if no substitutions are needed (might be already this way, just pointing it out)

mloubout:
@FabioLuporini discovered some deeper issue I have to work on so lots of that may change. Maybe we want to put a small PR for Diff objects with single function in it first (one line change)

devito/finite_differences/derivative.py
mloubout: Yes it is as all non `Add/Mul` are just functions (cos/sin/..) that are unary and don't requires special handling (except the open issue on having our own versions of these functions).

FabioLuporini: is it worth adding

```
else:
     assert len(self.expr.args) <= 1
```

?

devito/finite_differences/derivative.py
FabioLuporini: FWIW, I still think we need 

```
Examples
--------
...
```

devito/finite_differences/derivative.py
```diff
@@ -289,6 +315,10 @@ def _eval_fd(self, expr):
         else:
             res = generic_derivative(expr, *self.dims, self.fd_order, self.deriv_order,
                                      matvec=self.transpose, x0=self.x0)
+
+        # Step 3: Evaluate remaining part of expression
+        res = res.evaluate
+        # Step 4: Apply substitution
```
FabioLuporini, line 321: nitpicking: blank line before Step 4

devito/finite_differences/differentiable.py
```diff
@@ -115,6 +120,14 @@ def _eval_at(self, func):
             return self
         return self.func(*[getattr(a, '_eval_at', lambda x: a)(func) for a in self.args])
 
+    @property
+    def _eval_deriv(self):
+        return self.func(*[getattr(a, '_eval_deriv', a) for a in self.args])
+
+    @property
+    def _fd_priority(self):
+        return .75 if self.is_TimeDependent else .5
```
FabioLuporini, line 129: magic numbers?

devito/finite_differences/differentiable.py
```diff
@@ -304,8 +325,46 @@ class Mul(DifferentiableOp, sympy.Mul):
     __sympy_class__ = sympy.Mul
     __new__ = DifferentiableOp.__new__
 
+    @property
+    def _gather_for_diff(self):
```
FabioLuporini, line 329: I wonder whether this method should appear in the superclass too, clearly returning something dummy, or the object ``self``` itself?

devito/finite_differences/differentiable.py
```diff
@@ -304,8 +325,46 @@ class Mul(DifferentiableOp, sympy.Mul):
...
+        """
+        We handle Mul arguments by hand in case of staggered inputs
+        such as `f(x)*g(x + h_x/2)`
```
FabioLuporini, line 332: fullstop at the end, and perhaps no need to break line

devito/finite_differences/differentiable.py
```diff
@@ -304,8 +325,46 @@ class Mul(DifferentiableOp, sympy.Mul):
...
+        In that case, priority of indexing is applied to have single indices
```
FabioLuporini, line 333: the docstring was clear until this point. Then, as of this sentence, I'm a bit confused. I think a clearer example would help

devito/finite_differences/differentiable.py
FabioLuporini: how about, instead of this property, we just add it to `TimeFunction` and `Function` respectively?

devito/types/basic.py
```diff
@@ -664,6 +664,15 @@ def __init_finalize__(self, *args, **kwargs):
 
     __hash__ = Cached.__hash__
 
+    def subs(self, *args, **kwargs):
+        on_grid = kwargs.pop('on_grid', self.on_grid)
+        newobj = super(AbstractFunction, self).subs(*args, **kwargs)
+        try:
+            newobj._on_grid = on_grid
```
FabioLuporini, line 671: changing state, not acceptable

also, don't understand how can you have an AttributeError here, since you're setting an attribute

devito/types/basic.py
```diff
@@ -622,7 +622,7 @@ def __new__(cls, *args, **kwargs):
 
         if obj is not None:
             newobj = sympy.Function.__new__(cls, *args, **options)
-            newobj.__init_cached__(key)
+            newobj.__init_cached__(key, ignore=['_on_grid'])
```
FabioLuporini, line 625: this **looks** hacky

devito/types/basic.py
```diff
@@ -724,9 +733,22 @@ def dimensions(self):
         """Tuple of Dimensions representing the object indices."""
         return self._dimensions
 
+    @property
+    def _eval_deriv(self):
+        return self
+
+    @property
+    def on_grid(self):
```
FabioLuporini, line 741: why do you need to check the extra attribute `on_grid` ? can you not just check the staggered mask?

devito/types/dense.py
```diff
@@ -951,16 +951,28 @@ def __init_finalize__(self, *args, **kwargs):
         # parameter has to be computed at x + hx/2)
         self._is_parameter = kwargs.get('parameter', False)
 
+    @cached_property
+    def _fd_priority(self):
+        if self.staggered is None:
```
FabioLuporini, line 956: aka

```
if self.staggered in [None, NODE]:
    return 1
else:
    return 2
```
or

```
return 1 if ... else 2

File: devito/finite_differences/derivative.py (hunk 1 above)
Comment (mloubout) [reply devito/finite_differences/derivative.py:274]: Why?

File: devito/finite_differences/derivative.py (hunk 8 above)
Comment (mloubout) [reply devito/finite_differences/derivative.py:294]: I think it's pretty clear as it is, it's an internal method

File: devito/finite_differences/differentiable.py (hunk 12 above)
Comment (mloubout) [reply devito/finite_differences/differentiable.py:129]: They have one, this is an Expression and needs its own priority that is lower than the Function one.

File: devito/types/basic.py (hunk 17 above)
Comment (mloubout) [reply devito/types/basic.py:625]: It's not, the same way we don;t cache on `indices` but don's see it since it's in sympy.Function. This is an attribute that is different for different subs/indices and should not be cached as the original function. And in generality, there is not reason ine would not want to do partial caching of the dict.

File: devito/types/basic.py (hunk 16 above)
Comment (mloubout) [reply devito/types/basic.py:671]: - Not changing state, this is creating a new object and initializing it, not changing an existing one
- Because `Dummy` ends up here too in sympy.solve for example that thros an AttributeError.

File: devito/types/basic.py (hunk 18 above)
Comment (mloubout) [reply devito/types/basic.py:741]: Checking the staggered mask requires to remove offsets from FD, then remove staggered offset then check it's on the grid as below which is way more expensive.

File: devito/types/basic.py (hunk 16 above)
Comment (FabioLuporini) [reply devito/types/basic.py:671]: this is changing state

if you don't want it to be changing state, it must occur through `__init__` or `__new__`

File: devito/types/basic.py (hunk 18 above)
Comment (FabioLuporini) [reply devito/types/basic.py:741]: can you not save it as a flag at construction time, ie at some point in `__new__` or `__init__` or `__init_finalize__` , where you have more raw information?

I guess I should just spend some serious time to actually understand what's really going on, rather than thinking on the level of individual commits...

File: devito/finite_differences/differentiable.py (hunk 13 above)
Comment (mloubout) [reply devito/finite_differences/differentiable.py:334]: done

File: devito/types/basic.py (hunk 17 above)
Comment (mloubout) [reply devito/types/basic.py]: changed, was making my life complicated, much easier to call new directly at subs.

File: devito/finite_differences/derivative.py (hunk 1 above)
Comment (FabioLuporini) [reply devito/finite_differences/derivative.py:274]: to assert this:
> Yes it is as all non Add/Mul are just functions (cos/sin/..) that are unary

perhaps nitpicking . Just a bit concerned about us forgetting about something or SymPy changing over time. You can probably ignore this

File: devito/finite_differences/derivative.py [hunk 20]
```diff
@@ -267,7 +269,14 @@ def _eval_at(self, func):
         setup where one could have Eq(u(x + h_x/2), v(x).dx)) in which case v(x).dx
         has to be computed at x=x + h_x/2.
         """
-        x0 = dict(zip(func.dimensions, func.indices_ref))
+        # Split the Add case to handle different staggereing
```
FabioLuporini, line 272: nitpicking: typo

devito/finite_differences/derivative.py
```diff
@@ -267,7 +269,14 @@ def _eval_at(self, func):
...
+        x0 = dict(func.indices_ref._getters)
+        if self.expr.is_Add:
+            args = [self._new_from_self(expr=a, x0=x0) for a in self.expr._args_diff]
+            args += [a for a in self.expr.args if a not in self.expr._args_diff]
```
FabioLuporini, line 276: extend

FabioLuporini: or these two lines, more compactly?

```
args = [self._new_from_self(expr=a, x0=x0) if a in self.expr._args_diff else a
       for a in self.expr.args]
```

devito/finite_differences/differentiable.py
FabioLuporini: ah right sorry I'm dumb

devito/finite_differences/differentiable.py
```diff
@@ -8,7 +8,8 @@
 from cached_property import cached_property
 from devito.finite_differences.lazy import Evaluable
 from devito.logger import warning
-from devito.tools import EnrichedTuple, filter_ordered, flatten
+from devito.tools import filter_ordered, flatten
+from devito.types.utils import DimensionTuple
 
 __all__ = ['Differentiable']
```
FabioLuporini, line 15: I don't know where to write this, so I'll write it here.

Shouldn't `_fd_priority` be defined in `Differentiable` so that any future subclasses have it defined (even though with a dummy value). That is

```
class Differentiable(...):

    _fd_priority = some dummy value
     """thourhgout explanation of what this represents."""
```

devito/symbolics/manipulation.py
```diff
@@ -344,3 +344,14 @@ def indexify(expr):
         except AttributeError:
             pass
     return expr.xreplace(mapper)
+
+
+def aligned_ind(i, j, spacing):
```
FabioLuporini, line 349: ind -> indices ?

ind also a prefix for indirection

devito/types/basic.py
```diff
@@ -724,9 +725,20 @@ def dimensions(self):
         """Tuple of Dimensions representing the object indices."""
         return self._dimensions
 
+    @property
+    def _eval_deriv(self):
+        return self
+
+    @property
```
FabioLuporini, line 732: cached as it's got cost > 0

devito/types/basic.py
```diff
@@ -724,9 +725,20 @@ def dimensions(self):
...
+    def on_grid(self):
```
FabioLuporini, line 733: all introspection properties in Devito should start with `is_` or `has_`, for homogeneity. How about `is_grid_aligned` ?

also:
* it needs a docstring
* it needs to be private

devito/types/basic.py
mloubout: Well that'd defeat the all purpose of it caching it as `f(x)` and `f(x+dx/2)` would both return True instead of True and False.

FabioLuporini: but those are different objects, and you're caching on the instance, not on the class

also: `self.indices, self.indices_ref, self.dimensions` they all are immutable (as expected), so the output of this property is deterministic and immutable too

devito/finite_differences/differentiable.py
mloubout: It is defined in Differentiable below, just different value whether time dependent or not

devito/finite_differences/derivative.py
rhodrin: My brain is a bit shot atm, but this will ok if `expr` is and `Add` object with embeded `Mul`'s etc right? (i.e. `expr=f*g+f*h`).

devito/finite_differences/derivative.py
```diff
@@ -277,8 +286,25 @@ def evaluate(self):
...
 
+    @property
+    def _eval_deriv(self):
+        return self._eval_fd(self.expr)
+
     def _eval_fd(self, expr):
-        expr = getattr(expr, 'evaluate', expr)
+        """
+        Evaluate finite difference approximation of the Derivative.
+        The evaluation goes in four steps:
```
rhodrin, line 296: 'goes' doesn't really fit here, maybe `Evaluation is carried out via the following four steps:`?

devito/finite_differences/derivative.py
```diff
@@ -277,8 +286,25 @@ def evaluate(self):
...
+        - 1: Evaluate derivatives within the expression. For example `f.dx * g` will
```
rhodrin, line 297: `... expression. For example, given `f.dx * g`, `f.dx` will be evaluated first.`

devito/finite_differences/derivative.py
```diff
@@ -277,8 +286,25 @@ def evaluate(self):
...
+        evaluate `f.dx` first.
+        - 2: Evaluate the finite difference for the (new) expression.
+        - 3: Evaluate remaining (as `g` may need to be evaluated at a different point).
```
rhodrin, line 300: Evaluate the remaining terms (since `g` ...

devito/finite_differences/differentiable.py
```diff
@@ -115,6 +120,14 @@ def _eval_at(self, func):
             return self
         return self.func(*[getattr(a, '_eval_at', lambda x: a)(func) for a in self.args])
 
+    @property
+    def _eval_deriv(self):
+        return self.func(*[getattr(a, '_eval_deriv', a) for a in self.args])
+
+    @property
+    def _fd_priority(self):
```
rhodrin, line 128: Would you mind quickly outlining the design decisions behind this `_fd_priority`? i.e. why it needs/was chosen to be fractional number.

rhodrin: oh, forget it, just saw the resolved conversation below :p

tests/test_derivatives.py
```diff
@@ -421,7 +421,7 @@ def test_fd_adjoint(self, so, ndim, derivative, adjoint_name):
         deriv = getattr(f, derivative)
         coeff = 1 if derivative == 'dx2' else -1
         expected = coeff * getattr(f, derivative).evaluate.subs({x.spacing: -x.spacing})
-        assert deriv.T.evaluate == expected
+        assert simplify(deriv.T.evaluate) == simplify(expected)
```
rhodrin, line 424: Why is simplify now needed?

tests/test_staggered_utils.py
```diff
@@ -59,3 +76,87 @@ def test_is_param(ndim):
         for dd in d:
             avg = .5 * (avg + avg.subs({dd: dd - dd.spacing}))
         assert f2._eval_at(var).evaluate == avg
+
+
+@pytest.mark.parametrize('expr, expected', [
```
rhodrin, line 81: Worth having a more intricate composite expression?

tests/test_derivatives.py
mloubout: Because now subs is ours fully here and creates `(1/_h_x) * f(x)` instead of `f(x)/h_x` so doesn match without simplify (compiler refactorize so no flop increase issue)

rhodrin:
Looks good. Some nitpicking above.

devito/finite_differences/derivative.py
mloubout: Well not really. Currently only pure Add `a+b` and Mul `a*b` are going through this. For more complex expression, the assumption is that there is probably at least two different objects with the same priority and can't be safely shifted so should stay as is. I have some ideas for these cases but will need major refactor of a few pieces.

FabioLuporini: > Well not really. 

ahhhhh...that explains it. That was a big source of confusion for me, maybe I didn't explain it appropriately. 

>  I have some ideas for these cases but will need major refactor of a few pieces.

If addressing the general case would need a major refactor, then I suggest we at least:

* add a giant TODO here in the code
* add an issue to GH
* an x-failing test

then we can proceed I think

devito/finite_differences/differentiable.py
FabioLuporini: ah right sorry.

devito/finite_differences/differentiable.py
```diff
@@ -86,7 +87,11 @@ def dimensions(self):
     @property
     def indices_ref(self):
         """The reference indices of the object (indices at first creation)."""
-        return EnrichedTuple(*self.dimensions, getters=self.dimensions)
+        if len(self._args_diff) == 1:
```
FabioLuporini, line 90: I'm not sure I can discern between this `if` and the `elif` below. What does ``len(self._args_diff) == 1`` really mean, ie what case does it represent? and same for the elif below

Perhaps add a comment?

is it like, if == 1, then it's embedding a further Differentiable object, which therefore needs processing?

also, if == 0 is the base case, I suggest to swap the order of the ifs:

```
if len(self._args_diff) == 0:
    # Comment
    ...
elif len(self._args_diff) == 1:
    # Comment
    ....
else:
     # Comment
    ...
```

I'm also not sure why do == 2 , == 3...  need to be treated differently than == 1 ? Sorry, this is probably just me being slow...

devito/symbolics/manipulation.py
```diff
@@ -344,3 +344,14 @@ def indexify(expr):
         except AttributeError:
             pass
     return expr.xreplace(mapper)
+
+
+def aligned_indices(i, j, spacing):
+    """
+    Check if two indices are aligned. Two indices are aligned if they
+    differ by an Integer*spacing
```
FabioLuporini, line 352: full stop

devito/types/basic.py
```diff
@@ -724,9 +725,25 @@ def dimensions(self):
...
         return self._dimensions
 
+    @property
+    def _eval_deriv(self):
+        return self
+
+    @cached_property
+    def _is_on_grid(self):
+        """
+        Check whether the ibject is on the grid or need averaging.
```
FabioLuporini, line 735: typo, object

devito/symbolics/manipulation.py
FabioLuporini: probably nitpicking
I'm actually tempted to say that if you're using this function only in `is_on_grid`, since I really doubt it's ever going to be used elsewhere, you may as well drop it and move the content directly in that method

tests/test_staggered_utils.py
FabioLuporini: as I wrote above, yes, and perhaps x-fail it if it needs to

devito/finite_differences/derivative.py
mloubout: No, this is not a TODO or anything, this is not possible to support these kind of expression without hardocding hacks, allI am saying is that there is ways for specific cases but not here, so these would just  make thing worse pretending there is a way to do it..

devito/finite_differences/differentiable.py
mloubout: the len() == 1 is litteraly the linked issue

tests/test_staggered_utils.py
mloubout: No

devito/symbolics/manipulation.py
mloubout: I prefer it explicit and separated

tests/test_derivatives.py
mloubout: I could make it evaluate but that's be quite expensive for more complex expressions and not necessary.

devito/finite_differences/derivative.py
FabioLuporini: I still don't understand what's the semantic information encoded in `Add` , IOW, what `Add` represents to require a different treatment than `Mul`. But yeah, maybe it's just me being slow, maybe you can elaborate offline? I think I'm quite curious to understand at this point

FabioLuporini: cool now

PR #1253, opened 2020-04-24
Use OpenMP static scheduling for loops with small operation count

FabioLuporini:
also:

* drop DEVITO_DEBUG_COMPILER (now inside DEVITO_LOGGING=DEBUG)
* fix DEVITO_LOGGING=PERF


devito/ir/clusters/cluster.py
```diff
@@ -188,8 +188,8 @@ def dtype(self):
 
     @cached_property
     def ops(self):
-        """The Cluster operation count."""
-        return self.ispace.size*sum(estimate_cost(i) for i in self.exprs)
+        """Number of operations performed at each iteration."""
```
mloubout, line 191: at each grid point? I find iteration slightly missleading.

devito/passes/iet/openmp.py
```diff
@@ -122,13 +123,14 @@ def _make_construct(cls, parallel=False, **kwargs):
 
     @classmethod
     def _make_clauses(cls, ncollapse=None, chunk_size=None, nthreads=None,
-                      reduction=None, **kwargs):
+                      reduction=None, schedule=None, **kwargs):
```
mloubout, line 126: could almost make it an `opt` option like openmp/..... if people want to put their own ?

devito/passes/iet/openmp.py
```diff
@@ -230,6 +232,12 @@ class Ompizer(object):
     Coefficient to adjust the chunk size in parallelized non-affine Iterations.
     """
 
+    DYNAMIC_WORK = 10
```
mloubout, line 235: A CS individual counting in base 10, would have expected 8 or 16 ;)

devito/operator/profiling.py
```diff
@@ -382,7 +380,7 @@ def timings(self):
 
 def create_profile(name):
     """Create a new Profiler."""
-    if configuration['log-level'] == 'DEBUG':
+    if configuration['log-level'] in ['DEBUG', 'PERF']:
```
georgebisbas, line 383: yes

devito/ir/clusters/cluster.py
FabioLuporini: an iteration is a lower level concept than grid, and not all clusters (or in fact, not all devito codes) may be defined on a grid, so iteration is technically more correct

devito/passes/iet/openmp.py
FabioLuporini: I prefer them to play with JIT BACKDOOR and if they find out in some specific cases that a certain schedule is better than others, then I'd prefer to implement the logic/heuristic in the compiler directly. Long story short: I don't think it's a knob we should be exposing, because our experiments show that basically we always do the right thing

devito/passes/iet/openmp.py
FabioLuporini: lolz

PR #1261, opened 2020-04-30
bench: Add --dump-summary option

FabioLuporini:


benchmarks/user/benchmark.py
```diff
@@ -208,7 +214,23 @@ def run(problem, **kwargs):
...
+            summary = retval[-1]
+            assert isinstance(summary, PerformanceSummary)
+            f.write(str(summary.globals['fdlike']))
+
+    dumpfile = kwargs.pop('dump_norms')
+    if dumpfile:
+        norms = ["'%s': %f" % (i.name, norm(i)) for i in retval[:-1]
+                 if isinstance(i, DiscreteFunction)]
+        with open(dumpfile, 'w') as f:
+            f.write("{%s}" % ', '.join(norms))
```
mloubout, line 231: I find this plain file not very usable, can we use so more explicit format.

PR #1263, opened 2020-04-30
jit: Smarter CustomCompiler

FabioLuporini:
by returning GNUCompiler, we eventually add `-fopenmp`  (as long as the gcc version is >= 4.9 as usual) , which ensures 1) our `#pragma omp simd` are interpreted and 2) no annoying warnings "unrecognised pragma omp simd" are emitted to screen

clearly, all of this useful only if one doesn't set DEVITO_ARCH

### DIFF
--- a/devito/compiler.py
+++ b/devito/compiler.py
@@ -489,8 +489,11 @@ class IntelKNLCompiler(IntelCompiler):
 
 
 class CustomCompiler(Compiler):
+
     """
     Custom compiler based on standard environment flags.
+
+    If no environment flags are found, defaults to the GNUCompiler.
 
     Notes
     -----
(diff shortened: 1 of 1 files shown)


devito/compiler.py
```diff
@@ -500,6 +503,14 @@ class CustomCompiler(Compiler):
     or otherwise default to ``-fopenmp``.
     """
 
+    def __new__(cls, *args, **kwargs):
+        if any(i in environ for i in ['CC', 'CXX', 'CFLAGS', 'LDFLAGS']):
+            obj = Compiler.__new__(cls, *args, **kwargs)
```
mloubout, line 508: why not just a super?

FabioLuporini: yes you're right

FabioLuporini: there are some odd rules with new and init that I forget every time. But I tried adding a `print` to `__init__` and between the two `embed()` below it gets called only once

```
            from IPython import embed; embed()
            obj = super().__new__(cls, *args, **kwargs)
            obj.__init__(*args, **kwargs)
            from IPython import embed; embed()
```

FabioLuporini: is it? I'm overriding `__new__` so how is that possible? 
it's not called upon `obj = super().__new__(cls, *args, **kwargs)` (I just tried), and then I'd just be returning `obj`, so...

mloubout: Oh ok, still confised by how these two play together sometimes,

PR #1268, opened 2020-05-02
Tweak asv workflow next

rhodrin:
Seemed like some strange hangs were occasionally happening since adding the deallocate. From testing, these tweaks appear to make the workflow more robust.


FabioLuporini:
Thanks, merged

PR #1270, opened 2020-05-04
ASV: only run on HEAD while caching all previous runs

FabioLuporini:


.github/workflows/asv.yml
```diff
@@ -82,13 +82,39 @@ jobs:
...
 
     - name: Run benchmarks
       run: |
-        asv run -v --show-stderr --config benchmarks/regression/asv.conf.json HASHFILE:merge_history.txt --cpu-affinity 0-7
+        asv run -v --show-stderr --config benchmarks/regression/asv.conf.json --cpu-affinity 0-7 --machine F8s_v2
+
+    - name: Stash benchmarks results
+      run: |
+        mkdir -p /tmp/asv-results/F8s_v2
+        mv benchmarks/regression/.asv/results/benchmarks.json /tmp/asv-results
```
FabioLuporini, line 94: if I don't do this, the `checkout` action below is gonna ditch all the files that are not in the git stage area (not sure why)

mloubout: So the artifact tool is not working I take as you do it by hand?

FabioLuporini: tell me more about this "artifact tool" ?

Doing it by hand because the checkout action below, otherwise, removes all the new files

.github/workflows/asv.yml
```diff
@@ -82,13 +82,43 @@ jobs:
...
+    - name: Stash benchmarks results
+      uses: actions/upload-artifact@v2
+      with:
+        name: asv-results
+        path: /benchmarks/regression/.asv/results/
+
+    - name: Checkout asv-results branch
+      uses: actions/checkout@v1
+      with:
+        ref: asv-results
```
FabioLuporini, line 100: I mean, `clean: false` here and no need for artefacts maybe? mmhhh

mloubout: If that's what ti does even better yes

FabioLuporini:
@mloubout @rhodrin OK, just updated it with `clean: false`

FabioLuporini:
OK, let's see what happens

Issue #1273, opened 2020-05-05
ConditionalDimension: Condition indexing not shifted by offset.


georgebisbas:
The condition generated is not shifted by the halo offset. (No lowering is happening, neither indexification.)

```
from devito import Grid, ConditionalDimension, Function, Eq, Operator
from devito.symbolics import CondEq
import pdb

grid = Grid(shape=(4, 4))  # Define grid

# Define Function
g = Function(name='g', shape=grid.shape, dimensions=grid.dimensions)
x, y = grid.dimensions

ci = ConditionalDimension(name='ci', parent=y, condition=CondEq(g[x, y], 0))

f = Function(name='f', shape=grid.shape, dimensions=(x, ci))

op = Operator(Eq(f[x, ci], g[x, y] + 1))
print(op.ccode)
op.apply()
pdb.set_trace()
```

generates:

```
#pragma omp parallel num_threads(nthreads_nonaffine)
  {.
.
.
  for (int y = y_m; y <= y_M; y += 1)
      {
        if (g[x][y] == 0)
        {
          f[x + 1][y + 1] = g[x + 1][y + 1] + 1;
        }
      }
    }
  }
.
.

```

while one would expect `if (g[x+1][y+1] == 0)`

georgebisbas:
From @FabioLuporini : https://devitocodes.slack.com/archives/C3W8GF3H9/p1588692793038800
```
@George Bisbas hook for 1273: https://github.com/devitocodes/devito/blob/master/devito/operator/operator.py#L300
hints. There are two issues:
- (AFAIR) not searching for ConditionalDimensions and their condition
- even if we did, our objects are immutable so we can’t really change the condition  directly
```

FabioLuporini:
fixed by #1276

PR #1276, opened 2020-05-07
ConditionalDimension: Fix1273 inception

georgebisbas:
Here is the inception of a fix for #1273 after a meeting I had with @FabioLuporini .
I will leave some Notes for reviewers as the PR is still WIP.

Things left to do:
- Enhance equation.py API so that there is subdomain support for condition expressions.
- Solve my import problem in dimension.py
- Others


devito/types/dimension.py
```diff
@@ -709,10 +711,42 @@ class ConditionalDimension(DerivedDimension):
     is_Conditional = True
 
     def __init_finalize__(self, name, parent, factor=None, condition=None,
-                          indirect=False):
+                          indirect=False, **kwargs):
         super().__init_finalize__(name, parent)
+
+        # condition = indexify_expr(condition)
+        # import pdb; pdb.set_trace()
```
georgebisbas, line 718: NOTE FOR REVIEWERS:
The duplicate code here needs of course to be removed and we should be calling `indexify_expr`.
The fact that there is no `subdomains` for types like And, CondEq needs to be fixed. Will be added as second part of this PR. The code below is left to show that it works this way.

devito/ir/support/utils.py
```diff
@@ -158,3 +159,41 @@ def detect_io(exprs, relax=False):
             writes.append(f)
 
     return filter_sorted(reads), filter_sorted(writes)
+
+
+def indexify_expr(expressions, **kwargs):
```
georgebisbas, line 164: Is this a good place to define this function?

FabioLuporini: Drop "_expr" from the name

devito/operator/operator.py
```diff
@@ -15,13 +15,13 @@
...
 from devito.ir.stree import stree_build
+from devito.ir.support import indexify_expr
 from devito.operator.registry import operator_selector
 from devito.operator.profiling import create_profile
 from devito.mpi import MPI
 from devito.parameters import configuration
 from devito.passes import Graph
-from devito.symbolics import (estimate_cost, retrieve_functions, retrieve_indexed,
-                              uxreplace)
+from devito.symbolics import (estimate_cost, retrieve_functions)
```
FabioLuporini, line 24: drop parentheses

devito/types/dimension.py
```diff
@@ -6,7 +6,9 @@
 
 from devito.data import LEFT, RIGHT
 from devito.exceptions import InvalidArgument
+# from devito.ir.support import indexify_expr
 from devito.logger import debug
+from devito.symbolics import (retrieve_indexed, retrieve_functions, uxreplace)
```
FabioLuporini, line 11: parentheses unnecessary

tests/test_dimension.py
```diff
@@ -830,6 +830,27 @@ def test_implicit_dims(self):
 
         assert np.all(f.data == F)
 
+    def test_condition_lowering(self):
+        """
+        Test the lowering of a ConditionalDimension's condition.
+        """
+
+        # This test makes an Operator that should indexify and lower the condition
```
FabioLuporini, line 838: part of the docstring

devito/types/dimension.py
FabioLuporini: > The fact that there is no subdomains for types like And, CondEq needs to be fixed

People won't understand what you're talking about

devito/ir/support/utils.py
FabioLuporini: Where should this function go is a very good question. 

`ir` uses `types`, so it can't go there, since this will have to be called from inside `ConditionalDimension`

an option would be `passes/expressions`, but then again it wouldn't be callable from within `types/dimension.py`

how about directly inside `types/equation.py` ? after all, this is to be called with eqns as input. I'm not 100% sure though...

Thoughts @mloubout @rhodrin ?

georgebisbas: I was thinking about that but there is another indexify fuction in manupulation.py . Other name?

georgebisbas: + we already use it in ours (line:176 below)

devito/operator/operator.py
georgebisbas: dropped

tests/test_dimension.py
georgebisbas: Fixed.

devito/types/dimension.py
georgebisbas: fixed

devito/ir/support/utils.py
georgebisbas: hmm.. if I add it in equation.py I get
```
...
  File "/home/gbisbas/workspace/devito/devito/types/equation.py", line 8, in <module>
    from devito.symbolics import retrieve_indexed, retrieve_functions, uxreplace
ImportError: cannot import name 'retrieve_indexed' from partially initialized module 'devito.symbolics' (most likely due to a circular import) (/home/gbisbas/workspace/devito/devito/symbolics/__init__.py)

```

devito/types/dimension.py
```diff
@@ -718,7 +720,17 @@ def __init_finalize__(self, name, parent, factor=None, condition=None,
         # import pdb; pdb.set_trace()
         processed = []
         expr = condition
-        dimension_map = {}
+
+        # Awful TOFIX
```
georgebisbas, line 724: NOTE FOR REVIEWERS:
We intend to support subdomains for ConditionalDimensions's conditions.
I am trying to migrate extended sympy's Classes to `equation.py`
 and inherit subdomain properties. 
Pure Sympy conditions like `And` that is heavily used in injection and interpolation do not yet support a subdomain
in their expression.
That's why the awful hack here that needs to be fixed.
As said in previous comment the code below is duplicate to indexify_expr.

devito/types/dimension.py
georgebisbas: Partially explained in comment below.

tests/test_dimension.py
```diff
@@ -830,6 +831,69 @@ def test_implicit_dims(self):
...
+        g.data[:] = 1
+        x, y = grid.dimensions
+
+        ci = ConditionalDimension(name='ci', parent=y, condition=CondEq(g[x, y], 1))
+
+        f = Function(name='f', shape=grid.shape, dimensions=(x, ci))
+        Operator(Eq(f[x, ci], g[x, y])).apply()
+        assert np.all(f.data == 1)
+
+    def test_condition_subdomains(self):
```
georgebisbas, line 854: NOTE FOR REVIEWERS: This test can be deleted later, I just leave it temporarily here
to check the output of this script.

tests/test_dimension.py
```diff
@@ -851,6 +851,154 @@ def test_condition_lowering(self):
         Operator(Eq(f[x, ci], g[x, y])).apply()
         assert np.all(f.data == 1)
 
+    def test_strictlessthan_relational(self):
```
georgebisbas, line 854: NOTE for reviewers: The tests below have a lot of code replication, can we reduce throusgh a paremtetrized call?

devito/ir/support/utils.py
FabioLuporini: I thought about that and I don't have a better suggestion than putting it in `ir/equations/algorithms.py`.

Please call it `lower_exprs` or something along the lines of `lower_...`

This most likely will create a circular import when you will attempt to import it from `types/dimension.py`. I don't have a solution to this ATM. So just use a local import for now:

```
class ConditionalDimension(...):
    def __init_finalize__(...):
        ....

        from devito.ir.equatios import lower_exprs
        ...
```

@mloubout any suggestion would be appreciated, but I don't think we can escape from this, as needed it in `types`, and in particular in `types/dimension.py`, doesn't give us many alternatives I feel. Not unless we start splitting the real API type classes from the support type classes, but jeez...that's for another day/life I think

devito/ir/support/utils.py
```diff
@@ -158,3 +159,41 @@ def detect_io(exprs, relax=False):
...
+            mapper[i] = f.indexed[indices]
+
+        subs = kwargs.get('subs')
+        if subs:
+            # Include the user-supplied substitutions, and use
+            # `xreplace` for constant folding
+            processed.append(expr.xreplace({**mapper, **subs}))
+        else:
+            processed.append(uxreplace(expr, mapper))
+    return processed
```
FabioLuporini, line 199: blank line needed before this return

devito/ir/support/utils.py
FabioLuporini: I also think this line:

https://github.com/devitocodes/devito/pull/1276/files#diff-5ebf1ee0cae059be380b472501c3d663R275

should be squashed inside this new `lower_exprs`, at the very end:

```
...

processed = [LoweredEq(e) for e in processed if e.is_Equality]

return processed
```

tests/test_dimension.py
```diff
@@ -830,6 +831,254 @@ def test_implicit_dims(self):
...
+
+        Eq1 = Eq(f[x, y], g[x, y])
+        Eq2 = Eq(f, 5)
+
+        Operator([Eq1, Eq2]).apply()
+
+        assert np.all(f.data[2:6, 4:6] == 5.)
+        assert np.all(f.data[:, :4] < 5.)
+
+    def test_strictgreaterthan_relational(self):
```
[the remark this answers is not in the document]
georgebisbas: I agree yes, this adresses my note here:
https://github.com/devitocodes/devito/pull/1276/files#r422971333
Suggestions? I understand from your comment that I can parametrze.

devito/ir/support/utils.py
georgebisbas: hm.. this link you pasted doens't give me a line?

devito/types/equation.py
```diff
@@ -158,9 +158,9 @@ def __str__(self):
     __repr__ = __str__
 
 
-class LessThan(sympy.LessThan):
+class Ne(sympy.Ne):
```
georgebisbas, line 161: Note: To discuss how to relate or not with CondNe.

tests/test_dimension.py
```diff
@@ -851,120 +851,14 @@ def test_condition_lowering(self):
...
-        Eq2 = Eq(f, 5)
-
-        Operator([Eq1, Eq2]).apply()
-
-        assert np.all(f.data[2:6, 4:6] == 5.)
-        assert np.all(f.data[:, :4] < 5.)
-
-    def test_strictgreaterthan_relational(self):
+    @pytest.mark.parametrize('setup_rel, rhs, c1, c2, c3, c4', [
+        # Carried dependence in both /t/ and /x/
```
georgebisbas, line 855: Copy-paste Leftover, to be dropped/

tests/test_dimension.py
```diff
@@ -851,120 +851,14 @@ def test_condition_lowering(self):
...
+        (Lt, 3, 2, 4, 4, -1), (Le, 2, 2, 4, 4, -1), (Ge, 3, 4, 6, 1, 4),
+        (Gt, 2, 4, 6, 1, 4), (CondNe, 5, 2, 6, 1, 1)
+    ])
+    def test_relational_classes(self, setup_rel, rhs, c1, c2, c3, c4):
         """
-        Test a ConditionalDimension applied in a subdomain.
+        Test relational classes in ConditionalDimension applied in a subdomain.
```
georgebisbas, line 861: TOimprove docstring

devito/ir/support/utils.py
```diff
@@ -158,3 +159,44 @@ def detect_io(exprs, relax=False):
             writes.append(f)
 
     return filter_sorted(reads), filter_sorted(writes)
+
+
+def expr_lowering(expressions, **kwargs):
```
FabioLuporini, line 164: I don't think this is where I suggested this should go and I don't think this is how it should be named

devito/ir/support/utils.py
```diff
@@ -158,3 +159,44 @@ def detect_io(exprs, relax=False):
...
+    # Indexification
```
FabioLuporini, line 165: and now this function needs a proper docstring

devito/types/dimension.py
```diff
@@ -711,8 +711,17 @@ class ConditionalDimension(DerivedDimension):
     def __init_finalize__(self, name, parent, factor=None, condition=None,
                           indirect=False):
         super().__init_finalize__(name, parent)
+
+        # Local import of expr_lowering
```
FabioLuporini, line 715: make it a TODO

devito/types/dimension.py
```diff
@@ -711,8 +711,17 @@ class ConditionalDimension(DerivedDimension):
...
+        from devito.ir.support import expr_lowering
+
+        expressions = []
+        expressions.append(condition)
+        processed = expr_lowering(expressions)
```
FabioLuporini, line 720: this is ugly, to say the least

```
self._condition = lower_exprs(condition)
```

devito/types/dimension.py
```diff
@@ -711,8 +711,17 @@ class ConditionalDimension(DerivedDimension):
...
+
         self._factor = factor
-        self._condition = condition
+        # import pdb; pdb.set_trace()
```
FabioLuporini, line 723: leftover

devito/types/equation.py
```diff
@@ -1,4 +1,4 @@
-"""User API to specify equations."""
+"""User API to specify equations and relations."""
```
FabioLuporini, line 1: an equation is a relation

devito/types/equation.py
```diff
@@ -7,7 +7,8 @@
 from devito.finite_differences import Evaluable, default_rules
 from devito.tools import as_tuple
 
-__all__ = ['Eq', 'Inc', 'solve']
+__all__ = ['Eq', 'Inc', 'solve', 'Le', 'Lt', 'Ge',
+           'Gt', 'Ne']
```
FabioLuporini, line 11: no need to break line

devito/types/equation.py
FabioLuporini: we don't

devito/types/equation.py
```diff
@@ -157,6 +158,174 @@ def __str__(self):
...
 
+class LessThan(sympy.LessThan):
+    """
+    An less-than ("<=") relation between two objects, the left-hand side and the
+    right-hand side.
+
+    The left-hand side may be a Function or a SparseFunction. The right-hand
+    side may be any arbitrary expressions with numbers, Dimensions, Constants,
+    Functions and SparseFunctions as operands.
+
```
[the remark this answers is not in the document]
FabioLuporini: I think the main section is fine

devito/types/equation.py
```diff
@@ -157,6 +158,174 @@ def __str__(self):
...
+    lhs : Function or SparseFunction
+        The left-hand side.
+    rhs : expr-like, optional
+        The right-hand side. Defaults to 0.
+    subdomain : SubDomain, optional
+        To restrict the evalaution of the relation to a particular sub-region in the
+        computational domain.
+
+    Examples
+    --------
```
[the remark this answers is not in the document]
FabioLuporini: for example, yes, and it should be extremely succint

devito/types/equation.py
```diff
@@ -157,6 +158,174 @@ def __str__(self):
...
+    def subdomain(self):
+        """The SubDomain in which the Eq is defined."""
+        return self._subdomain
+
+    def __str__(self):
+        return "%s(%s, %s)" % (self.__class__.__name__, self.lhs, self.rhs)
+
+    __repr__ = __str__
+
+
```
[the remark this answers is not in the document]
FabioLuporini: additional class

`Ne` is for the users, `CondNe` is used internally by the compiler. I think they are different

tests/test_dimension.py
```diff
@@ -830,6 +831,148 @@ def test_implicit_dims(self):
...
+        ci = ConditionalDimension(name='ci', parent=y, condition=CondEq(g[x, y], 1))
+
+        f = Function(name='f', shape=grid.shape, dimensions=(x, ci))
+        Operator(Eq(f[x, ci], g[x, y])).apply()
+        assert np.all(f.data == 1)
+
+    @pytest.mark.parametrize('setup_rel, rhs, c1, c2, c3, c4', [
+        # Relation, RHS, c1 to c4 used as indexes in assert
+        (Lt, 3, 2, 4, 4, -1), (Le, 2, 2, 4, 4, -1), (Ge, 3, 4, 6, 1, 4),
+        (Gt, 2, 4, 6, 1, 4), (CondNe, 5, 2, 6, 1, 1)
```
FabioLuporini, line 857: I don't think CondNe is necessary

tests/test_dimension.py
```diff
@@ -830,6 +831,148 @@ def test_implicit_dims(self):
...
+        # Eq1 : First we copy g to f
+        # Eq2 : Then Conditional subdomain
+        Operator([Eq1, Eq2]).apply()
+
+        assert np.all(f.data[2:6, 2:4] == 5.)
+        assert np.all(f.data[:, 4:] < 5.)
+
+    def test_relational_subdomains(self):
+        """
+        Test a ConditionalDimension applied in a subdomain II.
```
FabioLuporini, line 941: is this really a different test?

tests/test_dimension.py
georgebisbas: Dropped.

devito/types/equation.py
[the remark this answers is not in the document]
georgebisbas: Added a few let;s discuss again.

devito/types/equation.py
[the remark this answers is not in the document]
georgebisbas: I need to make it more succinct.

devito/types/equation.py
```diff
@@ -157,6 +158,174 @@ def __str__(self):
...
+        The right-hand side. Defaults to 0.
+    subdomain : SubDomain, optional
+        To restrict the evalaution of the relation to a particular sub-region in the
+        computational domain.
+
+    Examples
+    --------
+
+    Notes
+    -----
```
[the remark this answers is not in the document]
georgebisbas: Deleted

devito/types/equation.py
```diff
@@ -157,6 +158,174 @@ def __str__(self):
...
+
+class StrictLessThan(sympy.StrictLessThan):
+    """
+    A strict-less-than ("<") relation between two objects, the left-hand side and the
+    right-hand side.
+
+    The left-hand side may be a Function or a SparseFunction. The right-hand
+    side may be any arbitrary expressions with numbers, Dimensions, Constants,
+    Functions and SparseFunctions as operands.
+
```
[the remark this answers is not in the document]
georgebisbas: Marking as resolved.

devito/types/equation.py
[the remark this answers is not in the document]
georgebisbas: I have not yet managed to make `Ne` to work.

tests/test_dimension.py
georgebisbas: Removed. `Ne` will be added as soon as it is ready.

tests/test_dimension.py
georgebisbas: Accidentally leftover. Removed.

devito/types/dimension.py
```diff
@@ -711,8 +711,15 @@ class ConditionalDimension(DerivedDimension):
...
+
+        # TODO: Avoid local import of lower_exprs
+        from devito.ir.equations.algorithms import lower_exprs
+
+        expressions = []
+        expressions.append(condition)
+
         self._factor = factor
-        self._condition = condition
+        self._condition = lower_exprs(expressions).pop()
```
mloubout, line 722: wait so this is done at `__init__()`. that' a bit odd isn't it?
So I assume this is because Dimension are never lowered. It used to be indexified somewhere though jsut without the shift, where was that happening?

devito/types/equation.py
```diff
@@ -157,6 +157,211 @@ def __str__(self):
     __repr__ = __str__
 
 
+class Le(sympy.Le):
```
mloubout, line 160: we may wanna split that into `equation.py` and `relational.py`

devito/types/equation.py
```diff
@@ -157,6 +157,211 @@ def __str__(self):
...
+                return {d: ('middle', 2, 2) for d in dimensions}
+    >>> inner_domain = InnerDomain()
+    >>> grid = Grid(shape=(8, 8), subdomains=(inner_domain,))
+    >>> g = Function(name='g', shape=grid.shape, dimensions=grid.dimensions)
+    >>> x, y = grid.dimensions
+    >>> xi, yi = grid.subdomains['inner'].dimensions
+    >>> cond = Gt(g[xi, yi], rhs, subdomain=grid.subdomains['inner'])
+    >>> ci = ConditionalDimension(name='ci', parent=yi, condition=cond)
+    >>> f = Function(name='f', shape=grid.shape, dimensions=(xi, ci))
+    >>> Eq1 = Eq(f[x, y], g[x, y])
```
mloubout, line 351: I think some of these need to have the output in the docstring if it returns something when run

tests/test_dimension.py
```diff
@@ -830,6 +831,105 @@ def test_implicit_dims(self):
...
+        f = Function(name='f', shape=grid.shape, dimensions=(x, ci))
+        Operator(Eq(f, g)).apply()
+        assert np.all(f.data == 1)
+
+    @pytest.mark.parametrize('setup_rel, rhs, c1, c2, c3, c4', [
+        # Relation, RHS, c1 to c4 used as indexes in assert
+        (Lt, 3, 2, 4, 4, -1), (Le, 2, 2, 4, 4, -1), (Ge, 3, 4, 6, 1, 4),
+        (Gt, 2, 4, 6, 1, 4)
+    ])
+    def test_relational_classes(self, setup_rel, rhs, c1, c2, c3, c4):
```
mloubout, line 859: move to `test_equation` or make a new `test_relational` that'd be consitant with splkit equation/relational

devito/types/dimension.py
FabioLuporini: no, it's fine to do it at `__init__`. The issue is that `condition` is an actual expression, possible with lots of indexed, and this is the only place we can lower it without losing sanity. Because, if we instead lowered it later on, during Operator compilation, then to avoid changing state we would have to go through an incredible mess:
* locating all CDs
* getting their conditions and explicitly lowering them
* create new CDs with the new lowered conditions
* replace the old CDs with new CDs

and even then, this would lead to all sort of headaches. Not to mention when one uses the CD directly inside a Function...

so this is the only place we can perform this lowering, and since lowering is a local thing, I don't think there's anything bad with it

devito/types/equation.py
FabioLuporini: yes, this is a very good point. Let's do it

devito/ir/equations/algorithms.py
```diff
@@ -64,3 +65,51 @@ def handle_indexed(indexed):
...
+
+def lower_exprs(expressions, **kwargs):
+    """
+    Expression lowering and indexification:
+
+        * Indexify functions
+        * Align indexeds with the computational domain
+        * Consider user-applied subs
+        """
+    # Indexification
```
FabioLuporini, line 78: drop this line

devito/ir/equations/algorithms.py
```diff
@@ -64,3 +65,51 @@ def handle_indexed(indexed):
...
+    # E.g., f(x - 2*h_x, y) -> f[xi + 2, yi + 4]  (assuming halo_size=4)
```
FabioLuporini, line 79: I guess move to the docstring?

devito/types/equation.py
```diff
@@ -157,6 +157,211 @@ def __str__(self):
...
+                return {d: ('middle', 2, 2) for d in dimensions}
+    >>> inner_domain = InnerDomain()
+    >>> grid = Grid(shape=(8, 8), subdomains=(inner_domain,))
+    >>> g = Function(name='g', shape=grid.shape, dimensions=grid.dimensions)
+    >>> x, y = grid.dimensions
+    >>> xi, yi = grid.subdomains['inner'].dimensions
+    >>> cond = Le(g[xi, yi], rhs, subdomain=grid.subdomains['inner'])
+    >>> ci = ConditionalDimension(name='ci', parent=yi, condition=cond)
+    >>> f = Function(name='f', shape=grid.shape, dimensions=(xi, ci))
+    >>> Eq1 = Eq(f[x, y], g[x, y])
```
FabioLuporini, line 198: eq = ....

devito/types/equation.py
```diff
@@ -157,6 +157,211 @@ def __str__(self):
...
+    subdomain : SubDomain, optional
+        To restrict the evalaution of the relation to a particular sub-region in the
+        computational domain.
+
+    Examples
+    --------
+    Le may be used to express a relation. Below, a summation along
+    the user-defined Dimension ``i``.
+
+    >>> class InnerDomain(SubDomain):
```
FabioLuporini, line 185: I think this example is insanely more complicated than it strictly needs to be

devito/types/equation.py
FabioLuporini: though unfortunately the example isn't really computing anything...

tests/test_dimension.py
FabioLuporini: well, the relational for now are only needed inside ConditionalDimension, and we wanna make sure that CD does lower the condition, so... I'm not sure about this one

tests/test_dimension.py
FabioLuporini: isn't it still here?

devito/types/equation.py
georgebisbas: SImplified now.

devito/types/equation.py
georgebisbas: Added directly to the Operator now. Let me know if I should change.

devito/types/equation.py
georgebisbas: SHould I add the snippet of the generated code that contains loops+conditional?

tests/test_dimension.py
georgebisbas: Removed the previous copy.

tests/test_dimension.py
georgebisbas: It is a tricky decision, although I cannot see a reason to discretize.  Both of you have a point as relationals and CDs are tightly coupled when used,

devito/types/relational.py
```diff
@@ -0,0 +1,189 @@
...
+    Examples
+    --------
+    Le may be used to express a relation (e.g. in a Subdomain).
+
+    >>> from devito import Grid, Function, ConditionalDimension, Eq, Operator
+    >>> from devito.types import Le
+    >>> grid = Grid(shape=(8, 8))
+    >>> g = Function(name='g', shape=grid.shape, dimensions=grid.dimensions)
+    >>> x, y = grid.dimensions
+    >>> cond = Le(g, 1)
```
georgebisbas, line 38: This can be squashed to the next line to reduce code size.

tests/test_dimension.py
```diff
@@ -830,6 +831,105 @@ def test_implicit_dims(self):
...
+
+        g.data[:4, :4] = 1
+        g.data[4:, :4] = 2
+        g.data[4:, 4:] = 3
+        g.data[:4, 4:] = 4
+
+        x, y = grid.dimensions
+        xi, yi = grid.subdomains['inner'].dimensions
+
+        cond = And(g[xi, yi] < 3, subdomain=grid.subdomains['inner'])
```
georgebisbas, line 919: This test actually is here to check the generated code fo the unorthodox API code generation through `And`.
I can remove but as discussed this is still a hacky way to semantically have equivalent:
`cond = And(g[xi, yi] < 3, subdomain=grid.subdomains['inner']) `
`cond = Le(g[xi, yi], 3, subdomain=grid.subdomains['inner']) `

Though after this PR, we should drop probably.

mloubout: I'd rather not let people get used to the `And` one and only keep `Le`. I don't think many people use that yet so it shouldn't be an issue to only keep `Le`

devito/types/relational.py
```diff
@@ -0,0 +1,189 @@
...
+    >>> ci = ConditionalDimension(name='ci', parent=y, condition=cond)
+    >>> f = Function(name='f', shape=grid.shape, dimensions=(x, ci))
+    >>> op = Operator(Eq(g, f))
```
mloubout, line 41: I would almost skip these three and just do like a `>> cond._conditon` and make sure it outputs what we expect

devito/types/relational.py
```diff
@@ -0,0 +1,189 @@
...
+
+import sympy
+
+
+__all__ = ['Le', 'Lt', 'Ge', 'Gt']
+
+
+class Le(sympy.Le):
+    """
+    A less-than ("<=") relation between two objects, the left-hand side and the
```
mloubout, line 11: less-than or equal

devito/types/relational.py
```diff
@@ -0,0 +1,189 @@
...
+    right-hand side. Can be used to build conditionals but not directly to
+    construct an Operator.
+
+    The left-hand side may be a Function or a SparseFunction. The right-hand
+    side may be any arbitrary expressions with numbers, Dimensions, Constants,
+    Functions and SparseFunctions as operands.
+
+    Parameters
+    ----------
+    lhs : Function or SparseFunction
```
mloubout, line 21: Can't be an `Expr` right?

devito/types/relational.py
```diff
@@ -0,0 +1,189 @@
...
+        The left-hand side.
+    rhs : expr-like, optional
+        The right-hand side. Defaults to 0.
+    subdomain : SubDomain, optional
+        To restrict the evalaution of the relation to a particular sub-region in the
+        computational domain.
+
+    Examples
+    --------
+    Le may be used to express a relation (e.g. in a Subdomain).
```
mloubout, line 31: That's funny to read as a frenchman :D

devito/types/relational.py
```diff
@@ -0,0 +1,189 @@
...
+
+    @property
+    def subdomain(self):
+        """The SubDomain in which the Le is defined."""
+        return self._subdomain
+
+
+class Lt(sympy.Lt):
+    """
+    A strict-less-than ("<") relation between two objects, the left-hand side and the
```
mloubout, line 57: less-than. people usually just use less-than for stric and less-than (-or-) equal for nonstrict. Same for Ge,Gt

devito/types/relational.py
```diff
@@ -0,0 +1,189 @@
...
+        The left-hand side.
+    rhs : expr-like, optional
+        The right-hand side. Defaults to 0.
+    subdomain : SubDomain, optional
+        To restrict the evalaution of the relation to a particular sub-region in the
+        computational domain.
+
+    Examples
+    --------
+    Le may be used to express a relation (e.g. in a Subdomain).
```
mloubout, line 77: Lt

devito/types/relational.py
```diff
@@ -0,0 +1,189 @@
...
```
mloubout, line 123: Ge

devito/types/relational.py
```diff
@@ -0,0 +1,189 @@
...
+        The left-hand side.
+    rhs : expr-like, optional
+        The right-hand side. Defaults to 0.
+    subdomain : SubDomain, optional
+        To restrict the evalaution of the relation to a particular sub-region in the
+        computational domain.
+
+    Examples
+    --------
+    Gt may be used to express a relation.
```
mloubout, line 169: blank line

devito/types/dimension.py
mloubout: Makes sense. Not part of the repr so shouldn't be confusing neither.

devito/types/relational.py
georgebisbas: Yes, just need `cond`

devito/types/relational.py
georgebisbas: Fixed.
I agree absolutely with you. I had that changed to be consistent with SymPy terminology.

devito/types/relational.py
georgebisbas: Fixed

devito/types/relational.py
georgebisbas: Fixed

devito/types/relational.py
georgebisbas: Fixed

tests/test_dimension.py
georgebisbas: What should I do here?

tests/test_dimension.py
georgebisbas: Agreed. Removed.

devito/types/relational.py
```diff
@@ -36,9 +36,8 @@ class Le(sympy.Le):
     >>> g = Function(name='g', shape=grid.shape, dimensions=grid.dimensions)
     >>> x, y = grid.dimensions
     >>> cond = Le(g, 1)
-    >>> ci = ConditionalDimension(name='ci', parent=y, condition=cond)
-    >>> f = Function(name='f', shape=grid.shape, dimensions=(x, ci))
-    >>> op = Operator(Eq(g, f))
+    >>> cond
```
mloubout, line 39: last nitpicking, ca you also add the `>>cond._conditon` so that can see the indexed expression is shifted correctly?

georgebisbas: Yes, funny it is that I was about to ask that.

devito/types/dimension.py
mloubout: So this should be `self._condition = expression` and lowering moved to algorithms no?
https://github.com/devitocodes/devito/blob/6aa03bb38b74224f742beb1dbe8df814b45941a0/devito/ir/clusters/algorithms.py#L179

devito/types/relational.py
```diff
@@ -0,0 +1,249 @@
...
+    >>> from devito import Grid, Function, ConditionalDimension, Eq, Operator
+    >>> from devito.types import Le
+    >>> grid = Grid(shape=(8, 8))
+    >>> g = Function(name='g', shape=grid.shape, dimensions=grid.dimensions)
+    >>> x, y = grid.dimensions
+    >>> Le(g, 1)
+    g(x, y) <= 1
+    """
+
+    is_Relational = False
```
mloubout, line 41: I'd rather not have that if possible

devito/types/equation.py
georgebisbas: Is this fine now?

tests/test_dimension.py
georgebisbas: Reminder

mloubout: leave it here it's fine

devito/ir/clusters/algorithms.py
```diff
@@ -176,7 +177,8 @@ def guard(clusters):
                 if cd.condition is None:
                     condition.append(CondEq(cd.parent % cd.factor, 0))
                 else:
-                    condition.append(cd.condition)
+                    cond = lower_exprs(cd.condition).pop()
```
FabioLuporini: I don't see how can this work?  `cd.condition` is an expression (not a list of expressions), and you pass it to `lower_exprs` which, in turn, performs `for expr in expressions`, but `expressions` as I just wrote is not a list

in `lower_exprs`, you probably should:
```
for expr in expressions: -> for expr in as_tuple(expressions):
```
and, at the end, `return` as follows
```
if isinstance(expressions, Iterable):
    return processed  # a list
else:
    assert len(processed) == 1
    return processed.pop()
```

IOW, if `lower_exprs` is given a list, it should return a list; if it's given a single expression, it should return a single expression, not a list. This makes `lower_exprs` more flexible while the caller won't need this horrible `.pop()`

devito/ir/equations/algorithms.py
```diff
@@ -64,3 +65,52 @@ def handle_indexed(indexed):
     ordering = PartialOrderTuple(extra, relations=(relations | implicit_relations))
 
     return ordering
+
+
+def lower_exprs(expressions, **kwargs):
+    """
+    Expression lowering and indexification:
```
FabioLuporini, line 72: "Lowering an expression consists of the following passes:"

devito/ir/equations/algorithms.py
```diff
@@ -64,3 +65,52 @@ def handle_indexed(indexed):
...
+        * Indexify functions
+        * Align indexeds with the computational domain
+        * Consider user-applied subs
+
+    # E.g., f(x - 2*h_x, y) -> f[xi + 2, yi + 4]  (assuming halo_size=4)
+    """
+
+    processed = []
+    for expr in expressions:
+        try:
```
FabioLuporini, line 83: what's going on here? this is ugly

devito/types/dimension.py
```diff
@@ -722,8 +722,9 @@ class ConditionalDimension(DerivedDimension):
     def __init_finalize__(self, name, parent, factor=None, condition=None,
                           indirect=False):
         super().__init_finalize__(name, parent)
+
         self._factor = factor
-        self._condition = condition
+        self._condition = [condition]
```
FabioLuporini: Quite ugly, as I explained above

devito/types/dimension.py
```diff
@@ -749,8 +750,8 @@ def index(self):
     @property
     def free_symbols(self):
         retval = super(ConditionalDimension, self).free_symbols
-        if self.condition is not None:
-            retval |= self.condition.free_symbols
+        if self.condition[0] is not None:
```
FabioLuporini: the consequence of the ugliness -- more ugliness

devito/types/relational.py
FabioLuporini: it should. Try, perhaps add test, and change the docstring as "expr-like"

devito/types/relational.py
```diff
@@ -0,0 +1,260 @@
...
+        The left-hand side.
+    rhs : expr-like, optional
+        The right-hand side. Defaults to 0.
+    subdomain : SubDomain, optional
+        To restrict the evalaution of the relation to a particular sub-region in the
+        computational domain.
+
+    Examples
+    --------
+    Le may be used to express a relation (e.g. in a Subdomain).
```
FabioLuporini, line 30: quotes between Le. update all docstrings similarly

devito/types/relational.py
```diff
@@ -0,0 +1,260 @@
...
+
+    >>> from devito import Grid, Function, ConditionalDimension, Eq, Operator
+    >>> from devito.types import Le
+    >>> grid = Grid(shape=(8, 8))
+    >>> g = Function(name='g', shape=grid.shape, dimensions=grid.dimensions)
```
FabioLuporini, line 35: this is another punch in the eye:

```
```

devito/types/relational.py
```diff
@@ -0,0 +1,260 @@
...
+    >>> x, y = grid.dimensions
```
FabioLuporini, line 36: why do you need these?

devito/types/relational.py
```diff
@@ -0,0 +1,260 @@
...
+        computational domain.
+
+    Examples
+    --------
+    Lt may be used to express a relation (e.g. in a Subdomain).
+
+    >>> from devito import Grid, Function, ConditionalDimension, Eq, Operator
+    >>> from devito.types import Lt
+    >>> grid = Grid(shape=(8, 8))
+    >>> g = Function(name='g', shape=grid.shape, dimensions=grid.dimensions)
```
FabioLuporini, line 84: like above

devito/types/relational.py
```diff
@@ -0,0 +1,260 @@
...
+        return obj
+
+    @property
+    def subdomain(self):
+        """The SubDomain in which the Le is defined."""
+        return self._subdomain
+
+    @property
+    def negated(self):
+        return ops.get(self.func)(*self.args)
```
FabioLuporini, line 54: what is `ops` ? :O

FabioLuporini: edit: ah I see, at the end...

unnecessary I'd say?

why not just

```
return Gt(*self.args)
```

?

devito/types/relational.py
```diff
@@ -0,0 +1,260 @@
...
+        return obj
+
+    @property
+    def subdomain(self):
+        """The SubDomain in which the Lt is defined."""
+        return self._subdomain
+
+    @property
+    def negated(self):
+        return ops.get(self.func)(*self.args)
```
FabioLuporini, line 103: like above

devito/types/relational.py
```diff
@@ -0,0 +1,260 @@
...
+    mathematically. Source: https://docs.sympy.org/latest/modules/core.html
+    """
+
+    is_Relational = False
+
+    def __new__(cls, lhs, rhs=0, subdomain=None, **kwargs):
+        kwargs.update({'evaluate': False})
+        obj = sympy.Ne.__new__(cls, lhs, rhs, **kwargs)
+        obj._subdomain = subdomain
+        # import pdb; pdb.set_trace()
```
FabioLuporini, line 251: leftover

tests/test_dimension.py
```diff
@@ -830,6 +831,76 @@ def test_implicit_dims(self):
...
+        f = Function(name='f', shape=grid.shape, dimensions=(xi, ci))
+
+        Eq1 = Eq(f, g)
+        Eq2 = Eq(f, 5)
+
+        Operator([Eq1, Eq2]).apply()
+        assert np.all(f.data[2:6, c1:c2] == 5.)
+        assert np.all(f.data[:, c3:c4] < 5.)
+
+    def test_not_equal_form(self):
```
FabioLuporini, line 892: this is not for this file

tests/test_dimension.py
```diff
@@ -830,6 +831,76 @@ def test_implicit_dims(self):
...
+        assert np.all(f.data == 1)
+
+    @pytest.mark.parametrize('setup_rel, rhs, c1, c2, c3, c4', [
+        # Relation, RHS, c1 to c4 used as indexes in assert
+        (Lt, 3, 2, 4, 4, -1), (Le, 2, 2, 4, 4, -1), (Ge, 3, 4, 6, 1, 4),
+        (Gt, 2, 4, 6, 1, 4), (Ne, 5, 2, 6, 1, 2)
+    ])
+    def test_relational_classes(self, setup_rel, rhs, c1, c2, c3, c4):
+        """
+        Test ConditionalDimension using conditions constructed with overriden
```
FabioLuporini, line 860: "... using conditions based on Relations."

tests/test_dimension.py
```diff
@@ -830,6 +831,76 @@ def test_implicit_dims(self):
 
         assert np.all(f.data == F)
 
+    def test_condition_lowering(self):
```
FabioLuporini, line 834: test_lowering

tests/test_dimension.py
```diff
@@ -830,6 +831,76 @@ def test_implicit_dims(self):
...
+        grid = Grid(shape=(3, 3))  # Define grid
+
+        # Define Function
+        g = Function(name='g', shape=grid.shape, dimensions=grid.dimensions)
+        g.data[:] = 1
+        x, y = grid.dimensions
+        ci = ConditionalDimension(name='ci', parent=y, condition=Le(g, 1))
+
+        f = Function(name='f', shape=grid.shape, dimensions=(x, ci))
+        Operator(Eq(f, g)).apply()
```
FabioLuporini, line 850: you could also `assert` against the expected structure of the lowered condition, ie (pseudocode)

```
assert condition.lhs.indices == ...
```

tests/test_dimension.py
```diff
@@ -830,6 +831,76 @@ def test_implicit_dims(self):
...
+        g.data[4:, 4:] = 3
+        g.data[:4, 4:] = 4
+
+        xi, yi = grid.subdomains['inner'].dimensions
+
+        cond = setup_rel(g, rhs, subdomain=grid.subdomains['inner'])
+        ci = ConditionalDimension(name='ci', parent=yi, condition=cond)
+        f = Function(name='f', shape=grid.shape, dimensions=(xi, ci))
+
+        Eq1 = Eq(f, g)
```
FabioLuporini, line 885: variables should basically never start with capital letters

```
eqns = [Eq(...), Eq(...)]
```

since they're tiny

devito/types/relational.py
mloubout: Generallity, and matches sympy Relational

devito/ir/equations/algorithms.py
mloubout: would rather avoid try/except too but even if necessary, the `if` is completely unncecessary 
```
try: 
    dimension_map = expr.subdomain.dimension_map 
except AttributeError:
   dimension_map ={}
```
does the same.

devito/ir/clusters/algorithms.py
mloubout: agree

devito/ir/equations/algorithms.py
FabioLuporini: > Some condition expressions do not have a subdomain 

such as which ones?

devito/types/relational.py
georgebisbas: Added a test and seems to be all fine.

georgebisbas: Not sure yet if it works for subdomain correctly. Getting some random(OMG!) failures!

tests/test_dimension.py
georgebisbas: WIll do.

tests/test_dimension.py
georgebisbas: Fixed

tests/test_dimension.py
georgebisbas: Fixed

tests/test_dimension.py
georgebisbas: Fixed

tests/test_dimension.py
georgebisbas: test_ir ?

devito/ir/clusters/algorithms.py
georgebisbas: Fixed

devito/ir/equations/algorithms.py
```diff
@@ -64,3 +67,53 @@ def handle_indexed(indexed):
...
+
+def lower_exprs(expressions, **kwargs):
+    """
+    Lowering an expression consists of the following passes:
+
+        * Indexify functions
+        * Align indexeds with the computational domain
+        * Consider user-applied subs
+
+    # E.g., f(x - 2*h_x, y) -> f[xi + 2, yi + 4]  (assuming halo_size=4)
```
FabioLuporini, line 80:
```
Examples
---------
....
```

devito/ir/equations/algorithms.py
FabioLuporini: > such as which ones?

this is being discussed on Slack -- about overriding `__lt__` etc to return our own `Lt` etc, which do have a subdomain attached

devito/ir/equations/algorithms.py
```diff
@@ -64,3 +67,53 @@ def handle_indexed(indexed):
...
+        subs = kwargs.get('subs')
+        if subs:
+            # Include the user-supplied substitutions, and use
+            # `xreplace` for constant folding
+            processed.append(expr.xreplace({**mapper, **subs}))
+        else:
+            processed.append(uxreplace(expr, mapper))
+
+    if isinstance(expressions, Iterable):
+        return processed  # a list
```
FabioLuporini, line 116: drop comment

devito/types/dimension.py
```diff
@@ -722,6 +722,7 @@ class ConditionalDimension(DerivedDimension):
     def __init_finalize__(self, name, parent, factor=None, condition=None,
                           indirect=False):
         super().__init_finalize__(name, parent)
+
```
FabioLuporini, line 725: is this blank line homogeneous w.r.t. the other constructors in this module?

devito/types/relational.py
```diff
@@ -0,0 +1,255 @@
...
+
+import sympy
+
+__all__ = ['Le', 'Lt', 'Ge', 'Gt', 'Ne']
+
+
+class Le(sympy.Le):
+    """
+    A less-than or equal ("<=") relation between two objects, the left-hand side and the
+    right-hand side. Can be used to build conditionals but not directly to
```
FabioLuporini, line 11: It can be

devito/types/relational.py
```diff
@@ -0,0 +1,255 @@
...
+    construct an Operator.
+
+    The left-hand side may be a Function or a SparseFunction. The right-hand
```
FabioLuporini, line 14: Drop this whole paragraph, useless (clearly apply to all other docstrigns too)

devito/types/relational.py
FabioLuporini: I think it'd be cleaner to just return `Gt(...)` but OK

tests/test_dimension.py
```diff
@@ -830,6 +831,98 @@ def test_implicit_dims(self):
 
         assert np.all(f.data == F)
 
+    def test_lowering(self):
+        """
+        Test the lowering of a ConditionalDimension's condition.
+        This test makes an Operator that should indexify and lower the condition
+        passed in the Conditional Dimension
```
FabioLuporini, line 838: ConditionalDimension

tests/test_dimension.py
```diff
@@ -830,6 +831,98 @@ def test_implicit_dims(self):
...
+        """
+
+        grid = Grid(shape=(3, 3))  # Define grid
+
+        # Define Function
```
FabioLuporini, line 843: useless comment

tests/test_dimension.py
```diff
@@ -830,6 +831,98 @@ def test_implicit_dims(self):
...
         assert np.all(f.data == F)
 
+    def test_lowering(self):
+        """
+        Test the lowering of a ConditionalDimension's condition.
+        This test makes an Operator that should indexify and lower the condition
+        passed in the Conditional Dimension
+        """
+
+        grid = Grid(shape=(3, 3))  # Define grid
```
FabioLuporini, line 841: useless comment

tests/test_dimension.py
```diff
@@ -830,6 +831,98 @@ def test_implicit_dims(self):
 
         assert np.all(f.data == F)
 
+    def test_lowering(self):
+        """
+        Test the lowering of a ConditionalDimension's condition.
+        This test makes an Operator that should indexify and lower the condition
+        passed in the Conditional Dimension
+        """
+
```
FabioLuporini, line 840: no blank lines between docstrings and first line of code in tests

tests/test_dimension.py
FabioLuporini: OK, thx

tests/test_dimension.py
```diff
@@ -830,6 +831,98 @@ def test_implicit_dims(self):
...
+        assert np.all(f.data == 1)
+
+    def test_expr_like_lowering(self):
+        """
+        Test the lowering of an expr-like ConditionalDimension's condition.
+        This test makes an Operator that should indexify and lower the condition
+        passed in the Conditional Dimension
+        """
+
+        grid = Grid(shape=(3, 3))  # Define grid
```
FabioLuporini, line 860: drop comments

tests/test_dimension.py
```diff
@@ -830,6 +831,98 @@ def test_implicit_dims(self):
...
+        g = Function(name='g', grid=grid)
+        g.data[:] = 1
+        x, y = grid.dimensions
+        ci = ConditionalDimension(name='ci', parent=y, condition=Le(g, 1))
+
+        f = Function(name='f', shape=grid.shape, dimensions=(x, ci))
+        Operator(Eq(f, g)).apply()
+        assert np.all(f.data == 1)
+
+    def test_expr_like_lowering(self):
```
FabioLuporini, line 853: how is this test different than the previous one? maybe just keep this one that uses something slightly more complicated in the condition ie `g1+g2` ?

tests/test_dimension.py
```diff
@@ -830,6 +831,98 @@ def test_implicit_dims(self):
...
+        assert np.all(f.data[:] == g1.data[:] + g2.data[:])
+
+    @pytest.mark.parametrize('setup_rel, rhs, c1, c2, c3, c4', [
+        # Relation, RHS, c1 to c4 used as indexes in assert
+        (Lt, 3, 2, 4, 4, -1), (Le, 2, 2, 4, 4, -1), (Ge, 3, 4, 6, 1, 4),
+        (Gt, 2, 4, 6, 1, 4), (Ne, 5, 2, 6, 1, 2)
+    ])
+    def test_relational_classes(self, setup_rel, rhs, c1, c2, c3, c4):
+        """
+        Test ConditionalDimension using using conditions based on Relations.
```
FabioLuporini, line 883: double "using"

" ... on Relations over SubDomains" ?

tests/test_dimension.py
```diff
@@ -830,6 +831,98 @@ def test_implicit_dims(self):
...
+        Operator(Eq(f, g1+g2)).apply()
+
+        assert np.all(f.data[:] == g1.data[:] + g2.data[:])
+
+    @pytest.mark.parametrize('setup_rel, rhs, c1, c2, c3, c4', [
+        # Relation, RHS, c1 to c4 used as indexes in assert
+        (Lt, 3, 2, 4, 4, -1), (Le, 2, 2, 4, 4, -1), (Ge, 3, 4, 6, 1, 4),
+        (Gt, 2, 4, 6, 1, 4), (Ne, 5, 2, 6, 1, 2)
+    ])
+    def test_relational_classes(self, setup_rel, rhs, c1, c2, c3, c4):
```
FabioLuporini, line 881: I actually think this and the two tests above could be squashed into a single test that tests lowering in the most general case , ie an expression `g1+g2` over a SubDomain. 

What do you think?

This three tests smells like over-engineering

tests/test_dimension.py
FabioLuporini: I think you can safely drop this test

tests/test_dimension.py
FabioLuporini: Unless I'm missing something, they all are testing, basically, that lowering _is happening_ -- not that "lowering works". So one general test should cover everything ?

devito/ir/equations/algorithms.py
FabioLuporini: if we don't wanna fix it now, perhaps leave a comment

```
try:
    ...
except AttributeError:
    # Some Relationals may be pure SymPy objects, thus lacking the subdomain
    dimension_map = {}
```

devito/ir/equations/algorithms.py
georgebisbas: Dropped

devito/types/dimension.py
georgebisbas: Sorry, a leftover. Dropped.

devito/types/relational.py
georgebisbas: Fixed

devito/types/relational.py
georgebisbas: Dropped, applied.

devito/types/relational.py
georgebisbas: Can that be marked as resolved?

devito/types/relational.py
georgebisbas: resolved?

tests/test_dimension.py
georgebisbas: Fixed

tests/test_dimension.py
georgebisbas: Fixed

tests/test_dimension.py
georgebisbas: Dropped

tests/test_dimension.py
georgebisbas: Dropped

tests/test_dimension.py
georgebisbas: Now, the lowering is not happening upon creation. I added another test for lowering+index-shifting of the relationals. Not yet for CDs although I can do as well.

tests/test_dimension.py
georgebisbas: Yes

tests/test_dimension.py
georgebisbas: Dropped

tests/test_dimension.py
georgebisbas: Yes, enhanced.

tests/test_dimension.py
georgebisbas: Yes

tests/test_dimension.py
georgebisbas: Dropped

FabioLuporini:
Merged, thanks!

Issue #1183, opened 2020-03-23
`SubDimension` 'thickness' inconsitently defined


rhodrin:
Take a look at:
```
    @classmethod
    def left(cls, name, parent, thickness, local=True):
        lst, rst = cls._symbolic_thickness(name)
        return cls(name, parent,
                   left=parent.symbolic_min,
                   right=parent.symbolic_min+lst-1,
                   thickness=((lst, thickness), (rst, 0)),
                   local=local)

    @classmethod
    def right(cls, name, parent, thickness, local=True):
        lst, rst = cls._symbolic_thickness(name)
        return cls(name, parent,
                   left=parent.symbolic_max-rst+1,
                   right=parent.symbolic_max,
                   thickness=((lst, 0), (rst, thickness)),
                   local=local)

    @classmethod
    def middle(cls, name, parent, thickness_left, thickness_right, local=False):
        lst, rst = cls._symbolic_thickness(name)
        return cls(name, parent,
                   left=parent.symbolic_min+lst,
                   right=parent.symbolic_max-rst,
                   thickness=((lst, thickness_left), (rst, thickness_right)),
                   local=local)
```

The 'thickness' properties mean something different for middle vs left/right - this should probably be made consistent or the actual properties changed?

mloubout:
Maybe we do not need to change the user code but only how the input is processed to create the subdomain:
- Turn the thickness left/right into a dim_width 
- Remove thickness internally
- Size is just dim_width

This would prbably mess up lot of sthings so maybe not as trivial as I picture it.

FabioLuporini:
I'm looping in @tjb900  because it's a relevant discussion for him

I'll take a proper look tmr

rhodrin:
Note: Temporary fix in #1186.

FabioLuporini:
@rhodrin reminder that in your next PR you may wanna expand that comment in _arg_defaults with a reference to this issue

PR #1284, opened 2020-05-13
Fix subdimension thickness

rhodrin:
Owing the inconsistency in how the thickness of subdimensions is defined (see #1183) a subdomain defined as, e.g.,
```
class sd0(SubDomain):
    name = 'sd0'
    def define(self, dimensions):
        x, y = dimensions
        return {x: x, y: ('right', 10)}
```
on a grid with shape `(N, 10)` doesn't currently work. This PR fixes that. Note that it doesn't fix the underlying issue raised in #1183.

### DIFF
--- a/devito/types/dimension.py
+++ b/devito/types/dimension.py
@@ -622,8 +622,8 @@ class SubDimension(DerivedDimension):
                                                    self.thickness.left[1]-1, LEFT)
                 rtkn = grid.distributor.glb_to_loc(self.root,
                                                    self.thickness.right[1]-1, RIGHT)
-                ltkn = ltkn+1 if ltkn else 0
-                rtkn = rtkn+1 if rtkn else 0
+                ltkn = ltkn+1 if ltkn is not None else 0
+                rtkn = rtkn+1 if rtkn is not None else 0
             else:
                 # dimension is of type ``middle``
                 ltkn = grid.distributor.glb_to_loc(self.root, self.thickness.left[1],
--- a/devito/types/dimension.py
+++ b/devito/types/dimension.py
@@ -615,10 +615,21 @@ class SubDimension(DerivedDimension):
     def _arg_defaults(self, grid=None, **kwargs):
         if grid is not None and grid.is_distributed(self.root):
             # Get local thickness
-            ltkn = grid.distributor.glb_to_loc(self.root,
-                                               self.thickness.left[1], LEFT) or 0
-            rtkn = grid.distributor.glb_to_loc(self.root,
-                                               self.thickness.right[1], RIGHT) or 0
+            if self.local:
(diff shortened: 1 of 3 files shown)


devito/types/dimension.py
```diff
@@ -615,10 +615,21 @@ def _offset_right(self):
...
             # Get local thickness
-            ltkn = grid.distributor.glb_to_loc(self.root,
-                                               self.thickness.left[1], LEFT) or 0
-            rtkn = grid.distributor.glb_to_loc(self.root,
-                                               self.thickness.right[1], RIGHT) or 0
+            if self.local:
+                # dimension is of type ``left``/right`` - compute the 'offset'
+                # and then add 1 to get the appropriate thickness
+                ltkn = grid.distributor.glb_to_loc(self.root,
+                                                   self.thickness.left[1]-1, LEFT)
```
mloubout, line 622: so this removes one (-1) from thickness and add it back below? looks weird.

rhodrin: the `glb_to_loc` is computing the offset, not the thickness. Consider the following example: we have a grid:
```
0-1-2-3-4
```
Then, i.e. `glb_to_loc(d, 0, LEFT)` would return 0. `glb_to_loc(d,4,LEFT)` would return 4 and `glb_to_loc(d,5,LEFT)` would return `None` since it's OOB. Hence we take the 1 off to get the offset but then we need to add it back to get the thickness (as defined for subdimensions of type `left`/`right` which have a minimum 'thickness' of 1).

mloubout: Ok makes sense.

rhodrin:
> So empty subdomain are allowed now?

Are you referring to the `y: ('right', 10)`? If so, no, I think this is the api being confusing again - that's saying that the dimension spans the entire `y` dimension i.e. equivalent to `y: y`.

mloubout:
> i.e. equivalent to y: y.

right my bad, brain not very great today :D

devito/types/dimension.py
```diff
@@ -615,10 +615,21 @@ def _offset_right(self):
     def _arg_defaults(self, grid=None, **kwargs):
         if grid is not None and grid.is_distributed(self.root):
             # Get local thickness
-            ltkn = grid.distributor.glb_to_loc(self.root,
-                                               self.thickness.left[1], LEFT) or 0
-            rtkn = grid.distributor.glb_to_loc(self.root,
-                                               self.thickness.right[1], RIGHT) or 0
+            if self.local:
+                # dimension is of type ``left``/right`` - compute the 'offset'
```
FabioLuporini, line 619: we could/should also explicitly refer the issue on github (the API inconsistency).

not necessarily in this PR though

rhodrin: Yeah, I guess we can add that in to the next PR?

PR #1291, opened 2020-05-15
Fix pickling with MPI full mode

### LINKED ISSUE #1286: MPIRegion pickle_args inconsistent with constructor
Hi!

We're hitting this error, since MPIRegion needs to be pickled when pickling an Operator - as needed to avoid building operators on every node of a job.

```
TypeError: __init__() missing 1 required positional argument: 'owned'
```

It appears to be caused by this:

```
class MPIRegion(CompositeObject):

    def __init__(self, name, key, arguments, owned):
        name = "%s%d" % (name, key)
        pname = "region%d" % key
...

    # Pickling support
    _pickle_args = ['name', 'arguments', 'owned']
```

I don't have an MFE to hand yet, sorry.

FabioLuporini:
fixes #1286

### DIFF
--- a/devito/mpi/routines.py
+++ b/devito/mpi/routines.py
@@ -925,17 +925,19 @@ class MPIMsg(CompositeObject):
     else:
         c_mpirequest_p = type('MPI_Request', (c_void_p,), {})
 
-    def __init__(self, name, function, halos, fields=None):
+    fields = [
+        (_C_field_bufs, c_void_p),
+        (_C_field_bufg, c_void_p),
+        (_C_field_sizes, POINTER(c_int)),
+        (_C_field_rrecv, c_mpirequest_p),
+        (_C_field_rsend, c_mpirequest_p),
+    ]
+
+    def __init__(self, name, function, halos):
         self._function = function
         self._halos = halos
-        fields = (fields or []) + [
-            (MPIMsg._C_field_bufs, c_void_p),
-            (MPIMsg._C_field_bufg, c_void_p),
-            (MPIMsg._C_field_sizes, POINTER(c_int)),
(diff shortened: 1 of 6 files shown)


devito/mpi/routines.py
```diff
@@ -1058,6 +1058,7 @@ def __init__(self, name, key, arguments, owned):
         name = "%s%d" % (name, key)
```
tjb900, line 1058: Actually...  isn't this going to make it different when it's unpickled?  Since the name property is no longer the name passed to init?

FabioLuporini: oohhh right nice catch! I even wrote an MFE for MPIMsg but not for MPIRegion. 

should be a trivial fix, will do on Monday

FabioLuporini:
@tjb900 should be fine now?

FabioLuporini:
jeez... this was a rabbit hole: https://github.com/sympy/sympy/issues/4297

causing pickling to break with MPI_fullmode + OpenMP

I just pushed a workaround

devito/symbolics/extended_sympy.py
```diff
@@ -293,6 +294,40 @@ def __str__(self):
...
+
+        https://github.com/sympy/sympy/issues/4297
+    """
+
+    def __new__(cls, name, arguments=None):
+        arguments = as_tuple(arguments)
+        obj = Function.__new__(cls, name, *arguments)
+        obj._name = name
+        obj._arguments = arguments
+        return obj
```
mloubout, line 310: That's...... :D But reading through the issue no really other way around it.

FabioLuporini: yes I ditched 1 hour on this this morning than I gave up as it's just not worth it

PR #1292, opened 2020-05-15
bench: fix results makedir to prevent exception with multiple MPI processes

jaimesouza:
When running the benchmark with MPI, all ranks try to make the result directory to save the json file. This might generate an exception, when a process tries to create an existing directory. This PR fix this issue by adding a parameter to the `os.makedirs` function. With the parameter `exist_ok = True`, the function does not generate an error when the directory already exists.


FabioLuporini:
thanks!

PR #1295, opened 2020-05-17
Fix MPI benchmark json results

jaimesouza:
Fix issues with MPI benchmark measures. 

1. Reduce values of all ranks to one rank only (rank 0)

    - Global gflops is the sum of all gflops

    - Global io is the maximum of all io

    - Global timing is the maximum of all timing 

2. Generate only one json file (rank 0) with the unified measures


benchmarks/user/benchmark.py
```diff
@@ -342,7 +342,16 @@ def bench(problem, **kwargs):
 
     bench = get_ob_bench(problem, resultsdir, kwargs)
     bench.execute(get_ob_exec(run), warmups=0, repeats=repeats)
-    bench.save()
+
+    # With MPI, only rank0 writes to disk
+    try:
```
FabioLuporini, line 347: I'd rewrite this whole block as:

```
try:
    if MPI.COMM_WORLD.rank == 0:
        bench.save()
except AttributeError:
    pass
```

benchmarks/user/tools/executor.py
```diff
@@ -43,6 +43,49 @@ def run(self, **kwargs):
         """
         raise NotImplementedError("No custom executor function specified")
 
+    def unify_mpi_measures(self, gflopss, oi, timings):
```
FabioLuporini, line 46: this routine is probably useless. All the information you're after is already carried by the summary dictionary that an Operator returns after `op.apply(...)`:

https://github.com/devitocodes/devito/blob/master/devito/operator/profiling.py#L230

so we somewhat need to exploit that.

An Operator exploits the `summary` to produce both rank-local and global performance data: https://github.com/devitocodes/devito/blob/master/devito/operator/operator.py#L745

I suppose the global performance data is what we need here right?

jaimesouza: Ok, I will take a look.

jaimesouza: I just want to change the dictionary at this point (https://github.com/devitocodes/devito/blob/master/benchmarks/user/benchmark.py#L519), cause I know that the change here will not affect others functions that I don't know about.

Some log messages need the full dictionary and could be compromised.

And I need only the benchmark json file to contain the overall result for each section.

FabioLuporini: yes...? 

at that line, `self.func(...)` returns a bunch of objects. The last of them (which is currently dropped on the floor due to the `_`) is the Operator `summary` which contains all you need

jaimesouza: I printed the fourth object `(_)` in `gflopss, oi, timings, _ = self.func(*args, **kwargs)` and It seems to be the record values (like the values in the grid). 

I just want to reduce the values in `gflopss, oi, timings` and remove entries with rank > 0, like I did before (adding a method for that). It worked, but somehow fails in CI tests.

FabioLuporini: > I just want to reduce the values in gflopss, oi, timings

yeah but as I said, all this information is within such `summary` dictionary

FabioLuporini: @georgebisbas do you see what I mean or am I talking nonsense?

mloubout: All of these are in summary yeah. a simple `[(k, v.oi) for k,v in summary..items()]` shoudl already give you everything back then just need to sum per section.

georgebisbas: After a quick look at the acoustic example, I think @jaimesouza is right;
Acoustic_example returns:
`    return summary.gflopss, summary.oi, summary.timings, [rec, u.data]`
which is mapped to:
`gflopss, oi, timings, _' or `gflopss, oi, timings, summary'.

So except if am misleaded here, summary per se in benchmarks doesn;t hold profiling data, but just the receiver and field data.

georgebisbas: Of course, we can act the other way round and change the returned args in the respective examples. (?)

FabioLuporini: so we rewrite it so that instead of `run` we use `setup` and then call `solver.forward`, just like I dif for the `run` mode. Then from `solver.forward` we have `summary`.

> Of course, we can act the other way round and change the returned args in the respective examples. (?)
 
that's wrong

FabioLuporini: @jaimesouza

georgebisbas: Seems fine plan.

jaimesouza: Ok guys. I'm a litle busy right now, but I'm going to check it out soon.

jaimesouza:
Yes, I don't know why it was generating N json files before (one for each rank). But now only rank 0 saves the json file. Besides, the information in this json is now the global measures. And is global in terms of MPI (summaried measures of all ranks) and sections (there's no individual measure for each specific section). 

In order to get summary.globals I also had to change fdlike option to carry oi information as well. So, I recover the global measures using summary.globals['fdlike']. I could use 'vanilla' option, but I only works for MPI executions (mpirun python ...). 

Why did you have the measures for each sections before? Is it important for some specific analysis? I ask because, I am not sure if I can just change that and save only global measures.

What do you think about all of this? Is it correct or am I missing something?

georgebisbas:
@jaimesouza we were indeed generating one file per rank so as to have the performance of each rank. To be honest I am not sure that we need them, @FabioLuporini WDYT?
Different sections are for different `tasks` in the c-code, For example ususally section0 holds the PDE updates, section 1 has source injextion, section 2 has receivers ...Some problems like TTI have more sections....etc..

In my personal opinion I do not mind having one file generated only. @FabioLuporini , @mloubout , @rhodrin ?

benchmarks/user/benchmark.py
```diff
@@ -25,22 +25,22 @@
     'viscoelastic': {
         'run': viscoelastic_run,
         'setup': viscoelastic_setup,
-        'default-section': 'section0'
+        'default-section': 'global'
```
FabioLuporini, line 28: @georgebisbas @mloubout I think this is a nice catch

benchmarks/user/benchmark.py
```diff
@@ -516,12 +525,19 @@ def __init__(self, func):
...
+
+            summary = retval[-1]
+            assert isinstance(summary, PerformanceSummary)
+            globals = summary.globals['fdlike']
 
-            for key in timings.keys():
-                self.register(gflopss[key], measure="gflopss", event=key.name)
-                self.register(oi[key], measure="oi", event=key.name)
-                self.register(timings[key], measure="timings", event=key.name)
+            # before there was a measure for each section
```
FabioLuporini, line 535: you can drop this comment.

But what you've done makes complete sense to me

devito/operator/profiling.py
```diff
@@ -364,13 +364,16 @@ def add_glb_fdlike(self, points, time):
         gpointss = gpoints/time
 
         if self.input:
+            traffic = sum(v.traffic for v in self.input.values())
             ops = sum(v.ops for v in self.input.values())
```
FabioLuporini, line 368: nitpicking: blank line after this line? this would separate the known quantities (traffic, ops) from the derived metrics (gflops, gflopss, oi, ..)

jaimesouza:
@georgebisbas and @FabioLuporini I agree with you. I am gonna do as you suggested.

georgebisbas:
@jaimesouza thanks for the latest commit (I think it needs to be rebased to latest master as well.). I did a test drive for all dump formats. I think it is very nice and we are almost there, a question to trigger discussion @FabioLuporini . Currently, the default method (-df global) produces the metrics for all sections and ranks...so globals have gflopss for the whole Kernel, rather than per section. Do we want the prementioned, do we want globals per section, or both?

FabioLuporini:
Merged.

PR #1296, opened 2020-05-18
Expose CIRE knobs

FabioLuporini:


devito/core/cpu.py
```diff
@@ -35,6 +35,19 @@ class CPU64NoopOperator(OperatorCore):
...
+    Minimum operation count of a Dimension-invariant aliasing expression to be
+    optimized away. Dimension-invariant aliases are lifted outside of one or more
+    invariant loop(s), so they require tensor temporaries that can be potentially
+    very large (e.g., the whole domain in the case of time-invariant aliases).
+    """
+
+    CIRE_MINCOST_SOPS = 10
+    """
+    Minimum operation count of a sum-of-product aliasing expression to be optimized away.
+    """
```
mloubout, line 49: can't we move these defaults to `OperatorCore` instead, unnecessary reapeat here.

devito/core/gpu_openmp.py
```diff
@@ -230,6 +230,19 @@ class DeviceOpenMPNoopOperator(OperatorCore):
...
+    Minimum operation count of a Dimension-invariant aliasing expression to be
+    optimized away. Dimension-invariant aliases are lifted outside of one or more
+    invariant loop(s), so they require tensor temporaries that can be potentially
+    very large (e.g., the whole domain in the case of time-invariant aliases).
+    """
+
+    CIRE_MINCOST_SOPS = 10
+    """
+    Minimum operation count of a sum-of-product aliasing expression to be optimized away.
+    """
```
mloubout, line 244: same

devito/passes/clusters/aliases.py
```diff
@@ -80,22 +65,32 @@ def cire(cluster, template, mode, options, platform):
...
-    assert all(i > 0 for i in options['cire-repeats'].values())
-
     # Relevant options
     min_storage = options['min-storage']
+    min_cost = options['cire-mincost']
+    repeats = options['cire-repeats']
+
+    # Sanity checks
+    assert mode in ['invariants', 'sops']
+    assert all(i > 0 for i in repeats.values())
```
mloubout, line 75: do we want to allow `0` to get a zero optization operator? just a thought

devito/core/cpu.py
FabioLuporini: no, they will change at some point for GPUs. And there already are a few differences, the CIRE_REPEATS_INV, for example, is 2 in the GPU one. As you may guess, I thought carefully about centralizing this, but it was growing into a big mess, with partial overrides in the individual operators, so I gave up. I prefer the **root** Operator classes (the CPU root, and the GPU root) to define them explicitly

devito/passes/clusters/aliases.py
FabioLuporini: I could allow it... but if one really wants to disable it, the way to go is to use a CustomOperator as we do in some tests btw

devito/core/cpu.py
mloubout: Ok fine with me

devito/passes/clusters/aliases.py
mloubout: I get that but this would be an "easier" way to have quick access to a basic noop operator.

FabioLuporini: OK, fine by me. Do you mind if I do that in the PR that sits right on top of this? it's the patch for the bug we discovered yesterday. So we don't have to respin CI, rebase, etc

mloubout: Completely fine yes. Played with that other one and looks like it fixes the issue so can probably open the PR (that big try/except not sure but does the job).

PR #1299, opened 2020-05-19
docs: Update index.rst and relative subpages

FabioLuporini:


README.md
```diff
@@ -111,12 +111,17 @@ If you are interested in any of the following
 
 * Generation of parallel code (CPU, GPU, multi-node via MPI);
 * Performance tuning;
-* Benchmarking Devito;
-* Any other aspect concerning application performance;
+* Benchmarking operators;
```
mloubout, line 114: We may wanna add a small section or link to the env variables, lots of people complaining thay can't find how to set openmp/compiler/....

docs/source/index.rst
```diff
@@ -22,19 +22,17 @@ You can get instructions on how to download and install Devito
...
-have a look at this :doc:`material </compiler>`.
-
 .. title:: The Devito project
 
 .. toctree::
    :hidden:
 
    Download <download>
-   Tutorials <tutorials>
+   Tutorials <https://github.com/devitocodes/devito/tree/master/examples>
```
mloubout, line 36: nbviewer is a lot nicer for tutorials so I prefer the previous way. May need some reorga to make it easier to link though.

README.md
FabioLuporini: they are in the FAQ now

docs/source/index.rst
FabioLuporini: oh I see, I had missed that. I was just trying to drop redundant places -- we have a nice description of what our tutorials are in `examples/README.md` and we had another one, incomplete and obsolete (!),  in `tutorials.rst`, which therefore I'm keen to drop, otherwise, it's just maintenance effort. 

How about we add a link to nbviewer in `examples/README.md` ?

README.md
mloubout: People don't look that far or don't find it (there is a lot of tabs in the Wiki) and and easy access to it is really necessary. Can guarantee you I get people complaining very often they can't find how to set it

docs/source/index.rst
mloubout: I think that'd be fine

README.md
FabioLuporini: FAQ now linked from here: https://www.devitoproject.org/ and from this README too under Resources.

Two bullets above: "Generation of parallel code (CPU, GPU, multi-node via MPI);" -> go-to `benchmarks/user/README.md` (where all relevant env vars are described)

Do you really think linking the FAQ **also** from here would make such a big difference?  IMHO unnecessary, but I can do that if you're sure that this is the central point of the README

mloubout: I think it's fine, not trying to create extra work. People will always find way to complain anyway.

docs/source/index.rst
FabioLuporini: done now, have a look

PR #1300, opened 2020-05-19
Fix scheduling of dimension-invariant aliasing expressions

FabioLuporini:


tests/test_dse.py
```diff
@@ -475,6 +475,49 @@ def test_uncontracted_shape(self):
...
+        """
+        grid = Grid(shape=(6, 6, 6))
+        x, y, z = grid.dimensions  # noqa
+        t = grid.stepping_dim
+
+        f = Function(name='f', grid=grid)
+        f.data_with_halo[:] =\
+            np.linspace(-1, 1, f.data_with_halo.size).reshape(*f.shape_with_halo)
+        u = TimeFunction(name='u', grid=grid, space_order=3)
+        u.data_with_halo[:] = 0.5
```
mloubout, line 490: any reason to add valeu to halo?

tests/test_dse.py
```diff
@@ -475,6 +475,49 @@ def test_uncontracted_shape(self):
...
+        assert a.symbolic_shape[0] == x_size
+        assert a.symbolic_shape[1] == y_size
+        assert Add(*a.symbolic_shape[2].args) == z_size + 1
+
+        # Check numerical output
+        op0(time_M=1)
+        exp = np.copy(u.data[:])
+        u.data_with_halo[:] = 0.5
+        op1(time_M=1)
+        assert np.allclose(u.data, exp, rtol=10e-7)
```
mloubout, line 518: I'd add an extra assert on the value of `u` (supposed to e sqrt(2))

devito/passes/clusters/aliases.py
```diff
@@ -699,12 +699,17 @@ def iter(self, ispace):
...
+            # Becomes True as soon as a Dimension in `ispace` is found to
+            # be independent of `intervals`
+            flag = False
             writeto = []
-            for n, i in enumerate(ispace.intervals):
-                interval = mapper.get(i.dim)
-                if interval is not None:
-                    if not writeto and interval == interval.zero():
+            for i in ispace.intervals:
+                try:
```
mloubout, line 710:
```
interval = mapper.get(i.dim)
if not interval:
    if any(i.dim in d._defines for d in mapper):
         # E.g., `i.dim` is `x0_blk0` in `x0_blk0[0,0]<0>`
          pass
     else:
           # E.g., `t[0,0]<0>` in the case of t-invariant aliases
          flag = True
```

would avoid large try/except.

tests/test_dse.py
FabioLuporini: not in this example, just copy-paste from the other tests. I can tweak it

tests/test_dse.py
FabioLuporini: yep, I can add an extra assert -- but what "value of u" are you specifically referring to?

devito/passes/clusters/aliases.py
FabioLuporini: Hang on, I'll try to make it nicer

FabioLuporini: now?

mloubout: but do we really need the try/except?

FabioLuporini: it's completely pythonic and I'm sure we use the same pattern somewhere else in Devito

so: "really need" -> no , "is it fine to have" -> yes

mloubout: ok gtg

tests/test_dse.py
mloubout: Sorry, for some reason read `f.data=1` forget what I said.

PR #1303, opened 2020-05-19
Ensure av runner always shutdown

rhodrin:
Tiny patch to ensure the asv runner is shutdown even if, for some reason, the benchmark fails.


FabioLuporini:
Merged, thanks

PR #1306, opened 2020-05-20
asymmetric padding

mloubout:
Allow asymmetric padding such as in numpy with `pad((left, right))` and added test
 
Will be necessary for proper subdomain freesurface for example that would not pad the top.

### DIFF
--- a/devito/builtins.py
+++ b/devito/builtins.py
@@ -209,18 +209,23 @@ def nbl_to_padsize(nbl, ndim):
     """
     Creates the pad sizes from `nbl`. The output is a tuple of tuple
     such as ((40, 40), (40, 40)) where each subtuple is the left/right
-    pad size for the dimension
+    pad size for the dimension.
     """
-    nb_t = as_tuple(nbl)
-    if len(nb_t) == 1 and len(nb_t) < ndim:
-        nbl = ndim*((nb_t[0], nb_t[0]),)
-    elif len(nb_t) == ndim:
-        if len(as_tuple(nb_t[0])) == 1:
-            nbl = tuple((nbd, nbd) for nbd in nb_t)
+    nb_total = as_tuple(nbl)
+    if len(nb_total) == 1 and len(nb_total) < ndim:
+        nb_pad = ndim*((nb_total[0], nb_total[0]),)
+    elif len(nb_total) == ndim:
+        nb_pad = []
+        for nb_dim in nb_total:
+            if len(as_tuple(nb_dim)) == 1:
+                nb_pad.append((nb_dim, nb_dim))
+            else:
+                assert len(nb_dim) == 2
+                nb_pad.append(nb_dim)
     else:
-        raise ValueError("nbl must be an integer or tuple of integers of length" +
-                         " function.shape.")
-    return nbl
+        raise ValueError("`nbl` must be an integer or tuple of (tuple of) integers"
+                         "of length `function.ndim`.")
+    return tuple(nb_pad)
 
 
 def initialize_function(function, data, nbl, mapper=None, mode='constant',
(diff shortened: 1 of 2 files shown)


devito/builtins.py
```diff
@@ -205,6 +205,24 @@ def fset(f, g):
     return f
 
 
+def nbl_to_padsize(nbl, ndim):
```
rhodrin, line 208: I'm tried, so maybe reading this function wrong, but doesn't look like it currently supports input of the form `((20, 30), 20, 10)` etc. which it probably should i.e. Just need to modify this
```
        if len(as_tuple(nb_t[0])) == 1:
            nbl = tuple((nbd, nbd) for nbd in nb_t)
```
to loop through each dim and do the check.

mloubout: Rushed it indeed nice catch, fixed

devito/builtins.py
```diff
@@ -283,16 +306,10 @@ def initialize_function(function, data, nbl, mapper=None, mode='constant',
...
-    elif len(as_tuple(nbl)) == function.ndim:
-        pass
-    else:
-        raise ValueError("nbl must be an integer or tuple of integers of length" +
-                         " function.shape.")
+    nbl = nbl_to_padsize(nbl, function.ndim)
 
-    slices = tuple([slice(n, -n) for _, n in zip(range(function.grid.dim),
-                                                 as_tuple(nbl))])
+    slices = tuple([slice(n[0], -n[1]) for _, n in zip(range(function.grid.dim),
```
FabioLuporini, line 311: not a big fan of all these n[0] and n[1] ... we can instead make it implicit in the iteration, which has the added value of implicitly checking the size of the tuple 

```
=  tuple([slice(n0, -n1) for _, (n0, n1) in zip(range(function.grid.dim),
```

or alternatively here  you could have just unpacked `n` with `*n`

devito/builtins.py
```diff
@@ -307,21 +324,21 @@ def initialize_function(function, data, nbl, mapper=None, mode='constant',
...
 
         def buff(i, j):
-            return [(i + k - 2*max(nbl)) for k in j]
+            return [(i + k - 2*max(max(nbl))) for k in j]
 
         b = [min(l) for l in (w for w in (buff(i, j) for i, j in zip(local_size, halo)))]
         if any(np.array(b) < 0):
             raise ValueError("Function `%s` halo is not sufficiently thick." % function)
 
     for d, n in zip(function.space_dimensions, as_tuple(nbl)):
```
FabioLuporini, line 333: (n0, n1)

devito/builtins.py
```diff
@@ -205,6 +205,29 @@ def fset(f, g):
     return f
 
 
+def nbl_to_padsize(nbl, ndim):
+    """
+    Creates the pad sizes from `nbl`. The output is a tuple of tuple
+    such as ((40, 40), (40, 40)) where each subtuple is the left/right
+    pad size for the dimension
```
FabioLuporini, line 212: full stop

devito/builtins.py
```diff
@@ -205,6 +205,29 @@ def fset(f, g):
...
+    elif len(nb_total) == ndim:
+        nb_pad = []
+        for nb_dim in nb_total:
+            if len(as_tuple(nb_dim)) == 1:
+                nb_pad.append((nb_dim, nb_dim))
+            else:
+                assert len(nb_dim) == 2
+                nb_pad.append(nb_dim)
+    else:
+        raise ValueError("nbl must be an integer or tuple of integers of length" +
```
FabioLuporini, line 226: quotes around nbl and around function.shape

also, no need for '+'

devito/builtins.py
rhodrin: If going down this route I'd prefer `nl`/`nleft` and `nr`/`nright` over `n0`, `n1`.

mloubout: done

devito/builtins.py
mloubout: done

devito/builtins.py
mloubout: fixed

devito/builtins.py
mloubout: fixed

PR #1307, opened 2020-05-21
Fix lowering of ConditionalDimensions Fix #1271

### LINKED ISSUE #1271: ConditionalDimension indexing does not follow SteppingDimension
A ConditionalDimension that uses a stepping DImension in its condition does not generate the code expected.
The Conditional is indexed using `t` instead of `t0` or `t1`.

Related culprit:
https://github.com/devitocodes/devito/blob/master/devito/ir/iet/scheduler.py#L76

Slack discussion:
https://devitocodes.slack.com/archives/CHRKMGAE4/p1588667937209500

MFE:
```
from devito import Dimension, ConditionalDimension, Function, TimeFunction, Eq, Operator, Grid
from sympy import And
from devito.symbolics import CondEq
import pdb
nx, ny = 4 , 4

shape = (nx, ny)
grid = Grid(shape=shape)

x, y = grid.dimensions
t = grid.stepping_dim
time = grid.time_dim 

g = TimeFunction(name='g', shape=(2,nx,ny), dimensions=(t,x,y))

ci = ConditionalDimension(name='ci', parent=y, condition=CondEq(g[t,x,y], 0))
op = Operator(Eq(g.forward, g + 1, implicit_dims=ci))
print(op.ccode)
op.apply()
```

FabioLuporini:


tests/test_dimension.py
```diff
@@ -831,6 +831,29 @@ def test_implicit_dims(self):
 
         assert np.all(f.data == F)
 
+    def test_stepping_dim_in_condition_lowering(self):
+        """
+        Check that the compiler performs the following lowering::
+
+        if f[t, x] > 2 {            if f[time, x] > 2 {
```
mloubout, line 838: not sure I follow the docstring with the code below

tests/test_dimension.py
```diff
@@ -831,6 +831,29 @@ def test_implicit_dims(self):
...
+          ...                -->      ...
+        }                           }
+        """
+        grid = Grid(shape=(4, 4))
+        _, y = grid.dimensions
+
+        ths = 10
+        g = TimeFunction(name='g', grid=grid)
+
+        ci = ConditionalDimension(name='ci', parent=y, condition=Le(g, ths))
```
mloubout, line 848: ANy reason why `y` is the parent?

georgebisbas: Innermost loop so that you nest all dims. But this is not a good answer.
I ask myself? So if x was parent what shoud the value of y index on the condition be?
Ideally I need to write an MFE here or other example to have conditionals iterating over one dim while other is constant. i.e. explicitly:

`         ci = ConditionalDimension(name='ci', parent=x, condition=Le(g[x,5], ths)) `
THoughts?

devito/ir/iet/visitors.py
```diff
@@ -743,6 +743,12 @@ def __init__(self, mapper=None, replacer=None):
         super(XSubs, self).__init__()
         self.replacer = replacer or (lambda i: uxreplace(i, mapper))
 
+    def visit_Conditional(self, o):
+        condition = self.replacer(o.condition)
+        then_body = self._visit(o.then_body)
```
georgebisbas, line 748: As said on slack, I was not thinking near here. Thanks.

tests/test_dimension.py
georgebisbas: I thought a bit more,

Would we like to have somethingl ike that?

ci = ConditionalDimension(name='ci', parent=x, condition=(g[x, 2] > ths))

mloubout: Possibly yes.

devito/ir/iet/visitors.py
FabioLuporini: > I was not thinking near here.

it's symptomatic you have to keep studying how this works

FabioLuporini:
@mloubout I suggest @georgebisbas adds all the necessary edits to take this PR to a mergeable state as per your suggestions. I'll hit the merge button when you'll formally have approved it

georgebisbas:
> @mloubout I suggest @georgebisbas adds all the necessary edits to take this PR to a mergeable state as per your suggestions. I'll hit the merge button when you'll formally have approved it

ok

tests/test_dimension.py
georgebisbas: So conceptually:
```
nx, ny = 4 , 4

shape = (nx, ny)
grid = Grid(shape=shape)

x, y = grid.dimensions

h = Function(name='h', grid=grid, dimensions=(x,), shape=(shape[0],))

cond = Le(h, 5)
ci = ConditionalDimension(name='ci', parent=x, condition=cond)

eq0 = Eq(h, g[x, 1])
eq1 = Eq(g, g + 1, implicit_dims=ci)
op = Operator([eq0, eq1])
print(op.ccode)
op.apply()
```

produces:
```
    {
      h[x + 1] = g[x + 1][2];
      if (h[x + 1] <= 5)
      {
        #pragma omp simd aligned(g:32)
        {
          g[x + 1][y + 1] = g[x + 1][y + 1] + 1;
        }
      }
    }
```
Ultimate goal should be to have CD able to be explicitly defined.
Currently it's not. SHould that be a different PR?

georgebisbas: Ideally, we need a way to get rid of `h`.

tests/test_dimension.py
mloubout: - the function is called `g` not `f`
- why is there two if
- why is it ` > 2` when the test is `<10`

This docstring makes it really hard to follwo what this test is about

georgebisbas: ok, did not realize you were matching the docs absolutely.
- Matched docstring
- If before transformation, if after XSubs
- Matched docstring

PR #1308, opened 2020-05-22
add MatrixSparseTimeFunction to support multi-point sources

tjb900:
This adds a `MatrixSparseTimeFunction` class, which allows a single time-series of source data to be injected (with different weights) at, or sampled from, multiple physical locations in the grid.

The typical use case is to inject a dipole source, or sample a dipole receiver (where the weights are +1 for one injection point and -1 for the other).  This is used, for example, if one wishes to model ghosting-type behaviour without using a free surface, and without storing twice as much data to do so.

### DIFF
--- a/devito/base.py
+++ b/devito/base.py
@@ -48,6 +48,10 @@ class PrecomputedSparseTimeFunction(with_metaclass(_BackendSelector, sparse.Prec
     pass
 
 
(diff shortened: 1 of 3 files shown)


FabioLuporini:
superseded by #1603

PR #1313, opened 2020-05-25
optimisation of parameter substitution in MPIRegion.arg_values

tjb900:
I forget why xreplace didn't work here, but it didn't.

This saves more than 40 seconds of python overhead per operator.apply (admittedly on a CPU with fairly ordinary single-thread performance), in MPI mode.

The same code might well be reusable in the performance summary generation, where we've also seen poor performance due to `subs`.

### DIFF
--- a/devito/mpi/routines.py
+++ b/devito/mpi/routines.py
@@ -15,7 +15,7 @@ from devito.ir.iet import (Call, Callable, Conditional, Expression, ExpressionBu
 from devito.ir.support import AFFINE, PARALLEL
 from devito.mpi import MPI
 from devito.symbolics import (Byref, CondNe, FieldFromPointer, FieldFromComposite,
-                              IndexedPointer, Macro)
+                              IndexedPointer, Macro, subs_op_args)
 from devito.tools import OrderedSet, dtype_to_mpitype, dtype_to_ctype, flatten, generator
 from devito.types import Array, Dimension, Symbol, LocalObject, CompositeObject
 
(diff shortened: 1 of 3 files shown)


tjb900:
I think there might be some other problem causing CI to barf - at least the errors I've looked at do not seem to relate to my changes even remotely.

devito/mpi/routines.py
```diff
@@ -1108,11 +1118,11 @@ def _arg_values(self, args, **kwargs):
             for a in self.arguments:
                 if a.is_Dimension:
                     a_m, a_M = mapper[a]
-                    setattr(entry, a.min_name, mapper[a][0].subs(args))
-                    setattr(entry, a.max_name, mapper[a][1].subs(args))
+                    setattr(entry, a.min_name, _simple_subs(a_m, args))
+                    setattr(entry, a.max_name, _simple_subs(a_M, args))
```
mloubout, line 1122: sorry, what's the difference with `a_m.subs(args)`, `a_M.subs(args)` here? Not sure I get it. I guess the `k in exprfree_symbols` prefilters ?

tjb900: Exactly.  Otherwise we spend a lot of time attempting to substitute in values that actually don't appear in the LHS - but sympy takes forever to figure that out.

mloubout: Can it be applied other places or is it the only critical one?

FabioLuporini: I think we should apply it to profiling.py when we currently do `subs(args)`. I'll take a look later. We should also probably move this utility somewhere else, perhaps under symbolics

PR #1314, opened 2020-05-25
add guard for empty regions in mpi=full

tjb900:
(apologies for the lack of tests here, just putting up fixes that we needed for our code to work)

In `mpi=full`, we can sometimes have zero-extent parallel loops but with the block sizes set to nonzero, so the old check didn't work.  I think there were also cases where these loops were not alone in a function, so the `return` as a way of not executing had unintended consequences - so this instead nests the protected code inside a guarding conditional.

### DIFF
--- a/devito/passes/iet/openmp.py
+++ b/devito/passes/iet/openmp.py
@@ -3,13 +3,13 @@ import os
 
 import numpy as np
 import cgen as c
-from sympy import Or, Max, Not
+from sympy import And, Max, Not
 
 from devito.ir import (DummyEq, Conditional, Block, Expression, ExpressionBundle, List,
-                       Prodder, Iteration, While, FindSymbols, FindNodes, Return,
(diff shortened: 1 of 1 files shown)


FabioLuporini:
@mloubout I had a go at this, and I ended up pushing to a separate branch. Reason is twofold:

* with this patch, **all** ompized loops will be wrapped within an extra if-then:
```
if p_src_M - p_src_m + 1 > 0 {
  #pragma omp parallel {
    int chunk_size = ...
    #pragma omp for
    ...
}
```
which is a bit horrible, considering that this, AFAICT, is just a work around to a bug in the Intel compiler which segfaults with trip-count or incr == 0. My revised version instead limits the extra if-then to the blocked loops. Is it enough? not sure yet, we need MFEs

* hypothesis: only collapse(n) with n>1 triggers this bug. Again, we need MFEs

FabioLuporini:
closing as superseded by #1673

PR #1316, opened 2020-05-25
Avoid dererencing dangling pointers on GC'd objects

tjb900:
This patch ensures that python holds a reference to objects that it passes to C for the duration of the call.  It was possible for this to not happen since the ctypes pointers are not sufficient to keep objects alive, but memoization of `arg_defaults`/`arg_values` had been covering up this issue.

Previously we didn't run into this since some of the functions that returned these objects
were memoized, which means they hung onto the objects indefinitely.  Removal of the memoization caused them to become eligible for GC, resulting in segfaults and other fun.


FabioLuporini:
@mloubout @rhodrin I don't know who, how and when, but this one needs clean up and an MFE

PR #1317, opened 2020-05-25
deps: update for sympy 1.6

mloubout:


devito/types/caching.py
```diff
@@ -2,6 +2,8 @@
 import weakref
 
 import sympy
+import sympy.core.cache as cache
```
mloubout, line 5:
```
importing sympy.core.cache with 'from sympy import *' has been
deprecated since SymPy 1.6. Use import sympy.core.cache instead. See
https://github.com/sympy/sympy/issues/18245 for more info.
```

mloubout:
There is deeper issues with 1.6 I'm looking into (we end up with `Dummy` in the middle of the expression n some cases like acoustic OT4)

mloubout:
Depends on
https://github.com/sympy/sympy/pull/19431

PR #1318, opened 2020-05-27
Improve CIRE

FabioLuporini:
... also dropped yreplace!

~@mloubout I also need a favor from you. I also decided to enforce all Arrays on the heap to avoid segfaults due to allocating to much stuff on the stack as happened to @jkwashbourne with his VTI. The alternative was a runtime check, but then what? we would have needed an escape hatch (a further opt knob) for users to tell devito "sorry, I got this warning upon `op.apply(...)`, can you actually recompile the Operator with Arrays on the heap". Which is horrible. So, I was saying -- everything is on the heap now, and I ran a few tests and the penalty seems completely negligible with tiny grids such as `30**3` -- the penalty being around 1-2 % -- and virtually disappears with big ones (>= `300**3`). Can you run similar checks too with your in house TTIs ? or perhaps TTI adjoint, which I cannot easily run yet~

scratch that -- turns out it will require more work because I can't use the `private` clause, but rather I'd need the mallocs inside the parallel region. For another day


devito/core/cpu.py
```diff
@@ -30,7 +30,7 @@ class CPU64NoopOperator(OperatorCore):
     Number of CIRE passes to detect and optimize away Dimension-invariant expressions.
     """
 
-    CIRE_REPEATS_SOPS = 2
+    CIRE_REPEATS_SOPS = 5
```
FabioLuporini, line 33: note for reviewers: due to the change `2*n + 1` -> `n` that you'll see below

jkwashbourne:
Thanks for this, I have to tweak stack 2 different ways (limit and OMP_STACKSIZE) to get robust behavior. 

Do you need to allocate inside an initial parallel region to get scope for those things right? I mean can there be more than one parallel region using the same temps?

devito/passes/clusters/aliases.py
```diff
@@ -154,17 +122,123 @@ def callbacks_sops(context, n):
...
+    """
+
+    mode = None
+
+    def __new__(cls, context, n, min_cost):
+        min_cost = min_cost[cls.mode]
+        if callable(min_cost):
+            min_cost = min_cost(n)
+
+        return (partial(cls.extract, n, context, min_cost),
```
mloubout, line 139: Wait so this returns a tuple of class method wrapped in a partial? That feels a bit odd, can't it just be properties?
Also, I don't like partial (I know it's in FD and needs maybe rework), lambda are more efficient

FabioLuporini: with this pattern you will never be able to instantiate the class, but you can still use inheritance. What you get instead are three callbacks. I think this is fine, if you had properties you would have to call them separately, while instead I want them as a unique block because **all* of them are necessary at the same time. The Callbacks class defines what in Java would have been an Interface -- something not instantiable but that still defines what the subclasses must implement

about `partial` -- I think you're too stressed about it, this is called once or twice per compilation and we're probably just talking about the 0.0001% of the turnaround time

FabioLuporini:
btw, still missing a test, but I wanted to file a PR so that @jkwashbourne could try it as I wrote on Slack

PR #1319, opened 2020-05-28
fix pow_to_mul to handle float-represented integer powers

tjb900:
Without this, integer powers that are constructed from floats result in the following error:

```
            elif exp > 0:
>               return sympy.Mul(*[base]*exp, evaluate=False)
E               TypeError: can't multiply sequence by non-int of type 'Float'
```


FabioLuporini:
thanks, merging

PR #1321, opened 2020-05-29
tensor: fix matmat prod

mloubout:
Tiny fix, row/col was flipped

### DIFF
--- a/devito/types/tensor.py
+++ b/devito/types/tensor.py
@@ -133,7 +133,7 @@ class TensorFunction(AbstractTensor, Differentiable):
 
             def entry(i):
                 return sum(self[i, k]*other[k] for k in range(self.cols))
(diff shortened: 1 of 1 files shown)


devito/types/tensor.py
```diff
@@ -145,8 +145,9 @@ def entry(i):
 
             def entry(i, j):
                 return sum(self[i, k]*other[k, j] for k in range(self.cols))
-            comps = [[entry(i, j) for i in range(self.cols)]
-                     for j in range(self.rows)]
+            comps = [[entry(i, j) for j in range(self.cols)]
```
georgebisbas, line 148: nitpicking but Isn't it better to have first i and then j?
```             
comps = [[entry(i, j) for i in range(self.rows)] 
for j in range(self.cols)]             
```

PR #1322, opened 2020-05-30
Bench update

mloubout:
- Enables adjoint/born/gradien if implemented
- Addedd ssa operators

Needs tweaking for `born`
Needed for https://github.com/devitocodes/thematrix/issues/30


mloubout:
>but the run-jit-backdoor and bench mode won't support it

I think bench is fine as it just calls run a couple times, jit may need a tweak yeah.

benchmarks/user/benchmark.py
```diff
@@ -41,10 +43,43 @@
...
+    """
+    Initialize any necessary input and run the operator associated with the solver.
+    """
+    # Get the operator if exist
+    try:
+        op = getattr(solver, operator)
+    except AttributeError:
+        raise AttributeError("Operator %s not implemented for %s" % (operator, solver))
+
+    # THis is a bit ugly but not sure how to make clean input creation for different op
```
mloubout, line 65: @FabioLuporini  this is not very pretty at all, but each operator have different mandatory inputs (trying to avoid computing zeros for benchmark) so this do the job for now.

PR #1328, opened 2020-06-02
enforce user provided x0

mloubout:
Small fix that makes sure `eval_at` does not overwritte user provided x0 (via `u.dx(x0=....)`)

### DIFF
--- a/devito/finite_differences/derivative.py
+++ b/devito/finite_differences/derivative.py
@@ -269,7 +269,8 @@ class Derivative(sympy.Derivative, Differentiable):
         setup where one could have Eq(u(x + h_x/2), v(x).dx)) in which case v(x).dx
         has to be computed at x=x + h_x/2.
         """
-        x0 = dict(func.indices_ref._getters)
+        # If an x0 already exists do not overwrite it
+        x0 = self.x0 or dict(func.indices_ref._getters)
         if self.expr.is_Add:
             # Derivatives are linear and  the derivative of an Add can be treated as an
             # Add of derivatives which makes (u(x + h_x/2) + v(x)).dx` easier to handle
--- a/devito/types/basic.py
+++ b/devito/types/basic.py
@@ -950,7 +950,7 @@ class AbstractFunction(sympy.Function, Basic, Cached, Pickable, Evaluable):
         # Indices after substitutions
         indices = [(a - o + f).xreplace(s) for a, o, f, s in
                    zip(self.args, self.origin, shift, subs)]
-
+        indices = [i.xreplace({k: int(k) for k in i.atoms(sympy.Float)}) for i in indices]
         return self.indexed[indices]
 
     def __getitem__(self, index):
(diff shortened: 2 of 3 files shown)


mloubout:
Should take the opportunity to enforce INT indices

devito/finite_differences/derivative.py
```diff
@@ -269,7 +269,8 @@ def _eval_at(self, func):
         setup where one could have Eq(u(x + h_x/2), v(x).dx)) in which case v(x).dx
         has to be computed at x=x + h_x/2.
         """
-        x0 = dict(func.indices_ref._getters)
+        # If an x0 already exist do not overwrite it
```
rhodrin, line 272: exist**s**

devito/finite_differences/derivative.py
```diff
@@ -269,7 +269,8 @@ def _eval_at(self, func):
         setup where one could have Eq(u(x + h_x/2), v(x).dx)) in which case v(x).dx
         has to be computed at x=x + h_x/2.
         """
-        x0 = dict(func.indices_ref._getters)
+        # If an x0 already exists do not overwrite it
+        x0 = self.x0 or dict(func.indices_ref._getters)
```
FabioLuporini, line 273: btw, remind me what's the difference between func.indices._getters and func.dimensions ?

mloubout: It's a dict `{dim: index_ref}`. So if not staggered is `{d: d fr d in dimensions}` and for staggered `{d: d+ .5spacing if stagg is d else x fir d in dimensions}`

devito/types/basic.py
```diff
@@ -948,7 +948,7 @@ def indexify(self, indices=None, lshift=False, subs=None):
         # Add halo shift
         shift = self._size_nodomain.left if lshift else tuple([0]*len(self.dimensions))
         # Indices after substitutions
-        indices = [(a - o + f).xreplace(s) for a, o, f, s in
+        indices = [sympy.nsimplify((a - o + f).xreplace(s)) for a, o, f, s in
```
mloubout, line 951: This basically does `x + 1.0` => `x + 1` which has been floating around in some cases and erroring compilation.

georgebisbas: That means that somewhere the offset is being considered/transformed as/to float?

mloubout: depending on staggered/x0 yes, where you can have like `x + .5*x.spacing` for example

georgebisbas: ok, thanks

FabioLuporini: but this could cause internal reconstruction thus slowing down compilation. 

What's the cause of 1.0? I think we should rather intercept that and cast it to int

mloubout: > think we should rather intercept that and cast it to int

That's what it does, not sure what it would do better somewhere else, this is only doing it on the final index (i.e `x+1.0` which is tiny) rather than having to filter and `x + .5 * h_x` and figure out what to do with it and create Rationals, .....

Issue #1250, opened 2020-04-23
Non-padding of the halo could cause incorrect computed values


speglich:
The following example results in NaN values.

```python

    grid = Grid(shape=(5,5))
 
    v = VectorTimeFunction(name="v", grid=grid, time_order=1, space_order=4)
    p = TimeFunction(name="p2", grid=grid, time_order=1, space_order=4, staggered=NODE)

    v[0].data[:] = 5.
    v[1].data[:] = 5.
 
    den = np.zeros((5,5), dtype=np.float32)
    den[:] = 2.

    B = Function(name="B", grid=grid, space_order=4, staggered=NODE)
    initialize_function(B, den, 0, dtype=np.float32)
    b = 1./B

    eq = Eq(p, div(b*v))
    op = Operator([eq])
    op.apply(time_M=0)
```

 
This NaN values are coming from non-padding of the halo in B. More precisely this occurs because the halo, non padded, is used in the derivative on B leading to 1/0.

As suggest by @mloubout padding the B as ` B._data_with_outhalo[:] = 2 ` solve this problem manually for this case.

FabioLuporini:
a different symptom of the same issue is given in #1432

PR #1330, opened 2020-06-03
model: fix nan default dm

mloubout:


examples/seismic/model.py
```diff
@@ -336,7 +336,7 @@ def dm(self):
         as `dm = div(m)`.
```
FabioLuporini, line 336: ? div(vp)

examples/seismic/model.py
```diff
@@ -336,7 +336,7 @@ def dm(self):
...
         """
         dm = Function(name="dm", grid=self.grid, space_order=self.space_order)
-        Operator(Eq(dm, div(self.m)))()
+        Operator(Eq(dm, div(self.vp)), subs=self.spacing_map)()
```
FabioLuporini, line 339: I guess the fix is self.m -> self.vp ?

mloubout: Yes, `m=1/vp**2` which creates `nan` at the edge( #1250 )

mloubout:
Fixed the text

>Also, nothing needs to be modified where model.dm is called in benchmarks/user/benchmark.py?

No nothing to change there, only issue was that `model.dm` had NaNs due to div(1/vp) at the edge

rhodrin:
Cool, lets merge when green and give `TheMatrix` a spin!

PR #1348, opened 2020-06-10
Refine CFL condition in seismic

mloubout:
refine elastic cfl according to Sei paper
https://doi.org/10.1137/0916052


devito/types/basic.py
```diff
@@ -729,7 +729,7 @@ def dimensions(self):
     def _eval_deriv(self):
         return self
 
-    @cached_property
+    @property
```
FabioLuporini, line 732: this is **extremely** scary

examples/seismic/acoustic/acoustic_example.py
```diff
@@ -62,6 +63,15 @@ def run(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
     return summary.gflopss, summary.oi, summary.timings, [rec, u.data]
 
 
+@pytest.mark.parametrize('ndim', [1, 2, 3])
+def test_isoacoustic_stability(ndim):
+    shape = tuple([11]*ndim)
+    spacing = tuple([20]*ndim)
+    _, _, _, [rec, _] = run(shape=shape, spacing=spacing, tn=20000.0, nbl=0)
+    norm = lambda x: np.linalg.norm(x.data.reshape(-1))
```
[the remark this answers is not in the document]
FabioLuporini: indeed

examples/seismic/acoustic/acoustic_example.py
```diff
@@ -62,6 +63,15 @@ def run(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
...
+    assert np.isfinite(norm(rec))
```
FabioLuporini, line 72: should we also not assert against some pre-computed value? a value around 10**7 is finite, but ...

examples/seismic/model.py
```diff
@@ -1,5 +1,6 @@
 import numpy as np
-from sympy import sin, Abs, finite_diff_weights
+from sympy import sin, Abs
```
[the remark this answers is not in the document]
mloubout: fixed

examples/seismic/viscoelastic/viscoelastic_example.py
```diff
@@ -28,16 +29,26 @@ def run(shape=(50, 50), spacing=(20.0, 20.0), tn=1000.0,
     info("Applying Forward")
     # Define receiver geometry (spread across x, just below surface)
     rec1, rec2, v, tau, summary = solver.forward(autotune=autotune)
-
+    from devito import norm
+    print(norm(rec1), norm(rec2), norm(v[0]))
```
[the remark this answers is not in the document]
mloubout: fixed

examples/seismic/acoustic/acoustic_example.py
mloubout: wouldn't really matter. That's ~10k timesteps it's either finite or NaN if unstable so `isfinite` i all needed to know it's stable.

devito/types/basic.py
FabioLuporini: as discussed offline (slack) this is necessary due to our caching infrastructure which assigns `__dict__` (for performance) rather than copying it, thus preventing us from selectively dropping certain entries from the cached init.

examples/seismic/acoustic/operators.py
```diff
@@ -17,7 +17,8 @@ def freesurface(model, eq):
     eq : Eq
         Time-stepping stencil (time update) to mirror at the freesurface.
     """
-    lhs, rhs = eq.evaluate
+    eq = eq.evaluate
+    lhs, rhs = eq.lhs, eq.rhs
```
FabioLuporini, line 21: nitpicking: `lhs, rhs = eq.evaluate.args `

examples/seismic/acoustic/acoustic_example.py
FabioLuporini: ahhhh right fair point

devito/types/basic.py
```diff
@@ -729,7 +729,7 @@ def dimensions(self):
...
     def _is_on_grid(self):
         """
         Check whether the object is on the grid or need averaging.
```
rhodrin, line 735: I see it's not from this PR, but should be `... needs averaging.`.

rhodrin: or better still, `requires`.

examples/seismic/model.py
```diff
@@ -77,7 +76,7 @@ def __init__(self, so):
 
     def define(self, dimensions):
         """
-        Definition of the top part of the domain for wrapped indices FS
+        Definition of the top part of the domain for wrapped indices FS.
```
rhodrin, line 79: `upper section of the domain`?

examples/seismic/viscoelastic/wavesolver.py
```diff
@@ -29,7 +29,7 @@ def __init__(self, model, geometry, space_order=4, **kwargs):
         self.geometry = geometry
 
         self.space_order = space_order
-        self.dt = self.model.critical_dt
+        self.dt = self.model.dtype(.9 * self.model.critical_dt)
```
rhodrin, line 32: Not sure I like this:
1. We are seeing instabilities without this factor?
2. Why this way instead of modifying the critical dt calc?

mloubout: 1. yes. You can look at Thorbecke's implementation where he basically says that with Q, it's not always stable and need to play with dt.
2. I can try to put in model, but that would probably make it smaller for some other examples too.

rhodrin: 1. Yeah, had seen that before.
2. Indeed, it's a bit tricky - I'm not sure if it's better to have our 'default' `critical_dt` air on the side of caution for 'robustness' or not? What do you think? But maybe we could add a `factor` `kwarg` which is by default 0.9 or 1.0 (don't really mind)? At least this way the api will robust/trivial to change for any tinkering down the line - and then people will easily be able to look at the docstring to understand why such factors are sometimes needed. But I think that some_factor*critical_dt with no comment should be avoided at least.

PR #1349, opened 2020-06-10
Heapification of temporaries

FabioLuporini:
This PR moves the allocation of temporary arrays from the stack to the heap. This means that one doesn't need to manually enforce bigger stack sizes to avoid segfaults due to blowing up the stack, which we saw happening in fancy codes with lots of temporaries.

Also, it ensures that **all** temporary arrays are defined once and for all right upon jumping into C land. With OpenMP, each thread allocates (mallocs) its own portion of array, which minimizes the likelihood of false sharing (though I still need to enforce padding along the innermost dimension -- for another day I'd say) and page faulting should we ever run flatten OpenMP across multiple sockets.

As a bonus, it improves the codebase in quite a few places (last but not least the pass responsible to place variable definitions)


devito/operator/operator.py
```diff
@@ -165,6 +168,10 @@ def __new__(cls, expressions, **kwargs):
     def _normalize_kwargs(cls, **kwargs):
         return kwargs
 
+    @classmethod
+    def _symbol_registry(cls):
+        return None
```
mloubout, line 173: Not an empty one `SymbolRegistry()`?

devito/passes/iet/definitions.py
```diff
@@ -119,29 +174,44 @@ def place_definitions(self, iet, **kwargs):
...
+        already_defined = list(iet.parameters)
+
         for k, v in MapExprStmts().visit(iet).items():
             if k.is_Expression:
                 if k.is_definition:
                     site = v[-1] if v else iet
                     self._alloc_scalar_on_low_lat_mem(site, k, storage)
                     continue
                 objs = [k.write]
+            elif k.is_Dereference:
```
mloubout, line 186: is_Dereference"d"?

devito/types/dimension.py
```diff
@@ -1060,6 +1061,34 @@ def __new__(cls, d, name):
     _pickle_kwargs = []
 
 
+class ThreadDimension(BasicDimension):
```
mloubout, line 1064: Not sure about this one, why isn't it just a `DefaultDimension` or part of `Thread`

devito/tools/utils.py
```diff
@@ -52,16 +52,18 @@ def as_tuple(item, type=None, length=None):
...
-def as_mapper(iterable, key=None):
+def as_mapper(iterable, key=None, get=None):
     """
     Rearrange an iterable into a dictionary of lists in which keys are
     produced by the function ``key``.
     """
     if key is None:
         key = lambda i: i
+    if get is None:
+        get = lambda i: i
```
mloubout, line 63: `get = get or lambda i: i`?

devito/core/operator.py
```diff
@@ -1,13 +1,54 @@
...
+        # {name -> generator()} -- to create unique names for symbols, functions, ...
+        self.counters = {}
+
+        # Special symbols
+        self.nthreads = nthreads
+        self.nthreads_nested = nthreads_nested
+        self.nthreads_nonaffine = nthreads_nonaffine
+        self.threadid = threadid
+
+    def make_name(self, prefix=None):
```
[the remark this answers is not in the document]
FabioLuporini: it's making a name afterall

devito/ir/iet/nodes.py
```diff
@@ -734,6 +751,44 @@ def defines(self):
         return ()
 
 
+class Dereference(ExprStmt, Node):
+
+    """
+    A node encapsulating a dereferentiation from a PointerArray to an Array.
```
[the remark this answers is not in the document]
FabioLuporini: should be dereference, will fix it

devito/ir/iet/nodes.py
```diff
@@ -866,6 +921,24 @@ def periodic(self):
...
+
+class ParallelBlock(Block):
+
+    """
+    A sequence of Nodes, wrapped in a parallel block {...}.
+    """
+
+    is_ParallelBlock = True
+
+
```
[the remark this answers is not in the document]
FabioLuporini: it's deliberately vague/abstract because it might be anything (to be specified by the subclasses)

devito/operator/operator.py
FabioLuporini: On purpose None because this is an abstract Operator and not all possible concrete classes may need one. Our subclasses in `core` do, though

devito/passes/iet/definitions.py
FabioLuporini: no, it's for a Dereference object

devito/tools/utils.py
FabioLuporini: yeah, I'm gonna change this

devito/types/array.py
```diff
@@ -0,0 +1,200 @@
...
+    @classmethod
+    def __indices_setup__(cls, **kwargs):
+        return as_tuple(kwargs['dimensions']), as_tuple(kwargs['dimensions'])
+
+    @property
+    def shape(self):
+        return self.symbolic_shape
+
+
+class Array(ArrayBasic):
```
FabioLuporini, line 28: this is basically lifted from basic.py, no new changes

devito/types/array.py
```diff
@@ -0,0 +1,200 @@
...
+
+    @cached_property
+    def free_symbols(self):
+        return super().free_symbols - {d for d in self.dimensions if d.is_Default}
+
+    # Pickling support
+    _pickle_kwargs = AbstractFunction._pickle_kwargs + ['dimensions', 'scope', 'sharing']
+
+
+class PointerArray(ArrayBasic):
```
FabioLuporini, line 150: this is new instead

devito/types/basic.py
```diff
@@ -967,118 +968,6 @@ def _pickle_reconstruct(self):
         return self.__class__.__base__
 
 
-class Array(AbstractFunction):
```
FabioLuporini, line 970: just moved to array.py

devito/types/dimension.py
FabioLuporini: what is `Thread` ?

I need a special Dimension because I need to be aware of the concept of "thread dimension", such that the declarator can place an `omp parallel region` around some mallocs (see `definitions.py`).

devito/types/array.py
```diff
@@ -0,0 +1,200 @@
...
+
+    """
+    Symbol representing a pointer to an Array.
+
+    Parameters
+    ----------
+    name : str
+        Name of the symbol.
+    dimensions : Dimension
```
mloubout, line 159: or tuple of Dimension?

devito/types/array.py
```diff
@@ -0,0 +1,200 @@
...
+        return kwargs['array'].dtype
+
+    @property
+    def _C_typename(self):
+        return ctypes_to_cstr(POINTER(POINTER(dtype_to_ctype(self.dtype))))
+
+    @property
+    def dim(self):
+        """Shortcut for self.dimensions[0]."""
+        return self.dimensions[0]
```
mloubout, line 193: So single dim even if it points to n-dim Array?

devito/types/array.py
FabioLuporini: no, single one

devito/types/dimension.py
mloubout: Not sure what you mean by "special dimension", this is just a generic fixed size dimension (so a DefaultDimension) that can be attached to one of the `NThreads` classes or something.

devito/types/array.py
FabioLuporini: yes. The "pointer dimension" is just one, ie in C it's like `int (*a)[..][..][..]..` , cannot be for example `int (**a)[..][..][..]`. This generalisation would have complicated the code, and we don't need it.

Basically, we need it for the "thread dimension"

devito/types/dimension.py
FabioLuporini: I need `is_Thread`

devito/ir/iet/nodes.py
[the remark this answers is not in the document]
FabioLuporini: fixed

devito/passes/clusters/aliases.py
```diff
@@ -16,24 +16,24 @@
...
         Input Cluster, subject of the optimization pass.
-    template : callable
-        To build the symbols (temporaries) storing the redundant expressions.
     mode : str
         The transformation mode. Accepted: ['invariants', 'sops'].
         * 'invariants' is for sub-expressions that are invariant w.r.t. one or
           more Dimensions.
         * 'sops' stands for sums-of-products, that is redundancies are searched
           across all expressions in sum-of-product form.
+    sregistry : SymbolRegistry
```
mloubout, line 33: Rephrase? Like "Devito symbol registery for temporaries. Creates unique temporary names...."

devito/operator/symbols.py
```diff
@@ -0,0 +1,33 @@
...
+        # {name -> generator()} -- to create unique names for symbols, functions, ...
+        self.counters = {}
+
+        # Special symbols
+        self.nthreads = nthreads
+        self.nthreads_nested = nthreads_nested
+        self.nthreads_nonaffine = nthreads_nonaffine
+        self.threadid = threadid
+
+    def make_name(self, prefix=None):
```
mloubout, line 23: nitpicking: I know was somewhere but can't find it. It does make a name but sounds like it's creating the name of the SymbolRegistery. Maybe something like "next_registered_symbol" ?

devito/passes/iet/engine.py
```diff
@@ -94,19 +96,50 @@ def apply(self, func, **kwargs):
...
+            if not args:
+                continue
+
+            def filter_args(v, efunc=None):
+                processed = list(v)
+                for _a in args:
+                    try:
+                        # Should the arg actually be dropped?
+                        a, drop = _a
+                        if drop:
```
mloubout, line 108:
```
try:
    if a[1] and a[0] in processed:
        processed.remove(a[0])
    continue
except AttributeError:
    a = _a 
```

devito/types/dimension.py
```diff
@@ -1060,6 +1064,36 @@ def __new__(cls, d, name):
...
+        self._symbolic_size = symbolic_size
+
+    @property
+    def symbolic_size(self):
+        return self._symbolic_size
+
+    def _arg_defaults(self):
+        return {}
+
+    def _arg_values(self, *args, **kwargs):
```
mloubout, line 1090: the BasicDimension  `_arg` don't work? I gues purely symbolic so no actual size.

devito/types/dimension.py
```diff
@@ -1060,6 +1064,36 @@ def __new__(cls, d, name):
...
+        return {}
+
+    # Pickling support
+    _pickle_kwargs = ['symbolic_size']
```
mloubout, line 1094: `BasicDimension._pickle_kwargs +  ['symbolic_size'] ` ?

devito/operator/symbols.py
FabioLuporini: it may sound like, but it's not a classmethod, so it can't be that. I mean, one of the two main points of the SymbolRegistry is to create unique names, so...

devito/passes/clusters/aliases.py
FabioLuporini: done, updated all docstrings concerning `sregistry`

devito/passes/iet/engine.py
FabioLuporini: I don't like explicit indexing into `a` , and I have a feeling that what you wrote is slightly different (semantically) than what I need here

devito/types/dimension.py
FabioLuporini: indeed. This is on purpose a no-op, like we do for the other Dimensions that are expected not to provide any arguments

devito/types/dimension.py
FabioLuporini: no, only `symbolic_size` needs to be pickled (there's no "spacing", like instead in the superclass, here)

also added a pickling test

PR #1352, opened 2020-06-12
Improve detection of redundant halo exchanges

FabioLuporini:
patch + MFE

### DIFF
--- a/devito/passes/iet/mpi.py
+++ b/devito/passes/iet/mpi.py
@@ -108,6 +108,27 @@ def _merge_halospots(iet):
     Merge HaloSpots on the same Iteration tree level where all data dependencies
     would be honored.
     """
+
+    # Merge rules -- if the retval is True, then it means the input `dep` is not
+    # a stopper to halo merging
+
+    def rule0(dep, hs, *args):
+        # E.g., `dep=W<f,[t1, x]> -> R<f,[t0, x-1]>` => True
+        return not any(d in hs.dimensions or dep.distance_mapper[d] is S.Infinity
+                       for d in dep.cause)
+
+    def rule1(dep, *args):
+        # TODO This is apparently never hit, but feeling uncomfortable to remove it
+        return dep.is_regular and all(not any(dep.read.touched_halo(d.root))
+                                      for d in dep.cause)
+
+    def rule2(dep, hs, loc_indices, *args):
+        # E.g., `dep=W<f,[t1, x+1]> -> R<f,[t1, xl+1]>` and `loc_indices={t: t0}` => True
+        return all(dep.distance_mapper[d] == 0 and dep.source[d] is not v
+                   for d, v in loc_indices.items())
+
+    merge_rules = [rule0, rule1, rule2]
+
     # Analysis
     mapper = {}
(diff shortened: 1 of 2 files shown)


FabioLuporini:
OK there's I think a minor thing that is left to deal with -- will do next week

devito/passes/iet/mpi.py
```diff
@@ -108,6 +108,27 @@ def _merge_halospots(iet):
...
+    # Merge rules -- if the retval is True, then it means the input `dep` is not
+    # a stopper to halo merging
+
+    def rule0(dep, hs, *args):
+        # E.g., `dep=W<f,[t1, x]> -> R<f,[t0, x-1]>` => True
+        return not any(d in hs.dimensions or dep.distance_mapper[d] is S.Infinity
+                       for d in dep.cause)
+
+    def rule1(dep, *args):
+        # TODO This is apparently never hit, but feeling uncomfortable to remove it
```
mloubout, line 121: Bit scary but OK 😄

devito/passes/iet/mpi.py
```diff
@@ -108,6 +108,27 @@ def _merge_halospots(iet):
...
+
+    # Merge rules -- if the retval is True, then it means the input `dep` is not
+    # a stopper to halo merging
+
+    def rule0(dep, hs, *args):
+        # E.g., `dep=W<f,[t1, x]> -> R<f,[t0, x-1]>` => True
+        return not any(d in hs.dimensions or dep.distance_mapper[d] is S.Infinity
+                       for d in dep.cause)
+
+    def rule1(dep, *args):
```
mloubout, line 120: Can wee keep same signature for all three `(dep,  hs, loc_indices)` for safety since they are all called the same way (kinda lie a template if one want to add more rules).

devito/passes/iet/mpi.py
```diff
@@ -122,22 +143,22 @@ def _merge_halospots(iet):
         for hs in halo_spots[1:]:
             mapper[hs] = hs.halo_scheme
 
-            for f in hs.fmapper:
+            for f, (loc_indices, _) in hs.fmapper.items():
```
mloubout, line 146: why not the same as before and just pass `hs` to rule and let it take what it needs in it.

devito/passes/iet/mpi.py
FabioLuporini: this is in master. I tried to drop it and all tests still pass. I think `rule0`, which I introduced not so long ago, generalised rule1 plus the other rule that was here before it. I just hadn't thought about that until this PR. I prefer to stick it around for a little longer, but we can probably safely drop it

devito/passes/iet/mpi.py
FabioLuporini: OK, pushed fix

devito/passes/iet/mpi.py
FabioLuporini: as explained on Slack, not sure I can drop this loop, because `f` is used below. I agree I could rewrite stuff somewhere in halo_scheme.py such that eventually this loop body becomes less contrieved (I don't like it either to pass `hs` and `loc_indices` to `rule`) , but unless you insist I'd leave it for another day?

PR #1353, opened 2020-06-13
Dummy PR to demonstrate that we are relying on memoization for correctness

tjb900:
Not a real PR - I expect will fail CI.

Filing to trigger a CI run and show crashes/incorrect results when we remove memoization from AbstractSparseFunction._arg_defaults

Relates to #1316


mloubout:
Should be fixed in #1316 , letting CI run to show it then closing

mloubout:
Fixed, closing

PR #1354, opened 2020-06-14
Start with two files: env setup and 2D FWI demo

deepearth:
PR from Microsoft Qie Zhang, as discussed with Gerard


rhodrin:
Thanks for the PR Qie. I think the plan for this should be as follows:
- The first notebook need not be a notebook, a wiki entry would suffice. Also, some material from notebook 2 could be moved to this wiki entry - how about a wiki entry re. setting up an Azure VM and running notebooks on the VM via a Jupyterhub? (Note a remote desktop to run the notebooks on the VM is not a recommended method, this should be done via Jupyterhub - or something similar).
- We've been discussing where to put content such as that in the second notebook. The structure of the notebook is very similar to the tutorial we already have,  but it's nice seeing this being applied to Marmousi - hence we're thinking of setting up a 'Blogs and examples' repo where material such as this can go (to also stop the main Devito repo from becoming too large). Will give you more details on that soon.

deepearth:
> Thanks for the PR Qie. I think the plan for this should be as follows:
> 
> * The first notebook need not be a notebook, a wiki entry would suffice. Also, some material from notebook 2 could be moved to this wiki entry - how about a wiki entry re. setting up an Azure VM and running notebooks on the VM via a Jupyterhub? (Note a remote desktop to run the notebooks on the VM is not a recommended method, this should be done via Jupyterhub - or something similar).
> * We've been discussing where to put content such as that in the second notebook. The structure of the notebook is very similar to the tutorial we already have,  but it's nice seeing this being applied to Marmousi - hence we're thinking of setting up a 'Blogs and examples' repo where material such as this can go (to also stop the main Devito repo from becoming too large). Will give you more details on that soon.

Thank you Rhodri for sharing your thoughts. Those are helpful.
> * You have a great point of having a repo outside main Devito repo. I prefer that approach too. The name can be like "Azure_blogs_examples". We will have a focus of how Devito examples (e.g., FWI) can be set up and run on Azure.
> * The purpose of those demos is to show end-to-end workflow, so users can repeat easily with a good experience. Yes I agree with you the first one sounds like a wiki entry. But I would like to keep the instruction and notebook in one place in order, so it's easier for a new user to find and follow. For a new user who is not familiar with the github structure, he may not know the existence of wiki. But I am open to the idea of moving it to wiki.
> * Great point on "running notebooks on the VM via a Jupyterhub" - will look into it. For now let's use remote desktop to keep it simple. Because with that, you can do all these: setup env / install Devito, run nvtop to check gpu status, download PGI package from firefox inside RD... A typical geophysicist would use script instead of jupyter notebook.
> * The second notebook is not only about Marmousi 2D FWI. Another purpose is to show it can flexibly adapt to various GPU/CPU arch on Azure, and how to do it on a single VM basis. That is why the top section of environment variables is included with the 2D FWI.

rhodrin:
As discussed, this PR has been/will be split into wiki entries + a notebook in a devitocodes 'binderhub' repo. Hence closing.

PR #1359, opened 2020-06-18
Fix missing halo exchange over invariants fields

FabioLuporini:
Also drop the unused `loop_wrapping` pass, which in turn enabled dropping some other nearly-dead stuff in data dependence analysis

PR equipped with MFE

### DIFF
--- a/devito/ir/support/basic.py
+++ b/devito/ir/support/basic.py
@@ -294,31 +294,41 @@ class TimedAccess(IterationInstance):
         other : TimedAccess
             The TimedAccess w.r.t. which the distance is computed.
         """
-        queue = list(zip(self.itintervals, other.itintervals))
-
         ret = []
-        for i, o, fi in zip(self, other, self.findices):
-            try:
-                iit, oit = queue.pop(0)
-            except IndexError:
-                # E.g., `self=R<f,[0,1]>` and `self.itintervals=()`
-                ret.append(i - o)
-                continue
+        for sit, oit in zip(self.itintervals, other.itintervals):
+            n = len(ret)
 
-            if iit != oit:
-                # E.g., `self=R<f,[x + 2]>` and `other=W<f,[i + 1]>`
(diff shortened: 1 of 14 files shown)


devito/passes/iet/mpi.py
```diff
@@ -62,6 +62,23 @@ def _hoist_halospots(iet):
...
     """
+
+    # Hoisting rules -- if the retval is True, then it means the input `dep` is not
+    # a stopper to halo hoisting
+
+    def rule0(dep, candidates):
+        # E.g., `dep=W<f,[x]> -> R<f,[x-1]>` and `candidates=(time, x)` => False
+        # E.g., `dep=W<f,[t1, x, y]> -> R<f,[t0, x-1, y+1]>`, `dep.cause={t,time}` and
+        #       `candidates=(x,)` => True
+        return (all(d in dep.distance_mapper for d in candidates) and
```
FabioLuporini, line 73: this line is the key fix, but to write it, I had to tweak/generalise (perhaps fix?) `TimedAccess.distance`

devito/ir/support/basic.py
```diff
@@ -292,50 +285,48 @@ def lex_le(self, other):
...
+            if not ai or ai._defines & sit.dim._defines:
+                # E.g., `self=R<f,[t + 1, x]>`, `self.itintervals=(time, x)` and `ai=t`
+                if sit.direction is Backward:
+                    ret.append(other[n] - self[n])
                 else:
-                    ret.append(i - o)
-            else:
-                # Mismatching `itinterval` => Infinity
+                    ret.append(self[n] - other[n])
+            elif fi in sit.dim._defines:
```
mloubout, line 328: is that the only too possible cases? No default `else` for uncaught ones?

FabioLuporini: no, if none of the two ifs are entered (the `if` and the `elif`), then we should just keep iterating

PR #1361, opened 2020-06-18
checkpointing: set right size_ckp

ofmla:
The `o.size_allocated `already takes into account the time order, i.e., it is a slightly larger value than `(3*nx*nz)` for a 2D second order `TimeFunction`. In this way, for a large number of checkpoints the allocated memory would be larger than the used memory  without using checkpointing. 
I made a simple change to correct this


ofmla:
Sorry, I complicated myself correcting a long line of code

ofmla:
Sorry, I complicated myself correcting a long line of code :/

navjotk:
Discussed and agreed. This is indeed a bug that was allocating more memory than required. LGTM.

navjotk:
Only test I can think of is to explicitly set the same "formula" in a test and compare? @ofmla do you have another idea?

mloubout:
Can't a simple size check work as a test? like for nt checkpoints assert `DevitoCheckpoint.size == TimeFunction(..., save=nt).shape`

navjotk:
That exact line can't work because `DevitoCheckpoint` defines the size of a single checkpoint, i.e. two time steps. 

But I think the spirit of @mloubout 's comment is the same as mine. We should write a test to ensure memory consumption never goes beyond the full manually unrolled version. Might have to expose an API in pyRevolve to enable that test here but that's fine.

PR #1362, opened 2020-06-18
Symbolic speedup

mloubout:
Minor tweaks to improve the speed of `evaluate` that is quite slow going through all the nobs of sympy. Bypasses a lot of it via direct call to `_subs` and avoid very expensive `evalf` on expression while only needed on the coefficients.

timing improvement for 12th order 3D on my laptop for this branch

```
In [1]: %timeit u.laplace.laplace.evaluate                                                                                                                                           
1.04 s ± 71.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [2]: %timeit u.laplace.laplace.evaluate                                                                                                                                           
1.06 s ± 132 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
vs master:

```
In [1]: %timeit u.laplace.laplace.evaluate   
6.23 s ± 1.02 s per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [2]: %timeit u.laplace.laplace.evaluate                                                                                                                                           
7.16 s ± 2.23 s per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [3]: %timeit u.laplace.laplace.evaluate                                                                                                                                           
5.56 s ± 268 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

```

And aside, `setup.cfg` updated for `flake8-nb` if someone wants to play with it but not adding it to CI.


devito/finite_differences/differentiable.py
```diff
@@ -288,6 +288,15 @@ def subs(self, *args, **kwargs):
         return self.func(*[getattr(a, 'subs', lambda x: a)(*args, **kwargs)
```
FabioLuporini, line 288: so this one doesn't call `_subs` directly, but, I'd guess, only via the superclass? that's a bit odd? or maybe not, actually

devito/finite_differences/finite_difference.py
```diff
@@ -270,13 +272,10 @@ def indices_weights_to_fd(expr, dim, inds, weights, matvec=1):
     # Loop through weights
     terms = []
     for i, c in zip(inds, weights):
-        try:
-            iloc = i.xreplace(mapper)
-        except AttributeError:
-            iloc = i
-        subs = {expr.indices_ref[dim]: iloc}
-        terms.append(expr.subs(subs) * c)
+        iloc = sympify(i).xreplace(mapper)
```
FabioLuporini, line 275: this block perhaps need a comment about what's going on

is the `sympify` there for the usual reasons (need for integer indices?)

FabioLuporini: isn't try-catch (on top of your changes) potentially faster?

devito/types/basic.py
```diff
@@ -770,6 +770,15 @@ def evaluate(self):
             return self
         return weight * sum(avg_list)
 
+    def _subs(self, old, new, **hints):
```
FabioLuporini, line 773: this looks identical to the one in Differentiable? Since DiscreteFunction is a subclass of both AbstractFunction and Differentiable, why don't we just drop it?

examples/seismic/preset_models.py
```diff
@@ -219,7 +219,7 @@ def demo_model(preset, **kwargs):
                              dtype=dtype, spacing=spacing, nbl=nbl, epsilon=epsilon,
                              delta=delta, theta=theta, phi=phi, bcs="damp", **kwargs)
 
-        if kwargs.get('smooth', True):
+        if kwargs.get('smooth', False):
```
FabioLuporini, line 222: why?

do we now get smoothing by default? when I run `python tti_example.py` for example

devito/types/basic.py
mloubout: This is not in Differentiable, only in DifferentiableOp.

examples/seismic/preset_models.py
mloubout: Yes that's what ` kwargs.get('smooth', True)` does, currently needs to explicitly ask for no smoothing and not easy in bench/examples (need extra cli arg)

devito/finite_differences/finite_difference.py
mloubout: >isn't try-catch (on top of your changes) potentially faster?

Not from what I tried no. WIll add comment

devito/finite_differences/differentiable.py
mloubout: Need subs for when there is more than one arg. `_subs` only accept `u._subs(old, new)` for dict/tuple of tuple this one is used. This is still a bit more efficient than the default sympy.expr version of it.

devito/types/basic.py
mloubout: Also previously missed small cases that can happen only for a function (or symbol) where you can have `u._subs(u, v)` which is not the case for an expression.

devito/finite_differences/finite_difference.py
FabioLuporini: That's fine, but for what is worth, I'd use a try-catch. That's what is for, and not using it in this case only makes things potentially slower.

devito/types/basic.py
FabioLuporini: I'm not sure I follow that last past... the `if old is self` at least should apply to generic exprs too?

I have a feeling, potentially wrong, that there's something off with this _subs being different than what we got in DifferentiableOp.

Other question: why is `Differentiable.subs` different than `DifferentiableOp.subs`? Should we rather try to have just one general `_subs` in `Differentiable`?

examples/seismic/tti/operators.py
```diff
@@ -75,22 +74,18 @@ def Gzz_centered(model, field, costheta, sintheta, cosphi, sinphi, space_order):
...
-    Gzz = (first_derivative(Gz * sintheta * cosphi,
-                            dim=x, side=centered, fd_order=order1,
-                            matvec=transpose) +
-           first_derivative(Gz * sintheta * sinphi,
-                            dim=y, side=centered, fd_order=order1,
-                            matvec=transpose) +
-           first_derivative(Gz * costheta,
-                            dim=z, side=centered, fd_order=order1,
-                            matvec=transpose))
+    Gz = -(sintheta * cosphi * field.dx(fd_order=order1) +
```
FabioLuporini, line 77: about time 😎

tests/test_dse.py
```diff
@@ -1411,7 +1411,7 @@ def test_fullopt_w_mpi(self):
 
     @switchconfig(profiling='advanced')
     @pytest.mark.parametrize('space_order,expected', [
-        (8, 173), (16, 307)
+        (8, 173), (16, 309)
```
FabioLuporini, line 1414: how are they different now?

fun that it only happens with the higher order one

devito/finite_differences/finite_difference.py
mloubout: Actually it makes thig potentially better since it guaranties indices are already sympy object  for subsequent manipulation.

FabioLuporini: yeah, but then you can use `sympify` only in the catch? so eventually you systematically have sympy objects

devito/finite_differences/differentiable.py
```diff
@@ -143,6 +143,23 @@ def __getattr__(self, name):
             return self._fd[name][0](self)
         raise AttributeError("%r object has no attribute %r" % (self.__class__, name))
 
+    def _eval_subs(self, old, new, **hints):
```
FabioLuporini, line 146: I guess the name `_eval_subs` (instead of `_subs`) is what sympy uses underneath ? so they have `subs` (public) and `_subs` + `_eval_subs` (private) ?

mloubout: Forget about it, reverting to previous commit any combination of moving to differentiable is about two times slower than previously

devito/types/basic.py
mloubout: > just one general _subs in Differentiable?

Way slower (no idea why) and also does weird stuff for non Add/Mul so why was originally moved to DifferentialOp

tests/test_dse.py
mloubout: Probably due to different order, the `evalf` was on the full fancy equation and may have been doing slughtly different stuff.

devito/types/basic.py
FabioLuporini: so why the below 

```
        if old is self:
             return new
         if old is new:
             return self
```

doesn't apply to DifferentiableOp?

devito/finite_differences/differentiable.py
```diff
@@ -288,6 +288,15 @@ def subs(self, *args, **kwargs):
         return self.func(*[getattr(a, 'subs', lambda x: a)(*args, **kwargs)
                            for a in self.args], evaluate=False)
 
+    def _subs(self, old, new, **hints):
```
FabioLuporini, line 291: out of curiosity, how is this different than the one in sympy?

devito/types/dense.py
```diff
@@ -1327,15 +1327,15 @@ def forward(self):
         i = int(self.time_order / 2) if self.time_order >= 2 else 1
         _t = self.dimensions[self._time_position]
 
-        return self.subs({_t: _t + i * _t.spacing})
+        return self._subs(_t, _t + i * _t.spacing)
```
FabioLuporini, line 1330: out of curiosity: was this change necessary for functionality or for speed?

FabioLuporini: ah ok, it's because now subs accepts a single pair `old,new` as two separate args

devito/finite_differences/differentiable.py
FabioLuporini: so, I just checked sympy.Basic._subs

so now this overriding does something different, so we should probably at least update the docstring

as I wrote in another comment below, I'm still a bit baffled by the lack of a check for identity (ie something along the lines as `if self is old then ...`. 

Also check this out: https://github.com/sympy/sympy/blob/master/sympy/core/basic.py#L998

devito/finite_differences/differentiable.py
```diff
@@ -288,6 +288,15 @@ def subs(self, *args, **kwargs):
...
+        args = list(self.args)
+        for i, arg in enumerate(args):
+            try:
```
FabioLuporini, line 294: OK, hang on... so is this overriding basically to avoid sympification ?

mloubout:
>, I got a question: why do we override _subs instead of _eval_subs, which seems to be the stub we need here?

`aresame`  lot more expensive than the `old is new` in the one here.

devito/types/dense.py
```diff
@@ -102,6 +102,8 @@ def __eq__(self, other):
         # is that they are indeed the same exact object
         return self is other
 
+    _subs = Differentiable._subs
```
FabioLuporini, line 105: this line should appear below all special methods lines ie those starting with `__` such as `__hash__`

mloubout: I prefer it here with `__eq__` that is sympy overwrite first then our stuff.

FabioLuporini:
merged

PR #1363, opened 2020-06-23
MPI + OpenMP 5.0

italoaug:
Enables multiple GPU offloading using MPI, OpenMP 5.0, and GPU Direct.


devito/core/gpu_openmp.py
```diff
@@ -14,6 +14,16 @@
...
+##
+from devito.tools import generator
+from devito.mpi.distributed import MPICommObject
+from devito.symbolics import Byref
+from devito.types import Symbol
+from devito.ir.equations import DummyEq
+from devito.mpi.routines import MPICallable, MPICall, GPUDirectMPICall
+from devito.ir.iet import Call, LocalExpression, List, FindNodes, Transformer, Block
+from sympy import Function
+from copy import deepcopy
```
georgebisbas, line 26: Imports must be ordered alphabetically and grouped according to type

devito/core/gpu_openmp.py
```diff
@@ -14,6 +14,16 @@
                                ParallelTree, optimize_halospots, mpiize, hoist_prodders,
                                iet_pass)
 from devito.tools import as_tuple, filter_sorted, timed_pass
+##
```
georgebisbas, line 17: all of the pep8 stuff/ flake8 can be fixed locally very quickly... inspecting through:
`flake8 ....changed files...```

devito/core/gpu_openmp.py
```diff
@@ -14,6 +14,16 @@
                                ParallelTree, optimize_halospots, mpiize, hoist_prodders,
                                iet_pass)
 from devito.tools import as_tuple, filter_sorted, timed_pass
```
georgebisbas, line 16: Some general guidelines:
some of this is written here: https://github.com/devitocodes/devito/blob/master/CONTRIBUTING.md#coding-guidelines

devito/core/gpu_openmp.py
```diff
@@ -211,6 +221,83 @@ def _(iet):
...
+
+    return iet, {}
+
+@iet_pass
+def mpiOmpizer(iet, **kwargs):
+    """
+    Modifies functions to enable multiple GPUs performing GPU-Direct communication
+    """
+    @singledispatch
+    def _mpiOmpizer(iet):
```
georgebisbas, line 282: functions/methods should be named with underscores (eg foo_bar , not camel fooBar)

devito/core/gpu_openmp.py
```diff
@@ -211,6 +221,83 @@ def _(iet):
...
+            body = [rank_decl, rank_init, ngpus_init, devicenum_init]
+
+            init = List(header=c.Comment('Begin of OpenMP5.0+MPI setup'),
+                    body=body,
+                    footer=(c.Comment('End of OpenMP5.0+MPI setup'), c.Line()))
+        else:
+            devicenum_init = LocalExpression(DummyEq(devicenum, 0))
+            body = [devicenum_init]
+
+            init = List(header=c.Comment('Begin of OpenMP5.0 setup'),
```
georgebisbas, line 259: duplicate code?

devito/core/gpu_openmp.py
```diff
@@ -306,8 +393,11 @@ def _specialize_iet(cls, graph, **kwargs):
         data_manager.place_definitions(graph)
         data_manager.place_casts(graph)
 
-        return graph
+        # Initialize OpenMP environment
+        initialize(graph)
```
georgebisbas, line 397: should that name be changed?

devito/core/gpu_openmp.py
```diff
@@ -334,8 +424,11 @@ def _specialize_iet(cls, graph, **kwargs):
...
```
georgebisbas, line 428: same

devito/core/gpu_openmp.py
```diff
@@ -211,6 +221,83 @@ def _(iet):
...
+@iet_pass
+def mpiOmpizer(iet, **kwargs):
+    """
+    Modifies functions to enable multiple GPUs performing GPU-Direct communication
+    """
+    @singledispatch
+    def _mpiOmpizer(iet):
+        devicenum = Symbol(name='devicenum')
+        mapper = {}
+        # adds devicenum to functions signatures and calls
```
FabioLuporini, line 285: comments should start with capital letter

FabioLuporini: they should also be in first person (as if imperative)

adds -> Add

devito/core/gpu_openmp.py
FabioLuporini: no, it's the same as in gpu_openacc, it's fine

devito/core/gpu_openmp.py
```diff
@@ -402,4 +495,8 @@ def _specialize_iet(cls, graph, **kwargs):
         data_manager.place_definitions(graph)
         data_manager.place_casts(graph)
 
-        return graph
+        # Initialize OpenMP environment
+        initialize(graph)
+        mpiOmpizer(graph)
```
FabioLuporini, line 500: this `mpi_ompizer` should be dropped and added to `_make_passes_mapper` . This is a Custom Operator. It only applies the passes that users ask it to apply, aside from the vey basic things

devito/mpi/routines.py
```diff
@@ -354,9 +354,9 @@ def _make_sendrecv(self, f, hse, key, **kwargs):
         count = reduce(mul, bufs.shape, 1)
         rrecv = MPIRequestObject(name='rrecv')
         rsend = MPIRequestObject(name='rsend')
-        recv = Call('MPI_Irecv', [bufs, count, Macro(dtype_to_mpitype(f.dtype)),
+        recv = GPUDirectMPICall('MPI_Irecv', [bufs, count, Macro(dtype_to_mpitype(f.dtype)),
```
FabioLuporini, line 357: mpi.py should be free of "gpu" concepts.

You're on the right track, but I'd do it differently:

* Instead of `MPICall` and `GPUDirectMPICall`, I'd add `IrecvCall` and `IsendCall`, subclasses of `Call`, because after all that's all you need
* Avoid changing all mpi calls -- aside from `MPI_Irecv` and `MPI_Isend`, all can remain `Call`

devito/core/gpu_openmp.py
italoaug: I don't think so. The code in _if_ statement is for MPI+GPU offloading, and the code in _else_ statement is for GPU offloading without MPI. They have different header, body, and footer.

devito/core/gpu_openmp.py
```diff
@@ -2,18 +2,26 @@
...
+                                 MPICallable)
 from devito.passes.clusters import (Lift, cire, cse, eliminate_arrays, extract_increments,
                                     factorize, fuse, optimize_pows)
 from devito.passes.iet import (DataManager, Storage, Ompizer, OpenMPIteration,
                                ParallelTree, optimize_halospots, mpiize, hoist_prodders,
                                iet_pass)
+from devito.symbolics import Byref
 from devito.tools import as_tuple, filter_sorted, timed_pass
+from devito.types import Symbol
+from sympy import Function
```
FabioLuporini, line 24: this import should be moved under `import cgen as c` as it's not a devito import, but rather an import from a devito dependency

devito/core/gpu_openmp.py
```diff
@@ -2,18 +2,26 @@
 
 import cgen as c
 
+from copy import deepcopy
```
FabioLuporini, line 5: this import should be moved under `from functools import ...` as it's not a devito import, but rather an import from the python library

also, I haven't checked yet what you use deepcopy for, but it's 99% wrong

devito/core/gpu_openmp.py
```diff
@@ -212,6 +221,83 @@ def _(iet):
...
+    Modifies functions to enable multiple GPUs performing GPU-Direct communication
+    """
+    @singledispatch
+    def _mpi_ompizer(iet):
+        mapper = {}
+        # Modify MPI_Isend and MPI_Irecv to perform GPU Direct communication
+        for node in FindNodes((IsendCall, IrecvCall)).visit(iet):
+            header = c.Pragma('omp target data use_device_ptr(%s) device(devicenum)' %
+                              node._args['parameters'][0].name)
+            mapper[node] = Block(header=header, body=deepcopy(node))
```
FabioLuporini, line 289: indeed. We never ever do deepcopies. If you look up the repo for deepcopies, you'll find none. Why do you have it here? can you not just drop it?

devito/core/gpu_openmp.py
```diff
@@ -212,6 +221,83 @@ def _(iet):
...
+
+    return iet, {}
+
+
+@iet_pass
+def mpi_ompizer(iet, **kwargs):
+    """
+    Modifies functions to enable multiple GPUs performing GPU-Direct communication
+    """
+    @singledispatch
```
FabioLuporini, line 282: you probably don't strictly need the `singledispatch` in this pass. You just apply the pass to all IET routines without making assumptions about where the MPI send/recv are, and you should be fine. In fact, the majority of Devito IET passes don't need `singledispatch`. You need `singledispatch` only when you need a pass to be applied to one or a specific type(s) of Callables only

devito/core/gpu_openmp.py
```diff
@@ -212,6 +221,83 @@ def _(iet):
...
+            header = c.Pragma('omp target data use_device_ptr(%s) device(devicenum)' %
+                              node._args['parameters'][0].name)
+            mapper[node] = Block(header=header, body=deepcopy(node))
+
+        iet = Transformer(mapper, nested=True).visit(iet)
+
+        return iet
+
+    iet = _mpi_ompizer(iet)
+    devicenum = Symbol(name='devicenum')
```
FabioLuporini, line 296: for various reasons, it might be better to use `Constant`, just like we do for `nthreads` for example

devito/core/gpu_openmp.py
```diff
@@ -306,6 +392,10 @@ def _specialize_iet(cls, graph, **kwargs):
         data_manager.place_definitions(graph)
         data_manager.place_casts(graph)
 
+        # Initialize OpenMP environment
+        initialize(graph)
+        mpi_ompizer(graph)
```
FabioLuporini, line 397: This is the `DeviceOpenMPNoopOperator` -- the "noop" in the name implies that no optimizations are performed. So you probably wanna remove `mpi_ompizer` from here

devito/core/gpu_openmp.py
```diff
@@ -212,6 +221,83 @@ def _(iet):
...
+
+    iet = _initialize(iet)
+
+    return iet, {}
+
+
+@iet_pass
+def mpi_ompizer(iet, **kwargs):
+    """
+    Modifies functions to enable multiple GPUs performing GPU-Direct communication
```
FabioLuporini, line 280: "Modify MPI Callables to enable ... 

also, full stop at the end

devito/core/gpu_openmp.py
```diff
@@ -212,6 +221,83 @@ def _(iet):
...
+    def _(iet):
+        return iet
+
+    iet = _initialize(iet)
+
+    return iet, {}
+
+
+@iet_pass
+def mpi_ompizer(iet, **kwargs):
```
FabioLuporini, line 278: "_ompizer" is sort of implicit to this `gpu_openmp` module... how about `mpi_gpu_direct` ? or something along these lines that highlights that what the pass does is "optimizing" MPI send/recvs via GPU direct

devito/core/gpu_openmp.py
italoaug: I've changed from `deepcopy` to `copy`. Turning it in a simple assignment (`body=node`) leads to `RecursionError: maximum recursion depth exceeded`.

FabioLuporini: never use copy or deepcopy

can you write an MFE for the case `body=node`  (literally 4-5 lines of code)?

italoaug: I've changed one line of code from `iet = Transformer(mapper, nested=True).visit(iet)` to `iet = Transformer(mapper).visit(iet)` and now it works fine without importing `copy`.

devito/core/gpu_openmp.py
```diff
@@ -212,6 +220,82 @@ def _(iet):
...
+def mpi_gpu_direct(iet, **kwargs):
+    """
+    Modify MPI Callables to enable multiple GPUs performing GPU-Direct communication.
+    """
+    def _mpi_gpu_direct(iet):
+        mapper = {}
+        # Modify MPI_Isend and MPI_Irecv to perform GPU Direct communication
+        for node in FindNodes((IsendCall, IrecvCall)).visit(iet):
+            header = c.Pragma('omp target data use_device_ptr(%s) device(devicenum)' %
+                              node._args['parameters'][0].name)
```
FabioLuporini, line 287: just `node.name` ?

see [here](https://github.com/devitocodes/devito/blob/master/devito/ir/iet/nodes.py#L238)

devito/core/gpu_openmp.py
```diff
@@ -212,6 +220,82 @@ def _(iet):
...
+
+    return iet, {'args': devicenum}
+
+
+@iet_pass
+def mpi_gpu_direct(iet, **kwargs):
+    """
+    Modify MPI Callables to enable multiple GPUs performing GPU-Direct communication.
+    """
+    def _mpi_gpu_direct(iet):
```
FabioLuporini, line 282: why do you need this nested function?

since you're not using single dispatch, and since you're calling it just once, I think it's uselss?

devito/mpi/routines.py
```diff
@@ -555,10 +555,10 @@ def _make_sendrecv(self, f, hse, key, msg=None):
         count = reduce(mul, sizes, 1)
         rrecv = Byref(FieldFromPointer(msg._C_field_rrecv, msg))
         rsend = Byref(FieldFromPointer(msg._C_field_rsend, msg))
-        recv = Call('MPI_Irecv', [bufs, count, Macro(dtype_to_mpitype(f.dtype)),
-                                  fromrank, Integer(13), comm, rrecv])
-        send = Call('MPI_Isend', [bufg, count, Macro(dtype_to_mpitype(f.dtype)),
-                                  torank, Integer(13), comm, rsend])
+        recv = IrecvCall('MPI_Irecv', [bufs, count, Macro(dtype_to_mpitype(f.dtype)),
```
FabioLuporini, line 558: instead of harcoding `'MPI_Irecv'` here, I'd just change the corresponding class as

```
class IrecvCall(Call):

    def __init__(self, parameters):
          super(IrecvCall, sell).__init__('MPI_Irecv', parameters)
```

devito/mpi/routines.py
```diff
@@ -555,10 +555,10 @@ def _make_sendrecv(self, f, hse, key, msg=None):
...
+                                       fromrank, Integer(13), comm, rrecv])
+        send = IsendCall('MPI_Isend', [bufg, count, Macro(dtype_to_mpitype(f.dtype)),
```
FabioLuporini, line 560: same story as above

devito/operator/operator.py
```diff
@@ -956,6 +956,7 @@ def parse_kwargs(**kwargs):
     options.setdefault('cire-mincost-sops', opt_options.get('cire-mincost-sops', None))
     options.setdefault('openmp', openmp)
     options.setdefault('mpi', configuration['mpi'])
+    options.setdefault('gpu-direct', False)
```
FabioLuporini, line 959: after #1387 went in, this should now be moved somewhere around [here](https://github.com/devitocodes/devito/blob/master/devito/core/gpu_openmp.py#L266)

tests/test_gpu_openmp.py
```diff
@@ -22,17 +22,17 @@ def test_basic(self):
         assert len(trees) == 1
 
         assert trees[0][1].pragmas[0].value ==\
-            'omp target teams distribute parallel for collapse(3)'
-        assert op.body[1].header[0].value ==\
+            'omp target teams distribute parallel for device(devicenum) collapse(3)'
+        assert op.body[2].header[0].value ==\
```
FabioLuporini, line 26: why is this and the below ones changing?

FabioLuporini: oh, is it because of the new initialization bit?

if so, then we need a new test, at the very beginning, testing that initialization is producing the right code

devito/core/gpu_openmp.py
italoaug: It must be the name of the buffer rather than the name of the node:

(Pdb) node.name
'MPI_Irecv'
(Pdb) node._args['parameters'][0].name
'bufs'
[...]
(Pdb) node.name
'MPI_Isend'
(Pdb) node._args['parameters'][0].name
'bufg'

devito/core/gpu_openmp.py
italoaug: Agreed. I'll push that together with the next changes.

tests/test_gpu_openmp.py
italoaug:
```
(Pdb) print(op.body[1])
float (*restrict u)[u_vec->size[1]][u_vec->size[2]][u_vec->size[3]] __attribute__ ((aligned (64))) = (float (*)[u_vec->size[1]][u_vec->size[2]][u_vec->size[3]]) u_vec->data;
```

devito/core/gpu_openmp.py
FabioLuporini: oh yeah sorry. Then just `node.parameters[0].name` ?

tests/test_gpu_openmp.py
FabioLuporini: yeah, and `op.body[0]` is most likely the initialization code

devito/mpi/routines.py
italoaug: I'll push that together with the next changes.

devito/mpi/routines.py
italoaug: I'll push that together with the next changes.

devito/core/gpu_openmp.py
italoaug: `AttributeError: 'IrecvCall' object has no attribute 'parameters'`
Are you happy with `node.arguments[0].name` instead?

tests/test_gpu_openmp.py
italoaug: That's right:
```
(Pdb) print(op.body[0])
/* Begin of OpenMP5.0 setup */
int devicenum = 0;
/* End of OpenMP5.0 setup */
```

devito/core/gpu_openmp.py
FabioLuporini: yes, sorry, I meant arguments oc

devito/core/gpu_openmp.py
FabioLuporini: thanks!

tests/test_gpu_openmp.py
FabioLuporini: reminder:

> if so, then we need a new test, at the very beginning, testing that initialization is producing the right code

devito/core/gpu_openmp.py
italoaug: I'll push that together with the next changes.

tests/test_gpu_openmp.py
italoaug: Forgot to say it is already implemented locally, and I'll push that together with the next changes.

devito/operator/operator.py
italoaug: It seems like it is also necessary to add the gpu-direct somewhere else. By just moving it from `operator.py` to `gpu_openmp.py` leads to the following exception:
`devito.exceptions.InvalidOperator: Unsupported optimization options: [gpu-direct]`

FabioLuporini: a `git grep "Unsupported optimization options" shows that that exception is only raised in one point, here:

https://github.com/devitocodes/devito/blob/master/devito/core/gpu_openmp.py#L268

it happens because you have to `pop` `'gpu-direct'` from `oo`, so that devito can raise an exception if an illegal keyword has been provided. So `pop` is the way to go, just like done with all other options

FabioLuporini:
there also are conflicts to be resolved. Probably easy ones

devito/core/gpu_openmp.py
```diff
@@ -1,20 +1,27 @@
...
+from sympy import Function
 import numpy as np
 
 from devito.core.operator import OperatorCore
 from devito.data import FULL
 from devito.exceptions import InvalidOperator
-from devito.ir.iet import Callable, ElementalFunction, MapExprStmts
+from devito.ir.equations import DummyEq
+from devito.ir.iet import (Block, Call, Callable, ElementalFunction, FindNodes,
+                           LocalExpression, List, MapExprStmts, Transformer)
```
FabioLuporini, line 12: nitpicking: it wouldn't be bad to logically group the imported objects by "type" 

* Block, Call, Callable, ElementalFunction, and **List** are IET notes
* the others are Visitors

So I'd write them in this order:

```
Block, Call, Callable, ElementalFunction,  List, FindNodes, LocalExpression, MapExprStmts, Transformer
```

devito/core/gpu_openmp.py
```diff
@@ -263,6 +341,7 @@ def _normalize_kwargs(cls, **kwargs):
         o['par-chunk-nonaffine'] = oo.pop('par-chunk-nonaffine', cls.PAR_CHUNK_NONAFFINE)
         o['par-dynamic-work'] = np.inf  # Always use static scheduling
         o['par-nested'] = np.inf  # Never use nested parallelism
+        o['gpu-direct'] = oo.pop('gpu-direct')
```
FabioLuporini, line 344:
```
oo.pop('gpu-direct', False)
```

otherwise you get an error if the user doesn't provide the option explicitly...

... because (EDIT) you will have to remove the default from operator.py . See comment below

devito/operator/operator.py
FabioLuporini: this line should be removed as per my comment above

tests/test_gpu_openmp.py
```diff
@@ -1,15 +1,33 @@
...
+from devito.ir.iet import Block, FindNodes, retrieve_iteration_tree
+from devito.mpi.routines import IrecvCall, IsendCall
 from examples.seismic import TimeAxis, RickerSource, Receiver
 
 
 class TestCodeGeneration(object):
 
+    @pytest.mark.parallel(mode=1)
+    @switchconfig(platform='nvidiaX')
+    def test_gpu_direct(self):
```
FabioLuporini, line 17: this test should go at the end of the class not at the beginning, since it's a more advanced test...

tests/test_gpu_openmp.py
FabioLuporini: is this now pushed? can I resolve this comment?

italoaug: Yes. It is pushed.

devito/core/gpu_openmp.py
```diff
@@ -349,6 +431,11 @@ def _specialize_iet(cls, graph, **kwargs):
         data_manager.place_definitions(graph)
         data_manager.place_casts(graph)
 
+        # Initialize OpenMP environment
+        initialize(graph)
```
FabioLuporini, line 435: I think this should be the first pass of the pipeline, so it should be moved before the "Distributed-memory parallelism" block (pls don't forget the blank lines)

devito/core/gpu_openmp.py
```diff
@@ -349,6 +431,11 @@ def _specialize_iet(cls, graph, **kwargs):
...
+        if options['gpu-direct']:
+            mpi_gpu_direct(graph)
```
FabioLuporini, line 437: this should be right after `mpiize`

devito/core/gpu_openmp.py
```diff
@@ -417,4 +505,7 @@ def _specialize_iet(cls, graph, **kwargs):
         data_manager.place_definitions(graph)
         data_manager.place_casts(graph)
 
+        # Initialize OpenMP environment
+        initialize(graph)
```
FabioLuporini, line 509: same as before -- this should be at the top of the pipeline

devito/core/gpu_openmp.py
```diff
@@ -321,6 +400,9 @@ def _specialize_iet(cls, graph, **kwargs):
...
```
FabioLuporini, line 404: I think this should be the first pass of the pipeline, before the "Distributed-memory parallelism" block

examples/performance/01_gpu.ipynb
```diff
@@ -137,17 +137,13 @@
    "cell_type": "code",
```
FabioLuporini, line 137: why is this notebook changing all of a sudden?

tests/test_gpu_openmp.py
```diff
@@ -194,7 +209,29 @@ def test_timeparallel_reduction(self):
...
+        grid = Grid(shape=(3, 3, 3))
+
+        u = TimeFunction(name='u', grid=grid)
+
+        op = Operator(Eq(u.forward, u.dx+1), opt=('advanced', {'gpu-direct': True}))
+
+        assert str(op.body[0].body[0]) == 'int rank = 0;'
```
FabioLuporini, line 224: this should be a separate test -- test_init_omp_env or something along these lines

FabioLuporini: and it should be one of the first tests of the class

examples/performance/01_gpu.ipynb
italoaug: The generated code has changed for the current version including `device(devicenum)`

devito/core/gpu_openmp.py
italoaug: Doing that generates incorrect C code. Also, that is the same position `initialize` is in `gpu_openacc.py`.

devito/core/gpu_openmp.py
italoaug: Doing that generates incorrect C code. Also, that is the same position `initialize` is in `gpu_openacc.py`.

devito/core/gpu_openmp.py
italoaug: Doing that generates incorrect C code. Also, that is the same position `initialize` is in `gpu_openacc.py`.

devito/core/gpu_openmp.py
italoaug: Calling `mpi_gpu_direct` before "Symbol definitions" block would create `Blocks` before creating C variables. That would lead to `MPI_Request` variables being local to their blocks. This way, it would generate incorrect C code.

tests/test_gpu_openmp.py
```diff
@@ -263,3 +309,71 @@ def test_iso_ac(self):
         op(time=time_range.num-1, dt=dt)
 
         assert np.isclose(norm(rec), 490.55, atol=1e-2, rtol=0)
+
+    @pytest.mark.parallel(mode=1)
```
rhodrin, line 313: Why the `mode=1`?

.github/workflows/pytest-gpu.yml
```diff
@@ -59,7 +62,9 @@ jobs:
 
     - name: Test with pytest
       run: |
-        pgaccelinfo
+        if [ "${{ matrix.name }}" == 'pytest-gpu-acc' ]; then
```
georgebisbas, line 65: why not always printing it? It doesn't hurt...
Well, maybe we need to print it for openmp as well?
Why? It **may** reveal any cuda installation problems.
I would keep it.

georgebisbas: I mean I would not have it printed conditionally.

devito/core/gpu_openmp.py
```diff
@@ -203,6 +211,76 @@ def _(iet):
         return iet, {}
 
 
+@iet_pass
+def initialize(iet, **kwargs):
+    """
+    Initialize the OpenMP environment.
```
georgebisbas, line 217: OpenMP5.0 ? nitpicking...

.github/workflows/pytest-gpu.yml
rhodrin: That command doesn't work on machines without acc. Hence the conditional.

rhodrin:
Also, some of the examples should now also be tested with MPI.

devito/core/gpu_openmp.py
italoaug: I agree

tests/test_gpu_openmp.py
italoaug: As we discussed on Slack, it was a mistake. I'll change that for 2 and 4 MPI ranks.

.github/workflows/pytest-gpu.yml
```diff
@@ -34,12 +34,15 @@ jobs:
           arch: "clang"
           platform: "nvidiaX"
           language: "openmp"
+          # FIXME: Re-add and "gpu" and amend tags before merge.
```
FabioLuporini, line 37: need to address this?

devito/core/gpu_openacc.py
```diff
@@ -74,6 +74,11 @@ def _map_present(cls, f):
         return cls.lang['map-present'](f.name, ''.join('[0:%s]' % i
                                                        for i in cls._map_data(f)))
 
+    @classmethod
+    def _map_delete(cls, f):
```
FabioLuporini, line 78: note that `lang['map-release']` is mapped to `acc exit data delete` as well, just like the current `lang['map-exit-delete']`. This is my fault of course. I wonder whether we can clean up this mess...

devito/core/gpu_openmp.py
```diff
@@ -84,7 +93,10 @@ def _map_release(cls, f):
     @classmethod
     def _map_delete(cls, f):
         return cls.lang['map-exit-delete'](f.name, ''.join('[0:%s]' % i
-                                                           for i in cls._map_data(f)))
+                                                           for i in cls._map_data(f)),
```
FabioLuporini, line 96: this, written this way over 5 lines, is a bit chaotic. Can you rewrite it more cleanly?

devito/core/gpu_openmp.py
```diff
@@ -203,6 +215,76 @@ def _(iet):
         return iet, {}
 
 
+@iet_pass
+def initialize(iet, **kwargs):
+    """
+    Initialize the OpenMP 5.0 environment.
```
FabioLuporini, line 221: Drop 5.0

devito/core/gpu_openmp.py
```diff
@@ -203,6 +215,76 @@ def _(iet):
...
+            body = [rank_decl, rank_init, ngpus_init, devicenum_init]
+
+            init = List(header=c.Comment('Begin of OpenMP5.0+MPI setup'),
+                        body=body,
+                        footer=(c.Comment('End of OpenMP5.0+MPI setup'), c.Line()))
+        else:
+            devicenum_init = LocalExpression(DummyEq(devicenum, 0))
+            body = [devicenum_init]
+
+            init = List(header=c.Comment('Begin of OpenMP5.0 setup'),
```
FabioLuporini, line 254: Drop 5.0 (here and below)

devito/core/gpu_openmp.py
FabioLuporini: I might have a fix in a branch of mine to this issue. Can you leave a TODO comment? Along the lines

```
# TODO: This should be moved right below the `mpiize` pass, but currently ... <same explanation you wrote above>
if options[...
```

devito/core/gpu_openmp.py
```diff
@@ -369,7 +460,8 @@ def _make_passes_mapper(cls, **kwargs):
             'optcomms': partial(optimize_halospots),
             'openmp': partial(ompizer.make_parallel),
             'mpi': partial(mpiize, mode=options['mpi']),
-            'prodders': partial(hoist_prodders)
+            'prodders': partial(hoist_prodders),
+            'gpu-direct': partial(mpi_gpu_direct)
```
FabioLuporini, line 464: you also have to add `'gpu-direct' here: https://github.com/devitocodes/devito/pull/1363/files#diff-471f90b5151a59238f0735766abbda55R448

tests/test_gpu_openmp.py
```diff
@@ -194,7 +226,24 @@ def test_timeparallel_reduction(self):
...
+    @switchconfig(platform='nvidiaX')
+    def test_gpu_direct(self):
+        grid = Grid(shape=(3, 3, 3))
+
+        u = TimeFunction(name='u', grid=grid)
+
+        op = Operator(Eq(u.forward, u.dx+1), opt=('advanced', {'gpu-direct': True}))
+
+        for f in op._func_table.items():
```
FabioLuporini, line 241: a more pythonic way would be:

```
for f, v in op._func_table.items():
   for node in FindNodes(Block).visit(v.root):
      ...
```

tests/test_gpu_openmp.py
```diff
@@ -263,3 +312,71 @@ def test_iso_ac(self):
         op(time=time_range.num-1, dt=dt)
 
         assert np.isclose(norm(rec), 490.55, atol=1e-2, rtol=0)
+
+    @pytest.mark.parallel(mode=[2, 4])
+    @skipif('nodevice')
+    def test_op_apply_gpu_direct(self):
```
FabioLuporini, line 318: just "test_gpu_direct" ?

tests/test_gpu_openmp.py
```diff
@@ -263,3 +312,71 @@ def test_iso_ac(self):
...
+        assert 'omp target' in str(op)
+
+        time_steps = 1000
+        op.apply(time_M=time_steps)
+
+        assert np.all(np.array(u.data[0, :, :, :]) == time_steps)
+
+    @pytest.mark.parallel(mode=[2, 4])
+    @skipif('nodevice')
+    def test_iso_ac_gpu_direct(self):
```
FabioLuporini, line 335: this is 99% identical to `test_iso_ac` right? never ever duplicate code except for extremely rare situations (amount of copy is tiny, eg 1-2 lines, and paste only happens once... and even then it's debatable)

so I would do something along the lines of (just to give u the idea)

```
def iso_acoustic(self, **opt_options):
    ...
    ... Operator(..., opt=('advanced', opt_options))
    ...

@skipif('nodevice')
def test_iso_acoustic(self):
    iso_acoustic()


@pytest.mark.parallel(mode=[2, 4])
@pytest.mark.parametrize('gpu_direct', [False, True])
@skipif('nodevice')
def test_mpi_iso_acoustic(self):
    opt_options = {'gpu-direct': gpu_direct}
    iso_acoustic(**opt_options)
```

FabioLuporini (changes requested):

.github/workflows/pytest-gpu.yml
rhodrin: It won't cause problems but I'll need to mess around with a couple of runners after the merge. This comment can now be removed.

.github/workflows/pytest-gpu.yml
```diff
@@ -59,7 +63,9 @@ jobs:
 
     - name: Test with pytest
       run: |
-        pgaccelinfo
```
rhodrin, line 62: @italoaug Can you confirm that this command still does not work on the openmp runner?

italoaug: Not sure I got it. Should you be asking about `gpu-runner-04`, there is no PGI compiler installed in it.

devito/core/gpu_openacc.py
italoaug: TBH, I'm not familiar with OpenACC yet. So I'd be happy to address that issue in a new PR with a small patch.

devito/core/gpu_openmp.py
italoaug: I managed to reduce it to 4 lines.

devito/core/gpu_openmp.py
italoaug: ok

devito/core/gpu_openmp.py
italoaug: ok

devito/core/gpu_openmp.py
italoaug: All right

devito/core/gpu_openmp.py
italoaug: I'm not sure I understood it. Do you mean adding
```

def _mpi_gpu_direct(self, *args):
        return
```
after line 454 of `devito/mpi/routines.py`?

tests/test_gpu_openmp.py
italoaug: ok

tests/test_gpu_openmp.py
italoaug: ok

tests/test_gpu_openmp.py
italoaug: I had to change `iso_acoustic()` by `TestOperator().iso_acoustic()` otherwise I get an unkown variable error
But still I'm getting this error:
`undefined name 'gpu_direct'` for the line `opt_options = {'gpu-direct': gpu_direct}` 
Also `TestOperator.test_iso_acoustic` fails due to an `AssertionError` in `assert np.isclose(norm(rec), 490.55, atol=1e-2, rtol=0)`

devito/core/gpu_openmp.py
FabioLuporini: sorry, I mean, you have to add it here: https://github.com/devitocodes/devito/blob/master/devito/core/gpu_openmp.py#L360, next to `'prodders'` so that the compiler recognizes it

tests/test_gpu_openmp.py
FabioLuporini: yeah... the variable is obviously undefined. Clearly, mine was just pseudocode. Once you parametrize a test, you also have to supply the variable name to the test parameters:

```
@pytest.mark.parametrize('gpu_direct', [False, True])
test_mpi_iso_acoustic(self, gpu_direct):
    ....

italoaug: ok

devito/core/gpu_openmp.py
italoaug: NP

mloubout (approved):

FabioLuporini (approved):

.github/workflows/pytest-gpu.yml
rhodrin: Yes, that's what I was asking. Thanks for clarifying.

rhodrin (approved):

FabioLuporini (approved):
Really great job, thanks! Merging now

Issue #1185, opened 2020-03-24
Devito CFD tutorials


jkwashbourne:
### 01_convection.ipynb
 - [ ] Suggest noting that u[:,-1] notation is shorthand for the maximum
 - [ ] In cell [4] last sentence, it was not clear to me that the 3d array u[t,x,y] could be indexed as u[0] to provide u[0,x,y]. Is this a numpy shorthand?
 - [ ] Cell [6]: wondering why the "1.0*" doesn't disappear?
 - [ ] Cell [6]: you refer to "s" and "h" but they are not seen here. Did the output from print(eq) change? We do see "h_x" and "h_y" in cell [7].
 - [ ] Cell [7]: discussion uses "h" but stencil uses "h_x" and "h_y". Note there is continuing confusion about "s", "dt".
 - [ ] I needed "%matplotlib inline" here also, must be my setup is whack
 - [ ] Cell [9]: I was expecting a smaller tolerance after implementing the BCs, what gives? Should the Devito result with no optimization be float eps equivalent to the numpy one? 

### 01_convection_revisited.ipynb
 - [ ] s/exampe/example/
 - [ ] Cell [9] uses "nx - 1" instead of "-1" notation, consistency?
 - [ ] Again schizo on "-1" and "n-1" notation for end of array

### 02_convection_nonlinear.ipynb
 - [ ] Cell [1] here has "%matplotlib inline", suggest adding to the others in the first cell
 - [ ] "s/hte/the/"
 - [ ] "s/woth/with/"
 - [ ] Cell [8]: A couple more words on the "automated staggering of vectorial function" would be super useful

### 03_diffusion.ipynb
 - [ ] Cell [3]: omg such a python n00b, I had to go look up the "//" operator. Is it worth noting what that operator does? Maybe not since you have the plot just below.
 - [ ] Cell [4] is a partial duplicate of cell [3]?
 - [ ] Cell [7], [8] there is inconsistency in line feeds ... [not sure anyone should care about this level of pedantry]
 - [ ] Cell [7] and [8] seem redundant

### 03_diffusion_nonuniform.ipynb
 - [ ] Suggest clarification: "This example is an enhancement of 03_diffusion in terms of having non-uniform viscosity as opposed to 03_diffusion" --> "This example is an enhancement of 03_diffusion in terms of having non-uniform viscosity"
[...]

PR #1364, opened 2020-06-23
Multi-GPU + PGI dockerfile

kenhester:
A dockerfile and supporting plugins for Multi-GPU support in Devito container.

### DIFF
--- a/devito/archinfo.py
+++ b/devito/archinfo.py
@@ -1,13 +1,15 @@
 """Collection of utilities to detect properties of the underlying architecture."""
 
-from subprocess import PIPE, Popen
+from subprocess import PIPE, Popen, DEVNULL
 
-import numpy as np
+from cached_property import cached_property
 import cpuinfo
+import numpy as np
 import psutil
+import re
 
 from devito.logger import warning
-from devito.tools.memoization import memoized_func
+from devito.tools import all_equal, memoized_func
 
 __all__ = ['platform_registry',
            'INTEL64', 'SNB', 'IVB', 'HSW', 'BDW', 'SKX', 'KNL', 'KNL7210',
@@ -115,6 +117,135 @@ def get_cpu_info():
     cpu_info['physical'] = physical
 
     return cpu_info
+
+
+@memoized_func
+def get_gpu_info():
+    """Attempt GPU info autodetection."""
+
+    # Filter out virtual GPUs from a list of GPU dictionaries
+    def filter_real_gpus(gpus):
+        def is_real_gpu(gpu):
+            return 'virtual' not in gpu['product'].lower()
+        return filter(is_real_gpu, gpus)
+
+    # The following functions of the form cmd_gpu_info(...) attempt obtaining GPU
+    #   information using 'cmd'
+    # The currently supported ways of obtaining GPU information (in order of attempt) are:
+    #   - 'lshw' from the command line
+    #   - 'lspci' from the command line
+
+    def lshw_gpu_info(text):
+        def lshw_single_gpu_info(text):
+            # Separate the output into lines for processing
+            lines = text.replace('\\n', '\n')
+            lines = lines.splitlines()
+
+            # Define the processing functions
+            if lines:
+                def extract_gpu_info(keyword):
+                    for line in lines:
+                        if line.lstrip().startswith(keyword):
+                            return line.split(':')[1].lstrip()
+
+                def parse_product_arch():
+                    for line in lines:
+                        if line.lstrip().startswith('product') and '[' in line:
+                            arch_match = re.search(r'\[([\w\s]+)\]', line)
+                            if arch_match:
+                                return arch_match.group(1)
+                    return 'unspecified'
+
+                # Populate the information
+                gpu_info = {}
+                gpu_info['product'] = extract_gpu_info('product')
+                gpu_info['architecture'] = parse_product_arch()
+                gpu_info['vendor'] = extract_gpu_info('vendor')
+                gpu_info['physicalid'] = extract_gpu_info('physical id')
+
+                return gpu_info
+
+        # Parse the information for all the devices listed with lshw
+        devices = text.split('display')[1:]
+        gpu_infos = [lshw_single_gpu_info(device) for device in devices]
+        return filter_real_gpus(gpu_infos)
+
+    def lspci_gpu_info(text):
+        # Note: due to the single line descriptive format of lspci, 'vendor'
+        #   and 'physicalid' elements cannot be reliably extracted so are left None
+
+        # Separate the output into lines for processing
+        lines = text.replace('\\n', '\n')
+        lines = lines.splitlines()
+
+        gpu_infos = []
+        for line in lines:
+            # Graphics cards are listed as VGA or 3D controllers in lspci
+            if 'VGA' in line or '3D' in line:
+                gpu_info = {}
+                # Lines produced by lspci command are of the form:
+                #   xxxx:xx:xx.x Device Type: Name
+                #   eg:
+                #   0001:00:00.0 3D controller: NVIDIA Corp... [Tesla K80] (rev a1)
+                name_match = re.match(
+                    r'\d\d\d\d:\d\d:\d\d\.\d [\w\s]+: ([\w\s\(\)\[\]]*)', line
+                )
+                if name_match:
+                    gpu_info['product'] = name_match.group(1)
+                    arch_match = re.search(r'\[([\w\s]+)\]', line)
+                    if arch_match:
+                        gpu_info['architecture'] = arch_match.group(1)
+                    else:
+                        gpu_info['architecture'] = 'unspecified'
+                else:
+                    continue
+
+                gpu_infos.append(gpu_info)
+        return filter_real_gpus(gpu_infos)
+
+    # Run homogeneity checks on a list of GPU, return GPU with count if homogeneous,
+    #   otherwise None
+    def homogenise_gpus(gpus):
+        if gpu_infos == []:
+            warning('No graphics cards detected')
+            return None
+
+        if all_equal(gpu_infos):
+            gpu_infos[0]['ncards'] = len(gpu_infos)
+            return gpu_infos[0]
+
+        warning('Different models of graphics cards detected')
+
+        return None
+
+    # Obtain textual gpu info and delegate parsing to helper functions
+
+    try:
+        # First try is with detailed command lshw
+        info_cmd = ['lshw', '-C', 'video']
+        proc = Popen(info_cmd, stdout=PIPE, stderr=DEVNULL)
+        raw_info = str(proc.stdout.read())
+
+        gpu_infos = lshw_gpu_info(raw_info)
+        return homogenise_gpus(gpu_infos)
+
+    except OSError:
+        pass
+
+    try:
+        # Second try is with lspci, which is more readable and less detailed than lshw
+        info_cmd = ['lspci']
+        proc = Popen(info_cmd, stdout=PIPE, stderr=DEVNULL)
+        raw_info = str(proc.stdout.read())
+
+        # Parse the information for all the devices listed with lspci
+        gpu_infos = lspci_gpu_info(raw_info)
+        return homogenise_gpus(gpu_infos)
+
+    except OSError:
+        pass
+
+    return None
 
 
 @memoized_func
@@ -278,8 +409,6 @@ class Device(Platform):
         self.cores_physical = cores_physical
         self.isa = isa
 
-        self.march = self._detect_march()
-
     @classmethod
     def _mro(cls):
         # Retain only the Device Platforms
@@ -291,19 +420,27 @@ class Device(Platform):
                 break
         return retval
 
-    @classmethod
-    def _detect_march(cls):
+    @cached_property
+    def march(self):
         return None
 
 
 class NvidiaDevice(Device):
-    pass
+
+    @cached_property
+    def march(self):
+        info = get_gpu_info()
+        if info:
+            architecture = info['architecture']
+            if 'tesla' in architecture.lower():
+                return 'tesla'
+        return None
 
 
 class AmdDevice(Device):
 
-    @classmethod
-    def _detect_march(cls):
+    @cached_property
+    def march(cls):
         # TODO: this corresponds to Vega, which acts as the fallback `march`
         # in case we don't manage to detect the actual `march`. Can we improve this?
         fallback = 'gfx900'
--- a/tests/test_gpu_common.py
+++ b/tests/test_gpu_common.py
@@ -1,0 +1,11 @@
+from devito.archinfo import get_gpu_info
+
+from conftest import skipif
+
+
+class TestGPUInfo(object):
+
+    @skipif('nodevice')
+    def test_get_gpu_info(self):
+        info = get_gpu_info()
+        assert 'tesla' in info['architecture'].lower()


devito/compiler.py
```diff
@@ -448,8 +448,7 @@ def __init__(self, *args, **kwargs):
         self.cflags.remove('-O3')
         self.cflags.remove('-Wall')
         self.cflags += ['-fast', '-acc']
-        # self.cflags += ['-Minfo=accel'] # Print informational messages for compliation
-        # self.cflags += ['-ta=tesla']    # Compile for a target GPU architecture
+        self.cflags += ['-ta=tesla,host']    # Compile for a target GPU architecture
```
mloubout, line 451: Isn't it purely tesla specific? If yes maybe should stay commented.

requirements-nvidia.txt
```diff
@@ -0,0 +1,5 @@
+cupy>=6
+dask-cuda
+jupyterlab==1.2.1
```
mloubout, line 3: The first two yes, these three I don't think are gpu/nvidia specific. May want to add it into docker requriements

mloubout:
Thanks for this addition, really nice. Few minor changes and should be GTG

FabioLuporini:
this should fix https://github.com/devitocodes/devito/issues/1185 (at least the openacc side of it)

examples/seismic/tutorials/04_dask.ipynb
```diff
@@ -25,6 +25,31 @@
     "The first part of this tutorial closely follows [tutorial 3](https://github.com/devitocodes/devito/blob/master/examples/seismic/tutorials/03_fwi.ipynb) and consists of reading the velocity model and setting up the acquisition geometry. Subsequently, we will implement serial versions of forward modeling and FWI objective functions and then show how we can use Dask to implement parallel versions of these functions. Finally, we will show how to write a wrapper that lets us pass our objective function to scipy's optimize package and how to run a small 2D FWI example using a limited-memory Quasi-Newton method."
```
navjotk, line 25: <p>Nitpicking: Capitalization</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/devitocodes/devito/pull/1364/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

devito/compiler.py
kenhester: The targets above are instructing the compiler for gpu code and single threaded cpu code.  This is also the default for the PGI openacc compiler.

requirements-nvidia.txt
kenhester: dask-cuda is from the RAPIDS.ai project and enables multi-gpu dask tasks.  
cupy is a requirement for dask-cuda.
jupyter-lab requires some changes in the start script.  This is not gpu specific.

docker/Dockerfile.nvidia
```diff
@@ -0,0 +1,91 @@
...
+
+ENV CC="/opt/pgi/linux86-64-llvm/19.10/bin/pgcc"
+ENV CXX="/opt/pgi/linux86-64-llvm/19.10/bin/pgc++"
+RUN mv '/opt/pgi/linux86-64/19.10/mpi/openmpi-3.1.3/bin/mpicc.devito' '/opt/pgi/linux86-64/19.10/mpi/openmpi-3.1.3/bin/mpicc' 
+
+WORKDIR /app
+EXPOSE 8888
+ENTRYPOINT ["/docker-entrypoint.sh"]
+CMD ["/jupyter"]
+#CMD ["/bin/bash"]
```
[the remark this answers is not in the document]
kenhester: Yes, this was used during debug of the dockerfile build.

docker/Dockerfile.nvidia
```diff
@@ -0,0 +1,91 @@
...
+#Options: [unset, PERF, DEBUG]
+ENV DEVITO_LOGGING=DEBUG
+#Options: [unset, nvidiaX] Set for openacc
+ENV DEVITO_PLATFORM=nvidiaX
+#Options: [unset, 1] Only for PGI; Debug Timing 
+ENV PGI_ACC_TIME=1 
+
+ENV CC="/opt/pgi/linux86-64-llvm/19.10/bin/pgcc"
+ENV CXX="/opt/pgi/linux86-64-llvm/19.10/bin/pgc++"
+RUN mv '/opt/pgi/linux86-64/19.10/mpi/openmpi-3.1.3/bin/mpicc.devito' '/opt/pgi/linux86-64/19.10/mpi/openmpi-3.1.3/bin/mpicc'
```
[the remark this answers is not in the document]
kenhester: The NGC PGI container has the C, C++, and MPI set to compile using the PGI compilers and libraries.  The libraries and tools needed during the install of Devito require the gcc libraries, compilers, and tools.  The dockerfile set the environment for gcc at the top of the dockerfile and resets the environment to PGI at the end of the dockerfile.

docker/run-jupyterlab.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+PYTHONPATH=/app /venv/bin/jupyter-lab --ip=0.0.0.0 --port=8888 --allow-root --NotebookApp.token=''
```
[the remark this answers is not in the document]
kenhester: This does remove the security token, and simplifies getting into the notebook after the dockerfile is launched.  This is common for nvidia jupyter based containers.

docker/Dockerfile.nvidia
```diff
@@ -0,0 +1,91 @@
...
+# Reference: https://github.com/devitocodes/devito/wiki/FAQ#can-i-manually-modify-the-c-code-generated-by-devito-and-test-these-modifications
+# 
+#Options: [gcc, pgcc]
+ENV DEVITO_ARCH="pgcc" 
+#Options: [openmp, openacc]
+ENV DEVITO_LANGUAGE="openacc"
+#Options: [unset, 1] For PGI openacc; Should only be set after a first execution of the benchmark
+#ENV DEVITO_JIT_BACKDOOR=1 
+#Options: [unset, 1] Unset for openacc
+#ENV DEVITO_OPENMP=1
```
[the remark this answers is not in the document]
kenhester: I did not find these documented elsewhere, so I documented them while I was developing the dockerfile.

devito/compiler.py
FabioLuporini: yeah... but why 'tesla' specifically? that looks a bit odd

how could we sniff the underlying gpu architecture?

docker/Dockerfile.nvidia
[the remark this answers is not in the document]
FabioLuporini: The environment variables are documented here: https://github.com/devitocodes/devito/wiki/FAQ#what-environment-variables-control-how-devito-works

DEVITO_OPENMP is obsolete, you can safely drop it

docker/Dockerfile.nvidia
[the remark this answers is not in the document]
FabioLuporini: could you add an explanatory comment? I was baffled for a moment as well

docker/Dockerfile.nvidia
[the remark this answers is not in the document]
FabioLuporini: reminder

requirements-nvidia.txt
FabioLuporini: > cupy is a requirement for dask-cuda

so it can probably be dropped here? I'd be surprised if it were not listed in `dask-cuda`'s requirements already

> This is not gpu specific.

Strictly speaking, non of these are devito-GPU specific. These are the requirements to run DASK on a (multi-)GPU node/cluster. And DASK is used in one of the layers of abstraction on top of Devito.  So I think this file should be moved to a different location, or at least be renamed

and @mloubout , instead of three `requirements-*` files in the root dir, we should perhaps start thinking about a reorganization of the directory tree

devito/compiler.py
kenhester: I do not have code that checks for another architecture, so the user would need to change.  Maybe you could use lspci or lsmod for linux.

FabioLuporini: doesn't the PGI compiler have an option to return info about the underlying GPU architecture? for example, gcc has one, but only for CPUs

> Maybe you could use lspci or lsmod for linux.

so we either enhance `get_platform`, see [here](https://github.com/devitocodes/devito/blob/master/devito/archinfo.py#L141), or we'll have to take this line back:

```
self.cflags += ['-ta=tesla,host']
```

(commented out would be OK)

I'll see if I can get one of our summer students to work on this

docker/Dockerfile.nvidia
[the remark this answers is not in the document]
kenhester: cleanup complete

docker/Dockerfile.nvidia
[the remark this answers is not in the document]
kenhester: removed.  cleanup complete

docker/Dockerfile.nvidia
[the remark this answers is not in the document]
kenhester: added these comments to the dockerfile for explanation.  complete

devito/compiler.py
kenhester: comment and commented out as this is the default target for PGI.  The line of code is left in  place commented out as others may want to adjust the target in the future.

devito/compiler.py
```diff
@@ -448,8 +448,7 @@ def __init__(self, *args, **kwargs):
         self.cflags.remove('-O3')
         self.cflags.remove('-Wall')
         self.cflags += ['-fast', '-acc']
-        # self.cflags += ['-Minfo=accel'] # Print informational messages for compliation
-        # self.cflags += ['-ta=tesla']    # Compile for a target GPU architecture
+        #self.cflags += ['-ta=tesla,host']    # Default PGI compile for a target is GPU and single threaded host.
```
FabioLuporini, line 451: @kenhester thanks to #1383 (from our summer student @jack-lascala) we'll soon be able to auto-detect whether it's an actual tesla card what we got underneath !

examples/seismic/tutorials/04_dask.ipynb
```diff
@@ -43,24 +68,13 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
```
mloubout, line 71: This notebook needs to be rerun top to bottom on a cpu.

docker/Dockerfile.nvidia
```diff
@@ -0,0 +1,92 @@
...
+
+RUN chmod +x /print-defaults /jupyter /tests /docker-entrypoint.sh && \
+    /venv/bin/jupyter labextension install jupyterlab-nvdashboard && \
+    /venv/bin/jupyter labextension install dask-labextension && \
+    /venv/bin/jupyter serverextension enable dask_labextension && \
+    /venv/bin/jupyter lab workspaces import /app/nvdashboard.json
+
+# Environment Variables for OpenACC Builds
+# Reference: https://github.com/devitocodes/devito/wiki/FAQ#can-i-manually-modify-the-c-code-generated-by-devito-and-test-these-modifications
+# Options: [gcc, pgcc]
```
mloubout, line 69: If these are option, it may be good to move it as CLI args, i.e.
```
ARG DEVITO_ARCH="openacc"
ENV DEVITO_LANGUAGE=$DEVITO_ARCH
```

So that can be used as `docker rebuild ...... --build-arg DEVITO_ARCH'"gcc"` easily without having to dig into the dockerfile.

docker/Dockerfile.nvidia
```diff
@@ -0,0 +1,92 @@
+##############################################################
+# README with base system requirements
+# https://ngc.nvidia.com/catalog/containers/hpc:pgi-compilers
+#
+# BUILD: docker build --file docker/Dockerfile.pgi --tag devito:nvidia.pgi .
```
mloubout, line 5: I would add a READMe to the docker directory for that so that it is easily accessible to users.

docker/Dockerfile.nvidia
```diff
@@ -0,0 +1,92 @@
...
+# Options: [unset, PERF, DEBUG]
+ENV DEVITO_LOGGING=DEBUG
+# Options: [unset, nvidiaX] Set for openacc
+ENV DEVITO_PLATFORM=nvidiaX
+# Options: [unset, 1] Only for PGI; Debug Timing 
+ENV PGI_ACC_TIME=1 
+
+# Resetting the environment to PGI after finishing the install of the requirements for Devito and tools.
+ENV CC="/opt/pgi/linux86-64-llvm/19.10/bin/pgcc"
+ENV CXX="/opt/pgi/linux86-64-llvm/19.10/bin/pgc++"
```
mloubout, line 86: So this enforce pgcc even if `DEVITO_ARCH="gcc"` above?

docker/run-jupyterlab.sh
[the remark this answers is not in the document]
navjotk: Read up a bit. Looks like common practice indeed.

docker/Dockerfile.nvidia
kenhester: The base container is the  NGC PGI container.  I need to reset the env back to the original state to support PGI compiles.  Perhaps you are relating gcc to CPU and pgcc to GPU.  pgcc can compile cpu and gpu codes.

docker/Dockerfile.nvidia
kenhester: The goal of the dockerfile was to compile codes using openacc.  The base container is from NGC PGI is an openacc container.  I would have to leave the env args in an intermediate state to support compiling with gcc.  pgcc can compile for both CPU and GPU, so is there a need to switch between gcc and pgcc via CLI args?  Perhaps we leave the other dockerfile in place to support a gcc only env.

mloubout: I understand that,  and I am not asking to make a generic dockerfile. What I am saying is that this says `# Options: [gcc, pgcc]`. If it is actually optional, then it should be a CLI argument to make the switch easier and same for any ENV var that has multiple options such as LOGGING/MPI/PLATFORM/... below. If it is not optional and should be `pgcc` then it should indeed be `ENV DEVITO_ARCH="pgcc" ` but the comment should be changed to something like `# Set arch to pgcc`

examples/seismic/tutorials/04_dask.ipynb
kenhester: Results cleared and CUDA Dask Flag set to false to pass CI validation

docker/Dockerfile.nvidia
kenhester: dockerfile comments touched up for these environment variables

docker/Dockerfile.nvidia
FabioLuporini: @mloubout above I read `DEVITO_ARCH="pgcc"` that seems fine to me

examples/seismic/tutorials/02_rtm.ipynb
```diff
@@ -2,9 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "metadata": {
-    "collapsed": true
-   },
+   "metadata": {},
```
FabioLuporini, line 5: There should be absolutely 0 changes to this notebook. This needs to be amended, or it will affect our CI

docker/Dockerfile.nvidia
mloubout: Yes he changed it, looking good now.

requirements-nvidia.txt
mloubout: Yes we may need to reorganize a bit our requirements but for another PR I'd say

FabioLuporini:
@kenhester -- @mloubout @navjotk and I agreed that this is basically good to go. The only remaining thing is the partially execute RTM notebook, which needs to be undone, otherwise it will mess up with CI coverage. Other than that, all good !

examples/seismic/tutorials/02_rtm.ipynb
kenhester: I just loaded the latest 02_rtm.ipynb and all the results are cleared.  Before I checked in this last notebook, the notebook had results from a previous GPU run.  Do you want me to check in the notebook with results from a CPU run?

mloubout: I would say just checkout the notebook from master since there shouldn't be any change for it and push and should be gtg (basically this notebook shouldn't show up in the diff).

kenhester: The notebook had a previous run in the results in master.  It seems the results should be removed.  Agree ?  If so, that is the file differences the software is recording.

FabioLuporini: yeah, just remove all edits to 02_rtm, there should be none in this PR

FabioLuporini:
superseded by #1392

PR #1365, opened 2020-06-24
tests: Fix args and test viscoacoustic

georgebisbas:
Viscoacoustic examples is currently failing by default as its kernels are not in the usual ones. Adding explicit tests for them and restoring its working status.

### DIFF
--- a/examples/seismic/utils.py
+++ b/examples/seismic/utils.py
@@ -192,7 +192,8 @@ def seismic_args(description):
     parser.add_argument("--nbl", default=40,
                         type=int, help="Number of boundary layers around the domain")
     parser.add_argument("-k", dest="kernel", default='OT2',
-                        choices=['OT2', 'OT4', 'centered', 'staggered'],
+                        choices=['OT2', 'OT4', 'centered', 'staggered',
+                                 'blanch_symes', 'ren', 'deng_mcmechan'],
                         help="Choice of finite-difference kernel")
     parser.add_argument("--constant", default=False, action='store_true',
                         help="Constant velocity model, default is a two layer model")
--- a/examples/seismic/utils.py
+++ b/examples/seismic/utils.py
@@ -192,8 +192,7 @@ def seismic_args(description):
     parser.add_argument("--nbl", default=40,
                         type=int, help="Number of boundary layers around the domain")
     parser.add_argument("-k", dest="kernel", default='OT2',
-                        choices=['OT2', 'OT4', 'centered', 'staggered',
-                                 'blanch_symes', 'ren', 'deng_mcmechan'],
+                        choices=['OT2', 'OT4', 'centered', 'staggered'],
                         help="Choice of finite-difference kernel")
     parser.add_argument("--constant", default=False, action='store_true',
                         help="Constant velocity model, default is a two layer model")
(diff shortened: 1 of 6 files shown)


examples/seismic/utils.py
```diff
@@ -192,7 +192,8 @@ def seismic_args(description):
     parser.add_argument("--nbl", default=40,
                         type=int, help="Number of boundary layers around the domain")
     parser.add_argument("-k", dest="kernel", default='OT2',
-                        choices=['OT2', 'OT4', 'centered', 'staggered'],
+                        choices=['OT2', 'OT4', 'centered', 'staggered',
+                                 'blanch_symes', 'ren', 'deng_mcmechan'],
```
FabioLuporini, line 196: these ones only apply to viscoacoustic

examples/seismic/viscoacoustic/wavesolver.py
```diff
@@ -32,6 +32,9 @@ def __init__(self, model, geometry, space_order=4, kernel='blanch_symes', **kwar
 
         self.space_order = space_order
         self.dt = self.model.critical_dt
+        if kernel not in {'blanch_symes', 'ren', 'deng_mcmechan'}:
```
FabioLuporini, line 35: no...this should be made clean by having viscoacoustic use its own arg for kernel

examples/seismic/utils.py
georgebisbas: So now, after migrating options to seismic_args we need to find a more 'global' way to do that, how should that be handled?

examples/seismic/viscoacoustic/wavesolver.py
georgebisbas: Out of seismic_args I guess?

examples/seismic/utils.py
mloubout: Yes, same way `--fs` or `--noazimut` is done in acoustic and tti, example specific args in each example with according checks.

examples/seismic/viscoacoustic/viscoacoustic_example.py
```diff
@@ -41,6 +43,14 @@ def test_viscoacoustic():
     assert np.isclose(norm(rec), 18.7749, atol=1e-3, rtol=0)
 
 
+@pytest.mark.parametrize('ndim', [1, 2, 3])
+def test_viscoacoustic_stability(ndim):
```
mloubout, line 47: add kernel as parameter to test all three

examples/seismic/utils.py
georgebisbas: But kernel options needs to be in a dict right? It is not an extra arg.

examples/seismic/viscoacoustic/viscoacoustic_example.py
FabioLuporini: is this the only thing missing here? @mloubout @georgebisbas

mloubout: I fixed the other ones so if wanted yes, can always be added later if wanted and just merge

FabioLuporini: still missing?

georgebisbas: Yes it is still missing.

FabioLuporini:
Merged.

PR #1366, opened 2020-06-25
Prevent sympy cache blowup

mloubout:
Small change to AbstractFunction caching that prevents creating duplicates of function by caching on indices as well.

### DIFF
--- a/devito/types/basic.py
+++ b/devito/types/basic.py
@@ -334,28 +334,8 @@ class Symbol(AbstractSymbol, Cached):
     actual objects.
     """
 
-    @classmethod
-    def _cache_key(cls, *args, **kwargs):
-        args = list(args)
-        key = {}
-
-        # The base type is necessary, otherwise two objects such as
-        # `Scalar(name='s')` and `Dimension(name='s')` would have the same key
-        key['cls'] = cls
-
-        # The name is always present, and added as if it were an arg
-        key['name'] = kwargs.pop('name', None) or args.pop(0)
-
-        # From the args
-        key['args'] = tuple(args)
-
-        # From the kwargs
-        key.update(kwargs)
-
-        return frozendict(key)
-
     def __new__(cls, *args, **kwargs):
-        key = cls._cache_key(*args, **kwargs)
+        key = cls._make_key(*args, **kwargs)
         obj = cls._cache_get(key)
 
         if obj is not None:
@@ -377,6 +357,26 @@ class Symbol(AbstractSymbol, Cached):
 
         return newobj
 
+    @classmethod
+    def _make_key(cls, *args, **kwargs):
+        args = list(args)
+        key = {}
+
+        # The base type is necessary, otherwise two objects such as
+        # `Scalar(name='s')` and `Dimension(name='s')` would have the same key
+        key['cls'] = cls
+
+        # The name is always present, and added as if it were an arg
+        key['name'] = kwargs.pop('name', None) or args.pop(0)
+
+        # From the args
+        key['args'] = tuple(args)
+
+        # From the kwargs
+        key.update(kwargs)
+
+        return frozendict(key)
+
     __hash__ = Cached.__hash__
 
 
@@ -386,14 +386,8 @@ class DataSymbol(AbstractSymbol, Cached):
     A scalar symbol, cached by both Devito and SymPy, which carries data.
     """
 
-    @classmethod
-    def _cache_key(cls, *args, **kwargs):
-        """A DataSymbol caches on the class type itself."""
-        return cls
-
     def __new__(cls, *args, **kwargs):
-        key = cls._cache_key(*args, **kwargs)
-        obj = cls._cache_get(key)
+        obj = cls._cache_get(cls)
 
         if obj is not None:
             return obj
@@ -487,20 +481,14 @@ class AbstractTensor(sympy.ImmutableDenseMatrix, Basic, Cached, Pickable, Evalua
     is_TensorValued = True
     is_VectorValued = False
 
-    @classmethod
-    def _cache_key(cls, *args, **kwargs):
-        """An AbstractTensor caches on the class type itself."""
-        return cls
-
     def __new__(cls, *args, **kwargs):
         options = kwargs.get('options', {})
 
-        key = cls._cache_key(*args, **kwargs)
-        obj = cls._cache_get(key)
+        obj = cls._cache_get(cls)
 
         if obj is not None:
             newobj = sympy.Matrix.__new__(cls, *args, **options)
-            newobj.__init_cached__(key)
+            newobj.__init_cached__(cls)
             return newobj
 
         name = kwargs.get('name')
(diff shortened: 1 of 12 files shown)


examples/seismic/tti/wavesolver.py
```diff
@@ -170,7 +170,8 @@ def adjoint(self, rec, srca=None, p=None, r=None, vp=None,
         -------
         Adjoint source, wavefield and performance summary.
         """
-
+        if kernel != 'centered':
+            error('Only centered kernel is supported for the adjoint')
```
FabioLuporini, line 174: raise RuntimeError(...)

examples/seismic/tti/operators.py
```diff
@@ -73,7 +73,6 @@ def Gzz_centered(model, field, costheta, sintheta, cosphi, sinphi, space_order):
     Rotated second order derivative w.r.t. z.
     """
     order1 = space_order // 2
-    x, y, z = model.space_dimensions
```
FabioLuporini, line 76: no complaints from flake8? 😮

mloubout: Yeah I know right? WTH

devito/types/caching.py
```diff
@@ -69,7 +69,7 @@ def __init__(self, key):
             The cache key. It must be hashable.
         """
         # Precompute hash. This uniquely depends on the cache key
-        self._cache_key_hash = hash(key)
+        self._cache_key_hash = hash(alias or key)
```
mloubout, line 72: This one is a bit odd but basically we need to make sure that all `u[t+1], u[t+2], ...` have the same cache_key_hash or in some odd case `is` will not work (== will still work but for example replace for custom coefficients breaks without this).

examples/seismic/tti/operators.py
mloubout: I mean should have complained before this is not used anywher

mloubout: Yeah it is why it's weird.

tests/test_dse.py
```diff
@@ -1411,7 +1411,7 @@ def test_fullopt_w_mpi(self):
 
     @switchconfig(profiling='advanced')
     @pytest.mark.parametrize('space_order,expected', [
-        (8, 173), (16, 309)
+        (8, 173), (16, 307)
```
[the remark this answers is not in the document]
mloubout: I'm gonna start putting both keeps leap froging

devito/types/basic.py
```diff
@@ -335,7 +335,7 @@ class Symbol(AbstractSymbol, Cached):
     """
 
     @classmethod
-    def _cache_key(cls, *args, **kwargs):
+    def _make_key(cls, *args, **kwargs):
```
mloubout, line 338: Note: renamed it but only classmethod here not part of caching.

tests/test_caching.py
```diff
@@ -492,8 +492,8 @@ def test_sparse_function(self, operate_on_empty_cache):
 
         u = SparseFunction(name='u', grid=grid, npoint=1, nt=10)
 
-        # created: u, p_u, h_p_u, u_coords, d, h_d
-        ncreated = 6
+        # created: u, u(inds), p_u, h_p_u, u_coords, u_coords(inds), d, h_d
+        ncreated = 8
```
mloubout, line 496: since we now cache both `u` and `u(x)` to avoid recreating objects, this creates to cache entry per Function the first time. Same below and other tests

devito/types/basic.py
FabioLuporini: ok, but then move it below `__new__` ?

devito/types/basic.py
```diff
@@ -652,6 +640,8 @@ def __new__(cls, *args, **kwargs):
 
         # Store new instance in symbol cache
         Cached.__init__(newobj, newcls)
+        # Cache with indices as well to avoid recreating the base function
+        Cached.__init__(newobj, (newcls, *indices), alias=newcls)
```
FabioLuporini, line 644: calling `__init__` twice is really weird... you also end up changing state (`_cache_key_hash` in this case)!

see my other comment below

devito/types/caching.py
FabioLuporini: I think this would be cleaner?

```
def __init__(self, key, other_keys=None):
    self._cache_key_hash = hash(key)

    awr = AugmentedWeakRef(self, self._cache_meta())
    for i in (key,) + as_tuple(other_keys):
         _SymbolCache[i] = awr
```

and should also avoid the need for calling `Cached.__init__` twice (which is conceptually wrong)

tests/test_caching.py
```diff
@@ -670,12 +670,15 @@ def test_solve(self, operate_on_empty_cache):
         del eqn
         del grid
 
+        # Only deleted `u` shift reference still existing
         # `u` points to the various Dimensions, the Dimensions point to the various
-        # spacing symbols, hence, we need three sweeps to clear up the cache
-        assert len(_SymbolCache) == 12
+        # spacing symbols, hence, we need four sweeps to clear up the cache
```
FabioLuporini, line 675: this comment deserves an extension for the extra sweep

devito/types/basic.py
```diff
@@ -611,20 +599,20 @@ class AbstractFunction(sympy.Function, Basic, Cached, Pickable, Evaluable):
...
-    def _cache_key(cls, *args, **kwargs):
-        """An AbstractFunction caches on the class type itself."""
-        return cls
-
     def __new__(cls, *args, **kwargs):
         options = kwargs.get('options', {'evaluate': False})
 
-        key = cls._cache_key(*args, **kwargs)
-        obj = cls._cache_get(key)
+        # Is it an existing shifted index
```
FabioLuporini, line 605: annoying nitpicking: question mark at end

devito/types/basic.py
```diff
@@ -611,20 +599,20 @@ class AbstractFunction(sympy.Function, Basic, Cached, Pickable, Evaluable):
...
+        obj = cls._cache_get((cls, *args))
+        if obj is not None:
+            return obj
 
+        # Does the base object exist
```
FabioLuporini, line 610: annoying nitpicking: question mark at end

devito/types/basic.py
mloubout: Well you "change" the state by replacing it with itself so it's not actually changing but I get your point.

devito/types/caching.py
mloubout: But this is wrong as `u(x+dx)` will have a different `_cache_key_hash`

FabioLuporini: even if you call it as:

```
Cached.__init__(newobj, newcls, other_keys=[(newcls, *indices)])
```

?

devito/types/basic.py
```diff
@@ -611,20 +599,20 @@ class AbstractFunction(sympy.Function, Basic, Cached, Pickable, Evaluable):
...
-        """An AbstractFunction caches on the class type itself."""
-        return cls
-
     def __new__(cls, *args, **kwargs):
         options = kwargs.get('options', {'evaluate': False})
 
-        key = cls._cache_key(*args, **kwargs)
-        obj = cls._cache_get(key)
+        # Is it an existing shifted index?
+        obj = cls._cache_get((cls, *args))
```
FabioLuporini, line 606: so here `args==indices` basically?

mloubout: Yes, at subs/new/.... it's called as `func(*indices, **kwargs)`

tests/test_caching.py
FabioLuporini: reminder

tests/test_caching.py
```diff
@@ -670,12 +692,17 @@ def test_solve(self, operate_on_empty_cache):
...
         del grid
 
-        # `u` points to the various Dimensions, the Dimensions point to the various
-        # spacing symbols, hence, we need three sweeps to clear up the cache
-        assert len(_SymbolCache) == 12
+        # We only deleted `u`, however we also cache shifted version created by the
+        # finite difference (u.dt, u.dx2). In this case we have three extra references
+        # to u(t + dt), u(x - h_x) and u(x + h_x) that have to be cleared.
+        # Then `u` points to the various Dimensions, the Dimensions point to the various
+        # spacing symbols, hence, we need five sweeps to clear up the cache.
```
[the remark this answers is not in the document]
mloubout: fixed

tests/test_caching.py
mloubout: fixed

Issue #1184, opened 2020-03-24
devito Data object issues


tjb900:
Some of the methods of devito.Data rely on attributes set in `__new__` and `__array_finalize__`, when these are not always called.

For example, this code:

```
from devito import Grid, Function
import pickle

g = Grid((20, 20))
u = Function(name='u', grid=g)

d1 = u.data
d2 = pickle.loads(pickle.dumps(d1))

try:
    # Will fail in array_finalize due to lack of _index_stash
    print(d2)
except Exception:
    traceback.print_exc()

try:
    # Fails in __del__ due to lack of memfree_args
    del d2
except Exception:
    traceback.print_exc()
```

demonstrates the issue.

```
<snip>
AttributeError: 'Data' object has no attribute '_index_stash'
<snip>
AttributeError: 'Data' object has no attribute '_memfree_args'
<snip>
```

tjb900:
I think there might be a case to be made that not calling `__array_finalize__` at unpickling time might be considered a numpy bug, since this document https://docs.scipy.org/doc/numpy/user/basics.subclassing.html seems to claim that `__array_finalize__` covers all methods of instance creation:

> Because `__array_finalize__` is the only method that always sees new instances being created, it is the sensible place to fill in instance defaults for new object attributes, among other tasks.

rhodrin:
`Data` objects are not designed 'stripped' away from their `Function` in this manner. Is there any reason why
```
from devito import Grid, Function
import pickle
g = Grid((20, 20))
u = Function(name='u', grid=g)
u2 = pickle.loads(pickle.dumps(u))
try:
    # Will fail in array_finalize due to lack of _index_stash
    print(u.data)
except Exception:
    traceback.print_exc()
try:
    # Fails in __del__ due to lack of memfree_args
    del u2
except Exception:
    traceback.print_exc()
```
Would not achieve what you want?

tjb900:
Well, that might be true, but it's relatively easy to end up with standalone `Data` objects.  e.g. this returns a `Data`:

```
def f():
    # create devito stuff
    # run operator
    output_i_want = u.data[:] * 2
    return output_i_want
```

And when these resulting objects are used in dask/distributed, they are pickled and then it makes for sad times.

Now, it would work if everyone knew to cast/copy everything to ndarray.  But I think moving a couple of things around in data.py should make it a non-issue anyway?

rhodrin:
[...]

PR #1369, opened 2020-07-02
`Data` tweaks

rhodrin:
Start of a PR to address some of the issues raised in #1184.

Currently only a `__reduce__` override has been introduced to raise an error is directly pickling a `Data` object.

In #1184 @tjb900  also states:

> I read this to mean that you cannot rely even on __new__ being called, which means the self._memfree_args reference in __del__ causes an error. This is a simple case of setting _memfree_args to None in __array_finalize__.

But `_memfree_args` is set to None here: https://github.com/devitocodes/devito/blob/c55e7a7fe534124f0a4b4d2b48ccc5a6cc169d07/devito/data/data.py#L95

So I'm probably missing something?

### DIFF
--- a/devito/data/data.py
+++ b/devito/data/data.py
@@ -5,6 +5,7 @@ import numpy as np
 
 from devito.data.allocators import ALLOC_FLAT
 from devito.data.utils import *
+from devito.logger import warning
 from devito.parameters import configuration
 from devito.tools import Tag, as_tuple, as_list, is_integer
 
@@ -77,7 +78,8 @@ class Data(np.ndarray):
         self._memfree_args = None
 
     def __reduce__(self):
-        raise NotImplementedError("Pickling of `Data` objects is not supported")
+        warning("Pickling of `Data` objects is not supported. Casting to `numpy.ndarray`")
+        return np.array(self).__reduce__()
 
     def __array_finalize__(self, obj):
         # `self` is the newly created object
--- a/devito/data/data.py
+++ b/devito/data/data.py
@@ -76,6 +76,9 @@ class Data(np.ndarray):
         self._allocator.free(*self._memfree_args)
         self._memfree_args = None
 
+    def __reduce__(self):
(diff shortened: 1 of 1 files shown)


tjb900:
Ah, oops, you are right.  Earlier in the same bug I mention the fact that array_finalize doesn't seem to be called when unpicking.

https://github.com/devitocodes/devito/issues/1184#issuecomment-603075388

rhodrin:
So we can potentially pickle `Data` doing something along the lines of:

```
    def __reduce__(self):
        self._distributor = None
        self._decomposition = (None,)*len(self.shape)
        pickled_state = super(Data, self).__reduce__()
        new_state = pickled_state[2] + (self.__dict__,)
        return (pickled_state[0], pickled_state[1], new_state)

    def __setstate__(self, state):
        self.__dict__.update(state[-1])
        super(Data, self).__setstate__(state[0:-1])
```
But if we go down this route how do we want to treat `self._distributor` and `self._decomposition` during the pickling process? I think we do something for this when pickling `Function`'s maybe? Will keep looking into it anyway.

FabioLuporini:
Thinking about this again, I wonder whether we rather want to keep it simple and **explicitly** disallow pickling of Data. Users will pickle Function and if they really (really) want, they will pickle numpy arrays casted from Data. 

Rationale: upon unpickling , I imagine the unpickled Function to be only necessary since it's a placeholder in an unpickled Operator, not for the data it (used to) carries. 

@navjotk @rhodrin perhaps I've missed something -- but when would pickling a Data object be necessary?

rhodrin:
> but when would pickling a Data object be necessary?

I think as we said before - it'd just be for 'user' convenience. So what about the following - if a `Data` object is pickled we cast it a `ndarray` and throw a warning? Maybe that's the 'best of both' solution (for the time being at least)?

FabioLuporini:
Merged, thanks

PR #1376, opened 2020-07-06
mpi data gatherer

rhodrin:
This PR adds a method to `Function`'s to enable the gathering of distributed `Data` onto a single rank.

~NOTE: PR'ing for CI spin - needs lots of clean-up, fixes and many more tests. Will update here with more details when appropriate.~


devito/data/data.py
```diff
@@ -174,9 +175,9 @@ def __str__(self):
         return super(Data, self._local).__str__()
 
     @_check_idx
-    def __getitem__(self, glb_idx, comm_type):
+    def __getitem__(self, glb_idx, comm_type, _gather=None):
```
FabioLuporini, line 178: variable name == method name ?

devito/data/data.py
```diff
@@ -195,32 +196,63 @@ def __getitem__(self, glb_idx, comm_type):
...
-                    send_val = local_val.data[loc_ind]
-                    reqs = comm.isend([send_ind, send_val], dest=send_rank)
-                    reqs.wait()
-                elif rank == send[index]:
-                    # Current rank is required to receive data from this index
-                    recval = comm.irecv(source=owners[index])
-                    local_dat = recval.wait()
-                    loc_ind = local_si[local_dat[0]]
-                    retval.data[loc_ind] = local_dat[1]
+            # TODO: These statements can potentially be fused?
```
FabioLuporini, line 199: yeah I can see there's some overlap... also the while-if-elif-else structure is identical AFAICT. We might want to abstract it away into a separate routine which takes callbacks (the bodies each each if-branch) as input? or is it just painful over-engineering? 
in the future, we might want to add other gather-like operations -- scatter, broadcast, ...

devito/data/data.py
rhodrin: Yeah, the whole variable possibly needs an overhaul.

So I'm thinking should it be allowed to be a `tuple`? i.e. you can gather simultaneously on multiple ranks. I'm not sure how often users would want this though + it adds a decent bit of complexity to the routine to save users (probably?) a couple of lines of code.

So would only be worth it if users are going to frequently do something along the lines of
```
a = f.data_gather(rank='all')
```

devito/data/data.py
rhodrin: Yes, it is identical at the moment. It possibly should stay identical depending on what functionality we want to allow - but we can discuss that.

> We might want to abstract it away into a separate routine

Yes, we definitely want to do that - it will make the fusion much cleaner. I'm just keeping them separate for the time being until I get all required functionality working with the gather, after that I'll have a good idea re. the best way to fuse them.

devito/data/data.py
```diff
@@ -471,6 +484,31 @@ def _set_global_idx(self, val, idx, val_idx):
                 mapped_idx.append(None)
         return as_tuple(mapped_idx)
 
+    def _gather(self, start=None, stop=None, step=1, rank=0):
+        """
+        Method for gathering distributed data into a NumPy array
+        on a single rank. See the the public ``data_gather`` method
```
FabioLuporini, line 490: double the

devito/types/dense.py
```diff
@@ -408,6 +408,26 @@ def data(self):
...
+        Parameters
+        ----------
+        rank : int
+            The rank onto which the data will be gathered.
+        step : int or tuple of ints
+            The `slice` step in each dimension.
+        start : int or tuple of ints
+            The `slice` start in each dimension.
+        stop : int or tuple of ints
+            The final point of the `slice` to include.
```
FabioLuporini, line 424: blank line below

devito/types/dense.py
```diff
@@ -408,6 +408,26 @@ def data(self):
...
+        Notes
+        -----
+        Notes will go here.
```
FabioLuporini, line 427: TODO

maybe also add something along the lines of "users should be aware that calling this method might blow up the memory since... so use it judiciously" -- what do you say?

FabioLuporini (commented):
minor comments left, but basically lovely and ready to go

devito/data/data.py
```diff
@@ -179,9 +179,9 @@ def __str__(self):
         return super(Data, self._local).__str__()
 
     @_check_idx
-    def __getitem__(self, glb_idx, comm_type):
+    def __getitem__(self, glb_idx, comm_type, gather=None):
         loc_idx = self._index_glb_to_loc(glb_idx)
-        if comm_type is index_by_index:
+        if comm_type is index_by_index or isinstance(gather, int):
```
mloubout, line 184: Should we have a default `gather=True` equivalent to `gather=0`? Or is it overkill

mloubout, line 184: Also there is a lot of `isinstance(gather, int)`, wouldn't it be nicer to have `gather = int(gather) if isinstance(gather, int) else None` at the beginning and then only needs `if gather` everywhere below. The `int(gather)` probably not necessary actually but just here for safety.

tests/test_data.py
```diff
@@ -1259,6 +1259,92 @@ def test_misc_data(self):
             assert(np.all(c.data[-1] == [4., 4., 6., 9., 8.]))
 
 
+class TestDataGather(object):
+
+    @pytest.mark.parallel(mode=4)
+    @pytest.mark.parametrize('rank', [0, 1, 2, 3])
+    def test_simple_gather(self, rank):
+        """ Fill me in."""
```
mloubout, line 1267: leftover?

devito/data/data.py
rhodrin: `gather` can be, and often is, `0` which is why is instance `isinstance` is used. But we could have the `kwarg` be `gather_rank` and then introduce `gather` as a `True/False` - leads to one more line of code, but code is then a bit tidier I guess.

tests/test_data.py
rhodrin: Thanks. the 3 instances of it will be fixed on next push.

devito/data/data.py
```diff
@@ -179,9 +179,10 @@ def __str__(self):
         return super(Data, self._local).__str__()
 
     @_check_idx
-    def __getitem__(self, glb_idx, comm_type):
+    def __getitem__(self, glb_idx, comm_type, gather_rank=None):
         loc_idx = self._index_glb_to_loc(glb_idx)
-        if comm_type is index_by_index:
+        gather = True if isinstance(gather_rank, int) else False
```
FabioLuporini, line 184: nitpicking for code homogeneity, perhaps for another PR: typically we call these variables "is_..." in this case "is_gather". The "is_" encodes the information that the variable is a boolean

FabioLuporini (approved):

tests/test_data.py
```diff
@@ -1259,6 +1259,92 @@ def test_misc_data(self):
...
+    @pytest.mark.parallel(mode=4)
+    @pytest.mark.parametrize('start, stop, step', [
+        (None, None, None),
+        (None, None, 2),
+        (None, None, -1),
+        (None, None, -2),
+        (1, 8, 3),
+        ((0, 4, 4), None, (2, 1, 1))])
+    def test_sliced_gather_3D(self, start, stop, step):
+        """ Test gather for various 2D slices."""
```
mloubout, line 1322: 3D 😬

rhodrin: oh, bleh, thanks. I'll add this and the thing @FabioLuporini mentioned to #1457.

PR #1377, opened 2020-07-07
Function/Expr shift utility

mloubout:
Add simple shifting utility to avoid having to go through subs and test.

### DIFF
--- a/devito/finite_differences/differentiable.py
+++ b/devito/finite_differences/differentiable.py
@@ -222,6 +222,13 @@ class Differentiable(sympy.Expr, Evaluable):
     def name(self):
         return "".join(f.name for f in self._functions)
 
+    def shift(self, dim, shift):
+        """
+        Shift  expression by `shift` along the Dimension `dim`.
+        For example u.shift(x, x.spacing) = u(x + h_x).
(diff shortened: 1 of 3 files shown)


devito/finite_differences/differentiable.py
```diff
@@ -133,6 +133,13 @@ def _subs(self, old, new, **hints):
                 continue
         return self.func(*args, evaluate=False)
 
+    def shift(self, dim, shift):
```
FabioLuporini, line 136: can you move it below `def diff` ? here is right in between all the private methods...

devito/finite_differences/differentiable.py
```diff
@@ -133,6 +133,13 @@ def _subs(self, old, new, **hints):
...
+        """
+        Shift  expression by `shift` along the dimension `dim`.
```
FabioLuporini, line 138: Dimension (capital)

devito/finite_differences/differentiable.py
mloubout: Thought it'd be good just next to `_subs` since it's basically a shortcut to it but sure.

PR #1378, opened 2020-07-07
Cancel previous CI at new commit

mloubout:
Long-awaited, there is an action for that now, sweet.

### DIFF
--- a/.github/workflows/cancel.yml
+++ b/.github/workflows/cancel.yml
@@ -1,0 +1,16 @@
+name: Cancel
+
+on: [push]
+
+jobs:
+  cancel:
+    name: 'Cancel Previous Runs'
+    runs-on: ubuntu-latest
+    timeout-minutes: 3
+    steps:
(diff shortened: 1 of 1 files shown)


rhodrin:
Looks good. Looking at the documentation:
```
This is a GitHub Action that will cancel any previous runs that are not completed for a given workflow.

This includes runs with a status of queued or in_progress.
```
Did you check how it interacts with 'depends on' jobs? They aren't even queued until the job it depends on is completed or cancelled - from what it says above I'd 'assume' they still go ahead?

mloubout:
I think it cancels the whole workflow, not the step, so the depends on should be canceled too, but haven't tried. May be good to try if you have an example.

rhodrin:
> I think it cancels the whole workflow, not the step, so the depends on should be canceled too, but haven't tried. May be good to try if you have an example.

Yeah, would be nice if it did - but that wasn't completely obvious from a quick read of the docs.

PR #1388, opened 2020-07-13
Add docker publish

mloubout:
Bulid and publish docker images to dockerhub/devitocodes at:
- PR mege into `devitocodes/devito:cpu-dev`
- release into `devitocodes/devito:cpu-v***`

GPU to be added once #1364  is merged

Tested in my fork and seems to work.


.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Check event name
+        run: echo ${{ github.event_name }}
+
+      - name: CPU image
+        if: github.event_name == 'push'
+        uses: docker/build-push-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+          repository: mloubout/devito
```
FabioLuporini, line 26: does this need to be your workspace?

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,38 @@
...
+          dockerfile: ./docker/Dockerfile
```
FabioLuporini, line 27: eventually, we'll add another one for Dockerfile.nvidia right?

.github/workflows/docker-publish.yml
mloubout: Oups, debug leftover should be `devitocodes`

.github/workflows/docker-publish.yml
mloubout: yep the one fro #1364 will be added

mloubout: added

PR #1390, opened 2020-07-14
Tensor API: improve behavior and interleaving with sympy. 

mloubout:
Lots of cleanup and improvement of Tensor/Vector:

- No more caching as these are only carrying cached components
- remove all the by hand mul and uses sympy with proper `_new` (lot cleaner)
- Improve priority
- Supports expressions with pure sympy Matrices

TODO:
- Add/fix docstring
- Some will be removed (such as _mul__, __rmul__) once sympy 1.7 is out
- allow `.apply(v=tensor)`. This one is tricky since the operator only has the components (Function) in parameters
- ...


devito/finite_differences/differentiable.py
```diff
@@ -157,30 +148,39 @@ def __getattr__(self, name):
         raise AttributeError("%r object has no attribute %r" % (self.__class__, name))
 
     # Override SymPy arithmetic operators
+    @call_highest_priority('__radd__')
```
FabioLuporini, line 151: I guess this is a bug fix, and it's getting tested..

devito/types/dense.py
```diff
@@ -1270,7 +1270,9 @@ def __init_finalize__(self, *args, **kwargs):
         self.time_dim = kwargs.get('time_dim', self.dimensions[self._time_position])
         self._time_order = kwargs.get('time_order', 1)
         super(TimeFunction, self).__init_finalize__(*args, **kwargs)
-
+        # Dynamically add derivative short-cuts
+        self._fd = generate_fd_shortcuts(self.dimensions, self.space_order,
```
FabioLuporini, line 1274: that's gonna **re**construct the space derivatives as well? since we also get them via ``super``

also, we need surrounding blank lines here

devito/types/sparse.py
```diff
@@ -41,7 +41,8 @@ def __init_finalize__(self, *args, **kwargs):
         self._space_order = kwargs.get('space_order', 0)
 
         # Dynamically add derivative short-cuts
-        self._fd = generate_fd_shortcuts(self)
+        self._fd = generate_fd_shortcuts(self.dimensions, self.space_order,
+                                         to=kwargs.get('time_order', 0))
```
FabioLuporini, line 45: that should only appear in a SparseTimeFunction

devito/types/tensor.py
```diff
@@ -1,19 +1,20 @@
...
+from devito.tools import flatten
+from devito.types.basic import AbstractTensor
 from devito.types.dense import Function, TimeFunction
 from devito.types.utils import NODE
 
 __all__ = ['TensorFunction', 'TensorTimeFunction', 'VectorFunction', 'VectorTimeFunction']
 
 
-class TensorFunction(AbstractTensor, Differentiable):
+class TensorFunction(AbstractTensor):
```
FabioLuporini, line 17: No more Differentiable,  but below it still uses

```
_op_priority = Differentiable._op_priority + 1.
```

that's quite odd

devito/finite_differences/differentiable.py
mloubout: It's tested through the tensor tests yes. Was done manually before but basically this enforce that it always uses the add/mul for the one with the highest priority between self and other.

devito/types/tensor.py
mloubout: This is just for priority reasons, so that Tensor Mul is higher priority than Differentiable. I could just put the value to avoid the "dependency" but feels clearer with it.

devito/types/dense.py
mloubout: Good point, will rethink it

mloubout: fixed

devito/types/sparse.py
mloubout: Then would generate shortcuts twice.

devito/finite_differences/differentiable.py
```diff
@@ -72,7 +58,8 @@ def grid(self):
         try:
             return grids.pop()
         except KeyError:
-            raise ValueError("No grid found")
+            return None
+            # raise ValueError("No grid found")
```
FabioLuporini, line 62: leftover?

devito/finite_differences/differentiable.py
```diff
@@ -102,6 +89,10 @@ def staggered(self):
     def is_Staggered(self):
         return any([getattr(i, 'is_Staggered', False) for i in self._args_diff])
 
+    @cached_property
+    def is_TimeDependent(self):
+        return any([i.is_Time for i in self.dimensions])
```
FabioLuporini, line 94: square brackets are redundant within any/all (which would already produce generators)

devito/finite_differences/differentiable.py
```diff
@@ -157,30 +148,39 @@ def __getattr__(self, name):
...
 
+    @call_highest_priority('__add__')
     def __radd__(self, other):
         return Add(other, self)
 
+    @call_highest_priority('__sub__')
     def __sub__(self, other):
         return Add(self, -other)
 
+    @call_highest_priority('__rsub__')
```
FabioLuporini, line 167: rsub or sub ?

devito/finite_differences/tools.py
```diff
@@ -95,59 +95,50 @@ def deriv_name(dims, orders):
...
-                            dims=d, fd_order=o, side=right)
-            name_fd = 'd%sr' % name
-            desciption = 'right first order derivative w.r.t dimension %s' % d.name
-            derivatives[name_fd] = (deriv, desciption)
+        # Add centered first derivatives
+        deriv = partial(diff_f, deriv_order=1, dims=d, fd_order=o, side=centered)
+        name_fd = 'd%sc' % name
+        desciption = 'centered derivative staggered w.r.t dimension %s' % d.name
+        derivatives[name_fd] = (deriv, desciption)
+        # if not function.is_Staggered:
```
FabioLuporini, line 131: leftover

devito/types/basic.py
```diff
@@ -487,76 +487,60 @@ class AbstractTensor(sympy.ImmutableDenseMatrix, Basic, Cached, Pickable, Evalua
...
 
         return newobj
 
-    __hash__ = Cached.__hash__
+    __hash__ = sympy.ImmutableDenseMatrix.__hash__
+
+    def doit(self, **hint):
+        return self
 
     def __init_finalize__(self, *args, **kwargs):
```
FabioLuporini, line 515: nitpicking: this method should be moved right below `new` (or whatever closest thing to `new`...)

devito/types/dense.py
```diff
@@ -959,7 +959,8 @@ def __init_finalize__(self, *args, **kwargs):
             raise TypeError("`space_order` must be int or 3-tuple of ints")
 
         # Dynamically add derivative short-cuts
-        self._fd = generate_fd_shortcuts(self)
+        self._fd = generate_fd_shortcuts(self.dimensions, self.space_order,
```
FabioLuporini, line 962: so we're not passing `self` anymore?

devito/types/dense.py
```diff
@@ -959,7 +959,8 @@ def __init_finalize__(self, *args, **kwargs):
...
+                                         to=getattr(self, 'time_order', 0))
```
FabioLuporini, line 963: this is a bit horrible. The OO way to do this is to use methods and override them in the subclass

```
class Function(...):

     def _generate_fd_shortcuts(self):
            return generate_fd_shortcuts(self.dimensions, self.space_order)


class TimeFunction(...):

          def _generate_fd_shortcuts(self):
                return generate_fd_shortcuts(self.dimensions, self.space_order, self.time_order)
```

FabioLuporini: in fact, we should stick to our hooks convention, and call it something like `__fd_setup__`

devito/types/sparse.py
FabioLuporini: same as before, this should be moved within a `__fd_setup__` method

devito/types/tensor.py
```diff
@@ -31,16 +67,22 @@ class TensorFunction(AbstractTensor, Differentiable):
...
     _class_priority = 10
+    _op_priority = Differentiable._op_priority + 1.
 
     def __init_finalize__(self, *args, **kwargs):
-        self._is_symmetric = kwargs.get('symmetric', True)
-        self._is_diagonal = kwargs.get('diagonal', False)
-        self._staggered = kwargs.get('staggered', self.space_dimensions)
-        self._grid = kwargs.get('grid')
-        self._space_order = kwargs.get('space_order', 1)
+        if args:
```
FabioLuporini, line 76: `if len(args) != ... : raise ValueError(...)`

?

devito/types/tensor.py
```diff
@@ -31,16 +67,22 @@ class TensorFunction(AbstractTensor, Differentiable):
...
+            comps = flatten(args[2])
+            grid = comps[0].grid
+            dimensions = None if grid else comps[0].dimensions
```
FabioLuporini, line 79: you mean the other way round? if yes (maybe I'm wrong...) is this path actually tested? if not, what would `comps[0]` be to carry `dimensions` ? another tensorial object? 

to ease understanding of this class, you may consider adding some asserts and/or checks on the input values

devito/types/tensor.py
```diff
@@ -62,15 +104,19 @@ def __subfunc_setup__(cls, *args, **kwargs):
         stagg = kwargs.get("staggered", None)
         name = kwargs.get("name")
         symm = kwargs.get('symmetric', True)
+        diag = kwargs.get('diagonal', False)
+
+        funcs = []
```
FabioLuporini, line 109: actually, this line can be dropped I think?

and below in the for loop you may use `funcs` instead of `funcs2` as variable name

devito/types/tensor.py
```diff
@@ -79,211 +125,89 @@ def __subfunc_setup__(cls, *args, **kwargs):
...
-    def grid(self):
-        return self._grid
+    def space_order(self):
+        """The space order for all components."""
+        return ({a.space_order for a in self} - {None}).pop()
 
     @property
-    def name(self):
-        return self._name
+    def time_order(self):
```
FabioLuporini, line 197: this should only be in a `*Time*` class

devito/types/basic.py
```diff
@@ -487,76 +487,60 @@ class AbstractTensor(sympy.ImmutableDenseMatrix, Basic, Cached, Pickable, Evalua
...
-    @classmethod
-    def __dtype_setup__(cls, **kwargs):
-        """Extract the object data type from ``kwargs``."""
-        return None
+    def _eval_matrix_mul(self, other):
+        other_len = other.rows*other.cols
+        new_len = self.rows*other.cols
+        new_mat = [self.zero]*new_len
+
+        # if we multiply an n x 0 with a 0 x m, the
```
georgebisbas, line 523: WOW

devito/types/tensor.py
```diff
@@ -23,6 +24,41 @@ class TensorFunction(AbstractTensor, Differentiable):
...
+        Define how the Function is staggered.
+    initializer : callable or any object exposing the buffer interface, optional
+        Data initializer. If a callable is provided, data is allocated lazily.
+    allocator : MemoryAllocator, optional
+        Controller for memory allocation. To be used, for example, when one wants
+        to take advantage of the memory hierarchy in a NUMA architecture. Refer to
+        `default_allocator.__doc__` for more information.
+    padding : int or tuple of ints, optional
+        .. deprecated:: shouldn't be used; padding is now automatically inserted.
+
```
georgebisbas, line 59: empty line?

devito/types/tensor.py
```diff
@@ -79,211 +125,89 @@ def __subfunc_setup__(cls, *args, **kwargs):
...
-    def __getattr__(self, name):
+    @call_highest_priority('__mul__')
+    def __rmul__(self, other):
         """
-        Try calling a dynamically created FD shortcut.
-
-        Notes
-        -----
-        This method acts as a fallback for __getattribute__
+        Prevents Functions to be interpreted as matrices in 2D becaue of `.shape`
```
georgebisbas, line 131: because

devito/types/tensor.py
mloubout: >what would comps[0] be to carry dimensions
A differentiable expression for example, i.e `div(VectorFunction)` will be a Tensor with derivatives as component (or evaluated derivatives) or any expression carrying dimensions but no grid.

devito/finite_differences/differentiable.py
```diff
@@ -157,30 +147,39 @@ def __getattr__(self, name):
...
     def __mul__(self, other):
         return Mul(self, other)
 
+    @call_highest_priority('__mul__')
     def __imul__(self, other):
         return Mul(self, other)
 
+    @call_highest_priority('__mul__')
     def __rmul__(self, other):
         return Mul(other, self)
```
FabioLuporini, line 185: missing `call_highest_priority` for `__pow__`?

devito/finite_differences/tools.py
```diff
@@ -95,59 +95,49 @@ def deriv_name(dims, orders):
...
-                            dims=d, fd_order=o, side=right)
-            name_fd = 'd%sr' % name
-            desciption = 'right first order derivative w.r.t dimension %s' % d.name
-            derivatives[name_fd] = (deriv, desciption)
+        # Add centered first derivatives
+        deriv = partial(diff_f, deriv_order=1, dims=d, fd_order=o, side=centered)
+        name_fd = 'd%sc' % name
+        desciption = 'centered derivative staggered w.r.t dimension %s' % d.name
+        derivatives[name_fd] = (deriv, desciption)
+        # Left
```
FabioLuporini, line 131: oh, do we now have left-right also for higher order derivatives?

devito/types/basic.py
```diff
@@ -487,76 +487,60 @@ class AbstractTensor(sympy.ImmutableDenseMatrix, Basic, Cached, Pickable, Evalua
...
-        Cached.__init__(newobj, newcls)
+    def _new(cls, *args, **kwargs):
+        if args:
+            try:
+                # Constructor if input is (rows, cols, lambda)
+                newobj = super(AbstractTensor, cls)._new(*args)
+            except ValueError:
+                # Constructor if input is list of list as (row, cols, list_of_list)
+                # doesn't work as it expects a flattened.
+                newobj = super(AbstractTensor, cls)._new(args[2])
```
FabioLuporini, line 498: check `args[2]` really is what you expect it to be? or at least `if len(args) != 3 raise ValueError ...` ?

devito/types/basic.py
FabioLuporini: comments always start with capital letters

devito/types/basic.py
```diff
@@ -487,76 +487,60 @@ class AbstractTensor(sympy.ImmutableDenseMatrix, Basic, Cached, Pickable, Evalua
...
+            self_cols = self.cols
+            mat = self._mat
+            other_mat = other._mat
+            for i in range(new_len):
+                row, col = i // other.cols, i % other.cols
+                row_indices = range(self_cols*row, self_cols*(row+1))
+                col_indices = range(col, other_len, other.cols)
+                vec = [mat[a]*other_mat[b] for a, b in zip(row_indices, col_indices)]
+                new_mat[i] = sum(vec)
+        # Get new class and return product
```
FabioLuporini, line 535: blank line before this comment?

devito/types/basic.py
```diff
@@ -601,6 +585,9 @@ class AbstractFunction(sympy.Function, Basic, Cached, Pickable, Evaluable):
                                          interpolation scheme, instead of linear
                                          interpolators.
     """
```
FabioLuporini, line 587: probably blank line below (check other classes)

devito/types/dense.py
```diff
@@ -958,15 +958,17 @@ def __init_finalize__(self, *args, **kwargs):
...
-
+        self.__fd_setup__()
         # Flag whether it is a parameter or a variable.
         # Used at operator evaluation to evaluate the Function at the
         # variable location (i.e. if the variable is staggered in x the
         # parameter has to be computed at x + hx/2)
         self._is_parameter = kwargs.get('parameter', False)
 
+    def __fd_setup__(self):
+        # Dynamically add derivative short-cuts
```
FabioLuporini, line 969: docstring?

devito/types/dense.py
```diff
@@ -958,15 +958,17 @@ def __init_finalize__(self, *args, **kwargs):
         else:
             raise TypeError("`space_order` must be int or 3-tuple of ints")
 
-        # Dynamically add derivative short-cuts
-        self._fd = generate_fd_shortcuts(self)
-
+        self.__fd_setup__()
```
FabioLuporini, line 961: no, it should be:

```
self._fd = self.__fd_setup__()
```

that is, the assignment should be explicitly visible in `__init__`, not buried in the private method

devito/types/sparse.py
```diff
@@ -42,7 +42,11 @@ def __init_finalize__(self, *args, **kwargs):
         self._space_order = kwargs.get('space_order', 0)
 
         # Dynamically add derivative short-cuts
-        self._fd = generate_fd_shortcuts(self)
+        self.__fd_setup__()
```
FabioLuporini, line 45: same comments as before

devito/types/tensor.py
FabioLuporini: Btw, remind me why not Differentiable anymore?

devito/types/tensor.py
```diff
@@ -23,6 +24,41 @@ class TensorFunction(AbstractTensor, Differentiable):
...
+        (``rp``) sides of a generic point of interest.
+    shape : tuple of ints, optional
+        Shape of the domain region in grid points. Only necessary if ``grid`` isn't given.
+    dimensions : tuple of Dimension, optional
+        Dimensions associated with the object. Only necessary if ``grid`` isn't given.
+    dtype : data-type, optional
+        Any object that can be interpreted as a numpy data type. Defaults
+        to ``np.float32``.
+    staggered : Dimension or tuple of Dimension or Stagger, optional
+        Define how the Function is staggered.
```
FabioLuporini, line 50: how the TensorFunction

devito/types/tensor.py
```diff
@@ -23,6 +24,41 @@ class TensorFunction(AbstractTensor, Differentiable):
 
     Parameters
     ----------
+    name : str
+        Name of the symbol.
+    grid : Grid, optional
+        Carries shape, dimensions, and dtype of the Function. When grid is not
```
FabioLuporini, line 30: of the TensorFunction

devito/types/tensor.py
```diff
@@ -23,6 +24,41 @@ class TensorFunction(AbstractTensor, Differentiable):
 
     Parameters
     ----------
+    name : str
+        Name of the symbol.
```
FabioLuporini, line 28: symbol or TensorFunction?

do we really have "symbol" in the sibling docstrings?

devito/types/tensor.py
FabioLuporini: indeed, to be dropped

devito/types/tensor.py
```diff
@@ -31,16 +67,22 @@ class TensorFunction(AbstractTensor, Differentiable):
         Staggering of each component, needs to have the size of the tensor. Defaults
         to the Dimensions.
     """
```
FabioLuporini, line 69: blank line below

devito/types/tensor.py
```diff
@@ -31,16 +67,22 @@ class TensorFunction(AbstractTensor, Differentiable):
...
+    _is_TimeDependent = False
```
FabioLuporini, line 70: blank line below, `_sub_type` logically belongs to another group of class attributes

devito/types/tensor.py
```diff
@@ -31,16 +67,22 @@ class TensorFunction(AbstractTensor, Differentiable):
...
     _sub_type = Function
-    _op_priority = Differentiable._op_priority + 2.
     _class_priority = 10
+    _op_priority = Differentiable._op_priority + 1.
```
FabioLuporini, line 73: sorry for repeating myself, but the fact that you're accessing `Differentiable._op_priority` while _not_ subclassing it is symptomatic that there's something _conceptually_ off

devito/types/tensor.py
FabioLuporini: reminder

devito/finite_differences/differentiable.py
mloubout: Don't think so, `pow` is a bi weird, for like `a + b` it's fine for for `a^b` what would be the opposite? There is not `rpow` or concept of it.

devito/finite_differences/tools.py
mloubout: yes, less confusing to have it for all.

devito/types/tensor.py
mloubout: >while not subclassing it is symptomatic that there's something conceptually off

That's not true. This is purely a precedence concept, the same may a `mul` as precedence over an `add` but doesn't subclass it. There is nothing wrong with defining relative priorities between non-inherited object so that what to do when we have both is properly defined.

devito/types/basic.py
mloubout: That's just from sympy  so want to leave it as is, only needed because of line 534 that explicitly calls `sympy.Add` in sympy.

devito/types/basic.py
mloubout: sympy.Matrix already has proper checks and error message if it's not what expected, no reason to duplicate it.

devito/types/tensor.py
mloubout: Thought wrote it somewhere. The components are differentiable in a functional analysis sense, we don't support matricial differential operation for say.

devito/types/tensor.py
mloubout: Not needed, would already have been caugtht by sympy.

devito/types/tensor.py
mloubout: Lemme recheck it, think was tricky to only use one.

devito/types/tensor.py
mloubout: Well that one is nasty. Technically could do something like `v + u` where `v` is a VectorFunction and `u` a TimeFunction which would end up having a time_order but would need costly arg checks (this one is easy but may not be) to transfer to VectorTimeFunction so it's safer IMO to have it here in case a component end up being time-dependent.

devito/types/tensor.py
mloubout: because sympy assumes that everything with a `.shape` is a matrix but out Function have a `shape` that is not compatible and should not be processed as a "matrix shape". Basically fucks up sympy, that;s what the `is_Matrix=Flase, is_MatrixLike=False` are for but will only be processed properly in 1.7

devito/types/tensor.py
FabioLuporini: perhaps leave a TODO then?

also, the implementation looks broken? you're accessing `.space_order` instead of `.time_order`, and you're assuming the time orders are homogeneous. Should you not explicitly check that out, and raise an Exception if that is not the case?

about promotion to VectorTimeFunction... I guess you would do that in `_new` right? I don't understand why you say it'd be costly, do you not just have to check whether at least one arg is time-depedent?

devito/types/tensor.py
```diff
@@ -79,211 +124,89 @@ def __subfunc_setup__(cls, *args, **kwargs):
...
+        return ({a.space_order for a in self} - {None}).pop()
 
     @property
-    def name(self):
-        return self._name
+    def time_order(self):
+        """The time order for all components."""
+        return ({a.space_order for a in self} - {None}).pop()
 
     @property
```
FabioLuporini, line 200: cached_property? this check looks relatively costly

devito/types/tensor.py
```diff
@@ -293,140 +216,61 @@ def values(self):
...
 
 class VectorFunction(TensorFunction):
     """
     Vector valued space varying Function as a rank 1 tensor of Function.
     """
+
     is_VectorValued = True
     is_TensorValued = False
-
+    _is_TimeDependent = False
```
FabioLuporini, line 268: blank line below this one

there are a few other blank-line-related unaddressed comments. I know, tedious and quite useless, but as you know I like the code to be as homogenous as possible

devito/types/tensor.py
mloubout: > don't understand why you say it'd be costly, do you not just have to check whether at least one arg is time-dependent?

Then you end up with odd circular imports and dependency.

>you're accessing .space_order instead of .time_order.

That's a bug, fixed.

devito/types/sparse.py

devito/types/tensor.py
mloubout: fixed

devito/types/tensor.py
mloubout: fixed

devito/types/tensor.py
mloubout: fixed

devito/types/tensor.py
mloubout: fixed

FabioLuporini (approved):
LGTM

devito/types/tensor.py
```diff
@@ -79,211 +128,84 @@ def __subfunc_setup__(cls, *args, **kwargs):
...
-        return self._is_symmetric
-
-    @property
-    def indices(self):
-        return self._indices
-
-    @property
-    def staggered(self):
-        return self._staggered
+    def is_TimeDependent(self):
```
FabioLuporini, line 186: Reminder for next PR: this property is probably useless as soon as you rename the class attribute `_is_TimeDependent` as `is_TimeDependent`

FabioLuporini (approved):

PR #1406, opened 2020-07-22
 Example: add second-order viscoacoustic equations

nogueirapeterson:
@lauerami, @speglich, and I developed the second-order forward and adjoint viscoacoustic equations in Devito, for the approaches of SLS (Standard Linear Solid), Ren and Deng. We hope that the Devito team appreciates the contribution.


examples/seismic/viscoacoustic/operators.py
```diff
@@ -6,9 +6,9 @@
 from examples.seismic import PointSource, Receiver
 
 
-def blanch_symes(model, geometry, v, p, **kwargs):
+def sls_1st_stencil(model, geometry, v, p, **kwargs):
```
rhodrin, line 9: What's the significance behind the `1st` here? Terminology from one of the references?

examples/seismic/viscoacoustic/operators.py
```diff
@@ -63,7 +62,85 @@ def blanch_symes(model, geometry, v, p, **kwargs):
     return [u_v, u_r, u_p]
 
 
-def ren(model, geometry, v, p, **kwargs):
+def sls_2nd_stencil(model, geometry, v, p, **kwargs):
```
rhodrin, line 65: Ok, I see, but 1st and 2nd seems a bit vague for the code here. Are you able to be a bit more specific?

examples/seismic/viscoacoustic/operators.py
```diff
@@ -63,7 +62,85 @@ def blanch_symes(model, geometry, v, p, **kwargs):
...
+    """
+    Stencil created from Bai (2014) viscoacoustic wave equation.
```
rhodrin, line 67: And what exactly do you mean by 'stencil' here? Would something along the lines of 'Implementation of the 2nd order viscoacoustic wave-equation from Bai (2014)' be better?

examples/seismic/viscoacoustic/operators.py
```diff
@@ -63,7 +62,85 @@ def blanch_symes(model, geometry, v, p, **kwargs):
...
+    https://library.seg.org/doi/10.1190/geo2013-0030.1
+
+    Parameters
+    ----------
+    v : VectorTimeFunction
+        Particle velocity.
+    p : TimeFunction
+        Pressure field.
+    """
+    forward = kwargs.get('forward')
```
rhodrin, line 78: `forward = kwargs.get('forward', True)`

examples/seismic/viscoacoustic/operators.py
```diff
@@ -63,7 +62,85 @@ def blanch_symes(model, geometry, v, p, **kwargs):
...
+    t_s = (sp.sqrt(1.+1./qp**2)-1./qp)/f0
+
+    # The strain relaxation parameter
+    t_ep = 1./(f0**2*t_s)
+
+    # The relaxation time
+    tt = (t_ep/t_s)-1.
+
+    # Density
+    rho = 1. / b
```
rhodrin, line 97: Is there a reason this is not stored directly in `model`?

examples/seismic/viscoacoustic/operators.py
```diff
@@ -93,14 +170,93 @@ def ren(model, geometry, v, p, **kwargs):
...
-    u_p = Eq(p.forward, damp * pde_u)
+    pde_p = p - s * vp * vp * rho * div(v.forward) + \
+        s * ((vp * vp * rho) / (w0 * qp)) * div(b * grad(p))
+    u_p = Eq(p.forward, damp * pde_p)
 
     return [u_v, u_p]
 
 
-def deng_mcmechan(model, geometry, v, p, **kwargs):
+def ren_2nd_stencil(model, geometry, v, p, **kwargs):
```
rhodrin, line 180: Can't this and the above be combined and have a `'case'` `kwarg`?

examples/seismic/viscoacoustic/operators.py
```diff
@@ -63,7 +62,85 @@ def blanch_symes(model, geometry, v, p, **kwargs):
...
+    # The strain relaxation parameter
+    t_ep = 1./(f0**2*t_s)
+
+    # The relaxation time
+    tt = (t_ep/t_s)-1.
+
+    # Density
+    rho = 1. / b
+
+    r = TimeFunction(name="r", grid=model.grid, time_order=2, space_order=space_order,
```
rhodrin, line 99: Shouldn't this be part of `model`? If it was the appropriate stencil and case could be constricted depending on what fields and `kwargs` are passed right?

examples/seismic/viscoacoustic/viscoacoustic_example.py
```diff
@@ -40,31 +39,36 @@ def run(shape=(50, 50), spacing=(20.0, 20.0), tn=1000.0,
 
 def test_viscoacoustic():
     _, _, _, [rec] = run()
-    assert np.isclose(norm(rec), 18.7749, atol=1e-3, rtol=0)
+    assert np.isclose(norm(rec), 684.4527, atol=1, rtol=0)
```
rhodrin, line 42: What's going on here? This looks stange?

rhodrin:
Thanks again for this contribution. Looks good! I had a quick read through and added a few comments above, will do another pass soonish. Two main comments are:
1. It seems like there's some redundancy and through using a couple of `kwargs` and a few adjustments to `model` some tidying and cutting down can be done.
2. Maybe I missed them, but couldn't see any correctness tests being added (e.g. a test in the example added to ci which checks wave-field and receiver norms)? I only saw some tests checking things actually run?

examples/seismic/viscoacoustic/operators.py
nogueirapeterson: We used `stencil` based on available examples. But we agree with you, `order` is better. We will change.

examples/seismic/viscoacoustic/operators.py
nogueirapeterson: Yes. We agree. We will change.

examples/seismic/viscoacoustic/operators.py
nogueirapeterson: okay. We will fix.

examples/seismic/viscoacoustic/operators.py
nogueirapeterson: Yes. We chose this approach based on the available viscoelastic example.

examples/seismic/viscoacoustic/operators.py
nogueirapeterson: We understand this approach but based on the available examples we choose to make this case option on equation methods, before calling these methods.

examples/seismic/viscoacoustic/operators.py
nogueirapeterson: The variable `b` is already part of the model. The` rho` is computed from `b`.

examples/seismic/viscoacoustic/viscoacoustic_example.py
nogueirapeterson: These values for norm and tolerance are expected because we are considering the second-order SLS equation as a reference. We will fix it to compute all first and second-order equations tests.

nogueirapeterson:
@rhodrin Thanks for the review, we appreciate your comments and learn a lot with them! See you soon.

examples/seismic/viscoacoustic/operators.py
```diff
@@ -130,14 +287,157 @@ def deng_mcmechan(model, geometry, v, p, **kwargs):
...
+
+    eq_stencil = stencils[('sls', time_order)]
+    eqn = eq_stencil(model, geometry, v, p, forward=forward, save=kwargs.get('save'))
+
+    return eqn
+
+
+def ren(model, geometry, v, p, forward=True, **kwargs):
+    """
+    Implementation of the 1st order viscoacoustic wave-equation and 2nd order
```
georgebisbas, line 396: Maybe: Implementation of the 1st and 2nd order viscoacoustic wave-equation from Ren et al. (2014).

examples/seismic/viscoacoustic/operators.py
```diff
@@ -130,14 +287,157 @@ def deng_mcmechan(model, geometry, v, p, **kwargs):
...
+    eq_stencil = stencils[('deng_mcmechan', time_order)]
+    eqn = eq_stencil(model, geometry, v, p, forward=forward, save=kwargs.get('save'))
+
+    return eqn
+
+
+def ForwardOperator(model, geometry, space_order=4, kernel='sls', time_order=2,
+                    save=False, **kwargs):
     """
     Construct method for the forward modelling operator in a viscoacoustic media.
```
georgebisbas, line 442: maybe:
in viscoacoustic media
or
in a viscoacoustic medium

examples/seismic/viscoacoustic/operators.py
```diff
@@ -151,26 +451,29 @@ def ForwardOperator(model, geometry, space_order=4, kernel='blanch_symes', save=
     space_order : int, optional
         Space discretization order.
     kernel : selects a visco-acoustic equation from the options below:
```
georgebisbas, line 453: kernel: string, optional
Selects...

examples/seismic/viscoacoustic/operators.py
```diff
@@ -193,4 +496,57 @@ def ForwardOperator(model, geometry, space_order=4, kernel='blanch_symes', save=
                     name='Forward', **kwargs)
 
 
-kernels = {'blanch_symes': blanch_symes, 'ren': ren, 'deng_mcmechan': deng_mcmechan}
+def AdjointOperator(model, geometry, space_order=4, kernel='sls', time_order=2, **kwargs):
+    """
+    Construct an adjoint modelling operator in an viscoacoustic media.
```
georgebisbas, line 501: same as before

examples/seismic/viscoacoustic/viscoacoustic_example.py
```diff
@@ -19,17 +18,17 @@ def viscoacoustic_setup(shape=(50, 50), spacing=(15.0, 15.0), tn=500., space_ord
...
+        kernel='sls', time_order=2, **kwargs):
 
     solver = viscoacoustic_setup(shape=shape, spacing=spacing, nbl=nbl, tn=tn,
-                                 space_order=space_order, constant=constant,
-                                 kernel=kernel, **kwargs)
+                                 space_order=space_order, preset=preset,
+                                 kernel=kernel, time_order=time_order, **kwargs)
     info("Applying Forward")
 
     # Define receiver geometry (spread across x, just below surface)
```
georgebisbas, line 34: leftover?

examples/seismic/viscoacoustic/wavesolver.py
```diff
@@ -20,18 +20,22 @@ class ViscoacousticWaveSolver(object):
...
-    def __init__(self, model, geometry, space_order=4, kernel='blanch_symes', **kwargs):
+    def __init__(self, model, geometry, space_order=4, kernel='sls', time_order=2,
+                 **kwargs):
         self.model = model
         self.geometry = geometry
 
         self.space_order = space_order
         self.kernel = kernel
+        self.time_order = time_order
         # Cache compiler options
```
georgebisbas, line 39: not sure this comment here is helpful?

examples/seismic/viscoacoustic/wavesolver.py
```diff
@@ -87,16 +98,17 @@ def forward(self, src=None, rec=None, v=None, r=None, p=None, qp=None, b=None,
...
+                                    time_order=self.time_order,
+                                    space_order=self.space_order)
 
         # Create the forward wavefield if not provided
         p = p or TimeFunction(name="p", grid=self.model.grid, save=save_t,
-                              time_order=1, space_order=self.space_order,
+                              time_order=self.time_order, space_order=self.space_order,
                               staggered=NODE)
 
         # Memory variable:
```
georgebisbas, line 109: what is the memory variable?

speglich: Hi George, thank you for your review! The R variable is described as memory variable on geophysics domain and is not connected with any computacional expect. This variable contain information about energy loss of stress-strain, and for containing this information is described as a memory variable.

rhodrin:
@nogueirapeterson @speglich Thanks for the latest updates. Would you mind rebasing this on top of the latest master (so that all the tests run) and then myself and @mloubout will take a further look before the end of the week.

speglich:
Resolving these merge issues.

speglich:
Hi @rhodrin! I noticed that the pipeline changed. The new ** Examples with MPI ** test are failing, including in old versions of this PR (We checked this). An initial investigation of the error raised by this test, take us to think about a wrong code generation. We will investigate the error and open an issue, right?

examples/seismic/viscoacoustic/operators.py
```diff
@@ -63,9 +62,87 @@ def blanch_symes(model, geometry, v, p, **kwargs):
...
+        pde_r = r + s * (tt / t_s) * p - s * (1. / t_s) * r
+        u_r = Eq(r.backward, damp * pde_r)
+
+        u_h = Eq(h, (1. + tt) * rho * p)
+
+        u_v = Eq(v, b * grad(h))
+
+        u_g = Eq(g, rho * r.backward)
+
+        u_w = Eq(w, b * grad(g))
```
mloubout, line 134: Are these temps really necessary? Is there an issue if you just put
```
 pde_p = 2. * p - damp * p.forward + s * s * vp * vp * div(b * grad((1. + tt) * rho * p))) - \
            s * s * vp * vp * div(b * grad(rho * r.backward))

```

You can still split the expression in a few lines if you want to avoid one long expression but the extra function/vector shulnd't be necessary.

examples/seismic/viscoacoustic/operators.py
```diff
@@ -93,16 +170,96 @@ def ren(model, geometry, v, p, **kwargs):
...
+
+    # Bulk modulus
+    bm = rho * (vp * vp)
+
+    if forward:
+
+        # Auxiliary functions
+        h = TimeFunction(name="h", grid=model.grid, time_order=2, space_order=space_order,
+                         staggered=NODE)
+        w = VectorTimeFunction(name="w", grid=model.grid, time_order=2,
```
mloubout, line 219: Probably not needed tmps as well.

examples/seismic/viscoacoustic/operators.py
```diff
@@ -93,16 +170,96 @@ def ren(model, geometry, v, p, **kwargs):
...
+        return [u_h, u_v, u_w, u_p]
+
+    else:
+
+        # Auxiliary functions
+        w = VectorTimeFunction(name="w", grid=model.grid, time_order=2,
+                               space_order=space_order)
+        h = TimeFunction(name="h", grid=model.grid, time_order=2, space_order=space_order,
+                         staggered=NODE)
+        g = TimeFunction(name="g", grid=model.grid, time_order=2, space_order=space_order,
```
mloubout, line 242: same

examples/seismic/viscoacoustic/operators.py
```diff
@@ -130,16 +287,159 @@ def deng_mcmechan(model, geometry, v, p, **kwargs):
...
+
+    bm = rho * (vp * vp)
+
+    if forward:
+
+        # Auxiliary functions
+        w = VectorTimeFunction(name="w", grid=model.grid, time_order=2,
+                               space_order=space_order)
+        h = TimeFunction(name="h", grid=model.grid, time_order=2, space_order=space_order,
+                         staggered=NODE)
```
mloubout, line 334: same

examples/seismic/viscoacoustic/operators.py
```diff
@@ -130,16 +287,159 @@ def deng_mcmechan(model, geometry, v, p, **kwargs):
...
+        u_p = Eq(p.forward, damp * pde_p)
+
+        return [u_h, u_v, u_w, u_p]
+
+    else:
+
+        # Auxiliary functions
+        h = TimeFunction(name="h", grid=model.grid, time_order=2, space_order=space_order,
+                         staggered=NODE)
+        g = TimeFunction(name="g", grid=model.grid, time_order=2, space_order=space_order,
```
mloubout, line 353: same

examples/seismic/viscoacoustic/operators.py
```diff
@@ -183,7 +487,7 @@ def ForwardOperator(model, geometry, space_order=4, kernel='blanch_symes', save=
     eqn = eq_kernel(model, geometry, v, p, save=save)
 
     # The source injection term
-    src_term = src.inject(field=p.forward, expr=src * s)
+    src_term = src.inject(field=p.forward, expr=src * dt**2 / m)
```
mloubout, line 490: That factor depends on the PDE, the `dt**2/m` is only valid for 2nd order acoustic WE.

examples/seismic/viscoacoustic/operators.py
```diff
@@ -193,4 +497,57 @@ def ForwardOperator(model, geometry, space_order=4, kernel='blanch_symes', save=
...
+                       npoint=geometry.nsrc)
+    rec = Receiver(name='rec', grid=model.grid, time_range=geometry.time_axis,
+                   npoint=geometry.nrec)
+
+    # Equations kernels
+    eq_kernel = kernels[kernel]
+    eqn = eq_kernel(model, geometry, u, q, forward=False)
+
+    # Construct expression to inject receiver values
+    receivers = rec.inject(field=q.backward, expr=rec * dt**2 / m)
```
mloubout, line 539: 2nd order acoustic weight `dt**2 / m` pde dependent.

examples/seismic/viscoacoustic/wavesolver.py
```diff
@@ -121,3 +132,62 @@ def forward(self, src=None, rec=None, v=None, r=None, p=None, qp=None, b=None,
                                               b=b, vp=vp,
                                               dt=kwargs.pop('dt', self.dt), **kwargs)
         return rec, p, summary
+
+    def adjoint(self, rec, srca=None, u=None, q=None, vp=None, qp=None, b=None, r=None,
```
mloubout, line 136: `v=None`

examples/seismic/viscoacoustic/wavesolver.py
```diff
@@ -121,3 +132,62 @@ def forward(self, src=None, rec=None, v=None, r=None, p=None, qp=None, b=None,
...
+        rec : SparseTimeFunction or array-like
+            The receiver data. Please note that
+            these act as the source term in the adjoint run.
+        srca : SparseTimeFunction or array-like
+            The resulting data for the interpolated at the
+            original source location.
+        v: TimeFunction, optional
+            The computed wavefield.
+        vp : Function or float, optional
+            The time-constant velocity.
```
mloubout, line 153: qp, r, b, q missing from docstring.

What's `q` again, pressure? If pressure would not use `q` usually used for source or attenuation and name it something like `pa`. Would also not name particle velocity `u`, that's confusing with most forward using `u` for forward wavefield, maybe `va`?

examples/seismic/viscoacoustic/wavesolver.py
```diff
@@ -121,3 +132,62 @@ def forward(self, src=None, rec=None, v=None, r=None, p=None, qp=None, b=None,
...
+        # Execute operator and return wavefield and receiver data
+        if self.kernel == 'sls':
+            # Execute operator and return wavefield and receiver data
+            # With Memory variable
+            summary = self.op_adj().apply(srca=srca, rec=rec, q=q, r=r, b=b, vp=vp,
+                                          qp=qp, dt=kwargs.pop('dt', self.dt), **kwargs)
+        else:
+            summary = self.op_adj().apply(srca=srca, rec=rec, q=q, vp=vp, b=b, qp=qp,
+                                          dt=kwargs.pop('dt', self.dt), **kwargs)
+        return srca, q, summary
```
mloubout, line 193: May want to return particle velocity as well?

examples/seismic/viscoacoustic/operators.py
nogueirapeterson: We appreciate your review @mloubout .
These temps were used to avoid the known and described issue about multiple functions on the argument of `grad` operator `Ex: grad((1. + tt) * rho * p)` . 
So we compute each step of the parameters before passing them as arguments.

examples/seismic/viscoacoustic/operators.py
nogueirapeterson: Okay. We will fix it.

examples/seismic/viscoacoustic/operators.py
nogueirapeterson: Okay. We will fix it.

examples/seismic/viscoacoustic/wavesolver.py
nogueirapeterson: okay

examples/seismic/viscoacoustic/wavesolver.py
nogueirapeterson: Okay, we will follow your suggestions.

examples/seismic/viscoacoustic/operators.py
mloubout: Is it still an issue? I think `grad` is used some places on expressions and should support it. If it fails, could you open an issue with an MFE on it?

nogueirapeterson: Yes, until the day I did the PR, this is still an issue.
You fixed it for the `div`, but not for the `grad` yet.
Yes, I can open an issue. I will do this as soon as possible.

mloubout:
```
In [16]: div(u * grad((1+t) * u * v))                                                                                                                                                
Out[16]: Derivative(u(t, x, y)*Derivative(3*u(t, x, y)*v(x, y), x), x) + Derivative(u(t, x, y)*Derivative(3*u(t, x, y)*v(x, y), y), y)

```

Works fine for expressions this is weird.

nogueirapeterson: I see. But when the stencil is passed to the operator the results produce NAN. Anyway, I will check on Monday with the most updated version of Devito, and if confirmed I will open an issue. Otherwise, I will remove the tmps.

mloubout: >passed to the operator the results produce NAN.

Now that's something I can get behind, may need some parametrizable grad/div  to keep it nice and stable (i.e like how you can do `u.dx(x0=x + dx)` but for grad/div. I'll see if I can find time to play around with your PR and file a patch for it but may take a little bit so can leave the implementation like that.

nogueirapeterson: Got it. I will keep implementing it that way.
Thanks for the clarification.
On Monday I will make the suggested corrections.

examples/seismic/viscoacoustic/operators.py
```diff
@@ -63,9 +98,87 @@ def blanch_symes(model, geometry, v, p, **kwargs):
...
+        return [u_v, u_r, u_p]
+
+    else:
+
+        # Auxiliary functions
+        w = VectorTimeFunction(name="w", grid=model.grid, time_order=2,
+                               space_order=space_order)
+        h = TimeFunction(name="h", grid=model.grid, time_order=2, space_order=space_order,
+                         staggered=NODE)
+        g = TimeFunction(name="g", grid=model.grid, time_order=2, space_order=space_order,
```
mloubout, line 158: Since these are local tmps, you may only need `VectorFunction, Function` and not time dependent, same other places

speglich:
Hi! We updated the PR removing unnecessary/duplicated functions and increase the tolerance to viscoacoustic tests.

examples/seismic/viscoacoustic/operators.py
speglich: Hi @mloubout, we understood your review and we tried this approach. When we replace `TimeFunction` by `Function` we are experiencing seg fault with the operator 😭.

mloubout: That's quite odd :O

examples/seismic/viscoacoustic/operators.py
```diff
@@ -367,13 +363,14 @@ def deng_2nd_order(model, geometry, p, **kwargs):
     # Auxiliary Function
     l = VectorFunction(name="l", grid=model.grid, space_order=space_order)
 
+    h = TimeFunction(name="h", grid=model.grid, time_order=2, space_order=space_order,
```
nogueirapeterson, line 366: @FabioLuporini  these auxiliary functions are temporaries.
Some of them are like `TimeFunction` / `VectorTimeFunction`, however it would not be necessary. Just `Function` / `VectorFunction` should be enough. But the last one generates a segmentation fault.

FabioLuporini:
CI is currently failing because of issue #1459 . I've assigned it to me

Leitevmd:
I've opened #1483 which I think is the only problem right now (after the last commit) for this PR. I've have already proposed a solution. Just waiting for approval to submit a new PR.

FabioLuporini:
I've just merged master into this branch since the random OSX failure should have now disappeared. Let's see

nogueirapeterson:
thanks, @FabioLuporini.  Now disappeared the OSX failure. All tests are green!

mloubout (commented):
Really nice, glad all working without temporaries now. Did you check the result look ok? May need to use the shift in div/grad i.e `div(grad(expr, shift=.5), shift=-.5)` to get nice results

examples/seismic/viscoacoustic/operators.py
```diff
@@ -1,35 +1,71 @@
...
+    # Source symbol with input wavelet
+    src = PointSource(name="src", grid=model.grid, time_range=geometry.time_axis,
+                      npoint=geometry.nsrc)
+    rec = Receiver(name='rec', grid=model.grid, time_range=geometry.time_axis,
+                   npoint=geometry.nrec)
+
+    forward = kwargs.get('forward', True)
+    time_order = p.time_order
+
+    if forward:
```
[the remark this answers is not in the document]
FabioLuporini: > glad all working without temporaries now. 

can we assert the expected wavefield norms in the notebooks?

tests/test_adjoint.py
```diff
@@ -42,6 +44,30 @@ class TestAdjoint(object):
         # 3D TTI tests with varying space orders
         ('layers-tti', (30, 35, 40), 'centered', 8, tti_setup),
         ('layers-tti', (30, 35, 40), 'centered', 4, tti_setup),
+        # 2D SLS Viscoacoustic tests with varying space orders
```
FabioLuporini, line 47: @mloubout @rhodrin now that I note this... that's a lot of tests added! Do you think we actually need all of them?

examples/seismic/viscoacoustic/operators.py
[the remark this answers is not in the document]
FabioLuporini: just noted that we're already testing via `test_viscoacoustic`, so we should be good

nogueirapeterson:
> Really nice, glad all working without temporaries now. Did you check the result look ok? May need to use the shift in div/grad i.e `div(grad(expr, shift=.5), shift=-.5)` to get nice results

After removing the tmps, I didn't check the results. But all the tests worked well (stability, amplitude, and adjoint).
Anyway, I'm going to check for security.

nogueirapeterson:
> > Really nice, glad all working without temporaries now. Did you check the result look ok? May need to use the shift in div/grad i.e `div(grad(expr, shift=.5), shift=-.5)` to get nice results
> 
> After removing the tmps, I didn't check the results. But all the tests worked well (stability, amplitude, and adjoint).
> Anyway, I'm going to check for security.

@mloubout 
@FabioLuporini 
We did three experiments using the 2nd acoustic wave equation. First without any modification, with temporaries, after without temporaries, and finally, using the shift parameter:
shots

Attached with this comment, we leave a script explaining how we did these experiments.
[test_wo_tmps.txt](https://github.com/devitocodes/devito/files/5584273/test_wo_tmps.txt)

FabioLuporini:
@nogueirapeterson if your tests were still passing despite the images being "wrong" or "flawed" then it _also_ means that the provided tests are not sufficient. Does it make sense @mloubout or am I missing something?

nogueirapeterson:
> @nogueirapeterson if your tests were still passing despite the images being "wrong" or "flawed" then it _also_ means that the provided tests are not sufficient. Does it make sense @mloubout or am I missing something?

The existing tests analyze stability, amplitude, and the adjoint operation. Removing the temps did not directly affect these tests. 
However, it is producing a noise that I guess is related to the derivatives on a staggered grid.

nogueirapeterson:
@FabioLuporini and @mloubout We are waiting for PR #1551, once it is approved, possibly this PR will be completed.

FabioLuporini (approved):
Great, all good now

rhodrin (approved):

PR #1409, opened 2020-07-28
Tweak CIRE for GPUs

FabioLuporini:
The new `cire-maxpar` option is to be used to tradeoff storage for parallelism.

Defaults: true on GPU backends, false on CPU backends.

when on, generates:
```
for x
  for y
    r[x,y] = ...
for x
  for y
    ... = ... r[x, y+1] ...
```

instead of
```
for x
  for y
    r[y] = ...
  for y 
    ... = ... r[y+1] ...
```

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -58,32 +58,34 @@ jobs:
           dockerfile: ./docker/Dockerfile.nvidia
           tags: gpu-latest,'gpu-${{ github.event.release.tag_name }}'
 
-  test-cpu-image:
-    needs: deploy-docker
-    runs-on: ubuntu-latest
+  # Comment out until next release
+  #test-cpu-image:
+    #needs: deploy-docker
+    #runs-on: ubuntu-latest
 
-    strategy:
-      fail-fast: false
-      matrix: 
-        im-name: [cpu-dev, cpu-latest]
+    #strategy:
+      #fail-fast: false
+      #matrix:
+        #im-name: [cpu-dev, cpu-latest]
 
-    steps:
-        - name: Run simple test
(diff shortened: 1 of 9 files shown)


FabioLuporini:
> The code example seems to be quite similar to what min-storage does though (is it compatible with it too actually ? ).

Say you have mixed temporaries, e.g., some are 1D, some 2D, and some 3D. By default all of them get promoted to 3D temporaries to maximize data locality (heuristic). With `min-storage` 1D temps remain 1D, 2D temps remain 2D, and 3D temps remain 3D. Loops are shared where possible.

`cire-maxpar` doesn't care about the dimensionality of temporaries; instead, it enforces storage along each parallel dimension **and** it schedules such tensor-expanded temporaries into a separate loop nest. Now, these are two separate things, though the former is a prerequisite for the latter.

It is true that in some cases `min-storage` and `cire-maxpar` may lead to similar code, but they are conceived for different cases (with `min-storage` you shrink the working set, with `cire-maxpar` you may end up increasing it). I haven't tried them in combination but I'm 99% sure that `min-storage` will just take precedence leaving `cire-maxpar` with nothing to do

.github/workflows/docker-publish.yml
```diff
@@ -58,32 +58,34 @@ jobs:
           dockerfile: ./docker/Dockerfile.nvidia
           tags: gpu-latest,'gpu-${{ github.event.release.tag_name }}'
 
-  test-cpu-image:
-    needs: deploy-docker
-    runs-on: ubuntu-latest
+  # Comment out until next release
+  #test-cpu-image:
```
mloubout, line 62: Wait no we can keep this one for `cpu-dev`

PR #1411, opened 2020-07-29
Restrict pytest to < 6.0

rhodrin:
With the release of pytest 6.0.0 a previous warning seems to now be an error. See [here](https://github.com/computationalmodelling/nbval/issues/139) who have also noted this issue. Latest pytest issues [here](https://github.com/pytest-dev/pytest/issues).


rhodrin:
Geez, `MPI` tests hanging. Will investigate either a bit later or more likely in the morning now.

PR #1251, opened 2020-04-24
Add MPI+GPU runs to CI

georgebisbas:
This PR enhances testing on GPUs
- adds MPI+openacc runs for elastic, visco-{acoustic,elastic}

PR #1413, opened 2020-07-29
Single precision interpolation

### LINKED ISSUE #1357: MPI with SparseFunction on the mpi domain boundary
There seem to be an issue when the source is on or near the MPI halo in some cases.

MFE  to be made.

mloubout:
Fixes interpolation using the wrong points with MPI (or without MPI and no BCs) due to rounding errors around zero in single precision.

- Added test for it

Fixes #1357 

Should fix #1251

### DIFF
--- a/devito/types/sparse.py
+++ b/devito/types/sparse.py
@@ -538,7 +538,15 @@ class SparseFunction(AbstractSparseFunction):
 
     @cached_property
     def _position_map(self):
-        """Symbol for coordinate value in each dimension of the point."""
+        """
+        Symbols map for the postition of the sparse point relative to the
+        origin of the grid.
+        When computing `(coord - origin)/spacing`, it may be expanded as
+        `coord/spacing - origin/spacing` and round up with a +-1 error as an
+        interger therefor these position are cmputed individally first.
+        This also reduce to 1 the number of time this position is computed
+        instead of 6 time in the different interpolation steps.
+        """
         pos = tuple(Scalar(name='pos%s' % d, dtype=self.dtype)
                     for d in self.grid.dimensions)
         return {c - o: p for p, c, o in zip(pos, self._coordinate_symbols,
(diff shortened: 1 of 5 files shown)


devito/types/sparse.py
```diff
@@ -535,6 +536,27 @@ def _point_symbols(self):
...
+        """
+        Symbols map between grid origin and sparse point position
+        along each SpaceDimension.
+
+        Notes
+        -----
+        The expression `(coord - origin)/spacing` could also be computed in the
+        mathematically equivalent expanded form `coord/spacing -
+        origin/spacing`. This particular form is problematic when a sparse
+        point is in proximity of the grid origin, since due to a larger machine
```
rhodrin, line 550: "...is in close proximity..."

rhodrin:
Looks good thanks. Are there any more tests we should add?

Need to think about a couple of issues (e.g. how this will work with the additional shift with `Function`'s on `SubDomain`'s) but can do that at a later date (when the books sprint is over).

devito/types/sparse.py
```diff
@@ -535,6 +536,27 @@ def _point_symbols(self):
         return tuple(Scalar(name='p%s' % d, dtype=self.dtype)
                      for d in self.grid.dimensions)
 
+    @cached_property
+    def _position_map(self):
+        """
+        Symbols map between grid origin and sparse point position
```
rhodrin, line 542: "Symbolic map between the grid origin and sparse point position(s)..."?

mloubout: I prefer what I had, this sounds like it maps `{ origin : coordinate}` which it's something else.

devito/types/sparse.py
FabioLuporini: done

devito/types/sparse.py
FabioLuporini: reverted

FabioLuporini:
> Looks good thanks. Are there any more tests we should add?

not sure tbh?

PR #1418, opened 2020-08-04
ci: Test all critical seismic examples with MPI

FabioLuporini:


.github/workflows/pytest-core-mpi.yml
```diff
@@ -39,15 +39,23 @@ jobs:
...
+    - name: Test notebooks
       run : |
         # Currently not tested due to issue #859
         # ipcluster start --profile=mpi -n 4 --daemon
         # py.test --nbval examples/mpi
         # ipcluster stop --profile=mpi
 
-    - name: Test mpi fwi
-      run: mpirun -n 4 python examples/seismic/inversion/fwi.py
+    - name: Test seismic examples
```
rhodrin, line 49: We should make this (and the fwi example) a separate job so that it 'parallelizes'. We can then always add an additional runner is required.

.github/workflows/pytest-core-mpi.yml
```diff
@@ -39,15 +39,23 @@ jobs:
...
+      run: |
+        DEVITO_MPI=1 mpirun -n 4 py.test examples/seismic/tti/tti_example.py
+        DEVITO_MPI=1 mpirun -n 4 py.test examples/seismic/elastic/elastic_example.py
+        DEVITO_MPI=1 mpirun -n 4 py.test examples/seismic/viscoacoustic/viscoacoustic_example.py
+        DEVITO_MPI=1 mpirun -n 4 py.test examples/seismic/viscoelastic/viscoelastic_example.py
```
mloubout, line 54: why not just put `DEVITO_MPI=1` in the `env` since it's used everywhere?

.github/workflows/pytest-core-mpi.yml
```diff
@@ -17,11 +17,12 @@ on:
 jobs:
   build:
     name: pytest-mpi
-    runs-on: self-hosted
+    runs-on: [self-hosted, mpi]
 
     env:
-      DEVITO_ARCH: "gcc-7"
+      DEVITO_MPI: "1"
```
mloubout, line 23: Oh yeah no this need to go, pytest expect `DEVITO_MPI=0` and then runs it internally

PR #1421, opened 2020-08-05
Fix issue #1298

### LINKED ISSUE #1298: A function that appears in a ConditionalDimension, but not in the Operator input is not always being added to the parameters of the Operator.
A function that appears in a ConditionalDimension, but not in the Operator input is not always beeing added to the parameters of the Operator. May need to run a few times to reproduce.

**MFE**:
```
import numpy as np
from devito import ConditionalDimension, Function, Eq, Operator, Dimension
from devito.types import Lt

shape = (8, 8)
x = Dimension(name='x')
y = Dimension(name='y')
g = Function(name='g', shape=shape, dimensions=(x, y), dtype=np.float32)
cond = Lt(g, 2)
ci = ConditionalDimension(name='ci', parent=y, condition=cond)
f = Function(name='f', shape=shape, dimensions=(x, ci))
eq1 = Eq(f, 5)
op = Operator([eq1])
print(op.ccode)
import pdb; pdb.set_trace()
op.apply()
```
**Error log:**
```
  File "tests/mfe_args.py", line 23, in <module>
    op.apply()
  File "/home/gbisbas/workspace/devito/devito/operator/operator.py", line 655, in apply
  File "/home/gbisbas/workspace/devito/devito/operator/operator.py", line 537, in arguments
  File "/home/gbisbas/workspace/devito/devito/operator/operator.py", line 482, in _prepare_arguments
    p._arg_check(args, self._dspace[p])
AttributeError: 'Symbol' object has no attribute '_arg_check'
```

**Hook from @FabioLuporini :**

I think I understand the issue above
we’re lacking a small piece of code that searches cd.condition
https://github.com/devitocodes/devito/blob/master/devito/ir/support/utils.py#L144
```
devito/ir/support/utils.py:144
        for j in candidates:
```
among candidates there could a ConditionalDimension. That must be searched inside its condition
actually we can make it even simpler a (lowered) expression knows its conditionals: https://github.com/devitocodes/devito/blob/master/devito/ir/equations/equation.py#L159
```
devito/ir/equations/equation.py:159
        expr._conditionals = tuple(conditionals)
```
so I think we just need to extend the roots here https://github.com/devitocodes/devito/blob/master/devito/ir/support/utils.py#L131 with the conditionals (if any, clearly; as you know, very often there are none)
```
devito/ir/support/utils.py:131
            roots.extend(list(i.lhs.indices))
```
try that, in a separate, self-contained PR, which includes patch + MFE. Thanks!

FabioLuporini:
fixes #1298

### DIFF
--- a/devito/ir/equations/equation.py
+++ b/devito/ir/equations/equation.py
@@ -1,11 +1,12 @@
 from cached_property import cached_property
 import sympy
 
-from devito.ir.equations.algorithms import dimension_sort
+from devito.ir.equations.algorithms import dimension_sort, lower_exprs
 from devito.finite_differences.differentiable import diff2sympy
 from devito.ir.support import (IterationSpace, DataSpace, Interval, IntervalGroup,
                                Stencil, detect_accesses, detect_oobs, detect_io,
                                build_intervals, build_iterators)
+from devito.symbolics import CondEq
 from devito.tools import Pickable, frozendict
 from devito.types import Eq
 
@@ -56,18 +57,6 @@ class IREq(object):
     @cached_property
     def conditionals(self):
         return self._conditionals or frozendict()
-
-    @cached_property
-    def conditions(self):
-        ret = []
-        for d, v in self.conditionals.items():
-            if d.is_Conditional:
-                if d.condition is not None:
-                    ret.append(d.condition)
-            else:
-                # Already lowered
-                ret.append(v)
-        return tuple(ret)
 
     @property
     def directions(self):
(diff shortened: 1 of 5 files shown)


devito/ir/equations/equation.py
```diff
@@ -157,7 +167,7 @@ def __new__(cls, *args, **kwargs):
 
         expr._dspace = dspace
         expr._ispace = ispace
-        expr._conditionals = frozendict([(d, ()) for d in conditionals])
+        expr._conditionals = conditionals
```
mloubout, line 170: Isn't that the revert of yesterday? 🤣

FabioLuporini: the isolated line yeah... but in practice no! [these lines](https://github.com/devitocodes/devito/pull/1421/files#diff-14cbfe70a6d744024416f98cb417dcbaR153-R161) :grimacing:

PR #1425, opened 2020-08-07
examples: Fix `opt` arg-passing for elastic and TTI

georgebisbas:
Fixes the `opt` argument-passing to TTI and elastic.

### DIFF
--- a/examples/seismic/tti/tti_example.py
+++ b/examples/seismic/tti/tti_example.py
@@ -8,7 +8,8 @@ from examples.seismic.tti import AnisotropicWaveSolver
 
 
 def tti_setup(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
-              space_order=4, nbl=10, preset='layers-tti', **kwargs):
+              kernel='centered', space_order=4, nbl=10, preset='layers-tti',
+              **kwargs):
 
     # Two layer model for true velocity
     model = demo_model(preset, shape=shape, spacing=spacing,
@@ -17,7 +18,7 @@ def tti_setup(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
     # Source and receiver geometries
     geometry = setup_geometry(model, tn)
 
-    return AnisotropicWaveSolver(model, geometry, space_order=space_order)
+    return AnisotropicWaveSolver(model, geometry, space_order=space_order, **kwargs)
 
 
 def run(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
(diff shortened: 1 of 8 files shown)


examples/seismic/tti/tti_example.py
```diff
@@ -17,7 +17,7 @@ def tti_setup(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
     # Source and receiver geometries
     geometry = setup_geometry(model, tn)
 
-    return AnisotropicWaveSolver(model, geometry, space_order=space_order)
+    return AnisotropicWaveSolver(model, geometry, space_order=space_order, **kwargs)
```
mloubout, line 20: add kernel to the signature to avoid it being in the kwargs.

examples/seismic/tti/wavesolver.py
```diff
@@ -52,7 +52,7 @@ def op_fwd(self, kernel='centered', save=False):
         """Cached operator for forward runs with buffered wavefield"""
         return ForwardOperator(self.model, save=save, geometry=self.geometry,
                                space_order=self.space_order,
-                               kernel=kernel, **self._kwargs)
+                               **self._kwargs)
```
georgebisbas, line 55: NOTE FR: This kernel arg causes memoization error if I keep it.
NOt sure how to solve it.

examples/seismic/tti/tti_example.py
```diff
@@ -17,15 +18,16 @@ def tti_setup(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
     # Source and receiver geometries
     geometry = setup_geometry(model, tn)
 
-    return AnisotropicWaveSolver(model, geometry, space_order=space_order)
+    return AnisotropicWaveSolver(model, geometry, space_order=space_order,
+                                 kernel=kernel, **kwargs)
```
mloubout, line 22: It's not used in the init of the solver so no kernel shouldn't be there

examples/seismic/tti/wavesolver.py
mloubout: WDYM memoization error, please put it back so can be debugged it should be there (and in adjoint as well)

examples/seismic/tti/tti_example.py
georgebisbas: I thought your previous comment meant that I should add it here (?).
Which one is the signature?

mloubout: I meant need to be in the signature of `tti_setup` so that doesn't end up in kwargs

FabioLuporini (commented):
@mloubout what's missing here?

rhodrin (approved):

PR #1429, opened 2020-08-11
fix misspelling of openmp-targets for amdgcn-amd-amdhsa

paklui:
Fix a typo when trying to run devito openmp-target for amdgcn-amd-amdhsa with AOMP:
error: unable to create target: 'No available targets are compatible with triple "amdgcn-amd-amdhs"'


FabioLuporini:
Uncontroversial, thanks!

PR #1430, opened 2020-08-12
Potentially smarter CIRE via cire-rotate option

FabioLuporini:
By default disabled as AFAICT it needs (block shape) tuning to deliver a performance improvement

Also needs more experimentation by more people

Early local results are encouraging though

The PR, in general, does improve the code quality in various places. Tests shipped.


devito/passes/clusters/aliases.py
```diff
@@ -418,10 +449,198 @@ def choose(exprs, aliases, selector):
     return chosen, others
 
 
-def process(cluster, chosen, aliases, max_par, sregistry, platform):
+def make_schedule(cluster, aliases, in_writeto, options):
```
FabioLuporini, line 452: note for reviewers: this is _not_ an entirely new function -- this stems from the former `def iter` (now deleted, see below) though it's been changed a little

devito/ir/clusters/cluster.py
```diff
@@ -163,14 +163,44 @@ def is_scalar(self):
...
+    @cached_property
+    def grid(self):
+        if len(self.grids) == 1:
+            return self.grids[0]
+        raise ValueError("Cluster has no unique Grid")
+
+    @cached_property
+    def grids(self):
+        """
+        The Grid's over which the Cluster is defined.
```
georgebisbas, line 200: Grids' ?

devito/passes/clusters/aliases.py
```diff
@@ -97,19 +105,21 @@ def cire(cluster, mode, sregistry, options, platform):
...
 
         # Rule out aliasing expressions with a bad flops/memory trade-off
         chosen, others = choose(exprs, aliases, selector)
         if not chosen:
             # Do not waste time
             continue
 
-        # Create Aliases and assign them to Clusters
-        clusters, subs = process(cluster, chosen, aliases, max_par, sregistry, platform)
+        # AliasMapper -> Schedule -> [Clusters]
```
georgebisbas, line 116: Maybe fuse with old comment?

devito/passes/clusters/misc.py
```diff
@@ -53,17 +53,29 @@ def callback(self, clusters, prefix):
...
-                processed.append(c)
-                continue
-
-            # Contract iteration and data spaces for the lifted Cluster
+            # The contracted iteration and data spaces
             key = lambda d: d not in hope_invariant
             ispace = c.ispace.project(key).reset()
             dspace = c.dspace.project(key).reset()
 
+            # All of the inner Dimensions must appear in the write-to region
```
georgebisbas, line 61: ok, this is helpful

georgebisbas (commented):
> Also needs more experimentation by more people

Are there any particular examples we can try on several platforms?

FabioLuporini:
@georgebisbas

> Are there any particular examples we can try on several platforms?

any code exploiting CIRE would work, so TTI for example. Just add `cire-rotate: True` to the Operator and it will do the trick. Instructions in the updated performance notebook. Thanks for looking into this

FabioLuporini:
@georgebisbas

> Are there any particular examples we can try on several platforms?

any code exploiting CIRE would work, so TTI for example. Just add `cire-rotate: True` to the Operator and it will do the trick. Instructions in the updated performance notebook. Thanks for looking into this

FabioLuporini:
@georgebisbas

> Are there any particular examples we can try on several platforms?

any code exploiting CIRE would work, so TTI for example. Just add `cire-rotate: True` to the Operator and it will do the trick. Instructions in the updated performance notebook. Thanks for looking into this

devito/ir/clusters/cluster.py
FabioLuporini: I've been told that one should use an apostrophe for the plural of proper nouns... :grimacing:

devito/passes/clusters/misc.py
FabioLuporini: yeah, this is fixing a bug with lifting!

devito/passes/clusters/aliases.py
FabioLuporini: I think this one is fine. What's troubling you? basically it's describing the lowering from AliasMapper to a list of Clusters

devito/ir/support/space.py
```diff
@@ -148,9 +155,49 @@ def _rebuild(self):
 
     @cached_property
     def size(self):
-        upper_extreme = self.dim.symbolic_max + self.upper
-        lower_extreme = self.dim.symbolic_min + self.lower
-        return (upper_extreme - lower_extreme + 1) / self.dim.symbolic_incr
+        # The Interval size is typically a function of several symbols (e.g.,
```
mloubout, line 158: why a comment instead of docstring?

devito/ir/iet/utils.py
```diff
@@ -92,31 +92,33 @@ def compose_nodes(nodes, retrieve=False):
...
-def derive_parameters(nodes, drop_locals=False):
+def derive_parameters(iet, drop_locals=False):
     """
     Derive all input parameters (function call arguments) from an IET
     by collecting all symbols not defined in the tree itself.
     """
     # Pick all free symbols and symbolic functions from the kernel
-    functions = FindSymbols('symbolics').visit(nodes)
-    free_symbols = FindSymbols('free-symbols').visit(nodes)
+    functions = FindSymbols('symbolics').visit(iet)
```
georgebisbas, line 101: wouldn't something better be `sym_funcs = Find...(`sym_funcs`)....
Although we search for symbolic functions...
seems awkward to use 'symbolics' as a search key and assigning the returned stuff to funtions..
I know it has nothing to do with this PR per se... just some nitpicking personal view....

devito/ir/support/space.py
```diff
@@ -148,9 +155,49 @@ def _rebuild(self):
...
+            assert is_integer(d.symbolic_incr)
+            assert is_integer(self.dim.symbolic_max)
+            assert self.lower == self.upper == 0
+            npoints = self.dim.symbolic_max - d.symbolic_min + 1
+            # Iterations [1, ..., n-1]:
+            assert d.symbolic_incr == self.dim.symbolic_max
+            npoints += 1 * (n-1)
+            npoints /= n
+        else:
+            # Typicall we end up here (Dimension, SubDimension, IncrDimension)
```
georgebisbas, line 187: Typically

mloubout:
As a general comment. it would be good to check that all these opt mode do not break any of the test/examples. Could for example pass a random combination of it to any of the tests (like adjoint/....) so that we know still pass with it.

devito/passes/clusters/aliases.py
georgebisbas: hmm.. I felt like the verbs Create and assign to Clusters was better...but np..

devito/passes/clusters/aliases.py
```diff
@@ -97,22 +107,29 @@ def cire(cluster, mode, sregistry, options, platform):
...
+        # The [Clusters] must be ordered so as to reuse as many of the `cluster`'s
+        # IterationIntervals as possible in order to honor the write-to region. This
+        # also guarantees that fusion is maximum
+        processed.extend(clusters)
+        processed.sort(key=partial(cit, cluster))
 
-        # Rebuild `cluster` so as to use the newly created Aliases
-        rebuilt = rebuild(cluster, others, aliases, subs)
+        # Rebuild `cluster` so as to use the newly created aliases
+        rebuilt = rebuild(cluster, others, schedule, subs)
```
georgebisbas, line 130: Isn;t it better to use a cluster-like name rather than past tense of rebuild...?
Forgive if I forgot the art of immutability but can't we just use `cluster = rebuild(cluster,......)`

devito/ir/support/space.py
FabioLuporini: fair point. Class docstrings now refactored.

devito/ir/support/space.py
FabioLuporini: fixed

devito/passes/clusters/aliases.py
FabioLuporini: yes, I think we can use `cluster = ...` . Fixing

FabioLuporini:
@mloubout 

> As a general comment. it would be good to check that all these opt mode do not break any of the test/examples. Could for example pass a random combination of it to any of the tests (like adjoint/....) so that we know still pass with it.

This is a good point. I tried my best to test virtually all possible combinations. There are 800 line changes in test_dse.py; I've basically parametrized all of the CIRE tests to experiment with and without `cire-rotate`. We could add an option to also test TTI via the examples, but I can't pass a performance option (cire-rotate) from the command line yet, so... `TestAliases`, today, covers all of the critical cases (ie it also tests operators akin to what we have in our own TTI and TTi adjoint)

georgebisbas (commented):
Are we sure we like having `indices` and `indicess`. Seems to be a trap for future problems.....from a simple grep up to more complicated stuff....same for `aliased` and `aliaseds`

devito/ir/support/space.py
```diff
@@ -148,9 +150,57 @@ def _rebuild(self):
...
+
+            # Iteration 0:
+            assert is_integer(d.symbolic_min)
+            assert is_integer(d.symbolic_incr)
+            assert is_integer(self.dim.symbolic_max)
+            assert self.lower == self.upper == 0
+            npoints = self.dim.symbolic_max - d.symbolic_min + 1
+            # Iterations [1, ..., n-1]:
+            assert d.symbolic_incr == self.dim.symbolic_max
+            npoints += 1 * (n-1)
```
mloubout, line 187: `1 *` ?

tests/test_dse.py
```diff
@@ -675,56 +718,58 @@ def test_min_storage_in_isolation(self):
...
         try:
             Operator(eqn, opt=('advanced-fsg', {'openmp': True, 'min-storage': True}))
         except InvalidOperator:
-            # Incompatible `advanced-fsg` + `min-storage`
             assert True
         except:
             assert False
 
-        u1.data_with_halo[:] = 1.42
+        # Check that `cire-rotate=True` has no effect in this code has there's
```
georgebisbas, line 739: has -> as

devito/ir/support/space.py
```diff
@@ -148,9 +150,57 @@ def _rebuild(self):
...
+            # Special case 1)
+            # May be caused by the performance option `cire-rotate=True`
+            d = self.dim.symbolic_min
+            assert d.is_Modulo
+            n = d.parent.symbolic_size
+
+            # Iteration 0:
+            assert is_integer(d.symbolic_min)
+            assert is_integer(d.symbolic_incr)
+            assert is_integer(self.dim.symbolic_max)
```
mloubout, line 182: Shouldn't these checks be in `CustomDimension` instead?

tests/test_dse.py
```diff
@@ -1223,81 +1248,87 @@ def test_full_shape_big_temporaries(self):
...
+        op1(time_M=1, u=u1)
+        assert np.all(u.data == u1.data)
 
+    @pytest.mark.parametrize('rotate', [False, True])
     @switchconfig(profiling='advanced')
-    def test_extraction_from_lifted_ispace(self):
+    def test_extraction_from_lifted_ispace(self, rotate):
         """
         Test that the aliases are scheduled correctly when extracted from
         Clusters whose iteration space is lifted (ie, stamp != 0).
```
georgebisbas, line 1294: stamp?

devito/ir/support/space.py
```diff
@@ -148,9 +150,57 @@ def _rebuild(self):
...
+        `self.dim.symbolic_max`), and all such symbols must be mappable
+        to actual numbers at `op.apply` time (i.e., the runtime values).
+        When `self.dim` is an "unstructured Dimension", such as ModuloDimension,
+        things can get nasty since the symbolic min/max/incr can literally be
+        anything (any expression involving any Dimension/symbol/...), which makes
+        it extremely complicated to numerically compute the size. However,
+        the compiler only uses such unstructured Dimensions in well defined
+        circumstances, which we explicitly handle here. Ultimately, therefore,
+        we return a `size` that is made up of known symbols.
+        """
```
mloubout, line 171: So no case where we can end up here wit ha `Modulo/is_Conditional Dimension` right? WHat if one has a Subdimension of a modulo dimension?

devito/ir/support/utils.py
```diff
@@ -80,14 +80,16 @@ def build_iterators(mapper):
     for k, v in mapper.items():
         for d, offs in v.items():
             if d.is_Stepping:
-                sub_iterators = iterators.setdefault(d.parent, set())
-                sub_iterators.update({ModuloDimension(d, i, k._time_size)
-                                      for i in offs})
+                values = iterators.setdefault(d.parent, [])
+                for i in sorted(offs):
```
mloubout, line 84: So what does this change? Like comapred to #1438

examples/performance/00_overview.ipynb
```diff
@@ -1369,7 +1369,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "#### All optimizations together\n",
+    "#### Impact of CIRE in the `advanced` mode\n",
     "\n",
     "The `advanced` mode triggers all of the passes we've seen so far... and in fact, many more! Some of them, however, aren't visible in our running example (e.g., all of the MPI-related optimizations). These will be treated in a future notebook.\n",
```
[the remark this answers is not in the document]
FabioLuporini: treated is OK

devito/ir/support/space.py
FabioLuporini: > So no case where we can end up here wit ha Modulo/is_Conditional Dimension right? 

correct. The possible cases are constrained by our user API ie IOW by what the user can write

> WHat if one has a Subdimension of a modulo dimension?

ModuloDimension is not part of our user API so we're good on this front ...

... but part of our API are SteppingDimension and SubDimension, and yeah, one could compose them and do nasty things, but I'm not even sure we would be able to generate anything functional, so...

at the end of the day, I'm just _enhancing_ this method, not dropping stuff, so at worst I'm extending coverage

devito/ir/support/space.py
FabioLuporini: no, `CustomDimension` (as the name suggests) is to be used for custom ("unstructured") iteration spaces. **So far**, it's used only in a very specific way, hence the asserts here. Should one start using it differently, then the asserts here would catch it, suggesting an extension of the method is necessary. But here's the right place of the asserts, otherwise I wouldn't have created a generic `CustomDimension`, but rather something more specific

devito/ir/support/utils.py
FabioLuporini: I was reordering the `uindices` (sub-iterators) for deterministic code generation too far in the food chain:

https://github.com/devitocodes/devito/pull/1430/files#diff-774365b68457b07c3690671a8c99ae00L33

I can't reorder them there because now you can have

```
xs = 0, xsi = f(xs, ...)
```

and you can't swap `xsi` with `xs` otherwise `xs`, used to define `xsi`, would be undefined

but I still want a unique ordering of stepping iterators, which I'm enforcing here (where it should have been in the first place after all)

devito/ir/support/space.py
FabioLuporini: the line before ensures:

```
assert d.symbolic_incr == self.dim.symbolic_max
```

so if the loop is entered `n` times (https://github.com/devitocodes/devito/pull/1430/files#diff-c28c75af048eb0f18f5cfba78eb8a88cR177) then for `n-1` times only `1` iteration will be executed

mloubout: this is fine, the one just looks odd `+= n-1` would look nicer

devito/ir/support/space.py
mloubout: ok

devito/ir/support/utils.py
mloubout: But does it break #1438? I.e does it prevent to have same order with and without modulo for the standard time case.

FabioLuporini: no, AFAICT in 1438 you're just adding a `key` to `sorted`, so you can rebase that on top of this one once it lands, and just by adding the key back we'll be fine

georgebisbas (approved):

devito/ir/support/utils.py
mloubout: Ok good.

PR #1433, opened 2020-08-17
OPENMP: fix reduction clause for array

mloubout:
The reduction clause for arrays is currently incorrect 

```
#pragma omp for collapse(3) schedule(static,1) reduction(+:B[l])
```

while openmp except actually indices/range of the array `reduction(+:B[start:end])`

This patch fixes it but still needs:
- handling the halo (currently assume `dim_m:dim_M+1`)
- tests


tests/test_dle.py
```diff
@@ -523,6 +523,33 @@ def test_scheduling(self):
         assert not iterations[3].is_Affine
         assert 'schedule(dynamic,chunk_size)' in iterations[3].pragmas[0].value
 
+    @pytest.mark.parametrize('so', [0, 1, 2])
+    @pytest.mark.parametrize('dim', [1, 2])
+    def test_array_reduction(self, so, dim):
+        grid = Grid(shape=(3, 3, 3))
+        d = grid.dimensions[dim]
+
+        f = Function(name='f', shape=(3,), dimensions=(d,), grid=grid, space_order=so)
```
georgebisbas, line 532: I would add a parameter for shape sizes and would use irregular shapes.
`nx, ny, nz = 3, 6, 3`
f= ......shape=(shape[dim])..

mloubout: Not sure why the size would make any difference.

rhodrin:
> This patch fixes it but still needs to figure out why in some case the reduction is not detected (like when using x instead of z in the new test)

This sounds concerning?

Could we also add a test with MPI + openmp?

devito/passes/iet/openmp.py
```diff
@@ -135,7 +135,20 @@ def _make_clauses(cls, ncollapse=None, chunk_size=None, nthreads=None,
             clauses.append('num_threads(%s)' % nthreads)
 
         if reduction:
-            args = ','.join(str(i) for i in reduction)
+            reds = []
+            for i in reduction:
+                # OMP expects a size not an index as input of reduction, such as
```
FabioLuporini, line 140: this piece of code really looks complicated... Are we totally sure it's really necessary? could you point me to the specs where it states it?

mloubout: Haven't found the exact spec, only went from tutos on openmp 4.5. But 99% sure it's necessary.

tests/test_dle.py
```diff
@@ -523,6 +523,33 @@ def test_scheduling(self):
...
+                                                         d.symbolic_max, so+1)
+                    in iterations[1].pragmas[0].value)
+        else:
+            assert ("reduction(+:f[%s:%s + 1])" % (d.symbolic_min, d.symbolic_max)
+                    in iterations[1].pragmas[0].value)
+
+        try:
+            op(time_M=1)
+            assert np.allclose(f.data, 18)
+        except:
```
FabioLuporini, line 548: this is flawed, because the assert failing would make you fall into the except branch

devito/passes/iet/openmp.py
FabioLuporini: fine, I guess I can just try running the test and see how it fails without this patch

devito/passes/iet/openmp.py
```diff
@@ -135,7 +135,20 @@ def _make_clauses(cls, ncollapse=None, chunk_size=None, nthreads=None,
...
+                # reduction(+:f[1:3])
+                if i.is_Indexed:
+                    bounds = [(k,) if k.is_Number else
```
FabioLuporini, line 143: I don't think we need this `if k.is_Number` thing ...

```
[(k.subs(d, d.symbolic_min), k.subs(d, d.symbolic_max + 1)) for k, d, ...]
```

should suffice?

devito/passes/iet/openmp.py
```diff
@@ -135,7 +135,20 @@ def _make_clauses(cls, ncollapse=None, chunk_size=None, nthreads=None,
...
+                              (k.subs(d, d.symbolic_min), k.subs(d, d.symbolic_max + 1))
+                              for k, d in zip(i.indices, i.function.dimensions)]
+                    bounds = ','.join(('%s:%s' % v if len(v) > 1 else str(v[0]))
```
FabioLuporini, line 146: sorry I'm a bit slow this morning, what does this do?

devito/passes/iet/openmp.py
mloubout: May need extra `sympyify` to have it to work like that but yes.

devito/passes/iet/openmp.py
mloubout: This does `[start:end]` if there is more than one index or just `[index]` if just a single one

tests/test_dle.py
mloubout: Hum indeed, just `op(time_M=1)` should be in it and assert below it

devito/passes/iet/openmp.py
mloubout: Atually nom you need the `if k.is_Number` or you'd end up with `(k, k)` for the single index case instead of just `(k,)`

mloubout:
Included in #1446

PR #1434, opened 2020-08-19
Examples: Add tests and damp safety checks for backward compatibility

mloubout:
- Add tests for geometry/model
- Add check in wavesolvers in case damp incorrectly initialized


examples/seismic/test_seismic_utils.py
```diff
@@ -0,0 +1,97 @@
...
+    assert geometry.rec.shape == (251, nrec)
+    assert norm(geometry.rec) == 0
+    assert geometry.src.shape == (251, 1)
+    assert norm(geometry.new_src(src_type=None)) == 0
+
+    rec2 = geometry.rec.resample(num=501)
+    assert rec2.shape == (501, nrec)
+    assert rec2.grid == model.grid
+
+    assert geometry.new_rec(name="bonjour").name == "bonjour"
```
georgebisbas, line 63: ;-D

examples/seismic/utils.py
```diff
@@ -90,21 +91,19 @@ def resample(self, dt):
...
         return self._src_type
 
     @property
     def grid(self):
-        return self.model.grid
+        return self._grid
+
+    @property
+    def model(self):
+        warning("model is kept for backward compatibility but should not be"
```
georgebisbas, line 104: Model property

georgebisbas (approved):

examples/seismic/model.py
```diff
@@ -40,9 +40,10 @@ def initialize_damp(damp, nbl, spacing, abc_type="damp", fs=False):
             val = -val if abc_type == "mask" else val
             eqs += [Inc(damp.subs({d: dim_l}), val/d.spacing)]
         # right
+        dampcoeff = 1.5 * np.log(1.0 / 0.001) / (nbr)
         dim_r = SubDimension.right(name='abc_%s_r' % d.name, parent=d,
-                                   thickness=nbl)
-        pos = Abs((nbl - (d.symbolic_max - dim_r) + 1) / float(nbl))
+                                   thickness=nbr)
```
FabioLuporini, line 45: should we rather call these variables `nbll` and `nblr` ?

examples/seismic/model.py
```diff
@@ -10,7 +10,7 @@
            'ModelViscoelastic', 'ModelViscoacoustic']
 
 
-def initialize_damp(damp, nbl, spacing, abc_type="damp", fs=False):
+def initialize_damp(damp, padsizes, spacing, abc_type="damp", fs=False):
```
FabioLuporini, line 13: why not keeping `nbl` ?

examples/seismic/model.py
```diff
@@ -124,12 +125,23 @@ def __init__(self, origin, spacing, shape, space_order, nbl=20,
...
 
+    def check_damp(self, bcs="mask"):
+        if self.nbl != 0:
+            center = self.damp.data[tuple(s // 2 for s in self.damp.shape)]
+            re_init = (bcs == "mask" and center == 0) or (bcs == "damp" and center == 1)
+            if re_init:
+                initialize_damp(self.damp, self.padsizes, self.spacing,
+                                abc_type=bcs, fs=self.fs)
+        else:
+            self.damp = 1 if bcs == "mask" else 0
```
FabioLuporini, line 143: changing state on a public function (although I understand you mean to only call this from the inside)

also, this looks like a repetition of https://github.com/devitocodes/devito/pull/1434/files#diff-5535037d20b3ed51e9065da448791b3cR132

I think there's something off in the way `damp` is set/handled, with some redundant code

examples/seismic/utils.py
```diff
@@ -140,15 +139,18 @@ def dtype(self):
...
                         time_range=self.time_axis, npoint=self.nrec,
                         coordinates=self.rec_positions)
 
     @property
     def adj_src(self):
         if self.src_type is None:
-            warning("No surce type defined, returning uninitiallized (zero) shot record")
-            return self.rec
+            warning("No source type defined, returning uninitiallized (zero) shot record")
+            return self.new_rec()
```
FabioLuporini, line 153: this looks off, because then this isn't a property, it's a method?
can we turn it into a cached_property or does mutability forbid this?

examples/seismic/model.py
mloubout: I think nbl/nbr is fine, what we use in builtins as well.

examples/seismic/model.py
mloubout: Cuz may be non-symmetric (or will be)

examples/seismic/model.py
mloubout: This is just backward compatibility so yes it will change the state since it's fixing bad setup for some examples. WOuld want to remove at some point

examples/seismic/utils.py
mloubout: Not sure. These are all method that just return new src/rec object based on the geometry, these will be read/written to in an operator so I think cached_property would make it quite unsafe .

examples/seismic/model.py
FabioLuporini: can we at least turn it into a private method? also I found it would be cleaner if we called it
```
def _initialize_damp(self, bcs='mask'):
    if ...
        ....
        if re_init:
            initialize_damp(...)
else:
     self.damp = ...
```

also it needs a warning comment !

```
# HACK: This is re-initializing damp exploiting <...> because <...>
```

examples/seismic/utils.py
FabioLuporini: can we at least turn properties such as `adj_src` into actual methods?

FabioLuporini: because conceptually they're not really _properties_ of a given Geometry instance

examples/seismic/model.py
mloubout: revamped/cleaned up

examples/seismic/model.py
```diff
@@ -118,17 +119,30 @@ def __init__(self, origin, spacing, shape, space_order, nbl=20,
...
-            self._physical_parameters = ['damp']
-        else:
+        if self.nbl == 0:
             self.damp = 1 if bcs == "mask" else 0
-            self._physical_parameters = []
+            return
+
+        # First initialization
+        init = self.damp is None
+        # Get current Function if alread yinitialized
```
FabioLuporini, line 134: already initialized

examples/seismic/model.py
```diff
@@ -118,17 +119,30 @@ def __init__(self, origin, spacing, shape, space_order, nbl=20,
         else:
             self.grid = grid
 
+        self._physical_parameters = set()
+        self.damp = None
+        self._initialize_bcs(bcs=bcs)
```
FabioLuporini, line 124: why do we need to initialise the BCs here?

rhodrin (approved):

FabioLuporini (approved):
Approving (because the retrocompatibility hack is in the `examples` -- if it were in `devito` we would have acted differently)

FabioLuporini (approved):
LGTM. Any other comments, @rhodrin ?

PR #1438, opened 2020-08-24
Improve bitwise reproducibility

ggorman:
- Added configuration option DEVITO_SAFE_MATH (e.g. turns off -ffast-math on GCC). Off by default.
- Fixed ordering of expressions in Devito. Not a bulletproof solution due to ordering issues within SymPy.
- Added test cases checking for Devito floating-point errors.


devito/compiler.py
```diff
@@ -354,7 +354,8 @@ def __init__(self, *args, **kwargs):
         super(GNUCompiler, self).__init__(*args, **kwargs)
 
         self.cflags += ['-march=native', '-Wno-unused-result', '-Wno-unused-variable',
-                        '-Wno-unused-but-set-variable', '-ffast-math']
+                        '-Wno-unused-but-set-variable', '-ffast-math',
+                        '-fno-unsafe-math-optimizations']
```
FabioLuporini, line 358: hang on -- what's the performance penalty with this?

devito/ir/support/utils.py
```diff
@@ -80,6 +80,7 @@ def build_iterators(mapper):
     for k, v in mapper.items():
         for d, offs in v.items():
             if d.is_Stepping:
+                offs = sorted(offs, key=lambda x: -float("inf") if x == 0 else x)
```
FabioLuporini, line 83: update me on what's going on here ?

tests/test_gradient.py
```diff
@@ -9,31 +9,20 @@
 
 class TestGradient(object):
 
-    @pytest.mark.parametrize('space_order', [4])
-    @pytest.mark.parametrize('kernel', ['OT2'])
-    @pytest.mark.parametrize('shape', [(70, 80)])
-    def test_gradient_checkpointing(self, shape, kernel, space_order):
+    @pytest.mark.parametrize('dtype, opt', [
+        (np.float32, ('noop', {'openmp': True, 'par-dynamic-work': 1000})),
```
FabioLuporini, line 13: `par-dynamic-work` ? that's Navjot's hack around the original failure. Unnecessary here?

tests/test_gradient.py
```diff
@@ -49,7 +38,7 @@ def test_gradient_checkpointing(self, shape, kernel, space_order):
 
         gradient, _ = wave.jacobian_adjoint(residual, u0, vp=v0, checkpointing=True)
         gradient2, _ = wave.jacobian_adjoint(residual, u0, vp=v0, checkpointing=False)
-        assert np.allclose(gradient.data, gradient2.data)
+        assert np.allclose(gradient.data, gradient2.data, atol=0, rtol=0)
```
FabioLuporini, line 41: interesting, what's the difference without these additions?

tests/test_gradient.py
```diff
@@ -9,31 +9,20 @@
...
         """
-        spacing = tuple(10. for _ in shape)
-        wave = setup(shape=shape, spacing=spacing, dtype=np.float64,
-                     kernel=kernel, space_order=space_order,
-                     nbl=40)
+        wave = setup(shape=(70, 80), spacing=(10., 10.), dtype=dtype,
+                     kernel='OT2', space_order=4, nbl=40, opt=opt)
 
-        v0 = Function(name='v0', grid=wave.model.grid, space_order=space_order)
+        v0 = Function(name='v0', grid=wave.model.grid, space_order=4)
```
FabioLuporini, line 25: I'd keep parametrization over `space_order` for future extensions

devito/compiler.py
FabioLuporini: maybe we should add DEVITO_SAFEMATH escape hatch rather than adding it unconditionally

FabioLuporini: how about the other compilers?

devito/ir/support/utils.py
FabioLuporini: smells of hack

FabioLuporini (approved):

FabioLuporini (changes requested):
EDIT -- requesting changes, I erroneously approved it before

devito/ir/support/utils.py
mloubout: You can check the thread on slack 🤣  but basically just makes sure offsets are ordered the way sympy orders additions based on indices. So it's not a hack but just a compatibility reordering.

tests/test_gradient.py
mloubout: No, it's still necessary in single-precision since dynamic scheduling leads to non-bitwise reproducibility even with correct generated code.

tests/test_gradient.py
mloubout: `rtol=1e-05, atol=1e-08` that's the numpy defaults

devito/compiler.py
ggorman: There are lots of health warnings relating to the use of -funsafe-math-optimization ... because it's unsafe. I think the default should be -fno-unsafe-math-optimization and we should add a Devito compiler optimization option that specifically enables unsafe optimization options.

tests/test_gradient.py
ggorman: Is that really true - I mean have you run it and had it fail? I've been running this thousands of times on two different systems and never once an issue.

tests/test_gradient.py
ggorman: In effect you are asking if they are bitwise identical.

tests/test_gradient.py
mloubout: Maybe not, haven't tried latest versio

tests/test_gradient.py
```diff
@@ -9,31 +9,20 @@
...
-    @pytest.mark.parametrize('space_order', [4])
-    @pytest.mark.parametrize('kernel', ['OT2'])
-    @pytest.mark.parametrize('shape', [(70, 80)])
-    def test_gradient_checkpointing(self, shape, kernel, space_order):
+    @pytest.mark.parametrize('dtype, opt', [
+        (np.float32, ('noop', {'openmp': True, 'par-dynamic-work': 1000})),
+        (np.float64, 'advanced')
+    ])
+    def test_gradient_checkpointing(self, dtype, opt):
         r"""
```
georgebisbas, line 17: leftover?

tests/test_gradient.py
mloubout: @ggorman yep still neede, fails without it.

FabioLuporini: BTW, with @tjb900 we recently talked about implementing deterministic injection for the SparseFunction submodule inside dugwave

tests/test_gradient.py
FabioLuporini: so equivalent to `np.all(gradient.data == gradient2.data)` ?

devito/compiler.py
ggorman: DEVITO_SAFE_MATH configuration option added. The default is 0 (off).

tests/test_gradient.py
ggorman: I rewrote this so that only the gradient Function is double (np.float64) to avoid the accumulation of roundoff errors after thousands of interactions. No more noop.

tests/test_gradient.py
ggorman: Yes.

tests/test_gradient.py
mloubout: That's not good, can we revert to previous setup. Forcing the gradient to `np.float64` is basically bypassing the tst setup to do compute n double precision so we only pretend to have the single precision test passing.

tests/test_gradient.py
ggorman: Removed

devito/ir/support/utils.py
```diff
@@ -81,7 +81,7 @@ def build_iterators(mapper):
         for d, offs in v.items():
             if d.is_Stepping:
                 values = iterators.setdefault(d.parent, [])
-                for i in sorted(offs):
+                for i in sorted(offs, key=lambda x: -float("inf") if x == 0 else x):
```
ggorman, line 84: @mloubout - can you leave a comment to explain this piece of black magic.

mloubout: done

.github/workflows/tutorials.yml
```diff
@@ -90,12 +87,17 @@ jobs:
 
     - name: Seismic notebooks
       run: |
-        $RUN_CMD py.test --nbval --cov . --cov-config=.coveragerc --cov-report=xml:seis_coverage.xml $SKIP examples/seismic/tutorials/
-        $RUN_CMD py.test --nbval --cov . --cov-config=.coveragerc --cov-report=xml:acc_coverage.xml $SKIP examples/seismic/acoustic/accuracy.ipynb
+        $RUN_CMD py.test --nbval --cov . --cov-config=.coveragerc --cov-report=xml:seis_coverage.xml -k 'not dask' examples/seismic/tutorials/
```
mloubout, line 90: That's not super pretty but ` echo "::set-env name=SKIP::-k 'not dask'"` was erroring and don't have time to dig into why.

rhodrin (approved):

devito/ir/support/utils.py
FabioLuporini: OK. I'm not 100% sure I like this fix (what if SymPy changes its internal policy for canonical ordering? we should actually sort based on that somehow....), but I can buy it for now

tests/test_gradient.py
FabioLuporini: I'm not sure why was this conversation marked as resolved when it's clearly not. Also, we're still disabling dynamic parallelism (option `par-dynamic-work` ) for no reason)

tests/test_gradient.py
FabioLuporini: I'm also unresolving this one. I think we should reinstate `space_order` for parametrization and also add at least one test case for higher space order

FabioLuporini (changes requested):

devito/ir/support/utils.py
FabioLuporini: After my investigation yesterday (for other reasons), I can say that this ordering

```
(t, t-1, t+1)
```

is happening in SymPy because "Symbols always precede operations such as Add and Mul". So `t` is a Symbol while `t-1` and `t+1` are Add's, hence the ordering above. However, there are big TODOs in SymPy that this might change in the future (they're explicitly talking about dropping their `ordering_of_classes` (u may wanna `git grep ordering_of_classes` in SymPy to see what I mean). 

So just to be sure. When you say: 

```
follows sympy's index ordering (t, t-1, t+1) afer modulo replacement
```

what do you exactly mean?

mloubout: >what do you exactly mean?
I mean  `t0=t, t1=t-1, t2=t+1` because for sympy `t0<t1` and `t1<t2` , while curently it generates it by offset and offset order is {-1, 0, 1} and therefore generate  `t0=t-1, t1=t, t2=t+1` which has not the same order.

devito/compiler.py
```diff
@@ -154,7 +154,7 @@ def __init__(self, **kwargs):
             self.cc = self.MPICC if kwargs.get('cpp', False) is False else self.MPICXX
         self.ld = self.cc  # Wanted by the superclass
 
-        self.cflags = ['-O3', '-g', '-fPIC', '-Wall', '-std=c99']
+        self.cflags = ['-O3', '-g', '-fPIC', '-Wall', '-std=c99', '-pedantic']
```
FabioLuporini, line 157: I see why you added it, but do we really _need_ it? pragmatically speaking...

Did you check this is not causing any of the GCC compilers to emit warnings we would never want users to see? I think it really shouldn't since `pedantic`, if anything, should throw errors not warnings...

devito/ir/support/utils.py
```diff
@@ -81,7 +81,12 @@ def build_iterators(mapper):
         for d, offs in v.items():
             if d.is_Stepping:
                 values = iterators.setdefault(d.parent, [])
-                for i in sorted(offs):
+                # Offsets are sorted so that the semantic order (t0, t1, t2)
+                # follows sympy's index ordering (t, t-1, t+1) afer modulo replacement
```
FabioLuporini, line 85: "afer" typo

devito/ir/support/utils.py
```diff
@@ -81,7 +81,12 @@ def build_iterators(mapper):
...
+                # and guarantying bitwise reproducibility.
```
FabioLuporini, line 86: I would drop this, or at least relax it -- isn't it too strong?

tests/test_gradient.py
```diff
@@ -9,53 +9,52 @@
 
 class TestGradient(object):
 
-    @pytest.mark.parametrize('space_order', [4])
-    @pytest.mark.parametrize('kernel', ['OT2'])
-    @pytest.mark.parametrize('shape', [(70, 80)])
-    def test_gradient_checkpointing(self, shape, kernel, space_order):
+    @pytest.mark.parametrize('dtype', [np.float32, np.float64])
+    @pytest.mark.parametrize('opt', [('advanced',
+                                      {'openmp': True, 'par-dynamic-work': 0}),
```
FabioLuporini, line 14: I still don't understand why do we keep messing around with `par-dynamic-work` ? is it because we rather want to _enforce_ it?

tests/test_roundoff.py
```diff
@@ -0,0 +1,158 @@
+import pytest
+import numpy as np
+
+from devito import Grid, Constant, TimeFunction, Eq, Operator, switchconfig
+
+
+class TestRoundoff(object):
```
FabioLuporini, line 7: Is it worth adding a test which tries both `safe_math=False` and `safe_math=True` with given suitable tolerances ?

FabioLuporini (commented):
Aside from a few minor things, I think we're pretty much ready

devito/compiler.py
ggorman: dropped

devito/ir/support/utils.py
ggorman: Isn't this just revisiting the discussion with @mloubout above?

tests/test_gradient.py
ggorman: Dropped.

devito/ir/support/utils.py
mloubout: I think he just means drop "guarantying bitwise reproducibility." that is a bit too strong of a statement.

ggorman: Changed last line to 'so that associativity errors are consistent.'

tests/test_roundoff.py
ggorman: The test is explicitly designed to diverge when any floating-point error is made. Quantifying the magnitude of the error is not telling us anything additional is adds an additional layer of complexity to this test. If there is a wider accuracy issue then there are thousands of other tests already present to detect and isolate those errors.

devito/ir/support/utils.py
ggorman: fixed

tests/test_roundoff.py
FabioLuporini: OK

devito/compiler.py
```diff
@@ -382,6 +386,8 @@ def __init__(self, *args, **kwargs):
         super(ClangCompiler, self).__init__(*args, **kwargs)
 
         self.cflags += ['-Wno-unused-result', '-Wno-unused-variable', '-ffast-math']
+        if not configuration['safe-math']:
+            self.cflags.append('-ffast-math')
```
FabioLuporini, line 390: this is being added twice: see line 388

FabioLuporini (approved):
This is good to go now. Merging. Thanks

PR #1446, opened 2020-09-10
Fix parallel increments

FabioLuporini:
* Two bug fixes concerning the generation of `reduction` clauses and `pragma omp atomic` (now avoided if unnecessary)
* Ships tests
* Supersedes #1433

### DIFF
--- a/devito/passes/iet/openmp.py
+++ b/devito/passes/iet/openmp.py
@@ -4,6 +4,7 @@ import numpy as np
 import cgen as c
 from sympy import Or, Max, Not
 
+from devito.data import FULL
 from devito.ir import (DummyEq, Conditional, Dereference, Expression, ExpressionBundle,
                        List, Prodder, ParallelIteration, ParallelBlock, While,
                        FindSymbols, FindNodes, Return, COLLAPSED, VECTORIZED, Transformer,
@@ -135,8 +136,22 @@ class OpenMPIteration(ParallelIteration):
             clauses.append('num_threads(%s)' % nthreads)
 
         if reduction:
-            args = ','.join(str(i) for i in reduction)
-            clauses.append('reduction(+:%s)' % args)
+            args = []
+            for i in reduction:
+                if i.is_Indexed:
+                    # OpenMP expects a size not an index as input of reduction,
+                    # such as reduction(+:f[d_m:d_M+1])
+                    f = i.function
+                    bounds = []
+                    for k, d in zip(i.indices, f.dimensions):
+                        if k.is_Number:
+                            bounds.append('[%s:%s]' % (str(k), str(k + 1)))
+                        else:
+                            bounds.append('[0:%s]' % f._C_get_field(FULL, d).size)
+                    args.append('%s%s' % (i.name, ''.join(bounds)))
+                else:
+                    args.append(str(i))
+            clauses.append('reduction(+:%s)' % ','.join(args))
 
         return clauses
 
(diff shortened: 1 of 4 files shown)


tjb900 (approved):
Haven’t looked at above code I confess, but generated code looks good and fixes issue. Thanks!

georgebisbas (approved):

devito/passes/iet/openmp.py
```diff
@@ -135,8 +136,22 @@ def _make_clauses(cls, ncollapse=None, chunk_size=None, nthreads=None,
             clauses.append('num_threads(%s)' % nthreads)
 
         if reduction:
-            args = ','.join(str(i) for i in reduction)
-            clauses.append('reduction(+:%s)' % args)
+            args = []
+            for i in reduction:
+                if i.is_Indexed:
+                    # OpenMP expects a size not an index as input of reduction,
+                    # such as reduction(+:f[d_m:d_M+1])
```
mloubout, line 143: Need to update this since not the generated anymore.

tests/test_gpu_openmp.py
```diff
@@ -194,7 +194,7 @@ def test_timeparallel_reduction(self):
         assert not tree.root.pragmas
         assert len(tree[1].pragmas) == 1
         assert tree[1].pragmas[0].value ==\
-            'omp target teams distribute parallel for collapse(3) reduction(+:f[0])'
+            'omp target teams distribute parallel for collapse(3) reduction(+:f[0:1])'
```
mloubout, line 197: This should have been caught by ` if i.is_Indexed:` and still be `f[0]` since it's a single index. Is it from my PR?

devito/passes/iet/openmp.py
FabioLuporini: gotcha, fixed

tests/test_gpu_openmp.py
FabioLuporini: no, from mine, I was just handling both cases (number and symbol) homegenously. But I've now rolled back to previous behaviour

tests/test_dle.py
```diff
@@ -523,6 +521,62 @@ def test_scheduling(self):
         assert not iterations[3].is_Affine
         assert 'schedule(dynamic,chunk_size)' in iterations[3].pragmas[0].value
 
+    @pytest.mark.parametrize('so', [0, 1, 2])
+    @pytest.mark.parametrize('dim', [1, 2])
```
mloubout, line 525: Is the `dim=0` case fix with the other part of that PR (the reduction isn't picked up if most outer loop).

FabioLuporini: I'm gonna check now but I think so since the assert is now for all test cases

FabioLuporini: yes, it looks correct:

```
    {
      #pragma omp for collapse(3) schedule(static,1) reduction(+:f[0:f_vec->size[0]])
      {
        {
          {
            f[y] += u[t0][x + 1][y + 1][z + 1] + 1;
          }
        }
      }
    }
```

mloubout: That's still the `y` dimension not `x`, need to change  `@pytest.mark.parametrize('dim', [0, 1, 2])` to get that previously problematic case.

mloubout: Added `0` back to dims and seems to be working fine now.

mloubout (approved):

PR #1452, opened 2020-09-15
Moving and packaging the Intel Advisor 2020 scripts into benchmarks/user/advisor

jack-lascala:
Shifted the scripts to package to allow using the `run_advisor.py`, `roofline.py`, `advisor_to_json.py` scripts when installing Devito. Adds a small collection of logging functionality for the scripts under `advisor_logging.py`. Allows using roof lining scripts externally, such as for TheMatrix.


FabioLuporini (changes requested):
You shouldn't just copy the files over, you should also remove the whole advisor machinery from scripts/advisor

FabioLuporini (approved):
Never mind, ignore me. All looking good. Thanks

georgebisbas (approved):

PR #1454, opened 2020-09-16
caching: Protect iteration through _SymbolCache

FabioLuporini:


ggorman:
I believe we should withdraw this PR as it's superseded by  #1458

FabioLuporini:
> I believe we should withdraw this PR as it's superseded by #1458

indeed, closing

PR #1456, opened 2020-09-16
Intel Advisor jupyter notebook benchmarks/user/advisor scripts update

jack-lascala:
Changing the scripts used in the 02_advisor_roofline jupyter notebook to reflect the movement of the advisor scripts.


FabioLuporini (approved):
uncontroversial

Issue #1453, opened 2020-09-15
Multiple SubDomainSet with multiple equations per set


mloubout:
THe generated code is missing the bound declaration when there is multiple SubDomainSet and multiple equations. The MFE is basically just adding one equation to `tests/test_subdomains.py`::test_multisets:

```
        eq1 = Eq(f, f+2, subdomain=grid.subdomains['mydomains1'])
        eq2 = Eq(g, g+2, subdomain=grid.subdomains['mydomains2'])
        eq3 = Eq(f, f-1, subdomain=grid.subdomains['mydomains1'])

        op = Operator([eq1, eq2, eq3])
```

that produces the following code:

```
#pragma omp parallel num_threads(nthreads)
  {
    #pragma omp for collapse(1) schedule(static,1)
    for (int n = n_m; n <= n_M; n += 1)
    {
      int xi_n_ltkn = xi_n_m[n];
      int yi_n_ltkn = yi_n_m[n];
      int xi_n_rtkn = xi_n_M[n];
      int yi_n_rtkn = yi_n_M[n];
      for (int xi_n = x_m + xi_n_ltkn; xi_n <= x_M - xi_n_rtkn; xi_n += 1)
      {
        #pragma omp simd aligned(f:16)
        for (int yi_n = y_m + yi_n_ltkn; yi_n <= y_M - yi_n_rtkn; yi_n += 1)
        {
          f[xi_n + 1][yi_n + 1] = f[xi_n + 1][yi_n + 1] + 2;
        }
      }
    }
  }
  /* End section0 */
  gettimeofday(&end_section0, NULL);
  timers->section0 += (double)(end_section0.tv_sec-start_section0.tv_sec)+(double)(end_section0.tv_usec-start_section0.tv_usec)/1000000;
  struct timeval start_section1, end_section1;
  gettimeofday(&start_section1, NULL);
  /* Begin section1 */
  #pragma omp parallel num_threads(nthreads)
  {
    #pragma omp for collapse(1) schedule(static,1)
    for (int m = m_m; m <= m_M; m += 1)
    {
      int xi_m_ltkn = xi_m_m[m];
      int yi_m_ltkn = yi_m_m[m];
      int xi_m_rtkn = xi_m_M[m];
      int yi_m_rtkn = yi_m_M[m];
      for (int xi_m = x_m + xi_m_ltkn; xi_m <= x_M - xi_m_rtkn; xi_m += 1)
      {
        #pragma omp simd aligned(g:16)
[...]

PR #1457, opened 2020-09-16
Fix SubDomainSet bug

rhodrin:
Fixes a bug that meant that equations assocciated with a `SubDomainSet` had to be ordered in 'groups'. Fixes issue #1453. Thanks to @mloubout.


FabioLuporini (approved):
I love it when we fix bugs by actually dropping code :)

FabioLuporini:
merged, thanks

mloubout:
As an aside, some "cleanup" may be done at some point in the generated C code that now generates the implicit expressions for each equation and therefore split in two loops if there is two equations.

PR #1458, opened 2020-09-16
CI: skip dask on osc builds

mloubout:


ggorman (approved):
We have pretty much convinced ourselves that there is a dask/distributed bug on OSX. As distributed computing on OSX is not a priority, we are fine to skip testing dask tutorials on OSX.

PR #1461, opened 2020-09-21
Added origin_map attribute to Grid

EdCaunt:
Added and origin_map analogous to spacing_map. The idea is to have a shortcut to obtain the float values of the tuple rather than having to do [i.data for i in grid.origin]

### DIFF
--- a/devito/types/grid.py
+++ b/devito/types/grid.py
@@ -168,8 +168,7 @@ class Grid(ArgProvider):
     @property
     def origin_map(self):
         """Map between origin symbols and their values"""
-        origin_pos = [i.data for i in self.origin]
-        return dict(zip(self.origin, origin_pos))
+        return {o: o.data for o in self.origin}
 
     @property
     def dimensions(self):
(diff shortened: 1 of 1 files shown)


devito/types/grid.py
```diff
@@ -165,6 +165,12 @@ def origin(self):
         """Physical coordinates of the domain origin."""
         return self._origin
 
+    @property
+    def origin_map(self):
+        """Map between origin symbols and their values"""
+        origin_pos = [i.data for i in self.origin]
+        return dict(zip(self.origin, origin_pos))
```
mloubout, line 172: Can one-line
`{o: o.data for o in self.origin}`

EdCaunt: Good point, will do

rhodrin:
As an aside, since we don't have a `test_grid.py` is there much `Grid` related stuff we're no currently testing?

FabioLuporini:
> As an aside, since we don't have a test_grid.py is there much Grid related stuff we're no currently testing?

a Grid is more of a bag of things, so basically the testing occurs indirectly through regression tests

PR #1462, opened 2020-09-24
perf: Add global gflopss to operator log

georgebisbas:
Adds global gflops/s to the operator log

### DIFF
--- a/devito/operator/operator.py
+++ b/devito/operator/operator.py
@@ -726,8 +726,12 @@ class Operator(Callable):
 
             v = summary.globals.get('fdlike')
             if v is not None:
-                perf("%s* Achieved %.2f FD-GPts/s" % (indent, v.gpointss))
-
+                if v.gflopss is not None:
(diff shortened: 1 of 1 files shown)


devito/operator/operator.py
```diff
@@ -751,8 +751,12 @@ def _emit_apply_profiling(self, args):
 
             v = summary.globals.get('fdlike')
             if v is not None:
-                perf("%s* Achieved %.2f FD-GPts/s" % (indent, v.gpointss))
-
+                if v.gflopss is not None:
```
FabioLuporini, line 754: remind me when it could be None ?

georgebisbas: It happened when test_autotuning was running.
I will check again.

FabioLuporini: OK, I see.

Can we show the GFlops/s first and the FD-GPts/s later? to make it homogeneous with the other prints

georgebisbas: ok

FabioLuporini (commented):
Please rebase and squash into a single commit

PR #1463, opened 2020-09-24
Advisor lacking profiling information bug fix

jack-lascala:
When insufficient information is gathered from profiling, `roofline.py` and `advisor_to_json.py` scripts fail by not being able to access the columns of the project data. Implementation of guard against failing with an exception and fix for the bug.

### DIFF
--- a/benchmarks/user/advisor/advisor_to_json.py
+++ b/benchmarks/user/advisor/advisor_to_json.py
@@ -66,7 +66,11 @@ def advisor_to_json(name, project, scale, precision, mode, th):
     full_df = pd.DataFrame(rows).replace('', np.nan)
 
     # Narrow down the columns to those of interest
-    df = full_df[analysis_columns].copy()
+    try:
+        df = full_df[analysis_columns].copy()
+    except KeyError:
+        err('Could not read data columns from profiling. Not enough data has been '
+            'generated for the specified problem.')
 
     df.self_ai = df.self_ai.astype(float)
     df.self_gflops = df.self_gflops.astype(float)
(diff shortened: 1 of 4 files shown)


FabioLuporini (commented):
I see duplicated code in those two files. Can we somewhat factor it out?

georgebisbas:
I can handle that soon, I think.

georgebisbas:
Did a rebase here on top of master...in my short-term list

georgebisbas:
> I see duplicated code in those two files. Can we somewhat factor it out?

These two files indeed have duplicated code...a lot... they both do some checks and then extract data, one saves a roofline and the other a json.

My suggestion would be to merge those two on one file and add an arg option to choose an extraction method.
i.e.
instead of having: 
`python ...roofline.py`
`python ...advisor_top_json.py`

we will have:
`python extract_data.py --mode roofline`
`python extract_data.py --mode json`

Thoughts?

FabioLuporini:
@georgebisbas either you or I should clean up these files (ie drop the redundant code somehow) and bring this PR to completion

georgebisbas:
> @georgebisbas either you or I should clean up these files (ie drop the redundant code somehow) and bring this PR to completion

I will do, but do you agree with what I proposed above?

FabioLuporini:
then we would also have to change the documentation in quite a few places. I'd in principle be OK with it, but maybe in this specific case it's just simpler if we move the common stuff to a utils.py ?

georgebisbas:
> then we would also have to change the documentation in quite a few places. I'd in principle be OK with it, but maybe in this specific case it's just simpler if we move the common stuff to a utils.py ?

I agree to move common stuff in a utils.py.

georgebisbas:
Closed temporarily.

PR #1465, opened 2020-09-28
Improve detection and scheduling of CIREs

FabioLuporini:
Note: this is still WIP though nearly ready (I have a regression to fix). 

New features come equipped with tests


devito/finite_differences/finite_difference.py
```diff
@@ -280,7 +280,7 @@ def indices_weights_to_fd(expr, dim, inds, weights, matvec=1):
             iloc = sympify(i).xreplace(mapper)
         # Enforce fixed precision FD coefficients to avoid variations in results
         c = sympify(c).evalf(_PRECISION)
-        terms.append(expr._subs(expr.indices_ref[dim], iloc) * c)
+        terms.append(expr._subs(dim, iloc - (expr.indices_ref[dim] - dim)) * c)
```
FabioLuporini, line 283: @mloubout your bug fix incorporated here

devito/ir/equations/algorithms.py
```diff
@@ -68,6 +70,39 @@ def handle_indexed(indexed):
     return ordering
 
 
+def generate_implicit_exprs(expressions):
```
FabioLuporini, line 73: @rhodrin formerly `add_implicit`, now, as anticipated, moved here inside `/ir/equations/algorithms`

devito/finite_differences/lazy.py
```diff
@@ -47,3 +49,15 @@ def evaluate(self):
         args = self._evaluate_args()
         evaluate = not all(i is j for i, j in zip(args, self.args))
         return self.func(*args, evaluate=evaluate)
+
+
+# Custom SymPy types used upon evaluation
+
+
+class EvalDerivative(sympy.Add):
```
mloubout, line 57: Why is it a sympy Add and not ours?

devito/finite_differences/differentiable.py
```diff
@@ -402,6 +402,11 @@ class Mod(DifferentiableOp, sympy.Mod):
     __new__ = DifferentiableOp.__new__
 
 
+class EvalDiffDerivative(DifferentiableOp, sympy.Add):
+    __sympy_class__ = EvalDerivative
```
mloubout, line 406: why not just `(DifferentiableOp, EvalDerivative):`

devito/symbolics/manipulation.py
```diff
@@ -61,7 +61,8 @@ def _uxreplace_handle(expr, args):
 @_uxreplace_handle.register(sympy.Add)
 def _(expr, args):
     if all(i.is_commutative for i in args):
-        _addsort(args)  # In-place sorting
+        _addsort(args)
```
mloubout, line 64: I know it's not new, but I remember looking at `addsort/mulsort` and was quite expensive (why moved from `subs` to `_subs` in part), do we actually need it there?

mloubout (commented):
quick first pass, will have another look later this week

devito/symbolics/manipulation.py
FabioLuporini: yes we do, tried removing and all sorts of issues pop up. Here's the key problem: canonical ordering must be enforced upon expression lowering (e.g., CIRE needs it for max performance). Currently, canonical ordering is enforced during `lower_exprs` through this call to `uxreplace`. So if we want to pull it out of here, we need another pass to enforce the canonical ordering. I'm 100% in favor to it, but I'm not sure procrastinating the PR for this reason is the best way to go

devito/types/basic.py
```diff
@@ -1203,3 +1203,37 @@ def origin(self):
...
+            return super().compare(other)
+        for l, r in zip(self.indices, other.indices):
+            try:
+                expr = l - r
+                if expr > 0:
+                    c = 1
+                elif expr == 0:
+                    c = 0
+                else:
+                    c = -1
```
mloubout, line 1233: how is that different than `c = sign(l-r)`?

devito/symbolics/manipulation.py
mloubout: Ok, too bad, but I get it.

devito/operator/operator.py
```diff
@@ -292,24 +263,27 @@ def _lower_exprs(cls, expressions, **kwargs):
...
+        # Add in implicit expressions
+        expressions = generate_implicit_exprs(expressions)
 
-        # Unfold lazyiness
-        expressions = flatten([i.evaluate for i in expressions])
+        # Specialization is performed on unevaluated expressions
+        expressions = cls._specialize_exprs(expressions, **kwargs)
 
-        # Scalarize tensor expressions
+        # Lower finite-difference DSL
```
mloubout, line 279: nitpicking: Lower functional DSL

devito/finite_differences/finite_difference.py
mloubout: 👍

devito/operator/operator.py
FabioLuporini: Done

devito/finite_differences/lazy.py
FabioLuporini: Because EvalDerivative is what we get upon evaluation. And by the time evaluation is performed, all the FD types will have been lowered to pure SymPy ones (ie, inside the compiler after Equation lowering we enforce pure SymPy for homogeneity)

devito/types/basic.py
FabioLuporini: oops, indeed, good catch. Using `sign` now

devito/finite_differences/differentiable.py
FabioLuporini: Tried, it works, so switching to your suggestion

FabioLuporini (commented):
I'm about to push tweaks for all your suggestions @mloubout 

I've also updated `examples/performance/00_overview.ipynb` which was previously broken on CI

devito/finite_differences/lazy.py
mloubout: Inside the compiler sure, but evaluation is still a user level feature and it's output (the evaluated expression) should still be differentiable

FabioLuporini: > it's output (the evaluated expression) should still be differentiable

they are indeed because upon .evaluate you get objects of type `EvalDiffDerivative` (which is part of the `Differentiable` hierarchy) while _only inside the compiler_, I think within `LoweredEq(...)`, you get `EvalDerivative`

also, all tests are passing so we should be good on this front

mloubout: >.evaluate you get objects of type EvalDiffDerivative 

Right ok should be good then indeed.

devito/finite_differences/derivative.py
```diff
@@ -202,6 +202,10 @@ def _new_from_self(self, **kwargs):
         _kwargs.update(**kwargs)
         return Derivative(expr, *self.dims, **_kwargs)
 
+    @property
+    def func(self):
+        return lambda *a, **kw: self._new_from_self(expr=a[0])
```
mloubout, line 207: why are kwargs ignored? I would think that if we allow `func` we should all related `kwargs` such as `x0, ....`

FabioLuporini:
also approved offline, merging

PR #1469, opened 2020-10-08
Minor tweaks

### LINKED ISSUE #1474: SubDomainsSet generating non-unique Functions for subdomain bounds
ie, if two Eqs use the same SubDomainSet , we end up with two `xi_n_m(n)`, two `xi_n_M(n)`, ... that are syntactically identical in the generated code but different objects in Python

FabioLuporini:
A number of minor patches and improvements extracted from a branch of mine which was becoming unmanageable due to excessive size

fixes #1474


tests/test_iet.py
```diff
@@ -0,0 +1,240 @@
...
+    l2 = l1._rebuild(l1.body)
+    assert len(l2.body) == 0
+    assert str(l2) == str(l1)
+
+    l3 = l2._rebuild(l2.body, **l2.args_frozen)
+    assert len(l3.body) == 0
+    assert str(l3) == str(l2)
+
+
+def test_make_cpp_parfor():
```
FabioLuporini, line 142: note for reviewers: except for this, which is completely new, all other tests have been lifted from test_ir

devito/compiler.py
```diff
@@ -452,7 +452,7 @@ def __init__(self, *args, **kwargs):
         self.cflags.remove('-std=c99')
         self.cflags.remove('-O3')
         self.cflags.remove('-Wall')
-        self.cflags += ['-fast', '-acc']
+        self.cflags += ['-std=c++11', '-fast', '-acc', '-mp']
```
rhodrin, line 455: Out of interest, why the `-mp` flag?

devito/ir/equations/algorithms.py
```diff
@@ -90,9 +91,11 @@ def generate_implicit_exprs(expressions):
                 sub_dims.append(e.subdomain.implicit_dimension)
                 dims = [d for d in dims if d not in frozenset(sub_dims)]
                 dims.append(e.subdomain.implicit_dimension)
-                grid = list(retrieve_functions(e, mode='unique'))[0].grid
-                processed.extend([i.func(*i.args, implicit_dims=dims) for i in
-                                  e.subdomain._create_implicit_exprs(grid)])
+                if e.subdomain not in found:
```
rhodrin, line 94: Did we have an MFE relating to this change that could be added to `test_subdomains.py`?

devito/ir/iet/nodes.py
```diff
@@ -834,6 +886,49 @@ def is_tensor(self):
         return False
 
 
+class Lambda(Node):
```
rhodrin, line 889: I don't have a strong opinion, but could be called `LambdaC` (or something along those lines) to make it a little clearer? That's probably making it a little less 'nice' though.

devito/ir/iet/visitors.py
```diff
@@ -254,23 +255,30 @@ def visit_AugmentedExpression(self, o):
...
 
-        return c.Initializer(c.Value(o.expr.lhs._C_typedata, lhs),
-                             ccode(o.expr.rhs, dtype=o.dtype))
+        return c.Initializer(c.Value(ctype, lhs), ccode(o.expr.rhs, dtype=o.dtype))
 
     def visit_ForeignExpression(self, o):
         return c.Statement(ccode(o.expr))
 
-    def visit_Call(self, o):
+    def visit_Call(self, o, nested_call=False):
```
rhodrin, line 275: I should review the code in more detail, but is there a quick and easy explanation as to why `nested_call` is needed?

rhodrin:
@FabioLuporini Looks great, although I didn't get chance to look through the new test yet. Will do that ASAP.

devito/compiler.py
FabioLuporini: to compile OpenMP

devito/ir/equations/algorithms.py
FabioLuporini: honestly, I don't even remember how I bumped into it... what was the issue again? :grimacing:

FabioLuporini: I mean we, for example, process the same subdomain twice... and then what happens?

devito/ir/iet/visitors.py
FabioLuporini: I need to pass it to `MultilineCall`, IIRC to distinguish between Call statements (terminating with ';') and Call expressions (no ';', thus making it possible to nest Calls)

devito/ir/equations/algorithms.py
FabioLuporini: but IIRC I think that some tests now, with all the other changes shipped by this PR,  do fail w/o this patch

devito/compiler.py
rhodrin: Oh, I thought that flag did something different sorry :p

devito/ir/equations/algorithms.py
rhodrin: We were getting unnecessary replication of implicit equations. I think you posted an MFE in development a while back, but would need to dig it out.

FabioLuporini: now addressed via https://github.com/devitocodes/devito/pull/1469/files#diff-7b79cb6a68638992ebb71a66e8faf75f4508c905d6b68a8827830383a693eedfR258

devito/ir/iet/nodes.py
FabioLuporini: it'd actually need to be LambaCPP (C doesn't have lambdas), so I guess I'll just stick to Lambda, it don't think it's a potential source of confusion after all

rhodrin (approved):

PR #1470, opened 2020-10-09
Tidy gpu CI workflow

rhodrin:
- Removes repeated example runs
- Examples are now tested with and without MPI
- All examples tested on openacc + openmp

### DIFF
--- a/.github/workflows/pytest-gpu.yml
+++ b/.github/workflows/pytest-gpu.yml
@@ -66,31 +66,21 @@ jobs:
           pgaccelinfo
         fi
         pytest --cov --cov-config=.coveragerc --cov-report=xml ${{ matrix.test_file }}
-        pytest --cov --cov-config=.coveragerc --cov-report=xml examples/seismic/elastic/elastic_example.py
-        pytest --cov --cov-config=.coveragerc --cov-report=xml examples/seismic/acoustic/acoustic_example.py
-        pytest --cov --cov-config=.coveragerc --cov-report=xml examples/seismic/viscoacoustic/viscoacoustic_example.py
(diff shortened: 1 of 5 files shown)


rhodrin:
NOTE: After investigating the opemp fails need to remember to re-add codecov for examples.

.github/workflows/pytest-gpu.yml
```diff
@@ -66,31 +66,21 @@ jobs:
...
-    - name: Run examples
-      if: matrix.name == 'pytest-gpu-acc'
+    - name: Test examples
       run: |
-        python3 examples/seismic/elastic/elastic_example.py
-        python3 examples/seismic/elastic/elastic_example.py
-        python3 examples/seismic/tti/tti_example.py
-        python3 examples/seismic/viscoacoustic/viscoacoustic_example.py
-        python3 examples/seismic/viscoelastic/viscoelastic_example.py
+        pytest examples/seismic/elastic/elastic_example.py
```
FabioLuporini, line 72: should we keep the --cov etc ?

rhodrin: Yes, we should - see note above (in the PR that is, not in the yml).

rhodrin:
@italoaug When you get the chance, would you mind taking a look at what's going on here? Thanks.

.github/workflows/pytest-gpu.yml
```diff
@@ -66,31 +66,21 @@ jobs:
...
+        pytest examples/seismic/elastic/elastic_example.py
```
italoaug, line 73: The elastic example is tested twice without MPI.
The acoustic example is not tested with and without MPI.

FabioLuporini (approved):

PR #1475, opened 2020-10-21
Conditonals: Added diff2sympy to conditionals

EdCaunt:
Fix for the issue where derivatives in ConditionalDimension conditions would cause a python-level error.


FabioLuporini (approved):
LGTM, merging

PR #1481, opened 2020-10-23
Allow Add objects to be passed to generate_subs

EdCaunt:
Patch symbolic coefficients so that custom coefficients can be used with staggered grids.


devito/finite_differences/coefficients.py
```diff
@@ -83,6 +84,27 @@ def dimension(self):
         """The dimension to which the coefficients will be applied."""
         return self._dimension
 
+    @property
+    def offset_dimension(self):
+        """
+        The dimension to which the coefficients will be applied plus the offset
+        in that dimension if the function is staggered.
+        """
+        if self._function.staggered is None:
```
mloubout, line 93: you should be able to get this dimension directly from `self._function.indices_ref[self._dimension]`

EdCaunt: Thanks, noted, will update now

devito/finite_differences/coefficients.py
```diff
@@ -217,27 +227,38 @@ def generate_subs(i):
         return rules
 
 
+def extract_dim(dim):
+    if isinstance(dim, sympy.Add):
+        # Staggered function: need to extract dimension from dim
+        return list(dim.as_coefficients_dict())[-1]
```
mloubout, line 233: Is it always guarantied it's the last one?

tests/test_symbolic_coefficients.py
```diff
@@ -112,3 +112,59 @@ def test_coefficients_w_xreplace(self):
         expected = '0.1*u(x, y) - 0.6*u(x - h_x, y) + 0.6*u(x + h_x, y) + 2'
 
         assert expected == str(eq.evaluate.lhs)
+
+    @pytest.mark.parametrize('order', [2, 4, 6])
+    def test_staggered_array(self, order):
```
mloubout, line 117: Do we wanna add a symbolic test as well that checks the obtained expression is what we want?

mloubout (commented):
Looks nice, thankis, just minor comments

devito/finite_differences/coefficients.py
EdCaunt: Not 100% sure, I assume it would be, but need to double check. Could potentially take both and check which one is a Dimension instead?

tests/test_symbolic_coefficients.py
EdCaunt: Good point, I'll have a look into it

devito/finite_differences/coefficients.py
mloubout: Think of it, all you want is to know whether it's a time dimension or not right?

so you could have something like
```
def is_time(function, dim):
   any (getattr(i, 'is_Time', False) for i in dim.args)
```
and same for space (can inline in if below)

EdCaunt: In a couple of other places further down, I need the dimension on its own (without the offset). Would be nice to have a tidier/more reliable to get the dimension from the Add though

PR #1482, opened 2020-10-23
Fixing halo spot merge rule (issue 1459)

Leitevmd:
Fix rule2 of _merge_halospots(mpi) solving issue 1459
Add test to check proper haloupdate schedule

### DIFF
--- a/tests/test_mpi.py
+++ b/tests/test_mpi.py
@@ -946,6 +946,23 @@ class TestCodeGeneration(object):
             assert np.allclose(g.data_ro_domain[0, 5:], [4.8, 4.8, 4.8, 4.8, 2.], rtol=R)
 
     @pytest.mark.parallel(mode=2)
+    def test_haloudate_scheduling_with_time_plus_no_time_varying_funcs(self):
+        grid = Grid(shape=(10,))
+        x = grid.dimensions[0]
+        t = grid.stepping_dim
+
+        f = Function(name='f', grid=grid, space_order=1)
+        g = TimeFunction(name='g', grid=grid, space_order=1, time_order=2)
+
+        eqns = [Eq(f, g[t, x+1] + g[t, x-1]),
+                Eq(g.forward, f[x-1] + f[x+1])]
+
+        op = Operator(eqns)
+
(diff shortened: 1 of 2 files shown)


tests/test_mpi.py
```diff
@@ -945,6 +945,23 @@ def test_avoid_haloudate_if_flowdep_along_other_dim(self):
             assert np.allclose(f.data_ro_domain[0, 5:], [4.8, 4.8, 4.8, 4.8, 3.6], rtol=R)
             assert np.allclose(g.data_ro_domain[0, 5:], [4.8, 4.8, 4.8, 4.8, 2.], rtol=R)
 
+    @pytest.mark.parallel(mode=2)
+    def test_haloudate_scheduling_with_time_plus_no_time_varying_funcs(self):
```
FabioLuporini, line 949: nitpicking: would you mind changing the name to `test_unmerge_haloupdate_if_no_locindices` ? I find it a bit clearer (and ideally we'd need it clear since there's no docstring).

tests/test_mpi.py
```diff
@@ -945,6 +945,23 @@ def test_avoid_haloudate_if_flowdep_along_other_dim(self):
...
+        x = grid.dimensions[0]
+        t = grid.stepping_dim
+
+        f = Function(name='f', grid=grid, space_order=1)
+        g = TimeFunction(name='g', grid=grid, space_order=1, time_order=2)
+
+        eqns = [Eq(f, g[t, x+1] + g[t, x-1]),
+                Eq(g.forward, f[x-1] + f[x+1])]
+
+        op = Operator(eqns)
```
FabioLuporini, line 960: I think we should also attempt execution and asserting against some expected values? since it's a fairly delicate test

FabioLuporini (commented):
left a couple of minor comments, but both the patch and the test are fundamentally OK

tests/test_mpi.py
Leitevmd: I've used `np.allclose` assertion as is the approach used in previous tests.

FabioLuporini (approved):
Perfect, thanks a lot. Merging once CI is green.

FabioLuporini:
Thanks, merged.

PR #1487, opened 2020-11-02
Patch compilation time

FabioLuporini:
ASV spotted a regression in compilation time, see spike [here](https://www.devitoproject.org/devito-performance/#codegen.TTI.time_forward) on https://github.com/devitocodes/devito/commit/891bfc4c 

This patch fixes it.

It also speeds-up `filter_ordered` (now `key` only computed once) and adds a simple tree visitor test

### DIFF
--- a/devito/tools/utils.py
+++ b/devito/tools/utils.py
@@ -165,7 +165,14 @@ def filter_ordered(elements, key=None):
             return unordered[np.argsort(inds)].tolist()
         except:
             return sorted(list(set(elements)), key=elements.index)
-    return [e for e in elements if not (key(e) in seen or seen.add(key(e)))]
+    else:
+        ret = []
+        for e in elements:
+            k = key(e)
+            if k not in seen:
+                ret.append(e)
+                seen.add(k)
+        return ret
 
 
 def filter_sorted(elements, key=None):
--- a/devito/ir/iet/visitors.py
+++ b/devito/ir/iet/visitors.py
@@ -536,6 +536,14 @@ class MapNodes(Visitor):
 
 class FindSymbols(Visitor):
 
+    def quick_str(o):
+        """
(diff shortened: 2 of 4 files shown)


devito/tools/utils.py
```diff
@@ -165,7 +165,14 @@ def filter_ordered(elements, key=None):
...
             return sorted(list(set(elements)), key=elements.index)
-    return [e for e in elements if not (key(e) in seen or seen.add(key(e)))]
+    else:
+        ret = []
+        for e in elements:
+            k = key(e)
+            if k not in seen:
+                ret.append(e)
+                seen.add(k)
+        return ret
```
rhodrin, line 175: A faster filtering method (e.g. using `set`) is not in some way applicable here?

rhodrin:
And btw, are compilation times with this patch basically the same or better than before the regression?

devito/tools/utils.py
FabioLuporini: elaborate? we need to preserve ordering btw

FabioLuporini:
> And btw, are compilation times with this patch basically the same or better than before the regression?

TTI-forward would seem mildly better. Definitely not slower

rhodrin (approved):

devito/tools/utils.py
rhodrin: If ordering needs to be preserved then it won't make much difference, but a `sorted(set(mylist), key=mylist.index)` structure type would be a little quicker?

mloubout (approved):

devito/tools/utils.py
FabioLuporini: so the filtering is dictated by the `key`, but in your solution you're filtering based on the elements' `__eq__` (implicitly through `set`), so the outcome would be different right?

PR #1489, opened 2020-11-02
Tensor grid fixup

### LINKED ISSUE #1476: Tensor grid
Tensor.vector grid detection is flawed and need proper visit of elements rather than just first.

mloubout:
Add robustness grid detection in the constructor. Should prevent errors when a `TensorFunction` is created with components that are not devito types.

TODO:
- Cleanup and force return sympy Matrix at first construction as well

fixes #1476

### DIFF
--- a/devito/passes/equations/linearity.py
+++ b/devito/passes/equations/linearity.py
@@ -52,7 +52,19 @@ def _(c, deriv):
 
 @_is_const_coeff.register(sympy.Function)
 def _(c, deriv):
-    return not set(deriv.dims) & c.free_symbols
+    is_const = not set(deriv.dims) & c.free_symbols
+    try:
+        is_const &= not set([d.parent for d in deriv.dims]) & c.free_symbols
+    except AttributeError:
+        pass
+    # Need to check that there isn't any sub/super-dimension of deriv.dims
+    # in c
+    for c in c.free_symbols:
+        try:
+            is_const &= not set(deriv.dims) & {c.parent}
+        except AttributeError:
+            pass
+    return is_const
 
 
 @_is_const_coeff.register(sympy.Expr)
--- a/devito/passes/equations/linearity.py
+++ b/devito/passes/equations/linearity.py
@@ -52,19 +52,9 @@ def _(c, deriv):
 
 @_is_const_coeff.register(sympy.Function)
 def _(c, deriv):
-    is_const = not set(deriv.dims) & c.free_symbols
-    try:
-        is_const &= not set([d.parent for d in deriv.dims]) & c.free_symbols
-    except AttributeError:
-        pass
-    # Need to check that there isn't any sub/super-dimension of deriv.dims
-    # in c
-    for c in c.free_symbols:
-        try:
-            is_const &= not set(deriv.dims) & {c.parent}
-        except AttributeError:
-            pass
-    return is_const
+    c_dims = set().union(*[getattr(i, '_defines', i) for i in c.free_symbols])
+    deriv_dims = set().union(*[d._defines for d in deriv.dims])
+    return not c_dims & deriv_dims
 
 
 @_is_const_coeff.register(sympy.Expr)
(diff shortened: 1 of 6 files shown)


devito/types/basic.py
```diff
@@ -497,8 +497,24 @@ def _new(cls, *args, **kwargs):
...
+            # Filter grid and dimensions
+            grid = {getattr(c, 'grid', None) for c in newobj._mat} - {None}
+            dimensions = {d for c in newobj._mat
+                          for d in getattr(c, 'dimensions', ())} - {None}
+            # If none of the components are devito objects, returns a sympy Matrix
+            if len(grid) == 0 and len(dimensions) == 0:
+                return sympy.ImmutableDenseMatrix(*args)
+            elif len(grid) > 0:
+                dimensions = None
+                grid = grid.pop()
```
FabioLuporini, line 510: you probably wanna assert the uniqueness of the grid? ie `assert len(grids) == 1` ?

devito/types/basic.py
```diff
@@ -497,8 +497,24 @@ def _new(cls, *args, **kwargs):
                 # Constructor if input is list of list as (row, cols, list_of_list)
                 # doesn't work as it expects a flattened.
                 newobj = super(AbstractTensor, cls)._new(args[2])
+
+            # Filter grid and dimensions
+            grid = {getattr(c, 'grid', None) for c in newobj._mat} - {None}
```
FabioLuporini, line 502: when we gather objects, the convention in devito is to use plural names for variables, ie `grids`. It's when you reduce to a single grid, eg through `grids.pop()`, that you assign it to `grid`. But until this point you need to use `grids` as, conceptually, you still don't know how many there are

FabioLuporini (commented):
basically uncontroversial so far

devito/types/basic.py
mloubout: Indeed added

devito/types/basic.py
mloubout: True, typo, fixed.

devito/passes/equations/linearity.py
```diff
@@ -52,7 +52,19 @@ def _(c, deriv):
 
 @_is_const_coeff.register(sympy.Function)
 def _(c, deriv):
-    return not set(deriv.dims) & c.free_symbols
+    is_const = not set(deriv.dims) & c.free_symbols
+    try:
```
FabioLuporini, line 56: Probably you should use `d._defines` rather than the sole parent?

Then _maybe_ you could rewrite everything simply as
```
c_dims = set().union(*[i._defines for i in c.free_symbols if isinstance(i, Dimension)])
deriv_dims = set().union(*[d._defines for d in deriv.dims])
is_const = not c_dims & deriv_dims
```

? 

not sure if I'm missing something though

FabioLuporini (commented):
you probably forgot to push the new test file

devito/passes/equations/linearity.py
mloubout: yeah that works too, is replaced the isinstance by a getaatr on `defines` to avoid importing `Dimension` inside  `is_const` but works as expected.

devito/passes/equations/linearity.py
```diff
@@ -52,7 +52,9 @@ def _(c, deriv):
 
 @_is_const_coeff.register(sympy.Function)
 def _(c, deriv):
-    return not set(deriv.dims) & c.free_symbols
+    c_dims = set().union(*[getattr(i, '_defines', i) for i in c.free_symbols])
```
FabioLuporini, line 55: `isinstance(i, Dimension)` makes it more clear. `getattr` should be avoided whenever possible as it hides information away

FabioLuporini (commented):
LGTM

devito/passes/equations/linearity.py
FabioLuporini: also, it'd seem that if you don't change it to use `isinstance` you would end up adding other non-Dimension objects to `c_dims`?

mloubout: I don't think these extras are problematic, it won't e a match anyway. I just prefer `getattr` to local import that's all but can change.

FabioLuporini: hang on, why would it be a local import?

mloubout: There is a circular dependency somewhere (didn't go to the bottom of it just switched back to isinstance) so can't import Dimension here unless inside the function.

FabioLuporini (approved):

FabioLuporini:
Merged.

PR #1492, opened 2020-11-04
Fix MPI in Dockerfile.nvidia

kenhester:
Update to NVIDIA HPC SDK 20.9
Update to MPI 4.0.2 in container
Enable UCX support in MPI4PY
Resolve missing libraries


docker/Dockerfile.nvidia
```diff
@@ -6,19 +6,22 @@
...
 RUN apt-get update -y && \
     apt-get install -y -q \
         apt-utils \
+        vim \
         curl \
         mpich libmpich-dev && \    
     curl -sL https://deb.nodesource.com/setup_12.x | bash - && \
     apt-get install -y -q \
         nodejs \
+        liblapack-dev  \
```
FabioLuporini, line 23: out of curiosity, why are these needed?

FabioLuporini (commented):
I don't think I have major objections about splitting `requirements-optional` into `requirements-optional` and `requirements-mpi`, but why would it be necessary here?

FabioLuporini (approved):

rhodrin (approved):

PR #1493, opened 2020-11-05
Outhalo size warning

### LINKED ISSUE #1466: Erronous halo size warning
Running e.g. `DEVITO_MPI=1 mpirun -np 4 python acoustic_example.py` results in the following warnings:
```
A space order of 4 and a halo size of Size(left=4, right=4) has been
                                 set but the current rank (2) has a domain size of
                                 only 65
A space order of 4 and a halo size of Size(left=4, right=4) has been
                                 set but the current rank (1) has a domain size of
                                 only 65
```
but this shouldn't be the case.

Note: no warning is produced for `-np 2`.

### LINKED ISSUE #1483: Halo update outside time iterations
This is another situation where halo updates are misplaced, similar to #1459, but this time outside of time-iteration loop. This is not just invalidating the halo update, but also results in compilation error, since some indices (eg. t0) passed as arguments to halo update are declared at the loop definition. 

MFE:
```
from devito import (Eq, Operator, Grid, Function, TimeFunction)

grid = Grid(shape=(10,10))
x, y = grid.dimensions
t = grid.stepping_dim

f = Function(name="f", grid=grid, space_order=4)
g = TimeFunction(name="g", grid=grid, space_order=4, time_order=2)
h = TimeFunction(name="h", grid=grid, space_order=4, time_order=2)

eqn = []
eqn.append( Eq(f, h + 1) )
eqn.append( Eq(g, f) )
eqn.append( Eq(h.forward, h + g[t,x+1,y+1] + 1) )

op = Operator(eqn)

op.apply(time_M=10)
```
Comments

* This error is also affecting PR #1406

rhodrin:
Tiny PR to fix #1466 .

### DIFF
--- a/devito/types/dense.py
+++ b/devito/types/dense.py
@@ -329,8 +329,11 @@ class DiscreteFunction(AbstractFunction, ArgProvider, Differentiable):
                 if not self._distributor.is_boundary_rank:
                     warning(warning_msg)
                 else:
-                    for i, j, k, l in zip(left[-len(self._distributor.dimensions):],
-                                          right[-len(self._distributor.dimensions):],
+                    left_dist = [i for i, d in zip(left, self.dimensions) if d
+                                 in self._distributor.dimensions]
+                    right_dist = [i for i, d in zip(right, self.dimensions) if d
+                                  in self._distributor.dimensions]
+                    for i, j, k, l in zip(left_dist, right_dist,
                                           self._distributor.mycoords,
                                           self._distributor.topology):
                         if l > 1 and ((j > 0 and k == 0) or (i > 0 and k == l-1)):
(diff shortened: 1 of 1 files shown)


mloubout (commented):
Looks fine, any way to test it?

FabioLuporini:
I'm not sure I understand the fix. `self._decomposition` should have been already of the proper size, and so `left` and `right`, given that `left = [.... in zip(self._decomposition, ...)]`. Why do you need to further slice it?

rhodrin:
> 
> 
> I'm not sure I understand the fix. `self._decomposition` should have been already of the proper size, and so `left` and `right`, given that `left = [.... in zip(self._decomposition, ...)]`. Why do you need to further slice it?

`left` and `right` also contain the size of the time halo which we wan't to ignore for this spatial check.

rhodrin:
> Looks fine, any way to test it?

I was thinking about that, but came to the conclusion that a specific test was overkill. The incorrect warning currently (<- current master) shows up for the mpi seismic examples, but I'm not sure we wan't to play around with them or add a 'dummy' example that currently erronously produces the warning?

FabioLuporini (approved):

FabioLuporini:
Thanks, merged.

PR #1494, opened 2020-11-05
Fix halo update hoisting

### LINKED ISSUE #1483: Halo update outside time iterations

Leitevmd:
PR to fix #1483


devito/passes/iet/mpi.py
```diff
@@ -66,14 +66,15 @@ def _hoist_halospots(iet):
...
 
-    def rule0(dep, candidates):
+    def rule0(dep, candidates, loc_dims):
         # E.g., `dep=W<f,[x]> -> R<f,[x-1]>` and `candidates=(time, x)` => False
         # E.g., `dep=W<f,[t1, x, y]> -> R<f,[t0, x-1, y+1]>`, `dep.cause={t,time}` and
         #       `candidates=(x,)` => True
         return (all(d in dep.distance_mapper for d in candidates) and
-                not dep.cause & candidates)
+                not dep.cause & candidates and
+                not loc_dims & candidates)
```
FabioLuporini, line 75: this looks about the right thing to do

FabioLuporini (commented):
So turns out you have a nice patch after all?

Do all tests pass?

Can you add all of the new tests that you've shown in #1483 ?

devito/passes/iet/mpi.py
```diff
@@ -92,13 +93,15 @@ def rule1(dep, candidates):
         for hs in halo_spots:
             hsmapper[hs] = hs.halo_scheme
 
-            for f in hs.fmapper:
+            for f, (loc_indices, _) in hs.fmapper.items():
                 for n, i in enumerate(iters):
                     candidates = set().union(*[i.dim._defines for i in iters[n:]])
+                    loc_dims = frozenset().union([q for d in loc_indices
```
FabioLuporini, line 99: This can be moved outside of this inner loop -- and thus be computed once and for all -- as soon as `loc_indices` is available? ie, first statement of the `for f, (loc_indices, ...` loop

tests/test_mpi.py
```diff
@@ -1635,6 +1635,33 @@ def test_haloupdate_same_timestep(self):
 
         assert np.all(v.data_ro_domain[-1, :, 1:-1] == 6.)
 
+    @pytest.mark.parallel(mode=2)
+    def test_haloupdate_same_timestep_v2(self):
+        """
+        Similar to test_haloupdate_same_timestep, but switching the expression that
+        writes to subsequent time step. Also checks halo update call placement.
```
FabioLuporini, line 1642: Please add:

```
"""
Similar to ... .

MFE for issue #1483 
"""
```

FabioLuporini (changes requested):

devito/passes/iet/mpi.py
Leitevmd: My bad!!

FabioLuporini (approved):
Perfect, thanks!

FabioLuporini:
Thanks, merged

PR #1495, opened 2020-11-06
Fixup _drop_halospots

### LINKED ISSUE #1490: cire advanced option does not generate MPI code
the operator in the MFE below does not generate MPI code when run with `DEVITO_MPI=1`

```
from devito import Grid, TimeFunction, Eq, Operator
grid = Grid(shape=(41,41,11))
p0 = TimeFunction(name='p0', grid=grid, time_order=2, space_order=8)
update_p = (p0.dx).dx + (p0.dy).dy + (p0.dz).dz 
op = Operator([stencil_p0, src_term], name='MFE', 
    opt=('advanced', {'cire-repeats-sops': 9, 'cire-rotate': True}))
op.apply(time_M=10)
```

FabioLuporini:
fixes #1490

### DIFF
--- a/devito/passes/iet/mpi.py
+++ b/devito/passes/iet/mpi.py
@@ -7,7 +7,7 @@ from devito.ir.iet import (Call, Expression, HaloSpot, Iteration, FindNodes,
 from devito.ir.support import PARALLEL, Scope
 from devito.mpi import HaloExchangeBuilder, HaloScheme
 from devito.passes.iet.engine import iet_pass
-from devito.tools import filter_sorted, generator
+from devito.tools import as_mapper, filter_sorted, generator
 
 __all__ = ['optimize_halospots', 'mpiize']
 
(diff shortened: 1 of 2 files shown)


devito/passes/iet/mpi.py
```diff
@@ -39,7 +39,13 @@ def _drop_halospots(iet):
...
     # If a HaloSpot Dimension turns out to be SEQUENTIAL, then the HaloSpot is useless
     for hs, iterations in MapNodes(HaloSpot, Iteration).visit(iet).items():
-        if any(i.is_Sequential for i in iterations if i.dim.root in hs.dimensions):
+        dmapper = as_mapper(iterations, lambda i: i.dim.root)
+        flag = False
+        for d, v in dmapper.items():
+            if d in hs.dimensions and all(i.is_Sequential for i in v):
+                flag = True
+                break
+        if flag:
```
mloubout, line 48: Do you need that flag?

```
for d, v in dmapper.items():
            if d in hs.dimensions and all(i.is_Sequential for i in v):
                mapper[hs].update(set(hs.functions))
                break
```

doesn't this do the same?

mloubout (approved):

devito/passes/iet/mpi.py
FabioLuporini: oh yes ! I'm an idiot -- basically some leftover from some more complicated code I had before. Just running tests locally with your version. Will push the update if all remains green (as it should!)

PR #1497, opened 2020-11-06
Fix custom coefficients on staggered grids

EdCaunt:


devito/finite_differences/coefficients.py
```diff
@@ -217,27 +227,41 @@ def generate_subs(i):
         return rules
 
 
+def extract_dim(dim):
```
mloubout, line 230: I think this should be moved into `symbolics/insepction.py` as a `retreive_dimension`

mloubout (commented):
One minor comment, looks good and nice tests

devito/finite_differences/coefficients.py
EdCaunt: Done

devito/symbolics/inspection.py
```diff
@@ -63,6 +63,22 @@ def count(exprs, query):
     return dict(mapper)
 
 
+def retrieve_dim(dim):
```
mloubout, line 66: Oh sorry typed to quick, this should be in `search.py` as 
```
def retreive_dimension(exprs, mode='all', deep=False):
    """Shorthand to retrieve the dimensions in ``exprs``."""
    return search(exprs, q_dimension mode, 'dfs', deep)
```
 (rename `retreive_dimension`)  and then need the `q_dimension` in `queries.py` 

```
def q_dimension(expr):
    from devito.types.dimension import Dimension
    return isinstance(expr, Dimension)
```

devito/finite_differences/coefficients.py
```diff
@@ -176,6 +185,7 @@ def generate_subs(i):
             deriv_order = i.deriv_order
             function = i.function
             dim = i.dimension
+            offset_dim = i.offset_dimension
```
rhodrin, line 188: Why are `dim` and `offset_dim` both required?

devito/finite_differences/coefficients.py
```diff
@@ -220,24 +231,26 @@ def generate_subs(i):
 def default_rules(obj, functions):
 
     def generate_subs(deriv_order, function, dim):
+        extracted_dim = retrieve_dim(dim)
```
rhodrin, line 234: The naming here could do with a review, `extracted_dim = retrieve_dim(dim)` is a bit weird.
`dim = retrieve_dimension(some_suitable_name)` would be better.

rhodrin:
Looks good. Just a couple of comments above + Mathias's comment to address.

Did you test everything works ok with `TimeFunction`'s? A `TimeFunction` based test should also be added. Also a test where it's checked that coefficients revert to the correct default coefficients for staggered grids when no substitutions are provided.

EdCaunt (commented):
Made the fixes

devito/finite_differences/coefficients.py
rhodrin: So `offset_dim` is the `Add` object? If so, should probably have it's name changed to something related to `index`. And is this `None` when it's not required?

EdCaunt: It's equivalent to `.dimension` on non staggered grids, but on staggered grids it also includes the half-increment offset introduced by the staggering. `.dimension` would always be `x` regardless as to whether the function is staggered in the x direction or not, but `.offset_dimension` could be `x` or `x+h_x` depending on the stagger

rhodrin: The point is that you should never have repeated information and hence the routine should be updated such that this is the case.

devito/finite_differences/coefficients.py
```diff
@@ -220,24 +231,26 @@ def generate_subs(i):
 def default_rules(obj, functions):
 
     def generate_subs(deriv_order, function, dim):
+        retrieved_dim = retrieve_dimension(dim)[0]
```
rhodrin, line 234: Looks to me like the `[0]` should be in the function itself and not here?

rhodrin: But also, the naming still needs changing, e.g. `dim = retrieve_dim(index)` since `dim` is not always a `dim`.

devito/finite_differences/coefficients.py
EdCaunt: Wouldn't that require extra conditionals though? Since sometimes `dimension` is required, but other times both `dimension` and `dimension+spacing/2` are required, in which case the latter simply matches the former. This avoids the need to check the value of `offset_dim`, and act accordingly if it is None. Additionally, intuitively, `dimension+offset` where `offset` is 0 is just `dimension`, not `None`

devito/finite_differences/coefficients.py
EdCaunt: Made the change to the naming

mloubout (approved):

FabioLuporini (approved):

rhodrin (approved):

FabioLuporini:
Merged, thanks.

Issue #1498, opened 2020-11-06
Naming of global vs local methods


rhodrin:
Many concepts in Devito are 'global', but certain properties of `Function`'s are not e.g. `f.shape` returns the local shape - `f.shape_global` is also present but elsewhere the opposite is true, e.g. for distributed `SubDomain`'s I believe we have `shape_local`.

Since our philosophy is that moving from non-distributed to distributed is simply (as close to possible) flicking a switch we should update these: `Function.shape` -> `Function.shape_local`, `Function.shape_global` -> `Function.shape` etc.

mloubout:
I think that makes sense to me, would make the API more uniform from user point of view

FabioLuporini:
indeed!

Now the question is what would break if we did that

* internally, and
* in all user codes running with MPI

There's a non negligible risk we would break some code

One thing we could do is finalising some of the big things we're working on (eg GPU support, Functions on subdomains), change the API, and burn Devito 5. (@ggorman )

A major change in version is probably necessary here

rhodrin:
From some `grep`'ing, would be a 'fairly' quick and easy change on our side - a few things need updating and a few tests fixing but wouldn't be too painful.

But as you mention, I think the big question is whose mpi code would this break?

PR #1499, opened 2020-11-06
Function globabl size method

rhodrin:
Tiny PR adding a quick method to `Function`'s for returning `size_global`.

But see #1498 re. the naming.


FabioLuporini (approved):
Uncontroversial

PR #1504, opened 2020-11-11
ir: Fixup _lower_stepping_dims

### LINKED ISSUE #1486: Error with wrong indices in loops
The following MFE presents some cases in which the indexes used to access functions are not defined correctly. All these cases produce more than one time-iteration loop and that's when the error appears. 

```python
from devito import (Eq, Operator, Grid, TimeFunction)

grid = Grid(shape=(10,10))
x, y = grid.dimensions
t = grid.stepping_dim

f = TimeFunction(name="f", grid=grid, space_order=2, time_order=2) 
g = TimeFunction(name="g", grid=grid, space_order=2, time_order=2)

eqn = []
eqn.append( Eq( f.forward, f + 1 ) )

# Case 1 - Compilation error (t0 undeclared on the second loop, t0+2 should be replaced by t2)
# eqn.append( Eq( g.forward, f[t+2,x,y] ) )

# Case 2 - Segfault (t0 is declared, but t0+3 attempts to access wrong index, t0+3 should be replaced by t2)
eqn.append( Eq( g, f[t+3,x,y] ) ) 

# Case 3 - No error (t0+2 should be replaced by t2)
# eqn.append( Eq( g, f[t+2,x,y] ) )

op = Operator(eqn)

op.apply(time_M=10)
```

It seems substitutions of `SteppingDimensions` by `ModuloDimensions` in  `_lower_stepping_dims(iet)` (devito/ir/iet/scheduler.py) are being done in a misleading manner: All expressions in `iet` have their indices replaced with `uindices` (eg. `t+2` -> `t0+2`) available in the first iteration tree, which in these examples is the first time-iteration loop. That way, when we get to the second time-iteration loop, indices that were supposed to be replaced (eg. `t+2` -> `t2`) are not found because they have already been replaced (eg. `t+2` -> `t0+2`) and so they remain that way in the code, resulting in compilation or execution errors. This is because the second time-iteration loop may not even have that `t0` indice (Case 1) or if it has `t0` it can attempt to access wrong index (Case 2) since indexes are miscalculated (eg. `t0+2 != t2`, because `t0 = (time)%(3)`, `t2 = (time+2)%(3)`, so `t0+2` could reach 4, out-of-bounds, as soon as t0==2). Also, that could even not produce any errors at all, but still mess up the numerical result (Case 3).

Leitevmd:
This makes _lower_stepping_dims to replace indices by iteration.
Fix #1486

### DIFF
--- a/devito/ir/iet/scheduler.py
+++ b/devito/ir/iet/scheduler.py
@@ -86,8 +86,8 @@ def _lower_stepping_dims(iet):
             mapper = {d.origin: d for d in v}
             rule = lambda e: e.function.is_TimeFunction and e.function._time_size == k
             replacer = lambda e: xreplace_indices(e, mapper, rule)
-            i = XSubs(replacer=replacer).visit(i)
-        subs[root] = i
+            root = XSubs(replacer=replacer).visit(root)
+        subs[i] = root
 
     return Transformer(subs).visit(iet)
 
--- a/devito/ir/iet/scheduler.py
+++ b/devito/ir/iet/scheduler.py
@@ -2,7 +2,7 @@ from collections import OrderedDict
 
 from devito.ir.iet import (Expression, Increment, Iteration, List, Conditional,
                            Section, HaloSpot, ExpressionBundle, FindNodes, FindSymbols,
-                           XSubs)
+                           XSubs, Transformer)
 from devito.symbolics import IntDiv, xreplace_indices
 from devito.tools import as_mapper, timed_pass
 from devito.types import ConditionalDimension
@@ -69,6 +69,7 @@ def _lower_stepping_dims(iet):
 
     u[t1, x] = u[t0, x] + 1
     """
+    subs = {}
     for i in FindNodes(Iteration).visit(iet):
         if not i.uindices:
             # Be quick: avoid uselessy reconstructing nodes
@@ -80,13 +81,15 @@ def _lower_stepping_dims(iet):
         # two different calls to `xreplace`
         mindices = [d for d in i.uindices if d.is_Modulo]
         groups = as_mapper(mindices, lambda d: d.modulo)
+        i_org = i
         for k, v in groups.items():
             mapper = {d.origin: d for d in v}
             rule = lambda i: i.function.is_TimeFunction and i.function._time_size == k
             replacer = lambda i: xreplace_indices(i, mapper, rule)
-            iet = XSubs(replacer=replacer).visit(iet)
+            i = XSubs(replacer=replacer).visit(i)
+        subs[i_org] = i
 
-    return iet
+    return Transformer(subs).visit(iet)
 
 
 def _lower_conditional_dims(iet):
--- a/devito/ir/iet/scheduler.py
+++ b/devito/ir/iet/scheduler.py
@@ -81,13 +81,13 @@ def _lower_stepping_dims(iet):
         # two different calls to `xreplace`
         mindices = [d for d in i.uindices if d.is_Modulo]
         groups = as_mapper(mindices, lambda d: d.modulo)
-        i_org = i
+        root = i
         for k, v in groups.items():
             mapper = {d.origin: d for d in v}
-            rule = lambda i: i.function.is_TimeFunction and i.function._time_size == k
-            replacer = lambda i: xreplace_indices(i, mapper, rule)
+            rule = lambda e: e.function.is_TimeFunction and e.function._time_size == k
+            replacer = lambda e: xreplace_indices(e, mapper, rule)
             i = XSubs(replacer=replacer).visit(i)
-        subs[i_org] = i
+        subs[root] = i
 
     return Transformer(subs).visit(iet)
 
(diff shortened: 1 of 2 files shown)


tests/test_operator.py
```diff
@@ -213,6 +213,33 @@ def test_multiple_steppers(self, expr, exp_uindices, exp_mods):
...
+        f = Function(name="f", grid=grid, space_order=4)
+        g = Function(name="g", grid=grid, space_order=4)
+        h = TimeFunction(name="h", grid=grid, space_order=4, time_order=2)
+
+        eqn = [Eq(f, h + 1), Eq(g, f),
+               Eq(h.forward, h + g + 1)]
+
+        op = Operator(eqn)
+
+        for iter in [i for i in FindNodes(Iteration).visit(op) if i.dim.is_Time]:
```
FabioLuporini, line 233: Maybe you could also assert against the number of expected Iterations? w/o the patch, we would obtain more Iterations right? that'd be a simple yet effective check

devito/ir/iet/scheduler.py
```diff
@@ -80,13 +81,15 @@ def _lower_stepping_dims(iet):
         # two different calls to `xreplace`
         mindices = [d for d in i.uindices if d.is_Modulo]
         groups = as_mapper(mindices, lambda d: d.modulo)
+        i_org = i
         for k, v in groups.items():
             mapper = {d.origin: d for d in v}
             rule = lambda i: i.function.is_TimeFunction and i.function._time_size == k
```
FabioLuporini, line 87: Unrelated to this PR, but since you're editing this code.... instead of `lambda i` could you name it differently , since `i` is already used as loop iteration variable? Same for the line below

devito/ir/iet/scheduler.py
```diff
@@ -80,13 +81,15 @@ def _lower_stepping_dims(iet):
         # two different calls to `xreplace`
         mindices = [d for d in i.uindices if d.is_Modulo]
         groups = as_mapper(mindices, lambda d: d.modulo)
+        i_org = i
```
FabioLuporini, line 84: Nitpicking: throughout Devito, to identify and refer to the root of an IET we typically use the variable name "root" ie `root = i`

tests/test_operator.py
Leitevmd: The patch doesn't change the number of iterations. It just fix what is going to happen if there is more than one. The MFE would produce two iterations for time dimension with or without the patch. 

We could assert 2 time iterations, but I think this is more related to cluster fusion in `_specialize_clusters` (e.g. devito /core/cpu.py) which I guess could vary for some implementation. Let me know if you still think it's worth it.

devito/ir/iet/scheduler.py
Leitevmd: Sure!

devito/ir/iet/scheduler.py
Leitevmd: Just to make sure: Even if `i` does not refer to the root of `iet`? For example we have `iet` argument, so we would use root to refer to these iterations?
```
<List (0, 3, 0)>
  <[affine,sequential] Iteration time::time::(time_m, time_M, 1)>
  	...
  <TimedList (2, 1, 2)>
	...
  <[affine,sequential] Iteration time::time::(time_m, time_M, 1)>
  	...
```

tests/test_operator.py
FabioLuporini: Gotcha.

Any chance we can then assert the expected computed value? since this would be the only test of this kind, it could be quite useful

Leitevmd: Done. Very simple assertion. Re-check please!

Leitevmd:
As reported [here](https://devitocodes.slack.com/archives/C7JMLMSG0/p1605621700409000), tests are failing due to a deprecated command which needs to be replaced.

devito/ir/iet/scheduler.py
```diff
@@ -80,13 +81,15 @@ def _lower_stepping_dims(iet):
...
         groups = as_mapper(mindices, lambda d: d.modulo)
+        root = i
         for k, v in groups.items():
             mapper = {d.origin: d for d in v}
-            rule = lambda i: i.function.is_TimeFunction and i.function._time_size == k
-            replacer = lambda i: xreplace_indices(i, mapper, rule)
-            iet = XSubs(replacer=replacer).visit(iet)
+            rule = lambda e: e.function.is_TimeFunction and e.function._time_size == k
+            replacer = lambda e: xreplace_indices(e, mapper, rule)
+            i = XSubs(replacer=replacer).visit(i)
```
FabioLuporini, line 89: I just remembered something. Overriding iteration variables in Python can lead to all sort of breakages. 

How about instead:

```
for ...:
    ...
    root = XSubs(replacer=replacer).visit(root)
subs[i] = root
```

?

FabioLuporini (changes requested):

devito/ir/iet/scheduler.py
Leitevmd: Yeah, that really makes sense!!

tests/test_operator.py
FabioLuporini: perfect, thanks

FabioLuporini (approved):

PR #59, opened 2016-07-29
Point interpolation via `Iteration` blocks

mlange05:
This merge is primarily intended to fix issue #48 by extending the Propagator API to insert user-defined loops over low-level "indexed access" expressions to perform sparse point interpolation. For this purpose a new class `Iteration` is introduced that encapsulates a single loop over a defined dimension. In more detail this PR brings:
* A new symbolic data object `CoordinateData` that gets attached to `PointData` objects and stores the coordinates of the sparse points.
* Fixes to the Operators `input_params` and `output_params` lists, including a more general way of avoiding duplication in the function signature. It also makes the call to `initialize()` on input params optional.
* New `Iteration` class that encapsulates a single loop iteration over a defined iteration space. It auto-generates the loop code around the given stencil(s) and is accepted as input to the Propagator, either as a normal stencil, or a `time_loop_stencil`.
* Small hacks to the `CodePrinter` to enable custom functions `INT` and `FLOAT` for data type conversions. This is required to get around SymPy limitations.
* `SourceLike` objects for FWI now provide symbolic expressions for grid-to-point and point-to-grid interpolation, which in turn now get wrapped by an `Iteration` over the point dimension `p` to insert loop-based interpolation code into all FWI operators. I leave TTI for another PR...

Please note that the new loop-based way of performing interpolation adds flops to the time loop that we previously pre-computed in Python. In turn it brings down the compilation overhead for large data sets though and is more general; and the point interpolation should still be sparse enough to not be a performance bottleneck.

PR #1507, opened 2020-11-14
make test tools less random

ggorman:
`tests/test_tools.py::test_sorting` was failing on my local machine. It is not clear to me that the test makes sense unless you are choosing between two different implementations for performance reasons. I suggest @mloubout takes a look at this lines of code (git blame) - my suggested fix is to convert this to a warning and changed to higher precision timer.

### DIFF
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -23,12 +23,19 @@ def test_toposort(elements, expected):
 
 def test_sorting():
     key = lambda x: x
-    array = np.random.rand(10000)
+
+    # Need predictable random sequence or test will
+    # have inconsistent behaviour results between tests.
+    np.random.seed(0)
+    array = np.random.randint(-1000, 1000, 10000)
+
     t0 = time.time()
-    sort_key = filter_ordered(array, key=key)
+    for _ in range(100):
+        sort_key = filter_ordered(array, key=key)
     t1 = time.time()
-    sort_nokey = filter_ordered(array)
+    for _ in range(100):
+        sort_nokey = filter_ordered(array)
     t2 = time.time()
-    # This one is slightly faster
-    assert t2 - t1 < .5 * (t1 - t0)
+
+    assert t2 - t1 < 0.5 * (t1 - t0)
     assert sort_key == sort_nokey


mloubout (commented):
Why/when did this slow down.

ggorman:
> Why/when did this slow down.

Nothing changed other than the fact that if I run this on my desktop:
Linux kimogila 5.4.0-53-generic #59-Ubuntu SMP Wed Oct 21 09:38:44 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
Python 3.8.5
gcc version 9.3.0 (Ubuntu 9.3.0-17ubuntu1~20.04) 

I am not convinced this was a valid test in the first place - but I opted to set it to a warning rather than deleting this part of the test.

mloubout:
It's odd it fails on your maching, `no_key` should be like 10 times faster. Did `nokey` get slower or `key` get faster?

ggorman:
> It's odd it fails on your maching, `no_key` should be like 10 times faster. Did `nokey` get slower or `key` get faster? THis is there because we were relying too much on `key` which was making src/rec processing crazy slow, this test was there to make sure the no-key sorting, that defaults to numpy since this change, was indeed faster and this was not useless extra sorting branches.

So it might be the underlying packages as I'm running a more recent linux distribution than is used in the tests (ubuntu 10.04, numpy 1.17.4).

I'm not sure it's worth investigating a lot of time diving into this - would need to run cprofile for the two cases on different systems. For this PR all that is to be decided is whether this really should be a failing test or warning. I'm advocating for the latter because I don't see this as a effective means of performance regression - and this change in performance is down to the env rather than having anything got to do with Devito.

mloubout:
>I'm not sure it's worth investigating a lot of time diving into this 

I disagree. The sorting speed was a major issue with MPI so if it got slower we need to know and fix it.

ggorman:
> > I'm not sure it's worth investigating a lot of time diving into this
> 
> I disagree. The sorting speed was a major issue with MPI so if it got slower we need to know and fix it.

What part of what I said are you disagreeing with?

Why do you think one method got slower? How do you know it's not that the other implementation has gotten faster?

Looking at more detail, this is what you are testing:
`t2 - t1 < 0.5 * (t1 - t0)`
where `t1-t0` is the wall time for `sort_key = filter_ordered(array, key=key)` and where `t2 - t1` is the wall time for `sort_nokey = filter_ordered(array)`. So you are expecting the nokey variant be to at least twice as fast as the key variant. Now looking at the actual timings i get on my machine:

0.5*(t1 - t0) =  0.001161813735961914
(t2 - t1)     =  0.0014281272888183594

So this test is failing because of a difference 0.00026 seconds. ... this is not the kind of thing you can measure to any reliability. I tested this manually in a loop - some times it fails and other times it passes.

ggorman:
If we rewrite the test to run a multiple experiments then we can mostly eliminate the jitter. But it remains a fragile test that needs rethinking.
 
def test_sorting():
    key = lambda x: x
    array = np.random.rand(10000)
    
    start = time.perf_counter()
    for i in range(100):
        sort_key = filter_ordered(array, key=key)
    t0 = time.perf_counter() - start

    start = time.perf_counter()
    for i in range(100):
        sort_nokey = filter_ordered(array)
    t1 = time.perf_counter() - start

    assert t1 < 0.5 * t0
    assert sort_key == sort_nokey

mloubout:
>Why do you think one method got slower? How do you know it's not that the other implementation has gotten faster?

That's my point, we need to know which one it is. If `sort_key = filter_ordered(array, key=key)` got faster then it's perfectly fine, if `sort_nokey = filter_ordered(array)` got slower enough to make this fail then we have a problem. I'm not saying this doesn't need update/fix, I'm just saying we need to know which of the two cases it is.

ggorman:
Damn rabbit holes...

So - the test as it stands is crap and needs fixing... period. 

However, I took a look at the performance issue were trying to solve. It boils down to the fact that numpy.unique give you sorted values whether you asked for them or not and in this particular case you needed to keep the ordering. The popular solution to this is to simply use pandas.unique instead which does not sort and has the added benefit of being much faster due to it's lower computational complexity. 

This does mean we incur pandas as a core dependency (previously it was in requirements-optional.txt. However, I believe that this is justified in that pandas is pretty much everywhere these days and there are numerous opportunities where we could pandas methods rather than numpy when they are superior.

mloubout (commented):
I'm happy with pandas, thought about adding it too wasn't sure if was worth it.

Another nitpicky comment sorry.

ggorman:
> I'm happy with pandas, thought about adding it too wasn't sure if was worth it.
> 
> Another nitpicky comment sorry.

Your nitpicking ultimately led us to find the source of the original problem and propose a potential better fix. So don't say sorry when you are right!

If this PR is merged then we should look to see where else in Devito there might be performance/feature opportunities. I seem to remember that I also considered it for resampling data.

devito/tools/utils.py
```diff
@@ -161,8 +162,7 @@ def filter_ordered(elements, key=None):
     seen = set()
     if key is None:
         try:
-            unordered, inds = np.unique(elements, return_index=True)
-            return unordered[np.argsort(inds)].tolist()
+            return pd.unique(elements.ravel()).tolist()
```
mloubout, line 165: Putting back my comment since seems to have not saved it. THis shouldn't use `ravel` since we can have 
- lists
- array of the form `[ (1,2), (3,4), (1,2)]`
as input and `ravel` doesn't work on lists and would flatten the tuples in the second case.

FabioLuporini (changes requested):
I don't think we should add a huge dependency such as pandas to improve a single line of code

FabioLuporini (changes requested):
The performance aspect of this test should be moved to ASV devito

FabioLuporini:
@ggorman as we agreed yesterday, we should split the performance side of this test and move it to an ASV test

ggorman:
@FabioLuporini I stepped back from that direction as I later realised that the test was more fundamentally flawed. I rewrote the test and rebased on master. 

Details:
    Updated test to use an array of random integers with a fixed seed rather
    than random floats with random seed to that the test:
    * is deterministic between runs
    * duplicate keys are in fact created so unique actually does something.
    
    Additionally, the tic/toc timings are now over multiple calls to the
    function so that the variance is reduced.

FabioLuporini:
I think, however, it was still a good idea to move the timing part to ASV

ggorman:
Are you saying that should be part of this PR?

FabioLuporini:
yeah why not it's a small PR after all

FabioLuporini (approved):

PR #1509, opened 2020-11-17
Remove use of set-env from workflows

rhodrin:
Updating various workflows owing to the [deprecation](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/) of `set-env`.


ggorman (approved):

ggorman:
GTG - This is an important security update.

PR #1510, opened 2020-11-17
arch: fix archinfo and misc for ARM

jaimesouza:
This PR fixes some issues in archinfo.py and misc.py that generated errors when running on ARM (Isambard).

### DIFF
--- a/devito/passes/iet/misc.py
+++ b/devito/passes/iet/misc.py
@@ -6,6 +6,7 @@ from devito.ir.iet import (List, Prodder, FindNodes, Transformer, make_efunc,
                            compose_nodes, filter_iterations, retrieve_iteration_tree)
 from devito.passes.iet.engine import iet_pass
 from devito.tools import flatten, is_integer, split
+from devito import archinfo
 
 __all__ = ['avoid_denormals', 'hoist_prodders', 'relax_incr_dimensions']
 
@@ -18,7 +19,7 @@ def avoid_denormals(iet):
     are normally flushed when using SSE-based instruction sets, except when
     compiling shared objects.
     """
-    if iet.is_ElementalFunction:
+    if iet.is_ElementalFunction or isinstance(archinfo.get_platform(), archinfo.Arm):
         return iet, {}
 
     header = (cgen.Comment('Flush denormal numbers to zero in hardware'),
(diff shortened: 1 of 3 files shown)


devito/passes/iet/misc.py
```diff
@@ -18,7 +19,7 @@ def avoid_denormals(iet):
     are normally flushed when using SSE-based instruction sets, except when
     compiling shared objects.
     """
-    if iet.is_ElementalFunction:
+    if iet.is_ElementalFunction or isinstance(archinfo.get_platform(), archinfo.Arm):
```
FabioLuporini, line 22: The compilation passes (`iet_pass`) are independent of the underlying architecture.

If you don't want `avoid_denormals` on Arm (even though this appears to be a mere work around) the way to go is to create a suitable compilation pipeline for Arm, which is identical to Intel except for the lack of `avoid_denormals`.

There already exists one such Arm compilation pipeline: https://github.com/devitocodes/devito/blob/master/devito/core/__init__.py#L31 -- as you can see, named `ArmOperator`.

You should go to ArmOperator, intercept (override) `_specialize_iet` so as to drop `avoid_denormals`

FabioLuporini (changes requested):
@rhodrin maybe time to add an ARM action to our CI workflows?

FabioLuporini (approved):
LGTM. 

@georgebisbas can you try this patch?

@rhodrin can we add a free GH Actions runner?

georgebisbas:
> @georgebisbas can you try this patch?

Yes

georgebisbas:
I can confirm that the patch works and I could execute
acoustic and elastic operator.

mloubout:
>can we add a free GH Actions runner?

You mean you want a `CI-arm`? That easily doable or can be added to `CI-cpu` but there is no ARM chips on GH actions.

georgebisbas:
> > can we add a free GH Actions runner?
> 
> You mean you want a `CI-arm`? That easily doable or can be added to `CI-cpu` but there is no ARM chips on GH actions.

@mloubout : https://github.blog/changelog/2019-12-03-github-actions-self-hosted-runners-on-arm-architectures/
is that of use?

georgebisbas:
Closed as superseded by #1515

PR #1511, opened 2020-11-18
builtins: initialize FD halo

### LINKED ISSUE #1177: Reorganize builtins
- create builtins subdirectory
- sort functions into `builtins/maths.py` , `builtins/misc.py`
- potentially, split `test_builtins.py` accordingly: `test_builtins_maths.py` `test_builtins_misc.py`
- add rand for `SparseFunctions`, similar to `norm`, `mmin`, `mmax`

### LINKED ISSUE #1250: Non-padding of the halo could cause incorrect computed values

mloubout:
- Fixes #1250 
- Fixes #1177
- Adds shifted div/grad

### DIFF
--- a/.github/workflows/tutorials.yml
+++ b/.github/workflows/tutorials.yml
@@ -67,7 +67,7 @@ jobs:
       if: runner.os == 'macOS'
       run: |
         if [ "${{ matrix.compiler }}" = "gcc-9" ]; then
-          brew install gcc
+          brew install gcc@9
         else
           sudo xcode-select -s /Applications/Xcode_12.2.app/Contents/Developer
         fi
--- a/devito/builtins.py
+++ b/devito/builtins.py
@@ -229,7 +229,7 @@ def nbl_to_padsize(nbl, ndim):
 
 
 def initialize_function(function, data, nbl, mapper=None, mode='constant',
-                        name='padfunc', **kwargs):
+                        name='padfunc', pad_halo=True, **kwargs):
     """
     Initialize a Function with the given ``data``. ``data``
     does *not* include the ``nbl`` outer/boundary layers; these are added via padding
@@ -255,6 +255,8 @@ def initialize_function(function, data, nbl, mapper=None, mode='constant',
         accepted.
     name : str, optional
         The name assigned to the operator.
+    pad_halo : bool, optional
+        Whether to pad the FD halo as well.
 
     Examples
     --------
@@ -304,10 +306,11 @@ def initialize_function(function, data, nbl, mapper=None, mode='constant',
             function.data[:] = data.data[:]
         else:
             function.data[:] = data[:]
+        if pad_halo:
+            pad_fd_halo(function)
         return
 
     nbl = nbl_to_padsize(nbl, function.ndim)
-
     slices = tuple([slice(nl, -nr) for _, (nl, nr) in zip(range(function.grid.dim),
                                                           as_tuple(nbl))])
     if isinstance(data, dv.Function):
@@ -363,6 +366,29 @@ def initialize_function(function, data, nbl, mapper=None, mode='constant',
         options = None
 
     assign(lhs, rhs, options=options, name=name, **kwargs)
+
+    if pad_halo:
+        pad_fd_halo(function)
+
+
+def pad_fd_halo(function):
+    """ Pad finite-difference halo with edge values."""
+    h_shape = function._data_with_outhalo.shape
+    for i, h in enumerate(function._size_outhalo):
+        if h.left == 0 and h.right == 0:
+            continue
+        slices = [slice(None)]*len(function.shape)
+        slices_d = [slice(None)]*len(function.shape)
+        # left part
+        slices[i] = slice(0, h.left)
+        slices_d[i] = slice(h.left, h.left+1, 1)
+        function._data_with_outhalo[tuple(slices)] \
+            = function._data_with_outhalo[tuple(slices_d)]
+        # right part
+        slices[i] = slice(h_shape[i] - h.right, h_shape[i], 1)
+        slices_d[i] = slice(h_shape[i] - h.right - 1, h_shape[i] - h.right, 1)
+        function._data_with_outhalo[tuple(slices)] \
+            = function._data_with_outhalo[tuple(slices_d)]
 
 
 # Reduction-inducing builtins
--- a/devito/builtins.py
+++ b/devito/builtins.py
@@ -256,7 +256,7 @@ def initialize_function(function, data, nbl, mapper=None, mode='constant',
     name : str, optional
         The name assigned to the operator.
     pad_halo : bool, optional
-        Whether to pad the FD halo as well.
+        Whether to also pad the finite-difference halo.
 
     Examples
     --------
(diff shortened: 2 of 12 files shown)


devito/builtins.py
```diff
@@ -255,6 +255,8 @@ def initialize_function(function, data, nbl, mapper=None, mode='constant',
         accepted.
     name : str, optional
         The name assigned to the operator.
+    pad_halo : bool, optional
+        Whether to pad the FD halo as well.
```
rhodrin, line 259: *Whether to also pad the ...

devito/builtins.py
```diff
@@ -364,6 +367,29 @@ def buff(i, j):
 
     assign(lhs, rhs, options=options, name=name, **kwargs)
 
+    if pad_halo:
+        pad_fd_halo(function)
+
+
+def pad_fd_halo(function):
```
rhodrin, line 374: Is `builtins.py` the right place for this?

tests/test_builtins.py
```diff
@@ -239,6 +239,37 @@ def test_if_parallel(self):
             assert np.all(a[::-1, 3:6] - np.array(f.data[12:18, 9:12]) == 0)
             assert np.all(a[3:6, ::-1] - np.array(f.data[9:12, 12:18]) == 0)
 
+    @pytest.mark.parametrize('ndim, nbl', [
```
rhodrin, line 242: I think we should have a test with mpi also?

devito/builtins.py
```diff
@@ -364,6 +367,29 @@ def buff(i, j):
...
+    """ Pad finite-difference halo with edge values."""
+    h_shape = function._data_with_outhalo.shape
+    for i, h in enumerate(function._size_outhalo):
+        if h.left == 0 and h.right == 0:
+            continue
+        slices = [slice(None)]*len(function.shape)
+        slices_d = [slice(None)]*len(function.shape)
+        # left part
+        slices[i] = slice(0, h.left)
```
rhodrin, line 383: this works ok with mpi where on a rank you could end up with
```
        slices[i] = slice(0, 0)
        slices_d[i] = slice(0, 1, 1)
```
?

rhodrin: Same for the right bit.

devito/builtins.py
mloubout: It's just a local function because it's used at two different places, similar to the `nbl` setup function. Not sure where else would put it.

tests/test_builtins.py
mloubout: Yes we should, working on it

devito/builtins.py
mloubout: Not sure I follow

devito/builtins.py
mloubout: Fixed

tests/test_builtins.py
mloubout: Added MPI test

devito/builtins.py
mloubout: Fixed MPI case

.github/workflows/tutorials.yml
```diff
@@ -67,7 +67,7 @@ jobs:
       if: runner.os == 'macOS'
       run: |
         if [ "${{ matrix.compiler }}" = "gcc-9" ]; then
-          brew install gcc
+          brew install gcc@9
```
FabioLuporini, line 70: out of curiosity, why the change?

devito/builtins.py
```diff
@@ -255,6 +255,8 @@ def initialize_function(function, data, nbl, mapper=None, mode='constant',
         accepted.
     name : str, optional
         The name assigned to the operator.
+    pad_halo : bool, optional
+        Whether to also pad the finite-difference halo.
```
FabioLuporini, line 259: Drop "finite-difference" or replace with something else

devito/builtins.py
FabioLuporini: yeah, we need to move the builtins to /builtins such that we can also have a utils.py

since it's a private function, the name could start with a _

devito/builtins.py
```diff
@@ -364,6 +367,29 @@ def buff(i, j):
 
     assign(lhs, rhs, options=options, name=name, **kwargs)
 
+    if pad_halo:
+        pad_fd_halo(function)
+
+
+def pad_fd_halo(function):
+    """ Pad finite-difference halo with edge values."""
```
FabioLuporini, line 375: drop "finite-difference"

devito/finite_differences/differentiable.py
```diff
@@ -240,16 +240,17 @@ def laplace(self):
         derivs = tuple('d%s2' % d.name for d in space_dims)
         return Add(*[getattr(self, d) for d in derivs])
 
-    @property
-    def div(self):
+    def div(self, shift=None):
```
FabioLuporini, line 243: turning this property into a function -- should we consider it a major API change?

devito/finite_differences/differentiable.py
```diff
@@ -240,16 +240,17 @@ def laplace(self):
...
         space_dims = [d for d in self.dimensions if d.is_Space]
-        derivs = tuple('d%s' % d.name for d in space_dims)
-        return Add(*[getattr(self, d) for d in derivs])
+        return Add(*[getattr(self, 'd%s' % d.name)(x0=None if shift is None else
```
FabioLuporini, line 245: easier to read as

```
x0=shift or (d + shift * d.spacing)
```

?

devito/finite_differences/differentiable.py
```diff
@@ -240,16 +240,17 @@ def laplace(self):
...
+                                                   d + shift * d.spacing)
+                     for d in space_dims])
 
-    @property
-    def grad(self):
+    def grad(self, shift=None):
         from devito.types.tensor import VectorFunction, VectorTimeFunction
-        comps = [getattr(self, 'd%s' % d.name) for d in self.dimensions if d.is_Space]
+        comps = [getattr(self, 'd%s' % d.name)(x0=None if shift is None else
```
FabioLuporini, line 251: same as above?

devito/builtins.py
FabioLuporini: > yeah, we need to move the builtins to /builtins such that we can also have a utils.py

this also appears as an issue on GH. We could do it now

mloubout: Doable

.github/workflows/tutorials.yml
mloubout: brew now installs gcc 10 by default, was easier to fix just this line than the other ones for gcc 10.

devito/finite_differences/differentiable.py
mloubout: No this is not user facing, `div(f)` is and hasn't changed.

devito/finite_differences/differentiable.py
mloubout: That's also not the same and wrong.

`None if shift is None else  (d + shift * d.spacing)` 
- None => None
- value => d + value * spacing

`x0=shift or (d + shift * d.spacing)` : 
- None => d + None * d.spacing , that's not valid
- value => value

devito/finite_differences/differentiable.py
mloubout: same

devito/builtins.py
mloubout: But that's the FD halo from the space order, why droping it?

devito/builtins.py
mloubout: Done

devito/finite_differences/differentiable.py
FabioLuporini: ah, LOL, don't know what my mind was processing when I wrote that comment above, ignore me :grimacing:

FabioLuporini (commented):
I really like the new `builtins/`

devito/builtins.py
FabioLuporini: the halo is a concept that applies to any (un)structured grid computation, so it's broader than FD

tests/test_builtins.py
```diff
@@ -239,6 +239,72 @@ def test_if_parallel(self):
...
+            assert np.take(f._data_with_outhalo, -1, axis=1)[7, 7] == 6
+        else:
+            assert f._data_with_outhalo[7, 7] == 0
+            assert np.take(f._data_with_outhalo, 0, axis=-1)[7] == 1
+            assert np.take(f._data_with_outhalo, -1, axis=-1)[7] == 3
+
+    @skipif('nompi')
+    @pytest.mark.parametrize('nbl', [0, 2])
+    @pytest.mark.parallel(mode=4)
+    def test_if_halo_mpi(self, nbl):
```
FabioLuporini, line 278: I don't know why, but it looks like my comment about merging this and the non-MPI test into one test with `pytest.mark.parallel(mode=[1, 2, 4])` has disappeared?

mloubout: `mode=1` is still MPI with one rank, not non-mpi.

devito/builtins.py
mloubout: Changed to outer halo

FabioLuporini (approved):
GTG, thanks

rhodrin (approved):

PR #1514, opened 2020-11-19
Postpone lowering of Section into TimedList

FabioLuporini:
This is a first step towards putting sections around MPI_Wait (for MPI) and locks (for GPU asynchronous execution) for a fine-grained breakdown of the waiting times.

I will need beta-testing from as many people as possible to be 100% sure I haven't screwed the performance profiling emitted by an Operator after `apply`, w/ and w/o auto-tuning. 

Seems fine but I'd like more pair of eyes to look at this.


FabioLuporini:
@navjotk : as you may guess, this is a first step towards timing the GPU waits which you/we need

mloubout (commented):
Looks fine to me. So I guess the new `pragma` is for future use for the timer?

FabioLuporini:
> So I guess the new pragma is for future use for the timer?

yes and no. I was also hitting a stupid IET-level error because I was iterating over supposed-to-be Expressions, while some of them actually were `List(header=pragma, body=Expression)`. Now that the pragma(s) is(are) part of the Expression, it's all good

FabioLuporini:
@mloubout oops, sorry, I just pushed the forgotten `passes/iet/instrument.py` file (though it's a tiny one)

georgebisbas (approved):
Seems GTG for me, I run some operators, logging=debug, ac-el-tti, w-wo autotuning, w-wo openmp, no problems so far.

PR #1515, opened 2020-11-20
Fix ARM issues

georgebisbas:
Supersedes 1510
Fix ARM issue, improve platform autodetection.
- autodetect Features (flags in older versions) necessary for vectorization
- assert `lscpu` and `psutil` where possible
- Solves issues when splitting `:` due to https://askubuntu.com/questions/1248273/lscpu-vulnerabilities

TODO:
Improve par-nested override for a global solution

Tested on Isambard's Marvell ThunderX2


FabioLuporini:
should the title be changed?

devito/archinfo.py
```diff
@@ -71,6 +72,9 @@ def get_cpu_brand():
         cpu_info['flags'] = ci.get('flags')
         cpu_info['brand'] = ci.get('brand')
 
+        if cpu_info['brand'] is None:
```
FabioLuporini, line 75: Can you add a comment about where (what platform) we saw "this" happening , ie the flags!=None while brand still None.

devito/archinfo.py
```diff
@@ -114,8 +118,16 @@ def get_cpu_brand():
             except KeyError:
                 warning("Physical core count autodetection failed")
                 physical = 1
-    cpu_info['physical'] = physical
+    try:
+        if physical == (lscpu()['Core(s) per socket'] * lscpu()['Socket(s)']):
```
FabioLuporini, line 122: 1) the lines above also access `lscpu`, so it's time to assign `out = lscpu()`
2) now we're **always** calling `lscpu()` on startup, which has a cost... how long does it take? order of milliseconds I presume. Can you time it?
2) the  way you wrote this `if` is a bit ugly, isn't it better as

```
expected = out[...]*out[...]
if physical != expected:
    physical = expected
```

note that I dropped the `warning` which would only confuse the user. However, there's a bigger conceptual flaw... :

I would guess the check starting at line 94 (ie the body of the `if lines`) produces a **wrong** `physical` value on some ARM architectures, right? if so, why is that the case? you may say a bug in `psutil`. Fine, but I think we should rather change the code such that the `if` at line 110 **is entered** (this's the fallback branch after all), thus avoiding the need for 121-...

You're instead _adding_ new code (lines 121...) which will be executed _regardless_ of the content of physical, and this may slow down startup, which instead should be extremely quick.

IOW:

* lines 94-108 is where we should sanity-check the content of `lines` such that we're sure it will produce the right `physical` value. 
* lines 110-120 is where we use the fallback, which includes calling `lscpu`

Does this make sense?

devito/core/cpu.py
```diff
@@ -319,8 +319,79 @@ class Intel64FSGOpenMPOperator(Intel64FSGOperator, CPU64OpenMPOperator):
...
+        data_manager.place_definitions(graph)
+        data_manager.place_casts(graph)
+
+        return graph
+
+
+class ArmOpenMPOperator(CPU64OpenMPOperator):
+
+    PAR_NESTED = 4
+    """
```
FabioLuporini, line 358: drop this docstring, which is clearly redundant (docstrings are inherited)

FabioLuporini (changes requested):

devito/archinfo.py
georgebisbas: ok

mloubout (approved):

devito/archinfo.py
```diff
@@ -40,7 +40,9 @@ def get_cpu_info():
 
         def get_cpu_flags():
             try:
-                flags = [i for i in lines if i.startswith('flags')][0]
+                # ARM Thunder X2 is using 'Features' instead of 'flags'
+                flags = [i for i in lines if (i.startswith('Features')
+                         or i.startswith('flags'))][0]
```
FabioLuporini, line 45: nitpicking: the `or i.startswith('flags)'` should be aligned with `i.startswith('Featuers')`

devito/archinfo.py
```diff
@@ -64,12 +66,16 @@ def get_cpu_brand():
...
+        if cpu_info['brand'] is None:
+            cpu_info['brand'] = cpuinfo.get_cpu_info().get('raw_arch_string')
 
-    if not all(i in cpu_info for i in ('flags', 'brand')):
-        # Fallback
-        ci = cpuinfo.get_cpu_info()
-        cpu_info['flags'] = ci.get('flags')
-        cpu_info['brand'] = ci.get('brand')
+    try:
+        if not ('flags' in cpu_info):
```
FabioLuporini, line 73: `if 'flags' not in cpu_info`

FabioLuporini: actually, this should be moved [here](https://github.com/devitocodes/devito/blob/master/devito/archinfo.py#L46) in place of the `return None`. Right?

devito/archinfo.py
```diff
@@ -64,12 +66,16 @@ def get_cpu_brand():
 
         cpu_info['flags'] = get_cpu_flags()
         cpu_info['brand'] = get_cpu_brand()
+        if cpu_info['brand'] is None:
+            cpu_info['brand'] = cpuinfo.get_cpu_info().get('raw_arch_string')
```
FabioLuporini, line 70: Actually, this should be moved [here](https://github.com/devitocodes/devito/blob/master/devito/archinfo.py#L63) in place of the `return None` right?

devito/archinfo.py
FabioLuporini: if we do so, `cpuinfo.get_cpu_info()` will be called twice. Then we probably want to memoize the call.

```
@memoized_func
def _get_cpu_info_from_cpuinfo():
     return cpuinfo.get_cpu_info()
```

unless it's already memoized internally by cpuinfo itself. Can you find out and/or time it

devito/archinfo.py
```diff
@@ -107,15 +113,31 @@ def get_cpu_brand():
         # Fallback 1: it should now be fine to use psutil
         physical = psutil.cpu_count(logical=False)
         if not physical:
-            # Fallback 2: we might end up here on more exotic platforms such a Power8
-            # Hopefully we can rely on `lscpu`
             try:
-                physical = lscpu()['Core(s) per socket'] * lscpu()['Socket(s)']
+                lscpu_mapper = lscpu()
+                # Fallback 2: we might end up here
```
FabioLuporini, line 118: why was this comment moved?

devito/archinfo.py
```diff
@@ -107,15 +113,31 @@ def get_cpu_brand():
         # Fallback 1: it should now be fine to use psutil
         physical = psutil.cpu_count(logical=False)
         if not physical:
-            # Fallback 2: we might end up here on more exotic platforms such a Power8
-            # Hopefully we can rely on `lscpu`
             try:
-                physical = lscpu()['Core(s) per socket'] * lscpu()['Socket(s)']
+                lscpu_mapper = lscpu()
```
FabioLuporini, line 117: I just remember that this is memoized, so this extra variable is probably useless. Apologies if I complained about that earlier

devito/archinfo.py
```diff
@@ -107,15 +113,31 @@ def get_cpu_brand():
...
+        if cpu_info['brand'] == 'aarch64':
+            # In some ARM processors psutils and lscpu fail to detect cores correctly
+            physical = psutil.cpu_count(logical=False)
+            lscpu_mapper = lscpu()
+            if physical != (lscpu_mapper['Core(s) per socket']
+                            * lscpu_mapper['Socket(s)']):
+                warning("Platform core count detection is not reliable. Using lscpu()")
+                physical = (lscpu_mapper['Core(s) per socket']
+                            * lscpu_mapper['Socket(s)'])
+    except:
```
FabioLuporini, line 137: out of curiosity, we don't know when we could end up here right?

devito/archinfo.py
```diff
@@ -107,15 +113,31 @@ def get_cpu_brand():
...
+                # erroneous autodetection
+                physical = (lscpu_mapper['Core(s) per socket']
+                            * lscpu_mapper['Socket(s)'])
             except KeyError:
                 warning("Physical core count autodetection failed")
                 physical = 1
-    cpu_info['physical'] = physical
 
+    try:
+        if cpu_info['brand'] == 'aarch64':
```
FabioLuporini, line 128:
```
if cpu_info['brand'] in ('arm', 'aarch64'):
```
?

devito/archinfo.py
```diff
@@ -107,15 +113,31 @@ def get_cpu_brand():
...
+            # In some ARM processors psutils and lscpu fail to detect cores correctly
+            physical = psutil.cpu_count(logical=False)
+            lscpu_mapper = lscpu()
```
FabioLuporini, line 131: again

devito/archinfo.py
```diff
@@ -107,15 +113,31 @@ def get_cpu_brand():
...
+            if physical != (lscpu_mapper['Core(s) per socket']
+                            * lscpu_mapper['Socket(s)']):
+                warning("Platform core count detection is not reliable. Using lscpu()")
```
FabioLuporini, line 134: we can drop this?

devito/archinfo.py
```diff
@@ -107,15 +113,31 @@ def get_cpu_brand():
...
+                # on more exotic platforms such a Power8 or due to
+                # erroneous autodetection
+                physical = (lscpu_mapper['Core(s) per socket']
+                            * lscpu_mapper['Socket(s)'])
             except KeyError:
                 warning("Physical core count autodetection failed")
                 physical = 1
-    cpu_info['physical'] = physical
 
+    try:
```
FabioLuporini, line 127: this entire try-except should be moved up... 

probably around [here](https://github.com/devitocodes/devito/blob/master/devito/archinfo.py#L85)

then we update `cpu_info` and return without going through the generic detection code. After all, this is a special case, and the special cases should be dealt with **before** the general case, almost always

FabioLuporini (changes requested):

devito/archinfo.py
georgebisbas: we already have:
https://github.com/devitocodes/devito/pull/1515/files#diff-fc43288ece59fd09fd1aaeaddf92cd861a8289feb72e30d449719f1dae337f3bR20
```
@memoized_func
def get_cpu_info():

```
I guess it is enough?

devito/archinfo.py
georgebisbas: np

devito/archinfo.py
georgebisbas: Accidental, sorry, fixed.

devito/archinfo.py
georgebisbas: Moved up, hope its fine.

devito/archinfo.py
georgebisbas: What is the issue here? Older ARMs will return arm, while in cases we are not able to autodetect and get
the raw_arch_string we get aarch64. Let me know if we need more insight here

devito/archinfo.py
georgebisbas: ok

devito/archinfo.py
georgebisbas: dropped.

devito/archinfo.py
georgebisbas: To the best of my knowledge, no.

devito/archinfo.py
```diff
@@ -60,18 +62,36 @@ def get_cpu_brand():
...
 
         cpu_info['flags'] = get_cpu_flags()
         cpu_info['brand'] = get_cpu_brand()
 
-    if not all(i in cpu_info for i in ('flags', 'brand')):
-        # Fallback
-        ci = cpuinfo.get_cpu_info()
-        cpu_info['flags'] = ci.get('flags')
-        cpu_info['brand'] = ci.get('brand')
+    try:
```
georgebisbas, line 73: @FabioLuporini conceptually you are right to move this try-except up, but it does not go through at the place you mentioned, probably due to not finding lines in OSX. I do not have one OSX to debug it efficiently.

devito/archinfo.py
```diff
@@ -60,18 +62,36 @@ def get_cpu_brand():
...
+        if 'flags' not in cpu_info:
```
FabioLuporini, line 74: is it true that the only possibility for `flags` not to be in `cpu_info` is that `lines` is empty? because if we enter the `if lines:` at line 34, then we go through lines 70-71, hence `cpu_info['flags']` should be defined?

devito/archinfo.py
```diff
@@ -60,18 +62,36 @@ def get_cpu_brand():
...
+            # Fallback
+            ci = cpuinfo.get_cpu_info()
+            cpu_info['flags'] = ci.get('flags')
```
FabioLuporini, line 77: so going back to my original comment. What I was **actually** suggesting was to move the **content** of the try-except in place of current line 48 -- instead of returning `None`, we return the `cpuinfo.get_cpu_info().get('flags')` ?

devito/archinfo.py
```diff
@@ -60,18 +62,36 @@ def get_cpu_brand():
...
+    except:
+        pass
 
     # Detect number of logical cores
+    try:
+        if cpu_info['brand'] == 'aarch64':
+            # In some ARM processors psutils and lscpu fail to detect cores correctly
```
FabioLuporini, line 84: this comment doesn't make much sense does it? you say that `lscpu` fails, but then you actually use it below?

devito/archinfo.py
```diff
@@ -60,18 +62,36 @@ def get_cpu_brand():
...
+            logical = psutil.cpu_count(logical=True)
+            physical = psutil.cpu_count(logical=False)
+            if physical != (lscpu()['Core(s) per socket'] * lscpu()['Socket(s)']):
```
FabioLuporini, line 87:
```
if not A:
   B = A
```
=>
```
B = A
```

devito/archinfo.py
```diff
@@ -60,18 +62,36 @@ def get_cpu_brand():
             except:
                 pass
 
-            return None
+            try:
```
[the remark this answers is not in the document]
FabioLuporini: "Certain ARM cpus ... "

devito/archinfo.py
```diff
@@ -60,18 +62,36 @@ def get_cpu_brand():
...
+    try:
+        if 'flags' not in cpu_info:
+            # Fallback
+            ci = cpuinfo.get_cpu_info()
+            cpu_info['flags'] = ci.get('flags')
+    except:
+        pass
 
     # Detect number of logical cores
+    try:
```
FabioLuporini, line 82: why do you need a try except here? I think you don't

FabioLuporini: Ah... you need it in case `lscpu` fails?

devito/archinfo.py
```diff
@@ -60,18 +62,36 @@ def get_cpu_brand():
...
+        if cpu_info['brand'] == 'aarch64':
+            # In some ARM processors psutils and lscpu fail to detect cores correctly
+            logical = psutil.cpu_count(logical=True)
+            physical = psutil.cpu_count(logical=False)
+            if physical != (lscpu()['Core(s) per socket'] * lscpu()['Socket(s)']):
+                physical = (lscpu()['Core(s) per socket'] * lscpu()['Socket(s)'])
+            cpu_info['logical'] = logical
+            cpu_info['physical'] = physical
+            return cpu_info
```
FabioLuporini, line 91: We can rewrite this try-except as:

```
if cpu_info['brand'] == 'aarch64':
    # Special case: ...
    cpu_info['physical'] = lscpu()['Core(s) per socket'] * lscpu()['Socket(s)']
```

we don't need the `logical`; we can move it right below lines 95-104, which will still work. 

In particular, this should go below the comment "Detect number of physical cores"

devito/archinfo.py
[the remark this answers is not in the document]
georgebisbas: ok

devito/archinfo.py
georgebisbas: This is the part where I do not know what happens in OSX, tests fail only there.

devito/archinfo.py
georgebisbas: again, I did that the previous days and was failing on OSX

devito/archinfo.py
georgebisbas: yes

devito/archinfo.py
georgebisbas: omg nice catch, I do not remember what I was thinking at that time.

devito/archinfo.py
georgebisbas: True. This is a leftover from using it to print a warning message thaty was later removed.
```warning...core autodetection is not reliable....using lscpu....```

File: devito/archinfo.py (hunk 20 above)
Comment (georgebisbas) [reply devito/archinfo.py]: well... I guess it cannot fail though...so I will remove...

File: devito/archinfo.py (hunk 21 above)
Comment (georgebisbas) [reply devito/archinfo.py]: but then we will not be able to return cpu_info for the case of arm?

Comment (georgebisbas) [reply devito/archinfo.py:91]: note: `if cpu_info['brand']` fails on OSX, cpu_info does not have 'brand'

File: devito/archinfo.py (hunk 20 above)
Comment (FabioLuporini) [reply devito/archinfo.py:83]: it  can -- if there's no `lscpu` on the system

Comment (georgebisbas) [reply devito/archinfo.py:83]: But we are not aware of such a case so far right?

File: devito/archinfo.py [hunk 22]
```diff
@@ -107,15 +119,16 @@ def get_cpu_brand():
...
         physical = psutil.cpu_count(logical=False)
         if not physical:
-            # Fallback 2: we might end up here on more exotic platforms such a Power8
-            # Hopefully we can rely on `lscpu`
+            # Fallback 2: we might end up here
+            # on more exotic platforms such a Power8 or due to
+            # erroneous autodetection
             try:
-                physical = lscpu()['Core(s) per socket'] * lscpu()['Socket(s)']
+                physical = (lscpu()['Core(s) per socket'] * lscpu()['Socket(s)'])
```
FabioLuporini, line 126: why adding parentheses?

devito/archinfo.py
```diff
@@ -107,15 +119,16 @@ def get_cpu_brand():
         # Fallback 1: it should now be fine to use psutil
         physical = psutil.cpu_count(logical=False)
         if not physical:
-            # Fallback 2: we might end up here on more exotic platforms such a Power8
-            # Hopefully we can rely on `lscpu`
+            # Fallback 2: we might end up here
```
FabioLuporini, line 122: can we avoid breaking this line into three?

also, typo: 'such a' -> 'such as'

should we now drop the "or due to erroneous autodetection" part?

devito/archinfo.py
```diff
@@ -107,15 +119,16 @@ def get_cpu_brand():
...
+            # erroneous autodetection
             try:
-                physical = lscpu()['Core(s) per socket'] * lscpu()['Socket(s)']
+                physical = (lscpu()['Core(s) per socket'] * lscpu()['Socket(s)'])
             except KeyError:
                 warning("Physical core count autodetection failed")
                 physical = 1
-    cpu_info['physical'] = physical
 
+    cpu_info['physical'] = physical
```
FabioLuporini, line 131: this shouldn't have been moved two lines below

devito/archinfo.py
```diff
@@ -258,7 +271,8 @@ def lscpu():
     if output:
         lines = output.decode("utf-8").strip().split('\n')
         mapper = {}
-        for k, v in [tuple(i.split(':')) for i in lines]:
+        # Using split(':', 1) to avoid splitting lines with security issues
```
FabioLuporini, line 274: "with security issues" . Can you be more precise? Nobody will understand it

```
# Using split(.....
# Example: ...
```

devito/archinfo.py
```diff
@@ -32,18 +32,19 @@ def get_cpu_info():
...
         def get_cpu_flags():
             try:
-                flags = [i for i in lines if i.startswith('flags')][0]
+                # ARM Thunder X2 is using 'Features' instead of 'flags'
+                flags = [i for i in lines if (i.startswith('Features')
+                                              or i.startswith('flags'))][0]
                 return flags.split(':')[1].strip().split()
             except:
-                return None
+                pass
```
FabioLuporini, line 47: why this change? this will `return None` anyway, so it's clearer the other way

devito/archinfo.py
```diff
@@ -60,16 +61,27 @@ def get_cpu_brand():
...
 
         cpu_info['flags'] = get_cpu_flags()
         cpu_info['brand'] = get_cpu_brand()
-
-    if not all(i in cpu_info for i in ('flags', 'brand')):
-        # Fallback
-        ci = cpuinfo.get_cpu_info()
-        cpu_info['flags'] = ci.get('flags')
-        cpu_info['brand'] = ci.get('brand')
+    else:
```
FabioLuporini, line 72: what happens if we get into the `if lines`  branch but either of `flags` or `brand` are None? 

are we sure that this new version behaves like the previous one? I'm not

devito/archinfo.py
georgebisbas: Yes

devito/archinfo.py
georgebisbas: I think it is one line? Anyway, fixed.

devito/archinfo.py
georgebisbas: What about:
```
        # Using split(':', 1) to avoid splitting lines where lscpu shows vulnerabilities
        # on some CPUs: https://askubuntu.com/questions/1248273/lscpu-vulnerabilities
```

devito/archinfo.py
georgebisbas: Fixed

devito/archinfo.py
georgebisbas: I think I get your point. TO my understanding if the last pair of try-except in each function
`get_cpu_xxx`
is returning None, then we are fine.

devito/archinfo.py
georgebisbas: Added `return None` in the last try-except pair of each `get_cpu_xxx` function. I think now it has the same behaviour as before.

devito/archinfo.py
```diff
@@ -60,16 +61,39 @@ def get_cpu_brand():
...
+        cpu_info['brand'] = cpuinfo.get_cpu_info().get('brand')
+
+    # Detect number of physical and logical cores
+    try:
+        if 'arm' in cpu_info['brand']:
+            # Special case: In some ARM processors psutils fail to detect physical cores
+            # correctly so we use lscpu()
+            logical = psutil.cpu_count(logical=True)
+            physical = (lscpu()['Core(s) per socket'] * lscpu()['Socket(s)'])
+            cpu_info['logical'] = logical
```
FabioLuporini, line 92: I still think that the issue with ARM is the correct detection of number of physical cores, **not** the logical cores.  So I would:

* drop the `logical` detection here
* move this try-except after the "logical core detection" part, ie below line 110

FabioLuporini (changes requested):

devito/archinfo.py
georgebisbas: Yes, but getting the logical here helps as return `cpu_info` and avoid all the other lines after 110

devito/archinfo.py
```diff
@@ -60,16 +61,39 @@ def get_cpu_brand():
...
         cpu_info['flags'] = get_cpu_flags()
         cpu_info['brand'] = get_cpu_brand()
 
-    if not all(i in cpu_info for i in ('flags', 'brand')):
-        # Fallback
-        ci = cpuinfo.get_cpu_info()
-        cpu_info['flags'] = ci.get('flags')
-        cpu_info['brand'] = ci.get('brand')
+    try:
+        if cpu_info['flags'] is None:
```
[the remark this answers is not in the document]
georgebisbas: ok

devito/archinfo.py
```diff
@@ -60,16 +61,39 @@ def get_cpu_brand():
...
+    except KeyError:
+        cpu_info['flags'] = cpuinfo.get_cpu_info().get('flags')
+
+    try:
+        if cpu_info['brand'] is None:
+            cpu_info['brand'] = cpuinfo.get_cpu_info().get('brand')
+    except KeyError:
+        cpu_info['brand'] = cpuinfo.get_cpu_info().get('brand')
+
+    # Detect number of physical and logical cores
```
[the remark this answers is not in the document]
georgebisbas: ok

devito/archinfo.py
FabioLuporini: I was suggesting to move it right between lines 103 and 104, so you would still avoid "those other lines"

FabioLuporini: if we do so, I think, it'd become a bit neater

devito/archinfo.py
```diff
@@ -60,16 +61,32 @@ def get_cpu_brand():
...
-        ci = cpuinfo.get_cpu_info()
-        cpu_info['flags'] = ci.get('flags')
-        cpu_info['brand'] = ci.get('brand')
+    if not cpu_info.get('flags'):
+        cpu_info['flags'] = cpuinfo.get_cpu_info().get('flags')
+
+    if not cpu_info.get('brand'):
+        cpu_info['brand'] = cpuinfo.get_cpu_info().get('brand')
+
+    # Special case: in some ARM processors psutils fail to detect physical cores
```
FabioLuporini, line 79: nitpicking: fails

mloubout:
Considering the amount of `try/except` and `if/else`, would it be good to separate `archinfo` in arch specific case and just check them one by one like 

```
for arch in [intel, arm, ....]
     if found(arch):
         do what's needed
         break
```

devito/archinfo.py
```diff
@@ -83,6 +88,16 @@ def get_cpu_brand():
             logical = 1
     cpu_info['logical'] = logical
 
+    # Special case: in some ARM processors psutils fails to detect physical cores
+    # correctly so we use lscpu()
+    try:
+        if 'arm' in cpu_info['brand']:
+            physical = (lscpu()['Core(s) per socket'] * lscpu()['Socket(s)'])
+            cpu_info['physical'] = physical
```
FabioLuporini, line 96: can this be a single line or does it not fit the 90 chars limit?

FabioLuporini (commented):
I actually think I'm quite happy with the current state of this PR

devito/core/intel.py
```diff
@@ -0,0 +1,202 @@
...
+        clusters = Blocking(options).process(clusters)
+
+        return clusters
+
+
+class Intel64FSGOpenMPOperator(Intel64FSGOperator, CPU64OpenMPOperator):
+    _specialize_iet = CPU64OpenMPOperator._specialize_iet
+
+
+PowerOperator = CPU64Operator
```
FabioLuporini, line 74: what is this?

devito/core/intel.py
```diff
@@ -0,0 +1,202 @@
...
+PowerOpenMPOperator = CPU64OpenMPOperator
+
+ArmOperator = CPU64Operator
```
FabioLuporini, line 77: same

devito/core/intel.py
```diff
@@ -0,0 +1,202 @@
...
+                                    extract_increments, factorize, fuse, optimize_pows)
+from devito.passes.iet import (DataManager, Ompizer, avoid_denormals, mpiize,
+                               optimize_halospots, hoist_prodders, relax_incr_dimensions)
+from devito.tools import as_tuple, timed_pass
+
+__all__ = ['Intel64Operator', 'Intel64OpenMPOperator', 'Intel64FSGOperator',
+           'Intel64FSGOpenMPOperator']
+
+
+Intel64Operator = CPU64Operator
```
FabioLuporini, line 16: necessary?

FabioLuporini (changes requested):

devito/core/__init__.py
```diff
@@ -1,8 +1,10 @@
 from devito.archinfo import Cpu64, Intel64, Arm, Power, Device
 from devito.core.cpu import (CPU64NoopOperator, CPU64Operator, CPU64OpenMPOperator,
-                             Intel64Operator, Intel64OpenMPOperator, Intel64FSGOperator,
-                             Intel64FSGOpenMPOperator, ArmOperator, ArmOpenMPOperator,
-                             PowerOperator, PowerOpenMPOperator, CustomOperator)
+                             CustomOperator)
+from devito.core.intel import (Intel64Operator, Intel64OpenMPOperator,
+                               Intel64FSGOperator, Intel64FSGOpenMPOperator)
+from devito.core.arm import (ArmOperator, ArmOpenMPOperator)
```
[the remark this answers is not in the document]
georgebisbas: Fixed

devito/core/__init__.py
```diff
@@ -1,8 +1,10 @@
...
+from devito.core.power import (PowerOperator, PowerOpenMPOperator)
```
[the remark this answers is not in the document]
georgebisbas: Fixed

devito/core/arm.py
```diff
@@ -0,0 +1,81 @@
+from devito.core.cpu import CPU64Operator, CPU64OpenMPOperator
+from devito.passes.iet import (DataManager, Ompizer, mpiize,
+                               optimize_halospots, hoist_prodders, relax_incr_dimensions)
+from devito.tools import timed_pass
+
+
+__all__ = ['ArmOperator', 'ArmOpenMPOperator']
+
+
+ArmOperator = CPU64Operator
```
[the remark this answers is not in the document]
georgebisbas: HHmm... not needed but was it needed before?

devito/core/arm.py
```diff
@@ -0,0 +1,81 @@
...
+ArmOpenMPOperator = CPU64OpenMPOperator
```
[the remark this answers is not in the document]
georgebisbas: ok

devito/core/intel.py
georgebisbas: Yes, needed for registry...though it can be removed, but is better like that IMHO (as an intermediate between CPU64 and Intel64FSG)

devito/core/intel.py
georgebisbas: Dropped

devito/core/intel.py
georgebisbas: dropped

devito/core/arm.py
```diff
@@ -0,0 +1,77 @@
+from devito.core.cpu import CPU64Operator, CPU64OpenMPOperator
+from devito.passes.iet import (DataManager, Ompizer, mpiize,
+                               optimize_halospots, hoist_prodders, relax_incr_dimensions)
+from devito.tools import timed_pass
+
+
```
FabioLuporini, line 6: one extra blank line?

FabioLuporini (approved):
OK now, thanks

FabioLuporini (approved):
OK, good now, thanks. Merging

FabioLuporini:
Merged.

PR #1519, opened 2020-11-27
Add test that compares gradient equivalence

navjotk:
We know that the gradient relationship can be written as either of the following, since it comes from integration by parts:
- `grad -= u * v.dt2`
- `grad -= u.dt2 * v`

This PR adds a test that computes the gradient through both these relationships and asserts "equality" (within floating-point precision). 

Current state:
- Added a draft test, looking for feedback on whether this should be rewritten differently before merge.

- Test does not pass up to `rtol=1e-4`, `atol=1e-4`. What would be our expected rtol/atol values for this test to pass?


tests/test_gradient.py
```diff
@@ -51,6 +53,71 @@ def test_gradient_checkpointing(self, shape, kernel, space_order):
...
+
+        # Add expression for receiver injection
+        receivers = rec.inject(field=v.backward, expr=rec * s**2 / m)
+
+        # Substitute spacing terms to reduce flops
+        grad_op_v = Operator(eqn + receivers + [gradient_update_v],
+                             subs=model.spacing_map, name='Gradient')
+
+        grad_op_u = Operator(eqn + receivers + [gradient_update_u],
+                             subs=model.spacing_map, name='Gradient')
```
mloubout, line 113: can combine in one op

tests/test_gradient.py
```diff
@@ -51,6 +53,71 @@ def test_gradient_checkpointing(self, shape, kernel, space_order):
...
+        eqn = iso_stencil(u, model, kernel)
+
+        # Construct expression to inject source values
+        src_term = src.inject(field=u.forward, expr=src * s**2 / m)
+
+        # Create interpolation expression for receivers
+        rec_term = rec.interpolate(expr=u)
+
+        fwd_op = Operator(eqn + src_term + rec_term, subs=model.spacing_map,
+                          name='Forward')
```
mloubout, line 89: Can probaly get all of that with something like

```
_, _, _, u, rec = acoustic.run(...., sav=true)
```

tests/test_gradient.py
```diff
@@ -51,6 +53,71 @@ def test_gradient_checkpointing(self, shape, kernel, space_order):
         gradient2, _ = wave.jacobian_adjoint(residual, u0, vp=v0, checkpointing=False)
         assert np.allclose(gradient.data, gradient2.data)
 
+    @pytest.mark.parametrize('space_order', [4])
+    @pytest.mark.parametrize('kernel', ['OT2'])
+    @pytest.mark.parametrize('shape', [(70, 80)])
```
mloubout, line 58: that's quite small, may wanna go with something like `(101, 101)` and `tn=1500)

mloubout (commented):
i like it, as said on slack we may wanna add the other case `u.dt*v.dt`. Some comments on setup, time range may be the issue currently.

tests/test_gradient.py
navjotk: Since I have to setup the gradient operator by hand anyway, I thought it makes sense to keep it consistent between the forward and the gradient?

tests/test_gradient.py
navjotk: Those two would never be together in a realistic operator, correct? Plus there is the risk of the two expressions interacting in some way if they are in the same operator (e.g. CSE etc). What is the reason for combining them in one operator anyway?

tests/test_gradient.py
navjotk: Changed to these values. Test is still not passing though.

tests/test_gradient.py
mloubout: >What is the reason for combining them in one operator anyway?

Guarantees uses the same time bounds.

tests/test_gradient.py
mloubout: I get you, but since we only care about the gradient for a given `u` , makes the test bit less verbose. But if you prefer like that it's fine.

FabioLuporini:
So now the test is passing?

georgebisbas (approved):

navjotk:
Mathias pushed a passing version of the test with the following changes that I don't agree with:
- Using the WaveSolver class for the forward when we can't use it for the gradient computation anyway. I'd like to keep it consistent between the forward and the gradient computations. If we're not using the solver class for the gradient, we don't use it at all in the test. 
- Combining all the expressions in a single Operator. This is not realistic. A user will never have those three expressions in the same Operator. If we want this test to be representative, the three expressions should be used to create three separate Operators. If some "extra things" need to be done to make the test pass this way e.g. make sure that the bounds are the same, or that some objects are reset after running each operator, they should be made explicit instead of hiding them behind a single combined operator that will never be used in practice. 
- Changed to double precision. I think this test needs to pass in single precision. Probably not much more explanation required here. 

My latest commit here:
- Reverts all the above back to what I think is better. 
- Makes the test pass in single precision but the bounds are absurdly big tbh. 

If the test only passes in single precision for `rtol=1e-3, atol=1e-3`, is it really passing? More thought required on this.

mloubout:
>Using the WaveSolver class for the forward when we can't use it for the gradient computation anyway. I'd like to keep it consistent between the forward and the gradient computations. If we're not using the solver class for the gradient, we don't use it at all in the test.

This is not relevant for this test. You are testing `u.dt * v = u * v.dt2` which is basically a standard dot test, whether the `u` comes from the wavesolver or not doesn't matter as it should work no matter where `u` comes from (even a pure random array). Having all the `u` and forward definition in that test makes it unnecessarily verbose.

> Combining all the expressions in a single Operator. This is not realistic. A user will never have those three expressions in the same Operator.

A user will never use the three expressions but just choose one and go with it, I don't see the point here.

>Changed to double precision. I think this test needs to pass in single precision.

That's easily parametrizable.

navjotk:
> This is not relevant for this test. 
Agree. What I hear you saying is this is a style choice. I'd like to stand the ground on my style choice of consistency within the test. 

>A user will never use the three expressions but just choose one and go with it,
Which is why the Operators have to represent the three different choices the user might make instead of one Operator that nobody will ever create. If some CSE between the three expressions causes the results to change, for example - we want to make sure this test doesn't pass. 

>That's easily parametrizable. 
But as you might notice in my message above yours, the test essentially doesn't pass in single precision. Let's talk about that.

FabioLuporini:
Is it worth keeping both versions, as two different tests? the version with two Operators makes sense to me: @navjotk has a point about the potential impact of CSE and other symbolic transformations

navjotk:
If there is a point to having that combined test, it is orthogonal to the purpose of this PR.

navjotk:
I believe this should be good to go now.

navjotk:
Updated test to make sure we have a true and a smoothed model. Now we are really calculating the gradient.

tests/test_gradient.py
```diff
@@ -98,7 +104,7 @@ def test_gradient_equivalence(self, shape, kernel, space_order, preset, nbl, dty
         s = model.grid.stepping_dim.spacing
 
         eqn_adj = iso_stencil(v, model, kernel, forward=False)
-        receivers = rec.inject(field=v.backward, expr=rec * s**2 / m)
+        receivers = rec_true.inject(field=v.backward, expr=rec_true * s**2 / m_true)
```
mloubout, line 107: m0

navjotk: Does it matter which `m` was used to construct the expression, when all `op.apply()` calls are explicit about `m_true` and `m0`?

mloubout: Makes it more readable what is what

navjotk: Done

navjotk:
What remains for this @mloubout ?

PR #1520, opened 2020-11-27
GPU data streaming

### LINKED ISSUE #994: Simplify creation of Expression IET objects
Currently, to create an `Expression` one has two possibilities:

- `Expression(ClusterizedEq(Eq(lhs, rhs)))`
- `Expression(DummyEq(lhs, rhs))`

we need to simplify this. Often one needs to manually construct an IET and wouldn't want to create low level obejcts such as `ClusterizedEq/Eq/DummyEq ` to instantiate an Expression. We need to make it possible to

- `Expression(lhs, rhs)`

and then **internally**, in `Expression.__init__` , if we see that we've received two sympy expressions (lhs, rhs) instead of a `DummyEq` or a `ClusterizedEq`, we first of all create a `DummyEq` or `ClusterizedEq`, and then proceed as usual

### LINKED ISSUE #1479: Rework lowering of SteppingDimensions
There are two issues:

* Currently, index expressions involving SteppingDimensions are handled by the `iet_lower_dims` routine, which takes place right after constructing an IET from an STree. This is probably too early -- `iet_lower_dims` could be an actual iet_pass, first performing a visit of the IET to collect all uindices, and then applying the substitutions across all efuncs.

* ModuloDimensions are created from SteppingDimensions through the `build_iterators` routine, but this happens w/o using the SymbolRegistry. DerivedDimensions use instead a fairly rudimentary name cache, which should disappear in favor of the SymbolRegistry.

FabioLuporini:
Enjoy reviewing 😬 

fixes #994 
fixes #1479


devito/core/gpu_openacc.py
```diff
@@ -2,15 +2,15 @@
 
 import cgen as c
 
-from devito.core.gpu_openmp import (DeviceOpenMPNoopOperator, DeviceOpenMPIteration,
-                                    DeviceOmpizer, DeviceOpenMPDataManager)
-from devito.exceptions import InvalidOperator
+from devito.core.gpu_openmp import (DeviceOpenMPNoopOperator, DeviceOpenMPOperator,
```
ggorman, line 5: How about renaming gpu_openmp and it's members as it's used for both OpenMP and OpenACC.

FabioLuporini: yes. There is a TODO somewhere (I think in gpu_openacc) which explains that some major renaming (actually, refactoring) is necessary. This was like this before this PR, not sure if we really want to do that here or afterwards

devito/core/gpu_openmp.py
```diff
@@ -502,13 +646,72 @@ def _specialize_iet(cls, graph, **kwargs):
...
+        functions = (maybe_symbol.function,)
+    except AttributeError:
+        assert maybe_symbol.is_Node
+        iet = maybe_symbol
+        functions = set(FindSymbols().visit(iet))
+        if only_writes:
+            expressions = FindNodes(Expression).visit(iet)
+            functions &= {i.write for i in expressions}
+
+    return all(not (f.is_TimeFunction and f.save is not None and f not in gpu_fit)
```
mloubout, line 694: what about large buffered time function, like `Buffer(O(1e2))` should this be considered or is it too empirical?

devito/ir/clusters/algorithms.py
```diff
@@ -186,3 +194,100 @@ def guard(clusters):
...
+            # buffers (save=None) *but* different modulo extent. The `t+1`
+            # indices above are therefore conceptually different, so they will
+            # be replaced with the proper ModuloDimension through two different
+            # calls to `xreplace_indices`
+            exprs = c.exprs
+            groups = as_mapper(mds, lambda d: d.modulo)
+            for size, v in groups.items():
+                mapper = {md.origin: md for md in v}
+
+                func = partial(xreplace_indices, mapper=mapper, key=partial(rule, size))
```
mloubout, line 283: How expensive is that? `partial` always makes me worried for some reason.

devito/ir/equations/algorithms.py
```diff
@@ -141,6 +141,10 @@ def lower_exprs(expressions, **kwargs):
             indices = [(lower_exprs(a) + o) for a, o in
                        zip(i.indices, f._size_nodomain.left)]
 
+            # Substitute spacing (spacing only used in own dimension)
+            indices = [i.xreplace({d.spacing: 1, -d.spacing: -1})
+                       for i, d in zip(indices, f.dimensions)]
```
mloubout, line 146: why is this here now?

devito/ir/equations/equation.py
```diff
@@ -158,13 +168,15 @@ def __new__(cls, *args, **kwargs):
...
-        # Construct the conditionals
+        # Construct the conditionals and replace the ConditionalDimensions in `expr`
         conditionals = {}
         for d in conditional_dimensions:
             if d.condition is None:
                 conditionals[d] = CondEq(d.parent % d.factor, 0)
             else:
                 conditionals[d] = diff2sympy(lower_exprs(d.condition))
+            if d.factor is not None:
+                expr = uxreplace(expr, {d: IntDiv(d.index, d.factor)})
```
mloubout, line 179: what cases is this needed for?

mloubout (commented):
first pass

devito/core/cpu.py
```diff
@@ -371,14 +383,50 @@ def _make_iet_passes_mapper(cls, **kwargs):
...
+        'collect-derivs',
+        # Expressions
+        'buffering',
+        # Clusters
+        'blocking', 'topofuse', 'fuse', 'factorize', 'cire-sops', 'cse',
+        'lift', 'opt-pows',
+        # IET
+        'denormals', 'optcomms', 'openmp', 'mpi', 'simd', 'prodders',
+    )
+    _known_passes_disabled = ('tasking', 'streaming', 'gpu-direct')
```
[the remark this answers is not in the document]
FabioLuporini: added

devito/ir/clusters/algorithms.py
FabioLuporini: I don't know exactly how long that piece of code takes (I don't see why it should be slow though, I think you're biased when it gets to `partial`?) but yours, more generally is still a good question -- "how about the new compilation time for these Operators"?

So I took TTI so=4 and ran it with master and in my branch. I get:
* master: 0.27s for the pad Operators, 4.5s for the main TTI Operator
* this branch: 0.29s for the pad Operators, 5s for the main TTI Operator

So it is a bit slower now, I could run a profiler to find out why

FabioLuporini: EDIT: well there's a whole new Cluster-level pass, the `lower_dimensions`, which creates the ModuloDimension, rather than using that awful global cache attached to DerivedDimension, which btw was a hack (causing failures here) and therefore has now been dropped

devito/ir/equations/algorithms.py
FabioLuporini: I asked you about it some weeks ago in #development on Slack. We agreed that "we would have talked about it when I would have opened this PR".

IIRC, the reason is that the new optional GPU pass, `buffering`, introduces Arrays to implement the buffers. These buffers, just like any other Functions, require lowering -- they have a halo, may be using SubDimensions, and so on. 

So, I construct these copy-back `Eq`s , from the buffer to the original Function , e.g.

```
 Eq(u[time + dt, x, y], ub[sb0 + dt, x, y])]
```

`ub` is the buffer -- here accessed via an indexed Array, whose `dt` also needs to be turned into 1, just like any `dt` (any spacing symbol) appearing in any Indexed expressed with the Indexed notation

More generally, if we ever wanted to expose Array to the user API, or simply use it in the higher layers of the compiler (just like I'm doing here), I need them to be lowered too.

Does this make sense?

FabioLuporini: EDIT: and yeah, the buffering tests in `test_lower_exprs.py` start failing without it

devito/ir/equations/equation.py
FabioLuporini: it's the result of lifting [this routine](https://github.com/devitocodes/devito/blob/master/devito/ir/iet/scheduler.py#L95) from the guts of the compiler up here. So it's an improvement (logically, this is where this sort of lowering should be performed)

devito/ir/iet/algorithms.py
```diff
@@ -0,0 +1,45 @@
+from collections import OrderedDict
```
FabioLuporini, line 1: note @mloubout : this file previously was https://github.com/devitocodes/devito/blob/master/devito/ir/iet/scheduler.py

I changed its name since I finally managed to remove the iet_lower* routines (which were disgusting -- performing that sort of Dimension lowering when we already have an IET... ).

devito/core/gpu_openmp.py
FabioLuporini: if you use Buffer, then it means that `save != None`, so they will still be captured according to this default strategy and therefore be subjected to streaming. Does it make sense?

devito/core/gpu_openmp.py
```diff
@@ -502,13 +646,72 @@ def _specialize_iet(cls, graph, **kwargs):
...
+    Parameters
+    ----------
+    maybe_symbol : Indexed or Function or Node
+        The inspected object. May be a single Indexed or Function, or even an
+        entire piece of IET.
+    gpu_fit : list of Function
+        The Function's which are known to definitely fit in the device memory. This
+        information is given directly by the user through the compiler option
+        `gpu-fit` and is propagated down here through the various stages of lowering.
+    only_writes : boo, optional
```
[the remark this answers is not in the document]
FabioLuporini: fixed

devito/ir/clusters/algorithms.py
mloubout: That's a quite small variance I'd say so it's fine.

devito/ir/equations/algorithms.py
mloubout: But the `lower_exprs` the line just above should have indexified and do the replacement already so this really shouldn't be necessary

devito/ir/iet/utils.py
```diff
@@ -102,7 +99,7 @@ def derive_parameters(iet, drop_locals=False):
     free_symbols = FindSymbols('free-symbols').visit(iet)
 
     # Filter out function base symbols and use real function objects
-    function_names = [s.name for s in functions]
+    function_names = set(flatten([(s.name, s._C_name) for s in functions]))
```
mloubout, line 102: can probably use `set(*[...` rather than flatten

devito/ir/iet/visitors.py
```diff
@@ -542,13 +542,13 @@ def visit_Node(self, o, ret=None, parents=None, in_parent=False):
...
 
-    def quick_str(o):
+    def quick_key(o):
         """
         SymPy's __str__ is horribly slow so we stay away from it for as much
         as we can. A devito.Indexed has its own overridden __str__, which
         relies on memoization, which is acceptable.
         """
-        return str(o) if o.is_Indexed else o.name
+        return (str(o) if o.is_Indexed else o.name, type(o))
```
mloubout, line 551: Why do you need the type now? In case two objects have same name?

devito/ir/iet/utils.py
FabioLuporini: fairly sure I tried but for whatever reason it wasn't working, can try again

devito/ir/iet/visitors.py
FabioLuporini: yep, this is a bug fix, probably explained in the relative commit message

devito/ir/equations/algorithms.py
FabioLuporini: no, the timing/ordering is different (you can easily see it if you drop the line and run any of the buffering tests in test_lower_exprs.py)

devito/core/gpu_openmp.py
```diff
@@ -463,19 +614,19 @@ def _make_passes_mapper(cls, **kwargs):
...
+        'collect-derivs',
+        # Expressions
+        'buffering',
+        # Clusters
+        'blocking', 'tasking', 'streaming', 'factorize', 'fuse', 'lift',
+        'cire-sops', 'cse', 'opt-pows', 'topofuse',
+        # IET
+        'optcomms', 'openmp', 'mpi', 'prodders', 'gpu-direct'
+    )
+    _known_passes_disabled = ('denormals', 'simd')
```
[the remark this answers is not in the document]
FabioLuporini: added, resolving conversation

devito/types/array.py
```diff
@@ -46,15 +49,26 @@ class Array(ArrayBasic):
...
+        though proper synchronization is needed (i.e., will be introduced by
+        the Devito compiler) to ensure data coherence. 'uniform', instead,
+        means the Array will be used as if the underlying node had uniform
+        memory address space.
     scope : str, optional
-        Control memory allocation. Allowed values: 'heap', 'stack'. Defaults to 'heap'.
+        The scope in the given memory space. Allowed values: 'heap', 'stack'.
+        Defaults to 'heap'. This may not have an impact on certain platforms.
     sharing : str, optional
         Control data sharing. Allowed values: 'shared', 'local'. Defaults to 'shared'.
```
ggorman, line 69: This boat has sailed, but I feel 'local' is ambiguous; just because data is local doesn't mean it cannot be accessed 'remotely'. 'private' has a clearer intent and would be consistent with other programming models.

devito/types/sync.py
```diff
@@ -0,0 +1,328 @@
+"""
+Support types to implement asynchronous execution via e.g., threading,
```
ggorman, line 2: The comment appears to conflict with the file title. Add another sentence of explanation for us mere mortals?

examples/performance/01_gpu.ipynb
```diff
@@ -172,7 +172,7 @@
...
...IJHITYughQv30ioYWrUPP1ZtihYRBJmkuDDoY20pRIgfaRg54teiElwV/Vucd5g7MlCI5/Xe8z4/ODDn3OG+933nuRfOzLl34EBmPg38DuztcexDwInMfAbYVB5HlflHxFpgH/B8Zj4LPAS8Sd35P3C2b/s02j403b/t236r7UOD/dv+Attvq32YsP4zs/cN2AZ8NbI/C8zWGHtkzKPAK8AFYLocmwYu9DjmDF1oLwLHgaD7Z85TS63LfR77UeAy5X3cI8erzB9YC1wBVgNTZf6v1pr/uGy2b/uttr/UPFvo3/Ztf6l5ttB+uf8m+7f9RWth+w21X+57Yvqvdanz/ILMmyvHqoiIdcAW4CSwJjOvlZuuA2t6HPog8A7wT9l/HPgjM/8u+32uw1PALeDjctnFhxGxkkrzz8yrwPvAb8A14E/gNPXmPy5sv2P77bUPbfZv+7YPbbYPjfZv+4vYfqeJ9mGy+h/8h1tFxCPAF8DbmXl79LbsfgXRy8daR8RrwM3MPN3H/f8PU8BzwAeZuQW4w12XOPQ8/1XAbron4xPASmBnH2NpabZv+y17EP3bvu2PA1/76/dv++PB9n3t/y+1TnyvAk+O7M+UY72KiIfpngCfZuaRcvhGREyX26eBmz0N/wLwekT8CnxGd+nDIeCxiJgq39PnOswBc5l5suwfpntS1Jr/y8DlzLyVmX8BR+jWpNb8x4Xt236r7UN7/du+7c9rrX1ou3/bX2D7bbUPE9R/rRPfH4AN5dO9ltG94flYnwNGRAAfAecyc//ITceAPeXrPXTvA7jvMnM2M2cycx3dfL/NzLeA74A3Kox/HbgSERvLoZeAs1SaP93lDlsjYkX5WcyPX2X+Y8T2bb/V9qGx/m3f9kc01T4037/tL7D9ttqHSer/Xt4YfC8bsAv4GfgFeLfCeNvp/qT/E/Bj2XbRXXP/DXAR+BpYXeGx7ACOl6/XA98Dl4DPgeU9jrsZOFXW4EtgVc35A+8B54EzwCfA8przH5fN9m2/1fbLWjTZv+3bfqvtl8fSXP+2v2gtbD/bab+MPxH9R3mwkiRJkiQN0uA/3EqSJEmS1DZPfCVJkiRJg+aJryRJkiRp0DzxlSRJkiQNmie+kiRJkqRB88RXkiRJkjRonvhKkiRJkgbNE19JkiRJ0qD9C+HJjXGzJ4sIAAAAAElFTkSuQmCC\n",
```
ggorman, line 175: Can we clear output before committing Jupyter notebooks as some of these are huge. It will make a happier sized repo as well as less insane diff's.

examples/performance/README.md
```diff
@@ -26,6 +26,10 @@
     * `cire-repeats-inv` (int, 1): control detection of dimension-invariants
     * `cire-mincost-inv` (int, 50): minimum cost of a dimension-invariant candidate
     * `cire-onstack` (boolean, False): control scope of temporary arrays (on the heap by default)
+  * Device-specific:
+    * `gpu-fit` (boolean, False): control movement of saved TimeFunctions between host and device
```
ggorman, line 30: Not clear from comment why this is 'fit'

examples/performance/README.md
```diff
@@ -26,6 +26,10 @@
...
+    * `gpu-direct` (boolean, False): tells Devito a GPU-aware MPI distribution will be used
```
ggorman, line 31: Ambiguous - this is not what GPU DIrect means.

tests/test_dimension.py
```diff
@@ -1047,3 +1048,126 @@ def test_no_fusion_convoluted(self):
         assert len(exprs) == 3
         assert exprs[1].expr.rhs is exprs[0].output
         assert exprs[2].expr.rhs is exprs[0].output
+
+
+class TestMashup(object):
```
ggorman, line 1053: Give it a more instructive name?

ggorman (approved):
Appears GTG. I've added some minor comments.

devito/core/gpu_openacc.py
```diff
@@ -228,39 +270,37 @@ def _specialize_iet(cls, graph, **kwargs):
...
+        # Expressions
+        'buffering',
+        # Clusters
+        'blocking', 'tasking', 'streaming', 'factorize', 'fuse', 'lift',
+        'cire-sops', 'cse', 'opt-pows', 'topofuse',
+        # IET
+        'optcomms', 'openacc', 'orchestrate', 'mpi', 'prodders'
+    )
+    _known_passes_disabled = ('openmp', 'denormals', 'simd', 'gpu-direct')
+    assert not (set(_known_passes) & set(_known_passes_disabled))
```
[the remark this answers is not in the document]
georgebisbas: I agree on this.

devito/passes/iet/orchestration.py
```diff
@@ -0,0 +1,366 @@
...
+        iet = While(CondNe(FieldFromPointer(sdata._field_flag, sdata.symbolic_base), 0),
+                    iet)
+
+        return Callable(name, iet, 'void', parameters, 'static'), sdata
+
+    def __make_threads(self, value=None):
+        name = self.sregistry.make_name(prefix='threads')
+
+        if value is None:
+            threads = STDThreadArray(name=name, nthreads_std=1)
```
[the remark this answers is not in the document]
FabioLuporini: Good point. This has already been renamed to PThreads in the branch on top of this one. This branch really requries STD threads; C++'s STD uses pthreads to implement STD threads on Linux, but something else for other OSs. So the name STDThread is technically correct here. In the branch on top of this one we generate pure C, not C++ (for the reasons already discussed offline), so they become Pthreads.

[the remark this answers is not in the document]
FabioLuporini: sneak peek...

https://github.com/devitocodes/devito/blob/amd-gpu-from-clusters/devito/types/sync.py#L81
https://github.com/devitocodes/devito/blob/amd-gpu-from-clusters/devito/types/sync.py#L124

devito/types/array.py
FabioLuporini: noted. Will change

examples/performance/README.md
FabioLuporini: OK, improving

examples/performance/README.md
FabioLuporini: Right, changing

devito/operator/operator.py
```diff
@@ -174,7 +173,7 @@ def _symbol_registry(cls):
         nthreads = NThreads(aliases='nthreads0')
         nthreads_nested = NThreadsNested(aliases='nthreads1')
         nthreads_nonaffine = NThreadsNonaffine(aliases='nthreads2')
-        threadid = CustomDimension(name='tid', symbolic_size=nthreads)
+        threadid = ThreadID(nthreads)
```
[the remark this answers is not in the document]
FabioLuporini: Jeez, I had written a detailed explanation of why that is not the case here, but the browser crashed...

Basically it boils down to the fact that `ThreadID` _is_ a `Dimension` to iterate over host threads. We never ever need to refer explicitly to GPU threads. Also, this is the main Operator class, and there's no device-related concept here in this module

devito/tools/decorators.py
```diff
@@ -203,3 +203,10 @@ def __enter__(self):
     def __exit__(self, *args):
         self.timings[self.name] = time() - self.tic
         del timed_pass.timings[get_ident()]
+        try:
```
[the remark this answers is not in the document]
FabioLuporini: again, I had a detailed explanation and the browser crashed, sob...

Basically: we don't need to wrap the Operator construction with a try-except because there's no global state (!) and therefore any exceptions arising during compilation can be propagated to the user without the need for any cleanup code...

... aside from this profiling-related stuff. But that's exactly why this class implements a context manager. This `__exit__` method _always_ gets executed (even in presence of exceptions), so we can execute the cleanup here safely, without exposing the internals of this class to the outside world (the Operator class, in our case...)

devito/types/sync.py
FabioLuporini: OK, improving docstring

PR #1523, opened 2020-11-30
operator: Patch grid-carried argument overrides

### LINKED ISSUE #1518: spacing seems to come from grid that operator is originally made with
Hi!

Not sure when this one crept in, or how we haven't hit it until now.  I'm assuming this isn't expected behaviour...?

```
def test_spacing_from_new_grid():
    from devito import Grid, Function, Operator, Eq
    grid = Grid(shape=(10, 10), extent=(9, 9))
    u = Function(name='u', grid=grid, space_order=1)

    # A bogus operator that just assigns the x spacing into the array
    # Note, grid.dimensions[0].spacing here is not a number, it's the symbol h_x
    op = Operator(Eq(u, grid.dimensions[0].spacing))

    # Create a new grid with different spacing, and a function defined on it
    grid2 = Grid(shape=(5, 5), extent=(9, 9))
    u2 = Function(name='u', grid=grid2, space_order=1)
    op(u=u2)

    # Fails: the h_x that was passed to the C code is from `grid`, not `grid2`
    assert u2.data[2, 2] == grid2.spacing[0]
```

Thanks!

FabioLuporini:
fixes #1518

### DIFF
--- a/devito/types/dimension.py
+++ b/devito/types/dimension.py
@@ -6,7 +6,7 @@ from cached_property import cached_property
 
 from devito.data import LEFT, RIGHT
 from devito.exceptions import InvalidArgument
-from devito.logger import debug
+from devito.logger import debug, warning
 from devito.tools import Pickable, dtype_to_cstr, is_integer, memoized_meth
 from devito.types.args import ArgProvider
 from devito.types.basic import Symbol, DataSymbol, Scalar
@@ -233,7 +233,7 @@ class Dimension(ArgProvider):
         interval : Interval
             Description of the Dimension data space.
         grid : Grid
-            Only relevant in case of MPI execution; if ``self`` is a distributed
+            Used for spacing overriding and MPI execution; if ``self`` is a distributed
             Dimension, then ``grid`` is used to translate user input into rank-local
             indices.
         **kwargs
@@ -262,7 +262,20 @@ class Dimension(ArgProvider):
             except (AttributeError, TypeError):
                 pass
 
-        return {self.min_name: loc_minv, self.max_name: loc_maxv}
+        args = {self.min_name: loc_minv, self.max_name: loc_maxv}
+
+        # Maybe override spacing
+        if grid is not None:
+            try:
+                spacing_map = {k.name: v for k, v in grid.spacing_map.items()}
+                args[self.spacing.name] = spacing_map[self.spacing.name]
+            except KeyError:
+                pass
+            except AttributeError:
+                # See issue #1524
+                warning("Unable to override spacing")
+
+        return args
 
     def _arg_check(self, args, size, interval):
         """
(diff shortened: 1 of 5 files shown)


devito/types/dimension.py
```diff
@@ -262,7 +262,20 @@ def _arg_values(self, args, interval, grid, **kwargs):
...
                 pass
 
-        return {self.min_name: loc_minv, self.max_name: loc_maxv}
+        args = {self.min_name: loc_minv, self.max_name: loc_maxv}
+
+        # Maybe override spacing
+        if grid is not None:
+            try:
+                spacing_map = {k.name: v for k, v in grid.spacing_map.items()}
+                args[self.spacing.name] = spacing_map[self.spacing.name]
```
mloubout, line 271: why not just 
`args[self.spacing.name] = grid.spacing_map[self.spacing]`

FabioLuporini: see issue #1524 

bottom line: I tried but it breaks and I couldn't come up with a quick patch

mloubout: `args[self.spacing.root.name] = grid.spacing_map[self.spacing.root]` this should always be valid no?  The case where spacing is an expression rather than a symbol should in principle always link to the parent.

FabioLuporini: this one could work, but it still "feels weird" that we have to pick the root in a superclass method

FabioLuporini: with this patch:

```
-                spacing_map = {k.name: v for k, v in grid.spacing_map.items()}
-                args[self.spacing.name] = spacing_map[self.spacing.name]
+                args[self.root.spacing.name] = grid.spacing_map[self.root.spacing]
```

the new test fails with:

```
...

        # The h_x that was passed to the C code must be the one `grid2`, not `grid`
>       assert u2.data[2, 2] == grid2.spacing[0]
E       assert 1.0 == 2.25

test_operator.py:1198: AssertionError
...
```

mloubout: Ooohhhhh, ok now I see it
- `args[self.spacing.name] ` this one is always correct, that's the one to be replaced `self.root.spacing.name` is not in `args`
- Then the `spacing_map` part is needed because while the new `grid` may have the same spacing name `self.spacing` won't be recognized as a key for `grid.spacing_map`.

But now I'm wondering, what if the original gird is `h_x, h_y` and then the new grid is `d_x, d_y`  then `spacing_map['h_x']` wont work with `a `grid2` witch custom dimensions and spacing.
This feels like we need first a dimension ordering in grid, then check the same number of dim then replace spacing by order.

FabioLuporini: for the record:
as discussed offline, this is a bit tricky to do, and since virtually nobody cares about this feature, we decided to proceed with merging

Issue #1528, opened 2020-12-01
linalg.py failing with PGI openacc


georgebisbas:
Failing on the reduction clause
to reproduce:
pgi+openacc

```
export DEVITO_LANGUAGE=openacc
export DEVITO_PLATFORM=nvidiaX
export DEVITO_ARCH=pgcc
export DEVITO_LOGGING=DEBUG #optional
```

`python3 misc/linalg.py mat-vec`

Some discussion: https://devitocodes.slack.com/archives/CQ0AT90R0/p1607003860367600

FabioLuporini:
@Leitevmd have you guys ever encountered this issue?  this could _potentially_ be relevant to you

georgebisbas:
After merging 
https://github.com/devitocodes/devito/pull/2226
the generated code for the above MFE is:


```
  #pragma acc parallel loop present(A,b,x)
  for (int i = i_m; i <= i_M; i += 1)
  {
    for (int j = j_m; j <= j_M; j += 1)
    {
      b[i] += x[j]*A[i][j];
    }
  }
```

instead of

```
  START_TIMER(section0)
  #pragma acc parallel loop collapse(2) reduction(+:b[0:b_vec->size[0]]) present(A,b,x)
  for (int i = i_m; i <= i_M; i += 1)
  {
    for (int j = j_m; j <= j_M; j += 1)
    {
      b[i] += x[j]*A[i][j];
    }
  }
  STOP_TIMER(section0,timers)
```

georgebisbas:
The above generated code is working but probably not optimal.
Should we close this issue or rename it so as to improve with a 
reduction sub-pass?

georgebisbas:
This can be probably closed now, as reduction is avoided, in purpose as I understand from #2226

FabioLuporini:
have u tested it recently?

but regardless, probably it can indeed be closed

PR #1529, opened 2020-12-04
[WIP_TEST] examples: check reductions

georgebisbas:
WIP

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -79,6 +79,10 @@ jobs:
       shell: bash -l {0}
       run: |
         python examples/misc/linalg.py mat-vec mat-mat-sum transpose-mat-vec
+        python examples/misc/linalg.py mat-mat-sum --optimize
+        python examples/misc/linalg.py mat-mat --optimize
+        python examples/misc/linalg.py mat-vec --optimize
+        python examples/misc/linalg.py transpose mat-vec --optimize
 
     - name: CFD examples
       shell: bash -l {0}
--- a/examples/misc/linalg.py
+++ b/examples/misc/linalg.py
@@ -24,7 +24,7 @@ def option_basic(f):
 
     def callback_opts(ctx, param, value):
         if value is True:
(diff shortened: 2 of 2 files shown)


.github/workflows/examples.yml
```diff
@@ -79,6 +79,10 @@ jobs:
       shell: bash -l {0}
       run: |
         python examples/misc/linalg.py mat-vec mat-mat-sum transpose-mat-vec
+        python examples/misc/linalg.py mat-mat-sum --optimize
+        python examples/misc/linalg.py mat-mat --optimize
+        python examples/misc/linalg.py mat-vec --optimize
+        python examples/misc/linalg.py transpose-mat-vec --optimize
```
mloubout, line 85: what does this do again?

georgebisbas: Checks the linalg with `advanced`, it was `noop` only before.

FabioLuporini (approved):

FabioLuporini (changes requested):
as discussed offline, we should either made the command a one liner or drop the `--optimize` argument by defaulting to `advanced`

georgebisbas:
Temporarily on hold due to: https://github.com/devitocodes/devito/issues/1528

examples/misc/linalg.py
```diff
@@ -24,7 +24,7 @@ def callback_shape(ctx, param, value):
 
     def callback_opts(ctx, param, value):
         if value is True:
-            return ('blocking', 'simd', 'openmp', {'blockinner': True})
+            return 'advanced'
```
FabioLuporini, line 27: this is not doing the same blocking/simd as `master`, because `blockinner: True` is dropped. Why?

georgebisbas: I vaguely remember that we talked about this in the past, maybe I am wrong....hmmm.. or it was due to the fact that no blocking happening nevertheless? Anyway reinstated.

FabioLuporini (changes requested):
Two commits for such a tiny change -> they need rebasing and squashing

PR #1532, opened 2020-12-10
[WIP] Add gradient subsampling test

navjotk:
This PR is the next milestone leading up to the compression followup work. It adds a test that computes the gradient normally first, then does it again for `dt/2` and then a third time for `dt/2` but subsampling 1-in-2 in time. It then asserts equality between the three, ensuring that the subsampling implementation makes sense. 

Currently, the norms for the three are:
```
grad_base 1498.496997261012 grad_2t 2996.1773245927657 grad_sub 1498.087825898719
```

which means the second case (where `dt/2` is used without subsampling) probably needs a dt-related scaling factor somewhere. Can @mloubout point me to the best way to incorporate this factor so this test makes sense?

Also, even though norms are fairly close, test does not pass at `tolerance=1e-3`. Might be a silly mistake that crept in. Looking for overall feedback at this point.

### DIFF
--- a/tests/test_gradient.py
+++ b/tests/test_gradient.py
@@ -3,7 +3,7 @@ import numpy as np
 import pytest
 from numpy import linalg
 
-from devito import Function, info, TimeFunction, Operator, Eq
+from devito import Function, info, TimeFunction, Operator, Eq, ConditionalDimension
 from devito.parameters import switchconfig
 from examples.seismic.acoustic.acoustic_example import smooth, acoustic_setup as setup
 from examples.seismic.acoustic.operators import iso_stencil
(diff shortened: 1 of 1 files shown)


tests/test_gradient.py
```diff
@@ -139,6 +139,143 @@ def test_gradient_equivalence(self, shape, kernel, space_order, preset, nbl, dty
...
+    @pytest.mark.parametrize('tn', [750.])
+    @pytest.mark.parametrize('spacing', [(10, 10)])
+    @pytest.mark.parametrize("dtype, tolerance", [(np.float32, 1e-4),
+                                                  (np.float64, 1e-12)])
+    @pytest.mark.parametrize('nbl', [40])
+    @pytest.mark.parametrize('preset', ['layers-isotropic'])
+    @pytest.mark.parametrize('space_order', [4])
+    @pytest.mark.parametrize('kernel', ['OT2'])
+    @pytest.mark.parametrize('shape', [(101, 101)])
+    def test_gradient_subsampling(self, shape, kernel, space_order, preset, nbl, dtype,
```
mloubout, line 151: I said it for the previous PR but I find these tests completely unreadable with all the clutter and redefinitions. There is perfectly sable setup methods in the example for this.

That `parametrize` is also IMO way too verbose for single parameters.

navjotk: I understand, but the point is for this code to be used directly for the upcoming paper - and we would like that code to be pure Devito code (Perhaps @ggorman could help me out here), and we're reusing that code as tests for Devito. These are not meant to be unit tests - they are closer to functional tests.

Happy to take any more specific feedback that doesn't involve using the wrappers from 'examples', because I think we won't be able to use those for the paper.

FabioLuporini: I rather have an orthogonal comment. If we really want such a big test, we _at least_ need asserts against expected norms after each op.apply, because otherwise if something goes south earlier, go figure...

navjotk: I don't like the idea of hard-coding expected norms in the test tbh but if that's the only way you see of achieving the goal you have in mind, then so be it. 

What would these norms be of? Receiver data? We could equally reinterpolate all receiver data to the same dt, and assert equality/closeness between them - this way no hard-coded norms are required.

FabioLuporini: yeah, agree it's not ideal. Basically all we need is a mechanism to check correctness after each step (ie after each `apply`)

yes, receivers should be fine, as transitively that should (sort of...) imply correctness of the wavefield norms too

mloubout (commented):
Just to speed up, the integration rule translate to :

- Standard for any time axis ` sum (time.spacing * u.dt2 *v)`
- Subsampled `sum(tsave.spacing * u.dt2 * v)`
-
So whatever time axis you integrate a function` func` over it's `time.spacing * func(time)`

mloubout:
Stale, closing

PR #1537, opened 2020-12-17
Migrate gpu workflow to new runners

rhodrin:
As the title says.

A couple of things that still need fixing:
- Using `source` all the time needs fixing, but still need to look into the best way of doing this the recent actions security fixes.
- ~openmp+mpi examples are being skipped due to #1536. Note that openmp+mpi is still being tested somewhat in the default suite of tests.~
- Edit: openmpi+mpi tests are now run individually until #1536 is resolved.

PR #1542, opened 2021-01-04
CI: Update gpu workflow

rhodrin:
This, for the time being, supersedes #1537. GPU tests will now be run on demand via workflow dispatch and the runners started and stopped during the workflow.


FabioLuporini:
Merged.

PR #1547, opened 2021-01-06
Refactor and enhance section profiling

FabioLuporini:
Now also track the time spent in:

* halo-exchanges
* busy-waits when streaming data on/off device

Such finer breakdown is emitted when `DEVITO_PROFILING` is set to `advanced1` or `advanced2`.


devito/ir/iet/nodes.py
```diff
@@ -961,12 +968,13 @@ class Section(List):
...
 
-    def __init__(self, name, body=None):
+    def __init__(self, name, body=None, is_subsection=False):
         super(Section, self).__init__(body=body)
         self.name = name
+        self.is_subsection = is_subsection
 
     def __repr__(self):
-        return "<Section (%d)>" % len(self.body)
+        return "<Section (%s)>" % self.name
```
georgebisbas, line 977: Any chance ever two sections having the same name?

georgebisbas (approved):
Seems fine to me. Is there any strong reason why someone should use advanced1 instead of advanced2? IOW does their diff in terms of profiling justify the existence of both modes? Can we get away only with advanced2 ?

devito/ir/iet/nodes.py
FabioLuporini: I don't think so, no, they are unique by construction. Also, if they ever were, you'd get a compilation error due to duplicate variable names

FabioLuporini:
> Is there any strong reason why someone should use advanced1 instead of advanced2? IOW does their diff in terms of profiling justify the existence of both modes? Can we get away only with advanced2 ?

This PR is just providing the basic infrastructure for enhanced profiling, but details (and names) might change in the near future. Said that, the difference between  `advanced1` and `advanced2` is that the former groups consecutive halo exchanges together, while the latter provides a per-halo-exchange breakdown. I think you can see this with elastic_example. I will update the wiki once the PR lands

devito/ir/iet/nodes.py
```diff
@@ -742,14 +742,21 @@ class TimedList(List):
...
+                         body=body,
+                         footer=c.Line('STOP_TIMER(%s,%s)' % (lname, timer.name)))
+
+    @classmethod
+    def _start_timer_header(cls):
+        return ('START_TIMER(S)', ('struct timeval start_ ## S , end_ ## S ; '
+                                   'gettimeofday(&start_ ## S , NULL);'))
+
+    @classmethod
+    def _stop_timer_header(cls):
```
ggorman, line 756: Should this not be _stop_timper_footer?

devito/ir/iet/nodes.py
ggorman: How about explicitly catching duplicates so Devito can handle the error more gracefully rather than relying upon the end compiler.

devito/mpi/routines.py
```diff
@@ -898,6 +908,35 @@ class IrecvCall(Call):
...
+
+
+class HaloUpdateList(MPIList):
+    pass
+
+
+class HaloWaitList(MPIList):
+    pass
+
+
```
ggorman, line 939: Why are all these virtual definitions necessary?

devito/operator/operator.py
```diff
@@ -131,7 +131,7 @@ class Operator(Callable):
     refer to the relevant documentation.
     """
 
-    _default_headers = ['#define _POSIX_C_SOURCE 200809L']
+    _default_headers = [('_POSIX_C_SOURCE', '200809L')]
```
ggorman, line 134: Why is #define dropped?

ggorman (approved):
Looks GTG

devito/ir/iet/nodes.py
FabioLuporini: as I wrote above:

>  they are unique by construction

that is, there cannot be duplicates

devito/mpi/routines.py
FabioLuporini: Because they define special types that I need to search for in later stages of the compilation pipeline. For example, at some point I need to run an IET visitor catching all `HaloWaitCall` nodes, ie `FindNodes(HaloWaitCall).visit(...)`

devito/operator/operator.py
FabioLuporini: it's added by the actual code generator in the last stage of the compilation pipeline. The information provided here is now more structured than before, as the define _name_ and _value_ are explicitly separated

devito/ir/iet/nodes.py
FabioLuporini: good catch. Since this appears to be the only (tiny) oversight of this PR, I'll fix it in the next imminent PR

FabioLuporini:
All GPU tests were run and passed CI: https://github.com/devitocodes/devito/actions/runs/471793778

Merging

PR #1549, opened 2021-01-06
sympy: Support v1.8

### LINKED ISSUE #984: Sympy standard function with our types
So basically the issue is that all of sympy standard functions ignore our type or transform into their own

For example `sqrt(u)` is a `sympy.Pow` while we would want a `Differentiable.Pow`. I am unsure how to proceed about this, most of the ideas I would require to import everything in Devito and add our own version and would break the moment someone uses a function we didn;t think of (there is like 100 in sympy so not gonna redo everything)

Open to suggestions

mloubout:
Upgrade deps to sympy 1.8.

- Add our own linear solve that only evaluates time derivatives to allow CIRE to work on unevaluated expression
- Add all of sympy's elementary function as our own differentiable ones, fixes #984


devito/types/equation.py
```diff
@@ -242,8 +241,8 @@ def solve(eq, target, **kwargs):
     for e, t in zip(as_tuple(eq), as_tuple(target)):
         # Try first linear solver
         try:
-            cc = linear_coeffs(e._eval_at(t).evaluate, t)
-            sols.append(-cc[1]/cc[0])
+            sol, = sympy.linsolve([e._eval_at(t).evaluate], t)
```
FabioLuporini, line 244: was `linear_coeffs` removed? for retrocompatibility you may try:

```
try:
    from sympy.solvers.solveset import linear_coeffs
except ImportError:
    ...
```

FabioLuporini (commented):
yeah, if not too difficult we probably want at least 1.5? 1.4 is a bit too old I think?

devito/types/equation.py
mloubout: No that's the issue, it's still there but it's been modified so it doesn't work anymore. Basically it use to check if `eq.has(target)` and now it only checks `target in eq.free_symbols` which fails because only the dimensions are in it, so I'm tempted to make a dummy type just for solve that adds `target` to its free symbols to bypass it because `linsolve` is not an option it's way too slow.

devito/builtins/arithmetic.py
```diff
@@ -20,6 +19,7 @@ def norm(f, order=2):
     order : int, optional
         The order of the norm. Defaults to 2.
     """
+    from devito.finite_differences.differentiable import Pow
```
FabioLuporini, line 22:
```
Pow = dv.finite_differences.differentiable.Pow
```
?

devito/finite_differences/differentiable.py
```diff
@@ -347,6 +348,19 @@ def _eval_is_zero(self):
         return None
 
 
+class DifferentiableFunction(DifferentiableOp):
+
+    def __new__(cls, *args, **kwargs):
+        return cls.__sympy_class__.__new__(cls, *args, **kwargs)
```
FabioLuporini, line 354: so here we're 100% sure we have devito.Mul/Add/... otherwise we would get an AttributeError on `__sympy_class__`

devito/finite_differences/elementary.py
```diff
@@ -0,0 +1,805 @@
+import sympy
+
+from devito.finite_differences.differentiable import (DifferentiableFunction, diffify)
+
+
+class factorial(DifferentiableFunction, sympy.factorial):
+    __sympy_class__ = sympy.factorial
+    __new__ = DifferentiableFunction.__new__
```
FabioLuporini, line 8: all these `__new__ = ...` in this file can probably be dropped. I know you took inspiration from what I did in `Add/Mul/...` in an old PR of my (current master), but in my current branch I've removed them and everything still works like a charm, as expected I'd say, because w/ no override the chosen `__new__` is the one of the first superclass.

devito/operations/solve.py
```diff
@@ -0,0 +1,70 @@
...
+            return expr
+
+
+@singledispatch
+def factorize_target(expr, target):
+    return 1 if expr is target else 0
+
+
+@factorize_target.register(Add)
+@factorize_target.register(EvalDerivative)
```
FabioLuporini, line 51: so, `EvalDerivative` _is_ an Add, so this registration might actually be unnecessary?

devito/operations/solve.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def eval_t(expr):
+    """
+    Evaluate all time derivatives in the expression.
+    """
+    try:
+        assert any(d.is_Time for d in expr.dims)
+        return expr.evaluate
+    except:
```
FabioLuporini, line 38: I know this is what we have in master, but I can't really see this nested try-except with no exception specification. What exceptions should be caught here? Also, always good to de-nest the exceptions when possible (I think it is here, just add `pass` in the first one and de-nest)

EDIT: oh, the except is for the assert? but that's not how `assert` should be used... 

```
if any(d.is_Time for d in expr.dims):
    return expr.evaluate
else:
    try:
        return ...
    except X:
        return expr
```

how about above ?

devito/operator/operator.py
```diff
@@ -269,6 +269,7 @@ def _lower_exprs(cls, expressions, **kwargs):
             * Shift indices for domain alignment.
         """
         # Add in implicit expressions
+
```
FabioLuporini, line 272: drop edit

devito/passes/clusters/factorization.py
```diff
@@ -76,7 +76,7 @@ def collect_const(expr):
             # -> (a + c)
             add = Add(*v)
             # -> 3.*(a + c)
-            mul = Mul(k, add, evaluate=False)
+            mul = Mul(k, add, evaluate=False) if k != 1 else add
```
FabioLuporini, line 79: (nitpicking)

I know it used to be like that, but we're dropping the ternary operators everywhere, so the convention is now to write:

```
if k == 1:
    mul = add
else:
    mul = ...
```

sorry, conventions...

devito/symbolics/inspection.py
```diff
@@ -1,10 +1,9 @@
 from collections import Counter
-
```
FabioLuporini, line 2: this should remain -- imports from stdlib separated from imports from our own deps

devito/symbolics/inspection.py
```diff
@@ -113,7 +112,10 @@ def estimate_cost(exprs, estimate=False):
                     if all(i.function.is_const for i in terminals):
                         flops += 1
                     else:
-                        flops += trascendentals_cost.get(op.__class__, 1)
+                        try:
```
FabioLuporini, line 115: no, I wouldn't change this line, but I'd rather change this one above:

```
trascendentals_cost = {sin: 100, cos: 100, exp: 100, log: 100}
```

something along the lines of:

```
trascendentals_cost = {i: 100 for i in <all objects in elementary.py>}
```

devito/finite_differences/elementary.py
```diff
@@ -0,0 +1,805 @@
+import sympy
+
+from devito.finite_differences.differentiable import (DifferentiableFunction, diffify)
+
```
FabioLuporini, line 4: ahem... I guess we need `__all__ = [...]` or we'll end up exposing DifferentiableFunction, diffify, and sympy too, to the user API 😬

devito/types/basic.py
```diff
@@ -867,7 +867,7 @@ def symbolic_shape(self):
         halo = [sympy.Add(*i, evaluate=False) for i in self._size_halo]
         padding = [sympy.Add(*i, evaluate=False) for i in self._size_padding]
         domain = [i.symbolic_size for i in self.dimensions]
-        ret = tuple(sympy.Add(i, j, k, evaluate=False)
+        ret = tuple(sympy.Add(i, j, k)
```
FabioLuporini, line 870: out of curiosity, why? it helped visually separating the left halo/padding from the right halo/padding

devito/types/equation.py
```diff
@@ -1,14 +1,13 @@
...
 import sympy
-from sympy.solvers.solveset import linear_coeffs
 
 from cached_property import cached_property
 
 from devito.finite_differences import default_rules
 from devito.logger import warning
 from devito.tools import as_tuple
-from devito.types.lazy import Evaluable
+from devito.types.lazy import Evaluable, eval_t
```
FabioLuporini, line 10: any better name for `eval_t` ?

devito/types/equation.py
```diff
@@ -236,15 +235,15 @@ def solve(eq, target, **kwargs):
         Symbolic optimizations applied while rearranging the equation. For more
         information. refer to ``sympy.solve.__doc__``.
     """
+    from devito.operations.solve import linsolve, SolveError
```
FabioLuporini, line 238: imho, time to move `solve(...)` to `operations`

devito/types/lazy.py
```diff
@@ -51,13 +49,15 @@ def evaluate(self):
         return self.func(*args, evaluate=evaluate)
 
 
-# Custom SymPy types used upon evaluation
-
-
-class EvalDerivative(sympy.Add):
-
+def eval_t(expr):
```
FabioLuporini, line 52: to drop, already available in solve.py

examples/performance/00_overview.ipynb
```diff
@@ -723,7 +722,7 @@
       "#pragma omp parallel num_threads(nthreads)\n",
       "{\n",
       "  const int tid = omp_get_thread_num();\n",
-      "  posix_memalign((void**)&pr0[tid], 64, sizeof(float[y_size + 2 + 2][z_size]));\n",
+      "  posix_memalign((void**)&pr0[tid], 64, sizeof(float[y_size + 4][z_size]));\n",
```
FabioLuporini, line 725: this is going to create loads of conflicts in my branch, if we could reinstante the evaluate=False it wouldn't be too bad...

requirements.txt
```diff
@@ -1,6 +1,6 @@
 pip>=9.0.1
 numpy>1.16
-sympy<1.6
+sympy>=1.8
```
FabioLuporini, line 3: or <=1.8 ? or == ?
do we also support 1.5.1? 1.7?

devito/finite_differences/differentiable.py
mloubout: Elementary function don't have a `__base__` not need diffify (I'm pretty sure) and Mul/Add never ends up here.

devito/symbolics/inspection.py
mloubout: I still need the pure sympy ones as well since the compiler lowers it before ending up here. (`diff2sympy`).

devito/types/basic.py
mloubout: Because you ended up with explicit `[i_size + 0 + 0 + 0 + 0]` all the way to the generated c code.

requirements.txt
mloubout: Only 1.8 there is some backward incompatibility with 1.5, haven't checked 1.7

devito/operations/solve.py
mloubout: No it's necessarry dispatch works on class match not on superclass match.

devito/finite_differences/elementary.py
mloubout: `__all__` is nothing so nothing is exposed. Need to be imported explicitly.

devito/operations/solve.py
mloubout: Well no that won't work because `expr` may not have a `dims` od `d` may not have a `is_Time`.

devito/builtins/arithmetic.py
mloubout: done

devito/finite_differences/elementary.py
mloubout: done

devito/operations/solve.py
mloubout: reworked with dispatch :D another X2 speedup in the process.

devito/types/equation.py
mloubout: done

devito/builtins/arithmetic.py
```diff
@@ -33,11 +33,11 @@ def norm(f, order=2):
     with MPIReduction(f) as mr:
         op = dv.Operator([dv.Eq(s, 0.0)] +
                          eqns +
-                         [dv.Inc(s, Abs(Pow(p, order))), dv.Eq(mr.n[0], s)],
+                         [dv.Inc(s, dv.Abs(Pow(p, order))), dv.Eq(mr.n[0], s)],
```
FabioLuporini, line 36: why can we do `dv.Abs` but not `dv.Pow`? this looks odd...

devito/finite_differences/differentiable.py
```diff
@@ -379,7 +393,7 @@ def _gather_for_diff(self):
         for f in self.args:
             if f not in self._args_diff:
                 new_args.append(f)
-            elif f is func_args:
+            elif f is func_args or isinstance(f, DifferentiableFunction):
```
FabioLuporini, line 396: nitpicking: potentially nicer if you `.extend` the `func_args` with the DifferentiableFunctions upfront?

devito/finite_differences/elementary.py
```diff
@@ -0,0 +1,661 @@
+import sympy
+
+from devito.finite_differences.differentiable import (DifferentiableFunction, diffify)
```
FabioLuporini, line 3: nitpicking: no need for parentheses

devito/finite_differences/elementary.py
FabioLuporini: wait, what? IIRC, this is only true if a variable name begins with an `_`  -- then it's hidden away. Otherwise it's implicitly imported, unless `__all__` is present. [Here](https://docs.python.org/3/tutorial/modules.html#importing-from-a-package) seems to confirm this

FabioLuporini: assuming I'm correct (if not apologies in advance), note that albeit tedious, it's not a shame to have a long list of explicit exports, see for example [here](https://github.com/sympy/sympy/blob/master/sympy/__init__.py#L51)

devito/operations/solve.py
```diff
@@ -0,0 +1,120 @@
...
+    # We need to rebuild the vector/tensor as sympy.solve outputs a tuple of solutions
+    if len(sols) > 1:
+        return target.new_from_mat(sols)
+    else:
+        return sols[0]
+
+
+def linsolve(expr, target, **kwargs):
+    """
+    Linear solve for the targe in a single equation.
```
FabioLuporini, line 58: typo, "target"

devito/operations/solve.py
```diff
@@ -0,0 +1,120 @@
...
+    eq : expr-like
+        The equation to be rearranged.
+    target : symbol
+        The symbol w.r.t. which the equation is rearranged. May be a `Function`
+        or any other symbolic object.
+    **kwargs
+        Symbolic optimizations applied while rearranging the equation. For more
+        information. refer to ``sympy.solve.__doc__``.
+    """
+
```
FabioLuporini, line 35: nitpicking: no blankline at start of function

devito/operations/solve.py
```diff
@@ -0,0 +1,120 @@
...
+    if hasattr(eq, 'rhs'):
```
FabioLuporini, line 36: the pythonic way is a `try-except(AttributeError)`. We should always avoid `hasattr` and `getattr` for as much as we can

devito/operations/solve.py
```diff
@@ -0,0 +1,120 @@
...
+        eq = eq.lhs - eq.rhs if eq.rhs != 0 else eq.lhs
+    sols = []
```
FabioLuporini, line 38: here we wwant a blank line above

devito/operations/solve.py
```diff
@@ -0,0 +1,120 @@
...
+        # Try first linear solver
+        try:
+            sols.append(linsolve(eval_t(e), t))
+        except SolveError:
+            warning("Equation is not affine w.r.t the target, falling back to standard"
+                    "sympy.solve that may be slow")
+            kwargs['rational'] = False  # Avoid float indices
+            kwargs['simplify'] = False  # Do not attempt premature optimisation
+            sols.append(sympy.solve(e.evaluate, t, **kwargs)[0])
+    # We need to rebuild the vector/tensor as sympy.solve outputs a tuple of solutions
```
FabioLuporini, line 49: also here we want a blank line above. The concept is that we shold separate blocks of code which are meant to achieve different things

devito/operations/solve.py
```diff
@@ -0,0 +1,120 @@
...
+    if len(sols) > 1:
+        return target.new_from_mat(sols)
+    else:
+        return sols[0]
```
FabioLuporini, line 53: I think, though I may be wrong, that this would fail "weirdly" if one accidentally called `solve([], ...` ie with no equations. You probably want to have an as_tuple at the very beginning of this function, and sanity-check that `len(...) > 0` , so surely `return sols[0]` won't trigger an `IndexError`

devito/symbolics/inspection.py
```diff
@@ -113,7 +113,10 @@ def estimate_cost(exprs, estimate=False):
...
```
FabioLuporini, line 116: since the cost is 100 for all of them, I actually suggest to finally embrace it -- let's turn `transcental_cost` into a number, ir `trascendentals_cost = 100` and here let's just add `flops += trascendentals_cost`. This will also make it work if a function other than the four present now (sin/cos/log/exp) were used

devito/types/basic.py
FabioLuporini: is that the same in master too?

examples/userapi/01_dsl.ipynb
```diff
@@ -679,14 +682,14 @@
       "} ;\n",
       "\n",
       "\n",
-      "int Kernel(const float dt, const float h_x, const float h_y, struct dataobj *restrict u_vec, const int time_M, const int time_m, struct profiler * timers, const int x_M, const int x_m, const int y_M, const int y_m, const int nthreads)\n",
+      "int Kernel(const float dt, const float h_x, const float h_y, struct dataobj *restrict u_vec, const int time_M, const int time_m, const int x_M, const int x_m, const int y_M, const int y_m, const int nthreads, struct profiler * timers)\n",
```
FabioLuporini, line 685: I wonder why this change

FabioLuporini: probably stemming from an old PR

devito/finite_differences/elementary.py
mloubout: From your link
```
If __all__ is not defined, the statement from sound.effects import * does not import all submodules from the package sound.effects into the current namespace; it only ensures that the package sound.effects has been imported
``

devito/types/basic.py
mloubout: It's not no, it breaks some tests with the `evaluate=False`

devito/builtins/arithmetic.py
mloubout: because `Pow` is explicitly not imported `not in its own _all__` while `Abs` is not imported but its submodule is so `.Abs` is defined.

FabioLuporini: btw, probably now `p**order` would work too

PR #1551, opened 2021-01-11
WIP: FD: add missing shift option

Leitevmd:
Small patch to fix missing shift option implementation in some functions (div, grad, etc).
It may need more adjustments to cover other implementations.
This is related to image mismatching issue discussed in #1406 and changes from #1511.


Leitevmd:
Added a test with `div(grad(f))` which calculates a shifted div of a VectorFunction. This causes an error without the patch.

Leitevmd:
Maybe we should add shift also for div of tensor functions [here](https://github.com/devitocodes/devito/blob/55967596698e01cee50c9cc11a801ebcf2e3484c/devito/types/tensor.py#L223).

rhodrin:
Hi @Leitevmd, thanks for the PR.

Just looking quickly and my memory may be out of date/playing tricks again - but can't `shift` be a `tuple`, e.g `(0.5, 0, 0)` or `(0.5, 0.5, 0.5)` etc?

Leitevmd:
> Just looking quickly and my memory may be out of date/playing tricks again - but can't `shift` be a `tuple`, e.g `(0.5, 0, 0)` or `(0.5, 0.5, 0.5)` etc?

The way @mloubout did in #1511 for shifting div and grad functions ([here](https://github.com/devitocodes/devito/blob/55967596698e01cee50c9cc11a801ebcf2e3484c/devito/finite_differences/differentiable.py#L245) for example) uses the same shift value for all dimensions of the grid.

FabioLuporini:
Merged, thanks

PR #1556, opened 2021-01-15
Data streaming support with OpenMP offloading

FabioLuporini:
* Only tested (no surprise) with clang and aomp
* the adjoint operators do not work yet because of clang/aomp bugs, which have already been flagged
* Data streaming is now performed in pure C, not C++ (now using pthreads instead of std::threads -- this was necessary because of compatibility issues with clang/aomp)
* Further improvements to the compiler passes implementing data streaming

### DIFF
--- a/devito/ir/iet/efunc.py
+++ b/devito/ir/iet/efunc.py
@@ -6,9 +6,9 @@ import cgen as c
 
 from devito.ir.iet.nodes import (BlankLine, Call, Callable, Conditional, Dereference,
                                  DummyExpr, Iteration, List, PointerCast, Return, While)
-from devito.ir.iet.utils import derive_parameters
+from devito.ir.iet.utils import derive_parameters, diff_parameters
 from devito.symbolics import CondEq, CondNe, FieldFromComposite, FieldFromPointer, Macro
-from devito.tools import as_tuple, split
+from devito.tools import as_tuple
 from devito.types import PThreadArray, SharedData, Symbol, VoidPointer
 
 __all__ = ['ElementalFunction', 'ElementalCall', 'make_efunc',
@@ -160,9 +160,7 @@ def _make_thread_init(threads, tfunc, isdata, sdata, sregistry):
 def _make_thread_func(name, iet, root, threads, sregistry):
     # Create the SharedData, that is the data structure that will be used by the
     # main thread to pass information dows to the child thread(s)
-    required = derive_parameters(iet)
-    known = root.parameters + tuple(i for i in required if i.is_Array and i._mem_shared)
-    parameters, dynamic_parameters = split(required, lambda i: i in known)
+    required, parameters, dynamic_parameters = diff_parameters(iet, root)
     sdata = SharedData(name=sregistry.make_name(prefix='sdata'), npthreads=threads.size,
                        fields=required, dynamic_fields=dynamic_parameters)
 
--- a/devito/ir/iet/efunc.py
+++ b/devito/ir/iet/efunc.py
@@ -1,10 +1,14 @@
 from cached_property import cached_property
 
-from devito.ir.iet.nodes import Call, Callable
+from devito.ir.iet.nodes import (BlankLine, Call, Callable, Conditional, DummyExpr,
+                                 List, While)
 from devito.ir.iet.utils import derive_parameters
-from devito.tools import as_tuple
+from devito.symbolics import CondEq, CondNe, FieldFromPointer
+from devito.tools import as_tuple, split
+from devito.types import SharedData
 
-__all__ = ['ElementalFunction', 'ElementalCall', 'make_efunc']
+__all__ = ['ElementalFunction', 'ElementalCall', 'make_efunc',
+           'ThreadFunction', 'make_tfunc']
 
 
 class ElementalFunction(Callable):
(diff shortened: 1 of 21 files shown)


devito/ir/iet/efunc.py
```diff
@@ -83,3 +93,196 @@ def make_efunc(name, iet, dynamic_parameters=None, retval='void', prefix='static
     """
     return ElementalFunction(name, iet, retval, derive_parameters(iet), prefix,
                              dynamic_parameters)
+
```
FabioLuporini, line 96: Note for reviewers: most of this code here was just lifted from `passes/iet/orchestration.py`

Semantically, the changes are minimal and not particularly significant. The code overall looks much nicer now IMHO

devito/passes/iet/orchestration.py
```diff
@@ -46,122 +45,6 @@ def _P(self):
         """Shortcut for self._Parallelizer."""
         return self._Parallelizer
 
-    def __make_tfunc(self, name, iet, root, threads):
```
FabioLuporini, line 49: this is what I wrote above -- much of this has been moved to `ir/iet/efunc.py`

devito/ir/iet/efunc.py
ggorman: By 'lifted' do you mean transposed or copied? If it's the latter, I assume you considered and rejected creating an abstract class.

FabioLuporini: by lifted I mean that I moved it (cut-paste), plus there are a few changes so that we generate C rather than C++, as we agreed. 

I don't know why you'd want an abstract class here, but the `Callable` hierarchy is already pretty deep

devito/passes/iet/orchestration.py
```diff
@@ -177,45 +60,43 @@ def _make_waitlock(self, iet, sync_ops, *args):
     def _make_withlock(self, iet, sync_ops, pieces, root):
         locks = sorted({s.lock for s in sync_ops}, key=lambda i: i.name)
 
-        threads = self.__make_threads(value=min(i.size for i in locks))
+        npthreads = min(i.size for i in locks)
```
ggorman, line 63: Can you walk me through this piece if code as it's not clear to my why this gives the number of npthreads. And why is locks sorted?

devito/passes/iet/orchestration.py
```diff
@@ -177,45 +60,43 @@ def _make_waitlock(self, iet, sync_ops, *args):
...
+
+            preactions.append(List(body=[BlankLine, update, DummyExpr(s.handle, 1)]))
             postactions.append(DummyExpr(s.handle, 2))
         preactions.append(BlankLine)
         postactions.insert(0, BlankLine)
 
-        # Turn `iet` into an ElementalFunction so that it can be
-        # executed asynchronously by `threadhost`
+        # Turn `iet` into a ThreadFunction so that it can be executed
+        # asynchronously by a child host thread
```
ggorman, line 80: What's the difference between a 'child host thread' and a pthread?

Remark: Given the number of different uses for thread (pthreads, openmp, openacc, etc) perhaps use the associated  qualifier (ie pthread, OpenMP threads, etc)/

devito/tools/utils.py
```diff
@@ -221,17 +221,21 @@ def ctypes_to_cstr(ctype, toarray=None):
...
+        else:
+            name = ctype.__name__[2:]
         # A primitive datatype
         # FIXME: Is there a better way of extracting the C typename ?
         # Here, we're following the ctypes convention that each basic type has
         # either the format c_X_p or c_X, where X is the C typename, for instance
         # `int` or `float`.
-        if ctype.__name__.endswith('_p'):
-            return '%s *' % ctype.__name__[2:-2]
+        if name.endswith('_p'):
```
ggorman, line 233: The only issue with this piece of code is that it's behaviour is not defined if it gets an unexpected (or non-conventional) datatype). Is there any sanity check that can be performed?

tests/test_gpu_common.py
```diff
@@ -269,6 +270,7 @@ def test_tasking_lock_placement(self):
         assert str(sections[1].body[0].body[0].body[0].body[0]) ==\
             'while(lock0[t1] == 0);'
 
+    @skipif('device-openmp')  # TODO: Still unsupported with OpenMP, but soon will be
```
ggorman, line 273: Perhaps we need to change the policy here for TODO's. At the time of writing the developer likely has a clear understanding of what needs to be done; but clarity can quickly evaporate as the developer undertakes other tasks. 

One possibility is to link every TODO with a corresponsing issue number. There the developer can add their notes for their future self or whoever picks up the TODO later.

ggorman (approved):
GTG - just added minor questions and comments.

devito/passes/iet/orchestration.py
FabioLuporini: Ah, `locks` is sorted for deterministic code gen. I shall add a comment

About the `npthreads` -- in practice, this routine constructs the piece of tree responsible for the copy-back from the device to the host. For a copy-back to be feasile all of the locks protecting the copied fileds must have been acquired. The `min` is  used to pick the maximum degree of parallelism possible. For example, consider:

```
while (lock0[i] == 0 && lock1[j] == 0);  // busy wait
```

where `i` and `j` are modulo dimensions, `lock0` protects 3 entries (say `u[0], u[1], u[2]`), and `lock1` protects 2 entries (say `v[0], v[1]`). There can't be more than two `v`'s in flight at the same time, so it's pointless to assume that more than 2 threads will ever be needed. I will also add a comment

devito/passes/iet/orchestration.py
FabioLuporini: > What's the difference between a 'child host thread' and a pthread?

no difference. I will clarify

devito/tools/utils.py
FabioLuporini: this piece of code is quite ancient... it's been here for ages. It also needs to be quick. I remember I did try to improve it, but I couldn't come up with anything convincing

tests/test_gpu_common.py
FabioLuporini: OK. But the "future" you're talking about here is more like imminent future (hence the "soon"). Basically all these TODOs point to the AMD issues... will see how to improve it

FabioLuporini (commented):
@ggorman I've now addressed some of your comments, rebased on top of latest master, and pushed a few other minor edits. Since you've already approved it, I'll merge once green

FabioLuporini:
PR approved, CI-gpu manually run and green (https://github.com/devitocodes/devito/actions/runs/509421224 ), hence merging

PR #1561, opened 2021-01-19
Shift argument as a tuple

### LINKED ISSUE #1554: `div`, `grad` etc. should accept `shift` as a `tuple`
The tensor operators `div`, `grad` and `curl` currently only accept `shift` as a single `float`. Since the staggering can be different in each dimension however they should also be able to accept a `tuple` of `float` of length `dims`.

See, e.g., [here](https://github.com/devitocodes/devito/blob/1802de464921b2ca225ec5e18e1f7935a9fbddab/devito/types/tensor.py#L305).

Leitevmd:
Shift argument of div/grad accepted as a float or a tuple (Fix #1554)
Add shift option to divergence of TensorFunction and gradient of VectorFunction.

### DIFF
--- a/devito/finite_differences/differentiable.py
+++ b/devito/finite_differences/differentiable.py
@@ -7,6 +7,7 @@ from sympy.core.decorators import call_highest_priority
 from sympy.core.evalf import evalf_table
 
 from cached_property import cached_property
+from devito.finite_differences.tools import make_shift_x0
 from devito.logger import warning
 from devito.tools import filter_ordered, flatten
 from devito.types.lazy import Evaluable, EvalDerivative
@@ -242,17 +243,16 @@ class Differentiable(sympy.Expr, Evaluable):
 
     def div(self, shift=None):
         space_dims = [d for d in self.dimensions if d.is_Space]
-        return Add(*[getattr(self, 'd%s' % d.name)(x0=None if shift is None else
-                                                   d + shift[i] * d.spacing if type(shift)
-                                                   is tuple else d + shift * d.spacing)
+        shift_x0 = make_shift_x0(shift, (len(space_dims),))
+        return Add(*[getattr(self, 'd%s' % d.name)(x0=shift_x0(shift, d, None, i))
                      for i, d in enumerate(space_dims)])
 
     def grad(self, shift=None):
         from devito.types.tensor import VectorFunction, VectorTimeFunction
-        comps = [getattr(self, 'd%s' % d.name)(x0=None if shift is None else
-                                               d + shift[i] * d.spacing if type(shift)
-                                               is tuple else d + shift * d.spacing)
-                 for i, d in enumerate(self.dimensions) if d.is_Space]
+        space_dims = [d for d in self.dimensions if d.is_Space]
+        shift_x0 = make_shift_x0(shift, (len(space_dims),))
+        comps = [getattr(self, 'd%s' % d.name)(x0=shift_x0(shift, d, None, i))
+                 for i, d in enumerate(space_dims)]
         vec_func = VectorTimeFunction if self.is_TimeDependent else VectorFunction
         return vec_func(name='grad_%s' % self.name, time_order=self.time_order,
                         space_order=self.space_order, components=comps, grid=self.grid)
--- a/devito/finite_differences/differentiable.py
+++ b/devito/finite_differences/differentiable.py
@@ -243,14 +243,16 @@ class Differentiable(sympy.Expr, Evaluable):
     def div(self, shift=None):
         space_dims = [d for d in self.dimensions if d.is_Space]
         return Add(*[getattr(self, 'd%s' % d.name)(x0=None if shift is None else
-                                                   d + shift * d.spacing)
-                     for d in space_dims])
+                                                   d + shift[i] * d.spacing if type(shift)
+                                                   is tuple else d + shift * d.spacing)
+                     for i, d in enumerate(space_dims)])
 
     def grad(self, shift=None):
         from devito.types.tensor import VectorFunction, VectorTimeFunction
         comps = [getattr(self, 'd%s' % d.name)(x0=None if shift is None else
-                                               d + shift * d.spacing)
-                 for d in self.dimensions if d.is_Space]
+                                               d + shift[i] * d.spacing if type(shift)
+                                               is tuple else d + shift * d.spacing)
+                 for i, d in enumerate(self.dimensions) if d.is_Space]
         vec_func = VectorTimeFunction if self.is_TimeDependent else VectorFunction
         return vec_func(name='grad_%s' % self.name, time_order=self.time_order,
                         space_order=self.space_order, components=comps, grid=self.grid)
--- a/devito/finite_differences/tools.py
+++ b/devito/finite_differences/tools.py
@@ -5,7 +5,6 @@ import numpy as np
 from sympy import S, finite_diff_weights, cacheit, sympify
 
 from devito.tools import Tag, as_tuple
-from devito.finite_differences import Differentiable
 
 
 class Transpose(Tag):
@@ -47,6 +46,7 @@ centered = Side('centered', 0)
 def check_input(func):
     @wraps(func)
     def wrapper(expr, *args, **kwargs):
+        from devito.finite_differences import Differentiable
         if expr.is_Number:
             return S.Zero
         elif not isinstance(expr, Differentiable):
@@ -261,3 +261,24 @@ def generate_indices_staggered(func, dim, order, side=None, x0=None):
             ind = [start, start - diff]
 
     return start, tuple(ind)
+
+
+def make_shift_x0(shift, ndim):
+    """
+    Returns a callable that calculates a shifted origin for each derivative
+    of an operation derivatives scheme (given by ndim) given a shift object
+    which can be a None, a float or a tuple with shape equal to ndim
+    """
+    if shift is None:
+        return lambda s, d, i, j: None
+    elif isinstance(shift, float):
+        return lambda s, d, i, j: d + s * d.spacing
+    elif type(shift) is tuple and np.shape(shift) == ndim:
+        if len(ndim) == 1:
+            return lambda s, d, i, j: d + s[j] * d.spacing
+        elif len(ndim) == 2:
+            return lambda s, d, i, j: d + s[i][j] * d.spacing
+        else:
+            raise ValueError("ndim length must be equal to 1 or 2")
+    raise ValueError("shift parameter must be one of the following options: "
+                     "None, float or tuple with shape equal to %s" % (ndim,))
--- a/devito/finite_differences/tools.py
+++ b/devito/finite_differences/tools.py
@@ -46,14 +46,11 @@ centered = Side('centered', 0)
 def check_input(func):
     @wraps(func)
     def wrapper(expr, *args, **kwargs):
-        from devito.finite_differences import Differentiable
-        if expr.is_Number:
-            return S.Zero
-        elif not isinstance(expr, Differentiable):
-            raise ValueError("`%s` must be of type Differentiable (found `%s`)"
+        try:
+            return S.Zero if expr.is_Number else func(expr, *args, **kwargs)
+        except AttributeError:
+            raise ValueError("'%s' must be of type Differentiable, not %s"
                              % (expr, type(expr)))
-        else:
-            return func(expr, *args, **kwargs)
     return wrapper
 
 
--- a/devito/types/tensor.py
+++ b/devito/types/tensor.py
@@ -346,7 +346,10 @@ class VectorFunction(TensorFunction):
         Gradient of the VectorFunction, creates the gradient TensorFunction.
         """
         func = tens_func(self, self)
-        comps = [[getattr(f, 'd%s' % d.name) for d in self.space_dimensions]
+        comps = [[getattr(f, 'd%s' % d.name)(x0=None if shift is None else
+                                             d + shift[i] * d.spacing if type(shift)
+                                             is tuple else d + shift * d.spacing)
+                  for i, d in enumerate(self.space_dimensions)]
                  for f in self]
         return func._new(comps)
 
--- a/devito/types/tensor.py
+++ b/devito/types/tensor.py
@@ -7,6 +7,7 @@ from sympy.core.decorators import call_highest_priority
 from sympy.core.sympify import converter as sympify_converter
 
 from devito.finite_differences import Differentiable
+from devito.finite_differences.tools import make_shift_x0
 from devito.types.basic import AbstractTensor
 from devito.types.dense import Function, TimeFunction
 from devito.types.utils import NODE
@@ -219,11 +220,12 @@ class TensorFunction(AbstractTensor):
         """
         comps = []
         func = vec_func(self, self)
-        for j, d in enumerate(self.space_dimensions):
+        ndim = len(self.space_dimensions)
+        shift_x0 = make_shift_x0(shift, (ndim, ndim))
+        for i in range(len(self.space_dimensions)):
             comps.append(sum([getattr(self[j, i], 'd%s' % d.name)
-                              (x0=None if shift is None else d + shift[i] * d.spacing
-                               if type(shift) is tuple else d + shift * d.spacing)
-                              for i, d in enumerate(self.space_dimensions)]))
+                              (x0=shift_x0(shift, d, i, j))
+                              for j, d in enumerate(self.space_dimensions)]))
         return func._new(comps)
 
     @property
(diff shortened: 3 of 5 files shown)


FabioLuporini:
Could you add some tests?

Leitevmd:
> Could you add some tests?

Yes sure! I'm already working on it!

devito/finite_differences/differentiable.py
```diff
@@ -243,14 +243,16 @@ def laplace(self):
     def div(self, shift=None):
         space_dims = [d for d in self.dimensions if d.is_Space]
         return Add(*[getattr(self, 'd%s' % d.name)(x0=None if shift is None else
-                                                   d + shift * d.spacing)
-                     for d in space_dims])
+                                                   d + shift[i] * d.spacing if type(shift)
+                                                   is tuple else d + shift * d.spacing)
+                     for i, d in enumerate(space_dims)])
```
mloubout, line 248: Getting quite igly there too many if/if/....

devito/types/tensor.py
```diff
@@ -221,6 +221,8 @@ def div(self, shift=None):
         func = vec_func(self, self)
         for j, d in enumerate(self.space_dimensions):
             comps.append(sum([getattr(self[j, i], 'd%s' % d.name)
+                              (x0=None if shift is None else d + shift[i] * d.spacing
```
mloubout, line 224: If we wanna go that way then it needs to be a tuple of tuple for a tensor, not just a tuple.

devito/types/tensor.py
```diff
@@ -342,7 +346,10 @@ def grad(self, shift=None):
         Gradient of the VectorFunction, creates the gradient TensorFunction.
         """
         func = tens_func(self, self)
-        comps = [[getattr(f, 'd%s' % d.name) for d in self.space_dimensions]
+        comps = [[getattr(f, 'd%s' % d.name)(x0=None if shift is None else
+                                             d + shift[i] * d.spacing if type(shift)
+                                             is tuple else d + shift * d.spacing)
+                  for i, d in enumerate(self.space_dimensions)]
```
mloubout, line 352: same if/if/...

May need more checks too since may be a tuple of wrong length, .....

mloubout (commented):
As fabio said, it needs tests, left some comments

devito/finite_differences/differentiable.py
FabioLuporini: I agree with Mathias here. I think we cannot afford list comprehension here anymore.

FabioLuporini (changes requested):

devito/types/tensor.py
mloubout: That's only valid for a vector with one component per dimension. A tensorfunction is a `ndimxndim` matrix with one component per dimension pair. It's the same thing as staggerring setup, it is a tuple for Vector and a tuple of tuple for Tensor

Leitevmd: Nice!! I'm gonna look into that..

Leitevmd: Please review again. I think this is a good solution. We just need to visit shift values once. One thing to note here is that you can pass (ndim,) or (ndim,ndim) shaped shift (beyond a single float) when dealing with tensors (grad of vectors and div of tensors). A (a,b,c) tuple would be considered as ((a,b,c), (a,b,c), (a,b,c)). Let me know if this would be undesirable.

mloubout: >A (a,b,c) tuple would be considered as ((a,b,c), (a,b,c), (a,b,c)). Let me know if this is undesirable.

That's not a good default IMO. In general, these tenors are symmetric/antisymmetric at least in the staggering so the shift should be too. There is not "easy" way to go from a tuple to a tuple of tuple for tensors and only tuple of tuples should be accepted I think.

mloubout: Would have a single function yes. May be easier and more compact with some  try/catch to merge all cases.

Leitevmd:
@mloubout, Should we add shift option for [curl](https://github.com/devitocodes/devito/blob/432235ff0df23f523f8d6f7687d2a76e5599baa6/devito/types/tensor.py#L324) function as mentioned in #1554 ?
Would be something like that?
```python
def curl(self, shift=None):
    """
    Gradient of the (3D) VectorFunction, creates the curl VectorFunction.
    """

    if len(self.space_dimensions) != 3:
        raise AttributeError("Curl only supported for 3D VectorFunction")
    shift_x0_ = shift_x0(shift, (3, 3))
    # The curl of a VectorFunction is a VectorFunction
    derivs = ['d%s' % d.name for d in self.space_dimensions]
    comp1 = getattr(self[2], derivs[1])(x0=shift_x0_(shift, derivs[1], 2, 1))\
        - getattr(self[1], derivs[2])(x0=shift_x0_(shift, derivs[2], 1, 2))
    comp2 = getattr(self[0], derivs[2])(x0=shift_x0_(shift, derivs[2], 0, 2))\
        - getattr(self[2], derivs[0])(x0=shift_x0_(shift, derivs[0], 2, 0))
    comp3 = getattr(self[1], derivs[0])(x0=shift_x0_(shift, derivs[0], 1, 0))\
        - getattr(self[0], derivs[1])(x0=shift_x0_(shift, derivs[1], 0, 1))

    vec_func = VectorTimeFunction if self.is_TimeDependent else VectorFunction
    return vec_func._new(3, 1, [comp1, comp2, comp3])
```

Leitevmd:
@FabioLuporini , I completely forgot the commits tags on this PR. Thanks for reminding on Slack =). Should I rebase them?

FabioLuporini:
> Should I rebase them?

@Leitevmd yeah, thanks a lot!

devito/finite_differences/differentiable.py
```diff
@@ -494,6 +496,20 @@ def _diff2sympy(obj):
     return _diff2sympy(expr)[0]
 
 
+def shift_x0(shift, shape):
```
FabioLuporini, line 499: maybe this should be moved to `devito/finite_differences/tools` ? What do you think @Leitevmd  @mloubout

devito/types/tensor.py
```diff
@@ -403,6 +409,20 @@ def tens_func(func1, func2):
     return TensorTimeFunction if f1 or f2 else TensorFunction
 
 
+def shift_x0(shift, shape):
```
FabioLuporini, line 412: this is almost a copy-paste. Can you generalize the two `shift_x0` and turn them into a single function? and move it to tools? you could also probably use a better name

FabioLuporini (changes requested):

devito/types/tensor.py
Leitevmd: We have `from devito.finite_differences import Differentiable` in tools.py, so I couldn't call `from devito.finite_differences.tools import shift_x0` inside defferentiable.py due to circular import. Any idea to solve that?

Leitevmd: Nevermind.. solved!

Leitevmd: @FabioLuporini what do you think? Name chosen based on [this one](https://github.com/devitocodes/devito/blob/3fb0c067d2b8019302af28078537262d6f908518/devito/passes/clusters/utils.py#L91). Let me know about anymore suggestions.

```python
def make_get_shifted_oringin(shift, shape):
    """
    Returns a callable that calculates a shifted origin for each derivative
    of a given operation derivatives scheme (shape) and a shift object which
    can be a None, a single value or a tuple with the given shape.
    """
    if shift is None:
        return lambda s, d, i, j: None
    if type(shift) is float:
        return lambda s, d, i, j: d + s * d.spacing
    if type(shift) is tuple and np.shape(shift) == shape:
        if len(shape) == 1:
            return lambda s, d, i, j: d + s[j] * d.spacing
        if len(shape) == 2:
            return lambda s, d, i, j: d + s[i][j] * d.spacing
    raise ValueError("shift parameter must be one of the following options: "
                     "None, float or tuple with shape %s." % (shape,))
```

devito/finite_differences/tools.py
```diff
@@ -47,6 +46,7 @@ def adjoint(self, matvec):
 def check_input(func):
     @wraps(func)
     def wrapper(expr, *args, **kwargs):
+        from devito.finite_differences import Differentiable
```
FabioLuporini, line 49: this is too hacky and therefore not acceptable, we need to fix the cyclic import issue differently

devito/finite_differences/tools.py
```diff
@@ -261,3 +261,22 @@ def generate_indices_staggered(func, dim, order, side=None, x0=None):
             ind = [start, start - diff]
 
     return start, tuple(ind)
+
+
+def make_get_shifted_oringin(shift, shape):
```
FabioLuporini, line 266: there's a typo -- it'd be "origin" not "origin"

since you assign the function to a variable named "shift_x0", I'd call this function "make_shift_x0"

devito/finite_differences/tools.py
```diff
@@ -261,3 +261,22 @@ def generate_indices_staggered(func, dim, order, side=None, x0=None):
...
+    """
+    Returns a callable that calculates a shifted origin for each derivative
+    of a given operation derivatives scheme (shape) and a shift object which
+    can be a None, a single value or a tuple with the given shape.
+    """
+    if shift is None:
+        return lambda s, d, i, j: None
+    if type(shift) is float:
```
FabioLuporini, line 274: `if isinstance(shift, float)` 

this way it will work even if eg shift is a numpy.float32 . Besides being more pythonic

FabioLuporini, line 274: `elif` probably more elegant because it makes it clear that this branch logically belongs to the same cascade of if-then-else

devito/finite_differences/tools.py
```diff
@@ -261,3 +261,22 @@ def generate_indices_staggered(func, dim, order, side=None, x0=None):
...
+        return lambda s, d, i, j: d + s * d.spacing
+    if type(shift) is tuple and np.shape(shift) == shape:
```
FabioLuporini, line 276: same as before. Also `elif`

you may also use `isinstance(shift, (tuple, list))` or maybe even `isinstance(shift, Iterable)` to support more iterable types

devito/finite_differences/tools.py
```diff
@@ -261,3 +261,22 @@ def generate_indices_staggered(func, dim, order, side=None, x0=None):
...
+        if len(shape) == 1:
+            return lambda s, d, i, j: d + s[j] * d.spacing
+        if len(shape) == 2:
```
FabioLuporini, line 279: elif

devito/finite_differences/tools.py
```diff
@@ -261,3 +261,22 @@ def generate_indices_staggered(func, dim, order, side=None, x0=None):
...
+            return lambda s, d, i, j: d + s[i][j] * d.spacing
```
FabioLuporini, line 280: what happens if it's neither 1 nor 2 ?

are we missing tests?

devito/finite_differences/tools.py
```diff
@@ -261,3 +261,22 @@ def generate_indices_staggered(func, dim, order, side=None, x0=None):
             ind = [start, start - diff]
 
     return start, tuple(ind)
+
+
+def make_get_shifted_oringin(shift, shape):
+    """
+    Returns a callable that calculates a shifted origin for each derivative
+    of a given operation derivatives scheme (shape) and a shift object which
```
FabioLuporini, line 269: I don't think what you call shape here is the shape in the devito sense. 

This is what we call `shape` in devito: https://github.com/devitocodes/devito/blob/master/devito/types/tensor.py#L42

Probably what you're after here is the dimension of the space domain, ie `(3,)` if `f(x, y, z)` or `(1, 1)` in `[f(x), g(x)]` (here a VectorFunction). Am I correct? (@mloubout )

FabioLuporini (changes requested):

devito/finite_differences/tools.py
```diff
@@ -47,6 +46,7 @@ def adjoint(self, matvec):
 def check_input(func):
     @wraps(func)
     def wrapper(expr, *args, **kwargs):
+        from devito.finite_differences import Differentiable
         if expr.is_Number:
             return S.Zero
         elif not isinstance(expr, Differentiable):
```
FabioLuporini, line 52: @Leitevmd @mloubout 

To avoid the circular import we should stop importing Differentiable into tools.

To avoid using Differentiable in tools, we could turn this `if` into a `try-except`, thus making it more pythonic, and potentially much much faster, since this function is likely to be called loads of times, and `isinstance` is relatively slow, enough to be a PITA if called too often

Mathias, how about we just do:

```
try:
    return func(expr, *args, **kwargs)
except ValueError:  # Not sure if a better exception should be picked; run some tests and find out?
    raise ValueError("`%s` must be of type Differentiable, not `%s`" (expr, type(expr))
```

?

devito/finite_differences/tools.py
Leitevmd: It will raise `ValueError` down there, but I do think we could have an extra `else` here and raise a value error with the proper justification.

devito/finite_differences/tools.py
Leitevmd: `(3, 3)` for `[f(x), g(x)]` if they have `(x, y, z)` dimensions. Yes, you're right. "shape" here does not correspond to Devito's nomenclature. Should we call it `ndim`?

FabioLuporini: yeah, I think so

devito/finite_differences/tools.py
FabioLuporini: ah, yes, I think that's fine

devito/finite_differences/tools.py
Leitevmd: Tested! When no `Differentiable` is passed, it later raises` AttributeError`. So I'm gonna use it instead of `ValueError` so that we can see this step, right?

devito/finite_differences/tools.py
```diff
@@ -261,3 +261,24 @@ def generate_indices_staggered(func, dim, order, side=None, x0=None):
             ind = [start, start - diff]
 
     return start, tuple(ind)
+
+
+def make_shift_x0(shift, ndim):
```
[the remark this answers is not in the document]
Leitevmd: Just to make sure I understand your concerning, by "a lot" you mean by every `div` or `grad` call, right? I ran an example repeating 10,000 times the same `div(tensor)` with the same parameters, with and without cache. Although `make_shif_x0` calls goes from 10,000 to 1 (with memoized version) execution time remains almost the same. Which means this function spend almost the same time as cache checking (memoized).
```
$ cat cProfile.out | grep -E "\(div\)|\(make_shift_x0\)|seconds|ncalls"
         73498783 function calls (70283322 primitive calls) in 42.183 seconds
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    10000    0.014    0.000   38.734    0.004 operators.py:1(div)
    10000    0.155    0.000   38.720    0.004 tensor.py:217(div)
    10000    0.022    0.000    0.142    0.000 tools.py:264(make_shift_x0)

$ cat cProfile.out2 | grep -E "\(div\)|\(make_shift_x0\)|seconds|ncalls"
         73391815 function calls (70176243 primitive calls) in 41.873 seconds
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    10000    0.015    0.000   38.304    0.004 operators.py:1(div)
    10000    0.154    0.000   38.290    0.004 tensor.py:217(div)
        1    0.000    0.000    0.000    0.000 tools.py:263(make_shift_x0)

```

devito/types/tensor.py
```diff
@@ -342,8 +346,11 @@ def grad(self, shift=None):
         Gradient of the VectorFunction, creates the gradient TensorFunction.
         """
         func = tens_func(self, self)
```
georgebisbas, line 348: self, self ? same arg?

Leitevmd: You're right. The same happens on line 247. Also with `vec_func` on lines 222, 237, 322, 362. The only one which uses different argument is at 358. We could simplify this with:
```python 
def tens_func(*funcs):
    return (TensorTimeFunction if any([getattr(f, 'is_TimeDependent', False)
            for f in funcs]) else TensorFunction)
```
and the same with `vec_func`

devito/finite_differences/tools.py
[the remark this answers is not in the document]
FabioLuporini: good question and good answer, I think we can resolve this comment

devito/types/tensor.py
```diff
@@ -391,16 +398,14 @@ class VectorTimeFunction(VectorFunction, TensorTimeFunction):
                  (False, True): TensorFunction, (False, False): VectorFunction}
 
 
-def vec_func(func1, func2):
-    f1 = getattr(func1, 'is_TimeDependent', False)
-    f2 = getattr(func2, 'is_TimeDependent', False)
-    return VectorTimeFunction if f1 or f2 else VectorFunction
+def vec_func(*funcs):
+    return (VectorTimeFunction if any([getattr(f, 'is_TimeDependent', False)
```
FabioLuporini, line 402: nitpicking: if you have to split the line like that, you may as well make a multiline if-else 

```
if cond
   do A
else
   do B
```

devito/types/tensor.py
```diff
@@ -391,16 +398,14 @@ class VectorTimeFunction(VectorFunction, TensorTimeFunction):
...
+            for f in funcs]) else VectorFunction)
 
 
-def tens_func(func1, func2):
-    f1 = getattr(func1, 'is_TimeDependent', False)
-    f2 = getattr(func2, 'is_TimeDependent', False)
-    return TensorTimeFunction if f1 or f2 else TensorFunction
+def tens_func(*funcs):
+    return (TensorTimeFunction if any([getattr(f, 'is_TimeDependent', False)
```
FabioLuporini, line 407: same

FabioLuporini (approved):

FabioLuporini:
Merged, thanks

Issue #1513, opened 2020-11-19
Incorrect values are computed for `time_m` and `time_M` arguments


speglich:
Hello Folks, 

As noticed by @Nogueira, we are experiencing numeric precision errors with `np.float64` in adjoint tests, he introduced this errors on Slack ([here](https://devitocodes.slack.com/archives/C1HSXP6KB/p1605747162212800?thread_ts=1603204771.133200&cid=C1HSXP6KB)) and now we figured out what’s going on.  

The error isn’t connected with the precision used or with the equations implemented (e.g. acoustic self-adjoint), but with the time arguments (`time_m` and `time_M`) that is computed for forward and backward operators. The difference between `time_m` and `time_M` are the cause of this precision error, that only could be noticed when we used `np.float64`.

The adjoint test is attached on this issue, however we build a MFE to explain what we found. 

```python 
from devito import Grid, TimeFunction, Eq, Operator

grid = Grid(shape=(10,10))
p = TimeFunction(name="p", grid=grid,time_order=1, space_order=4)

stencil_forward = Eq(p.forward, p + 1.)
op1 = Operator(stencil_forward)
op1.apply(time=5)

print(op1.arguments(time=5))

stencil_backward = Eq(p.backward, p + 1.)
op2 = Operator(stencil_backward)
op2.apply(time=5)

print(op2.arguments(time=5))
```

In this MFE  we could notice that `time_m` parameter is different for both cases, causing different time steps for each operator. 

The `time_m` and `time_M ` values are computed by `_prepare_arguments` (if they are not passed by user) and uses `dspace` to avoid out of boundary acesses.
 As we know and expect the `dspace` computed in each operator are different, and are perfectly correct, but maybe the way that we construct the time range interval could be improved, once the out of boundary access don’t happen on this time dimensions. 

 
[adjoint_test.txt](https://github.com/devitocodes/devito/files/5567994/adjoint_test.txt)

mloubout:
So thinking about it, these arguments make sense for that MFE, in the reverse mode, the last computed element is `p[time_m -1]` so it has to be `time_m=1` to be valid.

In the forward mode, the first computed is `p[1]` and last is `p[6]`
In the reverse mode, the first computed is `p[4]` and last is `p[0]`

So I think with `time=6` in reverse mode should be adjoints, maybe  would be good to add a source to make sure the automatic bounds are correct.

nogueirapeterson:
[...]

PR #1567, opened 2021-01-21
examples: add first order adjoint viscoacoustic equations

nogueirapeterson:
Hi guys,

Once the issue (#1513)  was discussed here and also in [slack ](https://devitocodes.slack.com/archives/C1HSXP6KB/p1605747162212800?thread_ts=1603204771.133200&cid=C1HSXP6KB), we were able to develop a set of first-order adjoint viscoacoustic equations.

Thus, we finalized the implementation of the forward and adjoint viscoacoustic equations to first and second-order formulation.
 
The next step will be the implementation of the viscoacoustic Born operators!

### DIFF
--- a/examples/seismic/viscoacoustic/operators.py
+++ b/examples/seismic/viscoacoustic/operators.py
@@ -55,6 +55,8 @@ def sls_1st_order(model, geometry, p, **kwargs):
     p : TimeFunction
         Pressure field.
     """
+    forward = kwargs.get('forward', True)
+    space_order = p.space_order
     save = kwargs.get('save', False)
     s = model.grid.stepping_dim.spacing
     b = model.b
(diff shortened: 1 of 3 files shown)


mloubout:
@FabioLuporini @rhodrin With the amount of adjoint tests, we may want to make it a separate job in the workflow to avoid long CI.

georgebisbas (commented):
I am thinking of whether we could parameterize more each introduced implementation. So far quite a few bits of each viscoacoustic wave -equations are similar (if not the same). i.e. Look at the diff between the content of 
`def sls_1st_order` and `def sls_2nd_order` ?

Does it deserve to have 40+ lines of code while the actual diff is only 4-5 lines of code?

Just my two pennies as I am not an expert here. Can we think of sth to reduce the duplicate code ?

Again, thanks for the contribution!

speglich:
Hi @georgebisbas, we agree with you and refactor all viscoacoustic equations. We reduced a significant amount of code with this new approach.

georgebisbas (commented):
I think it is better now, thanks. Let's wait for the rest of the team,(more experienced in these topics to give a review). Thanks.

FabioLuporini:
> With the amount of adjoint tests, we may want to make it a separate job in the workflow to avoid long CI.

@mloubout that was a good point -- thanks to @rhodrin, done as of [this commit](https://github.com/devitocodes/devito/commit/363a12717e27f5dc68c3fb4236301cdc265cb63b)

@nogueirapeterson do you notice any variation in CI turnaround times? you may want to compare the timing of two `CI-core` runs -- latest master and this branch

examples/seismic/viscoacoustic/operators.py
```diff
@@ -42,29 +42,30 @@ def src_rec(p, model, geometry, **kwargs):
     return src_term + rec_term
 
 
-def sls_1st_order(model, geometry, p, **kwargs):
+def sls(model, geometry, p, **kwargs):
```
mloubout, line 45: I preferred it split into `1st_order` and `2nd order` lot easier to read and find things, you can always make a time order map to know which one to use (see how tti switch kernels for example).

speglich: We agree with you @mloubout when you say that is lot easier to read, our first implementation followed this approach. But following @georgebisbas suggestion, we believe that all code be more clean (without repetitions) and easy to understand, once order map is not necessary. We believe that both approaches are valid and if you don't agree we could undo this last modification.

FabioLuporini: I think I agree with Mathias @speglich 

I think the only situation in which we should factorize these functions is when the maths allows you to do so. So for example if a certain operator is a generalisation of the other, then go fo it (eg OT2 - OT4 kernels in iso-acoustic). But otherwise, keeping it split improves readability, even if the price to pay is some redundant code (and this is just setup code, not algorithms, so it's easier to swallow...)

FabioLuporini (approved):
LGTM, thanks

PR #1569, opened 2021-01-25
Introduce `deviceid` API to offload Operators on specific GPUs

FabioLuporini:


devito/core/gpu_openmp.py
```diff
@@ -325,41 +326,53 @@ def initialize(iet, **kwargs):
...
+            body = [Conditional(
+                CondNe(deviceid, -1),
+                Call('omp_set_default_device', [deviceid])
+            )]
 
-        return iet
+        init = List(header=c.Comment('Begin of OpenMP+MPI setup'),
+                    body=body,
+                    footer=(c.Comment('End of OpenMP+MPI setup'), c.Line()))
+        iet = iet._rebuild(body=(init,) + iet.body)
```
mloubout, line 371: Can't this be merged in a `setup(then, del, ran, ngpu, else)` with the acc one above, unless missread it looks like same call with just different then/else/decl.

mloubout: Actually then/else seem to be the only two different lines.

devito/operator/symbols.py
```diff
@@ -10,16 +10,15 @@ class SymbolRegistry(object):
...
         self.counters = {}
 
         # Special symbols
-        self.nthreads = nthreads
-        self.nthreads_nested = nthreads_nested
-        self.nthreads_nonaffine = nthreads_nonaffine
-        self.threadid = threadid
+        self.nthreads = NThreads(aliases='nthreads0')
+        self.nthreads_nested = NThreadsNested(aliases='nthreads1')
+        self.nthreads_nonaffine = NThreadsNonaffine(aliases='nthreads2')
```
mloubout, line 20: More explicit aliases names?

mloubout (commented):
Looks good, could use some common setup with the overlap.

tests/test_gpu_common.py
```diff
@@ -706,3 +706,32 @@ def test_xcor_from_saved(self, gpu_fit):
         op.apply(time_M=nt-1)
 
         assert np.all(g.data == 30)
+
+
+class TestDeviceScheduling(object):
+
+    def get_deviceid(self, op):
```
italoaug, line 713: Maybe this method would be more reusable if it was placed in the Operator class

devito/operator/symbols.py
FabioLuporini: I was actually thinking of dropping them since they're useless. The true names are what appears in the generated code and what people typically use in `op.apply(...)`

devito/core/gpu_openmp.py
FabioLuporini: It is indeed feasible. Do you mind if I do it here? https://github.com/devitocodes/devito/blob/refact-op-hierarchy/devito/passes/iet/specializations/openmp.py#L213
this is three branches after this one, and I'm in the middle of a big refactoring as you can see (I've moved and slightly changed the `initialize` methods) so I suggest I do it there. If it's OK with  you I can do it today and push it there directly, this would save some rebasing

tests/test_gpu_common.py
FabioLuporini: but I don't see why we should ever need it aside from testing

the better way, btw, would be adding it to the SymbolRegistry next to NThreads etc

devito/core/gpu_openmp.py
```diff
@@ -325,41 +326,53 @@ def initialize(iet, **kwargs):
 
     @singledispatch
     def _initialize(iet):
-        return iet
+        return iet, {}
 
     @_initialize.register(EntryFunction)
     def _(iet):
-        comm = None
+        # TODO: we need to pick the rank from `comm_shm`, not `comm`,
```
georgebisbas, line 333: Seems like duplicate code?

georgebisbas (commented):
So by default, for one rank, the computation will be offloaded to device? Right?

devito/core/gpu_openmp.py
mloubout: That's fine there yes, would avoid tracking duplicates omp/acc so would be nice.

devito/operator/symbols.py
mloubout: Then would drop aliases yeah.

FabioLuporini:
Approved offline by @mloubout , all other comments were minor and addressed or, as agreed, to be addressed in the upcoming PRs. Also, CI-gpu manually run and green: https://github.com/devitocodes/devito/actions/runs/509776872

Hence, merging

PR #1571, opened 2021-01-27
gpu: Add `devicerm` API for conditional deletions

FabioLuporini:
This PR provides an API to avoid deletion of GPU data before jumping back to Python land


FabioLuporini:
The tests make it fairly clear how this works in practice

tests/test_gpu_common.py
```diff
@@ -708,30 +708,48 @@ def test_xcor_from_saved(self, gpu_fit):
         assert np.all(g.data == 30)
 
 
-class TestDeviceScheduling(object):
+class TestAPI(object):
```
mloubout, line 711: bit generic, `TestGPUAPI`

mloubout (approved):

tests/test_gpu_common.py
FabioLuporini: ah, this is basically the previous name (I was using Device instead of GPU), I dropped it because this test already lives in `test_gpu_common.py` so it's implicit in the filename that the test is about the GPU API

FabioLuporini: the other classes follow the same convention

georgebisbas (commented):
We need MPI+GPU CI for this PR right? Do we have it ?

tests/test_gpu_common.py
mloubout: I guess yeah. But originally (and we made the labels) API was meant for the symbolic/stencil-DSL part not the compiler part. This is about the GPU compiler not the user API.

FabioLuporini: well, this test is actually testing GPU-related *user* API, in particular compiler-generated symbolic arguments, namely `deviceid=...` and `devicerm=...` that users can override at `op.apply(...)` time

FabioLuporini:
> We need MPI+GPU CI for this PR right? Do we have it ?

not sure what makes you think we need MPI+GPU here, but the answer is no (why?)

georgebisbas:
> > We need MPI+GPU CI for this PR right? Do we have it ?
> 
> not sure what makes you think we need MPI+GPU here, but the answer is no (why?)

I thought in general since the deviceid is used in MPI code as well:
```
  /* Begin of OpenMP+MPI setup */
  if (deviceid != -1)
  {
    omp_set_default_device(deviceid);
  }
  /* End of OpenMP+MPI setup */
```
Maybe we could a test and therefore needed ?

Just ignore me if I am overlooking sth!

PR #1572, opened 2021-01-27
Separate adjoint based tests

rhodrin:
This PR contains a small adjustment to the `pytest-core-nompi` workflow to 'parallelize' tests with names containing the string `'adjoint'` and those without.

Subject to no queues forming, this should speed up core CI by roughly 12 minutes.

### DIFF
--- a/.github/workflows/pytest-core-nompi.yml
+++ b/.github/workflows/pytest-core-nompi.yml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   pytest:
-    name: ${{ matrix.name }}
+    name: ${{ matrix.name }}-${{ matrix.set }}
     runs-on: "${{ matrix.os }}"
 
     env:
@@ -38,6 +38,7 @@ jobs:
            pytest-osx-py37-clang-omp,
            pytest-docker-py36-gcc-omp
         ]
+        set: [base, adjoint]
         include:
         - name: pytest-ubuntu-py36-gcc49-omp
           python-version: 3.6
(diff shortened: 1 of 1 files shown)


.github/workflows/pytest-core-nompi.yml
```diff
@@ -38,6 +38,7 @@ jobs:
            pytest-osx-py37-clang-omp,
            pytest-docker-py36-gcc-omp
         ]
+        set: [core, adjoint]
```
mloubout, line 41: Can we add it to the job name? Just to make it easier to find in the checks.

rhodrin: Yeah, having two jobs with the same name show up in CI isn't great - was looking for some suitable workaround but got disturbed. Will take another look later.

The first thing I did was just to add `pytest-docker-py36-gcc-omp-adjoint` etc to the matrix which is fine for now, but you end up with a long list and it's not very extendible if we want to add further separation in the future.

Ideally we can stick to something similar to this but where the job names are clearly distinguished. As I was writing I had an idea re. this, but will need to test later.

rhodrin: Oh, silly me, solution is simple, just modify [this](https://github.com/devitocodes/devito/blob/30c5b6cde649a7d9dd4924ca9a602adb1b83d25f/.github/workflows/pytest-core-nompi.yml#L15)...

rhodrin: >Can we add it to the job name? Just to make it easier to find in the checks.

I see that's what you meant now. Multitasking :p

PR #1583, opened 2021-02-09
Refactor Array sharing

FabioLuporini:
* Drop Array.sharing (was essentially a horrible way of carrying information around)
* Carry the information previously attached to Array via Properties (eg the new `PARALLEL_IF_PVT`)

This paves the way to an upcoming PR


devito/ir/iet/nodes.py
```diff
@@ -1177,6 +1181,54 @@ def root(self):
         return self.body[0]
 
 
+class ParallelBlock(Block):
```
FabioLuporini, line 1184: nothing new, just moved from `passes/iet/parpragma.py`

tests/test_dse.py
```diff
@@ -1716,44 +1717,6 @@ def test_blocking_options(self, rotate):
         assert np.isclose(expected, norm(u1), rtol=1e-5)
         assert np.isclose(expected, norm(u2), rtol=1e-5)
 
-    @pytest.mark.parametrize('rotate', [False, True])
-    def test_arrays_enforced_on_stack(self, rotate):
```
FabioLuporini, line 1720: Dropped as now the tensor temporaries can only be allocated on the heap (stack was flaky and performance-wise useless, so why keep it?)

devito/ir/iet/nodes.py
```diff
@@ -1122,14 +1127,13 @@ def _process_kwargs(cls, **kwargs):
...
-    A sequence of Nodes, wrapped in a parallel block {...}.
-    """
-
-    is_ParallelBlock = True
+    @cached_property
+    def collapsed(self):
+        ret = [self]
+        for i in range(self.ncollapsed - 1):
+            ret.append(ret[i].nodes[0])
+        assert all(i.is_Iteration for i in ret)
```
mloubout, line 1135: why not just `ret.append(ret[i].nodes[0] if ret[i].nodes[0]. is_Iteration)` ?

mloubout (approved):

devito/ir/iet/nodes.py
FabioLuporini: because all of the nodes are expected to be Iterations at that point, and an assertion should cause a failure if that's not the case

FabioLuporini:
CI-gpu is [green](https://github.com/devitocodes/devito/actions/runs/554647363)

Merging

PR #1584, opened 2021-02-10
Fix find_library issue to MacOS Big Sur

### LINKED ISSUE #1582: Couldn't find `libc`'s `posix_memalign` to allocate memory in MacOS Big Sur
In `MacOS Big Sur` dylibs are removed from the disk when they are added to the shared dyld cache. Thus, ctypes.utils.find_libary cannot find the libraries, as we were used to.

Seems that this issue was solved here for cpython 3.9.1 and any patch is necessary.  However it is only available in the release candidate of the Apple Universal Binary version.

Applications that must be compatible with older python versions are writing patches, like this:

```python
elif os.name == "posix" and sys.platform == "darwin":
    from ctypes.macholib.dyld import dyld_find as _dyld_find
    def find_library(name):
        possible = ['@executable_path/../lib/lib%s.dylib' % name,
                    'lib%s.dylib' % name,
                    '%s.dylib' % name,
                    '%s.framework/%s' % (name, name)]
        for name in possible:
            try:
                return _dyld_find(name)
            except ValueError:
                continue
        return None
```

Following the @FabioLuporini suggestion we could implement the same patch like this:

```python
try:
    ... = find_library(...)
except ...:
    # Special cases
    if os.name == "posix" and sys.platform == "darwin":
         ...
    else:
         <suitable error message>
```

speglich:
Closes #1582

### DIFF
--- a/devito/data/allocators.py
+++ b/devito/data/allocators.py
@@ -140,8 +140,7 @@ class PosixAllocator(MemoryAllocator):
             try:
                 cls.lib = ctypes.CDLL(handle)
             except OSError:
-                raise RuntimeError("Couldn't find `libc`'s `posix_memalign` to "
-                                   "allocate memory")
+                cls.lib = None
 
     def _alloc_C_libcall(self, size, ctype):
         if not self.available():
--- a/devito/data/allocators.py
+++ b/devito/data/allocators.py
@@ -130,6 +130,9 @@ class PosixAllocator(MemoryAllocator):
     def initialize(cls):
         handle = find_library('c')
 
+        # Special case: on MacOS Big Sur any code that attempts to check
+        # for dynamic library presence by looking for a file at a path
+        # will fail. For this case, a static path is defined.
         if handle is None and os.name == "posix" and sys.platform == "darwin":
             handle = '/usr/lib/libc.dylib'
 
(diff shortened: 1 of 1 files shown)


mloubout:
Any idea how this could be tested? don't think gh-action has bigsur

devito/data/allocators.py
```diff
@@ -128,6 +129,10 @@ class PosixAllocator(MemoryAllocator):
     @classmethod
     def initialize(cls):
         handle = find_library('c')
+
+        if handle is None and os.name == "posix" and sys.platform == "darwin":
```
FabioLuporini, line 133: Can you add a comment along the lines of:

```
# Special case: on MacOS Big Sur ....
```

speglich: Of course I can, sorry about that!

devito/data/allocators.py
```diff
@@ -128,8 +129,19 @@ class PosixAllocator(MemoryAllocator):
...
+        # will fail. For this case, a static path is provided.
+        if handle is None and os.name == "posix" and sys.platform == "darwin":
+            handle = '/usr/lib/libc.dylib'
+
         if handle is not None:
-            cls.lib = ctypes.CDLL(handle)
+            try:
+                cls.lib = ctypes.CDLL(handle)
+            except OSError:
+                raise RuntimeError("Couldn't find `libc`'s `posix_memalign` to "
```
FabioLuporini, line 143: no need to raise an error; just `cls.lib = None` and the error will be triggered, at a more suitable time (ie upon the first data allocation), by the next call to `_alloc_C_libcall`

PR #1585, opened 2021-02-11
compiler: Improve the cost model used by CIRE

### LINKED ISSUE #1586: CIRE's maxalias option may lead to getting stuck in infinite looping
MFE

```
grid = Grid(shape=(20, 20, 20)) 

f = Function(name='f', grid=grid)
u = TimeFunction(name='u', grid=grid, space_order=4)

eqn = Eq(u.forward, (2*f*f*u.dy).dy + (3*f*u.dy).dy)

op = Operator(eqn, opt=('advanced', {'openmp': True, 'cire-maxalias': True}))
```

FabioLuporini:
fixes #1586


devito/passes/clusters/aliases.py
```diff
@@ -211,9 +216,19 @@ def extract(cls, n, context, min_cost, max_alias, cluster, sregistry):
...
+    @classmethod
+    def _extract_rule(cls, context, min_cost, cluster):
+        return lambda e: (e.is_Pow and e.exp.is_Integer and e.exp < 0 and
+                          all(i.function.is_const for i in e.base.free_symbols))
+
     @classmethod
-    def selector(cls, min_cost, cost, naliases):
-        return cost >= min_cost and naliases >= 1
+    def selector(cls, min_cost, cost):
+        return int(cost > 0)
```
mloubout, line 231: not `> min_cost` ?

devito/symbolics/inspection.py
```diff
@@ -79,7 +79,7 @@ def estimate_cost(exprs, estimate=False):
             * Powers with integer exponent `n>0` count as n-1 ops (as if
               it were a chain of multiplications).
     """
-    trascendentals_cost = {sin: 50, cos: 50, exp: 50, log: 50}
+    trascendentals_cost = {sin: 100, cos: 100, exp: 100, log: 100}
```
mloubout, line 82: Why the increase?

examples/performance/01_gpu.ipynb
```diff
@@ -289,6 +289,9 @@
       "\n",
       "  #pragma omp target enter data map(to: u[0:u_vec->size[0]][0:u_vec->size[1]][0:u_vec->size[2]])\n",
       "\n",
+      "  float r3 = 1.0/dt;\n",
+      "  float r4 = 1.0/(h_x*h_x);\n",
+      "  float r5 = 1.0/(h_y*h_y);\n",
```
mloubout, line 294: that's nice

tests/test_dle.py
```diff
@@ -323,7 +323,7 @@ def test_cache_blocking_imperfect_nest_v2(blockinner):
     shape = (16, 16, 16)
     grid = Grid(shape=shape, dtype=np.float64)
 
-    u = TimeFunction(name='u', grid=grid, space_order=2)
+    u = TimeFunction(name='u', grid=grid, space_order=4)
```
mloubout, line 326: why the order increase?

devito/passes/clusters/aliases.py
```diff
@@ -211,9 +216,19 @@ def extract(cls, n, context, min_cost, max_alias, cluster, sregistry):
...
 
+
+class CallbacksDivs(CallbacksInvariants):
+
+    mode = 'divs'
+
+    @classmethod
+    def _extract_rule(cls, context, min_cost, cluster):
+        return lambda e: (e.is_Pow and e.exp.is_Integer and e.exp < 0 and
+                          all(i.function.is_const for i in e.base.free_symbols))
```
mloubout, line 227: So functions divs not considered?

FabioLuporini: they will be picked already by `cire(..., 'invariants', ...)`

rationale: here I'm looking for scalar divisions. These are negligible when running on CPUs (the jit compiler will hoist them out of the innermost loop, so their cost will be completely amortized out over a bunch of loop iterations), but they typically have an impact on GPUs, since it would be one division per GPU thread.

devito/passes/clusters/aliases.py
FabioLuporini: no, I want all scalar divisions to be hoisted out, so `min_cost` is ignored. In fact, in `divs` mode, `min_cost` contains a dummy value (I think `None`)

devito/symbolics/inspection.py
FabioLuporini: as the PR title says, I've tweaked the cost model for aliases. Basically now I'm not lifting expressions (and therefore I'm not allocating tensor temporaries, thus increasing the working set size) if the trade-off with the saved number of operations isn't favorable. I've tweaked the weights and the CIRE algorithms to improve the cost model

tests/test_dle.py
FabioLuporini: due to the new cost model, which is now not allocating temporaries if the reduction in operation count isn't sufficient according to the cost model heuristics. With so=4 we have larger expressions, so CIRE does its job

FabioLuporini:
CI-gpu green: https://github.com/devitocodes/devito/actions/runs/571020418

FabioLuporini:
Approved on slack by @mloubout , merging

PR #1588, opened 2021-02-16
Solve time only

mloubout:
Refactor `solve` to only evaluate time derivative by default  (will evaluate full equation if doesn't find a solution) and corresponding test.

### DIFF
--- a/devito/types/equation.py
+++ b/devito/types/equation.py
@@ -242,7 +242,6 @@ def solve(eq, target, **kwargs):
         # Try first linear solver
         try:
             sol = list(sympy.linsolve([eval_t(e)], [t]))
-            print(eval_t(e), sol)
             sols.append(sol[0][0])
         except ValueError:
             warning("Equation is not affine w.r.t the target, falling back to standard"
--- a/devito/types/equation.py
+++ b/devito/types/equation.py
@@ -1,7 +1,6 @@
 """User API to specify equations."""
 
 import sympy
-from sympy.solvers.solveset import linear_coeffs
 
 from cached_property import cached_property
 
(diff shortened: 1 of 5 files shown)


devito/types/equation.py
```diff
@@ -255,3 +254,17 @@ def solve(eq, target, **kwargs):
...
+
+
+def eval_t(expr):
+    """
+    Evaluate all time derivatives in the expression
+    """
+    try:
+        assert any(d.is_Time for d in expr.dims)
+        return expr.evaluate
+    except:
```
FabioLuporini, line 266: missing exception

devito/types/equation.py
```diff
@@ -255,3 +254,17 @@ def solve(eq, target, **kwargs):
...
+        try:
```
FabioLuporini, line 267: nested exceptions just be avoided. This is nitpicking (only two levels), but the more pythonic way is to `pass`, de-indent the nested `try-except`, add a comment, and go

devito/types/equation.py
```diff
@@ -255,3 +254,17 @@ def solve(eq, target, **kwargs):
         return target.new_from_mat(sols)
     else:
         return sols[0]
+
+
+def eval_t(expr):
```
FabioLuporini, line 259: better name? also, move to fd/tools?

devito/types/equation.py
```diff
@@ -255,3 +254,17 @@ def solve(eq, target, **kwargs):
...
+    """
+    Evaluate all time derivatives in the expression
```
FabioLuporini, line 261: full stop

tests/test_dse.py
```diff
@@ -1800,12 +1800,12 @@ def test_fullopt(self):
         assert len(op0._func_table) == 0
         assert len(op1._func_table) == 1  # due to loop blocking
 
-        assert summary0[('section0', None)].ops == 50
+        assert summary0[('section0', None)].ops == 46
```
FabioLuporini, line 1803: ah, very interesting...

FabioLuporini:
closing as unnecessary (discussed offline)

PR #1591, opened 2021-02-19
compiler: Add option to use Functions, in place of Arrays, for compiler-generated temporary

FabioLuporini:


devito/types/dense.py
```diff
@@ -1454,3 +1453,99 @@ def parent(self):
         return self._parent
 
     _pickle_kwargs = Function._pickle_kwargs + ['parent']
+
+
+class TempFunction(DiscreteFunction):
```
mloubout, line 1458: Can we have some docstrings?

devito/types/dense.py
```diff
@@ -1454,3 +1453,99 @@ def parent(self):
...
+        super().__init_finalize__(*args, **kwargs)
+
+        self._pointer_dim = kwargs.get('pointer_dim')
+
+    @classmethod
+    def __indices_setup__(cls, **kwargs):
+        pointer_dim = kwargs.get('pointer_dim')
+        dimensions = as_tuple(kwargs['dimensions'])
+        if pointer_dim not in dimensions:
+            # This is a bit hacky but it does work around duplicate dimensions when
```
mloubout, line 1472: That sounds safe 😓

devito/types/dense.py
```diff
@@ -1454,3 +1453,99 @@ def parent(self):
...
+        return dimensions, dimensions
+
+    def __halo_setup__(self, **kwargs):
+        pointer_dim = kwargs.get('pointer_dim')
+        dimensions = as_tuple(kwargs['dimensions'])
+        halo = as_tuple(kwargs.get('halo'))
+        if halo is None:
+            halo = tuple((0, 0) for _ in dimensions)
+        if pointer_dim is not None:
+            halo = ((0, 0),) + halo
```
mloubout, line 1484: So `TempFunction` can't have spatial order halo (or space_order for that matters)

devito/types/dense.py
```diff
@@ -1454,3 +1453,99 @@ def parent(self):
...
+    data_with_halo = data
+    data_ro_domain = data
+    data_ro_with_halo = data
+
+    @property
+    def pointer_dim(self):
+        return self._pointer_dim
+
+    @property
+    def dim(self):
```
mloubout, line 1502: why the need for both?

devito/types/dense.py
```diff
@@ -1454,3 +1453,99 @@ def parent(self):
...
+            Shape of the domain region in grid points.
+        initializer : callable or any object exposing the buffer interface, optional
+            Data initializer. If a callable is provided, data is allocated lazily.
+        """
+        # Sanity check
+        if len(shape) != self.ndim:
+            raise ValueError("`shape` must contain %d integers, not %d"
+                             % (self.ndim, len(shape)))
+
+        return Function(name=self.name, dtype=self.dtype, dimensions=self.dimensions,
```
mloubout, line 1529: What if any of the dimensions is time?

mloubout (commented):
Still not sure what this is for, the test makes it look like it's not really a user side object and very intricate to use (requires to know what the generated code exactly is).

FabioLuporini:
@mloubout 

> Still not sure what this is for, the test makes it look like it's not really a user side object and very intricate to use (requires to know what the generated code exactly is).

You're right. I'm going to add docstrings in my next push.

As you correctly guessed, this is _not_ a user API object. These may be created by Devito in place of Arrays to implement tensor temporaries. The key difference is that an Array mallocs and frees memory directly in C-land, while a TempFunction is gonna give control of memory management to the user. I do not expect the average user to care about this (TempFunctions are never going to be used unless the user requests them through an Operator option).

devito/types/dense.py
FabioLuporini: adding

devito/types/dense.py
FabioLuporini: indeed. Irrelevant as the users doesn't create them explicitly

devito/types/dense.py
FabioLuporini: Make it usable where we have a PointerArray, as it's `PointerArray.dim`. It'd be cleaner with a Mixin, I admit... I may add one

devito/types/dense.py
FabioLuporini: undefined behaviour. But temporaries are never created along time, so we should be fine

devito/types/dense.py
```diff
@@ -1454,3 +1453,133 @@ def parent(self):
...
+
+    Examples
+    --------
+    The `make` method makes the TempFunction create a new Function. For more info,
+    refer to TempFunction.make.__doc__.
+
+      .. code-block:: python
+
+        op = Operator(...)
+        cfuncs = [i for i in op.input if i.is_TempFunction]
```
mloubout, line 1484: May wanna add an `op.tmp_function` property for ease of use.

devito/types/dense.py
mloubout: Can we add a check there is no time dim for safety?

mloubout (commented):
Minor comments

devito/types/dense.py
FabioLuporini: Done, added `op.temporaries`

devito/types/dense.py
FabioLuporini: Done, added a few lines above in `__indices_setup__` , and I've actually made it stronger by checking for `not any(i.is_NonlinearDerived ...)`

tests/test_dse.py
```diff
@@ -1787,6 +1787,67 @@ def test_blocking_options(self, rotate):
...
+        op0 = Operator(eqn, opt=('noop', {'openmp': True}))
+        op1 = Operator(eqn, opt=('advanced', {'openmp': True, 'cire-mincost-sops': 1,
+                                              'cire-ftemps': True}))
+        op2 = Operator(eqn, opt=('advanced-fsg', {'openmp': True, 'cire-mincost-sops': 1,
+                                                  'cire-ftemps': True}))
+
+        op0(time_M=1, nthreads=nthreads)
+
+        # TempFunctions expect an override
+        try:
```
tjb900, line 1826: Just saw this when looking for this example code. I think pytest gives you a mechanism for asserting that exceptions are thrown (and indeed, as I look again I see that this code _won't_ currently error in the case that no exception is thrown)

```
with pytest.raises(InvalidArgument):
    op1(...)
```

tests/test_dse.py
```diff
@@ -1787,6 +1787,67 @@ def test_blocking_options(self, rotate):
...
+            op1(time_M=1, u=u1)
+        except InvalidArgument:
+            assert True
+        except:
+            assert False
+
+        # Prepare to run op1
+        shape = [nthreads, x0_blk0_size, y0_blk0_size, grid.shape[-1]]
```
tjb900, line 1834: ohhh...  I had not thought about this. How is one supposed to know what to pass to `make`, here - if it's not always the `grid.shape`?

This seems to mean that the caller has to be aware of the details of the generated code. I guess we can live with that, but I'd prefer more safeguards if possible.  Could there potentially be some protection in `arg_values` that ensures the passed temporary is sufficiently large, given the rest of the arguments?

The sequence of events I'm worried about is:
 * devito version X produces C code that needs temporaries of a certain shape
 * application hard-codes temporaries of that shape using a construction very similar to this line
 * devito version X+1 produces C code that needs a slightly larger temporary, but code still passes same size
 * segfaults, incorrect results, chaos reigns
 
Alternatively, or perhaps as well, could the arguments to `make` be _all_ the arguments that will be passed to the operator, and make can extract whatever values it needs from those?

FabioLuporini: >This seems to mean that the caller has to be aware of the details of the generated code.

Well, the idea was that you query the TempFunction's shape. This gives you the _symbolic_ shape of the TempFunction from which you can infer what's required.

> Alternatively, or perhaps as well, could the arguments to make be all the arguments that will be passed to the operator, and make can extract whatever values it needs from those?

yes, if this is feasible on your side, it should be straightforward on our side. So the idea would be (sort of):

```
def make(..., **kwargs):
    # Sanity checks...
  
    try:  
        shape = tuple(i.subs(kwargs) for I in self.shape)
    except ...:
        raise ValueError('Don't know how to derive shape for ...') 
```

?

tests/test_dse.py
FabioLuporini: ah, good to know! Will update the test.

> as I look again I see that this code won't currently error in the case that no exception is thrown)

That's my mistake. I should have added an exception below `op1(...)`

tests/test_dse.py
tjb900: Ah, yeah I hadn't appreciated that we could get the symbolic shape. That resolves my concern for now, I think.

FabioLuporini:
@tjb900 :

* rebased on master
* changed to `with pytest.raises(...)` (thanks!)
* added prototypal API so that you can do `i.make(**kwargs)` instead of providing the shape explicitly. Test updated accordingly
* More roboust sanity checks inside `make`

Issue #1589, opened 2021-02-18
Symbolic coefficients on staggered stencils not working correctly


EdCaunt:
Setting `coefficents=symbolic` on a function with `staggered` and not specifying stencil coefficients results in derivatives defaulting to incorrect values. It appears that specifying coefficients does not change them from this incorrect default.

A short example:
```
from devito import Grid, Function, NODE, Eq, Operator, Coefficient, Substitutions
import numpy as np
shape = (11,)
extent = (10.,)
grid = Grid(shape=shape, extent=extent)
x = grid.dimensions[0]
f = Function(name='f', grid=grid, space_order=2, staggered=NODE)
g = Function(name='g', grid=grid, space_order=2, staggered=NODE,
             coefficients='symbolic')
h = Function(name='h', grid=grid, space_order=2, staggered=NODE,
             coefficients='symbolic')
vf = Function(name='v_f', grid=grid, space_order=2, staggered=x)
vg = Function(name='v_g', grid=grid, space_order=2, staggered=x)
vh = Function(name='v_h', grid=grid, space_order=2, staggered=x)
h_coeffs = Coefficient(1, h, x, np.array([0., -1., 1.]))
coeffs = Substitutions(h_coeffs)
eqf = Eq(vf, f.dx)
eqg = Eq(vg, g.dx)
eqh = Eq(vh, h.dx, coefficients=coeffs)
# Should all result in the same kernel
opf = Operator(eqf)
print(opf.ccode)
# Different stencil to the normal one. Should default to same
opg = Operator(eqg)
print(opg.ccode)
# Different stencil to normal one. Manually set to same
oph = Operator(eqh)
print(oph.ccode)
```

One would expect all of these operators to have the same ccode (ignoring differences in variable names). However, this is not the case. The latter two do not have the same stencil operation. `Eq(vf, f.dxc).evaluate` returns `Eq(v_f(x + h_x/2), -f(x)/h_x + f(x + h_x)/h_x)` whilst `Eq(vg, g.dx).evaluate` and `Eq(vh, h.dx, coefficients=coeffs)` both return `Eq(v_g(x + h_x/2), g(x + h_x)/(2*h_x))`

EdCaunt:
This was also found to affect substitutions at forward and backward time steps (although this may have been fixed since?). However, a new MFE is:
```
import numpy as np
from devito import *

grid = Grid(shape=(10, 10))
[...]

PR #1595, opened 2021-02-22
dsl: Patch symbolic coefficients with staggered grids

EdCaunt:
Fix for the bug where derivatives with symbolic coefficients on staggered grids default to incorrect values. Fixes one of the two issues detailed in #1589.

### DIFF
--- a/devito/finite_differences/coefficients.py
+++ b/devito/finite_differences/coefficients.py
@@ -259,7 +259,6 @@ def default_rules(obj, functions):
     # Determine which 'rules' are missing
     sym = get_sym(functions)
     terms = obj.find(sym)
-    # DEBUG: Only seem to have two terms here?
     args_present = filter_ordered(term.args[1:] for term in terms)
 
     subs = obj.substitutions
@@ -277,8 +276,6 @@ def default_rules(obj, functions):
 
     rules = {}
     for i in not_provided:
-        # DEBUG: For every not provided, generates a set of rules and appends to
-        # DEBUG: existing dictionary
         rules = {**rules, **generate_subs(*i)}
 
     return rules
--- a/devito/finite_differences/coefficients.py
+++ b/devito/finite_differences/coefficients.py
@@ -3,7 +3,7 @@ import numpy as np
 from cached_property import cached_property
 
 from devito.finite_differences import generate_indices
-from devito.tools import filter_ordered, as_tuple
+from devito.tools import filter_ordered, as_tuple, frozendict
 from devito.symbolics.search import retrieve_dimensions
 
 __all__ = ['Coefficient', 'Substitutions', 'default_rules']
@@ -243,8 +243,10 @@ def default_rules(obj, functions):
 
         subs = {}
 
+        mapper = frozendict({dim: index})
+
         indices, x0 = generate_indices(function, dim,
-                                       fd_order, side=None)
+                                       fd_order, side=None, x0=mapper)
 
         coeffs = sympy.finite_diff_weights(deriv_order, indices, x0)[-1][-1]
 
@@ -257,10 +259,12 @@ def default_rules(obj, functions):
     # Determine which 'rules' are missing
     sym = get_sym(functions)
     terms = obj.find(sym)
+    # DEBUG: Only seem to have two terms here?
     args_present = filter_ordered(term.args[1:] for term in terms)
 
     subs = obj.substitutions
     if subs:
+        # .index is not including function offset correctly
         args_provided = [(i.deriv_order, i.function, i.index)
                          for i in subs.coefficients]
     else:
@@ -273,6 +277,8 @@ def default_rules(obj, functions):
 
     rules = {}
     for i in not_provided:
+        # DEBUG: For every not provided, generates a set of rules and appends to
+        # DEBUG: existing dictionary
         rules = {**rules, **generate_subs(*i)}
 
     return rules
--- a/devito/finite_differences/coefficients.py
+++ b/devito/finite_differences/coefficients.py
@@ -3,7 +3,7 @@ import numpy as np
 from cached_property import cached_property
 
 from devito.finite_differences import generate_indices
-from devito.tools import filter_ordered, as_tuple, frozendict
+from devito.tools import filter_ordered, as_tuple
 from devito.symbolics.search import retrieve_dimensions
 
 __all__ = ['Coefficient', 'Substitutions', 'default_rules']
@@ -243,7 +243,7 @@ def default_rules(obj, functions):
 
         subs = {}
 
-        mapper = frozendict({dim: index})
+        mapper = {dim: index}
 
         indices, x0 = generate_indices(function, dim,
                                        fd_order, side=None, x0=mapper)
@@ -263,7 +263,6 @@ def default_rules(obj, functions):
 
     subs = obj.substitutions
     if subs:
-        # .index is not including function offset correctly
         args_provided = [(i.deriv_order, i.function, i.index)
                          for i in subs.coefficients]
     else:
--- a/devito/finite_differences/tools.py
+++ b/devito/finite_differences/tools.py
@@ -163,7 +163,7 @@ def generate_indices(func, dim, order, side=None, x0=None):
         Order of the finite-difference scheme
     side: Side
         Side of the scheme, (centered, left, right)
-    x0: Dimension or Expr or Number
+    x0: Dict of {Dimension: Dimension or Expr or Number}
         Origin of the scheme, ie. `x`, `x + .5 * x.spacing`, ...
 
     Returns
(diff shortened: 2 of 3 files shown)


devito/finite_differences/coefficients.py
```diff
@@ -243,8 +243,10 @@ def generate_subs(deriv_order, function, index):
 
         subs = {}
 
+        mapper = frozendict({dim: index})
```
mloubout, line 246: I still don't get why this all piece of code exist (`default_rule`). In `finite_difference.py` the indices are computed already should only need to add the coefficients to it (but maybe I'm missing something @rhodrin i the loop).

tests/test_symbolic_coefficients.py
```diff
@@ -119,6 +119,83 @@ def test_coefficients_w_xreplace(self):
...
+    @pytest.mark.parametrize('order', [1, 2, 6, 8])
+    @pytest.mark.parametrize('extent', [1., 10., 100.])
+    @pytest.mark.parametrize('conf', [{'l': 'NODE', 'r1': 'x', 'r2': None},
+                                      {'l': 'NODE', 'r1': 'y', 'r2': None},
+                                      {'l': 'NODE', 'r1': '(x, y)', 'r2': None},
+                                      {'l': 'x', 'r1': 'NODE', 'r2': None},
+                                      {'l': 'y', 'r1': 'NODE', 'r2': None},
+                                      {'l': '(x, y)', 'r1': 'NODE', 'r2': None},
+                                      {'l': 'NODE', 'r1': 'x', 'r2': 'y'}])
+    def test_default_rules_equation(self, order, extent, conf):
```
mloubout, line 132: I would have a pure simple symbolic test first, i.e just check lhs/rhs of the symbolic evaluated equation.

EdCaunt: As in do `str(expr)` of each `Eq` and check for equality?

mloubout: Yes, for example
https://github.com/devitocodes/devito/blob/5938e7f198f8eb9cfd5e8889c8ee9ba203eb4a44/tests/test_derivatives.py#L342

EdCaunt: Is there some way to get the coefficients as float? The only roadblock is that the symbolic coefficients default to fractions, whilst the standard coefficients are in floating point form, so comparison obvs returns False

EdCaunt: Done in the latest push

devito/finite_differences/coefficients.py
EdCaunt: To my understanding, the process is: Stencils evaluated but with `w(x+h_x)` etc instead of standard coefficients -> Fill in the coefficients for which no substitutions have been given -> Fill in the coefficients for which substitutions have been given. I presume the indices are 'baked in' and discarded once step one has been carried out? As far as I can tell, they are not readily available at the start of the second step

tests/test_symbolic_coefficients.py
```diff
@@ -119,6 +119,88 @@ def test_coefficients_w_xreplace(self):
...
+            b_std.data[::2, ::2] = 1.
+            b_sym.data[::2, ::2] = 1.
+        else:
+            b_std, b_sym = 0., 0.
+
+        eq_std = get_eq(u_std, a_std, b_std, conf)
+        eq_sym = get_eq(u_sym, a_sym, b_sym, conf)
+
+        evaluated_std = eq_std.evaluate.evalf(_PRECISION)
+        evaluated_sym = eq_sym.evaluate.evalf(_PRECISION)
```
mloubout, line 195: Sorry nitpicky, but this still only test get_eq against itself, can we get explicit symbolic expressions check so that no matter what happens to `.dx`, .... it will always have to pass on the aboslute true answer.

EdCaunt: I'm a little confused here? `get_eq` is just a function I put in the test to tidy stuff up so I'm not repeating code (it just returns the equation as specified by the config). What the test is doing is setting up two identical equations, one with symbolic coefficients and one without and making sure they produce the same result (and that the string of the expressions is the same). So `{'l': 'NODE', 'r1': '(x, y)', 'r2': None}` results in checking `Eq(u, a.dx + a.dy)` where `a` is staggered in `(x, y)` and `{'l': 'NODE', 'r1': 'x', 'r2': 'y'} ` results in checking `Eq(u, a.dx + b.dy)` where each is staggered in the specified direction. Unless you mean have a set of known answers fed in by `parameterize`?

mloubout: But you still rely on the shortcuts. SO the shortcut may change and the test still pass while wrong because it only check  "u.dx == u.dx" not ""u.dx == the hard coded expression it's supposed to be"

mloubout: You want to check:
- The symbolics are correct. This needs no data or operator, only purely checking the expression. For example `assert u.dx == 1/dx (u[x-1] - u[x+1])`
- The result is what I expect. This runs the operator and checks it produces the correct numbers after the compiler.

EdCaunt: Ah, so say if there was a change such that all `.dx` were broken, it wouldn't pick it up unless you were comparing to the explicitly written answers? I may have to reduce the number of scenarios tested, because the list of answers is going to get really big otherwise.

mloubout: You can just split it on two test, keep this one as is and add a smaller one for the pure symbolic part with just a few test cases (basically just add the test case was discussed on slack).

EdCaunt: Ok

tests/test_symbolic_coefficients.py
```diff
@@ -119,6 +119,110 @@ def test_coefficients_w_xreplace(self):
...
+                                      {'l': 'x', 'r1': 'NODE', 'r2': None},
+                                      {'l': 'y', 'r1': 'NODE', 'r2': None},
+                                      {'l': '(x, y)', 'r1': 'NODE', 'r2': None},
+                                      {'l': 'NODE', 'r1': 'x', 'r2': 'y'}])
+    def test_default_rules_vs_standard(self, order, extent, conf):
+        """
+        Test that equations containing default symbolic coefficients evaluate to
+        the same expressions as standard coefficients for the same function.
+        """
+        def function_setup(name, grid, order, stagger):
```
FabioLuporini, line 136: maybe you can pull these private functions out of the test?  might they be useful for other tests too?

tests/test_symbolic_coefficients.py
```diff
@@ -119,6 +119,110 @@ def test_coefficients_w_xreplace(self):
...
+                                                  ' + 3*f(x + 3*h_x)/(640*h_x)')])
+    def test_default_rules_vs_string(self, so, expected):
+        grid = Grid(shape=(11,), extent=(10.,))
+        x = grid.dimensions[0]
+        f = Function(name='f', grid=grid, space_order=so, staggered=NODE,
+                     coefficients='symbolic')
+        g = Function(name='g', grid=grid, space_order=so, staggered=x,
+                     coefficients='symbolic')
+        eq = Eq(g, f.dx)
+        print(eq.evaluate.rhs)
```
FabioLuporini, line 222: leftover debug print

tests/test_symbolic_coefficients.py
```diff
@@ -138,6 +242,7 @@ def test_staggered_array(self, order):
 
         eq_f = Eq(f, f.dx2, coefficients=Substitutions(coeffs_f))
         eq_g = Eq(g, g.dx2, coefficients=Substitutions(coeffs_g))
+        # Evaluate and check against one another?
```
FabioLuporini, line 245: ?

EdCaunt: Comments I left for myself when writing and forgot to clean up. Will remove

tests/test_symbolic_coefficients.py
EdCaunt: I think leave them where they are for now, but maybe add a comment to say that they should be pulled outside if another test needs them in the future?

rhodrin:
Looks good @EdCaunt. I think it just needs those additional tests we discussed today. Ping me when it's ready for me to take a final look.

EdCaunt:
> Looks good @EdCaunt. I think it just needs those additional tests we discussed today. Ping me when it's ready for me to take a final look.

I thought they were already in there? Unless I forgot one?

devito/finite_differences/tools.py
```diff
@@ -163,7 +163,7 @@ def generate_indices(func, dim, order, side=None, x0=None):
         Order of the finite-difference scheme
     side: Side
         Side of the scheme, (centered, left, right)
-    x0: Dimension or Expr or Number
+    x0: Dict of {Dimension: Dimension or Expr or Number}
```
FabioLuporini, line 166: nitpicking (you may fix it the next time): type name is "dict" not "Dict"

FabioLuporini (approved):

PR #1604, opened 2021-02-26
mpi: Add ability to specify MPI topology used for grid division

FabioLuporini:


mloubout (commented):
Can we have an example/test with this one?

FabioLuporini:
Yes, I will add one on Monday

FabioLuporini:
@mloubout: test added

Will update the Wiki FAQ once merged.

PR #1605, opened 2021-02-26
misc: Use pickled soname rather than generating a new one

FabioLuporini:
This makes the pickling-unpickling process faster by avoiding re-compilation (ie Operator code regeneration) upon unpickling.


devito/operator/operator.py
```diff
@@ -828,28 +829,14 @@ def __getnewargs_ex__(self):
         return (None,), {}
 
     def __setstate__(self, state):
-        soname = state.pop('_soname', None)
+        soname = state.pop('soname', None)
         binary = state.pop('binary', None)
         for k, v in state.items():
             setattr(self, k, v)
-        # If the `sonames` don't match, there *might* be a hidden bug as the
```
mloubout, line 835: So that's not an issue anymore?

FabioLuporini: That wasn't really an issue strictly speaking -- more of a sanity check upon unpickling aimed to spot bugs somewhere else. Which has non negligible overhead

mloubout: Ok. Is there a test for it?

FabioLuporini: yes, pickling tests for _all_ (many are picked recursively) Devito objects are available in `test_pickle.py`.

georgebisbas (approved):

PR #1607, opened 2021-02-26
mpi: Pass MPI_COMM_SELF to Distributor upon unpickling

FabioLuporini:


mloubout (approved):

tests/test_pickle.py
```diff
@@ -454,6 +454,20 @@ def test_mpi_objects():
     assert obj.dtype == new_obj.dtype
 
 
+@skipif(['nompi'])
+@pytest.mark.parallel(mode=[2])
+def test_mpi_grid():
+    grid = Grid(shape=(4, 4, 4))
+
+    pkl_grid = pickle.dumps(grid)
+    new_grid = pickle.loads(pkl_grid)
```
tjb900, line 463: So this makes sense to me, and shows that we're getting a single-rank grid when we unpickle.

The original motivation though is that unpickling works (i.e. does not hang) even if only one rank calls `loads` - is it worth having the test replicate that requirement?

tjb900 (approved):
LGTM, just a possible suggestion to make the test closer reflect the requirement.

tests/test_pickle.py
FabioLuporini: Good point. I have updated the test. I have also reproduced the hanging issue if we don't use COMM_SELF

PR #1608, opened 2021-02-26
dsl: Enable overriding over SubDimension thickness

FabioLuporini:


mloubout (approved):
Seems ok, @rhodrin mpi part looks sane right?

mloubout (approved):

PR #1615, opened 2021-03-02
arch: Add nvidia-smi parser to GPU checking

Leitevmd:


devito/arch/archinfo.py
```diff
@@ -216,13 +216,36 @@ def lspci_gpu_info(text):
...
+
     # Run homogeneity checks on a list of GPU, return GPU with count if homogeneous,
     #   otherwise None
     def homogenise_gpus(gpu_infos):
         if gpu_infos == []:
             warning('No graphics cards detected')
             return None
 
+        for gpu_info in gpu_infos:
+            gpu_info.pop('physicalid', None)
```
Leitevmd, line 247: Different physical IDs were resulting in None info of GPUs?

devito/arch/archinfo.py
```diff
@@ -234,7 +257,20 @@ def homogenise_gpus(gpu_infos):
     # Obtain textual gpu info and delegate parsing to helper functions
 
     try:
-        # First try is with detailed command lshw
+        # First try is with nvidia-smi, which can have more info about NVIDIA cards
```
Leitevmd, line 260: Put here because if nvidia-smi is present, we have greater chances of get proper information. Depending on how the driver was installed, lshw and lspci can't retrieve product information.

devito/arch/archinfo.py
FabioLuporini: Can you add a comment?

devito/arch/archinfo.py
FabioLuporini: yeah, I suppose if we have an AMD card instead this check will just quickly fail, and the execution will proceed smoothly with `lshw` right?

devito/arch/archinfo.py
```diff
@@ -216,13 +216,36 @@ def lspci_gpu_info(text):
                 gpu_infos.append(gpu_info)
         return filter_real_gpus(gpu_infos)
 
+    def nvidiasmi_gpu_info(text):
```
mloubout, line 219: Why not use the nvidia-smi python package? 
https://pypi.org/project/nvidia-ml-py3/

May be overkill I guess but may be safer.

FabioLuporini: but a compulsory dependence? 

I'd say no, and then what if one runs w/o it? do we ask him to install it raising a warning?

georgebisbas: Can `nvaccelinfo` do the work as well ?

devito/arch/archinfo.py
Leitevmd: Yes

devito/arch/archinfo.py
Leitevmd: Ok

devito/arch/archinfo.py
Leitevmd: Yes, but it requires loading SDK environment

Leitevmd:
I'm going to update this patch for using `nvidia-smi -L` flag from which we can get more reliable info. Using only `nvidia-smi` can retrieve incomplete product names.

devito/arch/archinfo.py
FabioLuporini: let's stick to the current approach. My workstation has nvidia-smi from 5 years ago and the output is in the same format as today's version. We'll revisit and consider adding that optional dependency in the future if ever necessary

FabioLuporini (approved):

PR #1617, opened 2021-03-02
examples: Add viscoacoustic Born operator to 2nd sls equation

nogueirapeterson:
Hi Guys,

We developed the second-order SLS viscoacoustic Born operator. We hope that the Devito team appreciates the contribution.


examples/seismic/viscoacoustic/operators.py
```diff
@@ -117,7 +117,7 @@ def sls_1st_order(model, geometry, p, **kwargs):
         return [u_r, u_v, u_p]
 
 
-def sls_2nd_order(model, geometry, p, **kwargs):
+def sls_2nd_order(model, geometry, p, r=None, **kwargs):
```
EdCaunt, line 120: Maybe want to document r in docstring?

examples/seismic/viscoacoustic/operators.py
```diff
@@ -543,6 +546,101 @@ def AdjointOperator(model, geometry, space_order=4, kernel='sls', time_order=2,
...
+        receivers (SparseTimeFunction) and their position.
+    space_order : int, optional
+        Space discretization order.
+    kernel : str, optional
+        Type of discretization, centered or shifted.
+    """
+    # Create wavefields and a dm field
+    p = TimeFunction(name='p', grid=model.grid, save=None,
+                     time_order=time_order, space_order=space_order, staggered=NODE)
+    P = TimeFunction(name='P', grid=model.grid, save=None,
```
EdCaunt, line 615: Maybe a slightly different naming could be more clear here? The `p` vs `P` is probably not ideal imo

examples/seismic/viscoacoustic/wavesolver.py
```diff
@@ -204,3 +222,136 @@ def adjoint(self, rec, srca=None, va=None, pa=None, vp=None, qp=None, b=None, r=
...
+            r = TimeFunction(name="r", grid=self.model.grid, time_order=self.time_order,
+                             space_order=self.space_order, staggered=NODE,
+                             save=self.geometry.nt)
+
+            summary = self.op_grad().apply(rec=rec, grad=grad, pa=pa, p=p, vp=vp,
+                                           r=r, qp=qp, b=b, dt=dt, **kwargs)
+
+        return grad, summary
+
+    def jacobian(self, dmin, src=None, rec=None, p=None, P=None, vp=None, qp=None,
```
EdCaunt, line 301: Again, maybe rethink the `p` vs `P` naming. Maybe `p` and `p_lin` or similar

EdCaunt (commented):
Looks good, just a few comments on docstrings and variable names

examples/seismic/viscoacoustic/operators.py
nogueirapeterson: Thanks for the review.
We did it that way because we were based on the acoustic example. There ( in `operator.py`), `u` and `U` are used to represent the background and scattered wavefields, respectively.
The difference is that in our case we are using `p` and `P`.

tests/test_adjoint.py
```diff
@@ -155,12 +161,19 @@ def test_adjoint_J(self, mkey, shape, kernel, space_order, setup_func):
         # Compute initial born perturbation from m - m0
         dm = (solver.model.vp.data**(-2) - model0.vp.data**(-2))
 
-        du = solver.jacobian(dm, vp=model0.vp)[0]
+        if setup_func is viscoacoustic_setup:
+            du = solver.jacobian(dm, vp=model0.vp, qp=model0.qp, b=model0.b)[0]
```
FabioLuporini, line 165: @mloubout wonder whether would be useful to extend `.jacobian` (and its siblings) to optionally accept the `model` directly so that they can autonomously and transparently extract all the kwargs they need

... rather than having this sort of branching in user or test code

@nogueirapeterson I'm interested in your thoughts as well

mloubout: Yes definitely, I know we talked about it at some point just was low priority so wasn't done.

nogueirapeterson: > @nogueirapeterson I'm interested in your thoughts as well

Yes, I was discussing this with @lauerami and we agreed with you. However, to keep the code standardized, this extension is also required for other models (for example acoustic, TTI...). What is your suggestion?

mloubout: An extension/rework of
https://github.com/devitocodes/devito/blob/1e619c0208ecf5f3817d614cd49b8df89d0beb22/examples/seismic/tti/wavesolver.py#L271 

should generalize to any model I think

lauerami: So, according to your suggestion, I have understood that the extension would be as follows (eg. viscoacoustic wavesolver):

```
def jacobian_adjoint(self, rec, p, pa=None, grad=None, vp=None, qp=None, b=None,
                         v=None, va=None, r=None, checkpointing=False,
                         model=None, **kwargs):
        model = model or self.model
        # Pick vp from model unless explicitly provided
        kwargs.update(model.physical_params(
            vp=vp, qp=qp, b=b)
        )
```
Thus, should we extend it for all `class ...WaveSolver(object):`?

mloubout: I'd say something like 
```
def jacobian_adjoint(self, rec, p, pa=None, grad=None, vp=None, qp=None, b=None,
                         v=None, va=None, r=None, checkpointing=False,
                         model=None, **kwargs):
        # Pick vp from model unless explicitly provided
        kwargs.update(model.physical_params(model=model))
```

lauerami: So, `qp`, `b`, `vp` will not be input parameters or should have a conditional `if model is none`?

I would like to know about it because in `test_gradient.py::TestGradient::test_gradientJ` 

```
for i in range(0, 7):
            # Add the perturbation to the model
            def initializer(data):
                data[:] = np.sqrt(v0.data**2 * v**2 /
                                  ((1 - H[i]) * v**2 + H[i] * v0.data**2))
            vloc = Function(name='vloc', grid=wave.model.grid, space_order=space_order,
                            initializer=initializer)
            # Data for the new model
            d = wave.forward(vp=vloc)[0]
            delta_d = (d.data - rec.data).reshape(-1)
            # First order error F(m0 + hdm) - F(m0)
            error1[i] = np.linalg.norm(delta_d, 1)
            # Second order term F(m0 + hdm) - F(m0) - J dm
            error2[i] = np.linalg.norm(delta_d - H[i] * Jdm.data.reshape(-1), 1)
```
where,

In this case, `vp` is overwritten inside the loop. Then, with this change, should we need to instantiate a `model`?

mloubout: Could have both as input yes or as part of `kwargs` with something like `kwargs.update(model.physical_params(model=model, **kwargs))`

lauerami: Hi @mloubout 

We already did the modifications suggested by you. However, we would like to know your opinion regarding docstrings. Taking TTI example:

```
def forward(self, src=None, rec=None, u=None, v=None, model=None,
                save=False, kernel='centered', **kwargs):
        """
        Forward modelling function that creates the necessary
        data objects for running a forward modelling operator.

        Parameters
        ----------
        geometry : AcquisitionGeometry
            Geometry object that contains the source (SparseTimeFunction) and
            receivers (SparseTimeFunction) and their position.
        u : TimeFunction, optional
            The computed wavefield first component.
        v : TimeFunction, optional
            The computed wavefield second component.
        model : Model, optional
            Object containing the physical parameters.
        vp : Function or float, optional
            The time-constant velocity.
        epsilon : Function or float, optional
            The time-constant first Thomsen parameter.
        delta : Function or float, optional
            The time-constant second Thomsen parameter.
        theta : Function or float, optional
            The time-constant Dip angle (radians).
        phi : Function or float, optional
            The time-constant Azimuth angle (radians).
        save : bool, optional
            Whether or not to save the entire (unrolled) wavefield.
        kernel : str, optional
            Type of discretization, centered or shifted.
```

Should we remove the input parameters `vp`, `epsilon`, `delta`, `theta` and `phi` from docstrings? Do you have any suggestions?

examples/seismic/acoustic/wavesolver.py
```diff
@@ -105,15 +105,16 @@ def forward(self, src=None, rec=None, u=None, vp=None, save=None, **kwargs):
                               save=self.geometry.nt if save else None,
                               time_order=2, space_order=self.space_order)
 
+        model = model or self.model
         # Pick vp from model unless explicitly provided
-        vp = vp or self.model.vp
+        kwargs.update(model.physical_params(model=model, **kwargs))
```
FabioLuporini, line 110: why do you also need `model=model` ?

examples/seismic/viscoacoustic/operators.py
```diff
@@ -127,6 +127,8 @@ def sls_2nd_order(model, geometry, p, **kwargs):
     ----------
     p : TimeFunction
         Pressure field.
+    r : TimeFunction
+        Memory variable.
```
FabioLuporini, line 131: "Memory"?

examples/seismic/viscoacoustic/operators.py
```diff
@@ -163,7 +167,8 @@ def sls_2nd_order(model, geometry, p, **kwargs):
         u_r = Eq(r.forward, damp * pde_r)
 
         pde_p = 2. * p - damp * p.backward + s * s * bm * (1. + tt) * \
-            div(b * grad(p, shift=.5), shift=-.5) - s * s * vp * vp * r.forward
+            div(b * grad(p, shift=.5), shift=-.5) - s * s * vp * vp * \
```
FabioLuporini, line 170: also (nitpicking) `... - s**2 * vp**2 * r.forward + s**2 * vp**2 * q`

examples/seismic/viscoacoustic/operators.py
```diff
@@ -543,6 +548,101 @@ def AdjointOperator(model, geometry, space_order=4, kernel='sls', time_order=2,
...
+        1st order - Blanch and Symes (1995) / Dutta and Schuster (2014)
+        viscoacoustic equation
+        2nd order - Bai et al. (2014) viscoacoustic equation
+        ren - Ren et al. (2014) viscoacoustic equation
+        deng_mcmechan - Deng and McMechan (2007) viscoacoustic equation
+        Defaults to sls 2nd order.
+    """
+    # Gradient symbol and wavefield symbols
+    grad = Function(name='grad', grid=model.grid)
+    p = TimeFunction(name='p', grid=model.grid, save=geometry.nt if save
```
FabioLuporini, line 578: Cuter (nitpicking) if `save=geometry.nt if save else None` goes on its own dedicated line

examples/seismic/viscoacoustic/operators.py
```diff
@@ -543,6 +548,101 @@ def AdjointOperator(model, geometry, space_order=4, kernel='sls', time_order=2,
...
+                     else None, time_order=2, space_order=space_order,
+                     staggered=NODE)
+    pa = TimeFunction(name='pa', grid=model.grid, save=None,
```
FabioLuporini, line 581: nitpicking: save=None unnecessary (default value)

FabioLuporini (commented):
Looks great, the latest changes are a considerable improvement to our examples IMHO

examples/seismic/viscoacoustic/operators.py
nogueirapeterson: yes, this is just the name of a variable in the PDE. 
There is no relationship with the computational part.

examples/seismic/acoustic/wavesolver.py
nogueirapeterson: yes, no need. We already fixed it.

examples/seismic/tti/wavesolver.py
```diff
@@ -171,16 +160,8 @@ def adjoint(self, rec, srca=None, p=None, r=None, vp=None,
             The computed wavefield first component.
         r : TimeFunction, optional
             The computed wavefield second component.
-        vp : Function or float, optional
-            The time-constant velocity.
-        epsilon : Function or float, optional
```
mloubout, line 176: Can we leave these in the docstring so that it's easier to know what inputs can be expected.

examples/seismic/viscoacoustic/operators.py
```diff
@@ -264,15 +270,15 @@ def ren_2nd_order(model, geometry, p, **kwargs):
     # Density
     rho = 1. / b
 
-    eta = (vp * vp) / (w0 * qp)
+    eta = (vp**2) / (w0 * qp)
```
mloubout, line 273: nipicking, vp**2 doesn't need parenthesis.

mloubout (commented):
Looks fine to me, what's the stopper?

examples/seismic/viscoacoustic/operators.py
```diff
@@ -151,19 +156,20 @@ def sls_2nd_order(model, geometry, p, **kwargs):
     rho = 1. / b
 
     # Bulk modulus
-    bm = rho * vp * vp
+    bm = rho * vp**2
 
-    r = TimeFunction(name="r", grid=model.grid, time_order=2, space_order=space_order,
-                     save=geometry.nt if save else None, staggered=NODE)
+    r = r or TimeFunction(name="r", grid=model.grid, time_order=2,
```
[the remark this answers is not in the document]
mloubout: Hum ok

mloubout (approved):
Needs rebase then GTG

FabioLuporini (approved):

FabioLuporini:
Great, merging!

PR #1619, opened 2021-03-04
dsl: patch for custom coefficients with staggered equations

### LINKED ISSUE #1589: Symbolic coefficients on staggered stencils not working correctly

EdCaunt:
Fixed #1589. Allows for custom coefficients to be set on equations with staggered functions.

### DIFF
--- a/devito/finite_differences/coefficients.py
+++ b/devito/finite_differences/coefficients.py
@@ -187,7 +187,6 @@ class Substitutions(object):
             dim = i.dimension
             index = i.index
             weights = i.weights
-            print(deriv_order, function, dim, index, weights)
 
             if isinstance(weights, np.ndarray):
                 fd_order = len(weights)-1
@@ -286,11 +285,8 @@ def default_rules(obj, functions):
         return subs
 
     # Determine which 'rules' are missing
-    print("Functions", functions)
     sym = get_sym(functions)
-    print("Sym", sym)
     terms = obj.find(sym)
-    print("Terms", terms)
     args_present = filter_ordered(term.args[1:] for term in terms)
 
     subs = obj.substitutions
@@ -303,8 +299,6 @@ def default_rules(obj, functions):
     # NOTE: Do we want to throw a warning if the same arg has
     # been provided twice?
     args_provided = list(set(args_provided))
-    print("Args present", args_present)
-    print("Args provided", args_provided)
 
     # Need to compare dimensions,, not indices, then take the index from
     # args_present to pass to generate_subs()
--- a/devito/finite_differences/coefficients.py
+++ b/devito/finite_differences/coefficients.py
@@ -187,6 +187,7 @@ class Substitutions(object):
             dim = i.dimension
             index = i.index
             weights = i.weights
+            print(deriv_order, function, dim, index, weights)
 
             if isinstance(weights, np.ndarray):
                 fd_order = len(weights)-1
@@ -203,6 +204,7 @@ class Substitutions(object):
             # passed as a dictionary of the form {pos: w} (or something similar).
             if isinstance(weights, np.ndarray):
                 for j in range(len(weights)):
+                    # Index here is incorrect. Need to pull it from elsewhere
                     subs.update({function._coeff_symbol
                                  (indices[j], deriv_order,
                                   function, index): weights[j]})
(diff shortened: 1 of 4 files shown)


devito/finite_differences/coefficients.py
```diff
@@ -224,6 +224,33 @@ def generate_subs(i):
         rules = {}
         for i in self.coefficients:
             rules.update(generate_subs(i))
+        return rules
+
+    def update_rules(self, obj):
```
FabioLuporini, line 229: I don't like the word "update" because it makes one think you're changing the internal state of the object but (thanks God) you aren't

devito/types/equation.py
```diff
@@ -97,7 +97,9 @@ def evaluate(self):
             # all rules to be expunged during this procress.
             rules = default_rules(eq, eq._symbolic_functions)
             try:
-                eq = eq.xreplace({**eq.substitutions.rules, **rules})
+                eq = eq.xreplace(rules)
+                custom_rules = eq.substitutions.update_rules(eq)
+                eq = eq.xreplace(custom_rules)
```
FabioLuporini, line 102: I don't like two xreplace back to back. I'm confident there's a way of working around this overhead (i.e., just one replace)

devito/finite_differences/coefficients.py
```diff
@@ -224,6 +224,33 @@ def generate_subs(i):
...
+        """Update the specified rules to reflect staggering in an equation"""
```
FabioLuporini, line 230: format for docstrings in devito:
```
"""
<line>.
"""
```
So each docstring takes at least three lines, and it always terminates with a full stop

devito/finite_differences/coefficients.py
```diff
@@ -224,6 +224,33 @@ def generate_subs(i):
...
+        # Determine which 'rules' are expected
+        sym = get_sym(self._function_list)
+        terms = obj.find(sym)
+        args_expected = filter_ordered(term.args[1:] for term in terms)
+        args_expected_dim = [(arg[0], arg[1], retrieve_dimensions(arg[2])[0])
```
FabioLuporini, line 235: you're often using `arg[0], arg[1], ...` but what are these objects? I almost never use explicit indexing into tuples/lists throughout devito. I think you're after a `namedtuple` here: https://docs.python.org/3/library/collections.html#collections.namedtuple (if yes, `git grep namedtuple` for examples oh how it's used in devito)

FabioLuporini:
ah, one more comment @EdCaunt : these commits:

* dsl: docstring tweaks
* dsl: minor tidy up

can and should be squashed away via `git rebase`

devito/finite_differences/coefficients.py
EdCaunt: Any suggestions for a replacement? 'bend_rules', 'modify_rules', 'adapt_rules'? I personally prefer the latter

rhodrin (commented):
Is [this](https://github.com/devitocodes/devito/blob/fe1bc971eb38eeddef80644f4559a8ff17eb597d/devito/finite_differences/coefficients.py#L221) still true?

EdCaunt:
Should still be true. The change I made makes it modify this base set of rules to match the equation if needs be.

devito/finite_differences/coefficients.py
rhodrin, line 229: I don't think this process should be happening here. This is the job of the `Coefficient` object. This is where any staggering etc should be being taken into account and hence allowing this logic to be abstracted away from the `Substitutions` object.

devito/types/equation.py
rhodrin: I don't think `equation.py` should need any changes at all. Rules can, and should, be structured such that `eq = eq.xreplace({**eq.substitutions.rules, **rules})` still deals with everything.

devito/finite_differences/coefficients.py
```diff
@@ -224,6 +224,33 @@ def generate_subs(i):
         rules = {}
         for i in self.coefficients:
             rules.update(generate_subs(i))
+        return rules
+
+    def update_rules(self, obj):
+        """Update the specified rules to reflect staggering in an equation"""
+        # Determine which 'rules' are expected
+        sym = get_sym(self._function_list)
```
rhodrin, line 232: Sorry, which symbol is this getting again? Do functions with symbolic coefficients not already have a property for this?

devito/finite_differences/coefficients.py
rhodrin: What are `arg[0]` etc. exactly? Are they not already available from some object we have access to?

FabioLuporini:
closing as superseded by #1636

PR #1620, opened 2021-03-05
compiler: Add skewing pass towards Temporal Blocking

georgebisbas:
This is the first step towards Temporal Blocking automation.

This PR aims to add skewing as a pass to:
- change iteration bounds
- skew accesses

A new Queue subclass is implemented.

- `skewinner` option is added to skew or not the innermost loop

No changes are happening in the computation order. Nothing is optimized.
Codegen tests have been added, covering simple time-stepping computations, autotuning, and subdomains.


.github/workflows/examples.yml
```diff
@@ -79,6 +79,7 @@ jobs:
       shell: bash -l {0}
       run: |
         python examples/misc/linalg.py mat-vec mat-mat-sum transpose-mat-vec
+        python examples/misc/linalg.py mat-vec -o mat-mat -o transpose-mat-vec -o
```
FabioLuporini, line 82: why?

devito/passes/clusters/temporal.py
```diff
@@ -0,0 +1,78 @@
+from devito.passes.clusters.utils import cluster_pass
```
FabioLuporini, line 1: This comment is about the file, not this specific line (I didn't know where to put it)

I don't like the name `temporal`. Clearly I know where it comes from, but I like to stick to the idea that time is no special in Devito. What makes time so special? Let's see if we can devise a better name from answering this question

devito/core/cpu.py
```diff
@@ -90,6 +90,8 @@ def _normalize_kwargs(cls, **kwargs):
         o['blockinner'] = oo.pop('blockinner', False)
         o['blocklevels'] = oo.pop('blocklevels', cls.BLOCK_LEVELS)
 
+        o['skewing'] = oo.pop('skewing', False)
```
FabioLuporini, line 93: Each though it will be a different pass, I _think_ you can put this option right below "blocklevels" (so no blank line) because essentially "skewing" is made for "blocking" (and nothing else, right?)

devito/passes/clusters/temporal.py
```diff
@@ -0,0 +1,78 @@
...
+from devito.ir.support import IterationSpace, SEQUENTIAL
+from devito.ir.support.space import Interval
```
FabioLuporini, line 3: Squash with previous line?

devito/passes/clusters/temporal.py
```diff
@@ -0,0 +1,78 @@
...
+from devito.symbolics import xreplace_indices
+from devito.logger import warning
```
FabioLuporini, line 5: Try to honour alphabetical ordering of imports

devito/passes/clusters/temporal.py
```diff
@@ -0,0 +1,78 @@
...
+    for i = 2, n-1
+        for j = 2+i, m-1+i
+            a[i,j-i] = (a[a-1,j-i] + a[i,j-1-i] + a[i+1,j-i] + a[i,j+1-i]) / 4
+        end for
+    end for
+    """
+    processed = []
+    # What dimensions do we target?
+    # a) SEQUENTIAL Dimensions
+    skew_dims = []
```
FabioLuporini, line 36: use a list comprehension, not four lines. That's the pythonic way

devito/passes/clusters/temporal.py
```diff
@@ -0,0 +1,78 @@
...
+    for i in cluster.ispace:
+        if SEQUENTIAL in cluster.properties[i.dim]:
+            skew_dims.append(i.dim)
+
+    skewable = {i.dim for i in cluster.ispace.intervals
```
FabioLuporini, line 41: This is to rule out already blocked loops? or what? remind us (add comment?)

devito/passes/clusters/temporal.py
```diff
@@ -0,0 +1,78 @@
...
+                if not i.dim.symbolic_incr.is_Symbol}
+
+    # Remove candidates
+    mapper, intervals = {}, []
+
+    try:
+        if len(skew_dims) > 1:
```
FabioLuporini, line 48: this `if` should probably go right after the computation of `skew_dims`

devito/passes/clusters/temporal.py
```diff
@@ -0,0 +1,78 @@
...
+            raise warning("More than 1 dimensions that can be skewed.\
+                          Skewing the first one")
+        skew_dim = skew_dims[0]  # Skew first one
+        intervals.append(Interval(skew_dim, 0, 0))
+        index = intervals.index(Interval(skew_dim, 0, 0))
+    except BaseException:
```
FabioLuporini, line 54: Too vague

devito/passes/clusters/temporal.py
```diff
@@ -0,0 +1,78 @@
...
+    # Remove candidates
+    mapper, intervals = {}, []
+
+    try:
+        if len(skew_dims) > 1:
+            raise warning("More than 1 dimensions that can be skewed.\
+                          Skewing the first one")
+        skew_dim = skew_dims[0]  # Skew first one
+        intervals.append(Interval(skew_dim, 0, 0))
+        index = intervals.index(Interval(skew_dim, 0, 0))
```
FabioLuporini, line 53: this is horrible. You probably wanted:

```
index = Interval(skew_dim, 0, 0)
...
```

but even then, why are you even assigning `index` here? this is getting overridden right after the try-except, not?

devito/passes/clusters/temporal.py
```diff
@@ -0,0 +1,78 @@
...
+        # No time dimensions -> nothing to do
+        return cluster
+
+    # Skew dim will not be none here:
+    # Initializing a default skewed dim index position in loop
+    index = 0
+
+    for i in cluster.ispace.intervals:
+        if i.dim not in skew_dims:
+            if index < cluster.ispace.intervals.index(i) and i.dim in skewable:
```
FabioLuporini, line 64: `skew_dims` and `skewable` should probably be fused. Just make sure that only the SEQUENTIAL ones are in skewable...

tests/test_skewing.py
```diff
@@ -0,0 +1,393 @@
...
+    op = Operator(Eq(u.forward, 0.5 * u.laplace + u), opt=opt)
+
+    blocksizes0 = get_blocksizes(op, opt, grid, blockshape0, 0)
+    blocksizes1 = get_blocksizes(op, opt, grid, blockshape1, 1)
+    op.apply(u=u, t=10, **blocksizes0, **blocksizes1)
+
+    return u.data[1, :], op
+
+
+class TestAutotuningWithSkewing(object):
```
FabioLuporini, line 80: No, too much code replication. Maybe what you actually want to do is parametrizing tests in test_autotuning?

tests/test_skewing.py
```diff
@@ -0,0 +1,393 @@
...
+
+        trees = retrieve_iteration_tree(op)
+        assert len(trees) == expected
+        op()
+
+        assert np.allclose(u.data[2, :, :], 3.0)
+        assert np.allclose(v.data[1, 1:-1, 1:-1], 1.0)
+
+
+class TestDleWithSkewing(object):
```
FabioLuporini, line 272: "Dle" is an archaic name

tests/test_skewing.py
FabioLuporini: and, if necessary, add a few ones there? autotuning is a very specific features, and anything concerning autotuning should go there I feel

tests/test_skewing.py
```diff
@@ -0,0 +1,393 @@
...
+    """
```
FabioLuporini, line 273: And same as before, maybe you want to parametrize the the already existing tests in test_dle, not copy-pasting code. 

I appreciate the fact that we said we were going to put everything in test_skewing, but if the price to pay is duplicated code (or nearly duplicated code), thanks but no?

devito/passes/clusters/temporal.py
```diff
@@ -0,0 +1,78 @@
...
+                skewable.remove(i.dim)
+            else:
+                intervals.append(Interval(i.dim, 0, 0))
+
+        processed = xreplace_indices(cluster.exprs, mapper)
+
+    ispace = IterationSpace(intervals, cluster.ispace.sub_iterators,
+                            cluster.ispace.directions)
+
+    rebuilt_cluster = cluster.rebuild(exprs=processed, ispace=ispace)
```
FabioLuporini, line 76: "rebuilt" is enough

devito/passes/clusters/temporal.py
```diff
@@ -0,0 +1,78 @@
...
+    skew_dims = []
+    for i in cluster.ispace:
+        if SEQUENTIAL in cluster.properties[i.dim]:
+            skew_dims.append(i.dim)
+
+    skewable = {i.dim for i in cluster.ispace.intervals
+                if not i.dim.symbolic_incr.is_Symbol}
+
+    # Remove candidates
+    mapper, intervals = {}, []
```
FabioLuporini, line 45: mapper should be defined later on?

FabioLuporini (changes requested):
Out of curiosity, have you considered using an implicit tree visitors for Clusters (ie a subclass of `Queue`) as for example done here https://github.com/devitocodes/devito/blob/master/devito/passes/clusters/blocking.py#L12

?

A few questions:
* Did you understand how a Queue works as for example in Blocking?
* Did you think it's not necessary here? As in, do you think that working on individual clusters in isolation is the way to go? (Haven't really thought about it, honestly, for this specific pass)

The other key comment I have is the following: when you implement a pass (and, in general, any algorithm in your life), you need to think not once, not ten times, but better one hundred times about the complexity of your implementation. Useful questions to ask yourself to drive the implementation "are all these lists/dict/... really necessary?" , "can I get away with fewer data structures and/or loops?" , etc. For example here you have both `skew_dims` and `skewable` -- that doesn't make much sense to me. Also `remove` from `skewable` is hardly necessary... I'm fairly sure there's a neater way of achieving this. You also have two nested `if`s... first one is `if i.dim not in skew_dims` while the nested one is `if index < .... and i.dim in skewable` . I hardly believe that comparing indices and checking again for the presence of a dimension in `skewable` is the simplest way to go

devito/passes/clusters/temporal.py
georgebisbas: Needs to be dropped, right.

devito/passes/clusters/temporal.py
georgebisbas: I think the var names I have used here are not helping. Will change. I need:
```
    # What dimensions should we skew against?
    # e.g. SEQUENTIAL Dimensions (Usually time in FD solvers, but not only limited
    # to this)
    # What dimensions should be skewed?
    # e.g. PARALLEL dimensions (x, y, z) but not blocking ones (x_blk0, y_blk0)
```
skew_dims .... like time...
skewable... like space...

devito/passes/clusters/temporal.py
georgebisbas: ok

devito/passes/clusters/temporal.py
georgebisbas: ok

devito/passes/clusters/temporal.py
georgebisbas: Yes, I updated the comments.

devito/passes/clusters/temporal.py
georgebisbas: True

devito/passes/clusters/temporal.py
georgebisbas: Yes

devito/passes/clusters/temporal.py
georgebisbas: dropped, redundant

tests/test_skewing.py
georgebisbas: Sure, I agree, I was aware of that. I just added a lot of duplicate/redundant test code for this draft state.
Will fix accordingly DSE/DLE insprired ones

tests/test_skewing.py
georgebisbas: Absolutely agree, I can get rid of the duplicate code.

devito/core/cpu.py
```diff
@@ -322,7 +323,8 @@ def _make_clusters_passes_mapper(cls, **kwargs):
             'cire-divs': lambda i: cire(i, 'divs', sregistry, options, platform),
             'cse': lambda i: cse(i, sregistry),
             'opt-pows': optimize_pows,
-            'topofuse': lambda i: fuse(i, toposort=True)
+            'topofuse': lambda i: fuse(i, toposort=True),
+            'skewing': skewing
```
mloubout, line 327: So probably not a fun question, but is that the "best" order? I.e does applying skewing earlier in the passes change the output?

devito/passes/clusters/temporal.py
mloubout: I think skewing is dangerous, just because of the DSL side (skewing can mean symmetrizing too). But depending on the complexity (this seem short) I would think it just belongs to blocking.

devito/passes/clusters/temporal.py
```diff
@@ -0,0 +1,82 @@
...
+    skew_dims, skewable = [], []
+
+    for i in cluster.ispace:
+        if SEQUENTIAL in cluster.properties[i.dim]:
+            skew_dims.append(i.dim)
+        elif PARALLEL in cluster.properties[i.dim] and not i.dim.symbolic_incr.is_Symbol:
+            skewable.append(i.dim)
+
+    if len(skew_dims) > 1:
+        raise warning("More than 1 dimensions that can be skewed.\
```
mloubout, line 52: No nested skewing? :D

devito/passes/clusters/temporal.py
```diff
@@ -0,0 +1,82 @@
...
+                      Skewing the first in the list")
+    elif len(skew_dims) == 0:
+        # No dimensions to skew against -> nothing to do, return
+        return cluster
+
+    # Skew dim will not be none here:
+    # Initializing a default skewed dim index position in loop
+    skew_index = 0
+    skew_dim = skew_dims[skew_index]  # Skew first one
```
mloubout, line 61: lenght supposed to be one so just `skew_dim = skew_dims.pop()`

devito/passes/clusters/temporal.py
```diff
@@ -0,0 +1,82 @@
...
+        if skew_index < cluster.ispace.intervals.index(i) and i.dim in skewable:
+            mapper[i.dim] = i.dim - skew_dim
+            intervals.append(Interval(i.dim, skew_dim, skew_dim))
+            skewable.remove(i.dim)
+        # Do not touch otherwise
+        else:
+            intervals.append(i)
+
+        processed = xreplace_indices(cluster.exprs, mapper)
+
```
mloubout, line 76: assert skewable empty

devito/passes/clusters/temporal.py
```diff
@@ -0,0 +1,82 @@
...
+    # Skew dim will not be none here:
+    # Initializing a default skewed dim index position in loop
+    skew_index = 0
+    skew_dim = skew_dims[skew_index]  # Skew first one
+
+    mapper, intervals, processed = {}, [], []
+
+    for i in cluster.ispace.intervals:
+        # Skew dim if nested under time i.dim not in skew_dims:
+        if skew_index < cluster.ispace.intervals.index(i) and i.dim in skewable:
```
mloubout, line 67: explicit `>0` since `skew_index` can't be anything else?

mloubout: Actually what happens if you have an overarching subdomain `n` before the time dimension?

devito/passes/clusters/temporal.py
```diff
@@ -0,0 +1,82 @@
...
+            mapper[i.dim] = i.dim - skew_dim
+            intervals.append(Interval(i.dim, skew_dim, skew_dim))
+            skewable.remove(i.dim)
```
mloubout, line 70: I think fabio usually use a queue and `pop` rather than remove.

tests/test_skewing.py
```diff
@@ -0,0 +1,393 @@
+import pytest
+from functools import reduce
+from operator import mul
+
+
```
mloubout, line 5: single blank line

tests/test_skewing.py
```diff
@@ -0,0 +1,393 @@
...
+from sympy import Add, cos, sin, sqrt  # noqa
+import numpy as np
+
+from devito import Grid, TimeFunction, Eq, Operator, configuration, switchconfig, Le, Lt, Ge, Gt  # noqa
+
```
mloubout, line 10: no blank line

tests/test_skewing.py
```diff
@@ -0,0 +1,393 @@
...
+from devito.core.autotuning import options  # noqa
+from devito import (NODE, Eq, Inc, Constant, Function, TimeFunction, SparseTimeFunction,  # noqa
```
mloubout, line 12: two `from devito ` fix imports

devito/core/cpu.py
georgebisbas: Before submitting this PR, I incorporated the pass in the advanced mode, looking at two cases
a) applied, just after blocking
b) applied as a last pass

AFAIR there was no diff in the resulting operators' OI. Will need to pay more attention though.Since the only parameter that may affect skewing from a kernel point of view is the space order, I think it will make no diff.

As a said, will pay more attention later on this.

devito/passes/clusters/temporal.py
georgebisbas: Will need tests to add for that, though in my mind I just need to ensure that we skew only loops that are nested inside the time_loop, without care for outer loops, hence `>0 `

georgebisbas: AHHH hang on I see the issue now! Thanks! Will fix it! (I thought it was already solved!)

devito/passes/clusters/temporal.py
georgebisbas: Sure.

devito/passes/clusters/temporal.py
georgebisbas: Maybe in the future should be sub-pass like addition in blocking? (This is what you mean ?)

mloubout: No I mean skewing probably belings to `blocking.py` rather than its own file

devito/passes/clusters/temporal.py
mloubout: warning or error?

devito/passes/clusters/temporal.py
FabioLuporini: Now that Mathias brings this up, I think I agree with him.  A separate file smells sounds like an overkill

devito/passes/clusters/temporal.py
```diff
@@ -0,0 +1,84 @@
...
+    # e.g. PARALLEL dimensions (x, y, z) but not blocking ones (x_blk0, y_blk0)
+    # Iterate over the iteration space and assign dimensions to skew or skewable
+    # depending on their properties
+
+    skew_dims, skewable = [], []
+
+    for i in cluster.ispace:
+        if SEQUENTIAL in cluster.properties[i.dim]:
+            skew_dims.append(i.dim)
+        elif PARALLEL in cluster.properties[i.dim] and not i.dim.symbolic_incr.is_Symbol:
```
FabioLuporini, line 48: remind me what's wrong with skewing IncrDimensions ? Would it lead to generating broken code?

devito/passes/clusters/temporal.py
FabioLuporini: > No nested skewing? :D

why not?

devito/passes/clusters/temporal.py
FabioLuporini: btw, 

```
not i.dim.symbolic_incr.is_Symbol
```

a less ugly way is:

```
not i.dim.is_Incr
```

devito/passes/clusters/temporal.py
FabioLuporini: I agree with @mloubout we should add a test involving subdomains

I still think using indices here is the symptom of a hack, or at least something that could be achieved more easily in another way

For example, instead of precomputing `skew_dims` and `skewable`, why don't you do everything greedily?
You could start iterating over the intervals (don't know you need the `.intervals` in `for i in cluster.ispace.intervals` btw) and keep a flag that essentially tells you if you've found your skewing dimension yet or not. Do you see what I mean?

devito/passes/clusters/temporal.py
FabioLuporini: and on second thought, why would you need to `remove` (or `pop`) ?

devito/passes/clusters/temporal.py
```diff
@@ -0,0 +1,84 @@
...
+
+    # Skew dim will not be none here:
+    # Initializing a default skewed dim index position in loop
+    skew_index = 0
+    skew_dim = skew_dims.pop()  # Skew first one
+
+    mapper, intervals, processed = {}, [], []
+
+    for i in cluster.ispace.intervals:
+        # Skew dim if nested under time i.dim not in skew_dims:
```
FabioLuporini, line 66: what is time?

tests/test_ir.py
```diff
@@ -549,6 +550,7 @@ def test_index_mode_detection(self, indexed, expected):
         """
         grid = Grid(shape=(4, 4, 4))
         x, y, z = grid.dimensions  # noqa
+        time = grid.time_dim  # noqa
```
FabioLuporini, line 553: why is time so important in this test?

tests/test_skewing.py
FabioLuporini: OK, poke me once done

FabioLuporini (changes requested):

devito/passes/clusters/temporal.py
georgebisbas: Fixed

devito/passes/clusters/temporal.py
georgebisbas: This is not working? They are all is_Incr..

devito/passes/clusters/blocking.py
```diff
@@ -157,3 +160,102 @@ def decompose(ispace, d, block_dims):
     directions.update({bd: ispace.directions[d] for bd in block_dims})
 
     return IterationSpace(intervals, sub_iterators, directions)
+
+
+class Skewing(Queue):
+
+    def __init__(self, options):
```
FabioLuporini, line 167: why?

devito/passes/clusters/blocking.py
```diff
@@ -157,3 +160,102 @@ def decompose(ispace, d, block_dims):
...
+
+        super(Skewing, self).__init__()
+
+    @timed_pass(name='skewing')
+    def process(self, clusters):
```
FabioLuporini, line 172: why?

devito/passes/clusters/blocking.py
```diff
@@ -157,3 +160,102 @@ def decompose(ispace, d, block_dims):
...
+        return super(Skewing, self).process(clusters)
+
+    def _process_fdta(self, clusters, level, prefix=None):
```
FabioLuporini, line 175: this is not how object oriented programming is supposed to work

devito/passes/clusters/blocking.py
```diff
@@ -157,3 +160,102 @@ def decompose(ispace, d, block_dims):
...
+            else:
+                intervals.append(i)
+
+        processed = xreplace_indices(exprs, mapper)
+        ispace = IterationSpace(intervals, ispace.sub_iterators,
+                                ispace.directions)
+    else:
+        return ispace, exprs
+
+    return ispace, processed
```
FabioLuporini, line 261: move at the bottom of the `if` above -- would be clearer

devito/passes/clusters/blocking.py
```diff
@@ -157,3 +160,102 @@ def decompose(ispace, d, block_dims):
...
+            skew_flag = True
+            break
+
+    if skew_flag:
+        for i in ispace.intervals:
+            # Skew a dim if nested under skew_dim and is prefix:
+            if skew_pos < ispace.intervals.index(i) and i.dim == d:
+                mapper[i.dim] = i.dim - skew_dim
+                intervals.append(Interval(i.dim, skew_dim, skew_dim))
+            # Do not touch otherwise
```
FabioLuporini, line 251: move inside else?

FabioLuporini: probably unnecessary

devito/passes/clusters/blocking.py
```diff
@@ -157,3 +160,102 @@ def decompose(ispace, d, block_dims):
...
+        if SEQUENTIAL in properties[i.dim] and not i.dim.symbolic_incr.is_Symbol:
+            skew_dim = i.dim
+            skew_pos = n
+            skew_flag = True
+            break
+
+    if skew_flag:
+        for i in ispace.intervals:
+            # Skew a dim if nested under skew_dim and is prefix:
+            if skew_pos < ispace.intervals.index(i) and i.dim == d:
```
FabioLuporini, line 248: `i.dim is d` will be enough (do you appreciate the difference?)

devito/passes/clusters/blocking.py
```diff
@@ -157,3 +160,102 @@ def decompose(ispace, d, block_dims):
...
+
+    # Skew dim will not be none here:
+    # Initializing a default skewed dim index position in loop
+
+    mapper, intervals, processed = {}, [], []
+
+    # Identify a skew_dim and its position
+    skew_flag = False
+    for n, i in enumerate(ispace):
+        if SEQUENTIAL in properties[i.dim] and not i.dim.symbolic_incr.is_Symbol:
```
FabioLuporini, line 239:
```
... and not i.dim.is_Incr
```
?

this should be fine -- eventually the compilation pipeline will be such that skewing is applied before blocking, so we should be good?

tests/test_autotuner.py
```diff
@@ -123,9 +123,29 @@ def test_mixed_blocking_nthreads():
...
 
-def test_tti_aggressive():
+@pytest.mark.parametrize('openmp, expected', [
+    (False, 2), (True, 3)
+])
+def test_mixed_blocking_w_skewing(openmp, expected):
+    grid = Grid(shape=(96, 96, 96))
+    f = TimeFunction(name='f', grid=grid)
+
+    op = Operator(Eq(f.forward, f + 1.), opt=('blocking', 'skewing',
```
FabioLuporini, line 133: I'd imagine skewing gets applied before blocking? OK, both ways could work, but logically I think of skewing as a preliminary transformation?

devito/passes/clusters/blocking.py
FabioLuporini: also, what happens if you have many nested SEQUENTIAL Dimensions?

devito/passes/clusters/blocking.py
```diff
@@ -157,3 +160,102 @@ def decompose(ispace, d, block_dims):
...
+    # Iterate over the iteration space and assign dimensions to skew or skewable
+    # depending on their properties
+
+    # Skew dim will not be none here:
+    # Initializing a default skewed dim index position in loop
+
+    mapper, intervals, processed = {}, [], []
+
+    # Identify a skew_dim and its position
+    skew_flag = False
```
FabioLuporini, line 237: this piece of code is flawed in many ways. First of all, it will fail if no SEQUENTIAL Dimensions are detected, since skew_dim is only defined inside the loop body.

Then, you're using three variables to carry one piece of information...  that's such a useless complication

I _think_ you could also get away with a _single_ `for i in ispace` loop -- you only need to track the first SEQUENTIAL Dimension, then you skew the inner PARALLEL loops afterwards

devito/passes/clusters/blocking.py
```diff
@@ -157,3 +160,102 @@ def decompose(ispace, d, block_dims):
...
+    # to this)
+    # What dimensions should be skewed?
+    # e.g. PARALLEL dimensions (x, y, z) but not blocking ones (x_blk0, y_blk0)
+    # Iterate over the iteration space and assign dimensions to skew or skewable
+    # depending on their properties
+
+    # Skew dim will not be none here:
+    # Initializing a default skewed dim index position in loop
+
+    mapper, intervals, processed = {}, [], []
```
FabioLuporini, line 234: processed definition is useless

the other two should have been defined closer to their first use

devito/passes/clusters/blocking.py
```diff
@@ -157,3 +160,102 @@ def decompose(ispace, d, block_dims):
...
+    def callback(self, clusters, prefix):
+        if not prefix:
+            return clusters
+
+        d = prefix[-1].dim
+
+        processed = []
+        for c in clusters:
+            if PARALLEL in c.properties[d] and not d.symbolic_incr.is_Symbol:
+                ispace, exprs = skew(c.ispace, c.exprs, c.properties, d)
```
FabioLuporini, line 188: see my comments below first then come back here 


---


In light of what I wrote below, I think we don't need an extra `skew` function. Everything could be written directly here, thus making the implementation much simpler. Instead of traversing the **whole** iteration space (thus having to keep the `skew_pos` around) you could instead traverse the `prefix`, which terminates with `d`, so all Intervals in the prefix are definitely before `d`. Then you don't have to search for anything else... it's just  `mapper[d] = ...`

devito/core/cpu.py
```diff
@@ -316,6 +316,7 @@ def _make_clusters_passes_mapper(cls, **kwargs):
 
         return {
             'blocking': Blocking(options).process,
+            'skewing': Skewing(options).process,
```
FabioLuporini, line 319: Eventually, instead of exposing `Blocking(...)` to the pipelines, we could rather expose 

```
def blocking(...)
   ...
```

which calls Skewing (if requested) and Blocking back to back (plus any other routines to achieve temporal blocking?) 

so in the pipelines we'll have 

```
clusters = blocking(clusters)
```

This will make the things a bit more structured.

After all, skewing always goes hand in hand with blocking (when temporal blocking is requested)

FabioLuporini (changes requested):

devito/passes/clusters/blocking.py
georgebisbas: ok

tests/test_autotuner.py
georgebisbas: hmm.. answered above

devito/passes/clusters/blocking.py
georgebisbas: dropped

devito/passes/clusters/blocking.py
georgebisbas: dropped

devito/passes/clusters/blocking.py
georgebisbas: I think it is better now, according to your comments

devito/passes/clusters/blocking.py
```diff
@@ -157,3 +160,71 @@ def decompose(ispace, d, block_dims):
...
+                # Traverse cluster's iteration space
+                for n, i in enumerate(c.ispace):
+                    # Identify a skew_dim and its position
+                    if (SEQUENTIAL in c.properties[i.dim]
+                       and not i.dim.symbolic_incr.is_Symbol):
+                        skew_dim = i.dim
+                        intervals.append(i)
+                    # Since we are here, prefix is skewable and nested under a
+                    # SEQUENTIAL loop. Do not skew innermost loop
+                    elif (i.dim == d and not i.dim == c.ispace[-1].dim and
```
georgebisbas, line 213: Avoid skewing for innermost loop

devito/passes/clusters/temporal.py
georgebisbas: Fixed

devito/passes/clusters/temporal.py
georgebisbas: Discussed regarding doing this before or after blocking.

devito/passes/clusters/blocking.py
georgebisbas: Maybe not, add a note to fix

FabioLuporini:
```
i.dim == c.ispace[-1].dim
```

As I wrote above, you probably just want to skew the TILABLE ones, rather than using "complicated" conditions such as this one

devito/passes/clusters/blocking.py
```diff
@@ -157,3 +160,74 @@ def decompose(ispace, d, block_dims):
...
+                    # skewed in one cluster but not in other
+                    elif (i.dim == d and not i.dim == c.ispace[-1].dim and
+                          skew_dim is not None):
+                        mapper[i.dim] = i.dim - skew_dim
+                        intervals.append(Interval(i.dim, skew_dim, skew_dim))
+                    # End-up here if dimensions is neither skewable or skewed against
+                    else:
+                        intervals.append(i)
+
+                processed_exprs = xreplace_indices(c.exprs, mapper)
```
FabioLuporini, line 224: exprs = ... (code homogeneity)

devito/passes/clusters/blocking.py
```diff
@@ -157,3 +160,74 @@ def decompose(ispace, d, block_dims):
...
+                ispace = IterationSpace(intervals, c.ispace.sub_iterators,
+                                        c.ispace.directions)
+                processed_clusters.append(c.rebuild(exprs=processed_exprs, ispace=ispace,
```
FabioLuporini, line 227: processed.append(...)  (code homogeneity)

FabioLuporini (changes requested):

devito/core/gpu.py
```diff
@@ -7,7 +7,7 @@
 from devito.passes.equations import collect_derivatives, buffering
 from devito.passes.clusters import (Blocking, Lift, Streaming, Tasker, cire, cse,
                                     eliminate_arrays, extract_increments, factorize,
-                                    fuse, optimize_pows)
+                                    fuse, optimize_pows, Skewing)
```
FabioLuporini, line 10: nitpicking: Skewing before Streaming (we do our best to keep alphabetical ordering)

devito/core/gpu.py
```diff
@@ -69,6 +69,7 @@ def _normalize_kwargs(cls, **kwargs):
         # Blocking
         o['blockinner'] = oo.pop('blockinner', True)
         o['blocklevels'] = oo.pop('blocklevels', cls.BLOCK_LEVELS)
+        o['skewing'] = oo.pop('skewing', False)
```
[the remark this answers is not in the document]
FabioLuporini: that's a 1986 article, so I'm not buying that today there's any value in doing skewing+interchange only (especially with devito apps in mind), unless you now come back with (compelling) evidence

also, this would break our assumption that we see skewing as a subpass of blocking?

IMHO:

* if we see it as a subpass of blocking, then it makes sense to keep everything in blocking.py, and it also makes sense to call it blockskew -- just like all other blocking options it should start with "block". Here skewing would be the trigger to enable tiling along the skewing (not skewed) dimensions too?
* if it's not a subpass of blocking, then OK with calling it `skewing`, but then we should revisit the idea of keeping everything inside `blocking.py`. Maybe just use `misc.py` then.

devito/passes/clusters/blocking.py
```diff
@@ -157,3 +160,85 @@ def decompose(ispace, d, block_dims):
...
+
+
+class Skewing(Queue):
+
+    """
+    Rebuild clusters with skewed expressions and skewed IterationSpace.
+
+    Notes
+    -----
+    This transformation is applying Loop skewing to derive the
```
FabioLuporini, line 172: Loop -> loop

devito/passes/clusters/blocking.py
```diff
@@ -157,3 +160,85 @@ def decompose(ispace, d, block_dims):
...
+    wavefront method of execution of nested loops.  Loop skewing is
+    a simple transformation of loop bounds and is combined with loop
+    interchanging to generate the wavefront [1]_.
```
FabioLuporini, line 175: so, "interchange" is part of skewing?

devito/passes/clusters/blocking.py
```diff
@@ -157,3 +160,85 @@ def decompose(ispace, d, block_dims):
...
+
+    .. code-block:: python
+
+        for i = 2, n-1
+            for j = 2, m-1
+                a[i,j] = (a[a-1,j] + a[i,j-1] + a[i+1,j] + a[i,j+1]) / 4
+            end for
+        end for
+
+        to
```
FabioLuporini, line 190: "to" outside of the code-block, then open a new code-block?

also, the ::python doesn't make sense since you're not using python syntax....unless you drop the `end for` (at least)

FabioLuporini (changes requested):

devito/core/gpu.py
georgebisbas: ok

devito/core/gpu.py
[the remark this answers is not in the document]
georgebisbas: Well, I agree. That is why I said 'no perf gains == no value.'
So my opinion: 
- It should be a sub-pass of blocking, but it should also be tested for correctness as a standalone operation.
99% of the cases will be used in combination with blocking.

devito/passes/clusters/blocking.py
georgebisbas: No. `interchange` is part of wavefront method.
Wavefront needs skewing+interchange.

Wavefront with skewing+BLOCKING+interchange = even better wavefront.

devito/passes/clusters/blocking.py
georgebisbas: Sure.

devito/passes/clusters/blocking.py
```diff
@@ -45,6 +49,9 @@ def process(self, clusters):
 
         processed = super(Blocking, self).process(processed)
 
+        if self.skewing:
```
FabioLuporini, line 52: This is now what I suggested on Slack, and putting it here doesn't make much sense to me (nested Queues?)

devito/passes/clusters/blocking.py
```diff
@@ -65,6 +72,9 @@ def callback(self, clusters, prefix):
 
         d = prefix[-1].dim
 
+        if self.levels == 0:
```
FabioLuporini, line 75: this is in Italo's PR too btw.

And it belongs to `def process` no?

georgebisbas: Fixed

devito/passes/clusters/blocking.py
georgebisbas: Fixed

devito/passes/clusters/blocking.py
```diff
@@ -161,3 +181,96 @@ def decompose(ispace, d, block_dims):
...
+            if (PARALLEL not in c.properties[d] or d.symbolic_incr.is_Symbol):
+                return clusters
+
+            if d is c.ispace[-1].dim and self.skewinner is False:
+                return clusters
+
+            mapper, intervals = {}, []
+            skew_dim = None
+
+            relations = c.ispace.intervals.relations
```
FabioLuporini, line 246: `c.ispace.relations` would be enough, and this looks like a useless line

devito/passes/clusters/blocking.py
```diff
@@ -161,3 +181,96 @@ def decompose(ispace, d, block_dims):
...
+
+            properties = {k: v for k, v in c.properties.items()}
+            for n, i in enumerate(c.ispace):
```
FabioLuporini, line 249: should you not rather loop over prefix, once and for all **before** the `for c in clusters` loop starts?

devito/passes/clusters/blocking.py
```diff
@@ -161,3 +181,96 @@ def decompose(ispace, d, block_dims):
...
+                # Identify a skew_dim and its position
+                if (SEQUENTIAL in c.properties[i.dim] and
+                   not i.dim.symbolic_incr.is_Symbol):
```
FabioLuporini, line 252: Remind me what is `not i.dim.symbolic_incr.is_Symbol` for?

devito/passes/clusters/blocking.py
```diff
@@ -161,3 +181,96 @@ def decompose(ispace, d, block_dims):
...
+                    skew_dim = i.dim
+            if not skew_dim:
+                # return if no skewing dimension found
+                return clusters
+
+            # Since we are here, prefix is skewable and nested under a
+            # SEQUENTIAL loop.
+            mapper[d] = d - skew_dim
+
+            for i in c.ispace:
```
FabioLuporini, line 262: `intervals = []` should be defined right on top of this loop , not far away in the code

devito/passes/clusters/blocking.py
```diff
@@ -161,3 +181,96 @@ def decompose(ispace, d, block_dims):
...
+                if i.dim is d:
+                    intervals.append(Interval(i.dim, skew_dim, skew_dim))
+                else:
+                    intervals.append(i)
+
+            intervals = IntervalGroup(intervals, relations=relations)
+
+            exprs = xreplace_indices(c.exprs, mapper)
```
FabioLuporini, line 270: instead of carrying `mapper` around, just `{d: d - skew_dim}` here?

devito/passes/clusters/blocking.py
```diff
@@ -161,3 +181,96 @@ def decompose(ispace, d, block_dims):
...
+            # Since we are here, prefix is skewable and nested under a
+            # SEQUENTIAL loop.
+            mapper[d] = d - skew_dim
+
+            for i in c.ispace:
+                if i.dim is d:
+                    intervals.append(Interval(i.dim, skew_dim, skew_dim))
+                else:
+                    intervals.append(i)
+
```
FabioLuporini, line 267: no need for blankline

devito/passes/clusters/blocking.py
```diff
@@ -161,3 +181,96 @@ def decompose(ispace, d, block_dims):
...
+            intervals = IntervalGroup(intervals, relations=relations)
```
FabioLuporini, line 268: as I wrote above, just `, c.ispace.relations)`

devito/passes/clusters/blocking.py
```diff
@@ -161,3 +181,96 @@ def decompose(ispace, d, block_dims):
...
+
+            exprs = xreplace_indices(c.exprs, mapper)
+            ispace = IterationSpace(intervals, c.ispace.sub_iterators,
```
FabioLuporini, line 271: move this line above, right below the construction of `intervals`

FabioLuporini (changes requested):

devito/passes/clusters/blocking.py
georgebisbas: Why should I do this? I cannot understand TBH.

devito/passes/clusters/blocking.py
georgebisbas: Actually, this is a remainder from a blocked time loop. I drop it now since it not needed for this PR, but will definitely need to take some care later...

devito/passes/clusters/blocking.py
georgebisbas: ok

devito/passes/clusters/blocking.py
georgebisbas: Done

devito/passes/clusters/blocking.py
georgebisbas: Done

devito/ir/clusters/analysis.py
```diff
@@ -192,3 +193,34 @@ def _callback(self, clusters, d, prefix):
             return
 
         return TILABLE
+
+
+class Skewing(Detector):
```
georgebisbas, line 198: A lot of duplicate code. Currently when sth is TILABLE, it is SKEWABLE as well.
Though this may change.

devito/passes/clusters/blocking.py
```diff
@@ -6,7 +6,21 @@
...
+from devito.ir.support import SEQUENTIAL, PARALLEL, Interval
+from devito.symbolics import xreplace_indices
+
+__all__ = ['blocking']
+
+
+@timed_pass(name='blocking')
+def blocking(clusters, options):
+
+    processed = Blocking(options).process(clusters)
```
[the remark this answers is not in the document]
FabioLuporini: you resolved, but u didn;t do it ;)

devito/passes/clusters/blocking.py
```diff
@@ -161,3 +203,86 @@ def decompose(ispace, d, block_dims):
...
+                return clusters
+
+            skew_dim = None
+            # Search for a skew_dim candidate
+            for i in c.ispace:
+                if SEQUENTIAL in c.properties[i.dim]:
+                    skew_dim = i.dim
+                    continue
+            if not skew_dim:
+                # return if no skewing dimension found
```
FabioLuporini, line 269: useless comment, drop

devito/passes/clusters/blocking.py
```diff
@@ -161,3 +203,86 @@ def decompose(ispace, d, block_dims):
...
+
+            if d is c.ispace[-1].dim and not self.skewinner:
+                return clusters
+
+            skew_dim = None
+            # Search for a skew_dim candidate
+            for i in c.ispace:
+                if SEQUENTIAL in c.properties[i.dim]:
+                    skew_dim = i.dim
+                    continue
```
FabioLuporini, line 267: this is a bit ugly, what happens if you have two SEQUENTIAL dimensions? you stick to the latest one? why don't you collect all of them, and skew along all of them , such as `x - time - time2...` ?

FabioLuporini: ie
```skew_dims = {i.dim for i in c.ispace if SEQUENTIAL in c.properties[i.dim]}```

File: devito/passes/clusters/blocking.py [hunk 63]
```diff
@@ -161,3 +203,86 @@ def decompose(ispace, d, block_dims):
...
+            for i in c.ispace:
+                if i.dim is d:
+                    intervals.append(Interval(d, skew_dim, skew_dim))
+                else:
+                    intervals.append(i)
+            intervals = IntervalGroup(intervals, relations=c.ispace.relations)
+            ispace = IterationSpace(intervals, c.ispace.sub_iterators,
+                                    c.ispace.directions)
+
+            exprs = xreplace_indices(c.exprs, {d: d - skew_dim})
```
FabioLuporini, line 284: in light of my comment above, this would become something along the lines of `d - sum(-i for i in skew_dims)`

tests/test_autotuner.py
```diff
@@ -265,16 +287,19 @@ def test_hierarchical_blocking():
 
 
 @switchconfig(platform='cpu64-dummy')  # To fix the core count
-def test_multiple_threads():
+@pytest.mark.parametrize('opt_dict', [{'skewing': False}, {'skewing': True}])
```
FabioLuporini, line 290: opt_options is the conventional name

tests/test_autotuner.py
```diff
@@ -265,16 +287,19 @@ def test_hierarchical_blocking():
...
+def test_multiple_threads(opt_dict):
     """
     Test autotuning when different ``num_threads`` for a given OpenMP parallel
     region are attempted.
     """
     grid = Grid(shape=(96, 96, 96))
 
+    opt_dict.update({'openmp': True})
```
FabioLuporini, line 298: drop, replace with (a few lines above)

```
opt=('blocking', {**opt_options, 'openmp': True})
```

FabioLuporini: this comment applies to all other tests in which ur doing analogous things

FabioLuporini (changes requested):

devito/ir/clusters/analysis.py
```diff
@@ -192,3 +193,19 @@ def _callback(self, clusters, d, prefix):
...
+
+    """
+    Detect the SKEWABLE Dimensions.
+    """
+
+    def process(self, elements):
```
FabioLuporini, line 204: this is probably unnecessary (why?)

devito/passes/clusters/blocking.py
```diff
@@ -1,12 +1,44 @@
...
+    ----------
+    clusters : tuple of Clusters
+        Input Clusters, subject of the optimization pass.
+    options : dict
+        The optimization options.
+        * `blockinner` (boolean, False): enable/disable loop blocking along the
+           innermost loop.
+        * `blocklevels` (int, 1): 1 => classic loop blocking; 2 for two-level
+           hierarchical
+           blocking.
```
FabioLuporini, line 28: inline with line above

devito/passes/clusters/blocking.py
```diff
@@ -1,12 +1,44 @@
...
+        * `skewing` (boolean, False): enable loop skewing.
```
FabioLuporini, line 29: enable/disable (for consistency)

devito/passes/clusters/blocking.py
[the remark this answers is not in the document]
FabioLuporini: again, this is still called `processed` (it's nitpicking, fine...), but it's not really "processed" yet

devito/passes/clusters/blocking.py
```diff
@@ -24,28 +56,12 @@ def __init__(self, options):
...
-            else:
-                processed.append(c)
-
-        processed = super(Blocking, self).process(processed)
-
-        return processed
+        if self.levels < 1:
+            return clusters
+        else:
+            processed = super(Blocking, self).process(clusters)
```
FabioLuporini, line 63: `return super(...)` (because it's a one-liner scope)

devito/passes/clusters/blocking.py
```diff
@@ -88,10 +104,14 @@ def callback(self, clusters, prefix):
...
+                # TILABLE property is dropped after the blocking.
+                # SKEWABLE is dropped as well, but only from the new
+                # block dimensions.
                 properties = dict(c.properties)
                 properties.pop(d)
                 properties.update({bd: c.properties[d] - {TILABLE} for bd in block_dims})
-
+                properties.update({bd: c.properties[d] - {SKEWABLE}
+                                  for bd in block_dims[:-1]})
                 processed.append(c.rebuild(exprs=exprs, ispace=ispace,
```
FabioLuporini, line 115: no blank line anymore? why?

devito/passes/clusters/blocking.py
```diff
@@ -88,10 +104,14 @@ def callback(self, clusters, prefix):
...
 
                 # The new Cluster properties
+                # TILABLE property is dropped after the blocking.
+                # SKEWABLE is dropped as well, but only from the new
+                # block dimensions.
                 properties = dict(c.properties)
                 properties.pop(d)
                 properties.update({bd: c.properties[d] - {TILABLE} for bd in block_dims})
-
+                properties.update({bd: c.properties[d] - {SKEWABLE}
```
FabioLuporini, line 113: replace `bd` with `i` if it makes it fit on a single line?

FabioLuporini: if so, change `bd` also in the line above

devito/passes/clusters/blocking.py
```diff
@@ -161,3 +203,82 @@ def decompose(ispace, d, block_dims):
...
+
+    def callback(self, clusters, prefix):
+        if not prefix:
+            return clusters
+
+        d = prefix[-1].dim
+
+        processed = []
+        for c in clusters:
+            # Explore skewing possibilities only in case d is SKEWABLE
```
FabioLuporini, line 254: feels like a useless comment

devito/passes/clusters/blocking.py
```diff
@@ -161,3 +203,82 @@ def decompose(ispace, d, block_dims):
...
+            # and not innermost
```
FabioLuporini, line 255: this piece of comment also doesn't seem to apply here

devito/passes/clusters/blocking.py
```diff
@@ -161,3 +203,82 @@ def decompose(ispace, d, block_dims):
...
+            if SKEWABLE not in c.properties[d]:
+                return clusters
+
+            if d is c.ispace[-1].dim and not self.skewinner:
+                return clusters
+
+            # Search for a skew_dim candidate, if more than one, we keep the last
```
FabioLuporini, line 262: that's not what it's happening below, because if len()>1 you return. You may even drop  this comment

devito/passes/clusters/blocking.py
georgebisbas: ok

devito/passes/clusters/blocking.py
georgebisbas: ok

devito/passes/clusters/blocking.py
georgebisbas: ok

devito/passes/clusters/blocking.py
georgebisbas: 94>90 :-(

tests/test_subdomains.py
```diff
@@ -364,7 +367,10 @@ class Inner(SubDomainSet):
 
         assert((np.array(result) == np.array(fex.data[:])).all())
 
-    def test_multi_sets_eq(self):
+    @pytest.mark.parametrize('opt', ['advanced', 'blocking',
+                                     ('blocking', {'skewing': True}),
+                                     ('blocking', {'skewing': True, 'blockinner': True})])
+    def test_multi_sets_eq(self, opt):
```
FabioLuporini, line 373: why not just two cases

```
['advanced',
 ('advanced', {'skewing': True})]
```

tests/test_subdomains.py
```diff
@@ -159,7 +159,10 @@ def define(self, dimensions):
 
         assert u0.data.all() == u1.data.all() == u2.data.all() == u3.data.all()
 
-    def test_iterate_NDomains(self):
+    @pytest.mark.parametrize('opt', ['advanced', 'blocking',
```
FabioLuporini, line 162: why not just two cases

```
['advanced',
 ('advanced', {'skewing': True})]
```

devito/passes/clusters/blocking.py
georgebisbas: True

devito/passes/clusters/blocking.py
georgebisbas: right!

tests/test_subdomains.py
georgebisbas: One case out of four is redundant, though I need to be checking innermost skewing, in order not to harm subdomains, as subds appear in innermost position in other clusters.
Though I admit this issue was resolved since the introduction of SKEWABLE, cause of course subds are not TILABLE and therefore not SKEWABLE. If it is not an issue, I would prefer to keep the test with blockinner.

So from the four tests I can keep 3 but better not 2.

tests/test_subdomains.py
georgebisbas: same

devito/passes/clusters/blocking.py
```diff
@@ -1,12 +1,43 @@
...
+        * `blocklevels` (int, 1): 1 => classic loop blocking; 2 for two-level
+           hierarchical blocking.
+        * `skewing` (boolean, False): enable/disable loop skewing.
+
+    Notes
+    ------
+    In case of skewing, if 'blockinner' is enabled, the innermost loop is also skewed.
+    """
+    clusters = preprocess(clusters, options)
+    processed = Blocking(options).process(clusters)
```
FabioLuporini, line 35: it would be more homogeneous if you could lift the `if self.levels < 1` up here

devito/passes/clusters/blocking.py
```diff
@@ -1,12 +1,43 @@
...
+           innermost loop.
+        * `blocklevels` (int, 1): 1 => classic loop blocking; 2 for two-level
+           hierarchical blocking.
+        * `skewing` (boolean, False): enable/disable loop skewing.
+
+    Notes
+    ------
+    In case of skewing, if 'blockinner' is enabled, the innermost loop is also skewed.
+    """
+    clusters = preprocess(clusters, options)
```
FabioLuporini, line 34: nitpicking: I'd leave a blank line between the preprocessing and the subpasses

devito/passes/clusters/blocking.py
FabioLuporini: can fit if you also drop the spaces between the minus (both lines)

georgebisbas: Unfortunately still doesn;t fit.

devito/passes/clusters/blocking.py
georgebisbas: oops

devito/passes/clusters/blocking.py
```diff
@@ -1,12 +1,45 @@
...
+           innermost loop.
+        * `blocklevels` (int, 1): 1 => classic loop blocking; 2 for two-level
+           hierarchical blocking.
+        * `skewing` (boolean, False): enable/disable loop skewing.
+
+    Notes
+    ------
+    In case of skewing, if 'blockinner' is enabled, the innermost loop is also skewed.
+    """
+    processed = preprocess(clusters, options)
```
FabioLuporini, line 34: this was pointed out a few times, "processed" is only at the end, when it needs no more processing... anyway, nitpicking

devito/passes/clusters/blocking.py
```diff
@@ -88,9 +98,14 @@ def callback(self, clusters, prefix):
                 exprs = [uxreplace(e, {d: bd}) for e in c.exprs]
 
                 # The new Cluster properties
+                # TILABLE property is dropped after the blocking.
```
FabioLuporini, line 101: # comments don't take full stop at end

FabioLuporini: after the blocking -> after blocking

FabioLuporini: sounds like a useless comment though

devito/passes/clusters/blocking.py
```diff
@@ -103,6 +118,28 @@ def callback(self, clusters, prefix):
         return processed
 
 
+def preprocess(clusters, options):
+    # Preprocess: heuristic: drop TILABLE from innermost Dimensions to
```
FabioLuporini, line 122: this could maybe be turned into a docstring now that preprocess has its own function

FabioLuporini (approved):
left some nitpicks, addressable in the next PR

This now looks great ! Thanks for the patience :)

devito/passes/clusters/blocking.py
georgebisbas: ok

devito/passes/clusters/blocking.py
georgebisbas: ok

PR #1626, opened 2021-03-08
arch: Review get_gpu_info

Leitevmd:
@FabioLuporini , please check review comments. We could also improve another points (e.g.,  get physical id also from `lspci` and `nvidia-smi`)

### DIFF
--- a/devito/arch/archinfo.py
+++ b/devito/arch/archinfo.py
@@ -227,8 +227,14 @@ def get_gpu_info():
                 gpu_info = {}
                 match = re.match(r'GPU *[0-9]*\: ([\w]*) (.*) \(', line)
                 if match:
-                    gpu_info['architecture'] = match.group(1)
-                    gpu_info['product'] = match.group(2)
+                    if match.group(1) == 'Graphics':
+                        gpu_info['architecture'] = 'unspecified'
+                    else:
+                        gpu_info['architecture'] = match.group(1)
+                    if match.group(2) == 'Device':
+                        gpu_info['product'] = 'unspecified'
+                    else:
+                        gpu_info['product'] = match.group(2)
                     gpu_info['vendor'] = 'NVIDIA'
                     gpu_infos.append(gpu_info)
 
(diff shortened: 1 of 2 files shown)


devito/arch/archinfo.py
```diff
@@ -239,7 +245,7 @@ def nvidiasmi_gpu_info(text):
     def homogenise_gpus(gpu_infos):
         if gpu_infos == []:
             warning('No graphics cards detected')
-            return None
+            return {}
```
Leitevmd, line 248: @FabioLuporini this is what you requested

devito/arch/archinfo.py
```diff
@@ -251,7 +257,7 @@ def homogenise_gpus(gpu_infos):
 
         warning('Different models of graphics cards detected')
 
-        return None
+        return {}
```
Leitevmd, line 260: Also here

devito/arch/archinfo.py
```diff
@@ -227,8 +227,14 @@ def nvidiasmi_gpu_info(text):
                 gpu_info = {}
                 match = re.match(r'GPU *[0-9]*\: ([\w]*) (.*) \(', line)
                 if match:
-                    gpu_info['architecture'] = match.group(1)
-                    gpu_info['product'] = match.group(2)
+                    if match.group(1) == 'Graphics':
```
Leitevmd, line 230: @FabioLuporini This is for the output you showed me on slack.. let me know what you think

tests/test_gpu_common.py
```diff
@@ -19,9 +19,10 @@ def test_get_gpu_info(self):
         info = get_gpu_info()
         try:
             assert info['architecture'].lower() in ['tesla', 'geforce', 'unspecified']
-        except TypeError:
+        except KeyError:
```
Leitevmd, line 22: Here we need to define what we want.. let pass only these values or even empty info (which was previous behavior)

FabioLuporini: yeah I think the current behaviour is OK

tests/test_gpu_common.py
```diff
@@ -19,9 +19,10 @@ def test_get_gpu_info(self):
...
             # There might be than one GPUs, but for now we don't care
             # as we're not really exploiting this info yet...
+            print("KeyError")
```
FabioLuporini, line 25: leftover, drop?

devito/arch/archinfo.py
FabioLuporini: Looks good to me

FabioLuporini:
Thanks, merging

PR #1628, opened 2021-03-10
compiler: Fix processing of grid spacing

### LINKED ISSUE #1629: Thickness overriding when using a different grid is broken
Explanation. Currently:

* `Grid` creates `Dimensions`; each `Dimension` is provided with a `Constant` spacing. Therefore, the physical value is attached to `Dimension` via the spacing.
* Two grids created via `Grid(shape=(...))` are defined over different `Dimension`s .
* However, this means that `grid.is_distributed(self.root)` in `SubDimension._arg_values(...)` will return false when `grid` is not the Grid used to construct the `Operator`. This breaks the thickness overriding.
* One could then say "OK, I'm then going to create the second grid via `Grid(shape=(...), dimensions=grid1.dimensions)`", but the issue here is that then both grids will **share the same state** (in particular, the Dimension spacing).

Reproducer:

```
from devito import *
import numpy as np

grid = Grid(shape=(10, 10))
x, y = grid.dimensions

xi = SubDimension.middle(name='xi', parent=x, thickness_left=3, thickness_right=3)
yi = SubDimension.middle(name='yi', parent=y, thickness_left=3, thickness_right=3)

u = TimeFunction(name='u', grid=grid, space_order=2, time_order=0)

eqn = Eq(u.forward, u + 1).subs({x: xi, y: yi})

op = Operator(eqn)

grid2 = Grid(shape=(10, 10), dimensions=(x, y)) 

u2 = TimeFunction(name='u', grid=grid2, space_order=2, time_order=0)

op.apply(time_M=0, u=u2)

assert np.all(u2.data[0, 3:-3, 3:-3] == 1.) 
```

FabioLuporini:
fixes #1629

### DIFF
--- a/devito/operator/operator.py
+++ b/devito/operator/operator.py
@@ -456,6 +456,7 @@ class Operator(Callable):
             raise ValueError("Multiple Grids found")
         try:
             grid = grids.pop()
+            args.update(grid._arg_values(**kwargs))
         except KeyError:
             grid = None
 
@@ -517,8 +518,16 @@ class Operator(Callable):
     @cached_property
     def _known_arguments(self):
         """The arguments that can be passed to ``apply`` when running the Operator."""
-        ret = set.union(*[set(i._arg_names) for i in self.input + self.dimensions])
-        return tuple(sorted(ret))
+        ret = set()
+        for i in self.input:
+            ret.update(i._arg_names)
+            try:
+                ret.update(i.grid._arg_names)
+            except AttributeError:
+                pass
+        for d in self.dimensions:
+            ret.update(d._arg_names)
+        return frozenset(ret)
 
     def _autotune(self, args, setup):
         """Auto-tuning to improve runtime performance."""
--- a/devito/types/grid.py
+++ b/devito/types/grid.py
@@ -1,14 +1,16 @@
 from collections import namedtuple
+from math import floor
 
 import numpy as np
 from sympy import prod
-from math import floor
+from cached_property import cached_property
 
 from devito.data import LEFT, RIGHT
+from devito.logger import warning
 from devito.mpi import Distributor, MPI
 from devito.tools import ReducerMap, as_tuple, memoized_meth
 from devito.types.args import ArgProvider
-from devito.types.constant import Constant
+from devito.types.basic import Scalar
 from devito.types.dense import Function
 from devito.types.dimension import (Dimension, SpaceDimension, TimeDimension,
                                     SteppingDimension, SubDimension)
@@ -111,8 +113,8 @@ class Grid(ArgProvider):
             # Create the spatial dimensions and constant spacing symbols
             assert(self.dim <= 3)
             dim_names = self._default_dimensions[:self.dim]
-            dim_spacing = tuple(Constant(name='h_%s' % n, value=v, dtype=self.dtype)
-                                for n, v in zip(dim_names, self.spacing))
+            dim_spacing = tuple(Scalar(name='h_%s' % n, dtype=self.dtype, is_const=True)
+                                for n in dim_names)
             self._dimensions = tuple(SpaceDimension(name=n, spacing=s)
                                      for n, s in zip(dim_names, dim_spacing))
         else:
@@ -128,16 +130,16 @@ class Grid(ArgProvider):
         self._subdomains = subdomains
 
         self._origin = as_tuple(origin or tuple(0. for _ in self.shape))
-        self._origin_symbols = tuple(Constant(name='o_%s' % d.name, value=v,
-                                              dtype=self.dtype)
-                                     for d, v in zip(self.dimensions, self.origin))
+        self._origin_symbols = tuple(Scalar(name='o_%s' % d.name, dtype=dtype,
+                                            is_const=True)
+                                     for d in self.dimensions)
 
         # Sanity check
         assert (self.dim == len(self.origin) == len(self.extent) == len(self.spacing))
 
         # Store or create default symbols for time and stepping dimensions
         if time_dimension is None:
-            spacing = Constant(name='dt', dtype=self.dtype)
+            spacing = Scalar(name='dt', dtype=dtype, is_const=True)
             self._time_dim = TimeDimension(name='time', spacing=spacing)
             self._stepping_dim = SteppingDimension(name='t', parent=self.time_dim)
         elif isinstance(time_dimension, TimeDimension):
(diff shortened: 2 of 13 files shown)


devito/types/grid.py
```diff
@@ -258,36 +267,69 @@ def is_distributed(self, dim):
...
+        # Dimensions size
         for k, v in self.dimension_map.items():
             args.update(k._arg_defaults(_min=0, size=v.loc))
 
+        # Dimensions spacing
+        try:
+            args.update({k.name: v for k, v in self.spacing_map.items()})
+        except AttributeError:
+            # See issue #1524
+            # We check whether we're in the special case {n*h_x: v, ...}, which is
```
mloubout, line 297: Wouldnt it be "easier" to fix the map so that it returns `{h_x: v/n, ...}`?

devito/types/grid.py
```diff
@@ -127,21 +129,23 @@ def __init__(self, shape, extent=None, origin=None, dimensions=None,
...
 
         # Sanity check
         assert (self.dim == len(self.origin) == len(self.extent) == len(self.spacing))
 
         # Store or create default symbols for time and stepping dimensions
         if time_dimension is None:
-            spacing = self._const(name='dt', dtype=self.dtype)
-            self._time_dim = self._make_time_dim(spacing)
-            self._stepping_dim = self._make_stepping_dim(self.time_dim, name='t')
+            spacing = Scalar(name='dt', dtype=dtype, is_const=True)
```
mloubout, line 142: I would still have a default value (dt=1) similar to how we have a default value for the extent. But not sure if breaks stuff.

mloubout (commented):
Looks fine overall, some design comments.

devito/types/grid.py
FabioLuporini: I actually totally agree with you. I've just changed it to do as you suggest

devito/types/grid.py
FabioLuporini: I'm all but convinced about this one. I think it could lead to various user level errors

devito/types/grid.py
```diff
@@ -211,22 +226,35 @@ def spacing(self):
...
-        grid_origin = [min(i) for i in self.distributor.glb_numb]
-        assert len(grid_origin) == len(self.spacing)
-        return tuple(i*h for i, h in zip(grid_origin, self.spacing))
+            # Typical case: `Grid.dimensions` -> (x, y, z)` where `x, y, z` are
+            # the SpaceDimensions
+            if d.is_Space:
+                mapper[d.spacing] = s
+                continue
+
+            raise RuntimeError("Unable to build `spacing_map` for `%s`" % self)
```
mloubout, line 255: Would add something along the lines of `"Unsupported dimension %s type for dimension %s " (d.name, d.type)` for clarity

devito/operator/operator.py
```diff
@@ -517,8 +518,16 @@ def _postprocess_arguments(self, args, **kwargs):
...
-        return tuple(sorted(ret))
+        ret = set()
+        for i in self.input:
+            ret.update(i._arg_names)
+            try:
+                ret.update(i.grid._arg_names)
+            except AttributeError:
+                pass
+        for d in self.dimensions:
+            ret.update(d._arg_names)
```
mloubout, line 529: So:
- `grid` arguments are reprocessed for all inputs?
- `dimensions` are reprocessed one more time afterward?
Seems a bit overkill. By here we should have check the grid is unique so should be something like

```
        for i in self.input:
            ret.update(i._arg_names)
            try:
                ret.update(i.grid._arg_names)
               break
            except AttributeError:
                pass
```

devito/types/grid.py
mloubout: Ok, fine with me.

devito/types/grid.py
```diff
@@ -211,22 +226,35 @@ def spacing(self):
...
+        for d, s in zip(self.dimensions, self.spacing):
+            # Special case subsampling: `Grid.dimensions` -> (xb, yb, zb)` where
+            # `xb, yb, zb` are ConditionalDimensions whose parents are SpaceDimensions
+            if d.is_Conditional:
+                try:
+                    a, b = d.spacing.args
+                    mapper[b] = s/self.dtype(a)
+                    continue
+                except (AttributeError, ValueError):
+                    pass
```
mloubout, line 247: Don't all ConditionalDimension have a `factor`, i.e `mapper[d.parent.spacing] = s/d.factor` should always be valid  (it's 1 by default for non=subsampled one` instead of the try/except.

FabioLuporini: yeah, but `factor` isn't always used by `ConditionalDimension` (it could be None). Anyway, you may have a point, I could assume `factor!=None` and otherwise we raise the exception

mloubout: No it can't be none, if `d._factor` is `None` then `d.factor` returns 1

mloubout: https://github.com/devitocodes/devito/blob/d1b601b4670ad758d22738f33570d0d61afe646e/devito/types/dimension.py#L769

devito/operator/operator.py
FabioLuporini: I don't follow. This is just the `_known_arguments` property, suitably patched (flawed in current master) so that we catch user errors in supplying arguments to `op.apply(...)`. BTW, it's a `cached_property`, so this runs only once?

And `_arg_names` isn't a recursive call

devito/types/grid.py
FabioLuporini: OK, doing it

FabioLuporini: Actually, this should be caught in the constructor, and this exception should be turned into an `assert False` if you see what I mean. Will do this

devito/types/grid.py
FabioLuporini: ah! good catch

devito/types/grid.py
```diff
@@ -211,22 +234,34 @@ def spacing(self):
...
     def spacing_map(self):
         """Map between spacing symbols and their values for each SpaceDimension."""
-        return dict(zip(self.spacing_symbols, self.spacing))
+        mapper = {}
+        for d, s in zip(self.dimensions, self.spacing):
+            if d.is_Conditional:
+                # Special case subsampling: `Grid.dimensions` -> (xb, yb, zb)`
+                # where `xb, yb, zb` are ConditionalDimensions whose parents
+                # are SpaceDimensions
+                try:
```
mloubout, line 251: Not sure why the try/except is still needed since the `try` part is always valid, but except that GTG

FabioLuporini: you're right again. Will drop. Great!

mloubout (approved):

PR #1630, opened 2021-03-12
mpi: Add diag2 mode

FabioLuporini:


tjb900 (approved):
LGTM! Generated code makes sense. Thanks very much!

mloubout (approved):

georgebisbas (commented):
What are the main advantages of this mode compared to the others?

tjb900:
> What are the main advantages of this mode compared to the others?

The motivations were:

* like "basic" mode, don't do compute/communication overlap, because in some circumstances it's not always a win (e.g. due to the serialisation overhead due to the loop over the remainder regions)
* like "overlap2" mode, use preallocated buffers because they avoid memory area registration/deregistration when using RDMA

PR #1634, opened 2021-03-16
compiler: Work around clang[10,11,?] omp-offloading bug

FabioLuporini:


JonChesterfield:
Haven't reviewed, but can confirm that applying this locally generated openmp code without the alignment parameter which then got much further through running the benchmark

PR #1637, opened 2021-03-18
ci: Create new environment for examples mpi tests

Leitevmd:


rhodrin:
Why is this needed? We could give the `asv` env a more generic name (and even tweak it if necessary) but it's cleaner to have one env for both these workflows.

FabioLuporini:
closing as the `asv` environment is reusable for this workflow too

PR #1640, opened 2021-03-23
dsl: better fix for custom coefficients on staggered grids

EdCaunt:
Also fixes the issue where custom coefficients would not be applied to forward or backward timesteps


FabioLuporini:
closing, as superseded by #1644

PR #1644, opened 2021-03-24
dsl: Overhaul custom fd weights

rhodrin:
Work in progress. TBC.

Aside from the actual implementation, we'll also need to discuss legacy support at some stage.


devito/finite_differences/derivative.py
```diff
@@ -171,18 +174,21 @@ def _process_kwargs(cls, expr, *dims, **kwargs):
...
         return new_dims, orders, fd_orders, variable_count
 
-    def __call__(self, x0=None, fd_order=None, side=None):
+    def __call__(self, x0=None, coeffs=None, fd_order=None, side=None):
         if self.ndims == 1:
             _fd_order = fd_order or self._fd_order
             _side = side or self._side
             new_x0 = {self.dims[0]: x0} if x0 is not None else self.x0
-            return self._new_from_self(fd_order=_fd_order, side=_side, x0=new_x0)
+            new_coeffs = coeffs
```
FabioLuporini, line 182: `coeffs or self._coeffs` ?

FabioLuporini (commented):
Looking really really nice :)

FabioLuporini:
@rhodrin 

> Aside from the actual implementation, we'll also need to discuss legacy support at some stage.

First thought: drop and burn release

mloubout:
RIght what I had in mind. The it'll be trivial to extend it to allowing `(ind, weight)` pairs as coeffs for people who really want custom stencils in general (like 3x3 kernel for first derivative)

For legacy, not completely sure yes to be discussed.

devito/finite_differences/derivative.py
rhodrin: Need to check - something wasn't working last night so wrote it in this way for now but quite a few things in there need checking and fixing still.

rhodrin:
> First thought: drop and burn release
>
> Second thought: provide a mini (mini) tutorial explaining how to transition from old to new

Sounds good to me.

rhodrin:
> RIght what I had in mind. The it'll be trivial to extend it to allowing (ind, weight) pairs as coeffs for people who really want > custom stencils in general (like 3x3 kernel for first derivative)
>
> For legacy, not completely sure yes to be discussed.

Yeah, indeed. Will implement that functionality. It's just some minor editing here and there.

Re. legacy support, we could for the time being add a decorator to equation that based on the equation + presence of a `coefficients` object just reassembles the equation in the new form. It's not pretty, but should be fairly simple to implement and then cut out when we deprecate.

PR #1645, opened 2021-03-26
gpu: Add gpu-fit value for all functions (fix #1642)

### LINKED ISSUE #1642: Segfault error in test_adjoint with GPU+OpenACC
Issue related to https://github.com/devitocodes/devito/pull/1639/checks?check_run_id=2176365412#step:5:43

Leitevmd:
Fix #1642


devito/core/gpu.py
```diff
@@ -93,7 +93,7 @@ def _normalize_kwargs(cls, **kwargs):
         o['par-nested'] = np.inf  # Never use nested parallelism
         o['par-disabled'] = oo.pop('par-disabled', True)  # No host parallelism by default
         o['gpu-direct'] = oo.pop('gpu-direct', True)
-        o['gpu-fit'] = as_tuple(oo.pop('gpu-fit', None))
+        o['gpu-fit'] = as_tuple(oo.pop('gpu-fit', 'all-fallback'))
```
FabioLuporini, line 96: Could you rather read the value from a class attribute? something like `cls.GPU_FIT)

devito/passes/iet/misc.py
```diff
@@ -146,5 +147,12 @@ def is_on_device(maybe_symbol, gpu_fit, only_writes=False):
             expressions = FindNodes(Expression).visit(iet)
             functions &= {i.write for i in expressions}
 
+    if 'all-fallback' in gpu_fit and any(f.is_TimeFunction and f.save is not None
+                                         for f in functions):
+        warning("gpu-fit option didn't include save-full TimeFunctions [...] which "
```
FabioLuporini, line 152: I would:

```
fsave = [f for f in functions if f.is_TimeFunction and f.save is not None]
```

then

```
if 'all-fallback in gpu_fit and  fsave:
    warning(".... spell out the fsave functions here....")
```

also, maybe we can drop

```
make sure their runtime size honors this
```

(because the other obvious alternative would be to start using `gpu-fit`

FabioLuporini:
I had forgotten about one thing @Leitevmd sorry 

Look at the errors here, for example: https://github.com/devitocodes/devito/pull/1645/checks?check_run_id=2203380066#step:9:84

So, those warnings should actually be avoided (for reasons I'm explaining below), so `gpu-fit` should probably also support a new value, `'all'`

`gpu-fit` is *not* a "compiler pass", but rather an "optimization option" (see [here](https://github.com/devitocodes/devito/blob/master/examples/performance/00_overview.ipynb) for a definition of "optimization option") for two passes, `tasking` and `streaming`. So, if `tasking` or `streaming` appear in `kwargs` inside `normalize_kwargs`, then `gpu-fit` should *not* be set to `cls.GPU_FIT` (== `all-fallback`) -- and thus it should remain `None`. I suggest to put the `gpu-fit`setup logic in a separate class method (such as `_normalize_gpu_fit` -- better names welcome) so that the behaviour can be customized by subclasses

devito/passes/iet/misc.py
```diff
@@ -146,5 +147,11 @@ def is_on_device(maybe_symbol, gpu_fit, only_writes=False):
             expressions = FindNodes(Expression).visit(iet)
             functions &= {i.write for i in expressions}
 
+    fsave = [f for f in functions if f.is_TimeFunction and f.save is not None]
+    if 'all-fallback' in gpu_fit and fsave:
+        warning("gpu-fit option didn't include save-full TimeFunctions %s which "
+                "however will be assumed to fit the gpu memory" % fsave)
+        return True
+
     return all(not (f.is_TimeFunction and f.save is not None and f not in gpu_fit)
```
FabioLuporini, line 156: nitpicking: use `fsave` here too?

devito/passes/iet/misc.py
```diff
@@ -146,5 +147,11 @@ def is_on_device(maybe_symbol, gpu_fit, only_writes=False):
             expressions = FindNodes(Expression).visit(iet)
             functions &= {i.write for i in expressions}
 
+    fsave = [f for f in functions if f.is_TimeFunction and f.save is not None]
+    if 'all-fallback' in gpu_fit and fsave:
+        warning("gpu-fit option didn't include save-full TimeFunctions %s which "
```
FabioLuporini, line 152: nitpicking: maybe something "easier" such as "TimeFunction %s assumed to fit the GPU memory" ?

devito/core/gpu.py
```diff
@@ -103,6 +108,12 @@ def _normalize_kwargs(cls, **kwargs):
 
         return kwargs
 
+    @classmethod
+    def _normalize_gpu_fit(cls, **kwargs):
+        if [i for i in ['tasking', 'streaming'] if i in kwargs['mode']]:
```
FabioLuporini, line 113: nitpicking: `any(i for ...)`

devito/core/gpu.py
```diff
@@ -54,6 +54,11 @@ class DeviceOperatorMixin(object):
     Coefficient to adjust the chunk size in non-affine parallel loops.
     """
 
+    GPU_FIT = 'all-fallback'
+    """
+    Assuming all functions fit into the gpu memory
```
FabioLuporini, line 59: nitpicking: full stop at the end

FabioLuporini (approved):
Just left a bunch of nitpick-level comment . PR looks good to me. Not sure why tests hanging though?

mloubout (commented):
Looks fine to me, not sure why it hangs especially for F that doesn't even need much memory.

FabioLuporini:
Merged, thanks a lot @Leitevmd !

PR #1649, opened 2021-04-01
advisor: merge roofline and json

georgebisbas:
Json files are merged with rooflines. Drops duplicate code, enhances rooflines with callstacks and memory access patterns.

Main improvements:
- Add Memory Access Pattern collections
- Stability for more platforms, so far was mostly tested on Azure ones.
- Added workarounds for a few bugs that may easily lead to no data collected.


FabioLuporini:
Hey @jack-lascala , would you mind taking a quick look at these changes?

benchmarks/user/advisor/run_advisor.py
```diff
@@ -100,21 +100,24 @@ def run_with_advisor(path, output, name, exec_args):
     ]
     advisor_cmd = [
         'advixe-cl',
-        '-q',  # Silence advisor
+        #'-q',  # Silence advisor
```
georgebisbas, line 103: I would propose to get rid of silencing cuz it also silences bugs and failures!

benchmarks/user/advisor/roofline.py
```diff
@@ -101,6 +110,13 @@ def roofline(name, project, scale, precision, mode, th):
     gflops_min = 2**0
     width = ai_max
 
+    # Declare the dictionary that will hold the JSON information
```
FabioLuporini, line 113: "Declare the dictionary" is obvious...

benchmarks/user/advisor/roofline.py
```diff
@@ -101,6 +110,13 @@ def roofline(name, project, scale, precision, mode, th):
...
+    roofline_data = {}
+
+    # Declare the two types of rooflines dictionaries
```
FabioLuporini, line 116: so this is to generate the four lines L1/L2/L3/DRAM? and likewise for compute (I don't remember what they are, but you get the point)?

FabioLuporini (commented):
so `advisor_to_json` dropped because it was 90% redundant stuff? do I get it right?

georgebisbas:
> so `advisor_to_json` dropped because it was 90% redundant stuff? do I get it right?

Yes, look here: https://github.com/devitocodes/devito/pull/1463#pullrequestreview-496364136

benchmarks/user/advisor/roofline.py
georgebisbas: Yes, for compute...sp_flops, dp_flops...etc

benchmarks/user/advisor/roofline.py
georgebisbas: WIll drop then..leftover

PR #1667, opened 2021-04-16
compiler: Check consistency between shape and grid for TimeFunction, too

tjb900:
This duplicates the logic used for `Function.__shape_setup__` into `TimeFunction` as well - essentially sanity-checking the provided shape matches the grid and dimensions, if both are provided.

This allows for `TimeFunction`s over smaller numbers of dimensions to still work correctly with MPI.

### DIFF
--- a/devito/types/dense.py
+++ b/devito/types/dense.py
@@ -1341,7 +1341,8 @@ class TimeFunction(Function):
     def __shape_setup__(cls, **kwargs):
         grid = kwargs.get('grid')
         save = kwargs.get('save') or None  # Force to None if 0/False/None/...
-        shape = kwargs.get('shape')
+        dimensions = kwargs.get('dimensions')
+        shape = kwargs.get('shape', kwargs.get('shape_global'))
         time_order = kwargs.get('time_order', 1)
 
         if grid is None:
@@ -1360,6 +1361,14 @@ class TimeFunction(Function):
                 shape.insert(cls._time_position, save)
             else:
                 raise TypeError("`save` can be None, int or Buffer, not %s" % type(save))
+        elif dimensions is None:
+            raise TypeError("`dimensions` required if both `grid` and "
+                            "`shape` are provided")
+        else:
+            shape = super(TimeFunction, cls).__shape_setup__(
+                grid=grid, shape=shape, dimensions=dimensions
+            )
+
         return tuple(shape)
 
     @cached_property
--- a/tests/test_mpi.py
+++ b/tests/test_mpi.py
@@ -50,6 +50,31 @@ class TestDistributor(object):
         }
         assert f.shape == expected[distributor.nprocs][distributor.myrank]
 
+    @pytest.mark.parallel(mode=[2, 4])
+    def test_partitioning_fewer_dims_timefunc(self):
+        """Test domain decomposition for Functions defined over a strict subset
+        of grid-decomposed dimensions."""
+        size_x, size_y = 16, 16
+        grid = Grid(shape=(size_x, size_y))
+        x, y = grid.dimensions
+
+        # A function with fewer dimensions that in `grid`
+        f = TimeFunction(
+            name='f',
+            grid=grid,
+            dimensions=(grid.time_dim, x,),
+            shape=(10, size_x,),
+        )
+
+        distributor = grid.distributor
+        expected = {  # nprocs -> [(rank0 shape), (rank1 shape), ...]
+            2: [(8,), (8,)],
+            4: [(8,), (8,), (8,), (8,)]
+        }
+        assert len(f.shape) == 2
+        assert f.shape[0] == 10
+        assert f.shape[1:] == expected[distributor.nprocs][distributor.myrank]
+
     @pytest.mark.parallel(mode=9)
     def test_neighborhood_horizontal_2d(self):
         grid = Grid(shape=(3, 3))


devito/types/dense.py
```diff
@@ -1360,6 +1361,31 @@ def __shape_setup__(cls, **kwargs):
...
+            # Dimensions in `dimensions` also appearing in `grid` have same size
+            # (given by `shape`) as that provided in `grid`
+            if len(shape) != len(dimensions):
+                raise ValueError("`shape` and `dimensions` must have the "
+                                 "same number of entries")
+            loc_shape = []
+            for d, s in zip(dimensions, shape):
+                if d in grid.dimensions:
+                    size = grid.dimension_map[d]
+                    if size.glb != s and s is not None:
```
mloubout, line 1378: can `s` ever be None?

devito/types/dense.py
```diff
@@ -1360,6 +1361,31 @@ def __shape_setup__(cls, **kwargs):
...
+                        raise ValueError("Dimension `%s` is given size `%d`, "
```
mloubout, line 1379: `InputError` may be more adapted.

tests/test_mpi.py
```diff
@@ -50,6 +50,31 @@ def test_partitioning_fewer_dims(self):
...
+            name='f',
+            grid=grid,
+            dimensions=(grid.time_dim, x,),
+            shape=(10, size_x,),
+        )
+
+        distributor = grid.distributor
+        expected = {  # nprocs -> [(rank0 shape), (rank1 shape), ...]
+            2: [(8,), (8,)],
+            4: [(8,), (8,), (8,), (8,)]
```
mloubout, line 72: shouldn't this be 4?

mloubout (commented):
Looks sane, few minor comments. Thanks!

devito/types/dense.py
tjb900: This is just lifted from the `Function` implementation. But I think the idea is that you shouldn't have to specify the shape component for dimensions that are part of the Grid.

So you should be able to `TimeFunction(name='u', grid=grid, dimensions=(grid.time_dim, grid.dimensions[0]), shape=(10, None))`

tests/test_mpi.py
tjb900: Decomposition for 4 ranks will do 2x2, so the x size doesn't change.

devito/types/dense.py
tjb900: Again, have lifted from `Function` but happy to change both (as a second commit). Let me know.

tests/test_mpi.py
mloubout: Yes, stupid me.

mloubout (approved):

devito/types/dense.py
mloubout: Ok we'll probably go over it another time then,

tests/test_mpi.py
mloubout: @rhodrin @FabioLuporini how is handled the duplication in that case?

tjb900: Yeah good question!

devito/types/dense.py
```diff
@@ -1360,6 +1361,31 @@ def __shape_setup__(cls, **kwargs):
                 shape.insert(cls._time_position, save)
             else:
                 raise TypeError("`save` can be None, int or Buffer, not %s" % type(save))
+        elif dimensions is None:
+            raise TypeError("`dimensions` required if both `grid` and "
+                            "`shape` are provided")
+        else:
+            # Got `grid`, `dimensions`, and `shape`. We sanity-check that the
```
FabioLuporini, line 1368: does it fail in some ways that I'm currently missing if we rather just do

```
else:
    return super().__shape_setup__(grid=grid, shape=shape, dimensions=dimensions)
```

FabioLuporini (commented):
Thanks a lot @tjb900 ! Minor question added, but otherwise gtg

devito/types/dense.py
tjb900: I don't think it fails without this PR, no. But, as things stand before this change:

*  `TimeFunction` is inconsistent with `Function`. The shape that must currently be provided for grid-owned `Dimension`s is the local shape not the global shape, whereas for `Function` it's the global shape (or None).
* it doesn't verify early that the provided shape is consistent with the `Grid` - you only get that error later, at `apply`/`arguments` time

FabioLuporini: Oh, with my message above I really meant something else, sorry. I was rather proposing a potential way of avoiding a piece of duplicated code (at least, it seemed duplicated code at first glance).

tjb900: Ah, sorry, I get you now. Yes, good plan. Change pushed.

tjb900:
Argh, sorry, flake8 issue. I'm so used to being spoiled by pre-commit and black. Once this is all approved let me squash these together into a single commit.

FabioLuporini (approved):

FabioLuporini:
@tjb900 : thanks, approved now! 

Since you're rebasing, you may want to take a look at the set of tags we've established for commit messages and PR titles: https://github.com/devitocodes/devito/wiki/Tags-for-commit-messages-and-PR-titles

FabioLuporini:
Thanks a lot, merging!

PR #1669, opened 2021-04-16
compiler: Tweak nested-par candidate condition

georgebisbas:
compiler: Tweak nested-par candidate condition.
Can help preserve parallelism in cases where bounds are expressions.

### DIFF
--- a/devito/passes/iet/parpragma.py
+++ b/devito/passes/iet/parpragma.py
@@ -307,7 +307,7 @@ class PragmaShmTransformer(PragmaSimdTransformer):
             # within a block)
             candidates = []
             for i in inner:
-                if self.key(i) and any((j.dim.parent is i.dim.root) for j in outer):
+                if self.key(i) and any((j.dim.root is i.dim.root) for j in outer):
                     candidates.append(i)
                 elif candidates:
                     # If there's at least one candidate but `i` doesn't honor the
--- a/devito/passes/iet/parpragma.py
+++ b/devito/passes/iet/parpragma.py
@@ -11,7 +11,7 @@ from devito.symbolics import CondEq, INT, ccode
 from devito.passes.iet.engine import iet_pass
 from devito.passes.iet.langbase import LangBB, LangTransformer, DeviceAwareMixin
 from devito.passes.iet.misc import is_on_device
-from devito.tools import as_tuple, is_integer, prod
+from devito.tools import as_tuple, prod
 from devito.types import Symbol, NThreadsBase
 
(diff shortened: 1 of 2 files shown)


devito/passes/iet/parpragma.py
```diff
@@ -307,7 +307,7 @@ def _make_nested_partree(self, partree):
             # within a block)
             candidates = []
             for i in inner:
-                if self.key(i) and any(is_integer(j.step-i.symbolic_size) for j in outer):
+                if self.key(i) and any((j.dim.parent is i.dim.root) for j in outer):
```
FabioLuporini, line 310: I'm not sure this will work if we have SubDimensions

could you add a nested parallelism over SubDimensions test here: https://github.com/devitocodes/devito/blob/master/tests/test_dle.py#L620

probably using `grid.interior` suffices

tests/test_dle.py
```diff
@@ -157,6 +158,27 @@ def test_cache_blocking_structure_subdims():
         tree[3].dim.root is y
     assert not tree[4].dim.is_Incr and tree[4].dim is zi and tree[4].dim.parent is z
 
+    # Non-local SubDimension -> nested blocking can works as expected
```
FabioLuporini, line 161: this test is for nested parallelism, but you added it to `test_cache_blocking_...`. I think you should rather write a new test under `TestNestedParallelism`. This one is out of place

FabioLuporini: also, does hierarchical blocking with nested parallelism over subdimensions work as expected? do we have one such test?

georgebisbas: Test moved. Nested+hierarchical+subdims seems ok, test added.

tests/test_dle.py
```diff
@@ -737,3 +737,67 @@ def test_multiple_subnests_v1(self):
         assert trees[1][3].pragmas[0].value == ('omp parallel for collapse(1) '
                                                 'schedule(dynamic,1) '
                                                 'num_threads(nthreads_nested)')
+
+    def test_nested_cache_blocking_structure_subdims(self):
```
FabioLuporini, line 741: there's a huge fraction of duplicated code in this test, it's like two tests in one. This is quite ugly.
 
I would rather parametrize over `blocklevels`. You may then use the `blocklevels` value (1 or 2) to check for additional entries in `trees`

FabioLuporini (changes requested):

FabioLuporini (approved):
Thanks, merging

Issue #1655, opened 2021-04-05
GPU + openacc broken for odd number of GPUs on same node


jkwashbourne:
Devito with variables `DEVITO_PLATFORM=nvidiaX DEVITO_ARCH=nvc DEVITO_LANGUAGE=openacc DEVITO_MPI=1` only works for certain numbers of MPI ranks. On a machine with 8 GPUs, only the numbers [1,2,4,8] work.

### Test platforms: 
* azure `ND40rs_v2` with 8 v100 GPUs
    * working number of MPI ranks: [1, 2, 4, 8] 
* azure `NC24_v3` with 4 v100 GPUs  
    * working number of MPI ranks: [1, 2, 4] 

### Reproducer
`DEVITO_PLATFORM=nvidiaX DEVITO_ARCH=nvc DEVITO_LANGUAGE=openacc DEVITO_MPI=1 mpirun -n 3 python3 ./examples/seismic/acoustic/acoustic_example.py 
`

### Output log from command
```
Operator `initdamp` ran in 6.08 s
Operator `initdamp` ran in 6.08 s
Operator `initdamp` ran in 6.08 s
Operator `pad_vp` ran in 0.01 s
Operator `pad_vp` ran in 0.01 s
Operator `pad_vp` ran in 0.01 s
Applying Forward
Applying Forward
Applying Forward
hostptr=0x2426ce0,stride=1,size=3,extent=-3,eltsize=4,name=src_coords,flags=0x200=present,async=-1,threadid=1
Present table dump for device[1]: NVIDIA Tesla GPU 0, compute capability 7.0, threadid=1
...empty...
deleted block   device:0x7eff17800000 size:3184128 threadid=1 
deleted block   device:0x7eff17c00000 size:3941888 threadid=1 
Present table dump for device[3]: NVIDIA Tesla GPU 2, compute capability 7.0, threadid=1
host:0x205a8e0 device:0x7efef5b7a800 size:2448 presentcount:0+1 line:76 name:rec_coords
host:0x2426ce0 device:(nil) size:0 presentcount:0+1 line:76 name:src_coords
host:0x317f000 device:0x7efee1000000 size:3941484 presentcount:0+1 line:76 name:vp
host:0x36c1e40 device:(nil) size:0 presentcount:1+1 line:76 name:src
host:0x37e8000 device:0x7efef5afa000 size:526320 presentcount:0+1 line:76 name:rec
[...]

PR #1684, opened 2021-04-26
gpu: Fix offloading when zero-size arrays

### LINKED ISSUE #1641: Source injection/receiver interpolation fail with MPI+OpenACC
Running an operator with either source injection or receiver interpolation using OpenACC and MPI domain decomposition results in a crash when the number of ranks becomes large.

In the 2D iso-acoustic example, this problem seems to appear for any domain size when `nranks >= 3`.

The problem seems to get solved when changing:

```
#pragma acc parallel ....
```

to:

```
#pragma acc parallel .... if (p_src_M - p_src_m >= 0)
```

around the injection and interpolation.

FabioLuporini:
fixes #1641 and #1655

### DIFF
--- a/tests/test_gpu_common.py
+++ b/tests/test_gpu_common.py
@@ -2,7 +2,7 @@ import pytest
 import numpy as np
 
 from devito import (Constant, Eq, Inc, Grid, Function, ConditionalDimension,
-                    SubDimension, SubDomain, TimeFunction, Operator)
+                    SparseTimeFunction, SubDimension, SubDomain, TimeFunction, Operator)
 from devito.arch import get_gpu_info
 from devito.exceptions import InvalidArgument
 from devito.ir import Expression, Section, FindNodes, FindSymbols, retrieve_iteration_tree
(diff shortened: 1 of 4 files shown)


tests/test_gpu_common.py
```diff
@@ -877,3 +879,39 @@ def test_npthreads(self):
...
+        grid = Grid(shape=(4, 4), extent=(3.0, 3.0))
+
+        f = TimeFunction(name='f', grid=grid, space_order=0)
+        f.data[:] = 1.
+        sf1 = SparseTimeFunction(name='sf1', grid=grid, npoint=0, nt=10)
+        sf2 = SparseTimeFunction(name='sf2', grid=grid, npoint=0, nt=10)
+        assert sf1.size == 0
+        assert sf2.size == 0
+
+        eqns = sf1.inject(field=f, expr=sf1 + sf2 + 1.)
```
mloubout, line 899: can you have something else than 1? Bit confusing whether it does nothing and stays at 1 or adds one here.

mloubout (approved):

tests/test_gpu_common.py
FabioLuporini: does nothing -- see `npoint=0` above. sf1/sf2/sf -- they all are 0-size Functions

mloubout: I know that, but the equation is ambiguous IMO that's all, but GTG

PR #1691, opened 2021-05-10
API: fixed callback of PrecomputedInterpolator injection

ccuetom:
Although the interpolation using PrecomputedInterpolator was working fine, the injection tries to access the `gridpoints` and `interpolation_coeffs` in `self` instead of in `self.obj`.

Also, when injecting with the PrecomputedInterpolator, the preparation of the arguments for the operator calls the `self._parent._arg_defaults` for the gridpoints SubFunction, which does not exist. By changing the parent of the SubFunction to `self.obj` this is solved.

### DIFF
--- a/tests/test_interpolation.py
+++ b/tests/test_interpolation.py
@@ -178,15 +178,13 @@ def test_precomputed_injection():
     origin = (0, 0)
     result = 0.25
 
-    r = 2  # Constant for linear interpolation
-    #  because we interpolate across 2 neighbouring points in each dimension
+    # Constant for linear interpolation
+    # because we interpolate across 2 neighbouring points in each dimension
+    r = 2
 
     m = unit_box(shape=shape)
     m.data[:] = 0.
 
-    r = 2  # Constant for linear interpolation
-    #  because we interpolate across 2 neighbouring points in each dimension
-    
     gridpoints, interpolation_coeffs = precompute_linear_interpolation(coords,
                                                                        m.grid, origin)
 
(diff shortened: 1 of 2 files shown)


tests/test_interpolation.py
```diff
@@ -169,6 +169,42 @@ def test_precomputed_interpolation_time():
...
+def test_precomputed_injection():
+    """Test injection with PrecomputedSparseFunction which accepts
+       precomputed values for interpolation coefficients
+    """
+    shape = (11, 11)
+    coords = [(.05, .95), (.45, .45)]
+    origin = (0, 0)
+    result = 0.25
+
+    r = 2  # Constant for linear interpolation
```
georgebisbas, line 181: nitpicking:
Comment should go before variable?

tests/test_interpolation.py
```diff
@@ -169,6 +169,42 @@ def test_precomputed_interpolation_time():
...
+    #  because we interpolate across 2 neighbouring points in each dimension
+
+    m = unit_box(shape=shape)
+    m.data[:] = 0.
+
+    r = 2  # Constant for linear interpolation
```
georgebisbas, line 187: Duplicate?

georgebisbas (changes requested):

tests/test_interpolation.py
ccuetom: I have now moved the comments before the variable

tests/test_interpolation.py
ccuetom: I have removed the duplicate variable

FabioLuporini:
Thanks a lot, merging!

Issue #1696, opened 2021-05-20
Not-parallel loops marked as parallel leading to race-conditions


georgebisbas:
The MFE adapted from tests/test_dse.py::TestLifting::test_edge_cases is:
Run with: DEVITO_LANGUAGE=openmp DEVITO_ARCH=gcc

```python
import numpy as np
from devito import Eq, dimensions, TimeFunction, Operator

t, x, y = dimensions('t x y')

h1 = TimeFunction(name='h1', shape=(1, 3), dimensions=(t, y), time_order=0)
h1.data[0, :] = 0

op = Operator([Eq(h1[0, 0], y, implicit_dims=(t, x, y))])

op.apply()
print(h1.data[:, :]
```

the generated code is:
```c
  /* Begin section0 */
  START_TIMER(section0)
  #pragma omp parallel num_threads(nthreads)
  {
    #pragma omp for collapse(1) schedule(static,1)
    for (int y = y_m; y <= y_M; y += 1)
    {
      h1[0][0] = y;
    }
  }
  STOP_TIMER(section0,timers)
  /* End section0 */

```

Clearly the y loop is not parallel.

FabioLuporini:
does using `Inc` instead of `Eq` change anything?

georgebisbas:
Inc asks to define x_M. For some x_M code is:

```
40   START_TIMER(section0)
41   #pragma omp parallel num_threads(nthreads)
42   {
43     #pragma omp for collapse(2) schedule(static,1) reduction(+:h1[0][0])
44     for (int t = t_m; t <= t_M; t += 1)
45     {
46       for (int x = x_m; x <= x_M; x += 1)
47       {
48         #pragma omp simd aligned(h1:32)
49         for (int y = y_m; y <= y_M; y += 1)
50         {
51           h1[0][0] += y;
52         }
53       }
54     }
55   }
56   STOP_TIMER(section0,timers)
57   /* End section0 */
```


No race conds AFAICU

FabioLuporini:
Note that the MFE above is nonsensical even without parallelism. You would be overwriting the same entry in `h` with a different `y` value at every iteration. Can you make it "realistic enough"?

georgebisbas:
I agree with what you said. And we have also discussed that in the past. 
The question is whether we want the compiler to catch this whether unrealistic or not.
If it breaks... it breaks...
Shouldn't the compiler take care of a "bad" user?

Thoughts?

FabioLuporini:
[...]

PR #1697, opened 2021-05-24
compiler: Fixes #1695 by prioritising innermost vectorizable candidates

### LINKED ISSUE #1695: Compiler fails to correctly parallelize edge cases of imperfectly nested loops
The MFE adapted from `tests/test_dse.py::TestLifting::test_edge_cases` is:
Run with: `DEVITO_LANGUAGE=openmp DEVITO_ARCH=gcc`

```python
import numpy as np
from devito import Eq, dimensions, TimeFunction, Operator

t, x, y = dimensions('t x y')

g = TimeFunction(name='g', shape=(1, 3), dimensions=(t, x),
                 time_order=0, dtype=np.int32)
g.data[0, :] = [0, 1, 2]
h1 = TimeFunction(name='h1', shape=(1, 3), dimensions=(t, y), time_order=0)
h1.data[0, :] = 0

op = Operator([Eq(y.symbolic_max, g[0, x], implicit_dims=(t, x)),
              Eq(h1[0, y], y, implicit_dims=(t, x, y))])
op.apply()

```

the generated code is:
```c
  /* Begin section0 */
  START_TIMER(section0)
  {
    #pragma omp for collapse(1) schedule(static,1)
    for (int t = t_m; t <= t_M; t += 1)
    {
      #pragma omp simd aligned(g,h1:32)
      {
        int y_M = g[0][x];
        {
          h1[0][y] = y;
        }
      }
    }
  }
  STOP_TIMER(section0,timers)
  /* End section0 */

```

Clearly, loop y should have been the vectorized one.

[make_simd function](https://github.com/devitocodes/devito/blob/967267c7f131250a5f34156eac954dada44f314b/devito/passes/iet/parpragma.py#L42)

marks both x and y loops as vectorizable, but the Transformer prioritizes the x loop.

georgebisbas:
This PR is fixing issue #1695 . Though there are tests:
`tests/test_dse.py::TestLifting::test_edge_cases`
where race conditions occur due to #1696.
So this PR will either be enhanced with a solution for #1696, or should be considered to be merged after #1696 is solved.

UPDATE:
Now decoupled from #1696, can be separate PRs


tests/test_dle.py
```diff
@@ -616,6 +616,73 @@ def test_incs_no_atomic(self):
...
+         0),
+        (['Eq(y.symbolic_min, g[0, 2], implicit_dims=(t, x))',
+          'Eq(h1[0, x], y.symbolic_min, implicit_dims=(t, x))'],
+         0),
+        (['Eq(y.symbolic_min, g[0, x], implicit_dims=(t, x))',
+          'Eq(y.symbolic_max, g[0, x]-1, implicit_dims=(t, x))',
+          'Eq(h1[0, y], y, implicit_dims=(t, x, y))'],
+         0)
+    ])
+    def test_omp_pragmas_edge_cases(self, exprs, exp_simd):
```
FabioLuporini, line 666: nitpicking: "omp" redundant as you already are in a class testing openmp stuff. "pragmas" also unnecessary probably

devito/passes/iet/parpragma.py
```diff
@@ -55,6 +55,9 @@ def make_simd(self, iet):
             if not candidate.is_Parallel:
                 continue
 
+            if not IsPerfectIteration(depth=candidates[-2]).visit(candidate):
```
FabioLuporini, line 58: it deserves a comment

georgebisbas: TOADD small example

georgebisbas: Added

tests/test_dle.py
```diff
@@ -617,6 +617,73 @@ def test_incs_no_atomic(self):
...
+        (['Eq(y.symbolic_min, g[0, 0], implicit_dims=(t, x))',
+          'Inc(h1[0, y], x, implicit_dims=(t, x, y))'],
+         1),
+        (['Eq(y.symbolic_min, g[0, 2], implicit_dims=(t, x))',
+          'Inc(h1[0, x], y.symbolic_min, implicit_dims=(t, x))'],
+         0),
+        (['Eq(y.symbolic_min, g[0, 2], implicit_dims=(t, x))',
+          'Inc(h1[0, x], y.symbolic_min, implicit_dims=(t, x, y))'],
+         0),
+        (['Eq(y.symbolic_min, g[0, 2], implicit_dims=(t, x))',
```
FabioLuporini, line 659: this seems identical to the one starting at line 653 ? can u check them all, please?

FabioLuporini (changes requested):

tests/test_dle.py
georgebisbas: 654 is Inc(h1[0, x]), not the same, will check

mloubout (approved):

FabioLuporini (changes requested):
why would you not want to SIMD-ize outer iterations? 

OK, this is maybe my bad, there should be a test catching this...

However, consider the following scenario

```
for x  #PARALLEL
  for y  #PARALLEL
    for z  # SEQUENTIAL
  for y  #PARALLEL
    for z  # SEQUENTIAL
```

the z loops cannot be SIMD-ized
the y loops cannot be collapsed

so

you parallelise x
you SIMD-ize y

Now, this was supposed to give you a motivation behind allowing outer-loop vectorisation (and yes, I've seen it resulting in some sort of speedup with the intel compiler, clearly veeeery far from linear in the number of vector lanes because outer-product vectorisation introduces overhead for shuffles etc, but still, better than nothing...)

With this in mind, this PR is conceptually broken. Perhaps you need instead some sort of prioritisation strategy, just like the one we have for selecting OMP parloops. Just a thought, though...

georgebisbas:
> why would you not want to SIMD-ize outer iterations?
> 
> OK, this is maybe my bad, there should be a test catching this...
> 
> However, consider the following scenario
> 
> ```
> for x  #PARALLEL
>   for y  #PARALLEL
>     for z  # SEQUENTIAL
>   for y  #PARALLEL
>     for z  # SEQUENTIAL
> ```
> 
> the z loops cannot be SIMD-ized the y loops cannot be collapsed
> 
> so
> 
> you parallelise x you SIMD-ize y
> 
> Now, this was supposed to give you a motivation behind allowing outer-loop vectorisation (and yes, I've seen it resulting in some sort of speedup with the intel compiler, clearly veeeery far from linear in the number of vector lanes because outer-product vectorisation introduces overhead for shuffles etc, but still, better than nothing...)
> 
> With this in mind, this PR is conceptually broken. Perhaps you need instead some sort of prioritisation strategy, just like the one we have for selecting OMP parloops. Just a thought, though...

Sequential loops are already ruled out earlier:
```
            # Only fully-parallel Iterations will be SIMD-ized (ParallelRelaxed
            # might not be enough then)
            if not candidate.is_Parallel:
                continue
```
The new tree to be simdized will be a shortened one, and the candidates[-1] will be `y` loop. In other words:

```
> why would you not want to SIMD-ize outer iterations?
```
I am still SIMDizing outer.
```
> you parallelise x you SIMD-ize y
```
I am expecting this to work as you said.


```
Perhaps you need instead some sort of prioritisation strategy, just like the one we have for selecting OMP parloops. Just a thought, though.
```
Indeed in case both loops are vectorizable candidates, this PR prioritizes the inner one.

georgebisbas:
That means probably that my docstring is giving the wrong message. If you wrote the above by judging from the docstring, then you are absolutely right. I communicate the wrong message

FabioLuporini:
@georgebisbas ah, OK, I understand now

However, I wonder whether a much neater solution would be turning the tree search from

```
for tree in retrieve_iteration_tree(iet)
```

into

```
for tree in retrieve_iteration_tree(let, more='superset')
```

see [here](https://github.com/devitocodes/devito/blob/master/devito/ir/iet/utils.py#L38)

EDIT: well, maybe we would need a `subset` mode, but you get the principle -- don't we just need a search that retrieves only one iteration tree per innermost Iteration, or IOW, `len(innermost Iterations) == len(retrieved trees)`

georgebisbas:
> EDIT: well, maybe we would need a `subset` mode, but you get the principle -- don't we just need a search that retrieves only one iteration tree per innermost Iteration, or IOW, `len(innermost Iterations) == len(retrieved trees)`

I think there is a danger to miss vectorization opportunities with this. What if you get the `subset` and subset contains sequential innermost?
I think that your proposal here comes against this scenario:
```
for x  #PARALLEL
  for y  #PARALLEL
    for z  # SEQUENTIAL
  for y  #PARALLEL
    for z  # SEQUENTIAL
```

This COULD have worked, if we would also be iterating through candidates to find the next possible one. Since we are not doing this and only picking the last one per tree...maybe its overkill. (Just my two cents.)

tests/test_dse.py
```diff
@@ -258,9 +258,6 @@ class TestLifting(object):
         (['Eq(y.symbolic_max, g[0, x], implicit_dims=(t, x))',
          'Inc(h1[0, 0], 1, implicit_dims=(t, x, y))'],
          [6., 0., 0.]),
-        (['Eq(y.symbolic_max, g[0, x], implicit_dims=(t, x))',
```
FabioLuporini, line 261: why dropping this?

tests/test_dle.py
```diff
@@ -610,6 +610,76 @@ def test_incs_no_atomic(self):
         assert 'collapse(1)' in str(op1)
         assert 'atomic' not in str(op1)
 
+    @pytest.mark.parametrize('exprs,simd_level', [
```
FabioLuporini, line 613: out of curiosity, do we really need all these cases? or are some of them actually nearly redundant?

FabioLuporini (commented):
couple of minor questions, but otherwise this LGTM, thanks

FabioLuporini:
@georgebisbas : ironically, I just found out that this issue also crops up in

`test_dse.py::TestAliases::test_space_invariant_v3`

Obviously it's not caught there because that test makes other sort of checks. Could you check that this PR is catching the pattern in this use case too?

tests/test_dse.py
georgebisbas: This example is wrong in 2 ways: #1695 #1696.
It should not be tested. The result we get is `by miracle` correct because the two issues manage to auto-cancel each other.

tests/test_dle.py
georgebisbas: hm... I would not say that all of them are "difficult corner cases", however, they are all different.
Just added resulting data asserts as well.

georgebisbas: Dropped another that should be dropped.

georgebisbas:
> @georgebisbas : ironically, I just found out that this issue also crops up in
> 
> `test_dse.py::TestAliases::test_space_invariant_v3`
> 
> Obviously it's not caught there because that test makes other sort of checks. Could you check that this PR is catching the pattern in this use case too?

Yes, indeed this PR can catch this pattern. Added a test as well.

tests/test_dle.py
```diff
@@ -610,6 +611,99 @@ def test_incs_no_atomic(self):
...
+        assert 'omp for collapse' in iterations[0].pragmas[0].value
+        if simd_level:
+            assert 'omp simd' in iterations[simd_level].pragmas[0].value
+
+        op.apply()
+        assert (h1.data == expected).all()
+
+    def test_simd_space_invariant(self):
+        """
+        Similar to test_space_invariant_v3, testing simd vectorization happens
```
FabioLuporini, line 686: I _think_ we could rather add an extra check directly inside `test_space_invariant_v3` and decorate it with a comment along the lines of

```
# NOTE: checks issue #1695
```

instead of a nearly duplicate test? or alternatively, you could trim down `eq` -- I don't think you need all those terms to expose the issue, do you?

if you instead think this current way is better, that's fine then

tests/test_dse.py
FabioLuporini: OK

tests/test_dle.py
georgebisbas: > I _think_ we could rather add an extra check directly inside `test_space_invariant_v3` and decorate it with a comment along the lines of
> 
> ```
> # NOTE: checks issue #1695
> ```
> 
> instead of a nearly duplicate test? or alternatively, you could trim down `eq` -- I don't think you need all those terms to expose the issue, do you?
> 
> if you instead think this current way is better, that's fine then

test_dse does not have a single test for omp or simd checks so I guess we should not diverge for one case.
I trimmed down to a format that still covers the case. Hope it is better now.

FabioLuporini: LGTM

FabioLuporini (approved):
LGTM, thanks

PR #1702, opened 2021-06-07
compiler: Revamp data streaming

FabioLuporini:
This PR:

* Fixes a host-side memory leak when streaming data from the host to device. This was due to the pthreads allocating so-called pinned memory under-the-hood to perform the transfers; the issue was that this memory never got freed. Now such pinned memory regions are created by the parent thread, through the use of buffers (hence the significant changes to the `buffering` pass) and by requesting the OpenACC runtime to always allocate pinned memory (option `gpu=pinned` added to `NvidiaCompiler` class)
* Enables data streaming on AMD cards via OpenMP offloading (reusing part of the technology implemented for the bug fix above!)
* Reworks and improves various components revolving around the entire data streaming process
* Makes data streaming , and in particular host-to-device prefetching, more efficient
* Adds quite a few new data streaming tests


devito/passes/iet/parpragma.py
```diff
@@ -486,6 +495,27 @@ def _make_sections_from_imask(cls, f, imask):
...
+        assert len(imask) == len(datasize)
+        offset = 0
+        shape = []
+        for n, (i, j) in enumerate(zip(imask, datasize)):
+            if i is FULL:
+                shape.append(j)
+            else:
+                nofs, size = i
+                if isinstance(nofs, Mod):
+                    # NOTE: work around alleged SymPy bug which doesn't allow
```
mloubout, line 512: As said on slack, it's not a sympy bug, it's a choice of devito to go with `%` while sympy uses `fmod` which is a function call and therefore always has precedence over Mul.
I think it makes much more sense to fix `print_Mod` rather than adding this odd extra type.

mloubout (commented):
Except the Precedence thing I don't think makes sense, no other comments loks fine to me.

devito/passes/iet/parpragma.py
mloubout: It's actually very easy to fix if you use our`Mod` instead of the sympy one and add
```
@property
def precedence(self):
    return 49  # precedence(Mul)-1
```

since the first thing sympy checks is if it's defined before going to the defaults
https://github.com/sympy/sympy/blob/126f80578140e752ad5135aac77b8ff887eede3e/sympy/printing/precedence.py#L123

FabioLuporini: right. However, as explained on Slack, I forgot that this piece of code is actually dead, so I just dropped it

devito/passes/clusters/buffering.py
```diff
@@ -0,0 +1,602 @@
...
+
+    @cached_property
+    def initmap(self):
+        """
+        A mapper from contracted Dimensions to indices representing the min points
+        for buffer initialization. For example, in the case of a forward-propagating
+        `time` Dimension, we could have `{time: (time_m + db0) % 2, (time_m + db0)}`;
+        likewise, for backwards, `{time: (time_M - 2 + db0) % 4, time_M - 2 + db0}`.
+        """
+        mapper = {}
```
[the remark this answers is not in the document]
FabioLuporini: yeah, you're very right, but the gencode name convention and the variable name convention in the python code is sometimes different. I'll see what needs to be done to fix this

mloubout (approved):

PR #1703, opened 2021-06-09
gpu: Enable tile clause in place of collapse with OpenACC

FabioLuporini:
this should generally provide better performance, but there might be exceptions (it would appear 3D elastic being one of them)


georgebisbas (commented):
Maybe add a test with elastic-like structure...?

mloubout (approved):

georgebisbas (approved):

benchmarks/user/benchmark.py
```diff
@@ -1,19 +1,13 @@
-from collections import OrderedDict
-import sys
-
 import numpy as np
 import click
 import os
-from devito import (clear_cache, configuration, info, warning, set_log_level,
-                    switchconfig, norm)
+from devito import configuration, info, warning, set_log_level, switchconfig, norm
```
georgebisbas, line 4: ab-order ?

devito/passes/iet/languages/openacc.py
```diff
@@ -151,12 +155,32 @@ def _map_update_wait_device(cls, f, imask=None, queueid=None):
...
 
+    def _make_partree(self, candidates, nthreads=None):
+        assert candidates
+        root = candidates[0]
+
+        if self._is_offloadable(root) and self.par_tile:
+            if isinstance(self.par_tile, tuple):
+                tile = self.par_tile
+            else:
+                # (32,4,4,...) is typically a decent choice
```
georgebisbas, line 173: a decent choice in...V100, RTX3080 etc.. ?

tests/test_benchmark.py
```diff
@@ -8,10 +8,10 @@
 
 
 @pytest.mark.parametrize('mode, problem, op', [
-    ('bench', 'acoustic', 'forward'), ('run', 'acoustic', 'adjoint'),
-    ('run', 'acoustic', 'jacobian'), ('bench', 'acoustic', 'jacobian_adjoint'),
-    ('bench', 'tti', 'forward'), ('bench', 'elastic', 'forward'),
-    ('bench', 'viscoelastic', 'forward'), ('run', 'acoustic_sa', 'forward'),
+    ('run', 'acoustic', 'forward'), ('run', 'acoustic', 'adjoint'),
```
georgebisbas, line 11: ok

benchmarks/user/README.md
```diff
@@ -195,6 +190,12 @@ grid:
 ```
 python benchmark.py run -P tti -d 512 402 890 -so 12 -a basic --tn 100
 ```
+Same as before, but telling devito not to use temporaries to store the
+intermediate values which stem from mixed derivatives:
+```
+python benchmark.py run -P tti -d 512 402 890 -so 12 -a basic --tn 100 --opt
```
georgebisbas, line 196: Good

georgebisbas (commented):
Just an idea: Not sure if it is really needed, but why not add some tests to check norms between those two methods. Also we could add some notes for the 32,4,8 tile that this may or may not work in all GPUs. Also, shouldn't this PR include an additional entry (or two entries) in the GPU notebook? Actually no, the GPU notebook is for openmp, but we could show collapse vs tile generated code in one notebook...

examples/seismic/utils.py
```diff
@@ -193,6 +193,23 @@ def __call__(self, parser, args, values, option_string=None):
             values = {'float32': np.float32, 'float64': np.float64}[values]
             setattr(args, self.dest, values)
 
+    class _opt_action(Action):
```
mloubout, line 196: that's duplicate from benchmark, can we put it somewhere common?

FabioLuporini: they use differ arg parsers so I'm not sure?

mloubout: Hum slightly different reutrn type indeed, I guess it's fine

mloubout (approved):

PR #1706, opened 2021-06-14
examples: Switch to adjoint time derivative

mloubout:
Switch time derivative positon for the gradient as discussed on slack

tests pas locally so should be green.


FabioLuporini:
Since this has implications on GPU+data streaming, we should wait for #1702 to land first, which also "fixes" the current mismatching tag issue that is causing the omp offloading job to get scheduled to `wampa` rather than `kimogila`

FabioLuporini (approved):

mloubout:
sure

FabioLuporini:
Merged, thanks

PR #1708, opened 2021-06-15
dsl: Improve support for running operators concurrently in python threads

tjb900:
This change is in a few parts (so far), so that people can go away and run the test without the fix, and so on.

## Step 1: add the test

The first commit adds a test which runs the same, already-constructed-and-compiled, `Operator` on many threads simultaneously. This, on my system, fails with various issues mentioned in the subsequent sections.

## Step 2: add a mutex for initialising sympy objects

This addresses this error:

```
tests/test_threading.py:25: in do_run
    u_private = TimeFunction(name='u', grid=grid_private)
devito/types/basic.py:697: in __new__
    newobj = sympy.Function.__new__(newcls, *indices, **options)
/redacted/python3.9/site-packages/sympy/core/cache.py:74: in wrapper
    retval = func(*args, **kwargs)
/redacted/python3.9/site-packages/sympy/core/function.py:457: in __new__
    if n not in cls.nargs:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = Naturals0, other = 4

    def __contains__(self, other):
        other = _sympify(other)
        c = self._contains(other)
        b = tfn[c]
        if b is None:
            # x in y must evaluate to T or F; to entertain a None
            # result with Set use y.contains(x)
>           raise TypeError('did not evaluate to a bool: %r' % c)
E           TypeError: did not evaluate to a bool: None

/redacted/python3.9/site-packages/sympy/sets/sets.py:692: TypeError
```

## Step 3: use more thread-safe constructs when manipulating Devito's SymbolCache

This fixes this class of issue:

```
devito/types/dense.py:1389: in forward
    return self._subs(_t, _t + i * _t.spacing)
devito/finite_differences/differentiable.py:126: in _subs
    return self.func(*args, evaluate=False)
devito/types/basic.py:683: in __new__
    newobj.__init_cached__(obj)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = u(t + dt, x, y, z), key = u(t, x, y, z)

    def __init_cached__(self, key):
        """
        Initialise `self` with a cached object state.
    
        Parameters
        ----------
        key : object
            The cache key of the object whose state is used to initialize `self`.
            It must be hashable.
        """
>       self.__dict__ = _SymbolCache[key]().__dict__.copy()
E       KeyError: u(t, x, y, z)

devito/types/caching.py:98: KeyError
```

## Step 4: split the loop over the SymbolCache copy and the copy itself into separate lines

Very occasionally, when running the test in a loop for many hours, another strange error occurs which I don't currently have the backtrace for. This change is to provide better error context when that occurs, so we know if the issue arises in the `dict.copy()` or in the for loop itself.

## Step 5: a stab in the dark at fixing the error above

I actually don't think this is it, in retrospect, but since there were shenanigans in the behaviour of the python dictionary I went looking for possible causes. And it seems an implicitly good thing to ensure that `__eq__` and `__hash__` are mutually compatible (i.e. that `a == b` implies `hash(a) == hash(b)`).


tests/test_threading.py
```diff
@@ -0,0 +1,42 @@
+from concurrent.futures import ThreadPoolExecutor
+from devito import Operator, TimeFunction, Grid, Eq
+import numpy as np
+from threading import get_native_id
+
+
+def test_concurrent_excecuting_operators():
```
tjb900, line 7: - [ ] fix typo

devito/types/caching.py
```diff
@@ -85,17 +90,19 @@ def __init__(self, key, *aliases):
...
-    def __init_cached__(self, key):
+    def __init_cached__(self, cached_obj):
         """
         Initialise `self` with a cached object state.
 
+        "obj" must have been returned, non-None, from _cache_get
+
         Parameters
         ----------
         key : object
```
FabioLuporini, line 101: needs updating

devito/types/caching.py
```diff
@@ -165,6 +172,17 @@ def clear(cls, force=True):
...
+        #  during iteration" error we know which operation caused it
+        cache_copied = _SymbolCache.copy()
+        for key in cache_copied:
+            obj = _SymbolCache.get(key)
+            if obj is None:
+                # deleted by another thread
+                continue
             if obj() is None:
-                del _SymbolCache[key]
+                # does not error if already gone
```
FabioLuporini, line 187: either this comment, or the one above, you may want to mention that all of this is to support multi-threading

devito/types/dense.py
```diff
@@ -103,7 +103,8 @@ def __eq__(self, other):
 
     _subs = Differentiable._subs
 
-    __hash__ = AbstractFunction.__hash__  # Required since we're overriding __eq__
+    def __hash__(self):
+        return id(self)
```
FabioLuporini, line 107: ah, that's quite an interesting change

devito/types/parallel.py
```diff
@@ -250,11 +250,12 @@ def __str__(self):
 
     __repr__ = __str__
 
-    __hash__ = sympy.Basic.__hash__
-
     def __eq__(self, other):
         return type(self) == type(other) and self.args == other.args
 
+    def __hash__(self):
+        return hash(type(self), self.args)
```
FabioLuporini, line 257: good catch

FabioLuporini (approved):
This LGTM

FabioLuporini:
@mloubout also may want to look at these changes

devito/types/dense.py
tjb900: I don't think the previous implementation is problematic in this case actually, since `__eq__` is tweaked in this class to define equality to be true iff the two objects are actually the same object. Which in turn means any hash function should work fine - but in that case `id` is the cheapest since it doesn't need to do anything. I was just on a bit of a crusade to make the `__hash__` implementations match the equality tests in case that was causing an issue.

I should put it next to the `__eq__` override, actually, as it's much clearer that it's correct when viewed next to that.

Or can get rid of it. I think I've realised since making the change that it's not important.

tjb900:
Next error to fix, hit after looping on the test for ~8 hours:

```
tests/test_threading.py:47: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/redacted/python3.9/concurrent/futures/_base.py:445: in result
    return self.__get_result()
/redacted/python3.9/concurrent/futures/_base.py:390: in __get_result
    raise self._exception
/redacted/python3.9/concurrent/futures/thread.py:52: in run
    result = self.fn(*self.args, **self.kwargs)
tests/test_threading.py:29: in do_run
    u_private.data[:] = val
devito/types/dense.py:424: in data
    return self.data_domain
devito/types/dense.py:117: in wrapper
    CacheManager.clear(force=False)
devito/types/caching.py:165: in clear
    elif any(i.nbytes > cls.gc_ths for i in _SymbolCache.values()):
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

.0 = <dict_valueiterator object at 0x7f5b03142e50>

>   elif any(i.nbytes > cls.gc_ths for i in _SymbolCache.values()):
E   RuntimeError: dictionary changed size during iteration

devito/types/caching.py:165: RuntimeError
```

devito/types/caching.py
tjb900: done

devito/types/caching.py
tjb900: done

tjb900:
Above error case should be fixed now - that appears to be the one hit in CI also.

tjb900:
Once @mloubout is happy I'll consolidate a few of these commits, write the commit messages in an approved format, and force-push

tests/test_threading.py
```diff
@@ -0,0 +1,47 @@
...
+        val = current_thread().ident % 100000
+
+        grid_private = Grid(shape=shape)
+        u_private = TimeFunction(name='u', grid=grid_private)
+        u_private.data[:] = val
+
+        op(u=u_private, time_m=1, time_M=100)
+        assert np.all(u_private.data[1, :, :, :] == val + 100)
+
+    print("First running serially to demonstrate it works")
```
mloubout, line 34: use devito log instead

devito/types/caching.py
```diff
@@ -149,13 +159,21 @@ def clear(cls, force=True):
         sympy.polys.fields._field_cache.clear()
         sympy.polys.domains.modularinteger._modular_integer_cache.clear()
 
+        # Take a copy of the dictionary so we can safely iterate over it
+        # even if another thread is making changes
+
+        # mydict.copy() is safer than list(mydict) for getting an unchanging list
+        # See https://bugs.python.org/issue40327 for terrifying discussion
+        # on this issue.
+        cache_copied = _SymbolCache.copy()
```
mloubout, line 168: Isn't it potentially dangerous and may blow up if the cache contains large objects? 

From my understanding, the del/pop part is where we may try to delete already removed object, but it only needs the list of key, not the the full dictionary so  `_SymbolCache.keys()` may be the only thing needed to be copied then access `_SymbolCache[copied_key]`

mloubout (commented):
Nice, pretty cool addition and fix. Some small comments, mainly I would prefer to avoid making copy of the cache, but except that looks good.

tests/test_threading.py
tjb900: Will do 👍🏼

devito/types/caching.py
tjb900: This shouldn’t be too bad I think - as it’s a shallow copy. So won’t duplicate any large objects, just pointers to them. The issue with using keys() is that you have to iterate over it in a non- (less-) gil-safe manner, so we’d need to add more locking. Eg see the iteration over items which failed in CI above. Whereas copy is _mostly_ atomic (notwithstanding some  memory allocation operations which can apparently drop the gil).

mloubout: The memory is fine yes, but because we make a copy, don't we increase the reference count and prevent the actual cache clearing? I guess not since the test pass but seems risky to me, but if all is green I'm probably wrong.

tjb900: Hmmm. I’ll have another look tomorrow to be sure. I _think_ IIRC we’ll just take an extra reference to the weakref object, which won’t keep anything alive that wouldn’t otherwise be.

tjb900: Indeed, the contents of the cache dict are all `AugmentedWeakRef` objects, so I don't think this is an issue. The existence of the copied dict won't change any of the checks in the loop starting at (new) line 186.

tjb900:
Rebased against latest master, consolidated commits and updated commit messages to be in-spec.
Would you prefer I move the new test to after the fixes, to help with bisections?

devito/types/caching.py
FabioLuporini: before we had `list(...)` so wasn't it the same thing practically?

tjb900: Ah, yeah you had `list(x.items())`! So yes, actually we are not copying anything now that we were not copying before.

FabioLuporini (approved):

PR #1719, opened 2021-07-01
Generalised MatrixSparseTimeFunction

tjb900:
Main goal of this series is to introduce some changes to `MatrixSparseTimeFunction`, as described below. However, this necessitated some other work as well. 

The `ignore_dimsort` patch allows (ideally only internal) types to circumvent the requirement for the topological sort of all dimensions, as in some cases the presence of an `Indexed[dim1, dim2]` does not actually carry the requirement that the loops be ordered in that way. In particular here, the read-only coefficient arrays were the issue that broke the toposort, as the deterministic injection needs to access them in a different order than the interpolation - and resolving that would have involved duplicating the array.

This also augments the tests for `MatrixSparseTimeFunction` to cover some of the new cases.

commit 50247b29588e5fe8351d162e3e5853bf995c7770
Author: Tim Burgess <timb@dugeo.com>
Date:   Tue Jun 29 16:43:02 2021 +0800

    dse: add ignore_dimsort backdoor to Eq and Inc

    Currently, implicit_dims adds to the set of dimension relations
    that feed into the topological sort. Specifying this additional
    flag instead causes the implicit_dims to be used as _the_ answer,
    which allows devito-internal users (such as MatrixSparseTimeFunction)
    to override the toposort for cases where it cannot be resolved.

commit 3bbbc1632dd8f8f03fd930fc3f2c2910443fa3e4
Author: Tim Burgess <timb@dugeo.com>
Date:   Mon Jun 21 16:39:17 2021 +0800

    types/sparse: generalise MatrixSparseTimeFunction with non-cubic shapes and parallelism over not-first-dimension

    This augments MatrixSparseTimeFunction with some additional features:

     * non-cubic injection stencils. This means that we can have longer injection/interpolation stencils
       in some dimensions compared to others
     * relaxes the requirement for the length of stencil to be even
     * allows some stencil dimensions to have radius `None`, which indicates that the source/receiver
       points are not localised in this grid dimension, so there is no location element in that
       dimension, and the coefficients are declared over e.g. `x` rather than `rx`. The use case
       here is 2.5D propagation, where the grid dimension in question is actually a wavenumber.
     * the ability to specify which grid dimension is used for the parallelisation of the
       deterministic injection. Previously, this was locked to the first dimension of the
       `grid`. However, if the first dimension of the `grid` represents a wavenumber, then all
       sources interact with all wavenumbers, and so the efficiency of the parallel
       injection is reduced.

commit 251b8c68c9cd251d3603aa39df16b738d1d99dfd
Author: Fabio Luporini <f.luporini12@imperial.ac.uk>
Date:   Tue Jun 29 15:14:55 2021 +0200

    compiler: Patch CSE in presence of implicit dependencies


FabioLuporini (commented):
The only thing I don't like is the addition of a new attribute for ignoring the dimsort, which results in a cascade of additions throughout various classes. I think we could (should) use `implicit_dims`, thus generalising its semantics (we don't need to change the name as this would cause an API change, and therefore the need for deprecation etc etc -- a pain I'm not asking you to suffer)

All the rest, including the new tests and the edits to MSF, look good to me!

tjb900:
FWIW I did start to go down that path before, I just found it a bit ugly using type-overloading to pass an additional attribute that is kind of separate to the implicit dims themselves.

Anyway, have added commits with the requested change (revert of old one plus the new one) so you can compare. Will rebase out the commit/revert pair once review is done.

environment-dev.yml
```diff
@@ -3,6 +3,8 @@ channels:
   - defaults
   - conda-forge
 dependencies:
+  - openmpi
```
FabioLuporini, line 6: I think I'm OK with this, after all it's a dev install. @mloubout ?

FabioLuporini (approved):
GTG

environment-dev.yml
tjb900: Oh, my bad - didn't mean to commit that part. Will remove.

tjb900:
Have rebased against master, removed the environment.yml accidental change, removed the commit/revert pair and force-pushed.

FabioLuporini:
all good, thanks, merging!

PR #1720, opened 2021-07-02
mpi: Mitigate SparseFunction setup costs

mloubout:
Some better use of numpy broadcasting to setup sparse functions

I get a ~X5 speedup on 256*256 receivers on my desktop.

### DIFF
--- a/devito/mpi/distributed.py
+++ b/devito/mpi/distributed.py
@@ -10,7 +10,7 @@ from cgen import Struct, Value
 
 from devito.data import LEFT, CENTER, RIGHT, Decomposition
 from devito.parameters import configuration
-from devito.tools import EnrichedTuple, as_tuple, ctypes_to_cstr, is_integer
+from devito.tools import EnrichedTuple, as_tuple, ctypes_to_cstr, filter_ordered
 from devito.types import CompositeObject, Object
 
 
@@ -308,29 +308,27 @@ class Distributor(AbstractDistributor):
 
         Parameters
         ----------
-        index : int or list of ints
-            The index, or list of indices, for which the owning MPI rank(s) is
+        index : array of ints
+            The index, or array of indices, for which the owning MPI rank(s) is
             retrieved.
         """
-        assert isinstance(index, (tuple, list))
-        if len(index) == 0:
+        # assert isinstance(index, np.array)
+        if index.shape[0] == 0:
             return None
-        elif is_integer(index[0]):
-            # `index` is a single point
-            indices = [index]
-        else:
-            indices = index
-        ret = []
-        for i in indices:
-            assert len(i) == self.ndim
-            found = False
-            for r, j in enumerate(self.all_ranges):
-                if all(v in j[d] for v, d in zip(i, self.dimensions)):
-                    ret.append(r)
-                    found = True
-                    break
-            assert found
-        return as_tuple(ret)
+        elif sum(index.shape) == 1:
+            return index
+        assert index.shape[1] == self.ndim
+        ret = {}
+
+        for r, j in enumerate(self.all_ranges):
+            mins = np.array([b[0] for b in j]).reshape(1, -1, 1)
+            maxs = np.array([b[-1] for b in j]).reshape(1, -1, 1)
+            inds = np.where(((index <= maxs) & (index >= mins)).all(axis=1))
+            if inds[0].size == 0:
+                continue
+            ret[r] = filter_ordered(inds[0])
+        return ret
+
 
     @property
     def neighborhood(self):
--- a/devito/mpi/distributed.py
+++ b/devito/mpi/distributed.py
@@ -312,12 +312,19 @@ class Distributor(AbstractDistributor):
             The index, or array of indices, for which the owning MPI rank(s) is
             retrieved.
         """
-        # assert isinstance(index, np.array)
+        assert isinstance(index, np.ndarray)
         if index.shape[0] == 0:
             return None
         elif sum(index.shape) == 1:
             return index
+        print(index.shape, self.ndim)
         assert index.shape[1] == self.ndim
+
+        # Add singleton dimension at the end if only single gridpoint is passed
+        # instead of support.
+        if len(index.shape) == 2:
+            index = np.expand_dims(index, axis=2)
+
         ret = {}
 
         for r, j in enumerate(self.all_ranges):
@@ -328,7 +335,6 @@ class Distributor(AbstractDistributor):
                 continue
             ret[r] = filter_ordered(inds[0])
         return ret
-
 
     @property
     def neighborhood(self):
--- a/devito/mpi/distributed.py
+++ b/devito/mpi/distributed.py
@@ -317,7 +317,7 @@ class Distributor(AbstractDistributor):
             return None
         elif sum(index.shape) == 1:
             return index
-        print(index.shape, self.ndim)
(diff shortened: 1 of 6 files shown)


devito/types/sparse.py
```diff
@@ -709,22 +694,28 @@ def _dist_scatter(self, data=None):
         comm = distributor.comm
         mpitype = MPI._typedict[np.dtype(self.dtype).char]
 
+        # Compute dist map only once
+        dmap = self._dist_datamap
+
         # Pack sparse data values so that they can be sent out via an Alltoallv
-        data = data[self._dist_scatter_mask]
+        data = data[self._dist_scatter_mask(dmap=dmap)]
```
mloubout, line 701: @FabioLuporini @rhodrin This is basically 90-95% of the cost because we index a numpy array with individual indices so it's quite inefficient. But we can't really get out of it since we can't guaranty that the indices will be consecutive so gonna be really hard to get better.

mloubout: Same below for coords indexing.

devito/types/sparse.py
```diff
@@ -113,92 +113,81 @@ def inject(self, *args, **kwargs):
...
         injection/interpolation operators.
         """
-        ret = []
-        for i in self.gridpoints:
-            support = [range(max(0, j - self._radius + 1), min(M, j + self._radius + 1))
-                       for j, M in zip(i, self.grid.shape)]
-            ret.append(tuple(product(*support)))
-        return tuple(ret)
+        max_shape = np.array(self.grid.shape).reshape(1, self.grid.dim)
+        minmax = lambda arr: np.minimum(max_shape, np.maximum(0, arr))
```
georgebisbas, line 128: maybe change this name?

mloubout: why?

devito/mpi/distributed.py
```diff
@@ -308,29 +308,33 @@ def glb_to_rank(self, index):
...
-            for r, j in enumerate(self.all_ranges):
-                if all(v in j[d] for v, d in zip(i, self.dimensions)):
-                    ret.append(r)
-                    found = True
-                    break
-            assert found
-        return as_tuple(ret)
+        elif sum(index.shape) == 1:
+            return index
+        print(index.shape, self.ndim)
```
FabioLuporini, line 320: leftover

devito/mpi/distributed.py
```diff
@@ -308,29 +308,33 @@ def glb_to_rank(self, index):
...
+
+        ret = {}
+
+        for r, j in enumerate(self.all_ranges):
+            mins = np.array([b[0] for b in j]).reshape(1, -1, 1)
+            maxs = np.array([b[-1] for b in j]).reshape(1, -1, 1)
+            inds = np.where(((index <= maxs) & (index >= mins)).all(axis=1))
+            if inds[0].size == 0:
+                continue
+            ret[r] = filter_ordered(inds[0])
```
FabioLuporini, line 336: how big is `inds`? Haven't tried to catch up with the new logic, but if big, we may want to numpy-ize this too? if small, not an issue obv

devito/types/sparse.py
```diff
@@ -113,92 +113,81 @@ def inject(self, *args, **kwargs):
...
         Mapper ``M : MPI rank -> required sparse data``.
         """
-        ret = {}
-        support = support or self._support
-        for i, s in enumerate(support):
-            # Sparse point `i` is "required" by the following ranks
-            for r in self.grid.distributor.glb_to_rank(s):
-                ret.setdefault(r, []).append(i)
-        return {k: filter_ordered(v) for k, v in ret.items()}
+        dmap = self.grid.distributor.glb_to_rank(self._support)
```
FabioLuporini, line 137: single line `return self. ...` ?

devito/types/sparse.py
```diff
@@ -113,92 +113,81 @@ def inject(self, *args, **kwargs):
...
-    @property
-    def _dist_subfunc_scatter_mask(self):
+    def _dist_subfunc_scatter_mask(self, dmap=None):
         """
         This method is analogous to :meth:`_dist_scatter_mask`, although
         the mask is now suitable to index into self's SubFunctions, rather
         than into ``self.data``.
         """
-        return self._dist_scatter_mask[self._sparse_position]
+        return self._dist_scatter_mask(dmap=dmap)[self._sparse_position]
```
FabioLuporini, line 160: this is, essentially, to avoid recomputation of `_dist_datamap` when processing eg coordinates?

devito/mpi/distributed.py
```diff
@@ -308,29 +308,33 @@ def glb_to_rank(self, index):
...
+        print(index.shape, self.ndim)
+        assert index.shape[1] == self.ndim
+
+        # Add singleton dimension at the end if only single gridpoint is passed
+        # instead of support.
+        if len(index.shape) == 2:
+            index = np.expand_dims(index, axis=2)
+
+        ret = {}
+
```
FabioLuporini, line 329: no blank line

devito/mpi/distributed.py
```diff
@@ -308,29 +308,33 @@ def glb_to_rank(self, index):
...
+        for r, j in enumerate(self.all_ranges):
+            mins = np.array([b[0] for b in j]).reshape(1, -1, 1)
```
FabioLuporini, line 331: this list comprehension, and the one below... I'd guess they take up some time? does it get any better if you unzip ie

```
b = list(zip(*j))
c, d = b[0], b[-1]
```

just speculating here...

devito/types/sparse.py
mloubout: yes, it's called like 8 times so this gain quite a bit to do it only once.

devito/mpi/distributed.py
mloubout: No this is tiny only for loop over dimensions

devito/mpi/distributed.py
mloubout: `filter_ordered` is already numpy-ized so wouldn't really gain much

devito/mpi/distributed.py
```diff
@@ -308,29 +308,33 @@ def glb_to_rank(self, index):
...
         ----------
-        index : int or list of ints
-            The index, or list of indices, for which the owning MPI rank(s) is
+        index : array of ints
+            The index, or array of indices, for which the owning MPI rank(s) is
             retrieved.
         """
-        assert isinstance(index, (tuple, list))
-        if len(index) == 0:
+        assert isinstance(index, np.ndarray)
```
georgebisbas, line 315: Maybe this fixes the python 3.9 issue as well?

devito/types/sparse.py
georgebisbas: minmax is a bit ambiguous...?

FabioLuporini: it's the min of the maximum -- pretty clear to me, resolving

devito/mpi/distributed.py
FabioLuporini: aside from this print, is this PR basically good to go?

devito/mpi/distributed.py
mloubout: possibly, where is the issue for it again?

georgebisbas: I think no issue was opened, need to check

PR #1721, opened 2021-07-05
mpi: Towards compatibility for mode!=basic and GPU

FabioLuporini:


georgebisbas (approved):

devito/types/dense.py
```diff
@@ -1610,3 +1610,30 @@ def _arg_values(self, **kwargs):
...
+
+    is_AliasFunction = True
+
+    __indices_setup__ = Function.__indices_setup__
+    __shape_setup__ = Function.__shape_setup__
+
+    @property
+    def data(self):
+        # Any attempt at allocating data by the user should fail miserably
+        raise TypeError("TempFunction cannot allocate data")
```
mloubout, line 1634: AliasFunction

mloubout (commented):
What's with the imports changes?

FabioLuporini:
> What's with the imports changes?

circular dependency otherwise mpi->types->mpi IIRC

mloubout (approved):
Looks good.

Should we make `Alias/Temp` internal classes `_Alias/_Temp` just a though may be a pain to update.

FabioLuporini:
> Should we make Alias/Temp internal classes _Alias/_Temp just a though may be a pain to update.

I think you're right. I'm merging, but I'm gonna fix this now in #1703 . Not with the underscore, but by removing them from the `__all__`

FabioLuporini:
ah, no, I cannot merge because nexu is offline. So will fix it here now

FabioLuporini:
actually, that applies to AliasFunction, but not so sure about TempFunction. So I'll pull off AliasFunction for now

PR #1724, opened 2021-07-13
compiler: Add access linearization pass

FabioLuporini:
A pass to turn multi-dimensional array accesses into one-dimensional, or "flatten", array accesses. For example `a[i][j] -> a[i*n + j]`.

Sometimes it makes the code go faster, sometimes not, so it's off by default.


FabioLuporini:
closing. Will be superseded by another upcoming PR

PR #1726, opened 2021-07-14
Add cython and versioneer to requirements.txt

jkwashbourne:
Add cython and versioneer so `pip install` works without errors

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,3 +18,5 @@ distributed<2021.8
 pytest>=3.6,<7.0
 pytest-runner
 pytest-cov
+versioneer
+cython


jkwashbourne:
pyrevolve install fails. 
if you first install cython and versioneer pyrevolve install.

mloubout:
That's quite odd, I've install it from scratch in multiple docker images that have nothing and it installs fine. What version of python and such is it?

jkwashbourne:
@mloubout  `Python 3.6.10` 

Might be due to our conda install from inside julia (via adding our Devito.jl package). 

But `conda` and `versioneer` seem to be requirements for pyrevolve, correct? Maybe the issue is that `pyrevolve` only has `flake8` in `requirements.txt`

jkwashbourne:
moved to pyrevolve

PR #1736, opened 2021-08-12
compiler: Add optimization option to fuse WithLocks tasks

FabioLuporini:


tests/test_gpu_common.py
```diff
@@ -211,6 +211,56 @@ def test_tasking_unfused_two_locks(self):
...
+                Eq(v.forward, tmp1, subdomain=bundle0)]
+
+        op = Operator(eqns, opt=('tasking', 'fuse', 'orchestrate', {'fuse-tasks': True}))
+
+        # Check generated code
+        assert len(retrieve_iteration_tree(op)) == 5
+        assert len([i for i in FindSymbols().visit(op) if isinstance(i, Lock)]) == 2
+        sections = FindNodes(Section).visit(op)
+        assert len(sections) == 3
+        assert (str(sections[1].body[0].body[0].body[0].body[0]) ==
```
georgebisbas, line 238: Any good to create something like
retrieve_locks...? to avoid this lengthy search?

FabioLuporini: well, not sure & not easy -- here I need to check that the lock is actually placed at that specific depth in the tree

georgebisbas (approved):

FabioLuporini:
@mloubout any objections?

mloubout (approved):

PR #1743, opened 2021-09-01
install: Update to HPC SDK 21.7, Update to Jupyter>=3.0

kenhester:


requirements.txt
```diff
@@ -18,3 +18,4 @@ distributed<2021.9
 pytest>=3.6,<7.0
 pytest-runner
 pytest-cov
+fsspec
```
mloubout, line 21: what's this one?

FabioLuporini: this is a leftover, right? (@kenhester )

FabioLuporini:
superseded by #1760 (identical to this PR, but without the `fsspec` dependency, which turns out to be unnecessary

Many thanks @kenhester !

PR #1762, opened 2021-09-28
compiler: Patch and improve SubDomainSet

### LINKED ISSUE #1761: SubDomainSet broken (both functionally and performance)
Consider the following MFE:

```
        class DummySubdomains(SubDomainSet):
            name = 'dummydomain'
        dummy = DummySubdomains(N=1, bounds=(1, 1, 1, 1))

        grid = Grid(shape=(10, 10), subdomains=(dummy,))

        g = TimeFunction(name='g', grid=grid)
        theta = Function(name='theta', grid=grid)
        phi = Function(name='phi', grid=grid)

        eqns = [Eq(f.forward, f*sin(phi), subdomain=grid.subdomains['dummydomain']),
                Eq(g.forward, g*sin(theta), subdomain=grid.subdomains['dummydomain'])]

        op = Operator(eqns)
```

There are three issues:

1) functional: as @mloubout pointed out on Slack , this generates invalid code as the thickness assignments (eg `xi_n_ltkn = xi_n_m[n]`) are generated N times (with N number of equations, here N=2) . Not only are these assignments redundant, but also flawed since the first one assigns to a `const int` variable, which is later assigned to again, thus causing a compilation failure
2) the generation of redundant assignments prevents the fusion of the two eqns into the same loop (performance issue)
3) the hoisted temporaries are redundantly computed within the time loop, while they should go outside and be computed over the whole grid once and for all

FabioLuporini:
fixes #1761

### DIFF
--- a/devito/passes/clusters/misc.py
+++ b/devito/passes/clusters/misc.py
@@ -391,25 +391,33 @@ class MSDOptimizer(Queue):
             return clusters
 
         # Remove redundant thicknesses assignments
-        candidates = flatten(list(d._thickness_map) for d in msds)
-        schedulable = set(candidates)
+
+        thicknesses = set().union(*[list(i._thickness_map) for i in msds])
+        candidates = [c for c in clusters if set(c.scope.writes) & thicknesses]
+
+        # First of all, make sure we analyze all and only the thicknesses assignments
+        # at the same depth
+        d = prefix[-1].dim
+        if any(c.itintervals[-1].dim is not d for c in candidates):
+            return clusters
+
+        # Then, attempt extirpation of redundancies
+        schedulable = set(thicknesses)
         processed = []
         for c in clusters:
-            exprs = []
-
-            for e in c.exprs:
-                if e.lhs in candidates:
+            if c in candidates:
+                exprs = []
+                for e in c.exprs:
                     try:
                         schedulable.remove(e.lhs)
                         exprs.append(e)
                     except KeyError:
                         # Already scheduled, no-op
                         pass
-                else:
-                    exprs.append(e)
-
-            if exprs:
-                processed.append(c.rebuild(exprs=exprs))
+                if exprs:
+                    processed.append(c.rebuild(exprs=exprs))
+            else:
+                processed.append(c)
 
         # Sanity check
         assert len(schedulable) == 0
--- a/devito/passes/clusters/misc.py
+++ b/devito/passes/clusters/misc.py
@@ -7,9 +7,10 @@ from devito.passes.clusters.utils import cluster_pass
 from devito.symbolics import pow_to_mul
 from devito.tools import DAG, Stamp, as_tuple, flatten, frozendict, timed_pass
 from devito.types import Symbol
+from devito.types.grid import MultiSubDimension
 
 __all__ = ['Lift', 'fuse', 'optimize_pows', 'extract_increments',
-           'fission']
+           'fission', 'optimize_msds']
 
 
 class Lift(Queue):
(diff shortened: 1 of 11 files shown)


devito/types/grid.py
```diff
@@ -505,7 +527,88 @@ def define(self, dimensions):
...
+class MultiSubDimension(SubDimension):
+
+    """
+    A special SubDimension for graceful lowering of MultiSubDomains.
+    """
+
+    def __init_finalize__(self, name, parent):
+        lst, rst = self._symbolic_thickness(name)
+        left = parent.symbolic_min + lst
+        right = parent.symbolic_max - rst
```
mloubout, line 539: isn't there a -1 for the right part?

devito/types/grid.py
```diff
@@ -505,7 +527,88 @@ def define(self, dimensions):
...
+                bounds_m[j] = dec.index_glb_to_loc(m[j], LEFT)
+
+            if lmax < dec.loc_abs_min:
+                bounds_m[j] = dec.loc_abs_max
+                bounds_M[j] = dec.loc_abs_max
+                continue
+            elif lmax >= dec.loc_abs_max:
+                bounds_M[j] = 0
+            else:
+                bounds_M[j] = dec.index_glb_to_loc(M[j], RIGHT)
```
mloubout, line 587: I think this can simplify a bit 
```
           if lmin < dec.loc_abs_min:
                bounds_m[j] = 0
            elif lmin > dec.loc_abs_max:
                bounds_m[j] = dec.loc_abs_max
            else:
                bounds_m[j] = dec.index_glb_to_loc(m[j], LEFT)

            if lmax < dec.loc_abs_min:
                bounds_M[j] = dec.loc_abs_max
            elif lmax >= dec.loc_abs_max:
                bounds_M[j] = 0
            else:
                bounds_M[j] = dec.index_glb_to_loc(M[j], RIGHT)
```

and that should cover all the cases

mloubout (commented):
Some minor comment. Elastic on SubdomainSet still fails due to alternating subdomains, now the second loop over a subdomain doesn't have the bounds declaration. MFE:

```
from devito import * 

n = Dimension(name='n')
m = Dimension(name='m')

class DummySubdomains(SubDomainSet):
    name = 'dummydomain'
    implicit_dimension = m

dummy = DummySubdomains(N=1, bounds=(1, 1, 1, 1))

class DummySubdomains2(SubDomainSet):
    name = 'dummydomain2'
    implicit_dimension = n

dummy2 = DummySubdomains2(N=1, bounds=(1, 1, 1, 1))

grid = Grid(shape=(10, 10), subdomains=(dummy, dummy2))

g = TimeFunction(name='g', grid=grid)
theta = Function(name='theta', grid=grid)
phi = Function(name='phi', grid=grid)

eqns = [Eq(f.forward, f*sin(phi), subdomain=grid.subdomains['dummydomain']),
        Eq(g.forward, g*sin(theta), subdomain=grid.subdomains['dummydomain2']),
        Eq(f.forward, f*tan(phi), subdomain=grid.subdomains['dummydomain'])]

op = Operator(eqns)
op.cfunction
```

devito/types/grid.py
FabioLuporini: ah, this is just lifted from what is now a subclass -- this is code that @rhodrin wrote a time ago... did you see my ranting on slack about it? basically this would benefit from a rewrite regardless

devito/types/grid.py
FabioLuporini: I don't think so? this should behave like SubDimension.middle

mloubout: ok

devito/passes/clusters/misc.py
```diff
@@ -365,3 +366,68 @@ def fission(clusters):
...
+        if not prefix or any(isinstance(i.dim, MultiSubDimension) for i in prefix):
+            return clusters
+
+        msds = {d for d in set().union(*[c.dimensions for c in clusters])
+                if isinstance(d, MultiSubDimension)}
+        if not msds:
+            return clusters
+
+        # Remove redundant thicknesses assignments
+
```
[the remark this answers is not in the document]
FabioLuporini: no because the whole block of code below logically pertains to this comment

tests/test_dse.py
```diff
@@ -2565,7 +2565,7 @@ def test_fullopt(self):
 
         # Check expected opcount/oi
         assert summary[('section1', None)].ops == 90
-        assert np.isclose(summary[('section1', None)].oi, 1.508, atol=0.001)
+        assert np.isclose(summary[('section1', None)].oi, 1.511, atol=0.001)
```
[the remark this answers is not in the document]
FabioLuporini: good question. Look at the commit that introduces this change. Basically we're counting data space points along incr dimensions that we were skipping before. Just O(1) basically so you virtually never see an impact except in this test. Or at least, this is my best explanation

devito/types/grid.py
```diff
@@ -505,7 +527,88 @@ def define(self, dimensions):
...
+
+        return bounds_m, bounds_M
+
+    def __subdomain_finalize__(self, dimensions, shape, distributor=None, **kwargs):
+        """
+        Must be overridded by subclasses.
```
georgebisbas, line 593: *overridden

devito/types/grid.py
```diff
@@ -505,7 +527,88 @@ def define(self, dimensions):
...
+        """
+        raise NotImplementedError
+
+    def _create_implicit_exprs(self, grid):
+        """
+        Must be overridded by subclasses.
```
georgebisbas, line 599: same

devito/types/grid.py
```diff
@@ -505,7 +527,88 @@ def define(self, dimensions):
...
+        """
+        raise NotImplementedError
+
+    @property
+    def implicit_dimensions(self):
+        """
+        The Dimensions implicitly defined by the MultiSubDomain.
+        """
+        return ()
```
georgebisbas, line 608: needs completeness right...

tests/test_subdomains.py
```diff
@@ -427,3 +429,69 @@ class MySubdomains2(SubDomainSet):
                              [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]], dtype=np.int32)
 
         assert((np.array(f.data[:]+g.data[:]) == expected).all())
+
+    def test_issue_1761(self):
```
georgebisbas, line 433: Any test with N>1 ? Like N=3?

FabioLuporini: it doesn't matter, you can put N=1000. It's irrelevant for this issue/test. The only important thing are the *eqns* , the SubDomainSet is arbitrary

devito/types/grid.py
FabioLuporini: I think I'll fix it in #1764 to avoid a CI spin

PR #1764, opened 2021-09-30
compiler: Add machinery for custom memory allocators and MPI

FabioLuporini:


devito/passes/iet/definitions.py
```diff
@@ -260,17 +261,24 @@ def place_definitions(self, iet, **kwargs):
...
+    def derive_transfers(self, iet):
+        """
+        Collect all symbols that cause host-device data transfer, distinguishing
+        between reads and writes.
+        """
+        return ([], [])
+
     @iet_pass
-    def map_onmemspace(self, iet, **kwargs):
+    def place_transfers(self, iet, **kwargs):
```
georgebisbas, line 273: map_transfers?

FabioLuporini: no, it's more of a placement than a mapping. "map" is the process of binding (ie mapping) two things. Nitpicking, anyway ;)

devito/data/allocators.py
```diff
@@ -375,19 +375,37 @@ def alloc(self, shape, dtype):
...
 
 
-def default_allocator():
+def default_allocator(name=None):
     """
-    Return a suitable MemoryAllocator for the architecture on which the process
-    is running. Possible allocators are: ::
+    Return a MemoryAllocator for the underlying architecture.
 
+        * ALLOC_GUARD: Only used in so-called "develop mode", to trigger SIGSEGV as
```
georgebisbas, line 405: develop-mode or develop_mode just for 'grepping' homogeneity

georgebisbas (commented):
Any tests needed in this PR?

FabioLuporini:
> Any tests needed in this PR?

Basically no because I'm not adding any new functionalities

mloubout (approved):
Any easy way to add a test/example of custom allocator?

FabioLuporini:
> Any easy way to add a test/example of custom allocator?

not really... we could create a mapping for one of the existing ones... but what for?

FabioLuporini:
as in, just to clarify. All the code added by this PR **is** already tested automatically by the current tests, as the execution flow goes through all the touched classes and functions. That's why I wrote above that we're basically good to go

devito/types/grid.py
```diff
@@ -590,13 +590,13 @@ def _bounds_glb_to_loc(cls, dec, m, M):
 
     def __subdomain_finalize__(self, dimensions, shape, distributor=None, **kwargs):
         """
-        Must be overridded by subclasses.
+        Must be overridden by subclasses.
```
FabioLuporini, line 593: @georgebisbas as promised, fixed it here

PR #1770, opened 2021-10-05
[WIP, RFC] compiler: Add preliminary temporal blocking support

georgebisbas:
This PR adds temporal blocking optimisation to a limited class of mostly Jacobi-like operators.
It is still work in progress.

Supports auto-tuning.
Supports hierarchical blocking.

Not supporting sources/receivers.
Not supporting subdomains.
Not supporting tti-like sets of loops (back to back mutual read-write dependencies).


devito/passes/clusters/blocking.py
```diff
@@ -191,13 +215,48 @@ def decompose(ispace, d, block_dims):
     sub_iterators.pop(d, None)
     sub_iterators.update({bd: ispace.sub_iterators.get(d, []) for bd in block_dims})
 
+    # When blocking sequentially, update modulo subiterators with new parents
```
georgebisbas, line 218: I need to find better way

devito/passes/iet/misc.py
```diff
@@ -92,8 +92,24 @@ def relax_incr_dimensions(iet, **kwargs):
         # Get root's `symbolic_max` out of each outer Dimension
         roots_max = {i.dim.root: i.symbolic_max for i in outer}
 
+        # Get the sequential dimensions and the skewing dim (0 if not applicable) filter
+        # out sequential iterations from inner, outer, take care of skewing factor in
+        # outer and seq_dims
+        skew_inner = (inner[0].dim if inner[0].is_onlyAffineSequential else 0)
```
georgebisbas, line 98: I would prefer to have less code lines here: switch to some call to a function?

devito/passes/iet/misc.py
```diff
@@ -145,3 +164,89 @@ def is_on_device(obj, gpu_fit):
         return True
 
     return all(f in gpu_fit for f in fsave)
+
+
+def evalmin(a, b):
```
georgebisbas, line 169: Maybe move those to some util file?

tests/test_skewing.py
```diff
@@ -1,39 +1,41 @@
 import pytest
```
georgebisbas, line 1: Comment for the whole file: Will remove some tests  later, some can also be merged.

tests/test_skewing.py
```diff
@@ -154,25 +154,366 @@ def test_skewing_codegen(self, expr, expected, skewing, blockinner):
...
+        u.data[:] = init_value
+
+        op2 = Operator(eq0, opt=('advanced', {'openmp': True,
+                                              'wavefront': True, 'blocklevels': 2}))
+
+        op2.apply(time_M=nt, dt=dt, a=nu)
+        assert np.isclose(norm(u), norm_u, atol=1e-5, rtol=0)
+
+    @pytest.mark.parametrize('nt', [2, 24, 8, 17, 16, 19, 41])
+    @pytest.mark.parametrize('shape',
```
georgebisbas, line 445: "fuzzy testing" to cover corner cases for exotic min/max time/space block combos

PR #1777, opened 2021-10-13
compiler: Add evalmin, evalmax utilities

georgebisbas:
A small PR towards making MIN, MAX code easier to control (especially for future temporal blocking code).


FabioLuporini:
Merged, thanks

PR #1780, opened 2021-10-15
compiler: add relaxation template as part of IncrDimension

georgebisbas:
This PR moves the control of MIN/MAX bounds upon creation of an IncrDimension though:
```
    @cached_property
    def relaxed_max(self):
```
This way an IncrDimension holds info of its MIN/MAX bounds at cluster level.

At IET level, we only add the offsets through replace.

This helps to define (in future) more exotic patterns of blocking schemes at cluster level only.


georgebisbas:
Closing as discussed on weekly meeting @16.11.2021

PR #1789, opened 2021-11-01
compiler: Nested indexification

mloubout:
Fxes bug that comes with the test (fails to compile on master since the functuion `u1` only appear as an index and is never lowered and indexified)

### DIFF
--- a/devito/ir/equations/algorithms.py
+++ b/devito/ir/equations/algorithms.py
@@ -134,7 +134,7 @@ def lower_exprs(expressions, **kwargs):
             dimension_map = {}
 
         # Handle Functions (typical case)
-        mapper = {f: f.indexify(lshift=True, subs=dimension_map)
+        mapper = {f: lower_exprs(f.indexify(subs=dimension_map), **kwargs)
                   for f in retrieve_functions(expr)}
 
         # Handle Indexeds (from index notation)
@@ -160,6 +160,7 @@ def lower_exprs(expressions, **kwargs):
         mapper.update(dimension_map)
         # Add the user-supplied substitutions
         mapper.update(subs)
+        # Apply mapper to expression
         processed.append(uxreplace(expr, mapper))
 
     if isinstance(expressions, Iterable):
--- a/devito/types/basic.py
+++ b/devito/types/basic.py
@@ -1069,7 +1069,7 @@ class AbstractFunction(sympy.Function, Basic, Cached, Pickable, Evaluable):
         """
         return default_allocator().guaranteed_alignment
 
-    def indexify(self, indices=None, lshift=False, subs=None):
+    def indexify(self, indices=None, subs=None):
         """Create a types.Indexed from the current object."""
         if indices is not None:
             return Indexed(self.indexed, *indices)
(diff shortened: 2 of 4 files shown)


mloubout:
`eq0 = Eq(u0[u1,y], 2*u0)` well yeah but it's not very practical when you only care about the first dimension irrespective of the other ones or call `subs` on the equation directly to avoid multiple subs.

georgebisbas:
> `eq0 = Eq(u0[u1,y], 2*u0)` well yeah but it's not very practical when you only care about the first dimension irrespective of the other ones or call `subs` on the equation directly to avoid multiple subs.

I definitely agree. Maybe we should add like a test or two with more complex Eq ?

tests/test_operator.py
```diff
@@ -317,6 +317,26 @@ def test_nested_lowering(self):
...
+        grid = Grid(shape=(4, 4), dtype=np.int32)
+        x, y = grid.dimensions
+
+        u0 = Function(name="u0", grid=grid)
+        u1 = Function(name="u1", grid=grid, dtype=np.int32)
+
+        u0.data[:, :] = 2
+        u1.data[:, :] = 1
+
+        eq0 = Eq(u0._subs(x, u1), 2*u0)
```
FabioLuporini, line 333: why not `subs` instead of `_subs`?

georgebisbas (approved):

tests/test_operator.py
mloubout: Oh it does the same just quicker through our `_subs` that bypass some sympy check and stuff, just habit can switch it.

devito/ir/equations/algorithms.py
```diff
@@ -160,6 +160,10 @@ def lower_exprs(expressions, **kwargs):
         mapper.update(dimension_map)
         # Add the user-supplied substitutions
         mapper.update(subs)
+        # Apply mapper to itself for nested indexing
+        mapper = {k: uxreplace(v, mapper) if k.is_Function else v
```
FabioLuporini, line 164: does this work for doubly (or arbitrarily) nested Functions?

should this not become (somewhat) a recursive call?

going through the whole `mapper` here feels weird

mloubout: >should this not become (somewhat) a recursive call?

Yeah I think so this will only catch single recursion

devito/symbolics/manipulation.py
```diff
@@ -44,7 +44,7 @@ def _uxreplace(expr, rule):
     if expr in rule:
         v = rule[expr]
         if not isinstance(v, dict):
-            return v, True
+            return _uxreplace(v, rule)[0], True
```
mloubout, line 47: @FabioLuporini this makes the mapper processign nested so should handle any depth, will add a test for it.

devito/types/basic.py
```diff
@@ -1078,11 +1078,9 @@ def indexify(self, indices=None, lshift=False, subs=None):
         subs = subs or {}
         subs = [{**{d.spacing: 1, -d.spacing: -1}, **subs} for d in self.dimensions]
 
-        # Add halo shift
-        shift = self._size_nodomain.left if lshift else tuple([0]*len(self.dimensions))
```
FabioLuporini, line 1082: uh? 😮

mloubout: Nested lower_expr takes care of it through the indexed part of it

FabioLuporini: awesome!

FabioLuporini: doesn't it apply to `subs` as well?

FabioLuporini (commented):
this is a great change

does it improve or worsen the compilation speed? try with tti so=16?

devito/types/basic.py
mloubout: No subs is needed because of sudomain dimension map above, because `f.indexe` don't have a subdomain

FabioLuporini (approved):
Thanks, LGTM!

examples/seismic/tutorials/14_creating_synthetics.ipynb
```diff
@@ -30,6 +30,7 @@
     "    import gempy as gp\n",
     "except ModuleNotFoundError:\n",
     "    # Need to install these\n",
+    "    ! pip install aiohttp==3.7.1\n",
```
georgebisbas, line 33: how does that help?

georgebisbas: Any chance you checked whether now, we can have working version without installing specific gempy and specific pyvista? @EdCaunt ?

Issue #1788, opened 2021-11-01
Fix boolean masking on function data


EdCaunt:
Currently the following mfe throws an error:

```
import numpy as np
from devito import Grid, Function

shape = (5,)
extent = (6)

grid = Grid(shape=shape, extent=extent)

f = Function(name='f', grid=grid)

bool_arr = np.array([True, True, False, False, True])

f.data[bool_arr] = 1

print(f.data)
```

It should return [1, 1, 0, 0, 1], but instead you get the following traceback:
```
Traceback (most recent call last):
  File "/home/ed/Documents/FD_Code/mfes/function_data_boolean_indexing_mfe.py", line 13, in <module>
    f.data[bool_arr] = 1
  File "/home/ed/devito/devito/data/data.py", line 176, in wrapper
    return func(data, *args, **kwargs)
  File "/home/ed/devito/devito/data/data.py", line 263, in __setitem__
    loc_idx = self._index_glb_to_loc(glb_idx)
  File "/home/ed/devito/devito/data/data.py", line 435, in _index_glb_to_loc
    if NONLOCAL in loc_idx:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

EdCaunt:
Appears this is an issue in 3.9 only

PR #1790, opened 2021-11-01
mpi: Add MPI support for python3.9

georgebisbas:
Add to core as well.

### DIFF
--- a/.github/workflows/pytest-core-mpi.yml
+++ b/.github/workflows/pytest-core-mpi.yml
@@ -14,6 +14,9 @@ jobs:
   build:
     name: pytest-mpi
     runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        python-version: ['3.8','3.9']
 
     env:
       DEVITO_LANGUAGE: "openmp"
@@ -22,6 +25,10 @@ jobs:
       CXX: "g++-9"
 
     steps:
+    - uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+
     - name: Checkout devito
       uses: actions/checkout@v2
 
--- a/devito/data/data.py
+++ b/devito/data/data.py
@@ -432,7 +432,7 @@ class Data(np.ndarray):
             loc_idx.append(v)
 
         # Deal with NONLOCAL accesses
-        if NONLOCAL in loc_idx:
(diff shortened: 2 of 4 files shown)


.github/workflows/pytest-core-mpi.yml
```diff
@@ -14,6 +14,9 @@ jobs:
   build:
     name: pytest-mpi
     runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        python-version: ['3.9', '3.8']
```
FabioLuporini, line 19: this is doubling the cost of `pytest-core-mpi`, which is already quite significant.

I wonder whether we should forget about 3.8 for MPI testing @mloubout @rhodrin . Other options welcome

devito/data/data.py
```diff
@@ -432,7 +432,7 @@ def _index_glb_to_loc(self, glb_idx):
             loc_idx.append(v)
 
         # Deal with NONLOCAL accesses
-        if NONLOCAL in loc_idx:
+        if any(NONLOCAL is j for j in loc_idx):
```
FabioLuporini, line 435: `j is NONLOCAL ...` a bit cleaner

.github/workflows/pytest-core-mpi.yml
rhodrin: This will spawn a job for each concurrently, hence unless we're hitting the limit of runners we can use at any one time (which I don't think we currently are?) won't affect overall test times.

FabioLuporini: ah, right. How can we check?

.github/workflows/pytest-core-mpi.yml
```diff
@@ -22,6 +25,9 @@ jobs:
       CXX: "g++-9"
 
     steps:
+    - uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
```
mloubout, line 30: blank line

mloubout:
add the MFE from #1788  as a test

mloubout:
Reference issues it's fixing so that it closes it and we don't have to go by hand in it.

mloubout (approved):

FabioLuporini (approved):
LGTM, thanks

PR #1793, opened 2021-11-05
compiler: Move lambda level to IncrDimension property

georgebisbas:


mloubout (approved):

devito/passes/clusters/blocking.py
```diff
@@ -255,15 +263,15 @@ def callback(self, clusters, prefix):
...
-            # The level of a given Dimension in the hierarchy of block Dimensions, used
-            # to skew over the outer level of loops.
-            level = lambda dim: len([i for i in dim._defines if i.is_Incr])
-
             # Since we are here, prefix is skewable and nested under a
             # SEQUENTIAL loop.
             intervals = []
             for i in c.ispace:
-                if i.dim is d and level(d) <= 1:  # Skew only at level 0 or 1
+                # Skew if not Incr
```
georgebisbas, line 270: This is to avoid level property to be at SpaceDimension class. This way it can only be for IncrDimension.
We could have it higher in the hierarchy if DImension Classes and actually return 0 if no blocking in the game. This was the behaviour so far. Thoughts?

mloubout: I would probably have it at the highest level return 0 and only have IncrDimension return whatever is needed

georgebisbas: So like add in :

```python
class SpaceDimension:

    @cached_property
    def level(self):
        return 0
```
?

mloubout: All the way to `Dimension`

devito/types/dimension.py
```diff
@@ -1095,6 +1095,11 @@ def symbolic_incr(self):
         except (TypeError, ValueError):
             return self.step
 
+    @cached_property
+    def level(self):
+        # The level of an `IncrDimension` in the hierarchy of `IncrDimension`s
+        return len([i for i in self._defines if i.is_Incr])
```
FabioLuporini, line 1101: the level should trivially be defined as the depth in the hierarchy, and it could well be a DerivedDimension property (and Dimension.level returning 0 as Mathias was suggesting above)

mloubout (approved):

devito/types/dimension.py
```diff
@@ -445,6 +449,11 @@ def spacing(self):
     def _defines(self):
         return frozenset({self}) | self.parent._defines
 
+    @cached_property
+    def level(self):
+        # The depth in the hierarchy of `IncrDimension`s
+        return len([i for i in self._defines if i.is_Incr])
```
FabioLuporini, line 455: No, this is wrong. You don't need to search for `is_Incr` ... And it is wrong twice, because you're searching for a subclass in a superclass, and superclasses don't see subclasses. 

In fact, maybe we don't even need to override? how about just 

```
class Dimension(...)
     @property
     def level(self):
          return len(self._defines)
```

is this enough?

georgebisbas: Subdim blocking e.g. lots of tests in TestSubDimension. I guess obvious why.

For non subdims tests:

This is enough because in space blocking you do relative level comparison. So the returned value will be altered now but still valid as it is gonna be compared with another altered value.
SO like instead of having  
3>=2
in e.g.
```python
if i.dim.level >= bd.level:
```
we now gonna have
4>=3

Still valid, but different valued hierarchy

georgebisbas: Then we just the property from Derived to Incr  as before?

FabioLuporini: > . I guess obvious why.

can u elaborate pls?

georgebisbas: >. I guess obvious why.
> can u elaborate pls?

So there is not problem.
I apologize. It breaks only with skewing options, as the `relations` strategy is different. Will fix and recommit!

georgebisbas: So the issue is that in master skewing I need to know the absolute position in Incr-only hierarchy... the rest as I wrote on slack

devito/types/dimension.py
```diff
@@ -175,6 +175,10 @@ def min_name(self):
     def max_name(self):
         return "%s_M" % self.name
 
+    @cached_property
```
FabioLuporini, line 178: just property, obv

FabioLuporini:
Also, it should be a private property, not a public one.

devito/passes/clusters/blocking.py
```diff
@@ -247,23 +255,18 @@ def callback(self, clusters, prefix):
...
-            # The level of a given Dimension in the hierarchy of block Dimensions, used
-            # to skew over the outer level of loops.
-            level = lambda dim: len([i for i in dim._defines if i.is_Incr])
-
             # Since we are here, prefix is skewable and nested under a
             # SEQUENTIAL loop.
             intervals = []
             for i in c.ispace:
-                if i.dim is d and level(d) <= 1:  # Skew only at level 0 or 1
+                if i.dim is d and not d.is_Incr:  # Skew if non-Incr
```
FabioLuporini, line 267: how is this different than (as I said on Slack)

```
if i.dim is d and (not d.is_Incr or d._depth == 1):
    intervals.append(Interval(d, skew_dim, skew_dim))
else:
    intervals.append(i)
```

devito/types/dimension.py
```diff
@@ -1050,6 +1050,11 @@ def __init_finalize__(self, name, parent, _min, _max, step=None, size=None):
     def size(self):
         return self._size
 
+    @property
+    def _depth(self):
+        # The depth in the hierarchy of dimensions
```
FabioLuporini, line 1055:
```
"""
The depth of `self` in the hierarchy of IncrDimensions.
"""
```

devito/passes/clusters/blocking.py
georgebisbas: I was dumb and used `and` instead of `or` when discussing on slack.. thanks!

PR #1795, opened 2021-11-08
ci: add python 3.10, gcc-10 option

georgebisbas:
Tested python 3.10 via docker seems ok, just spinning a python 3.10 CI

### DIFF
--- a/.github/workflows/pytest-core-nompi.yml
+++ b/.github/workflows/pytest-core-nompi.yml
@@ -39,63 +39,63 @@ jobs:
         set: [base, adjoint]
         include:
         - name: pytest-ubuntu-py37-gcc5-omp
-          python-version: 3.7
+          python-version: '3.7'
           os: ubuntu-18.04
           arch: "gcc-5"
           language: "openmp"
           sympy: 1.8
 
         - name: pytest-ubuntu-py38-gcc6-omp
-          python-version: 3.8
+          python-version: '3.8'
           os: ubuntu-18.04
           arch: "gcc-6"
           language: "openmp"
           sympy: 1.7
 
         - name: pytest-ubuntu-py36-gcc7-omp
-          python-version: 3.6
+          python-version: '3.6'
           os: ubuntu-18.04
           arch: "gcc-7"
           language: "openmp"
           sympy: 1.8
 
         - name: pytest-ubuntu-py37-gcc7-noomp
-          python-version: 3.7
+          python-version: '3.7'
           os: ubuntu-18.04
           arch: "gcc-7"
           language: "C"
           sympy: 1.8
 
         - name: pytest-ubuntu-py38-gcc8-omp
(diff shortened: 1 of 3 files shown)


mloubout:
scipy doesn't support 3.10 and numpy only in some case so until these move towards it I wouldn't even look at it.

georgebisbas:
makes sense

mloubout:
Looks liek scipy got a new release 3 days ago that support 3.10

.github/workflows/pytest-core-nompi.yml
```diff
@@ -86,12 +86,12 @@ jobs:
           language: "C"
           sympy: 1.8
 
-        - name: pytest-docker-py36-gcc-omp
-          python-version: 3.6
-          os: ubuntu-18.04
+        - name: pytest-docker-py310-gcc-omp
```
mloubout, line 89: I wouldn't change the docker build we wanna keep a safe standard version for that one.

.github/workflows/pytest-core-nompi.yml
```diff
@@ -29,65 +29,72 @@ jobs:
...
            pytest-ubuntu-py39-gcc9-omp,
            pytest-osx-py37-clang-omp,
-           pytest-docker-py36-gcc-omp
+           pytest-docker-py37-gcc-omp
         ]
         set: [base, adjoint]
         include:
         - name: pytest-ubuntu-py37-gcc5-omp
-          python-version: 3.7
+          python-version: '3.7'
```
FabioLuporini, line 41: why do we need the '...' now?

.github/workflows/pytest-core-nompi.yml
```diff
@@ -29,65 +29,72 @@ jobs:
...
 
         - name: pytest-ubuntu-py39-gcc9-omp
-          python-version: 3.9
+          python-version: '3.9'
           os: ubuntu-20.04
           arch: "gcc-9"
           language: "openmp"
           sympy: 1.8
 
+        - name: pytest-ubuntu-py310-gcc10-omp
```
FabioLuporini, line 82: adding a new one, but not dropping an old one? is this PR replacing a job, or adding a new one?

FabioLuporini:
why is the docker build also changing?

.github/workflows/pytest-core-nompi.yml
georgebisbas: We discused in the meeting yesterday to maybe replace a 3.7 job, ` ' ' ` are needed from now for github actions.

georgebisbas: https://github.com/actions/setup-python/issues/160

georgebisbas:
> why is the docker build also changing?

It had wrong name before... now it is fixed. It is using 3.7

FabioLuporini (approved):
LTGM

PR #1801, opened 2021-11-23
gpu: Loop tiling for GPU

italoaug:
Enable code generation of loop tiling for GPU

### DIFF
--- a/tests/test_gpu_openacc.py
+++ b/tests/test_gpu_openacc.py
@@ -178,7 +178,7 @@ class TestOperator(object):
 
         assert np.all(np.array(u.data[0, :, :, :]) == time_steps)
 
-    def iso_acoustic(self, **opt_options):
+    def iso_acoustic(self, opt):
         shape = (101, 101)
         extent = (1000, 1000)
         origin = (0., 0.)
@@ -217,8 +217,7 @@ class TestOperator(object):
         src_term = src.inject(field=u.forward, expr=src * dt**2 / m)
         rec_term = rec.interpolate(expr=u.forward)
 
-        op = Operator([stencil] + src_term + rec_term, opt=('advanced', opt_options),
-                      language='openacc')
+        op = Operator([stencil] + src_term + rec_term, opt=opt, language='openacc')
 
         # Make sure we've indeed generated OpenACC code
         assert 'acc parallel' in str(op)
@@ -227,12 +226,12 @@ class TestOperator(object):
 
         assert np.isclose(norm(rec), 490.56, atol=1e-2, rtol=0)
 
-    @pytest.mark.parametrize('opt_options', [
-        {},
-        {'linearize': True},
+    @pytest.mark.parametrize('opt', [
+        'advanced',
+        ('blocking', {'linearize': True}),
     ])
-    def test_iso_acoustic(self, opt_options):
-        TestOperator().iso_acoustic(**opt_options)
+    def test_iso_acoustic(self, opt):
+        TestOperator().iso_acoustic(opt)
 
 
 class TestMPI(object):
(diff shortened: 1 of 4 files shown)


georgebisbas (commented):
This PR adds blocking code generation as a secondary option. i.e. not in the advanced pipeline but as a custom pass. I checked myself and seems to be doing what we want. Performance-wise there is no reason to be added in the advanced pipeline as it does not deliver performance. This is expected since nested parallelism pragmas need to be added. This can happen at later PR.

tests/test_gpu_openacc.py
```diff
@@ -227,12 +226,14 @@ def iso_acoustic(self, **opt_options):
 
         assert np.isclose(norm(rec), 490.56, atol=1e-2, rtol=0)
 
-    @pytest.mark.parametrize('opt_options', [
-        {},
-        {'linearize': True},
+    @pytest.mark.parametrize('opt', [
+        ('advanced', {}),
+        ('advanced', {'linearize': True}),
```
FabioLuporini, line 231: this one is probably overkill (never forget that running tests has a cost) provided that we keep the last one, ie `(blocking, {linearize=true})`

tests/test_gpu_openmp.py
```diff
@@ -314,12 +335,14 @@ def iso_acoustic(self, **opt_options):
 
         assert np.isclose(norm(rec), 490.55, atol=1e-2, rtol=0)
 
-    @pytest.mark.parametrize('opt_options', [
-        {},
-        {'linearize': True},
+    @pytest.mark.parametrize('opt', [
+        ('advanced', {}),
+        ('advanced', {'linearize': True}),
```
FabioLuporini, line 340: this one is probably overkill (never forget that running tests has a cost) provided that we keep the last one, ie (blocking, {linearize=true})

tests/test_gpu_openacc.py
```diff
@@ -227,12 +226,14 @@ def iso_acoustic(self, **opt_options):
...
+        ('blocking', ),
```
FabioLuporini, line 232: this one is probably overkill since blocking is already tested below in tandem with linearize

tests/test_gpu_openmp.py
```diff
@@ -314,12 +335,14 @@ def iso_acoustic(self, **opt_options):
...
+        ('blocking', ),
```
FabioLuporini, line 341: this one is probably overkill since blocking is already tested below in tandem with linearize

tests/test_gpu_openmp.py
```diff
@@ -341,4 +364,4 @@ def test_mpi_nocomms(self):
     @skipif('device-aomp')
     @pytest.mark.parallel(mode=[2, 4])
     def test_mpi_iso_acoustic(self):
-        TestOperator().iso_acoustic()
+        TestOperator().iso_acoustic(opt=('advanced', {}))
```
FabioLuporini, line 367: opt='advanced'

tests/test_gpu_openmp.py
```diff
@@ -314,12 +335,14 @@ def iso_acoustic(self, **opt_options):
 
         assert np.isclose(norm(rec), 490.55, atol=1e-2, rtol=0)
 
-    @pytest.mark.parametrize('opt_options', [
-        {},
-        {'linearize': True},
+    @pytest.mark.parametrize('opt', [
+        ('advanced', {}),
```
FabioLuporini, line 339: 'advanced' should suffice

FabioLuporini (approved):

FabioLuporini (changes requested):
The code is OK, but the commit history is not. There are two main issues:

* Far too many commits for just a few line edits. You'll need to squash away some of them
* The commit messages do not honor the format. See [here](https://github.com/devitocodes/devito/wiki/Tags-for-commit-messages-and-PR-titles)

FabioLuporini (approved):

FabioLuporini:
Merging, thanks

PR #1803, opened 2021-11-30
compiler: add nvidia gpu compute capability auto-detection

georgebisbas:
Add compute capability autodetection to Nvidia GPUs

### DIFF
--- a/devito/arch/archinfo.py
+++ b/devito/arch/archinfo.py
@@ -7,11 +7,12 @@ import cpuinfo
 import numpy as np
 import psutil
 import re
+import ctypes
 
 from devito.logger import warning
 from devito.tools import all_equal, memoized_func
 
-__all__ = ['platform_registry', 'get_cpu_info', 'get_gpu_info',
+__all__ = ['platform_registry', 'get_cpu_info', 'get_gpu_info', 'detect_cc',
            'Platform', 'Cpu64', 'Intel64', 'Amd', 'Arm', 'Power', 'Device',
            'NvidiaDevice', 'AmdDevice',
            'INTEL64', 'SNB', 'IVB', 'HSW', 'BDW', 'SKX', 'KNL', 'KNL7210',  # Intel
@@ -300,6 +301,41 @@ def get_gpu_info():
         pass
 
     return None
+
+
+@memoized_func
+def detect_cc():
+    CUDA_SUCCESS = 0
+
+    libnames = ('libcuda.so', 'libcuda.dylib', 'cuda.dll')
+    for libname in libnames:
+        try:
+            cuda = ctypes.CDLL(libname)
+        except OSError:
+            continue
+        else:
+            break
+    else:
+        raise OSError("could not load any of: " + ' '.join(libnames))
+
+    nGpus = ctypes.c_int()
+    cc_major = ctypes.c_int()
+    cc_minor = ctypes.c_int()
+    result = ctypes.c_int()
+    device = ctypes.c_int()
+    error_str = ctypes.c_char_p()
+
+    result = cuda.cuInit(0)
+    if result != CUDA_SUCCESS:
+        cuda.cuGetErrorString(result, ctypes.byref(error_str))
+        print("cuInit failed with error code %d: %s" % (result, error_str.value.decode()))
+        return 1
+    result = cuda.cuDeviceGetCount(ctypes.byref(nGpus))
+    print("Found %d device(s)." % nGpus.value)
+    for i in range(nGpus.value):
+        if (cuda.cuDeviceComputeCapability(ctypes.byref(cc_major),
+           ctypes.byref(cc_minor), device) == CUDA_SUCCESS):
+            return 10*cc_major.value + cc_minor.value
 
 
 @memoized_func
--- a/devito/arch/archinfo.py
+++ b/devito/arch/archinfo.py
@@ -12,7 +12,7 @@ import ctypes
 from devito.logger import warning
 from devito.tools import all_equal, memoized_func
 
-__all__ = ['platform_registry', 'get_cpu_info', 'get_gpu_info', 'detect_cc',
+__all__ = ['platform_registry', 'get_cpu_info', 'get_gpu_info', 'get_nvidia_cc',
            'Platform', 'Cpu64', 'Intel64', 'Amd', 'Arm', 'Power', 'Device',
            'NvidiaDevice', 'AmdDevice',
            'INTEL64', 'SNB', 'IVB', 'HSW', 'BDW', 'SKX', 'KNL', 'KNL7210',  # Intel
@@ -304,9 +304,7 @@ def get_gpu_info():
 
 
 @memoized_func
-def detect_cc():
-    CUDA_SUCCESS = 0
-
+def get_nvidia_cc():
     libnames = ('libcuda.so', 'libcuda.dylib', 'cuda.dll')
     for libname in libnames:
         try:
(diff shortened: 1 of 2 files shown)


devito/arch/archinfo.py
```diff
@@ -302,6 +303,41 @@ def homogenise_gpus(gpu_infos):
...
+    nGpus = ctypes.c_int()
+    cc_major = ctypes.c_int()
+    cc_minor = ctypes.c_int()
+    result = ctypes.c_int()
+    device = ctypes.c_int()
+    error_str = ctypes.c_char_p()
+
+    result = cuda.cuInit(0)
+    if result != CUDA_SUCCESS:
+        cuda.cuGetErrorString(result, ctypes.byref(error_str))
```
georgebisbas, line 330: toremove multiple gpus

devito/arch/archinfo.py
```diff
@@ -302,6 +303,41 @@ def homogenise_gpus(gpu_infos):
...
+        print("cuInit failed with error code %d: %s" % (result, error_str.value.decode()))
+        return 1
+    result = cuda.cuDeviceGetCount(ctypes.byref(nGpus))
+    print("Found %d device(s)." % nGpus.value)
+    for i in range(nGpus.value):
+        if (cuda.cuDeviceComputeCapability(ctypes.byref(cc_major),
+           ctypes.byref(cc_minor), device) == CUDA_SUCCESS):
+            return 10*cc_major.value + cc_minor.value
```
georgebisbas, line 338: can be done better (?)

devito/arch/archinfo.py
```diff
@@ -302,6 +303,41 @@ def homogenise_gpus(gpu_infos):
...
+
+    result = cuda.cuInit(0)
+    if result != CUDA_SUCCESS:
+        cuda.cuGetErrorString(result, ctypes.byref(error_str))
+        print("cuInit failed with error code %d: %s" % (result, error_str.value.decode()))
+        return 1
+    result = cuda.cuDeviceGetCount(ctypes.byref(nGpus))
+    print("Found %d device(s)." % nGpus.value)
+    for i in range(nGpus.value):
+        if (cuda.cuDeviceComputeCapability(ctypes.byref(cc_major),
```
georgebisbas, line 336: update deprecated

devito/arch/compiler.py
```diff
@@ -397,7 +397,7 @@ def __init__(self, *args, **kwargs):
             # Add flags for OpenMP offloading
             if language in ['C', 'openmp']:
                 # TODO: Need a generic -march setup
-                # self.cflags += ['-Xopenmp-target', '-march=sm_37']
+                self.cflags += ['-Xopenmp-target', '-march=sm_'+str(detect_cc())]
```
georgebisbas, line 400: join str

devito/arch/archinfo.py
```diff
@@ -302,6 +303,41 @@ def homogenise_gpus(gpu_infos):
...
+        try:
+            cuda = ctypes.CDLL(libname)
+        except OSError:
+            continue
+        else:
+            break
+    else:
+        raise OSError("could not load any of: " + ' '.join(libnames))
+
+    nGpus = ctypes.c_int()
```
georgebisbas, line 321: rmv

devito/arch/compiler.py
FabioLuporini: as you know, this is not the way we do string concatenation in devito

devito/arch/compiler.py
```diff
@@ -396,8 +396,10 @@ def __init__(self, *args, **kwargs):
             self.cflags.remove('-std=c99')
             # Add flags for OpenMP offloading
             if language in ['C', 'openmp']:
-                # TODO: Need a generic -march setup
-                # self.cflags += ['-Xopenmp-target', '-march=sm_37']
+                # Autodetect GPU compute capability
+                if detect_cc():
+                    self.cflags += ['-Xopenmp-target', ''.join(['-march=sm_',
```
mloubout, line 401: why do you need `join`. just `'-march=sm_%s' % detect_cc()` should work no

devito/arch/archinfo.py
```diff
@@ -302,6 +303,38 @@ def homogenise_gpus(gpu_infos):
...
+        else:
+            break
+    else:
+        # warning("could not load any of: " + ' '.join(libnames))
+        return None
+
+    cc_major = ctypes.c_int()
+    cc_minor = ctypes.c_int()
+    result = ctypes.c_int()
+    device = ctypes.c_int()
```
mloubout, line 325: I'm not sure I get the subtlety between these four but seems needed.

devito/arch/archinfo.py
```diff
@@ -302,6 +303,38 @@ def homogenise_gpus(gpu_infos):
...
+    libnames = ('libcuda.so', 'libcuda.dylib', 'cuda.dll')
+    for libname in libnames:
+        try:
+            cuda = ctypes.CDLL(libname)
+        except OSError:
+            continue
+        else:
+            break
+    else:
+        # warning("could not load any of: " + ' '.join(libnames))
```
mloubout, line 319: remove

devito/arch/archinfo.py
```diff
@@ -302,6 +303,38 @@ def homogenise_gpus(gpu_infos):
     return None
 
 
+@memoized_func
+def detect_cc():
```
FabioLuporini, line 307: this is an NVidia-only function, so I'd name it differently to denote that. For example, `get_nvidia_cc` ?

devito/arch/archinfo.py
FabioLuporini: wdym? the called underlying function expects three arguments and returns one value, all of type c_int, so in total we need four values, no?

devito/arch/archinfo.py
```diff
@@ -302,6 +303,38 @@ def homogenise_gpus(gpu_infos):
...
+    device = ctypes.c_int()
+    error_str = ctypes.c_char_p()
+
+    result = cuda.cuInit(0)
+    if result != CUDA_SUCCESS:
+        cuda.cuGetErrorString(result, ctypes.byref(error_str))
+        # print("cuInit failed with error code %d: %s" % (result, error_str.value.decode()))
+        return None
+    if (cuda.cuDeviceComputeCapability(ctypes.byref(cc_major),
+       ctypes.byref(cc_minor), device) == CUDA_SUCCESS):
```
FabioLuporini, line 334: this is not how we indent code in devito, would be surprised if flake8 wouldn't throw an error

devito/arch/compiler.py
FabioLuporini: indeed...

FabioLuporini (changes requested):

devito/arch/compiler.py
FabioLuporini: also, it's quite ugly, as you're calling `detect_cc` twice, relying on the assumption that it gets memorized

And, finally, we want `'-Xopenmp-target'` to always be added at least, no? so that we'll try to compile, even if the CC couldn't be detected for whatever reason.

FabioLuporini (changes requested):

devito/arch/compiler.py
georgebisbas: >And, finally, we want '-Xopenmp-target' to always be added at least, no? so that we'll try to compile, even if the CC couldn't be detected for whatever reason.

We were not using it before, why should we do now?

devito/arch/archinfo.py
georgebisbas: Yes

devito/arch/archinfo.py
georgebisbas: Yes, my bad

devito/arch/archinfo.py
georgebisbas: We need them yes.

devito/arch/archinfo.py
FabioLuporini: what is the call to `cuda.cuGetErrorString` for, if you don't actually use the return value `error_str` ?

devito/arch/compiler.py
```diff
@@ -396,8 +396,10 @@ def __init__(self, *args, **kwargs):
             self.cflags.remove('-std=c99')
             # Add flags for OpenMP offloading
             if language in ['C', 'openmp']:
-                # TODO: Need a generic -march setup
-                # self.cflags += ['-Xopenmp-target', '-march=sm_37']
+                # Autodetect GPU compute capability
```
FabioLuporini, line 399: drop comment? quite self-explanatory I'd say

devito/arch/archinfo.py
georgebisbas: Mainly for debugging reasons if needed in the future. I am happy to drop. Let me know your preference.

devito/arch/compiler.py
georgebisbas: ok

mloubout (approved):

devito/arch/archinfo.py
georgebisbas: Better now?

italoaug (approved):
Looks good

FabioLuporini (commented):
I think it's OK now. @mloubout ?

FabioLuporini (approved):

PR #1806, opened 2021-12-15
[WIP] install: Add Intel dockerfile

georgebisbas:
This PR adds a dockerfile that uses standalone intel compilers
Tested on test_operator.py


docker/third-party-programs.txt
```diff
@@ -0,0 +1,1403 @@
+Intel(R) oneAPI Toolkits Third Party Programs File
```
FabioLuporini, line 1: what's this file for?

FabioLuporini (commented):
Thanks for this PR!

Complexity is increasing... this should be a discussion item for our next group meeting on Tuesday (@mloubout )

docker/third-party-programs.txt
georgebisbas: It is needed if we are to pull from Intel dockerfile directly. If we want to get rid of it we may need to explcitly add each itel dockerfile command and remove the command needed for this file...

In other words: we pull from this file, and need to handle this line somehow: https://github.com/intel/oneapi-containers/blob/master/images/docker/hpckit/Dockerfile.ubuntu-18.04#L35

FabioLuporini (changes requested):
TODO

- split `pytest-core-nompi` into `pytest-core-nompi-local` (better name?) and `pytest-core-nompi-docker`
- split `tutorials` into `tutorials-local` (better name?) and `tutorials-docker`
- Factorize common docker instruction into parent Dockerfile (example: the following lines are copy-paste...)

```
ADD ./requirements.txt /app/requirements.txt
ADD ./requirements-optional.txt /app/requirements-optional.txt
ADD ./requirements-mpi.txt /app/requirements-mpi.txt

RUN python3 -m venv /venv
RUN /venv/bin/pip install --no-cache-dir --upgrade pip
RUN /venv/bin/pip install --no-cache-dir jupyter
RUN /venv/bin/pip install --no-cache-dir wheel
RUN /venv/bin/pip install --no-cache-dir -r /app/requirements.txt
RUN /venv/bin/pip install --no-cache-dir -r /app/requirements-optional.txt
RUN . /opt/intel/oneapi/mpi/latest/env/vars.sh && /venv/bin/pip install --no-cache-dir -r /app/requirements-mpi.txt
RUN rm -rf ~/.cache/pip

...
```


- [optional?] inheritance from a thinner image? `intel/oneapi-hpckit:latest` contains *everything* 
- drop `third-part-programs.txt` somehow

.github/workflows/docker-publish.yml
```diff
@@ -95,6 +95,50 @@ jobs:
             devitocodes/devito:gpu-latest
             devitocodes/devito:gpu-${{ github.event.release.tag_name }}
 
+  deploy-docker-intel:
```
mloubout, line 98: Can't this be easily be just added as a different case in a `matrix` rather than duplicate all these steps just for a dockerfile name?

docker/Dockerfile.intel
```diff
@@ -0,0 +1,90 @@
...
+#
+# RUN a few tests:
+#   docker run --rm --name testrun 'devito/intel:1.0' python -m pytest -m "not parallel" tests/test_operator.py
+#
+##############################################################
+# Inherited from: https://github.com/intel/oneapi-containers/blob/master/images/docker/hpckit/Dockerfile.ubuntu-18.04
+# Copyright (c) 2019-2020 Intel Corporation.
+# SPDX-License-Identifier: BSD-3-Clause
+
+FROM intel/oneapi-hpckit:latest
```
mloubout, line 21: This and language are the oly difference with the normal dockerfile. Would be better to be merged and handled via cmd args. i.e first line having
```
ARG BASE=python:3.7
FROM $BASE
```

and call it with `build-args:` in the CI file. (same for DEVITO_ARCH)

docker/Dockerfile.intel
```diff
@@ -0,0 +1,90 @@
...
+
+# Update toolkit packages first (existing sources in image)
+RUN apt-get update -y && \
+    apt-get install -y --no-install-recommends -o=Dpkg::Use-Pty=0
+
+RUN apt-get update -y && \
+    apt-get install -y --no-install-recommends -o=Dpkg::Use-Pty=0 \
+    apt-utils \
+    python3-dev \
+    python3-venv
```
mloubout, line 31: intel hpckit doesn't have python :O

docker/Dockerfile.intel
```diff
@@ -0,0 +1,90 @@
...
+    chown app:app /docker-entrypoint.sh && \
+    chown app:app /print-defaults && \
+    chown app:app /tests && \
+    chown app:app /jupyter
+
+# Change to the app user.
+USER app
+WORKDIR /app/devito
+
+ARG ICC_HOME=/opt/intel/oneapi/compiler/latest/
```
mloubout, line 85: This should not be needed that's what `/opt/intel/..../setvars.sh` is for and should b in entrypoint

.github/workflows/docker-publish.yml
FabioLuporini: agree

.github/workflows/pytest-core-nompi.yml
```diff
@@ -43,55 +44,84 @@ jobs:
...
+          image: "devito_img"
+          test_suite: "full"
+
+        - name: pytest-docker-py38-icc-omp
+          python-version: '3.8'
+          os: ubuntu-18.04
+          arch: "icc"
+          language: "openmp"
+          sympy: 1.8
+          container: true
```
FabioLuporini, line 121: what's the purpose of the container flag? just set `dockerfile: ""` in the other entries, no?

docker/Dockerfile.intel
```diff
@@ -0,0 +1,90 @@
+##############################################################
```
FabioLuporini, line 1: there's a massive amount of code duplication here -- sigh

as I pointed out in Italo's PR for OpenMP offloading, which should rather establish a hierarchy of dockerfiles and reuse code as much as possible

we don't duplicate code -- ever !

docker/Dockerfile.intel
FabioLuporini: ah, lol, I just wrote about this code duplication issue a few lines above

docker/Dockerfile.intel
mloubout: ok

FabioLuporini:
`docker: squash` is not an acceptable commit message

docker/Dockerfile.intel
```diff
@@ -0,0 +1,53 @@
...
+
+FROM devitocodes/devito:cpu-dev
+
+USER root
+
+# Update toolkit packages first (existing sources in image)
+RUN apt-get update -y && \
+    apt-get install -y --no-install-recommends -o=Dpkg::Use-Pty=0 \
+    apt-utils \
+    ncurses-term
```
georgebisbas, line 29: could be dropped

docker/Dockerfile.intel
georgebisbas: will need to educate myself a bit for this

.github/workflows/pytest-core-nompi.yml
```diff
@@ -43,55 +44,83 @@ jobs:
           arch: "gcc-5"
           language: "openmp"
           sympy: "1.7"
+          dockerfile: ''
+          container: false
```
FabioLuporini, line 48: drop

.github/workflows/pytest-core-nompi.yml
```diff
@@ -43,55 +44,83 @@ jobs:
...
+          test_suite: "reduced"
 
         - name: pytest-docker-py37-gcc-omp
           python-version: '3.7'
           os: ubuntu-18.04
           arch: "gcc"
           language: "openmp"
           sympy: "1.10"
+          dockerfile: "docker/Dockerfile"
+          image: "devito_img"
```
FabioLuporini, line 112: shouldn't this just be "latest" @mloubout @georgebisbas ? the tag should be uniform no?

georgebisbas:
Superseded by #1946 .

PR #1809, opened 2021-12-16
examples: add adjoint, born, gradient, checkpointing to the tti example

mloubout:
@FabioLuporini now supports same command line args as acoustic.


FabioLuporini (approved):
Perfect. GTG

Note: this would benefit from a common template shared by all seismic propagators, since it's basically the same code everywhere (anyway, nitpicking, since it's "just" an example)

mloubout:
> this would benefit from a common template shared by all seismic propagators

Yeah thought about that. Don't have much of an opinion but I think it 's a bit more readable as a standalone example like that. Maybe next time.

PR #1811, opened 2021-12-16
compiler: Define Iteration bounds through hierarchy-related rules

georgebisbas:
Definitely needs some improvement in terms of presenting better the theoretic background of it.

Main concept : This is a system of rules to handle iteration bounds in a parametric way based upon known assumptions of their relations. 

Clearly there are more lines of code, and considering that devito supports only one type of blocking shapes this may seem an overkill. However this opens up space for new blocking scheme implementations where more complex iteration bounds are needed.

Opening now, to get reviewers familiar with the reasoning and the code.

### DIFF
--- a/devito/passes/iet/misc.py
+++ b/devito/passes/iet/misc.py
@@ -106,7 +106,7 @@ def relax_incr_dimensions(iet, **kwargs):
         roots_max = {i.dim.root: i.symbolic_max for i in outer}
 
         # Process inner iterations and adjust their bounds
-        for n, i in enumerate(inner):
+        for i in inner:
             # This part of the code aims to collect the data necessary to build the
             # MIN/MAX bounds of an Iteration. The main challenge is to correctly
             # take care of interval offsets (incurred by several Devito optimizations),
@@ -138,18 +138,18 @@ def relax_incr_dimensions(iet, **kwargs):
             # Step 2: Generate assumptions deriving from hierarchical blocking. Assuming
             # lower blocking levels perfectly fit within outer levels, we use depth of
             # IncrDimensions as a rule to generate assumptions such as:
-            # Defines are all the ancestors of `i.dim`. Starting from the parent up the
-            # root.
             # e.g. for 2-level blocking [x (inner Incr), x0_blk1, x0_blk0, x (outer root)]
             # [x0_blk1 > x0_blk0, x0_blk1_size > x0_blk0_size]
+            # `defines` keeps all the ancestors and descendants of `i.dim`.
             defines = sorted(i.dim._defines, key=lambda x: (not x.is_Incr or -x._depth))
-            # Collect assumptions
+
+            # Collect assumptions based on hierarchy
             acs = [j for j in defines if j is not i.dim and j.is_Incr]
             assumptions = []
             assumptions.extend(chain.from_iterable((Gt(j.step, k.step), Gt(j, k)) for j, k
                                in product(acs, acs) if j._depth > k._depth))
 
-            # Reduce defines min/max candidates using assumptions on defines list
+            # Step 3: Reduce defines min/max candidates using assumptions on `defines`
             defmin = evalrel(max, defines, assumptions, eval=False)
             defmax = evalrel(min, defines, assumptions, eval=False)
 
@@ -157,38 +157,30 @@ def relax_incr_dimensions(iet, **kwargs):
             defmin = [j.symbolic_min for j in defmin]
             defmax = [j.symbolic_max for j in defmax]
 
-            # [x, x0_blk0, x0_blk0, x]
-            # [x0_blk1 + x0_blk1_size - 1, x_M, x_M, x_M]
-
             # Drop duplicates from defmin/defmax and keep them Ordered
             defmin = list(OrderedDict.fromkeys(defmin))
             defmax = list(OrderedDict.fromkeys(defmax))
 
-            # Drop a candidate if it is the symbolic_min/max of other
-            # symbol candidates
+            # Step 4: Drop a candidate if it is the symbolic_min/max of symbol candidates
             defmin = [j for j in defmin if j not in (k.symbolic_min for
                       k in defmin if k.is_Symbol and not k.is_Scalar)]
             defmax = [j for j in defmax if j not in (k.symbolic_max for
                       k in defmax if k.is_Symbol and not k.is_Scalar)]
-            # [x0_blk1]
-            # [x0_blk1 + x0_blk1_size - 1, x_M, x_M, x_M]
-            # [x0_blk1 + x0_blk1_size - 1, x_M]
 
             # At this point the candidates for the start and end of the `Iteration`
-            # comprise of candidates stemming from i.dim and candidates stemming from
-            # other dims, (in space blocking other dims is usually root.) The terms
-            # stemming from 'i.dim' can be compared and for those we need the minimum
-            # of them for the start and the maximum of them for the end so as to span
-            # the whole extent of the iteration.
+            # comprise of candidates stemming (i) from `i.dim` and (ii) candidates
+            # stemming from other dims, (in space blocking other dims is usually root.)
+            # The canidates stemming from 'i.dim' can be compared and for those we need
+            # the minimum of them for the start and the maximum of them for the end so
+            # as to span the whole extent of the Iteration's Dimension (i.dim iteration
+            # extent).
+            # We split on comparable and uncomparable. Comparable ones include
+            # subdimensions (offsets incurred) and non-comparable ones include
+            # dimensions with no clear rule of simplification
             # For the start it is MAX(uncomparable, MIN(comparable))
-            # For the end it is MAX(uncomparable, MIN(comparable))
-            # Comparable ones include subdimensions(offsets)
+            # For the end it is MIN(uncomparable, MAX(comparable))
             defmin = evalrel(min, defmin, assumptions=None, eval=False)
             defmax = evalrel(max, defmax, assumptions=None, eval=False)
-
-            # Split on comparable and uncomparable
-            # Comparable ones include subdimensions(offsets)
-            # Non-comparable ones include dimensions with no clear rule of simplification
 
             # Apply offset if needed, else leave as it is
             defmin = [j.subs(min_map) if j in min_map else j for j in as_list(defmin)]
@@ -198,6 +190,9 @@ def relax_incr_dimensions(iet, **kwargs):
             iter_min = evalrel(max, defmin)
             iter_max = evalrel(min, defmax)
 
+            # Final iteration form will look like:
+            # Iteration i <starting from MAX(uncomparable, MIN(comparable))
+            # to MIN(uncomparable, MAX(comparable))>
             mapper[i] = i._rebuild(limits=(iter_min, iter_max, i.step))
 
     if mapper:
--- a/devito/passes/iet/misc.py
+++ b/devito/passes/iet/misc.py
@@ -7,7 +7,7 @@ from devito.ir import (Forward, List, Prodder, FindNodes, Transformer,
 from devito.passes.iet.engine import iet_pass
 from devito.symbolics import MIN, MAX, reduce_relation, rfunc
 from devito.types.relational import Gt
-from devito.tools import is_integer, split, as_list
+from devito.tools import split, as_list
 
 __all__ = ['avoid_denormals', 'hoist_prodders', 'relax_incr_dimensions']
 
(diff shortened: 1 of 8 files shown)


devito/types/relational.py
```diff
@@ -9,6 +9,7 @@ class AbstractRel(object):
     """
```
FabioLuporini, line 9: and one between the class and the docstring!!

devito/symbolics/manipulation.py
```diff
@@ -334,6 +334,9 @@ def evalrel(func=min, input=None, assumptions=None):
     assumptions : list
         A list of assumptions formed as relationals between candidates, assumed to be
         True. Defaults to None.
+    eval : bool, optional
```
FabioLuporini, line 337: maybe have two functions, `reduce_relation` and `eval_relation`, where the latter essentially calls the former plus one extra line for evaluation?

I don't like much two different output types here

devito/passes/iet/misc.py
```diff
@@ -104,9 +113,78 @@ def relax_incr_dimensions(iet, **kwargs):
...
+            defmin = [j.symbolic_min for j in defmin]
+            defmax = [j.symbolic_max for j in defmax]
+
+            # Drop duplicates from defmin/defmax and keep them Ordered
+            defmin = list(OrderedDict.fromkeys(defmin))
+            defmax = list(OrderedDict.fromkeys(defmax))
+
+            # Step 4: Drop a candidate if it is the symbolic_min/max of symbol candidates
+            defmin = [j for j in defmin if j not in (k.symbolic_min for
+                      k in defmin if k.is_Symbol and not k.is_Scalar)]
```
FabioLuporini, line 157: > if k.is_Symbol and not k.is_Scalar

ugly (to say the least)

devito/symbolics/manipulation.py
georgebisbas: DOne

devito/passes/iet/misc.py
```diff
@@ -104,9 +113,88 @@ def relax_incr_dimensions(iet, **kwargs):
...
+                    pass
+
+            for k in defmax.copy():
+                try:
+                    if k.symbolic_max in defmax:
+                        defmax.remove(k.symbolic_max)
+                except:
+                    pass
+
+            # At this point the candidates for the start and end of the `Iteration`
```
FabioLuporini, line 170: Writing comments is an art (and I'm bad at it)... but isn't this way too long?

FabioLuporini: sometimes *tiny* examples are enough

devito/passes/iet/misc.py
```diff
@@ -104,9 +113,88 @@ def relax_incr_dimensions(iet, **kwargs):
...
+            # lower blocking levels perfectly fit within outer levels, we use depth of
+            # IncrDimensions as a rule to generate assumptions such as:
+            # e.g. for 2-level blocking [x (inner Incr), x0_blk1, x0_blk0, x (outer root)]
+            # [x0_blk1 > x0_blk0, x0_blk1_size > x0_blk0_size]
+            # `defines` keeps all the ancestors and descendants of `i.dim`.
+            defines = sorted(i.dim._defines, key=lambda x: (not x.is_Incr or -x._depth))
+
+            # Collect assumptions based on hierarchy
+            acs = [j for j in defines if j is not i.dim and j.is_Incr]
+            assumptions = []
```
FabioLuporini, line 139: no need?

devito/passes/iet/misc.py
```diff
@@ -104,9 +113,88 @@ def relax_incr_dimensions(iet, **kwargs):
             # `i.dim.symbolic_max = x0_blk0 + x0_blk0_size - 1` then the generated
             # maximum will be `MIN(x0_blk0 + x0_blk0_size + 1, x_M + 2)`
 
+            # Step 1: Interval offset care
+            # Here we create two dictionaries to keep track of offsets incurring in
+            # 'i' and 'i.root' iteration
+            min_map = {}  # For the Iteration minimum
```
FabioLuporini, line 119: just 

```
min_map = {
     i.dim.root.symbolic_min: root_min,
     i.dim.symbolic_min: i.symbolic_min
}
```

isn't it more readable like this?

FabioLuporini: regardless, this definition line alone isn't really useful

devito/passes/iet/misc.py
```diff
@@ -104,9 +113,88 @@ def relax_incr_dimensions(iet, **kwargs):
...
+
+            # Collect symbolic_min, symbolic_max from defmin/defmax lists
+            defmin = [j.symbolic_min for j in defmin]
+            defmax = [j.symbolic_max for j in defmax]
+
+            # Drop duplicates from defmin/defmax and keep them Ordered
+            defmin = list(OrderedDict.fromkeys(defmin))
+            defmax = list(OrderedDict.fromkeys(defmax))
+
+            # Step 4: Drop a candidate if it is the symbolic_min/max of symbol candidates
```
FabioLuporini, line 155: I'm not sure I understand what you're trying to do here, or better, why?

devito/passes/iet/misc.py
```diff
@@ -104,9 +113,88 @@ def relax_incr_dimensions(iet, **kwargs):
...
+            # For the end: MIN(uncomparable, MAX(comparable)) --> MIN(uncomparable)
+            defmin = reduce_relation(min, defmin, assumptions=None)
+            defmax = reduce_relation(max, defmax, assumptions=None)
+
+            # Apply offset if needed, else leave as it is
+            defmin = [j.subs(min_map) if j in min_map else j for j in as_list(defmin)]
+            defmax = [j.subs(max_map) if j in max_map else j for j in as_list(defmax)]
+
+            # Final MIN/MAX processing
+            iter_min = evaluate_relation(max, defmin)
```
FabioLuporini, line 191: so you eventually call `reduce_relation` 3 times for MIN (and 3 times for MAX).

Why do you need 3 times for MIN (MAX) and not just one? that's really really odd at first glance

3 because one is buried inside `evaluate_relation`

FabioLuporini (changes requested):

devito/passes/iet/misc.py
georgebisbas: >regardless, this definition line alone isn't really useful

Which line? Do not really get it. (?)

devito/passes/iet/misc.py
georgebisbas: yes, but I will add as_list to next line.

devito/passes/iet/misc.py
georgebisbas: The first one is applied with assumptions on defines (no min max).
Second one is applied without assumptions on min/max.
The last one is indeed redundant, just removed.

devito/passes/iet/misc.py
FabioLuporini: ping

devito/passes/iet/misc.py
FabioLuporini: ping

devito/passes/iet/misc.py
```diff
@@ -90,10 +93,16 @@ def relax_incr_dimensions(iet, **kwargs):
         outer, inner = split(iterations, lambda i: not i.dim.parent.is_Block)
 
         # Get root's `symbolic_max` out of each outer Dimension
+        roots_min = {i.dim.root: i.symbolic_min for i in outer}
         roots_max = {i.dim.root: i.symbolic_max for i in outer}
 
         # Process inner iterations and adjust their bounds
-        for n, i in enumerate(inner):
+        for i in inner:
+            # This part of the code aims to collect the data necessary to build the
```
FabioLuporini, line 101: This whole comment can probably be dropped

devito/passes/iet/misc.py
```diff
@@ -104,9 +113,89 @@ def relax_incr_dimensions(iet, **kwargs):
             # `i.dim.symbolic_max = x0_blk0 + x0_blk0_size - 1` then the generated
             # maximum will be `MIN(x0_blk0 + x0_blk0_size + 1, x_M + 2)`
 
+            # Step 1: Interval offset care
```
FabioLuporini, line 116: drop "care"? it sounds weird

devito/passes/iet/misc.py
```diff
@@ -104,9 +113,89 @@ def relax_incr_dimensions(iet, **kwargs):
...
+            # the minimum of them for the start and the maximum of them for the end so
+            # as to span the whole extent of the Iteration's Dimension (i.dim iteration
+            # extent).
+            # We split on comparable and uncomparable. Comparable ones can be further
+            # simplfied, (subdimensions, offsets incurred) and non-comparable ones
+            # include candidates that cannot be simplified more (no clear rule of
+            # further simplification exists)
+            # For the start: MAX(uncomparable, MIN(comparable)) --> MAX(uncomparable)
+            # For the end: MIN(uncomparable, MAX(comparable)) --> MIN(uncomparable)
+            defmin = reduce_relation(min, defmin, assumptions=None)
```
FabioLuporini, line 184: I still don't understand why we need 2x2 calls to `reduce_relation` rather than 2x1 (ie once for MIN and once for MAX)

Please write example

devito/passes/iet/misc.py
georgebisbas: Improved

georgebisbas:
>IOW we need to add that if a > c, b > d then a+b>c+d.

I realised that there is a bit of redundant codegen in some subdimension code (no tests or tutorials catch it). The above addition will fix it.
Focusing on this.

devito/passes/iet/misc.py
```diff
@@ -90,23 +93,96 @@ def relax_incr_dimensions(iet, **kwargs):
...
-            # Iteration e.g. `x0_blk0 + x0_blk0_size - 1` and (b) the `symbolic_max`
-            # of the current Iteration's root Dimension e.g. `x_M`. The generated
-            # maximum will be `MIN(x0_blk0 + x0_blk0_size - 1, x_M)
+        for i in inner:
+            # Usually the Iteration's maximum is the MIN of (a) the `symbolic_max` of
+            # current Iteration e.g. `x0_blk0 + x0_blk0_size - 1` and (b) the
+            # `symbolic_max` of the current Iteration's root Dimension e.g. `x_M`. The
+            # generated maximum will be `MIN(x0_blk0 + x0_blk0_size - 1, x_M)
 
             # In some corner cases an offset may be added (e.g. after CIRE passes)
```
FabioLuporini, line 106: these are not corner cases

devito/passes/iet/misc.py
```diff
@@ -90,23 +93,96 @@ def relax_incr_dimensions(iet, **kwargs):
...
+
             root_max = roots_max[i.dim.root] + i.symbolic_max - i.dim.symbolic_max
-            iter_max = evalrel(min, [i.symbolic_max, root_max])
-            mapper[i] = i._rebuild(limits=(i.symbolic_min, iter_max, i.step))
+            max_map = {
+                i.dim.root.symbolic_max: root_max,
+                i.dim.symbolic_max: i.symbolic_max
+            }
+
+            # Step 2: Generate assumptions deriving from hierarchical blocking. Assuming
```
FabioLuporini, line 126: Note that more than "assuming", it is "by construction" (ie, user should get an exception if he/she tries to set the mid BlockDimension to anything that is not a perfect division of the parent block size... something like this. I hope this is enforced somewhere, perhaps at args processing time?)

georgebisbas: This is kinda not-related to this part of code? I mean it is something irrelevant of this PR. (right?)
So maybe it just deserves another small PR?

devito/passes/iet/misc.py
```diff
@@ -90,23 +93,96 @@ def relax_incr_dimensions(iet, **kwargs):
...
         roots_max = {i.dim.root: i.symbolic_max for i in outer}
 
         # Process inner iterations and adjust their bounds
-        for n, i in enumerate(inner):
-            # The Iteration's maximum is the MIN of (a) the `symbolic_max` of current
-            # Iteration e.g. `x0_blk0 + x0_blk0_size - 1` and (b) the `symbolic_max`
-            # of the current Iteration's root Dimension e.g. `x_M`. The generated
-            # maximum will be `MIN(x0_blk0 + x0_blk0_size - 1, x_M)
+        for i in inner:
+            # Usually the Iteration's maximum is the MIN of (a) the `symbolic_max` of
```
FabioLuporini, line 101: Why "usually" ?

devito/passes/iet/misc.py
```diff
@@ -90,23 +93,96 @@ def relax_incr_dimensions(iet, **kwargs):
...
+            # current Iteration e.g. `x0_blk0 + x0_blk0_size - 1` and (b) the
+            # `symbolic_max` of the current Iteration's root Dimension e.g. `x_M`. The
+            # generated maximum will be `MIN(x0_blk0 + x0_blk0_size - 1, x_M)
 
-            # In some corner cases an offset may be added (e.g. after CIRE passes)
+            # There are cases were optimizations may add an offset (e.g. CIRE passes)
```
FabioLuporini, line 106: Why "There are cases"... I mean, it's always the MIN of the same two quantities, but one quantity carries implicitly an offset, no...?

That is, the way it's written here, it's like there are two different ways of computing it

FabioLuporini (changes requested):

georgebisbas:
Pushed branch to georgebisbas/devito.

PR #1814, opened 2021-12-22
compiler: gpu cc detection fix

georgebisbas:
` cuda.cuInit(0) ` needs to be called before cuda.cuDeviceComputeCapability and not inside.
Tests were passing accidentally before. Now it should be fine before the offloading dockerfile comes in.


italoaug (approved):
I've tried it in a VM, and it did show the expected behaviour.

FabioLuporini (approved):
OK, thanks, merging

PR #1815, opened 2021-12-24
builtins: Add devito version of NumPy's count_nonzero, nonzero

georgebisbas:
Add a Devito version of NumPy's count_nonzero, nonzero:
https://numpy.org/doc/stable/reference/generated/numpy.count_nonzero.html
This PR is needed towards implementing steps 1,2 of p4 in https://arxiv.org/pdf/2010.10248.pdf.
In the paper, I was using NumPy, but now to automate through Devito, I would like to generate it through Devito.
Numpy's version is faster unless we change to collapse(1). Then Devito is faster for like int32 (20000, 20000) sized functions.
If reviewers require more benchmarking, happy to do so.

Generated code is like:
```
  /* Begin section0 */
  START_TIMER(section0)
  {
    #pragma omp for collapse(2) schedule(static,1) reduction(+:h1[1])
    {
      {
        if (f[x + 1][y + 1] != 0)
        {
          int r0 = 1;
          h1[1] += r0;
        }
      }
    }
  }
  STOP_TIMER(section0,timers)
  /* End section0 */

  return 0;
```

### DIFF
--- a/devito/builtins/arithmetic.py
+++ b/devito/builtins/arithmetic.py
@@ -4,7 +4,7 @@ import devito as dv
 from devito.builtins.utils import MPIReduction
 
 
-__all__ = ['norm', 'sumall', 'inner', 'mmin', 'mmax']
+__all__ = ['norm', 'sumall', 'inner', 'mmin', 'mmax', 'count_nonzero']
 
 
 @dv.switchconfig(log_level='ERROR')
@@ -161,3 +161,21 @@ def mmax(f):
         return mr.v.item()
     else:
         raise ValueError("Expected Function, not `%s`" % type(f))
+
+
+def count_nonzero(f):
+    """
+    Retrieve the count of nonzero elements of a (Time)Function.
+
+    Parameters
+    ----------
+    f : array_like or Function
+        Input operand.
+    """
+    ci = dv.ConditionalDimension(name='ci', parent=f.dimensions[-1],
+                                 condition=dv.Ne(f, 0))
+    g = dv.Function(name='g', shape=(1,), dimensions=(ci,), dtype=np.int32)
+    eq0 = dv.Inc(g[0], 1, implicit_dims=(f.dimensions + (ci,)))
+    op0 = dv.Operator([eq0])
+    op0.apply()
+    return g.data[0]
--- a/devito/builtins/arithmetic.py
+++ b/devito/builtins/arithmetic.py
@@ -4,7 +4,7 @@ import devito as dv
 from devito.builtins.utils import MPIReduction
 
 
-__all__ = ['norm', 'sumall', 'inner', 'mmin', 'mmax', 'count_nonzero']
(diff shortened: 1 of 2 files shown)


FabioLuporini (commented):
> Numpy's version is faster unless we change to collapse(1)

you may use the `par-collapse` opt-option also

But anyway, if you write it as a built-in, I don't see how you can glue it with a compiler pass, unless you copy-paste the equation definition?

georgebisbas:
> But anyway, if you write it as a built-in, I don't see how you can glue it with a compiler pass, unless you copy-paste the equation definition?

Yes, though I am still thinking about the best way for doing that.

georgebisbas:
> But anyway, if you write it as a built-in, I don't see how you can glue it with a compiler pass, unless you copy-paste the equation definition?

Yes, since it will be added as part of the pass, we will need to use the equations of it. However, are we opposed to having devito versions of this?

devito/builtins/arithmetic.py
```diff
@@ -161,3 +161,66 @@ def mmax(f):
...
+    ci = dv.ConditionalDimension(name='ci', parent=f.dimensions[-1],
+                                 condition=dv.Ne(f, 0))
+    g = dv.Function(name='g', shape=(1,), dimensions=(ci,), space_order=0, dtype=np.int32)
+
+    # Extra Scalar used to avoid reduction in gcc-5
+    i = dv.types.Scalar(name='i', dtype=np.int32)
+    eqi = dv.Eq(i, 0)
+
+    eq0 = dv.Inc(g[i], 1, implicit_dims=(f.dimensions + (ci,)))
+    op0 = dv.Operator([eqi, eq0], opt=('openmp', {'par-collapse-ncores': 30}))
```
mloubout, line 184: Is this really the most efficient way to compute this?

georgebisbas: wrt par-collpase?

mloubout: No I mean the equation  + conditional but I guess not really simpler way.

mloubout:
Where would this be used? And when you say faster does it count the code-generation overhead because it would re-generate the operator at each call so numpy is still probably more efficient.

georgebisbas:
> Where would this be used? And when you say faster does it count the code-generation overhead because it would re-generate the operator at each call so numpy is still probably more efficient.

Sorry to reply late Mathias. With regards to the use, maybe it is not needed anymore for the time-tiling work, since I can use this: https://github.com/devitocodes/devito/pull/1817/files#diff-2da23b1e569f0f2b184634790719bca628a15fdd92739ddf6e9f0452e5b9bc6eR614

wrt to efficiency I admit I did not check the operator generation.

The question is: Since we probably do not need this functionality, I am thinking of whether this can be an enhancement/tutorial of how to use conditional dimension? Thoughts @FabioLuporini ?

georgebisbas:
I agree.

PR #1817, opened 2021-12-31
compiler: source injection to nested field

georgebisbas:
Opening for discussion.
This helps to work around a bug when injecting to more complex fields such as f(t, u(x,y,z)).
Helps to implement Listing 3 of https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=9460483

### DIFF
--- a/devito/operations/interpolators.py
+++ b/devito/operations/interpolators.py
@@ -282,6 +282,14 @@ class LinearInterpolator(GenericInterpolator):
         """
         implicit_dims = as_tuple(implicit_dims) + self.sfunction.dimensions
 
+        def get_field_offset(field):
+            # In case of nested Functions, use them as a field
+            for ind in field.indices:
+                if ind.is_Function:
+                    return ind.origin
+
+            return field.origin
+
         def callback():
             # Derivatives must be evaluated before the introduction of indirect accesses
             try:
@@ -293,7 +301,8 @@ class LinearInterpolator(GenericInterpolator):
             variables = list(retrieve_function_carriers(_expr)) + [field]
 
             # Need to get origin of the field in case it is staggered
-            field_offset = field.origin
+            field_offset = get_field_offset(field)
+
             # List of indirection indices for all adjacent grid points
             idx_subs, temps = self._interpolation_indices(
                 variables, offset, field_offset=field_offset, implicit_dims=implicit_dims
--- a/devito/operations/interpolators.py
+++ b/devito/operations/interpolators.py
@@ -301,7 +301,11 @@ class LinearInterpolator(GenericInterpolator):
             variables = list(retrieve_function_carriers(_expr)) + [field]
 
             # Need to get origin of the field in case it is staggered
-            field_offset = get_field_offset(field)
+            field_offset = field.origin
+            try:
+                assert (len(field_offset) == len(self.grid.dimensions))
+            except:
+                field_offset = tuple([0] * len(self.grid.dimensions))
 
             # List of indirection indices for all adjacent grid points
             idx_subs, temps = self._interpolation_indices(
--- a/devito/operations/interpolators.py
+++ b/devito/operations/interpolators.py
@@ -282,14 +282,6 @@ class LinearInterpolator(GenericInterpolator):
         """
         implicit_dims = as_tuple(implicit_dims) + self.sfunction.dimensions
 
-        def get_field_offset(field):
-            # In case of nested Functions, use them as a field
-            for ind in field.indices:
-                if ind.is_Function:
-                    return ind.origin
-
-            return field.origin
-
         def callback():
             # Derivatives must be evaluated before the introduction of indirect accesses
             try:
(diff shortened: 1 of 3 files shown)


devito/operations/interpolators.py
```diff
@@ -268,6 +268,15 @@ def inject(self, field, expr, offset=0):
         offset : int, optional
             Additional offset from the boundary.
         """
+
+        def get_field_offset(field):
```
georgebisbas, line 272: Could be done recursively, but I highly ever doubt this will be used. Not to mention that even for more than one level of nesting depth the indirections get really complex for more than one or two dimensions.

devito/operations/interpolators.py
```diff
@@ -268,6 +268,15 @@ def inject(self, field, expr, offset=0):
...
+            # In case of nested Functions, use them as a field
+            for ind in field.indices:
+                if ind.is_Function:
```
FabioLuporini, line 275: this already fails if `u[..., 1 + f[...]]`

FabioLuporini (changes requested):

FabioLuporini:
I cannot access the paper. Can you paste/rewrite the snippet here?

devito/operations/interpolators.py
FabioLuporini: perhaps you need a SparseFunction subclass overriding `origin`? just spitballing ideas, I don't even remember what `origin` is supposed to return

(@mloubout )

tests/test_interpolation.py
```diff
@@ -590,3 +590,85 @@ def test_msf_interpolate():
     assert np.all(np.unique(nzt) == np.array([1, 2, 3, 4]))
     # 12 points x 4 timesteps
     assert nzt.size == 48
+
+
+@pytest.mark.parametrize('shape', [(50, 50, 50)])
+@pytest.mark.parametrize('so', (2, 4, 8))
+@pytest.mark.parametrize('tn', (40, 40, 200))
+def test_decompose_to_aligned(shape, so, tn):
```
FabioLuporini, line 598: is this really an MFE? It's huge

devito/operations/interpolators.py
mloubout: `origin` is used for staggered functions interp/inject since they can have a half grid shift. Origin is the shift

georgebisbas: The recursion should be fixed now. Tests added.

devito/operations/interpolators.py
georgebisbas: should be fixed now, tests added

georgebisbas:
> I cannot access the paper. Can you paste/rewrite the snippet here?

Helps to implement Listing 3  of https://arxiv.org/pdf/2010.10248.pdf

tests/test_interpolation.py
georgebisbas: It is not an MFE for nested source injection, but has just the necessary extra stuff to test nested source injection + listing 3 of the paper. So it is more than an MFE but still all of it seems useful to my eyes.

devito/operations/interpolators.py
FabioLuporini: I'm really not happy... what is this function supposed to do exactly? 

also, if you have two indirections this function prefers the first one... ?

```
a[b(), c()]
```

will return `b.origin` ?

as I said, maybe you should just subclass Function or SparseFunction to your needs e.g. overriding `def origin` s.t.` it returns what you expect it to return? 

this really looks like a hack for whatever you're attempting to achieve, but happy to be proven wrong

@mloubout thoughts?

mloubout: I honestly don't think this makes much sense to me. For a given (staggered) function it has an origin for each of its indices. But I don't see how/why anyone would index a staggered function with "staggered indices", i.e what this does is something like

$ f(x + h_x, g(y + h_y) + h_y)$ which doesn't quite make any sense to me. It should be either $ f(x + h_x, g(y + h_y))$  or $ f(x + h_x, g(y) + h_y)$ to make sense IMO.

So as @FabioLuporini said and actual example would be useful on where such an odd case would show up.

georgebisbas: The main problem derives from the field_offset for `save_src[time, s(x, y, z)]` is (0,0) instead of (0,0,0).
This, a bit later in the code leads to no-generation of `pz`: 
```
    def _coordinate_bases(self, field_offset):
        """Symbol for the base coordinates of the reference grid point."""
        return tuple([cast_mapper[self.dtype](c - o - idx * i.spacing)
                      for c, o, idx, i, of in zip(self._coordinate_symbols,
                                                  self.grid.origin_symbols,
                                                  self._coordinate_indices,
                                                  self.grid.dimensions[:self.grid.dim],
                                                  field_offset)])
```

In all existing cases in master the field has the same space dimensions as self.grid

gird is also used for:

```
    @cached_property
    def _coordinate_indices(self):
        """Symbol for each grid index according to the coordinates."""
        return tuple([INT(FLOOR((c - o) / i.spacing))
                      for c, o, i in zip(self._coordinate_symbols,
                                         self.grid.origin_symbols,
                                         self.grid.dimensions
```

What about in case of not equal len of offset and grid dimensions to resort in grid dimensions always?
Thoughts?

devito/operations/interpolators.py
```diff
@@ -280,6 +280,11 @@ def callback():
 
             # Need to get origin of the field in case it is staggered
             field_offset = field.origin
+            try:
```
FabioLuporini, line 283: Nope, neither the concept nor the implementation

devito/operations/interpolators.py
FabioLuporini: > The main problem derives from the field_offset for save_src[time, s(x, y, z)] is (0,0) instead of (0,0,0).

I get it, but as asked earlier, could you point us to the line of code that triggers the error without your edits? I guess it must be something along the lines of

```
sf.inject(...)
OR
sf.interpolate(...)
```

PR #1818, opened 2022-01-02
gpu: move blocking pass from custom to advanced mode

italoaug:


georgebisbas:
Hi, what is this PR offering compared to the previous implementation? Is loop blocking expected to deliver performance by default now?

devito/core/gpu.py
```diff
@@ -185,6 +188,10 @@ def _specialize_iet(cls, graph, **kwargs):
         # Distributed-memory parallelism
         mpiize(graph, sregistry=sregistry, options=options)
 
+        # Loop tiling
+        if options['blocklevels']:
```
FabioLuporini, line 192: does it fail without this?

FabioLuporini (commented):
Minor question about the need for `if options[blocklevels] ...` but otherwise looking good :ok_hand: 

@georgebisbas : see latest discussion in `gpu-dev`

devito/core/gpu.py
georgebisbas: Should be fine to remove yes? Unless we miss sth.

tests/conftest.py
```diff
@@ -311,14 +311,8 @@ def assert_blocking(operator, exp_nests):
...
-                       ('openacc', 'blocking', {'skewing': True}),
-                       ('openacc', 'blocking', {'skewing': True, 'blockinner': True})]
-
-
-# A list of optimization options/pipelines to be used in testing
-# regarding GPU openmp spatial and/or temporal blocking.
-opts_openmp_tiling = [('openmp', 'blocking'),
-                      ('openmp', 'blocking', {'skewing': True}),
-                      ('openmp', 'blocking', {'skewing': True, 'blockinner': True})]
+# regarding GPU spatial and/or temporal blocking.
```
georgebisbas, line 314: If we ask for `blocking` only, do we get blocklevels=1 now?

devito/core/gpu.py
italoaug: You're right. It works fine without the conditional. So I'm removing it for the next commit.

FabioLuporini (approved):
LGTM, merging

PR #1819, opened 2022-01-14
CUDA/HIP/SYCL preliminaries + misc improvements

FabioLuporini:
I want to use the JIT backdoor to play with CUDA/HIP/SYCL.

This PR goes in this direction adding misc features concerning both execution and code generation

### DIFF
--- a/devito/arch/compiler.py
+++ b/devito/arch/compiler.py
@@ -482,6 +482,25 @@ class NvidiaCompiler(PGICompiler):
         self.MPICXX = 'mpicxx'
 
 
+class CudaCompiler(Compiler):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, cpp=True, **kwargs)
+
+        self.cflags.remove('-std=c99')
+        self.cflags.remove('-Wall')
+        self.cflags.remove('-fPIC')
+        self.cflags += ['-std=c++11', '-Xcompiler', '-fPIC']
+
+        self.src_ext = 'cu'
+
+    def __lookup_cmds__(self):
+        self.CC = 'nvcc'
+        self.CXX = 'nvcc'
+        self.MPICC = 'mpic++'
+        self.MPICXX = 'mpicxx'
+
+
 class IntelCompiler(Compiler):
 
     def __init__(self, *args, **kwargs):
@@ -597,8 +616,9 @@ compiler_registry = {
     'pgcc': PGICompiler,
     'pgi': PGICompiler,
     'nvc': NvidiaCompiler,
-    'nvcc': NvidiaCompiler,
+    'nvc++': NvidiaCompiler,
     'nvidia': NvidiaCompiler,
+    'cuda': CudaCompiler,
     'osx': ClangCompiler,
     'intel': IntelCompiler,
     'icpc': IntelCompiler,
--- a/devito/ir/clusters/analysis.py
+++ b/devito/ir/clusters/analysis.py
@@ -135,7 +135,7 @@ class Rounding(Detector):
         # so that `i`'s trip count can safely be rounded up
         # Note: autopadding guarantees that the padding size along the
         # Fastest Varying Dimension is a multiple of the SIMD vector length
-        functions = [f for f in scope.functions if f.is_Tensor]
+        functions = [f for f in scope.functions if f.is_AbstractFunction]
         if any(not f._honors_autopadding for f in functions):
             return
 
--- a/devito/passes/iet/langbase.py
+++ b/devito/passes/iet/langbase.py
@@ -3,11 +3,13 @@ from abc import ABC
 
 import cgen as c
 
-from devito.ir import (BlankLine, DummyExpr, Call, Conditional, List, Prodder,
-                       ParallelIteration, ParallelBlock, PointerCast, EntryFunction,
-                       ThreadFunction)
+from devito.ir import (BlankLine, DummyExpr, Call, Conditional, Expression,
+                       List, Prodder, ParallelIteration, ParallelBlock,
+                       PointerCast, EntryFunction, ThreadFunction, FindNodes,
+                       FindSymbols)
 from devito.mpi.distributed import MPICommObject
 from devito.passes.iet.engine import iet_pass
+from devito.passes.iet.misc import is_on_device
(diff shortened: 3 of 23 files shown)


devito/passes/iet/langbase.py
```diff
@@ -43,6 +46,48 @@ class LangBB(object, metaclass=LangMeta):
...
 
+    @classmethod
+    def _map_data(cls, f):
+        if f.is_Array:
+            return f.symbolic_shape
+        else:
+            return tuple(f._C_get_field(FULL, d).size for d in f.dimensions)
+
+    @classmethod
+    def _make_symbolic_sections_from_imask(cls, f, imask):
```
FabioLuporini, line 57: note: this is NOT new -- just lifted from `parpragma.py`

devito/passes/iet/langbase.py
```diff
@@ -43,6 +46,48 @@ class LangBB(object, metaclass=LangMeta):
     Prodder = Prodder
     PointerCast = PointerCast
 
+    @classmethod
+    def _map_data(cls, f):
```
FabioLuporini, line 50: note: this is NOT new -- just lifted from` parpragma.py`

devito/passes/iet/langbase.py
```diff
@@ -298,3 +343,16 @@ def _(iet):
             return iet, {}
 
         return _initialize(iet)
+
+    def _is_offloadable(self, iet):
```
FabioLuporini, line 347: note: this is NOT new -- just lifted from `parpragma.py`

devito/passes/iet/parpragma.py
```diff
@@ -485,41 +474,6 @@ def _make_nested_partree(self, partree):
 
 class PragmaLangBB(LangBB):
 
-    @classmethod
-    def _make_symbolic_sections_from_imask(cls, f, imask):
```
FabioLuporini, line 489: see comments above... lifted to langbase.py

devito/types/basic.py
```diff
@@ -1315,7 +1321,7 @@ class LocalObject(AbstractObject):
 # - To override SymPy caching behaviour
 
 
-class IndexedData(sympy.IndexedBase, Pickable):
+class IndexedData(sympy.IndexedBase, Pickable, CodeSymbol):
```
FabioLuporini, line 1324: the changes here are to pass a Function by "data pointer" rather by "struct dataobj *", ie it allows the generation of

```
void foo(float * f)
```

besides

```
void foo(struct dataobj * f_vec)
```

devito/arch/compiler.py
```diff
@@ -597,8 +616,9 @@ def __lookup_cmds__(self):
     'pgcc': PGICompiler,
     'pgi': PGICompiler,
     'nvc': NvidiaCompiler,
-    'nvcc': NvidiaCompiler,
+    'nvc++': NvidiaCompiler,
     'nvidia': NvidiaCompiler,
+    'nvcc': CudaCompiler,
```
mloubout, line 621: would probably calll it `cuda` , nvc/nvc++ bit too close.

devito/ir/clusters/analysis.py
```diff
@@ -135,7 +135,7 @@ def _callback(self, clusters, d, prefix):
         # so that `i`'s trip count can safely be rounded up
         # Note: autopadding guarantees that the padding size along the
         # Fastest Varying Dimension is a multiple of the SIMD vector length
-        functions = [f for f in scope.functions if f.is_Tensor]
+        functions = [f for f in scope.functions if f.is_AbstractFunction]
```
mloubout, line 138: So why this relaxation?

mloubout (commented):
Looks good, minor comments. Is that callable now "nestable", i.e could you in theory add call to external functions?

georgebisbas (commented):
Had a very quick look, nothing really caught my eye.  Question: Is the CUDA code gonna be linearized? Maybe better:
```
__global__
void saxpy(int n, float a, float *x, float *y)
{
  int i = blockIdx.x*blockDim.x + threadIdx.x;
  if (i < n) y[i] = a*x[i] + y[i];
}

```

Is thelinearization gonna help produce:
`int i = blockIdx.x*blockDim.x + threadIdx.x;` somehow?

devito/ir/clusters/analysis.py
FabioLuporini: oh no, actually just `is_Tensor` being completely redundant, ie hitting same cases as `is_AbstractFunction` (which was more widespread and "sensible" as name)

FabioLuporini:
@georgebisbas ultimately yes, but we're a long way from that :)

FabioLuporini:
@mloubout 

> Is that callable now "nestable", i.e could you in theory add call to external functions?

what sort of external functions?

I guess not MPI routines for example :) they are external too... you mean your own C functions? if so:

* no, as in, Callables aren't really neastable; Calls are though (but that's not your point right?). 
* However, if that's what you up to, I got a prototypal branch in devitopro that adds an **API** to inject whatever custom call you like

devito/arch/compiler.py
FabioLuporini: that would overload `DEVITO_LANGUAGE=cuda`

I agree the names are close. But these are the NVidia names, really.

* nvcc is the NVidia CUDA Compiler
* nvc is the NVidia C compiler for C/C++/OpenACC/OpenMP/etc (but not CUDA)

mloubout:
@FabioLuporini 
>what sort of external functions?

Was thinking stuff like blas/mkl/..... So yeah related to second point. I don't have a case for now just wondering.

devito/arch/compiler.py
mloubout: I get you but I still find the name confusing, in particular with the redundant ones (`nvidia`, `nvc`) that make it even more confusing. 
Why is it an issue to have same name for language and arch?

georgebisbas (approved):

devito/arch/compiler.py
FabioLuporini: Not an issue. And I don't have strong objections. So changed it to `cuda` as per your suggestion

mloubout (approved):

PR #1822, opened 2022-01-23
arch: Support OSX+M1

FabioLuporini:


FabioLuporini:
for the record: `-march=native` doesn't work on Apple M1 https://stackoverflow.com/questions/65966969/why-does-march-native-not-work-on-apple-m1

devito/arch/compiler.py
```diff
@@ -581,10 +586,14 @@ class CustomCompiler(Compiler):
     """
 
     def __new__(cls, *args, **kwargs):
+        platform = kwargs.pop('platform', configuration['platform'])
+
         if any(i in environ for i in ['CC', 'CXX', 'CFLAGS', 'LDFLAGS']):
             obj = super().__new__(cls, *args, **kwargs)
             obj.__init__(*args, **kwargs)
             return obj
+        elif platform is M1:
```
mloubout, line 595: I'd probably put it as first `if` in the unlikely case it's an M! but someone declared CC

devito/arch/archinfo.py
```diff
@@ -482,7 +488,7 @@ def _mro(cls):
 
     def _detect_isa(self):
         for i in reversed(self.known_isas):
-            if any(j.startswith(i) for j in get_cpu_info()['flags']):
+            if any(j.startswith(i) for j in as_tuple(get_cpu_info()['flags'])):
```
mloubout, line 491: `get_cpu_info().get('flag', None)`

FabioLuporini: that is our `get_cpu_info`, which is guaranteed by construction to have a 'flags' entry

PR #1826, opened 2022-01-26
Corrected 2D kernel when sintheta==0

dimitri-voytan:
When sin(theta) is 0, the Zero object has no attribute `dy`


dimitri-voytan:
I noticed that this version is inconsistent with the style of the 3D implementation , so going to close and resubmit

PR #1827, opened 2022-01-26
Fix 2D Gzz when sintheta==0

dimitri-voytan:
When sin(theta) == 0, Gz*sintheta has no attribute .dx. This correction is now consistent with the 3D Gzz centered kernel


examples/seismic/tti/operators.py
```diff
@@ -110,9 +110,13 @@ def Gzz_centered_2d(model, field, costheta, sintheta, space_order):
...
     order1 = space_order // 2
     Gz = -(sintheta * field.dx(fd_order=order1) +
-           costheta * field.dy(fd_order=order1))
-    Gzz = ((Gz * sintheta).dx(fd_order=order1).T +
-           (Gz * costheta).dy(fd_order=order1).T)
+              costheta * field.dy(fd_order=order1))
+    Gzz = (Gz * costheta).dy(fd_order=order1).T
+
+    # Add rotated derivative if angles are not zero. If angles are
+    # zeros then `0*Gz = 0` and doesn't have any `.dy` ....
```
FabioLuporini, line 117: @mloubout actually, on second thought, the neat(er) way to go here would be subclassing SymPy's Zero and One and make them differentiable objects

mloubout: That would only work in some very special cases where we set the zero ourselves, in general, it's just gonna be random zero popping up from an expression we can't control so it'll still be a standard Sympy Zero if not just the zero number (or any number).

mloubout: @FabioLuporini thoughts?

georgebisbas:
We still need renaming of commits here as shown in: https://github.com/devitocodes/devito/wiki/Tags-for-commit-messages-and-PR-titles

examples/seismic/tti/operators.py
FabioLuporini: I'm not entirely sure there is no way of returning our own zero, but I agree this is work we don't have the time to undertake now, so for me it's a go provided that, as @georgebisbas pointed out, commit messages are rewritten to follow the naming conventions

mloubout:
@dimitri-voytan could you rebase that branch. I used the github browser to update the branch which removed your commit ownership so for the sake of recognizing this is your PR, it would be good to have you rebase it to put back your name on it.

dimitri-voytan:
Rebased to main

mloubout:
Thanks. Looks like something went wrong though. Sorry for all the steps

mloubout:
Took care of it in #1855, lemme know if you ok with it, it should still show you are the author of the commit

dimitri-voytan:
Looks good to me!

mloubout:
Fixed in #1855

PR #1836, opened 2022-02-07
gpu: Adding DPCPP compiler support

reguly:
This adds Intel GPU support through the clang-based dpcpp/icx compiler.
`DEVITO_ARCH=dpcpp`
However, without setting `DEVITO_PLATFORM=nvidiaX`, CPU-side OpenMP is generated. Please advise on how to extend the list of platforms so that GPU-side code is generated.

Furthermore, I am getting an issue out of codepy:
in jit.py's `get_dep_structure()` function, the following command `toolchain.get_dependencies(source_paths)` includes  a bunch of header files, as well as `host-x86_64-unknown-linux-gnu` and `openmp-spir64` which are not files that the system can find, so it crashes. My hack included removing these.

Tested on a Gen9 GPU in Intel DevCloud. The Iris XE GPUs lack double precision support and some automatic type conversions in the generated code currently results in double precision ops.


mloubout:
- For the platform a new `IntelGPU` or some good name for it should be added to the known arches
https://github.com/devitocodes/devito/blob/master/devito/arch/archinfo.py

- We will check the trigo functions type and such, it should indeed produce the correct ones for single precision
- Concerning `get_dep_structure` there may be details with intel GPU popping up, to be investigated.

reguly:
Thanks - created `intelgpuX` platform, added it to the appropriate places.
Filed pull request for codepy: https://github.com/inducer/codepy/pull/26

reguly:
This should work on any integrated GPU, and only requires the Base Toolkit:
https://www.intel.com/content/www/us/en/developer/tools/oneapi/base-toolkit-download.html

You can test on Intel Devcloud (free to register):
https://www.intel.com/content/www/us/en/developer/tools/devcloud/overview.html

which has Gen9 GPUs, and Iris XE GPUs (which still have the issue with single/double precision math functions).

Though if you only want to test functionality, it will work on Intel CPUs too.

georgebisbas:
Maybe for the testing we can speedup https://github.com/devitocodes/devito/pull/1806
and combine both to an Intel pipeline?
However, the hardware availability (Intel GPU) is the primary question.

devito/arch/archinfo.py
```diff
@@ -571,6 +571,11 @@ def memavail(self, deviceid=0):
         except (AttributeError, KeyError):
             return None
 
+class IntelDevice(Device):
```
FabioLuporini, line 574: Before I trigger CI -- classes require two blank lines before and after their definition, so this needs two more blank lines, one here and one at the bottom, or it will fail the flake8 build

devito/arch/compiler.py
```diff
@@ -455,6 +455,18 @@ def __lookup_cmds__(self):
         self.MPICC = 'mpicc'
         self.MPICXX = 'mpicxx'
 
+class DPCPPCompiler(Compiler):
```
FabioLuporini, line 458: same here

FabioLuporini (commented):
aside from the trivial pep8 issues, this is good to go for me. Many thanks!

georgebisbas:
I think we also need renaming of commit messages according to this: https://github.com/devitocodes/devito/wiki/Tags-for-commit-messages-and-PR-titles
?

devito/arch/archinfo.py
```diff
@@ -657,7 +665,8 @@ def march(cls):
     'power8': POWER8,
     'power9': POWER9,
     'nvidiaX': NVIDIAX,  # Generic NVidia GPU
-    'amdgpuX': AMDGPUX   # Generic AMD GPU
+    'amdgpuX': AMDGPUX,   # Generic AMD GPU
+    'intelgpuX': INTELGPUX   #Generic Intel GPU
```
georgebisbas, line 669: yes, needs space after hash

georgebisbas (changes requested):

georgebisbas:
> sorry, newbie issues :)

heh, no worries at all Istvan! Thank you very much for your contribution! :-)

georgebisbas (approved):

FabioLuporini:
@reguly Thanks Istvan. I've now enabled CI.

devito/arch/archinfo.py
```diff
@@ -14,11 +14,11 @@
 
 __all__ = ['platform_registry', 'get_cpu_info', 'get_gpu_info', 'get_nvidia_cc',
            'Platform', 'Cpu64', 'Intel64', 'Amd', 'Arm', 'Power', 'Device',
-           'NvidiaDevice', 'AmdDevice',
+           'NvidiaDevice', 'AmdDevice','IntelDevice',
```
georgebisbas, line 17: `, 'IntelDevice',`
This whitespace  after comma should fix everything in flake8 
https://github.com/devitocodes/devito/runs/5155237472?check_suite_focus=true

georgebisbas (approved):

devito/arch/compiler.py
```diff
@@ -456,6 +456,20 @@ def __lookup_cmds__(self):
         self.MPICXX = 'mpicxx'
 
 
+class DPCPPCompiler(Compiler):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, cpp=False, **kwargs)
```
FabioLuporini, line 462: @reguly shouldn't this be `cpp=True` ?

reguly: No, I found the icx compiler packaged with dpcpp which is willing to compile C code. Though I guess cpp=False is the default, so this is unnecessary?

FabioLuporini: yep probably unnecessary, but would also be useful if you could add a comment about this:

> I found the icx compiler packaged with dpcpp which is willing to compile C code.

This is news to me -- I thought it expected cpp (hence the compiler name...)

reguly: Added some comments, hope that clarifies this. Intel now has too many compilers! :)

FabioLuporini (approved):

PR #1840, opened 2022-02-17
compiler: Add SEPARABLE property

FabioLuporini:


devito/ir/support/properties.py
```diff
@@ -65,6 +73,7 @@ def normalize_properties(*args):
     elif len(args) == 1:
         return args[0]
 
+    # SEQUENTIAL > PARALLEL_w_condition > PARALLEL
```
georgebisbas, line 76: This comment can be improved

devito/ir/support/properties.py
```diff
@@ -51,6 +51,14 @@ def __init__(self, name, val=None):
 vector length thanks to the presence of enough padding.
 """
 
+ZEROABLE = Property('zeroable')
+"""
+A Dimension whose upper and lower limits could be safely set to zero, that is
+without altering the correctness of the computation. They reason they may
```
georgebisbas, line 57: The reason

tests/test_dse.py
```diff
@@ -2493,6 +2493,24 @@ def test_hoisting_pow_one(self):
...
+        op = Operator(Eq(u.forward, u.dx.dx))
+        iterations = FindNodes(Iteration).visit(op)
+        v = iterations.pop(2)
+        assert ZEROABLE in v.properties
+        assert not any(ZEROABLE in i.properties for i in iterations)
+
+        # Fusion makes ZEROABLE disappear
+        for eq in [Eq(u.forward, u.dx.dx + u.dy.dy),
+                   Eq(u.forward, u.dx.dx + u.dx.dy)]:
+            op = Operator(eq)
```
georgebisbas, line 2510: I would recommend to op.apply() as well.

georgebisbas (changes requested):

tests/test_dse.py
FabioLuporini: testing for what?

georgebisbas: It reminded me of an issue I had with fcc compiler. In case the loop was going from 0 to 0 openmp was segfaulting.
We should be safe for what we have now, but it would not harm I guess. In any case it is nitpicking

georgebisbas: I totally get it. I am just wondering whether there is a chance that the loop goes from 0 to 0.

devito/ir/support/properties.py
FabioLuporini: fixed

devito/ir/support/properties.py
FabioLuporini: tweaked

tests/test_dse.py
```diff
@@ -2493,6 +2493,24 @@ def test_hoisting_pow_one(self):
...
 
+    def test_zeroable_property(self):
+        grid = Grid(shape=(10, 10))
+
+        u = TimeFunction(name='u', grid=grid, space_order=4)
+
+        op = Operator(Eq(u.forward, u.dx.dx))
+        iterations = FindNodes(Iteration).visit(op)
+        v = iterations.pop(2)
+        assert ZEROABLE in v.properties
```
mloubout, line 2504: Now I'm confused. Shouldn't be `y` the only zeroable here?

mloubout (approved):

tests/test_dse.py
FabioLuporini: yes, that's what the assert checks! `v` is `y`

mloubout: I should not do that late. Of course it's y that's what pop does

mloubout (approved):

FabioLuporini:
I want to change a few things, hence back to WIP

devito/symbolics/printer.py
```diff
@@ -151,6 +152,19 @@ def _print_TrigonometricFunction(self, expr):
...
 
+    def _print_Or(self, expr):
+        # Enforce parentheses if Or(And() ...)
+        # This is only to avoid compiler warnings, because technically
+        # && has precedence over || in C (see for example
+        # https://en.cppreference.com/w/c/language/operator_precedence)
+        args = []
+        for i in expr.args:
+            v = self._print(i)
+            if isinstance(i, And):
```
mloubout, line 163: Would probably generic check if Relational

devito/passes/clusters/aliases.py
```diff
@@ -1191,6 +1196,16 @@ def free_symbols(self):
...
 
+    @property
+    def is_frame(self):
+        """
+        An Alias is said to be a "frame" if the `distances` of the `aliaseds`
+        expressions from the `pivot` form a frame in the linear algebra sense.
+        In essence, any two distances are either orthogonal or one is a multiple
+        of the other by a scalar constant (i.e., a "redundant" vector w.r.t. a basis).
+        """
+        return all(len([e for e in i if e != 0]) <= 1 for i in self.distances)
```
mloubout, line 1207: Not sure how this match the def

tests/test_dse.py
```diff
@@ -2493,6 +2493,24 @@ def test_hoisting_pow_one(self):
 
         assert len([i for i in FindSymbols().visit(op) if i.is_Array]) == 0
 
+    @pytest.mark.parametrize('expr,expected', [
+        ('u.dy.dy', False),  # No STAR as hyperplane degenerates to 1D space
+        ('u.dx.dx + u.dy.dy', True),
+        ('u.dx.dx + u.dx.dy', True)
```
mloubout, line 2499: Not sure Star is a very representative name, these are square stencil, `u.dx + u.dy` is what's called a star stencil

mloubout (commented):
Some comments but overall looks good

devito/passes/clusters/aliases.py
FabioLuporini: a distance is a vector in Rn

tests/test_dse.py
FabioLuporini: nope, after CIRE they become stars -- that's the whole point of CIRE

devito/symbolics/printer.py
FabioLuporini: uhm, I'll think about this one

tests/test_dse.py
mloubout: That's not possible. It either gets separated and becomes 1D lines or it's a square, there no possible way to compute `dx.dy` with a star stencil

devito/passes/clusters/aliases.py
mloubout: I know that, but this expression still doesn't really make sense to me w.r.t being  a frame (maybe not the best time to think about it)

tests/test_dse.py
FabioLuporini: yes it is possible possible the u.dx is assigned to a temporary
so conceptually it's a square, but the implementation makes it a star

for me a square means that in a loop nest you have accesses sucha s `u[x+3][y+2]`. Not the case here

mloubout: Assigning it to a temporary means you are separating the dimensions. In which case you don't have a start you have a sequence of 1D line stencils

devito/ir/support/properties.py
```diff
@@ -58,13 +58,22 @@ def __init__(self, name, val=None):
 and "regular".
 """
 
+STAR = Property('star')
+"""
+Two or more Dimensions defining a star-shaped stencil.
```
georgebisbas, line 63: Can add another sentence for what is a star-shaped stencil.

georgebisbas (approved):

tests/test_dse.py
FabioLuporini: yes, but if you take the two loop nests together, it's a star

devito/passes/clusters/aliases.py
FabioLuporini: no?

the inner 

```
len([e for e in i if e != 0])
```

tells you how many non-zero entries in an Rn vector. And the inner condition is s.t. it gives True iff at most one element != 0. The outer condition imposes such condition to all Rn vectors in `self.distances`. Hence it is True if for example `(0, 0, 0, 0)`, `(0, 1, 0, 0)`, `(0, 2, 0, 0)`, `(0, 0, 1, 0)`. And here the Rn vectors are either pairwise orthogonal or scaled. No?

devito/passes/clusters/aliases.py
```diff
@@ -1229,18 +1258,34 @@ def filter(self, key):
     def aliaseds(self):
         return flatten(i.aliaseds for i in self._list)
 
+    @property
+    def is_frame(self):
+        """
+        An AliasList is said to be a "frame" if all of its Aliases are frames.
```
georgebisbas, line 1264: IF -> IFF?

devito/passes/clusters/misc.py
```diff
@@ -432,3 +432,17 @@ def optimize_msds(clusters):
     Optimize clusters defined over MultiSubDomains.
     """
     return MSDOptimizer().process(clusters)
+
+
+@timed_pass()
+def optimize_hyperplanes(clusters):
+    """
+    At the moment this is just a dummy no-op pass that we only use
```
georgebisbas, line 440: At the moment, ....

devito/types/misc.py
```diff
@@ -113,6 +113,17 @@ class Global(Symbol):
     pass
 
 
+class Hyperplane(tuple):
+
+    """
+    A collection of Dimensions defining an hyperplane.
```
georgebisbas, line 119: minor: I think 'a hyperplane'

devito/passes/clusters/aliases.py
mloubout: This only checks sufficient (see slack discussion) so would not put IFF

mloubout (approved):

PR #1843, opened 2022-02-21
api: Add op.cinterface for C-level interoperability

FabioLuporini:


devito/operator/operator.py
```diff
@@ -603,6 +603,39 @@ def cfunction(self):
 
         return self._cfunction
 
+    def cinterface(self, force=False):
```
ggorman, line 606: How about adding name=None to the API (and a `if not name` before line 621) so that the caller can change the default?

FabioLuporini: the user can control that already via `Operator(..., name=X)` -- so here (unlike "classic" devito use model) we don't use the `.soname` but rather the actual `Operator` name.

devito/ir/iet/visitors.py
```diff
@@ -487,25 +513,50 @@ def visit_Operator(self, o):
...
+
+        return c.Module(headers + includes + typedecls +
                         esigns + [blankline, kernel] + efuncs)
 
 
+class CInterface(CGen):
+
+    def _operator_includes(self, o):
+        includes = super()._operator_includes(o)
+        includes.append(c.Include("%s.h" % o.name, system=False))
```
mloubout, line 536: why this not in `_operator_includes` ?

mloubout (approved):

devito/ir/iet/visitors.py
FabioLuporini: well when we normally execute devito and run `op.apply` there's no Operator-specific header file to include --it's all in the `.c`, no? :)

devito/ir/iet/visitors.py
```diff
@@ -467,9 +468,34 @@ def visit_HaloSpot(self, o):
         body = flatten(self._visit(i) for i in o.children)
         return c.Collection(body)
 
-    def visit_Operator(self, o):
-        blankline = c.Line("")
+    # Operator-handle machinery
```
georgebisbas, line 471: handling?

devito/ir/iet/visitors.py
```diff
@@ -467,9 +468,34 @@ def visit_HaloSpot(self, o):
...
+            xfilter = lambda i: True
+        else:
+            public_types = (AbstractFunction, CompositeObject)
+            if mode == 'public':
+                xfilter = lambda i: isinstance(i, public_types)
+            else:
+                xfilter = lambda i: not isinstance(i, public_types)
+
+        typedecls = [i._C_typedecl for i in o.parameters
+                     if xfilter(i) and i._C_typedecl is not None]
```
georgebisbas, line 488: `is not None`  can be safely dropped I think?

georgebisbas, line 488: `is not None`  can be safely dropped I think?

devito/ir/iet/visitors.py
```diff
@@ -467,9 +468,34 @@ def visit_HaloSpot(self, o):
...
+        for i in o._func_table.values():
+            if not i.local:
+                continue
+            typedecls.extend([j._C_typedecl for j in i.root.parameters
+                              if xfilter(j) and j._C_typedecl is not None])
```
georgebisbas, line 493: `is not None`  can be safely dropped I think?

georgebisbas, line 493: `is not None`  can be safely dropped I think?

mloubout:
CI error is a stupid codecov failing to upload this is GTG

PR #1844, opened 2022-02-21
CI: prevent  conflict with pytest

mloubout:
Supersedes #1833 

`pytest` now has a `setup` function that was badly conflicting with an import, this should fix compat.


georgebisbas (approved):

FabioLuporini (commented):
LGTM, thx

PR #1847, opened 2022-02-23
Installation: update setup.py and cleanup docker files

mloubout:
Add mpi and nvidia options to setup.py

Small cleanup of docker file to actually install devito rather than setting python path for every command

### DIFF
--- a/docker/Dockerfile.nvidia
+++ b/docker/Dockerfile.nvidia
@@ -79,18 +79,10 @@ ENV CUDA_ROOT $HPCSDK_HOME/cuda/bin
 ENV PATH $HPCSDK_HOME/compilers/bin:$HPCSDK_HOME/cuda/bin:$HPCSDK_HOME/comm_libs/mpi/bin:${PATH}
 ENV LD_LIBRARY_PATH $HPCSDK_HOME/cuda/lib:$HPCSDK_HOME/cuda/lib64:$HPCSDK_HOME/compilers/lib:$HPCSDK_HOME/math_libs/lib64:$HPCSDK_HOME/comm_libs/mpi/lib:$HPCSDK_CUPTI/lib64:bitcomp_DIR:${LD_LIBRARY_PATH}
 
-ADD ./requirements.txt /app/requirements.txt
-ADD ./requirements-optional.txt /app/requirements-optional.txt
-ADD ./requirements-nvidia.txt /app/requirements-nvidia.txt
-ADD ./requirements-mpi.txt /app/requirements-mpi.txt
-
-RUN python3 -m venv /venv && \
-    /venv/bin/pip install --no-cache-dir --upgrade pip && \
-    /venv/bin/pip install --no-cache-dir wheel && \
-    /venv/bin/pip install --no-cache-dir -r /app/requirements.txt && \
-    /venv/bin/pip install --no-cache-dir -r /app/requirements-optional.txt && \
-    /venv/bin/pip install --no-cache-dir -r /app/requirements-nvidia.txt && \
-    rm -rf ~/.cache/pip
+# Copy Devito
+ADD . /app/devito
+# Remove git files
+RUN rm -rf /app/devito/.git
 
 # MPI ROOT USER DEFAULTS
 ENV OMPI_ALLOW_RUN_AS_ROOT=1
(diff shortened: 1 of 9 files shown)


docker/Dockerfile.nvidia
```diff
@@ -125,16 +116,14 @@ RUN if [ "x$MPI_VER" = "x4" ]; then \
...
-ADD ./examples /app/examples
-ADD ./benchmarks /app/benchmarks
-COPY setup.cfg /app/
+# install
+RUN python3 -m venv /venv && \
+    /venv/bin/pip install --no-cache-dir --upgrade pip && \
+    /venv/bin/pip install --no-cache-dir wheel && \
+    /venv/bin/pip install --no-cache-dir -e /app/devito[extras,mpi,nvidia]
+
+ENV CFLAGS=
```
georgebisbas, line 127: ?

docker/Dockerfile.nvidia
```diff
@@ -125,16 +116,14 @@ RUN if [ "x$MPI_VER" = "x4" ]; then \
...
-    rm -rf ~/.cache/pip
-ENV CFLAGS=
 
-ADD ./devito /app/devito
-ADD ./tests /app/tests
-ADD ./scripts /app/scripts
-ADD ./examples /app/examples
-ADD ./benchmarks /app/benchmarks
-COPY setup.cfg /app/
+# install
```
georgebisbas, line 121: Maybe enrich...
# Upgrade pip, install wheel, devito and rest....

docker/Dockerfile.nvidia
```diff
@@ -178,7 +167,7 @@ RUN mkdir -p /app && \
     
 # Change to the app user.
 USER app
-WORKDIR /app
+WORKDIR /app/devito
```
georgebisbas, line 171: I guess we are not anymore backwards compatible with any people that will be developing or running scripts upon docker installation?

georgebisbas (commented):
Nice cleanup

docker/Dockerfile.nvidia
mloubout: We are the root dir is different but the structure is the same

docker/Dockerfile.nvidia
mloubout: Was already there

docker/Dockerfile.nvidia
```diff
@@ -125,16 +117,14 @@ RUN if [ "x$MPI_VER" = "x4" ]; then \
...
                $HPCSDK_HOME/comm_libs/mpi ; \
     fi;  \
-    /venv/bin/pip install --no-cache-dir -r /app/requirements-mpi.txt && \
-    rm -rf ~/.cache/pip
-ENV CFLAGS=
 
-ADD ./devito /app/devito
-ADD ./tests /app/tests
-ADD ./scripts /app/scripts
-ADD ./examples /app/examples
```
FabioLuporini, line 135: How are the examples/benchmarks now brought in?

FabioLuporini (commented):
nice cleanup! I just have a minor question about the examples and benchmarks folders that my sleepy brain is still failing to catch

docker/Dockerfile.nvidia
```diff
@@ -79,18 +79,10 @@ ENV CUDA_ROOT $HPCSDK_HOME/cuda/bin
...
-    /venv/bin/pip install --no-cache-dir --upgrade pip && \
-    /venv/bin/pip install --no-cache-dir wheel && \
-    /venv/bin/pip install --no-cache-dir -r /app/requirements.txt && \
-    /venv/bin/pip install --no-cache-dir -r /app/requirements-optional.txt && \
-    /venv/bin/pip install --no-cache-dir -r /app/requirements-nvidia.txt && \
-    rm -rf ~/.cache/pip
+# Devito
+ADD . /app/devito
+# Remove git files
+RUN rm -rf /app/devito/.git
```
FabioLuporini, line 85: why?

FabioLuporini:
PR name also needs a suitable tag

docker/Dockerfile.nvidia
mloubout: Same as it was and as it is in devito, this was just copying each subdir one by one

docker/Dockerfile.nvidia
mloubout: Git history takes space and the docker user doesn't have a git account anyway

docker/Dockerfile.nvidia
mloubout: Fixed

Issue #1739, opened 2021-08-18
test_hoisting_scalar_divs fails if any of CC/CFLAGS/LDFLAGS/CXX are set?


tjb900:
So I'm trying to build a local environment where the tests pass and noticed this test seems to have a problem. This is with current `master` == `02bbefb7e380d299a2508fef2923c1a4fbd5c59d`. CI doesn't seem to hit this. I suspect that none of those environment variables (CC, CFLAGS, LDFLAGS, CXX) are set in the environment, so this code path is not exercised?

I'm not sure that overriding `__new__` in this way is the neatest way to go about this?

```
__________________________________________________________________________________________________ TestAliases.test_hoisting_scalar_divs ___________________________________________________________________________________________________

self = <test_dse.TestAliases object at 0x7f66c7a63d30>

    def test_hoisting_scalar_divs(self):
        """
        Test that scalar divisions are hoisted out of the inner loops.
        """
        grid = Grid(shape=(3, 3))
    
        u = TimeFunction(name="u", grid=grid, time_order=2, space_order=4)
        m = Function(name='m', grid=grid, space_order=4)
    
        pde = m * u.dt2 - u.laplace
        eq = Eq(u.forward, solve(pde, u.forward))
    
        # Check that different backends behave the same
        op0 = Operator(eq, opt=('advanced', {'openmp': False}))
>       op1 = Operator(eq, platform='nvidiaX', language='openacc')

tests/test_dse.py:1553:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
devito/operator/operator.py:144: in __new__
    kwargs = parse_kwargs(**kwargs)
devito/operator/operator.py:972: in parse_kwargs
    configuration['compiler'].__new_from__(platform=kwargs['platform'],
devito/arch/compiler.py:192: in __new_from__
    return self.__class__(suffix=kwargs.pop('suffix', self.suffix),
[...]

PR #1850, opened 2022-02-27
arch: Fix CustomCompiler instantiation

AtilaSaraiva:
Hello guys,
I was here packaging devito for my distribution and got into a bug saying
```
TypeError: object.__new__() takes exactly one argument (the type to instantiate)
```
for the line I changed with this simple commit. As a consequence I simply fixed the issue. I've made some tests and this shouldnt and mustn't break anything on anyone's computer.

This may sound simple and non important, but Devito is written in Python, which is a distro agnostic language. It is expected that the code will run on any linux or mac, and if it doesn't I think it should be treated as a bug (at least in my humble opinion).

I have every intention to make Devito work on [NixOS package manager (Nix)](https://nixos.org/) since it is completely reproducible and distro agnostic, which will relieve me from the necessity of using a Docker container for developement.

Thx for your time.

### DIFF
--- a/devito/arch/compiler.py
+++ b/devito/arch/compiler.py
@@ -612,7 +612,7 @@ class CustomCompiler(Compiler):
         platform = kwargs.pop('platform', configuration['platform'])
 
         if any(i in environ for i in ['CC', 'CXX', 'CFLAGS', 'LDFLAGS']):
-            obj = super().__new__(cls, *args, **kwargs)
+            obj = super().__new__(cls)
             obj.__init__(*args, **kwargs)
             return obj
         elif platform is M1:


FabioLuporini:
I, thanks for this. I think we should just drop the `*args` ?

AtilaSaraiva:
well maybe we should, I'm only concerned about this CI check that failed on OSX. Any clue what's that about?

FabioLuporini:
Sometimes the osx build for the jupyter notebooks does that (@mloubout ). Nothing related to this PR I think!

AtilaSaraiva:
I see, then I'm glad to help. I actually a geophysicist so I'll probably be supporting more in the seismic modelling front, but I do have some ideas for the more devops front. Hope we can work towards a better software overall.

FabioLuporini:
absolutely -- thanks for this! We do welcome and love contributions from everybody. See also: https://github.com/devitocodes/devito/blob/master/CONTRIBUTING.md

BTW, I'm changing the PR tag from "ir" to "arch"

FabioLuporini:
also edited the PR title to reflect better what's it actually doing

mloubout:
Always happy to get new people on board. As @FabioLuporini said the CI fail isn't relevant just osx being not always nice. I think `**kwargs` can stay too and should be GTG thanks.

Fixes #1739 I think.

AtilaSaraiva:
> absolutely -- thanks for this! We do welcome and love contributions from everybody. See also: https://github.com/devitocodes/devito/blob/master/CONTRIBUTING.md
> 
> BTW, I'm changing the PR tag from "ir" to "arch"

No prob, I was actually confused to which tag to use haha.



> Always happy to get new people on board. As @FabioLuporini said the CI fail isn't relevant just osx being not always nice. I think `**kwargs` can stay too and should be GTG thanks.
> 
> Fixes #1739 I think.

Let me test it real quick.

AtilaSaraiva:
Seems golden @mloubout.  If u guys could squash the first commit with the wrong name and use the last one I'd appreciate it.

@FabioLuporini even those tests that were giving the wrong norm I posted on slack are working fine now with this **kwargs argument, have no idea why though.

mloubout:
>Hm, after testing using **kwargs along with the cls argument, I've encountered this error:

Ok well I guess kwargs need to be removed as well, thanks for checking

AtilaSaraiva:
yeah, it seems like it.

PR #1864, opened 2022-03-08
compiler: Patch norm accumulation

### LINKED ISSUE #1863: improve handling of precision promotion for norm calculation
in devito/builtins/arithmetic.py 31, the accumulator is set to float64 to reduce roundoff, in effect assuming the array type is float32.

If f.type is int this convert to floating arithmetic might be unexpected.

What happens if f.type is already float64? Then you'd really want to promote to float128.

An alternative approach is to add an xarg to allow the caller to specify the type of the accumulator.

Another approach is to add a function that returns the promoted type for a given type, ie
float8 -> float16
float16 -> float32
float32 -> float64
float64 -> float128

FabioLuporini:
fixes #1863

### DIFF
--- a/devito/builtins/arithmetic.py
+++ b/devito/builtins/arithmetic.py
@@ -6,6 +6,21 @@ from devito.builtins.utils import MPIReduction
 
 __all__ = ['norm', 'sumall', 'inner', 'mmin', 'mmax']
 
+accumulator_mapper = {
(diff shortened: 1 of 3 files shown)


mloubout (approved):

georgebisbas (approved):

ggorman (approved):
GTG

PR #1870, opened 2022-03-11
compiler: Add GuardOverflow

FabioLuporini:
Used in DevitoPRO but added here because 1) why not and 2) I suspect I'm going to use it in the near future for other stuff

### DIFF
--- a/devito/ir/support/guards.py
+++ b/devito/ir/support/guards.py
@@ -4,7 +4,7 @@ of the compiler to express the conditions under which a certain object
 (e.g., Eq, Cluster, ...) should be evaluated at runtime.
 """
 
-from sympy import Le, Gt, Mul, true
+from sympy import Ge, Gt, Le, Lt, Mul, true
 from sympy.core.operations import LatticeOp
 
 from devito.ir.support.space import Forward, IterationDirection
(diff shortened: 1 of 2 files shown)


mloubout (approved):

devito/ir/support/guards.py
```diff
@@ -194,5 +214,6 @@ def transform_guard(expr, guard_type, callback):
     GuardBoundLe: GuardBoundGt,
     GuardBoundGt: GuardBoundLe,
     GuardBoundNextLe: GuardBoundNextGt,
-    GuardBoundNextGt: GuardBoundNextLe
+    GuardBoundNextGt: GuardBoundNextLe,
+    GuardOverflow: GuardOverflowLt
```
georgebisbas, line 218: I think for homogeneity we should have:

GuardOverflowGe: GuardOverflowLt
GuardOverflowLt: GuardOverflowGe

georgebisbas (commented):
Two very small-small-minor comments, otherwise GTG

devito/ir/support/guards.py
```diff
@@ -4,15 +4,15 @@
 (e.g., Eq, Cluster, ...) should be evaluated at runtime.
 """
 
-from sympy import Le, Gt, Mul, true
+from sympy import Le, Lt, Ge, Gt, Mul, true
```
georgebisbas, line 7: AB order?

devito/ir/support/guards.py
FabioLuporini: fixed

georgebisbas (approved):

PR #1871, opened 2022-03-14
compiler: Drop unnecessary check_indices

FabioLuporini:


mloubout (approved):
Should be tagged as API not compiler

georgebisbas (approved):

FabioLuporini:
> Should be tagged as API, not compiler

I had changed it to API, but now back to compiler, because I just realized that `_check_indices` is a private method used inside for lowering purposes.

PR #1872, opened 2022-03-14
compiler: Move derivative.evaluate

FabioLuporini:
This is in preparation for the "derivative evaluation overhaul" (DEO from now on!) that we've talked about recently.

-- A glimpse of the future --

I will need to introduce an `evaluate_partial` to `Derivative` (I cannot touch `evaluate` because we don't want API-level surprises). Such `evaluate_partial` (better name needed) will expand the `Derivative` to an intermediate high-level object, for example

```
Derivative(f, x).dx.evaluate_partial
->
DotDerivative(f, coeff)
```

Representing the dot product between `f` and an array of coefficients, that is, a "rolled" Derivative. This will be the object used by Devito throughout compilation and in particular throughout (and until) Cluster specialization.

So, back to this PR... moving that `.evaluate` will allow me to share the first two steps of `Derivative.evaluate` with `Derivative.evaluate_partial` by moving them to a common function, or perhaps better, by having `Derivative.evaluate` calling `Derivative.evaluate_partial` first. Not sure yet, still very much work in progress, and I'm doing this only part time

### DIFF
--- a/devito/finite_differences/finite_difference.py
+++ b/devito/finite_differences/finite_difference.py
@@ -280,11 +280,19 @@ def indices_weights_to_fd(expr, dim, inds, weights, matvec=1):
             iloc = sympify(i).xreplace(mapper)
         # Enforce fixed precision FD coefficients to avoid variations in results
         c = sympify(c).evalf(_PRECISION)
-        terms.append(expr._subs(dim, iloc - (expr.indices_ref[dim] - dim)) * c)
+
+        # The FD term
+        term = expr._subs(dim, iloc - (expr.indices_ref[dim] - dim)) * c
+
+        # Re-evaluate any off-the-grid Functions potentially impacted by the FD
+        try:
+            term = term.evaluate
+        except AttributeError:
+            # Pure number
+            pass
+
+        terms.append(term)
 
     deriv = EvalDerivative(*terms, base=expr)
-
-    # Evaluate remaining part of expression
-    deriv = deriv.evaluate
 
     return deriv
--- a/devito/finite_differences/finite_difference.py
+++ b/devito/finite_differences/finite_difference.py
@@ -284,4 +284,7 @@ def indices_weights_to_fd(expr, dim, inds, weights, matvec=1):
 
     deriv = EvalDerivative(*terms, base=expr)
 
+    # Evaluate remaining part of expression
+    deriv = deriv.evaluate
+
     return deriv
--- a/devito/finite_differences/derivative.py
+++ b/devito/finite_differences/derivative.py
@@ -325,14 +325,15 @@ class Derivative(sympy.Derivative, Differentiable):
     def _eval_fd(self, expr):
         """
         Evaluate finite difference approximation of the Derivative.
-        Evaluation is carried out via the following four steps:
+        Evaluation is carried out via the following three steps:
 
         - 1: Evaluate derivatives within the expression. For example given
             `f.dx * g`, `f.dx` will be evaluated first.
         - 2: Evaluate the finite difference for the (new) expression.
-        - 3: Evaluate remaining terms (as `g` may need to be evaluated
-        at a different point).
-        - 4: Apply substitutions.
+             This in turn is a two-step procedure, for Functions that may
+             may need to be evaluated at a different point due to e.g. a
+             shited derivative.
+        - 3: Apply substitutions.
         """
         # Step 1: Evaluate derivatives within expression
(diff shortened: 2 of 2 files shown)


devito/finite_differences/finite_difference.py
```diff
@@ -284,4 +284,7 @@ def indices_weights_to_fd(expr, dim, inds, weights, matvec=1):
 
     deriv = EvalDerivative(*terms, base=expr)
 
+    # Evaluate remaining part of expression
+    deriv = deriv.evaluate
```
mloubout, line 288: AS I said this is dangerous because now you can prevent shifts cancelling out and by forcing the "avg" to be evaluated this can lead to quite worse accuracy. This evaluate should always be the most last part in case there is half-node indices left.

FabioLuporini: there are two issues here:

1) there's no failing test in the test suite capturing the behavior you're describing. Could you add one? Pushing to this PR would be completely fine

2) 
> This evaluate should always be the most last part in case there is half-node indices left.

I'm not sure I see how my change would alter the order? The `return deriv` below would return control to `derivative.py._eval_fd`, so... what am I missing? perhaps with the MFE I'll see it

mloubout: my brain is not working very well these days, it doesn't change anything to have moved it.

mloubout (approved):

devito/finite_differences/finite_difference.py
FabioLuporini: OK, no problem

devito/finite_differences/finite_difference.py
```diff
@@ -280,11 +280,19 @@ def indices_weights_to_fd(expr, dim, inds, weights, matvec=1):
...
+        term = expr._subs(dim, iloc - (expr.indices_ref[dim] - dim)) * c
+
+        # Re-evaluate any off-the-grid Functions potentially impacted by the FD
+        try:
+            term = term.evaluate
+        except AttributeError:
+            # Pure number
+            pass
+
+        terms.append(term)
```
mloubout, line 294: you could probably postponn the `*c` here as well, may be slightly more efficient (one less `Mul` built and evaluated). This also prevents the full travers of EvalDerivative so good catch

FabioLuporini: Good catch. Note that this horribly conflicts with the changes in #1875 😬  so I'm going to rebase #1875 now on top of this one and will make the change there; I'll tag you there so that later you can give me green light here

FabioLuporini: I tried, but I'm getting (expected) unevaluation errors:

```
self = <test_derivatives.TestFD object at 0x109a5dfa0>

    def test_fd_new_lo(self):
        grid = Grid((10,))
        x = grid.dimensions[0]
        u = Function(name="u", grid=grid, space_order=2)

        dplus = "-1.0*u(x)/h_x + u(x + 1.0*h_x)/h_x"
        dminus = "u(x)/h_x - 1.0*u(x - 1.0*h_x)/h_x"
>       assert str(u.dx(x0=x + .5 * x.spacing).evaluate) == dplus
E       AssertionError: assert '-1.0*u(x)/h_... 1.0*h_x)/h_x' == '-1.0*u(x)/h_... 1.0*h_x)/h_x'
E         - -1.0*u(x)/h_x + u(x + 1.0*h_x)/h_x
E         + -1.0*u(x)/h_x + 1.0*u(x + 1.0*h_x)/h_x
E         ?                 ++++

test_derivatives.py:368: AssertionError
```

mloubout: Looks like it's introducing a weird `1.0*` for some reason sympy being annoying again

devito/finite_differences/finite_difference.py
```diff
@@ -280,7 +280,18 @@ def indices_weights_to_fd(expr, dim, inds, weights, matvec=1):
             iloc = sympify(i).xreplace(mapper)
         # Enforce fixed precision FD coefficients to avoid variations in results
         c = sympify(c).evalf(_PRECISION)
-        terms.append(expr._subs(dim, iloc - (expr.indices_ref[dim] - dim)) * c)
+
+        # The FD term
+        term = expr._subs(dim, iloc - (expr.indices_ref[dim] - dim)) * c
+
+        # Re-evaluate any off-the-grid Functions potentially impacted by the FD
```
georgebisbas, line 287: off-the-grid!

georgebisbas (approved):

PR #1874, opened 2022-03-15
mpi: Do not send padding region when doing pythonland halo exchange

tjb900:
This change is both a tiny optimisation and also a fix for a potential correctness issue in the case where the application provides a customised padding, since the provided padding is not required to be the same on all ranks.

There are three commits here which hopefully make sense. 

* add a timeout to MPI tests
* add a test which fails with the two ranks having different padding (it hangs without the first commit)
* a fix

The timeout one can be a separate PR if you like, or we can drop it altogether. It's needed for the tests to behave as expected (i.e. fail) with the second commit, but after the third commit they pass again (and obviously don't hang) so I don't mind either way. 

I'm also not sure if 30s is an appropriate timeout, or if we should make it e.g. an environment variable to be set in the test environment script?

### DIFF
--- a/devito/types/dense.py
+++ b/devito/types/dense.py
@@ -565,8 +565,11 @@ class DiscreteFunction(AbstractFunction, ArgProvider, Differentiable):
         self._is_halo_dirty = True
         offset = getattr(getattr(self, '_offset_%s' % region.name)[dim], side.name)
         size = getattr(getattr(self, '_size_%s' % region.name)[dim], side.name)
-        index_array = [slice(offset, offset+size) if d is dim else slice(None)
-                       for d in self.dimensions]
+        index_array = [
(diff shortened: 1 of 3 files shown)


FabioLuporini (approved):
Uncontroversial. Many thanks for this.

FabioLuporini:
@tjb900 the only thing maybe is that the timeout is a bit too short as IIRC there are a few tests that will take much longer

@mloubout thoughts?

mloubout:
I'll double check the logs to see if there is some test that are borderline but seems to all pass so timeout must be long enough. 
Thanks @tjb900

mloubout:
Everything looks legit in the logs. And I don't think the timeout can actually cause issue, since it only concerns non-responding mpi not long running script with working comm (in my understanding)

FabioLuporini:
Thank you again @tjb900. Merging.

PR #1875, opened 2022-03-15
compiler: Minor FD refactorings

FabioLuporini:
Not sure you'll like everything @mloubout ... anyway, opened for debate

### DIFF
--- a/devito/finite_differences/derivative.py
+++ b/devito/finite_differences/derivative.py
@@ -314,8 +314,8 @@ class Derivative(sympy.Derivative, Differentiable):
     @property
     def evaluate(self):
         # Evaluate finite-difference.
-        # Note: evaluate and _eval_fd splitted for potential future different
-        # types of discretizations.
+        # NOTE: `evaluate` and `_eval_fd` split for potential future different
+        # types of discretizations
         return self._eval_fd(self.expr)
 
     @property
@@ -324,7 +324,7 @@ class Derivative(sympy.Derivative, Differentiable):
 
     def _eval_fd(self, expr):
         """
-        Evaluate finite difference approximation of the Derivative.
+        Evaluate the finite-difference approximation of the Derivative.
         Evaluation is carried out via the following three steps:
 
         - 1: Evaluate derivatives within the expression. For example given
@@ -336,7 +336,10 @@ class Derivative(sympy.Derivative, Differentiable):
         - 3: Apply substitutions.
         """
         # Step 1: Evaluate derivatives within expression
-        expr = getattr(expr, '_eval_deriv', expr)
+        try:
+            expr = expr._eval_deriv
+        except AttributeError:
+            pass
 
         # Step 2: Evaluate FD of the new expression
         if self.side is not None and self.deriv_order == 1:
--- a/devito/finite_differences/derivative.py
+++ b/devito/finite_differences/derivative.py
@@ -2,6 +2,7 @@ from collections import OrderedDict
 from collections.abc import Iterable
 
 from cached_property import cached_property
+import numpy as np
 import sympy
 
 from devito.finite_differences.finite_difference import (generic_derivative,
@@ -10,9 +11,11 @@ from devito.finite_differences.finite_difference import (generic_derivative,
 from devito.finite_differences.differentiable import Differentiable
 from devito.finite_differences.tools import direct, transpose
 from devito.tools import as_mapper, as_tuple, filter_ordered, frozendict
+from devito.types.array import Array
+from devito.types.dimension import StencilDimension
 from devito.types.utils import DimensionTuple
 
-__all__ = ['Derivative']
+__all__ = ['Derivative', 'Weights']
 
 
 class Derivative(sympy.Derivative, Differentiable):
@@ -358,3 +361,29 @@ class Derivative(sympy.Derivative, Differentiable):
             res = res.xreplace(e)
 
         return res
+
+
+class Weights(Array):
+
+    """
+    The weights (or coefficients) of a finite-difference expansion.
+    """
+
+    def __init_finalize__(self, *args, **kwargs):
+        dimensions = as_tuple(kwargs.get('dimensions'))
+        weights = kwargs.get('initvalue')
+
+        assert len(dimensions) == 1
+        d = dimensions[0]
+        assert isinstance(d, StencilDimension) and d.symbolic_size == len(weights)
+        assert isinstance(weights, (list, tuple, np.ndarray))
+
+        kwargs['scope'] = 'static'
+
+        super().__init_finalize__(*args, **kwargs)
+
+    @property
+    def dimension(self):
+        return self.dimensions[0]
+
+    weights = Array.initvalue
--- a/devito/finite_differences/differentiable.py
+++ b/devito/finite_differences/differentiable.py
@@ -10,11 +10,11 @@ from sympy.core.evalf import evalf_table
 from cached_property import cached_property
 from devito.finite_differences.tools import make_shift_x0
 from devito.logger import warning
-from devito.tools import filter_ordered, flatten, split
+from devito.tools import as_tuple, filter_ordered, flatten, is_integer, split
 from devito.types.lazy import Evaluable
 from devito.types.utils import DimensionTuple
 
-__all__ = ['Differentiable', 'EvalDerivative']
+__all__ = ['Differentiable', 'DotDerivative', 'EvalDerivative']
 
 
 class Differentiable(sympy.Expr, Evaluable):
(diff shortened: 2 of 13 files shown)


devito/finite_differences/differentiable.py
```diff
@@ -469,6 +469,64 @@ class Mod(DifferentiableOp, sympy.Mod):
     __sympy_class__ = sympy.Mod
 
 
+class IndexSum(DifferentiableOp):
```
FabioLuporini, line 472: Still unused... but I think you see where I'm going with this... I shall add some tests, will do tmr

devito/finite_differences/finite_difference.py
```diff
@@ -235,52 +173,59 @@ def generic_derivative(expr, dim, fd_order, deriv_order, symbolic=False,
...
-        # Apply replacements to indices. Needs to be sympified in case
-        # the indices are pure numbers
+    for i, c in zip(indices, weights):
+        # Transpose FD if necessary through `mapper`
         try:
             iloc = i.xreplace(mapper)
         except AttributeError:
+            # Pure number -> sympify
             iloc = sympify(i).xreplace(mapper)
+        # Shift index due to staggering, if any
```
FabioLuporini, line 222: Is this comment correct? just to be sure...

devito/types/dimension.py
```diff
@@ -1058,6 +1052,80 @@ class IncrDimension(AbstractIncrDimension):
     is_SubIterator = True
 
 
+class BlockDimension(AbstractIncrDimension):
```
FabioLuporini, line 1055: This is just reshuffled code, nothing new under the sun

devito/finite_differences/differentiable.py
```diff
@@ -469,6 +469,64 @@ class Mod(DifferentiableOp, sympy.Mod):
...
+        obj = sympy.Expr.__new__(cls, expr)
+        obj.dimensions = dimensions
+
+        return obj
+
+
+class Dot(IndexSum):
+
+    """
+    Dot-product, as the IndexSum of a binary Mul.
```
mloubout, line 510: Would probably say instead that this is a symbolic dot product of two symbolic vectors and add a little example

mloubout (commented):
Mostly comments and discussion points but all looks good to me.

FabioLuporini:
@mloubout addressed to minor comments, I'll work on IndexSub and related stuff tomorrow (will ad simple tests for the symbolics at least)

georgebisbas (commented):
Not sure StencilDimension is a good name...but will review again tmr

devito/finite_differences/differentiable.py
```diff
@@ -469,6 +469,64 @@ class Mod(DifferentiableOp, sympy.Mod):
...
+        if not dimensions:
+            return expr
+        for d in dimensions:
+            try:
+                if d.is_Dimension and is_integer(d.symbolic_size):
+                    continue
+            except AttributeError:
+                pass
+            raise ValueError("Expected Dimension with numeric size, "
+                             "got `%s` instead" % d)
```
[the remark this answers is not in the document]
FabioLuporini: There are a few options in theory... even DefaultDimension and CustomDimension would work. Not that we care about those at this point, though...

devito/finite_differences/differentiable.py
```diff
@@ -469,6 +469,64 @@ class Mod(DifferentiableOp, sympy.Mod):
...
+        for d in dimensions:
+            if d not in expr.free_symbols:
```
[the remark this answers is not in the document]
FabioLuporini: hence not going to use it unless we bump the SymPy requirement, which we do not want to do?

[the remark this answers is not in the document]
mloubout:
```
In [12]: expr = div(u.laplace * u).evaluate

In [9]: expr.free_symbols
Out[9]: {h_x, h_y, t, x, y}

In [10]: %timeit x in expr.free_symbols
1.44 ms ± 114 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [11]: %timeit expr.has_free(x)
1.38 µs ± 74.1 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

I disagree the difference in perf is huge (that's a X1000) and it's just the little things here and there. Should also change it throughout

devito/finite_differences/finite_difference.py
mloubout: yes

devito/types/dimension.py
```diff
@@ -1187,6 +1255,38 @@ def _symbolic_thickness(cls, name):
                 Scalar(name="%s_rtkn" % name, dtype=np.int32, nonnegative=True))
 
 
+class StencilDimension(BasicDimension):
+
+    def __init_finalize__(self, name, size):
+        self._spacing = sympy.S.One
```
mloubout, line 1261: may need start/end in case non-symmetric

FabioLuporini: yepper, I have local edits that I haven't pushed yet which do exactly that

devito/finite_differences/differentiable.py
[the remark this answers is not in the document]
FabioLuporini: are you suggesting to override `has_free` in `Differentiable` ?

```
def has_free(self, x):
    try:
         return super().has_free(x)
    except AttributeError:
         <for loops>
```

?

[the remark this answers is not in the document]
mloubout: Either that or a generic util function for all our sympy 

```
def has_free(expr, x):
    try:
         return expr.has_free(x)
    except AttributeError:
         <for loops>
```

Probably not for this PR but could be good. I know it's not  used that much but a gain is a gain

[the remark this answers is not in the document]
FabioLuporini: I don't think I like a free function, but I'm happy to add a method for now as per my previous message... then we can take it from there

devito/finite_differences/differentiable.py
```diff
@@ -469,6 +470,91 @@ class Mod(DifferentiableOp, sympy.Mod):
...
+                raise ValueError("Dimension `%s` must appear in `expr`" % d)
+        for i in expr.find(IndexSum):
+            for d in dimensions:
```
georgebisbas, line 497: You can shorten this with if any(d in i.dims... for d in ...)

devito/finite_differences/derivative.py
```diff
@@ -355,3 +361,29 @@ def _eval_fd(self, expr):
...
+    """
+
+    def __init_finalize__(self, *args, **kwargs):
+        dimensions = as_tuple(kwargs.get('dimensions'))
+        weights = kwargs.get('initvalue')
+
+        assert len(dimensions) == 1
+        d = dimensions[0]
+        assert isinstance(d, StencilDimension) and d.symbolic_size == len(weights)
+        assert isinstance(weights, (list, tuple, np.ndarray))
```
mloubout, line 379: So what's the plan when the stencil spans multiple dimension?

devito/finite_differences/differentiable.py
mloubout: Considering what this does and how it's defined, could we name it something a bit more "explicit",like `WeightedContraction` ? current name is a bit odd

devito/finite_differences/differentiable.py
```diff
@@ -469,6 +470,91 @@ class Mod(DifferentiableOp, sympy.Mod):
     __sympy_class__ = sympy.Mod
 
 
+class IndexSum(DifferentiableOp):
+
+    """
+    A tensor contraction over one or more Dimensions.
```
mloubout, line 476: weighted

devito/finite_differences/differentiable.py
```diff
@@ -469,6 +470,91 @@ class Mod(DifferentiableOp, sympy.Mod):
...
+        return self._expr
+
+    @property
+    def dimensions(self):
+        return self._dimensions
+
+    @property
+    def evaluate(self):
+        # Overrides sympy.Expr.expand
+        values = product(*[list(range(d._min, d._max + 1)) for d in self.dimensions])
```
mloubout, line 525: can't have a `step>` ? i.e a stencil using every ther two offsets

mloubout: Would probably be better as `d.indices/range/values` with the `StencilDimension` having its own range defined as a property. Or even just return the mapper?

devito/finite_differences/differentiable.py
```diff
@@ -469,6 +470,91 @@ class Mod(DifferentiableOp, sympy.Mod):
...
+        terms = []
+        for i in values:
+            mapper = dict(zip(self.dimensions, i))
+            terms.append(self.expr.xreplace(mapper))
+        return sum(terms)
+
+
+class IndexDerivative(IndexSum):
```
mloubout, line 533: Index or Indexed? Not sure need "Derivative" no? This is just a standard "stencil expression"

devito/finite_differences/differentiable.py
```diff
@@ -469,6 +470,91 @@ class Mod(DifferentiableOp, sympy.Mod):
...
+    def weights(self):
+        return self._weights
+
+    @property
+    def evaluate(self):
+        expr = super().evaluate
+
+        w = self.weights
+        d = w.dimension
+        mapper = {w.subs(d, i): w.weights[n]
```
mloubout, line 551: Same would make it a `weight` property with a `weight.mapper` and `d.range`

devito/finite_differences/finite_difference.py
```diff
@@ -235,54 +173,60 @@ def generic_derivative(expr, dim, fd_order, deriv_order, symbolic=False,
...
+        iloc -= expr.indices_ref[dim] - dim
+
         # Enforce fixed precision FD coefficients to avoid variations in results
-        c = sympify(c).evalf(_PRECISION)
+        v = sympify(c).evalf(_PRECISION)
```
mloubout, line 226: why `c -> v` c was for "coefficient"

devito/types/dimension.py
```diff
@@ -1193,6 +1261,59 @@ def _symbolic_thickness(cls, name):
...
+    name : str
+        Name of the dimension.
+    _min : expr-like
+        The minimum point of the stencil.
+    _max : expr-like
+        The maximum point of the stencil.
+    """
+
+    def __init_finalize__(self, name, _min, _max):
+        self._spacing = sympy.S.One
```
mloubout, line 1280: would make this oneinput as well for non unitary steps

devito/finite_differences/differentiable.py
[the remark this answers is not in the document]
FabioLuporini: Check out the new commits. I'm now using `has_free` as per the above discussion.

Also fixed `IndexSum.free_symbols` and new test added

devito/finite_differences/differentiable.py
FabioLuporini: > can't have a step> ? i.e a stencil using every ther two offsets

yeah, we could add that feature, but that'd be equivalent to adding the multiplier in the expression where the StencilDimension appears

> Would probably be better as d.indices/range/values with the StencilDimension having its own range defined as a property. Or even just return the mapper?

Adding `range` property now

devito/finite_differences/differentiable.py
FabioLuporini: It needs "Derivative" because it will ultimately be used for unexpanded derivatives in place of the fully unrolled Adds (`EvalDerivative`) that we generate today.

I'm honestly not sure about the name, but since this is far from the final state, I think we can postpone or discuss offline on Slack to converge to a better name

devito/finite_differences/finite_difference.py
FabioLuporini: because you should not override the iteration variables

devito/types/dimension.py
FabioLuporini: Right. Done and added test

devito/finite_differences/differentiable.py
FabioLuporini: Now using `d.range`, but not `weight.mapper` because it sounds like overkill (I only expect it to be needed here)

mloubout (approved):

PR #1876, opened 2022-03-15
Add xdsl as requirement

tobiasgrosser:


mloubout:
Thanks. Two small questions (@FabioLuporini ):
- how big is xdsl and are we sure there isn't any requirement conflicts. 
- Do we want to make a separate one (like nvidia) in case need more stuff down the road?

tobiasgrosser:
Sorry, did not want to propose this to devito.

PR #1878, opened 2022-03-15
compiler: Do not evaluate unevaluable expressions

mloubout:
diff2sympy patch in case the expresssion is already a non-evaluable sympy expression

### DIFF
--- a/devito/finite_differences/differentiable.py
+++ b/devito/finite_differences/differentiable.py
@@ -587,7 +587,12 @@ def diff2sympy(expr):
             # Won't lower (e.g., EvalDerivative)
             pass
         if flag:
-            return obj.func(*args, evaluate=False), True
+            try:
+                return obj.func(*args, evaluate=False), True
+            except TypeError:
+                # In case of indices using other Function, evaluate
+                # may not be a supported argument.
+                return obj.func(*args), True
         else:
             return obj, False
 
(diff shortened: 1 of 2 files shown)


devito/finite_differences/differentiable.py
```diff
@@ -587,7 +587,12 @@ def _diff2sympy(obj):
             # Won't lower (e.g., EvalDerivative)
             pass
         if flag:
-            return obj.func(*args, evaluate=False), True
+            try:
+                return obj.func(*args, evaluate=False), True
+            except TypeError:
+                # In case of indices using other Function evaluate
+                # may not be a asupported argument.
```
georgebisbas, line 594: typo

tests/test_lower_exprs.py
```diff
@@ -181,3 +183,20 @@ def test_nocollection_staggered(self):
         leq = collect_derivatives.func([eq])[0]
 
         assert eq == leq
+
+    def test_lower_func_as_ind(self):
```
FabioLuporini, line 187: You should create another class because this is specifically about the compiler pass "collect_derivatives". Since this is a test for a pice of code that is executed unconditionally, I'd put it into a "TestLowering" class at the top of the file

FabioLuporini (commented):
Virtually uncontroversial -- The test should be moved to a separate class and there's a typo as per comments, but other than that all good

PR #1880, opened 2022-03-16
compiler: Avoid generating unnecessary conditionals

FabioLuporini:


mloubout (approved):
GTG. Is there a basic example/test for it or is it for some non trivial case?

georgebisbas (approved):
Agree with Mathias for a test, otherwise GTG

FabioLuporini:
I thought I had responded, but my comment not showing up? 
Anyway...

> or is it for some non trivial case?

The opposite -- this tiny patch is just for super trivial cases and as I said it's just about the aesthetics of the generated code, to avoid generating uselss

```
if (true)
```

FabioLuporini:
I thought I had responded, but my comment not showing up? 
Anyway...

> or is it for some non trivial case?

The opposite -- this tiny patch is just for super trivial cases and as I said it's just about the aesthetics of the generated code, to avoid generating uselss

```
if (true)
```

PR #1883, opened 2022-03-21
compiler: Fix guards fusion

FabioLuporini:
Patch to correctly fuse guards outside of any loops

Instead of

```
if (A)
   for x
if (A)
   for w
```

it now correctly generates

```
if (A)
  for x
  for w
```

Also added an API to express such loop-free guards (so far they were only generated internally by the compiler).

Tests should be self-explanatory.

### DIFF
--- a/devito/ir/clusters/algorithms.py
+++ b/devito/ir/clusters/algorithms.py
@@ -283,7 +283,7 @@ class Stepper(Queue):
         def rule(size, e):
             try:
                 return e.function.shape_allocated[d] == size
-            except (AttributeError, KeyError):
+            except (AttributeError, KeyError, ValueError):
                 return False
 
         # Reconstruct the Clusters
--- a/devito/ir/clusters/analysis.py
+++ b/devito/ir/clusters/analysis.py
@@ -134,7 +134,12 @@ class Rounding(Detector):
         # Note: autopadding guarantees that the padding size along the
         # Fastest Varying Dimension is a multiple of the SIMD vector length
         functions = [f for f in scope.functions if f.is_AbstractFunction]
-        if any(not f._honors_autopadding for f in functions):
+        try:
+            if any(not f._honors_autopadding for f in functions):
+                return
+        except ValueError:
+            # E.g., lazily allocated Functions don't have an accessible
+            # `.shape` until after the first call to `f._arg_values`
             return
 
         # Mixed data types (e.g., float and double) is unsupported


tests/test_dimension.py
```diff
@@ -942,6 +942,27 @@ def test_grouping(self):
...
+        within the same (also shared) ConditionalDimension.
+        """
+        grid = Grid(shape=(10, 10))
+        time = grid.time_dim
+        cond = ConditionalDimension(name='cond', parent=None, condition=time < 5)
+
+        u = TimeFunction(name='u', grid=grid)
+        v = TimeFunction(name='v', grid=grid)
+
+        eqns = [Eq(u.forward, u + 1., implicit_dims=[cond]),
```
georgebisbas, line 957: No [ ] needed in implicit_dims

devito/ir/stree/algorithms.py
```diff
@@ -35,9 +36,13 @@ def stree_schedule(clusters):
     """
     stree = ScheduleTree()
 
-    prev = None
+    prev = Cluster(None)
     mapper = DefaultOrderedDict(lambda: Bunch(top=None, bottom=None))
 
+    def reuse_metadata(c0, c1, d):
+        return (c.guards.get(d) == prev.guards.get(d) and
```
georgebisbas, line 43: what I am i missing?
should not c/prev be c0/c1 ?

FabioLuporini: indeed, CI failing miserably ! copy-paste mistake in my last commit. Fixing now...

devito/types/dimension.py
```diff
@@ -755,7 +755,13 @@ class ConditionalDimension(DerivedDimension):
 
     def __init_finalize__(self, name, parent, factor=None, condition=None,
```
georgebisbas, line 756: parent=None ?

tests/test_dimension.py
```diff
@@ -942,6 +942,27 @@ def test_grouping(self):
...
         assert len(retrieve_iteration_tree(conds[0].then_body)) == 2
 
+    def test_grouping_v2(self):
+        """
+        Test that two equations sharing a strict subset of loops get scheduled
+        within the same (also shared) ConditionalDimension.
+        """
+        grid = Grid(shape=(10, 10))
+        time = grid.time_dim
+        cond = ConditionalDimension(name='cond', parent=None, condition=time < 5)
```
georgebisbas, line 952: So with the above you can avoid adding None as an arg here?
So we can eventually define COnditional Dimension without specifying parent at all?

FabioLuporini: good catch, fixed

devito/types/dimension.py
FabioLuporini: yep, fixed

tests/test_dimension.py
FabioLuporini: fixed

devito/types/dimension.py
```diff
@@ -753,9 +753,15 @@ class ConditionalDimension(DerivedDimension):
     is_NonlinearDerived = True
     is_Conditional = True
 
-    def __init_finalize__(self, name, parent, factor=None, condition=None,
+    def __init_finalize__(self, name, parent=None, factor=None, condition=None,
                           indirect=False):
+        # `parent=None` degenerates to a ConditionalDimension outside of
+        # any iteration space
```
georgebisbas, line 759: outside of any iteration space -->
with an empty IterationSpace
?

FabioLuporini: no, we obv don't know about IterationSpaces here

georgebisbas (approved):

devito/ir/stree/algorithms.py
```diff
@@ -64,14 +73,15 @@ def attach_metadata(cluster, d, tip):
...
             # new properties
-            mapper[it0].top.ispace = IterationSpace.union(mapper[it0].top.ispace,
-                                                          c.ispace.project([d]))
-            mapper[it0].top.properties = normalize_properties(mapper[it0].top.properties,
-                                                              c.properties[it0.dim])
-
-            # Different guards or syncops cannot be further nested
-            if c.guards.get(d) != prev.guards.get(d) or \
-               c.syncs.get(d) != prev.syncs.get(d):
+            mapper[it0].top.ispace = IterationSpace.union(
```
FabioLuporini, line 76: Aesthetics...

FabioLuporini:
Also approved by @mloubout offline, merging

PR #1887, opened 2022-03-23
WIP xDSL notebook

superlopuh:
Adds a jupyter notebook with some prototype code to translate a devito node into an xdsl structure.


superlopuh:
Should have been a PR in the cloned repository

PR #1890, opened 2022-03-24
CI-Docker: Publish CPU image with tag latest

navjotk:
I think this is the only change needed. Now we publish CPU images under 3 tags - `devitocodes/devito:latest`, `devitocodes/devito:cpu-latest`, `devitocodes/devito:version`. 

So the command `docker pull devitocodes/devito` which maps automatically to `devitocodes/devito:latest`, should successfully download the latest cpu image.


mloubout (approved):

FabioLuporini:
Thank you, merged!

PR #1898, opened 2022-04-08
compiler: Patch iteration space scheduling

FabioLuporini:


tests/test_operator.py
```diff
@@ -1929,3 +1930,22 @@ def test_scheduling_with_free_dims(self):
         assert tree[0].dim is time
         assert tree[1].dim is x
         assert tree[2].dim is y
+
+    def test_issue_1897(self):
+        grid = Grid(shape=(11, 11, 11))
+
+        v = VectorTimeFunction(name='v', grid=grid, time_order=1, space_order=4)
```
georgebisbas, line 1937: Orders can be dropped..space_order, time_order

georgebisbas (commented):
Should a GPU test be added as well? Or redundant? It shouldn't;t fail I guess, but since John spotted a similar behaviour with tile(...) just saying...

jkwashbourne (approved):
Wow that is a fast fix! Much impressed.

devito/ir/clusters/algorithms.py
```diff
@@ -84,7 +84,7 @@ class Schedule(QueueStateful):
 
     @timed_pass(name='schedule')
     def process(self, clusters):
-        return self._process_fdta(clusters, 1)
+        return self._process_fatd(clusters, 1)
```
mloubout, line 87: Was really just a typo? 😮

mloubout (approved):

devito/ir/clusters/algorithms.py
mloubout: Not kidding was just wondering if was intentional to have `fdta` or if was meant to be `fatd` from the beginning but wasn't noticed. anyway all good

PR #1899, opened 2022-04-08
deps: Drop deprecated distutils imports

georgebisbas:
Had this locally (maybe from 3 months ago?)
did a minor change...PRing for a test...
If anyone wants to continue on this..feel free..

It helps to avoid all the warnings created by this:
https://peps.python.org/pep-0632/#:~:text=In%20Python%203.10%20and%203.11,platforms%20will%20not%20be%20added.

### DIFF
--- a/devito/arch/compiler.py
+++ b/devito/arch/compiler.py
@@ -1,7 +1,7 @@
 from functools import partial
 from hashlib import sha1
 from os import environ, path
-from distutils import version
+from packaging.version import Version
 from subprocess import DEVNULL, PIPE, CalledProcessError, check_output, check_call, run
 import platform
 import warnings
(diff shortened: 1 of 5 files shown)


mloubout:
distutils still used in `setup.py` for install though, should we change it too

mloubout:
@georgebisbas as above we may wanna completely get rid of it i.e cyamfe the `setup.py`. Pip is moving away from distuyils anyway.

georgebisbas:
> anyway.

My understanding is that setuptools is the first option when it comes for versioning. Where do you that distutils is used?

mloubout:
Nevermind confused myself somehow

mloubout (approved):

FabioLuporini (approved):
Excellent. Merging

PR #1903, opened 2022-04-12
compiler: Patch race conditions due to storage-related dependencies

### LINKED ISSUE #1901: Undetected storage-related dependence with buffering in corner cases
When one uses buffering in a dummy Operator with a parallel time dimension there is a race condition due to OpenMP-izing the time loop.

Reproducer:

```
    grid = Grid(shape=(2, 2))
    time = grid.time_dim
    x, y = grid.dimensions

    usave = TimeFunction(name='usave', grid=grid, save=10)
    v = TimeFunction(name='v', grid=grid)

    eq = [Eq(v[time, x, y], usave)]

    op = Operator(eq, opt='buffering')

    trees = retrieve_iteration_tree(op)
    assert len(trees) == 2
    assert trees[1].root.dim is time
    assert not trees[1].root.is_Parallel
    assert trees[1].root.is_Sequential  # Obv
```

The issue is that buffering creates a storage-related dependence and what was previously discovered to be PARALLEL should now be downgraded to SEQUENTIAL

### LINKED ISSUE #1902: Undetected storage-related dependence causes Dimension to be tagged as PARALLEL
Reproducer

```
from devito import *

grid = Grid(shape=(2, 2))
t = grid.stepping_dim
x, y = grid.dimensions

f = Function(name='f', grid=grid)
v = TimeFunction(name='v', grid=grid)

eqns = [
    Eq(f[0, y], u),
    Eq(v, f[0, y])
]

op = Operator(eqns, openmp=True)
print(op)
```

FabioLuporini:
fixes #1901 
fixes #1902

### DIFF
--- a/devito/ir/support/basic.py
+++ b/devito/ir/support/basic.py
@@ -100,8 +100,6 @@ class IterationInstance(LabeledVector):
                 retval.append(dims.pop())
             elif isinstance(i, Dimension):
                 retval.append(i)
-            elif q_constant(i):
-                retval.append(fi)
             else:
                 retval.append(None)
         return DimensionTuple(*retval, getters=self.findices)
@@ -262,10 +260,14 @@ class TimedAccess(IterationInstance):
         # space Dimensions
         positions = []
         for d in self.aindices:
-            for n, i in enumerate(self.intervals):
-                if i.dim._defines & d._defines:
-                    positions.append(n)
-                    break
+            try:
+                for n, i in enumerate(self.intervals):
+                    if i.dim._defines & d._defines:
+                        positions.append(n)
+                        break
+            except AttributeError:
+                # `d is None` due to e.g. constant access
+                continue
         return positions == sorted(positions)
 
     def __lt__(self, other):
@@ -549,6 +551,15 @@ class Dependence(object):
         return self.function.is_Symbol
 
     @memoized_meth
+    def is_const(self, dim):
+        """
+        True if a constant depedence, that is no Dimensions involved, False otherwise.
+        """
+        return (self.source.aindices[dim] is None and
+                self.sink.aindices[dim] is None and
+                self.distance_mapper[dim] == 0)
+
+    @memoized_meth
     def is_carried(self, dim=None):
         """Return True if definitely a dimension-carried dependence, False otherwise."""
         try:
@@ -623,9 +634,10 @@ class Dependence(object):
         cause the access of the same memory location, False otherwise.
         """
         for d in self.findices:
-            if (d._defines & set(as_tuple(dims)) and
-                    any(i.is_NonlinearDerived for i in d._defines)):
-                return True
+            if d._defines & set(as_tuple(dims)):
+                if any(i.is_NonlinearDerived for i in d._defines) or \
+                   self.is_const(d):
+                    return True
         return False
 
 
--- a/devito/passes/clusters/buffering.py
+++ b/devito/passes/clusters/buffering.py
@@ -5,7 +5,8 @@ from cached_property import cached_property
 import numpy as np
 
 from devito.ir import (Cluster, Forward, GuardBound, Interval, IntervalGroup,
-                       IterationSpace, PARALLEL, Queue, Vector, lower_exprs, vmax, vmin)
+                       IterationSpace, PARALLEL, Queue, SEQUENTIAL, Vector,
+                       lower_exprs, normalize_properties, vmax, vmin)
 from devito.exceptions import InvalidOperator
 from devito.logger import warning
 from devito.symbolics import retrieve_function_carriers, uxreplace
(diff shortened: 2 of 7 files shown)


devito/passes/clusters/buffering.py
```diff
@@ -233,7 +243,16 @@ def callback(self, clusters, prefix, cache=None):
                 expr = lower_exprs(uxreplace(Eq(lhs, rhs), b.subdims_mapper))
                 ispace = b.written
 
-                processed.append(c.rebuild(exprs=expr, ispace=ispace))
+                # Buffering creates a storage-related dependence along the
+                # contracted dimensions
+                properties = dict(c.properties)
```
mloubout, line 248: why the duplicate code?

tests/test_buffering.py
```diff
@@ -603,6 +603,25 @@ def test_multi_access():
...
 
+def test_issue_1901():
+    grid = Grid(shape=(2, 2))
+    time = grid.time_dim
+    x, y = grid.dimensions
+
+    usave = TimeFunction(name='usave', grid=grid, save=10)
+    v = TimeFunction(name='v', grid=grid)
+
+    eq = [Eq(v[time, x, y], usave)]
```
mloubout, line 614: nitpicking: `Eq(usave, v)` bit more readable and makes bit more "sense"

mloubout (approved):
Minor comments, but looks GTG

tests/test_buffering.py
FabioLuporini: I want `time` in place of `t` since otherwise the loop will be tagged SEQUENTIAL due to the presence of uindices. But I want it fully PARALLEL to trigger the issue

devito/core/cpu.py
```diff
@@ -143,14 +143,15 @@ class Cpu64NoopOperator(Cpu64OperatorMixin, CoreOperator):
     def _specialize_iet(cls, graph, **kwargs):
         options = kwargs['options']
         platform = kwargs['platform']
+        compiler = kwargs['compiler']
```
mloubout, line 146: Note: not for here but long run, should be gather all of these into a `CompilerOptions` class to avoid carrying all these multiple arguments everywhere.

devito/passes/clusters/buffering.py
mloubout: ok

tests/test_buffering.py
mloubout: Yeah that's fin,e I meant having `usave` as lhs mostly but again just nitpicking

devito/passes/iet/parpragma.py
```diff
@@ -213,9 +218,18 @@ def _make_reductions(self, partree):
         if not any(i.is_ParallelAtomic for i in partree.collapsed):
             return partree
 
+        def supported():
+            # Not all backend compilers support array reduction!
+            # Here are the known unsupported ones:
+            if isinstance(self.compiler, GNUCompiler) and \
```
mloubout, line 224: SHould we havd something similar for `osx clang + openmp`?

FabioLuporini: ah, yeah, probably..

tests/test_buffering.py
FabioLuporini: ah, I see what you mean now... 
well in backward propagators you do have usave's on the RHS :)

mloubout (approved):

devito/ir/support/basic.py
```diff
@@ -548,6 +550,15 @@ def is_cross(self):
     def is_local(self):
         return self.function.is_Symbol
 
+    @memoized_meth
+    def is_const(self, dim):
+        """
+        True if a constant depedence, that is no Dimensions involved, False otherwise.
```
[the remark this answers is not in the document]
FabioLuporini: fixed

tests/test_dle.py
```diff
@@ -731,9 +731,16 @@ def test_edge_cases(self, exprs, simd_level, expected):
         op = Operator(exprs, opt=('advanced', {'openmp': True}))
 
         iterations = FindNodes(Iteration).visit(op)
-        assert 'omp for collapse' in iterations[0].pragmas[0].value
-        if simd_level:
-            assert 'omp simd' in iterations[simd_level].pragmas[0].value
+        try:
+            assert 'omp for collapse' in iterations[0].pragmas[0].value
```
georgebisbas, line 735: Does that mean that you do not see "omp for collapse" in every case as before?
If this is the case I would prefer another variable in parametrization to whether "omp for collapse" is expected or not.
If we still see "omp for collapse" it should be out of the try-except.

Ideally we could use parameters to avoid the try-except. THese parameters would reflect the expected positions of omp and simd pragmas.

FabioLuporini: false, it is caught in the except

georgebisbas: Then why do we not have:
```
        assert 'omp for collapse' in iterations[0].pragmas[0].value
        try:
            if simd_level:
                assert 'omp simd' in iterations[simd_level].pragmas[0].value
        except:
            # E.g. gcc-5 doesn't support array reductions, so the compiler will
            # generate different legal code
            assert not Ompizer._support_array_reduction(configuration['compiler'])
```

The "omp for collapse is always in iterations[0], right?

georgebisbas: If it just false we can add a similar var to simd_level, like omp_level to get either 0 or None...
i.e.
```
if omp_level:
    assert 'omp for collapse' in iterations[0].pragmas[0].value
```

FabioLuporini: > The "omp for collapse is always in iterations[0], right?

no, with gcc-5 it will be in the innermost loop*, as it prefers thread parallelism over simd parallelism


* it's a bit more complicated than that... it actually depends on the number of physical and logical cores of the underlying platform, as collapsing may further change what the candidate-selector ends up choosing. Hence the loose(ish) `any(...)` in the except branch

georgebisbas: okk

georgebisbas (approved):
GTG from me once green!

mloubout (approved):

PR #1905, opened 2022-04-14
mpi: Always generate MPIComm with MPI enabled

### LINKED ISSUE #1728: Operators allocate memory blocks on the same GPU when using multiple GPUs with domain decomposition
When using multiple GPUs for wave propagation, certain operators like 'initdamp', when run first, do not appropriately initialize the MPI environment, so they allocate the different model blocks on the same GPU instead of distributing them across multiple GPUs.

This behavior can be produced with one of Devito provided examples:

DEVITO_PLATFORM=nvidiaX DEVITO_ARCH=pgcc DEVITO_LANGUAGE=openacc DEVITO_MPI=1 OMPI_CC=pgcc mpirun -np 2 python3 elastic_example.py

And nvidia-smi shows that two memory blocks are allocated on the same GPU, by the operator 'initdamp'.

snapshot

And a snippet of the generated code for the operator 'initdamp':

initdamp

FabioLuporini:
fixes #1728 

@nmbader

### DIFF
--- a/devito/passes/iet/mpi.py
+++ b/devito/passes/iet/mpi.py
@@ -299,8 +299,20 @@ def make_mpi(iet, **kwargs):
         mapper[hs] = heb.make(hs)
 
     efuncs = sync_heb.efuncs + user_heb.efuncs
-    objs = filter_sorted(sync_heb.objs + user_heb.objs)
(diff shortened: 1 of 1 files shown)


mloubout (approved):

georgebisbas (approved):
Unless no tests are required for this, all good

PR #1906, opened 2022-04-14
compiler: Fix checkpoint size when save != None

### LINKED ISSUE #1907: Size of checkpoint could be incorrect when save != None
When using Checkpoint and save != None (eg. save != time_order + 1) the value returned for size of a TimeFunction will be larger than necessary. 

Actually on `devito/checkpoint/checkpoint.py:72`: 

```python

     @property
     def size(self):
         """The memory consumption of the data contained in a checkpoint."""
         return sum([int((o.size_allocated/(o.time_order+1))*o.time_order)
         return sum([int((o.size_allocated/(o.time_size))*o.time_order)
                    for o in self.objects])
```

The size returned to each object must be `o.size_allocated/(o.time_size)` the assumption of `time_size = time_order + 1`  is only true with `save=None` or `save=time_order +1`

speglich:
When using Checkpoint and `save != None` (eg. `save != time_order + 1`) the value returned for size of a TimeFunction will be larger than necessary. 

Closes #1907

### DIFF
--- a/devito/checkpointing/checkpoint.py
+++ b/devito/checkpointing/checkpoint.py
@@ -72,7 +72,7 @@ class DevitoCheckpoint(Checkpoint):
     @property
     def size(self):
         """The memory consumption of the data contained in a checkpoint."""
-        return sum([int((o.size_allocated/(o.time_order+1))*o.time_order)
+        return sum([int((o.size_allocated/(o.time_size))*o.time_order)
                    for o in self.objects])
 
     def save(*args):
--- a/devito/types/dense.py
+++ b/devito/types/dense.py
@@ -1411,6 +1411,10 @@ class TimeFunction(Function):
         return self.shape_allocated[self._time_position]
 
     @property
+    def time_size(self):
+        return self._time_size
+
+    @property
     def _time_buffering(self):
         return not is_integer(self.save)
 


FabioLuporini:
Thanks for this!

Before we review, could you take a look at https://github.com/devitocodes/devito/wiki/Tags-for-commit-messages-and-PR-titles ? Thanks!

devito/checkpointing/checkpoint.py
```diff
@@ -72,7 +72,7 @@ def get_data_location(self, timestep):
     @property
     def size(self):
         """The memory consumption of the data contained in a checkpoint."""
-        return sum([int((o.size_allocated/(o.time_order+1))*o.time_order)
+        return sum([int((o.size_allocated/(o.time_size))*o.time_order)
```
FabioLuporini, line 75: why does this property need to become public? can/should you not just use `_time_size`? can it be useful on the user side ? (@navjotk )

navjotk: I believe this is the "get" version of the property that the user sets with `save=...`. I think it is useful for the user to be able to read the value if they are allowed to set it - especially since we accept defaults. 

Having said that, I don't feel too strongly about it so happy with it being public or private.

PR #1908, opened 2022-04-15
compiler: Revamp generation of implicit equations

FabioLuporini:
At last...

The reason I had to rework this now is that I have a compiler pass in Pro where, during clustering (!), I create a further equation that requires the MultiSubDomain thickness assignments, which wouldn't be generated due to premature lowering without this PR

### DIFF
--- a/devito/passes/clusters/implicit.py
+++ b/devito/passes/clusters/implicit.py
@@ -4,7 +4,7 @@ Passes to gather and form implicit equations from MultiSubDomains.
 
 from devito.ir import Cluster, Interval, IntervalGroup, IterationSpace, Queue
 from devito.symbolics import retrieve_dimensions
-from devito.tools import filter_ordered, timed_pass
+from devito.tools import filter_sorted, timed_pass
 from devito.types.grid import MultiSubDimension
 
 __all__ = ['generate_implicit']
@@ -59,13 +59,14 @@ class LowerMultiSubDimensions(Queue):
         if msdim(dd):
             return clusters
 
+        idx = len(prefix)
+
         seen = set()
         tip = None
 
         processed = []
         for c in clusters:
             try:
-                idx = len(prefix)
                 dd = c.ispace[idx].dim
                 d = msdim(dd)
             except IndexError:
@@ -82,7 +83,9 @@ class LowerMultiSubDimensions(Queue):
             exprs = d.msd._implicit_exprs
 
             # The implicit Dimensions and iterators induced by the MultiSubDomain
-            dims = filter_ordered(retrieve_dimensions(exprs, deep=True))
+            # NOTE: `filter_sorted` is for deterministic code generation, should
+            # there ever be a crazy MultiSubDomain with multiple implicit Dimensions
+            dims = filter_sorted(retrieve_dimensions(exprs, deep=True))
             idims = tuple(i for i in dims if not i.is_SubIterator)
             intervals = [Interval(i, 0, 0) for i in idims]
             sub_iterators = {i.root: i for i in dims if i.is_SubIterator}
--- a/devito/passes/clusters/implicit.py
+++ b/devito/passes/clusters/implicit.py
@@ -48,6 +48,9 @@ class LowerMultiSubDimensions(Queue):
         Cluster([Eq(f[t1, xi_n, yi_n], f[t0, xi_n, yi_n] + 1)])
     """
 
+    def _make_key_hook(self, cluster, level):
+        return (tuple(cluster.guards.get(i.dim) for i in cluster.itintervals[:level]),)
+
     def callback(self, clusters, prefix):
         try:
             dd = prefix[-1].dim
--- a/devito/passes/clusters/implicit.py
+++ b/devito/passes/clusters/implicit.py
@@ -3,7 +3,8 @@ Passes to gather and form implicit equations from MultiSubDomains.
 """
 
 from devito.ir import Interval, IntervalGroup, IterationSpace, Queue
-from devito.tools import timed_pass
+from devito.symbolics import retrieve_dimensions
+from devito.tools import filter_ordered, timed_pass
 from devito.types.grid import MultiSubDimension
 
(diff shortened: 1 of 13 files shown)


devito/types/grid.py
```diff
@@ -739,12 +742,15 @@ def __subdomain_finalize__(self, grid, **kwargs):
             # equivalent.
             self._local_bounds = self._global_bounds
 
-    def _create_implicit_exprs(self, grid):
+    @cached_property
+    def _implicit_exprs(self):
```
FabioLuporini, line 746: for another day: we could probably detach the generation logic from this class and move it inside the compiler pass directly, it'd be neater probably

devito/tools/threading.py
```diff
@@ -1,6 +1,26 @@
 import threading
 
-__all__ = ['sympy_mutex']
+__all__ = ['sympy_mutex', 'safe_dict_copy']
 
 
 sympy_mutex = threading.RLock()
+
+
+def safe_dict_copy(v):
```
mloubout, line 9: when/where is that an issue? Caching I see but why is it problematic now?

devito/types/grid.py
```diff
@@ -702,11 +708,12 @@ def __init__(self, **kwargs):
         self._global_bounds = kwargs.get('bounds', None)
 
     def __subdomain_finalize__(self, grid, **kwargs):
+        self._grid = grid
         self._dtype = grid.dtype
 
         # Create the SubDomainSet SubDimensions
         self._dimensions = tuple(
-            MultiSubDimension('%si_%s' % (d.name, self.implicit_dimension.name), d)
+            MultiSubDimension('%si_%s' % (d.name, self.implicit_dimension.name), d, self)
```
mloubout, line 716: That's some odd circular dependent thing here no?

mloubout (commented):
Some minor comments but GTG I think.

devito/tools/threading.py
FabioLuporini: PR was triggering this failure:
https://github.com/devitocodes/devito/runs/6078671948?check_suite_focus=true

Discussed a few alternatives with @tjb900 and @ggorman , eventually picked this one

devito/types/grid.py
```diff
@@ -542,20 +542,32 @@ class MultiSubDimension(SubDimension):
...
         lst, rst = self._symbolic_thickness(name)
         left = parent.symbolic_min + lst
         right = parent.symbolic_max - rst
 
         super().__init_finalize__(name, parent, left, right, ((lst, 0), (rst, 0)), False)
 
+    def __hash__(self):
+        # There is no possibility for two MultiSubDimensions to ever hash the same, since
+        # a MultiSubDimension carries a reference to a MultiSubDomain, which is unique
+        return id(self)
```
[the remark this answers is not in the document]
FabioLuporini: ?

devito/types/grid.py
mloubout: I agree was just wondering but don't have a way around neither looking at it. Maybe just add a TODO/Comment for later

devito/tools/threading.py
mloubout: Ok. Weird never triggered before.

devito/types/grid.py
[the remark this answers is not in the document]
FabioLuporini: oh, fair point. Here I'm just overriding Sympy's default behavior tho

devito/passes/clusters/implicit.py
```diff
@@ -0,0 +1,133 @@
...
+
+            idx = c.ispace.index(dd)
+            ispace0 = c.ispace[:idx]
+            ispace1 = c.ispace[idx:]
+
+            # The implicit expressions introduced by the MultiSubDomain
+            exprs = d.msd._implicit_exprs
+
+            # The implicit Dimensions and iterators induced by the MultiSubDomain
+            dims = filter_ordered(retrieve_dimensions(exprs, deep=True))
```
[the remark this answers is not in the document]
FabioLuporini: in fact, we need filter_sorted I suspect (will patch now). If we had one crazy MultiSubDomain with multiple implicit_dimensions, we would want a deterministic ordering of the iteration space (eg always `n, m` and not randomly `n, m` vs `m, n`)

devito/passes/clusters/implicit.py
```diff
@@ -0,0 +1,133 @@
...
+        if msdim(dd):
+            return clusters
+
+        seen = set()
+        tip = None
+
+        processed = []
+        for c in clusters:
+            try:
+                idx = len(prefix)
```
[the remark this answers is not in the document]
FabioLuporini: yes

devito/passes/clusters/implicit.py
```diff
@@ -0,0 +1,133 @@
...
+            else:
+                nxt = self._make_tip(c, ispaceN)
+                if tip is None or tip != nxt:
+                    processed.append(c.rebuild(exprs=exprs, ispace=ispace))
+                    tip = nxt
+                else:
+                    # Special case: we're reusing the previously constructed ispaceN
+                    pass
+
+            ispace = IterationSpace.union(c.ispace, ispaceN)
```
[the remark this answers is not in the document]
FabioLuporini: not really

The former ispace is for the thickness assignments
The latter ispace is for the reconstructed cluster

While the former may occasionally be skipped (optimization to avoid redundant assignments), the latter is compulsory

devito/types/grid.py
FabioLuporini: will do now

devito/tools/threading.py
```diff
@@ -1,6 +1,26 @@
...
+
+    Notes
+    -----
+    See https://bugs.python.org/issue40327
+    """
+    while True:
+        try:
+            return v.copy()
+        except RuntimeError:
+            pass
```
mloubout, line 26: nitpicking but a `sleep(small)` may be better just to give some delay

mloubout (approved):

PR #1914, opened 2022-04-29
examples: invoke tti example with --constant argument

ofmla:
I just realized that `python ~/devito/examples/seimic/tti/tti_example.py --constant` does not use the `constant-tti` model. This PR adds a very simple solution which will work for the `constant-tti `and `layers-tti` models


examples/seismic/tti/tti_example.py
```diff
@@ -37,12 +37,14 @@ def run(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
     # Define receiver geometry (spread across `x, y`` just below surface)
     rec, u, v, summary = solver.forward(save=save, autotune=autotune)
 
-    if preset == 'constant':
+    if preset == 'constant-tti':
         # With  a new m as Constant
         v0 = Constant(name="v", value=2.0, dtype=np.float32)
-        solver.forward(save=save, vp=v0)
+        solver.model.update('vp', v0.value)
```
FabioLuporini, line 43: this shouldn't be required strictly speaking?

examples/seismic/tti/tti_example.py
```diff
@@ -37,12 +37,14 @@ def run(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
...
+        solver.forward(save=save)
         # With a new vp as a scalar value
-        solver.forward(save=save, vp=2.0)
+        solver.model.update('vp', 2.0)
```
FabioLuporini, line 46: perhaps the update should be performed internally to `forward`? 

if you had multiple calls to `forward` and each time you had to update one or more parameters "by hand", it would be painful

examples/seismic/tti/tti_example.py
```diff
@@ -81,8 +83,15 @@ def test_tti_stability(shape, kernel):
...
-    preset = 'layers-tti-noazimuth' if args.azi else 'layers-tti'
+    # if user defines constant and noazimuth arguments at the same time raise exception
+    if args.constant and args.azi:
+        parser.error('argument --noazimuth: not allowed with argument --constant')
+
+    preset = 'layers-tti'
+    if args.constant:
+        preset = 'constant-tti'
+    elif args.azi:
+        preset = 'layers-tti-noazimuth'
```
FabioLuporini, line 94:
```
else:
    preset = 'layers-tii'
```

examples/seismic/model.py
```diff
@@ -375,9 +375,9 @@ def update(self, name, value):
             param = getattr(self, name)
         except AttributeError:
             # No physical parameter with tha name, create it
-            setattr(self, name, self._gen_phys_param(name, value, self.space_order))
+            setattr(self, name, self._gen_phys_param(value, name, self.space_order))
```
FabioLuporini, line 378: We need to put this under CI somehow

FabioLuporini (commented):
Thanks a lot! I just left a few comments

examples/seismic/tti/tti_example.py
mloubout: I prefer the runtime argument `vp=v0` to avoid keeping track of what was changed in the model

examples/seismic/tti/tti_example.py
mloubout: Should probably yes be internal so the `dt` directly computed from runtime vo

examples/seismic/tti/tti_example.py
mloubout: Don't particularly like these additional presets. There is also no reason constant would not be compatible with no azimuth

mloubout (commented):
I overall don't have an issue with this as it adds test/example cases but I think it should be done as a proper constant tti as currently it just slightly modifies the current VP constant only

examples/seismic/tti/tti_example.py
ofmla: Well, `v0` is a constant and passing it directly to `solver.forward` won't work. It gives the following error:
```
  File "devito/examples/seismic/tti/tti_example.py", line 103, in <module>
    run(shape=shape, spacing=spacing, nbl=args.nbl, tn=tn,
  File "devito/examples/seismic/tti/tti_example.py", line 48, in run
    solver.forward(vp=2., save=save)
  File "/home/oscar/devito/examples/seismic/tti/wavesolver.py", line 155, in forward
    summary = self.op_fwd(save).apply(src=src, rec=rec, u=u, v=v,
  File "/home/oscar/devito/devito/operator/operator.py", line 708, in apply
  File "/home/oscar/devito/devito/operator/operator.py", line 557, in arguments
  File "/home/oscar/devito/devito/operator/operator.py", line 449, in _prepare_arguments
    args.update(p._arg_values(**kwargs))
  File "/home/oscar/devito/devito/types/dense.py", line 843, in _arg_values
    for i, s in zip(self.dimensions, new.shape):
AttributeError: 'float' object has no attribute 'shape'
```


and for that reason I used `model.update`

examples/seismic/tti/tti_example.py
ofmla: So, the same applies here: passing an scalar to `solver.forward() `would lead to an error
```
  File "devito/examples/seismic/tti/tti_example.py", line 103, in <module>
    run(shape=shape, spacing=spacing, nbl=args.nbl, tn=tn,
  File "devito/examples/seismic/tti/tti_example.py", line 48, in run
    solver.forward(vp=2., save=save)
  File "/home/oscar/devito/examples/seismic/tti/wavesolver.py", line 155, in forward
    summary = self.op_fwd(save).apply(src=src, rec=rec, u=u, v=v,
  File "/home/oscar/devito/devito/operator/operator.py", line 708, in apply
  File "/home/oscar/devito/devito/operator/operator.py", line 557, in arguments
  File "/home/oscar/devito/devito/operator/operator.py", line 449, in _prepare_arguments
    args.update(p._arg_values(**kwargs))
  File "/home/oscar/devito/devito/types/dense.py", line 843, in _arg_values
    for i, s in zip(self.dimensions, new.shape):
AttributeError: 'float' object has no attribute 'shape'
```
If you planned for this to be possible, then it is likely that there is some error somewhere else. If this is not the case, you need to use `model.update `before

examples/seismic/tti/tti_example.py
ofmla: I wrote it in that way because there is not a` 'constant-noazimuth' `preset model. In any case you would have to play with conditionals I think

examples/seismic/model.py
ofmla: > I overall don't have an issue with this as it adds test/example cases but I think it should be done as a proper constant tti as currently it just slightly modifies the current VP constant only

I tried to fix a mistake (constant case was not running) without changing the test design. The test only update the `vp` parameter, so I left the things as they were.

ofmla:
> I overall don't have an issue with this as it adds test/example cases but I think it should be done as a proper constant tti as currently it just slightly modifies the current VP constant only

I tried to fix a mistake (constant case was not running) without changing the test design. The test only update the vp parameter, so I left the things as they were.

examples/seismic/tti/tti_example.py
mloubout: This looks like it ends up in the wrong spot (dense) instead of constant.

ofmla: So, are you saying you planned that the user could set a physical parameter i.e. velocity as a `Constant `or scalar  and pass it to the forward method of the `solver`, and it should work smoothly?

examples/seismic/tti/tti_example.py
```diff
@@ -81,8 +83,15 @@ def test_tti_stability(shape, kernel):
                         help="Choice of finite-difference kernel")
     args = parser.parse_args()
 
-    # Switch to TTI kernel if input is acoustic kernel
-    preset = 'layers-tti-noazimuth' if args.azi else 'layers-tti'
+    # if user defines constant and noazimuth arguments at the same time raise exception
+    if args.constant and args.azi:
+        parser.error('argument --noazimuth: not allowed with argument --constant')
```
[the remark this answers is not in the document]
ofmla: That is what the documentation says:

ArgumentParser.error(message)[¶](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.error)
This method prints a usage message including the message to the standard error and terminates the program with a status code of 2.

ofmla:
> What's still missing:
> 
> * CI test for the constant case
> * (maybe) moving the "model.update" inside `solver.forward`

Well, the constant case is already tested in  `tests/test_tti.py`, I mean the forward modeling for TTI case is executed using a constant model. It seems to me that  there is not any wave-equation_example.py (i.e. `tti_example.py`, `acoustic_example.py`) in CI and that's OK. The point of this PR is that the `tti_example.py` script uses a layered model by default and although the user passes the argument `--constant` the exemple won't run on a constant model. The PR changes this.

On the other hand the `solver.forward`  gets the physical parameters from `model`  or `kwargs`. If I pass `vp` as argument to `solver.forward`  it is passed to `self.op_fwd(save).apply(...,**kwargs)` even updating `model` via `model.update` within `solver.forward`. As result I get the same errors I mentioned in the coments above. That is why I use `model.update` before. If `vp` is a numpy array there is no problem passing it to `solver.forward` but if it is a scalar or `Constant` breakage will occur.

examples/seismic/tti/tti_example.py
```diff
@@ -81,8 +83,15 @@ def test_tti_stability(shape, kernel):
                         help="Choice of finite-difference kernel")
     args = parser.parse_args()
 
-    # Switch to TTI kernel if input is acoustic kernel
-    preset = 'layers-tti-noazimuth' if args.azi else 'layers-tti'
+    # if user defines constant and noazimuth arguments at the same time raise exception
```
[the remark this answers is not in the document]
ofmla: Is it worth spending a commit here? 😅

examples/seismic/model.py
ofmla: All that comes into my mind is 
``` 
from examples.seismic import demo_model
import numpy as np


def test_model_updt(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0),
                    space_order=4, nbl=10):

    # Layered model (tti)
    tti_model = demo_model('layers-tti', shape=shape, spacing=spacing,
                           space_order=space_order, nbl=nbl)
    tpl1_set = set(tti_model.physical_parameters)

    # Layered model (isotropic)
    iso_model = demo_model('layers-isotropic', shape=shape, spacing=spacing,
                           space_order=space_order, nbl=nbl)
    tpl2_set = set(iso_model.physical_parameters)

    # Physical parameters in either set but not in the intersection.
    diff_phys_par = tuple(tpl1_set ^ tpl2_set)

    # Convert iso model in tti model
    for i in diff_phys_par:
        iso_model.update(i, getattr(tti_model, i).data[nbl:-nbl, nbl:-nbl, nbl:-nbl])
        assert np.array_equal(getattr(iso_model, i).data, getattr(tti_model, i).data)
```
I can add this script to the `tests` folder if you agree

examples/seismic/tti/tti_example.py
```diff
@@ -81,8 +74,15 @@ def test_tti_stability(shape, kernel):
                         help="Choice of finite-difference kernel")
     args = parser.parse_args()
 
-    # Switch to TTI kernel if input is acoustic kernel
-    preset = 'layers-tti-noazimuth' if args.azi else 'layers-tti'
+    preset = 'layers-tti'
```
FabioLuporini, line 77: I would maybe rather move this inside the nested `else`

```
if ...
    ...
else:
    if args.azi:
        ....
    else:
        preset = 'layers-tti'
```

tests/test_model_update.py
```diff
@@ -0,0 +1,26 @@
+from examples.seismic import demo_model
```
FabioLuporini, line 1: We shouldn't test `examples` inside `tests` because in theory the examples folder shouldn't (couldn't) even be accessible.

So I've been thinking about this...

How about we add it somewhere here? https://github.com/devitocodes/devito/blob/master/examples/seismic/tti/tti_example.py#L72

It should then be captured by CI through this line? 
https://github.com/devitocodes/devito/blob/master/.github/workflows/examples.yml#L49 
(@mloubout )

examples/seismic/tti/tti_example.py
```diff
@@ -37,13 +37,6 @@ def run(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
     # Define receiver geometry (spread across `x, y`` just below surface)
     rec, u, v, summary = solver.forward(save=save, autotune=autotune)
 
-    if preset == 'constant':
```
FabioLuporini, line 40: sorry I'm confused now... why is this being entirely dropped?

examples/seismic/tti/tti_example.py
ofmla: Ok

tests/test_model_update.py
ofmla: I moved the model update test to tti_exemple.py as suggested

examples/seismic/tti/tti_example.py
ofmla: As I see it, those lines are not longer necessary

FabioLuporini (approved):
LGTM. Thanks!

examples/seismic/tti/tti_example.py
```diff
@@ -71,6 +64,29 @@ def test_tti_stability(shape, kernel):
     assert np.isfinite(norm(rec))
 
 
+@pytest.mark.parametrize('shape', [(51, 51), (16, 16, 16)])
+def test_model_updt(shape):
```
mloubout, line 68: Some nitpicking:
- Can this be moved in `model.py` 
- Can we test all cases, i.e 1) replacing the value of existing param by an array, 2) replacing existing param with a new Function 3) new param array, 4) new param Function

examples/seismic/tti/tti_example.py
```diff
@@ -81,8 +97,16 @@ def test_tti_stability(shape, kernel):
                         help="Choice of finite-difference kernel")
     args = parser.parse_args()
 
-    # Switch to TTI kernel if input is acoustic kernel
-    preset = 'layers-tti-noazimuth' if args.azi else 'layers-tti'
+    if args.constant:
```
mloubout, line 100: Can probably make it a dict, but fine like that too.

mloubout (commented):
Thanks for the patience and updates. Left a minor comment on the test.

I still have one main point of discussion. So the constant TTI case is not `Constant` right, it's `Function` with constant values. 
The main idea behind the constant isotropic was to have the `Constant` case to see the perf difference when there is no variation. I don't think there is much interest in a constant model if using Function personally (@FabioLuporini feel free to shit me down) since it's just a normal one with different values. So I'd rather make this an actual true constant model or remove it if there is no need for it. 

Would still keep the update to Model and test OC.

ofmla:
> Thanks for the patience and updates. Left a minor comment on the test.
> 
> I still have one main point of discussion. So the constant TTI case is not `Constant` right, it's `Function` with constant values. The main idea behind the constant isotropic was to have the `Constant` case to see the perf difference when there is no variation. I don't think there is much interest in a constant model if using Function personally (@FabioLuporini feel free to shit me down) since it's just a normal one with different values. So I'd rather make this an actual true constant model or remove it if there is no need for it.
> 
> Would still keep the update to Model and test OC.

You are right. I reversed the changes made and kept the model updating, thx

mloubout:
Looks like this broke some notebooks and flake8 not passing.

ofmla:
> Looks like this broke some notebooks and flake8 not passing.

I made some changes. I hope that it will work.

examples/seismic/acoustic/acoustic_example.py
```diff
@@ -39,7 +39,7 @@ def run(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
     # Define receiver geometry (spread across x, just below surface)
     rec, u, summary = solver.forward(save=save, autotune=autotune)
 
-    if preset == 'constant':
+    if preset == 'constant-isotropic':
         # With  a new m as Constant
```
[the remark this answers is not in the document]
FabioLuporini: I will merge it as is because it's been around for too long, but we should check if there's a pep8 rule for flake8 to intercept it

examples/seismic/tti/tti_example.py
```diff
@@ -37,12 +37,23 @@ def run(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
     # Define receiver geometry (spread across `x, y`` just below surface)
     rec, u, v, summary = solver.forward(save=save, autotune=autotune)
 
-    if preset == 'constant':
-        # With  a new m as Constant
-        v0 = Constant(name="v", value=2.0, dtype=np.float32)
-        solver.forward(save=save, vp=v0)
-        # With a new vp as a scalar value
-        solver.forward(save=save, vp=2.0)
+    if preset in ['constant-tti', 'constant-tti-noazimuth']:
```
FabioLuporini, line 40: @ofmla wait... so this way it will run twice if preset=constant-tti for example ? right before this `if` and then inside?

examples/seismic/tti/tti_example.py
```diff
@@ -81,8 +92,16 @@ def test_tti_stability(shape, kernel):
                         help="Choice of finite-difference kernel")
     args = parser.parse_args()
 
-    # Switch to TTI kernel if input is acoustic kernel
-    preset = 'layers-tti-noazimuth' if args.azi else 'layers-tti'
+    if args.constant:
```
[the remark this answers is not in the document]
ofmla: Well, it is never easy to satisfy all parties. I had written without `else` but one of you wanted an if-else. That is a matter of taste.

examples/seismic/acoustic/acoustic_example.py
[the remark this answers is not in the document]
ofmla: I don't see any problem here

examples/seismic/tti/tti_example.py
```diff
@@ -37,12 +37,23 @@ def run(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
...
+        phi = None
+        if len(shape) > 2 and preset not in ['constant-tti-noazimuth']:
+            phi = .4
+        solver.forward(save=save, vp=vp, epsilon=epsilon, delta=delta,
+                       theta=theta, phi=phi)
+        # With new physical parameters as Constants
+        d = {'vp': vp, 'epsilon': epsilon, 'delta': delta, 'theta': theta, 'phi': phi}
+        for k, v in d.items():
+            v = Constant(name=k, value=v, dtype=np.float32)
+        solver.forward(save=save, vp=vp, epsilon=epsilon, delta=delta,
```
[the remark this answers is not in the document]
mloubout: It should be something like

```
k, v in d.items():
            d[k] = Constant(name=k, value=v, dtype=np.float32)
solver.forward(save=save, **d)
```

examples/seismic/acoustic/acoustic_example.py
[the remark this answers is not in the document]
mloubout: He means there is an extra space between `With` and `a`

examples/seismic/tutorials/15_tti_qp_pure.ipynb
```diff
@@ -85,8 +85,8 @@
     "origin  = (0.,0.)  \n",
     "nbl = 0  # number of pad points\n",
     "\n",
-    "model = demo_model('constant-tti', spacing=spacing, space_order=8,\n",
-    "                   shape=shape, nbl=nbl)\n",
+    "model = demo_model('layers-tti', spacing=spacing, space_order=8,\n",
```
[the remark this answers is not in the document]
ofmla: I got an error in cell 6 and I just didn't want to spend much time on it. The error is the following
```
  File "tti_qpure.py", line 77, in <module>
    optime=Operator([update_p] + src_term + rec_term)
  File "/home/oscar/devito/devito/operator/operator.py", line 159, in __new__
    op = cls._build(expressions, **kwargs)
  File "/home/oscar/devito/devito/operator/operator.py", line 183, in _build
    expressions = cls._lower_exprs(expressions, **kwargs)
  File "/home/oscar/devito/devito/tools/timing.py", line 76, in __call__
    retval = self.func(*args, **kwargs)
  File "/home/oscar/devito/devito/operator/operator.py", line 283, in _lower_exprs
    expressions = flatten([i.evaluate for i in expressions])
  File "/home/oscar/devito/devito/operator/operator.py", line 283, in <listcomp>
    expressions = flatten([i.evaluate for i in expressions])
  File "/home/oscar/anaconda3/envs/devito/lib/python3.8/site-packages/cached_property.py", line 36, in __get__
  File "/home/oscar/devito/devito/types/equation.py", line 86, in evaluate
    lhs, rhs = self.lhs.evaluate, self.rhs._eval_at(self.lhs).evaluate
  File "/home/oscar/devito/devito/types/lazy.py", line 47, in evaluate
    args = self._evaluate_args()
  File "/home/oscar/devito/devito/types/lazy.py", line 42, in _evaluate_args
    return [Evaluable._evaluate_maybe_nested(i) for i in self.args]
  File "/home/oscar/devito/devito/types/lazy.py", line 42, in <listcomp>
    return [Evaluable._evaluate_maybe_nested(i) for i in self.args]
  File "/home/oscar/devito/devito/types/lazy.py", line 16, in _evaluate_maybe_nested
    return maybe_evaluable.evaluate
  File "/home/oscar/devito/devito/finite_differences/derivative.py", line 322, in evaluate
    return self._eval_fd(self.expr)
  File "/home/oscar/devito/devito/finite_differences/derivative.py", line 356, in _eval_fd
    res = generic_derivative(expr, *self.dims, self.fd_order, self.deriv_order,
  File "/home/oscar/devito/devito/finite_differences/tools.py", line 50, in wrapper
    return S.Zero if expr.is_Number else func(expr, *args, **kwargs)
  File "/home/oscar/devito/devito/finite_differences/tools.py", line 67, in wrapper
    return func(expr, *args, **kwargs)
  File "/home/oscar/devito/devito/finite_differences/finite_difference.py", line 195, in generic_derivative
    return make_derivative(expr, dim, fd_order, deriv_order, side, matvec, x0, symbolic)
  File "/home/oscar/devito/devito/finite_differences/finite_difference.py", line 208, in make_derivative
    return indices_weights_to_fd(expr, dim, indices, weights, matvec=matvec.val)
  File "/home/oscar/devito/devito/finite_differences/finite_difference.py", line 223, in indices_weights_to_fd
    iloc -= expr.indices_ref[dim] - dim
  File "/home/oscar/devito/devito/tools/data_structures.py", line 48, in __getitem__
    return self.__getitem_hook__(key)
  File "/home/oscar/devito/devito/types/utils.py", line 27, in __getitem_hook__
    raise KeyError
KeyError

``` 
I can open an issue with a MFE after you merge the PR, it is the easier way. What do you think

examples/seismic/tti/tti_example.py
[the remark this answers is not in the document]
ofmla: ok

examples/seismic/tutorials/15_tti_qp_pure.ipynb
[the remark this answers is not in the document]
mloubout: Ok that's fine.

mloubout (approved):

mloubout:
Needs rebase and GTG

ofmla:
> Needs flake8, rebase and GTG

The flake8 errors are in files that I didn't touch. It is weird

mloubout:
>The flake8 errors are in files that I didn't touch. It is weird

Must be the new standards, is fixed in #1979 so I'd say wait for it to be merged then rebase and should be all good

examples/seismic/tti/tti_example.py
```diff
@@ -37,12 +37,22 @@ def run(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=250.0,
...
+        theta = .75
+        phi = None
+        if len(shape) > 2 and preset not in ['constant-tti-noazimuth']:
+            phi = .4
+        solver.forward(save=save, vp=vp, epsilon=epsilon, delta=delta,
+                       theta=theta, phi=phi)
+        # With new physical parameters as Constants
+        d = {'vp': vp, 'epsilon': epsilon, 'delta': delta, 'theta': theta, 'phi': phi}
+        for k, v in d.items():
+            d[k] = Constant(name=k, value=v, dtype=np.float32)
```
FabioLuporini, line 54: shuold be a dict comprehension

PR #1918, opened 2022-05-04
gpu: Update to HPC 22.3, reduced image size

kenhester:


docker/Dockerfile.nvidia
```diff
@@ -16,41 +16,52 @@
...
+#   docker run --gpus all --rm -it devito:nvidia python examples/seismic/acoustic/acoustic_example.py
 #
 #   docker run --gpus all --rm -it -p 8888:8888 -p 8787:8787 -p 8786:8786 --device=/dev/infiniband/uverbs0 --device=/dev/infiniband/rdma_cm  devito:nvidia 
 #
 # to run in user context on a cluster with shared filesystem, you can add the correct user config as docker options e.g.:
 #   docker run --gpus all --rm -it -v `pwd`:`pwd` -w `pwd` -u $(id -u):$(id -g) devito:nvidia python examples/seismic/acoustic/acoustic_example.py
 #
 ##############################################################
-FROM python:3.8
+#FROM python:3.9
```
georgebisbas, line 36: Are the comments to be removed?

georgebisbas (commented):
Thanks!

docker/Dockerfile.nvidia
kenhester: I was just try to illustrate that there was an alternate non-slim version of the base container. `#FROM python:3.9 ` could be removed, yes.

georgebisbas: I think, the slimmer, the better right?

docker/Dockerfile.nvidia
```diff
@@ -16,41 +16,52 @@
...
 RUN echo 'deb [trusted=yes] https://developer.download.nvidia.com/hpc-sdk/ubuntu/amd64 /' > /etc/apt/sources.list.d/nvhpc.list && \
     apt-get update -y && \
     apt-get install -y -q \
-        apt-utils curl\   
-        vim && \
+        apt-utils curl wget \   
     curl -sL https://deb.nodesource.com/setup_12.x | bash - && \
     apt-get install -y -q \
-        nodejs \
+#        nvhpc-22-3-cuda-multi \
```
georgebisbas, line 48: same I guess

docker/Dockerfile.nvidia
kenhester: That is the ask from some of our the users of the container.

mloubout:
Could we make this automatic somehow with a cron workflow to avoid the effort and PRs. Like just check once a week if there is a new sdk and update the image

FabioLuporini:
> Could we make this automatic somehow with a cron workflow to avoid the effort and PRs. Like just check once a week if there is a new sdk and update the image

That's a lovely idea -- tho that's probably on us ;)

docker/Dockerfile.nvidia
mloubout: Two previous cuda version and everythng that goes with it. Takes a lot of space usually yeah and not needed.

FabioLuporini:
@mloubout @georgebisbas can you make any sense of the conflict with the existing Dockerfile.nvidia ?

georgebisbas:
> @mloubout @georgebisbas can you make any sense of the conflict with the existing Dockerfile.nvidia ?

Can I see it somehow?

georgebisbas:
> @mloubout @georgebisbas can you make any sense of the conflict with the existing Dockerfile.nvidia ?

I wonder whether this conflict comes from @kenhester 's fork?
@kenhester could you kindly double-check?

mloubout:
Conflict is from the changes to dockerfile that were made in between yes since the COPY has slightly changed.

mloubout:
Should be GTG

FabioLuporini:
Thanks a lot @kenhester ! Merging, sorry it took a while, we have had a long list of PRs in the queue

PR #1919, opened 2022-05-04
operator: made the unrecognized argument error clearer

AtilaSaraiva:
Hey folks,

I was working on something on devito, and got stuck up to an hour trying to figure out that this error:
```
ValueError: Unrecognized argument damp=damp(x, y).
```
meant the operator expressions didn't have the function damp. Thinking back it is obvious but the error message could have been cleaerer. I'm just making this commit as way to prevent another person to go through this. If I encounter other situations such as this I will open another PR so that we can make the error messages and error catching mechanisms work better.

I'm not sure if that is the final solution to the problem, but I hope this can at least start a discussion.

Cheers!

### DIFF
--- a/devito/operator/operator.py
+++ b/devito/operator/operator.py
@@ -439,7 +439,7 @@ class Operator(Callable):
         if not configuration['ignore-unknowns']:
             for k, v in kwargs.items():
                 if k not in self._known_arguments:
-                    raise ValueError("Unrecognized argument %s=%s" % (k, v))
+                    raise ValueError("Unrecognized argument %s=%s. It most likely is missing from the expressions passed to the operator." % (k, v))
 
         overrides, defaults = split(self.input, lambda p: p.name in kwargs)
 


FabioLuporini:
Thanks!

I'm not entirely convinced that's the most likely case one could bump into such an exception? Not sure. @mloubout thoughts ?

FabioLuporini:
for example, it could also be a name mismatch

```
u0 = TimeFunction(name='u', ...)
u1 = TimeFunction(name='u', ...)

eq = Eq(u.forward, u + 1)

op = Operator(eq)

op(u0=u1)
```

I've seen this happening too

mloubout:
I get the message can be a little bit cryptic now but not sure this update covers all the case or make it more clear. It may be a bit messy to try to cover all the cases as it would need to infer what the user is trying to pass to the Operator.

AtilaSaraiva:
> I get the message can be a little bit cryptic now but not sure this update covers all the case or make it more clear. It may be a bit messy to try to cover all the cases as it would need to infer what the user is trying to pass to the Operator.

I agree, it doesn't. I'm not so sure if this is right message, but maybe just improving the message a little could help. Maybe we can start an issue and close this PR until the discussion takes shape.

FabioLuporini:
I agree an issue with a list of all possible situations ending up there would be useful. After that, we can think of refining the error handling. Basically we need a clear picture of all possible scenarios

AtilaSaraiva:
I posted the issue #1928 as suggested. I'm going to close this PR so that we can focus on the issue.

PR #1924, opened 2022-05-23
api: Add support for subdomain allocation for Functions

AtilaSaraiva:
Hello folks,

This is the first commit for a series of commits adding support for functions to be allocated within a subdomain (let's call those subfunctions for brevity), and operated along other normal functions . This only first commit only adds allocation support for the Function class, I still need to add it for the TimeFunction class, not to mention that I still have to add support for it for the Operator class.
This is only but the first step in the right direction, or so I hope.

Cheers!

Authored by Átila Saraiva Quintela Soares (atilasaraiva@gmail.com)
SENAI CIMATEC geophysicist researcher.

### DIFF
--- a/devito/types/grid.py
+++ b/devito/types/grid.py
@@ -136,6 +136,28 @@ class CartesianDiscretization(ABC):
             setattr(self, k, v)
         self._distributor = Distributor(self.shape, self.dimensions, MPI.COMM_SELF)
 
+    @property
+    def time_dim(self):
+        """Time dimension associated with this Grid."""
+        return self._time_dim
+
+    @property
+    def stepping_dim(self):
+        """Stepping dimension associated with this Grid."""
+        return self._stepping_dim
+
+    def time_dimension_handle (self, time_dimension, dtype):
+        # Store or create default symbols for time and stepping dimensions
+        if time_dimension is None:
+            spacing = Scalar(name='dt', dtype=dtype, is_const=True)
+            self._time_dim = TimeDimension(name='time', spacing=spacing)
+            self._stepping_dim = SteppingDimension(name='t', parent=self.time_dim)
+        elif isinstance(time_dimension, TimeDimension):
+            self._time_dim = time_dimension
+            self._stepping_dim = SteppingDimension(name='%s_s' % self.time_dim.name,
+                                                   parent=self.time_dim)
+        else:
+            raise ValueError("`time_dimension` must be None or of type TimeDimension")
 
 class Grid(CartesianDiscretization, ArgProvider):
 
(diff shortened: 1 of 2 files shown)


devito/types/grid.py
```diff
@@ -380,7 +383,7 @@ def __init__(self, *args, **kwargs):
             self.name = self.__class__.__name__
 
         # All other attributes get initialized upon `__subdomain_finalize__`
-        super().__init__()
+        # super().__init__()
```
FabioLuporini, line 386: leftover

FabioLuporini (commented):
Thanks a lot for this! I've added it to my backlog of "PRs to be reviewed". 

At first glance, it looks good to me. Basically, you're lifting some functionality because you need the distributor machinery to be available in the SubDomains (or MultiSubDomains), right?

FabioLuporini:
so yeah I think this is going in the right direction, my other suggestion is to keep PRs as small as possible so that we also merge them as quickly as possible :)

AtilaSaraiva:
> so yeah I think this is going in the right direction, my other suggestion is to keep PRs as small as possible so that we also merge them as quickly as possible :)

Oh, do you mean in number of lines written, the number of commits, or number of modified files?

> Thanks a lot for this! I've added it to my backlog of "PRs to be reviewed".

No problem, thanks for the early review, but I still have to fix it for TimeFunction as well.

> At first glance, it looks good to me. Basically, you're lifting some functionality because you need the distributor machinery to be available in the SubDomains (or MultiSubDomains), right?

Yes that is correct, no big changes, just moving things around. I wonder if it would be more appropriate to split the CartesianDiscretization in to classes, one for the distributor machinery and other for the rest. Lemme know what you think.

FabioLuporini:
> I wonder if it would be more appropriate to split the CartesianDiscretization in two classes, one for the distributor machinery and other for the rest. Lemme know what you think.

Mhh, what do you have in mind? Something like a top abstract class (CartesianDiscretization) and a mixin for "all the CartesianDiscretization that can 1) be used to define a (Time)Function and 2) be decomposed (with MPI) ?

devito/types/grid.py
AtilaSaraiva: Sorry, fixed!

AtilaSaraiva:
Hello Fabio! I'm studying the code for the time being, but you did say you would backport some one dimensional array functionality so I kinda need that to progress further.

Either way, I gonna need some more in depth understanding code generation and allocation to change the iteration space so Im digging the code for now.

FabioLuporini:
> Hello Fabio! I'm studying the code for the time being, but you did say you would backport some one dimensional array functionality so I kinda need that to progress further.

ah, yes sure, in my backlog... it will take a while to get there, but we will!

> Either way, I gonna need some more in depth understanding code generation and allocation to change the iteration space so Im digging the code for now.

Sure, that makes a lot of sense to me. Get in touch on slack if you have any questions (feel free to use the #development channel)

FabioLuporini:
what's the state of this?

FabioLuporini:
no followup, closing

PR #1926, opened 2022-05-23
Boundary reconstruction

fffarias:
Hello everybody. 

This PR tries to suggest reconstructing the source wave field from the one saved in the border. This should be an option for the gradient computation, in which the forward field is reconstructed inside the backwards loop. It was necessary to put a boolean argument inside the SparseTimeFunction's inject method, since I needed the injected field not to be incremented. Given that I changed an important part of the code, I added a test to validate the option of not incrementing. As an example, I also add a demonstration tutorial and application of the method in FWI.

Thanks a lot


devito/operations/interpolators.py
```diff
@@ -285,13 +288,18 @@ def callback():
...
             # Substitute coordinate base symbols into the interpolation coefficients
-            eqns = [Inc(field.xreplace(vsub), _expr.xreplace(vsub) * b,
-                        implicit_dims=self.sfunction.dimensions)
-                    for b, vsub in zip(self._interpolation_coeffs, idx_subs)]
+            if increment:            
+                eqns = [Inc(field.xreplace(vsub), _expr.xreplace(vsub) * b,
+                            implicit_dims=self.sfunction.dimensions)
+                        for b, vsub in zip(self._interpolation_coeffs, idx_subs)]
+            else: 
+                eqns = [Eq(field.xreplace(idx_subs[0]), _expr.xreplace(idx_subs[0]),
```
FabioLuporini, line 296: why are you picking `idx_subs[0]`? should you not traverse the whole list of `idx_subs` like in the `increment` case?

Could you do something like:

```
if increment:
   cls: Inc
else:
   cls: Eq
eqns = [cls(field.xreplace(....... .....) for b, vsub in zip(....)]
```
?

tests/test_interpolation.py
```diff
@@ -417,6 +429,31 @@ def test_inject_time_shift(shape, coords, result, npoints=19):
     assert np.allclose(a.data[indices], result, rtol=1.e-5)
 
 
+@pytest.mark.parametrize('shape, result, increment',  [
+    ((10, 10), 1., False),
+    ((10, 10), 5., True),
+    ((10, 10, 10), 1., False),
+    ((10, 10, 10), 5., True)        
+])
+def test_inject_time_increment(shape, result, increment):
```
FabioLuporini, line 438: Thank you for attaching tests to this PR!

FabioLuporini:
Thank you, Fernanda! We will review this PR carefully. Looks nice!

Meanwhile, note that some of the commit messages would have to be edited, via rebasing, to use valid tags: https://github.com/devitocodes/devito/wiki/Tags-for-commit-messages-and-PR-titles

devito/operations/interpolators.py
fffarias: because I thought in this case very specifically to apply boundary conditions, not source injection, for example. The ` ```increment=False``` option as I suggested should not be used for off-grid points.

mloubout: While this is true for your case, the generic case is still a standard potentially off the grid point injection without increment. So there is two option
- Use the injection as is knowing the coefficient will still be zero for the interpolation if the point is on the grid
- Define somehow an "on-the-grid" sparse function that ignores the interpolation.

The first case is definitely easier and should still work for your case and would boil down to what @FabioLuporini wrote earlier

examples/seismic/tutorials/16_source_wavefield_reconstruction.ipynb
```diff
@@ -0,0 +1,1069 @@
+{
```
mloubout, line 1: Nice tutorial. Some generic comments.

I think it would be beneficial to split and/or extend this tutorial for readability. I would 
- first focus purely on the wavefield itself and show snapshots/movie/... of the forward wavefield and its reconstruction (i.e check the snapshot tutorial for ploting)
- Second detail the gradient computation

And on the into side, It would be beneficial to use slightly more explicit variable names, such as `u` and `u_boundary` or such. I would also completely skip the illumination part. While this is indeed very useful in practice, it adds to the complexity of the tutorial.

tests/test_interpolation.py
```diff
@@ -53,6 +53,18 @@ def time_points(grid, ranges, npoints, name='points', nt=10):
...
 
 
+def time_grid_points(grid, name='points', nt=10):
+    """Create a SparseTimeFunction field with coordinates 
+       filled in by all grid points"""
+    npoints = prod(grid.shape) 
+    a = SparseTimeFunction(name=name, grid=grid, npoint=npoints, nt=nt)
+    dims = tuple([np.linspace(0., 1., d) for d in grid.shape]) 
+    for i in range(len(grid.shape)):
+        a.coordinates.data[:,i] = np.meshgrid(*dims)[i].flatten()
```
mloubout, line 63: can probably compute meshgrid only once before the for loop

mloubout (commented):
Thanks so much for the contribution @fffarias this is very much appreciated. I left a couple comments for the tutorial to make it a bit more pedagogic, happy to discuss what's doable without adding too much on your plate.

devito/operations/interpolators.py
mloubout: Indeed good point I missed that gotta think about it

examples/seismic/tutorials/16_source_wavefield_reconstruction.ipynb
EdCaunt, line 1: <p>Might be good to have this presented as percentage of max wavefield value?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/devitocodes/devito/pull/1926/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

mloubout:
Generic question. Would it be easier to use 1D `Function` instead of Sparse functions for this?

devito/operations/interpolators.py
mloubout: Getting back here since been looking into sparse, I think that the "best" way to go for `increment=Flase` is to do what `interpolation` does i.e 
- Define `sum` and do the standard `sum = interp(expr)`
- assign sum afterward `field = sum`

This would cover the off-the-grid case as well

PR #1929, opened 2022-05-31
compiler: Add blockrelax tests and refresh advisor profiling

georgebisbas:
- Avoid autotuning when children's block shapes are the same as the parent's block shapes
- Slightly refresh advisor

### DIFF
--- a/benchmarks/user/advisor/run_advisor.py
+++ b/benchmarks/user/advisor/run_advisor.py
@@ -31,7 +31,7 @@ from benchmarks.user.advisor.advisor_logging import (check, log, progress,
 def run_with_advisor(path, output, name, exec_args):
     path = Path(path)
     check(path.is_file(), '%s not found' % path)
-    check(path.suffix == '.py', '%s not a regular Python file' % path)
+    check(path.suffix == '.py', '%s not a Python file' % path)
 
     # Create a directory to store the profiling report
     if name is None:
@@ -49,15 +49,20 @@ def run_with_advisor(path, output, name, exec_args):
         output = Path(output).joinpath(name)
         output.mkdir(parents=True, exist_ok=True)
 
-    # Intel Advisor must be available through either Intel Parallel Studio
-    # or Intel oneAPI (currently tested versions include IPS 2020 Update 2 and
+    # Intel Advisor and Intel compilers must be available through either Intel Parallel
+    # Studio or Intel oneAPI (currently tested versions include IPS 2020 Update 2 and
     # oneAPI 2021 beta08)
     try:
         ret = check_output(['advixe-cl', '--version']).decode("utf-8")
     except FileNotFoundError:
         check(False, "Error: Couldn't detect `advixe-cl` to run Intel Advisor.")
 
-    # If Advisor is available, so is the Intel compiler
+    try:
+        ret = check_output(['icc', '--version']).decode("utf-8")
+    except FileNotFoundError:
+        check(False, "Error: Couldn't detect Intel Compiler (icc).")
+
+    # All good, Intel compiler and advisor are available
     os.environ['DEVITO_ARCH'] = 'intel'
 
     # Tell Devito to instrument the generated code for Advisor
@@ -68,12 +73,13 @@ def run_with_advisor(path, output, name, exec_args):
     if devito_logging is None:
         os.environ['DEVITO_LOGGING'] = 'WARNING'
 
-    with progress('Set up multi-threading environment'):
-        # Roofline analyses only make sense with threading enabled
+    with progress('Setting up multi-threading environment'):
+        # Roofline analyses are recommended with threading enabled
         os.environ['DEVITO_LANGUAGE'] = 'openmp'
 
-        # We must be able to do thread pinning, otherwise any results would be
-        # meaningless. Currently, we only support doing that via numactl
+        # Thread pinning is strongly recommended for reliable results.
+        # This script is using numactl for this purpose. Users may want to set their
+        # own pinning: https://hpc-wiki.info/hpc/Binding/Pinning
         try:
             ret = check_output(['numactl', '--show']).decode("utf-8")
             ret = dict(i.split(':') for i in ret.split('\n') if i)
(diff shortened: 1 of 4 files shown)


benchmarks/user/advisor/run_advisor.py
```diff
@@ -68,12 +73,12 @@ def run_with_advisor(path, output, name, exec_args):
...
-    with progress('Set up multi-threading environment'):
-        # Roofline analyses only make sense with threading enabled
+    with progress('Setting up multi-threading environment'):
+        # Roofline analyses are recommended with threading enabled
         os.environ['DEVITO_LANGUAGE'] = 'openmp'
 
-        # We must be able to do thread pinning, otherwise any results would be
-        # meaningless. Currently, we only support doing that via numactl
+        # Thread pinning is strongly recommended for reliable results.
+        # We support thread pinning via numactl
```
mloubout, line 81: or `KMP_AFFINITY`?

FabioLuporini:
please review: https://github.com/devitocodes/devito/wiki/Tags-for-commit-messages-and-PR-titles

benchmarks/user/advisor/run_advisor.py
FabioLuporini: "we support" doesn't also make much sense. We don't support anything, thread/process placement is up to the user (though we might one day devise an option to do that automatically at least for openmp since there are special for-loop clauses that allow you to specify the thread affinity for that loop. But not today)

KMP_AFFINITY is also intel stuff

The OpenMP standard env vars for thread pinning are OMP_PLACES and OMP_PROC_BIND , IIRC

devito/core/autotuning.py
```diff
@@ -75,16 +75,13 @@ def autotune(operator, args, level, mode):
     # Detect the time-stepping Iteration; shrink its iteration range so that
     # each autotuning run only takes a few iterations
     steppers = {i for i in flatten(trees) if i.dim.is_Time}
-    if len(steppers) == 0:
-        stepper = None
-        timesteps = 1
-    elif len(steppers) == 1:
+    if len(steppers) == 1:
```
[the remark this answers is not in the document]
FabioLuporini: if CI doesn't fail, either test suite is flawed or we somewhat changed things over time such that it's never the case.

Now, the second case to me is unlikely...one could definitely try auto-tuning without a time loop. Hence, the test suite deserves an update !

devito/core/autotuning.py
```diff
@@ -319,13 +316,14 @@ def generate_block_shapes(blockable, args, level):
         for bs in list(ret):
             handle = []
             for v in options['blocksize-l1']:
-                # To be a valid blocksize, it must be smaller than and divide evenly
-                # the parent's block size
-                if all(v <= i and i % v == 0 for _, i in bs):
+                # To be a valid block size, it must be smaller than
+                # and divide evenly the parent's block size.
+                # Blocksizes equal to the parent's block size are not included
```
FabioLuporini, line 321: also why not? it would be sort-of like defaulting to the non-hierarchical case no?

benchmarks/user/advisor/run_advisor.py
georgebisbas: I rephrased a bit. Hope its better now

benchmarks/user/advisor/run_advisor.py
```diff
@@ -109,21 +117,20 @@ def run_with_advisor(path, output, name, exec_args):
         '-run-pass-thru=--no-altstack',  # Avoids `https://software.intel.com/en-us/vtune-amplifier-help-error-message-stack-size-is-too-small`  # noqa
         '-run-pass-thru=-timestamp=sys',  # Avoids 'VTune Amplifier may detect which timer source to use incorrectly on Intel® Xeon® processor E5-XXXX processors (200287361)' # noqa
         '-strategy ldconfig:notrace:notrace',  # Avoids `https://software.intel.com/en-us/forums/intel-vtune-amplifier-xe/topic/779309`  # noqa
-        '-start-paused',  # The generated code will enable/disable Advisor on a loop basis
+        '-start-paused',  # The generated code will enable/disable Advisor on a loop basis according to the decorated pragmas  # noqa
```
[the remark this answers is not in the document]
georgebisbas: Just long comment, similar to the way the others are (lines 117-120)

devito/core/autotuning.py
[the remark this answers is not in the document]
georgebisbas: The 0 case is removed since devito does not apply loop blocking for non time-iterative computations. 
https://github.com/devitocodes/devito/blob/45c75a7591e8f098c8b48c21ea639d5bf1bf368a/devito/passes/clusters/blocking.py#L148

CI passes.

Idea: blockrelax should allow blocking non-time-iterative loops, right? It skips the heuristics. Then we need to add some tests for this.

i.e. block relax on matrix multiplication example?
I need to check

devito/core/autotuning.py
georgebisbas: I would expect that if someone asks for >1 levels would not be interested in that....

devito/core/autotuning.py
```diff
@@ -319,13 +316,14 @@ def generate_block_shapes(blockable, args, level):
...
+                if all(v < i and i % v == 0 for _, i in bs):
                     ret.append(bs + tuple((d.step, v) for d in level_1))
             ret.remove(bs)
 
-    # Generate level-n (n > 1) block shapes
+    # Generate level-n (n > 2) block shapes
```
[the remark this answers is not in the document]
georgebisbas: It is just typo-like.
Nothing changes

devito/core/autotuning.py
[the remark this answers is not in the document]
FabioLuporini: sounds good to me yeah

devito/core/autotuning.py
FabioLuporini: as in, why would you remove a bunch of points from the exploration space?

mloubout (approved):

PR #1931, opened 2022-06-06
arch: Support OpenMP on Apple M1

mloubout:
Add flags and libs for openmp support on apple m1. Leads to decent performance gain (over gcc-11 on my laptop)

### DIFF
--- a/devito/arch/archinfo.py
+++ b/devito/arch/archinfo.py
@@ -378,7 +378,7 @@ def get_cuda_path():
 
 
 @memoized_func
-def get_m1_llvm_path():
+def get_m1_llvm_path(language):
     # Check if Apple's llvm is installed (installable via Homebrew), which supports
     # OpenMP.
     # Check that we are on apple system
@@ -395,14 +395,17 @@ def get_m1_llvm_path():
                   if "InstalledDir" in v][0][:-4]
         # ** Fourth check if the libraries are installed
         if os.path.exists(os.path.join(prefix, 'lib', 'libomp.dylib')):
-            libs = "%s/lib" % prefix
-            includes = "%s/include" % prefix
-            return {'libs': libs, 'includes': includes}
-        else:
+            libs = os.path.join(prefix, "lib")
+            include = os.path.join(prefix, "include")
+            return {'libs': libs, 'include': include}
+        elif language == "openmp":
             warning("Apple's arm64 clang found but openmp libraries not found."
                     "Install Apple LLVM for OpenMP support, i.e. `brew install llvm`")
+        else:
+            pass
     else:
-        warning("Apple's x86 clang found, OpenMP is not supported.")
+        if language == "openmp":
+            warning("Apple's x86 clang found, OpenMP is not supported.")
     return None
 
 
--- a/devito/arch/archinfo.py
+++ b/devito/arch/archinfo.py
@@ -1,6 +1,6 @@
 """Collection of utilities to detect properties of the underlying architecture."""
 
-from subprocess import PIPE, Popen, DEVNULL
+from subprocess import PIPE, Popen, DEVNULL, run
 
 from cached_property import cached_property
 import cpuinfo
@@ -9,12 +9,13 @@ import numpy as np
 import psutil
 import re
 import os
+import sys
 
 from devito.logger import warning
 from devito.tools import as_tuple, all_equal, memoized_func
 
 __all__ = ['platform_registry', 'get_cpu_info', 'get_gpu_info', 'get_nvidia_cc',
-           'get_cuda_path', 'check_cuda_runtime',
+           'get_cuda_path', 'check_cuda_runtime', 'get_m1_llvm_path',
            'Platform', 'Cpu64', 'Intel64', 'Amd', 'Arm', 'Power', 'Device',
            'NvidiaDevice', 'AmdDevice', 'IntelDevice',
            'INTEL64', 'SNB', 'IVB', 'HSW', 'BDW', 'SKX', 'KNL', 'KNL7210',  # Intel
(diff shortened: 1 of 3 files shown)


devito/arch/compiler.py
```diff
@@ -410,9 +412,15 @@ def __init__(self, *args, **kwargs):
...
-            # openmp unusable
-            pass
+            # This requires to install Aplle's llvm (through brew) to install the openmp
+            # librairies and c++ includes. Once installed, both Apple's default clang and
+            # howebrew clang support openmp. Check that openmnp lib exist.
+            if path.exists('/opt/homebrew/opt/llvm/lib/libomp.dylib'):
+                self.ldflags += ['-mcpu=apple-m1', '-fopenmp']
+                self.ldflags += ['-L/opt/homebrew/opt/llvm/lib']
+            else:
+                warning("Openmp librairies not foud. Install howebrew llvm:"
```
georgebisbas, line 422: typo

devito/arch/compiler.py
```diff
@@ -410,9 +412,15 @@ def __init__(self, *args, **kwargs):
                 self.ldflags += ['-march=%s' % platform.march]
         elif platform is M1:
             # NOTE:
-            # -march=native unsupported
-            # openmp unusable
-            pass
+            # This requires to install Aplle's llvm (through brew) to install the openmp
+            # librairies and c++ includes. Once installed, both Apple's default clang and
+            # howebrew clang support openmp. Check that openmnp lib exist.
+            if path.exists('/opt/homebrew/opt/llvm/lib/libomp.dylib'):
```
FabioLuporini, line 418: It'd be nice to add a (memoized) `get_clang_path` to `arch/archinfo.py`  and then call it here , just like we do with other paths (e.g., CUDA). Hardcoding a path here is a bit ugly

devito/arch/compiler.py
```diff
@@ -410,9 +412,15 @@ def __init__(self, *args, **kwargs):
                 self.ldflags += ['-march=%s' % platform.march]
         elif platform is M1:
             # NOTE:
-            # -march=native unsupported
-            # openmp unusable
-            pass
+            # This requires to install Aplle's llvm (through brew) to install the openmp
```
FabioLuporini, line 415: there are a few typos:

Apple
libraries

devito/arch/compiler.py
```diff
@@ -410,9 +412,15 @@ def __init__(self, *args, **kwargs):
...
+            # librairies and c++ includes. Once installed, both Apple's default clang and
+            # howebrew clang support openmp. Check that openmnp lib exist.
```
FabioLuporini, line 417: no full stop at the end

devito/arch/compiler.py
```diff
@@ -410,9 +412,15 @@ def __init__(self, *args, **kwargs):
...
+            if path.exists('/opt/homebrew/opt/llvm/lib/libomp.dylib'):
+                self.ldflags += ['-mcpu=apple-m1', '-fopenmp']
+                self.ldflags += ['-L/opt/homebrew/opt/llvm/lib']
+            else:
+                warning("Openmp librairies not foud. Install homebrew llvm:"
```
FabioLuporini, line 422: OpenMP libraries 
found
Homebrew
LLVM

FabioLuporini, line 422: this warning could also be moved inside the aforementioned `get_clang_path` (or `get_llvm_path` FWIW)

Optionally, you could also return a dict from this function -- one entry for the include dir, one entry for the libs (where you would find `libomp.dylib`), etc

devito/arch/compiler.py
mloubout: Done

devito/tools/memoization.py
```diff
@@ -22,16 +22,17 @@ def __init__(self, func):
         self.func = func
         self.cache = {}
 
-    def __call__(self, *args):
+    def __call__(self, *args, **kw):
```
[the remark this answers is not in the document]
FabioLuporini: it's OK

FabioLuporini (approved):
Sweeeet! Love what you did here

Approved!

devito/arch/archinfo.py
```diff
@@ -376,6 +377,35 @@ def get_cuda_path():
...
+    # *** Second check if this clang version targets arm64 (M1)
+    if "arm64-apple" in ver:
+        # *** Third extract install path. clang version command contains the line:
+        # InstalledDir: /path/to/llvm/bin
+        # which points to the llvm root directory we need for libraries and includes
+        prefix = [v.split(': ')[-1] for v in ver.split('\n')
+                  if "InstalledDir" in v][0][:-4]
+        # ** Fourth check if the libraries are installed
+        if os.path.exists(os.path.join(prefix, 'lib', 'libomp.dylib')):
+            libs = "%s/lib" % prefix
```
FabioLuporini, line 398: `os.path.join(prefix, lib)` a bit more portable, but hey, nitpicking here...

devito/arch/compiler.py
```diff
@@ -408,11 +410,15 @@ def __init__(self, *args, **kwargs):
                                  '-fopenmp-targets=amdgcn-amd-amdhsa',
                                  '-Xopenmp-target=amdgcn-amd-amdhsa']
                 self.ldflags += ['-march=%s' % platform.march]
-        elif platform is M1:
+        elif platform is M1 and language == 'openmp':
```
FabioLuporini, line 413: shouldn't this `and language == 'openmp'` be moved down in `and` with `if llvmm1` ?

FabioLuporini (changes requested):
I had approved it, but I think the if-condition must be tweaked or it will break for non-llvm compilers

devito/arch/archinfo.py
mloubout: You mean like the line above 😅 will fix

devito/arch/compiler.py
mloubout: Not sure it matters much since we in ClangCompiler in practice but would be more logical yeah will fix

devito/arch/archinfo.py
mloubout: fixed

devito/arch/compiler.py
mloubout: fixed

devito/arch/archinfo.py
```diff
@@ -376,6 +377,35 @@ def get_cuda_path():
...
+def get_m1_llvm_path():
+    # Check if Apple's llvm is installed (installable via Homebrew), which supports
+    # OpenMP.
+    # Check that we are on apple system
+    if sys.platform != 'darwin':
+        raise ValueError('Apple LLVM is only available on Mac OS X.')
+    # *** First check if LLVM is installed
+    ver = run(["clang", "--version"], stdout=PIPE, stderr=DEVNULL).stdout.decode("utf-8")
+    # *** Second check if this clang version targets arm64 (M1)
+    if "arm64-apple" in ver:
```
georgebisbas, line 390: Shouldn;t this be try except? How can you ensure that Apple's x86 clang found?

devito/arch/compiler.py
```diff
@@ -410,9 +412,13 @@ def __init__(self, *args, **kwargs):
                 self.ldflags += ['-march=%s' % platform.march]
         elif platform is M1:
             # NOTE:
-            # -march=native unsupported
-            # openmp unusable
-            pass
+            # Apple M1 supports OpenMP through Apple's LLVM compiler.
+            # The compiler can be installed with Homebrew or can be built from scratch.
+            # Check if installed and set compiler flags accordingly
+            llvmm1 = get_m1_llvm_path()
```
georgebisbas, line 418: I would switch to llvm_m1 for readability
or....m1_llvm for more homogeneity with the get_m1...function

devito/arch/archinfo.py
mloubout: Because apple system comes with clang as factory default the same way any linux comes with GCC 4.9.
This function also can only be hit from withing ClangCompiler

PR #1932, opened 2022-06-08
dsl: Implementation of new interpolation methods

guaacoelho:
Hi Guys @FabioLuporini @mloubout @navjotk @georgebisbas ,

I've been working with @nogueirapeterson to include new interpolation techniques on Devito. We implement the cubic and sinc interpolation methods. Thus, we create two new classes: CubicInterpolator and SincInterpolator, which contains, respectively, the cubic and sinc interpolation methods.

The cubic interpolator was implemented using the cubic convolution technique (10.1109/TASSP.1981.1163711) and the sinc interpolator was implemented using Kaiser windowed sinc functions (https://doi.org/10.1190/1.1451454).

We hope you appreciate this contribution and help us to improve this implementation with your great review!

### DIFF
--- a/devito/operations/interpolators.py
+++ b/devito/operations/interpolators.py
@@ -588,11 +588,11 @@ class CubicInterpolator(GenericInterpolator):
 
             # Checks if the data is 3D or 2D
             if len(dim_pos) == 3:
-                eqs = [np.sum(v) for v in self._tricubic_equations(_expr,
+                eqs = [sympy.Add(*v) for v in self._tricubic_equations(_expr,
                                                                    idx_subs,
                                                                    dim_pos=dim_pos)]
             else:
-                eqs = [np.sum(v) for v in self._bicubic_equations(_expr,
+                eqs = [sympy.Add(*v) for v in self._bicubic_equations(_expr,
                                                                   idx_subs,
                                                                   dim_pos=dim_pos)]
 
@@ -840,7 +840,7 @@ class SincInterpolator(GenericInterpolator):
             else:
                 result = self._sinc_equations3D(_expr, idx_subs, dim_pos=dim_pos)
 
-            eqs = [np.sum(v) for v in result]
+            eqs = [sympy.Add(*v) for v in result]
 
             summands.extend([Inc(rhs, v, implicit_dims=self.sfunction.dimensions)
                              for v in eqs])
--- a/devito/operations/interpolators.py
+++ b/devito/operations/interpolators.py
@@ -1,6 +1,4 @@
 from abc import ABC, abstractmethod
-# Polly interpolator é a que contem as duas interpolações
-# Backup cubic e sinc são as originais
 import sympy
 import numpy as np
 from cached_property import cached_property
--- a/devito/operations/interpolators.py
+++ b/devito/operations/interpolators.py
@@ -297,6 +297,7 @@ class PolynomialInterpolator(GenericInterpolator):
 
         return eqns
 
+
 class LinearInterpolator(PolynomialInterpolator):
 
     """
--- a/devito/operations/interpolators.py
+++ b/devito/operations/interpolators.py
@@ -406,6 +406,43 @@ class CubicInterpolator(GenericInterpolator):
     def grid(self):
         return self.sfunction.grid
 
+    def _interpolation_indices(self, variables, offset=0, field_offset=0):
+        """
+        Generate interpolation indices for the DiscreteFunctions in ``variables``.
+        """
+        index_matrix, points = self.sfunction._index_matrix(offset)
+
+        idx_subs = []
+        for i, idx in enumerate(index_matrix):
+            # Introduce ConditionalDimension so that we don't go OOB
+            mapper = {}
+            for j, d in zip(idx, self.grid.dimensions):
+                p = points[j]
+                lb = sympy.And(p >= d.symbolic_min - self.sfunction._radius,
+                               evaluate=False)
+                ub = sympy.And(p <= d.symbolic_max + self.sfunction._radius,
+                               evaluate=False)
+                condition = sympy.And(lb, ub, evaluate=False)
+                mapper[d] = ConditionalDimension(p.name, self.sfunction._sparse_dim,
+                                                 condition=condition, indirect=True)
+
+            # Track Indexed substitutions
+            idx_subs.append(mapper)
+
+        # Temporaries for the position
+        temps = [Eq(v, k, implicit_dims=self.sfunction.dimensions)
+                 for k, v in self.sfunction._position_map.items()]
+
+        temps.extend([Eq(v, k.subs(self.sfunction._position_map), implicit_dims=self.sfunction.dimensions)
+                     for k, v in self.sfunction._relative_position_map.items()])
+
+        # Temporaries for the indirection dimensions
+        temps.extend([Eq(v, k.subs(self.sfunction._position_map),
+                         implicit_dims=self.sfunction.dimensions)
+                      for k, v in points.items()])
+
+        return idx_subs, temps
+
     def interpolate(self, expr, offset=0, increment=False, self_subs={}):
         """
         Generate equations interpolating an arbitrary expression into ``self``.
@@ -432,8 +469,11 @@ class CubicInterpolator(GenericInterpolator):
             # Need to get origin of the field in case it is staggered
             # TODO: handle each variable staggereing spearately
             field_offset = variables[0].origin
-
-            return 
+            
+            # List of indirection indices for all adjacent grid points
+            idx_subs, temps = self._interpolation_indices(variables, offset,
+                                                          field_offset=field_offset)
+            return temps
 
         return Interpolation(expr, offset, increment, self_subs, self, callback)
 
--- a/devito/operations/interpolators.py
+++ b/devito/operations/interpolators.py
@@ -10,7 +10,7 @@ from devito.tools import powerset, flatten, prod
 from devito.types import (ConditionalDimension, Dimension, DefaultDimension, Eq, Inc,
                           Evaluable, Symbol, SubFunction)
 
-__all__ = ['LinearInterpolator', 'PrecomputedInterpolator']
+__all__ = ['LinearInterpolator', 'PrecomputedInterpolator', 'CubicInterpolator']
 
 
 class UnevaluatedSparseOperation(sympy.Expr, Evaluable):
@@ -386,5 +386,83 @@ class PrecomputedInterpolator(GenericInterpolator):
             rhs = prod(coeffs) * _expr
             _field = _field.subs(dim_subs)
             return [Eq(_field, _field + rhs.subs(dim_subs))]
+
+        return Injection(field, expr, offset, self, callback)
+
+
+class CubicInterpolator(GenericInterpolator):
+    """
+    Concrete implementation of GenericInterpolator implementing a cubic interpolation
+    scheme.
+    Parameters
+    ----------
+    sfunction: The SparseFunction that this Interpolator operates on.
+    """
+
+    def __init__(self, sfunction):
+        self.sfunction = sfunction
+
+    @property
+    def grid(self):
+        return self.sfunction.grid
+
+    def interpolate(self, expr, offset=0, increment=False, self_subs={}):
+        """
+        Generate equations interpolating an arbitrary expression into ``self``.
+        Parameters
+        ----------
+        expr : expr-like
+            Input expression to interpolate.
+        offset : int, optional
+            Additional offset from the boundary.
+        increment: bool, optional
+            If True, generate increments (Inc) rather than assignments (Eq).
+        """
+        def callback():
+
+            # Derivatives must be evaluated before the introduction of indirect accesses
+            try:
+                _expr = expr.evaluate
+            except AttributeError:
+                # E.g., a generic SymPy expression or a number
+                _expr = expr
+
+            variables = list(retrieve_function_carriers(_expr))
+
+            # Need to get origin of the field in case it is staggered
+            # TODO: handle each variable staggereing spearately
+            field_offset = variables[0].origin
+
+            return 
+
+        return Interpolation(expr, offset, increment, self_subs, self, callback)
+
+    def inject(self, field, expr, offset=0):
+        """
+        Generate equations injecting an arbitrary expression into a field.
+        Parameters
+        ----------
+        field : Function
+            Input field into which the injection is performed.
+        expr : expr-like
+            Injected expression.
+        offset : int, optional
+            Additional offset from the boundary.
+        """
+        def callback():
+            # Derivatives must be evaluated before the introduction of indirect accesses
+            try:
+                _expr = expr.evaluate
+            except AttributeError:
+                # E.g., a generic SymPy expression or a number
+                _expr = expr
+
+            variables = list(retrieve_function_carriers(_expr)) + [field]
+
+            # Need to get origin of the field in case it is staggered
+            field_offset = field.origin
+
+
+            return 
 
         return Injection(field, expr, offset, self, callback)
--- a/devito/operations/interpolators.py
+++ b/devito/operations/interpolators.py
@@ -626,7 +626,27 @@ class CubicInterpolator(GenericInterpolator):
             # Need to get origin of the field in case it is staggered
             field_offset = field.origin
 
+             # List of indirection indices for all adjacent grid points
+            idx_subs, temps = self._interpolation_indices(variables, offset,
+                                                          field_offset=field_offset)
 
-            return 
+            dim_pos = self.sfunction.grid.dimensions
+            if len(dim_pos) == 3:
+                eqs = [v for v in self._tricubic_equations(_expr,
+                                                           idx_subs,
+                                                           dim_pos=dim_pos)]
+            else:
+                eqs = [v for v in self._bicubic_equations(_expr,
+                                                          idx_subs,
+                                                          dim_pos=dim_pos)]
+
+            size = np.prod(np.shape(eqs))
+            eqs = [e[0] for e in np.reshape(eqs, (size, 1))]
+
+            eqns = [Inc(field.xreplace(vsub), b,
+                        implicit_dims=self.sfunction.dimensions)
+                    for b, vsub in zip(eqs, idx_subs)]
+
+            return temps + eqns
 
         return Injection(field, expr, offset, self, callback)
--- a/devito/operations/interpolators.py
+++ b/devito/operations/interpolators.py
@@ -524,14 +524,18 @@ class CubicInterpolator(GenericInterpolator):
             mapper = {}
             for j, d in zip(idx, self.grid.dimensions):
                 p = points[j]
-                lb = sympy.And(p >= d.symbolic_min - self.sfunction._radius,
-                               evaluate=False)
-                ub = sympy.And(p <= d.symbolic_max + self.sfunction._radius,
-                               evaluate=False)
-                condition = sympy.And(lb, ub, evaluate=False)
-                mapper[d] = ConditionalDimension(p.name, self.sfunction._sparse_dim,
-                                                 condition=condition, indirect=True)
-
+                # Only needs Conditional Dimensions if radius > variables.space_order
+                if all(list(map((lambda x: self.sfunction._radius <= x.space_order
+                       if hasattr(x, 'space_order') else False), variables))):
+                    mapper[d] = p
+                else:
+                    lb = sympy.And(p >= d.symbolic_min - self.sfunction._radius,
+                                   evaluate=False)
+                    ub = sympy.And(p <= d.symbolic_max + self.sfunction._radius,
+                                   evaluate=False)
+                    condition = sympy.And(lb, ub, evaluate=False)
+                    mapper[d] = ConditionalDimension(p.name, self.sfunction._sparse_dim,
+                                                     condition=condition, indirect=True)
             # Track Indexed substitutions
             idx_subs.append(mapper)
 
@@ -589,12 +593,12 @@ class CubicInterpolator(GenericInterpolator):
             # Checks if the data is 3D or 2D
             if len(dim_pos) == 3:
                 eqs = [sympy.Add(*v) for v in self._tricubic_equations(_expr,
-                                                                   idx_subs,
-                                                                   dim_pos=dim_pos)]
+                                                                       idx_subs,
+                                                                       dim_pos=dim_pos)]
             else:
                 eqs = [sympy.Add(*v) for v in self._bicubic_equations(_expr,
-                                                                  idx_subs,
-                                                                  dim_pos=dim_pos)]
+                                                                      idx_subs,
+                                                                      dim_pos=dim_pos)]
 
             # Creates Sympy equation using the values in eq as
             # the right side of the equation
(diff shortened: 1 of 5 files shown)


mloubout (commented):
Great! Thanks a lot. I'll have a thorough walk-through.

So `SincInterpolation` is not testable, right. It would be good to make it testable in some way. Also, there is a hardcoded path to `lookuptable.h` that will break the rest of the code.

tests/test_interpolation_cubic.py
```diff
@@ -0,0 +1,269 @@
+import numpy as np
+import pytest
+from devito import (Grid, Operator, Dimension, SparseFunction, SparseTimeFunction,
+                    Function, TimeFunction)
+
+
+def unit_box(name='a', shape=(11, 11), grid=None):
```
mloubout, line 7: Redefinition of existing fixture.
same below

tests/test_interpolation_cubic.py
```diff
@@ -0,0 +1,269 @@
...
+    for i, r in enumerate(ranges):
+        points.coordinates.data[:, i] = np.linspace(r[0], r[1], npoints)
+    return points
+
+
+@pytest.mark.parametrize('shape, coords', [
+    ((101, 101), [(3.4, 95.), (3., 92)]),
+    ((101, 101, 101), [(3.4, 95.), (3., 92), (10., 10.)])
+])
+def test_interpolate(shape, coords, npoints=20):
```
mloubout, line 103: Can this all be gathered under a common test class?

Also I don't think it needs a new file, `test_interpolation.py` exists for this

tests/test_interpolation_cubic.py
```diff
@@ -0,0 +1,269 @@
...
+    """
+
+    a = exp_box(shape=shape)
+    p = points(a.grid, coords, npoints=npoints, cubic=True)
+    expr = p.interpolate(a)
+    op = Operator(expr)
+    # print(p.coordinates.data)
+    op(a=a)
+
+    # Precisão 3D nao bate em 10-6 mas sim em 10-5
```
mloubout, line 115: There is leftover Portugues throughout

tests/test_interpolation_cubic.py
```diff
@@ -0,0 +1,269 @@
...
+def test_interpolate(shape, coords, npoints=20):
+    """Test generic point interpolation testing the x-coordinate of an
+    abitrary set of points going across the grid.
+    """
+
+    a = exp_box(shape=shape)
+    p = points(a.grid, coords, npoints=npoints, cubic=True)
+    expr = p.interpolate(a)
+    op = Operator(expr)
+    # print(p.coordinates.data)
```
mloubout, line 112: debug?

tests/test_interpolation_cubic.py
```diff
@@ -0,0 +1,269 @@
...
+    expected_values = [((point[1]**2) + 1) for point in p.coordinates.data]
+
+    assert np.allclose(p.data[:], expected_values, rtol=1e-5)
+
+
+@pytest.mark.parametrize('shape, coords', [
+    ((101, 101), [(3.4, 95.), (3., 92)]),
+    ((101, 101, 101), [(3.4, 95.), (3., 92), (10., 10.)])
+])
+def test_interpolate_time_shift(shape, coords, npoints=20):
```
mloubout, line 145: Are these all from `test_interpolation`? This looks like it could be a single `parametrize` change there to test linear and cubic cases.

devito/operations/interpolators.py
```diff
@@ -388,3 +389,520 @@ def callback():
...
+
+    def _interpolation_indices(self, variables, offset=0, field_offset=0):
+        """
+        Generate interpolation indices for the DiscreteFunctions in ``variables``.
+        """
+        index_matrix, points = self.sfunction._index_matrix(offset)
+
+        idx_subs = []
+        for i, idx in enumerate(index_matrix):
+            # Introduce ConditionalDimension so that we don't go OOB
```
mloubout, line 520: This should only be necessary if `r > variables.space_order`

devito/operations/interpolators.py
```diff
@@ -388,3 +389,520 @@ def callback():
...
+            eqs = [np.sum(v) for v in result]
+
+            summands.extend([Inc(rhs, v, implicit_dims=self.sfunction.dimensions)
+                             for v in eqs])
+
+            # Write/Incr `self`
+            lhs = self.sfunction.subs(self_subs)
+            last = [Inc(lhs, rhs)] if increment else [Eq(lhs, rhs)]
+
+            # Creates the symbolic equation that calls the populate function
```
mloubout, line 849: What's this?

devito/operations/interpolators.py
```diff
@@ -388,3 +389,520 @@ def callback():
...
+            idx_subs, temps = self._interpolation_indices(variables, offset,
+                                                          field_offset=field_offset)
+
+            rhs = Symbol(name='sum', dtype=self.sfunction.dtype)
+            summands = [Eq(rhs, 0., implicit_dims=self.sfunction.dimensions)]
+
+            dim_pos = self.sfunction.grid.dimensions
+
+            # Checks if the data is 3D or 2D
+            if len(dim_pos) == 3:
```
mloubout, line 587: this can be setup at `__init_` with `self._eqs = self.bicubic if self.sfunction.grid.dim = 2 else ...`

devito/operations/interpolators.py
```diff
@@ -388,3 +389,520 @@ def callback():
...
+                eqs = [np.sum(v) for v in self._tricubic_equations(_expr,
+                                                                   idx_subs,
+                                                                   dim_pos=dim_pos)]
+            else:
+                eqs = [np.sum(v) for v in self._bicubic_equations(_expr,
```
mloubout, line 592: why doesn't `..._equations` directly return the sum?
Also why `np.sum`? Isn't it still partially symbolic?

devito/operations/interpolators.py
```diff
@@ -388,3 +389,520 @@ def callback():
...
+            # Need to get origin of the field in case it is staggered
+            # TODO: handle each variable staggereing spearately
+            field_offset = variables[0].origin
+
+            # List of indirection indices for all adjacent grid points
+            idx_subs, temps = self._interpolation_indices(variables, offset,
+                                                          field_offset=field_offset)
+
+            rhs = Symbol(name='sum', dtype=self.sfunction.dtype)
+            summands = [Eq(rhs, 0., implicit_dims=self.sfunction.dimensions)]
```
mloubout, line 582: everything till here is the same as in `LinearInterpolation` right? and after the `bicubic_equation` too. So may be nice to gather those under a `PolynomialInterpolation` with `order=1` or `order=3` that specialize on the equations only since it's the only thing that changes

devito/operations/interpolators.py
```diff
@@ -388,3 +389,520 @@ def callback():
             return [Eq(_field, _field + rhs.subs(dim_subs))]
 
         return Injection(field, expr, offset, self, callback)
+
+
+class CubicInterpolator(GenericInterpolator):
+    """
+    Concrete implementation of GenericInterpolator implementing a cubic interpolation
+    scheme.
```
FabioLuporini, line 397: blank line

devito/operations/interpolators.py
```diff
@@ -388,3 +389,520 @@ def callback():
...
+
+        eqs = []
+        eqs.extend([self._cubic_equation(expr, pos, dim_pos,
+                    idx_subs=idx_subs[(ind*4)+(16*ii):((1+ind)*4)+16*ii],
+                    idx2d=ind, idx3d=ii)
+                    for ii in range(4) for ind in range(4)])
+
+        return eqs
+
+    def _interpolation_indices(self, variables, offset=0, field_offset=0):
```
FabioLuporini, line 512: isn't there anything we could factorize in a common superclass?

devito/operations/interpolators.py
```diff
@@ -388,3 +389,520 @@ def callback():
...
+    Concrete implementation of GenericInterpolator implementing a Sinc interpolation
+    scheme, using a method that is windowed by a Kaiser Function.
+    Parameters
+    ----------
+    sfunction: The SparseFunction that this Interpolator operates on.
+    """
+
+    def __init__(self, sfunction):
+        self.sfunction = sfunction
+        self.r = 4
```
FabioLuporini, line 671: This should probably be a class property

tests/test_interpolation_cubic.py
```diff
@@ -0,0 +1,269 @@
+import numpy as np
```
FabioLuporini, line 1: IMHO this file should be squashed with `test_interpolation`, and as much code and existing tests as possible should be reused, through `pytest.parametrize` for example

FabioLuporini (changes requested):
Thank you for this contribution. Looking great! I just left a bunch of comments

tests/test_interpolation_cubic.py
```diff
@@ -0,0 +1,269 @@
...
+        points.coordinates.data[:, i] = np.linspace(r[0], r[1], npoints)
+    return points
+
+
+@pytest.mark.parametrize('shape, coords', [
+    ((101, 101), [(3.4, 95.), (3., 92)]),
+    ((101, 101, 101), [(3.4, 95.), (3., 92), (10., 10.)])
+])
+def test_interpolate(shape, coords, npoints=20):
+    """Test generic point interpolation testing the x-coordinate of an
```
EdCaunt, line 104: Opening triple quotes want to be on line above for consistency with the rest of the codebase, i.e.
```
"""
Test generic...
"""
```

devito/operations/interpolators.py
```diff
@@ -388,3 +389,520 @@ def callback():
...
+        return self.sfunction.grid
+
+    def _cubic_equation(self, expr, position, dim_pos, idx_subs, idx2d, idx3d=None):
+        """
+        Generate the basic cubic equation.
+        Parameters
+        ----------
+        expr : expr-like
+            Input expression to interpolate.
+        position: List
```
EdCaunt, line 417: Should be `list` (all lowercase)

devito/operations/interpolators.py
```diff
@@ -388,3 +389,520 @@ def callback():
...
+            Elements which represents the position of the interpolation point
+            multiplied by their respective coefficients.
+        dim_pos: Symbol
+            Symbol representing the dimension wich will be used to form the sinc
+            equations.
+        idx_subs: dict
+            Structure responsible for mapping the order of substitution of dimensions
+            in expr.
+        idx_2d: Integer
```
EdCaunt, line 693: Should be `int` I think?

devito/operations/interpolators.py
```diff
@@ -388,3 +389,520 @@ def callback():
...
+        # n = number of neighbor points
+        n = self.r*2
+
+        # Generation of equations
+        eqs = [self._sinc_equation(expr, pos, dim_pos, idx_subs[ii*n:(ii+1)*n], idx2d=ii)
+               for ii in range(n)]
+
+        return eqs
+
+    def _sinc_equations3D(self, expr, idx_subs, dim_pos=None):
```
EdCaunt, line 744: Again, can these be consolidated into `_sinc_equationsND()`? Seems like a decent amount of lines are the same and would be nice to have a generalization to N-D

tests/test_interpolation_cubic.py
EdCaunt, line 1: Would it be worth having a test to check the convergence of the cubic interpolator? Not sure what convergence would be expected for the sinc one

mloubout: It isn't as I pointed out above

EdCaunt: Maybe I'm just being stupid and missing it, but where?

mloubout: Sorry completely missed what you were talking about only saw the `numpy`.

Convergence may be tricky, it's usually easier to just do an analytical check where you set the field to polynomial values and check the interpolation is exact.

devito/operations/interpolators.py
guaacoelho: To implement the sinc interpolation I used a look up table that holds the values of the coefficients. Before it could be accessed, it was necessary to initialize this table with the values. This symbolic Function `populate()` creates a call to the function responsible for initialize the look up table.

This function is defined inside look up table.h, that I apparently forgot to add here. I'm attaching the lookuptable.h in this comment, this way you can see how it works

[lookuptable.zip](https://github.com/devitocodes/devito/files/8872262/lookuptable.zip)

devito/operations/interpolators.py
guaacoelho: So if  `r <= variables.space_order`  there is no need to use  ConditionalDimension in this case?

devito/types/sparse.py
```diff
@@ -519,6 +553,11 @@ def _position_map(self):
     @cached_property
     def _point_increments(self):
         """Index increments in each dimension for each point symbol."""
+        if self.cubic:
+            return tuple(product(range(-1, 3), repeat=self.grid.dim))
+        elif self.sinc:
+            r = self.interpolator.r
+            return tuple(product(range(-r + 1, r+1), repeat=self.grid.dim))
```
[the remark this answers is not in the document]
guaacoelho: You're right! `range(-r + 1, r+1)` works for all r.

 I'll change this

devito/types/sparse.py
```diff
@@ -494,6 +504,30 @@ def _point_symbols(self):
...
+        """
+        Symbols map for the position of the sparse points relative to the indices
+        of the vector wich contain and acess the data
+        Notes
+        -----
+        The expression `(coord - origin)/spacing` could also be computed in the
+        mathematically equivalent expanded form `coord/spacing -
+        origin/spacing`. This particular form is problematic when a sparse
+        point is in close proximity of the grid origin, since due to a larger
+        machine precision error it may cause a +-1 error in the computation of
```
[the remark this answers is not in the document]
guaacoelho: Actually this was my mistake, I took the same comments from position_map function and forgot to update after the rebase. So there is no error in here

devito/operations/interpolators.py
guaacoelho: Not returning directly the sum, the return of `..._equations` can be used for both injection and interpolation. In interpolation it sum the values and in injection just assign the values for its respective grid point.

About the use of numpy: Yeah, it is still a symbolic. I wanted to generate a symbolic equation that sum all the individual values, that's why i used `np.sum`. Is there some problem in using it?

devito/operations/interpolators.py
mloubout: Yes

devito/operations/interpolators.py
```diff
@@ -388,3 +389,520 @@ def callback():
...
+        When idx_subs is not None, expr is used as element which represents the neighbors
+        points. Otherwise, pn is used.
+        """
+
+        def cubicConv(arg, option):
+            """
+            option = 1, ponto imediatamente vizinho ao ponto de intepolação.
+            Onde 0 < |arg| < 1
+            option = 2, pontos extremos da vizinhança.
+            Onde 1 < |arg| < 2
```
navjotk, line 435: Am I seeing an old version? We probably want our documentation in English :)

devito/operations/interpolators.py
navjotk: ...also lets try to setup as much at init as possible?

devito/operations/interpolators.py
navjotk: Does this actually work? Numpy summing over sympy symbols?

devito/operations/interpolators.py
```diff
@@ -388,3 +389,520 @@ def callback():
...
+            # List of indirection indices for all adjacent grid points
+            idx_subs, temps = self._interpolation_indices(variables, offset,
+                                                          field_offset=field_offset)
+
+            dim_pos = self.sfunction.grid.dimensions
+            if len(dim_pos) == 3:
+                eqs = [v for v in self._tricubic_equations(_expr,
+                                                           idx_subs,
+                                                           dim_pos=dim_pos)]
+            else:
```
navjotk, line 643: Can we do this with inheritance? Or in a way that generalises beyond 2/3 dimensions?

devito/operations/interpolators.py
navjotk: Is it a problem to let the user choose?

devito/operator/operator.py
```diff
@@ -133,7 +134,8 @@ class Operator(Callable):
     """
 
     _default_headers = [('_POSIX_C_SOURCE', '200809L')]
-    _default_includes = ['stdlib.h', 'math.h', 'sys/time.h']
+    _default_includes = ['stdlib.h', 'math.h', 'sys/time.h',
+                         f'{os.path.expanduser("~")}/lookuptable.h']
```
navjotk, line 138: I need help understanding what this header file is doing. Also, is this a static header file that should become a standard part of Devito? Alternately, can this be generated using the JIT infra?

devito/types/sparse.py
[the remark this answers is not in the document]
navjotk: So in the end we won't have this conditional, I hope?

navjotk (commented):
Left some minor comments. 

Also have two higher-level questions:
1. Is `SincInterpolator` the best name for what this is? (Sinc is only a part of the interpolation function)
2. How does this approach compare to using the machinery provided in `PrecomputedInterpolator`, i.e. the Sinc class could inherit from that, precompute the coefficients (which I think you're doing in some way anyway), and hand them off to `PrecomputedInterpolator` for the rest?

devito/operations/interpolators.py
guaacoelho: Yes, it works summing over sympy symbols. But I will take a look at Devito.Add(*v) to use it

devito/operator/operator.py
guaacoelho: To implement the sinc interpolation I used a look up table that holds the values of the coefficients. So this header contains the look up table and the methods responsible for initialize and get the values from it.

Access to these functions is defined from the definition of symbolic functions during the generation of symbolic equations.

I updated the PR adding the lookuptable.h file, this way you can take a look if you want to.

guaacoelho: About using JIT to generate the code, I don't know how it could be using JIT but I will take a look.

And yes, the static header file should become a standard part of Devito.

devito/types/sparse.py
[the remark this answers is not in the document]
guaacoelho: I updated the PR with this fix.

devito/operations/interpolators.py
guaacoelho: I updated the PR with this fix

guaacoelho: I adjust this part of the code and updated here. 

You'll see that I added a conditional`if hasattr(x, 'space_order') else False`. I needed to do this because in some cases `variables` does not have the `space_order` attribute

mloubout:
The `.h` file shouldn't really be necessary. Bessel function are defined in `math.h` so should be able to make a symbolic expression directly calling it rather than the lookup table

navjotk:
Also, I want to understand if this is significantly different from using PrecomputedInterpolator (i.e. this class could extend and use the functionality there)

guaacoelho:
> Also, I want to understand if this is significantly different from using PrecomputedInterpolator (i.e. this class could extend and use the functionality there)

At the same way that linear interpolation, both cubic and sinc interpolation also can be done by using PrecomputedInterpolator, if the right coefficients are used. So I can say that the main difference is that this new classes removes the responsibility of obtaining the interpolation coefficients from the user.

devito/operations/interpolators.py
guaacoelho: > Is it a problem to let the user choose?

The Bessel function was implemented inside the .h file considering `r = 4`. So it would be a problem to allow the user to choose the value of r

devito/operations/interpolators.py
guaacoelho: > isn't there anything we could factorize in a common superclass?

Yes, I think it has. I'm working on it

navjotk:
@guaacoelho But LinearInterpolator does not do any pre computation. If this class is doing pre computation of coefficients, it could store them in the properties provided by `PrecomputedInterpolator` and use the rest of the infrastructure there. i.e. This class precomputes the coefficients and hands them off to `PrecomputedInterpolator`, not the user.

devito/operations/interpolators.py
guaacoelho: I added a PolynomialInterpolator class that works as a parent class. It's not exactly what you just said, but I think it solves the code reuse problem

devito/operations/interpolators.py
guaacoelho: I updated the code by adding the superclass you suggested

devito/operations/interpolators.py
```diff
@@ -388,3 +389,520 @@ def callback():
...
+            Structure containing the spatial dimensions of the field
+        """
+        pos = self.sfunction._point_symbols
+
+        eqs = [self._cubic_equation(expr, pos, dim_pos,
+               idx_subs=idx_subs[ii*4:(ii+1)*4], idx2d=ii)
+               for ii in range(4)]
+        return eqs
+
+    def _tricubic_equations(self, expr, idx_subs, dim_pos=None):
```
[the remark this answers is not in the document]
guaacoelho: Yes, it could. I changed the code and made this change you suggested

devito/operations/interpolators.py
```diff
@@ -1,5 +1,6 @@
 from abc import ABC, abstractmethod
-
+# Polly interpolator é a que contem as duas interpolações
```
georgebisbas, line 2: Translate and avoid comment in this area

devito/operations/interpolators.py
```diff
@@ -294,44 +446,113 @@ def callback():
...
+        idx_subs, temps = super(CubicInterpolator, self)._interpolation_indices(
+            variables, offset, field_offset)
+        temps.extend(self.eq_relative_position())
+
+        return idx_subs, temps
 
     def interpolate(self, expr, offset=0, increment=False, self_subs={}):
         """
         Generate equations interpolating an arbitrary expression into ``self``.
-
```
georgebisbas, line 334: add line back

devito/operations/interpolators.py
```diff
@@ -342,27 +563,203 @@ def interpolate(self, expr, offset=0, increment=False, self_subs={}):
...
+                _expr = expr.evaluate
+            except AttributeError:
+                # E.g., a generic SymPy expression or a number
+                _expr = expr
 
-            return [Eq(lhs, lhs + rhs)]
+            variables = list(retrieve_function_carriers(_expr))
+
+            # Need to get origin of the field in case it is staggered
+            # TODO: handle each variable staggereing spearately
```
georgebisbas, line 577: typo

devito/operator/operator.py
```diff
@@ -133,7 +134,8 @@ class Operator(Callable):
     """
 
     _default_headers = [('_POSIX_C_SOURCE', '200809L')]
-    _default_includes = ['stdlib.h', 'math.h', 'sys/time.h']
+    _default_includes = ['stdlib.h', 'math.h', 'sys/time.h',
+                         f'{os.path.expanduser("~/devito")}/lookuptable.h']
```
georgebisbas, line 138: Ι _think_ this .h is not something we encourage

georgebisbas (commented):
Thanks for this contribution!

mloubout:
FYI sparse interpolation is beeing revamp in #2128  which should help updae this PR.

mloubout:
sinc interpolation added in #2342 , closing.

Feel free to open a new PR for cubic with the new API.

PR #1934, opened 2022-06-13
install: Update actions, checking `latest` not showing up

georgebisbas:
Checking why devitocodes/devito:latest is not being published


.github/workflows/docker-publish.yml
```diff
@@ -41,7 +41,7 @@ jobs:
 
       - name: CPU image release
         if: github.event_name == 'release'
-        uses: docker/build-push-action@v2.4.0
+        uses: docker/build-push-action@v3.0.0
```
mloubout, line 44: There is some v3 and some v3.0.0 would pick one

mloubout (approved):

PR #1941, opened 2022-06-14
Adapt CUDA compatability for CUDA 11.1

ggorman:


devito/data/allocators.py
```diff
@@ -431,3 +432,45 @@ def default_allocator(name=None):
             return ALLOC_NUMA_LOCAL
     else:
         return ALLOC_FLAT
+
+class MmapAllocator(MemoryAllocator):
```
FabioLuporini, line 436: wrong branch

devito/arch/archinfo.py
```diff
@@ -394,9 +394,20 @@ def check_cuda_runtime():
...
+        # 2. version returned as (1000 major + 10 minor), eg CUDA 9.2 -> 9020
         driver_version = driver_version.value
         runtime_version = runtime_version.value
-        if driver_version < runtime_version:
+        if runtime_version < 11010:
+            cuda_incompat = driver_version < runtime_version
+        else:
+            driver_version_major = int(driver_version.value/1000)
+            runtime_version_major = int(runtime_version.value/1000)
+            cuda_incompat = driver_versioni_major < runtime_version_major
```
mloubout, line 408: why is the /1000 .... needed, wouldn't `driver_version < runtie_version` still be true without it it's just a scaling.
Also `.value` taken twice, already done line 401

PR #1942, opened 2022-06-14
api: Fix Derivative substitution

mloubout:
Fixes @EdCaunt issue with substitutions.

### DIFF
--- a/devito/finite_differences/derivative.py
+++ b/devito/finite_differences/derivative.py
@@ -228,7 +228,11 @@ class Derivative(sympy.Derivative, Differentiable):
         """
         # Check if trying to replace the whole expression
         if self in subs:
-            return subs[self]
+            new = subs.pop(self)
+            try:
+                return new._xreplace(subs)
+            except AttributeError:
+                return new, True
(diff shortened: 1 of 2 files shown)


FabioLuporini (approved):

devito/finite_differences/derivative.py
```diff
@@ -209,6 +209,11 @@ def subs(self, *args, **kwargs):
         """
         Bypass sympy.Subs as Devito has its own lazy evaluation mechanism.
         """
+        # Check if we are calling subs(self, old, new, **hint) in which case
+        # return the standard substitution. Need to check `==` rather than `is`
+        # because a new derivative could be created i.e `f.dx.subs(f.dx, y)`
+        if len(args) == 2 and args[0] == self:
```
EdCaunt, line 215: Would it be worth also having a check if len(args) == 1 and type(args[0]) == list for the sympy use case where you supply a list of tuples of substitutions? Not sure how straightforward that would be though

mloubout: I don't think a list if a valid substitution in sympy, it expects either (old, new) or a dict.

georgebisbas (approved):

PR #1944, opened 2022-06-15
[WIP] examples: Add elastic adjoint operator 

nogueirapeterson:
Hi guys,
This PR contains the implementations of elastic adjoint equations. 
The implementation is based on Feng and Schuster (2017): Elastic least-squares reverse time migration (https://doi.org/10.1190/geo2016-0254.1).
I hope that you appreciate this contribution.


examples/seismic/elastic/operators.py
```diff
@@ -3,38 +3,134 @@
...
+        rec_vy = Receiver(name='rec_vy', grid=model.grid, time_range=geometry.time_axis,
+                          npoint=geometry.nrec)
+    name = "rec_tau" if forward else "rec"
+    rec = Receiver(name="%s" % name, grid=model.grid, time_range=geometry.time_axis,
+                   npoint=geometry.nrec)
 
-    # Create interpolation expression for receivers
-    rec_term1 = rec1.interpolate(expr=tau[-1, -1])
-    rec_term2 = rec2.interpolate(expr=div(v))
+    if forward:
```
mloubout, line 25: You don't really need this if/else, just a 

```
def sparce_expr(recs, src, v, tau):
   .....
```

and you just do `sparse_expr(src, recs, v, tau)` in the adjoint

examples/seismic/elastic/operators.py
```diff
@@ -3,38 +3,134 @@
...
+        src_xx = src.inject(field=tau[0, 0].forward, expr=src * s)
+        src_zz = src.inject(field=tau[-1, -1].forward, expr=src * s)
+        src_expr = src_xx + src_zz
+        if model.grid.dim == 3:
+            src_yy = src.inject(field=tau[1, 1].forward, expr=src * s)
+            src_expr += src_yy
+        # Create interpolation expression for receivers
+        rec_term_vx = rec_vx.interpolate(expr=v[0])
+        rec_term_vz = rec_vz.interpolate(expr=v[-1])
+        expr = tau[0, 0] + tau[-1, -1]
```
mloubout, line 37: exprt = sympy.Trace(tau).simplify()

examples/seismic/elastic/operators.py
```diff
@@ -3,38 +3,134 @@
...
+        """
+        Implementation of the elastic wave-equation from:
+        1 - Feng and Schuster (2017): Elastic least-squares reverse time migration
+        https://doi.org/10.1190/geo2016-0254.1
+
+        """
+
+        lpmu = lam + 2. * mu
+
+        pde_vx = rho * v[0].dt.T - (lpmu * tau[0, 0]).dx - (lam * tau[1, 1]).dx - \
```
mloubout, line 91: I'm sorry but I don't really like this. We have a nice Tensor forward PDE that looks very close to the math and clean and the adjoint has a Tensor expression as well this is not very readable and I would prefer a tensor adjoint equation as well.

examples/seismic/elastic/wavesolver.py
```diff
@@ -93,6 +106,59 @@ def forward(self, src=None, rec1=None, rec2=None, v=None, tau=None,
...
         # Execute operator and return wavefield and receiver data
-        summary = self.op_fwd(save).apply(src=src, rec1=rec1, rec2=rec2,
-                                          dt=kwargs.pop('dt', self.dt), **kwargs)
-        return rec1, rec2, v, tau, summary
+        summary = self.op_fwd(save).apply(src=src, rec_tau=rec_tau, rec_vx=rec_vx,
+                                          rec_vz=rec_vz, dt=kwargs.pop('dt', self.dt),
+                                          **kwargs)
+        if self.model.grid.dim == 3:
+            return rec_tau, rec_vx, rec_vz, v, tau, summary, rec_vy
+        return rec_tau, rec_vx, rec_vz, v, tau, summary
```
mloubout, line 114: thios change of order with a rec at the end in 3d isn't very good

mloubout (commented):
Thanks once again for the nice contribution.

I would prefer a vectorial/tensorial formulation to keep the nice and clean mathematical definitions though.

examples/seismic/elastic/operators.py
nogueirapeterson: Thanks @mloubout for the review. 
I also totally agree that adjoint formulation is not very readable. But I couldn't find the vectorial/tensorial adjoint 
formulation appropriate for Devito.
Actually, I have derived a vectorial/tensorial adjoint equation but I need to calculate the `diag(grad(TensorTimeFunction))` and the Devito doesn't have this option. Because of that, I decided to use this kind of implementation.
Could you give some work or the math steps for finding the adjoint wave equation with a friendly Tensor expression for Devito?
Please see my vectorial/tensorial adjoint equation derivation attached.
[Elastic notes.pdf](https://github.com/devitocodes/devito/files/8919521/Elastic.notes.pdf)

nogueirapeterson: @mloubout 
I am investigating the possibilities of using a compact formulation for the elastic adjoint wave equation (equation 5 of this [paper](https://doi.org/10.1190/geo2016-0254.1)). I would like to know if you are sure that it is possible to write this equation using vectorial/tensorial formulation with the current features of Devito.
In my point of view, I suppose it is necessary to implement some features to do this, but in fact, I'm not sure.
So, from your answer, I decide if I keep the PR and persist trying to implement the vector/tensorial formulation or if I remove the PR and move   my efforts to implement the neccessary features that allow a vectorial/tensorial implementation of elastic adjoint equation.

examples/seismic/elastic/operators.py
FabioLuporini: could also add `def trace` to our TensorFunctions API

examples/seismic/elastic/wavesolver.py
FabioLuporini: agreed.

Perhaps we could return a `namedtuple` instead? or just None... so that we always return the same number of arguments, in a given order

examples/seismic/elastic/operators.py
FabioLuporini: what would you expect `diag(grad(TensorTimeFunction))`to do ? you would be creating an N-dimensional tensor here, I'm not sure I follow

```
In [26]: from devito import diag

In [27]: print(diag.__doc__)

    Creates a diagonal tensor with func on its diagonal.

    Parameters
    ----------
    func : Differentiable or scalar
```

FabioLuporini: ah wait... I've now looked at your notes. Did you actually mean `diag(div)` ? that's definitely supported, I wouldn't see why it shouldn't work

FabioLuporini: is it an error in the maths ? you would be summing `diag(grad(TensorTimeFunction)` to `div(TensorTimeFunction)` which is a Vector, so that can't possibly be right

FabioLuporini (commented):
Thank you for contributing this ! Review in progress, I just left a few comments.

I agree with Mathias we need tensor notation. If something lacks in devito, we are keen to provide support

FabioLuporini:
closing as superseded by #2133

PR #1888, opened 2022-03-23
Fixing MPI slicing issue

rhodrin:
Fix for issue #1862. Lets see how CI reacts.

#### TODO:
- Tidy up the fix.
- Add lots of tests.

Issue #1862, opened 2022-03-08
MPI slicing broken


FabioLuporini:
Reproducer

```
import numpy as np
from devito import Grid, Function, norm

nx =  100
ny =  100

shape   = (nx,ny)

grid = Grid(shape=shape, dtype=np.float32)
f1 = Function(name='f1', grid=grid)

f1.data[:] = f1.data[:]
```

Run as:

```
DEVITO_MPI=1 mpirun -n 4 python prova2.py
```

rhodrin:
The problem appears to be within the `_setitem_` method when setting part/all of one Devito `Function`'s MPI distributed `Data` to another `Function`'s MPI distributed `Data`. Certain slicing short hands do not currently work.

In the above for example
`f1.data[:, :] = f1.data[:]`
would work, as would
```
b = np.array(f1.data_gather())
f.data[:]=b
```
This should be a one or two line fix. I previously fixed a similar bug in the `_getitem_` method.

rhodrin:
Fix is under way here: https://github.com/devitocodes/devito/tree/mpi_slicing_fix

In this branch, the above MFE is fixed but below is a new failing example with `-np 4`:
```
import numpy as np
from devito import Grid, Function, norm

nx = 8
ny = 8
nz = 8
shape1 = (nx, ny)
shape2 = (nx, ny, nz)
grid1 = Grid(shape=shape1, dtype=np.float32)
grid2 = Grid(shape=shape2, dtype=np.float32)
f = Function(name='f', grid=grid1)
g = Function(name='g', grid=grid2)

dat1 = np.arange(64, dtype=np.int32).reshape(grid1.shape)
g.data[0, :, :] = dat1

f.data[:] = g.data[0, :] # Result is incorrect
f.data[:, :] = g.data[0, :] # Result is correct
```
The above example is more complex in that `g.data[0,:]` is contained on ranks 0 and 1 but needs to be distributed to ranks 0, 1, 2 and 3. The fix will probably be an extension of the patch already in place, but need to dig a little further.

georgebisbas:
Fixed by https://github.com/devitocodes/devito/pull/1949

Issue #1010, opened 2019-11-20
Devito `Data` operations: RHS's with data distributed across different arrays doesn't work in the same way as serial


rhodrin:
Consider the following example distributed across, say, 4 ranks:
```
grid = Grid(shape=(8, 8, 8))
f = Function(name='f', grid=grid)
f.data[:] = 1
f.data[:, 5, :] = f.data[:, 1, :] + f.data[:, 7, :]
```
Currently this does not work since `f.data[:, 1, :]` and `f.data[:, 7, :]` have their data stored on different ranks and hence the addition on the RHS is 'not well defined'.
The above operation would currently have to be carried out via:
```
f.data[:, 5, :] = f.data[:, 1, :]
f.data[:, 5, :] += f.data[:, 7, :]
```
For convenience, we may want `f.data[:, 5, :] = f.data[:, 1, :] + f.data[:, 7, :]` to work directly for distributed arrays. Since `f.data[:, 1, :] + f.data[:, 7, :]` will require data to be transferred across ranks prior to `__setitem__` being called we could do this through overriding `__add__`.

PR #1947, opened 2022-06-21
mpi: Fix data distribution bugs [part 1]

rhodrin:
This set of PRs will supersede #1888.

Later PRs will deal specifically with issues #1010 and #1862.

This PR fixes some MPI data slicing issues. See the newly added [tests](https://github.com/devitocodes/devito/blob/62a9574e6bfb0d65fa817c696b3f8292e40c2b99/tests/test_data_parallel.py#L8) for details.

Note that these tests are located in a new `test_data_parallel.py` file to avoid `test_data.py` becoming too unwieldy. We may wish to refactor these two files in a little in this or an upcoming PR?


devito/data/data.py
```diff
@@ -247,7 +247,12 @@ def __getitem__(self, glb_idx, comm_type, gather_rank=None):
                 else:
                     pass
                 it.iternext()
-            return retval
+            reshape = tuple([s for s, i in zip(retval.shape, loc_idx)
+                             if type(i) is not np.int64])
+            if reshape and 0 not in reshape and reshape != retval.shape:
```
mloubout, line 252: maybe some parenthesis for readability?

devito/data/utils.py
```diff
@@ -389,7 +389,7 @@ def flip_idx(idx, decomposition):
                 stop = i.stop
             processed.append(slice(start, stop, i.step))
         else:
-            processed.append(i)
+            processed.append(slice(i, i+1, 1))
```
mloubout, line 392: `1` or `None`?

tests/test_data_parallel.py
```diff
@@ -0,0 +1,58 @@
+import pytest
+import numpy as np
+
+from devito import Grid, Function
+from devito.data import loc_data_idx
+
+
+class TestDataParallel(object):
```
mloubout, line 8: Can maybe be moved in `test_mpi`?

mloubout (commented):
Sweeet thanks @rhodrin for finding the root of it

devito/data/data.py
```diff
@@ -247,7 +247,12 @@ def __getitem__(self, glb_idx, comm_type, gather_rank=None):
                 else:
                     pass
                 it.iternext()
-            return retval
+            reshape = tuple([s for s, i in zip(retval.shape, loc_idx)
```
FabioLuporini, line 250: I would add a comment about what's being done here

FabioLuporini:
IMHO we either move the test inside here: https://github.com/devitocodes/devito/blob/master/tests/test_data.py#L475
or move the class inside test_data_parallel

other than that, uncontroversial (perhaps add a couple of comments to the new code as per my comment above)

devito/data/data.py
rhodrin: Ta, added those.

devito/data/utils.py
rhodrin: Doesn't matter I don't think? `1` is just more 'explicit'?

tests/test_data_parallel.py
rhodrin: I've moved them to `TestDistributedData` in `test_data.py` now.

rhodrin:
> Not sure I like a whole new file for just one single test

Yeah, move back to `test_data.py` for now. (Although note that the reason a new file was added for this PR was that many more tests will be added in the next couple of PRs).

tests/test_data.py
```diff
@@ -1273,6 +1273,56 @@ def test_misc_data(self):
             assert(np.all(c.data[-2:-1] == [3., 3., 5., 8., 7.]))
             assert(np.all(c.data[-1] == [4., 4., 6., 9., 8.]))
 
+    @pytest.mark.parallel(mode=4)
```
FabioLuporini, line 1276: perhaps `mode=[4, 8]` is it worth it or overkill?
or perhaps `mode=[4, 7]` 😬

Issue #1892, opened 2022-03-29
MPI gather for TimeFunction


jkwashbourne:
MPI `data_gather` broken for `TimeFunction`. MFE below

```
# env DEVITO_LOGGING=DEBUG DEVITO_MPI=1 mpirun -n 2 python mfe_gather2.py

import numpy as np
from devito import Grid, TimeFunction, Eq, Operator, MPI

grid = Grid(extent=(1,1,1), shape=(11,11,11), origin=(0.0,0.0,0.0), dtype=np.float32)

ft = TimeFunction(name='ft', grid=grid, time_order=1, space_order=2)

op = Operator([Eq(ft.forward,ft+1)], name="Op")
op.apply(time_m=1, time_M=10)

# works
ft_gather1 = ft.data[1]._gather(rank=0)

print("ft.data.shape;         ", ft.data.shape)
print("ft.data[1].shape;      ", ft.data[1].shape)
print("ft_gather1.data.shape; ", ft_gather1.data.shape)

# does not work ... error about dimensionality
ft_gather2 = ft.data_gather(rank=0)
```

jkwashbourne:
additional comment: isnt a very high performance version of this gather method absolutely **essential** for practical use of MPI at scale?

rhodrin:
I should note that this is now fixed in [fix_mpi_slicing_p2](https://github.com/devitocodes/devito/pull/1949). PR needs some tidying but is mostly ready.

georgebisbas:
Fixed by https://github.com/devitocodes/devito/pull/1949

PR #1949, opened 2022-06-22
mpi: Fix data distribution bugs [part 2]

rhodrin:
Fix for issues #1862 and  #1892 + associated tests.

### DIFF
--- a/devito/data/data.py
+++ b/devito/data/data.py
@@ -167,7 +167,11 @@ class Data(np.ndarray):
         @wraps(func)
         def wrapper(data, *args, **kwargs):
             glb_idx = args[0]
-            if len(args) > 1 and isinstance(args[1], Data) \
+            # NOTE: Consolidate with that appearing in __getitem__
+            is_gather = True if (kwargs and isinstance(kwargs['gather_rank'], int)) else False
+            if is_gather and all(i == slice(None, None, 1) for i in glb_idx):
+                comm_type = gather
+            elif len(args) > 1 and isinstance(args[1], Data) \
                     and args[1]._is_mpi_distributed:
                 comm_type = index_by_index
             elif data._is_mpi_distributed:
@@ -197,7 +201,44 @@ class Data(np.ndarray):
     def __getitem__(self, glb_idx, comm_type, gather_rank=None):
         loc_idx = self._index_glb_to_loc(glb_idx)
         is_gather = True if isinstance(gather_rank, int) else False
-        if comm_type is index_by_index or is_gather:
+        if is_gather and comm_type == gather:
+            comm = self._distributor.comm
+            size = comm.Get_size()
+            rank = comm.Get_rank()
+
+            sendbuf = np.array(self[:].flatten())
+            sendcounts = np.array(comm.gather(len(sendbuf), gather_rank))
+
+            if rank == gather_rank:
+                recvbuf = np.zeros(sum(sendcounts), dtype=self.dtype.type)
+            else:
+                recvbuf = None
+            comm.Gatherv(sendbuf=sendbuf, recvbuf=(recvbuf, sendcounts), root=gather_rank)
+
+            # Reshape the gathered data to produce the output
+            # NOTE: Make cleverer so that new array doesn't need to be allocated
+            if rank == gather_rank:
+                if len(self.shape) > len(self._distributor.glb_shape):
+                    glb_shape = list(self._distributor.glb_shape)
+                    for i in range(len(self.shape) - len(self._distributor.glb_shape)):
+                        glb_shape.insert(i, self.shape[i])
+                retval = np.zeros(glb_shape, dtype=self.dtype.type)
+                start, stop, step = 0, 0, 1
+                for i, s in enumerate(sendcounts):
+                    if i > 0:
+                        start += sendcounts[i-1]
+                    stop += sendcounts[i]
+                    data_slice = recvbuf[slice(start, stop, step)]
+                    shape = [r.stop-r.start for r in self._distributor.all_ranges[i]]
+                    idx = [slice(r.start, r.stop, r.step) for r in self._distributor.all_ranges[i]]
+                    for i in range(len(self.shape) - len(self._distributor.glb_shape)):
+                        shape.insert(i, glb_shape[i])
+                        idx.insert(i, slice(0, glb_shape[i]+1, 1))
+                    retval[tuple(idx)] = data_slice.reshape(tuple(shape))
+                return retval
+            else:
+                return None
+        elif comm_type is index_by_index or is_gather:
             # Retrieve the pertinent local data prior to mpi send/receive operations
             data_idx = loc_data_idx(loc_idx)
             self._index_stash = flip_idx(glb_idx, self._decomposition)
@@ -545,4 +586,5 @@ class Data(np.ndarray):
 class CommType(Tag):
     pass
 index_by_index = CommType('index_by_index')  # noqa
-serial = CommType('serial')
+serial = CommType('serial')  # noqa
+gather = CommType('gather')  # noqa
--- a/devito/data/data.py
+++ b/devito/data/data.py
@@ -89,13 +89,6 @@ class Data(np.ndarray):
         warning("Pickling of `Data` objects is not supported. Casting to `numpy.ndarray`")
         return np.array(self).__reduce__()
 
-    def reshape_data(self, shape):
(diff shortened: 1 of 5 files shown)


devito/data/data.py
```diff
@@ -155,12 +155,24 @@ def _global(self, glb_idx, decomposition):
         ret._is_distributed = any(i is not None for i in decomposition)
         return ret
 
+    def _prune_shape(self, shape):
```
georgebisbas, line 158: _reduce_eq_shape?

devito/data/data.py
```diff
@@ -155,12 +155,24 @@ def _global(self, glb_idx, decomposition):
...
+        # Reduce distributed MPI `Data`'s shape to that of an equivalently
+        # sliced numpy array.
+        decomposition = tuple([d for d in self._decomposition if d.size > 1])
```
FabioLuporini, line 161: nitpicking

```
tuple(d for d in ...)
```

is fine -- you don't need the inner list comprehension

```
tuple([...])
```

devito/data/data.py
```diff
@@ -155,12 +155,24 @@ def _global(self, glb_idx, decomposition):
...
+        retval._decomposition = decomposition
+        return retval
+
     def _check_idx(func):
         """Check if __getitem__/__setitem__ may require communication across MPI ranks."""
         @wraps(func)
         def wrapper(data, *args, **kwargs):
             glb_idx = args[0]
-            if len(args) > 1 and isinstance(args[1], Data) \
+            is_gather = True if (kwargs and isinstance(kwargs['gather_rank'], int)) \
```
FabioLuporini, line 171:
```
is_gather = kwargs and isinstance(...)
```

devito/data/data.py
```diff
@@ -190,7 +202,45 @@ def __str__(self):
     def __getitem__(self, glb_idx, comm_type, gather_rank=None):
         loc_idx = self._index_glb_to_loc(glb_idx)
         is_gather = True if isinstance(gather_rank, int) else False
```
FabioLuporini, line 204:
```
is_gather = isinstance(...)
```

devito/data/utils.py
```diff
@@ -240,6 +240,16 @@ def mpi_index_maps(loc_idx, shape, topology, coords, comm):
         dat_len[coords[j]] = comm.bcast(shape, root=j)
         if any(k == 0 for k in dat_len[coords[j]]):
             dat_len[coords[j]] = as_tuple([0]*len(dat_len[coords[j]]))
+    if len(np.amax(dat_len)) > len(topology):
```
FabioLuporini, line 243: Leave a blank line, add a comment about what's going on here?

devito/data/data.py
```diff
@@ -249,10 +300,13 @@ def __getitem__(self, glb_idx, comm_type, gather_rank=None):
                 it.iternext()
             # Check if dimensions of the view should now be reduced to
             # be consistent with those of an equivalent NumPy serial view
-            reshape = tuple([s for s, i in zip(retval.shape, loc_idx)
-                             if type(i) is not np.int64])
-            if reshape and (0 not in reshape) and (reshape != retval.shape):
-                return retval.reshape(reshape)
+            if not is_gather:
+                newshape = tuple([s for s, i in zip(retval.shape, loc_idx)
```
FabioLuporini, line 304: as before, you may drop the `[ ]`

devito/data/data.py
```diff
@@ -190,7 +202,45 @@ def __str__(self):
     def __getitem__(self, glb_idx, comm_type, gather_rank=None):
         loc_idx = self._index_glb_to_loc(glb_idx)
         is_gather = True if isinstance(gather_rank, int) else False
-        if comm_type is index_by_index or is_gather:
+        if is_gather and comm_type == gather:
```
FabioLuporini, line 205:
```
comm_type is gather
```

devito/data/data.py
```diff
@@ -155,12 +155,24 @@ def _global(self, glb_idx, decomposition):
         ret._is_distributed = any(i is not None for i in decomposition)
         return ret
 
+    def _prune_shape(self, shape):
+        # Reduce distributed MPI `Data`'s shape to that of an equivalently
+        # sliced numpy array.
+        decomposition = tuple([d for d in self._decomposition if d.size > 1])
+        retval = self.reshape(shape)
+        retval._decomposition = decomposition
```
FabioLuporini, line 163: `reshape` creates a new `Data` I presume, so why do we have to reset the `_decomposition` here? I would expect this to happen in the constructor or at least `__array_finalize__`, so why it's not the case?

devito/data/data.py
FabioLuporini, line 205: Refresh my mind please -- what does the user need to write so that we end up here?

FabioLuporini (commented):
Great! Only minor comments left.

devito/data/data.py
rhodrin: It's not an equation, just a `Data` object, but I'm open to alternative names.

devito/data/data.py
rhodrin: For MPI gathers, the decorator will infer how to gather based on the slice (or lack of) supplied by the user. (That is, `gather` vs `index_by_index` is set by the decorator, not by the user).

tests/test_data.py
```diff
@@ -1323,6 +1323,52 @@ def test_inversions(self, gslice):
...
+        dat1 = np.arange(64, dtype=np.int32).reshape(grid1.shape)
+
+        f1.data[:] = dat1
+        f2.data[:] = f1.data[:]
+        assert np.all(f2.data[:] == f1.data[:])
+
+        g.data[0, :, :] = dat1
+        f1.data[:] = g.data[0, ::-1, ::-1]
+        result = np.array(f1.data[:])
+        if LEFT in glb_pos_map[x] and LEFT in glb_pos_map[y]:
```
georgebisbas, line 1351: These lines can be done parametrically as well, but its pretty clear this way, not sure if it is gonna be better parametrically

devito/data/data.py
mloubout: Doesn't numpy has a no copy reshape with `view`?

devito/data/data.py
```diff
@@ -155,12 +155,23 @@ def _global(self, glb_idx, decomposition):
...
+        return retval
+
     def _check_idx(func):
         """Check if __getitem__/__setitem__ may require communication across MPI ranks."""
         @wraps(func)
         def wrapper(data, *args, **kwargs):
             glb_idx = args[0]
-            if len(args) > 1 and isinstance(args[1], Data) \
+            is_gather = kwargs and isinstance(kwargs['gather_rank'], int)
```
mloubout, line 171: guarantied to have `gather_rank`? maybe `is_gather = isinstance(kwargs.get('gather_rank', None) int)`

devito/data/data.py
```diff
@@ -189,8 +200,46 @@ def __str__(self):
     @_check_idx
     def __getitem__(self, glb_idx, comm_type, gather_rank=None):
         loc_idx = self._index_glb_to_loc(glb_idx)
-        is_gather = True if isinstance(gather_rank, int) else False
-        if comm_type is index_by_index or is_gather:
+        is_gather = isinstance(gather_rank, int)
+        if is_gather and comm_type is gather:
```
mloubout, line 204: why not just `if gather_rank and ....`

devito/data/data.py
```diff
@@ -189,8 +200,46 @@ def __str__(self):
...
+            comm = self._distributor.comm
+            rank = comm.Get_rank()
+
+            sendbuf = np.array(self[:].flatten())
```
mloubout, line 208: `flattent` makes a copy no? wouldn something like `self.flat` work here (1D view of ndarray without copy)

devito/data/data.py
rhodrin: If able, the method returns a `view`:

> This will be a new view object if possible; otherwise, it will be a copy. Note there is no guarantee of the memory layout (C- or Fortran- contiguous) of the returned array.

devito/data/data.py
rhodrin: Yes, indeed. Changed this to `self.flat[:]`.

devito/data/data.py
rhodrin: `0` is a valid rank so I think we need this?

FabioLuporini (approved):

georgebisbas (approved):

mloubout (approved):

PR #1971, opened 2022-07-23
compiler: Infer type generation from _C_ctype

FabioLuporini:
CodeSymbol subclasses were required to override a subset of `_C_typedata`, `_C_typename`, `_C_typedecl`, `_C_ctype`. This was messy, difficult to understand, challenging to maintain, and also a bit hacky.

Now CodeSymbol subclasses are only requested to specialize `_C_ctype`.  All other information are derived from `_C_ctype` through suitable utility functions (e.g., a `str` representation of the C type for code generation will be generated via the utility `ctypes_to_cstr`).

This is in my opinion a remarkable improvement to the codegen infrastracture, that was long due.

The `ArrayMapped` type is also introduced.


devito/types/array.py
```diff
@@ -215,6 +215,19 @@ def _make_pointer(self, dim):
...
 
 
+class ArrayMapped(Array):
+
+    _C_structname = 'array'
+    _C_field_data = 'data'
+    _C_field_nbytes = 'nbytes'
+    _C_field_dmap = 'dmap'
+
+    _C_ctype = POINTER(type(_C_structname, (Structure,),
```
FabioLuporini, line 225: you see? a ney symbolic type, `ArrayMapped`, but the only thing that needs to be defined is the `_C_ctype`. This is enough to create a `ctype`, a `str` representing its C type, etc etc.

tests/test_operator.py
```diff
@@ -1269,120 +1269,6 @@ def test_loose_kwargs(self):
 @skipif('device')
 class TestDeclarator(object):
 
-    def test_heap_1D(self):
```
FabioLuporini, line 1272: all these tests were obsolete, a massive PITA to maintain, and basically unnecessary. Hence, dropping

devito/tools/utils.py
```diff
@@ -257,16 +265,54 @@ def ctypes_to_cstr(ctype, toarray=None):
...
+    # Add qualifiers, if any
+    if qualifiers:
+        retval = '%s %s' % (' '.join(qualifiers), retval)
+
+    return retval
+
+
+def ctypes_to_cgen(ctype, fields=None):
+    """
+    Create a cgen.Structure off a ctypes.Struct, or (possibly nested) ctypes.Pointer
```
georgebisbas, line 281: gap after period

georgebisbas (commented):
Overall not any comment to make.

georgebisbas (approved):

PR #1976, opened 2022-07-26
misc: Add FAQ page (lifted from the wiki)

FabioLuporini:
if we do this, we should also update the FAQ link in the wiki (and drop the page there)


georgebisbas (commented):
If you look at the rich diff some of them do not link to the place where the question is answered.
e.g.: Do Devito Operators release the GIL when executing C code?

FabioLuporini:
> If you look at the rich diff some of them do not link to the place where the question is answered.
> e.g.: Do Devito Operators release the GIL when executing C code?

same in the wiki?

georgebisbas:
> > If you look at the rich diff some of them do not link to the place where the question is answered.
> > e.g.: Do Devito Operators release the GIL when executing C code?
> 
> same in the wiki?

As I see, yes.

mloubout (approved):

PR #1979, opened 2022-07-29
compiler: Patch premature lowering of EvalDerivative

### LINKED ISSUE #1978: Premature lowering of EvalDerivative
Reproducer:

```
        grid = Grid(shape=(10, 10))

        u = TimeFunction(name='u', grid=grid, space_order=4)

        # Not really a custom derivative, but the enforced pre-evaluation makes
        # it behaves as if it were one
        mock_custom_deriv = u.dx.dy.evaluate

        # This symbolic operation -- creating an Add between an arbitray object
        # and an EvalDerivative -- caused the EvalDerivative to be prematurely
        # simplified being flatten into an Add
        expr0 = u.dt - mock_custom_deriv
        expr = -expr0

        eq = Eq(u.forward, expr)

        op = Operator(eq)
```

flop count blows up here because `EvalDerivatives` get lost during the early symbolic manipulations

FabioLuporini:
fixes #1978

### DIFF
--- a/devito/finite_differences/differentiable.py
+++ b/devito/finite_differences/differentiable.py
@@ -423,7 +423,7 @@ class Mul(DifferentiableOp, sympy.Mul):
         # a*1 -> a
         args = [i for i in args if i != 1]
 
-        # a*-1 -> a
+        # a*-1 -> a*-1
         # a*-1*-1 -> a
         # a*-1*-1*-1 -> a*-1
         nminus = len([i for i in args if i == sympy.S.NegativeOne])
@@ -437,10 +437,10 @@ class Mul(DifferentiableOp, sympy.Mul):
         # `sympy.Mul.flatten(coeff, Add)` flattens out nested Adds within Add,
         # which would destroy `EvalDerivative`s if present. So here we perform
         # a similar thing, but cautiously construct an evaluated Add, which
-        # will preserve the integrity of `EvalDerivative`s, if any.
+        # will preserve the integrity of `EvalDerivative`s, if any
         try:
             a, b = args
-            if not a.is_zero and a.is_Rational:
+            if a.is_Rational:
                 r, b = b.as_coeff_Mul()
                 if r is sympy.S.One and type(b) is Add:
                     return Add(*[_keep_coeff(a, bi) for bi in b.args], evaluate=False)
--- a/devito/finite_differences/differentiable.py
+++ b/devito/finite_differences/differentiable.py
@@ -5,7 +5,7 @@ from functools import singledispatch
 import numpy as np
 import sympy
 from sympy.core.add import _addsort
-from sympy.core.mul import _mulsort
+from sympy.core.mul import _keep_coeff, _mulsort
 from sympy.core.decorators import call_highest_priority
 from sympy.core.evalf import evalf_table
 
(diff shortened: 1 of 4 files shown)


georgebisbas (approved):

devito/finite_differences/differentiable.py
```diff
@@ -423,14 +423,30 @@ def __new__(cls, *args, **kwargs):
         # a*1 -> a
         args = [i for i in args if i != 1]
 
+        # a*-1 -> a
```
mloubout, line 426: -a ?

devito/finite_differences/differentiable.py
```diff
@@ -423,14 +423,30 @@ def __new__(cls, *args, **kwargs):
...
         # a*-1*-1 -> a
+        # a*-1*-1*-1 -> a*-1
```
mloubout, line 428: -a ?

devito/finite_differences/differentiable.py
```diff
@@ -423,14 +423,30 @@ def __new__(cls, *args, **kwargs):
...
         # Reorder for homogeneity with pure SymPy types
         _mulsort(args)
 
+        # `sympy.Mul.flatten(coeff, Add)` flattens out nested Adds within Add,
+        # which would destroy `EvalDerivative`s if present. So here we perform
+        # a similar thing, but cautiously construct an evaluated Add, which
+        # will preserve the integrity of `EvalDerivative`s, if any.
+        try:
+            a, b = args
+            if not a.is_zero and a.is_Rational:
```
mloubout, line 443: `a.is_zero` is un-necessary would have been cought line 420 already.

devito/finite_differences/differentiable.py
```diff
@@ -423,14 +423,30 @@ def __new__(cls, *args, **kwargs):
...
+                r, b = b.as_coeff_Mul()
+                if r is sympy.S.One and type(b) is Add:
```
mloubout, line 445: Can this be generalized to like `r is sympy.Number` or `r.is_Symbol` and merge it with `a`?

mloubout (commented):
Some comments but nice catch and yest these `1.0` where a pain

devito/finite_differences/differentiable.py
FabioLuporini: for the record, I lifted it from sympy

https://github.com/sympy/sympy/blob/master/sympy/core/mul.py#L285

tho admittedly a slightly older version than the one I have installed locally

I will try to improve it

devito/finite_differences/differentiable.py
FabioLuporini: the arguments are `a` and `-1` (unless `a` it's an integer) that's why it's not factored out

devito/finite_differences/differentiable.py
FabioLuporini: good catch. will put `a*-1` (see below)

FabioLuporini:
pushed edits

FabioLuporini:
note latest pep8-related commit necessary due to https://github.com/devitocodes/devito/runs/7605926638?check_suite_focus=true

EdCaunt (approved):

mloubout (approved):

PR #1980, opened 2022-07-29
compiler: Generalize lowering of reductions

FabioLuporini:


devito/core/cpu.py
```diff
@@ -76,6 +76,12 @@ class Cpu64OperatorMixin(object):
     than this threshold.
     """
 
+    MAPIFY_REDUCE = False
+    """
+    Vector-expand all scalar reductions to turn them into explicit map-reductions,
```
georgebisbas, line 81: cool

devito/ir/clusters/algorithms.py
```diff
@@ -368,19 +369,46 @@ def pull_indexeds(expr, subs, mapper, parent=None):
...
+            v = Symbol(name=name, dtype=e.dtype)
             processed.extend([e.func(v, e.rhs, operation=None),
                               e.func(e.lhs, v)])
+
+        elif e.is_Reduction and e.lhs.is_Symbol and opt_mapify_reduce:
+            # Transform `e` into what is in essence an explicit map-reduce
+            # For example, turn:
+            # `s += f(u[x], v[x], ...)`
+            # into
+            # `r[x] = f(u[x], v[x], ...)`
```
mloubout, line 403: Is this a block size temp or full size temp?

mloubout: Can this be turned off with `min-storage` ?

FabioLuporini: full size temporary

> Can this be turned off with `min-storage`

You're overthinking it probably. This is by default disabled because `opt_mapify_reduce` is set to False in all backends

FabioLuporini: turning it into into a block-size temporary would require quite a bit of work, but I agree in principle it should be like you say

mloubout (approved):

georgebisbas (commented):
Need to try whether this can have any impact on: https://github.com/devitocodes/devito/issues/1528

georgebisbas:
> Need to try whether this can have any impact on: #1528

no difference

georgebisbas (approved):

PR #1982, opened 2022-08-01
docker: Fix docker tags for publish

mloubout:
Tested and working as can bee seen for dev:

https://github.com/mloubout/CIDebug/actions/runs/2776343404

and release (latest fixed in this PR as well):
https://github.com/mloubout/CIDebug/runs/7615747425?check_suite_focus=true

### DIFF
--- a/.github/workflows/docker-devito.yml
+++ b/.github/workflows/docker-devito.yml
@@ -82,21 +82,21 @@ jobs:
           # `enable` is a "if" filter fir that tag
           # `type` is the type of tag
           tags: |
-            type=ref,event=pr,value=${{ matrix.tag }}-dev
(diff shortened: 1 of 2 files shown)


FabioLuporini (commented):
Uncontroversial

Random thought: I appreciate it's just "tags"... but why don't we just remove the old tags and write a message on Slack that they've been replaced with something else?  That would be less code to maintain

FabioLuporini (approved):

mloubout:
For the old tags, I agree it would be less to maintain, just thought would be good to keep them for a little bit before dropping them. Honestly it's setup so shouldn't really need to touch it so I'm fine either way

PR #1988, opened 2022-08-17
dsl: Patched cross-derivative fd_order bug

EdCaunt:
Fix for manually specifying the derivative order of cross-derivatives via the call API. Makes `f.dxdy(fd_order={x: 2, y: 2})` behave as expected rather than raising `TypeError`.

### DIFF
--- a/tests/test_derivatives.py
+++ b/tests/test_derivatives.py
@@ -689,6 +689,18 @@ class TestTwoStageEvaluation(object):
         assert Add(*term0.expand().args) == term1.expand()
 
 
+class TestCallAPI(object):
+    """Tests for the f.dx(x0=..., fd_order=...) API"""
+    def test_xderiv_order(self):
+        grid = Grid(shape=(11, 11), extent=(10., 10.))
+        x, y = grid.dimensions
+        f = Function(name='f', grid=grid, space_order=4)
+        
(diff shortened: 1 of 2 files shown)


tests/test_derivatives.py
```diff
@@ -689,6 +689,18 @@ def test_dxdy(self):
         assert Add(*term0.expand().args) == term1.expand()
 
 
+class TestCallAPI(object):
```
mloubout, line 692: I don't mind that but I think we have some test here and there from previous bugs (like `x0=...`) that should belong there as well then.

Blank line between class def and first docstring

EdCaunt: Which tests did you have in mind?

mloubout: `test_fd_new_order`, `test_fd_new_side`, `test_fd_new_x0`, `test_fd_new_lo` there is quite a few so I would probably just move this with those in the normal `TestFD`. This actually probably belongs in `test_fd_new_order` as a `parametrized` case

EdCaunt: Moved it up to be with the other tests, and added a similar test for x0 with x-derivs

mloubout (approved):

FabioLuporini (approved):
LGTM

FabioLuporini:
Thanks, merged

PR #1989, opened 2022-08-18
api: add dimension-wise summing builtin and tests

mloubout:
- Add missing tests for `sumall`
- Add `sum(f, dims=...)`. The dimension input must be a dimension not a number

### DIFF
--- a/devito/builtins/arithmetic.py
+++ b/devito/builtins/arithmetic.py
@@ -4,7 +4,7 @@ import devito as dv
 from devito.builtins.utils import MPIReduction
 
 
-__all__ = ['norm', 'sumall', 'inner', 'mmin', 'mmax']
+__all__ = ['norm', 'sumall', 'sum', 'inner', 'mmin', 'mmax']
 
 accumulator_mapper = {
     # Integer accumulates on Float64
@@ -56,6 +56,42 @@ def norm(f, order=2):
     v = np.power(mr.v, 1/order)
 
     return f.dtype(v)
+
+
+@dv.switchconfig(log_level='ERROR')
+def sum(f, dims=None):
+    """
+    Compute the sum of the Function data over specified dimensions.
+    Defaults to sum over all dmensions
+
+    Parameters
+    ----------
+    f : Function
+        Input Function.
+    dims : Dimension or tuple
+        Dimensions to sum over
+    """
+    if dims is None or dims == f.dimensions:
+        return sumall(f)
+
+    dims = dv.tools.as_tuple(dims)
+    new_dims = tuple(d for d in f.dimensions if d not in dims)
+    shape = tuple(f._size_domain[d] for d in new_dims)
+    if f.is_TimeFunction and f.time_dim not in dims:
+        out = dv.TimeFunction(name="%ssum" % f.name, grid=f.grid,
+                              space_order=f.space_order, shape=shape,
+                              dimensions=new_dims, time_dim=f.time_dim,
+                              time_order=f.time_order)
+    else:
+        out = dv.Function(name="%ssum" % f.name, grid=f.grid,
+                          space_order=f.space_order, shape=shape,
+                          dimensions=new_dims)
+    kw = {}
+    if f.is_TimeFunction:
+        kw = {'time_m': 0, 'time_M': f.time_order}
+    op = dv.Operator(dv.Eq(out, out + f))
+    op(**kw)
+    return out
 
 
 @dv.switchconfig(log_level='ERROR')
--- a/tests/test_builtins.py
+++ b/tests/test_builtins.py
@@ -6,7 +6,7 @@ from scipy import misc
 from conftest import skipif
 from devito import Grid, Function, TimeFunction, switchconfig
 from devito.builtins import (assign, norm, gaussian_smooth, initialize_function,
-                             inner, mmin, mmax)
+                             inner, mmin, mmax, sum, sumall)
 from devito.data import LEFT, RIGHT
 from devito.tools import as_tuple
 from devito.types import SubDomain, SparseTimeFunction
@@ -345,12 +345,40 @@ class TestBuiltinsResult(object):
         grid = Grid((101, 101), extent=(1000., 1000.))
 
         nrec = 101
-        rec0 = SparseTimeFunction(name='rec0', grid=grid, nt=1001, npoint=nrec)
+        rec0 = TimeFunction(name='rec0', grid=grid, nt=1001, npoint=nrec)
 
         rec0.data[:, :] = 1 + np.random.rand(*rec0.shape).astype(grid.dtype)
         term1 = np.linalg.norm(rec0.data)
         term2 = norm(rec0)
         assert np.isclose(term1/term2 - 1, 0.0, rtol=0.0, atol=1e-5)
+
+    def test_sum(self):
+        """
+        Test that sum produces the correct result against numpy
+        """
+        grid = Grid((11, 11))
+        x, y = grid.dimensions
+        t = grid.stepping_dim
+        f = TimeFunction(name='f', grid=grid)
+        f.data.fill(1)
+        assert sum(f) == sumall(f)
+        assert sum(f, dims=None) == sumall(f)
+        assert sum(f, dims=f.dimensions) == sumall(f)
+        assert sumall(f) == np.prod(f.shape)
+
+        fsum = sum(f, dims=x)
+        assert fsum.dimensions == (t, y)
+        assert np.all(fsum.data == 11)
+
+        fsum = sum(f, dims=t)
+        assert not fsum.is_TimeFunction
+        assert fsum.dimensions == (x, y)
+        assert np.all(fsum.data == 2)
+
+        fsum = sum(fsum, dims=x)
+        assert not fsum.is_TimeFunction
+        assert fsum.dimensions == (y,)
+        assert np.all(fsum.data == 2 * 11)
 
     def test_min_max_sparse(self):
         """


devito/builtins/arithmetic.py
```diff
@@ -58,6 +58,42 @@ def norm(f, order=2):
...
+    """
+    Compute the sum of the Function data over specified dimensions.
+    Defaults to sum over all dmensions
+
+    Parameters
+    ----------
+    f : Function
+        Input Function.
+    dims : Dimension or tuple
+        Dimensions to sum over
```
FabioLuporini, line 72: full stop

devito/builtins/arithmetic.py
```diff
@@ -58,6 +58,42 @@ def norm(f, order=2):
...
+def sum(f, dims=None):
+    """
+    Compute the sum of the Function data over specified dimensions.
+    Defaults to sum over all dmensions
+
+    Parameters
+    ----------
+    f : Function
+        Input Function.
+    dims : Dimension or tuple
```
FabioLuporini, line 71: or tuple of Dimensions

devito/builtins/arithmetic.py
```diff
@@ -58,6 +58,42 @@ def norm(f, order=2):
...
+        Dimensions to sum over
+    """
+    if dims is None or dims == f.dimensions:
+        return sumall(f)
+
+    dims = dv.tools.as_tuple(dims)
+    new_dims = tuple(d for d in f.dimensions if d not in dims)
+    shape = tuple(f._size_domain[d] for d in new_dims)
+    if f.is_TimeFunction and f.time_dim not in dims:
+        out = dv.TimeFunction(name="%ssum" % f.name, grid=f.grid,
```
FabioLuporini, line 81: I've recently implemented `_rebuild` see [here](https://github.com/devitocodes/devito/blob/master/devito/passes/iet/asynchrony.py#L130) for example

so you should be able to just do: 
```
f._rebuild(name=..., shape=shape, dimensions=new_dims)
```

examples/seismic/tutorials/07_DRP_schemes.ipynb
```diff
@@ -415,6 +415,7 @@
     }
    ],
    "source": [
+    "# NBVAL_IGNORE_OUTPUT\n",
```
FabioLuporini, line 418: why are these failing now all of a sudden?

FabioLuporini (commented):
Before you ask: I'm sick in bed, so why not reviewing a PR 😬

examples/seismic/tutorials/07_DRP_schemes.ipynb
EdCaunt: Matplotlib changed the default plot size iirc

mloubout: New version of matplotlib yes

devito/builtins/arithmetic.py
mloubout: In some cases yeah but not all. There is still corner cases (like sparse) I need to add and because the class can change depending on the dimension summed rebuild won't work. (E.g summing a SparseTime over the rec dim makes a TimeFunc not a rebuild)

navjotk (commented):
Looks good to me. Thank you for this.

devito/builtins/arithmetic.py
```diff
@@ -39,17 +39,12 @@ def norm(f, order=2):
     if f.is_TimeFunction and f._time_buffering:
         kwargs[f.time_dim.max_name] = f._time_size - 1
 
-    # Protect SparseFunctions from accessing duplicated (out-of-domain) data,
-    # otherwise we would eventually be summing more than expected
-    p, eqns = f.guard() if f.is_SparseFunction else (f, [])
```
mloubout, line 44: @FabioLuporini so I guess this is up for discussion but I think this is wrong. If someone wants to compute the norm/sum/inner/... of a sparse function, then all you comput is `sum_{t, p} rec[t, p]` so whether or not it lies within the grid is not relevant, The guard should only be used if the physical location `u[posz,...]`  in an interpolation/injection to make sure that index exists in the function.

mloubout: Oh wait no, MPI halo with duplicates right. reverting.

devito/operator/operator.py
```diff
@@ -495,7 +495,11 @@ def _prepare_arguments(self, autotune=None, **kwargs):
                 # E.g., SubFunctions
                 continue
             for k, v in p._arg_values(**kwargs).items():
-                if k in args and args[k] != v:
+                try:
+                    val_test = k in args and (args[k] != v).all()
+                except AttributeError:
```
georgebisbas, line 500: Maybe add a comment about this attribute error?

examples/seismic/tutorials/10_nmo_correction.ipynb
```diff
@@ -291,6 +291,7 @@
     }
    ],
    "source": [
+    "#NBVAL_IGNORE_OUTPUT\n",
```
georgebisbas, line 294: THere is an increasing number of nbval as we discussed.
What if we add some norm checking in some places?
Here we could check the norms of rec

devito/builtins/arithmetic.py
```diff
@@ -58,6 +57,58 @@ def norm(f, order=2):
     return f.dtype(v)
 
 
+@dv.switchconfig(log_level='ERROR')
+def sum(f, dims=None):
```
FabioLuporini, line 61: how about we call this `reduce` and add an optional argument `op='+'` ? no need to implement the other cases in this PR, OC

devito/builtins/arithmetic.py
```diff
@@ -58,6 +57,58 @@ def norm(f, order=2):
...
+    """
+    Compute the sum of the Function data over specified dimensions.
```
FabioLuporini, line 63: with the above, this should be changed accordingly

devito/builtins/arithmetic.py
FabioLuporini: sure, but why not using it where possible?

devito/operator/operator.py
```diff
@@ -495,7 +495,11 @@ def _prepare_arguments(self, autotune=None, **kwargs):
                 # E.g., SubFunctions
                 continue
             for k, v in p._arg_values(**kwargs).items():
-                if k in args and args[k] != v:
+                try:
```
FabioLuporini, line 498: wait... what's this one for?

devito/operator/operator.py
```diff
@@ -495,7 +495,11 @@ def _prepare_arguments(self, autotune=None, **kwargs):
...
+                    val_test = k in args and (args[k] != v).all()
```
FabioLuporini, line 499: the .all here is potentially expensive if a big numpy array? there's something weird going on here

devito/operator/operator.py
mloubout: The cordinates end up here in some sparse reduces case. Not sure exactly why.

devito/builtins/arithmetic.py
mloubout: Was erroring, some kwargs missing haven't had time to dig into it

devito/builtins/arithmetic.py
```diff
@@ -58,6 +57,58 @@ def norm(f, order=2):
...
+    f : Function
+        Input Function.
+    dims : Dimension or tuple of Dimension
+        Dimensions to sum over.
+    """
+    dims = dv.tools.as_tuple(dims)
+    if dims == () or dims == f.dimensions:
+        return sumall(f)
+
+    # Get dimensions and shape of the result.
```
georgebisbas, line 77: no period here

devito/builtins/arithmetic.py
```diff
@@ -58,6 +57,58 @@ def norm(f, order=2):
...
+    else:
+        out = dv.Function(name="%ssum" % f.name, grid=f.grid,
+                          space_order=f.space_order, shape=shape,
+                          dimensions=new_dims)
+
+    kwargs = {}
+    if f.is_TimeFunction and f._time_buffering:
+        kwargs[f.time_dim.max_name] = f._time_size - 1
+
+    # Only need one guard has they have the same coordiantes and dimensions
```
georgebisbas, line 105: has they have?

devito/types/dense.py
```diff
@@ -1104,7 +1107,8 @@ def __halo_setup__(self, **kwargs):
                 halo = (left_points, right_points)
             else:
                 raise TypeError("`space_order` must be int or 3-tuple of ints")
-            return tuple(halo if i.is_Space else (0, 0) for i in self.dimensions)
+            halo = tuple(halo if i.is_Space else (0, 0) for i in self.dimensions)
+            return DimensionTuple(*halo, getters=self.dimensions)
```
FabioLuporini, line 1111: you can probably factor 

```
return DimensionTuple(...)
```

out of the `if-else`. This way, you'll also catch the case at line 1100, which might still return a plain tuple otherwise

devito/types/dense.py
```diff
@@ -1128,13 +1132,20 @@ def __padding_setup__(self, **kwargs):
...
+            return DimensionTuple(*padding, getters=self.dimensions)
         elif isinstance(padding, int):
-            return tuple((0, padding) if d.is_Space else (0, 0) for d in self.dimensions)
+            padding = tuple((0, padding) if d.is_Space else (0, 0)
+                            for d in self.dimensions)
+            return DimensionTuple(*padding, getters=self.dimensions)
         elif isinstance(padding, tuple) and len(padding) == self.ndim:
-            return tuple((0, i) if isinstance(i, int) else i for i in padding)
+            padding = tuple((0, i) if isinstance(i, int) else i for i in padding)
+            return DimensionTuple(*padding, getters=self.dimensions)
```
FabioLuporini, line 1148: same story as above -- you can factor out all these `return DimensionTuple(..)`

devito/builtins/arithmetic.py
FabioLuporini: has -> as

also, "Dimensions" (to be picky)

devito/builtins/arithmetic.py
FabioLuporini: ping

devito/operator/operator.py
```diff
@@ -495,7 +495,8 @@ def _prepare_arguments(self, autotune=None, **kwargs):
                 # E.g., SubFunctions
                 continue
             for k, v in p._arg_values(**kwargs).items():
-                if k in args and args[k] != v:
+                val_test = k in args and args[k] != v
```
FabioLuporini, line 498: no need for `val_test` -- can remain as in master

FabioLuporini (commented):
some minor comments (mostly aesthetics), but otherwise GTG

sum -> reduce is IMHO a better way to go though (you might have missed that comment)

devito/builtins/arithmetic.py
mloubout: It's a bit of a pain cuz these reudction (+, -, ...) are not easy to carry around but would be a good extension at some point yes

examples/seismic/tutorials/10_nmo_correction.ipynb
mloubout: >THere is an increasing number of nbval as we discussed.

These are on figure cells. the test/asserts shoulds still all be fine as they are in their own (at least they should be)

devito/tools/data_structures.py
```diff
@@ -29,18 +30,33 @@ def __init__(self, **kwargs):
         self.__dict__.update(kwargs)
 
 
-class EnrichedTuple(tuple):
+class EnrichedTuple(Pickable, tuple):
```
FabioLuporini, line 33: this should be `Reconstructable` not `Pickable`

but even then

why is it needed at all? Can you write an MFE?

devito/tools/data_structures.py
```diff
@@ -29,18 +30,33 @@ def __init__(self, **kwargs):
...
 
     """
     A tuple with an arbitrary number of additional attributes.
     """
 
     def __new__(cls, *items, getters=None, **kwargs):
+        # Check kwargs e.g for pickled input
```
FabioLuporini, line 40: Assuming this stays (though I suspect the actual issue is somewhere else), the comments should be changed accordingly as pickling is another thing

pickling uses reconstructability 

but reconstructability is a more general concept

devito/builtins/arithmetic.py
FabioLuporini: we could have thrown a NotImplementedError for all but +, but okay can postpone to another PR, no big deal

devito/tools/data_structures.py
mloubout: The MFE is the new test. And consequently all halo/padding are reconstructed wrong

devito/tools/data_structures.py
mloubout: The error is not somewhere else. This is reconstruced as a pure tuple and because of the constructor `__new__(cls, *items, getters=None, **kwargs)` and the fact that it modifies `__dict__` it is reconstucted as `__new__(cls, (*items,))`. So it misses all the extra nad it also reconstuct with the wrong items because `(2,3)` is then reconstructed as `((2, 3),)`

devito/builtins/arithmetic.py
```diff
@@ -58,6 +57,57 @@ def norm(f, order=2):
     return f.dtype(v)
 
 
+@dv.switchconfig(log_level='ERROR')
+def sum(f, dims=None):
+    """
+    Compute the sum of the Function data over specified dimensions.
+    Defaults to sum over all dmensions
```
georgebisbas, line 64: typo dimensions

devito/builtins/arithmetic.py
```diff
@@ -58,6 +57,57 @@ def norm(f, order=2):
...
+    # Get dimensions and shape of the result
+    new_dims = tuple(d for d in f.dimensions if d not in dims)
+    shape = tuple(f._size_domain[d] for d in new_dims)
+    if f.is_TimeFunction and f.time_dim not in dims:
+        out = f._rebuild(name="%ssum" % f.name, shape=shape, dimensions=new_dims,
+                         initializer=np.empty(0))
+    elif f.is_SparseTimeFunction:
+        if f.time_dim in dims:
+            # Sum over time -> SparseFunction
+            coords = f.coordinates._rebuild(name="%ssum_coords" % f.name)
```
georgebisbas, line 86: I would rename this as new_coords to be consistent with new_dims above...or rename new_dims to dims

examples/seismic/tutorials/10_nmo_correction.ipynb
georgebisbas: OK, we good then

tests/test_builtins.py
```diff
@@ -352,6 +352,66 @@ def test_norm_sparse(self):
         term2 = norm(rec0)
         assert np.isclose(term1/term2 - 1, 0.0, rtol=0.0, atol=1e-5)
 
+    def test_sum(self):
+        """
+        Test that sum produces the correct result against numpy
```
georgebisbas, line 357: NumPy ?

tests/test_builtins.py
```diff
@@ -352,6 +352,66 @@ def test_norm_sparse(self):
...
+    def test_sum_sparse(self):
+        """
+        Test that sum produces the correct result against numpy
+        """
+        grid = Grid((11, 11))
+        nx = 101
+        nt = 1001
+        s = SparseTimeFunction(name='s', grid=grid, nt=nt, npoint=nx)
+        s.data.fill(1.)
+        s.coordinates.data[:] = np.random.rand(*s.coordinates.data.shape)
```
georgebisbas, line 392: I think we avoided so far to have randomness in tests (?)

tests/test_pickle.py
```diff
@@ -43,6 +44,19 @@ def test_dimension():
...
 
 
+def test_enrichedtuple():
+    # Dummy enriched tuple
+    tup = EnrichedTuple(11, 31, getters=('a', 'b'), left=[3, 4], right=[5, 6])
+
+    pkl_t = pickle.dumps(tup)
+    new_t = pickle.loads(pkl_t)
+
+    assert new_t == tup  # This only test the actual tuple
```
georgebisbas, line 54: typo: only tests

georgebisbas (commented):
Some small comments

tests/test_builtins.py
EdCaunt: Is there some way to set a seed on the rand so it generates a consistent array?

mloubout: We avoid *computation* on random values, this is just whatever value never changed.

FabioLuporini (approved):
LGTM now

georgebisbas (approved):

FabioLuporini:
Merged, thanks everybody

PR #1992, opened 2022-08-24
compiler: Fix MPI mode diag2 does not need a MPIRegion

mmohrhard:
`Diag2HaloExchangeBuilder` does not use or require the `MPIRegion` that the base class `Overlap2HaloExchangeBuilder` wants to create.

### DIFF
--- a/devito/mpi/routines.py
+++ b/devito/mpi/routines.py
@@ -797,6 +797,9 @@ class Diag2HaloExchangeBuilder(Overlap2HaloExchangeBuilder):
         compute()
     """
 
+    def _make_region(self, hs, key):
+        return
+
     def _make_compute(self, hs, key, *args):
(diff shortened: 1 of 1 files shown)


mloubout:
Thnaks! Would you have a small test that fails without that change?

mmohrhard:
> Thnaks! Would you have a small test that fails without that change?

Not a simple one or one that I can share. In our case it works around a bug that we are still trying to track down where the MPIRegion object fails with specific sub domain constructions.

However, unrelated to the issue this PR is fixing an inconsistency as the diag2 mode does not need the MPIRegion object.

FabioLuporini (approved):
As long as it passes CI, good to go

FabioLuporini:
Merged, thanks

PR #1997, opened 2022-09-06
api: fix Derivative dict substitution with self as key

### LINKED ISSUE #1995: Derivatives get duplicated when a derivative is substituted for expression containing derivatives
If an unevaluated derivative is substituted for some expression containing derivatives, any derivatives in the substituted expression are duplicated unless they are the original derivative. As the `Derivative` object gets duplicated, subsequent algebraic simplifications are missed during the lowering process as they are considered separate symbols and thus not canceled.

If this expression is then used in building a kernel, it will take much longer than usual to build as a result of the increased number of derivative objects.

MFE:
```
from devito import Grid, Function, Derivative

grid = Grid(shape=(11,), extent=(10.,))
f = Function(name='f', grid=grid, space_order=4)

# To make sure that derivative objects are the same
fdx = f.dx
fdx2 = f.dx2

expr = fdx + fdx2

assert expr.find(Derivative) == expr.subs({fdx: fdx + fdx2, fdx2: fdx2}).find(Derivative)
```

mloubout:
Fixes #1995

### DIFF
--- a/devito/finite_differences/derivative.py
+++ b/devito/finite_differences/derivative.py
@@ -211,8 +211,18 @@ class Derivative(sympy.Derivative, Differentiable):
         # Check if we are calling subs(self, old, new, **hint) in which case
         # return the standard substitution. Need to check `==` rather than `is`
         # because a new derivative could be created i.e `f.dx.subs(f.dx, y)`
-        if len(args) == 2 and args[0] == self:
+        if len(args) == 2 and args[0] is self:
             return args[1]
+        # Check if we are calling subs(self, {old: new}, **hint) in which case
+        # return the standard substitution.
+        elif len(args) == 1 and self in args[0]:
+            new = args[0].pop(self)
+            if len(kwargs) == 0 and len(args) == 0:
+                return new
+            try:
+                return new.subs(args[0], **kwargs)
+            except AttributeError:
+                return new
         try:
             rules = dict(*args)
         except TypeError:
--- a/tests/test_derivatives.py
+++ b/tests/test_derivatives.py
@@ -535,6 +535,18 @@ class TestFD(object):
         expr3 = expr.subs({f.dx: f.dx2, 'x0': 2})
         assert simplify(expr3 - (f.dx2(x0=2) + f + 1)) == 0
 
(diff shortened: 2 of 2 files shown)


mloubout:
@EdCaunt ok so this is not as minor as anticipated. Mainly the cause of the issue you raised is that, unless you do `subs(self, new)` then the substitution is delayed and the dictionary stored as a property. it means if you do `f.dx.subs({'a': 2})` then it will keep that dict and it will be part of the `==` comparison.


So in your case you end up with two different `fdx`. The original one and a new one that is "fdx but remember {fdx2: ....}" which even though in this case won't do anything leads to it being a different object. There isn't really an easy way out of it because some of the keys may not exist until after evaluation so there is no way to know if that dict needs to be saved to not.

So to summarize,  your issue is just a side effect of the lazy evaluation and there is no straightforward way to get out of it.

PS: renaming as WIP and draft cuz may not be doable nicely.

devito/finite_differences/derivative.py
```diff
@@ -213,6 +213,14 @@ def subs(self, *args, **kwargs):
         # because a new derivative could be created i.e `f.dx.subs(f.dx, y)`
         if len(args) == 2 and args[0] == self:
             return args[1]
+        # Check if we are calling subs(self, {old: new}, **hint) in which case
+        # return the standard substitution.
+        elif len(args) == 1 and self in args[0]:
+            new = args[0].pop(self)
```
FabioLuporini, line 219: just use `self`?

devito/finite_differences/derivative.py
```diff
@@ -213,6 +213,14 @@ def subs(self, *args, **kwargs):
         # because a new derivative could be created i.e `f.dx.subs(f.dx, y)`
         if len(args) == 2 and args[0] == self:
```
FabioLuporini, line 214: is self

devito/finite_differences/derivative.py
mloubout: No this is `args[0][self]` not `self`. I.e this extract what to replace `self` with.

Issue #2003, opened 2022-09-20
local_indices errors with SparseFunctions and SparseTimeFunctions running with MPI and 1 point


deckerla:
call this script example.py
(this behavior also holds for `SparseTimeFunctions`)
```
import devito
grd = devito.Grid(shape=(4,5))
f = devito.Function(name="f", grid=grd)
sf = devito.SparseFunction(name="sf", grid=grd, npoint=1)
print(sf.local_indices)
```
if I run 
`DEVITO_MPI=1 mpirun -n 2 python example.py`
it errors with
```
(slice(0, 1, None),)
Traceback (most recent call last):
  File "/home/cvx/.julia/dev/Devito/ex.py", line 5, in <module>
    print(sf.local_indices)
  File "/home/cvx/.conda/envs/conda_jl/lib/python3.10/site-packages/cached_property.py", line 36, in __get__
    value = obj.__dict__[self.func.__name__] = self.func(obj)
  File "/home/cvx/.conda/envs/conda_jl/lib/python3.10/site-packages/devito/types/dense.py", line 643, in local_indices
    return tuple(self._distributor.glb_slices.get(d, slice(0, s))
  File "/home/cvx/.conda/envs/conda_jl/lib/python3.10/site-packages/devito/types/dense.py", line 643, in <genexpr>
    return tuple(self._distributor.glb_slices.get(d, slice(0, s))
  File "/home/cvx/.conda/envs/conda_jl/lib/python3.10/site-packages/cached_property.py", line 36, in __get__
    value = obj.__dict__[self.func.__name__] = self.func(obj)
  File "/home/cvx/.conda/envs/conda_jl/lib/python3.10/site-packages/devito/mpi/distributed.py", line 116, in glb_slices
    return {d: slice(min(i), max(i) + 1)
  File "/home/cvx/.conda/envs/conda_jl/lib/python3.10/site-packages/devito/mpi/distributed.py", line 116, in <dictcomp>
    return {d: slice(min(i), max(i) + 1)
ValueError: min() arg is an empty sequence
```
At least for the purposes of Devito.jl we would love for it to return a tuple  `(0:-1,)` for ranks where the sparsefunction memory isn't allocated.

FabioLuporini:
@deckerla this is what your PR has fixed right Luke? can we close it?

mloubout:
Yes it is, PR was saying close instead of fix so didn't auto close. closing

PR #2004, opened 2022-09-20
return slice(0,-1) for glb_slices if glb_numb empty on an mpi rank

deckerla:


deckerla:
Closes #2003

deckerla:
`local_indices` tests in MPI seem to be turned off at the moment, so I am not sure what the best way is to have unit tests for this PR

https://github.com/devitocodes/devito/blob/6e1dcc502a36b59716378cfa5e6924fcbbdb5bd5/tests/test_mpi.py#L2336

mloubout:
>tests in MPI seem to be turned off a

No this is the "main" to debug, i.e so that you can run `python test_mpi.py`. Pytest ignores this it runs all tests functions named `test_*`. So you can just use an extra case to the test here:

https://github.com/devitocodes/devito/blob/6e1dcc502a36b59716378cfa5e6924fcbbdb5bd5/tests/test_mpi.py#L364

deckerla:
@mloubout thanks for the advice!  most recent push includes modified unit test covering the behavior

tests/test_mpi.py
```diff
@@ -395,24 +395,28 @@ def test_ownership(self, shape, coords, points):
         assert list(ownership.keys()) == [grid.distributor.myrank]
 
     @pytest.mark.parallel(mode=4)
-    @pytest.mark.parametrize('coords,expected', [
-        ([(0.5, 0.5), (1.5, 2.5), (1.5, 1.5), (2.5, 1.5)], [[0.], [1.], [2.], [3.]]),
+    @pytest.mark.parametrize('coords,expected,expectedinds', [
+        ([(0.5, 0.5), (1.5, 2.5), (1.5, 1.5), (2.5, 1.5)], [[0.], [1.], [2.], [3.]], [(slice(0,1),),(slice(1,2),),(slice(2,3),),(slice(3,4),)]),
+        ([(1.5, 1.5),], [[],[],[],[0.]], [(slice(0,-1),),(slice(0,-1),),(slice(0,-1),),(slice(0,1),)])
```
mloubout, line 400: Looks good thanks. Can you check line length? We require <92 characters per line in CI.

You can run `flakke8 tests/test_mpi.py` to check (it should be installed as a devitio dependency).

mloubout (commented):
Looks good thanks.

One minor CI comment but should be goog to go

tests/test_mpi.py
deckerla: thanks for pointing that out! should now be resolved.  the line also needed some white space after commas :)

PR #2005, opened 2022-09-21
dependencies: sympy 1.11 compatibility

mloubout:
Small PR than fixes the newly introduced bug in sympy for `_print_Add`. We should keep that bug fix in our printer even once 1.12 is released so that we stay compatible with sympy 1.11

### DIFF
--- a/devito/symbolics/printer.py
+++ b/devito/symbolics/printer.py
@@ -3,8 +3,10 @@ Utilities to turn SymPy objects into C strings.
 """
 
 import numpy as np
+import sympy
 
 from mpmath.libmp import prec_to_dps, to_str
+from packaging.version import Version
 from sympy.logic.boolalg import BooleanFunction
 from sympy.printing.precedence import PRECEDENCE_VALUES, precedence
 from sympy.printing.c import C99CodePrinter
@@ -115,6 +117,27 @@ class CodePrinter(C99CodePrinter):
         func = "abs" if is_integer else "fabs"
         return "%s(%s)" % (func, self._print(expr.args[0]))
 
+    def _print_Add(self, expr, order=None):
+        """"
+        Print an addition.
+        """
+        terms = self._as_ordered_terms(expr, order=order)
+
+        PREC = precedence(expr)
+        l = []
+        for term in terms:
+            t = self._print(term)
+            if precedence(term) < PREC:
+                l.extend(["+", "(%s)" % t])
+            elif t.startswith('-'):
+                l.extend(["-", t[1:]])
+            else:
+                l.extend(["+", t])
+        sign = l.pop(0)
+        if sign == '+':
+            sign = ""
+        return sign + ' '.join(l)
+
     def _print_Float(self, expr):
         """Print a Float in C-like scientific notation."""
         prec = expr._prec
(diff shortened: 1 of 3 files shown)


FabioLuporini (approved):
Uncontroversial

georgebisbas (approved):
I guess notebooks should be fixed, and ready to go

mloubout:
>and ready to go

No there is still a bug with `str(expr)` looking into best way to fix it

devito/symbolics/printer.py
```diff
@@ -223,3 +251,8 @@ def ccode(expr, **settings):
         the input ``expr`` itself.
     """
     return CodePrinter(settings=settings).doprint(expr, None)
+
+
+# Sympy 1.11 has introduced a bug in `_print_Add`, so we enforce here
+# to always use the correct one from our printer
+setattr(sympy.printing.str.StrPrinter, '_print_Add', CodePrinter._print_Add)
```
mloubout, line 258: @FabioLuporini quite hacky but the only "simple" way to have `str` work properly without too much issues while avoiding skipping a sympy version

devito/symbolics/printer.py
```diff
@@ -115,6 +117,30 @@ def _print_Abs(self, expr):
...
+        keep the implementation here.
+        """
+        terms = self._as_ordered_terms(expr, order=order)
+
+        PREC = precedence(expr)
+        l = []
+        for term in terms:
+            t = self._print(term)
+            if precedence(term) < PREC:
+                l.extend(["+", "(%s)" % t])
```
mloubout, line 134: @FabioLuporini that's the part I don't like in sympy. They have
```
sign = "-" if t.startswith("-") else "="
l.extend([sign, "(%s)" % t[1:]])
```
Which is weird. Because they assume you can refactorize `-xxxxxx` into `-(xxxxxx)` no matter what `t` is the precedence is lower. I can't find a quick erroring case but it looks wrong.

mloubout:
Ok sorry for the delay, looks like we GTG.

Thecnically fixed in sympy (https://github.com/sympy/sympy/pull/24030) from original issue (https://github.com/sympy/sympy/issues/24022) but from looking at issues/PRs and links I thnk is safer to keep it internal.

devito/symbolics/printer.py
```diff
@@ -115,6 +117,30 @@ def _print_Abs(self, expr):
         func = "abs" if is_integer else "fabs"
         return "%s(%s)" % (func, self._print(expr.args[0]))
 
+    def _print_Add(self, expr, order=None):
+        """"
+        Lifted from sympy's PR #24030 that fixes the _print_Add in sympy 1.11
```
georgebisbas, line 122: Add a link to PR?

devito/symbolics/printer.py
FabioLuporini: I see. Can you open an issue and/or propose a PR? 

We can give them a day or two to respond, if they don't, we merge ours? 

And if this procrastinate the sympy syncing (because e.g. a further patch would have to wait for 1.13) then we can merge ours regardless

FabioLuporini:
@mloubout to be extra clear, I don't mind merging our PR now, so it's a go for me 👍 Above I was just assessing alternatives

devito/symbolics/printer.py
mloubout: Yeah gonna try to make an MFE, I think they merged a PR that fixes the MFE but in a weird way will let you know

PR #2009, opened 2022-09-27
compiler: Misc refactorings towards serialization support

FabioLuporini:


devito/operator/profiling.py
```diff
@@ -230,10 +235,16 @@ def summary(self, args, dtype, reduce_over=None):
...
-                                     for k, v in i.items()])
-                        for i in data.itermaps]
-            itershapes = tuple(tuple(i.values()) for i in itermaps)
+            try:
+                itermaps = [OrderedDict([(k, int(subs_op_args(v, args)))
+                                         for k, v in i.items()])
+                            for i in data.itermaps]
+                itershapes = tuple(tuple(i.values()) for i in itermaps)
+            except TypeError:
+                # E.g., a section comprising just function calls
```
georgebisbas, line 244: redundant?

devito/passes/clusters/buffering.py
```diff
@@ -250,18 +257,30 @@ def callback(self, clusters, prefix, cache=None):
...
+                # complementary guards, hence only one will be execute. In such a
+                # case, we can split all equations over separate IterationSpaces
+                key0 = lambda: Stamp()
+            elif b.is_readonly:
+                # `b` is read multiple times -- this could just be the case of
+                # coupled equations, so we more cautiously perform a
+                # "buffer-wise" splitting of the IterationSpaces (i.e., only
+                # relevant if there are at least two read-only buffers)
+                stamp = Stamp()
+                key0 = lambda: stamp
```
georgebisbas, line 272: are lines 271-272 not doing the same thing with line 265?

devito/passes/iet/orchestration.py
```diff
@@ -46,35 +45,25 @@ def _make_waitlock(self, iet, sync_ops):
 
         return iet, []
 
-    def _make_releaselock(self, iet, sync_ops):
-        preactions = []
-        preactions.append(BusyWait(Or(*[CondNe(s.handle, 2) for s in sync_ops])))
-        preactions.extend(DummyExpr(s.handle, 0) for s in sync_ops)
+    def _make_releaselock(self, iet, sync_ops, *args):
+        pre = []
```
georgebisbas, line 49: Previous name looked better

georgebisbas (commented):
I cannot claim I follow everything. Will take another look.

devito/passes/clusters/asynchrony.py
```diff
@@ -48,14 +47,12 @@ def callback(self, clusters, prefix):
 
         d = prefix[-1].dim
 
-        if not all(SEQUENTIAL in c.properties[d] for c in clusters):
```
mloubout, line 51: Why can this be dropped?

devito/passes/clusters/buffering.py
mloubout: No 272 reuses the same stamp everytime while 265 creates a new one every time ke0 is called

mloubout (commented):
Minor comments looks fine

devito/operator/profiling.py
FabioLuporini: I wouldn't say so?

devito/passes/clusters/buffering.py
FabioLuporini: indeed

devito/passes/clusters/asynchrony.py
FabioLuporini: it doesn't really matter whether time is parallel or sequential. The time dimensions is, regardless of its computational nature, designated to be the dimension along which streaming occurs. "Designated" -> I mean by the compiler callbacks, which one could in theory customize, but why on earth :)

So I've made things a little more general...

This is because I had a simple test with `Op(Eq(usave, time))` where `time` clearly is parallel but streaming wouldn't trigger because of that annoying condition

devito/operator/profiling.py
georgebisbas: It is:
```
                # E.g., a section comprising just function calls    
                # E.g., a section comprising just function calls, or at least
                # a sequence of unrecognized or non-conventional expr statements
```
Why the part:
`# E.g., a section comprising just function calls`
is duplicated?

devito/passes/clusters/buffering.py
georgebisbas: Thanks for this. I feel I never got the Stamps right.

FabioLuporini: you don't need to understand what Stamp achieves to pick up the difference here though

it's basically as Mathias said

the underlying idea is actually very straightforward. There are a few caveats but in essence: ** different stamp => no fusion possible **  (the converse is generally false btw, and I'd say also of couse)

mloubout (approved):

devito/operator/profiling.py
```diff
@@ -230,10 +235,16 @@ def summary(self, args, dtype, reduce_over=None):
...
+                # E.g., a section comprising just function calls, or at least
```
georgebisbas, line 245: duplicate?

georgebisbas (approved):

PR #2010, opened 2022-09-30
Update FAQ.md

FabioLuporini:
I've just realized that `par-tile` is outrageously under-documented. Not to mention the fact that maybe it should be the default, rather than collapse...

At a bare minimum it should be added here:
https://github.com/devitocodes/devito/blob/master/examples/performance/00_overview.ipynb

and here:
https://github.com/devitocodes/devito/blob/master/examples/performance/README.md

Potentially reusable running example:
https://github.com/devitocodes/devito/blob/master/tests/test_gpu_openacc.py#L101

Can anyone look into this?

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -29,6 +29,7 @@
 - [How do I find the source of my bug quickly and get support](#how-do-i-find-the-source-of-my-bug-quickly-and-get-support)
 - [Is there a way to get the performance of an Operator](#is-there-a-way-to-get-the-performance-of-an-operator)
 - [How does devito compute the performance of an Operator](#how-does-devito-compute-the-performance-of-an-operator)
+- [How do I fairly compare the performance to that of my in-house code](#how-do-i-fairly-compare-the-performance-to-that-of-my-in-house-code)
 - [Is there is list of refereed papers related to the Devito project](#is-there-a-list-of-refereed-papers-related-to-the-devito-project)
 - [How do I cite Devito](#how-do-i-cite-devito)
 - [Where did the name Devito come from?](#where-did-the-name-devito-come-from)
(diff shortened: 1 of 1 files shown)


georgebisbas:
I can have a look from next week, full tied up today.

FAQ.md
```diff
@@ -668,6 +669,21 @@ Compiler ninjas may wonder about the counting of loop-invariant sub-expressions,
 
 [top](#Frequently-Asked-Questions)
 
+## How do I fairly compare the performance to that of my in-house code
+
+It's challenging! Here's a potentially non-exhaustive list of things to check:
+
+* Check the physics — e.g., there are many TTI variations out there, are the two codes using the same one? You wanna ensure the PDEs are the same.
+* Check the discretization — do the two codes use the same spatial order? most importantly, is it the same spatial order in all problem dimensions? carefully optimized codes often use lower order stencils in some dimensions. This significantly impacts performance
```
georgebisbas, line 677: I would add spatial **discretisation** order.

FAQ.md
```diff
@@ -668,6 +669,21 @@ Compiler ninjas may wonder about the counting of loop-invariant sub-expressions,
...
+* Did you try tuning the performance of the Devito `Operator`? E.g., on GPUs it is worth giving the `par-tile` option a go.
```
georgebisbas, line 678: > Did you try tuning the performance of the Devito `Operator`? E.g., on GPUs it is worth giving the `par-tile` option a go.
Did you try tuning the performance of the Devito `Operator`? E.g., both on CPUs and GPUs cache blocking size is a key parameter in performance. It is worth giving the `par-tile` option a go or use DEVITO_AUTOTUNING=aggressive.

FAQ.md
```diff
@@ -668,6 +669,21 @@ Compiler ninjas may wonder about the counting of loop-invariant sub-expressions,
...
+* Snapshotting: compression, serialization, asynchronous streaming if running on device... Many legacy codes have these features enabled out-of-the-box. Are you using them?
```
georgebisbas, line 679: I would move this towards the end of the list.

FAQ.md
FabioLuporini: I mention par-tile because it's the only lacking proper documentation

and probably we should make it the default instead of collapse

FAQ.md
FabioLuporini: why?

PR #2011, opened 2022-10-03
types: Minor fixes to sparse function indices and implicit dims

mloubout:
- Fixes remaining hardcoded `-1` for the position of the sparse index
- - Fix missing implicit dims for the final equation of interpolation

Added test for it

### DIFF
--- a/devito/operations/interpolators.py
+++ b/devito/operations/interpolators.py
@@ -256,7 +256,8 @@ class LinearInterpolator(GenericInterpolator):
 
             # Write/Incr `self`
             lhs = self.sfunction.subs(self_subs)
-            last = [Inc(lhs, rhs)] if increment else [Eq(lhs, rhs)]
+            ecls = Inc if increment else Eq
+            last = [ecls(lhs, rhs, implicit_dims=implicit_dims)]
 
             return temps + summands + last
 
--- a/tests/test_interpolation.py
+++ b/tests/test_interpolation.py
@@ -5,7 +5,7 @@ import pytest
 from sympy import Float
 
 from devito import (Grid, Operator, Dimension, SparseFunction, SparseTimeFunction,
-                    Function, TimeFunction,
+                    Function, TimeFunction, DefaultDimension, Eq,
                     PrecomputedSparseFunction, PrecomputedSparseTimeFunction,
                     MatrixSparseTimeFunction)
 from examples.seismic import (demo_model, TimeAxis, RickerSource, Receiver,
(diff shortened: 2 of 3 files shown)


devito/operations/interpolators.py
```diff
@@ -256,7 +256,8 @@ def callback():
 
             # Write/Incr `self`
             lhs = self.sfunction.subs(self_subs)
-            last = [Inc(lhs, rhs)] if increment else [Eq(lhs, rhs)]
+            last = [Inc(lhs, rhs, implicit_dims=implicit_dims)] if increment else \
```
FabioLuporini, line 259: imho more readable as:

```
if increment:
    cls = Inc
else:
    cls = eq
last = [cls(lhs, rhs, ...)]
```

tests/test_interpolation.py
```diff
@@ -14,6 +14,11 @@
 import scipy.sparse
 
 
+class SparseFirst(SparseFunction):
```
FabioLuporini, line 17: why not making this  test-local ?

FabioLuporini (approved):
great

devito/operations/interpolators.py
mloubout: Yeah thought about doing it and forgot before pushing will fix

tests/test_interpolation.py
mloubout: Wasn't sure if was gonna need more tests for it in the future (I'm sure someone will come up with more bugs) so thought was better but happy to move it for now

tests/test_interpolation.py
```diff
@@ -590,3 +590,38 @@ def test_msf_interpolate():
...
+
+    # Operator
+    u = TimeFunction(name="u", grid=grid, time_order=1)
+    fs = Function(name="fs", grid=grid, dimensions=(*dims, ds), shape=(11, 11, 3))
+
+    eqs = [Eq(u.forward, u+1), Eq(fs, u)]
+    # No time dependence so need the implicit dim
+    rec = s.interpolate(expr=s+fs, implicit_dims=grid.stepping_dim)
+    op = Operator(eqs + rec)
+
```
georgebisbas, line 624: I would add a check of the order of dimensions:
using assert_structure(op, ....)

georgebisbas (commented):
Good thanks, left a comment to add a test

tests/test_interpolation.py
mloubout: There is still some simd warning and op "issues" to be resolved, this mainly makes sure result is correct and that the legacy hardcoded value are fixed.

Issue #1435, opened 2020-08-21
ConditionalDimension: Devito produces a code referencing a dimension before its declaration


maksgraczyk:
When ConditionalDimension is used with the intended inner-most dimension as its parent and more than one dimension in the condition, Devito sometimes produces a code referencing one of the dimensions before its declaration. This leads to a compilation error.

MFE:
```
from devito import *
from sympy import And, Or

t1, t2, t3, t4, t5, t6, t7, t8, t9, t10 = \
    dimensions('t1 t2 t3 t4 t5 t6 t7 t8 t9 t10')

f0 = Function(name='f0', grid=Grid(shape=(2, 2, 4, 4),
                                   dimensions=(t1, t2, t3, t4)))
f1 = Function(name='f1', grid=Grid(shape=(2, 2, 3, 3),
                                   dimensions=(t5, t6, t7, t8)))
f2 = Function(name='f2', grid=f1.grid)

cd = ConditionalDimension(name='cd', parent=t10,
                          condition=Or(Gt(f0[t5, t6, t7 + t9,
                                             t8 + t10],
                                          f1[t5, t6, t7, t8]),
                                       And(~Ne(f0[t5, t6, t7 + t9,
                                                  t8 + t10],
                                               f1[t5, t6, t7, t8]),
                                           Lt(2 * t9 + t10,
                                              f2[t5, t6, t7, t8]))))

op = Operator([Eq(f1[t5, t6, t7, t8], f0[t5, t6, t7 + t9, t8 + t10],
                  implicit_dims=cd),
               Eq(f2[t5, t6, t7, t8], 2 * t9 + t10, implicit_dims=cd)])

op.apply(t9_M=1, t10_M=1)
```

Traceback:
```
/tmp/devito-jitcache-uid1000/54ae8b2b7dd776276090529a8f3645f1098562ce.c: In function ‘Kernel’:
/tmp/devito-jitcache-uid1000/54ae8b2b7dd776276090529a8f3645f1098562ce.c:48:43: error: ‘t8’ undeclared (first use in this function); did you mean ‘t9’?
   48 |             if (f0[t5][t6][t7 + t9][t10 + t8] == f1[t5][t6][t7][t8] && t10 + 2*t9 < f2[t5][t6][t7][t8] || f0[t5][t6][t7 + t9][t10 + t8] > f1[t5][t6][t7][t8])
      |                                           ^~
      |                                           t9
[...]

PR #2012, opened 2022-10-04
compiler: Further misc improvements

### LINKED ISSUE #1927: No default arg-value for saved TimeFunctions over ConditionalDimension
on going work here: https://github.com/devitocodes/devito/tree/timeM-argvalue

FabioLuporini:
fixes #1927 

proves that issue #1435 has actually been fixed, by adding the attached test

Some more misc improvements, polishing, tweaks

### DIFF
--- a/devito/builtins/initializers.py
+++ b/devito/builtins/initializers.py
@@ -2,12 +2,13 @@ import numpy as np
 
 import devito as dv
 from devito.tools import as_tuple, as_list
-from devito.builtins.utils import nbl_to_padsize, pad_outhalo
+from devito.builtins.utils import abstract_args, nbl_to_padsize, pad_outhalo
 
 __all__ = ['assign', 'smooth', 'gaussian_smooth', 'initialize_function']
 
 
-def assign(f, rhs=0, options=None, name='assign', **kwargs):
+@abstract_args
+def assign(f, rhs=0, options=None, name='assign', argmap=None, **kwargs):
     """
     Assign a list of RHSs to a list of Functions.
 
@@ -61,7 +62,7 @@ def assign(f, rhs=0, options=None, name='assign', **kwargs):
     else:
         for i, j in zip(as_list(f), rhs):
             eqs.append(dv.Eq(i, j))
-    dv.Operator(eqs, name=name, **kwargs)()
+    dv.Operator(eqs, name=name, **kwargs)(**argmap)
 
 
 def smooth(f, g, axis=None):
--- a/devito/builtins/initializers.py
+++ b/devito/builtins/initializers.py
@@ -7,6 +7,7 @@ from devito.builtins.utils import abstract_args, nbl_to_padsize, pad_outhalo
 __all__ = ['assign', 'smooth', 'gaussian_smooth', 'initialize_function']
 
 
+@dv.switchconfig(log_level='ERROR')
 @abstract_args
 def assign(f, rhs=0, options=None, name='assign', argmap=None, **kwargs):
     """
--- a/devito/builtins/initializers.py
+++ b/devito/builtins/initializers.py
@@ -2,7 +2,7 @@ import numpy as np
 
 import devito as dv
 from devito.tools import as_tuple, as_list
-from devito.builtins.utils import abstract_args, nbl_to_padsize, pad_outhalo
+from devito.builtins.utils import nbl_to_padsize, pad_outhalo
 
 __all__ = ['assign', 'smooth', 'gaussian_smooth', 'initialize_function']
 
--- a/devito/builtins/initializers.py
+++ b/devito/builtins/initializers.py
@@ -8,8 +8,7 @@ __all__ = ['assign', 'smooth', 'gaussian_smooth', 'initialize_function']
 
 
 @dv.switchconfig(log_level='ERROR')
-@abstract_args
-def assign(f, rhs=0, options=None, name='assign', argmap=None, **kwargs):
+def assign(f, rhs=0, options=None, name='assign', **kwargs):
     """
     Assign a list of RHSs to a list of Functions.
 
@@ -63,7 +62,7 @@ def assign(f, rhs=0, options=None, name='assign', argmap=None, **kwargs):
     else:
         for i, j in zip(as_list(f), rhs):
             eqs.append(dv.Eq(i, j))
-    dv.Operator(eqs, name=name, **kwargs)(**argmap)
+    dv.Operator(eqs, name=name, **kwargs)()
 
 
 def smooth(f, g, axis=None):
--- a/devito/builtins/utils.py
+++ b/devito/builtins/utils.py
@@ -1,7 +1,12 @@
+from functools import wraps
+
 import numpy as np
 
 import devito as dv
+from devito.symbolics import uxreplace
 from devito.tools import as_tuple
+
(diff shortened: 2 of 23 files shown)


tests/test_derivatives.py
```diff
@@ -416,9 +416,13 @@ def test_subsampled_fd(self):
 
         eqns = [Eq(u.forward, u + 1.), Eq(u2.forward, u2.dx)]
         op = Operator(eqns)
-        op.apply(time_M=nt-2)
-        # Verify that u2[1, x,y]= du2/dx[0, x, y]
 
+        # NOTE: To assert against u.data[-1] we need to provide explicit bounds
```
FabioLuporini, line 420: @mloubout note

devito/builtins/initializers.py
```diff
@@ -2,11 +2,12 @@
 
 import devito as dv
 from devito.tools import as_tuple, as_list
-from devito.builtins.utils import nbl_to_padsize, pad_outhalo
+from devito.builtins.utils import abstract_args, nbl_to_padsize, pad_outhalo
```
mloubout, line 5: where is this used?

tests/test_derivatives.py
mloubout: Not sure I follow why, if these are needed now in such case it's a bit of a pain to have the user needing to infer those.

devito/builtins/utils.py
```diff
@@ -98,3 +103,32 @@ def pad_outhalo(function):
             # Need to access it so that that worker is not blocking exectution since
             # _data_with_outhalo requires communication
             function._data_with_outhalo._local[0] = function._data_with_outhalo._local[0]
+
+
+def abstract_args(func):
```
mloubout, line 108: Can't find where this is use

devito/operator/operator.py
```diff
@@ -511,7 +522,7 @@ def _prepare_arguments(self, autotune=None, **kwargs):
                 # E.g., SubFunctions
                 continue
             for k, v in p._arg_values(**kwargs).items():
-                if k in args and args[k] != v:
+                if k in args and args[k] not in as_tuple(v) and k not in futures:
```
mloubout, line 525: why the `as_tuple` and wouldn't it be dangerous if `v` is an array (which we saw in some previous PR can happen in some very weird corner cases)

mloubout (commented):
Minor comments, looks good

devito/builtins/initializers.py
FabioLuporini: I tried it here: https://github.com/devitocodes/devito/pull/2012/commits/65cd098196bec4e3e466c24693bcab334fb09698

then dropped it...

I was seeing a weird error in PRO, but to be fair it might just have been a confounder since I was fighting a batch of concurrent minor bugs

I left it here around because I think it's got potential and I hope to either get back to it soon or that someone will 😬

devito/builtins/utils.py
FabioLuporini: see comment above

devito/operator/operator.py
FabioLuporini: now v can be a set of legal values such as `{14, 15, 16}` or a plain number such as `14`. I hadn't thought about the numpy case with is... crucial 😬 

so yes, gonna change this

as for why `v` can now be a set, it's part of the 1927 fix. Take a look at `ConditionalDimension._arg_defaults`

tests/test_derivatives.py
FabioLuporini: no, not necessary in general.

Just in this case because you don't just care about `u2`, but also about the modulo buffered TimeFunction namely `u`

I could have also changed the assert to

```
assert np.allclose(u.data[-2], nt-2)
```

Iterating up to 11 or 10 is legal -- in both cases `u2` will have tracked its last value.
Whether `u.data[-1]` is of interest or not is up to the user

devito/operator/operator.py
FabioLuporini: actually... 

```
args[k] != v:
```

this, in master, would cause a whole-array comparison too??

mloubout: No it wouldn't it would complain that you need to use `.all()` in boolean context

mloubout: You en up here for example when you use a `SubFunction` with a weird name and it tries to compare coordinate values

devito/tools/data_structures.py
```diff
@@ -201,11 +201,17 @@ class OrderedSet(OrderedDict, MutableSet):
 
     Notes
     -----
-    Originally extracted from:
+    Readapted from:
 
         https://stackoverflow.com/questions/1653970/does-python-have-an-ordered-set
     """
 
+    def __init__(self, *args, **kwargs):
```
georgebisbas, line 209: Is this offering something additional?

examples/compiler/04_iet-B.ipynb
```diff
@@ -475,7 +475,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": "Python 3 (ipykernel)",
```
georgebisbas, line 478: These can be discarded but ok

tests/test_builtins.py
```diff
@@ -79,6 +79,18 @@ def define(self, dimensions):
                                            [0, 2, 2, 0],
                                            [0, 0, 0, 0]])
 
+    def test_assign_subsampled_timefunction(self):
```
georgebisbas, line 82: Why in builtins and not test_dimension?

tests/test_dimension.py
```diff
@@ -1214,6 +1214,76 @@ def test_sparse_time_function(self):
...
+                                                       f1[t5, t6, t7, t8]),
+                                                   Lt(2 * t9 + t10,
+                                                      f2[t5, t6, t7, t8]))))
+
+        op = Operator([Eq(f1[t5, t6, t7, t8], f0[t5, t6, t7 + t9, t8 + t10],
+                          implicit_dims=cd),
+                       Eq(f2[t5, t6, t7, t8], 2 * t9 + t10, implicit_dims=cd)])
+
+        # Check it compiles correctly! See issue report
+        op.cfunction
```
georgebisbas, line 1243: Why don't we op.apply to check results as well? It does not make sense?

mloubout (approved):

tests/test_dimension.py
FabioLuporini: let me know what number you want me to assert against :)

tests/test_builtins.py
FabioLuporini: because it's using a builtin

devito/tools/data_structures.py
FabioLuporini: yes

tests/test_dimension.py
georgebisbas: Added here in this branch2PR: https://github.com/devitocodes/devito/pull/2015/files

devito/operator/operator.py
FabioLuporini: just a minor note that this was addressed on slack

tests/test_dimension.py
FabioLuporini: done

PR #2014, opened 2022-10-05
compiler: restrict qualifier as property

georgebisbas:


devito/types/basic.py
```diff
@@ -945,6 +945,10 @@ def name(self):
         """The name of the object."""
         return self._name
 
+    @property
+    def _C_type_qualifier(self):
```
georgebisbas, line 949: Should any docstring be added, or obvious?

FabioLuporini:
I think it should be inferred from this instead https://github.com/devitocodes/devito/blob/44dc63b00360ce70ff93c0e19191148d27bdc36f/devito/tools/utils.py#L279

georgebisbas:
> I think it should be inferred from this instead
> 
> https://github.com/devitocodes/devito/blob/44dc63b00360ce70ff93c0e19191148d27bdc36f/devito/tools/utils.py#L279

for  class IndexedData(IndexedBase) as well?

for AbstractFunction, ok

devito/types/basic.py
mloubout, line 949: Any reason why we don't want it in `CodeSymbol` with the rest?

georgebisbas: right, thanks

mloubout:
What happend here?

georgebisbas:
> What happend here?

hell, I pushed from my fork! WIll cleanup...

FabioLuporini:
the `unexpansion` branch probably has the machinery to do this right, but it's weeks away from merging state

FabioLuporini:
closing as superseded by #2060

FabioLuporini:
@georgebisbas you decide whether it's worth keeping the branch around or not

georgebisbas:
I guess as soon as I rebase my devito-xdsl-fork on top of latest master I will let you know. I am fine to close now, thanks for the fix in your PR

PR #2017, opened 2022-10-12
ci: Remove dangling docker layers

FabioLuporini:


.github/workflows/pytest-gpu.yml
```diff
@@ -78,7 +78,10 @@ jobs:
 
     - name: Build docker image
       run: |
-          docker build . --pull --file docker/Dockerfile.devito --tag devito_img --build-arg base=${{ matrix.base }}
+          docker build . --rm --pull --file docker/Dockerfile.devito --tag devito_img --build-arg base=${{ matrix.base }}
```
mloubout, line 81: Isn't `--rm` a run option

.github/workflows/pytest-gpu.yml
```diff
@@ -78,7 +78,10 @@ jobs:
...
+
+    - name: Remove dangling layers
```
mloubout, line 83: So why is it needed again?

.github/workflows/docker-devito.yml
```diff
@@ -107,6 +107,9 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           build-args: base=devitocodes/${{ matrix.base }}
 
+      - name: Remove dangling layers
```
georgebisbas, line 110: I had the impression this was happening already, ok

.github/workflows/pytest-gpu.yml
FabioLuporini: it's a build too -- it removes the dangling layers!

.github/workflows/pytest-gpu.yml
FabioLuporini: ~300 GB of layers on v1003 and counting, which I purged manully for now

if we pushed too much in a good week before docker-build kicks in (which does purge IIRC), we might end up saturating the hard drive

though now that I think about it, that's weird, the base image shouldn't get rebuilt each time... mh

mloubout: If there is ~300 GB of layers then there is something else going on probably due to
- cancelled workflow
- Dev on the machine

So this may "fix" some of it but it should be safer to have the actual machine have a safety cron job with that

mloubout (approved):

PR #2018, opened 2022-10-12
symbolics: Fix absolute value warning for integer input

### LINKED ISSUE #1957: Absolute value of an integer produces C warning
Here is an MFE:
```
import numpy as np
from sympy import sign
from devito import Function, Grid, Eq, Operator

gridR = Grid(shape=(2,))

index = Function(name="index", grid=gridR, dtype=np.int32)
R = Function(name="R", grid=gridR)
a = gridR.dimensions

eqs = [Eq(index[a], abs(sign(R[a])))]


op = Operator(eqs)
op.apply()
```

produces the following warnings:
```
...snipped.c:36:25: warning: using floating point absolute value function 'fabs' when argument is of integer type [-Wabsolute-value]
    index[x + 1] = fmin(fabs((((B[x + 1]) > 0) - ((B[x + 1]) < 0))), fabs((((R[x + 1]) > 0) - ((R[x + 1]) < 0))));
                        ^
...snipped.c:36:25: note: use function 'abs' instead
    index[x + 1] = fmin(fabs((((B[x + 1]) > 0) - ((B[x + 1]) < 0))), fabs((((R[x + 1]) > 0) - ((R[x + 1]) < 0))));
                        ^~~~
                        abs
...snipped.c:36:70: warning: using floating point absolute value function 'fabs' when argument is of integer type [-Wabsolute-value]
    index[x + 1] = fmin(fabs((((B[x + 1]) > 0) - ((B[x + 1]) < 0))), fabs((((R[x + 1]) > 0) - ((R[x + 1]) < 0))));
                                                                     ^
...snipped.c:36:70: note: use function 'abs' instead
    index[x + 1] = fmin(fabs((((B[x + 1]) > 0) - ((B[x + 1]) < 0))), fabs((((R[x + 1]) > 0) - ((R[x + 1]) < 0))));
                                                                     ^~~~
                                                                     abs
2 warnings generated.
```

I think the correct behaviour would be to generate `abs` instead of `fabs` here. 

Happy to file a PR fixing it if people agree this is incorrect/unwanted behaviour. 

Can @FabioLuporini or @mloubout  or @georgebisbas point me to where to look in order to change the mapping?

mloubout:
Fixes #1957


georgebisbas (approved):

devito/symbolics/printer.py
```diff
@@ -94,6 +96,23 @@ def _print_Mod(self, expr):
         result = '%'.join(args)
         return result
 
+    def _print_Abs(self, expr):
+        """Print an absolute value. Use `abs` if can infer it is an Integer"""
+        # `abs` undefined in soem offloading cases
+        if isinstance(configuration['platform'], Device):
```
FabioLuporini, line 102: we can't access configuration in a printer... it escapes all abstractions (e.g., this would break if platform=cpu but one creates `Operator(..., platform=nvidiaX)`

nvidia definitely supports abs/fabs/... see for example here: https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__SINGLE.html#group__CUDA__MATH__SINGLE_1gb00f8593e1bfb1985526020fbec4e0fc

where did you see the failure? perhaps it's just missing an include? though math.h should suffice...

devito/symbolics/printer.py
```diff
@@ -94,6 +96,23 @@ def _print_Mod(self, expr):
...
+            return "fabs(%s)" % self._print(expr.args[0])
+
+        # Check if argument is an integer
+        is_integer = True
+        for a in expr.args[0].args:
```
FabioLuporini, line 107: should you rather use `atoms` or something like that? otherwise you might miss nested symbols if a deep expression

(test to be added accordingly if so)

tests/test_symbolics.py
```diff
@@ -214,6 +214,12 @@ def test_extended_sympy_arithmetic():
     assert ccode(-1 + bar) == '-1 + o->bar'
 
 
+def test_integer_abs():
```
georgebisbas, line 217: If the conf issue mentioned here : https://github.com/devitocodes/devito/pull/2018/files#r994249271
is fixed, is it worth adding any GPU tests?

devito/symbolics/printer.py
mloubout: Is an AMD issue with aomp looks like

https://github.com/devitocodes/devito/actions/runs/3242390619/jobs/5315585337

devito/symbolics/printer.py
mloubout: There is probably room for improvement, this is a "basic" one to remove the most obvious ones

devito/symbolics/printer.py
FabioLuporini: Let me take a look at his

georgebisbas (approved):

FabioLuporini (approved):
Thanks, LGTM

devito/operator/operator.py
```diff
@@ -639,6 +639,21 @@ def _soname(self):
         """A unique name for the shared object resulting from JIT compilation."""
         return Signer._digest(self, configuration)
 
+    @cached_property
+    def ccode(self):
+        """
```
FabioLuporini, line 644: docstrings (`__doc__` attribute) are inherited as well, so you should drop it here. Can do in the subsequent PR if this one is ready

FabioLuporini (commented):
Very very nice cleanup 👍🏻

devito/operator/operator.py
mloubout: Done in next PR

PR #2019, opened 2022-10-12
compiler: Fix subdim arguemnt mismatch

mloubout:
On top of #2018 

Fix a small corner case where you can end with something like `[nt, frozenset{nt, nt+1, ...}]` needing to be reduced/unique'ed with subdimensions and apply overrides.


tests/test_dimension.py
```diff
@@ -542,7 +542,17 @@ def test_basic(self):
...
         eqns = [Eq(u.forward, u + 1.), Eq(u2.forward, u2 + 1.), Eq(usave, u)]
         op = Operator(eqns)
-        op.apply(t_M=nt-2)
+        op.apply()
+        assert np.all(np.allclose(u.data[(nt-1) % 3], nt-1))
+        assert np.all([np.allclose(u2.data[i], i) for i in range(nt)])
+        assert np.all([np.allclose(usave.data[i], i*factor)
+                      for i in range((nt+factor-1)//factor)])
+
+        # Check overrides
```
FabioLuporini, line 551: is this worth a separate test *but* with just equation (so simpler than this one in terms of symbols and equations) *and* the override? might help future debugging

devito/tools/data_structures.py
```diff
@@ -113,7 +113,10 @@ def compare_to_first(v):
             if isinstance(first, np.ndarray) or isinstance(v, np.ndarray):
                 return (first == v).all()
             else:
-                return first == v
+                try:
+                    return v.issuperset({first})
```
FabioLuporini, line 117: AKA 

```
first in v
```

?

FabioLuporini (commented):
basically no objections here, just minor comments

devito/tools/data_structures.py
mloubout: Well no (pushing a change wit hseparate test later). needs to be `v.issuperset(frozenset(first))` because `first` could be a set as well if there is multiple subdims in which case `first in v` would be false but `issuperset` would be true

tests/test_dimension.py
mloubout: done

FabioLuporini (approved):
LGTM, thanks

devito/operator/operator.py
```diff
@@ -641,16 +641,6 @@ def _soname(self):
 
     @cached_property
     def ccode(self):
-        """
-        Generate C code.
```
georgebisbas, line 645: I would keep this

mloubout: https://github.com/devitocodes/devito/pull/2018#discussion_r997811708

PR #2023, opened 2022-10-21
[draft] Unified Memory Allocator

guaacoelho:
Hello everyone,

We of SENAI CIMATEC are working in a Unified Memory Allocator to Devito trough CuPy library.

The first results using this new allocator have impressive results when using checkpointing compared to default allocator in GPU.

The impact of performance in our experiment using Overthrust (894x884x299) are close to three times* compared to default allocator in Devito.

With this approach we expect to be able to allocate memory beyond the GPU capacity in the future.

We will open a Draft PR to standardize this allocator with Devito patterns, fix possible bugs and open to community use :)

A version enabling this through External Allocator is in development too and we expect share this soon.

All feedbacks are welcome :)

Thank you all CIMATEC and Devito team, to make this possible :slightly_smiling_face:

*All experiments were using Nvidia V100 with 32 GB of memory.

### DIFF
--- a/devito/data/allocators.py
+++ b/devito/data/allocators.py
@@ -5,6 +5,7 @@ import mmap
 import os
 import sys
 
+import cupy as cp
 import numpy as np
 import ctypes
 from ctypes.util import find_library
@@ -15,7 +16,7 @@ from devito.tools import dtype_to_ctype
 
 __all__ = ['ALLOC_FLAT', 'ALLOC_NUMA_LOCAL', 'ALLOC_NUMA_ANY',
            'ALLOC_KNL_MCDRAM', 'ALLOC_KNL_DRAM', 'ALLOC_GUARD',
-           'default_allocator']
+           'CUPY_ALLOC', 'default_allocator']
 
 
 class MemoryAllocator(object):
@@ -317,6 +318,33 @@ class NumaAllocator(MemoryAllocator):
         return self._node == 'local'
 
 
+class CupyAllocator(MemoryAllocator):
+
+    """
+    Memory allocator based on ``posix`` functions. The allocated memory is
+    aligned to page boundaries.
+    """
+
+    is_Posix = True
+
+    def __init__(self):
+        cp.cuda.set_allocator(cp.cuda.MemoryPool(cp.cuda.malloc_managed).malloc)
+    
+    @classmethod
+    def initialize(cls):
+        pass
+
+
+    def _alloc_C_libcall(self, size, ctype):
+
+        mem_obj = cp.zeros(size, dtype=cp.float64)
+        return mem_obj.data.ptr, mem_obj
+
+    def free(self, c_pointer):
+        pass
+
+
+
 class ExternalAllocator(MemoryAllocator):
 
     """
@@ -373,6 +401,7 @@ ALLOC_KNL_DRAM = NumaAllocator(0)
 ALLOC_KNL_MCDRAM = NumaAllocator(1)
 ALLOC_NUMA_ANY = NumaAllocator('any')
 ALLOC_NUMA_LOCAL = NumaAllocator('local')
+CUPY_ALLOC = CupyAllocator()
 
 custom_allocators = {}
 """User-defined allocators."""
--- a/devito/data/allocators.py
+++ b/devito/data/allocators.py
@@ -344,7 +344,6 @@ class CupyAllocator(MemoryAllocator):
         pass
 
 
-
 class ExternalAllocator(MemoryAllocator):
 
     """
--- a/devito/data/data.py
+++ b/devito/data/data.py
@@ -3,7 +3,7 @@ from functools import wraps
 
 import numpy as np
 
-from devito.data.allocators import ALLOC_FLAT
+from devito.data.allocators import ALLOC_FLAT, CUPY_ALLOC
 from devito.data.utils import *
 from devito.logger import warning
 from devito.parameters import configuration
@@ -82,7 +82,8 @@ class Data(np.ndarray):
             # Dask/Distributed context), which may (re)create a Data object
             # without going through `__array_finalize__`
             return
-        self._allocator.free(*self._memfree_args)
+        if self._allocator is not CUPY_ALLOC:
+            self._allocator.free(*self._memfree_args)
         self._memfree_args = None
 
     def __reduce__(self):
--- a/devito/passes/iet/definitions.py
+++ b/devito/passes/iet/definitions.py
@@ -20,6 +20,8 @@ from devito.symbolics import (Byref, DefFunction, FieldFromPointer, IndexedPoint
 from devito.tools import as_mapper, as_list, as_tuple, filter_sorted, flatten
 from devito.types import DeviceRM, Symbol
 
+from devito.data.allocators import CUPY_ALLOC
+
 __all__ = ['DataManager', 'DeviceAwareDataManager', 'Storage']
 
 
@@ -434,6 +436,9 @@ class DeviceAwareDataManager(DataManager):
         synchronize the host and device copies, while the latter does not.
         """
         mmap = self.lang._map_to(obj)
+
+        if obj._allocator is CUPY_ALLOC:
+            return
 
         if read_only is False:
             unmap = [self.lang._map_update(obj),


guaacoelho:
It was used the version 8.3.0 of Cupy

speglich:
ToDo: 

* Distributed GPU allocation

devito/data/data.py
```diff
@@ -82,7 +82,8 @@ def __del__(self):
             # Dask/Distributed context), which may (re)create a Data object
             # without going through `__array_finalize__`
             return
-        self._allocator.free(*self._memfree_args)
+        if self._allocator is not CUPY_ALLOC:
```
FabioLuporini, line 85: No, you should override `free` instead

devito/data/allocators.py
```diff
@@ -15,7 +16,7 @@
 
 __all__ = ['ALLOC_FLAT', 'ALLOC_NUMA_LOCAL', 'ALLOC_NUMA_ANY',
            'ALLOC_KNL_MCDRAM', 'ALLOC_KNL_DRAM', 'ALLOC_GUARD',
-           'default_allocator']
+           'CUPY_ALLOC', 'default_allocator']
```
FabioLuporini, line 19: ALLOC_CUPY for homogeneity

devito/data/allocators.py
```diff
@@ -317,6 +318,32 @@ def put_local(self):
         return self._node == 'local'
 
 
+class CupyAllocator(MemoryAllocator):
+
+    """
+    Memory allocator based on ``posix`` functions. The allocated memory is
```
FabioLuporini, line 324: copy-paste docstring

devito/data/allocators.py
```diff
@@ -317,6 +318,32 @@ def put_local(self):
...
+    aligned to page boundaries.
+    """
+
+    is_Posix = True
```
FabioLuporini, line 328: leftover.....

devito/data/allocators.py
```diff
@@ -317,6 +318,32 @@ def put_local(self):
...
+    def initialize(cls):
+        pass
+
+
+    def _alloc_C_libcall(self, size, ctype):
+
+        mem_obj = cp.zeros(size, dtype=cp.float64)
+        return mem_obj.data.ptr, mem_obj
+
+    def free(self, c_pointer):
```
FabioLuporini, line 343: CuPy frees the memory right? unless we explicitly tell it at some point?

devito/data/data.py
FabioLuporini: in fact, `free` is a no-op, so why the need to special-case here?

devito/passes/iet/definitions.py
```diff
@@ -435,6 +437,9 @@ def _map_function_on_high_bw_mem(self, site, obj, storage, devicerm, read_only=F
         """
         mmap = self.lang._map_to(obj)
 
+        if obj._allocator is CUPY_ALLOC:
```
FabioLuporini, line 440: this breaks the abstraction in all sort of ways, unfortunately

You can't access the allocator here to steer compilation

what *really* matters at this point is the `_mem_space`of the object: https://github.com/devitocodes/devito/blob/master/devito/types/basic.py#L39

you shouldn't actually end up here, because GPU-allocated functions shuold have a local mem-space, which in turns naturally prevents them ever enter this point

FabioLuporini (commented):
Thanks for the contribution! I left a few comments.

There also are other issues, in particular lack of tests, update to installation, etc. You may want to take a look at this file: https://github.com/devitocodes/devito/blob/master/CONTRIBUTING.md#contributing-to-devito

speglich:
@FabioLuporini, we added tests to this allocator, could we proceed with this PR?

tests/test_data.py
```diff
@@ -206,6 +207,26 @@ def test_indexing_into_sparse(self):
         sf.data[1:-1, 0] = np.arange(8)
         assert np.all(sf.data[1:-1, 0] == np.arange(8))
 
+    def test_uma_allocation(self):
```
FabioLuporini, line 210: There's a `test_external_allocator` somewhere in this file. Could you move that test, and this test, within a new class, say  `TestAllocators`?

FabioLuporini:
> @FabioLuporini, we added tests to this allocator, could we proceed with this PR?

The issue is that CuPy is an optional dependency (shipped by the NVidia SDK I'm guessing?), so if I try to run CI here, it will undoubtedly break. 

So, we need conditional imports, a mechanism like [this](https://github.com/devitocodes/devito/pull/2023/files#diff-ab72c15c798471001f2f7a29225ce8bd24648e3241ab548c709e5cd13b6a8914R276) to emit suitable error messages in case one attempts to use the allocator but the allocator isn't available, etc etc

FabioLuporini (changes requested):
Based on what I said above, a few tweaks are still necessary here

mloubout:
Should this be added to the GPU CI and actually tested on the nvidia run?

tests/test_data.py
guaacoelho: I updated the PR with this change

speglich:
> Should this be added to the GPU CI and actually tested on the nvidia run?

Probably @mloubout

devito/data/allocators.py
```diff
@@ -13,9 +13,15 @@
 from devito.parameters import configuration
 from devito.tools import dtype_to_ctype
 
+# try:
+#     from mpi4py import MPI  # noqa
+# except ImportError:
+#     MPI = None
```
mloubout, line 19: Leftover

devito/data/allocators.py
```diff
@@ -317,6 +326,51 @@ def put_local(self):
...
+
+    @classmethod
+    def initialize(cls):
+
+        try:
+            import cupy as cp
+            cls.lib = cp
+            cls._initialize_shared_memory()
+            try:
+                from mpi4py import MPI
```
mloubout, line 344: Why reimport and not import it from `devito.mpi`?

devito/data/allocators.py
```diff
@@ -317,6 +326,51 @@ def put_local(self):
...
+                cls.MPI = MPI
+                cls._set_device_for_mpi()
+            except:
```
mloubout, line 347: except `ImportError` other errors should be caught

devito/data/allocators.py
```diff
@@ -317,6 +326,51 @@ def put_local(self):
...
+    def _set_device_for_mpi(cls):
+        if cls.MPI.Is_initialized():
+            n_gpu = cls.lib.cuda.runtime.getDeviceCount()
+            rank_l = cls.MPI.COMM_WORLD.Split_type(cls.MPI.COMM_TYPE_SHARED).Get_rank()
+            cls.lib.cuda.runtime.setDevice(rank_l % n_gpu)
+
+    def _alloc_C_libcall(self, size, ctype):
+        if not self.available():
+            raise ImportError("Couldn't initialize cupy or MPI elements of alocation")
+        mem_obj = self.lib.zeros(size, dtype=self.lib.float64)
```
mloubout, line 367: always `float64`? I thoiught size was for UInt8

devito/data/allocators.py
```diff
@@ -76,10 +82,13 @@ def alloc(self, shape, dtype):
...
-        # cast to 1D array of the specified size
-        ctype_1d = ctype * size
-        buf = ctypes.cast(c_pointer, ctypes.POINTER(ctype_1d)).contents
-        pointer = np.frombuffer(buf, dtype=dtype)
+        if c_pointer:
+            # cast to 1D array of the specified size
+            ctype_1d = ctype * size
+            buf = ctypes.cast(c_pointer, ctypes.POINTER(ctype_1d)).contents
+            pointer = np.frombuffer(buf, dtype=dtype)
+        else:
```
mloubout, line 90: How can we end up here? the `c_pointer is None` case is already above

tests/test_data.py
```diff
@@ -1,8 +1,10 @@
 import pytest
 import numpy as np
+import cupy as cp
```
mloubout, line 3: Need to be added somehow to the test requirements and this step should be decoratred with a `skipif(device)`

mloubout (changes requested):

devito/data/allocators.py
guaacoelho: During the execution in MPI, domain splitting can generate a situation where the allocated data size is zero, as we have observed with Sparse Functions. When this occurs, Cupy returns a pointer with a value of zero. This conditional statement was defined for this case.

georgebisbas: Could you add a comment noting this, until some better solution is around?

mloubout:
Can you check if #2171 fixes the install issue

speglich:
@mloubout, definitely, we will test it. BTW, this docker image worked too, thanks to you.

```dockerfile

FROM nvcr.io/nvidia/nvhpc:23.5-devel-cuda12.1-ubuntu22.04 AS devel

RUN apt-get update -y && \
    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
        git \
        make \
        wget && \
    rm -rf /var/lib/apt/lists/*

RUN apt-get update -y && \
    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
        python3 \
        python3-dev \
        python3-pip \
        python3-setuptools \
        python3-wheel \
        python3.10-venv && \
    rm -rf /var/lib/apt/lists/*

RUN pip3 --no-cache-dir install cupy-cuda12x
```

devito/data/allocators.py
speglich: We will push it, George.

mloubout:
Ok the nvidia setup has been updated so please:
- rebase and answer comments
- add the test to the nvidia test suite
For the second point you may have to modify the `pytest-gpu.yml` workflow file to add that extra test to it

devito/data/allocators.py
```diff
@@ -317,6 +327,51 @@ def put_local(self):
...
+    """
+    Memory allocator based on Unified Memory concept. The allocation is made using Cupy.
+    """
+    _mempool = None
+
+    @classmethod
+    def initialize(cls):
+
+        try:
+            import cupy as cp
```
mloubout, line 341: Move `try import` for cupy at the top

```
try:
  import cupy as cp
except importerror:
  cp = None
```

and just check if None here.

MPI should just be imported at the top from devito.mpi

tests/test_data.py
```diff
@@ -1473,6 +1475,55 @@ def test_gather_time_function(self):
             assert ans == np.array(None)
 
 
+class TestAllocators(object):
```
mloubout, line 1478: move to `test_gpu_common` as `TestCupyAllocator` with an nividia device skip so it's added to GPU CI

kmk142789 (approved):

PR #2025, opened 2022-10-24
More tweaks towards support shared memory in parlang backends

FabioLuporini:
"parlang" is the new designated word to generically refer to low level languages for device programming, such as CUDA, HIP, SYCL. 

Note that currently for pragma-based programming models (OpenMP, OpenACC) we use "parpragma" 

Anyway, this PR is just yet another batch of fixes, tweaks, generalizations... nothing of particular relevance


tests/test_dle.py
```diff
@@ -262,33 +262,61 @@ def test_cache_blocking_structure_optrelax_linalg(opt, expected):
...
+
+        bns, _ = assert_blocking(op, {'x0_blk0', 'x1_blk0'})
+        assert len(bns) == len(expected)
+        for root, v in zip(bns.values(), expected):
+            iters = FindNodes(Iteration).visit(root)
+            iters = [i for i in iters if i.dim.is_Block and i.dim._depth == 1]
+            assert len(iters) == len(v)
+            assert all(i.step == j for i, j in zip(iters, v))
+
+    def test_structure_2p5D(self):
```
georgebisbas, line 295: 2p5D?

tests/test_dle.py
```diff
@@ -262,33 +262,61 @@ def test_cache_blocking_structure_optrelax_linalg(opt, expected):
...
+        par_tile = (16, 4)
 
-    eqns = [Eq(u.forward, u.dx),
-            Eq(v.forward, u.forward.dx)]
+        op = Operator(eqns, opt=('advanced', {'par-tile': par_tile,
+                                              'blockinner': True}))
 
-    op = Operator(eqns, opt=('advanced', {'par-tile': par_tile,
-                                          'blockinner': True}))
+        # 3D grid, but par-tile has only 2 entries => generates so called
```
georgebisbas, line 309: 2.5D blocking is usually used for 2D+simd, which is what by default would be constructed even without par-tile. 
Maybe rephrase this comment?

georgebisbas (changes requested):
Need change of PR name

devito/mpi/halo_scheme.py
```diff
@@ -113,15 +117,27 @@ def union(self, halo_schemes):
             for k, v in i.fmapper.items():
                 hse = fmapper.setdefault(k, v)
 
-                # The `loc_indices` must match
                 if hse.loc_indices != v.loc_indices:
-                    raise ValueError("Cannot compute the union of one or more HaloScheme "
-                                     "when the `loc_indices` differ")
+                    # The `loc_dirs` must match otherwise it'd be a symptom there's
+                    # something horrinbly broken somewhere!
```
georgebisbas, line 122: typos

tests/test_dle.py
FabioLuporini: no why? I think the comment makes sense as it stands

tests/test_dle.py
FabioLuporini: "2 point 5" -- makes sense, I think, as it's within the `TestBlockingParTile` class (so it's like "2.5D given par-tile")

devito/mpi/halo_scheme.py
FabioLuporini: yes, fixing (these and those from the old PR in profiling.py, as promised)

tests/test_dle.py
georgebisbas: ah ok, I thought it was referring to 2 dimensional 5 point jacobi stencil!

tests/test_dimension.py
```diff
@@ -45,6 +45,38 @@ def test_multi_buffer_long_time(self):
         for i in range(7):
             assert np.allclose(g.data[i], 14+i+1)
 
+    def test_modulo_dims_generation(self):
+        grid = Grid((3, 3))
+
+        f = TimeFunction(name="f", grid=grid, save=Buffer(2))
```
mloubout, line 51: Can we add a test that

```
f = TimeFunction(name="f", grid=grid, save=Buffer(2), time_order=2)
f.dt2 == ...
```

It works correctly but would be good to have it as a safety because this `Buffer(2)` will be mostly used in those cases

mloubout (approved):

georgebisbas (approved):

mloubout (approved):

FabioLuporini:
Thanks everybody for the patience and the reviews. Now it's finally time to merge. Landing 🛩️ !

PR #2028, opened 2022-11-03
ci: Streamlining

ggorman:


.github/workflows/docker-devito.yml
```diff
@@ -22,13 +22,13 @@ jobs:
...
             test: 'tests/test_gpu_openacc.py tests/test_gpu_common.py'
-            runner: ["self-hosted", "gpu", "docker", "v1004"]
+            runner: ["self-hosted", "nvidiagpu"]
 
           - base: 'bases:nvidia-clang'
             tag: 'nvidia-clang'
             flag: '--gpus all'
             test: 'tests/test_gpu_openmp.py tests/test_gpu_common.py'
-            runner: ["self-hosted", "gpu", "docker", "kimogila"]
+            runner: ["self-hosted", "nvidiagpu", "docker"]
```
mloubout, line 31: `amdgpu` below?

FabioLuporini: yes

and why do we need "docker" here and not in the one above?

FabioLuporini (changes requested):
The asv action is not addressed here

Remember: it used to run on kimogila and it kicks in only after merges into master

And I don't remember whether a new ASV profile needs to be created on the node before it can run anything. This should maybe be tested with a temporary commit (to be dropped before merge) that enables the ASV action upon pushed to this branch

FabioLuporini:
@mloubout @ggorman 

yes, ASV needs changing, at least here https://github.com/devitocodes/devito/blob/master/.github/workflows/asv.yml#L58

and the `asv` tag (https://github.com/devitocodes/devito/blob/master/.github/workflows/asv.yml#L26) should be attached to the new designated node

FabioLuporini:
and, one last comment: as per the [contributing guidelines](https://github.com/devitocodes/devito/blob/master/CONTRIBUTING.md#making-changes), you are expected to prefix all your commit messages as well as the PR tile with a suitable tag among those available

FabioLuporini (approved):
Now GTG. Thanks

mloubout:
Wasn't the asv one supposed to be changed too?

PR #2029, opened 2022-11-04
compiler: init support for HPE Cray compiler

georgebisbas:


devito/arch/compiler.py
```diff
@@ -472,6 +474,37 @@ def __lookup_cmds__(self):
         self.MPICXX = 'mpicxx'
 
 
+class CrayCompiler(Compiler):
```
mloubout, line 477: That's a lot of duplicates cannot it just inherit from the amd compiler? Also my understanding is that aompcc is getting deprecated and should switch to amdclang with all the flags always

FabioLuporini: it would be conceptually bad to inherit from AOMPCompiler

I think I'm OK with a certain degree of duplication here

What's the CRAY compiler based on underneath? gnu or clang?

devito/arch/compiler.py
```diff
@@ -472,6 +474,37 @@ def __lookup_cmds__(self):
...
+
+    """HPE Cray's compiler."""
+
+    def __init__(self, *args, **kwargs):
+        super(CrayCompiler, self).__init__(*args, **kwargs)
+
+        self.cflags += ['-Wno-unused-result', '-Wno-unused-variable']
```
FabioLuporini, line 484: we're incrementally converting all these += into extend/append

devito/arch/compiler.py
```diff
@@ -472,6 +474,37 @@ def __lookup_cmds__(self):
...
+        self.cflags.append('-march=native')
+
+        language = kwargs.pop('language', configuration['language'])
+        if language == 'openmp':
+            self.ldflags += ['-fopenmp']
+
+        # For MPI, mpicc is compiled against amdclang not aompcc, so need the flags back.
+        if kwargs.get('mpi'):
+            self.ldflags.extend(['-target', 'x86_64-pc-linux-gnu'])
+            self.ldflags.extend(['-fopenmp'])
```
FabioLuporini, line 498: append

devito/arch/compiler.py
FabioLuporini: but yeah, regardless, in my experience these compiler classes tend to require too much special-casing, so it's in general difficult and not particularly simplifying to share code in some more abstract base classes... I think

georgebisbas: It is clang underneath. I will see what I can drop, this is why I marked the PR as draft.
I will explore these options, and will ping you on any more advice

mloubout: >it would be conceptually bad to inherit from AOMPCompiler

Yes but the way it is going with aomp, both this one and aomp should be a `ClangCompiler` with their few specific flags

georgebisbas: So I think we can inherit from Clang so far; I do not see any difference apart from the compiler invocation name.

devito/arch/compiler.py
```diff
@@ -436,6 +438,20 @@ def __lookup_cmds__(self):
         self.MPICXX = 'mpicxx'
 
 
+class CrayCompiler(ClangCompiler):
+
+    """HPE Cray's Clang compiler."""
+
+    def __init__(self, *args, **kwargs):
```
FabioLuporini, line 445: Why overriding the constructor?

devito/arch/compiler.py
```diff
@@ -46,6 +46,8 @@ def sniff_compiler_version(cc):
         compiler = "gcc"
     elif ver.startswith("clang"):
         compiler = "clang"
+    elif ver.startswith("cray"):
```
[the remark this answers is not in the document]
georgebisbas: ok

devito/arch/compiler.py
```diff
@@ -436,6 +438,20 @@ def __lookup_cmds__(self):
         self.MPICXX = 'mpicxx'
 
 
+class CrayCompiler(ClangCompiler):
+
+    """HPE Cray's Clang compiler."""
```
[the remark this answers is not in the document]
FabioLuporini: so why is it called "cray" ?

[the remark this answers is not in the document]
georgebisbas: This is the official docs term: https://support.hpe.com/hpesc/public/docDisplay?docId=a00115299en_us&page=About_the_Cray_Clang_C_and_C___Quick_Reference.html

FabioLuporini (approved):

FabioLuporini:
Merged, thanks

PR #2032, opened 2022-11-21
misc: Add Omnitrace to Dockerfile.amd

FabioLuporini:


docker/Dockerfile.amd
```diff
@@ -45,6 +45,16 @@ ENV PATH=${PATH}:/opt/openmpi/bin:$AOMP/bin
...
 
+# Install Omnitrace, an AMDResearch profiling tool which is de facto the best profiler from AMD
+RUN wget https://github.com/AMDResearch/omnitrace/releases/download/v1.7.3/omnitrace-1.7.3-ubuntu-20.04-ROCm-50100-PAPI-OMPT-Python3.sh -q
+RUN mv omnitrace-1.7.3-ubuntu-20.04-ROCm-50100-PAPI-OMPT-Python3.sh omnitrace-installer.sh && \
+    chmod +x omnitrace-installer.sh && \
+    mkdir -p /opt/omnitrace && \
+    mv omnitrace-installer.sh /opt/omnitrace && \
+    cd /opt/omnitrace && \
+    ./omnitrace-installer.sh --skip-license && \
+    cd -
```
mloubout, line 56: no needed automatically exits back to base dir at the ned of RUN

docker/Dockerfile.amd
```diff
@@ -45,6 +45,16 @@ ENV PATH=${PATH}:/opt/openmpi/bin:$AOMP/bin
 ENV LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/opt/openmpi/lib:$AOMP/lib
 ENV OMPI_CC=$AOMP/bin/clang
 
+# Install Omnitrace, an AMDResearch profiling tool which is de facto the best profiler from AMD
+RUN wget https://github.com/AMDResearch/omnitrace/releases/download/v1.7.3/omnitrace-1.7.3-ubuntu-20.04-ROCm-50100-PAPI-OMPT-Python3.sh -q
```
mloubout, line 49: Should this be somehow linked to the `rocm` version at the top?

docker/Dockerfile.amd
mloubout: Can be simplified to one line with just 
```
RUN ./omnitrace-1.7.3-ubuntu-20.04-ROCm-50100-PAPI-OMPT-Python3.sh --prefix=/opt/omnitrace --exclude-subdir --skip-license && rm omnitrace-1.7.3-ubuntu-20.04-ROCm-50100-PAPI-OMPT-Python3.sh
```

docker/entrypoint.sh
```diff
@@ -15,4 +15,9 @@ if [[ -z "${DEPLOY_ENV}" ]]; then
     ./codecov -t -t ${CODECOV_TOKEN} -F "${DEVITO_ARCH}-${DEVITO-PLATFORM}"
 else
     exec "$@"
-fi
\ No newline at end of file
+fi
+
+if [[ -n "${ROCM_VERSION}" ]]; then
+    echo "sourcing Omnitrace"
+    source /opt/omnitrace/share/omnitrace/setup-env.sh
```
mloubout, line 22: may need a `COPY chown /opt` (pseudo code) for safety

mloubout (commented):
Some nitpicking

docker/Dockerfile.amd
FabioLuporini: Probably, but... it's a bit tedious now

Also the OS right?

What if I leave a comment at the top to remember to update this accordingly if the ROCM_VERSION changes?

Because whatever I do here it still sounds quite fragile regardless. But anyway open to ideas... what did you have in mind? parse the ROCM_VERSION and transform it in bash into "50100" or something like that?

FabioLuporini: I will leave a question in their github

FabioLuporini: Done: https://github.com/AMDResearch/omnitrace/issues/217

docker/Dockerfile.amd
FabioLuporini: OK

docker/entrypoint.sh
FabioLuporini: I don't think so those files are readable

docker/Dockerfile.amd
mloubout: That's a bad naming convention jeez. Yeah may have to just stick to a given version

jrmadsen: The naming convention arises from all the optional 3rd party library support. PAPI support is optional, so is OpenMP tools, so it Python, and there's also build variants with full MPI support (it will aggregate profiling data), and you can even build it with support with callbacks to others profilers like TAU, Caliper, etc. There's actually a fair amount excluded at this point: Kokkos-Tools, partial MPI support, and RCCL support are included but I gave up detailing them. I've considered dropping OMPT and PAPI since those do not require runtime link dependencies similar to Kokkos-Tools/partial MPI/RCCL but I've tried to avoid changing the convention

docker/entrypoint.sh
jrmadsen: Since you are using modulefiles for omniperf, might as well use `module use /opt/omnitrace/share/modulefiles`

docker/Dockerfile.amd
jrmadsen: Ah, in fact it may be worth noting, ROCm support is optional too. You can use omnitrace on pure CPU code

docker/Dockerfile.amd
```diff
@@ -45,6 +45,27 @@ ENV PATH=${PATH}:/opt/openmpi/bin:$AOMP/bin
...
 ENV OMPI_CC=$AOMP/bin/clang
 
+# Install Omnitrace, an AMDResearch profiling tool akin to Nvidia's Nsight-systems
+RUN wget https://github.com/AMDResearch/omnitrace/releases/download/v1.7.3/omnitrace-1.7.3-ubuntu-20.04-ROCm-50100-PAPI-OMPT-Python3.sh -q
+RUN ./omnitrace-1.7.3-ubuntu-20.04-ROCm-50100-PAPI-OMPT-Python3.sh --prefix=/opt/omnitrace --exclude-subdir --skip-license && \
+    rm omnitrace-1.7.3-ubuntu-20.04-ROCm-50100-PAPI-OMPT-Python3.sh
+
+# Install Omniperf, an AMDResearch profiling tool akin to Nvidia's Nsight-compute
+RUN apt-get update && apt-get install -y software-properties-common && \   # cmake3.19 required by Omniperf...
+    wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | sudo apt-key add - && \
```
mloubout, line 55: no sudo in docker

docker/Dockerfile.amd
```diff
@@ -45,6 +45,27 @@ ENV PATH=${PATH}:/opt/openmpi/bin:$AOMP/bin
 ENV LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/opt/openmpi/lib:$AOMP/lib
 ENV OMPI_CC=$AOMP/bin/clang
 
+# Install Omnitrace, an AMDResearch profiling tool akin to Nvidia's Nsight-systems
+RUN wget https://github.com/AMDResearch/omnitrace/releases/download/v1.7.3/omnitrace-1.7.3-ubuntu-20.04-ROCm-50100-PAPI-OMPT-Python3.sh -q
+RUN ./omnitrace-1.7.3-ubuntu-20.04-ROCm-50100-PAPI-OMPT-Python3.sh --prefix=/opt/omnitrace --exclude-subdir --skip-license && \
+    rm omnitrace-1.7.3-ubuntu-20.04-ROCm-50100-PAPI-OMPT-Python3.sh
+
+# Install Omniperf, an AMDResearch profiling tool akin to Nvidia's Nsight-compute
+RUN apt-get update && apt-get install -y software-properties-common && \   # cmake3.19 required by Omniperf...
```
mloubout, line 54: move `software-properties-common` to top with the ap update

docker/Dockerfile.amd
```diff
@@ -45,6 +45,27 @@ ENV PATH=${PATH}:/opt/openmpi/bin:$AOMP/bin
...
+    wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | sudo apt-key add - && \
+    apt-add-repository 'deb https://apt.kitware.com/ubuntu/ bionic main' && \
+    apt-get update && \
+    apt install -y -q cmake
+RUN apt-get update && apt-get install lmod  # to eventually bring it in PATH via the entrypoint
```
mloubout, line 59: redundant `update`, same as `software-properties` just add `lmod` to initial extra apt install line 15

docker/Dockerfile.amd
```diff
@@ -45,6 +45,27 @@ ENV PATH=${PATH}:/opt/openmpi/bin:$AOMP/bin
...
+ENV OMNIPERF_DIR=/app/omniperf
+RUN wget https://github.com/AMDResearch/omniperf/releases/download/v1.0.4/omniperf-1.0.4.tar.gz -q
+RUN tar xfz omniperf-1.0.4.tar.gz && \
+    cd omniperf-1.0.4/ && \
+    /venv/bin/pip install --no-cache-dir -r requirements.txt && \
+    mkdir build && cd build/ && \
+    cmake -DCMAKE_INSTALL_PREFIX=${OMNIPERF_DIR}/1.0.4 -DPYTHON_DEPS=${OMNIPERF_DIR}/python-libs -DMOD_INSTALL_PATH=${OMNIPERF_DIR}/modulefiles .. && \
+    make install
+
```
mloubout, line 68: `rm omniperf-1.0.4.tar.gz` and `rm mniperf-1.0.4/` after install?

mloubout (commented):
some extra comments

docker/Dockerfile.amd
FabioLuporini: uops yeah this was a copy-paste from the Internet

docker/entrypoint.sh
FabioLuporini: thanks, will do that

docker/Dockerfile.amd
FabioLuporini: OK

docker/Dockerfile.amd
FabioLuporini: yes

FabioLuporini (commented):
@mloubout one question

I had to switch from

```
ENV OMNIPERF_DIR=/opt/omniperf
```

to 

```
ENV OMNIPERF_DIR=/app/omniperf
```

because for some reason the package wouldn't appear in `opt`. Any idea what the cause might be? 😮

FabioLuporini:
@mloubout I've pushed some edits, but I still have to test this: https://github.com/AMDResearch/omnitrace/issues/217#issuecomment-1323574921

mloubout:
That's weird for `/opt` that's where nvhpc is installed and it works fine.

Did you look at the log? May be just that need to create the directory first

FabioLuporini:
log was fine AFAICT

Also, this env var

```
ENV OMNIPERF_DIR=/app/omniperf
```

Doesn't persist in the container. LIterally, it's not there. Any idea @mloubout ?

FabioLuporini:
I pushed several edits. Could you take another look @mloubout ?

docker/Dockerfile.amd
```diff
@@ -1,18 +1,36 @@
-##############################################################
-# This Dockerfile contains AMD compilers
-##############################################################
+####################################################################################
+# This Dockerfile contains the AMD ROCm SDK as well as other useful tools for Devito
```
georgebisbas, line 2: as well as AMD profiling tools (Omnitrace, Omniperf) useful to Devito... or something like this...

docker/Dockerfile.amd
```diff
@@ -1,18 +1,36 @@
...
+########################################################################
+FROM rocm/dev-ubuntu-20.04:${ROCM_VERSION}-complete as sdk-base
 
 ENV DEBIAN_FRONTEND noninteractive
 
 ARG ROCM_VERSION
 ENV rocm=/opt/rocm-${ROCM_VERSION}
 ENV AOMP=/opt/rocm-${ROCM_VERSION}/llvm
 
+# Bring in cmake3.19 apt repo
```
georgebisbas, line 18: is it fixed to cmake3.19?

docker/Dockerfile.amd
```diff
@@ -45,11 +63,52 @@ ENV PATH=${PATH}:/opt/openmpi/bin:$AOMP/bin
 ENV LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/opt/openmpi/lib:$AOMP/lib
 ENV OMPI_CC=$AOMP/bin/clang
 
-# Devito env
+# Install Omnitrace, an AMDResearch profiling tool akin to Nvidia's Nsight-systems
```
georgebisbas, line 66: I would drop akin to....

docker/Dockerfile.amd
```diff
@@ -45,11 +63,52 @@ ENV PATH=${PATH}:/opt/openmpi/bin:$AOMP/bin
...
+ENV OMNITRACE_DIR=/opt/omnitrace
+RUN wget https://github.com/AMDResearch/omnitrace/releases/download/v1.7.3/omnitrace-1.7.3-ubuntu-20.04-ROCm-50100-PAPI-OMPT-Python3.sh -q && \
+    chmod +x omnitrace-1.7.3-ubuntu-20.04-ROCm-50100-PAPI-OMPT-Python3.sh && \
+    mkdir -p ${OMNITRACE_DIR} && \
+    ./omnitrace-1.7.3-ubuntu-20.04-ROCm-50100-PAPI-OMPT-Python3.sh --prefix=${OMNITRACE_DIR} --exclude-subdir --skip-license && \
+    rm omnitrace-1.7.3-ubuntu-20.04-ROCm-50100-PAPI-OMPT-Python3.sh
+
+# Install Omniperf, an AMDResearch profiling tool akin to Nvidia's Nsight-compute
```
georgebisbas, line 74: same

docker/Dockerfile.amd
```diff
@@ -45,11 +63,52 @@ ENV PATH=${PATH}:/opt/openmpi/bin:$AOMP/bin
...
+########################################################################
+# AOMP for GPUs via OpenMP config
+########################################################################
+FROM sdk-base as aomp
+
 ENV DEVITO_ARCH="aomp"
 ENV DEVITO_PLATFORM="amdgpuX"
 ENV DEVITO_LANGUAGE="openmp"
 
+# This will only trigger if arch is hip since the final stage depends on it
```
georgebisbas, line 97: same

docker/Dockerfile.amd
```diff
@@ -45,11 +63,52 @@ ENV PATH=${PATH}:/opt/openmpi/bin:$AOMP/bin
...
+    cd omniperf-1.0.4/ && \
+    mkdir -p ${OMNIPERF_DIR} && \
+    python3 -m pip install -t ${OMNIPERF_DIR}/python-libs -r requirements.txt && \
+    mkdir build && cd build/ && \
+    cmake -DCMAKE_INSTALL_PREFIX=${OMNIPERF_DIR}/1.0.4 -DPYTHON_DEPS=${OMNIPERF_DIR}/python-libs -DMOD_INSTALL_PATH=${OMNIPERF_DIR}/modulefiles .. && \
+    make install && \
+    rm -rf omniperf-1.0.4.tar.gz omniperf-1.0.4/
+ENV PYTHONPATH ${OMNIPERF_DIR}/python-libs:${PYTHONPATH}
+
+# This will only trigger if arch is aomp since the final stage depends on it
```
georgebisbas, line 87: arch or base?

georgebisbas (commented):
Added some minor comments

docker/Dockerfile.amd
```diff
@@ -45,11 +63,52 @@ ENV PATH=${PATH}:/opt/openmpi/bin:$AOMP/bin
...
 
+# This will only trigger if arch is hip since the final stage depends on it
+########################################################################
+# HIPCC for GPUs via HIP config
+########################################################################
+FROM sdk-base as hip
+
+ENV DEVITO_ARCH="hip"
+ENV DEVITO_PLATFORM="amdgpuX"
+ENV DEVITO_LANGUAGE="hip"
```
mloubout, line 105: Should this be added to the cron CI? 

Also maybe a line in the Readme that it's an option now

docker/Dockerfile.amd
jrmadsen: Omniperf is very similar to Nsight-Compute but the capabilities of Omnitrace far exceed the capabilities of Nsight-Systems. Intel's VTune is closer analogue.

docker/Dockerfile.amd
FabioLuporini: We can and should add it to the cron jobs, I agree. 

We shouldn't isntead write anything in the README until support is complete. This is to be regarded as a dev utility for now

docker/Dockerfile.amd
FabioLuporini: that's what omniperf wants...

docker/Dockerfile.amd
FabioLuporini: Did you see Dockerfile.nvidia?

FabioLuporini: I *think* I should have done the same thing as there but please correct me if I've made a mistake. Testing locally seems fine.

mloubout: arch is correct. This will only be triggered when the final image, defined by arch, is aomp

georgebisbas: Yeah sorry, I literally thought I had deleted this comment as I realised it directly

docker/Dockerfile.amd
georgebisbas: ok

mloubout:
This should be gtg right

FabioLuporini:
I'm waiting for this https://github.com/AMDResearch/omniperf/issues/45#issuecomment-1344741068

I'll tag u in that issue so that you can track progress

mloubout:
Status on this? Still wanted?

FabioLuporini:
@mloubout yes, but some story as Dockerfile.amd -- I want to be able to pull in from something they maintain , especially since the project is young

But we definitely want the AMD profilers in

FabioLuporini:
Just checked -- They have added something recently, but ...

https://github.com/AMDResearch/omniperf/blob/main/docker/ubuntu20.04/Dockerfile

an amdgpu package? unsure..

mloubout:
So what's the plan for this?

FabioLuporini:
HOnestly don't know yet, I have to look at their latest installation instructions to see if it got any simpler

the way it is at the moment is too painful and I wouldn't want to maintain it

That said, having the AMD profiler around in our Docker image would be quite useful...

jrmadsen:
```bash
wget https://github.com/AMDResearch/omnitrace/releases/latest/download/omnitrace-install.py
python3 ./omnitrace-install.py --prefix /opt/omnitrace/rocm-5.4 --rocm 5.4
```

Omnitrace installation is quite straightforward

PR #2033, opened 2022-11-21
compiler: Ensure order invariance of candidates in ReducerMap.unique

ccuetom:
When looking for unique values in `ReducerMap.unique`, the order of candidates alters the result:

- If the candidates are `[1, frozenset({1, 2, 3})]` the method will run as expected.
- If the candidates are `[frozenset({1, 2, 3}), 1]` the method will fail to find a unique value among the candidates

This is fixed by adding another branch to the local function `compare_to_first` checking if `first` is a `Set`.

### DIFF
--- a/devito/tools/data_structures.py
+++ b/devito/tools/data_structures.py
@@ -118,10 +118,7 @@ class ReducerMap(MultiDict):
                 else:
                     return first in v
             elif isinstance(first, Set):
-                if isinstance(v, Set):
-                    return not first.isdisjoint(v)
-                else:
-                    return v in first
+                return v in first
             else:
                 return first == v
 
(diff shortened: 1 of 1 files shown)


mloubout:
Thanks. Don't think need to wrapp in a `comapre_set(v, first) or compare_set(first, v)` 

Could you add a small test for what triggered this?

ccuetom:
@mloubout sorry, not sure I understand what you mean by:

> Don't think need to wrapp in a comapre_set(v, first) or compare_set(first, v)

@FabioLuporini said it would be overkill to add a test for this, but happy either way.

ccuetom:
I think all cases should be covered, as the code would have never visited the branch that I removed, so it would have always ended up in `return v in first`. What cases do you think wouldn't be covered?

mloubout:
Nevermind brain not working today

mloubout (approved):

PR #2038, opened 2022-12-01
misc: Updated contact information in the Code of Conduct.

ggorman:


mloubout:
FYI you can make those organisation wide so that they automatically are in every repo in the organisation.

You just need to create a `.github` repository and put them in it. You can also put a README in that `.gi thub` repo that will show when you are at the orga page (like [here](https://github.com/slimgroup) the About section) see for example our SLIM one https://github.com/slimgroup/.github

mloubout (approved):

PR #2042, opened 2023-01-02
compiler: Extensions for parlang backends

FabioLuporini:
This PR includes:

* Some bug fixes (in particular, see `buffering.py`)
* New classes for new features (e.g., `Bundle` for vector types such as `float2`)
* Several refactorings

PR equipped with tests

### DIFF
--- a/devito/arch/archinfo.py
+++ b/devito/arch/archinfo.py
@@ -378,6 +378,17 @@ def get_cuda_path():
 
 
 @memoized_func
+def get_hip_path():
+    # *** First try: via commonly used environment variables
+    for i in ['HIP_HOME']:
+        hip_home = os.environ.get(i)
+        if hip_home:
+            return hip_home
+
+    return None
+
+
+@memoized_func
 def get_m1_llvm_path(language):
     # Check if Apple's llvm is installed (installable via Homebrew), which supports
     # OpenMP.
--- a/devito/arch/archinfo.py
+++ b/devito/arch/archinfo.py
@@ -15,7 +15,7 @@ from devito.logger import warning
 from devito.tools import as_tuple, all_equal, memoized_func
 
 __all__ = ['platform_registry', 'get_cpu_info', 'get_gpu_info', 'get_nvidia_cc',
-           'get_cuda_path', 'check_cuda_runtime', 'get_m1_llvm_path',
+           'get_cuda_path', 'get_hip_path', 'check_cuda_runtime', 'get_m1_llvm_path',
            'Platform', 'Cpu64', 'Intel64', 'Amd', 'Arm', 'Power', 'Device',
            'NvidiaDevice', 'AmdDevice', 'IntelDevice',
            'INTEL64', 'SNB', 'IVB', 'HSW', 'BDW', 'SKX', 'KNL', 'KNL7210',  # Intel
--- a/devito/builtins/initializers.py
+++ b/devito/builtins/initializers.py
@@ -8,7 +8,7 @@ __all__ = ['assign', 'smooth', 'gaussian_smooth', 'initialize_function']
 
 
 @dv.switchconfig(log_level='ERROR')
-def assign(f, rhs=0, options=None, name='assign', **kwargs):
+def assign(f, rhs=0, options=None, name='assign', domain_only=True, **kwargs):
     """
     Assign a list of RHSs to a list of Functions.
 
@@ -23,6 +23,9 @@ def assign(f, rhs=0, options=None, name='assign', **kwargs):
         be passed to Eq.
     name : str, optional
         Name of the operator.
+    domain_only : bool, optional
+        Restrict the assignment to `f`'s physical domain (True, default) or include
+        the halo as well.
 
     Examples
     --------
@@ -52,6 +55,7 @@ def assign(f, rhs=0, options=None, name='assign', **kwargs):
     """
     if not isinstance(rhs, list):
         rhs = len(as_list(f))*[rhs, ]
+
     eqs = []
     if options:
         for i, j, k in zip(as_list(f), rhs, options):
@@ -62,6 +66,17 @@ def assign(f, rhs=0, options=None, name='assign', **kwargs):
     else:
         for i, j in zip(as_list(f), rhs):
             eqs.append(dv.Eq(i, j))
+
+    if not domain_only:
+        subs = {}
+        for d, h in zip(f.dimensions, f._size_halo):
+            if sum(h) == 0:
+                continue
+            subs[d] = dv.CustomDimension(name=d.name, parent=d,
+                                         symbolic_min=d.symbolic_min - h.left,
+                                         symbolic_max=d.symbolic_max + h.right)
+        eqs = [eq.xreplace(subs) for eq in eqs]
+
     dv.Operator(eqs, name=name, **kwargs)()
 
 
--- a/devito/builtins/initializers.py
+++ b/devito/builtins/initializers.py
@@ -8,7 +8,7 @@ __all__ = ['assign', 'smooth', 'gaussian_smooth', 'initialize_function']
 
 
 @dv.switchconfig(log_level='ERROR')
-def assign(f, rhs=0, options=None, name='assign', domain_only=True, **kwargs):
+def assign(f, rhs=0, options=None, name='assign', assign_halo=False, **kwargs):
     """
     Assign a list of RHSs to a list of Functions.
 
@@ -23,9 +23,9 @@ def assign(f, rhs=0, options=None, name='assign', domain_only=True, **kwargs):
         be passed to Eq.
     name : str, optional
         Name of the operator.
-    domain_only : bool, optional
-        Restrict the assignment to `f`'s physical domain (True, default) or include
-        the halo as well.
+    assign_halo : bool, optional
+        Include the halo region in the assignment (False, default) or restrict it
+        to `f`'s physical domain.
 
     Examples
     --------
@@ -67,7 +67,7 @@ def assign(f, rhs=0, options=None, name='assign', domain_only=True, **kwargs):
         for i, j in zip(as_list(f), rhs):
             eqs.append(dv.Eq(i, j))
 
-    if not domain_only:
+    if assign_halo:
         subs = {}
         for d, h in zip(f.dimensions, f._size_halo):
             if sum(h) == 0:
--- a/devito/ir/iet/nodes.py
+++ b/devito/ir/iet/nodes.py
@@ -60,6 +60,12 @@ class Node(Signer):
     appears in this list are treated as traversable fields.
     """
 
+    _ccode_handler = None
+    """
+    Customizable by subclasses, in particular Operator subclasses which define
+    backend-specific nodes and, su such, require node-specific handlers.
+    """
+
     def __new__(cls, *args, **kwargs):
         obj = super(Node, cls).__new__(cls)
         argnames, _, _, defaultvalues, _, _, _ = inspect.getfullargspec(cls.__init__)
@@ -93,8 +99,11 @@ class Node(Signer):
               from devito.ir.iet import CGen
               CGen().visit(self)
         """
-        from devito.ir.iet.visitors import CGen
-        return CGen().visit(self)
+        try:
+            return self._ccode_handler().visit(self)
+        except (AttributeError, TypeError):
+            from devito.ir.iet.visitors import CGen
+            return CGen().visit(self)
 
     @property
     def view(self):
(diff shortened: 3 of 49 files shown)


devito/ir/iet/nodes.py
```diff
@@ -60,6 +60,12 @@ class Node(Signer):
     appears in this list are treated as traversable fields.
     """
 
+    _ccode_handler = None
+    """
+    Customizable by subclasses, in particular Operator subclasses which define
+    backend-specific nodes and, su such, require node-specific handlers.
```
georgebisbas, line 66: typo

devito/ir/stree/algorithms.py
```diff
@@ -146,9 +146,8 @@ def stree_make_halo(stree):
             spot = k
             ancestors = [n for n in k.ancestors if n.is_Iteration]
             for n in ancestors:
-                # Place the halo exchange right before the first
-                # distributed Dimension which requires it
-                if any(i.dim in n.dim._defines for i in v.halos):
+                # Place the halo exchange before the first distributed Dimension
+                if n.dim._defines.intersection(f._dist_dimensions):
```
georgebisbas, line 150: ok

FabioLuporini:
@mloubout @georgebisbas and readers

Note that part of the reviewing is in #2043 , which is essentially this PR plus a few edits from @ggorman

devito/ir/iet/nodes.py
FabioLuporini: noted, will fix

FabioLuporini: ... in my upcoming PR, if you don't mind, as apparently no other changes are requested here given yours and @mloubout 's comments

devito/arch/archinfo.py
```diff
@@ -377,6 +377,17 @@ def get_cuda_path():
     return None
 
 
+@memoized_func
+def get_hip_path():
+    # *** First try: via commonly used environment variables
```
georgebisbas, line 382: But there is no other try yet, right?

devito/ir/iet/nodes.py
georgebisbas: ok

devito/operator/operator.py
```diff
@@ -975,6 +962,34 @@ def __setstate__(self, state):
...
+    else:
+        cls = operator_selector(**kwargs)
+
+    # Tweak the compilation kwargs
+    options = dict(kwargs['options'])
+    # NOTE: it is not only pointless to apply the following passes recursively
+    # (because once, during the main compilation phase, is simply enough), but
+    # also dangerous as a handful of compiler passes, the minority, might break
+    # in some circumstances if applied in cascade (e.g., `linearization` on top
+    # of `linearization`)
```
georgebisbas, line 982: good

examples/seismic/model.py
```diff
@@ -133,7 +133,8 @@ def _initialize_bcs(self, bcs="damp"):
         # First initialization
         init = self.damp is None
         # Get current Function if alread yinitialized
```
georgebisbas, line 135: typo but can be done in other PR

georgebisbas (approved):
gtg, left a few minor comments

devito/arch/archinfo.py
FabioLuporini: no, but akin to get_nvidia_path. We might improve it in the future

examples/seismic/model.py
FabioLuporini: oki

devito/ir/support/guards.py
```diff
@@ -136,12 +136,12 @@ def __new__(cls, d, direction, **kwargs):
...
-                # values on the floor. E.g., `iteration=time - 1`, `v=2`, then when
-                # `time=0` we want the Mul to evaluate to -1, not to 0, which is
-                # what C's integer division would give us
-                p1 = Mul(((p1 - 1) / FLOAT(v)), v, evaluate=False)
+                # Round `p1 - 1` down to the nearest sub-multiple of `v`
+                # NOTE: we use ABS to make sure we handle negative values properly.
+                # Once `p1 - 1` is negative (e.g. `iteration=time - 1` and `time=0`),
+                # as long as we get a negative number, rather than 0 and even if it's
+                # not `-v`, we're good
+                p1 = (p1 - 1) - abs(p1 - 1) % v
```
mloubout, line 144: But that's not a multiple of `v` if `p1-1` is negative, it should be 
`p1 = (p1 - 1) - sign(p1 - 1) * (abs(p1 - 1) % v)`
to always be a multiple of v irrespective of sign.

devito/operator/operator.py
```diff
@@ -663,8 +647,11 @@ def _soname(self):
 
     @cached_property
     def ccode(self):
-        from devito.ir.iet.visitors import CGen
-        return CGen(compiler=self._compiler).visit(self)
+        try:
+            return self._ccode_handler(compiler=self._compiler).visit(self)
+        except (AttributeError, TypeError):
+            from devito.ir.iet.visitors import CGen
+            return CGen(compiler=self._compiler).visit(self)
```
mloubout, line 654: Note for me from other PR, cannot be made class default due to circular imports.

devito/symbolics/extended_sympy.py
```diff
@@ -544,6 +548,9 @@ def __str__(self):
 
     __repr__ = __str__
 
+    def _sympystr(self, printer):
```
mloubout, line 551: What's that?

devito/types/array.py
```diff
@@ -345,3 +346,217 @@ def dim(self):
...
+    @property
+    def is_shared(self):
+        if self.c0.is_Array:
+            return self.c0._is_shared
+        else:
+            return False
+
+    # CodeSymbol overrides defaulting to self.c0's behaviour
+
+    for i in ['_mem_internal_eager', '_mem_internal_lazy', '_mem_local',
```
mloubout, line 468: Why not `getattr` default like in Differentiable or Tensor? Would avoid keeping track of all the function that needs to default to `c0`

devito/types/array.py
```diff
@@ -345,3 +346,217 @@ def dim(self):
...
+    def _C_ctype(self):
+        if self._mem_mapped:
+            return ArrayMapped._C_ctype
+        else:
+            return POINTER(dtype_to_ctype(self.dtype))
+
+
+class ComponentAccess(Expr, Reconstructable):
+
+    _component_names = ('x', 'y', 'z', 'w')
```
mloubout, line 512: `w`?

devito/types/array.py
```diff
@@ -345,3 +346,217 @@ def dim(self):
...
+        return super()._hashable_content() + (self._index,)
+
+    def __str__(self):
+        return "%s.%s" % (self.base, self.sindex)
+
+    __repr__ = __str__
+
+    func = Reconstructable._rebuild
+
+    def _sympystr(self, printer):
```
mloubout, line 537: Still not sure why need this, everything in sympy goes through `_print` which should be safer to use no?

devito/types/array.py
```diff
@@ -345,3 +346,217 @@ def dim(self):
     @property
     def array(self):
         return self._array
+
+
+class Bundle(ArrayBasic):
```
mloubout, line 351: So how different is it com[ared to a VectorFunction? Seem to abstract the same type of object.

mloubout (commented):
Left some comments but looks mostly good

devito/builtins/initializers.py
```diff
@@ -23,6 +23,9 @@ def assign(f, rhs=0, options=None, name='assign', **kwargs):
         be passed to Eq.
     name : str, optional
         Name of the operator.
+    domain_only : bool, optional
```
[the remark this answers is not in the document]
FabioLuporini: Done

devito/ir/support/basic.py
```diff
@@ -65,7 +65,11 @@ class IterationInstance(LabeledVector):
     """
 
     def __new__(cls, access):
-        findices = tuple(access.function.dimensions)
+        try:
+            findices = tuple(access.function.dimensions)
+        except AttributeError:
+            # E.g., Objects, which don't have `dimensions`
```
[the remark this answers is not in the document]
FabioLuporini: No, I think it's correct. See below:

```
In [7]: Object(name='a', dtype=np.int32)
Out[7]: a

In [8]: Object(name='a', dtype=np.int32).function
Out[8]: a

In [10]: type(Object(name='a', dtype=np.int32).function)
Out[10]: devito.types.object.Object

In [11]: Object(name='a', dtype=np.int32).function.dimensions
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In [11], line 1
----> 1 Object(name='a', dtype=np.int32).function.dimensions

AttributeError: 'Object' object has no attribute 'dimensions'
```

devito/ir/support/guards.py
FabioLuporini: yes, but the comment above explains why we can tolerate it

I agree it's far from ideal, but in practice it always works (unless we start allowing time dimensions in negative ranges...)

The reason I haven't done it like that is because I don't know how to neatly turn `sign` into C code. It would require a non trivial addition to the compiler, basically a further compiler pass intercepting "high order functions", such as `sign`, that don't have a one-to-one matching with routines in clib (math.h)

If we wanted explicit, inline lowering, we could use one of these forms https://stackoverflow.com/questions/1903954/is-there-a-standard-sign-function-signum-sgn-in-c-c
But it feels like it would clutter the generated code...

Unless I'm missing the obvious, I'd prefer to open an issue rather than tackling this now

devito/symbolics/extended_sympy.py
FabioLuporini: eheh

SymPy dark voodoo magic :)

try a `git grep _sympystr` in sympy, you will see

basically a magic method that one needs to override to make sure Basic subclasses emit their actual `repr` when part of a larger expression

E.g.

Say you have an `IntDiv(a, 4)`. When you are in the terminal, you want to see `a / 4` (or `a // 4`, whatever), not `IntDiv(a, 4)`. This works in master... but only if not part of a larger expression. For example, we would see `IntDiv(a, 4) + 3`. Overriding `_sympystr` allows you to display `(a / 4) + 3`. See the tests that I've added

devito/types/array.py
FabioLuporini: CUDA/HIP convention for float4 types

devito/types/array.py
FabioLuporini: see comment above, `_print` is only when you use `str(..)` basically

devito/types/array.py
FabioLuporini: Initially I tried to use a VectorFunction, but I encountered several issues... I also tried to subclass it, but no luck. I honestly don't remember what the actual culprit was, but basically I would have had to subclass VectorFunction and override nearly all of its methods, and even then, such subclass would still appear as a user-provided object to the eyes of the compiler, which is not ideal

As an Array resembles a DiscreteFunction but it's not a DiscreteFunction, a Bundle resembles a VectorFunction but it's not a VectorFunction

I'm not entirely convinced this is the best course of action (so u may have a point), but this is where I landed after a non negligible amount of thinking and trial-and-error

devito/types/array.py
FabioLuporini: correct me if I'm wrong, but IIRC `__getattr__` gets called only when an attribute with the given name could not be found. Here the attributes exist, as they are defined in a parent superclass (`CodeSymbol`), so we need explicit overriding... no?

examples/seismic/model.py
FabioLuporini: fixed now since I had to push again anyway

devito/ir/iet/nodes.py
FabioLuporini: fixed now as I had to push anyway

devito/ir/support/guards.py
mloubout: Sign is a defined function we use it for the free surface for example I'm the acoustic propagator there is nothing needed in the compiler

devito/ir/support/basic.py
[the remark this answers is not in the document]
mloubout: Ok

devito/types/array.py
mloubout: Hum now I'm in doubt thought was always called  you probably right

devito/types/array.py
mloubout: I guess the atmpy matrix internals are too much for it ok.

devito/ir/support/guards.py
FabioLuporini: Sorry, I forgot to copy-paste what I wrote on Slack here. Here it is:

It crashes in (at least) one case
So the guard that gets generated lets the execution flow pass through at `time = [4, 2, 0]` while the expectation is to let it go through at `time = [4, 2]`

PR #2043, opened 2023-01-02
Parlang extensions ghacks

ggorman:
Testing tweaks to @FabioLuporini's PR

### DIFF
--- a/.github/workflows/docker-devito.yml
+++ b/.github/workflows/docker-devito.yml
@@ -33,7 +33,7 @@ jobs:
           # Runtime gpu flags from https://hub.docker.com/r/rocm/tensorflow/
           - base: 'bases:amd'
             tag: 'amd'
-            flag: '--network=host --device=/dev/kfd --device=/dev/dri --ipc=host --group-add video'
+            flag: '--network=host --device=/dev/kfd --device=/dev/dri --ipc=host --group-add video --group-add 109 --cap-add=SYS_PTRACE --security-opt seccomp=unconfined'
             test: 'tests/test_gpu_openmp.py'
             runner: ["self-hosted", "amdgpu"]
 
--- a/devito/builtins/initializers.py
+++ b/devito/builtins/initializers.py
@@ -8,7 +8,7 @@ __all__ = ['assign', 'smooth', 'gaussian_smooth', 'initialize_function']
 
 
 @dv.switchconfig(log_level='ERROR')
-def assign(f, rhs=0, options=None, name='assign', **kwargs):
+def assign(f, rhs=0, options=None, name='assign', domain_only=True, **kwargs):
     """
     Assign a list of RHSs to a list of Functions.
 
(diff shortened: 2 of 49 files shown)


devito/operator/operator.py
```diff
@@ -663,8 +647,11 @@ def _soname(self):
 
     @cached_property
     def ccode(self):
-        from devito.ir.iet.visitors import CGen
-        return CGen(compiler=self._compiler).visit(self)
+        try:
+            return self._ccode_handler(compiler=self._compiler).visit(self)
```
mloubout, line 651: So this is the door for custom operators in pro/... right? Wouldn't making `CGen` the default instead of `None` avoid this try block?

devito/operator/registry.py
```diff
@@ -26,7 +26,7 @@ class OperatorRegistry(OrderedDict, metaclass=Singleton):
     """
 
     _modes = ('noop', 'advanced', 'advanced-fsg')
-    _languages = ('C', 'openmp', 'openacc', 'cuda')
+    _languages = ('C', 'openmp', 'openacc', 'cuda', 'hip')
```
mloubout, line 29: So what does it actually do if you use this without pro? Just the standard cpup code but compiled with HIP?

devito/builtins/initializers.py
```diff
@@ -62,6 +66,17 @@ def assign(f, rhs=0, options=None, name='assign', **kwargs):
     else:
         for i, j in zip(as_list(f), rhs):
             eqs.append(dv.Eq(i, j))
+
+    if not domain_only:
+        subs = {}
+        for d, h in zip(f.dimensions, f._size_halo):
+            if sum(h) == 0:
+                continue
+            subs[d] = dv.CustomDimension(name=d.name, parent=d,
```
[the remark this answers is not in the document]
FabioLuporini: no, because remember those are global values, so it will break with MPI

Secondly, I wanted to abstract that complexity ("how much should x_m, x_M, y_m ... be?") away

devito/operator/registry.py
FabioLuporini: no you will get an InvalidOperator exception as soon as you try to build an Operaator

devito/operator/operator.py
FabioLuporini: it catches both an AttributeError and a TypeError but I don't remember exactly why, to be fair. My bad. I will see if I can improve it

.github/workflows/docker-devito.yml
```diff
@@ -33,7 +33,7 @@ jobs:
           # Runtime gpu flags from https://hub.docker.com/r/rocm/tensorflow/
           - base: 'bases:amd'
             tag: 'amd'
-            flag: '--network=host --device=/dev/kfd --device=/dev/dri --ipc=host --group-add video'
+            flag: '--network=host --device=/dev/kfd --device=/dev/dri --ipc=host --group-add video --group-add 109 --cap-add=SYS_PTRACE --security-opt seccomp=unconfined'
```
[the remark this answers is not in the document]
FabioLuporini: yes

mloubout:
#2042 merged, closing

PR #2045, opened 2023-01-04
compiler: add cluster guard to AliasKey for safety

mloubout:


georgebisbas (approved):

georgebisbas (commented):
GTG, minor question:
Is it worth adding a result check? i.e. run the op for time_M=x and check?

FabioLuporini (approved):
Uncontroversial. Thank you

PR #2050, opened 2023-01-23
compiler: Fix placement of ConditionalDimension in subdomain

georgebisbas:


devito/ir/clusters/algorithms.py
```diff
@@ -206,7 +206,13 @@ def guard(clusters):
...
                 else:
                     dims = pull_dims(cd.condition)
-                    k = max(dims, default=d, key=lambda i: c.ispace.index(i))
+                    # If `cd` uses more dimensions than the ispace,
+                    # stay under parent
+                    if (not dims.issubset(set(c.ispace.dimensions)) and
+                       cd.parent in dims):
+                        k = cd.parent
+                    else:
+                        k = max(dims, default=d, key=lambda i: c.ispace.index(i))
```
georgebisbas, line 215: Solution 2 with False flag

```
                # If `cd` uses, as condition, an arbitrary SymPy expression, then
                # we must ensure to nest it inside the last of the Dimensions
                # appearing in `cd.condition`
                if cd._factor is not None:
                    k = d
                else:
                    dims = pull_dims(cd.condition, flag=False)
                    k = max(dims, default=d, key=lambda i: c.ispace.index(i))
```

tests/test_subdomains.py
```diff
@@ -628,3 +628,101 @@ class Dummy(SubDomainSet):
...
+        ci = ConditionalDimension(name='ci', condition=condition,
+                                  parent=mid.dimensions[0])
+
+        op = Operator(Eq(f, f + 10, implicit_dims=ci,
+                      subdomain=my_grid.subdomains['middle']))
+        op.apply()
+
+        assert_structure(op, ['i1x'], 'i1x')
+
+    def test_condition_w_subdomain_II(self):
```
FabioLuporini, line 666: usually we use v0, v1, v2... instead of I, II, III

tests/test_subdomains.py
```diff
@@ -628,3 +628,101 @@ class Dummy(SubDomainSet):
...
+
+        f = Function(name='f', grid=my_grid)
+
+        sdf = Function(name='sdf', grid=my_grid)
+        sdf.data[5:] = 1
+
+        condition = Lt(sdf[mid.dimensions[0]], 1)
+
+        ci = ConditionalDimension(name='ci', condition=condition,
+                                  parent=mid.dimensions[0])
```
FabioLuporini, line 658: mid.dimensions[0] is a root dimension already, do we actually need this test? 

IOW, would this test actually fail in current master?

FabioLuporini (commented):
Do the new tests actually fail under master?

georgebisbas:
>Do the new tests actually fail under master?

Yes, they fail in the current master with the python-land exception
(I guess the same you think of here: https://devitocodes.slack.com/archives/C7JMLMSG0/p1674301184495359?thread_ts=1674231704.845469&cid=C7JMLMSG0)

tests/test_subdomains.py
georgebisbas: https://github.com/devitocodes/devito/pull/2050#issuecomment-1408497079

georgebisbas:
@FabioLuporini plans for this?

FabioLuporini:
@georgebisbas plan imho is to find a neater/simpler/better way

devito/ir/clusters/algorithms.py
georgebisbas: Do you also consider the second solution not good enough?

FabioLuporini: It's not clear _why_ passing `flag=False` would fix the issue here?

mloubout:
What's the status on this?

georgebisbas:
@FabioLuporini requested a better solution on this.
Any ides?

mloubout:
As per my comment above, I don't think it's an issue anymore and should work without the change

georgebisbas:
@mloubout I cannot see any comment of yours (?)

pytest tests/test_subdomains.py::TestSubDomain_w_condition

I tried to run the tests and they fail

FabioLuporini (approved):
This is now good to go imho

EdCaunt (approved):

mloubout (commented):
Much cleaner than before, one minro comment

devito/ir/clusters/algorithms.py
```diff
@@ -238,7 +238,7 @@ def guard(clusters):
                 if cd._factor is not None:
                     k = d
                 else:
-                    dims = pull_dims(cd.condition)
+                    dims = pull_dims(cd.condition, flag=False)
```
mloubout, line 241: I think this should be `flag=cd.indirect` for the case where the conditional is used as the indexing dimension

tests/test_subdomains.py
```diff
@@ -693,3 +693,101 @@ class Dummy(SubDomainSet):
...
+
+        f = Function(name='f', grid=my_grid)
+        op = Operator(Eq(f, f + 10, implicit_dims=ci,
+                      subdomain=my_grid.subdomains['middle']))
+
+        op.apply()
+
+        assert_structure(op, ['xy'], 'xy')
+
+    def test_condition_w_subdomain_v2(self):
```
georgebisbas, line 763: @mloubout could you help sketching a test about what you mention as the conditional being the indexing dimension?
or at least show me an example with some pseducode?

mloubout: Ok sorry after bit of digging, forgot this algorithm section is only for non indirect (see line 223 in algorithms.py)

GTG

mloubout (approved):

tests/test_subdomains.py
georgebisbas: ah right, this is what happens if I have long time to see a PR

PR #2051, opened 2023-01-23
arch: Add ICX support

georgebisbas:
- Add ICX support : https://www.intel.com/content/www/us/en/developer/articles/guide/porting-guide-for-icc-users-to-dpcpp-or-icx.html
- Add icx docker image (same with icc one)
- Adds platforms SKX, KLX, CLC, CLK (@FabioLuporini )
- Adds a condition option for switching test confs
- Drops support for icc version<15


devito/arch/compiler.py
```diff
@@ -501,6 +501,22 @@ def __lookup_cmds__(self):
         self.MPICXX = 'mpicxx'
 
 
+class ICXCompiler(Compiler):
```
mloubout, line 504: Shouldn't this be an IntelCompiler

mloubout: Did you try it? How does it compare with ICC in perf?

georgebisbas: I wrote it in a similar way to DPCPP. So as not inheriting from Intel.

Did not see any perf advantage on a few acoustic/tti problems locally. 
icc is gonna be deprecated soon and icx will be the default from Intel

FabioLuporini: I agree that not inheriting from either IntelCompiler or ClangCompiler is a bit odd, but I can live with it for now, I think...

georgebisbas: Conceptually I would agree, but there is not much overlap in flags etc...

mloubout (commented):
Could we at least add it to the icc-cpu docker CI

devito/arch/compiler.py
```diff
@@ -528,6 +528,58 @@ def __lookup_cmds__(self):
...
+
+        if platform is INTELGPUX:
+            self.cflags.remove('-g')  # -g disables some optimizations in IGC
+            self.cflags.append('-gline-tables-only')
+            self.cflags.append('-fdebug-info-for-profiling')
+
+        # Make sure the MPI compiler uses `icx` underneath -- whatever the MPI distro is
+        if kwargs.get('mpi'):
+            ver = check_output([self.MPICC, "--version"]).decode("utf-8")
+            if not ver.startswith("Intel(R) oneAPI"):
```
FabioLuporini, line 570: No here we should rather use `sniff_mpi_distro` (which potentially needs to be updated by adding "ver.startswith(Intel oneAPI "...")

And it's not just to avoid code duplication, but also because sniff_mpi_distro is memoized

docker/Dockerfile.cpu
```diff
@@ -55,6 +55,32 @@ ENV DEVITO_LANGUAGE="openmp"
 # MPICC compiler for mpi4py
 ENV MPICC=$I_MPI_ROOT/bin/mpiicc
 
+##############################################################
+# ICX image
+##############################################################
+FROM base as icx
+
+ # Download the key to system keyring
```
mloubout, line 63: This whole part is a waste , just do `FROM icc as icx` and just set the DEVITO env vars

tests/conftest.py
```diff
@@ -69,6 +71,12 @@ def skipif(items, whole_module=False):
            isinstance(configuration['platform'], Cpu64):
             skipit = "`icc+cpu64` won't work with this test"
             break
+        # Skip if it won't run with OneAPICompiler
+        if i == 'cpu64-icpx' and \
```
mloubout, line 75: I don't think we need that, if everything is a proper `IntelCompiler` then we just need `cpu64-intel`

mloubout (commented):
I still don't like having separate OneApi and Intel compiler

tests/conftest.py
georgebisbas: Can do that, but we have tests that, at least temporarily, pass with ICC but not with ICX.
What you propose should be fine with the next release and at some point, we are gonna drop ICC entirely.
I do not have any string preference

devito/arch/compiler.py
```diff
@@ -528,6 +528,58 @@ def __lookup_cmds__(self):
...
+
+        if language == 'sycl':
+            self.cflags.append('-fsycl')
+            if platform is NVIDIAX:
+                self.cflags.append('-fsycl-targets=nvptx64-cuda')
+            else:
+                self.cflags.append('-fsycl-targets=spir64')
+
+        if language == 'openmp':
+            self.cflags.append('-fopenmp')
```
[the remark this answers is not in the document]
georgebisbas: added

tests/test_dimension.py
```diff
@@ -1336,6 +1336,8 @@ def test_affiness(self):
         iterations = [i for i in FindNodes(Iteration).visit(op) if i.dim is not time]
         assert all(i.is_Affine for i in iterations)
 
+    # Skipping this test with icx, as it requires safe-math
```
FabioLuporini, line 1339: You just gave me an idea

We could rather extend `switchconfig` to swallow an additional parameter, let's call it for example `condition`. Example:

```
@switchconfig(condition=lambda i: isinstance(configuration['compiler'], OneapiCompiler),
                           safe_math=True)
```

This way we wouldn't have to skip the test

FabioLuporini (commented):
Let's try taking this to completion

tests/test_dimension.py
georgebisbas: ok

.github/workflows/pytest-core-nompi.yml
georgebisbas, line 1: Should I rename to intel-icc, intel-icx?

.github/workflows/pytest-core-nompi.yml
```diff
@@ -121,21 +129,26 @@ jobs:
...
         python-version: ${{ matrix.python-version }}
 
-    - name: Build docker image
-      if: contains(matrix.name, 'docker')
+    - name: Build docker image for GCC variants
+      if: "contains(matrix.name, 'docker') && contains(matrix.name, 'gcc')"
       run: |
           docker build . --file docker/Dockerfile.devito --tag devito_img --build-arg base=devitocodes/bases:cpu-${{ matrix.arch }}
 
+    - name: Build docker image for INTEL variants
```
mloubout, line 137: Don't really like to duplicate that step makes a lot of "single case" steps which arent' really good

FabioLuporini (changes requested):

devito/arch/compiler.py
```diff
@@ -793,6 +847,8 @@ def __lookup_cmds__(self):
     'intel': IntelCompiler,
     'icpc': IntelCompiler,
     'icc': IntelCompiler,
```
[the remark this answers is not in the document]
georgebisbas: ok

tests/test_dimension.py
georgebisbas: ok

tests/test_dimension.py
```diff
@@ -1382,6 +1383,9 @@ def test_affiness(self):
...
+    @switchconfig(condition=isinstance(configuration['compiler'],
+                  OneapiCompiler), safe_math=True)
```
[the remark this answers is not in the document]
georgebisbas: So you mean whether icpc also fails? e.g. we have more than on intel compiler mapping to ONEAPI Compiler...are you asking whether all of them fail, or whether I give the correct arg to the condition?

.github/workflows/pytest-core-nompi.yml
georgebisbas: >Maybe we can add the tags icc/icx to the dovcker build CI so we only build it once but add aliases for easy pull


For some reason I had not seen this command, indeed it solves the problems stated form you and Fabio

devito/parameters.py
```diff
@@ -241,6 +241,14 @@ def __init__(self, **params):
     def __call__(self, func, *args, **kwargs):
         @wraps(func)
         def wrapper(*args, **kwargs):
+            # Do not switch if condition is False
```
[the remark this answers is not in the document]
georgebisbas: Yes

.github/workflows/pytest-core-nompi.yml
georgebisbas: οκ

.github/workflows/pytest-core-nompi.yml
```diff
@@ -124,18 +135,18 @@ jobs:
...
       if: contains(matrix.name, 'docker')
       run: |
-          docker build . --file docker/Dockerfile.devito --tag devito_img --build-arg base=devitocodes/bases:cpu-${{ matrix.arch }}
+          docker build . --file docker/Dockerfile.devito --tag devito_img --build-arg base=devitocodes/bases:cpu-${{ matrix.tag }}
 
     - name: Set run prefix
       run: |
           if [[ "${{ matrix.name }}" =~ "docker" ]]; then
-              echo "RUN_CMD=docker run --rm -e CODECOV_TOKEN=${{ secrets.CODECOV_TOKEN }} --name testrun devito_img"  >> $GITHUB_ENV
+              echo "RUN_CMD=docker run --rm -e CODECOV_TOKEN=${{ secrets.CODECOV_TOKEN }} -e DEVITO_ARCH=${{ matrix.arch }} --name testrun devito_img"  >> $GITHUB_ENV
```
FabioLuporini, line 143: no, this is (has to be) encoded in the Dockerfile already

devito/arch/compiler.py
```diff
@@ -692,11 +745,12 @@ def __init__(self, *args, **kwargs):
                 # Append the OpenMP flag regardless of configuration['language'],
                 # since icc15 and later versions implement OpenMP 4.0, hence
                 # they support `#pragma omp simd`
+                # This class of IntelCompiler, will be dropped in future releases
+                # in favour of the OneApi IntelCompiler
```
FabioLuporini, line 749: OneapiCompiler

devito/parameters.py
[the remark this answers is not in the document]
FabioLuporini: Can you convince me what I proposed doesn't really work? I can't believe we need five lines to express one trivial if-not-condition-then-return

tests/conftest.py
```diff
@@ -65,7 +66,7 @@ def skipif(items, whole_module=False):
             break
         # Skip if it won't run with IntelCompiler
         if i == 'cpu64-icc' and \
-           isinstance(configuration['compiler'], IntelCompiler) and \
+           isinstance(configuration['compiler'], (IntelCompiler, OneapiCompiler)) and \
```
FabioLuporini, line 69: This wouldn't be needed if OneapiCompiler were a subclass of IntelCompiler as I think Mathias pointed out at some point, but yeah, nitpicking

devito/arch/compiler.py
```diff
@@ -528,6 +528,60 @@ def __lookup_cmds__(self):
...
+        language = kwargs.pop('language', configuration['language'])
+        platform = kwargs.pop('platform', configuration['platform'])
+
+        self.cflags.append("-xHost")
+        self.cflags.append("-qopt-zmm-usage=high")
+
+        if configuration['safe-math']:
+            self.cflags.append("-fp-model=strict")
+        else:
+            self.cflags.append('-fast')
```
FabioLuporini, line 545: -fp-model=fast like in IntelCompiler? or just -fast ?

FabioLuporini (commented):
@georgebisbas (@mloubout ) in retrospect I don't see why OneapiCompiler shouldn't be a subclass of IntelCompiler

.github/workflows/pytest-core-nompi.yml
mloubout: Well no the dockerfile and image are exactly the same for `icc` and `icx` with the exception of this single env var so it's quite a pain to have to rebuild a whole image in the build CI just to change this one var that can be set at runtime easily like that.

georgebisbas: I think I can consider this resolved, as we only want to build on image and use this env var to distinguish between ICC and icx

devito/arch/compiler.py
georgebisbas: -fast here works fine, should be the same

devito/arch/compiler.py
georgebisbas: But it is an IntelCompiler class so far

georgebisbas: Gotcha

devito/parameters.py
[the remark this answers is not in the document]
georgebisbas: The key for your apporach to work is to set by default condition to None.
We actually have 3 states of condition (No condition, True , False).
'No condition' should not have the same result with a False condition.
With this change we can indeed write in 2 lines

.github/workflows/pytest-core-nompi.yml
```diff
@@ -95,13 +96,23 @@ jobs:
...
           arch: "icc"
+          tag: "icc"
+          language: "openmp"
+          sympy: "1.11"
+
+        - name: pytest-docker-py38-icx-omp
+          python-version: '3.8'
+          os: ubuntu-22.04
+          arch: "icx"
+          tag: "icc"
```
mloubout, line 115: I would just add the `icx` tag to the docker build (i,e just push the ICC image with icx tag) to avoid those

devito/arch/compiler.py
```diff
@@ -737,6 +737,54 @@ def __init__(self, *args, **kwargs):
...
+            if platform is INTELGPUX:
+                self.cflags.append('-fopenmp-targets=spir64')
+                self.cflags.append('-fopenmp-target-simd')
+
+        if platform is INTELGPUX:
+            self.cflags.remove('-g')  # -g disables some optimizations in IGC
+            self.cflags.append('-gline-tables-only')
+            self.cflags.append('-fdebug-info-for-profiling')
+
+        # Make sure the MPI compiler uses `icx` underneath -- whatever the MPI distro is
```
mloubout, line 772: Get his through `super`

devito/arch/compiler.py
```diff
@@ -737,6 +737,54 @@ def __init__(self, *args, **kwargs):
...
+
+        if language == 'sycl':
+            self.cflags.append('-fsycl')
+            if platform is NVIDIAX:
+                self.cflags.append('-fsycl-targets=nvptx64-cuda')
+            else:
+                self.cflags.append('-fsycl-targets=spir64')
+
+        if language == 'openmp':
+            # TODO: To be switched to `-fiopenmp` or `-qopenmp` as soon as
```
mloubout, line 758: `super` then `remove(`-qopenmp`)

devito/arch/compiler.py
```diff
@@ -737,6 +737,54 @@ def __init__(self, *args, **kwargs):
...
 
+class OneapiCompiler(IntelCompiler):
+
+    def __init__(self, *args, **kwargs):
+        super(OneapiCompiler, self).__init__(*args, **kwargs)
+
+        language = kwargs.pop('language', configuration['language'])
+        platform = kwargs.pop('platform', configuration['platform'])
+
+        self.cflags.append("-qopt-zmm-usage=high")
```
mloubout, line 748: That's oddly specific, this works on any platform?

FabioLuporini: no, indeed. 

Can you please `git cherry-pick` (and adjust?) this commit of mine from #2114 ? 
https://github.com/devitocodes/devito/pull/2114/commits/4afe4f849027d4720af19444a115a6e96fb84ae5

devito/arch/compiler.py
FabioLuporini: indeed, otherwise what's the point of inhereting

devito/parameters.py
```diff
@@ -236,11 +236,15 @@ class switchconfig(object):
     """
 
     def __init__(self, **params):
```
FabioLuporini, line 238: so what's the error trace if `condition=None` were part of the signature instead? Apologies if I had already asked ...

georgebisbas: https://devitocodes.slack.com/archives/C02LA241E57/p1683717242731189?thread_ts=1683714842.075059&cid=C02LA241E57

devito/arch/compiler.py
georgebisbas: I do not think it is needed anymore

tests/conftest.py
```diff
@@ -68,6 +70,12 @@ def skipif(items, whole_module=False):
            isinstance(configuration['platform'], Cpu64):
             skipit = "`icc+cpu64` won't work with this test"
             break
+        # Skip if it won't run with OneAPICompiler
+        if i == 'cpu64-icx' and \
```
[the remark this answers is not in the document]
FabioLuporini: Out of curiosity, not even with `switchconfig(condition, safe_math=True)`

devito/parameters.py
FabioLuporini: I've added comments on Slack at the thread above

devito/arch/compiler.py
FabioLuporini: ping

also, rewrite warning msg to drop the "DPC++" thing

.github/workflows/pytest-core-nompi.yml
FabioLuporini: yes I agree we don't want an extra field like "tag" here (useless complexity)

devito/arch/compiler.py
```diff
@@ -375,13 +376,22 @@ class GNUCompiler(Compiler):
...
         if configuration['safe-math']:
             self.cflags.append('-fno-unsafe-math-optimizations')
         else:
             self.cflags.append('-ffast-math')
 
+        if isinstance(platform, IntelSkylake):
+            # The default is `=256` because avx512 slows down the CPU frequency;
+            # however, we empirically found that stencils generally benefit
+            # from `=512`
+            self.cflags.append('-mprefer-vector-width=512')
```
FabioLuporini, line 393: @jkwashbourne we're adding this to gcc's default compilation flags, so, once this PR lands, should u see regressions, please let us know. I've only seen improvements so far (though only up to 5%)

devito/arch/compiler.py
```diff
@@ -729,14 +730,60 @@ class IntelKNLCompiler(IntelCompiler):
     def __init__(self, *args, **kwargs):
         super(IntelKNLCompiler, self).__init__(*args, **kwargs)
 
-        self.cflags += ["-xMIC-AVX512"]
+        self.cflags.append('-xMIC-AVX512')
```
FabioLuporini, line 733: good

devito/arch/compiler.py
```diff
@@ -729,14 +730,60 @@ class IntelKNLCompiler(IntelCompiler):
...
         language = kwargs.pop('language', configuration['language'])
 
         if language != 'openmp':
             warning("Running on Intel KNL without OpenMP is highly discouraged")
 
 
+class OneapiCompiler(IntelCompiler):
+
+    def __init__(self, *args, **kwargs):
+        super(OneapiCompiler, self).__init__(*args, **kwargs)
```
FabioLuporini, line 744: Instead of 

```
super(OneapiCompiler, self).__init__(*args, **kwargs)
```

the following suffices:

```
super().__init__(*args, **kwargs)
```

Updating all instances of this pattern throughout the compiler, incrementally when I see them...

devito/arch/compiler.py
```diff
@@ -729,14 +730,60 @@ class IntelKNLCompiler(IntelCompiler):
...
+
+        platform = kwargs.pop('platform', configuration['platform'])
+        language = kwargs.pop('language', configuration['language'])
+
+        if language == 'openmp':
+            self.ldflags.remove('-qopenmp')
+            self.ldflags.append('-fopenmp')
```
FabioLuporini, line 751: uff, really? does ICC support -fopenmp instead of -qopenmp (or perhaps both?)

FabioLuporini (commented):
I think we're finally converging

.github/workflows/pytest-core-nompi.yml
georgebisbas: @mloubout ?

devito/arch/compiler.py
georgebisbas: ok

devito/arch/compiler.py
georgebisbas: Typically it supports both, but until this fix we discussed through mail is in, they will not work.
This can be temporary and will be dropped after we also drop intel compiler

devito/arch/compiler.py
```diff
@@ -671,37 +681,28 @@ def __lookup_cmds__(self):
 class IntelCompiler(Compiler):
 
     def __init__(self, *args, **kwargs):
-        super(IntelCompiler, self).__init__(*args, **kwargs)
-
-        self.cflags.append("-xhost")
+        super().__init__(*args, **kwargs)
 
-        language = kwargs.pop('language', configuration['language'])
+        self.cflags.append("-xHost")
```
FabioLuporini, line 686: ideally this line goes after pulling platform and language, but we can address it in another PR if it can spare a useless CI spin

devito/arch/compiler.py
```diff
@@ -727,16 +728,62 @@ def __lookup_cmds__(self):
...
+                self.cflags.append('-fopenmp-targets=spir64')
+                self.cflags.append('-fopenmp-target-simd')
+
+        if platform is INTELGPUX:
+            self.cflags.remove('-g')  # -g disables some optimizations in IGC
+            self.cflags.append('-gline-tables-only')
+            self.cflags.append('-fdebug-info-for-profiling')
+
+        # Make sure the MPI compiler uses `icx` underneath -- whatever the MPI distro is
+        if kwargs.get('mpi'):
```
FabioLuporini, line 772: Wait, can we now not drop this? Why was the conversation marked as "resolved" ? We should now be getting the MPI warning via super()?

tests/test_buffering.py
```diff
@@ -701,7 +701,6 @@ def test_everything():
     assert np.all(u.data == u1.data)
 
 
-@skipif('cpu64-icc')
```
FabioLuporini, line 704: it somehow works now?

devito/arch/compiler.py
georgebisbas: ok

tests/test_buffering.py
georgebisbas: Yes!

FabioLuporini (approved):
OK, thanks for the patience. This is now for me good to go.

mloubout (approved):

devito/arch/compiler.py
FabioLuporini: ping (merging)

FabioLuporini: @jkwashbourne

PR #2052, opened 2023-01-26
symbolics: use devito floor instead of Undefined Function

### LINKED ISSUE #2053: pickling of simple forward acoustic operator fails
I got the error below when tried to pickle a forward acoustic `Operator` object built as shown in tuto 1. The code to reproduce the issue is in attachment. An even shorter code  (MFE) that results in the same error message is also provided.

```
/home/oscar/geodev/test_pickle.py:66: DeprecationWarning: Please use `floor` from the `scipy.special` namespace, the `scipy.special.orthogonal` namespace is deprecated.
  pkl_op = pickle.dumps(op)
  File "/home/oscar/geodev/test_pickle.py", line 71, in <module>
    main()
  File "/home/oscar/geodev/test_pickle.py", line 66, in main
    pkl_op = pickle.dumps(op)
_pickle.PicklingError: Can't pickle floor: attribute lookup floor on __main__ failed
```
[test_pickle_op.py.txt](https://github.com/devitocodes/devito/files/10509752/test_pickle_op.py.txt)
[mfe.py.txt](https://github.com/devitocodes/devito/files/10509758/mfe.py.txt)

mloubout:
Fix issue with floor function

Fixes #2053


FabioLuporini (approved):
uncontroversial

devito/symbolics/extended_sympy.py
```diff
@@ -701,6 +696,31 @@ def rfunc(func, item, *args):
...
 
 rfunc_mapper = {
-    min: MIN,
-    max: MAX,
+    min: Min,
+    max: Max,
 }
+
+
+def integer_args(*args):
```
FabioLuporini, line 704: If this function is used just for dispatching to the correct printing routing, should we not keep it inside `printer.py`?

devito/symbolics/extended_sympy.py
```diff
@@ -701,6 +696,31 @@ def rfunc(func, item, *args):
...
+    Used to choose the function printed in the c-code
+    """
+    if len(args) == 0:
+        return False
+
+    if len(args) == 1:
+        try:
+            return np.issubdtype(args[0].dtype, np.integer)
+        except AttributeError:
+            return args[0].is_integer
```
FabioLuporini, line 716: blank line ideally here

tests/test_symbolics.py
```diff
@@ -354,25 +355,25 @@ def test_multibounds_op(self):
         assert fnorm == fnorm2
 
     @pytest.mark.parametrize('op, expr, assumptions, expected', [
-        ([min, '[a, b, c, d]', '[]', 'MIN(a, MIN(b, MIN(c, d)))']),
-        ([max, '[a, b, c, d]', '[]', 'MAX(a, MAX(b, MAX(c, d)))']),
+        ([min, '[a, b, c, d]', '[]', 'Min(a, Min(b, Min(c, d)))']),
```
FabioLuporini, line 358: Being C macros, it would be nicer if they were all uppercase letters 😬 

see here https://stackoverflow.com/questions/20269669/define-why-uppercase for example

devito/mpi/routines.py
```diff
@@ -513,6 +520,14 @@ class OverlapHaloExchangeBuilder(DiagHaloExchangeBuilder):
         remainder()
     """
 
+    @property
+    def headers(self):
```
FabioLuporini, line 524: Here's a potentially better way of achieving the same thing, which generalizes to future Macros as well..

We could have a final IET pass that parses the AST looking for special symbolic objects, such as Min, Max, ... "special" here means for example inheriting from a mixin class "MacroGenerator" or something along the likes of it

Such IET pass would then very very simply return 

```
macro_generators = ...
headers = [n.header for n in macro_generators]

return iet, {'headers': headers}
```

This may or may not be an addition we want to do with this PR, but if not, I would open an issue

By doing that, we wouldn't require each individual pass using a macro having to care about it !

devito/symbolics/extended_sympy.py
```diff
@@ -701,6 +696,31 @@ def rfunc(func, item, *args):
...
+    res = True
+    for a in args:
+        try:
+            if len(a.args) > 0:
+                res = res and integer_args(*a.args)
+            else:
+                res = res and integer_args(a)
+        except AttributeError:
+            res = res and integer_args(a)
+    return res
```
FabioLuporini, line 726: So the implementation of this recursive function is a bit odd

What I would do (haven't tested the code though)

```
def is_integer_expr(expr):
    if expr.is_integer:
        return True

    try:
        return np.issubdtype(expr.dtype, np.integer)
    except AttributeError:
        pass

     # We must look inside the args
     return all(is_integer_expr(a) for a in expr.args)
```

FabioLuporini (changes requested):

devito/symbolics/extended_sympy.py
```diff
@@ -7,14 +7,15 @@
...
+from devito.finite_differences.elementary import Min, Max
 from devito.tools import Pickable, as_tuple, is_integer
 from devito.types import Symbol
 
 __all__ = ['CondEq', 'CondNe', 'IntDiv', 'CallFromPointer', 'FieldFromPointer',
            'FieldFromComposite', 'ListInitializer', 'Byref', 'IndexedPointer', 'Cast',
            'DefFunction', 'InlineIf', 'Keyword', 'String', 'Macro', 'MacroArgument',
-           'CustomType', 'Deref', 'INT', 'FLOAT', 'DOUBLE', 'VOID', 'CEIL',
-           'FLOOR', 'MAX', 'MIN', 'Null', 'SizeOf', 'rfunc', 'cast_mapper',
+           'CustomType', 'Deref', 'INT', 'FLOAT', 'DOUBLE', 'VOID',
```
georgebisbas, line 17: nit: These 3 ilnes of imports can be 2 I guess

devito/symbolics/extended_sympy.py
georgebisbas, line 704: Isn't this name a bit misleading? Can we use sth better?

mloubout: What do you mean , misleading? It check if the args are integer

mloubout: Yeah couldn't decide happy either way

devito/symbolics/extended_sympy.py
mloubout: Because we have nested function with a lot of case that have `is_integer` defined but returns `None` because it's a function/expression so it returns wrong result with the simple recursive (had it exactly like that at initially)

examples/performance/00_overview.ipynb
```diff
@@ -340,9 +340,9 @@
       "    {\n",
       "      for (int y0_blk0 = y_m; y0_blk0 <= y_M; y0_blk0 += y0_blk0_size)\n",
       "      {\n",
-      "        for (int x = x0_blk0; x <= MIN(x0_blk0 + x0_blk0_size - 1, x_M); x += 1)\n",
+      "        for (int x = x0_blk0; x <= MIN(x_M, x0_blk0 + x0_blk0_size - 1); x += 1)\n",
```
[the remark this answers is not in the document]
mloubout: It's symbolic sorting, it's deterministic it just sorts the arguments based on sympy rules (string sort, numbers, ...). I think it's a really bad idea to enforce our own sorting, it's a lot of extra work that needs to be tested and maintain rather than using an existing and pretty robust symbolic sorting.

devito/symbolics/extended_sympy.py
mloubout: Done

devito/symbolics/extended_sympy.py
mloubout: done

devito/symbolics/extended_sympy.py
mloubout: moved

devito/mpi/routines.py
mloubout: That's a good point, not sure the scope of this PR. Not sure which one is easier to maintain/extend some headers/macro may be tricky to infer.

FabioLuporini: The thing I suggested above would be, by experience, easier.

Basically: the more we automate, the better

I agree we can park it for another day though. Could u open an enhancement-issue once we land this PR?

devito/symbolics/extended_sympy.py
FabioLuporini: maybe `has_integer_args` -- but nitpicking here

georgebisbas: yEAH, sth like has_....check....test...integer_args..

mloubout: done

devito/mpi/routines.py
```diff
@@ -510,6 +517,14 @@ class OverlapHaloExchangeBuilder(DiagHaloExchangeBuilder):
...
+        """
+        Overlap Mode uses MIN/MAX that need to be defined
```
georgebisbas, line 523: ok

devito/symbolics/printer.py
```diff
@@ -101,20 +101,23 @@ def _print_Mod(self, expr):
         args = ['(%s)' % self._print(a) for a in expr.args]
         return '%'.join(args)
 
+    def _print_Min(self, expr):
```
georgebisbas, line 104: ok

tests/test_linearize.py
```diff
@@ -191,7 +192,7 @@ def test_codegen_quality1():
     assert all('const long' not in str(i) for i in exprs[-3:])
 
     # Only two access macros necessary, namely `uL0` and `r1L0` (the other five
-    # obviously are _POSIX_C_SOURCE, MIN, MAX, START_TIMER, STOP_TIMER)
+    # obviously are _POSIX_C_SOURCE, Min, Max, START_TIMER, STOP_TIMER)
```
georgebisbas, line 195: MIN, MAX

mloubout: fixed

mloubout:
To be improved, see #2064

PR #2058, opened 2023-01-31
compiler: Switch from aomp to clang for amd 

mloubout:


.github/workflows/pytest-gpu.yml
```diff
@@ -46,31 +46,34 @@ jobs:
...
-          # https://github.com/ROCm-Developer-Tools/aomp/issues/397
-          test_files: "tests/test_gpu_openmp.py"
+          test_files: "tests/test_adjoint.py tests/test_gpu_common.py tests/test_gpu_openmp.py"
           tags: ["self-hosted", "amdgpu"]
           base: "devitocodes/bases:amd"
           test_drive_cmd: "rocm-smi"
           # Attach the AMD GPU devices `/dev` and add user to video and render (109 on wampa) group
           # Options from https://rocmdocs.amd.com/en/latest/ROCm_Virtualization_Containers/ROCm-Virtualization-&-Containers.html
           flags: "--network=host --device=/dev/kfd --device=/dev/dri --ipc=host --group-add video --group-add 109 --cap-add=SYS_PTRACE --security-opt seccomp=unconfined --rm --name testrun-amd"
+          mpi_flags: "--mca btl '^openib' -x UCX_TLS=sm,self,rocm_copy,rocm_ipc  --mca pml ucx"
```
FabioLuporini, line 76: Can we make them part of the dockerfile somehow?

docker/Dockerfile.cpu
```diff
@@ -24,7 +24,7 @@ RUN apt-get install -y libgl1-mesa-glx
 ##############################################################
 FROM base as gcc
 
-RUN apt-get install -y mpich libmpich-dev 
+RUN apt-get install -y mpich libmpich-dev
```
FabioLuporini, line 27: extra space?

.github/workflows/pytest-gpu.yml
```diff
@@ -46,31 +46,34 @@ jobs:
           pytest-gpu-acc-nvidia,
           pytest-gpu-omp-amd
         ]
+        test_examples: ["examples/seismic/tti/tti_example.py examples/seismic/acoustic/acoustic_example.py examples/seismic/viscoacoustic/viscoacoustic_example.py examples/seismic/viscoelastic/viscoelastic_example.py examples/seismic/elastic/elastic_example.py"]
+
         include:
         - name: pytest-gpu-omp-nvidia
           test_files: "tests/test_adjoint.py tests/test_gpu_common.py tests/test_gpu_openmp.py"
```
FabioLuporini, line 53: Now that they are the same for all builds, we can pull this variable outside the `include` scope, and maybe put it right under the newly introduced `test_examples` ?

mloubout: They are not the openacc build is different

.github/workflows/pytest-gpu.yml
mloubout: We could be this makes it slightly easier for ci

FabioLuporini: but OK why do these options only matter for CI and not for the end users?

mloubout: No it does too and should be moved to the docker at some point, but I would need to re-build the docker separately , test it, then restart that PR so it's not very practical that's all

docker/Dockerfile.cpu
mloubout: No removed extra space

.github/workflows/pytest-gpu.yml
mloubout: As mentionned in slack these should go in the `mpiexec.conf`

.github/workflows/pytest-gpu.yml
FabioLuporini: ah right !

.github/workflows/pytest-gpu.yml
mloubout: fixed

PR #2065, opened 2023-02-13
compiler: Patch data dependencies across Jumps

FabioLuporini:


devito/ir/support/basic.py
```diff
@@ -822,7 +823,10 @@ def __init__(self, exprs, rules=None):
                     f = a.function
                     indices = [i if d in e.ispace else S.Infinity
                                for i, d in zip(a, a.aindices)]
-                    mock = f.indexify(indices)
+                    try:
+                        mock = f.indexify(indices)
+                    except IndexException:
```
mloubout, line 828: What would make this fail? Bad indices?

FabioLuporini: yeah just empty `indices`, trivial sympy exception when you attempt to create an Indexed w/o supplying indices 😬

georgebisbas: Should you add some related message or failing test?

georgebisbas (approved):

mloubout (approved):

devito/ir/support/basic.py
FabioLuporini: Definitely not a message; probably a unit test...

However, creating an artificial case exciting this code path here w/o PRO will take O(hour) time, which I don't have, so given that this is a minuscule thing and that, after all, we do have a test exciting this in PRO, despite agreeing w/ you this ain't ideal, I suggest we proceed

PR #2067, opened 2023-02-20
misc: Address Nvidia CUDA_HOME has been deprecated warning

ggorman:
Addresses the compiler warning

nvc++-Warning-CUDA_HOME has been deprecated. Please, use NVHPC_CUDA_HOME instead.

### DIFF
--- a/docker/Dockerfile.nvidia
+++ b/docker/Dockerfile.nvidia
@@ -90,7 +90,7 @@ RUN echo "$HPCSDK_HOME/cuda/lib" >> /etc/ld.so.conf.d/nvidia.conf && \
     echo "$HPCSDK_HOME/math_libs/lib64" >> /etc/ld.so.conf.d/nvidia.conf    
     
 # Compiler, CUDA, and Library paths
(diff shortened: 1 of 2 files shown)


docker/Dockerfile.nvidia
```diff
@@ -90,7 +90,7 @@ RUN echo "$HPCSDK_HOME/cuda/lib" >> /etc/ld.so.conf.d/nvidia.conf && \
     echo "$HPCSDK_HOME/math_libs/lib64" >> /etc/ld.so.conf.d/nvidia.conf    
     
 # Compiler, CUDA, and Library paths
-ENV CUDA_HOME $HPCSDK_HOME/cuda
+ENV NVHPC_CUDA_HOME $HPCSDK_HOME/cuda
```
mloubout, line 93: I would keep both, CUDA_HOME is still used my most stuff for cuda path

FabioLuporini: but then you would still get the warning?

ggorman:
Moved changes to #2079

PR #2069, opened 2023-02-21
compiler: Relax _mark_overlappable

FabioLuporini:
Works around an issue in PRO in which diag2 fallbacks to basic due to too conservative data dependence analysis

the tests are in PRO... I can't think of an easy way of testing this here. Not ideal, but this fixes an issue @jkwashbourne is facing

### DIFF
--- a/.github/workflows/pytest-core-nompi.yml
+++ b/.github/workflows/pytest-core-nompi.yml
@@ -54,7 +54,7 @@ jobs:
 
         - name: pytest-ubuntu-py36-gcc7-omp
           python-version: '3.6'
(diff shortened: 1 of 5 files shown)


mloubout (approved):

georgebisbas (approved):
Can you use any func_table content to test this?

FabioLuporini:
> Can you use any func_table content to test this?

@georgebisbas 

not really in the sense that I don't have the means to generate that particular code (it would theoretically be possible, but it's a lot of work, and I don't have the time right now)

PR #2079, opened 2023-03-07
tests: Fix warnings

georgebisbas:


devito/finite_differences/differentiable.py
```diff
@@ -228,9 +228,13 @@ def __neg__(self):
         return Mul(sympy.S.NegativeOne, self)
 
     def __eq__(self, other):
-        return super(Differentiable, self).__eq__(other) and\
-            all(getattr(self, i, None) == getattr(other, i, None)
-                for i in self.__rkwargs__)
+        if super(Differentiable, self).__eq__(other) is not True:
```
mloubout, line 231: can you not just do `super(Differentiable, self).__eq__(other) is True and\

georgebisbas: Thanks @mloubout , seems to work after a quick local test.

pytest.ini
```diff
@@ -0,0 +1,4 @@
+[pytest]
+filterwarnings =
+    ignore::pytest.PytestCacheWarning
```
georgebisbas, line 3: I would add a comment to remember where it comes from

PR #2080, opened 2023-03-07
arch: Support aws graviton

### LINKED ISSUE #1106: Add ARM testing CI through github actions

mloubout:
AWS gravition arch and aws-ec2 on-demand self-hosted runners

Fixes #1106

### DIFF
--- a/.github/workflows/cancel.yml
+++ b/.github/workflows/cancel.yml
@@ -12,5 +12,5 @@ jobs:
         with:
           # Ids to cancel core/mpi/gpu/examples/tutorials
           # https://api.github.com/repos/devitocodes/devito/actions/workflows
-          workflow_id: 167582, 203470, 203471, 434862, 501517
+          workflow_id: 167582, 203470, 203471, 434862, 501517, 50473775
           access_token: ${{ github.token }}
--- a/devito/arch/compiler.py
+++ b/devito/arch/compiler.py
@@ -12,7 +12,7 @@ import numpy.ctypeslib as npct
 from codepy.jit import compile_from_string
 from codepy.toolchain import GCCToolchain
 
-from devito.arch import (AMDGPUX, Cpu64, M1, NVIDIAX, SKX, POWER8, POWER9,
+from devito.arch import (AMDGPUX, Cpu64, M1, NVIDIAX, SKX, POWER8, POWER9, GRAVITON,
                          get_nvidia_cc, check_cuda_runtime, get_m1_llvm_path)
 from devito.exceptions import CompilationError
 from devito.logger import debug, warning, error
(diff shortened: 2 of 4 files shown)


devito/arch/compiler.py
```diff
@@ -373,7 +373,7 @@ def add_ldflags(self, flags):
 class GNUCompiler(Compiler):
 
     def __init__(self, *args, **kwargs):
-        super(GNUCompiler, self).__init__(*args, **kwargs)
```
ggorman, line 376: I don't understand why this line has changed for graviton support.

devito/arch/compiler.py
```diff
@@ -400,6 +400,17 @@ def __lookup_cmds__(self):
         self.MPICXX = 'mpicxx'
 
 
+class ArmCompiler(GNUCompiler):
```
ggorman, line 403: How is this different from previous ARM support?

mloubout: We did not have an ARM compiler before only gcc, this adds the extra flags for graviton and I'm sure other "brands" will need some flags too.

devito/arch/compiler.py
mloubout: Oh it's just we moving towards `super()` everywhere in the codebase rather than with args so changed it while there

ggorman:
Doesn't this fix https://github.com/devitocodes/devito/issues/1106

.github/workflows/cancel.yml
```diff
@@ -12,5 +12,5 @@ jobs:
         with:
           # Ids to cancel core/mpi/gpu/examples/tutorials
           # https://api.github.com/repos/devitocodes/devito/actions/workflows
-          workflow_id: 167582, 203470, 203471, 434862, 501517
+          workflow_id: 167582, 203470, 203471, 434862, 501517, 50473775
```
ggorman, line 15: Ok - school me... what does this do?

mloubout: Oh this cancel the workflows currently running when you push again so we don't have thousands of concurrent runs from different commits on the same branch

mloubout: In the url line 14

devito/arch/compiler.py
```diff
@@ -400,6 +400,17 @@ def __lookup_cmds__(self):
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        platform = kwargs.pop('platform', configuration['platform'])
+        # Graviton flag
+        if platform is GRAVITON:
+            self.cflags += ['-mcpu=neoverse-n1']
```
FabioLuporini, line 411: really =native doesn't do the same?

FabioLuporini (approved):
I left a small question, but otherwise GTG. Great, thanks

georgebisbas (approved):

devito/arch/compiler.py
mloubout: According to the aws this flag should be used yeah

ggorman (approved):

PR #2081, opened 2023-03-08
ci: remove jupyter formatting from output as it breaks with ipynb+mpi.

ggorman:
Depending on the version of python/jupyter, there are some differences in the formatting of objects when you let jupyter+ipyparallel/mpi simply echo the object by writing the object as a statement at the end of a cell vs running the notebook using nbval. 

Resolved by explicitly passing the object to `print()` at the end of the cell.

### DIFF
--- a/examples/mpi/overview.ipynb
+++ b/examples/mpi/overview.ipynb
@@ -159,193 +159,27 @@
    },
    "outputs": [
     {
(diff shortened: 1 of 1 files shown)


ggorman:
Move changes into #2079

PR #2087, opened 2023-03-13
compiler: only use offloading flags for amd gpu not cpu

mloubout:


FabioLuporini (approved):
uncontroversial

How did you notice it?

mloubout:
>How did you notice it?

My home laptop is amd cpu.

PR #2094, opened 2023-03-29
compiler: Do block with partial data reuse

FabioLuporini:


EdCaunt (approved):

devito/passes/clusters/blocking.py
```diff
@@ -196,6 +196,10 @@ def callback(self, clusters, prefix):
         if self._has_short_trip_count(d):
             return clusters
 
+        # Pointless if there's no data reuse
+        if all(not self._has_data_reuse(c) for c in clusters):
```
georgebisbas, line 200: nit: Slight preference for but not important

```suggestion
        if not any(self._has_data_reuse(c) for c in clusters):
```

georgebisbas (approved):

devito/passes/clusters/blocking.py
FabioLuporini: agree, I will change it in a subsequent branch

PR #2095, opened 2023-03-29
compiler: Change tile use in DeviceAcczier to allow multiple tile sizes

gpc1064:
A proposal to allow the use of multiple tile sizes in the same operator, through the _part-tile_ flag.
Passing multiple tuples to the _part-tile_ flag is already supported, but only the first tuple is applied in the __make_partree_ method of _DeviceAcczier_ class.

The changes consist of supplying the nesting index in the _make_parallel_ method of _PragmaShmTransformer_ to the __make_partree_ method and using the tile size with the respective index.

Cases with different tile and nesting sizes are handled. Given `n` tile sizes for an operator with `m` nests:

If `n = m`: The i-th tile size is assigned to the i-th nesting.
If `n > m`: The first `m` tile sizes are assigned to the `m` nests (i-th tile size -> i-th nesting), and the remaining `n-m` is not used.
If `n < m`: The `n` tile sizes are assigned to the first `n` nestings, and the remaining `m-n` nestings replicate the last tile size.

### DIFF
--- a/devito/passes/iet/languages/openacc.py
+++ b/devito/passes/iet/languages/openacc.py
@@ -180,13 +180,8 @@ class DeviceAccizer(PragmaDeviceAwareTransformer):
         if self._is_offloadable(root) and \
            all(i.is_Affine for i in [root] + collapsable) and \
            self.par_tile:
-            # TODO: still unable to exploit multiple par-tiles (one per nest)
-            # This will require unconditionally applying blocking, and then infer
-            # the tile clause shape from the BlockDimensions' step
-            par_tile_length = len(self.par_tile)
-            idx = index if index < par_tile_length else par_tile_length - 1
+            idx = min(index, len(self.par_tile) - 1)
             tile = self.par_tile[idx]
-
             assert isinstance(tile, tuple)
             nremainder = (ncollapsable + 1) - len(tile)
             if nremainder >= 0:
--- a/devito/passes/iet/languages/openacc.py
+++ b/devito/passes/iet/languages/openacc.py
@@ -171,7 +171,7 @@ class DeviceAccizer(PragmaDeviceAwareTransformer):
 
     lang = AccBB
 
-    def _make_partree(self, candidates, nthreads=None):
+    def _make_partree(self, candidates, nthreads=None, index=0):
         assert candidates
 
         root, collapsable = self._select_candidates(candidates)
@@ -183,7 +183,10 @@ class DeviceAccizer(PragmaDeviceAwareTransformer):
             # TODO: still unable to exploit multiple par-tiles (one per nest)
             # This will require unconditionally applying blocking, and then infer
             # the tile clause shape from the BlockDimensions' step
-            tile = self.par_tile[0]
+            par_tile_length = len(self.par_tile)
+            idx = index if index < par_tile_length else par_tile_length - 1
+            tile = self.par_tile[idx]
+
             assert isinstance(tile, tuple)
             nremainder = (ncollapsable + 1) - len(tile)
             if nremainder >= 0:
--- a/devito/passes/iet/parpragma.py
+++ b/devito/passes/iet/parpragma.py
@@ -395,7 +395,7 @@ class PragmaShmTransformer(PragmaSimdTransformer):
                 continue
 
             # Outer parallelism
-            root, partree = self._make_partree(candidates, None, i)
+            root, partree = self._make_partree(candidates, index=i)
             if partree is None or root in mapper:
                 continue
 
--- a/devito/passes/iet/parpragma.py
+++ b/devito/passes/iet/parpragma.py
@@ -388,7 +388,7 @@ class PragmaShmTransformer(PragmaSimdTransformer):
     def _make_parallel(self, iet):
         mapper = {}
         parrays = {}
-        for i, tree in enumerate(retrieve_iteration_tree(iet, mode='superset')) :
+        for i, tree in enumerate(retrieve_iteration_tree(iet, mode='superset')):
             # Get the parallelizable Iterations in `tree`
             candidates = filter_iterations(tree, key=self.key)
             if not candidates:
--- a/devito/passes/iet/parpragma.py
+++ b/devito/passes/iet/parpragma.py
@@ -388,14 +388,14 @@ class PragmaShmTransformer(PragmaSimdTransformer):
     def _make_parallel(self, iet):
         mapper = {}
         parrays = {}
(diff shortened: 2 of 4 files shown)


devito/passes/iet/languages/openacc.py
```diff
@@ -183,7 +183,10 @@ def _make_partree(self, candidates, nthreads=None):
             # TODO: still unable to exploit multiple par-tiles (one per nest)
```
FabioLuporini, line 183: You may now drop this comment

georgebisbas (changes requested):
Thanks for this! As a first thing to improve, before the rest, could you add tests to this PR?

devito/passes/iet/languages/openacc.py
```diff
@@ -183,7 +183,10 @@ def _make_partree(self, candidates, nthreads=None):
...
             # This will require unconditionally applying blocking, and then infer
             # the tile clause shape from the BlockDimensions' step
-            tile = self.par_tile[0]
+            par_tile_length = len(self.par_tile)
```
FabioLuporini, line 186: how about simply `n = ...` ?

devito/passes/iet/languages/openacc.py
```diff
@@ -183,7 +183,10 @@ def _make_partree(self, candidates, nthreads=None):
...
+            idx = index if index < par_tile_length else par_tile_length - 1
```
FabioLuporini, line 187: we prefer explicit if-then-else:

```
if ...:
    idx = ...
else:
    idx = ...
```

devito/passes/iet/parpragma.py
```diff
@@ -388,14 +388,14 @@ def _make_nested_partree(self, partree):
...
-        for tree in retrieve_iteration_tree(iet, mode='superset'):
+        for i, tree in enumerate(retrieve_iteration_tree(iet, mode='superset')) :
             # Get the parallelizable Iterations in `tree`
             candidates = filter_iterations(tree, key=self.key)
             if not candidates:
                 continue
 
             # Outer parallelism
-            root, partree = self._make_partree(candidates)
+            root, partree = self._make_partree(candidates, None, i)
```
FabioLuporini, line 398: instead of passing None:

```
self._make_partree(candidates, index=i)
```

devito/passes/iet/languages/openacc.py
FabioLuporini: and in practice this is just 

```
index = min(index, par_tile_length-1)
```

FabioLuporini (changes requested):
This PR needs to be equipped with tests, somewhere here: https://github.com/devitocodes/devito/blob/master/tests/test_dle.py#L265

and at least one in `test_gpu_openacc.py` as well

I've also left a batch of small comments

devito/passes/iet/languages/openacc.py
gpc1064: Which one is more suitable?
```
tile = self.par_tile[min(index, len(self.par_tile) - 1)]
```

```
idx = min(index, len(self.par_tile) - 1)
tile = self.par_tile[idx]
```

FabioLuporini: I'd say the latter for homogeneity with the rest of the codebase.

gpc1064:
> Thanks for this! As a first thing to improve, before the rest, could you add tests to this PR?

> This PR needs to be equipped with tests, somewhere here: https://github.com/devitocodes/devito/blob/master/tests/test_dle.py#L265
> 
> and at least one in `test_gpu_openacc.py` as well
> 
> I've also left a batch of small comments

Added [test_openacc_structure](https://github.com/gpc1064/devito/blob/multiple_par_tile/tests/test_dle.py#L329) in `test_dle.py::TestBlockingParTile` to ensure correct blocking while used with OpenACC's tile.

Added [test_openacc_multi_tile_structure](https://github.com/gpc1064/devito/blob/multiple_par_tile/tests/test_dle.py#L350) in `test_dle.py::TestBlockingParTile` to ensure correct blocking and applying multiple tile sizes with OpenACC.

Added [test_multiple_tile_sizes](https://github.com/gpc1064/devito/blob/multiple_par_tile/tests/test_gpu_openacc.py#L116) in `test_gpu_openacc.py::TestCodeGeneration` ensuring that multiple tile sizes are applied correctly across nests.

gpc1064:
Requested code changes also committed

tests/test_dle.py
```diff
@@ -318,6 +318,60 @@ def test_structure_2p5D(self):
...
 
+    
+    @pytest.mark.parametrize('par_tile,expected', [
+        ((32, 4, 4), ((4, 4, 32), (4, 4, 32))),
+        (((16, 4, 4), (16, 16, 16)), ((4, 4, 16), (16, 16, 16))),
+        (((32, 4, 4), 1), ((4, 4, 32), (4, 4, 32))),
+        (((32, 4, 4), 1, 'tag0'), ((4, 4, 32), (4, 4, 32))),
+        ((((32, 4, 8), 1, 'tag0'), ((32, 8, 4), 2)), ((8, 4, 32), (4, 8, 32))),
+    ])
+    def test_openacc_structure(self, par_tile, expected):
```
FabioLuporini, line 329: Sorry, I made a mistake; clearly these tests belong to `test_gpu_openacc`, not to `test_dle` which constitutes a platform-independent set of tests

FabioLuporini: (By moving them there you'll have to rename them as well -- dropping "openacc" for example)

tests/test_dle.py
```diff
@@ -318,6 +318,60 @@ def test_structure_2p5D(self):
...
+        bns, _ = assert_blocking(op, {'x0_blk0', 'x1_blk0'})
+        assert len(bns) == len(expected)
+        for root, v in zip(bns.values(), expected):
+            iters = FindNodes(Iteration).visit(root)
+            iters = [i for i in iters if i.dim.is_Block and i.dim._depth == 1]
+            assert len(iters) == len(v)
+            assert all(i.step == j for i, j in zip(iters, v))
+
+
+    def test_openacc_multi_tile_structure(self):
```
FabioLuporini, line 350: how is this test conceptually different than the one above?

FabioLuporini (changes requested):

tests/test_dle.py
gpc1064: Indeed the first one was kind of unnecessary. I've dropped it and kept the second one, which is more relevant

FabioLuporini (approved):
OK I think this is now good to go. Enabling tests...

PR #2096, opened 2023-03-30
reqs: Move pyrevolve to optionals and introduce testing-only reqs

georgebisbas:


.github/workflows/pytest-core-mpi.yml
```diff
@@ -36,7 +36,7 @@ jobs:
       run: |
         sudo apt-get update && sudo apt install mpich -y
         pip3 install --upgrade pip
-        pip3 install -e .[extras,mpi]
+        pip3 install -e .[extras,mpi,testd]
```
mloubout, line 39: typo (sorry I know I made it on slack 😅 )

requirements.txt
```diff
@@ -10,11 +10,7 @@ py-cpuinfo<10
 cgen>=2020.1
 codepy>=2019.1
 click<9.0
-codecov
 multidict
 anytree>=2.4.3,<=2.8
-pyrevolve>=2.1.3
 distributed<2023.4
 pytest>=7.2,<8.0
```
mloubout, line 16: pytest too

setup.py
```diff
@@ -24,7 +27,8 @@
         reqs += [ir]
 
 extras_require = {}
-for mreqs, mode in zip([optionals, mpis, nvidias], ['extras', 'mpi', 'nvidia']):
+for mreqs, mode in (zip([optionals, mpis, nvidias, testing],
+                        ['extras', 'mpi', 'nvidia', 'testd'])):
```
mloubout, line 31: Need to make `exclude` below conditional on `tests`

georgebisbas: Not sure I understand this. Could you issue a commit on this?

tests/conftest.py
```diff
@@ -23,19 +24,17 @@ def skipif(items, whole_module=False):
...
         raise ValueError("Illegal skipif argument(s) `%s`" % unknown)
     skipit = False
     for i in items:
         # Skip if no MPI
-        if i == 'nompi':
-            if MPI is None:
-                skipit = "mpi4py/MPI not installed"
-                break
-            continue
+        if i == 'nompi' and MPI is None:
```
FabioLuporini, line 35: why de-nesting?

devito/__init__.py
```diff
@@ -23,7 +23,11 @@
 from devito.data.allocators import *  # noqa
 from devito.logger import error, warning, info, set_log_level  # noqa
 from devito.mpi import MPI  # noqa
-from devito.checkpointing import DevitoCheckpoint, CheckpointOperator  # noqa
+from devito.checkpointing import pyrevolve  # noqa
+try:
+    from devito.checkpointing import DevitoCheckpoint, CheckpointOperator  # noqa
+except ImportError:
+    pass
```
FabioLuporini, line 30: I think we can avoid a lot of 

```
if pyrevolve:
    ...
```

and a lot of `devito.DevitoCheckpoint`

by doing the following:

in `checkpointing.py` we could add:

```
class Noop(object):
    """
    Dummy replacement in case pyrevolve isn't available.
    """

        raise ImportError("Missing required dependency `pyrevolve`; cannot use checkpointing")

class NoopCheckpointOperator(Noop):
    pass

class NoopCheckpoint(Noop):
    pass
```

then here (line 30 of `__init__.py`) you import:

```
from devito.checkpointing import NoopCheckpoint as DevitoCheckpoint, NoopCheckpointOperator as CheckpointOperator
```

and then in theory we should just be good to go avoiding most of the changes we've made to examples and tests

(clearly in `test_checkpoint` we would still need the `skipif`)

tests/conftest.py
FabioLuporini: In retrospect, I don't even know whether the skipif(nompi) makes sense at all

FabioLuporini (approved):
LGTM now. Thanks a lot.

.github/workflows/examples.yml
```diff
@@ -42,8 +42,8 @@ jobs:
 
     - name: Install dependencies
       run: |
-        pip install -e .
-        pip install matplotlib
+        pip install -e .[tests]
+        pip install matplotlib pyrevolve
```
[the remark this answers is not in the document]
mloubout: `pip` always install the latest compatible version.

PR #2097, opened 2023-03-31
docker: Build icx docker image

georgebisbas:


mloubout:
I think it's quite overkill as can just use a common intel docker and call it with `-e DEVITO_ARCH=icx/icc` I don't mind renaming the `cpu-icc` to `cpu-intel` and then just have the runtime choice

PR #2099, opened 2023-04-05
compiler: Prevent topofusion of homogeneous sync Clusters

FabioLuporini:
This affects only PRO.

While CI runs, I'll think about whether there's a way of testing this in OSS as well (in PRO it's tested, it's trivial)

### DIFF
--- a/devito/ir/clusters/visitors.py
+++ b/devito/ir/clusters/visitors.py
@@ -27,6 +27,7 @@ class Queue(object):
     _q_ispace_in_key = True
     _q_guards_in_key = False
     _q_properties_in_key = False
+    _q_syncs_in_key = False
 
     def callback(self, *args):
         raise NotImplementedError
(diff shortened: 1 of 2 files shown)


FabioLuporini:
I honestly think it's impossible because the only to trigger the faulty scenario is by having a compiler pass that injects a specific equation at a specific point in the pipeline -- an equation that shouldn't be allowed to flaot around in the sequence of Clusters, that's what the patch prevents

I can't express that equation in the user level because of the way tasks are selected, that is on an individual (Eq) basis. In other words, I can't express a task comprising two Eqns.

```
Eq(usave, ...)
```

with `usave` being a saved TimeFunction will be casted into an asynchronous task, but I (user) can't write

```
eqns = [Eq(s, 0.53),
              Eq(usave, ... s ...)]
```

The first eqn simply won't be part of the task (the compiler will still produce correct code, but it will pass `s` to the asynchrnous task as a struct member

I do realise most of this might not make sense to you if you're not familiar with the kind of streaming code that PRO generates

However, I hope the explanation is convincing enough to justify the absence of a unit test

The test is however present in PRO; for the record, it's `devitopro/tests/test_abox_funcs.py::TestIsotropicAcoustic::test_forward`

mloubout:
I think I get the gist of it , don't completely follow, but sounds sane. GTG

georgebisbas (approved):
Looks ok I think

PR #2104, opened 2023-04-13
install: Overhaul Dockerfile.amd for MPI support

FabioLuporini:


docker/Dockerfile.amd
```diff
@@ -2,72 +2,129 @@
...
-ENV OMPI_CXX=$AOMP/bin/amdclang++
-ENV OMPI_F90=$AOMP/bin/flan
-
-# Set mpiexec flags for gpu aware mpi
-RUN echo "--mca btl '^openib' -x UCX_TLS=sm,self,rocm_copy,rocm_ipc --mca pml ucx" > /opt/openmpi/etc/mpiexec.conf
+    apt-get install -y dh-autoreconf python3-venv python3-dev python3-pip
 
 # Build mpi4py against amdclang
-RUN python3 -m venv /venv && \
+RUN which mpicc && \
```
georgebisbas, line 121: Would it hurt to keep?

georgebisbas (approved):

docker/Dockerfile.amd
FabioLuporini: which part?

devito/arch/compiler.py
```diff
@@ -487,24 +487,24 @@ def __init__(self, *args, **kwargs):
...
+                self.ldflags += ['-fopenmp']
+                self.ldflags += ['--offload-arch=%s' % platform.march]
         elif platform in [POWER8, POWER9]:
             # It doesn't make much sense to use AOMP on Power, but it should work
             self.cflags.append('-mcpu=native')
         else:
             self.cflags.append('-march=native')
 
-        # Generic amd flags
-        self.ldflags.extend(['-fopenmp', '-target', 'x86_64-pc-linux-gnu'])
```
mloubout, line 501: I know these can be removed but you still need the other flag `offload-arch`

docker/Dockerfile.amd
```diff
@@ -2,72 +2,129 @@
 # This Dockerfile contains AMD compilers
 ##############################################################
 
+ARG IMAGE="rocm/dev-ubuntu-22.04:5.4.2-complete"
 ARG arch="aomp"
 
-ARG ROCM_VERSION=5.4.2
```
mloubout, line 7: This seems easier to maintain it's the exact same but can easily switch version and much easier to use

devito/arch/compiler.py
mloubout: Also need to fix the arch detection to use `offload-arch` not `mygpu`

docker/Dockerfile.amd
```diff
@@ -2,72 +2,129 @@
...
-                 --prefix=/opt/openmpi/ --with-ucx=/opt/ucx \
-                 --enable-mca-no-build=btl-uct --enable-mpi1-compatibility && \
-    make -j $(( $(lscpu | awk '/^Socket\(s\)/{ print $2 }') * $(lscpu | awk '/^Core\(s\) per socket/{ print $4 }') )) && \
-    make install && \
-    rm -rf /deps/openmpi
-
-# Set OpenMPI path
-ENV PATH=${PATH}:/opt/openmpi/bin:$AOMP/bin
-ENV LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/opt/openmpi/lib:$AOMP/lib
-ENV OMPI_CC=$AOMP/bin/amdclang
```
mloubout, line 58: Pretty sure need those (and hip one for hip)

mloubout, line 58: Need those (and hip one for hip build)

docker/Dockerfile.amd
```diff
@@ -2,72 +2,129 @@
...
+    && make install
+
+# Install OpenMPI
+RUN cd /tmp \
+    && git clone --recursive https://github.com/open-mpi/ompi.git -b ${OMPI_BRANCH} \
+    && cd ompi \
+    && ./autogen.pl \
+    && mkdir build \
+    && cd build \
+    && ../configure --prefix=$OMPI_HOME --with-ucx=$UCX_HOME \
```
mloubout, line 86: If you don't build openmpi against amdclang you need to change the amdclang compiler to have `self.mpicc` as amdclang or it's gonna compile with gcc

mloubout (changes requested):
I think there is a few things that need change this looks like a start but will create lots of issues

devito/arch/compiler.py
FabioLuporini: ah OK, probably still using the old image?

mloubout: Unless you pushed a new one by hand yes it's still using the old one

docker/Dockerfile.amd
```diff
@@ -2,94 +2,170 @@
 # This Dockerfile contains AMD compilers
 ##############################################################
```
FabioLuporini, line 3: Perhaps we add a reference to https://github.com/amd/InfinityHub-CI/tree/main/base-gpu-mpi-rocm-docker ?

docker/Dockerfile.amd
FabioLuporini: we can drop `which mpicc` yeah

docker/Dockerfile.amd
```diff
@@ -2,94 +2,170 @@
...
 
+
 ########################################################################
 # HIPCC for GPUs (HIP)
 ########################################################################
 # This will only trigger if arch is hip since the final stage depends on it
 FROM sdk-base as hip
 
+# MPI env
+ENV OMPI_CC="hipcc"
```
FabioLuporini, line 160: Note: in my experiments, I haven't tried OMPI_CC=hipcc. 

And even if I did, nothing would have changed.

Remember that the HipCompiler class sets mpicc=hipcc already , so we can actually drop these lines and, if we want to, leave a comment about the above

docker/Dockerfile.amd
```diff
@@ -2,94 +2,170 @@
...
+    build-essential \
+    autoconf \
+    automake \
+    libtool \
+    pkg-config \
+    libnuma* \
+    gfortran \
+    flex \
+    hwloc \
+  && apt-get clean
```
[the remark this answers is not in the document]
mloubout: Tmux is there in the list and o added back tmpi below already

docker/Dockerfile.amd
```diff
@@ -2,94 +2,170 @@
...
+
+# Until rocm base has it fixed
+RUN ln -s /opt/rocm/llvm/bin/offload-arch /opt/rocm/bin/offload-arch
```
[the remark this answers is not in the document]
mloubout: Ok lemme check it they pushed and will remove

docker/Dockerfile.amd
mloubout: I would keep it for homogeneity in case for example someone uses a fancy version of customcompiler

FabioLuporini: OK

docker/Dockerfile.amd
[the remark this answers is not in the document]
mloubout: Nope not pushed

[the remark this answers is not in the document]
mloubout: No that's not fixed it's a hack in their dockerfile the same way we have it here, but it has to be fixed directly into `rocm` and no idea when they'll do that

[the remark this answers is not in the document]
FabioLuporini: sorry you're right, when I wrote the stuff above my head was still in INCLUDE+ mode. And clearly "fixed" was poor wording. OK

docker/Dockerfile.nvidia
```diff
@@ -185,7 +187,7 @@ RUN cd /llvm-project/build && \
     -DCLANG_OPENMP_NVPTX_DEFAULT_ARCH=sm_86 \
     -DLIBOMPTARGET_NVPTX_COMPUTE_CAPABILITIES=all \
     -DCMAKE_INSTALL_PREFIX=/llvm && \
-    make -j $(( $(lscpu | awk '/^Socket\(s\)/{ print $2 }') * $(lscpu | awk '/^Core\(s\) per socket/{ print $4 }') )) && \
+    make -j ${nproc} && \
```
FabioLuporini, line 190: ah! what's this?

mloubout: I think it's a predefined env var by docker makes thing much simpler

mloubout (approved):

PR #2107, opened 2023-04-14
compiler: Introduce gpu-create parameter for buffers initialised on device

ccuetom:
Introduces a parameter `devicecreate` that defines a list of buffers to be created on the device and initialised to zero instead of copied from host to device. 

For a buffer `u` to be created on the device, an operator can be created using `op(..., devicecreate=u, ...)`.

For a series of buffers `u` and `v` the operator can be created using `op(..., devicecreate=(u, v), ...)`.

The PR also marks recursive compilations with the flag `is_rcompile` and changes the `iet_pass` decorator to accept an optional parameter `iet_pass(skipif_rcompile=True)` that skips passes if part of a recursive compilation. This prevents infinite recursions when calling `rcompile` from an `iet_pass`.

### DIFF
--- a/devito/core/cpu.py
+++ b/devito/core/cpu.py
@@ -70,6 +70,7 @@ class Cpu64OperatorMixin(object):
         # Recognised but unused by the CPU backend
         oo.pop('par-disabled', None)
         oo.pop('gpu-fit', None)
+        oo.pop('devicecreate', None)
 
         if oo:
             raise InvalidOperator("Unrecognized optimization options: [%s]"
--- a/devito/core/cpu.py
+++ b/devito/core/cpu.py
@@ -180,7 +180,7 @@ class Cpu64AdvOperator(Cpu64OperatorMixin, CoreOperator):
         hoist_prodders(graph)
 
         # Symbol definitions
-        cls._Target.DataManager(**kwargs).process(graph)
+        cls._Target.DataManager(**kwargs).process(graph, **kwargs)
 
         # Linearize n-dimensional Indexeds
         linearize(graph, **kwargs)
--- a/devito/core/cpu.py
+++ b/devito/core/cpu.py
@@ -70,7 +70,7 @@ class Cpu64OperatorMixin(object):
         # Recognised but unused by the CPU backend
         oo.pop('par-disabled', None)
         oo.pop('gpu-fit', None)
-        oo.pop('devicecreate', None)
+        oo.pop('gpu-create', None)
 
         if oo:
             raise InvalidOperator("Unrecognized optimization options: [%s]"
--- a/devito/core/cpu.py
+++ b/devito/core/cpu.py
@@ -66,6 +66,7 @@ class Cpu64OperatorMixin(object):
         o['linearize'] = oo.pop('linearize', False)
         o['mapify-reduce'] = oo.pop('mapify-reduce', cls.MAPIFY_REDUCE)
         o['index-mode'] = oo.pop('index-mode', cls.INDEX_MODE)
+        o['place-transfers'] = oo.pop('place-transfers', True)
 
         # Recognised but unused by the CPU backend
         oo.pop('par-disabled', None)
@@ -180,7 +181,7 @@ class Cpu64AdvOperator(Cpu64OperatorMixin, CoreOperator):
         hoist_prodders(graph)
 
         # Symbol definitions
-        cls._Target.DataManager(**kwargs).process(graph, **kwargs)
+        cls._Target.DataManager(**kwargs).process(graph)
 
         # Linearize n-dimensional Indexeds
         linearize(graph, **kwargs)
--- a/devito/core/gpu.py
+++ b/devito/core/gpu.py
@@ -81,6 +81,7 @@ class DeviceOperatorMixin(object):
         o['linearize'] = oo.pop('linearize', False)
         o['mapify-reduce'] = oo.pop('mapify-reduce', cls.MAPIFY_REDUCE)
         o['index-mode'] = oo.pop('index-mode', cls.INDEX_MODE)
+        o['place-transfers'] = oo.pop('place-transfers', True)
 
         if oo:
             raise InvalidOperator("Unsupported optimization options: [%s]"
@@ -210,7 +211,7 @@ class DeviceAdvOperator(DeviceOperatorMixin, CoreOperator):
         hoist_prodders(graph)
 
         # Symbol definitions
-        cls._Target.DataManager(**kwargs).process(graph, **kwargs)
+        cls._Target.DataManager(**kwargs).process(graph)
 
         # Linearize n-dimensional Indexeds
         linearize(graph, **kwargs)
--- a/devito/core/gpu.py
+++ b/devito/core/gpu.py
@@ -73,7 +73,7 @@ class DeviceOperatorMixin(object):
         o['par-nested'] = np.inf  # Never use nested parallelism
         o['par-disabled'] = oo.pop('par-disabled', True)  # No host parallelism by default
         o['gpu-fit'] = as_tuple(oo.pop('gpu-fit', cls._normalize_gpu_fit(**kwargs)))
-        o['devicecreate'] = as_tuple(oo.pop('devicecreate', ()))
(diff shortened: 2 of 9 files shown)


tests/test_gpu_common.py
```diff
@@ -1128,6 +1128,44 @@ def test_xcor_from_saved(self, opt, opt_options, gpu_fit):
 
         assert np.all(g.data == 30)
 
+    def test_devicecreate_forward(self):
```
FabioLuporini, line 1131: on second thought, I think we need to check codegen. Something very minimal -- eg to make sure that there's indeed a create(...) and not a mapto(). 

```
if openacc:
    assert 'create(u)' in str(...)
elif openmp
    assert '...'
```

(it's target-language specific; you can pull the target language from `configuration`)

devito/core/cpu.py
```diff
@@ -179,7 +180,7 @@ def _specialize_iet(cls, graph, **kwargs):
         hoist_prodders(graph)
 
         # Symbol definitions
-        cls._Target.DataManager(**kwargs).process(graph)
+        cls._Target.DataManager(**kwargs).process(graph, **kwargs)
```
FabioLuporini, line 183: see comments below -- but basically you don't need to pass kwargs here

devito/passes/iet/definitions.py
```diff
@@ -393,7 +393,7 @@ def place_casts(self, iet, **kwargs):
 
         return iet, {}
 
-    def process(self, graph):
+    def process(self, graph, **kwargs):
```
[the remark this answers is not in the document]
ccuetom: These `**kwargs` are needed to have `is_rcompile` reach the `iet_pass` wrapper

devito/passes/iet/definitions.py
```diff
@@ -448,15 +449,43 @@ def _map_function_on_high_bw_mem(self, site, obj, storage, devicerm, read_only=F
...
+            name = self.sregistry.make_name(prefix='init')
+            efunc = make_callable(name, init)
+            init = Call(name, efunc.parameters)
+
+            mmap = (mmap, init)
+        else:
+            mmap = self.lang._map_to(obj)
+            efunc = ()
 
         if read_only is False:
```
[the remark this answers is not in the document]
ccuetom: probably not now that I think of it

devito/passes/__init__.py
```diff
@@ -39,6 +39,36 @@ def is_on_device(obj, gpu_fit):
...
+
+    fcreate = []
+    for i in functions:
+        try:
+            f = i.alias or i
+        except AttributeError:
+            f = i
+        fcreate.append(f)
+
+    return all(f in gpu_create for f in fcreate)
```
mloubout, line 69: nitpicking in case list is "long". You don't actually need other `fcreate` you just need to return `false` at the first `i` not in `gpu_create` and return true at the end

devito/passes/iet/definitions.py
```diff
@@ -448,15 +450,34 @@ def _map_function_on_high_bw_mem(self, site, obj, storage, devicerm, read_only=F
...
+
+                name = self.sregistry.make_name(prefix='init')
+                efunc = make_callable(name, init)
+                init = Call(name, efunc.parameters)
+
+                mmap = (mmap, init)
+            else:
+                mmap = self.lang._map_to(obj)
+                efunc = ()
+
```
mloubout, line 472: Conceptual question. If the Function is initialized on device, the "assumption" is that that function can be considered "temporary" so should the `_map_update` be optional as well, i.e never sent back to host just freed

devito/passes/iet/definitions.py
```diff
@@ -556,3 +579,18 @@ def process(self, graph):
         self.place_devptr(graph)
         self.place_bundling(graph)
         self.place_casts(graph)
+
+
+def make_zero_init(obj):
+    cdims = []
+    for d, (h0, h1), s in zip(obj.dimensions, obj._size_halo, obj.symbolic_shape):
```
mloubout, line 586: I thought we already had a version of that in `initialize_function` no?

tests/test_gpu_common.py
```diff
@@ -1128,6 +1128,60 @@ def test_xcor_from_saved(self, opt, opt_options, gpu_fit):
...
+            assert 'create(u' in str(op)
+        elif language == 'openmp':
+            assert 'map(alloc: u' in str(op)
+        assert 'init0(u_vec' in str(op)
+
+        op.apply(time_M=nt - 2)
+
+        assert np.all(u.data[0] == 36)
+        assert np.all(u.data[1] == 35)
+
```
mloubout, line 1184: Should `place-transfers` be tested out as well separatrly

mloubout (commented):
Some minor comments, looks nice

devito/passes/iet/definitions.py
ccuetom: I think there might be cases when it's temporary and others in which it's not. The assumption is that these Functions are zero-initialised and therefore do not need to be copied from host, but their final value might be relevant to the user. The example I can think of is the computed wavefield in a forward solve. It would be good to have separate levers to control this, though

devito/passes/iet/definitions.py
ccuetom: There is some overlap with `initialize_function`, although in that case the initialisation is applied in-place when the function is called using the host-side buffer. In the case here, the initialisation is called within the operator. Not sure if there is some code redundancy that could be avoided?

tests/test_gpu_common.py
ccuetom: I can do that. Should I add the test to `test_gpu_common.py::TestStreaming`?

devito/core/gpu.py
```diff
@@ -73,6 +73,7 @@ def _normalize_kwargs(cls, **kwargs):
         o['par-nested'] = np.inf  # Never use nested parallelism
         o['par-disabled'] = oo.pop('par-disabled', True)  # No host parallelism by default
         o['gpu-fit'] = as_tuple(oo.pop('gpu-fit', cls._normalize_gpu_fit(**kwargs)))
+        o['devicecreate'] = as_tuple(oo.pop('devicecreate', ()))
```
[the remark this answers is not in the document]
FabioLuporini: agreed

devito/passes/__init__.py
```diff
@@ -39,6 +39,37 @@ def is_on_device(obj, gpu_fit):
...
+    False otherwise. Objects created in the device memory are zero-initialised.
+
+    Parameters
+    ----------
+    obj : Indexed or Function or collection of Functions
+        The target object.
+    gpu_create : list of Function
+        The Function's which are expected to be created in device memory. This
+        information is given directly by the user through the compiler option
+        `devicecreate` and is propagated down here through the various stages of lowering.
```
FabioLuporini, line 54: gpu-create

devito/passes/__init__.py
```diff
@@ -39,6 +39,37 @@ def is_on_device(obj, gpu_fit):
     return all(f in gpu_fit for f in fsave)
 
 
+def is_gpu_create(obj, gpu_create):
+    """
+    True if the given object is created and not copied in the device memory,
```
FabioLuporini, line 44: :%s/object is/objects are

?

devito/passes/iet/definitions.py
FabioLuporini: > should the _map_update be optional as well

I think it's impossible to say in general. If I'm writing a test, I for sure want an update. Maybe an application code won't care. I agree in the future we may need further opt-options to e.g. distinguish between what is to be updated back and what not

devito/passes/iet/definitions.py
FabioLuporini: yeah, this **might** be one of those rare cases in which it's preferable to keep them separated, as there are minor differences which would make generalisation probably more tricky to understand and maintain than anything... but open for debate

FabioLuporini (approved):
Left tiny comments, but imho we can land this

FabioLuporini (approved):
LGTM, running tests

FabioLuporini (approved):
GTG. Next in the merge pipeline... Thanks!

PR #2109, opened 2023-04-15
mpi: Packed gathers and scatters

FabioLuporini:


devito/ir/support/basic.py
```diff
@@ -534,6 +534,67 @@ def distance_mapper(self):
                 retval[d] = j
         return retval
 
+    @cached_property
```
FabioLuporini, line 537: all these properties here are simply lifted from what is now a subclass of Relation, namely Dependence , so this is not new code

docker/Dockerfile.cpu
```diff
@@ -24,7 +24,20 @@ RUN apt-get install -y libgl1-mesa-glx
 ##############################################################
 FROM base as gcc
 
-RUN apt-get install -y mpich libmpich-dev
+RUN mkdir -p /deps && mkdir -p /opt/openmpi && cd /deps && \
```
FabioLuporini, line 27: @mloubout for homogeneity, switching to OpenMPI here

mloubout: I don't mind switching but this is the expensive route, for standard CPU `apt install libopenmpi-dev` should be enough (gotta double check name)

FabioLuporini: but last I checked you would get openmpi 3, this way you get 4 (but I may be wrong)

devito/ir/support/basic.py
```diff
@@ -1013,3 +1030,28 @@ def d_from_access(self, accesses):
...
+        Generate the Relations of the Scope.
+        """
+        for f in self.functions:
+            v = self.reads.get(f, []) + self.writes.get(f, [])
+
+            for a0, a1 in product(v, v):
+                if a0 is a1:
+                    continue
+
+                r = Relation(a0, a1)
```
mloubout, line 1046: Is `Relation` commutative? product is quite brutal (might be ok if small set) so if it's commutative you are better of with two for loop `(i, a0) in enumerate(v), a1 in v[I:-1]`

devito/tools/dtypes_lowering.py
```diff
@@ -24,34 +24,70 @@
     "double": np.float64
 }
 
+
+def build_dtypes_vector(field_names, counts):
+    ret = {}
+    for base_name, base_dtype in mapper.items():
```
mloubout, line 30: `mapper` might be dangerously generic of a name with how much we use it internally

docker/Dockerfile.cpu
mloubout: Ok, it's fine

`mkdir -p` lets you do any path so just `mkdir -p /opt/openmpi`

devito/ir/support/basic.py
FabioLuporini: it's not commutative because the distance would flip (positive -> negative or negative -> positive)

devito/mpi/halo_scheme.py
```diff
@@ -362,20 +362,30 @@ def arguments(self):
...
+        corresponding to the provided `functions`.
         """
         fmapper = {f: v for f, v in self.fmapper.items() if f not in as_tuple(functions)}
         return HaloScheme.build(fmapper, self.honored)
 
+    def add(self, f, hse):
+        """
+        Create a new HaloScheme that contains all entries in `self` plus the one
+        passed in input. If `f` already exists in `self`, the old value is
+        overridden.
```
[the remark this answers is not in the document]
FabioLuporini: it's classic dict behaviour:

```
a[2] = 3
a[2] = 4
```

Now `a[2]` contains 4

IOW, it does _not_ behave as `dict.setdefault`

devito/tools/dtypes_lowering.py
FabioLuporini: I think it's fine here, it's used loosely for generic maps, and it's not exported to the outside modules of course, otherwise yes it'd be a pretty bad choice

devito/ir/support/basic.py
```diff
@@ -482,20 +482,20 @@ def touched_halo(self, findex):
         return (touch_halo_left, touch_halo_right)
 
 
-class Dependence(object):
+class Relation(object):
 
     """
-    A data dependence between two TimedAccess objects.
+    A relation between two TimedAccess objects.
```
georgebisbas, line 488: Why not also keep the data dependence notion in the docstring?

devito/tools/dtypes_lowering.py
```diff
@@ -78,21 +114,18 @@ def dtype_to_ctype(dtype):
 
 def dtype_to_mpitype(dtype):
     """Map numpy types to MPI datatypes."""
```
georgebisbas, line 116: NumPy?

devito/mpi/halo_scheme.py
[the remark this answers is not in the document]
mloubout: I meant what is the reason for the new value to have priority.

devito/tools/dtypes_lowering.py
mloubout: ok

docker/Dockerfile.cpu
mloubout: missread `deps` for `openmpi` my bad

mloubout (approved):
Some minor comments but looks good

devito/mpi/halo_scheme.py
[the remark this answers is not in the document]
FabioLuporini: It's just the semantics I've established for this method. It seemed more convenient at the caller sites

devito/mpi/routines.py
```diff
@@ -67,6 +68,11 @@ def make(self, hs):
         # Sanity check
         assert all(f.is_Function and f.grid is not None for f in hs.fmapper)
 
+        # Pack-up Functions into Bundles. Worst case scenario, we have Bundles
+        # with just one components each. This is to maximize the likelihood of
```
georgebisbas, line 72: one component

devito/mpi/routines.py
```diff
@@ -307,32 +347,38 @@ def _make_copy(self, f, hse, key, swap=False):
...
 
+        eqns = []
+        eqns.extend([Eq(d.symbolic_min, 0) for d in dims])
+        eqns.extend([Eq(d.symbolic_max, d.symbolic_size - 1) for d in dims])
+
+        bdims = [CustomDimension(name='vd', symbolic_size=f.ncomp)] + dims
+        buf = Array(name='buf', dimensions=bdims, dtype=f.c0.dtype, padding=0)
         if swap is False:
-            eq = Eq(buf[dims], f[f_indices])
+            swap = lambda i, j: (i, j)
```
georgebisbas, line 357: For readability, ok!

devito/tools/dtypes_lowering.py
```diff
@@ -24,34 +24,70 @@
...
+    ret = {}
+    for base_name, base_dtype in mapper.items():
+        for count in counts:
+            name = "%s%d" % (base_name, count)
+
+            titles = field_names[:count]
+
+            padded_count = count
+            if count == 3:
+                padded_count = 4
```
georgebisbas, line 38: Could this be better documented/explained? Not sure I get it

georgebisbas (approved):

PR #2114, opened 2023-04-19
compiler: Add fission-for-pressure subpass

FabioLuporini:
So far, this isn't enabled by default. More experimentation required

This PR also adds a new compiler flag to GNUCompiler when running on Skylake and later processors. I've tested it in a few places, and it seems to generally improve things. I'd like people to reproduce it. Note that you need to ensure DEVITO_PLATFORM is properly set to skx, clk, ... the autodetection mechanism, with the recent proliferation of intel architectures, doesn't seem to be as good as it was.

### DIFF
--- a/devito/passes/clusters/fission.py
+++ b/devito/passes/clusters/fission.py
@@ -1,6 +1,7 @@
 from itertools import groupby
 
 from devito.ir import Queue, Scope
+from devito.symbolics import retrieve_terminals
 from devito.tools import Stamp, flatten, frozendict, timed_pass
 
 __all__ = ['fission']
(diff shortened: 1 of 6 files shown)


devito/passes/clusters/fission.py
```diff
@@ -0,0 +1,165 @@
...
+            functions0 = {i.function for i in terminals0 if i.is_Indexed}
+            functions1 = {i.function for i in terminals1 if i.is_Indexed}
+            functions_shared = functions0.intersection(functions1)
+
+            n0 = len(functions0)
+            n1 = len(functions1)
+            ns = len(functions_shared)
+
+            if not ns:
+                ns = .001
```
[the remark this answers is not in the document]
georgebisbas: Maybe add some comment for this, even though not really necessary? Just to justify this option

tests/test_fission.py
```diff
@@ -3,125 +3,204 @@
...
-def test_nofission_as_unprofitable():
-    """
-    Test there's no fission if not gonna increase number of collapsable loops.
-    """
-    grid = Grid(shape=(20, 20))
-    x, y = grid.dimensions
-    t = grid.stepping_dim
+    def test_nofission_as_unprofitable(self):
+        """
+        Test there's no fission if not gonna increase number of collapsable loops.
```
EdCaunt, line 43: "gonna" -> "going to"?

tests/test_fission.py
```diff
@@ -3,125 +3,204 @@
...
-            Eq(u.forward, f[y + 1] + 1.)]
+        op = Operator(eqns, opt='fission')
 
-    op = Operator(eqns, opt='fission')
+        assert_structure(op, ['t,x,y', 't,x,y'], 't,x,y,y')
 
-    assert_structure(op, ['t,x,y', 't,x,y'], 't,x,y,y')
+    def test_fission_partial(self):
+        """
+        Test there's no fission if not gonna increase number of collapsable loops.
```
EdCaunt, line 81: See previous comment

PR #2116, opened 2023-04-26
compiler: Add host-*-pin handles; more volatile with pthreads

FabioLuporini:


devito/ir/iet/utils.py
```diff
@@ -122,3 +123,34 @@ def derive_parameters(iet, drop_locals=False):
...
+
+
+def maybe_alias(obj, candidate):
+    """
+    True if `candidate` can act as an alias for `obj`, False otherwise.
+    """
+    if obj is candidate:
+        return True
+
+    # Names are unique throughout compilation, so this is another case we can handle
```
georgebisbas, line 135: the compilation (?)

devito/passes/iet/asynchrony.py
```diff
@@ -208,3 +213,20 @@ def lower_async_calls(iet, track=None, sregistry=None):
...
 
     return iet, {'efuncs': tuple(efuncs.values())}
+
+
+# *** Utils
+
+def sanitize_ncfields(ncfields):
+    # Due to a bug in the NVC compiler (v<=22.7 and potentially later),
+    # we have to use C's `volatile` more extensively than strictly necessary
+    # to avoid flaky optimizations that would cause fauly behaviour in rare,
```
georgebisbas, line 223: faulty*

devito/ir/iet/utils.py
```diff
@@ -122,3 +123,34 @@ def derive_parameters(iet, drop_locals=False):
...
+    # straightforwardly. It might happen that we have an alias used in a subroutine
+    # with different type qualifiers (e.g., const vs not const, volatile vs not
+    # volatile), but if the names match, they definitely represent the same
+    # logical object
+    if obj.name == candidate.name:
+        return True
+
+    if obj.is_AbstractFunction:
```
mloubout, line 143: Are these case really needed? Will obviously always return false passed this line no matter

mloubout (commented):
Looks mostly fine, not sure what this volatile is but I trust you on it

devito/ir/iet/utils.py
FabioLuporini: not strictly speaking and not today, but they were in the back of my mind already, so I put them there for the future

PR #2119, opened 2023-05-05
ci: switch to concurrency settings rather than extra action

mloubout:
See

https://docs.github.com/en/actions/using-jobs/using-concurrency


georgebisbas (approved):
Looks fine I would say

FabioLuporini (approved):
uncontroversial

PR #2123, opened 2023-05-11
deps: sympy 1.12 compat

mloubout:


georgebisbas (approved):

devito/core/autotuning.py
```diff
@@ -273,6 +274,14 @@ def calculate_nblocks(tree, blockable):
 
 
 def generate_block_shapes(blockable, args, level):
+    # Make sure all params are substitutable (ie cannot raise SympifyError)
```
mloubout, line 277: Note: I know this is not the pretiest but this errors in the latest sympy because we have some `CParams` in there that are not sympify-able, see here for new sympy behavior

https://github.com/sympy/sympy/pull/23887

FabioLuporini: OK

PR #2124, opened 2023-05-11
mpi: Enhance flexibility for custom topologies

georgebisbas:
This PR enhances the available custom topology MPI decompositions

The requirements for the number of stars to evenly divide the number of MPI procs is dropped:
    N must evenly divide the number of `'*'`, otherwise a ValueError exception
    is raised.

You can see some topology to decomposition combos in the added tests: (e.g.)

```
    @pytest.mark.parametrize('comm_size, topology, dist_topology', [
        (1, (1, '*', '*'), (1, 1, 1)),
        (2, (1, '*', '*'), (1, 2, 1)),
        (3, (1, '*', '*'), (1, 3, 1)),
        (3, ('*', 1, '*'), (3, 1, 1)),
        (3, ('*', '*', 1), (3, 1, 1)),
        (6, ('*', 1, '*'), (3, 1, 2)),
        (6, ('*', '*', 1), (3, 2, 1)),
        (6, (1, '*', '*'), (1, 3, 2)),
        (6, ('*', '*', '*'), (2, 3, 1)),
        (8, ('*', 1, '*'), (4, 1, 2)),
        (8, ('*', '*', 1), (4, 2, 1)),
        (8, (1, '*', '*'), (1, 4, 2)),
        (8, ('*', '*', '*'), (2, 2, 2)),
        (9, ('*', '*', '*'), (3, 3, 1)),
        (11, (1, '*', '*'), (1, 11, 1)),
    ])
```

### DIFF
--- a/devito/mpi/distributed.py
+++ b/devito/mpi/distributed.py
@@ -1,7 +1,9 @@
+from abc import ABC, abstractmethod
 from ctypes import c_int, c_void_p, sizeof
 from itertools import groupby, product
-from math import ceil
-from abc import ABC, abstractmethod
+from math import ceil, pow
+from sympy import root, divisors
+
 import atexit
 
 from cached_property import cached_property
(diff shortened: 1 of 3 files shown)


mloubout (approved):

tests/test_mpi.py
```diff
@@ -180,6 +180,44 @@ def test_custom_topology(self):
         assert f2.shape == expected[distributor.myrank]
         assert f2.size_global == f.size_global
 
+    @pytest.mark.parametrize('topology, dist_topology, expected', [
+        (('*', 1, 1), (4, 1, 1), [(4, 15, 15), (4, 15, 15), (4, 15, 15), (3, 15, 15)]),
+        ((1, '*', 1), (1, 4, 1), [(15, 4, 15), (15, 4, 15), (15, 4, 15), (15, 3, 15)]),
+        ((1, 1, '*'), (1, 1, 4), [(15, 15, 4), (15, 15, 4), (15, 15, 4), (15, 15, 3)]),
+        (('*', 1, '*'), (2, 1, 2), [(8, 15, 8), (8, 15, 7), (7, 15, 8), (7, 15, 7)]),
+    ])
+    @pytest.mark.parallel(mode=[4])
```
FabioLuporini, line 189: Can we test the class CustomTopology "in isolation", that is without launching an actual MPI test? MPI tests are slow (because they fork new processes etc), so if avoidable it would be better. Here they should be avoidable with some expedients

georgebisbas: As far as I am aware, all the tests that use a Distributor are marked parallel, i.e. launching jobs.
Do we have an example of not doing that?

georgebisbas: I can sketch something but not sure I like it or want it.
Also I believe it is worth to have it running w mpi, but up to you.

georgebisbas: I pushed an idea with a dummy communicator class, but only for the custom topology.
The current Distributor implementation resorts to (1,1,1) in case of a non-MPI init communicator present.

FabioLuporini: > As far as I am aware, all the tests that use a Distributor are marked parallel, i.e. launching jobs.
Do we have an example of not doing that?

yes in this file IIRC

the new test with the dummy communicator is exactly what I had in mind. It's lightweight and it achieves the objective of the test we need to add

tests/test_mpi.py
```diff
@@ -12,10 +12,19 @@
                            retrieve_iteration_tree)
 from devito.mpi import MPI
 from devito.mpi.routines import HaloUpdateCall, MPICall
+from devito.mpi.distributed import CustomTopology, Distributor
 from examples.seismic.acoustic import acoustic_setup
 
 pytestmark = skipif(['nompi'], whole_module=True)
 
+class Dummy_input_comm():
```
mloubout, line 20: can get away with just a namedtuple but that's fine with me like that too

FabioLuporini: also note the chosen class name ain't pythonic

devito/mpi/distributed.py
```diff
@@ -608,7 +621,21 @@ def __new__(cls, items, input_comm):
                 raise ValueError("Custom topology must be only 1 or *")
 
             v = input_comm.size // nstars
-            processed = [i if i == 1 else v for i in items]
+
+            # In case more than one stars are present, this routine
```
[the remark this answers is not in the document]
georgebisbas: Fixing this now

devito/mpi/distributed.py
```diff
@@ -583,13 +584,33 @@ class CustomTopology(tuple):
 
     * `(1, '*', 1)`: the wildcard `'*'` tells the runtime to decompose the y
                      Dimension into N chunks
-    * `('*', '*', 1)`: the wildcard `'*'` tells the runtime to decompose both the
-                       x and y Dimensions into N / 2 chunks respectively.
+    * `('*', '*', 1)`: the wildcard `'*'` tells the runtime to decompose both
```
georgebisbas, line 587: improving these lines a bit

georgebisbas:
Closing temporarily, have some ideas to further improve. Will re-open

PR #2128, opened 2023-05-15
api: Revamp interpolation/injection

### LINKED ISSUE #920: DSE uses wrong np.dtype for some temporaries
Reproducer below

```
import numpy as np
from devito import Grid, Function, Eq, Operator, generic_derivative


a = np.arange(64).reshape((8, 8))
grid = Grid(shape=a.shape, extent=(8, 8))

so = 3
dims = grid.dimensions
f = Function(name='f', grid=grid, space_order=so, dtype=np.int32)
f.data[:] = a

fo = Function(name='fo', grid=grid, space_order=so, dtype=np.int32)

expr = Eq(fo, f.dx)

op = Operator(expr)
op.apply()
```

basically the DSE generates `int r0 = 1.0/h_x;`, inferring `int` from the dtype of the searched expression. But then `r0` will be 0, and so everything will be 0 at the end. We may wanna force the temporary dtype to float. Have to think about the potential side effects of doing that though

### LINKED ISSUE #2147: source injection onto multiple time functions when source term has non-grid dimensions fails
```
import devito

nt = 10
nfreq = 2
g = devito.Grid(shape=(5,6))
u1 = devito.TimeFunction(name="u1", grid=g, time_order=0)
u2 = devito.TimeFunction(name="u2", grid=g, time_order=0)

stf = devito.SparseTimeFunction(name="stf", nt=nt, npoint=1, grid=g)
stf.data[:] = 1.
stf.coordinates_data[:] = 0.5
freq_dim = devito.DefaultDimension(name="freq_dim", default_value=nfreq)
f = devito.Function(name="f", dimensions=(freq_dim,), shape=(nfreq,), grid=g)
f.data[:] = 2.

inj_u1 = stf.inject(field=u1.forward, expr=stf*f)
inj_u2 = stf.inject(field=u2.forward, expr=stf*f)

op1 = devito.Operator([inj_u1], opt=("advanced", {"min-storage":True}), name="inj_u1")
print(op1.parameters)
op1.apply()

op2 = devito.Operator([inj_u2], opt=("advanced", {"min-storage":True}), name="inj_u2")
print(op2.parameters)
op2.apply()

op3 = devito.Operator([inj_u1, inj_u2], opt=("advanced", {"min-storage":True}), name="Failing")
print(op3.parameters)
op3.apply()
```
this results in:
```
cvx@cbox-lukedecker-ela2d:~/.julia/dev/JetPackDevitoPSD/test$ python mfe.py 
(f(freq_dim), stf(time, p_stf), stf_coords(p_stf, d), u1(t, x, y), x_M, x_m, y_M, y_m, freq_dim_M, freq_dim_m, h_x, h_y, o_x, o_y, p_stf_M, p_stf_m, time_M, time_m, nthreads_nonaffine, timers)
Operator `inj_u1` ran in 0.04 s
(f(freq_dim), stf(time, p_stf), stf_coords(p_stf, d), u2(t, x, y), x_M, x_m, y_M, y_m, freq_dim_M, freq_dim_m, h_x, h_y, o_x, o_y, p_stf_M, p_stf_m, time_M, time_m, nthreads_nonaffine, timers)
Operator `inj_u2` ran in 0.01 s
(f(freq_dim), stf(time, p_stf), stf_coords(p_stf, d), u1(t, x, y), u2(t, x, y), x_M, x_m, y_M, y_m, freq_dim_M, freq_dim_m, h_x, h_y, o_x, o_y, p_stf_M, p_stf_m, posx, posy, time_M, time_m, nthreads_nonaffine, timers)
  File "/home/cvx/.julia/dev/JetPackDevitoPSD/test/mfe.py", line 29, in <module>
    op3.apply(posx=0.5, posy=0.5)
  File "/home/cvx/.conda/envs/conda_jl/lib/python3.10/site-packages/devito/operator/operator.py", line 812, in apply
  File "/home/cvx/.conda/envs/conda_jl/lib/python3.10/site-packages/devito/operator/operator.py", line 653, in arguments
  File "/home/cvx/.conda/envs/conda_jl/lib/python3.10/site-packages/devito/operator/operator.py", line 604, in _prepare_arguments
    p._arg_check(args, self._dspace[p], am=self._access_modes.get(p))
AttributeError: 'Symbol' object has no attribute '_arg_check'
```

### LINKED ISSUE #2189: Handing of non-integer sparse time function coordinates with MPI
Consider the following MFE:
```
import devito as dv 
import numpy as np 

nx = 12
ny = 12
nz = 12
# extent_x = nx-1
# extent_y = ny-1
# extent_z = nz-1
for extent in (nx-1, 1):
    extent_x = extent
    extent_y = extent
    extent_z = extent
    g = dv.Grid(shape=(ny,ny,nz), extent=(extent_x,extent_y,extent_z))

    coords = np.zeros((nx*ny,3))
    
    cx = np.zeros(12*12)
    cy = np.zeros(12*12)
    cz = np.zeros(12*12)

    dx = extent_x/(nx-1)
    dy = extent_y/(ny-1)
    dz = extent_z/(nz-1)
    for ix in range(0,12):
        my_x = ix*dx
        for iy in range(0,12):
            indx = ix*(12) + iy
            my_y = iy*dy 
            cx[indx] = my_x
            cy[indx] = my_y
            cz[indx] = dz

    coords[:,0] = cx
    coords[:,1] = cy 
    coords[:,2] = cz

    (x,y,z) = g.dimensions

    f = dv.TimeFunction(name="f", grid=g)

    stf = dv.SparseTimeFunction(name="stf", npoint=12*12, nt=2, grid=g, coordinates=coords)

    op = dv.Operator([dv.Eq(f,1), stf.interpolate(expr=f)])

    op.apply()
    print(stf.data[1,:])
```

Running without MPI produces this:
```
cvx@cbox-lukedecker-mpi:~/.julia/dev/Devito/test$ DEVITO_MPI=0 python testrecs.py 
Operator `Kernel` ran in 0.03 s
[1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1.
 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1.
 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1.
 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1.
 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1.
 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
Operator `Kernel` ran in 0.01 s
[1.        1.        1.        1.        1.        1.        1.
 1.        1.        1.        1.        1.0000004 1.        1.
 1.        1.        1.        1.        1.        1.        1.
 1.        1.        1.0000004 1.        1.        1.        1.
 1.        1.        1.        1.        1.        1.        1.
 1.0000004 1.        1.        1.        1.        1.        1.
 1.        1.        1.        1.        1.        1.0000004 1.
 1.        1.        1.        1.        1.        1.        1.
 1.        1.        1.        1.0000004 1.        1.        1.
 1.        1.        1.        1.        1.        1.        1.
 1.        1.0000004 1.        1.        1.        1.        1.
 1.        1.        1.        1.        1.        1.        1.0000004
 1.        1.        1.        1.        1.        1.        1.
 1.        1.        1.        1.        1.0000004 1.        1.
 1.        1.        1.        1.        1.        1.        1.
 1.        1.        1.0000004 1.        1.        1.        1.
 1.        1.        1.        1.        1.        1.        1.
 1.0000004 1.        1.        1.        1.        1.        1.
 1.        1.        1.        1.        1.        1.0000004 1.0000004
 1.0000004 1.0000004 1.0000004 1.0000004 1.0000
```

Running with MPI everything looks fine for the first case where each dimension's spacing is 1.  But when the extent is default, so the spacing of x = 1/(nx-1) things get strange:

```
cvx@cbox-lukedecker-mpi:~/.julia/dev/Devito/test$ DEVITO_MPI=1 mpirun -n 2 python testrecs.py 
Operator `Kernel` ran in 1.65 s
Operator `Kernel` ran in 1.65 s
[1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1.
 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1.
 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
[1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1.
 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1.
 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]
Operator `Kernel` ran in 2.28 s
Operator `Kernel` ran in 2.28 s
[1.0000000e+00 1.0000000e+00 1.0000000e+00 1.0000000e+00 1.0000000e+00
 1.0000000e+00 1.0000000e+00 1.0000000e+00 1.0000000e+00 1.0000000e+00
 1.0000000e+00 1.0000004e+00 2.9802322e-07 2.9802322e-07 2.9802322e-07
 2.9802322e-07 2.9802322e-07 2.9802322e-07 2.9802322e-07 2.9802322e-07
 2.9802322e-07 3.2782555e-07 2.9802322e-07 3.2782566e-07 0.0000000e+00
 0.0000000e+00 0.0000000e+00 0.0000000e+00 0.0000000e+00 0.0000000e+00
 0.0000000e+00 0.0000000e+00 0.0000000e+00 0.0000000e+00 0.0000000e+00
 0.0000000e+00 0.0000000e+00 0.0000000e+00 0.0000000e+00 0.0000000e+00
 0.0000000e+00 0.0000000e+00 0.0000000e+00 0.0000000e+00 0.0000000e+00
 0.0000000e+00 0.0000000e+00 0.0000000e+00 0.0000000e+00 0.0000000e+00
 0.0000000e+00 0.0000000e+00 0.0000000e+00 0.0000000e+00 0.0000000e+00
 0.0000000e+00 0.0000000e+00 0.0000000e+00 0.0000000e+00 0.0000000e+00
 1.0000000e+00 1.0000000e+00 1.0000000e+00 1.0000000e+00 1.0000000e+00
 1.0000000e+00 1.0000000e+00 1.0000000e+00 1.0000000e+00 1.0000000e+00
 1.0000000e+00 1.0000004e+00]
[1.        1.        1.        1.        1.        1.        1.
 1.        1.        1.        1.        1.0000004 1.        1.
 1.        1.        1.        1.        1.        1.        1.
 1.        1.        1.0000004 1.        1.        1.        1.
 1.        1.        1.        1.        1.        1.        1.
 1.0000004 1.        1.        1.        1.        1.        1.
 1.        1.        1.        1.        1.        1.0000004 1.
 1.        1.        1.        1.        1.        1.        1.
 1.        1.        1.        1.0000004 1.        1.        1.
 1.        1.        1.        1.        1.        1.        1.
 1.        1.0000004]
```

FabioLuporini:
Revamp sparse function.

- Improve hierachy to include PrecomputedSparseFunction
- Add MPI support to PrecomputedSparseFunction
- Revamp interpolation to ease implementation of new Weighted interpolation
- Introduce multi-inject/interp so the user can inject into multiple fields at once

Issue fixed:

Fixes #920 
Fixes #2147
Fixes #2189


devito/types/sparse.py
```diff
@@ -931,23 +936,99 @@ class PrecomputedSparseFunction(AbstractSparseFunction):
...
 
         # Grid points per sparse point (2 in the case of bilinear and trilinear)
         r = kwargs.get('r')
+        if not is_integer(r):
+            raise TypeError('Need `r` int argument')
+        if r <= 0:
+            raise ValueError('`r` must be > 0')
+        self._r = r
+
+        coordinates = kwargs.get('coordinates', kwargs.get('coordinates_data'))
```
mloubout, line 959: Should wrap this copy paste in to an `_coordinate_setup`

devito/types/sparse.py
```diff
@@ -931,23 +936,99 @@ class PrecomputedSparseFunction(AbstractSparseFunction):
...
+        elif coordinates is None:
+            # Unlike `gridpoints` or `interpolation_coeffs`, not strictly necessary
+            self._coordinates = None
+        else:
+            raise ValueError("`coordinates` must be either SubFunction or iterable "
+                             "(e.g., list, np.ndnarray)")
+
         gridpoints = kwargs.get('gridpoints')
-        interpolation_coeffs = kwargs.get('interpolation_coeffs')
+        gridpoints = kwargs.get('gridpoints', kwargs.get('gridpoints_data'))
```
mloubout, line 980: `gridpoints` and `coordinates` cannot be a valid combination only one of the two can be provided.

devito/types/sparse.py
```diff
@@ -931,23 +936,99 @@ class PrecomputedSparseFunction(AbstractSparseFunction):
...
+                initializer=gridpoints_data, alias=self.alias,
+                distributor=self._distributor,
+            )
+        else:
+            raise ValueError("`gridpoints` must be either SubFunction or iterable "
+                             "(e.g., list, np.ndnarray)")
+
+        interpolation_coeffs = kwargs.get('interpolation_coeffs',
+                                          kwargs.get('interpolation_coeffs_data'))
+        if isinstance(interpolation_coeffs, SubFunction):
```
mloubout, line 999: I think could be cleanedup wiht `_coordinate_setup(name=gripoint/coeffs/cords/...` that's a lot of repetition to maintain.

devito/types/sparse.py
```diff
@@ -931,23 +936,99 @@ class PrecomputedSparseFunction(AbstractSparseFunction):
...
-        super(PrecomputedSparseFunction, self).__init_finalize__(*args, **kwargs)
+        super().__init_finalize__(*args, **kwargs)
 
         # Grid points per sparse point (2 in the case of bilinear and trilinear)
         r = kwargs.get('r')
+        if not is_integer(r):
+            raise TypeError('Need `r` int argument')
+        if r <= 0:
+            raise ValueError('`r` must be > 0')
+        self._r = r
```
mloubout, line 957: Need to check against `interpolation_coeffs_data` that radius matches coeffs

mloubout (commented):
Sorry bunch of comments.  I know this hasn't been touched in quite a few yers ut yes is needed to revamp and cleanup

devito/operations/interpolators.py
```diff
@@ -108,15 +107,12 @@ def interpolate(self, *args, **kwargs):
...
 
     """
-    Concrete implementation of GenericInterpolator implementing a Linear interpolation
-    scheme, i.e. Bilinear for 2D and Trilinear for 3D problems.
-
-    Parameters
-    ----------
-    sfunction: The SparseFunction that this Interpolator operates on.
+    Represent an Interpolation operation on a SparseFunction that is separable
+    in space, meaning hte coefficient are defined for each Dimension separately
```
FabioLuporini, line 114: the coefficientS

devito/operations/interpolators.py
```diff
@@ -126,90 +122,96 @@ def __init__(self, sfunction):
...
+    @property
+    def _weights(self):
+        raise NotImplementedError
+
+    @property
+    def _psym(self):
+        return self.sfunction._point_symbols
+
+    @property
+    def _gdim(self):
```
FabioLuporini, line 134: imho, just "dimensions" as we do everywhere else

devito/operations/interpolators.py
```diff
@@ -305,106 +305,64 @@ def callback():
...
-            Input expression to interpolate.
-        offset : int, optional
-            Additional offset from the boundary.
-        increment: bool, optional
-            If True, generate increments (Inc) rather than assignments (Eq).
-        """
-        def callback():
-            _expr = indexify(expr)
+    """
+    Concrete implementation of GenericInterpolator implementing a Linear interpolation
```
FabioLuporini, line 319: of WeightedInterpolator

devito/operations/interpolators.py
```diff
@@ -305,106 +305,64 @@ def callback():
...
-                rd = DefaultDimension(name="r%s" % d.name, default_value=self.r)
-                dim_subs.append((d, INT(rd + self.obj.gridpoints[p, i])))
-                coeffs.append(self.obj.interpolation_coeffs[p, i, rd])
-            rhs = prod(coeffs) * _expr
-            _field = _field.subs(dim_subs)
-            return [Inc(_field, rhs.subs(dim_subs))]
+class PrecomputedInterpolator(WeightedInterpolation):
 
-        return Injection(field, expr, offset, self, callback)
+    def _positions(self, implicit_dims):
```
FabioLuporini, line 341: aren't `implicit_dims` an instance property? do we actually need to pass them or can we turn this into a property? 

btw, some of these properties *might* benefit from `cached_property`

devito/operations/interpolators.py
```diff
@@ -126,90 +122,96 @@ def __init__(self, sfunction):
...
+    @property
+    def _psym(self):
+        return self.sfunction._point_symbols
+
+    @property
+    def _gdim(self):
+        return self.grid.dimensions
+
+    def implicit_dims(self, implicit_dims):
+        return as_tuple(implicit_dims) + self.sfunction.dimensions
```
FabioLuporini, line 138: you know I'm thinking, maybe instead of `sfunction` we could use just `function`, like we do everywhere else

devito/types/sparse.py
```diff
@@ -76,6 +82,67 @@ def __shape_setup__(cls, **kwargs):
...
+            n = 2  # (Sparse points, Grid Dimensions)
+            dtype = self.dtype
+        else:
+            if not isinstance(key, np.ndarray):
+                key = np.array(key)
+                # Correct for corner case of single coordinate
+                n = max(key.ndim, 2)
+            else:
+                n = key.ndim
+            # Need to fix this check to get global npoint, global_shape broken
```
FabioLuporini, line 117: Can we raise an error here if there's an unsupported use case? and perhaps add an x-failing MFE?

devito/types/sparse.py
```diff
@@ -89,10 +156,89 @@ def space_order(self):
         """The space order."""
         return self._space_order
 
+    @property
+    def r(self):
```
FabioLuporini, line 160: what I'm saying here applies everywhere, even to the new WeightedInterpolator class. 

Instance/class properties/methods should always be logically grouped together according to some logic. For example, first all of the private properties, then all of the public properties, then...

Here `r` is in between private stuff, some being methods and some properties

devito/types/sparse.py
```diff
@@ -218,25 +368,159 @@ def _dist_alltoall(self, dmap=None):
...
+    def _dist_subfunc_scatter(self, subfunc):
+        # If not using MPI, don't waste time
+        if self.distributor.nprocs == 1:
+            return {subfunc: subfunc.data}
+
+        # Compute dist map only once
+        dmap = self._dist_datamap
+        mask = self._dist_scatter_mask(dmap=dmap)
+
+        # Pack (reordered) coordinates so that they can be sent out via an Alltoallv
```
FabioLuporini, line 439: instead of coordinates, here you mean "subfunc values" ?

devito/types/sparse.py
```diff
@@ -218,25 +368,159 @@ def _dist_alltoall(self, dmap=None):
...
+
+        # Send out the sparse point coordinates
+        _, scount, sdisp, rshape, rcount, rdisp = \
+            self._dist_subfunc_alltoall(subfunc, dmap=dmap)
+        scattered = np.empty(shape=rshape, dtype=subfunc.dtype)
+        self.comm.Alltoallv([sfuncd, scount, sdisp, self._smpitype[subfunc]],
+                            [scattered, rcount, rdisp, self._smpitype[subfunc]])
+        sfuncd = scattered
+
+        # Translate global coordinates into local coordinates
```
FabioLuporini, line 450: same?

devito/types/sparse.py
```diff
@@ -451,35 +730,11 @@ class SparseFunction(AbstractSparseFunction):
...
-
-    def __distributor_setup__(self, **kwargs):
-        """
-        A `SparseDistributor` handles the SparseFunction decomposition based on
-        physical ownership, and allows to convert between global and local indices.
-        """
-        return SparseDistributor(kwargs['npoint'], self._sparse_dim,
-                                 kwargs['grid'].distributor)
+        self._coordinates = self.__subfunc_setup__(coordinates, 'coords')
+        self._dist_origin = {self._coordinates: self.grid.origin_offset}
```
FabioLuporini, line 737: (cached) property ?

devito/types/sparse.py
```diff
@@ -624,116 +835,6 @@ def _decomposition(self):
...
-        sshape, scount, sdisp, _, rcount, rdisp = self._dist_alltoall(dmap=dmap)
-        gathered = np.empty(shape=sshape, dtype=self.dtype)
-        mpitype = MPI._typedict[np.dtype(self.dtype).char]
-        comm.Alltoallv([data, rcount, rdisp, mpitype],
-                       [gathered, scount, sdisp, mpitype])
-        # Unpack data values so that they follow the expected storage layout
-        gathered = np.ascontiguousarray(np.transpose(gathered, self._dist_reorder_mask))
-        self._data[mask] = gathered[:]
-
-        if coords is not None:
```
FabioLuporini, line 721: I love it that subfunc and data scattering/gathering is now explicitly separated :)

devito/types/sparse.py
```diff
@@ -931,61 +1035,157 @@ class PrecomputedSparseFunction(AbstractSparseFunction):
...
+        self._interpolation_coeffs = \
+            self.__subfunc_setup__(interpolation_coeffs, 'interp_coeffs')
+        self._dist_origin.update({self._interpolation_coeffs: None})
+        # Make sure it matches the radius
+        if self._interpolation_coeffs.shape[-1] != r:
+            nr = self._interpolation_coeffs.shape[-1]
+            raise ValueError("Interpolation coefficients shape %d do "
+                             "not match specified radius %d" % (r, nr))
+
+        warning("Ensure that the provided interpolation coefficient and grid "
```
FabioLuporini, line 1093: Can we remove this warning, or move it to a place s.t. it's triggered only if it's clear the user might have made a mistake?

devito/types/sparse.py
```diff
@@ -931,61 +1035,157 @@ class PrecomputedSparseFunction(AbstractSparseFunction):
...
+        try:
+            return self.gridpoints.data._local
+        except AttributeError:
+            return None
 
-    def _dist_scatter(self, data=None):
-        data = data if data is not None else self.data
-        distributor = self.grid.distributor
+    @property
+    def coordinates(self):
```
FabioLuporini, line 1133: this should be in the superclass? also the one below?

devito/types/sparse.py
```diff
@@ -89,10 +156,89 @@ def space_order(self):
...
+    @property
+    def r(self):
+        return self._radius
+
     @property
     def _sparse_dim(self):
         return self.dimensions[self._sparse_position]
 
+    @property
+    def _mpitype(self):
```
FabioLuporini, line 168: I didn't know about `MPI._typedict`! We should move the logic this and the property below to a separate tools function, somewhere around this one: https://github.com/devitocodes/devito/blob/master/devito/tools/dtypes_lowering.py#L115

then get the MPI type explicitly by a call to the tool function where needed

devito/types/sparse.py
```diff
@@ -117,10 +263,14 @@ def inject(self, *args, **kwargs):
         """
         return self.interpolator.inject(*args, **kwargs)
 
+    @cached_property
+    def _point_increments(self):
+        """Index increments in each dimension for each point symbol."""
```
FabioLuporini, line 268: Dimension (nitpicking)

devito/types/sparse.py
```diff
@@ -931,61 +1035,157 @@ class PrecomputedSparseFunction(AbstractSparseFunction):
...
+        self.interpolator = PrecomputedInterpolator(self)
+
+    @property
+    def r(self):
+        return self._radius
+
+    @cached_property
+    def _point_increments(self):
+        """Index increments in each dimension for each point symbol."""
+        return tuple(product(range(-self.r//2+1, self.r//2+1), repeat=self.grid.dim))
```
FabioLuporini, line 1106: considering that `r == _radius`, could we move this property to one of the superclasses?

devito/types/sparse.py
```diff
@@ -931,61 +1035,157 @@ class PrecomputedSparseFunction(AbstractSparseFunction):
...
+
+    @cached_property
+    def _point_support(self):
```
FabioLuporini, line 1109: same as above?

devito/types/sparse.py
```diff
@@ -931,61 +1035,157 @@ class PrecomputedSparseFunction(AbstractSparseFunction):
...
+        return np.array(self._point_increments)
+
+    @property
+    def _coords_indices(self):
```
FabioLuporini, line 1113: superclass?

devito/types/sparse.py
```diff
@@ -931,61 +1035,157 @@ class PrecomputedSparseFunction(AbstractSparseFunction):
...
-        if distributor.nprocs == 1:
-            return
+    @property
+    def interpolation_coeffs_data(self):
+        return self.interpolation_coeffs.data._local
 
-        raise NotImplementedError
+    @cached_property
+    def _coordinate_symbols(self):
+        """Symbol representing the coordinate values in each dimension."""
```
FabioLuporini, line 1158: Dimension (nitpicking)

devito/types/sparse.py
```diff
@@ -931,61 +1035,157 @@ class PrecomputedSparseFunction(AbstractSparseFunction):
...
+        mathematically equivalent expanded form `coord/spacing -
+        origin/spacing`. This particular form is problematic when a sparse
+        point is in close proximity of the grid origin, since due to a larger
+        machine precision error it may cause a +-1 error in the computation of
+        the position. We mitigate this problem by computing the positions
+        individually (hence the need for a position map).
+        """
+        if self.gridpoints is not None:
+            ddim = self.gridpoints.dimensions[-1]
+            return tuple([self.gridpoints._subs(ddim, di) for di in range(self.grid.dim)])
```
FabioLuporini, line 1185: I prefer explicit `else:`

FabioLuporini, line 1185: I think I prefer `n` for numbers instead of `di` (or `d`) which is typically used for Dimensions instead

devito/types/sparse.py
```diff
@@ -931,61 +1035,157 @@ class PrecomputedSparseFunction(AbstractSparseFunction):
...
+        return tuple([INT(floor(p / i.spacing))
+                      for p, i in zip(self._position_map.keys(),
```
FabioLuporini, line 1187: no need for `keys()`

devito/types/sparse.py
```diff
@@ -1974,11 +2175,11 @@ def _dist_scatter(self, data=None):
 
     # The implementation in AbstractSparseFunction now relies on us
     # having a .coordinates property, which we don't have.
-    def _arg_apply(self, dataobj, alias=None):
+    def _arg_apply(self, *dataobj, alias=None):
```
FabioLuporini, line 2178: I would rather have:

```
def _arg_apply(self, dataobj, *subfuncs, alias=None):
```

or `*args`; so that you don't have to use `[0]` below. Also, `dataobj` indicates the true Function data

tests/test_dle.py
```diff
@@ -942,17 +941,8 @@ def test_parallel_prec_inject(self):
...
-            in iterations[2].pragmas[0].value
-
-        op1 = Operator(eqns, opt=('advanced', {'openmp': True,
-                                               'par-collapse-ncores': 1,
-                                               'par-collapse-work': 1}))
-        iterations = FindNodes(Iteration).visit(op1)
         assert not iterations[0].pragmas
-        assert 'omp for collapse(3) schedule(dynamic,chunk_size)'\
-            in iterations[1].pragmas[0].value
+        assert 'omp for' in iterations[1].pragmas[0].value
```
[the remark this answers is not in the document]
mloubout: The loop structure is now the exact same as for the standard linear interpretation `p, conditional` so no more collapse

devito/operations/interpolators.py
```diff
@@ -108,15 +107,12 @@ def interpolate(self, *args, **kwargs):
         pass
 
 
-class LinearInterpolator(GenericInterpolator):
+class WeightedInterpolation(GenericInterpolator):
```
georgebisbas, line 110: For homogeneity with the other classes I would rename it as WeightedInterpolator.
Unless you want to rename the other classes to xxxInterpolation..?

devito/types/grid.py
```diff
@@ -222,11 +222,16 @@ def origin_map(self):
...
-        """Offset of the local (per-process) origin from the domain origin."""
+    def origin_ioffset(self):
+        """Offset index of the local (per-process) origin from the domain origin."""
         grid_origin = [min(i) for i in self.distributor.glb_numb]
         assert len(grid_origin) == len(self.spacing)
-        return tuple(i*h for i, h in zip(grid_origin, self.spacing))
+        return tuple(grid_origin)
+
+    @property
+    def origin_offset(self):
```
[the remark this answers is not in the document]
mloubout: Not sure how that would work, it's two different things one is the physical origin the other one is the index of the [physical origin in the grid.

devito/operations/interpolators.py
mloubout: No these are not the dimensions these are the dimensions of the underlying grid which are different than the dimension of the Sparse object and associated interpolator

devito/operations/interpolators.py
mloubout: `implicit_dims` is an optional `kwarg` of `.interpolat/inject` in case the user wants to make it conditional so this has to process the user provided implicit dims in addition to the default sparse function dimensions. So cannot be made a property only

devito/types/sparse.py
mloubout: It's not like there is an unsupported case it's just we need to add a way to get the global number of points, on the todo list

devito/types/sparse.py
mloubout: es

devito/types/sparse.py
mloubout: yes

devito/operations/interpolators.py
```diff
@@ -126,90 +122,96 @@ def __init__(self, sfunction):
...
+
+    def implicit_dims(self, implicit_dims):
+        return as_tuple(implicit_dims) + self.sfunction.dimensions
+
+    @property
+    def r(self):
+        return self.sfunction.r
+
+    @property
+    def _interp_points(self):
```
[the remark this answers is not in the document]
mloubout: No, in all generality the points might not be contiguous and could be just a list of points

devito/types/sparse.py
mloubout: done

devito/types/sparse.py
mloubout: done

devito/types/sparse.py

devito/types/sparse.py
mloubout: fixed

devito/types/sparse.py
mloubout: Had it like that in abstract one fixed here

tests/test_dse.py
```diff
@@ -2798,7 +2798,7 @@ def test_fullopt(self):
         bns, _ = assert_blocking(op1, {'x0_blk0'})  # due to loop blocking
 
         assert summary0[('section0', None)].ops == 50
-        assert summary0[('section1', None)].ops == 148
+        assert summary0[('section1', None)].ops == 122
```
georgebisbas, line 2801: why that happens?

devito/operations/interpolators.py
```diff
@@ -108,108 +107,111 @@ def interpolate(self, *args, **kwargs):
...
+    Represent an Interpolation operation on a SparseFunction that is separable
+    in space, meaning the coefficients are defined for each Dimension separately
+    and multiplied at a given point: `w[x, y] = wx[x] * wy[y]`
     """
 
     def __init__(self, sfunction):
         self.sfunction = sfunction
 
-    @property
+    @cached_property
```
FabioLuporini, line 121: imho overkill

devito/operations/interpolators.py
```diff
@@ -108,108 +107,111 @@ def interpolate(self, *args, **kwargs):
...
+        return self.sfunction._point_symbols
+
+    @cached_property
+    def _gdim(self):
+        return self.grid.dimensions
+
+    def implicit_dims(self, implicit_dims):
+        return as_tuple(implicit_dims) + self.sfunction.dimensions
+
+    @cached_property
```
FabioLuporini, line 140: just property imho

devito/operations/interpolators.py
FabioLuporini: indeed we should change that

devito/operations/interpolators.py
```diff
@@ -108,108 +107,111 @@ def interpolate(self, *args, **kwargs):
...
+    def r(self):
+        return self.sfunction.r
+
+    @property
+    def _interp_points(self):
+        return range(-self.r+1, self.r+1)
+
+    @property
+    def _nd_points(self):
```
FabioLuporini, line 149: what does "nd" stand for ?

devito/operations/interpolators.py
```diff
@@ -108,108 +107,111 @@ def interpolate(self, *args, **kwargs):
...
-        return A.inv().T * p
+    def _positions(self, implicit_dims):
+        return [Eq(v, k, implicit_dims=implicit_dims)
+                for k, v in self.sfunction._position_map.items()]
+
+    def subs_coords(self, _expr, *idx_subs):
+        return [_expr.xreplace(v_sub) * b.xreplace(v_sub)
+                for b, v_sub in zip(self._interpolation_coeffs, idx_subs)]
+
+    def subs_coords_eq(self, field, _expr, *idx_subs, implicit_dims=None):
```
FabioLuporini, line 170: **if** they're called in just **one** place, probably no need for a separate method. Regardless, it's a private method, so it should with `_`

devito/operations/interpolators.py
```diff
@@ -108,108 +107,111 @@ def interpolate(self, *args, **kwargs):
...
-        right = dict((b, dim.spacing) for b, dim in zip(indices2, self.grid.dimensions))
-        reference_cell = {**left, **right}
-        # Substitute in interpolation matrix
-        A = A.subs(reference_cell)
-        return A.inv().T * p
+    def _positions(self, implicit_dims):
+        return [Eq(v, k, implicit_dims=implicit_dims)
+                for k, v in self.sfunction._position_map.items()]
+
+    def subs_coords(self, _expr, *idx_subs):
```
FabioLuporini, line 166: `_apply_coords_subs` ? 
`_sub_coords` ?

At first glance, `subs_coords` makes me think a substitution mapper is returned, but that's clearly not the case

devito/operations/interpolators.py
```diff
@@ -108,108 +107,111 @@ def interpolate(self, *args, **kwargs):
...
-        for i, idx in enumerate(index_matrix):
-            # Introduce ConditionalDimension so that we don't go OOB
-            mapper = {}
-            for j, d in zip(idx, self.grid.dimensions):
-                p = points[j]
-                lb = sympy.And(p >= d.symbolic_min - self.sfunction._radius,
-                               evaluate=False)
-                ub = sympy.And(p <= d.symbolic_max + self.sfunction._radius,
-                               evaluate=False)
+        points = {d: [] for d in self._gdim}
```
FabioLuporini, line 181: defaultdict(list)

devito/operations/interpolators.py
```diff
@@ -108,108 +107,111 @@ def interpolate(self, *args, **kwargs):
...
+
+        # Coefficient symbol expression
+        temps.extend(self._coeff_temps(implicit_dims))
+
+        # Create positions and indices temporaries/indirections
+        for ((di, d), pos) in zip(enumerate(self._gdim), pmap):
+            for (ri, r) in enumerate(self._interp_points):
+                p = Symbol(name='ii_%s_%s_%d' % (self.sfunction.name, d.name, ri))
+                points[d].append(p)
+                # Conditionals to avoid OOB
```
FabioLuporini, line 198: nitpicking: blank line before comment

devito/operations/interpolators.py
```diff
@@ -230,7 +232,7 @@ def interpolate(self, expr, offset=0, increment=False, self_subs={},
             interpolation expression, but that should be honored when constructing
             the operator.
         """
-        implicit_dims = as_tuple(implicit_dims) + self.sfunction.dimensions
+        implicit_dims = self.implicit_dims(implicit_dims)
```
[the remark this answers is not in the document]
FabioLuporini: also "implicit_dims" should be a private method.

Maybe `_augment_implicit_dims`?

devito/operator/operator.py
```diff
@@ -621,9 +621,10 @@ def _postprocess_arguments(self, args, **kwargs):
         """Process runtime arguments upon returning from ``.apply()``."""
         for p in self.parameters:
             try:
-                p._arg_apply(args[p.name], args[p.coordinates.name], kwargs.get(p.name))
+                subfuncs = (args[s] for s in p._subfunc_names)
```
FabioLuporini, line 624: Do we actually need the `_subfunc_names` attribute? I think we should drop it

```
subfuncs = [args[s.name] for s in p._sub_functions]
```

devito/tools/dtypes_lowering.py
```diff
@@ -128,6 +128,11 @@ def dtype_to_mpitype(dtype):
     }[dtype]
 
 
+def dtype_to_mpidtype(dtype):
```
FabioLuporini, line 131: What does this function return exactly?

devito/types/dense.py
```diff
@@ -1471,6 +1471,13 @@ def __padding_setup__(self, **kwargs):
     def _halo_exchange(self):
         return
 
+    @property
+    def origin(self):
```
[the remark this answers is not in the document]
FabioLuporini: even if it remained here, it should be moved next to the other public attributes; here it's in between two private methods, hence it's misplaced

devito/types/grid.py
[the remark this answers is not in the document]
FabioLuporini: well the caller may need both occasionally?

`grid.origin_offset.index` vs `grid.origin_offset.phy`

But again, nitpicking (basically my obsession with trying to keep the class interfaces as small as possible, but this is probably just syntactic sugar)

devito/types/sparse.py
FabioLuporini: I don't like commented out code, it's something we've never done.

It's perfectly fine to spell out it's a TODO... then maybe reference an issue on github and move the pseudocode in there?

devito/types/sparse.py
```diff
@@ -76,6 +83,62 @@ def __shape_setup__(cls, **kwargs):
...
+
+            # Infer dtype
+            if np.issubdtype(key.dtype.type, np.integer):
+                dtype = np.int32
+            else:
+                dtype = self.dtype
+
+        if key is not None and key.ndim > 2:
+            shape = (*shape, *key.shape[2:])
+            # Safely assume there is at most 3 (3D) extra dimensions
```
FabioLuporini, line 124: assert ? or raise NotImplementedError if not the case

devito/types/sparse.py
FabioLuporini: you can now drop this property

devito/types/sparse.py
```diff
@@ -89,21 +152,141 @@ def space_order(self):
...
+        sfuncs = [getattr(self, s) for s in self._sub_functions
+                  if getattr(self, s) is not None]
+        return {s: dtype_to_mpidtype(s.dtype) for s in sfuncs}
+
+    @property
+    def comm(self):
+        return self.grid.distributor.comm
+
+    @property
+    def distributor(self):
```
FabioLuporini, line 182: private

devito/types/sparse.py
```diff
@@ -89,21 +152,141 @@ def space_order(self):
...
+        return dtype_to_mpidtype(self.dtype)
+
+    @property
+    def _smpitype(self):
+        sfuncs = [getattr(self, s) for s in self._sub_functions
+                  if getattr(self, s) is not None]
+        return {s: dtype_to_mpidtype(s.dtype) for s in sfuncs}
+
+    @property
+    def comm(self):
```
FabioLuporini, line 178: private

return self._distributor.comm

devito/types/sparse.py
```diff
@@ -212,31 +390,165 @@ def _dist_alltoall(self, dmap=None):
...
+            return {subfunc: subfunc.data}
+
+        # Compute dist map only once
+        dmap = self._dist_datamap
+        mask = self._dist_scatter_mask(dmap=dmap)
+
+        # Pack (reordered) subfunc values so that they can be sent out via an Alltoallv
+        sfuncd = subfunc.data._local[mask[self._sparse_position]]
+
+        # Send out the sparse point coordinates
```
FabioLuporini, line 470: "Send out the SubFunction values"

devito/types/sparse.py
FabioLuporini: prod

devito/types/sparse.py
FabioLuporini: I just find this warning a bit misleading. Is it worht keeping it?

FabioLuporini: sorry, "unclear", more than "misleading"

devito/types/sparse.py
```diff
@@ -1581,13 +1771,13 @@ def inject(self, field, expr, offset=0, u_t=None, p_t=None):
     @classmethod
     def __indices_setup__(cls, **kwargs):
         """
-        Return the default dimension indices for a given data shape.
+        Return the default Dimension indices for a given data shape.
         """
-        dimensions = kwargs.get('dimensions')
-        if dimensions is None:
-            dimensions = (kwargs['grid'].time_dim, Dimension(
+        Dimensions = kwargs.get('dimensions')
```
FabioLuporini, line 1776: too much :D

devito/types/sparse.py
```diff
@@ -1974,7 +2164,7 @@ def _dist_scatter(self, data=None):
 
     # The implementation in AbstractSparseFunction now relies on us
     # having a .coordinates property, which we don't have.
-    def _arg_apply(self, dataobj, alias=None):
+    def _arg_apply(self, dataobj, *subfunc, alias=None):
```
FabioLuporini, line 2167: "*subfuncs" to be pedantinc

devito/types/sparse.py
mloubout: I like to use `(namei, name) in zip(...)` so that's why I have `di` since the "name" is `d` but can change to n

.github/workflows/examples.yml
```diff
@@ -50,7 +50,13 @@ jobs:
         pip install matplotlib pyrevolve
 
     - name: Tests in examples
-      run: py.test --cov --cov-config=.coveragerc --cov-report=xml examples/
+      run: |
+        py.test --cov --cov-config=.coveragerc --cov-report=xml examples/
```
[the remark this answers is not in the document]
mloubout: No it only pick up the files starting with test_

devito/symbolics/inspection.py
```diff
@@ -260,3 +261,26 @@ def has_integer_args(*args):
...
+
+def sympy_dtype(expr, default):
+    """
+    Try to infer the data type of the expression
+    returns the default if non is found
+    """
+    args = expr.args
+    # We can only infer the dtype for addition/multiplication or Symbols
+    # For other case the epxression function may modify the infered dtype
+    if not (isinstance(expr.func, Add) or isinstance(expr.func, Mul)) or \
```
FabioLuporini, line 274: what happens w/o this `if`? it feels a bit weird...

devito/types/dense.py
```diff
@@ -1458,6 +1458,13 @@ def _arg_values(self, **kwargs):
     def parent(self):
         return self._parent
 
+    @property
+    def origin(self):
+        """
+        SubFunction have zero origin
```
FabioLuporini, line 1464: probably just a simple comment, not a docstring

devito/types/dimension.py
```diff
@@ -107,6 +107,8 @@ class Dimension(ArgProvider):
     is_Incr = False
     is_Block = False
 
+    indirect = False
```
FabioLuporini, line 110: this may need some work

tests/test_dle.py
```diff
@@ -187,13 +187,20 @@ def test_cache_blocking_structure_optrelax():
 
     op = Operator(eqns, opt=('advanced', {'blockrelax': True}))
 
-    bns, _ = assert_blocking(op, {'x0_blk0', 'p_src0_blk0'})
+    bns, _ = assert_blocking(op, {'p_src0_blk0', 'x0_blk0', 'p_src1_blk0'})
```
FabioLuporini, line 190: memo for me: I will have to look into this. Why two loop nests over p_src?

devito/types/sparse.py
```diff
@@ -76,9 +74,67 @@ def __shape_setup__(cls, **kwargs):
...
         return shape
 
-    def _halo_exchange(self):
-        # no-op for SparseFunctions
-        return
+    def __fd_setup__(self):
+        """
+        Dynamically add derivative short-cuts.
+        """
+        return generate_fd_shortcuts(self.dimensions, self.space_order)
```
FabioLuporini, line 81: 👀

devito/types/sparse.py
```diff
@@ -89,38 +145,67 @@ def space_order(self):
...
-    def inject(self, *args, **kwargs):
-        """
-        Implement an injection operation from a sparse point onto the grid
-        """
-        return self.interpolator.inject(*args, **kwargs)
+    @property
+    def _coords_indices(self):
+        if self.gridpoints_data is not None:
+            return self.gridpoints_data
+        else:
```
FabioLuporini, line 174: no need for this else

devito/passes/clusters/aliases.py
```diff
@@ -832,7 +833,11 @@ def lower_schedule(schedule, meta, sregistry, ftemps):
     subs = {}
     for pivot, writeto, ispace, aliaseds, indicess, _ in schedule:
         name = sregistry.make_name()
-        dtype = meta.dtype
+        # Infer the dtype for the pivot
+        # This prevents cases such as `floor(a*b)` with `a` and `b` floats
+        # that would creat a temporary `int r = b` leading to erronous numerical results
+        # Such cases happen with the positions for sparse functions for example.
+        dtype = sympy_dtype(pivot, meta.dtype) or meta.dtype
```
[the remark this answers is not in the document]
mloubout: WIth a small tweak yes and fixes it.

tests/test_dle.py
mloubout: One for the temp (r0 = src_coords ...) then on for the interpolation. First `p_src` loop is outside the time loop why is before `x0_block`

devito/tools/dtypes_lowering.py
FabioLuporini: up

devito/types/grid.py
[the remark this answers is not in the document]
FabioLuporini: up

devito/passes/iet/languages/openacc.py
```diff
@@ -164,8 +164,7 @@ def _make_partree(self, candidates, nthreads=None, index=0):
         if self._is_offloadable(root) and \
            all(i.is_Affine for i in [root] + collapsable) and \
            self.par_tile:
-            idx = min(index, len(self.par_tile) - 1)
-            tile = self.par_tile[idx]
+            tile = self.par_tile.next()
```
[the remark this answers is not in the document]
mloubout: See your comment below, basically allow to avoid corner case where the user assumes the number of loop nest wrong so it reuses the last one if there is missing inputs.

devito/passes/iet/mpi.py
```diff
@@ -45,6 +45,13 @@ def _drop_halospots(iet):
             if f in hs.fmapper and all(i.is_reduction for i in v):
                 mapper[hs].add(f)
 
+    # If a HaloSpot is outside any iteration it is not needed
```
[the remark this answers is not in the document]
mloubout: Comment is a bit misleading. It is not droping any it is dropping HaloSpot that have an iter (loc_indices) but go lifted out of it meaning there is no need for it, it still kepps halospot that are out of iteration that don't have loc_indices, i.e initial ones (there is one of the test that would fail if it wasn't)

devito/symbolics/inspection.py
```diff
@@ -260,3 +261,26 @@ def has_integer_args(*args):
         except AttributeError:
             res = res and has_integer_args(a)
     return res
+
+
+def sympy_dtype(expr, default):
```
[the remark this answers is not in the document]
mloubout: default cannot be None it needs to be a proper dtype

devito/tools/dtypes_lowering.py
mloubout: the MPI data type needed by the mpi comm routines such as alltoall

devito/types/dimension.py
mloubout: It's jus a Dimension class property the same as all the `is_...` above

devito/types/grid.py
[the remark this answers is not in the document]
mloubout: Will see, not sure it helps much as this isn't really a user level property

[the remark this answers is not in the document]
FabioLuporini: it helps by reducing the number of instance methods by one, but I agree it's mostly syntactic sugar...

requirements-testing.txt
```diff
@@ -4,4 +4,5 @@ pytest-cov
 codecov
 flake8>=2.1.0
 nbval
-scipy
\ No newline at end of file
+scipy
+pooch; python_version >= "3.8"
```
FabioLuporini, line 8: ?

tests/test_caching.py
```diff
@@ -625,9 +625,10 @@ def test_sparse_function(self, operate_on_empty_cache):
 
         i = u.inject(expr=u, field=u)
 
-        # created: ii_u_0*2 (Symbol and ConditionalDimension), ii_u_1*2, ii_u_2*2,
-        # ii_u_3*2, px, py, posx, posy, u_coords (as indexified),
-        ncreated = 2+2+2+2+2+1+1+1
+        # created: rxu, rxy (radius dimensions) and spacings and conditionals
+        # conditional sparse dim
```
FabioLuporini, line 629: leftover

devito/types/sparse.py
FabioLuporini: up

devito/types/sparse.py
FabioLuporini: up

devito/types/sparse.py
```diff
@@ -140,6 +225,99 @@ def _dist_datamap(self):
         """
         return self.grid.distributor.glb_to_rank(self._support) or {}
 
+    @cached_property
+    def dist_origin(self):
```
FabioLuporini, line 229: group with public attributes

devito/types/sparse.py
```diff
@@ -140,6 +225,99 @@ def _dist_datamap(self):
...
+        origin.
+        """
+        return OrderedDict([((c - o)/d.spacing, p)
+                            for p, c, d, o in zip(self._pos_symbols,
+                                                  self._coordinate_symbols,
+                                                  self.grid.dimensions,
+                                                  self.grid.origin_symbols)])
+
+    @cached_property
+    def _dist_reorder_mask(self):
```
FabioLuporini, line 259: misplaced, group with sibling properties/methods

devito/types/sparse.py
FabioLuporini: up

devito/types/sparse.py
```diff
@@ -849,11 +889,9 @@ def interpolate(self, expr, offset=0, u_t=None, p_t=None, increment=False):
         if p_t is not None:
             subs = {self.time_dim: p_t}
 
-        return super(SparseTimeFunction, self).interpolate(expr, offset=offset,
-                                                           increment=increment,
-                                                           self_subs=subs)
+        return super().interpolate(expr, increment=increment, self_subs=subs)
 
-    def inject(self, field, expr, offset=0, u_t=None, p_t=None, implicit_dims=None):
+    def inject(self, field, expr, u_t=None, p_t=None, implicit_dims=None):
```
FabioLuporini, line 894: excellent

tests/test_caching.py
```diff
@@ -625,9 +625,10 @@ def test_sparse_function(self, operate_on_empty_cache):
 
         i = u.inject(expr=u, field=u)
 
-        # created: ii_u_0*2 (Symbol and ConditionalDimension), ii_u_1*2, ii_u_2*2,
-        # ii_u_3*2, px, py, posx, posy, u_coords (as indexified),
-        ncreated = 2+2+2+2+2+1+1+1
+        # created: rxu, rxy (radius dimensions) and spacings and conditionals
```
georgebisbas, line 628: rux-ruy?

devito/types/sparse.py
mloubout: It's just moved from above, space_order is zero this doesn't really do anything but needs to be implemented at some point

devito/types/sparse.py
mloubout: THough you said we prefer if/else rather than just if and then more code but can change it

devito/types/sparse.py
mloubout: All is ordered

_(cls
property
cached_property

devito/types/sparse.py
mloubout: It is a property in the super class yes

tests/test_dimension.py
```diff
@@ -1416,8 +1416,7 @@ def test_affiness(self):
         iterations = [i for i in FindNodes(Iteration).visit(op) if i.dim is not time]
         assert all(i.is_Affine for i in iterations)
 
-    @switchconfig(condition=isinstance(configuration['compiler'],
-                  (IntelCompiler, OneapiCompiler)), safe_math=True)
+    @switchconfig(safe_math=True)
```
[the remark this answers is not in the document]
mloubout: no

devito/passes/iet/mpi.py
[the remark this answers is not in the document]
mloubout: So
- Without this pass `test_mpi.py::test_injection_wodup_wtime` fails because you end up with an HaloSpot for a write-to only function outside the time loop (that HaloSpot is not there in master but appears here due to the revamp of dimensions/... of sparse inject). So it actually correctly removes it
- If removed but does not check that loc_indices isn't empty then `test_hoist_haloupdate_from_innerloop` fails as it wrongly removes the read halospot with no loc_indices (Function)  for `f`

devito/passes/iet/parpragma.py
```diff
@@ -284,6 +285,10 @@ def _select_candidates(self, candidates):
                 if i.is_Vectorized:
                     break
 
+                # Also, we do not want to collapse small atomic reductions
+                if i.is_ParallelAtomic and i.dim.is_Custom:
```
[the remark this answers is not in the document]
mloubout: It "small" but not amount of work small. it's more that collapsing atomic reduction over Custom dimension is very unlikely to improve performance as it will only increase race conditions

requirements-testing.txt
mloubout: scipy>=1.11 (which only supports python >= 3.8) requires pooch for scipy.datasets but is not part of their requirments so need it (needed for scipy.ascent in the builtin tests)

devito/ir/equations/algorithms.py
```diff
@@ -51,28 +50,33 @@ def handle_indexed(indexed):
...
+    extra = filter_sorted(extra)
 
     # Add in implicit relations for parent dimensions
     # -----------------------------------------------
     # 1) Note that (d.parent, d) is what we want, while (d, d.parent) would be
     # wrong; for example, in `((t, time), (t, x, y), (x, y))`, `x` could now
     # preceed `time`, while `t`, and therefore `time`, *must* appear before `x`,
     # as indicated by the second relation
-    implicit_relations = {(d.parent, d) for d in extra if d.is_Derived}
+    implicit_relations = {(d.parent, d) for d in extra if d.is_Derived and not d.indirect}
```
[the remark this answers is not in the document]
FabioLuporini: I was more of referring to the fact that I remembered (before looking at the rest of your PR) that .indirect was a property of ConditionalDimension only. Resolving!

devito/passes/iet/mpi.py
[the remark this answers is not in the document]
FabioLuporini: > write-to only function

What function is that? 
Why do we have a write-only function? is it just a feature of the test?

If the symptom is the presence of a HaloSpot for a write-only function, then the actual issue, I think, is that that HaloSpot should have never been generated in the first place. So the actual problem might be earlier on in the compilation pipeline, in particular here:

https://github.com/devitocodes/devito/blob/master/devito/ir/clusters/algorithms.py#L374

If this were correct, then all we have to do at that point is avoiding to add stuff to `points` if `a.function` doesn't appear in `scope.reads`.

Before I develop any further this concept (if necessary), let's see if we agree on the first part, namely:

> that HaloSpot should have never been generated in the first place

devito/passes/iet/parpragma.py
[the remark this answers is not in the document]
FabioLuporini: CustomDimensions can be anything, they might even be equivalent in all to SpaceDimensions

What kind of CustomDimensions do you have in mind here? What loops are we trying to avoid to collapse?

devito/passes/iet/parpragma.py
```diff
@@ -424,6 +429,11 @@ def _make_nested_partree(self, partree):
         if self.nhyperthreads <= self.nested:
             return partree
 
+        # Loop nest with atomic reductions are more likely to have less latency
+        # keep outer loop parallel
```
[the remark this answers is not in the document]
FabioLuporini: ah sorry, read quickly and the lack of capital letter at the beginning tricked me :)

devito/tools/dtypes_lowering.py
FabioLuporini: memo

devito/types/dimension.py
FabioLuporini: prod

requirements-testing.txt
FabioLuporini: one day we shall rewrite those tests not to depend on scipy

tests/test_dimension.py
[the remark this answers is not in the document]
FabioLuporini: why it didn't before? what changed? different loop nest?

tests/test_dle.py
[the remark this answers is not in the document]
FabioLuporini: sweet

tests/test_dse.py
```diff
@@ -2657,11 +2672,14 @@ def test_fullopt(self):
...
-        assert summary0[('section1', None)].ops == 148
+        assert summary0[('section1', None)].ops == 44
         assert np.isclose(summary0[('section0', None)].oi, 2.851, atol=0.001)
 
-        assert summary1[('section0', None)].ops == 31
-        assert np.isclose(summary1[('section0', None)].oi, 1.767, atol=0.001)
+        assert summary1[('section0', None)].ops == 9
+        assert summary1[('section1', None)].ops == 9
+        assert summary1[('section2', None)].ops == 31
+        assert summary1[('section3', None)].ops == 26
```
FabioLuporini, line 2681: why the new sections?

FabioLuporini: I should maybe just run the code. I will, later on

tests/test_pickle.py
```diff
@@ -393,25 +430,6 @@ def test_receiver(self, pickle):
...
         assert np.all(new_rec.coordinates.data == [[0.], [1.], [2.]])
 
-    def test_alias_sparse_function(self, pickle):
-        grid = Grid(shape=(3,))
-        sf = SparseFunction(name='sf', grid=grid, npoint=3, space_order=2,
-                            coordinates=[(0.,), (1.,), (2.,)])
-        sf.data[0] = 1.
-
-        # Create alias
-        f0 = sf._rebuild(name='f0', alias=True)
```
FabioLuporini, line 403: Are we dropping the test for this alias=True case?

the new test that you've added above doesn't appear to be testing this case

devito/types/sparse.py
FabioLuporini: No no I meant something completely different... but not sure now what my brain was thinking. Ignore me, resolving

devito/types/sparse.py
FabioLuporini: this one appearts to be in the middle between two private properties, `_dist_datamap` and `_pos_symbols`

devito/types/sparse.py
FabioLuporini: likewise, there's a `_dist_scatter_mask` approx 60 lines below, and in the middle between the two, public methods

tests/test_dse.py
mloubout: There is no one section per sparse coord computing temps for the position

devito/passes/iet/mpi.py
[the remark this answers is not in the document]
mloubout: So had a quick look, it seems to be a bit more involved bedause some how it has the wrong properties by the time it gets there. i.e you have
```
r0 = ....
f[ ...] += r0
```

but it think it's a read only (is_write False and is_read True) for `f` I'll see if can find the source of the issue

devito/passes/iet/parpragma.py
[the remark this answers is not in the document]
mloubout: Just trying to prevent 
```
for (p_src ...)
 px = 
 py = 
 # omp for collapse
 for (rx = ...)
    for (ry = ...)
```

I could maybe make a special dims for this

tests/test_dimension.py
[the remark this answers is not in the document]
mloubout: Different loop next a interp weight computation

tests/test_dle.py
mloubout:
```
  #pragma omp parallel num_threads(nthreads_nonaffine)
  {
    int chunk_size = (int)(fmax(1, (1.0F/3.0F)*(p_sf_M - p_sf_m + 1)/nthreads_nonaffine));
    #pragma omp for schedule(dynamic,chunk_size)
    for (int p_sf = p_sf_m; p_sf <= p_sf_M; p_sf += 1)
    {
      r1[p_sf] = floor(r3*(-o_x + sf_coords[p_sf][0]));
      r2[p_sf] = floor(r4*(-o_y + sf_coords[p_sf][1]));
    }
  }

```

I.e you have those now

tests/test_dse.py
mloubout: Same as above
```
  #pragma omp parallel num_threads(nthreads_nonaffine)
  {
    int chunk_size = (int)(fmax(1, (1.0F/3.0F)*(p_sf_M - p_sf_m + 1)/nthreads_nonaffine));
    #pragma omp for schedule(dynamic,chunk_size)
    for (int p_sf = p_sf_m; p_sf <= p_sf_M; p_sf += 1)
    {
      r1[p_sf] = floor(r3*(-o_x + sf_coords[p_sf][0]));
      r2[p_sf] = floor(r4*(-o_y + sf_coords[p_sf][1]));
    }
  }
```

devito/passes/iet/parpragma.py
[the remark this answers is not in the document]
FabioLuporini: > It "small" but not amount of work small 

These loops should be pretty small, and the way "work" is defined later is essentially "number of iterations"

So I'm not sure I follow? These CustomDimensions have numeric size (right? or is this what I'm getting wrong?) so it should be feasible?

tests/test_caching.py
FabioLuporini: up

devito/types/sparse.py
FabioLuporini: up

devito/types/sparse.py
FabioLuporini: What I meant to say is that `_dist_origin` should be a (cached?) property, not attached to self at `__init__` time

mloubout: It depends on the input in Precomputed and has to be set as __init__

devito/passes/iet/parpragma.py
```diff
@@ -294,7 +295,8 @@ def _select_candidates(self, candidates):
                     except TypeError:
                         pass
 
-                collapsable.append(i)
+                if not i.is_ParallelAtomic or nested:
```
FabioLuporini, line 298: Can you add a short explanatory comment?

devito/symbolics/inspection.py
```diff
@@ -260,3 +263,23 @@ def has_integer_args(*args):
...
+    """
+    Try to infer the data type of the expression
```
FabioLuporini, line 270: I'd rewrite as

```
"""
Infer the dtype of the expression, or None it could not be determined.
"""
```

devito/symbolics/printer.py
```diff
@@ -102,6 +102,10 @@ def _print_Mod(self, expr):
         args = ['(%s)' % self._print(a) for a in expr.args]
         return '%'.join(args)
 
+    def _print_Mul(self, expr):
+        term = super()._print_Mul(expr)
+        return term.replace("(-1)*", "-")
```
FabioLuporini, line 107: I don't liks this -- the symptom points to a sympy-reconstruction bug somewhere

For example, I have a factorizer patch in my branch that fixes an identical symptom:

https://github.com/devitocodes/devito/commit/9a4a31a65143d29b572218a385a63b2c4c871bee#diff-a3ab417a5b298b20b70b45a84b15b8324f0526f6fe12a588a3f2912ed057d377R145

Does it help it here?

We surely shouldn't hack away the underlying sympy representation here, but rather fix the actual bug -- hopefully my patch+test does it!

devito/types/dimension.py
```diff
@@ -181,6 +181,14 @@ def min_name(self):
     def max_name(self):
         return "%s_M" % self.name
 
+    @property
+    def indirect(self):
+        return False
+
+    @property
+    def index(self):
+        return self if self.indirect is True else getattr(self, 'parent', self)
```
FabioLuporini, line 190: A class shouldn't have any knowledge of its subclasses

Here you're trying to access the parent, which is subclass-only stuff, otherwise you wouldn't use `getattr` (which should always be avoided unless in very specific cases)

the way to go is:

* in the superclass: systematically `return self `
* in DerivedDimension: use the if-else, then no need for getattr

devito/passes/iet/mpi.py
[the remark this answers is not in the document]
FabioLuporini: copy-pasting from slack for future reference:

I tried dropping [this entire piece of code ](https://github.com/devitocodes/devito/pull/2128/files#diff-e2c8e1150ab045d6300aee639c2af8a7deba5ef56290603d94a6bba2895cc83bR49-R56)and I’ve just got a full pass on test_mpi.py
this seems to contradict what you were saying [here](https://github.com/devitocodes/devito/pull/2128/files#r1273671513) —  in particular, the fact that:
If removed but does not check that loc_indices isn't empty then test_hoist_haloupdate_from_innerloop fails as it wrongly removes the read halospot with no loc_indices (Function) for f

devito/passes/iet/parpragma.py
FabioLuporini: Actually we may want a careful explanation why we're peeking at is_ParallelAtomic here

devito/passes/iet/parpragma.py
```diff
@@ -566,7 +571,7 @@ def __init__(self, sregistry, options, platform, compiler):
         super().__init__(sregistry, options, platform, compiler)
 
         self.gpu_fit = options['gpu-fit']
-        self.par_tile = options['par-tile']
+        self.par_tile = UnboundTuple(options['par-tile'])
```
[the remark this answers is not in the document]
FabioLuporini: Unsure , now, actually. Why isn't UnboundedMultiTuple(...) actually just usable here?

devito/symbolics/printer.py
mloubout: There is a small issue with some sympy version that pickling does not lead to the same expression between `-expr` and `(-1) * expr` not sure exactly why.

devito/types/dimension.py
mloubout: Except we have dimensions such as `CustomDimension` that are not `DerivedDimension` but can be indirect and have a parent in some cases.

devito/passes/iet/parpragma.py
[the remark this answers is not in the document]
mloubout: `UnboundedMultiTuple` does not itearte infinitely over each tuple and throws an error after reaching the last element of a subtuple instead of returning it forever, I assume this is intended behaviour so did not want to modify it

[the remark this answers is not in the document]
FabioLuporini: you could easily add

```
self.par_tile.next(default='last')
```

or even

```
umt = UnboundMultiTuple(..., default='last')
```

to change default behavior

I **think** what we want here is an UnboundMultiTuple to actually support the user providing `par-tile: ((16, 16, 16), (4, 4, 4), (16, 4, 7), ...)`

devito/symbolics/printer.py
FabioLuporini: What's the MFE? I can look into it

devito/types/sparse.py
FabioLuporini: What's the failing test if I turn it into a property?

the `_dist_origin` dictionary maps two objects that belong to `self`. Since `self` and all its attributes are immutable (clearly aside from the .data), I struggle to see why this should be a problem

devito/symbolics/printer.py
mloubout: `test_pickle.py` at `TestOperator::test_usave_sampled`

devito/types/sparse.py
mloubout: THis one is a cached property and groupped with it, the dist_scatter/gather are methods don't think can really move this one in a better place

requirements-testing.txt
mloubout: yeah, it's litterally only used to load an image and that's it so we should try to ditch it

tests/test_pickle.py
mloubout: Slightly moved above

mloubout:
As discussed on slack, issue opened for SubFunction and merging, SUbFunction cleanup top of the list next.

devito/ir/equations/algorithms.py
```diff
@@ -33,14 +33,19 @@ def handle_indexed(indexed):
...
-                relation.extend([d for d in filter_sorted(i.free_symbols)
-                                 if isinstance(d, Dimension)])
+                relation.extend(filter_sorted(i.atoms(Dimension)))
 
         # StencilDimensions are lowered subsequently through special compiler
         # passes, so they can be ignored here
         relation = tuple(d for d in relation if not d.is_Stencil)
 
-        return relation
+        # Make sure indices appearance order satisfies "parent" order. Foe example
```
georgebisbas, line 42: For example

devito/types/sparse.py
```diff
@@ -82,9 +80,128 @@ def __shape_setup__(cls, **kwargs):
...
         return shape
 
-    def _halo_exchange(self):
-        # no-op for SparseFunctions
-        return
+    def func(self, *args, **kwargs):
+        # Rebuild subfunctions first to avoid new data creation as we have to use `_data`
+        # as a reconstruction kwargs to avoid the circular dependency
+        # with the parent in SubFunction
+        # This is also necessary to avoid shaoe issue in the SubFunction with mpi
```
georgebisbas, line 87: shaoe?

devito/passes/iet/parpragma.py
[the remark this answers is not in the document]
FabioLuporini: I still think that two classes for unbounded tuples is overkill, but I also appreciate that at this stage you want to move on...

devito/symbolics/printer.py
FabioLuporini: I think you can probably cherry-pick this commit 

https://github.com/devitocodes/devito/pull/2183/commits/9a4a31a65143d29b572218a385a63b2c4c871bee

to get rid of this hack

tests/test_gradient.py
```diff
@@ -15,6 +15,7 @@
 class TestGradient(object):
 
     @skipif(['chkpnt', 'cpu64-icc'])
+    @switchconfig(safe_math=True)
```
FabioLuporini, line 18: is it correct to say that with the new interpolation loop nest structure we're somehow losing some accuracy?

how significant is the failure without safe_math?

devito/types/sparse.py
FabioLuporini: ah, OK.

But if the Precomputed*'s constructor overrides this, then I think you might rather want to use a `__setup_origin__` method. 

Anyway, nitpicking for another day

devito/types/sparse.py
```diff
@@ -402,7 +724,7 @@ class SparseFunction(AbstractSparseFunction):
         Discretisation order for space derivatives. Defaults to 0.
     shape : tuple of ints, optional
         Shape of the object. Defaults to ``(npoint,)``.
-    dimensions : tuple of Dimension, optional
+    Dimensions : tuple of Dimension, optional
```
georgebisbas, line 727: dimensions

devito/types/sparse.py
```diff
@@ -924,32 +969,35 @@ class PrecomputedSparseFunction(AbstractSparseFunction):
...
+        So for `r=6`, we will store 18 coefficients per sparse point (instead of
+        potentially 216).  Must be a three-dimensional array of shape
+        `(npoint, grid.ndim, r)`.
     space_order : int, optional
         Discretisation order for space derivatives. Defaults to 0.
     shape : tuple of ints, optional
-        Shape of the object. Defaults to ``(npoint,)``.
-    dimensions : tuple of Dimension, optional
+        Shape of the object. Defaults to `(npoint,)`.
+    Dimensions : tuple of Dimension, optional
```
georgebisbas, line 995: dimensions?

tests/test_gradient.py
mloubout: It's quite small, about 1e-5.

PR #2129, opened 2023-05-17
compiler: Avoid generating collapse(1)

FabioLuporini:
Reason: NVC (at least version <= 23.1 , possible future as well) has a **compilation** bug (as in, the NVC compiler throws an internal fatal error while attempting to compile perfectly legal C+OpenMP code) that sometimes crops up when the `collapse(1)` clause is used. 

Since (from the OpenMP specification)

> The collapse clause associates one or more loops with the directive on which it appears for the purpose of identifying the portion of the depth of the canonical loop nest to which to apply the semantics of the directive. The argument n speciﬁes the number of loops of the associated loop nest to which to apply those semantics. On all directives on which the collapse clause may appear, the eﬀect is as if a value of one was speciﬁed for n if the collapse clause is not speciﬁed.

We here work around the annoying NVC issue by simply not generating `collapse(1)`.

Bug hunting credits to @deckerla


georgebisbas (commented):
I do not see any problem with the changes, though there are failing tests

mloubout (approved):

PR #2133, opened 2023-05-17
examples: Add forward and elastic operators using compact formulation

nogueirapeterson:
Sirs,

@FabioLuporini @mloubout 

This PR is an implementation of forward and adjoint elastic operators in a compact formulation. In this PR, we did an implementation based on the explicit sffiness matrix.

This PR aims to replace #1944 .

We hope this contribution is welcome.
Thanks!


examples/seismic/stiffness/utils.py
```diff
@@ -0,0 +1,78 @@
+from devito.types.tensor import tens_func
+import numpy as np
+
+
+def D(self, shift=None):
+    """
+    Returns the result of matrix D applied over the TensorFunction.
```
EdCaunt, line 7: Would it be worth showing what matrix D contains or pointing to a paper using this notation in the docstring?

examples/seismic/utils.py
```diff
@@ -4,11 +4,26 @@
...
 from devito.tools import Pickable
 
+from sympy import symbols, Matrix
+
 from .source import *
 
 __all__ = ['AcquisitionGeometry', 'setup_geometry', 'seismic_args']
 
 
+def matriz_init(model):
```
EdCaunt, line 14: Why "matriz"?

examples/seismic/stiffness/utils.py
EdCaunt: Maybe just repeat the Feng and Schuster (2017): Elastic least-squares reverse time migration https://doi.org/10.1190/geo2016-0254.1 here?

EdCaunt (commented):
Looks good to me. Couple of minor comments, but nothing code-related

PR #2134, opened 2023-05-22
mpi: Enhance flexibility for custom topologies

georgebisbas:
This PR enhances the available custom topology MPI decompositions

The requirements for the number of stars to evenly divide the number of MPI procs are dropped.
You can see some topology to decomposition combos in the added tests


devito/mpi/distributed.py
```diff
@@ -600,17 +616,57 @@ class CustomTopology(tuple):
...
-            if input_comm.size % nstars != 0:
-                raise ValueError("Invalid `topology` for given nprocs")
-            if any(i not in ('*', 1) for i in items):
-                raise ValueError("Custom topology must be only 1 or *")
-
-            v = input_comm.size // nstars
-            processed = [i if i == 1 else v for i in items]
-        else:
+        alloc_procs = np.prod([i for i in items if i != '*'])
+        remprocs = int(input_comm.size // alloc_procs)
```
mloubout, line 622: should be `-` so not sure how the rest can work

georgebisbas: Not sure I understand your comment. Do you mean `-` instead of `//` ?

mloubout: Sorry looked at it too quickly ignore this

devito/mpi/distributed.py
```diff
@@ -600,17 +616,53 @@ class CustomTopology(tuple):
...
+
+            # Get star and integer indices
+            int_pos = [i for i, item in enumerate(items) if isinstance(item, int)]
+            int_vals = [item for item in items if isinstance(item, int)]
+            star_pos = [i for i, item in enumerate(items) if not isinstance(item, int)]
+
+            # Decompose the processes remaining for allocation to prime factors
+            alloc_procs = np.prod([i for i in items if i != '*'])
+            remprocs = int(input_comm.size // alloc_procs)
+            prime_factors = primefactors(remprocs)
```
mloubout, line 637: This looks a bit overly intricate. 

If you use `factorint(remproc)` you get directly the dict of factors that you just need to split i.e 

```
factors = factorint(remprocs)
vals = [k for (k, v) in factors.items() for _ in range(v)]
vals = vals + [1 for _ in range(nstars - len(vals))]
startvals = (*vals[:nstart-1], prod(vals[nstars-1:])
```

and should be it

georgebisbas: Well, not realy, since you do not prioritise neither the outermost dimension, nor the overall balance:
e.g. your approach will give:
```
FAILED tests/test_mpi.py::TestDistributor::test_custom_topology_3d_dummy[24-topology20-dist_topology20] - assert (2, 2, 6) == (6, 2, 2)
FAILED tests/test_mpi.py::TestDistributor::test_custom_topology_3d_dummy[32-topology21-dist_topology21] - assert (2, 2, 8) == (4, 4, 2)
```

tests/test_mpi.py
```diff
@@ -180,6 +187,57 @@ def test_custom_topology(self):
...
+        (128, ('*', '*', 1), (16, 8, 1)),
+        (231, ('*', '*', '*'), (11, 7, 3)),
+        (256, (1, '*', '*'), (1, 16, 16)),
+        (256, ('*', 1, '*'), (16, 1, 16)),
+        (256, ('*', '*', 1), (16, 16, 1)),
+        (256, ('*', '*', '*'), (8, 8, 4)),
+        (256, ('*', '*', 2), (16, 8, 2)),
+        (256, ('*', 32, 2), (4, 32, 2)),
+    ])
+    def test_custom_topology_3d_dummy(self, comm_size, topology, dist_topology):
```
FabioLuporini, line 236: isn't this... a bit too much?

georgebisbas: Well, maybe a few can be dropped, but I wanted to extensively test this for some big number of ranks for the strong scaling experiments.
What is maybe redundant is the prime combs like 11, 231 etc...
Overall the above tests are very cheap (no MPI_Init, as we discussed), so I would not worry.

FabioLuporini: thoughts @mloubout ?

devito/mpi/distributed.py
```diff
@@ -600,17 +616,53 @@ class CustomTopology(tuple):
...
+            # then cyclically-iteratively decompose as evenly as possible until
+            # decomposing to the number of `remprocs`
+            while remprocs != 1:
+                star_i = star_i + 1
+                star_i = star_i % nstars
+                prime_factors = primefactors(remprocs)
+                dd_list[star_i] = dd_list[star_i]*max(prime_factors)
+                remprocs = remprocs // max(prime_factors)
+
+            if int_pos:
```
FabioLuporini, line 652: why do you need this if?

devito/mpi/distributed.py
```diff
@@ -600,17 +616,53 @@ class CustomTopology(tuple):
...
+                for index, value in zip(int_pos, int_vals):
+                    processed[index] = value
+
+            if dd_list:
```
FabioLuporini, line 656: why do you need this if?

FabioLuporini: and it's probably doable with a comprehension

devito/mpi/distributed.py
```diff
@@ -600,17 +616,53 @@ class CustomTopology(tuple):
...
+            # Decompose the processes remaining for allocation to prime factors
+            alloc_procs = np.prod([i for i in items if i != '*'])
+            remprocs = int(input_comm.size // alloc_procs)
+            prime_factors = primefactors(remprocs)
+
+            star_i = -1
+            dd_list = [1] * nstars
+
+            # Start by using the max prime factor at the first starred position,
+            # then cyclically-iteratively decompose as evenly as possible until
```
FabioLuporini, line 643: cyclically-iteratively ?

devito/mpi/distributed.py
```diff
@@ -583,15 +584,30 @@ class CustomTopology(tuple):
...
-    N must evenly divide the number of `'*'`, otherwise a ValueError exception
-    is raised.
-    If the wildcard `'*'` is used, then the CustomTopology can only contain either
-    `'*'` or 1's, otherwise a ValueError exception is raised.
+    * `('*', '*', 1)`: the wildcard `'*'` tells the runtime to decompose both
+                       the x and y Dimensions in `nstars` factors of N, prioritizing
+                       the outermost dimension
+
+    Assuming N=6 and requested topology is `('*', '*', 1)`,
+    since there is no integer k, so that k*k=6, we resort to the closest factors to
```
FabioLuporini, line 592: Could you rewrite this (grammarly maybe)? it's not very clear to me at first glance

tests/test_mpi.py
```diff
@@ -12,11 +12,18 @@
                            retrieve_iteration_tree)
 from devito.mpi import MPI
 from devito.mpi.routines import HaloUpdateCall, MPICall
+from devito.mpi.distributed import CustomTopology
 from examples.seismic.acoustic import acoustic_setup
 
 pytestmark = skipif(['nompi'], whole_module=True)
 
 
+class DummyInputComm():
```
FabioLuporini, line 21: In theory you don't even need an extra class, you can just pass `Bunch(size=...)` to CustomTopology

devito/mpi/distributed.py
```diff
@@ -600,17 +616,53 @@ class CustomTopology(tuple):
...
+
+        # If no stars exist we are ready
+        if nstars == 0:
             processed = items
+        else:
+            # Init decomposition list
+            processed = [1] * len(items)
+
+            # Get star and integer indices
+            int_pos = [i for i, item in enumerate(items) if isinstance(item, int)]
```
FabioLuporini, line 630: instead of int_pos and int_vals, and the loop below, why don't you just do it explicitely? it would be much neater

```
for i, item in enumerate(items):
    if isinstance(item, int):
         processed ...
```

etc etc

clearly move this after the `dd_list` (better name pls?) computation

devito/mpi/distributed.py
```diff
@@ -600,17 +616,53 @@ class CustomTopology(tuple):
...
+            remprocs = int(input_comm.size // alloc_procs)
+            prime_factors = primefactors(remprocs)
+
+            star_i = -1
+            dd_list = [1] * nstars
+
+            # Start by using the max prime factor at the first starred position,
+            # then cyclically-iteratively decompose as evenly as possible until
+            # decomposing to the number of `remprocs`
+            while remprocs != 1:
```
FabioLuporini, line 645: nitpicking: if one passes a dummy comm with .size = 0, this remains entrapped in an infinite loop. Safer to use : `while remprocs > 1`

Also, I see an inconsisten use of `_` in variable names

devito/mpi/distributed.py
```diff
@@ -600,17 +616,53 @@ class CustomTopology(tuple):
...
+                star_i = star_i + 1
+                star_i = star_i % nstars
```
FabioLuporini, line 647: AKA

```
star_i = (star_i + 1) % nstars
```

...?

devito/mpi/distributed.py
FabioLuporini, line 637: instead of having this both here and inside the while loop, you should have it only once, at the top of the while loop...

which, if necessary, you can turn into:

```
while True:
    if rem_procs > 1:
        break
    ...
```

thus emulating a do-while loop

devito/mpi/distributed.py
```diff
@@ -600,17 +616,53 @@ class CustomTopology(tuple):
...
+
+            if dd_list:
+                for index, value in zip(star_pos, dd_list):
+                    processed[index] = value
+
+        # Final check that topology matches the communicator size
+        try:
+            assert np.prod(processed) == input_comm.size
+        except:
+            raise ValueError("Invalid `topology`", processed, " for given nprocs:",
```
FabioLuporini, line 664: if it's an assert, why bothering with this? Just leave the assert alone and that's it

FabioLuporini (changes requested):
Several improvements possible in my opinion

devito/mpi/distributed.py
mloubout: That's just a minor change  to the last line to cycle instead of multiply the remainder
```
split =  np.array_split(np.pad(vals, nstar)
starvals = np.prod([np.pad(s, (0, nstar-len(s)), constant_values=1) for s in split], dims=1)
```

mloubout: And pad `(nstar-len(s), 0)` if you want it left heavy instead

devito/mpi/distributed.py
georgebisbas: iteratively

devito/mpi/distributed.py
georgebisbas: Sure yes

devito/mpi/distributed.py
georgebisbas: I think it will be more helpful as a message to the user?
I can drop if you feel so

tests/test_mpi.py
georgebisbas: Good idea, I did not know this Class

devito/mpi/distributed.py
georgebisbas: Really? Like how without iterating O(len(processed))?

devito/mpi/distributed.py
georgebisbas: Switched to `rem_procs`.

devito/mpi/distributed.py
georgebisbas: star_vals is better, idea from @mloubout

devito/mpi/distributed.py
georgebisbas: I think I pushed a simplified version, lemme know if you like it better.
I would dare to say that I consider your approach less readable, at least for my eyes.
btw, I could not make it work?
Would you mind to try locally your code?

georgebisbas: Dropped and improved, lemme know if it is fine

devito/mpi/distributed.py
georgebisbas: I think passing 0 is bad use, maybe we can jump to the assert statement  directly ?
Asking cutom topo on 0 ranks looks bad to me

devito/mpi/distributed.py
georgebisbas: You are right, probably remainder from bad code earlier that it was needed

tests/test_mpi.py
mloubout: Looks a bit overkill yeah but it's very cheap so I guess it's fine

georgebisbas: I dropped a few as well

mloubout (approved):

mloubout (approved):

FabioLuporini (approved):

FabioLuporini:
Merged, thanks

PR #2138, opened 2023-05-31
compiler: Misc compiler fixes and improvements -- part II

FabioLuporini:


georgebisbas (commented):
What was the problem with u.dx2? Is there an issue somewhere about this?

FabioLuporini:
> What was the problem with u.dx2? Is there an issue somewhere about this?

These were PRO issues, and all new tests therefore a in PRO 

That said, I can assure you that there's no new feature here in OSS territory that hasn't been equipped with at least one test. Most changes here are compilation-related generalisations and improvements which are stress-tested by what de-facto are higher-order regression tests in PRO

devito/passes/iet/parpragma.py
```diff
@@ -527,7 +527,9 @@ def function(self):
     def imask(self):
         return self._imask
 
-    @cached_property
+    # TODO: cached_property here will break our pickling tests for reasons that
```
mloubout, line 530: out of curiosity which test?

devito/tools/abc.py
```diff
@@ -130,7 +130,11 @@ def __init__(self, a, b, c=4):
             * `a._rebuild(c=5) -> x(3, 5, 5)`
             * `a._rebuild(1, c=7) -> x(1, 5, 7)`
         """
-        args += tuple(getattr(self, i) for i in self.__rargs__[len(args):])
+        for i in self.__rargs__[len(args):]:
+            if i.startswith('*'):
```
mloubout, line 134: how do you end up here?

devito/types/array.py
```diff
@@ -203,6 +211,13 @@ def _mem_shared(self):
     def _mem_constant(self):
         return self._scope == 'constant'
 
+    @property
```
mloubout, line 214: Move to `AbstractFunction`  I'd say

devito/types/misc.py
```diff
@@ -68,20 +68,21 @@ class FIndexed(Indexed, Pickable):
     `uX[x*ny + y]`, where `X` is a string provided by the caller.
     """
 
-    __rargs__ = ('indexed', 'pname')
+    __rargs__ = ('base', '*indices')
```
mloubout, line 71: Not sure why `*`

tests/test_gpu_common.py
```diff
@@ -1,3 +1,5 @@
+import cloudpickle as pickle
```
mloubout, line 1: We might want to do both `picklle` and `cloudpickle` as parameters for all pickling tests

tests/test_gpu_common.py
```diff
@@ -1343,6 +1350,25 @@ def test_npthreads(self):
             assert op.arguments(time_M=2, npthreads0=5)
 
 
+class TestMisc(object):
+
+    def test_pickling(self):
```
mloubout, line 1355: That should be in `test_pickle` but I guess it's easier for gpu to dso it here

tests/test_pickle.py
```diff
@@ -1,7 +1,7 @@
 import pytest
 import numpy as np
 from sympy import Symbol
-import pickle
+import cloudpickle as pickle
```
mloubout, line 4: As above, import both and make all tests for each

devito/passes/iet/parpragma.py
FabioLuporini: PRO tests landing with the `fix-udx2-MPI` branch in PRO. For future reference: they are `test_layered_funcs.py::TestMisc::testpickling` and `test_composite_funcs.py::TestMisc::testpickling`

devito/tools/abc.py
FabioLuporini: the need to use something like `*indices` inside `__rargs__ = (..., '*indices')` for reconstruction of objects with variadic args (e.g., `FIndexed`)

devito/types/misc.py
FabioLuporini: see comment above

tests/test_gpu_common.py
FabioLuporini: AFAICT, cloudpickle >>> pickle though

Anyway, how would you implement it? Thinking about it, and it doesn't sound obvious... does it?

tests/test_gpu_common.py
FabioLuporini: mmhh in my view, the GPU tests are "more specific" than the pickling tests

Just like some MPI tests are found outside of `test_mpi.py`

tests/test_pickle.py
FabioLuporini: ah maybe i see what you mean now... are you saying something like

```
import pickle as pickle0
import cloudpickle as pickle1

@pytest.parametrize('pickle', [pickle0, pickle1])
def test_...(...):
    ...
```

?

isn't it overkill?

devito/ir/stree/algorithms.py
```diff
@@ -176,15 +177,35 @@ def reuse_partial_subtree(c0, c1, d=None):
...
-    if d in cluster.syncs:
-        tip = NodeSync(cluster.syncs[d], tip)
+    try:
+        syncs = cluster.syncs[d]
+        if all(isinstance(s, WaitLock) for s in syncs):
+            # Unlike all other SyncOps, a WaitLock "floats" in the stree, in that
+            # it doesn't need to wrap any subtree. Thus, a WaitLock acts like
+            # a barrier to what follows inside `d`
+            NodeSync(syncs, tip)
+            return tip
```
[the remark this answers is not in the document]
FabioLuporini: improved and fixed 👍

devito/types/array.py
FabioLuporini: I will see how I can improve things. Note that DiscreteFunction has a `_distributor` attribute too

tests/test_pickle.py
mloubout: Yes what I mean. Since there seems to be breaking difference between those two pickling implementation I think it would be safer to have it yeah. THese are very cheap tests so it shouldn't be an issue

devito/types/array.py
mloubout: I think so yes

FabioLuporini: done

tests/test_pickle.py
FabioLuporini: done. Please take a look, I've restructured a few things

FabioLuporini:
@mloubout update: addressed your comments and rebased on latest master

georgebisbas (approved):
Not sure I have something to comment. Should I look into something specific that you believe needs more attention?

mloubout (approved):

PR #2139, opened 2023-06-05
compiler: Pass operator arguments to downstream operators

ccuetom:
When changing platform, language, or compiler using keyword arguments to the operator, these are not recorded to be used by downstream operators, which instead default to the global configuration.

This makes the information available by storing it in the `ArgumentsMap`.


mloubout (approved):

FabioLuporini (approved):

FabioLuporini:
Merged, thanks

PR #2141, opened 2023-06-08
compiler: Revert "compiler: Relax WaitLock regions in a ScheduleTree"

FabioLuporini:
This reverts commit 4a184e0fcfc04553897e4bc39a8299d4b56c0f49.

Explanation: this was done only to change the profiling scope of the WaitLock regions . However, it turns out it breaks code generation in CUDA/HIP + streaming. Hence, reverting for now, until I will have the time to find a better solution

### DIFF
--- a/devito/mpi/routines.py
+++ b/devito/mpi/routines.py
@@ -1168,7 +1168,7 @@ class MPIMsg(CompositeObject):
         else:
             assert self.target.c0.is_Array
             assert args is not None
-            return int(v.subs(args))
+            return int(subs_op_args(v, args))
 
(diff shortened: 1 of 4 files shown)


devito/mpi/routines.py
```diff
@@ -1168,7 +1168,7 @@ def _as_number(self, v, args):
         else:
             assert self.target.c0.is_Array
             assert args is not None
-            return int(v.subs(args))
+            return int(subs_op_args(v, args))
```
FabioLuporini, line 1171: sympy 1.12 patch

PR #2142, opened 2023-06-08
compiler: Add opkwargs property to ArgumentsMap

ccuetom:
PR adds `opkwargs` property to `ArgumentsMap`. `opkwargs` returns a dictionary containing compiler, language and platform keyword arguments usable by downstream operators.

### DIFF
--- a/devito/operator/operator.py
+++ b/devito/operator/operator.py
@@ -1061,6 +1061,16 @@ class ArgumentsMap(dict):
         """The MPI communicator the arguments are collective over."""
         return self.grid.comm if self.grid is not None else MPI.COMM_NULL
 
+    @property
+    def opkwargs(self):
+        temp_registry = {v: k for k, v in platform_registry.items()}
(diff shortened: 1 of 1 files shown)


devito/operator/operator.py
```diff
@@ -1054,6 +1054,16 @@ def comm(self):
...
 
+    @property
+    def opkwargs(self):
+        temp_registry = {v: k for k, v in platform_registry.items()}
+        platform = temp_registry[self.platform]
+
+        temp_registry = {v: k for k, v in compiler_registry.items()}
+        compiler = temp_registry[self.compiler.__class__]
+
+        return {'platform': platform, 'compiler': compiler, 'language': self.language}
```
mloubout, line 1065: Why not `self.compiler` and `self.platform`

`opkwargs` is a bit vague since it only returns the "language" part

FabioLuporini (approved):
I think this is OK

PR #2148, opened 2023-06-17
compiler: Fix pickling of aliasing SparseFunction

FabioLuporini:


georgebisbas (approved):

devito/types/sparse.py
```diff
@@ -488,7 +488,10 @@ def coordinates(self):
 
     @property
     def coordinates_data(self):
-        return self.coordinates.data.view(np.ndarray)
+        try:
+            return self.coordinates.data.view(np.ndarray)
```
mloubout, line 492: `.data._local`

mloubout (commented):
._local already return a proper ndarray view

mloubout (approved):

devito/types/sparse.py
FabioLuporini: since it gets fixed in your PR, I'm merging

PR #2150, opened 2023-06-22
compiler: Add guards to prevent OOB when streaming buffers with ConditionalDimension

ccuetom:
Using the time derivative of a streamed buffer with a ConditionalDimension results in OOB within `init_to_host`. This was not captured in existing tests.

A new test has been added `test_gpu_common.py::test_streaming_multi_input_conddim_foward` that fails with this OOB; and a new guard is now introduced in `Buffering` to prevent it.

### DIFF
--- a/devito/passes/clusters/buffering.py
+++ b/devito/passes/clusters/buffering.py
@@ -179,8 +179,7 @@ class Buffering(Queue):
                 lhs = b.indexed[dims]._subs(dim, b.firstidx.b)
                 rhs = b.function[dims]._subs(dim, b.firstidx.f)
 
-                if dim.is_Conditional:
-                    guards[b.xd] = GuardBound(0, b.firstidx.f)
+                guards[b.xd] = GuardBound(0, b.firstidx.f)
 
             elif b.is_write and init_onwrite(b.function):
                 dims = b.buffer.dimensions
@@ -229,11 +228,8 @@ class Buffering(Queue):
 
                 expr = lower_exprs(Eq(lhs, rhs))
                 ispace = b.readfrom
+                guards = c.guards.andg(b.xd, GuardBound(0, b.firstidx.f))
                 properties = c.properties.sequentialize(d)
-
-                guards = c.guards
-                if b.dim.is_Conditional:
-                    guards = guards.andg(b.xd, GuardBound(0, b.firstidx.f))
 
                 processed.append(
                     c.rebuild(exprs=expr, ispace=ispace,
@@ -266,11 +262,8 @@ class Buffering(Queue):
 
                 expr = lower_exprs(uxreplace(Eq(lhs, rhs), b.subdims_mapper))
                 ispace = b.written
+                guards = c.guards.andg(b.xd, GuardBound(0, b.firstidx.f))
                 properties = c.properties.sequentialize(d)
-
-                guards = c.guards
-                if b.dim.is_Conditional:
-                    guards = guards.andg(b.xd, GuardBound(0, b.firstidx.f))
 
                 processed.append(
                     c.rebuild(exprs=expr, ispace=ispace,
--- a/devito/passes/clusters/buffering.py
+++ b/devito/passes/clusters/buffering.py
@@ -167,6 +167,8 @@ class Buffering(Queue):
             b = Buffer(f, dim, d, accessv, cache, self.options, self.sregistry)
             buffers.append(b)
 
+            guards = {}
+
             if b.is_read or not b.has_uniform_subdims:
                 # Special case: avoid initialization if not strictly necessary
                 # See docstring for more info about what this implies
@@ -176,6 +178,9 @@ class Buffering(Queue):
                 dims = b.function.dimensions
                 lhs = b.indexed[dims]._subs(dim, b.firstidx.b)
                 rhs = b.function[dims]._subs(dim, b.firstidx.f)
+
+                if dim.is_Conditional:
+                    guards[b.xd] = GuardBound(0, b.firstidx.f)
 
             elif b.is_write and init_onwrite(b.function):
                 dims = b.buffer.dimensions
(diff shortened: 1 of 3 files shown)


tests/test_gpu_common.py
```diff
@@ -576,6 +576,34 @@ def test_streaming_multi_input(self, opt, ntmps):
...
+        time_sub = ConditionalDimension(name="time_sub", parent=time_dim, factor=factor)
+
+        u = TimeFunction(name='u', grid=grid, time_order=2, save=nt, time_dim=time_sub)
+        v = TimeFunction(name='v', grid=grid)
+        v1 = TimeFunction(name='v', grid=grid)
+
+        for i in range(u.save):
+            u.data[i, :] = i
+
+        expr = u.dt2 + 3.*u.dt(x0=time_sub - time_sub.spacing)
```
mloubout, line 595: FYI this is just `u.dtl`

tests/test_buffering.py
```diff
@@ -589,7 +589,7 @@ def test_conddim_w_shifting():
     op1 = Operator(eqns, opt='buffering')
 
     # Check generated code
-    assert len(retrieve_iteration_tree(op1)) == 3
+    assert len(retrieve_iteration_tree(op1)) == 4
```
mloubout, line 592: So what changed in the code since it seems to change existing tests not related to `init_to_host`

mloubout (commented):
Looks good to me, is there some standard cond dim cases that were broken as well non related to your issue?

tests/test_gpu_common.py
ccuetom: good to know - I adapted the code from the next test there

tests/test_buffering.py
ccuetom: There tests also generate initialisation code that is affected by the addition of the guard. The loop:

```
  if (time_m <= time_M)
  {
    /* Begin section0 */
    START_TIMER(section0)
    for (int db0 = 0; db0 <= 0; db0 += 1)
    {
      {
        {
          ub0[(db0 + (time_M / factor))%(1)][x][y] = u[db0 + (time_M / factor)][x][y];
        }
      }
    }
    STOP_TIMER(section0,timers)
    /* End section0 */
  }
```

becomes:

```
  if (time_m <= time_M)
  {
    /* Begin section0 */
    START_TIMER(section0)
    for (int db0 = 0; db0 <= 0; db0 += 1)
    {
      if (0 <= db0 + (time_M / factor))
      {
        {
          {
            ub0[(db0 + (time_M / factor))%(1)][x][y] = u[db0 + (time_M / factor)][x][y];
          }
        }
      }
    }
    STOP_TIMER(section0,timers)
    /* End section0 */
  }
```

devito/passes/clusters/buffering.py
```diff
@@ -177,6 +179,9 @@ def callback(self, clusters, prefix, cache=None):
                 lhs = b.indexed[dims]._subs(dim, b.firstidx.b)
                 rhs = b.function[dims]._subs(dim, b.firstidx.f)
 
+                if dim.is_Conditional:
```
FabioLuporini, line 182: what about we place the guard regardless of whether it's a ConditionalDimension or a plain Dimension?

FabioLuporini: second question.

Since `db0` starts, by construction, at 0 and can only take positive values, how about this as a guard instead:

```
guards[pd] = GuardBound(0, b.firstidx.f - b.xd)  # aka: time_M / factor >= 0
```

then later on, you would have to use `andg` to chain this guard with the other `time_M >= time_m` guard

this would generalise to the no-ConditionalDimension case as well

this way, we would generate neater code in my opinion

ccuetom: I might be missing something, but I think the guard does need to be on `db0` - we need to check that `db0 + (time_m / factor) - 1 >= 0` or `db0 + (time_M / factor) - 1 >= 0` depending on whether it's forward or backward.

Using `guards[pd] = GuardBound(0, b.firstidx.f - b.xd)  # aka: time_M / factor >= 0` tests fail as expected.

tests/test_buffering.py
ccuetom: this is the iteration nest before adding the guard:

```
IterationTree(<WithProperties[affine,parallel]::Iteration db0; (0, 0, 1)>, <WithProperties[affine,parallel]::Iteration x; (x_m, x_M, 1)>, <WithProperties[affine,parallel]::Iteration y; (y_m, y_M, 1)>)
IterationTree(<WithProperties[affine,sequential]::Iteration time[t0,t1]; (time_m, time_M, 1)>,)
IterationTree(<WithProperties[affine,sequential]::Iteration time[t0,t1]; (time_m, time_M, 1)>, <WithProperties[affine,parallel,parallel=]::Iteration x; (x_m, x_M, 1)>, <WithProperties[affine,parallel,parallel=]::Iteration y; (y_m, y_M, 1)>)
```

and after adding the guard:

```
IterationTree(<WithProperties[affine,parallel]::Iteration db0; (0, 0, 1)>,)
IterationTree(<WithProperties[affine,parallel]::Iteration db0; (0, 0, 1)>, <WithProperties[affine,parallel]::Iteration x; (x_m, x_M, 1)>, <WithProperties[affine,parallel]::Iteration y; (y_m, y_M, 1)>)
IterationTree(<WithProperties[affine,sequential]::Iteration time[t0,t1]; (time_m, time_M, 1)>,)
IterationTree(<WithProperties[affine,sequential]::Iteration time[t0,t1]; (time_m, time_M, 1)>, <WithProperties[affine,parallel,parallel=]::Iteration x; (x_m, x_M, 1)>, <WithProperties[affine,parallel,parallel=]::Iteration y; (y_m, y_M, 1)>)
```

Iteration tree `IterationTree(<WithProperties[affine,parallel]::Iteration db0; (0, 0, 1)>,)` appears as a result. Not sure if this is expected or if you want me to probe into it.

FabioLuporini: no, it's OK!

devito/passes/clusters/buffering.py
```diff
@@ -226,8 +231,13 @@ def callback(self, clusters, prefix, cache=None):
                 ispace = b.readfrom
                 properties = c.properties.sequentialize(d)
 
+                guards = c.guards
```
FabioLuporini, line 234: nitpicking: guards are typically computed before properties

FabioLuporini (approved):
Thanks for all these changes. Appreciate that. Approved!

mloubout:
Rebase and GTG

FabioLuporini:
Many thanks! Merging...

PR #2155, opened 2023-06-28
Removing AWS from CI

ggorman:
Removing AWS from CI as we have run out of sponsored credits and it is too costly.


mloubout (approved):

FabioLuporini (approved):
OK

PR #2158, opened 2023-07-03
ci: Add python 3.11 and minor CI fixing

georgebisbas:


docker/Dockerfile.cpu
```diff
@@ -2,7 +2,7 @@
 # This Dockerfile contains the Devito codes and can be built using different base images.
 ##############################################################
 
-ARG pyversion=python:3.7
+ARG pyversion=python:3.8
```
EdCaunt, line 5: Why not get ahead of the curve and use Python 3.9 or 3.10 here (and elsewhere)? Looks like the GPU one uses 3.9

.github/workflows/pytest-core-nompi.yml
```diff
@@ -31,14 +31,14 @@ jobs:
...
            pytest-ubuntu-py310-gcc10-noomp,
            pytest-ubuntu-py38-gcc8-omp,
            pytest-ubuntu-py39-gcc9-omp,
-           pytest-osx-py37-clang-omp,
-           pytest-docker-py37-gcc-omp,
-           pytest-docker-py37-icc-omp,
+           pytest-osx-py38-clang-omp,
+           pytest-docker-py38-gcc-omp,
+           pytest-docker-py38-icc-omp,
            pytest-docker-py38-icx-omp
```
EdCaunt, line 42: If the others are bumped up a version, would it make sense to do the same for this one (3.8->3.9)?

.github/workflows/pytest-core-nompi.yml
```diff
@@ -50,15 +50,15 @@ jobs:
...
-        - name: pytest-ubuntu-py38-gcc12-omp
-          python-version: '3.8'
+        - name: pytest-ubuntu-py311-gcc12-omp
+          python-version: '3.11'
           os: ubuntu-22.04
           arch: "gcc-12"
           language: "openmp"
           sympy: "1.10"
 
         - name: pytest-ubuntu-py39-gcc7-omp
```
georgebisbas, line 60: tofix

docker/Dockerfile.cpu
georgebisbas: I am open to this, let's see what others think

.github/workflows/pytest-core-nompi.yml
georgebisbas: I should have another look to "keep some balance for all"

.github/workflows/examples.yml
```diff
@@ -87,7 +87,7 @@ jobs:
         python examples/cfd/example_diffusion.py
 
     - name: Upload coverage to Codecov
-      if: matrix.name != 'pytest-docker-py37-gcc-omp'
+      if: matrix.name != 'pytest-docker-py38-gcc-omp'
```
mloubout, line 90: This isn't right, there is no pytest or matrix name in this workflow it's only run once, if should be removed

docker/Dockerfile.cpu
mloubout: 3.9 would be a safe option yes

mloubout (commented):
What's going on with all those notebook changes

georgebisbas:
> What's going on with all those notebook changes

There is some different latex printing adding this \operatorname

devito/core/autotuning.py
```diff
@@ -122,7 +122,8 @@ def autotune(operator, args, level, mode):
             at_args.update(dict(run))
 
             # Drop run if not at least one block per thread
-            if not configuration['develop-mode'] and nblocks_per_thread.subs(at_args) < 1:
+            if not configuration['develop-mode'] and \
```
georgebisbas, line 125: Fix for Sympy1.12 from @mloubout

devito/tools/utils.py
georgebisbas, line 1: not related to this PR, but minor addition

.github/workflows/examples.yml
```diff
@@ -86,7 +86,7 @@ jobs:
         python examples/cfd/example_diffusion.py
 
     - name: Upload coverage to Codecov
-      if: matrix.name != 'pytest-docker-py37-gcc-omp'
+      if: matrix.name != 'pytest-docker-py39-gcc-omp'
```
mloubout, line 89: Leftover delete line

mloubout (commented):
I wouldn't change the tutorials it's unnecessary diff

mloubout (approved):

FabioLuporini (approved):

PR #2162, opened 2023-07-10
docker: revamp base deployment

mloubout:
- Revamp base deployment not to rebuild nvhpc. Reduces the number of jobs and should reduce CI time since all extras are mostly an env var on top of the existing image. Little bit more verbose but I don't think it's really an issue there
- Added icx as a separate image
- Fix MPI flags for oneapi

### DIFF
--- a/.github/workflows/docker-bases.yml
+++ b/.github/workflows/docker-bases.yml
@@ -101,16 +101,6 @@ jobs:
           build-args: 'arch=icx'
           tags: 'devitocodes/bases:cpu-icx'
 
-      - name: ICC image
-        uses: docker/build-push-action@v3
-        with:
-          context: .
-          file: './docker/Dockerfile.cpu'
-          push: true
-          target: 'icc'
-          build-args: 'arch=icc'
-          tags: 'devitocodes/bases:cpu-icc'
-
 #######################################################
 ################### Nvidia nvhpc ######################
 #######################################################
@@ -250,7 +240,7 @@ jobs:
           context: .
           file: './docker/Dockerfile.amd'
           push: true
-          target: 'aomp'
+          target: 'amdclang'
           tags: devitocodes/bases:amd
 
       - name: AMD HIP image
--- a/.github/workflows/docker-bases.yml
+++ b/.github/workflows/docker-bases.yml
@@ -15,68 +15,14 @@ on:
     - cron: "0 13 * * 1"
 
 jobs:
-  deploy-docker-bases:
+#######################################################
+################### CPU ###############################
+#######################################################
+  deploy-cpu-bases:
     name: ${{ matrix.tag }}
     runs-on: ${{ matrix.runner }}
     env:
       DOCKER_BUILDKIT: "1"
-
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
-          - tag: 'devitocodes/bases:nvidia-nvc'
-            arch: 'arch=nvc'
-            version: 'ver=nvhpc-22-7'
-            dockerfile: './docker/Dockerfile.nvidia'
-            runner: ["self-hosted", "nvidiagpu"]
-
-          - tag: 'devitocodes/bases:nvidia-nvcc'
-            arch: 'arch=nvcc'
-            version: 'ver=nvhpc-22-7'
-            dockerfile: './docker/Dockerfile.nvidia'
-            runner: ["self-hosted", "nvidiagpu"]
-
-          - tag: 'devitocodes/bases:nvidia-clang'
-            arch: 'arch=clang'
-            # clang is only officialy supporting cuda<11.5. The latest nvidia sdk with cuda 11.5 is 21.11 (last of 2021)
-            # We cannot use newer ones such as 22-1 because the package manager automatically installs the latest 2022
-            # that re-install the latest (11.7) version on top of it.
-            # For more info check https://llvm.org/docs/CompileCudaWithLLVM.html and the Prerequisites 
-            version: 'ver=nvhpc-21-11'
-            dockerfile: './docker/Dockerfile.nvidia'
-            runner: ["self-hosted", "nvidiagpu"]
-
-          - tag: 'devitocodes/bases:amd'
-            arch: ''
-            version: ''
-            dockerfile: './docker/Dockerfile.amd'
-            runner: ["self-hosted", "amdgpu"]
-
-          - tag: 'devitocodes/bases:amd-hip'
-            arch: 'arch=hip'
-            version: ''
-            dockerfile: './docker/Dockerfile.amd'
-            runner: ["self-hosted", "amdgpu"]
-          
-          # These ones are tiny runs on default
-          - tag: 'devitocodes/bases:cpu-gcc'
-            arch: 'arch=gcc'
-            version: ''
-            dockerfile: './docker/Dockerfile.cpu'
-            runner: ubuntu-latest 
-          
-          - tag: 'devitocodes/bases:cpu-icc, devitocodes/bases:cpu-icx'
-            arch: 'arch=icc'
-            version: ''
-            dockerfile: './docker/Dockerfile.cpu'
-            runner: ubuntu-latest 
-
-          - tag: 'devitocodes/bases:cpu-nvc'
-            arch: 'arch=nvc-host'
-            version: ''
-            dockerfile: './docker/Dockerfile.nvidia'
-            runner: ["self-hosted", "nvidiagpu"]
 
     steps:
       - name: Checkout devito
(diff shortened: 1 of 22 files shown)


docker/Dockerfile.cpu
```diff
@@ -30,7 +40,7 @@ RUN mkdir -p /deps && mkdir -p /opt/openmpi && cd /deps && \
     mkdir build &&  cd build && \
     ../configure --prefix=/opt/openmpi/ \
                  --enable-mca-no-build=btl-uct --enable-mpi1-compatibility && \
-    make -j $(( $(lscpu | awk '/^Socket\(s\)/{ print $2 }') * $(lscpu | awk '/^Core\(s\) per socket/{ print $4 }') )) && \
+    make -j ${nproc} && \
```
georgebisbas, line 43: why?

docker/Dockerfile.cpu
```diff
@@ -43,38 +53,50 @@ ENV DEVITO_ARCH="gcc"
 ENV DEVITO_LANGUAGE="openmp"
 
 ##############################################################
-# ICC image
+# Intel Oneapi base
 ##############################################################
-FROM base as icc
+FROM base as oneapi
```
georgebisbas, line 58: good

docker/Dockerfile.nvidia
```diff
@@ -98,15 +107,14 @@ ENV NVHPC_CUDA_HOME $HPCSDK_HOME/cuda
...
 ENV PATH $HPCSDK_HOME/compilers/bin:$HPCSDK_HOME/cuda/bin:$HPCSDK_HOME/comm_libs/mpi/bin:${PATH}
 ENV LD_LIBRARY_PATH $HPCSDK_HOME/cuda/lib:$HPCSDK_HOME/cuda/lib64:$HPCSDK_HOME/compilers/lib:$HPCSDK_HOME/math_libs/lib64:$HPCSDK_HOME/comm_libs/mpi/lib:$HPCSDK_CUPTI/lib64:bitcomp_DIR:${LD_LIBRARY_PATH}
-ENV CPATH $HPCSDK_HOME/comm_libs/mpi/include:${CPATH}
+ENV CPATH $HPCSDK_HOME/comm_libs/mpi/include:$HPCSDK_HOME/comm_libs/nvshmem/include:$HPCSDK_HOME/comm_libs/nccl/include:${CPATH}
 
 # MPI
 RUN if [ "x$MPI_VER" = "x4" ]; then \
         rm -f  $HPCSDK_HOME/comm_libs/mpi && \
-        ln -sf $HPCSDK_HOME/comm_libs/openmpi4/openmpi-4.0.5 \
+        ln -sf $HPCSDK_HOME/comm_libs/openmpi4/latest \
```
georgebisbas, line 115: Was this a problem with latest nvhpc?, I think it was `4.1.*`?

docker/Dockerfile.cpu
mloubout: Easier and standard defined env var

docker/Dockerfile.nvidia
mloubout: latest is always defined as a symlink so safer than hardcoded version

.github/workflows/pytest-core-mpi.yml
```diff
@@ -25,6 +25,7 @@ jobs:
     env:
       DEVITO_LANGUAGE: "openmp"
       DEVITO_ARCH: "gcc-9"
+      OMP_NUM_THREADS: "1"
```
georgebisbas, line 28: ?

.github/workflows/pytest-core-mpi.yml
```diff
@@ -52,3 +53,22 @@ jobs:
...
+      - name: Checkout devito
+        uses: actions/checkout@v3
+      
+      - name: Build docker image
+        run: |
+          docker build . --file docker/Dockerfile.devito --tag devito_img --build-arg base=devitocodes/bases:cpu-${{ matrix.arch }}
+      
+      - name: Test with pytest
+        run: |
+          docker run --rm -e CODECOV_TOKEN=${{ secrets.CODECOV_TOKEN }} -e OMP_NUM_THREADS=1 --name testrun devito_img pytest tests/test_mpi.py
```
georgebisbas, line 74: and here?

.github/workflows/pytest-core-mpi.yml
mloubout: Avoid unnecessary thread concurency sonce this is mpi

.github/workflows/pytest-core-mpi.yml
mloubout: same

docker/Dockerfile.amd
```diff
@@ -82,26 +82,8 @@ RUN cd /tmp/ \
...
-    && make -j ${nproc} \
-    && make install
-
 # Cleanup
-RUN rm -rf /tmp/ompi && rm -rf /tmp/ucx
-
+RUN rm -rf /tmp/ucx
 
 # Adding OpenMPI and UCX to Environment
 ENV PATH=$OMPI_HOME/bin:$UCX_HOME/bin:$PATH \
```
ggorman, line 89: You have removed OpenMPI but it's still present in the path.

docker/Dockerfile.devito
```diff
@@ -18,6 +23,7 @@ RUN python3 -m venv /venv && \
     /venv/bin/pip install --no-cache-dir --upgrade pip && \
     /venv/bin/pip install --no-cache-dir jupyter && \
     /venv/bin/pip install --no-cache-dir wheel && \
+    eval "$MPI4PY_FLAGS /venv/bin/pip install --no-cache-dir mpi4py" && \
```
ggorman, line 26: Where does MPI4PY_FLAGS come from? A quick google indicates zero hits. Add link/comment?

requirements-optional.txt
```diff
@@ -1,3 +1,5 @@
 matplotlib
 pandas
-pyrevolve
\ No newline at end of file
+pyrevolve
+scipy
```
ggorman, line 4: Do we need to update any install instructions - what happens when new users try the tutorials? Will they know that these requirements need to be installed?

requirements-testing.txt
```diff
@@ -1,4 +1,8 @@
 pytest>=7.2,<8.0
 pytest-runner
 pytest-cov
-codecov
\ No newline at end of file
+codecov
+flake8>=2.1.0
+nbval
+scipy
+cloudpickle
```
ggorman, line 8: I think that Cloudpickle is a hard requirement at this point because native Python pickle does not work with Devito objects.

ggorman (changes requested):
GTO once green and comments addressed.

docker/Dockerfile.amd
mloubout: I didn't remove open MPI I just moved it below with each compiler but the path is still the same for both images

docker/Dockerfile.devito
mloubout: It's just a made up one I made to handle all our different compilers and arches

requirements-optional.txt
mloubout: Install instruvtions should still be the same and examples should always be installed with `.[extras]` these were leftover that should have been moved here

requirements-testing.txt
mloubout: ok

conftest.py
```diff
@@ -21,6 +21,11 @@
     MPI = None
 
 
+def pytest_collectstart(collector):
+    if collector.fspath and collector.fspath.ext == '.ipynb':
+        collector.skip_compare += ('text/latex',)
```
mloubout, line 26: see https://nbval.readthedocs.io/en/latest/#Skipping-certain-output-types

This just avoid failures for different sympy version printing slightly different latex

mloubout: Also needed to be moved to root directory so that `conftest` is found when testing the examples

georgebisbas: Good to know, so this cancels out my LATEX changes in the other PR?

devito/arch/compiler.py
```diff
@@ -719,9 +732,9 @@ def __lookup_cmds__(self):
         # we try to use `mpiicc` first, while `mpicc` is our fallback, which may
         # or may not be an Intel distribution
         try:
-            check_output(["mpiicc", "--version"]).decode("utf-8")
+            check_output(["mpiicc", "-cc=%s" % self.CC, "--version"]).decode("utf-8")
             self.MPICC = 'mpiicc'
```
FabioLuporini, line 736: are we sure the OneAPI MPI compiler name is still mpiicc, rather than say mpiicx ?

devito/types/dimension.py
```diff
@@ -321,7 +321,8 @@ def _arg_check(self, args, size, interval):
                 upper = interval.upper
             else:
                 # Autopadding causes non-integer upper limit
-                upper = interval.upper.subs(args)
+                from devito.symbolics import normalize_args
```
FabioLuporini, line 324: sooner or later we will have to fix these cycling imports

docker/Dockerfile.amd
```diff
@@ -147,13 +153,31 @@ ENV DEVITO_ARCH="aomp"
...
 ENV DEVITO_LANGUAGE="openmp"
 
-
 ########################################################################
 # HIPCC for GPUs (HIP)
 ########################################################################
 # This will only trigger if arch is hip since the final stage depends on it
 FROM sdk-base as hip
 
+# Install OpenMPI
```
FabioLuporini, line 162: why was OpenMPI expanded out ? there must be a logic, but I'm missing it

requirements-optional.txt
```diff
@@ -1,3 +1,5 @@
...
+distributed
```
FabioLuporini, line 5: excellent

requirements-testing.txt
```diff
@@ -1,4 +1,7 @@
 pytest>=7.2,<8.0
 pytest-runner
 pytest-cov
-codecov
\ No newline at end of file
+codecov
+flake8>=2.1.0
+nbval
+scipy
```
FabioLuporini, line 7: excellent

requirements.txt
```diff
@@ -12,4 +9,4 @@ codepy>=2019.1
 click<9.0
 multidict
 anytree>=2.4.3,<=2.9.0
-distributed<2023.8
+cloudpickle
```
FabioLuporini, line 12: fantastic 👍

docker/Dockerfile.amd
FabioLuporini: ah, for CC=, CXX= etc...

.github/workflows/docker-bases.yml
```diff
@@ -15,68 +19,208 @@ on:
...
+        uses: docker/setup-buildx-action@v2
+
+      - name: Login to DockerHub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+
+      - name: cleanup
+        run: docker system prune -a -f
```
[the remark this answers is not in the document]
mloubout: It should not touch your images only the 'action" user one, always been there.

devito/arch/compiler.py
```diff
@@ -707,6 +709,17 @@ def __init__(self, *args, **kwargs):
             if mpi_distro != 'IntelMPI':
                 warning("Expected Intel MPI distribution with `%s`, but found `%s`"
                         % (self.__class__.__name__, mpi_distro))
+            self.cflags.insert(0, '-cc=%s' % self.CC)
+
+    def get_version(self):
```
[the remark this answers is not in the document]
mloubout: `codepy` does, it's basically copied from there but making sure the `-cc` doesn't make it error

devito/arch/compiler.py
mloubout: Yes

docker/Dockerfile.amd
mloubout: Yeah, decided to go concervative and force all the compiler for openmpi we can revisit

.github/workflows/pytest-core-mpi.yml
```diff
@@ -52,3 +53,22 @@ jobs:
       with:
         token: ${{ secrets.CODECOV_TOKEN }}
         name: pytest-mpi
+
+  test-mpi-docker:
+      name: pytest-mpi
+      runs-on: ubuntu-latest
+      strategy:
+        matrix:
+          arch: [gcc, icc, icx]
```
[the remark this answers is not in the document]
mloubout: Thought so too wasn't sure if we wanted to keep it until intel drop it

examples/seismic/source.py
```diff
@@ -58,10 +58,10 @@ def __init__(self, start=None, step=None, num=None, stop=None):
         if not isinstance(num, int):
             raise TypeError("input argument must be of type int")
 
-        self.start = start
-        self.stop = stop
-        self.step = step
-        self.num = num
+        self.start = float(start)
```
georgebisbas, line 61: I see

.github/workflows/pytest-core-mpi.yml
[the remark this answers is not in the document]
georgebisbas: I would agree to drop it. Worst case scenario if someone is so eager to use icc by default, we could always point out to use the latest stable release

.github/workflows/docker-bases.yml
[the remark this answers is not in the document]
FabioLuporini: wait, I was pretty sure "docker system prune" were a system-level action, that is as if it were issued by the `docker` group?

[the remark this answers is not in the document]
FabioLuporini: uhmmm

.github/workflows/pytest-core-mpi.yml
mloubout: the MPI+openmp is tested with the mpi examples but this is a small 2 core machine that is already oversubscribed with MPI, adding threads only make things slower for no reason

.github/workflows/docker-bases.yml
```diff
@@ -100,13 +132,124 @@ jobs:
...
+
+      - name: AMD image
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          file: './docker/Dockerfile.amd'
+          push: true
+          target: 'aomp'
```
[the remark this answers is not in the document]
mloubout: that's the name of the stage in the dockerfile was easier to keep it but can change it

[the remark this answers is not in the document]
mloubout: fixed

.github/workflows/pytest-core-mpi.yml
FabioLuporini: fair enough

for another day: use some kind of self-hosted runner for MPI+OpenMP build? if agree, maybe add to issue list

docker/README.md
```diff
@@ -10,9 +10,9 @@ Devito provides several images that target different architectures and compilers
 
 We provide two CPU images:
 - `devito:gcc-*` with the standard GNU gcc compiler.
-- `devito:icc-*` with the Intel C compiler for Intel architectures.
+- `devito:icx-*` with the Intel C compiler for Intel architectures.
```
FabioLuporini, line 13: good catch

FabioLuporini (approved):
Fantastic. GTG for me

PR #2164, opened 2023-07-20
compiler: Keep -qopenmp by default after icx 2023.2

georgebisbas:
Passes locally
After icx 2023.2.0, reductions should are passing

### DIFF
--- a/devito/arch/compiler.py
+++ b/devito/arch/compiler.py
@@ -763,7 +763,7 @@ class OneapiCompiler(IntelCompiler):
         platform = kwargs.pop('platform', configuration['platform'])
         language = kwargs.pop('language', configuration['language'])
 
-        # Earlier to versions to OneAPI 2023.2.0 (clang17 underneath), have an OpenMP bug
+        # Earlier versions to OneAPI 2023.2.0 (clang17 underneath), have an OpenMP bug
         if self.version < Version('17.0.0') and language == 'openmp':
             self.ldflags.remove('-qopenmp')
             self.ldflags.append('-fopenmp')
--- a/devito/arch/compiler.py
+++ b/devito/arch/compiler.py
@@ -52,6 +52,8 @@ def sniff_compiler_version(cc):
         compiler = "clang"
     elif ver.startswith("Homebrew clang"):
         compiler = "clang"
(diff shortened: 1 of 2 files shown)


devito/arch/compiler.py
```diff
@@ -761,10 +761,6 @@ def __init__(self, *args, **kwargs):
         platform = kwargs.pop('platform', configuration['platform'])
         language = kwargs.pop('language', configuration['language'])
 
-        if language == 'openmp':
```
mloubout, line 764: There is no guaranty an generic user will have icx >= 2023.2 the version number needs to be checked

georgebisbas: Right, thanks! I was thinking of our docker only! WIll fix

devito/arch/compiler.py
```diff
@@ -761,7 +763,9 @@ def __init__(self, *args, **kwargs):
...
+        # Since OneAPI 2023.2.0 (clang17 underneath), we can use '-qopenmp'
+        # for all our tests
+        if self.version >= Version('17.0.0') and language == 'openmp':
```
FabioLuporini, line 768: isn't this doing the opposite of what the comment says?

also, if you are to change this (maybe not, maybe I misunderstood...), could you drop "for all our tests" which doesn't mean much? what you should have said is that earlier versions have an OpenMP bug. The bug is in their compilers, it doesn't have anything to do with our tests (I mean, not directly)

georgebisbas: damn sorry, you are right

devito/arch/compiler.py
```diff
@@ -761,7 +763,8 @@ def __init__(self, *args, **kwargs):
         platform = kwargs.pop('platform', configuration['platform'])
         language = kwargs.pop('language', configuration['language'])
 
-        if language == 'openmp':
+        # Earlier to versions to OneAPI 2023.2.0 (clang17 underneath), have an OpenMP bug
```
FabioLuporini, line 766: typo, but fine, we can fix it in the next PR

FabioLuporini (approved):

mloubout (approved):

PR #2171, opened 2023-07-25
docker: add some tweaks  to nvidia docker

mloubout:
From @kenhester suggestions

Should fix issues with #2023 as cupy-cuda12x seems to install and work fine inside the image

To be discussed, do we keep nvidia-clang in CI


FabioLuporini:
> To be discussed, do we keep nvidia-clang in CI

Probably we can drop it

NVC for openmp offloading on NVidia cards
Amdclang for openmp offloading on AMD cards
icx for openmp offloading on Intel cards

That's completely fine by me

IIRC, we're still stuck with our own fork of clang, so one more reason to drop it. 

Can we try again NVC + openmp offloading, now that we finally have NVC 23.5 ?

georgebisbas (commented):
I would vote for dropping as well...not popular, not faster, cumbersome for non-docker....
But will we also drop related infrastructure?

mloubout:
>Can we try again NVC + openmp offloading

Yes


Ok removing clang build and replacing by nvc

FabioLuporini (approved):

.github/workflows/docker-devito.yml
```diff
@@ -24,9 +24,9 @@ jobs:
             test: 'tests/test_gpu_openacc.py tests/test_gpu_common.py'
             runner: ["self-hosted", "nvidiagpu"]
 
-          - base: 'bases:nvidia-clang'
-            tag: 'nvidia-clang'
-            flag: '--gpus all'
+          - base: 'bases:nvidia-nvc'
+            tag: 'nvidia-nvc-omp'
+            flag: '--gpus all --env DEVITO_LABGUAGE=openmp'
```
georgebisbas, line 29: LAN*

PR #2173, opened 2023-07-26
examples: free surface for coupled tti eqs

ofmla:
Add FS for TTI case taking advantage of the available structure (subdomain def in `model.py,` `freesurface` fctn with mirroring stuff, etc) created for the acoustic case. If I'm missing something, let me know.


tests/test_adjoint.py
```diff
@@ -90,7 +96,7 @@ def test_adjoint_F(self, mkey, shape, kernel, space_order, time_order, setup_fun
         location from data. This test uses the conventional dot test:
         < Fx, y> = <x, F^T y>
         """
-        tn = 500.  # Final time
+        tn = 1000. if mkey[-3:] == "-fs" else 500.  # Final time (twice the time for fs)
```
mloubout, line 99: Why?

tests/test_adjoint.py
```diff
@@ -155,7 +165,7 @@ def test_adjoint_J(self, mkey, shape, kernel, space_order, time_order, setup_fun
         dot test:
         < Jx, y> = <x ,J^T y>
         """
-        tn = 500.  # Final time
+        tn = 1000. if mkey[-3:] == "-fs" else 500.  # Final time (twice the time for fs)
```
mloubout, line 168: Why?

examples/seismic/tti/wavesolver.py
```diff
@@ -34,6 +34,12 @@ def __init__(self, model, geometry, space_order=4, kernel='centered',
         self.geometry = geometry
         self.kernel = kernel
 
+        if model.fs and kernel == 'staggered':
+            warning("Bad arguments in AnisotropicWaveSolver() - " +
+                    "if model.fs is True, kernel should be centered. " +
+                    "Switching to 'centered'")
+            self.kernel = 'centered'
```
mloubout, line 41: It should raise an error not change user set parameters

tests/test_adjoint.py
ofmla: the truth is that using 500. or 1000. doesn't make difference, the original value can be used.

tests/test_adjoint.py
ofmla: the data in the linearized shot record is close to the end of the register time with very low values and that makes the values of the dot product be very low, that's why I increased the time, although it can be a lower value for example 750.

examples/seismic/tti/wavesolver.py
ofmla: Ok, I will use `raise ValueError` then.

examples/seismic/tti/wavesolver.py
```diff
@@ -34,6 +34,10 @@ def __init__(self, model, geometry, space_order=4, kernel='centered',
         self.geometry = geometry
         self.kernel = kernel
 
+        if model.fs and kernel == 'staggered':
+            raise ValueError("Bad arguments in AnisotropicWaveSolver() - " +
```
mloubout, line 38: "Free surface only supported for centered TTI kernel"

examples/seismic/tti/operators.py
```diff
@@ -20,10 +21,16 @@ def second_order_stencil(model, u, v, H0, Hz, qu, qv, forward=True):
...
-    first_stencil = Eq(unext, stencilp)
-    second_stencil = Eq(vnext, stencilr)
+    first_stencil = Eq(unext, stencilp, subdomain=model.grid.subdomains['physdomain'])
+    second_stencil = Eq(vnext, stencilr, subdomain=model.grid.subdomains['physdomain'])
 
     stencils = [first_stencil, second_stencil]
+
+    # Add free surface
+    if model.fs:
+        stencils.append(freesurface(model, Eq(unext, stencilp)))
```
mloubout, line 31: No need to recreate the eq, just `freesurface(first_stencil)`

tests/test_adjoint.py
```diff
@@ -27,6 +29,8 @@ class TestAdjoint(object):
         ('layers', (60, 70), 'OT2', 8, 2, acoustic_setup),
         ('layers', (60, 70), 'OT2', 4, 2, acoustic_setup),
         ('layers', (60, 70), 'OT4', 2, 2, acoustic_setup),
+        # 2D test with 2 layers and freesurface
+        ('layers-fs', (200, 80), 'OT2', 4, 2, acoustic_setup),
```
mloubout, line 33: We trying to avoid these "larger" sizes for CI as it increases the test time quite a bit, would prefer the standard sizes used with the other cases and keep the `tn` as it is

examples/seismic/tti/operators.py
ofmla: `firs_stencil` is defined with the "physdomain" subdomain, for that reason I used `Eq` again

examples/seismic/tti/wavesolver.py
ofmla: Ok

tests/test_adjoint.py
ofmla: you're right , I'll keep the smaller sizes and 500. ms

mloubout (approved):

PR #2175, opened 2023-07-28
misc: Add deviceid to configuration and enhance switchconfig

ccuetom:
Adds `deviceid` to `configuration` and updates `switchconfig` so that it can act both as a decorator and a context manager.

### DIFF
--- a/devito/operator/operator.py
+++ b/devito/operator/operator.py
@@ -1186,7 +1186,7 @@ def parse_kwargs(**kwargs):
 
     # `allocator`
     kwargs['allocator'] = default_allocator(
-        '%s.%s.%s' % (kwargs['compiler'].__class__.__name__,
+        '%s.%s.%s' % (kwargs['compiler'].name,
                       kwargs['language'],
                       kwargs['platform'])
     )
--- a/devito/operator/operator.py
+++ b/devito/operator/operator.py
@@ -1186,7 +1186,9 @@ def parse_kwargs(**kwargs):
 
     # `allocator`
     kwargs['allocator'] = default_allocator(
-        '%s.%s' % (kwargs['compiler'].__class__.__name__, kwargs['language'])
+        '%s.%s.%s' % (kwargs['compiler'].__class__.__name__,
+                      kwargs['language'],
+                      kwargs['platform'])
     )
 
     return kwargs
--- a/devito/parameters.py
+++ b/devito/parameters.py
@@ -7,6 +7,7 @@ from functools import wraps
 from devito.logger import info, warning
 from devito.tools import Signer, filter_ordered
 
+
 __all__ = ['configuration', 'init_configuration', 'print_defaults', 'print_state',
            'switchconfig']
 
@@ -159,7 +160,8 @@ env_vars_mapper = {
     'DEVITO_FIRST_TOUCH': 'first-touch',
     'DEVITO_JIT_BACKDOOR': 'jit-backdoor',
     'DEVITO_IGNORE_UNKNOWN_PARAMS': 'ignore-unknowns',
-    'DEVITO_SAFE_MATH': 'safe-math'
+    'DEVITO_SAFE_MATH': 'safe-math',
+    'DEVITO_DEVICEID': 'deviceid'
 }
 
 env_vars_deprecated = {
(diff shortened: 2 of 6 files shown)


devito/types/parallel.py
```diff
@@ -252,7 +252,7 @@ class DeviceID(DeviceSymbol):
 
     @property
     def default_value(self):
-        return -1
+        return configuration['deviceid']
```
FabioLuporini, line 255: We can return a `default_value` based on global state

I think the better way to go here would be overriding `_arg_values` (from `Scalar`)

Within the override, we check, in order:

* whether kwargs has a user-override for `deviceid`
* whether `configuration['deviceid']` is != `default_value`
* else return `default_value`

FabioLuporini (changes requested):
The changes in switchconfig are solid -- thanks!

My only request here is to edit DeviceID (see comment in review)

devito/parameters.py
```diff
@@ -7,6 +7,7 @@
 from devito.logger import info, warning
 from devito.tools import Signer, filter_ordered
 
+
```
georgebisbas, line 10: ?

devito/operator/operator.py
```diff
@@ -1186,7 +1186,9 @@ def parse_kwargs(**kwargs):
 
     # `allocator`
     kwargs['allocator'] = default_allocator(
-        '%s.%s' % (kwargs['compiler'].__class__.__name__, kwargs['language'])
+        '%s.%s.%s' % (kwargs['compiler'].__class__.__name__,
```
mloubout, line 1189: Can we just make a `name` property in our Compilers so we don't have to do this

devito/types/parallel.py
```diff
@@ -254,6 +254,14 @@ class DeviceID(DeviceSymbol):
     def default_value(self):
         return -1
 
+    def _arg_values(self, **kwargs):
+        if self.name in kwargs:
+            return {self.name: kwargs.pop(self.name)}
+        elif configuration['deviceid'] != self.default_value:
+            return {self.name: configuration['deviceid']}
+        else:
+            return {self.name: self.default_value}
```
mloubout, line 263: why not just `{self.name: kwargs.pop(self.name, self.default_value)}` and have `default_value` return `configuration['deviceid']`

devito/parameters.py
```diff
@@ -232,31 +225,38 @@ def init_configuration(configuration=configuration, env_vars_mapper=env_vars_map
...
             self.params = {k.replace('_', '-'): v for k, v in params.items()}
         else:
             self.params = {}
+        self.previous = {}
+
+    def __enter__(self, condition=True, **params):
+        self.previous = {}
+        for k, v in self.params.items():
+            self.previous[k] = configuration[k]
+            configuration[k] = v
```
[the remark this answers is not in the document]
ccuetom: this has a slightly different effect, although I could write it more succinctly:

```
self.previous = {k: configuration[k] for k in self.params.keys()}
configuration.update(**self.params)
```

devito/parameters.py
```diff
@@ -232,31 +225,38 @@ def init_configuration(configuration=configuration, env_vars_mapper=env_vars_map
...
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        for k, v in self.params.items():
+            try:
+                configuration[k] = self.previous[k]
+            except ValueError:
+                # E.g., `platform` and `compiler` will end up here
+                try:
+                    configuration[k] = self.previous[k].name
+                except AttributeError:
+                    super(Parameters, configuration).__setitem__(k, self.previous[k])
```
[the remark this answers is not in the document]
ccuetom: Not exactly, this bypasses the `@_check_key_value` decorator and the `_update_functions` callback. The value we're setting back into the `configuration` has already gone through those and in some cases (e.g. compiler and platform) it has been transformed. So the previous checks and transformations need to be skipped.

devito/types/parallel.py
ccuetom: isn't that what the original code did? I thought the idea was to keep `default_value` and `configuration['deviceid']` separate?

FabioLuporini (approved):

georgebisbas (commented):
I guess no tests are needed for this PR for this new functionality?

mloubout (approved):

devito/types/parallel.py
FabioLuporini: yes, agree with Carlos here. We can't just return a (mutable) global value from `default_value` :-)

PR #2176, opened 2023-08-01
builtins: Support batched initialize_function

FabioLuporini:


devito/builtins/initializers.py
```diff
@@ -286,73 +346,45 @@ def initialize_function(function, data, nbl, mapper=None, mode='constant',
           [2, 3, 3, 3, 3, 2],
           [2, 2, 2, 2, 2, 2]], dtype=int32)
     """
-    name = name or 'pad_%s' % function.name
-    if isinstance(function, dv.TimeFunction):
+    if isinstance(function, (list, tuple)):
+        if not isinstance(data, (list, tuple)):
+            raise TypeError("Expected a list of `data`")
```
mloubout, line 351: I think we could support this case for `init([f,g,h], data)` (ie all function initialized to smae data) with just `data = [data]*len(function)`

devito/builtins/initializers.py
```diff
@@ -286,73 +346,45 @@ def initialize_function(function, data, nbl, mapper=None, mode='constant',
...
-        else:
-            function.data[:] = data[:]
-        if pad_halo:
-            pad_outhalo(function)
-        return
-
-    nbl, slices = nbl_to_padsize(nbl, function.ndim)
-    if isinstance(data, dv.Function):
-        function.data[slices] = data.data[:]
+        for f in functions:
```
mloubout, line 369: `for (f, d) in zip(functions, datas)`

devito/builtins/initializers.py
```diff
@@ -286,73 +346,45 @@ def initialize_function(function, data, nbl, mapper=None, mode='constant',
...
-                options.extend(options_extra)
-            else:
-                options.extend([options_extra])
-
-    if all(options is None for i in options):
-        options = None
+        assert len(lhss) == len(rhss) == len(optionss)
 
-    assign(lhs, rhs, options=options, name=name, **kwargs)
+        name = name or 'pad_%s' % '_'.join(f.name for f in functions)
```
mloubout, line 385: would use "init" instead of pad as the generic case

devito/builtins/initializers.py
```diff
@@ -286,73 +346,45 @@ def initialize_function(function, data, nbl, mapper=None, mode='constant',
...
-    if isinstance(function, dv.TimeFunction):
+    if isinstance(function, (list, tuple)):
+        if not isinstance(data, (list, tuple)):
+            raise TypeError("Expected a list of `data`")
+        elif len(function) != len(data):
+            raise ValueError("Expected %d `data` items, got %d" %
+                             (len(function), len(data)))
+
+        if mapper is not None:
+            raise NotImplementedError("Unsupported `mapper` with batching")
```
mloubout, line 357: Not even one mapper per function?

tests/test_builtins.py
```diff
@@ -327,6 +327,23 @@ def test_if_halo_mpi(self, nbl):
...
             assert np.all(f._data_with_outhalo._local == expected)
 
+    def test_batching(self):
+        grid = Grid(shape=(12, 12))
+
+        a = np.arange(16).reshape((4, 4))
+
+        f = Function(name='f', grid=grid, dtype=np.int32)
+        g = Function(name='g', grid=grid, dtype=np.int32)
+        h = Function(name='h', grid=grid, dtype=np.int32)
```
mloubout, line 337: Would mix dtype to test "general" case

mloubout (approved):

PR #2198, opened 2023-09-07
api: Cleanup and improve SubFunction

### LINKED ISSUE #2192: SubFunction. circular dependency
Having a subfunction keep its parent (in `__rkwargs__` and as a property) is a circular dependency problematic for object reconstruction and pickling.

This is partly related to SubFunction handling in general for operations on aprents such as `subs` or `xreplace`

mloubout:
Fixes #2192 

- Fixes `subs/rebuild` of SparseFunction to properly propagate through the subfunctions (coords, ...)
- Add tests

### DIFF
--- a/devito/passes/clusters/factorization.py
+++ b/devito/passes/clusters/factorization.py
@@ -142,13 +142,16 @@ def collect_nested(expr):
                     terms.append(i)
 
             # Collect common funcs
-            w_funcs = Add(*w_funcs, evaluate=False)
-            w_funcs = collect(w_funcs, funcs, evaluate=False)
-            try:
-                terms.extend([Mul(k, collect_const(v), evaluate=False)
-                              for k, v in w_funcs.items()])
-            except AttributeError:
-                assert w_funcs == 0
+            if len(w_funcs) > 1:
+                w_funcs = Add(*w_funcs, evaluate=False)
+                w_funcs = collect(w_funcs, funcs, evaluate=False)
+                try:
+                    terms.extend([Mul(k, collect_const(v), evaluate=False)
+                                  for k, v in w_funcs.items()])
+                except AttributeError:
+                    assert w_funcs == 0
+            else:
+                terms.extend(w_funcs)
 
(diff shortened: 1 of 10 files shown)


devito/passes/clusters/factorization.py
```diff
@@ -142,13 +142,16 @@ def run(expr):
...
 
             # Collect common funcs
-            w_funcs = Add(*w_funcs, evaluate=False)
-            w_funcs = collect(w_funcs, funcs, evaluate=False)
-            try:
-                terms.extend([Mul(k, collect_const(v), evaluate=False)
-                              for k, v in w_funcs.items()])
-            except AttributeError:
-                assert w_funcs == 0
+            if len(w_funcs) > 1:
```
mloubout, line 145: @FabioLuporini  this does not fix the pickle reconstruction issue with `(-1)*` so left the `_print_Mul` in printer for the time being

tests/test_interpolation.py
```diff
@@ -242,8 +242,8 @@ def test_precomputed_injection_time(r):
     sf = PrecomputedSparseTimeFunction(name='s', grid=m.grid, r=r, npoint=len(coords),
                                        gridpoints=gridpoints, nt=nt,
                                        interpolation_coeffs=interpolation_coeffs)
-
-    expr = sf.inject(m, Float(1.))
+    sf.data.fill(1.)
+    expr = sf.inject(m, sf)
```
mloubout, line 246: This one is actually a consequence of removing `parent` as injecting `1` akes an operator that does not process `sf` anymore as an argument as it was only processed through the coordinates as a parent

devito/passes/clusters/factorization.py
FabioLuporini: OK. Can you open an issue about that and refer to the failing test without the hack?

FabioLuporini (approved):
Thanks, uncontroversial

georgebisbas (approved):

PR #2200, opened 2023-09-08
compiler: Fix data dependency issue on dimension-wide WAR

georgebisbas:


FabioLuporini (commented):
Looking good to me!

Minors: can you rebase, merge together squashable commits, rename them to honor our commit message rules (@georgebisbas can explain and point to the right documentation?

Also (another nitpick), I'm incrementally trying to shorten our comments and to wrap them around the 80th(ish) character, while the previous limit wash 90 characters long rows. While this is still loosely applied, it would help to refactor the comments to keep this mind. Also, as I said, the more compact, the better. And, final nitpick: if a comment belongs to e.g. the body of an `if`, it should go inside the `if`, not outside

devito/ir/support/basic.py
```diff
@@ -426,7 +426,25 @@ def distance(self, other):
...
+                # We are writing over an entire dimension but reading from one point.
+                # If there are overlaps between the two then we would have a dependency
+                # This is a conservative estimation as there are cases (example below)
+                # where we may or may not have a dependency given that we don't write
+                # depending on domain size, which is not compilation-time known
+
+                # For example:
+                # Eq(u[0, y], 1)
+                # Eq(u[1, y+1], u[0, 1])
+                elif i.is_Number and not j.is_Number:
```
FabioLuporini, line 446: instead of `not j.is_Number` you can also use `j.is_Symbol`, they should be equivalent in this context

also you must cover the symmetric case:
```
elif (i.is_Number and j.is_Symbol) or (i.is_Symbol and j.is_Number):
```

devito/ir/support/basic.py
```diff
@@ -426,7 +426,25 @@ def distance(self, other):
...
                 v = i - j
                 if v.is_Number and v.is_finite:
-                    return Vector(S.ImaginaryUnit)
+                    # If i and j are numbers, we append an (I) distance
+                    if i.is_Number and j.is_Number:
+                        return Vector(S.ImaginaryUnit)
+                    # If both i and j are not numbers, there may be dimension-dependent
+                    # dependencies so we append the distance
+                    else:
+                        ret.append(v)
```
FabioLuporini, line 435: we might want to append Infinite here, otherwise I think we would have never ended up in this place, because we would have never left the first big loop at the top of the method over the iteration intervals

devito/ir/support/basic.py
FabioLuporini: (parametrize test?)

AndrewCheng827: > instead of `not j.is_Number` you can also use `j.is_Symbol`, they should be equivalent in this context
> 
> also you must cover the symmetric case:
> 
> ```
> elif (i.is_Number and j.is_Symbol) or (i.is_Symbol and j.is_Number):
> ```

For this George and I figured that we only need `i.is_Number and j.is_Symbol` case since we would only have a dependency if we are reading from one location and writing over an entire dimension. The other case would have been caught previously. 

Consider the equations
```
eq0 = Eq(u[0,y], 1)
eq1 = Eq(u[1,1], u[0, y+1])
```
The dependency between reading from `u[0, y+1]` and writing to `u[0,y]` is caught by
```
v = i - j
if v.is_Number and v.is_finite:
    # If i and j are numbers, we append an (I) distance
    if i.is_Number and j.is_Number:
        return Vector(S.ImaginaryUnit)
    # If both i and j are not numbers, 
    # there may be dimension-dependent dependencies 
    # so we append the distance
    else:
        ret.append(v)
``` 
The dependency between reading from `u[0,y+1]` and writing to `u[1,1]` is imaginary. Hence the algorithm doesn't append anything to the distance vector

devito/ir/support/basic.py
AndrewCheng827: The equation pair
```
eq0 = Eq(u[0,y], 1)
eq1 = Eq(u[1,1], u[0, y+1])
```
would resulting in the execution ending up at line 435

Also I'm confused to why adding an Infinite here would help with never reaching this part of the code

devito/ir/support/basic.py
```diff
@@ -426,7 +426,25 @@ def distance(self, other):
             else:
                 v = i - j
                 if v.is_Number and v.is_finite:
-                    return Vector(S.ImaginaryUnit)
+                    # If i and j are numbers, we append an (I) distance
+                    if i.is_Number and j.is_Number:
+                        return Vector(S.ImaginaryUnit)
+                    # If both i and j are not numbers, there may be dimension-dependent
```
FabioLuporini, line 432: this is also another useless comment, maybe we can instead use a tiny `E.g.`

FabioLuporini: possibly *inside* the *else*, not on the outside...

devito/ir/support/basic.py
FabioLuporini: I'm not so particularly convinced we shouldn't care about the symmetrict case

what if you have

```
Eq(u[0, 1], 1)
Eq(u[x, y], u[0, y])
```

?

Also:
* you're still using "not j.is_Number", instead of "is_Symbol"
* how is your solution any stronger and/or more robust than just having
```
if i.is_Symbol or j.is_Symbol:
   ..
```

FabioLuporini (changes requested):
History requires rebase and squashing

commit messages do not adhere to our format. Please see the contributing guidelines

FabioLuporini:
Also: pep8 isn't honored. See failing build

devito/ir/support/basic.py
```diff
@@ -426,7 +426,32 @@ def distance(self, other):
...
+                        # so we append the distance
+
+                        # For example:
+                        # Eq(u[0, y], 1)
+                        # Eq(u[1, 1], u[0, y+1])
+                        ret.append(S.Infinity)
+
+                # We are writing over an entire dimension
+                # but reading from one point.
+                # If there are overlaps between the two
```
FabioLuporini, line 443: Drop this entire paragraph (starting with "If there are overlaps...")

devito/ir/support/basic.py
```diff
@@ -426,7 +426,32 @@ def distance(self, other):
...
+                        # If both i and j are not numbers,
+                        # there may be dimension-dependent dependencies
+                        # so we append the distance
+
+                        # For example:
+                        # Eq(u[0, y], 1)
+                        # Eq(u[1, 1], u[0, y+1])
+                        ret.append(S.Infinity)
+
+                # We are writing over an entire dimension
```
FabioLuporini, line 441: you're breaking to a new line too soon. Use around 83~84 characters, then break line

devito/ir/support/basic.py
```diff
@@ -426,7 +426,32 @@ def distance(self, other):
             else:
                 v = i - j
                 if v.is_Number and v.is_finite:
-                    return Vector(S.ImaginaryUnit)
+                    if i.is_Number and j.is_Number:
+                        return Vector(S.ImaginaryUnit)
+                    else:
+                        # If both i and j are not numbers,
```
FabioLuporini, line 432: Drop this paragraph

devito/ir/support/basic.py
```diff
@@ -426,7 +426,32 @@ def distance(self, other):
...
+                        # there may be dimension-dependent dependencies
+                        # so we append the distance
+
+                        # For example:
```
FabioLuporini, line 436: First of all, why in this example is the first Eq even needed? is the dependency along `y` across the u[1,1] write and the u[0, y+1] read?

Also, rewrite for homogeneity (and assuming what I wrote above is correct) as

```
# E.g., `self=W<f,[1, 1]>` and `other=R<f[0, y+1]>`
```

devito/ir/support/basic.py
FabioLuporini: also simplify as "Writing (reading) over an entire dimension, reading (writing) from one point"

FabioLuporini (changes requested):
Left a few more comments

Also, given the nature and the size of the edits, I suggest to squash all of the 5 commits into a single commit with message "compiler: Patch ..."

devito/ir/support/basic.py
AndrewCheng827: The first equation is necessary since this `else` statement catches a dependency between writing into `[u0,y]` and reading from `u[0, y+1]` (we can' t schedule the two statements into the same loop because of this, or else the read from `u[0, y+1]` will read points not yet initialized by `u[0,y]`

AndrewCheng827: do I write it as `self=W<u,[0.y]` and `other=R<u,[0,y+1]>` then?

AndrewCheng827: And similarly for the example I gave about 10 lines below:
```
Eq(u[0,y], 1)
Eq(u[1,y+1], u[0,1])
```
I would write
`self=R<u,[1,2]> and other=W<u,[1,y+1]>` instead?

FabioLuporini: right, yes

georgebisbas:
Closing in favour of #2212

PR #2212, opened 2023-09-20
compiler: Fix issue 2194

AndrewCheng827:


mloubout (approved):

devito/ir/support/basic.py
```diff
@@ -426,7 +426,20 @@ def distance(self, other):
...
+                    if i.is_Number and j.is_Number:
+                        return Vector(S.ImaginaryUnit)
+                    else:
+                        # For example:
+                        # self=W<u,[0,y]> and other=R<u,[0,y+1]>
+                        ret.append(v)
+
+                # Writing (reading) over an entire dimension, reading (writing)
+                # from one point
+
```
FabioLuporini, line 438: nitpicking, I wouldn't normally say it, but since this is your learning phase and since this might be the first of many PRs... :

* Drop blank line;
* Move the "For example ..." on the same line as "# from one point. For example:" (after the for example, it's ok to break line)

PR #2218, opened 2023-09-27
compiler: prevent temporary for local reductions

mloubout:


devito/types/sparse.py
```diff
@@ -338,11 +338,11 @@ def guard(self, expr=None):
                                   condition=condition, indirect=True)
 
         if expr is None:
-            out = self.indexify().xreplace({self._sparse_dim: cd})
+            out = self.indexify()._subs(self._sparse_dim, cd)
         else:
             functions = {f for f in retrieve_function_carriers(expr)
                          if f.is_SparseFunction}
-            out = indexify(expr).xreplace({f._sparse_dim: cd for f in functions})
+            out = indexify(expr).subs({f._sparse_dim: cd for f in functions})
```
FabioLuporini, line 345: why subs and not _subs?

and why subs and not xreplace? just performance or because it can be tricky math exprs?

devito/ir/equations/equation.py
```diff
@@ -73,6 +73,11 @@ def apply(self, func):
         kwargs['conditionals'] = {k: func(v) for k, v in self.conditionals.items()}
         return self.func(*args, **kwargs)
 
+    def __repr__(self):
+        if not self.is_Reduction:
+            return super().__repr__()
+        else:
+            return '%s = %s(%s, %s)' % (self.lhs, self.operation, self.lhs, self.rhs)
```
FabioLuporini, line 80: Nitpicking, aesthetics, feel free to ignore...:

```
elif self.operation is OpInc:
    return 'Inc(%s, %s)' % ...
else:
    return '%s(%s, %s)' % (self.operation, ...)
```

devito/ir/clusters/cluster.py
```diff
@@ -230,7 +230,18 @@ def is_dense(self):
 
     @property
```
FabioLuporini, line 231: now deserves to be cached

devito/ir/clusters/cluster.py
```diff
@@ -230,7 +230,18 @@ def is_dense(self):
...
+        A cluster is sparse if it represent a sparse operation i.e if both
+
+            * The cluster contains sparse functions
+            * The cluster uses dense functions
+
+        If only the first case is true, the cluster only contains operation on the sparse
+        function itself without indirection and therefore only contains dense operations.
+        """
+        return (any(f.is_SparseFunction for f in self.functions) and
+                len([f for f in self.functions
```
FabioLuporini, line 243: what if an Array?

perhaps a bit safer:

```
return all(split(self.functions, lambda f: f.is_SparseFunction))
```

?

devito/ir/clusters/algorithms.py
```diff
@@ -471,7 +471,8 @@ def normalize_reductions(cluster, sregistry, options):
             processed.extend([e.func(v, e.rhs, operation=None),
                               e.func(e.lhs, v)])
 
-        elif e.is_Reduction and e.lhs.is_Symbol and opt_mapify_reduce:
+        elif e.is_Reduction and e.lhs.is_Symbol and opt_mapify_reduce \
+                and not cluster.is_sparse:
```
FabioLuporini, line 475: this pass is getting tricky...

maybe I would split it into two passes, `normalize_reductions_sparse` and `normalize_reductions_dense`. 

If u agree, to do that, change `@cluster_pass(mode='all')` to `@cluster_pass(mode='sparse')` and `@cluster_pass(mode='dense')` respectively

Since you're doing it... note that instead of

```
 dims = [d for d, v in cluster.properties.items() if PARALLEL_IF_ATOMIC in v]
```

you could add `is_parallel_atomic`, akin to https://github.com/devitocodes/devito/blob/master/devito/ir/support/properties.py#L208


Anyway, bottom line is: by decoupling this pass into two pass, things get way simpler

FabioLuporini (changes requested):

devito/types/sparse.py
mloubout: >why subs and not _subs?

Multile entries in the dict, `_subs` is for single replacement `_subs(a, b)` 

> and why subs and not replace?

xreplace bit more expensive and end just a bit of extra stuff before calling `subs`.

Note: `indexify` here is ugly but leads to massive issues when removing but for another time.

devito/ir/equations/equation.py
mloubout: Ok. It's not really "used" anywhere was just easier for debug with `embed`, will update is nicer indeed

devito/ir/clusters/cluster.py
mloubout: ok

devito/ir/clusters/cluster.py
mloubout: Not sure that's enough, that doesn't check if there actually is a sparse function I'll see if can simplify

devito/ir/clusters/algorithms.py
mloubout: done

devito/ir/clusters/cluster.py
FabioLuporini: ah OK

FabioLuporini (approved):

PR #2219, opened 2023-09-27
api: cleanup SubDimension and SubDomain

mloubout:


devito/passes/clusters/implicit.py
```diff
@@ -120,7 +121,7 @@ def callback(self, clusters, prefix):
                 # once and for all at the top of the current IterationInterval,
                 # and reuse them for one or more (potentially non-consecutive)
                 # `clusters`
-                if ispaceN not in seen:
+                if ispaceN not in seen or (not ispaceN and d not in seend):
```
FabioLuporini, line 124: two questions:

* Is the `not ispaceN` really necessary? we get to this point only if `len(ispaceN) == 0`, so it should be implicit?
* Instead of adding `seend`, can we not just put everything inside `seen`? that is, `seen` contains both IterationSpaces and plain Dimensions...

On second thought, I'm confused by the _original_ code. If `len(ispaceN) == 0`, then it means `ispaceN == IterationSpace([])` no? I think I'm not seeing something obvious, and therefore at least the first comment above might make no sense...

BTw, since you're touching this file... could you polish:
* `Interval(i, 0, 0) -> Interval(i)` (a few lines above)

mloubout: My understanding is that you could have an `IterationSpace` of length `0` but not empty in some weird corner case which is why was needed. but yeah can probably drop the `not ispaceN`

FabioLuporini (approved):
OK, GTG now

PR #2222, opened 2023-10-05
dsl: Add functionality to intersect SubDimensions

EdCaunt:
Introduce the ability to form a new `SubDimension` as the intersection of multiple` SubDimension`s

### DIFF
--- a/devito/types/dimension.py
+++ b/devito/types/dimension.py
@@ -607,6 +607,65 @@ class SubDimension(DerivedDimension):
                    thickness=((lst, thickness_left), (rst, thickness_right)),
                    local=local)
 
+    @classmethod
+    def intersection(cls, sdims):
+        # Check all SubDimensions supplied overlap
+        # NOTE: Think the test for overlap isn't strict enough?
+        if not all([sdims[0].overlap(sdim) for sdim in sdims[1:]]):
+            raise ValueError("SubDimensions do not all overlap")
+
+        name = '_n_'.join([sdim.name for sdim in sdims])
+        try:
+            [parent] = set([sdim.parent for sdim in sdims])
+        except ValueError:
+            raise ValueError("Inconsistent SubDimension parents")
+
+        local = any([sdim.local for sdim in sdims])
+
+        if all([sdim.is_left for sdim in sdims]):
+            # Only left SubDimensions results in left SubDimension
+            thickness = sympy.Min(*[sdim.thickness.left[1]
+                                    for sdim in sdims])
+            return cls.left(name, parent,
+                            thickness,
+                            local=local)
+        elif all([sdim.is_right for sdim in sdims]):
+            # Only right SubDimensions results in right SubDimension
+            thickness = sympy.Min(*[sdim.thickness.right[1]
+                                    for sdim in sdims])
+            return cls.right(name, parent,
+                             thickness,
+                             local=local)
+        else:
+            # Intersection of different types of subdimension or two middle
+            # will always be middle
+            width = parent.symbolic_max - parent.symbolic_min + 1
+
+            def get_left_thickness(sdim):
+                if sdim.is_left:
+                    return 0
+                elif sdim.is_right:
+                    return width - sdim.thickness.right[1]
+                else:
+                    return sdim.thickness.left[1]
+
+            def get_right_thickness(sdim):
+                if sdim.is_left:
+                    return width - sdim.thickness.left[1]
+                elif sdim.is_right:
+                    return 0
+                else:
+                    return sdim.thickness.right[1]
+
+            thickness_left = sympy.Max(*[get_left_thickness(sdim)
+                                         for sdim in sdims])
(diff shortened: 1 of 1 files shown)


devito/types/dimension.py
```diff
@@ -607,6 +607,65 @@ def middle(cls, name, parent, thickness_left, thickness_right, local=False):
                    thickness=((lst, thickness_left), (rst, thickness_right)),
                    local=local)
 
+    @classmethod
+    def intersection(cls, sdims):
```
mloubout, line 611: I think this should not be a class method, similar to `Set` that defines intersection as `set().intersectio` not `Set.intersection`.

devito/types/dimension.py
```diff
@@ -607,6 +607,65 @@ def middle(cls, name, parent, thickness_left, thickness_right, local=False):
...
+        # Check all SubDimensions supplied overlap
+        # NOTE: Think the test for overlap isn't strict enough?
+        if not all([sdims[0].overlap(sdim) for sdim in sdims[1:]]):
```
mloubout, line 614: Would make it easier here to with just `self.overlap ...`

devito/types/dimension.py
```diff
@@ -607,6 +607,65 @@ def middle(cls, name, parent, thickness_left, thickness_right, local=False):
...
+            raise ValueError("SubDimensions do not all overlap")
+
+        name = '_n_'.join([sdim.name for sdim in sdims])
```
mloubout, line 617: `_i_` ?

devito/types/dimension.py
```diff
@@ -607,6 +607,65 @@ def middle(cls, name, parent, thickness_left, thickness_right, local=False):
...
+        try:
+            [parent] = set([sdim.parent for sdim in sdims])
```
mloubout, line 619: {self.parent}.union(*[sdim.parent for sdim in sdims])

devito/types/dimension.py
```diff
@@ -607,6 +607,65 @@ def middle(cls, name, parent, thickness_left, thickness_right, local=False):
...
+        except ValueError:
+            raise ValueError("Inconsistent SubDimension parents")
+
+        local = any([sdim.local for sdim in sdims])
+
+        if all([sdim.is_left for sdim in sdims]):
+            # Only left SubDimensions results in left SubDimension
+            thickness = sympy.Min(*[sdim.thickness.left[1]
```
mloubout, line 627: use our `Min` not sympy

devito/types/dimension.py
```diff
@@ -607,6 +607,65 @@ def middle(cls, name, parent, thickness_left, thickness_right, local=False):
...
+                                    for sdim in sdims])
+            return cls.left(name, parent,
+                            thickness,
+                            local=local)
+        elif all([sdim.is_right for sdim in sdims]):
+            # Only right SubDimensions results in right SubDimension
+            thickness = sympy.Min(*[sdim.thickness.right[1]
```
mloubout, line 634: same, `devito.Min`

devito/types/dimension.py
```diff
@@ -607,6 +607,65 @@ def middle(cls, name, parent, thickness_left, thickness_right, local=False):
...
+                                    for sdim in sdims])
+            return cls.right(name, parent,
+                             thickness,
+                             local=local)
+        else:
+            # Intersection of different types of subdimension or two middle
+            # will always be middle
+            width = parent.symbolic_max - parent.symbolic_min + 1
+
+            def get_left_thickness(sdim):
```
mloubout, line 644: define as instance attribute outside, and then just `self/sdim.right_thickness` , no need for "get"

mloubout (commented):
Some comments, can discuss.

This needs bunch of tests including with MPI. Do we wanna define `union` as well?

devito/types/dimension.py
EdCaunt: I think it wants to be a class method since you could conceivably want the intersection of several `SubDimension`s, so it would want to be `intersect(subdimensions)` rather than `intersect(self, other)` imo

devito/types/dimension.py
EdCaunt: Was thinking it looks like the set intersection symbol if you squint at it?

EdCaunt:
> This needs bunch of tests including with MPI. Do we wanna define `union` as well?
Tests to come. Union is more complex and less urgent so was going to leave for now

devito/types/dimension.py
mloubout: You can still have multiple ones, again similar to a set `set(a).intersection(a,b,c,d,e,f)` there is nothing that prevents it. to have more than one without being a class method

devito/types/dimension.py
EdCaunt: Note that this is both getting the parent and checking they all share a parent, since the unpack will fail with `ValueError` if the set does not contain one item

FabioLuporini (changes requested):
add tests

EdCaunt:
Dead, closing

PR #2230, opened 2023-10-11
compiler: prevent multisubdimension expressions duplicates

mloubout:
Can't easily make a test for it here. Breaks `abox` with extra dims

### DIFF
--- a/devito/passes/clusters/implicit.py
+++ b/devito/passes/clusters/implicit.py
@@ -7,6 +7,7 @@ from math import floor
 
 from devito.ir import (Cluster, Interval, IntervalGroup, IterationSpace, Queue,
                        FetchUpdate, PrefetchUpdate, SEQUENTIAL)
+from devito.symbolics import retrieve_dimensions
 from devito.tools import as_tuple, timed_pass
 from devito.types import Eq
 from devito.types.grid import MultiSubDimension, SubDomainSet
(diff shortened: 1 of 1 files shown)


FabioLuporini (approved):
So a test is being added to PRO right?

Approved!

devito/passes/clusters/implicit.py
```diff
@@ -106,6 +107,13 @@ def callback(self, clusters, prefix):
             # The "implicit expressions" created for the MultiSubDomain
             exprs, dims, sub_iterators = make_implicit_exprs(d.msd, c)
 
+            # Make sure the "implicit expressions" aren't scheduled in
+            # an inner loop. E.g schedule both for `t, xi, yi` and `t, d, xi, yi`
```
FabioLuporini, line 111: what do you mean by "both" here? 
I think the example could be a bit more clear here
OK in another PR

devito/passes/clusters/implicit.py
```diff
@@ -106,6 +107,13 @@ def callback(self, clusters, prefix):
...
+            edims = set(retrieve_dimensions(exprs, deep=True))
+            if dim not in edims and any(d.dim in edims for d in prefix):
```
FabioLuporini, line 113: Actually, `dim` is in prefix, so the following should suffice:

```
if any(d.dim in edims for d in prefix):
```

In fact, the following might work as well:

```
if prefix.itdims & edims:
```

devito/passes/clusters/implicit.py
mloubout: For clusters with itspace `t, xi, yi` and `t, d, xi, yi` then this statement will be reached twice during the `self.process` once for `t` as a prefix and once for `t, d`

Issue #2235, opened 2023-10-13
default time_M value in presence of saved TimeFunction is wrong


FabioLuporini:
Reproducer

```
import numpy as np
from devito import Eq, Grid, TimeFunction, Operator

grid = Grid(shape=(4, 4))

u = TimeFunction(name='u', grid=grid)
usave = TimeFunction(name='usave', grid=grid, save=5)

eqns = [Eq(u.forward, u + 1),
        Eq(usave, u)]

op = Operator(eqns)

assert op.arguments()['time_M'] == 4

op.apply()

assert all(np.all(usave.data[i] == i) for i in range(5))
```

The issue stems from the fact that `op._dspace` sets [0,1] as the time dimension's data space bounds, and that "1" will cause the default value to become 3. Which is clearly wrong, since 4 is perfectly fine here.

This violates the principle of least surprise.

PR #2236, opened 2023-10-13
tests: Fix issue #2235

georgebisbas:
Opening for some first review


devito/ir/clusters/cluster.py
```diff
@@ -343,6 +343,21 @@ def dspace(self):
         intervals = intervals.promote(lambda d: not d.is_Sub)
         intervals = intervals.zero(set(intervals.dimensions) - oobs)
 
+        # Intersect with intervals from buffered dimensions. Unions of
```
mloubout, line 346: I'm not quite sure why would do this here, this really seems like a small tweak to `process_arguments`

PR #2237, opened 2023-10-13
compiler: Fix issue #2235

### LINKED ISSUE #2235: default time_M value in presence of saved TimeFunction is wrong

georgebisbas:


tests/test_buffering.py
```diff
@@ -752,3 +752,24 @@ def test_stencil_issue_1915_v2(subdomain):
     op1.apply(time_M=nt-2, u=u1)
 
     assert np.all(u.data == u1.data)
+
+
+def test_default_timeM():
```
FabioLuporini, line 757: this test doesn't have anything to do with the buffering pass

devito/ir/clusters/cluster.py
```diff
@@ -343,6 +343,21 @@ def dspace(self):
         intervals = intervals.promote(lambda d: not d.is_Sub)
         intervals = intervals.zero(set(intervals.dimensions) - oobs)
 
+        # Intersect with intervals from buffered dimensions. Unions of
+        # buffered dimension intervals may result in shrinking time size
```
FabioLuporini, line 347: "shrinking" feels odd here, since after all `time_M` will get a larger default

by shrinking here probably you mean that perhaps [0,1] becomes [0,0]

mloubout:
Also, potentially related, thos tests (and opertor)

https://github.com/devitocodes/devito/blob/b6f73081fd19ee4b5f0d7cd4ee0ff1cdbfa36135/tests/test_gpu_common.py#L563

Should be valid for `time_M=nt-1` but errors due to "OOB"

devito/ir/clusters/cluster.py
mloubout: The shrinking is introduced line 298, wouldn't it be cleaner to fix it there with slightly more checks

FabioLuporini: maybe, not sure, haven't thought about this

part of the exercise, I guess

devito/ir/clusters/cluster.py
```diff
@@ -332,6 +332,12 @@ def dspace(self):
         intervals = intervals.promote(lambda d: not d.is_Sub)
         intervals = intervals.zero(set(intervals.dimensions) - oobs)
 
+        # Buffered TimeDimensions should not shirnk their upper time offset
+        for f, v in parts.items():
```
FabioLuporini, line 336: this is better than before, but it's breaching through abstractions pretty heavily...

as you probably know, you shouldn't use the "time" inside the compiler, for obvious reasons

FabioLuporini: what's wrong with setting an interval to 0,0 if the dimension is something like `d.is_NonlinearDerived` OR `d.is_Stepping` or ... to be determined?

https://github.com/devitocodes/devito/pull/2237/files#diff-908f8895b205a02f261796e377b0b34ab9f3fe9998a5e152e2c8ec102e8ab3e7R305

FabioLuporini (changes requested):

devito/ir/clusters/cluster.py
georgebisbas: I have not looked into these properties TBH I am not sure if they apply to all cases
Defeinitly it is dangerous to zero the lower bound as well...
several checkpointing/gradient/buffering tests fail.
I will try to find some solution for this. Thanks for the ideas!

examples/userapi/02_apply.ipynb
```diff
@@ -449,14 +453,14 @@
...
-      "Operator `Kernel` run in 0.00 s\n"
+      "Operator `Kernel` ran in 0.01 s\n"
      ]
     },
     {
      "data": {
       "text/plain": [
-       "PerformanceSummary([('section0',\n",
-       "                     PerfEntry(time=3e-06, gflopss=0.021333333333333333, gpointss=0.010666666666666666, oi=0.16666666666666666, ops=2, itershapes=[(2, 4, 4)]))])"
+       "PerformanceSummary([(PerfKey(name='section0', rank=None),\n",
```
FabioLuporini, line 462: are we systematically displaying `rank`?

devito/ir/support/space.py
```diff
@@ -259,6 +259,11 @@ def negate(self):
     def zero(self):
         return Interval(self.dim, 0, 0, self.stamp)
 
+    def ceil(self, o):
```
FabioLuporini, line 262: I would turn the API into `ceil(self, v)` where `v` is just an integer. It doesn't make much sense to me to "ceil" based on an Interval

devito/ir/clusters/cluster.py
```diff
@@ -362,6 +362,13 @@ def dspace(self):
         intervals = intervals.promote(lambda d: not d.is_Sub)
         intervals = intervals.zero(set(intervals.dimensions) - oobs)
 
+        # Upper bound of intervals including dimensions classified for
+        # shifting should retain the "oobs" upper bound
+        for f, v in parts.items():
+            for i in v:
+                if i.dim in oobs:
+                    intervals = intervals.ceil(v[i.dim])
```
FabioLuporini, line 370: as per my comment below, probably this should be `i.upper` ?

devito/ir/clusters/cluster.py
```diff
@@ -362,6 +362,13 @@ def dspace(self):
         intervals = intervals.promote(lambda d: not d.is_Sub)
         intervals = intervals.zero(set(intervals.dimensions) - oobs)
 
+        # Upper bound of intervals including dimensions classified for
```
FabioLuporini, line 365: This comment doesn't sound too clear. I think shorter + tiny example might help? 

perhaps this is just me nitpicking...

devito/ir/clusters/cluster.py
```diff
@@ -362,6 +362,13 @@ def dspace(self):
...
+        # shifting should retain the "oobs" upper bound
+        for f, v in parts.items():
```
FabioLuporini, line 367:
```
for v in parts.values():
```

examples/userapi/02_apply.ipynb
```diff
@@ -246,14 +250,14 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "OOB detected due to time_M=2\n"
+      "OOB detected due to time_M=3\n"
```
FabioLuporini, line 253: I haven't looked at the example code yet, but is this sane? 

IOW -- was there an issue with the prior implementation?

tests/test_operator.py
```diff
@@ -2011,10 +2011,10 @@ def test_indirection(self):
         op = Operator(eqns)
 
         assert op._dspace[time].lower == 0
-        assert op._dspace[time].upper == 1
-        assert op.arguments()['time_M'] == nt - 2
+        assert op._dspace[time].upper == 0
+        assert op.arguments()['time_M'] == nt - 1
```
FabioLuporini, line 2015: at first glance, this looks wrong?

if `f` has `nt` slots, then it can be written up to `nt-2` included no? at `nt-2`, `f` will be written at `f[s, ...] = f[time+1, ...] = f[nt-2+1, ...] = f[nt-1, ...]` which is the extreme

So this looks borked

FabioLuporini (changes requested):

examples/userapi/02_apply.ipynb
georgebisbas: Yes

devito/ir/support/space.py
```diff
@@ -259,6 +259,9 @@ def negate(self):
     def zero(self):
         return Interval(self.dim, 0, 0, self.stamp)
 
+    def min_upper(self, v=0):
+        return Interval(self.dim, self.lower, v, self.stamp)
```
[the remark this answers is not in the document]
georgebisbas: better?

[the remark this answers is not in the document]
FabioLuporini: it looks like a jump in the class API, given that all other methos are more generic, see zero, flip, lift, etc

georgebisbas (commented):
@FabioLuporini just a gentle reminder for this

devito/ir/clusters/cluster.py
```diff
@@ -358,6 +358,18 @@ def dspace(self):
...
+        # not mapped to intervals e.g. issue #2235, resulting in reduced
+        # iteration size. Here, we relax this mapped upper interval, by
+        # intersecting intervals with matching only dimensions
+        for f, v in parts.items():
+            for i in v:
+                # oobs check is not required but helps reduce
+                # interval reconstruction
+                if i.dim in oobs and i.dim in f.dimensions:
+                    ii = intervals[i.dim].intersection(v[i.dim])
+                    intervals = intervals.set_upper(i.dim, ii.upper)
```
FabioLuporini, line 371: question. After this, is it true that `intervals[i.dim] == ii` ?

georgebisbas:
I tried to rebase this, but it failed on a gradient test. WIll check again

georgebisbas (commented):
So I know, it may be bit hard to remember where we have been here, but trying to refresh it.

georgebisbas:
I had honestly forgotten about this, please re-review when available

georgebisbas:
I am having a problem rebasing this due to:

```
    def test_function_coefficients_xderiv(self):
        p = Dimension('p')

        nstc = 8

        grid = Grid(shape=(51, 51, 51))
        x, y, z = grid.dimensions

        f = Function(name='f', grid=grid, space_order=(2*nstc, 0, 0),
                     coefficients='symbolic')
        g = Function(name='g', grid=grid, space_order=(2*nstc, 0, 0))
        ax = Function(name='DD2x', space_order=0, shape=(2*nstc + 1,),
                      dimensions=(p,))
        ay = Function(name='DD2y', space_order=0, shape=(2*nstc + 1,),
                      dimensions=(p,))
        stencil_coeffs_x_p1 = Coefficient(1, f, x, ax)
        stencil_coeffs_y_p1 = Coefficient(1, f, y, ay)
        stencil_coeffs = Substitutions(stencil_coeffs_x_p1, stencil_coeffs_y_p1)

        eqn = Eq(g, f.dxdy, coefficients=stencil_coeffs)

        op = Operator(eqn)
        op()
```

where domain should not be relaxed here...

devito/ir/support/space.py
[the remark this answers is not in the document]
EdCaunt: To add, should all of these properties be cached? They all build `Interval` objects, rather than simply returning a private variable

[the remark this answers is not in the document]
FabioLuporini: this one in particular isn't a property, it's a method

georgebisbas:
From @FabioLuporini 
> the solution seems particularly convoluted to me : https://github.com/devitocodes/devito/pull/2237/files#diff-908f8895b205a02f261796e377b0b34ab9f3fe9998a5e152e2c8ec102e8ab3e7R392
the comment is hard to decipher (or at least, I don’t understand it)
But most importantly, what’s really unclear to me is whether this is the right place where to inject such a fix. I’d have thought that a saved TimeFunction should have its own intervals “adjusted” in advance, so the union at line 390 returns the desired thing out of the box without any further processing. While here we do post-processing attempting to re-discovered information, so it’s a bit weird imho.
The introduction of set_upper is, imho, also debatable as we already have a translate that does a similar thing already (although not quite the same)

EdCaunt:
Also agree with @FabioLuporini that [this comment](https://github.com/devitocodes/devito/pull/2237/files#diff-908f8895b205a02f261796e377b0b34ab9f3fe9998a5e152e2c8ec102e8ab3e7R392) is quite cryptic. I also struggle to understand it

georgebisbas:
Now in:

https://github.com/devitocodes/devito/pull/2933

PR #2242, opened 2023-10-20
compiler: Fix unexpansion w custom coeffs

FabioLuporini:


devito/finite_differences/differentiable.py
```diff
@@ -637,6 +637,15 @@ def spacings(self):
 
     weights = Array.initvalue
 
+    def _xreplace(self, rule):
+        if self in rule:
+            return rule[self]
```
mloubout, line 642: , true

Same below

mloubout (commented):
Xreplace missing flag, except that gtg

devito/finite_differences/differentiable.py
FabioLuporini: thanks. Pushed fix

mloubout (approved):

PR #2248, opened 2023-10-26
compiler: Patch symbolic coefficients over cross derivatives

FabioLuporini:


tests/test_unexpansion.py
```diff
@@ -22,7 +22,7 @@ def test_backward_dt2(self):
         assert_structure(op, ['t,x,y'], 't,x,y')
 
 
-class TestSymbolicCoefficients(object):
+class TestSymbolicCoeffs(object):
```
EdCaunt, line 25: Think we should also test the case where different coefficients are passed for x and y derivatives with cross-derivatives, mixed derivatives, and non-perfect mixed derivatives. Also the case where only one of the x or y derivatives gets coefficients specified but bother are defined as symbolic.

tests/test_unexpansion.py
```diff
@@ -40,12 +40,22 @@ def test_fallback_to_default(self):
...
+        opt = ('advanced', {'expand': False})
+
+        # Pure derivative
+        Operator(Eq(u, u.dx2, coefficients=coeffs), opt=opt).cfunction
+
+        # Mixed derivative
+        Operator(Eq(u, u.dx.dx, coefficients=coeffs), opt=opt).cfunction
+
+        # Non-perfect mixed derivative
+        Operator(Eq(u, (u.dx + v.dx).dx, coefficients=coeffs), opt=opt).cfunction
```
EdCaunt, line 58: Need to sanity check these outputs. Print them and you will see that no errors are raised currently, but the resultant stencils are incorrect

EdCaunt (approved):
Looks good, should tide things over

PR #2257, opened 2023-11-03
types: Remove increments from SubDimension names

EdCaunt:
Requires a compiler pass to resolve clashing dimension names, which in turn requires additions to `uxreplace` so it can be used on `ConditionalDimension`s.

### DIFF
--- a/devito/ir/equations/algorithms.py
+++ b/devito/ir/equations/algorithms.py
@@ -5,6 +5,7 @@ from sympy import sympify
 
 from devito.symbolics import retrieve_indexed, uxreplace, retrieve_dimensions
 from devito.tools import Ordering, as_tuple, flatten, filter_sorted, filter_ordered
+from devito.ir.support import pull_dims
 from devito.types import Dimension, IgnoreDimSort
 from devito.types.basic import AbstractFunction
 
@@ -157,12 +158,18 @@ def separate_dimensions(expressions):
     resolutions = {}
     count = {}  # Keep track of increments on dim names
     processed = []
-    from devito.ir.support import pull_dims
     for e in expressions:
-        dims = dimension_sort(e)
-        dims = set(dims).union(*tuple(pull_dims(d.condition, flag=False)
-                                      for d in dims if d.is_Conditional
-                                      and d.condition is not None))
+        # Think dimension_sort is too eager?
+        # dims = set(dimension_sort(e))
+        # Dimensions in indices
+        dims = set().union(*tuple(set(i.function.dimensions)
+                                  for i in retrieve_indexed(e)))
+        # Dimensions in conditions and ConditionalDimension parents
+        dims = dims.union(*tuple(pull_dims(d.condition, flag=False)
+                                 for d in dims if d.is_Conditional
+                                 and d.condition is not None),
+                          set(d.parent for d in dims if d.is_Conditional))
+        # Sort for groupby
         dims = sorted(dims, key=lambda x: x.name)
 
         # Group dimensions by matching names
@@ -187,7 +194,6 @@ def separate_dimensions(expressions):
                         count[d.name] = 1
                     resolutions[d] = subs[d]
 
-        print(subs)
         # FIXME: ConditionalDimension parent not getting updated here
         processed.append(uxreplace(e, subs))
 
--- a/devito/ir/equations/algorithms.py
+++ b/devito/ir/equations/algorithms.py
@@ -157,25 +157,17 @@ def separate_dimensions(expressions):
     resolutions = {}
     count = {}  # Keep track of increments on dim names
     processed = []
+    from devito.ir.support import pull_dims
     for e in expressions:
+        dims = dimension_sort(e)
+        dims = set(dims).union(*tuple(pull_dims(d.condition, flag=False)
+                                      for d in dims if d.is_Conditional
+                                      and d.condition is not None))
+        dims = sorted(dims, key=lambda x: x.name)
+
         # Group dimensions by matching names
-        # More restrictive than dimension_sort()
-        # TODO: think it actually wants to be dimension_sort?
-
-        # Produces separate loop nests
-        # dims = sorted(set().union(*tuple(set(i.function.dimensions)
-        #                                  for i in retrieve_indexed(e))),
-        #               key=lambda x: x.name)
-
-        # Doesn't replace y with y0, weird loop order, weird Eq1 expression
-        dims = sorted(dimension_sort(e), key=lambda x: x.name)
-    
-        print("Dims:", dims)
-
         groups = tuple(tuple(g) for n, g in groupby(dims, key=lambda x: x.name))
         clashes = tuple(g for g in groups if len(g) > 1)
-
-        print("Clashes", clashes)
 
         subs = {}
         for c in clashes:
@@ -195,6 +187,8 @@ def separate_dimensions(expressions):
                         count[d.name] = 1
                     resolutions[d] = subs[d]
 
+        print(subs)
+        # FIXME: ConditionalDimension parent not getting updated here
         processed.append(uxreplace(e, subs))
 
     return processed
--- a/devito/ir/equations/algorithms.py
+++ b/devito/ir/equations/algorithms.py
@@ -156,14 +156,12 @@ def separate_dimensions(expressions):
     resolutions = {}
     count = {}  # Keep track of increments on dim names
     processed = []
+    from devito.ir.support import pull_dims
+    from itertools import groupby
     for e in expressions:
         # Group dimensions by matching names
-        name_map = {}
-        for d in e.dimensions:
-            try:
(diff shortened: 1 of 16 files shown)


devito/ir/equations/algorithms.py
```diff
@@ -147,3 +149,50 @@ def lower_exprs(expressions, **kwargs):
...
+    resolutions = {}
+    count = {}  # Keep track of increments on dim names
+    processed = []
+    for e in expressions:
+        # Just want dimensions which appear in the expression
+        # Dimensions in indices
+        dims = set().union(*tuple(set(i.function.dimensions)
+                                  for i in retrieve_indexed(e)))
+        # Dimensions in conditions and ConditionalDimension parents
+        dims = dims.union(*tuple(pull_dims(d.condition, flag=False)
```
FabioLuporini, line 167: you probably want to unclutter here and move most of this logic into `pull_dims`

ideally, all you need to do here would be:

```
dims = pull_dims(e, flag=...)

# Sort for groupby
dims = sorted(dims, key=lambda x: x.name)
```

because here you're performing a `search` which, in theory, is just one, atomic step of `separate_dimensions`

devito/symbolics/manipulation.py
```diff
@@ -172,21 +199,34 @@ class UxreplaceRegistry(list):
...
+            self.deep_cls.append(cls)
         self.append(cls)
         _uxreplace_handle.register(cls, _uxreplace_handle_reconstructable)
 
         for kls, callback in (rkwargs_callback_mapper or {}).items():
             _uxreplace_dispatch.register(kls, callback)
 
-    def dispatchable(self, obj):
-        return isinstance(obj, tuple(self))
+    def dispatchable(self, obj, deep=False):
```
FabioLuporini, line 215: So, I'm not sure I fully understand...

For `ConditionalDimension`:
* sometimes you want it `deep` 
* but some other times, you don't ?

is this correct? in particular, you want it deep **only** when you calol it inside `separate_dimensions`?

devito/ir/equations/algorithms.py
EdCaunt: IIrc, what's required here is more picky than `pull_dims` (pull dims selects more dimensions than are needed). It just wants dimensions in the indices, conditional parent, and conditional condition

devito/symbolics/manipulation.py
EdCaunt: Inside `separate_dimensions`, `uxreplace` wants to recurse into `ConditionalDimension`s and update all their args, kwargs, etc as necessary. However, elsewhere in the code, in particular [here](https://github.com/devitocodes/devito/blob/7c155e0f852afc15a493d1eacbc05cd706276b85/devito/ir/equations/algorithms.py#L143) when called from [here](https://github.com/devitocodes/devito/blob/7c155e0f852afc15a493d1eacbc05cd706276b85/devito/ir/equations/equation.py#L193), this behaviour may not be desired. In the above case, the `ConditionalDimension` may have already been indexified and shifted during the lowering of the expressions if it contains terms present in the `Eq` (in the case that `ConditionalDimension` is always affected by `uxreplace`), causing shifting to get applied twice. Hence, the need to be able to have some kind of control over the classes touched by `uxreplace` on a call-by-call basis

devito/types/dimension.py
```diff
@@ -200,6 +200,10 @@ def is_const(self):
     def root(self):
         return self
 
+    @property
+    def is_Root(self):
+        return self == self.root
```
FabioLuporini, line 205: is self.root

devito/symbolics/manipulation.py
```diff
@@ -180,13 +223,28 @@ def register(self, cls, rkwargs_callback_mapper=None):
...
+# Registry for default uxreplace
 _uxreplace_registry = UxreplaceRegistry()
 _uxreplace_registry.register(Eq)
 _uxreplace_registry.register(DefFunction)
 _uxreplace_registry.register(ComponentAccess)
+# Classes which only want uxreplacing when deep=True specified
+# _uxreplace_registry.register(ConditionalDimension, deep=True)
+
+# Registry for dimension uxreplace
+_dxreplace_registry = UxreplaceRegistry()
```
FabioLuporini, line 239: rename as `ReplaceRegistry` ?

devito/ir/equations/algorithms.py
```diff
@@ -147,3 +150,50 @@ def lower_exprs(expressions, **kwargs):
...
+                                 for d in dims if d.is_Conditional
+                                 and d.condition is not None),
+                          set(d.parent for d in dims if d.is_Conditional))
```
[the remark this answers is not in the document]
EdCaunt: The issue is that `retrieve_dimensions` collects too many things, not too few (even with flag=False). It wants to retrieve only indices in the equation, conditional dimension parent, and dimensions in the conditional dimension condition. Good point on the factor, that should probably be added in.

Fabio pointed out that `separate_equations` probably wants to be after creation of `LoweredEQ`, so may need to rethink this anyway.

devito/ir/equations/algorithms.py
```diff
@@ -147,3 +150,50 @@ def lower_exprs(expressions, **kwargs):
...
+        clashes = tuple(g for g in groups if len(g) > 1)
+
+        subs = {}
+        for c in clashes:
+            # Preferentially rename dims that aren't root dims
+            rdims = tuple(d for d in c if d.is_Root)
+            ddims = tuple(d for d in c if not d.is_Root)
+
+            for d in (ddims + rdims)[:-1]:
+                if d in resolutions.keys():
```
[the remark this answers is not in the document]
EdCaunt: I could always check that the incremented dimension name isn't in `dims` and incrementing until it is not?

devito/symbolics/manipulation.py
```diff
@@ -69,53 +80,85 @@ def _uxreplace(expr, rule):
...
+                args, aflag = _uxreplace_dispatch(v, rule, mode=mode)
+            else:
+                aflag = False
+
+            try:
+                # Get the underlying attribute originally used to build
+                # the object rather than the property, as the latter can
+                # sometimes have default values which do not match the
+                # default for the class
+                v = {i: getattr(expr, "_"+i) for i in expr.__rkwargs__}
```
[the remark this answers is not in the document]
EdCaunt: Not quite sure I understand?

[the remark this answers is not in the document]
mloubout: Some type have attribute such as `self.function` that are in `rkwargs` but `self._function` doesn't exist

devito/ir/equations/algorithms.py
```diff
@@ -147,3 +150,51 @@ def lower_exprs(expressions, **kwargs):
...
+        for c in clashes:
+            # Preferentially rename dims that aren't root dims
+            ddims = tuple(d for d in c if not d.is_Root)
+
+            for d in ddims:
+                if d in resolutions.keys():
+                    subs[d] = resolutions[d]
+                else:
+                    try:
+                        subs[d] = d._rebuild('%s%s' % (d.name, count[d.name]))
```
FabioLuporini, line 191: use the `sregistry` instead of `count`

You can plumb it down the compiler pass from the caller site -- it's one of the kwargs

devito/ir/equations/equation.py
```diff
@@ -220,8 +221,11 @@ def writes(self):
     def xreplace(self, rules):
         return LoweredEq(self.lhs.xreplace(rules), self.rhs.xreplace(rules), **self.state)
 
-    def func(self, *args):
-        return self._rebuild(*args, evaluate=False)
+    def func(self, *args, **kwargs):
```
FabioLuporini, line 224: since it's a Reconstructable, and in particular a Pickable, I think that the following should suffice:

```
func = Pickable._rebuild
```

devito/ir/support/space.py
```diff
@@ -1000,4 +1001,43 @@ def nonderived_directions(self):
...
+    lower, flag = _uxreplace_dispatch(expr.lower, rule, mode=mode)
+    changed |= flag
+    upper, flag = _uxreplace_dispatch(expr.upper, rule, mode=mode)
+    changed |= flag
+
+    return Interval(dim, lower=lower, upper=upper, stamp=expr.stamp), changed
+
+
+@_uxreplace_dispatch.register(IterationSpace)
+def _(expr, rule, mode='ux'):
```
FabioLuporini, line 1018: `expr` should be called `ispace`

devito/ir/equations/algorithms.py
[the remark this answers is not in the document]
FabioLuporini: also no need for .keys()

FabioLuporini (changes requested):

devito/ir/support/space.py
FabioLuporini: but most importantly, see comments on slack

PR #2258, opened 2023-11-03
api: Add shift and fd order option to all FD operators:

mloubout:
Add options so can do things like `div(u, order=so, shift=.5)` for left/right laplacian

### DIFF
--- a/devito/types/tensor.py
+++ b/devito/types/tensor.py
@@ -195,6 +195,14 @@ class TensorFunction(AbstractTensor):
     def div(self, shift=None, order=None):
         """
         Divergence of the TensorFunction (is a VectorFunction).
+
+        Parameters
+        ----------
+        shift: Number, optional, default=None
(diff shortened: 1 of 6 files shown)


devito/types/tensor.py
```diff
@@ -190,33 +192,44 @@ def values(self):
...
             comps.append(sum([getattr(self[j, i], 'd%s' % d.name)
-                              (x0=shift_x0(shift, d, i, j))
+                              (x0=shift_x0(shift, d, i, j), fd_order=order)
                               for j, d in enumerate(self.space_dimensions)]))
         return func._new(comps)
 
     @property
     def laplace(self):
+        """
+        Laplacian of the TensorFunction.
```
EdCaunt, line 213: Maybe add a note to the docstring about `laplace` vs `laplacian`?

mloubout: done

PR #2259, opened 2023-11-06
api: Always make subsampling factor symbolic

mloubout:
Fix issue with overwrties with different factor

### DIFF
--- a/devito/types/dense.py
+++ b/devito/types/dense.py
@@ -793,8 +793,8 @@ class DiscreteFunction(AbstractFunction, ArgProvider, Differentiable):
         args = ReducerMap({key.name: self._data_buffer})
 
         # Collect default dimension arguments from all indices
-        for i, s in zip(self.dimensions, self.shape):
-            args.update(i._arg_defaults(_min=0, size=s))
+        for a, i, s in zip(key.dimensions, self.dimensions, self.shape):
+            args.update(i._arg_defaults(_min=0, size=s, alias=a))
 
         return args
 
--- a/devito/types/dimension.py
+++ b/devito/types/dimension.py
@@ -11,6 +11,7 @@ from devito.logger import debug
 from devito.tools import Pickable, is_integer
 from devito.types.args import ArgProvider
 from devito.types.basic import Symbol, DataSymbol, Scalar
+from devito.types.constant import Constant
 
 __all__ = ['Dimension', 'SpaceDimension', 'TimeDimension', 'DefaultDimension',
            'CustomDimension', 'SteppingDimension', 'SubDimension', 'ConditionalDimension',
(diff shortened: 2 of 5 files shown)


devito/types/dense.py
```diff
@@ -793,8 +793,8 @@ def _arg_defaults(self, alias=None):
         args = ReducerMap({key.name: self._data_buffer})
 
         # Collect default dimension arguments from all indices
-        for i, s in zip(self.dimensions, self.shape):
-            args.update(i._arg_defaults(_min=0, size=s))
+        for a, i, s in zip(key.dimensions, self.dimensions, self.shape):
```
mloubout, line 796: @FabioLuporini quite surprised this never broke anything

FabioLuporini: I'm now confused. 

Shouldn't it be `key._data_buffer`?
Shouldn't it be `key.shape`?

Might just be early morning sleepy brain

devito/types/dimension.py
```diff
@@ -890,7 +900,7 @@ def _arg_defaults(self, _min=None, size=None, alias=None):
             return defaults
         try:
             # Is it a symbolic factor?
-            factor = defaults[dim._factor.name] = dim._factor.data
+            factor = defaults[dim._factor.name] = self._factor.data
```
FabioLuporini, line 903: why?

devito/types/dimension.py
```diff
@@ -848,13 +849,22 @@ def __init_finalize__(self, name, parent=None, factor=None, condition=None,
 
         super().__init_finalize__(name, parent)
 
-        self._factor = factor
+        # Always make the factor symbolic to allow overrides with different factor.
```
FabioLuporini, line 852: We totally sure we don't want to leave this to the user? 

Probably yes for the principle of least surprise right? the underlying issue is when you pass an override with a TimeFunction with a different factor -- correct?

A symbolic factor will make the generated code a bit more verbose (thinking about when it's used in tandem with lazy streaming), but yeah, it sounds safer...

devito/types/dense.py
mloubout: No:
- `key/alias` is the object and names used to build the operator, i.e the "key" of key-value pairs
- `self` is the runtime argument with the runtime shapes/values/data/....

So you want to return the {key: self} key-pair values

devito/types/dimension.py
mloubout: same as above, self is the runtime argument with the value for the operator, dim is the name of the dimension used to build the operator and argument name

devito/types/dimension.py
mloubout: I think it's safer to make it symbolic. It makes it a little it more verbose yes but at the price of avoiding bad surprises.

devito/types/dimension.py
```diff
@@ -848,13 +849,22 @@ def __init_finalize__(self, name, parent=None, factor=None, condition=None,
...
+        if factor is None:
+            self._factor = None
+        elif is_integer(factor):
+            self._factor = Constant(name="%sf" % name, value=factor, dtype=np.int32)
+        elif factor.is_Constant and is_integer(factor.data):
```
FabioLuporini, line 857: issubclass(factor, np.integer)

slightly cleaner

devito/types/dense.py
FabioLuporini: ah yes you're right of course

PR #2261, opened 2023-11-07
arch: support rocm for gpu info

### LINKED ISSUE #2260: get_gpu_info not working for AMD GPU
get_cpu_info() utility at least partially broken on AMD GPU returning `None` for available memory on the device

should implement a `romc-smi` mode similar to the `nvidia-smi`

- `rocm-smi --showmeminfo vram` for memory
- `rocm-smi --showproductname` for product/architecture

mloubout:
currently removes the `Xcompile` flag needed for host side as a duplicate flag

Fixes #2260

### DIFF
--- a/.github/workflows/pytest-core-nompi.yml
+++ b/.github/workflows/pytest-core-nompi.yml
@@ -80,7 +80,7 @@ jobs:
         - name: pytest-ubuntu-py39-gcc9-omp
           python-version: '3.9'
           os: ubuntu-20.04
-          arch: "gcc-9"
+          arch: "custom"
           language: "openmp"
           sympy: "1.9"
 
@@ -140,7 +140,7 @@ jobs:
       id: set-run
 
     - name: Install ${{ matrix.arch }} compiler
-      if: "runner.os == 'linux' && !contains(matrix.name, 'docker')"
+      if: "runner.os == 'linux' && !contains(matrix.name, 'docker') && matrix.arch !='custom' "
       run : |
         sudo apt-get install -y ${{ matrix.arch }}
 
@@ -166,8 +166,6 @@ jobs:
 
     - name: Test with pytest
       run: |
-        ${{ env.RUN_CMD }} ${{ matrix.arch }} --version
-        ${{ env.RUN_CMD }} python3 --version
         ${{ env.RUN_CMD }} pytest -k "${{ matrix.test-set }}" -m "not parallel" --cov --cov-config=.coveragerc --cov-report=xml ${{ env.TESTS }}
 
     - name: Upload coverage to Codecov
--- a/devito/arch/archinfo.py
+++ b/devito/arch/archinfo.py
@@ -10,6 +10,7 @@ import psutil
 import re
 import os
 import sys
+import json
 
 from devito.logger import warning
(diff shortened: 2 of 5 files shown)


georgebisbas (commented):
What was the mpi problem?

deckerla:
> What was the mpi problem?

Ill create an issue

devito/ir/stree/algorithms.py
```diff
@@ -150,7 +150,10 @@ def preprocess(clusters, options=None, **kwargs):
     for c in clusters:
         if c.is_halo_touch:
             hs = HaloScheme.union(e.rhs.halo_scheme for e in c.exprs)
-            queue.append(c.rebuild(halo_scheme=hs))
+            if hs.distributed_aindices:
```
FabioLuporini, line 153: After playing with the test a bit, I'm not sure this is the right fix.

The test _does_ need a halo exchange (`u` needs to be up-to-date before getting interpolated), but what this patch is essentially: "well, if the halo_scheme is empty, drop it and don't generate anything"

However, empty halo schemes should have been dropped earlier on, to avoid clusters-level pollution. In particular, here: https://github.com/devitocodes/devito/blob/master/devito/ir/clusters/algorithms.py#L395

So, it turns out that `is_void` gives False here because the `fmapper` is non-empty while the distributed-aindices are empty. And the latter is caused by the fact that no dimensions are used to index into `u` -- just plain symbols, AFAICT

So we might have to:
* fix this somewhere else
* make the test more robust, e.g. by placing the receivers at the MPI borders

About the fix: it might need to go somewhere here https://github.com/devitocodes/devito/blob/master/devito/mpi/halo_scheme.py#L470

by the looks of it, if `f.grid.is_distributed(d)` then:

* if there's an aindex, we good (most of the cases)
* if `i[d]` is a number, we should throw a user exception, since it's illegal MPI code (can't use pure numbers to index into distributed dimensions
* if `i[d]` isn't a number (e.g., like in our test, an expression of pure symbols), then we should use a dummy dimension placeholder, capturing the fact that such dimension does require a halo exchange

FabioLuporini (changes requested):

devito/ir/stree/algorithms.py
mloubout: It doesn't drop it it just keeps it where it is in the list of clusters. It's dropped somewhere else probably in the mpi pass but that's a different issue.

devito/ir/stree/algorithms.py
```diff
@@ -182,7 +189,33 @@ def preprocess(clusters, options=None, **kwargs):
...
+                aindex = max([itdims.index(d) for d in halo_scheme.distributed_aindices
+                              if d in itdims] + [0])
+                if dindex > aindex:
+                    hispace = c.ispace.project(itdims[:dindex])
+                else:
+                    hispace = None
+            else:
+                hispace = None
+
+            if hispace and options['mpi']:
```
FabioLuporini, line 212: whatever happens here, and regardless of how we end up doing it, we should always do it, irrespective of whether MPI is enabled or not. This will confer robustness to our compiler and, in particular, it will avoid "bugs to only pop up with MPI enabled"

FabioLuporini (approved):

PR #2270, opened 2023-11-20
deps: support python 3.12

mloubout:


devito/__init__.py
```diff
@@ -1,4 +1,5 @@
 from itertools import product
+from . import _version
```
mloubout, line 2: All below is autogenerated by versioneer, those files were outdated and not compatible with python3.12

PR #2271, opened 2023-11-21
compiler: Add missing numpy dtypes

mloubout:


devito/symbolics/extended_sympy.py
```diff
@@ -649,6 +659,8 @@ class CHARP(CastStar):
     (np.int8, '*'): CHARP,
     (np.uint8, '*'): CHARP,
     (int, '*'): INTP,  # noqa
+    (np.uint16, '*'): INTP,  # noqa
```
FabioLuporini, line 662: this should actually be USHORTP (to be added) ?

FabioLuporini (changes requested):

FabioLuporini (approved):

PR #2272, opened 2023-11-21
mpi: Fix haloupdate with inner dim [v2]

FabioLuporini:


devito/ir/clusters/algorithms.py
```diff
@@ -374,54 +374,53 @@ class Communications(Queue):
...
     B = Symbol(name='⊥')
 
-    @timed_pass(name='schedule')
+    @timed_pass(name='communications')
     def process(self, clusters):
         return self._process_fatd(clusters, 1, seen=set())
 
     def callback(self, clusters, prefix, seen=None):
-        if seen.issuperset(clusters):
+        if not prefix:
```
mloubout, line 382: Should this be both `not prefix or seen.issuperset(clusters)` ?

mloubout (approved):

devito/ir/clusters/algorithms.py
FabioLuporini: no the `seen` check is performed a few lines below on a per-cluster basis

PR #2280, opened 2023-12-11
misc: Switch off develop-mode

FabioLuporini:
This was a legacy thing. Pointless.

In fact, potentially harmful for data transfer performance since aligned based addresses will result in higher performance with most device backends

### DIFF
--- a/devito/__init__.py
+++ b/devito/__init__.py
@@ -122,9 +122,10 @@ configuration.add('autotuning', 'off', accepted, callback=autotune_callback,
                   impacts_jit=False)
 
 # In develop-mode:
-# - Some optimizations may not be applied to the generated code.
-# - The compiler performs more type and value checking
-configuration.add('develop-mode', True, [False, True])
+# - The ALLOC_GUARD data allocator is used. This will trigger segfaults as soon
+#   as an out-of-bounds memory access is performed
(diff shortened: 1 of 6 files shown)


devito/__init__.py
```diff
@@ -122,9 +122,10 @@ def autotune_callback(val):  # noqa
                   impacts_jit=False)
 
 # In develop-mode:
-# - Some optimizations may not be applied to the generated code.
-# - The compiler performs more type and value checking
-configuration.add('develop-mode', True, [False, True])
+# - The ALLOC_GUARD data allocator is used. This will trigger segfaults as soon
+#   as an out-of-bounds memory access is performed
+# - Some autoi-tuning optimizations are disabled
```
EdCaunt, line 127: Typo 'autoi'

tests/test_autotuner.py
```diff
@@ -7,7 +7,7 @@
 from devito.core.autotuning import options  # noqa
 
 
-@switchconfig(log_level='DEBUG')
+@switchconfig(log_level='DEBUG', develop_mode=True)
```
ggorman, line 10: It needs to be documented that setting log_level='DEBUG' also turns on develop_mode.

PR #2281, opened 2023-12-12
compiler: Introduce cluster-level Temp

georgebisbas:
A high-order TTI kernel resulted in requiring a mulsort between a "cluster-level" and a "devito" Temp, both having the same name and class name. This fix solves this issue. MFE test added.
Not sure, if we would like to have a TTI so16 kernel to compile in testing


tests/test_tools.py
```diff
@@ -144,3 +148,18 @@ def test_unbound_tuple():
     assert ub.next() == 2
     ub.iter()
     assert ub.next() == 1
+
+
+def test_temp_order():
```
FabioLuporini, line 153: This kinda belongs to test_dse

If not test_dse, definitely not here. The tools are those inside devito/tools

FabioLuporini (changes requested):

mloubout (approved):

FabioLuporini (approved):
GTG, thanks!

PR #2284, opened 2023-12-18
compiler: Fix handling of redundant derivatives

FabioLuporini:


devito/symbolics/inspection.py
```diff
@@ -229,13 +246,15 @@ def _(expr, estimate):
...
+@dont_count_if_seen
+def _(expr, estimate, seen):
+    return _estimate_cost(expr._evaluate(expand=False), estimate, seen)
 
 
 @_estimate_cost.register(IndexDerivative)
-def _(expr, estimate):
-    flops, _ = _estimate_cost(expr.expr, estimate)
+@dont_count_if_seen
+def _(expr, estimate, seen):
```
mloubout, line 256: Should this be also done for `EvalDerivative` to catch prematurely evaluated derivs?

mloubout (approved):

devito/symbolics/inspection.py
FabioLuporini: unnecessary
EvalDerivatives are Adds so they will be treated as such. EvalDerivatives are lower-level, more rudimentary objects, so it's OK to be treated as such

PR #2288, opened 2023-12-22
compiler: Improve IndexDerivative lowering

FabioLuporini:
Tiny enhancement to avoid useless stupid temporary variables in the generated code

### DIFF
--- a/devito/passes/clusters/derivatives.py
+++ b/devito/passes/clusters/derivatives.py
@@ -41,18 +41,36 @@ def _lower_index_derivatives(clusters, sregistry=None, **kwargs):
     for c in clusters:
         exprs = []
         for e in c.exprs:
-            expr, v = _core(e, c, weights, mapper, sregistry)
+            # Optimization 1: if the LHS is already a Symbol, then surely it's
+            # usable as a temporary for one of the IndexDerivatives inside `e`
+            if e.lhs.is_Symbol and e.operation is None:
+                reusable = {e.lhs}
+            else:
+                reusable = set()
+
+            expr, v = _core(e, c, weights, reusable, mapper, sregistry)
+
             if v:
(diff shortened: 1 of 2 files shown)


FabioLuporini:
"Fabio, have you thought about just performing a round of `cse()` instead"

"Yes I have, unfortunately it's not that easy because CSE is a Cluster-local pass, which doesn't work well here"

mloubout (approved):

georgebisbas:
> "Fabio, have you thought about just performing a round of `cse()` instead"
> 
> "Yes I have, unfortunately it's not that easy because CSE is a Cluster-local pass, which doesn't work well here"

Why it does not work well?

FabioLuporini:
> Why does it not work well? Not sure I understand TBH

`lower_index_derivative` might produce:

```
Cluster([
  r1 = 0
  r1 += <lowered index derivative>
  r0 = r1
])
```

CSE would be able to see that r0 is unnecessary here and safely removable. But `r0` is needed in later Clusters, so it would produce wrong code ultimately

Note that this is just a simplified example

You would need to understand `lower_index_derivatives` fully to appreciate all the different flavors of this issue

georgebisbas:
> > Why does it not work well? Not sure I understand TBH
> 
> `lower_index_derivative` might produce:
> 
> ```
> Cluster([
>   r1 = 0
>   r1 += <lowered index derivative>
>   r0 = r1
> ])
> ```
> 
> CSE would be able to see that r0 is unnecessary here and safely removable. But `r0` is needed in later Clusters, so it would produce wrong code ultimately
> 
> Note that this is just a simplified example
> 
> You would need to understand `lower_index_derivatives` fully to appreciate all the different flavors of this issue

Thanks

georgebisbas (approved):

PR #2294, opened 2024-01-09
example: small cleanup of tti for easier reuse

mloubout:
Minor cleanup for easier call from other places

### DIFF
--- a/devito/types/dimension.py
+++ b/devito/types/dimension.py
@@ -922,7 +922,7 @@ class ConditionalDimension(DerivedDimension):
         except AttributeError:
             factor = dim._factor
 
-        defaults[dim.parent.max_name] = range(1, factor*size - 1)
+        defaults[dim.parent.max_name] = range(0, factor*size - 1)
 
         return defaults
 
--- a/tests/test_dse.py
+++ b/tests/test_dse.py
@@ -2768,7 +2768,7 @@ class TestTTI(object):
         # TTI layered model for the tti test, no need for a smooth interace
         # bewtween the two layer as the compilation passes are tested, not the
         # physical prettiness of the result -- which ultimately saves time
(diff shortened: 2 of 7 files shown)


devito/types/dimension.py
```diff
@@ -922,7 +922,7 @@ def _arg_defaults(self, _min=None, size=None, alias=None):
         except AttributeError:
             factor = dim._factor
 
-        defaults[dim.parent.max_name] = range(1, factor*size - 1)
+        defaults[dim.parent.max_name] = range(0, factor*size - 1)
```
FabioLuporini, line 925: nitpicking: you can omit the `0, ` part

why was there a 1 before?
is it worth adding a test?

tests/test_dse.py
```diff
@@ -2874,7 +2874,7 @@ def test_opcounts(self, space_order, expected):
 
     @switchconfig(profiling='advanced')
     @pytest.mark.parametrize('space_order,exp_ops,exp_arrays', [
-        (4, 122, 6), (8, 221, 7)
+        (4, 122, 6), (8, 235, 7)
```
FabioLuporini, line 2877: why the increase? variable density on by default? that feels weird

devito/types/dimension.py
mloubout: I don't think the 1 was there for any specific reason, just a "run at least one step" but not really needed

tests/test_dse.py
mloubout: No it's because of the space order I think. THe model used to be setup with space order `4` while setting up the second case with space order `8` so not sure if it was creating a few low order derivatives.
All the other opcount case stay the same the FD hasn't changed (b is never setup in oss for tti)  it's just the space order change

EdCaunt (commented):
Nothing to add that Fabio hasn't already mentioned

PR #2296, opened 2024-01-16
compiler: Revamp data alignment

### LINKED ISSUE #1579: Drop deprecation of `padding` parameter
`configuration['autopadding'] = True` makes devito applying padding to the innermost dimension such that the size is a multiple of the SIMD vector length

There are other reasons one may want to pad, e.g. to minimize cache conflicts. And clearly not all of them can be supported in Devito.

So we should drop the deprecation of `padding`: https://github.com/devitocodes/devito/blob/master/devito/types/dense.py#L929

FabioLuporini:
Sorry about the branch name -- it's misleading to an extent

Note that one test is being dropped because I've removed a micro-optimization pass that was just legacy, rarely (if ever) used, and marginally (if any) impactful, all while being a massive PITA to maintain (basically everything revolving around the so called "ROUNDABLE" Property)

fixes #1579


devito/types/dense.py
```diff
@@ -1108,46 +1103,44 @@ def __halo_setup__(self, **kwargs):
...
+
             elif isinstance(space_order, tuple) and len(space_order) == 3:
-                _, left_points, right_points = space_order
-                halo = (left_points, right_points)
+                _, l, r = space_order
+                halo = [(l, r) if i.is_Space else (0, 0) for i in self.dimensions]
+
+            elif isinstance(space_order, tuple) and len(space_order) == 2:
+                _, space_halo = space_order
+                if not isinstance(space_halo, tuple) or \
```
mloubout, line 1115: can't be just an `int` to be same left/right?

mloubout (commented):
Some minor comments but looks good.

devito/arch/archinfo.py
```diff
@@ -649,9 +647,23 @@ def memavail(self, *args, **kwargs):
         """Available physical memory in bytes, or None if unknown."""
         return None
 
+    def max_mem_trans_size(self, dtype):
+        """
+        Number of items of type `dtype` that can be transferred in a single
+        memory transaction.
+        """
+        assert self.max_mem_trans_nbytes % np.dtype(dtype).itemsize == 0
```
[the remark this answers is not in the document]
FabioLuporini: @EdCaunt 
No, because we developers write these classes and upon testing a new one we would be thrown the assert exception right away if the case

@mloubout yes, == 0, the memory transaction nbytes is always a multiple of 2, just like a C base type size

devito/data/allocators.py
```diff
@@ -69,25 +69,41 @@ def alloc(self, shape, dtype):
...
-            raise RuntimeError("Unable to allocate %d elements in memory", str(size))
-
-        # cast to 1D array of the specified size
-        ctype_1d = ctype * size
+        # Add padding, if any
+        try:
+            padleft, padright = padding
+        except TypeError:
+            padleft, padright = padding, padding
+        if not isinstance(padleft, int) and not isinstance(padright, int):
```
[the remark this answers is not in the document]
FabioLuporini: ok fixing

devito/data/allocators.py
```diff
@@ -69,25 +69,41 @@ def alloc(self, shape, dtype):
...
+        array = np.frombuffer(buf, dtype=dtype)
+
+        # `array.reshape` should not be used here because it may introduce
+        # a copy. From `docs.scipy.org/doc/numpy/reference/generated/numpy.reshape`:
+        #   It is not always possible to change the shape of an array without
+        #   copying the data. If you want an error to be raised when the data
+        #   is copied, you should assign the new shape to the shape attribute
+        #   of the array:
+        array.shape = shape
+        ndarray = array  # At this point it's interpreted as an ndarray
```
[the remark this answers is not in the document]
FabioLuporini: nah, dropping

devito/operator/operator.py
```diff
@@ -624,7 +624,8 @@ def _prepare_arguments(self, autotune=None, **kwargs):
 
         # Sanity check
         for p in self.parameters:
-            p._arg_check(args, self._dspace[p], am=self._access_modes.get(p))
+            p._arg_check(args, self._dspace[p], am=self._access_modes.get(p),
+                         **kwargs)
```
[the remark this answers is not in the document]
FabioLuporini: Gonna be needed in PRO soon.

I also had a change in OSS while developing this branch, but then I dropped it. I left the **kwargs because I will need it in an upcoming branch.

If you think about it, it makes a lot of sense -- there's a lot of symbolic properties we might want to check they're honored before signing off on the override; properties we aren't checking today

devito/types/array.py
```diff
@@ -127,42 +125,23 @@ def __init_finalize__(self, *args, **kwargs):
...
-            return tuple(padding)
+            padding = ((0, 0),)*self.ndim
+        elif isinstance(padding, DimensionTuple):
+            padding = tuple(padding[d] for d in self.dimensions)
         elif isinstance(padding, int):
-            return tuple((0, padding) for _ in range(self.ndim))
+            padding = tuple((0, padding) for _ in range(self.ndim))
         elif isinstance(padding, tuple) and len(padding) == self.ndim:
-            return tuple((0, i) if isinstance(i, int) else i for i in padding)
+            padding = tuple((0, i) if isinstance(i, int) else i for i in padding)
```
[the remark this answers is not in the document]
FabioLuporini: that's all we need for data alignment, I'm yet to find a use case for left-padding, I doubt it exists but 🤷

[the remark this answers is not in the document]
FabioLuporini: OK switching to is_integer

devito/types/dense.py
FabioLuporini: unfortunately not, think about shifted derivatives and you may have Functions requiring 4 points on the left and 3 on the right

devito/types/dense.py
```diff
@@ -1108,46 +1103,44 @@ def __halo_setup__(self, **kwargs):
...
+                   len(space_halo) != len(self.space_dimensions):
+                    raise TypeError("Invalid `space_order`")
```
georgebisbas, line 1117: You could give some hint here on how the user can fix this

tests/test_data.py
```diff
@@ -316,9 +316,9 @@ def test_w_halo_w_autopadding(self):
         assert u0.shape_allocated == (4, 4, 16)
 
         assert u1._size_halo == ((3, 3), (3, 3), (3, 3))
-        assert u1._size_padding == ((0, 0), (0, 0), (0, 14))  # 14 stems from 6 + 8
-        assert u1._size_nodomain == ((3, 3), (3, 3), (3, 17))
-        assert u1.shape_allocated == (10, 10, 24)
+        assert u1._size_padding == ((0, 0), (0, 0), (0, 6))  # 14 stems from 6 + 8
```
georgebisbas, line 319: Fix comment

devito/types/dense.py
georgebisbas: You could possibly also add a minor test

tests/test_data.py
FabioLuporini: good catch, fixing

devito/types/dense.py
FabioLuporini: Giving a hint is a mess, far too many cases. However, it's a fair point. I'm extending the docstring to explain clearly what the possibilities are.

And yes, I also agree about a small test. Adding it...

tests/test_data.py
```diff
@@ -316,9 +316,21 @@ def test_w_halo_w_autopadding(self):
...
-        assert u1._size_nodomain == ((3, 3), (3, 3), (3, 17))
-        assert u1.shape_allocated == (10, 10, 24)
+        assert u1._size_padding == ((0, 0), (0, 0), (0, 6))  # 6 stems from 16-(3+4+3)
+        assert u1._size_nodomain == ((3, 3), (3, 3), (3, 9))
+        assert u1.shape_allocated == (10, 10, 16)
+
+    def test_w_halo_custom(self):
+        grid = Grid(shape=(4, 4))
+
+        # Custom halo with not enougn entries raises an exception
```
georgebisbas, line 326: small typo "enough"

mloubout (approved):

PR #2297, opened 2024-01-16
arch: patch compiler version

mloubout:


devito/arch/compiler.py
```diff
@@ -237,6 +228,21 @@ def __new_with__(self, **kwargs):
...
+        if self.suffix is not None:
+            try:
+                version = Version(str(float(self.suffix)))
+            except (TypeError, ValueError):
+                version = Version(self.suffix)
+        else:
+            try:
+                # Knowing the version may still be useful to pick supported flags
+                version = sniff_compiler_version(self.CC)
+            except (FileNotFoundError, OSError):
```
FabioLuporini, line 242: in case of FileNotFoundError a `sys.exit()` will be called inside `sniff_compiler_version` and the program will abort, so we might have to tweak that too

FabioLuporini: or we just remove this try-except and wait for the next use case to break

mloubout: Changed to something bit simpler

FabioLuporini (approved):
OK, thanks, merging

PR #2308, opened 2024-02-11
Correct the Poisson equation in the cavity flow example

rafael-fuente:
There is a sign mistake in the Poisson equation. It doesn't significantly affect the solution of this particular example, but it can be important when solving Navier Stokes for other problems.


mloubout (approved):

mloubout:
Good catch, thanks!

PR #2309, opened 2024-02-12
compiler: Block reductions irrespective of par-tile

FabioLuporini:
Test is in PRO

I also took the chance to create opt-options to allow specifying different par-tiles for reduction loops

### DIFF
--- a/devito/passes/clusters/blocking.py
+++ b/devito/passes/clusters/blocking.py
@@ -7,8 +7,8 @@ from devito.ir.support import (AFFINE, PARALLEL, PARALLEL_IF_ATOMIC, PARALLEL_IF
                                IntervalGroup, IterationSpace, Scope)
 from devito.passes import is_on_device
 from devito.symbolics import search, uxreplace, xreplace_indices
-from devito.tools import (UnboundedMultiTuple, as_tuple, filter_ordered,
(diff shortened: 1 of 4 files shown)


devito/passes/clusters/blocking.py
```diff
@@ -447,14 +449,42 @@ def __init__(self, par_tile):
...
         x = any(i.dim.is_Block for i in flatten(c.ispace for c in clusters))
         if not x:
             self.tip += 1
 
-        # TODO: This is for now exceptionally rudimentary
+        # Correctness -- enforce blocking where necessary.
+        # See also issue #276:PRO
+        if any(c.properties.is_parallel_atomic(d) for c in clusters):
+            if any(c.is_sparse for c in clusters):
+                if not x:
```
mloubout, line 479: Don't think you don't need this, it's just a tuple not a multi-tuple, same below

mloubout (commented):
Small comment, gtg

devito/passes/clusters/blocking.py
mloubout: Wait no you do need it to go back to first element if there is multiple sparse/reduce cluster nevermind

FabioLuporini: correct

PR #2311, opened 2024-02-13
compiler: Optimize normalize_reductions_dense

FabioLuporini:


devito/ir/clusters/algorithms.py
```diff
@@ -499,20 +503,25 @@ def normalize_reductions_dense(cluster, sregistry, options):
...
             # This makes it much easier to parallelize the map part regardless
             # of the target backend
+            lhs, rhs = e.args
 
-            if e.lhs.function.is_Array:
+            if lhs.function.is_Array:
                 # Probably a compiler-generated reduction, e.g. via
                 # recursive compilation; it's an Array already, so nothing to do
                 processed.append(e)
+            elif rhs in mapper:
```
mloubout, line 512: Should we do this for sparse too? the rhs of the atomic update for elastic is computed trice for example

FabioLuporini: the issue is not the redundant expressions (if there were, it's up to CSE to catch them, not here), but the fact that we were introducing redundant arrays (which means memory overhead). In the sparse case, instead, we only create scalars

mloubout (approved):

PR #2319, opened 2024-02-22
examples: Update MPI notebook to reference inner and outer halo terminology

EdCaunt:


examples/mpi/overview.ipynb
georgebisbas, line 1: I only see some deleted text? Is that the purpose?

FabioLuporini: agree, looks like a flaky commit

EdCaunt: Oh weird. Not sure what happened there. Will fix

PR #2322, opened 2024-02-27
compiler: Add Weights.value utility

FabioLuporini:


devito/finite_differences/differentiable.py
```diff
@@ -732,6 +732,24 @@ def _xreplace(self, rule):
                 pass
             return super()._xreplace(rule)
 
+    @cached_property
+    def _npweights(self):
+        # NOTE: `self.weights` cannot just be an array or SymPy will fail
```
georgebisbas, line 737: Break into two sentences? Reads a bit weird

EdCaunt (approved):

PR #2329, opened 2024-03-11
dsl: Introduce ability to limit injection and interpolation to a SubDomain

EdCaunt:


mloubout (changes requested):
As mentioned in Houston, I don't think a subdomain is a valid input for injection/interp.

An expression `sparse.inject(function)` contains everything it needs to know, namely the `function.grid/subdomain` and the `sparse.coordinates`. Adding a subdomain as input is just likely to create more issues than anything. 

The only thing that needs to be changed to be compatible with function on subdomain is to have `interpolator._gdims` to pull the dimensions from the function rather than the global dim

PR #2331, opened 2024-03-13
api: add priority to fd coefficients for mixed derivatives

mloubout:
So can sort fd coeffs types such as with drp


FabioLuporini (approved):
uncontroversial

PR #2333, opened 2024-03-13
compiler: Enable generation of templated function declarations

FabioLuporini:


FabioLuporini:
a test is added is PRO (the test you've seen would fail to compile without the changes here)

but I can maybe add a simple IET test too

FabioLuporini:
@mloubout tests added

mloubout (approved):

PR #2340, opened 2024-03-27
Ensure compiler check catches permission errors

### LINKED ISSUE #2336: Bad error message when missing compiler
https://github.com/devitocodes/devito/blob/fe47e96197253de924cd338e8569504442f22e56/devito/arch/compiler.py#L38-L50

The reason for catching the FileNotFoundError is likely to print out a nice error message if the compiler command can't be found. But the actual error that is thrown when someone tries to run a nonexistent executable is a permissions error, which is not being caught.

The documentation for subprocess indicates OSError will be thrown in case of missing program. (Search "OSError" on https://docs.python.org/3/library/subprocess.html)

If this makes sense, I can make a pull request to fix it next week.

gbruer15:
I added a test for #2336, but I'm not sure if it will actually error on the bug I was encountering, so I'd like to run the standard test suite to make sure.

Closes #2336

### DIFF
--- a/tests/test_error_checking.py
+++ b/tests/test_error_checking.py
@@ -1,8 +1,9 @@
 import pytest
+import re
 
 from devito import Grid, Function, TimeFunction, Eq, Operator, switchconfig
 from devito.exceptions import ExecutionError
-
+from devito.arch.compiler import sniff_compiler_version
 
 @switchconfig(safe_math=True)
 @pytest.mark.parametrize("expr", [
@@ -25,3 +26,11 @@ def test_stability(expr):
 
     with pytest.raises(ExecutionError):
         op.apply(time_M=200, dt=.1)
+
+@pytest.mark.parametrize("cc", [
+    "doesn'texist",
+    "/root/doesn'texist",
+])
+def test_sniff_compiler_version(cc):
+    with pytest.raises(SystemExit) as e:
+        sniff_compiler_version(cc)


tests/test_error_checking.py
```diff
@@ -25,3 +26,11 @@ def test_stability(expr):
...
     with pytest.raises(ExecutionError):
         op.apply(time_M=200, dt=.1)
+
+@pytest.mark.parametrize("cc", [
+    "doesn'texist",
+    "/root/doesn'texist",
+])
+def test_sniff_compiler_version(cc):
+    with pytest.raises(SystemExit) as e:
+        sniff_compiler_version(cc)
```
mloubout, line 36: I think it needs  to be

```
with pytest.raises(SystemExit) as e:
        sniff_compiler_version(cc)
 assert e == SystemExit
```

so there is a test.

mloubout (commented):
Thanks!.

You can the test locally to make sure it does what you want with 

`py.test -vsx tests/test_error_checking.py -k test_sniff_compiler_version`

(`-vsx` is just for verbosity)

tests/test_error_checking.py
```diff
@@ -1,8 +1,9 @@
 import pytest
+import re
```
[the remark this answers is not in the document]
gbruer15: Good catch, thanks. I was gonna check the error message to make sure it mentioned the compiler, but then I found it was too much effor to get the stderr.

gbruer15:
I tested locally and it's fine. The issue is I'm not sure how I should consistently trigger the `PermissionDenied` error to ensure it's being caught.

I could make a file without execute permissions and delete it afterwards. I'm not sure if that will be portable for the OS's that Devito supports, though.

devito/mpi/distributed.py
gbruer15, line 1: @mloubout While trying to run the tests, I encountered an error with `MPI.Init` being `None`. In line with this PR, I updated this file to print out the real error message, which should be more informative.

mloubout: Do you have an MFE for this?  he main goal of the content of except is to avoid raising an error an run in non-mpi mode, that's why .

This looks like this now will raise a delayed error making harder to figure out.

mloubout: - (2) yes, looks like it's missing the tag, see e.g [this test](https://github.com/devitocodes/devito/blob/39ea32a23fd0506b05d09293d01d6fac983a52b5/tests/test_operator.py#L1210) we might wanna wrap the flag into `parallel` directly

- (1) We probably wanna elaborate more on it yes. We have setup the project so that only minimal dependency are installed for usage case `pip install -e .[cases]` e.g `pip install -e .[tests,mpi]` to install the base + all tests and mpi dependencies. Currenly we only briefly mention in in the doc (https://www.devitoproject.org/download.html#conda-route) but could use some details and refreshing

We'll check about the error throwing.

gbruer15: Should I save these changes for a separate MR then? Or is it ok here?

mloubout: (2) should be fixed with

https://github.com/devitocodes/devito/pull/2342/commits/e9b2c29a5782e47ed54baf5ede38832aeedd61ef

devito/mpi/distributed.py
```diff
@@ -42,7 +42,10 @@ def cleanup():
         if init_by_devito and MPI.Is_initialized() and not MPI.Is_finalized():
             MPI.Finalize()
     atexit.register(cleanup)
-except ImportError:
```
mloubout, line 45: (@FabioLuporini) maybe instead of carrying that import error around we could just add

```
@classmethod
def Init(cls):
   raise e
```

to the dummy fallback below so that it only triggers if it tries to init MPI.

FabioLuporini: fine be me

tests/test_error_checking.py
FabioLuporini: Also: this test should be moved to some other files as this one, `test_error_checking`, is supposed to be for runtime (C-level) numerical and computational errors.

Perhaps `test_environment`?

FabioLuporini: `sys.exit(...)` is also quite brutal, perhaps we turn it into a RuntimeError?

gbruer15: `test_environment.py` seems to be making sure the test environment is usable, so this test doesn't fit there in my opinion. This test is independent of the test environment. Should there be a second test file for non-numerical runtime errors?

I agree that RuntimeError is better. I'll make that change.

devito/mpi/distributed.py
gbruer15: Note: due to scoping rules, I saved the exception in the class variable.

mloubout (approved):
Looks good to me. Not sure where else to put that test though but not critical IMO

tests/test_error_checking.py
FabioLuporini: > Should there be a second test file for non-numerical runtime errors?

perhaps `test_arch.py` where in the future we can also add all other `/devito/arch` related tests?

FabioLuporini (changes requested):
We just ened to find a better place for the test, `test_error_checking` is meant for something else, and perhaps we should rename it

FabioLuporini:
there are a few pep8 issues to be resolved

tests/test_error_checking.py
gbruer15: I moved the test to `test_arch.py` and fixed the pep8 errors, so as far as I know, this is ready to be merged.

mloubout: There is benchmark setup that needs fixing and rebase, I'll do it and get it merged

PR #2342, opened 2024-04-01
api: Add Hicks (sinc) interpolation api

### LINKED ISSUE #595: A utility for interpolation and resampling
Over time the work on resampling input data in time and the spatial interpolation of off-grid points seem to have a lot of overlap and this could be cleanly set out in a set of utility functions/classes. I was looking at some existing implementations that we should try and use wherever possible, with @ggorman a few weeks ago, and I am now putting these up in an issue to enable discussion to make this more concrete. 

I understand that [scipy.signal.firwin](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.firwin.html#scipy.signal.firwin) can be used to precompute the coefficients required to use the `PrecomputedSparseFunction` class in devito. This can also be used to resample incoming data in time. I also believe that [scipy.signal.lfilter](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.lfilter.html#scipy.signal.lfilter) is (functionally) scipy's counterpart of `PrecomputedSparseFunction` - although only in a single dimension. 

Scipy also provides full resampling implementations in the form of [scipy.signal.resample](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.resample.html#scipy.signal.resample) and [scipy.signal.resample_poly](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.resample_poly.html#scipy.signal.resample_poly). IIRC, `resample` only works for integer multiples because it uses FFT. I'm not sure why we can't use `resample_poly`, though. 

Scipy also provides direct functions for computing Kaiser-function [1] based interpolation like [scipy.signal.kaiserord](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.kaiserord.html#scipy.signal.kaiserord) and [scipy.signal.kaiser_beta](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.kaiser_beta.html#scipy.signal.kaiser_beta). 

I didn't find any ready implementations of Lanczos-filter based interpolation in scipy. However, I did find an [implementation](https://docs.obspy.org/_modules/obspy/signal/interpolation.html) in [ObsPy](http://www.seismosoc.org/Publications/SRL/SRL_81/srl_81-3_es/).

References:
[1] Hicks, Graham J. "Arbitrary source and receiver positioning in finite-difference schemes using Kaiser windowed sinc functions." Geophysics 67, no. 1 (2002): 156-165.

mloubout:
Add support for kaiser-windowed sinc interpolation

Fixes #595

### DIFF
--- a/.github/workflows/pytest-core-mpi.yml
+++ b/.github/workflows/pytest-core-mpi.yml
@@ -48,6 +48,11 @@ jobs:
         python3 scripts/clear_devito_cache.py
         python3 -m pytest --cov --cov-config=.coveragerc --cov-report=xml -m parallel tests/
 
+    - name: Test examples
+      run: |
+        python3 scripts/clear_devito_cache.py
+        python3 -m pytest --cov --cov-config=.coveragerc --cov-report=xml examples/seismic
+
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v4
       with:
@@ -79,3 +84,7 @@ jobs:
       - name: Test with pytest
         run: |
           docker run --rm -e CODECOV_TOKEN=${{ secrets.CODECOV_TOKEN }} -e OMP_NUM_THREADS=1 --name testrun devito_img pytest tests/test_mpi.py
+
+      - name: Test examples with MPI
+        run: |
+          docker run --rm --env DEVITO_MPI=1 --name examplerun mpiexec -n 2 pytest examples/seismic
--- a/.github/workflows/pytest-core-mpi.yml
+++ b/.github/workflows/pytest-core-mpi.yml
@@ -70,9 +70,12 @@ jobs:
             - name: gcc
               arch: gcc
               os: ubuntu-latest
+              mpiflag: ""
             - name: icx
               arch: icx
               os: ubuntu-latest
+              # Need safe math for icx due to inaccuracy with mpi+sinc interpolation
+              mpiflag: "-e DEVITO_SAFE_MATH=1"
       
       steps:
       - name: Checkout devito
@@ -85,3 +88,8 @@ jobs:
       - name: Test with pytest
         run: |
           docker run --rm -e CODECOV_TOKEN=${{ secrets.CODECOV_TOKEN }} -e OMP_NUM_THREADS=1 --name testrun devito_img pytest tests/test_mpi.py
+
+      - name: Test examples with MPI
+        run: |
+          docker run --rm ${{ matrix.mpiflag }} -e DEVITO_MPI=1 -e OMP_NUM_THREADS=1 --name examplerun devito_img mpiexec -n 2 pytest examples/seismic/acoustic
+          docker run --rm -e DEVITO_MPI=1 -e OMP_NUM_THREADS=1 --name examplerun devito_img mpiexec -n 2 pytest examples/seismic/tti
--- a/.github/workflows/pytest-core-mpi.yml
+++ b/.github/workflows/pytest-core-mpi.yml
@@ -51,7 +51,8 @@ jobs:
     - name: Test examples
       run: |
         python3 scripts/clear_devito_cache.py
-        python3 -m pytest --cov --cov-config=.coveragerc --cov-report=xml examples/seismic
+        DEVITO_MPI=1 mpirun -n 2 python3 -m pytest --cov --cov-config=.coveragerc --cov-report=xml examples/seismic/acoustic
+        DEVITO_MPI=1 mpirun -n 2 python3 -m pytest --cov --cov-config=.coveragerc --cov-report=xml examples/seismic/tti
 
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v4
@@ -87,4 +88,5 @@ jobs:
 
       - name: Test examples with MPI
         run: |
-          docker run --rm --env DEVITO_MPI=1 --name examplerun mpiexec -n 2 pytest examples/seismic
+          docker run --rm -e DEVITO_MPI=1 -e OMP_NUM_THREADS=1 --name examplerun devito_img mpiexec -n 2 pytest examples/seismic/acoustic
+          docker run --rm -e DEVITO_MPI=1 -e OMP_NUM_THREADS=1 --name examplerun devito_img mpiexec -n 2 pytest examples/seismic/tti
--- a/.github/workflows/pytest-core-mpi.yml
+++ b/.github/workflows/pytest-core-mpi.yml
@@ -85,8 +85,3 @@ jobs:
       - name: Test with pytest
         run: |
           docker run --rm -e CODECOV_TOKEN=${{ secrets.CODECOV_TOKEN }} -e OMP_NUM_THREADS=1 --name testrun devito_img pytest tests/test_mpi.py
-
-      - name: Test examples with MPI
-        run: |
-          docker run --rm -e DEVITO_MPI=1 -e OMP_NUM_THREADS=1 --name examplerun devito_img mpiexec -n 2 pytest examples/seismic/acoustic
-          docker run --rm -e DEVITO_MPI=1 -e OMP_NUM_THREADS=1 --name examplerun devito_img mpiexec -n 2 pytest examples/seismic/tti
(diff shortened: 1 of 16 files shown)


devito/operations/interpolators.py
```diff
@@ -421,3 +445,62 @@ def _weights(self):
...
+        return coeffs
+
+    @property
+    def _weights(self):
+        return Mul(*[w._subs(rd, rd-rd.parent.symbolic_min)
+                     for (rd, w) in self.interpolation_coeffs.items()])
+
+    def _arg_defaults(self, coords=None, sfunc=None):
+        args = {}
+        b = self._b_table[self.r]
```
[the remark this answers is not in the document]
mloubout: It will just throw a key-error but can add a check (I doubt anyone will run it this high that's a 20x20x20 cube in 3d super expensive)

devito/operations/interpolators.py
```diff
@@ -223,6 +235,14 @@ def _interp_idx(self, variables, implicit_dims=None):
                     for ((k, c), p) in zip(mapper.items(), pos)})
                     for v in variables}
 
+        # Position only replacement, not radius dependent.
+        # E.g src.inject(vp(x)*src) needs to use vp[posx] at all points
+        # not vp[posx + rx]
+        if pos_only is not None:
```
FabioLuporini, line 241: if you turn pos_only into `pos_only=()` (instead of None) you don't need this `if` I think?

devito/operations/interpolators.py
```diff
@@ -421,3 +445,62 @@ def _weights(self):
...
+    Arbitrary source and receiver positioning in finite‐difference schemes
+    using Kaiser windowed sinc functions
+
+    https://library.seg.org/doi/10.1190/1.1451454
+
+    """
+
+    _name = 'sinc'
+
+    # Table 1
```
FabioLuporini, line 463: What's Table 2? u mean Table 1 in the paper?

devito/operations/interpolators.py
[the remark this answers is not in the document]
FabioLuporini: agree to add a check. Users can be surprising...

devito/operations/interpolators.py
```diff
@@ -421,3 +445,62 @@ def _weights(self):
...
+        args = {}
+        b = self._b_table[self.r]
+        b0 = np.i0(b)
+        if coords is None or sfunc is None:
+            raise ValueError("No coordinates or sparse function provided")
+        # Coords to indices
+        coords = coords / np.array(sfunc.grid.spacing)
+        coords = coords - np.floor(coords)
+
+        # Precompute sinc
```
FabioLuporini, line 497: I think this wants a `NOTE: <comment about performance>` ? if you measured it's slowish? or is it super quick since the loops are very short?

devito/types/sparse.py
```diff
@@ -799,16 +804,23 @@ class SparseFunction(AbstractSparseFunction):
...
     def __init_finalize__(self, *args, **kwargs):
+        # Interpolation method
+        interp = kwargs.pop('interpolation', 'linear')
+        self.interpolation = interp
+        self.interpolator = _interpolators[interp](self)
+        self._radius = kwargs.pop('r', _default_radius[interp])
+        if interp == 'sinc' and self._radius < 2:
+            raise ValueError("The 'sinc' interpolator requires a radius of at least 2")
+
+        # Set space ordert to `r` for safety
```
FabioLuporini, line 822: order

FabioLuporini: but not sure why this comment ended up here

examples/seismic/acoustic/acoustic_example.py
```diff
@@ -60,21 +60,26 @@ def run(shape=(50, 50, 50), spacing=(20.0, 20.0, 20.0), tn=1000.0,
     solver.jacobian(dm, autotune=autotune)
     info("Applying Gradient")
     solver.jacobian_adjoint(rec, u, autotune=autotune, checkpointing=checkpointing)
-    return summary.gflopss, summary.oi, summary.timings, [rec, u.data]
+    return summary.gflopss, summary.oi, summary.timings, [rec, u.data._local]
```
FabioLuporini, line 63: this will pick the (dirty) padding as well I think? why not just data or data_with_halo?

examples/seismic/acoustic/operators.py
```diff
@@ -15,13 +14,21 @@ def freesurface(model, eq):
...
+    lhs, rhs = eq.args
     # Get vertical dimension and corresponding subdimension
-    zfs = model.grid.subdomains['fsdomain'].dimensions[-1]
+    fsdomain = model.grid.subdomains['fsdomain']
+    zfs = fsdomain.dimensions[-1]
     z = zfs.parent
 
-    # Functions present in the stencil
-    funcs = retrieve_functions(rhs)
+    # Retrieve vertical derivatives
```
FabioLuporini, line 23: what's all these new additions?

examples/seismic/acoustic/operators.py
```diff
@@ -15,13 +14,21 @@ def freesurface(model, eq):
...
+    Dz = {d for d in retrieve_derivatives(rhs) if z in d.dims}
```
FabioLuporini, line 24: dzs ?

devito/operations/interpolators.py
```diff
@@ -133,6 +136,13 @@ def inject(self, *args, **kwargs):
     def interpolate(self, *args, **kwargs):
         pass
 
+    @property
+    def name(self):
+        return self._name
+
+    def _arg_defaults(self, **args):
```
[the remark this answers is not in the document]
mloubout: No, the interpolator isn't an arg provide, it's sparwse function is and is calling it only.

devito/operations/interpolators.py
mloubout: Yes Table 1 in the paper

devito/operations/interpolators.py
mloubout: I haven't seen any drastic slow compute since it's numpified so not sure want to force user to be like "hey here is a warning should I nitpick here"

examples/seismic/acoustic/acoustic_example.py
mloubout: for MPI debug the _local is a lot more useful

examples/seismic/acoustic/operators.py
mloubout: it does the same, except it only evaluates `z` derivatives instead of everything since the rest isn't needed which is much cheaper especially for things like tti.

devito/operations/interpolators.py
FabioLuporini: no, the convention is to use immutable objects only, and tuple in that sense is fine ; dict/list/set aren't instead obv

for example, I'm using `=()` in ir/iet/nodes.py

alternatively, you may use `as_tuple`

.github/workflows/pytest-core-mpi.yml
```diff
@@ -78,4 +84,9 @@ jobs:
       
       - name: Test with pytest
         run: |
-          docker run --rm -e CODECOV_TOKEN=${{ secrets.CODECOV_TOKEN }} -e OMP_NUM_THREADS=1 --name testrun devito_img pytest tests/test_mpi.py
\ No newline at end of file
+          docker run --rm -e CODECOV_TOKEN=${{ secrets.CODECOV_TOKEN }} -e OMP_NUM_THREADS=1 --name testrun devito_img pytest tests/test_mpi.py
+
+      - name: Test examples with MPI
+        run: |
+          docker run --rm -e DEVITO_SAFE_MATH=1 -e DEVITO_MPI=1 -e OMP_NUM_THREADS=1 --name examplerun devito_img mpiexec -n 2 pytest examples/seismic/acoustic
```
FabioLuporini, line 91: comment why safe-math  needed

devito/operations/interpolators.py
```diff
@@ -347,10 +373,9 @@ def _inject(self, field, expr, implicit_dims=None):
         # summing temp that wouldn't allow collapsing
         implicit_dims = implicit_dims + tuple(r.parent for r in self._rdim)
 
-        variables = variables + list(fields)
-
         # List of indirection indices for all adjacent grid points
-        idx_subs, temps = self._interp_idx(variables, implicit_dims=implicit_dims)
+        idx_subs, temps = self._interp_idx(list(fields), implicit_dims=implicit_dims,
```
FabioLuporini, line 377: why does it need to be casted to a list? it was a tuple already, so it feels pretty odd

PR #2344, opened 2024-04-04
compiler: Support for C-level MPI_Allreduce

FabioLuporini:


devito/builtins/arithmetic.py
```diff
@@ -129,15 +129,15 @@ def sumall(f):
...
     dtype = accumulator_mapper[f.dtype]
+    n = make_retval(f.grid, dtype)
     s = dv.types.Symbol(name='sum', dtype=dtype)
 
-    with MPIReduction(f, dtype=dtype) as mr:
-        op = dv.Operator([dv.Eq(s, 0.0)] + eqns +
-                         [dv.Inc(s, p), dv.Eq(mr.n[0], s)],
-                         name='sum')
-        op.apply(**kwargs)
+    op = dv.Operator([dv.Eq(s, 0.0)] + eqns +
```
mloubout, line 135: nitpicking: can probalby wrap (since everywhere in this file) the `s=0, eqs, s+=expr, n[0]=s` into an `make_reduction(eqns, expr)`

devito/mpi/reduction_scheme.py
```diff
@@ -0,0 +1,48 @@
+import sympy
+
+from devito.tools import Reconstructable
+
+__all__ = ['DistReduce']
+
+
+class DistReduce(sympy.Function, Reconstructable):
```
mloubout, line 8: why do you need `sympy.Function` ?

devito/mpi/reduction_scheme.py
```diff
@@ -0,0 +1,48 @@
...
+
+    """
+    A SymPy object representing a distributed Reduction.
+    """
+
+    __rargs__ = ('var',)
+    __rkwargs__ = ('op', 'grid', 'ispace')
+
+    def __new__(cls, var, op=None, grid=None, ispace=None, **kwargs):
+        obj = sympy.Function.__new__(cls, var, **kwargs)
```
mloubout, line 18: `op.__new__` ?

devito/mpi/routines.py
```diff
@@ -1351,3 +1352,38 @@ def _arg_values(self, args=None, **kwargs):
...
+class ReductionBuilder(object):
+
+    """
+    Build IET routines performing MPI reductions.
+    """
+
+    mapper = {
+        OpInc: 'MPI_SUM',
+        OpMax: 'MPI_MAX',
+        OpMin: 'MPI_MIN',
```
mloubout, line 1372: missing `OpProd: MPI_PROD`

devito/types/sparse.py
```diff
@@ -128,21 +124,25 @@ def __subfunc_setup__(self, key, suffix, dtype=None):
...
-            # Need to rebuild so the dimensions match the parent SparseFunction
-            indices = (self.indices[self._sparse_position], *key.indices[1:])
-            return key._rebuild(*indices, name=name, shape=shape,
-                                alias=self.alias, halo=None)
-        elif key is not None and not isinstance(key, Iterable):
-            raise ValueError("`%s` must be either SubFunction "
-                             "or iterable (e.g., list, np.ndarray)" % key)
+            if d in key.dimensions and not self.alias:
+                # From a reconstruction which leaves `dimensions` intact
+                return key
```
mloubout, line 131: For future reference, this is very problematic (both if and else case) because now `key`'s parent does not match the sparsefunction that uses it as a subfunction so can lead to very problematic runtime args/post_process so we may need to completely avoid this rebuild and always create a new subfunc using the key's data (need a '_local' initalizer

mloubout (commented):
Overallm the reduction part looks good, but we need to think carefully about the sparse subfunc because the current changes I think just make our setup worse and harder to fix later. 
Can discuss if we wanna proceed but tag an "urgent" issue on it

devito/builtins/utils.py
```diff
@@ -1,52 +1,26 @@
...
-        else:
-            comm = self.grid.distributor.comm
-            self.v = comm.allreduce(np.asarray(self.n.data), self.op)[0]
+    if grid is None:
+        raise ValueError("Expected Grid, got None")
+
+    i = dv.Dimension(name='mri',)
+    n = dv.Function(name='n', shape=(1,), dimensions=(i,), grid=grid,
+                    dtype=dtype, space='host')
+    n.data[:] = 0
```
EdCaunt, line 22: Shouldn't this initialise to zero already? Also do you need to set the space order of n to 0 so that it doesn't assign a halo?

tests/test_builtins.py
```diff
@@ -453,7 +453,7 @@ def test_sum_sparse(self):
 
     def test_min_max_sparse(self):
         """
-        Test that mmin/mmax work on SparseFunction
+        Test that mmin/mmax work on SparseFunction.
```
EdCaunt, line 456: "SparseFunctions"

devito/builtins/utils.py
FabioLuporini: yeah I think the `n.data[:] = 0` is legacy and can go

halo isn't an issue

devito/mpi/reduction_scheme.py
FabioLuporini: no `op` is a devito internal type

FabioLuporini: (yes, we should start adding types...)

mloubout (approved):

devito/builtins/utils.py
georgebisbas: will this be dropped?

devito/ir/clusters/algorithms.py
```diff
@@ -432,6 +443,57 @@ def callback(self, clusters, prefix, seen=None):
...
+            # The IterationSpace within which the global distributed reduction
+            # must be carried out
+            ispace = c.ispace.prefix(lambda d: d in var.free_symbols)
+
+            fifo.append(DistReduce(var, op=op, grid=grid, ispace=ispace))
+
+        processed.append(c)
+
+    # Leftover reductions are placed at the very end
+    if fifo:
```
georgebisbas, line 490: An example of what goes here?

devito/operator/operator.py
```diff
@@ -271,9 +271,9 @@ def _lower(cls, expressions, **kwargs):
         return IRs(expressions, clusters, stree, uiet, iet), byproduct
 
     @classmethod
-    def _rcompile_wrapper(cls, **kwargs0):
-        def wrapper(expressions, **kwargs1):
-            return rcompile(expressions, {**kwargs0, **kwargs1})
+    def _rcompile_wrapper(cls, **kwargs):
```
georgebisbas, line 274: This kwargs0 was recently changed as well right, now back to it, ok

devito/ir/clusters/algorithms.py
FabioLuporini: wdym ? it's just code for draining `fifo` should it still contains elements once we exit the main loop, so it will contains reductions

georgebisbas (approved):

devito/builtins/arithmetic.py
FabioLuporini: will do in my next PR

tests/test_builtins.py
FabioLuporini: will fix in my next PR

PR #2348, opened 2024-04-08
compiler: Tweak device-aware blocking

FabioLuporini:


tests/test_dle.py
```diff
@@ -216,6 +217,26 @@ def test_cache_blocking_structure_optrelax_customdim():
                      'd,x0_blk0,y0_blk0,z0_blk0,x,y,z')
 
 
+def test_cache_blocking_structure_optrelax_defaultdim_alone():
```
EdCaunt, line 220: This is a horrible test name. How about `test_blocking_defaultdim_blockrelax`? Failing that, give it a shorter name and add a docstring explaining what it does.

tests/test_dle.py
```diff
@@ -216,6 +217,26 @@ def test_cache_blocking_structure_optrelax_customdim():
...
+
+    u = TimeFunction(name="u", grid=grid)
+    f = Function(name="f", grid=grid, dimensions=(d, x, y, z),
+                 shape=(2,) + grid.shape)
+
+    eqn = Inc(f, u*cos(time*d))
+
+    op = Operator(eqn, opt=('advanced', {'blockrelax': 'device-aware'}))
+
+    _, _ = assert_blocking(op, {'d0_blk0', 'x0_blk0'})
```
EdCaunt, line 234: Why do you need to create dummy variables here? Could you not just have `assert_blocking(op, {'d0_blk0', 'x0_blk0'})` on its own?

FabioLuporini: copy-paste leftover

tests/test_dle.py
FabioLuporini: don't check the commit time because I had to rebase to address your other comment below. So you've to do with faith this time

EdCaunt (approved):

georgebisbas (approved):

PR #2350, opened 2024-04-14
ci: tweak mpi setup to allow -s and hide output

mloubout:
Should fix @georgebisbas issue with too verbose logging for him
SHould now allow -s for embed

### DIFF
--- a/tests/test_linearize.py
+++ b/tests/test_linearize.py
@@ -152,7 +152,7 @@ def test_interpolation_msf():
     assert op1.cfunction
 
 
-@pytest.mark.parallel(mode=[(1, 'diag2')])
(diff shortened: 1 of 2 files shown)


georgebisbas (commented):
I get the sam error locally as the one appearing in CI

EdCaunt (approved):
Looks good to me. Rip `return "bonjour"`

tests/test_linearize.py
```diff
@@ -152,7 +152,7 @@ def test_interpolation_msf():
     assert op1.cfunction
 
 
-@pytest.mark.parallel(mode=[(1, 'diag2')])
+@pytest.mark.parallel(mode=[(2, 'diag2')])
```
FabioLuporini, line 155: I think 1 was fine here

PR #2357, opened 2024-04-17
compiler: Fix ConditionalDimensions on SubDomains

EdCaunt:
Tweak dimension extraction from conditions so that ConditionalDimensions can have SubDimension parents.

Now you can do
```
class Middle(SubDomain):
    name = 'middle'
    def define(self, dimensions):
        return {x: x, y: ('middle', 2, 4)}
mid = Middle() 
my_grid = Grid(shape = shape, subdomains = (mid, ))

_, yi = mid.dimensions
condition = Lt(sdf, 1).subs(y, yi)
ci = ConditionalDimension(name='ci', condition=condition, parent=yi)
```
and the `ConditionalDimension` will be nested inside the `SubDimension` in the generated loops.

### DIFF
--- a/devito/ir/clusters/algorithms.py
+++ b/devito/ir/clusters/algorithms.py
@@ -238,7 +238,7 @@ def guard(clusters):
                 if cd._factor is not None:
                     k = d
                 else:
-                    dims = pull_dims(cd.condition)
+                    dims = pull_dims(cd.condition, flag=False)  # Don't want parent dims
                     k = max(dims, default=d, key=lambda i: c.ispace.index(i))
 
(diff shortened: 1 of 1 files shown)


EdCaunt:
> So you still needs the subs eventhough the parent is `yi` ?
> 
> `condition = Lt(sdf.subs(y, yi), 1)` would maybe make more sense.
> 
> Is there a test/example for it?

Yeah, you still need the `.subs`. A test needs to be added, it would probably also be a good idea to add an example

EdCaunt:
Closed as replicates a fix already in #2050

PR #2359, opened 2024-04-22
reqs: Update cached_property to functools version

EdCaunt:
`cached_property` has been included in `functools` since Python 3.8. The original `cached_property` module hasn't had an update since. Since Python 3.7 is officially end-of-life, support can be dropped and we can move to using `functools.cached_property` throughout the codebase. This PR updates the requirements and all imports accordingly.


devito/data/decomposition.py
```diff
@@ -1,7 +1,7 @@
 from collections.abc import Iterable
 
 import numpy as np
-from cached_property import cached_property
+from functools import cached_property
```
mloubout, line 4: This is now a system import so convention throughout is to group it with system import (first group with collection)

.github/workflows/pytest-core-nompi.yml
```diff
@@ -32,11 +32,11 @@ jobs:
         name: [
            pytest-ubuntu-py311-gcc11-noomp,
            pytest-ubuntu-py38-gcc12-omp,
-           pytest-ubuntu-py37-gcc7-omp,
+           pytest-ubuntu-py38-gcc7-omp,
```
mloubout, line 35: Ok, I'll fix the branch rules once approved so it shows as green

devito/types/dense.py
```diff
@@ -243,7 +244,20 @@ def shape_with_halo(self):
         """
         return tuple(j + i + k for i, (j, k) in zip(self.shape, self._size_outhalo))
 
-    _shape_with_outhalo = shape_with_halo
+    @cached_property
+    def _shape_with_outhalo(self):
```
mloubout, line 248: We gotta find a better way, duplication like that are not good.

EdCaunt: Yeah, need to figure that one out. Wanted to get it working first though

EdCaunt: Now improved somewhat

mloubout: Well no, the docstring is not really the issue, the content of the function is which is twice the same code.

.github/workflows/pytest-core-nompi.yml
```diff
@@ -84,8 +88,13 @@ jobs:
           language: "openmp"
           sympy: "1.9"
 
+<<<<<<< HEAD
```
mloubout, line 91: rebase leftover

mloubout (approved):

FabioLuporini (approved):

PR #2366, opened 2024-04-30
dsl: Patch edge-case derivative specifications

EdCaunt:
Derivatives specified as `Derivative(f, (x, 2), (y, 0))` (zeroth-order wrt one dimension) are a valid request, but the resultant derivatives were not equal to those specified as `Derivative(f, (x, 2))` and certain routines such as substitution would fail. This PR patches this and adds a test.


FabioLuporini (approved):
uncontroversial

mloubout (approved):

PR #2368, opened 2024-05-03
api: cleanup FD tools and support zeroth order derivative

mloubout:


tests/test_derivatives.py
```diff
@@ -639,13 +639,22 @@ def test_zero_spec(self):
...
+        assert drv0.fd_order == 2
+        assert drv1.fd_order == (2, 2)
+        assert drv0.deriv_order == 2
+        assert drv1.deriv_order == (2, 0)
+
+        assert drv0.evaluate == drv1.evaluate
+        fmean = .5 * (f + f._subs(y, y + y.spacing))
+        drv1x0 = drv1(x0={y: y+y.spacing/2}).evaluate
+
+        assert simplify(fmean.dx2.evaluate - drv1x0) == 0
```
EdCaunt, line 657: It would also be a good idea to check equivalence between the `Derivative(f, y, deriv_order=0, x0=y+y.spacing/2)` and `Derivative(f, (y, 0), x0={y: y+y.spacing/2})` specifications.

There should also be tests for `Derivative(f, (x, 0), (y, 0), x0={x: x+x.spacing/2, y: y+y.spacing/2})` and `Derivative(f, x, y,, deriv_order=(0, 0), x0={x: x+x.spacing/2, y: y+y.spacing/2})`

jkwashbourne-oss:
Great work! What about a unit test interpolating an analytic function like a sin or cos?

tests/test_derivatives.py
mloubout: Added

mloubout:
>What about a unit test interpolating an analytic function like a sin or cos?

There is a unit test checking coefficient correctness.

tests/test_caching.py
```diff
@@ -831,7 +831,7 @@ def test_solve(self, operate_on_empty_cache):
         # created by the finite difference (u.dt, u.dx2). We would have had
         # three extra references to u(t + dt), u(x - h_x) and u(x + h_x).
         # But this is not the case anymore!
-        assert len(_SymbolCache) == 11
+        assert len(_SymbolCache) == 12
```
mloubout, line 834: Note: This is because now every order uses a StencilDImension even for `space_order=1` while we were only giving the list of indices for that case before.

devito/finite_differences/differentiable.py
```diff
@@ -876,6 +877,9 @@ def __new__(cls, *args, base=None, **kwargs):
 
     func = DifferentiableOp._rebuild
 
+    __eq__ = sympy.Add.__eq__
```
FabioLuporini, line 880: what's the issue without?

mloubout: replied on slack

devito/types/basic.py
```diff
@@ -1071,43 +1071,49 @@ def _eval_deriv(self):
         return self
 
     @property
-    def _is_on_grid(self):
+    def _grid_map(self):
         """
         Check whether the object is on the grid and requires averaging.
```
FabioLuporini, line 1076: requires docstring update

devito/finite_differences/differentiable.py
FabioLuporini: ok, but I could use a comment here

mloubout: done

devito/types/basic.py
mloubout: done

PR #2369, opened 2024-05-07
api: Always expand time derivatives

mloubout:


devito/finite_differences/finite_difference.py
```diff
@@ -230,6 +230,9 @@ def generic_derivative(expr, dim, fd_order, deriv_order, matvec=direct, x0=None,
 
 def make_derivative(expr, dim, fd_order, deriv_order, side, matvec, x0, coefficients,
                     expand):
+    # Always expand time derivatives
```
FabioLuporini, line 233: Can you add why? also there's a couple of leftover "I need to expand comments" from prior PR

mloubout: I already did the do string changes in previous PR

mloubout: added longer comment

PR #2374, opened 2024-05-23
Geom setup ex

mloubout:


devito/types/sparse.py
```diff
@@ -62,8 +66,24 @@ def __init_finalize__(self, *args, **kwargs):
...
     def __indices_setup__(cls, *args, **kwargs):
         dimensions = as_tuple(kwargs.get('dimensions'))
+        # Need this not to break MatrixSparseFunction
+        try:
+            _sub_funcs = tuple(cls._sub_functions)
+        except TypeError:
+            _sub_funcs = ()
+        # If a subfunction provided use the sparse dimension
+        for f in _sub_funcs:
+            if f in kwargs:
```
FabioLuporini, line 76: just:

```
try:
    sparse_dim = ...
except (AttributeError, KeyError):
    pass
```

FabioLuporini (commented):
this is such an outstanding cleanup... !

I've left one tiny coding comment, other than that we good

devito/types/sparse.py
```diff
@@ -62,8 +66,24 @@ def __init_finalize__(self, *args, **kwargs):
...
+                try:
+                    sparse_dim = kwargs[f].indices[0]
+                    break
+                except AttributeError:
+                    continue
+        else:
```
EdCaunt, line 82: Wrong indent level for this `else`?

devito/types/sparse.py
```diff
@@ -62,8 +66,24 @@ def __init_finalize__(self, *args, **kwargs):
...
+            sparse_dim = Dimension(name='p_%s' % kwargs["name"])
```
EdCaunt, line 83: This could be made the default value for `dimensions = as_tuple(kwargs.get('dimensions'))` which would shave off some lines by removing the need for the `if not dimensions` block. It would also allow the `if args` block to be moved to the top of the function, improving readability

devito/types/sparse.py
```diff
@@ -120,6 +140,22 @@ def __subfunc_setup__(self, key, suffix, dtype=None):
         if key is not None and not isinstance(key, SubFunction):
             key = np.array(key)
 
+        # Check if already a SubFunction
+        d = self.indices[self._sparse_position]
```
EdCaunt, line 144: Move definition of `d` inside the `if isinstance` block? It only gets used there right?

devito/types/sparse.py
```diff
@@ -120,6 +140,22 @@ def __subfunc_setup__(self, key, suffix, dtype=None):
...
+        if isinstance(key, SubFunction):
+            if d in key.indices:
+                # Can use as is, dimension already matches
+                if self.alias:
```
EdCaunt, line 148:
```
if self.alias:
    return key._rebuild(alias=self.alias, name=name)
return key
```

devito/types/sparse.py
```diff
@@ -120,6 +140,22 @@ def __subfunc_setup__(self, key, suffix, dtype=None):
...
+                    return key._rebuild(alias=self.alias, name=name)
+                else:
+                    return key
+            else:
```
EdCaunt, line 152: This `else` can be removed and the contents of the block unindented by a level.

devito/types/sparse.py
mloubout: No the `else` goes there with the `for`. It's a python construct "if for didn't exit then " that avoids wrapping the for in a try/if

devito/types/sparse.py
mloubout: done

devito/types/sparse.py
mloubout: tweaked

devito/types/sparse.py
mloubout: done

devito/types/sparse.py
mloubout: I like to keep the else for simple blocks and unindent when it's a decent code chunck that is the "main" part

devito/types/sparse.py
mloubout: same

devito/types/sparse.py
EdCaunt: Huh, TIL

FabioLuporini: yes and it's quite handy

devito/operations/interpolators.py
```diff
@@ -486,8 +486,7 @@ def interpolation_coeffs(self):
             sf = SubFunction(name="wsinc%s" % r.name, dtype=self.sfunction.dtype,
                              shape=shape, dimensions=dimensions,
                              space_order=0, alias=self.sfunction.alias,
-                             distributor=self.sfunction._distributor,
-                             parent=self.sfunction)
+                             parent=None)
```
FabioLuporini, line 489: nitpicking: we can probably save a line here by compacting these four lines into three, but obviously this is for another day

FabioLuporini (approved):

EdCaunt (approved):

PR #2388, opened 2024-06-18
API: fix printer dtype processing

mloubout:


devito/passes/iet/linearization.py
```diff
@@ -82,8 +82,9 @@ def key1(f, d):
     """
     if f.is_regular:
         # For paddable objects the following holds:
-        # `same dim + same halo => same (auto-)padding`
-        return (d, f._size_halo[d], f.is_autopaddable)
+        # `same dim + same halo + same dtype => same (auto-)padding`
+        print(f, f.dtype)
```
FabioLuporini, line 86: leftover

PR #2392, opened 2024-06-28
compiler: Patch CSE in presence of conditionals

FabioLuporini:


georgebisbas (approved):

devito/passes/clusters/cse.py
```diff
@@ -98,24 +98,26 @@ def _cse(maybe_exprs, make, min_cost=1, mode='default'):
...
-                pe, changed = _uxreplace(pe, {t.rhs: t.lhs})
-                if changed and t not in updated:
-                    updated.append(t)
+            for k, v in chosen:
+                pe, changed = _uxreplace(pe, {k: v})
+                if changed and v not in scheduled:
+                    updated.append(pe.func(v, k, operation=None))
+                    scheduled.append(v)
             updated.append(pe)
         processed = updated
```
EdCaunt, line 116: Why not just use `updated` henceforth, rather than using `processed`?

tests/test_dse.py
```diff
@@ -168,6 +168,29 @@ def test_cse_temp_order():
     assert type(args[2]) is CTemp
 
 
+def test_cse_w_conditionals():
+    grid = Grid(shape=(10, 10, 10))
+    x, _, _ = grid.dimensions
```
EdCaunt, line 173: Nitpick: `x = grid.dimensions[0]` would be tidier imo

FabioLuporini: tidier but (nitpicking) the one above performs an implicit assertion

devito/passes/clusters/cse.py
FabioLuporini: wdym ? we keep on looping and `processed` needs to be updated? what change are you proposing ?

EdCaunt: Ignore me then

EdCaunt: Missed an indent

EdCaunt (approved):
Comments responded, looks good

mloubout (approved):

mloubout (approved):

PR #2393, opened 2024-06-28
compiler: Drop SubDomainSet earlier during compilation for more graceful lowering

EdCaunt:


devito/types/grid.py
```diff
@@ -781,27 +784,27 @@ def __subdomain_finalize__(self, grid, counter=0, **kwargs):
...
+                idx = 2*i + j
+                if isinstance(self._local_bounds[idx], int):
+                    f[j].data[:] = np.full((self._n_domains,),
+                                           self._local_bounds[idx],
+                                           dtype=np.int32)
+                else:
+                    f[j].data[:] = self._local_bounds[idx]
+
+            d._implicit_dimension = i_dim
+            d._functions = f
```
mloubout, line 807: modifying `d` isn't great should rebuild it or create a new one

FabioLuporini: this is within the lazy constructor, so it's kinda acceptable

devito/passes/clusters/implicit.py
```diff
@@ -94,8 +91,12 @@ def callback(self, clusters, prefix):
                 processed.append(c)
                 continue
 
+            msdims = [msdim(i.dim) for i in c.ispace[idx:]]
```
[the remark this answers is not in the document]
EdCaunt: I suppose the mapper is unordered, so can use a set here. Good spot

devito/types/grid.py
EdCaunt: Was able to shuffle things around, remove the modification, and reduce the number of lines

mloubout (approved):

devito/types/grid.py
```diff
@@ -774,34 +763,38 @@ def __subdomain_finalize__(self, grid, counter=0, **kwargs):
...
+            dname = '%si%d' % (d.name, counter)
+            fnames = ["%s_%s_%s" % (self.name, dname, mM)
+                      for mM in ('m', 'M')]
+            f = tuple(Function(name=fname, grid=self._grid, shape=(self._n_domains,),
+                               dimensions=(i_dim,), dtype=np.int32) for fname in fnames)
 
-            functions.append(f)
-        self._functions = as_tuple(functions)
+            # Check if shorthand notation has been provided:
+            for j in range(2):
```
georgebisbas, line 785: why 2? looks hardcoded, wasn;t it better

EdCaunt: There's only ever going to be a minimum and maximum bound. This is actually enforced higher up in this function

devito/passes/clusters/implicit.py
```diff
@@ -94,8 +91,10 @@ def callback(self, clusters, prefix):
                 processed.append(c)
                 continue
 
-            # Get the dynamic thickness mapper for the given MultiSubDomain
-            mapper, dims = lower_msd(d.msd, c)
+            # Get all MultiSubDimensions in the cluster and get the dynamic thickness
+            # mapper for the associated MultiSubDomain
+            mapper, dims = lower_msd({msdim(i.dim) for i in c.ispace[idx:]} - {None})
```
FabioLuporini, line 96: I'm not sure I understand this change. If you are visiting `d`, why do you need its msd as well as any other nested msds ? 

I don't think this is particularly clean, but besides that, I'm failing to understand why

devito/passes/clusters/implicit.py
```diff
@@ -170,7 +169,7 @@ def callback(self, clusters, prefix):
                 continue
 
             # Get the dynamic thickness mapper for the given MultiSubDomain
-            mapper, dims = lower_msd(d.msd, c)
+            mapper, dims = lower_msd({i.dim for i in ispace})
```
FabioLuporini, line 172: same story as above

FabioLuporini (commented):
Left some comments; note that changing the `lower_msd` interface will also break PRO

devito/passes/clusters/implicit.py
EdCaunt: `lower_msd` has been completely replaced. Now stands for "lower multisubdimension" and takes all the multisubdimensions present in the cluster, rather than the multisubdomain attached to the outermost dimension (all references to which have been dropped at this point).

devito/types/grid.py
```diff
@@ -774,34 +764,39 @@ def __subdomain_finalize__(self, grid, counter=0, **kwargs):
...
 
-            functions.append(f)
-        self._functions = as_tuple(functions)
+        dimensions = []
+        for i, d in enumerate(grid.dimensions):
+            # Check if shorthand notation has been provided:
+            for j in range(2):
+                idx = 2*i + j
+                if isinstance(self._local_bounds[idx], int):
+                    sd_func.data[:, idx] = np.full((self._n_domains,),
```
mloubout, line 787: why not just `sd_func.data[:, idx] = self._local_bounds[idx]` ?

EdCaunt: Good spot. Also removes a bunch of extra complexity

devito/passes/clusters/implicit.py
```diff
@@ -170,7 +172,7 @@ def callback(self, clusters, prefix):
                 continue
 
             # Get the dynamic thickness mapper for the given MultiSubDomain
-            mapper, dims = lower_msd(d.msd, c)
+            mapper, dims = lower_msd({i.dim for i in ispace}, c)
```
FabioLuporini, line 175: this is just `ispace.itdims`

FabioLuporini (approved):

devito/passes/clusters/implicit.py
EdCaunt: It'll get forgotten if it isn't done here, so will just do it

mloubout (approved):

PR #2395, opened 2024-07-01
MPI: fix haloupdate inside factor conditional

mloubout:
On top of #2394

### DIFF
--- a/devito/passes/iet/mpi.py
+++ b/devito/passes/iet/mpi.py
@@ -6,6 +6,7 @@ from devito.ir.iet import (Call, Expression, HaloSpot, Iteration, FindNodes,
                            MapNodes, MapHaloSpots, Transformer,
                            retrieve_iteration_tree)
 from devito.ir.support import PARALLEL, Scope
+from devito.ir.support.guards import GuardFactorEq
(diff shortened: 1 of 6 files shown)


devito/passes/iet/mpi.py
```diff
@@ -182,7 +183,12 @@ def rule2(dep, hs, loc_indices):
             # within the same Conditional, otherwise we would break the control
             # flow semantics
             if cond_mapper.get(hs0) != cond_mapper.get(hs1):
-                continue
+                cond = cond_mapper.get(hs1).symmetric_difference(cond_mapper.get(hs0))
```
FabioLuporini, line 186: would it be neater to de-nest it by re-writing it as (pseudocode):

```
cond = cond_mapper.get(hs1).symmetric_difference(cond_mapper.get(hs0))
if len(cond) > 1:
    <explanation>
    continue
elif len(cond) == 1 and isinstance(cond.pop().condition, GuardFactorEq):
    <explanation>
    continue
```

mloubout: cleaner to just filter when making the cond_mapper

PR #2396, opened 2024-07-02
compiler: Yet another batch of compilation tweaks

FabioLuporini:
These are mostly to improve code generation and performance of elastic-like PDEs.

### DIFF
--- a/devito/__init__.py
+++ b/devito/__init__.py
@@ -101,8 +101,21 @@ configuration.add('jit-backdoor', 0, [0, 1], preprocessor=bool, impacts_jit=Fals
 # optimisations.
 configuration.add('safe-math', 0, [0, 1], preprocessor=bool, callback=reinit_compiler)
 
+
 # Enable/disable automatic padding for allocated data
-configuration.add('autopadding', False, [False, True, np.float32, np.float64])
+def _preprocess_autopadding(v):
+    return {
+        '0': False,
+        '1': np.float32,
+        True: np.float32,
+        'fp16': np.float16,
+        'fp32': np.float32,
+        'fp64': np.float64
+    }.get(v, v)
+
+configuration.add('autopadding', False,  # noqa: E305
+                  [False, True, 0, 1, np.float16, np.float32, np.float64],
+                  preprocessor=_preprocess_autopadding)
 
 # Select target device
 configuration.add('deviceid', -1, preprocessor=int, impacts_jit=False)
--- a/devito/__init__.py
+++ b/devito/__init__.py
@@ -102,7 +102,7 @@ configuration.add('jit-backdoor', 0, [0, 1], preprocessor=bool, impacts_jit=Fals
 configuration.add('safe-math', 0, [0, 1], preprocessor=bool, callback=reinit_compiler)
 
 # Enable/disable automatic padding for allocated data
-configuration.add('autopadding', False, [False, True])
+configuration.add('autopadding', False, [False, True, np.float32, np.float64])
 
 # Select target device
 configuration.add('deviceid', -1, preprocessor=int, impacts_jit=False)
--- a/devito/core/gpu.py
+++ b/devito/core/gpu.py
@@ -120,17 +120,17 @@ class DeviceOperatorMixin:
         options0 = kwargs0.pop('options')
 
         def wrapper(expressions, mode='default', options=None, **kwargs1):
-            options = {**options0, **(options or {})}
             kwargs = {**kwargs0, **kwargs1}
 
             if mode == 'host':
-                par_disabled = options['par-disabled']
+                options = options or {}
                 target = {
                     'platform': 'cpu64',
-                    'language': 'C' if par_disabled else 'openmp',
+                    'language': 'C' if options0['par-disabled'] else 'openmp',
                     'compiler': 'custom'
                 }
             else:
+                options = {**options0, **(options or {})}
                 target = None
 
             return rcompile(expressions, kwargs, options, target=target)
--- a/devito/ir/clusters/algorithms.py
+++ b/devito/ir/clusters/algorithms.py
@@ -343,6 +343,11 @@ class Stepper(Queue):
         # Reconstruct the Clusters
         processed = []
         for c in clusters:
+            exprs = c.exprs
+
+            sub_iterators = dict(c.ispace.sub_iterators)
+            sub_iterators[d] = [i for i in sub_iterators[d] if i not in subiters]
+
             # Apply substitutions to expressions
             # Note: In an expression, there could be `u[t+1, ...]` and `v[t+1,
             # ...]`, where `u` and `v` are TimeFunction with circular time
@@ -350,17 +355,19 @@ class Stepper(Queue):
             # indices above are therefore conceptually different, so they will
             # be replaced with the proper ModuloDimension through two different
             # calls to `xreplace_indices`
-            exprs = c.exprs
             groups = as_mapper(mds, lambda d: d.modulo)
             for size, v in groups.items():
-                subs = {md.origin: md for md in v}
-                func = partial(xreplace_indices, mapper=subs, key=partial(rule, size))
+                key = partial(rule, size)
+                if size == 1:
+                    # Optimization -- avoid useless "% 1" ModuloDimensions
+                    subs = {md.origin: 0 for md in v}
+                else:
+                    subs = {md.origin: md for md in v}
+                    sub_iterators[d].extend(v)
+
+                func = partial(xreplace_indices, mapper=subs, key=key)
                 exprs = [e.apply(func) for e in exprs]
 
-            # Augment IterationSpace
-            sub_iterators = dict(c.ispace.sub_iterators)
-            sub_iterators[d] = tuple(i for i in sub_iterators[d] + tuple(mds)
-                                     if i not in subiters)
             ispace = IterationSpace(c.ispace.intervals, sub_iterators,
                                     c.ispace.directions)
 
--- a/devito/ir/iet/visitors.py
+++ b/devito/ir/iet/visitors.py
@@ -480,7 +480,7 @@ class CGen(Visitor):
             code = c.Assign(lhs, rhs)
 
         if o.pragmas:
-            code = c.Module(list(o.pragmas) + [code])
+            code = c.Module(self._visit(o.pragmas) + (code,))
 
         return code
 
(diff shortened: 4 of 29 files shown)


devito/ir/support/basic.py
```diff
@@ -1330,6 +1326,10 @@ def is_regular(self):
 
 # *** Utils
 
+def vinf(entries):
```
georgebisbas, line 1329: nitpick, would do vec_inf

devito/ir/support/basic.py
```diff
@@ -2,6 +2,7 @@
 from functools import cached_property
 
 from sympy import S
+import sympy
```
mloubout, line 5: jusit import Interval above no?

devito/passes/iet/languages/utils.py
```diff
@@ -0,0 +1,5 @@
+__all__ = ['joins']
+
+
+def joins(*symbols):
+    return ",".join(sorted([i.name for i in symbols]))
```
mloubout, line 5: why sorted?

tests/test_linearize.py
```diff
@@ -611,4 +611,4 @@ def test_different_dtype():
 
     # Check generated code has different strides for different dtypes
     assert "bL0(x,y) b[(x)*y_stride0 + (y)]" in str(op1)
-    assert "L0(x,y) f[(x)*y_stride1 + (y)]" in str(op1)
+    assert "L0(x,y) f[(x)*y_stride0 + (y)]" in str(op1)
```
mloubout, line 614: Can we parametrize this test with autopadding so `y_stride1` case is there for "mixed" padding

mloubout (commented):
Mostly cleanup, look goog.

Few minorcomments

devito/ir/support/basic.py
FabioLuporini: I would, but to make a neat distinction with our own Interval...

devito/ir/support/basic.py
FabioLuporini: but we have vmax, vmin already, so for homogeneity vinf stands better no? we would have to change them all otherwise

tests/test_linearize.py
FabioLuporini: it's not possible by construction anymore; autopadding is now a global choice, in essence. This means that if you have an fp32 and an fp64 the data will be padded to either fp32 or fp64; in the former case, the performance will be suboptimal. But regardless, we will always only generate one stride per dimension. See `types/basic.py::AbstractFunction::__padding_dtype__` logic

That said, I've now parametrized the test as you suggested, but the expected outcome is always `y_stride0`

devito/ir/support/basic.py
georgebisbas: hm sorry did not pay attention to that

georgebisbas (approved):
I do not have something more to add.

devito/core/gpu.py
```diff
@@ -120,17 +120,17 @@ def _rcompile_wrapper(cls, **kwargs0):
         options0 = kwargs0.pop('options')
 
         def wrapper(expressions, mode='default', options=None, **kwargs1):
-            options = {**options0, **(options or {})}
             kwargs = {**kwargs0, **kwargs1}
 
             if mode == 'host':
-                par_disabled = options['par-disabled']
+                options = options or {}
```
EdCaunt, line 126: Could you set the kwarg default to `options={}` rather than `options=None`, thereby removing this line, and simplifying line 133 below?

devito/ir/clusters/algorithms.py
```diff
@@ -343,24 +343,31 @@ def rule(size, e):
         # Reconstruct the Clusters
         processed = []
         for c in clusters:
+            exprs = c.exprs
```
EdCaunt, line 346: This doesn't look like it gets used before being overwritten in the `for` loop below

devito/ir/iet/visitors.py
```diff
@@ -480,7 +480,7 @@ def visit_Expression(self, o):
             code = c.Assign(lhs, rhs)
 
         if o.pragmas:
```
EdCaunt, line 482: These four lines are repeated twice. Worth constructing a utility function?

devito/ir/support/basic.py
```diff
@@ -1345,3 +1345,54 @@ def retrieve_accesses(exprs, **kwargs):
...
+        where zl is a left SubDimension with thickness, say, 4
+        The test will return True, as the two index access functions never
+        overlap.
+    """
+    if e0 == e1:
+        return False
+
+    if d.is_Custom:
+        subs = {}
+    elif d.is_Sub and d.is_left:
```
EdCaunt, line 1378: Why can middle/right subdimensions be ignored? Genuinely curious

devito/ir/support/basic.py
```diff
@@ -1345,3 +1345,54 @@ def retrieve_accesses(exprs, **kwargs):
...
+        subs = {d.root.symbolic_min: 0, **dict([d.thickness.left])}
+    else:
+        return False
+
+    m = it.symbolic_min.subs(subs)
+    M = it.symbolic_max.subs(subs)
+
+    p00 = e0._subs(d, m)
```
EdCaunt, line 1386: Explanatory comment of the 00, 01, etc would be helpful here

devito/passes/iet/linearization.py
```diff
@@ -129,14 +125,25 @@ def __init__(self, mode, dtype, sregistry):
         self.strides_dynamic = {}  # Strides varying regularly across iterations
         self.dists = {}
 
+    def _select_indexing_dtype(self, f):
+        # Some objects may contain an extremely large number of elements, so we
+        # conservatively use int64 to avoid potential overflows regardless of
+        # what the user requested via `index-mode`
+        if f.is_SparseTimeFunction:
```
EdCaunt, line 132: Just `SparseFunction` I would say, given that the issue occurs due to a large number of points, rather than a large nt iirc

devito/symbolics/manipulation.py
```diff
@@ -286,12 +286,12 @@ def pow_to_mul(expr):
         except TypeError:
             # E.g., a Symbol, or possibly a generic expression
             return expr
-        if exp > 10 or exp < -10 or int(exp) != exp or exp == 0:
+        if exp > 10 or exp < -10 or exp == 0:
             # Large and non-integer powers remain untouched
```
EdCaunt, line 290: "non-integer" part of this comment needs moving to the `elif`

devito/types/basic.py
```diff
@@ -999,30 +999,41 @@ def __padding_setup__(self, **kwargs):
...
-                # Already a multiple of `mmts`, no need to pad
-                return nopadding
+        if not self.__padding_dtype__:
+            return nopadding
 
-            dpadding = (0, (mmts - remainder))
-            padding = [(0, 0)]*self.ndim
-            padding[self.dimensions.index(d)] = dpadding
+        # The padded Dimension
+        candidates = self.space_dimensions
```
EdCaunt, line 1020: Could shorten to:
```
if not self.space_dimensions:
    return nopadding
d = self.space_dimensions[-1]
```

tests/test_dimension.py
```diff
@@ -210,6 +210,21 @@ def test_modulo_dims_generation_v2(self):
         assert np.all(f.data[3] == 2)
         assert np.all(f.data[4] == 4)
 
+    def test_degenerate_to_zero(self):
```
EdCaunt, line 213: Docstring or a comment explaining what this does would help understanding at a glance

devito/core/gpu.py
FabioLuporini: setting kwargs to mutable objects is strongly discouraged as it potentially leads to subtle bugs, so we just never do it

devito/ir/clusters/algorithms.py
FabioLuporini: yes, but I like to stick to the conventional ordering "first we pull/transform/handle the exprs, then the ispace (and its related metadata such as the sub-iterators), then the guards, syncs, properties, etc". Basically the order of the kwargs in the Cluster constructor.

devito/ir/iet/visitors.py
FabioLuporini: which four lines? I only see two, starting at `if o.pragmas` -- which is too little to deserve a separate function

devito/ir/support/basic.py
FabioLuporini: we just don't treat them as we don't -- practically speaking need to

as per the docstring:

"A rudimentary test ..."

and

"Our implementation focuses on tiny yet relevant cases"

so basically this is a simplistic implementation, to be refined in the future if we ever will have to

devito/ir/support/basic.py
FabioLuporini: just dummy variable names because you have two "dimensions":

* e0 and e1
* m and M

which leads to four objects

devito/passes/iet/linearization.py
FabioLuporini: order of millions is fine, order of billions (due to time...) isn't

devito/symbolics/manipulation.py
FabioLuporini: good catch, fixing

devito/types/basic.py
FabioLuporini: right, fixing

tests/test_dimension.py
FabioLuporini: adding docstring

devito/ir/iet/visitors.py
EdCaunt:
```
if o.pragmas:
            code = c.Module(self._visit(o.pragmas) + (code,))

return code
```
but tbh, you're probably right

devito/passes/iet/linearization.py
EdCaunt: Sure, but those are different array dimensions right? Unless the array is linearised?

devito/__init__.py
```diff
@@ -102,7 +102,7 @@ def reinit_compiler(val):
 configuration.add('safe-math', 0, [0, 1], preprocessor=bool, callback=reinit_compiler)
 
 # Enable/disable automatic padding for allocated data
-configuration.add('autopadding', False, [False, True])
+configuration.add('autopadding', False, [False, True, np.float32, np.float64])
```
mloubout, line 105: How are you passing `np.float32` as en env var? This needs to accept a string of the type or like "float32/f32"

mloubout (changes requested):

devito/__init__.py
FabioLuporini: fixed

mloubout (approved):

EdCaunt (approved):

PR #2399, opened 2024-07-04
compiler: Default to index-mode=int32

FabioLuporini:
This is based #2396 hence the huge diff, otherwise it's just ~15 lines change

This is the diff: https://github.com/devitocodes/devito/pull/2399/commits/7f6762a0540b01062df5cb8ea4236dcfb2773474


mloubout (approved):

georgebisbas (approved):
The b2b diff looks ok

devito/__init__.py
```diff
@@ -101,8 +101,21 @@ def reinit_compiler(val):
 # optimisations.
 configuration.add('safe-math', 0, [0, 1], preprocessor=bool, callback=reinit_compiler)
 
+
 # Enable/disable automatic padding for allocated data
-configuration.add('autopadding', False, [False, True])
+def _preprocess_autopadding(v):
+    return {
+        '0': False,
```
EdCaunt, line 108: Does this also need a `{False: False}` mapping?

devito/ir/support/basic.py
```diff
@@ -2,6 +2,7 @@
 from functools import cached_property
 
 from sympy import S
+import sympy
```
EdCaunt, line 5: Nitpick: could you not import `Interval` and `Vector` from SymPy on the line above? It would remove this line and tidy up some of the other code

devito/passes/iet/engine.py
```diff
@@ -398,18 +398,21 @@ def abstract_efunc(efunc):
     return efunc
 
 
-def abstract_objects(objects, sregistry=None):
+def abstract_objects(objects0, sregistry=None):
     """
     Proxy for `abstract_object`.
     """
+    # Expose hidden objects for complete reconstruction
+    objects = []
```
EdCaunt, line 406: If order doesn't matter, then deepcopying `objects0` or using `objects = list(object0)` instead of initialising an empty list would remove the `objects.append(i)` line

devito/passes/iet/engine.py
```diff
@@ -398,18 +398,21 @@ def abstract_efunc(efunc):
...
+
     # Precedence rules make it possible to reconstruct objects that depend on
     # higher priority objects
-    priority = {
-        Array: 1,
-        DiscreteFunction: 2,
-        AbstractIncrDimension: 3,
-        BlockDimension: 4,
-    }
+    keys = [Bundle, Array, DiscreteFunction, AbstractIncrDimension, BlockDimension]
```
EdCaunt, line 414: Nitpick: worth making this a tuple?

devito/passes/clusters/blocking.py
```diff
@@ -569,7 +574,8 @@ def next(self, prefix, d, clusters):
 class SynthesizeSkewing(Queue):
 
     """
-    Construct a new sequence of clusters with skewed expressions and iteration spaces.
+    Construct a new sequence of clusters with skewed expressions and
```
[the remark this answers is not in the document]
georgebisbas: This is some pre-requisite for generating loops for temporal blocking.
You may find more info on this here if interested:

https://link.springer.com/content/pdf/10.1007/BF01407876.pdf

devito/__init__.py
FabioLuporini: no

```
>>> False == 0
True
>>> hash(False) == hash(0)
True
```

devito/ir/support/basic.py
FabioLuporini: I could, but don't want to, for clarity. We have our own `Interval` and `Vector` too

devito/passes/iet/engine.py
FabioLuporini: note that these comments belong to the parent PR as per this PR's description

PR #2375, opened 2024-05-26
API: Introducing complex (np.complex64/128) native support 

mloubout:
Support complex float data type.

PR #2403, opened 2024-07-11
compiler: Fix complex arguments and implement float16 lowering

enwask:
Fixes some issues that prevented complex float arguments from being passed correctly. Also implements lowering for half-precision floats with the `_Float16` type.

Some remaining tasks:

- MPI with complex floating types
- Look into MPI with float16, if possible
- Type conversion is done by modifying the global `ctypes_vector_mappet`; need some more tests for operator building/rebuilding/pickling with float16, or potentially a different (hopefully less brittle) solution
- More test coverage of existing behavior applied to float16 types

### DIFF
--- a/devito/data/allocators.py
+++ b/devito/data/allocators.py
@@ -92,8 +92,12 @@ class MemoryAllocator(AbstractMemoryAllocator):
         return
 
     def alloc(self, shape, dtype, padding=0):
-        datasize = int(reduce(mul, shape))
-        ctype = dtype_to_ctype(dtype)
+        # For complex number, allocate double the size of its real/imaginary part
+        alloc_dtype = dtype(0).real.__class__
+        c_scale = 2 if np.issubdtype(dtype, np.complexfloating) else 1
+
+        datasize = int(reduce(mul, shape) * c_scale)
+        ctype = dtype_to_ctype(alloc_dtype)
 
         # Add padding, if any
         try:
--- a/devito/data/allocators.py
+++ b/devito/data/allocators.py
@@ -11,8 +11,7 @@ import numpy as np
 
 from devito.logger import logger
 from devito.parameters import configuration
-from devito.tools import is_integer
-from devito.tools.dtypes_lowering import dtype_alloc_ctype
+from devito.tools import is_integer, dtype_alloc_ctype
 
 __all__ = ['ALLOC_ALIGNED', 'ALLOC_NUMA_LOCAL', 'ALLOC_NUMA_ANY',
            'ALLOC_KNL_MCDRAM', 'ALLOC_KNL_DRAM', 'ALLOC_GUARD',
--- a/devito/data/allocators.py
+++ b/devito/data/allocators.py
@@ -11,7 +11,8 @@ import numpy as np
 
 from devito.logger import logger
 from devito.parameters import configuration
-from devito.tools import dtype_to_ctype, is_integer
+from devito.tools import is_integer
+from devito.tools.dtypes_lowering import dtype_alloc_ctype
 
 __all__ = ['ALLOC_ALIGNED', 'ALLOC_NUMA_LOCAL', 'ALLOC_NUMA_ANY',
            'ALLOC_KNL_MCDRAM', 'ALLOC_KNL_DRAM', 'ALLOC_GUARD',
@@ -92,12 +93,8 @@ class MemoryAllocator(AbstractMemoryAllocator):
         return
 
     def alloc(self, shape, dtype, padding=0):
-        # For complex number, allocate double the size of its real/imaginary part
-        alloc_dtype = dtype(0).real.__class__
-        c_scale = 2 if np.issubdtype(dtype, np.complexfloating) else 1
-
+        ctype, c_scale = dtype_alloc_ctype(dtype)
         datasize = int(reduce(mul, shape) * c_scale)
-        ctype = dtype_to_ctype(alloc_dtype)
 
         # Add padding, if any
         try:
--- a/devito/ir/iet/nodes.py
+++ b/devito/ir/iet/nodes.py
@@ -1056,15 +1056,13 @@ class Dereference(ExprStmt, Node):
             assert issubclass(self.pointer._C_ctype, ctypes._Pointer), \
                    "Scalar dereference must have a pointer ctype"
             ret.extend([self.pointer._C_symbol, self.pointee._C_symbol])
+        elif self.pointer.is_PointerArray or self.pointer.is_TempFunction:
+            ret.extend([self.pointer.indexed, self.pointee.indexed])
+            ret.extend(flatten(i.free_symbols
+                               for i in self.pointee.symbolic_shape[1:]))
+            ret.extend(self.pointer.free_symbols)
         else:
-            ret.append(self.pointer.indexed)
-            if self.pointer.is_PointerArray or self.pointer.is_TempFunction:
-                ret.append(self.pointee.indexed)
-                ret.extend(flatten(i.free_symbols
-                                   for i in self.pointee.symbolic_shape[1:]))
-                ret.extend(self.pointer.free_symbols)
-            else:
-                ret.append(self.pointee._C_symbol)
+            ret.extend([self.pointer.indexed, self.pointee._C_symbol])
         return tuple(filter_ordered(ret))
 
     @property
--- a/devito/ir/iet/nodes.py
+++ b/devito/ir/iet/nodes.py
@@ -1033,7 +1033,6 @@ class Dereference(ExprStmt, Node):
           `pointee` is a field in `pointer`.
         * `pointer` is a Symbol with its _C_ctype deriving from ct._Pointer, and
           `pointee` is a Symbol representing the dereferenced value.
-          
     """
 
     is_Dereference = True
@@ -1054,15 +1053,16 @@ class Dereference(ExprStmt, Node):
     def expr_symbols(self):
         ret = []
         if self.pointer.is_Symbol:
-            assert (issubclass(self.pointer._C_ctype, ctypes._Pointer),
-                    "Scalar dereference must have a pointer ctype")
+            assert issubclass(self.pointer._C_ctype, ctypes._Pointer), \
+                   "Scalar dereference must have a pointer ctype"
             ret.append(self.pointer._C_symbol)
             ret.append(self.pointee._C_symbol)
         else:
             ret.append(self.pointer.indexed)
             if self.pointer.is_PointerArray or self.pointer.is_TempFunction:
                 ret.append(self.pointee.indexed)
-                ret.extend(flatten(i.free_symbols for i in self.pointee.symbolic_shape[1:]))
+                ret.extend(flatten(i.free_symbols
+                                   for i in self.pointee.symbolic_shape[1:]))
                 ret.extend(self.pointer.free_symbols)
             else:
                 ret.append(self.pointee._C_symbol)
--- a/devito/ir/iet/nodes.py
+++ b/devito/ir/iet/nodes.py
@@ -1,6 +1,7 @@
 """The Iteration/Expression Tree (IET) hierarchy."""
 
 import abc
+import ctypes
 import inspect
 from functools import cached_property
 from collections import OrderedDict, namedtuple
@@ -1030,6 +1031,9 @@ class Dereference(ExprStmt, Node):
         * `pointer` is a PointerArray or TempFunction, and `pointee` is an Array.
         * `pointer` is an ArrayObject representing a pointer to a C struct, and
           `pointee` is a field in `pointer`.
+        * `pointer` is a Symbol with its _C_ctype deriving from ct._Pointer, and
+          `pointee` is a Symbol representing the dereferenced value.
+          
     """
 
     is_Dereference = True
@@ -1048,13 +1052,20 @@ class Dereference(ExprStmt, Node):
 
     @property
     def expr_symbols(self):
-        ret = [self.pointer.indexed]
-        if self.pointer.is_PointerArray or self.pointer.is_TempFunction:
-            ret.append(self.pointee.indexed)
-            ret.extend(flatten(i.free_symbols for i in self.pointee.symbolic_shape[1:]))
-            ret.extend(self.pointer.free_symbols)
-        else:
+        ret = []
+        if self.pointer.is_Symbol:
+            assert (issubclass(self.pointer._C_ctype, ctypes._Pointer),
+                    "Scalar dereference must have a pointer ctype")
+            ret.append(self.pointer._C_symbol)
             ret.append(self.pointee._C_symbol)
+        else:
+            ret.append(self.pointer.indexed)
+            if self.pointer.is_PointerArray or self.pointer.is_TempFunction:
+                ret.append(self.pointee.indexed)
+                ret.extend(flatten(i.free_symbols for i in self.pointee.symbolic_shape[1:]))
+                ret.extend(self.pointer.free_symbols)
+            else:
+                ret.append(self.pointee._C_symbol)
         return tuple(filter_ordered(ret))
 
     @property
--- a/devito/ir/iet/nodes.py
+++ b/devito/ir/iet/nodes.py
@@ -1055,8 +1055,7 @@ class Dereference(ExprStmt, Node):
         if self.pointer.is_Symbol:
             assert issubclass(self.pointer._C_ctype, ctypes._Pointer), \
                    "Scalar dereference must have a pointer ctype"
-            ret.append(self.pointer._C_symbol)
-            ret.append(self.pointee._C_symbol)
+            ret.extend([self.pointer._C_symbol, self.pointee._C_symbol])
         else:
             ret.append(self.pointer.indexed)
             if self.pointer.is_PointerArray or self.pointer.is_TempFunction:
--- a/devito/ir/iet/visitors.py
+++ b/devito/ir/iet/visitors.py
@@ -10,6 +10,7 @@ from itertools import chain, groupby
 import ctypes
 
 import cgen as c
+import numpy as np
 from sympy import IndexedBase
 from sympy.core.function import Application
 
@@ -188,6 +189,21 @@ class CGen(Visitor):
     }
     _restrict_keyword = 'restrict'
 
+    def _complex_type(self, ctypestr, dtype):
+        # Not complex
+        try:
+            if not np.issubdtype(dtype, np.complexfloating):
+                return ctypestr
+        except TypeError:
+            return ctypestr
+        # Complex only supported for float and double
+        if ctypestr not in ('float', 'double'):
+            return ctypestr
+        if self._compiler._cpp:
+            return 'std::complex<%s>' % ctypestr
+        else:
+            return '%s _Complex' % ctypestr
+
     def _gen_struct_decl(self, obj, masked=()):
         """
         Convert ctypes.Struct -> cgen.Structure.
@@ -243,10 +259,10 @@ class CGen(Visitor):
                       if getattr(obj.function, k, False) and v not in masked]
 
         if (obj._mem_stack or obj._mem_constant) and mode == 1:
-            strtype = obj._C_typedata
+            strtype = self._complex_type(obj._C_typedata, obj.dtype)
             strshape = ''.join('[%s]' % ccode(i) for i in obj.symbolic_shape)
         else:
-            strtype = ctypes_to_cstr(obj._C_ctype)
+            strtype = self._complex_type(ctypes_to_cstr(obj._C_ctype), obj.dtype)
             strshape = ''
             if isinstance(obj, (AbstractFunction, IndexedData)) and mode >= 1:
                 if not obj._mem_stack:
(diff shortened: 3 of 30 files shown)


devito/passes/iet/definitions.py
```diff
@@ -414,6 +414,7 @@ def place_casts(self, iet, **kwargs):
 
     @iet_pass
     def make_langtypes(self, iet):
+        iet, _ = lower_scalar_half(iet, self.lang, self.sregistry)
```
mloubout, line 417: Merge int oa lower_dtypes

devito/passes/iet/dtypes.py
```diff
@@ -2,8 +2,44 @@
...
+            continue
+
+        ptr = s._rebuild(dtype=half_p)
+        val = Symbol(name=sregistry.make_name(prefix='hf'), dtype=half, is_const=True)
+
+        params_mapper[s] = ptr
+        body_mapper[s] = val
+        body.append(Dereference(val, ptr))  # val = *ptr
+
+    body.extend(as_tuple(Uxreplace(body_mapper).visit(iet.body)))
```
mloubout, line 38: why does it. need extend? `Uxreplace(body_mapper).visit(iet.body)` should doo all it needs no?

mloubout (commented):
Some quick comments

devito/symbolics/printer.py
```diff
@@ -216,13 +226,17 @@ def _print_Float(self, expr):
...
+            rv = '%sF16' % rv
 
         return rv
 
     def _print_ImaginaryUnit(self, expr):
         if self.compiler._cpp:
             if self.single_prec(with_f=True):
                 return '1if'
+            elif self.half_prec(with_f=True):
+                return '1if16'
```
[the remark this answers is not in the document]
enwask: ah, I tested and this works on my end but I guess it's a clang extension. I'll remove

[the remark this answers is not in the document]
mloubout: You tested half precision complex numbers?

devito/tools/dtypes_lowering.py
```diff
@@ -145,6 +145,9 @@ def dtype_to_ctype(dtype):
         # Bypass np.ctypeslib's normalization rules such as
         # `np.ctypeslib.as_ctypes_type(ctypes.c_void_p) -> ctypes.c_ulong`
         return dtype
+    elif dtype == np.float16:
```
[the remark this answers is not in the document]
enwask: I think this should be fine since it will only catch the types before they're mapped to ours; numpy docs suggest using equality for checking dtypes if you mean the `==` is dangerous as opposed to `is`. But yes will move to allocator

devito/symbolics/extended_dtypes.py
```diff
@@ -15,6 +18,52 @@
...
+    # Ctype for non-scalar half floats
+    @classmethod
+    def from_param(cls, val):
+        return cls(np.float16(val).view(np.uint16))
+
+
+c_float16 = type('_Float16', (_c_half,), {})
+
+
+class _c_half_p(ct.POINTER(c_float16)):
```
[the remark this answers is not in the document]
enwask: Correct, primarily because we need the bespoke `from_param` handling

devito/passes/iet/dtypes.py
```diff
@@ -2,8 +2,44 @@
...
 from devito.ir import FindSymbols, Uxreplace
+from devito.ir.iet.nodes import Dereference
+from devito.tools.utils import as_tuple
+from devito.types.basic import Symbol
 
-__all__ = ['lower_complex']
+__all__ = ['lower_scalar_half', 'lower_complex']
+
+
+def lower_scalar_half(iet, lang, sregistry):
```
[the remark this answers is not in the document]
enwask: Ok, will refactor. I suppose `lower_dtypes` will first handle the float16 scalars, then call `_complex_headers` for the includes, then finally apply the primary `types` mapping?

devito/passes/iet/dtypes.py
enwask: Well this prepends whatever I've put in `body` within the loop (those deref expressions). What would be a better way to do this? I can just append to `iet.body` I guess and do the mapping on the whole thing, including what I've added

devito/passes/iet/dtypes.py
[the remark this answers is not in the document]
mloubout: not sure why three steps, just one step that does the type mapping then on step that does the headers/...

devito/symbolics/printer.py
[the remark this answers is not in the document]
enwask: Well I just manually added them into a kernel and printed the values. Can't really test with devito as there's no `np.complex32`

devito/passes/iet/dtypes.py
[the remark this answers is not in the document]
enwask: True, I can group the type mappings but I do need to do the scalars in their own loop before the other mappings (to avoid them getting mapped along with any array/indexed/whatever else, since those mapped types don't work for scalars)

devito/symbolics/printer.py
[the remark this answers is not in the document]
mloubout: Your kernel is a C kernel, (not sure what `i` is in that case), this part of the printer is for `std::complex`  with openacc/sycl or other c++ kenrels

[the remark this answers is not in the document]
enwask: Huh, apparently the literal suffix is also a GNU extension. Either way, oops, did not see the cpp compiler check. I'll revert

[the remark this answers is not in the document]
mloubout: >the literal suffix is also a GNU extension

Yes that's what you added to the print_Float and such, this is is the definition of complex unit

devito/ir/iet/nodes.py
```diff
@@ -1030,6 +1031,8 @@ class Dereference(ExprStmt, Node):
         * `pointer` is a PointerArray or TempFunction, and `pointee` is an Array.
         * `pointer` is an ArrayObject representing a pointer to a C struct, and
           `pointee` is a field in `pointer`.
+        * `pointer` is a Symbol with its _C_ctype deriving from ct._Pointer, and
+          `pointee` is a Symbol representing the dereferenced value.
```
georgebisbas, line 1035: Do all these lines need some concatenation?

enwask: What sort of concatenation?

georgebisbas: My bad, I did not see that these are following cases

devito/ir/iet/visitors.py
```diff
@@ -208,7 +209,7 @@ def _gen_struct_decl(self, obj, masked=()):
             while issubclass(ctype, ctypes._Pointer):
                 ctype = ctype._type_
 
-            if not issubclass(ctype, ctypes.Structure):
+            if not issubclass(ctype, ctypes.Structure) or issubclass(ctype, NoDeclStruct):
```
EdCaunt, line 212: `NoDeclStruct` subclasses `ctypes.Structure`, so the `or` shouldn't be needed right?

devito/passes/iet/dtypes.py
```diff
@@ -2,11 +2,53 @@
...
+    """
+    # Include complex headers if needed (before we replace complex dtypes)
+    metadata = _complex_includes(iet, lang, compiler)
+
+    body_prefix = []  # Derefs to prepend to the body
+    body_mapper = {}
+    params_mapper = {}
+
+    # Lower scalar float16s to pointers and dereference them
+    if lang.get('half_types') is not None:
```
EdCaunt, line 24: Could this just be `if lang.get('half_types)`?

devito/symbolics/extended_dtypes.py
```diff
@@ -15,6 +18,45 @@
 }
 
 
+class NoDeclStruct(ct.Structure):
+    # ctypes.Structure that does not generate a struct definition
```
EdCaunt, line 22: Should be docstring

devito/symbolics/extended_dtypes.py
```diff
@@ -1,10 +1,13 @@
+import ctypes as ct
```
EdCaunt, line 1: For consistency with the rest of the codebase: `import ctypes` rather than `import ctypes as ct`

devito/symbolics/extended_dtypes.py
```diff
@@ -15,6 +18,45 @@
...
+    # Ctype for non-scalar half floats
+    @classmethod
+    def from_param(cls, val):
+        return cls(np.float16(val).view(np.uint16))
+
+
+class c_half_p(ct.POINTER(c_half)):
+    # Ctype for half scalars; we can't directly pass _Float16 values so
+    # we use a pointer and dereference (see `passes.iet.dtypes`)
+    @classmethod
```
EdCaunt, line 54: Line break before `@classmethod`

devito/ir/iet/visitors.py
enwask: No, the check is inverted—we break out here if the type _is not_ a `Structure`, or if it _is_ a `NoDeclStruct`

devito/passes/iet/dtypes.py
enwask: It could be but I believe pep8 suggests this syntax

devito/symbolics/extended_dtypes.py
enwask: Ok, presumably same for the `c_complex` etc.?

EdCaunt: Yeah, I think so

devito/data/allocators.py
```diff
@@ -11,7 +11,8 @@
 
 from devito.logger import logger
 from devito.parameters import configuration
-from devito.tools import dtype_to_ctype, is_integer
+from devito.tools import is_integer
+from devito.tools.dtypes_lowering import dtype_alloc_ctype
```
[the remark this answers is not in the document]
enwask: It is in `__all__`; just didn't think to import it from tools. I'll replace

devito/operator/operator.py
```diff
@@ -285,6 +290,11 @@ def _lower(cls, expressions, **kwargs):
     def _rcompile_wrapper(cls, **kwargs0):
         raise NotImplementedError
 
+    @classmethod
+    def _load_dtype_mappings(cls, **kwargs):
+        lang: type[LangBB] = cls._Target.DataManager.lang
+        ctypes_vector_mapper.update(lang.mapper.get('types', {}))
```
mloubout, line 296: Now that's a bit tricky because this updates a global `ctypes_vector_mapper` which might lead to odd behavior building multiple operators with different languages.
Do you know where it's called and needs those `types` ? I.e can the mapper be "local" to the operator and passed there?

devito/passes/iet/dtypes.py
```diff
@@ -1,14 +1,54 @@
...
+        # body with a dereference (using the original symbol's dtype)
+        ptr = s._rebuild(dtype=Float16P, is_const=True)
+        val = Symbol(name=sregistry.make_name(prefix='hf'), dtype=s.dtype,
+                     is_const=s.is_const)
+
+        params_mapper[s], body_mapper[s] = ptr, val
+        body_prefix.append(Dereference(val, ptr))  # val = *ptr
+
+    # Apply the replacements
+    body = body_prefix + as_list(Uxreplace(body_mapper).visit(iet.body))
```
FabioLuporini, line 42: extend

devito/symbolics/extended_dtypes.py
```diff
@@ -15,6 +18,62 @@
 }
 
 
+class NoDeclStruct(ctypes.Structure):
```
FabioLuporini, line 21: @ZoeLeibowitz I may be wrong but don't you need this type too? please see also where it's used and comment

devito/symbolics/extended_dtypes.py
```diff
@@ -15,6 +18,62 @@
...
+    """A ctypes.Structure that does not generate a struct definition"""
```
FabioLuporini, line 22: can you add a line explaining why it's needed ? basically for foreign types, but yeah, should be mentioned I think

devito/passes/iet/dtypes.py
```diff
@@ -1,14 +1,54 @@
...
+from devito.symbolics.extended_dtypes import Float16P
+from devito.tools.utils import as_list
+from devito.types.basic import Symbol
 
-__all__ = ['lower_complex']
+__all__ = ['lower_dtypes']
 
 
-def lower_complex(iet, lang, compiler):
+def lower_dtypes(iet, lang, compiler, sregistry):
```
FabioLuporini, line 14: how can this not be an `@iet_pass` and still work?

enwask: It's called from an `@iet_pass` [in `definitions` here](https://github.com/devitocodes/devito/blob/516b4ad0adc0f9eca18f5bd072a80051557cccf4/devito/passes/iet/definitions.py#L416)

devito/symbolics/printer.py
```diff
@@ -137,6 +144,9 @@ def _print_Pow(self, expr):
             if expr.exp == -1 and self.single_prec():
                 PREC = precedence(expr)
                 return '1.0F/%s' % self.parenthesize(expr.base, PREC)
+            if expr.exp == -1 and self.half_prec():
+                PREC = precedence(expr)
```
EdCaunt, line 148: You can pull `PREC = precedence(expr)` outside the `if` statements

EdCaunt (commented):
Needs tests

devito/passes/iet/dtypes.py
```diff
@@ -1,14 +1,54 @@
...
-    Add headers for complex arithmetic
+    Lowers float16 scalar types to pointers since we can't directly pass their
+    value. Also includes headers for complex arithmetic if needed.
+    """
+
+    iet, metadata = _complex_includes(iet, lang, compiler)
+
+    # Lower float16 parameters to pointers and dereference
+    body_prefix = []
+    body_mapper = {}
```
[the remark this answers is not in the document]
enwask: I think we do because the symbol in the parameters needs to be mapped to one with a different dtype but same name; in the body the same symbol has to be mapped to a new one with a different name (which we assign by the `Dereference`)

devito/ir/iet/nodes.py
```diff
@@ -1048,13 +1051,20 @@ def functions(self):
...
-            ret.extend(self.pointer.free_symbols)
+        ret = []
+        if self.pointer.is_Symbol:
+            assert issubclass(self.pointer._C_ctype, ctypes._Pointer), \
+                   "Scalar dereference must have a pointer ctype"
+            ret.extend([self.pointer._C_symbol, self.pointee._C_symbol])
         else:
-            ret.append(self.pointee._C_symbol)
+            ret.append(self.pointer.indexed)
+            if self.pointer.is_PointerArray or self.pointer.is_TempFunction:
```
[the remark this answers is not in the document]
enwask: Sure. Can do this with an `extend` to avoid the repetition as well

devito/passes/iet/dtypes.py
FabioLuporini: ah OK , I maybe have told Mathias already, but imho that thing goes straight into `operator/operator.py::Operator::_lower_iet`

FabioLuporini:
>  but still needs tests

seconded, this is crucial

enwask:
Yeah my plan was to write some more extensive tests for complex (since the existing ones didn't catch all of the break cases I've found) and add in a few for half. I want to get all of these changes out of the way first, and then there's a few more things to work out such as math functions being assigned to float16 arrays. Maybe this should be a draft actually?

devito/ir/iet/visitors.py
```diff
@@ -977,6 +982,7 @@ def _defines_aliases(n):
     rules = {
         'symbolics': lambda n: n.functions,
         'basics': lambda n: [i for i in n.expr_symbols if isinstance(i, Basic)],
+        'scalars': lambda n: [i for i in n.expr_symbols if isinstance(i, AbstractSymbol)],
```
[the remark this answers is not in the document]
enwask: Looks like no, some types seem to be missing the attribute

devito/passes/iet/dtypes.py
```diff
@@ -1,14 +1,54 @@
...
+    params_mapper = {}
+
+    # Lower scalar float16s to pointers and dereference them
+    for s in FindSymbols('scalars').visit(iet):
```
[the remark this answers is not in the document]
enwask: looks like not, since the actual types I'm mapping (e.g. `Constant`) aren't nodes they are only caught by `FindSymbols` if it's as a reference within some other expression, so we need to visit the body. That said I can probably make this marginally more efficient by making a set of parameters beforehand and checking membership that way

enwask:
Also worth noting that I'll need to rebase the complex branch later for the commit that dropped Python 3.9 support, as I'm using some of the 3.9 type hinting syntax. Hopefully won't be too bad

devito/symbolics/extended_dtypes.py
ZoeLeibowitz: We have different use cases because I need a local composite type which still generates a struct in the header file (i.e still produces a struct definition)

devito/passes/iet/dtypes.py
enwask: Ah okay, I can make this change

tests/test_operator.py
```diff
@@ -644,16 +644,18 @@ def test_tensor(self, func1):
     def test_complex(self, dtype):
```
EdCaunt, line 644: The introduction of an override is a good idea here.

I think this wants separating out into a couple of tests:
* Test that the correct expressions are generated
* Test that you can specify derivatives of complex fields
* Test that you can override complex values
* Test that you can mix complex and real symbols on the RHS

Note: do we also need to test that halo exchanges work with complex?

EdCaunt (commented):
Tests looking good. I think there should also be a test to ensure that the use of fp16/complex gets propagated into the variables introduced by CSE

tests/test_operator.py
enwask: Ok, will add some tests for those points.

Re: halo exchanges, I'll get back to you when I figure out what a halo exchange is

EdCaunt: A halo exchange is where each MPI rank swaps the data along its edges with its neighbours before advancing to the next timestep. It would probably worth checking that the MPI communication doesn't garble the data somehow when moving it between ranks. You would probably need some equations like:
```
from devito import Grid, TimeFunction, Eq, Operator

grid = Grid(shape=(4, 4))
x, y = grid.dimensions
t = grid.stepping_dim

v = TimeFunction(name='v', grid=grid)

eqns = [Eq(u[t+1, x, y], 1+sympy.I), Eq(v[t+1, x, y], sympy.I*u[t, x, y+1])]

op = Operator(eqns)
```

tests/test_dtypes.py
```diff
@@ -0,0 +1,202 @@
...
+    strings in generated code.
+    """
+
+    # Retrieve the language-specific type mapping
+    lang_types: dict[np.dtype, type] = _get_language(**kwargs).get('types')
+
+    # Set up an operator
+    grid = Grid(shape=(3, 3))
+    x: Dimension
+    y: Dimension
```
georgebisbas, line 68: These can be dropped later when ready to merge, I guess they are helpful for you now

tests/test_dtypes.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+def test_half_params() -> None:
+    """
+    Tests float16 input parameters: scalars should be lowered to pointers
+    and dereferenced; other parameters should keep the original dtype.
+    """
+
+    grid = Grid(shape=(5, 5), dtype=np.float16)
+    x: Dimension
```
georgebisbas, line 91: same

tests/test_operator.py
enwask: Ok. I think what I'll do is leave these general complex arithmetic tests where they are (in `test_operator` and `test_gpu_common`) and add these lower-level ones to the `test_dtypes` module. Except this halo exchange one goes in `test_mpi` I guess

devito/symbolics/extended_dtypes.py
```diff
@@ -15,6 +18,68 @@
...
+
+    @classmethod
+    def from_param(cls, val):
+        arr = np.array(val, dtype=np.float16)
+        return arr.ctypes.data_as(cls)
+
+
+class Float16P(np.float16):
+    """
+    Dummy dtype for a scalar float16 value that's been mapped to a pointer.
```
ZoeLeibowitz, line 75: that has been

tests/test_dtypes.py
```diff
@@ -0,0 +1,328 @@
...
+    # Check against expected result
+    dx, dy = grid.spacing_map.values()
+    xx, yy = np.meshgrid(np.linspace(0, 4, 5, dtype=dtype),
+                         np.linspace(0, 4, 5, dtype=dtype))
+    expected = xx * dx + yy * dy * dtype(0.0 + 1.0j)
+    assert np.allclose(g.data.T, expected)
+
+
+@pytest.mark.parametrize('dtype', [np.complex64, np.complex128])
+def test_complex_space_deriv(dtype: np.dtype[np.complexfloating]) -> None:
```
EdCaunt, line 302: Test derivatives on grids with complex spacing?

enwask: Hm? Is this a suggestion for rewriting the docstring? Or you mean that should be a test of its own? That's essentially what this case covers

EdCaunt: Oh yeah, must have misread it. Ignore

mloubout:
Merged into #2375

PR #2405, opened 2024-07-12
dsl: Remove counters for SubDomains and MultiSubDomains

EdCaunt:
Remove all counters used for `SubDomain`s and `SubDomainSet`s, instead dynamically renaming `SubDimension`s during compilation using a symbol registry.

Also ships some bugfixes and tests for `SubDomainSet`.

### DIFF
--- a/devito/ir/equations/algorithms.py
+++ b/devito/ir/equations/algorithms.py
@@ -202,6 +202,8 @@ def _(dim, sregistry, rebuilt):
         dimensions = list(dim.functions.dimensions)
         dimensions[0] = idim
 
+        # TODO: Requires a name change for some reason not to break things. Why?
+        # TODO: Dig into workings of ._rebuild() and .function
         f_name = sregistry.make_name(prefix=dim.functions.name)
         func = dim.functions._rebuild(name=f_name, dimensions=tuple(dimensions),
                                       halo=None, padding=None)
--- a/devito/ir/equations/algorithms.py
+++ b/devito/ir/equations/algorithms.py
@@ -204,6 +204,9 @@ def _(dim, sregistry, rebuilt):
 
         # TODO: Requires a name change for some reason not to break things. Why?
         # TODO: Dig into workings of ._rebuild() and .function
+        # NOTE: Rebuild is necessary here to break the link created by .function
+        # otherwise the old implicit dimension gets pulled by derive_parameters
+        # rather than the new one.
         f_name = sregistry.make_name(prefix=dim.functions.name)
         func = dim.functions._rebuild(name=f_name, dimensions=tuple(dimensions),
                                       halo=None, padding=None)
@@ -234,15 +237,14 @@ def rename_thicknesses(mapper, sregistry, rebuilt):
 def rebuild_cdims(expr, rebuilt):
     """
     Rebuild expression using ConditionalDimensions where their parent
-    dimension is a MultiSubDimension which has been rebuilt to have a unique
-    name.
+    dimension is a SubDimension which has been rebuilt to have a unique
+    name or unique thicknesses.
     """
     i_dims = expr.implicit_dims
     rebuilt_dims = []
     for d in i_dims:
         try:
             parent = rebuilt[d.parent]
-            # FIXME: Condition substitution should maybe be moved
             cond = d.condition.subs(rebuilt)
             rebuilt_dims.append(d._rebuild(parent=parent, condition=cond,
                                            factor=None))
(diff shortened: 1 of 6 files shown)


devito/passes/clusters/implicit.py
```diff
@@ -103,7 +106,13 @@ def callback(self, clusters, prefix):
                 processed.append(c)
                 continue
 
-            exprs, thickness = make_implicit_exprs(mapper, self.sregistry)
+            try:
```
mloubout, line 109: just an `if dim in thicknesses` should be fine not need for try

devito/ir/equations/algorithms.py
```diff
@@ -99,18 +102,30 @@ def lower_exprs(expressions, subs=None, **kwargs):
     --------
     f(x - 2*h_x, y) -> f[xi + 2, yi + 4]  (assuming halo_size=4)
     """
-    return _lower_exprs(expressions, subs or {})
+    return _lower_exprs(expressions, subs or {}, **kwargs)
 
 
-def _lower_exprs(expressions, subs):
+def _lower_exprs(expressions, subs, **kwargs):
+    # FIXME: Not sure why you can get this, but not access with index
```
mloubout, line 109: ?

devito/ir/equations/algorithms.py
```diff
@@ -99,18 +102,30 @@ def lower_exprs(expressions, subs=None, **kwargs):
...
+    sregistry = kwargs.get('sregistry')
```
mloubout, line 110: From below looks like it's mandatory not optional so need a valid default one not None

devito/ir/equations/algorithms.py
FabioLuporini: second

devito/ir/equations/algorithms.py
FabioLuporini: the following is OK:

```
def _lower_exprs(..., sregistry=None, **kwargs):
    <use sregistry>
```

devito/ir/equations/algorithms.py
```diff
@@ -99,18 +102,30 @@ def lower_exprs(expressions, subs=None, **kwargs):
...
     processed = []
+    rebuilt = {}
     for expr in as_tuple(expressions):
         try:
             dimension_map = expr.subdomain.dimension_map
         except AttributeError:
             # Some Relationals may be pure SymPy objects, thus lacking the subdomain
             dimension_map = {}
 
+        if sregistry and dimension_map:
```
FabioLuporini, line 121: sregistry cannot be none except for perhaps some low-level tests, in which case you may just pass a dummy one instead

devito/ir/equations/algorithms.py
```diff
@@ -99,18 +102,30 @@ def lower_exprs(expressions, subs=None, **kwargs):
...
+            # Give SubDimension thicknesses, SubDimensionSet functions and
+            # SubDomainSet implicit dimensions unique names
+            rename_thicknesses(dimension_map, sregistry, rebuilt)
+            # Rebuild ConditionalDimensions using rebuilt subdimensions
+            # The expression is then rebuilt with this ConditionalDimension
+            expr = rebuild_cdims(expr, rebuilt)
```
FabioLuporini, line 127: imho `rename_thicknesses` and `rebuild_cdims` should all be part of the same dispatch logic; you may use a ranking system (see https://github.com/devitocodes/devito/blob/master/devito/passes/iet/engine.py#L412-L415 for an example) to ensure certain objects are rebuilt before others.

nitpicking: `uniquify_symbols` sounds like a better name to me

devito/passes/clusters/implicit.py
```diff
@@ -92,6 +93,8 @@ def callback(self, clusters, prefix):
                 d = None
             if d is None:
                 processed.append(c)
+                # If no MultiSubDomain present in this cluster, then tip should be reset
+                tip = None
```
FabioLuporini, line 97: with `thicknesses` now in place, do we actually still need this `tip` thing?

EdCaunt: Yeah, it keeps track of whether consecutive expressions are on the same SubDomainSet, and avoids generating additional unnecessary thickness expressions in this case. However, it did not previously reset if there was an expression without a SubDomainSet in the middle, which meant that you could create cases where the thickness expressions were erroneously omitted.

devito/ir/equations/algorithms.py
```diff
@@ -146,3 +161,93 @@ def _lower_exprs(expressions, subs):
...
+
+    except KeyError:
+        idim_name = sregistry.make_name(prefix=dim.implicit_dimension.name)
+        idim = dim.implicit_dimension._rebuild(name=idim_name)
+        rebuilt[dim.implicit_dimension] = idim
+
+        dimensions = list(dim.functions.dimensions)
+        dimensions[0] = idim
+
+        # TODO: Requires a name change for some reason not to break things. Why?
```
FabioLuporini, line 205: there's a lot of TODOs here

EdCaunt:
Superseded by #2431

PR #2408, opened 2024-07-15
compiler: Patch bug in SubDomainSet lowering

EdCaunt:
Separated from #2405, since I think I need to go back to the drawing board on that one, but this patch fixes a bug currently in master.


FabioLuporini (approved):
uncontroversial

mloubout (approved):

FabioLuporini (approved):

PR #2409, opened 2024-07-15
compiler: Hack in thickness

FabioLuporini:
This is beyond horrible, but it's gonna fixed in #2405


EdCaunt (commented):
I think it's as fine as it's going to get, but you should add a test so we can be certain this issue is fixed when I revert the change

FabioLuporini:
Closing as superseded by #2411

PR #2410, opened 2024-07-15
compiler: fix cse with different conditionals

mloubout:
Fix cse to make sure expression from different conditionals get properly handled.

avoids things like

```
if (a){
  f1 = expr
 }
if (b){
  f1 = expr
 }
```

to be converted to

```
if (a){
  r0 = expr
  f1 = r0
 }
if (b){
  f1 = r0
 }
```

CSE now counts expression based on the equation's conditional. We could also just lift all `rxx` out of conditionals but that would be computing those cse temps way more than needed which I don't think it great

### DIFF
--- a/devito/passes/clusters/cse.py
+++ b/devito/passes/clusters/cse.py
@@ -1,6 +1,7 @@
-from collections import Counter, OrderedDict
+from collections import Counter, OrderedDict, namedtuple
 from functools import singledispatch
 
+import sympy
 from sympy import Add, Function, Indexed, Mul, Pow
 try:
     from sympy.core.core import ordering_of_classes
@@ -9,14 +10,17 @@ except ImportError:
     from sympy.core.basic import ordering_of_classes
 
 from devito.finite_differences.differentiable import IndexDerivative
-from devito.ir import Cluster, Scope, cluster_pass, ClusterizedEq
+from devito.ir import Cluster, Scope, cluster_pass
 from devito.passes.clusters.utils import makeit_ssa
 from devito.symbolics import estimate_cost, q_leaf
 from devito.symbolics.manipulation import _uxreplace
-from devito.tools import as_list
+from devito.tools import as_list, frozendict
 from devito.types import Eq, Symbol, Temp
 
 __all__ = ['cse']
+
+
+Counted = namedtuple('Candidate', 'expr, conditionals')
 
 
 class CTemp(Temp):
@@ -90,14 +94,12 @@ def _cse(maybe_exprs, make, min_cost=1, mode='default'):
 
     while True:
         # Detect redundancies
-        counted = count(processed, None).items()
-        targets = OrderedDict([(k, estimate_cost(k[0], True))
+        counted = count(processed).items()
+        targets = OrderedDict([(k, estimate_cost(k.expr, True))
                                for k, v in counted if v > 1])
-
         # Rule out Dimension-independent data dependencies
         targets = OrderedDict([(k, v) for k, v in targets.items()
-                               if not k[0].free_symbols & exclude])
-
+                               if not k.expr.free_symbols & exclude])
         if not targets or max(targets.values()) < min_cost:
             break
 
@@ -112,13 +114,12 @@ def _cse(maybe_exprs, make, min_cost=1, mode='default'):
         updated = []
         for e in processed:
             pe = e
-            pe_c = e.conditionals
-            for (k, c), v in chosen:
-                if not c == pe_c:
+            for k, v in chosen:
+                if not k.conditionals == e.conditionals:
                     continue
-                pe, changed = _uxreplace(pe, {k: v})
+                pe, changed = _uxreplace(pe, {k.expr: v})
                 if changed and v not in scheduled:
-                    updated.append(pe.func(v, k, operation=None))
+                    updated.append(pe.func(v, k.expr, operation=None))
                     scheduled.append(v)
             updated.append(pe)
         processed = updated
@@ -160,34 +161,39 @@ def _compact_temporaries(exprs, exclude):
 
 
 @singledispatch
-def count(expr, conds):
+def count(expr):
     """
     Construct a mapper `expr -> #occurrences` for each sub-expression in `expr`.
     """
     mapper = Counter()
     for a in expr.args:
-        mapper.update(count(a, None))
+        mapper.update(count(a))
     return mapper
 
 
 @count.register(list)
 @count.register(tuple)
-def _(exprs, conds):
+def _(exprs):
     mapper = Counter()
     for e in exprs:
-        mapper.update(count(e, None))
+        mapper.update(count(e))
+
     return mapper
 
 
-@count.register(ClusterizedEq)
-def _(exprs, conds):
-    conditionals = exprs.conditionals
-    return count(exprs.rhs, conditionals)
+@count.register(sympy.Eq)
+def _(expr):
+    mapper = count(expr.rhs)
+    try:
+        cond = expr.conditionals
+    except AttributeError:
+        cond = frozendict()
+    return {Counted(e, cond): v for e, v in mapper.items()}
 
 
 @count.register(Indexed)
 @count.register(Symbol)
-def _(expr, conds):
+def _(expr):
     """
     Handler for objects preventing CSE to propagate through their arguments.
     """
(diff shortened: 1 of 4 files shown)


devito/passes/clusters/cse.py
```diff
@@ -90,12 +90,13 @@ def _cse(maybe_exprs, make, min_cost=1, mode='default'):
 
     while True:
         # Detect redundancies
-        counted = count(processed).items()
-        targets = OrderedDict([(k, estimate_cost(k, True)) for k, v in counted if v > 1])
+        counted = count(processed, None).items()
+        targets = OrderedDict([(k, estimate_cost(k[0], True))
```
georgebisbas, line 94: why k[0] specifically?

tests/test_dse.py
```diff
@@ -2734,7 +2783,7 @@ def test_fullopt(self):
 
         assert summary1[('section0', None)].ops == 31
         assert summary1[('section1', None)].ops == 88
-        assert summary1[('section2', None)].ops == 25
+        assert summary1[('section2', None)].ops == 28
```
FabioLuporini, line 2786: what's this change due to?

devito/passes/clusters/cse.py
```diff
@@ -156,53 +160,59 @@ def _compact_temporaries(exprs, exclude):
...
     mapper = Counter()
     for e in exprs:
-        mapper.update(count(e))
+        mapper.update(count(e, None))
     return mapper
 
 
+@count.register(ClusterizedEq)
+def _(exprs, conds):
+    conditionals = exprs.conditionals
```
FabioLuporini, line 184: useless assignment

devito/passes/clusters/cse.py
```diff
@@ -156,53 +160,59 @@ def _compact_temporaries(exprs, exclude):
...
+def _(exprs, conds):
     mapper = Counter()
     for e in exprs:
-        mapper.update(count(e))
+        mapper.update(count(e, None))
     return mapper
 
 
+@count.register(ClusterizedEq)
+def _(exprs, conds):
```
FabioLuporini, line 183: expr, not exprs

devito/passes/clusters/cse.py
FabioLuporini: see the `count` output

but yes, we should use a namedtuple or a Bunch, see my other comments below

devito/passes/clusters/cse.py
```diff
@@ -156,53 +160,59 @@ def _compact_temporaries(exprs, exclude):
...
+    conditionals = exprs.conditionals
+    return count(exprs.rhs, conditionals)
```
FabioLuporini, line 185: I'll write it here, but it applies to the whole dispatching logic.

I don't think you need to propagate `conds` through the various `count` recursive calls.

In particular here `count` returns a dict; all I think you have to do here is:

```
mapper = count(expr.rhs)
if expr.conditionals:
     mapper = {(k, expr.conditionals): v for k, v in mapper.items()}
return mapper
```

To be more precise, we shouldn't just use a 2-tuple as key, but rather either a namedtuple or a Bunch, so that the keys are homogeneous and you don't have to do `k[0]` at the call site

devito/passes/clusters/cse.py
```diff
@@ -111,7 +112,10 @@ def _cse(maybe_exprs, make, min_cost=1, mode='default'):
         updated = []
         for e in processed:
             pe = e
-            for k, v in chosen:
+            pe_c = e.conditionals
```
FabioLuporini, line 115: `pe` changes, `pe_c` doesn't, so you don't need an assignment here

following on from the other comments below, here I'd have:

```
for k, v in chosen:
    if k.conditionals != e.conditionals:
         continue
    ...
```

devito/passes/clusters/cse.py
FabioLuporini: `namedtuple` is probably OK, but you could also introduce a `Candidate` class such that `conditionals` in an optional argument; or preset the `namedtuple` so that `conditionals` defaults to None.

I don't know if one such `Candidate` class would help elsewhere, for example we could enrich it with properties such as `estimate` and `free_symbols` to decentralize complexity, though admittedly this is getting close to nitpicking

Anyway, I think it's good to keep the complexity of CSE as under control as possible

devito/passes/clusters/cse.py
```diff
@@ -90,12 +90,13 @@ def _cse(maybe_exprs, make, min_cost=1, mode='default'):
...
-        counted = count(processed).items()
-        targets = OrderedDict([(k, estimate_cost(k, True)) for k, v in counted if v > 1])
+        counted = count(processed, None).items()
+        targets = OrderedDict([(k, estimate_cost(k[0], True))
+                               for k, v in counted if v > 1])
 
         # Rule out Dimension-independent data dependencies
         targets = OrderedDict([(k, v) for k, v in targets.items()
-                               if not k.free_symbols & exclude])
+                               if not k[0].free_symbols & exclude])
```
EdCaunt, line 99: Nitpick: we use `and` elsewhere in the codebase

devito/passes/clusters/cse.py
```diff
@@ -156,53 +160,59 @@ def _compact_temporaries(exprs, exclude):
...
 @singledispatch
-def count(expr):
+def count(expr, conds):
     """
     Construct a mapper `expr -> #occurrences` for each sub-expression in `expr`.
     """
     mapper = Counter()
     for a in expr.args:
-        mapper.update(count(a))
+        mapper.update(count(a, None))
```
EdCaunt, line 169: Is it possible to one-line these loops? `Counter(count(a, None) for a in expr.args)`

devito/passes/clusters/cse.py
mloubout:
```
mapper = count(expr.rhs)
if expr.conditionals:
     mapper = 
```

Thqt would require a by hand merge for list/..., propagating conditionals makes the Counter update much easier and simpler

I'm fine with a `namedtuple` and will fix, `Candidate` class would have to bee too intricated to handle different use case but long term can be looked into.

devito/passes/clusters/cse.py
mloubout: > , so you don't need an assignment here

Just habit to assign if resused but yes not really necessary

devito/passes/clusters/cse.py
mloubout: No that's not the same, `&` is the set intersection not the logical and that returns something different

tests/test_dse.py
mloubout: Tt goes from

```
float r4 = floorf(r7);
int posx = (int)r4;
float px = -r4 + r7;
```

To 

```
int posx = (int)(floorf(r4));
float px = r4 - floorf(r4);
```

for the interpolation section, don't think much of a difference but will double check

devito/passes/clusters/cse.py
EdCaunt: Fair, I hadn't spotted they were sets

devito/passes/clusters/cse.py
```diff
@@ -172,9 +176,20 @@ def _(exprs):
     mapper = Counter()
     for e in exprs:
         mapper.update(count(e))
+
     return mapper
 
 
+@count.register(Eqs)
```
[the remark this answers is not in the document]
FabioLuporini: with this in place, imho the PR gets neat enough to be merged

devito/passes/clusters/cse.py
```diff
@@ -1,7 +1,7 @@
-from collections import Counter, OrderedDict
+from collections import Counter, OrderedDict, namedtuple
 from functools import singledispatch
 
-from sympy import Add, Function, Indexed, Mul, Pow
+from sympy import Add, Function, Indexed, Mul, Pow, Eq as Eqs
```
FabioLuporini, line 4: I would just `import sympy` and then use `sympy.Eq`. The explicit way might be more verbose, but it makes the code easier to understand

mloubout: done

devito/passes/clusters/cse.py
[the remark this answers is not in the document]
FabioLuporini: single dispatch prioritizes subclasses over classes. It traverses the class hierarchy and pick the first handler starting from the bottom, so a `devito.Eq` would only enter its own handler

anyway, nitpicking, I'm fine the way it is

PR #2412, opened 2024-07-17
compiler: Bump default block size

FabioLuporini:
It's virtually always faster

### DIFF
--- a/devito/types/dimension.py
+++ b/devito/types/dimension.py
@@ -1307,10 +1307,8 @@ class BlockDimension(AbstractIncrDimension):
             return ()
 
     def _arg_defaults(self, **kwargs):
-        # TODO: need a heuristic to pick a default incr size
-        # TODO: move default value to __new__
         try:
-            return {self.step.name: 8}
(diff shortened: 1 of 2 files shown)


georgebisbas (approved):

devito/types/dimension.py
```diff
@@ -1223,10 +1223,8 @@ def _arg_names(self):
             return ()
 
     def _arg_defaults(self, **kwargs):
-        # TODO: need a heuristic to pick a default incr size
-        # TODO: move default value to __new__
         try:
-            return {self.step.name: 8}
+            return {self.step.name: 16}
```
mloubout, line 1227: nitpicking: Can we add a "TODO" to check the args for space order and have some empirical `2*space_order` or something?

mloubout (approved):

devito/types/dimension.py
FabioLuporini: doing in a sec in upcoming PR, merging this one now

PR #2413, opened 2024-07-17
api: Support combination of condition and factor for ConditionalDimension

### LINKED ISSUE #2273: Conflict in time_M when using saved buffer and conditional dimension
When using a conditional dimension that contains `Ge` / `Le` types of conditions and a saved buffer defined over that conditional dimension, there is a conflict in the calculation of `time_M` for different objects:

```
from devito import *
from devito.symbolics import CondEq


grid = Grid(shape=(10, 10))
time = grid.time_dim

nt = 200
bounds = (10, 100)
factor = 5

condition = sympy.And(CondEq(time % 7, 0),
                      Ge(time, bounds[0]),
                      Le(time, bounds[1]), )

time_under = ConditionalDimension(name='timeu',
                                  parent=time,
                                  factor=7,
                                  condition=condition)

buffer_size = (bounds[1] - bounds[0] + factor) // factor + 1

rec = SparseTimeFunction(name='rec', grid=grid, npoint=1, nt=nt)

u = TimeFunction(name='u', grid=grid, space_order=2)
usaved = TimeFunction(name='usaved', grid=grid, space_order=2, time_dim=time_under, save=buffer_size)

eq = [Eq(usaved, u)] + rec.interpolate(u)

op = Operator(eq)
op(time_m=1, time_M=nt-1, usaved=usaved, rec=rec)
```
produces the error `ValueError: Unable to find unique value for key time_M, candidates: [199, range(1, 139)]`.

mloubout:
Fixes #2273

### DIFF
--- a/devito/ir/equations/equation.py
+++ b/devito/ir/equations/equation.py
@@ -191,9 +191,12 @@ class LoweredEq(IREq):
             if d.condition is None:
                 conditionals[d] = GuardFactor(d)
             else:
-                conditionals[d] = diff2sympy(lower_exprs(d.condition))
+                cond = diff2sympy(lower_exprs(d.condition))
+                if d.factor is not None:
+                    cond = sympy.And(cond, GuardFactor(d))
+                conditionals[d] = cond
             if d.factor is not None:
-                expr = uxreplace(expr, {d: IntDiv(d.index, d.factor)})
+                expr = uxreplace(expr, {d: IntDiv(d.fact_index, d.factor)})
         conditionals = frozendict(conditionals)
 
         # Lower all Differentiable operations into SymPy operations
--- a/devito/ir/equations/equation.py
+++ b/devito/ir/equations/equation.py
@@ -8,7 +8,7 @@ from devito.ir.support import (GuardFactor, Interval, IntervalGroup, IterationSp
                                Stencil, detect_io, detect_accesses)
 from devito.symbolics import IntDiv, uxreplace
 from devito.tools import Pickable, Tag, frozendict
-from devito.types import Eq, Inc, ReduceMax, ReduceMin
+from devito.types import Eq, Inc, ReduceMax, ReduceMin, relational_min
 
 __all__ = ['LoweredEq', 'ClusterizedEq', 'DummyEq', 'OpInc', 'OpMin', 'OpMax']
 
@@ -192,11 +192,15 @@ class LoweredEq(IREq):
                 conditionals[d] = GuardFactor(d)
             else:
                 cond = diff2sympy(lower_exprs(d.condition))
-                if d.factor is not None:
+                if d._factor is not None:
                     cond = sympy.And(cond, GuardFactor(d))
                 conditionals[d] = cond
-            if d.factor is not None:
-                expr = uxreplace(expr, {d: IntDiv(d.fact_index, d.factor)})
+            # Replace dimension with index
+            index = d.index
+            if d.condition is not None:
+                index = index - relational_min(d.condition, d.parent)
+            expr = uxreplace(expr, {d: IntDiv(index, d.factor)})
+
         conditionals = frozendict(conditionals)
 
         # Lower all Differentiable operations into SymPy operations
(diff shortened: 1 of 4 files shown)


devito/types/dimension.py
```diff
@@ -892,6 +895,18 @@ def condition(self):
...
+    @property
+    def fact_index(self):
+        if self.condition is None or self._factor is None:
+            return self.index
+
+        # This is the corner case where both a condition and a factor are provided
+        # the index will need to be `self.parent - min(self.condition)` to avoid
+        # shifted indexing. E.g if you have `factor=2` and `condition=Ge(time, 10)`
+        # then the lowered index needs to be `(time - 10)/ 2`
+        ltkn = relational_min(self.condition, self.parent)
```
FabioLuporini, line 907: ltkn stands for "left thickness" -- is it really the name u wanna use here?

(nitpicking)

devito/types/dimension.py
```diff
@@ -927,7 +942,7 @@ def _arg_defaults(self, _min=None, size=None, alias=None):
         # `factor` endpoints are legal, so we return them all. It's then
         # up to the caller to decide which one to pick upon reduction
         dim = alias or self
-        if dim._factor is None or size is None:
+        if dim.condition is not None or size is None:
```
FabioLuporini, line 945: so we don't check _factor anymore here?

devito/types/dimension.py
mloubout: Yeah that's intentional that the actual left thickness that is required to be shifted by.

devito/types/dimension.py
mloubout: Yeah, basically if there is a condition then you can't safely use the factor to infer arg bounds

devito/types/dimension.py
```diff
@@ -892,6 +895,18 @@ def condition(self):
     def indirect(self):
         return self._indirect
 
+    @property
+    def fact_index(self):
```
FabioLuporini, line 899: can you not move this whole logic into `__init_finalize__` and set index to the proper value from the get go?

devito/ir/equations/equation.py
```diff
@@ -188,12 +188,19 @@ def __new__(cls, *args, **kwargs):
         for d in ordering:
             if not d.is_Conditional:
                 continue
-            if d.condition is None:
+            if d._factor is not None:
```
FabioLuporini, line 191: why not `d.factor`?

FabioLuporini (approved):

devito/ir/equations/equation.py
mloubout: `d.factor` is never None it returns `1` if there is no factor for various reasons

devito/types/relational.py
```diff
@@ -208,3 +209,44 @@ def __new__(cls, lhs, rhs=0, subdomain=None, **kwargs):
 
 ops = {Ge: Lt, Gt: Le, Le: Gt, Lt: Ge}
 rev = {Ge: Le, Gt: Lt, Lt: Gt, Le: Ge}
+
+
+def relational_min(expr, s):
```
georgebisbas, line 214: Any chance you may reuse:

`def evalrel(func=min, input=None, assumptions=None):`

have not looks closely but maybe we can find some common way

devito/types/relational.py
```diff
@@ -208,3 +209,44 @@ def __new__(cls, lhs, rhs=0, subdomain=None, **kwargs):
...
+    """
+    Infer the minimum valid value for symbol `s` in the expression `expr`.
+    For example
+        - if `expr` is `s < 10`, then the minimum valid value for `s` is 0
+        - if `expr` is `s >= 10`, then the minimum valid value for `s` is 10
```
georgebisbas, line 219: or at least this rang a bell when reading the docstring here

devito/types/relational.py
mloubout: No `evalrel` simplifies/... a set of relations, this does not evaluate the condition to some minimal case, it finds the smallest. value for symbol `s` that satisfies `expr` in order to get a index bound

georgebisbas (approved):

PR #2415, opened 2024-07-18
arch: Support more platforms

mloubout:
Generalizes gcc
suport graviton 2-3-4 and add cflags to use gcc with it as well

### DIFF
--- a/devito/arch/compiler.py
+++ b/devito/arch/compiler.py
@@ -434,10 +434,8 @@ class GNUCompiler(Compiler):
         if platform in [POWER8, POWER9]:
             # -march isn't supported on power architectures, is -mtune needed?
             self.cflags = ['-mcpu=native'] + self.cflags
-        elif platform is Graviton:
-            # Graviton flag
-            mx = platform.march
-            self.cflags = ['-mcpu=%s' % mx] + self.cflags
+        elif isinstance(platform, Graviton):
+            self.cflags = ['-mcpu=%s' % platform.march] + self.cflags
         else:
             self.cflags = ['-march=native'] + self.cflags
 
@@ -466,9 +464,8 @@ class ArmCompiler(GNUCompiler):
         platform = kwargs.pop('platform', configuration['platform'])
 
         # Graviton flag
-        if platform is Graviton:
-            mx = platform.march
-            self.cflags += ['-mcpu=%s' % mx]
+        if isinstance(platform, Graviton):
+            self.cflags += ['-mcpu=%s' % platform.march]
 
 
 class ClangCompiler(Compiler):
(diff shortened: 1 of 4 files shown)


devito/arch/compiler.py
```diff
@@ -434,6 +434,10 @@ def __init_finalize__(self, **kwargs):
         if platform in [POWER8, POWER9]:
             # -march isn't supported on power architectures, is -mtune needed?
             self.cflags = ['-mcpu=native'] + self.cflags
+        elif platform is Graviton:
```
FabioLuporini, line 437: it shuold be `isinstance(platform, Graviton)` no?

devito/arch/compiler.py
```diff
@@ -434,6 +434,10 @@ def __init_finalize__(self, **kwargs):
...
+            # Graviton flag
```
FabioLuporini, line 438: Drop useless comment

devito/arch/compiler.py
```diff
@@ -434,6 +434,10 @@ def __init_finalize__(self, **kwargs):
...
+            mx = platform.march
```
FabioLuporini, line 439: probably also inlinable

devito/arch/compiler.py
```diff
@@ -462,8 +466,9 @@ def __init_finalize__(self, **kwargs):
         platform = kwargs.pop('platform', configuration['platform'])
 
         # Graviton flag
-        if platform is GRAVITON:
-            self.cflags += ['-mcpu=neoverse-n1']
+        if platform is Graviton:
```
FabioLuporini, line 469: `isinstance(platform, Graviton)`

devito/arch/compiler.py
```diff
@@ -962,7 +967,26 @@ def __new_with__(self, **kwargs):
...
+    """
+
+    def __getitem__(self, key):
+        if key.startswith('gcc-'):
+            i = key.split('-')[1]
+            return partial(GNUCompiler, suffix=i)
+        return super().__getitem__(key)
+
+    def has_key(self, k):
+        return k in self.keys() or k.startswith('gcc-')
```
FabioLuporini, line 983: no need for "()" after `keys`

FabioLuporini (changes requested):

FabioLuporini:
I've also changed the PR title to start with "arch: ..." rather than "compiler: ..."

georgebisbas (approved):

ggorman:
Reinstating Graviton runner so this can actually be tested.

ggorman (approved):

devito/arch/compiler.py
mloubout: Fixe

devito/arch/compiler.py
mloubout: Yes you do

FabioLuporini (approved):
GTG now

PR #2416, opened 2024-07-19
tests: Add test and patch for issue 2325

### LINKED ISSUE #2325: Rebuilding a `ConditionalDimension` with no factor adds a factor.
If a `ConditionalDimension` is created with `factor=None`, then `self._factor` is set to `None`, but `self.factor` will return 1. As such, when rebuilding a `ConditionalDimension` with no factor, a factor of 1 is added unexpectedly, changing the generated code.

EdCaunt:
Fixes #2325.

### DIFF
--- a/devito/types/dimension.py
+++ b/devito/types/dimension.py
@@ -966,6 +966,10 @@ class ConditionalDimension(DerivedDimension):
 
     @property
     def factor(self):
+        return self._factor
+
+    @property
+    def factor_default(self):
         return self._factor if self._factor is not None else 1
 
     @property
@@ -982,7 +986,7 @@ class ConditionalDimension(DerivedDimension):
         if self.condition is not None:
             retval |= self.condition.free_symbols
         try:
-            retval |= self.factor.free_symbols
+            retval |= self.factor_default.free_symbols
         except AttributeError:
             pass
         return retval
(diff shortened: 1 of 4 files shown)


mloubout (changes requested):
I don't really see what's the issue there and why we would need two separate properties. 
If the factor is None it will rebuild with factor=1 that is treated as None so it will rebuild correctly so I don't see where the issue is

mloubout:
I don't see anything wrong on master

```

In [1]: from devito import *

In [2]: x = Dimension(name='x')
   ...: cd = ConditionalDimension(name='cd', parent=x, condition=Ge(x, 1))

In [3]: cd._rebuild()._factor

In [4]: cd._rebuild().factor
Out[4]: 1

In [5]: cd.factor
Out[5]: 1

In [6]: cd._factor


```

devito/types/dimension.py
```diff
@@ -966,6 +966,10 @@ def spacing(self):
 
     @property
     def factor(self):
+        return self._factor
```
georgebisbas, line 969: Cant this be `return self._factor or 1`
and drop `def factor_default` ?

EdCaunt:
Looks like this was actually fixed in a separate patch (blame says @mloubout fixed it 5 days ago). Closing.

PR #2417, opened 2024-07-19
compiler: Fix parlang reductions over >= 4 loops

FabioLuporini:


devito/ir/clusters/algorithms.py
```diff
@@ -594,31 +591,49 @@ def normalize_reductions_minmax(cluster):
...
+    """
+    # The candidate Dimensions along which to perform the map part
+    candidates = [d for d in cluster.ispace.itdims
+                  if cluster.properties.is_parallel_atomic(d)]
+    if not candidates:
         return cluster
 
+    # If there are more parallel dimensions than the maximum allowed by the
+    # target platform, we must restrain the number of candidates
+    max_par_dims = platform.limits()['max-par-dims']
```
mloubout, line 626: why not just `platform.max-par-dims` as a property?

devito/arch/archinfo.py
```diff
@@ -664,6 +664,16 @@ def max_mem_trans_size(self, dtype):
         assert self.max_mem_trans_nbytes % np.dtype(dtype).itemsize == 0
         return int(self.max_mem_trans_nbytes / np.dtype(dtype).itemsize)
 
+    def limits(self, compiler=None, language=None):
+        """
+        Return the architecture-specific limits for the given compiler and
+        language.
+        """
+        return {
+            'max-par-dims': np.inf,
```
mloubout, line 673: I think `np.inf` might be problematic if used for indexing

georgebisbas (commented):
Nothing to report, apart from fixing the failing tests.
Just OOC, how did this came up as motivation, are people doing any fancy parallelizing stuff for over 4 loops?

devito/arch/archinfo.py
```diff
@@ -664,6 +664,16 @@ def max_mem_trans_size(self, dtype):
         assert self.max_mem_trans_nbytes % np.dtype(dtype).itemsize == 0
         return int(self.max_mem_trans_nbytes / np.dtype(dtype).itemsize)
 
+    def limits(self, compiler=None, language=None):
```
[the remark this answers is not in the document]
FabioLuporini: they will in the future yeah, some limits may be compiler and/or language dependent

devito/ir/clusters/algorithms.py
FabioLuporini: because (see comment above) some of these limits may be e.g. programming-model dependent, besides being architecture-dependent

devito/arch/archinfo.py
FabioLuporini: and indeed one test is failing because of this :D

EdCaunt (approved):

devito/ir/clusters/algorithms.py
mloubout: The it's the same as having the compiler store it's complex/f16 dtype it doesn't belong there and should be part of a pass with if/dispatch

PR #2418, opened 2024-07-19
misc: Fix EnrichedTuple rebuilding

enwask:
Fixes `EnrichedTuple` (and its hierarchy) rebuilding to an empty instance.

### DIFF
--- a/devito/tools/data_structures.py
+++ b/devito/tools/data_structures.py
@@ -43,11 +43,22 @@ class EnrichedTuple(tuple, Pickable):
     A tuple with an arbitrary number of additional attributes.
     """
 
+    __rargs__ = ('*items',)
+    __rkwargs__ = ('getters',)
+
     def __new__(cls, *items, getters=None, **kwargs):
         obj = super().__new__(cls, items)
         obj.__dict__.update(kwargs)
-        obj._getters = OrderedDict(zip(getters or [], items))
+        # Convert to list if we're getting an OrderedDict from rebuild
+        obj.getters = OrderedDict(zip(list(getters or []), items))
         return obj
+
+    def _rebuild(self, *args, **kwargs):
+        # Need to explicitly apply any additional attributes
+        _kwargs = dict(self.__dict__)
+        _kwargs.update(**kwargs)
+
+        return super()._rebuild(*args, **_kwargs)
 
     def __getitem__(self, key):
         if isinstance(key, int):
@@ -60,21 +71,26 @@ class EnrichedTuple(tuple, Pickable):
             start = key.start or 0
             stop = key.stop if key.stop is not None else len(self)
             kwargs = dict(self.__dict__)
-            kwargs['getters'] = list(self._getters)[start:stop]
+            kwargs['getters'] = list(self.getters)[start:stop]
             return EnrichedTuple(*items, **kwargs)
         else:
             return self.__getitem_hook__(key)
 
     def __getitem_hook__(self, key):
-        return self._getters[key]
+        return self.getters[key]
 
     def __getnewargs_ex__(self):
         # Bypass default reconstruction logic since this class spawns
         # objects with varying number of attributes
-        return (tuple(self), dict(self.__dict__))
+        return tuple(self), dict(self.__dict__)
 
     def get(self, key, val=None):
-        return self._getters.get(key, val)
+        return self.getters.get(key, val)
+
+    @property
+    def items(self) -> tuple:
+        # Needed for rargs
+        return tuple(self)
 
 
 class ReducerMap(MultiDict):
--- a/devito/types/utils.py
+++ b/devito/types/utils.py
@@ -25,9 +25,9 @@ CELL = Stagger('cell')
 class DimensionTuple(EnrichedTuple):
 
     def __getitem_hook__(self, dim):
-        for d in self._getters:
+        for d in self.getters:
             if d._defines & dim._defines:
-                return self._getters[d]
+                return self.getters[d]
         raise KeyError
 
 
--- a/devito/finite_differences/derivative.py
+++ b/devito/finite_differences/derivative.py
@@ -222,7 +222,7 @@ class Derivative(sympy.Derivative, Differentiable, Reconstructable):
             raise TypeError("Side only supported for first order single"
                             "Dimension derivative such as `.dxl` or .dx(side=left)")
         # Cross derivative
-        _fd_order = dict(self.fd_order._getters)
+        _fd_order = dict(self.fd_order.getters)
         try:
             _fd_order.update(fd_order or {})
             _fd_order = tuple(_fd_order.values())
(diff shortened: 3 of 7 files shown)


devito/tools/data_structures.py
```diff
@@ -43,10 +43,15 @@ class EnrichedTuple(tuple, Pickable):
     A tuple with an arbitrary number of additional attributes.
     """
 
+    __rargs__ = ('*values',)
```
mloubout, line 46: keep consistent with signature so rename as items or change to values in __new__

devito/tools/data_structures.py
```diff
@@ -43,10 +43,15 @@ class EnrichedTuple(tuple, Pickable):
...
+    __rkwargs__ = ('getters',)
+
     def __new__(cls, *items, getters=None, **kwargs):
         obj = super().__new__(cls, items)
         obj.__dict__.update(kwargs)
-        obj._getters = OrderedDict(zip(getters or [], items))
+        if getters is not None or '_getters' not in kwargs:
```
mloubout, line 52: can `_getters

mloubout: Also this will be problematic if `getters` is `None` because `obj._getters` will not exist, that's what the `getters or ()` is for so that `obj._getters` always exists.

devito/tools/data_structures.py
FabioLuporini: seconded, this should be called "items" though in practice it might work the same since it's prefixed with "*"

devito/tools/data_structures.py
FabioLuporini: I don't see why adding an `if` would be necessary?

devito/tools/data_structures.py
```diff
@@ -76,6 +81,16 @@ def __getnewargs_ex__(self):
     def get(self, key, val=None):
         return self._getters.get(key, val)
 
+    @property
+    def values(self) -> tuple:
```
FabioLuporini, line 85: "items"

devito/tools/data_structures.py
enwask: My reasoning was that otherwise, if reconstructed by something other than `Reconstructable._rebuild` (I believe there's some pickling system that uses `__getnewargs_ex__` instead, no?) then `kwargs['_getters']` will have the old value to be unpickled, but `getters` will equal `None`. Then if we don't do this check, the constructor will overwrite whatever the old value obtained from unpicking was, and instead overwrite with an empty dict

enwask: Wait that was my old reasoning, I forgot I changed the way I did this so that's not accurate anymore. I'll take a look again in a minute

enwask: Well, have found at least one case of extra kwargs being used `left` and `right` passed to a `DimensionTuple` constructor from `AbstractFunction`. I feel like in the long run that really be specialized rather than abusing the object `__dict__` this way but I'll leave it for now. 

Which means that we do need this check still, I think

devito/types/utils.py
```diff
@@ -24,6 +24,12 @@ class Stagger(Tag):
 
 class DimensionTuple(EnrichedTuple):
 
+    __rkwargs__ = EnrichedTuple.__rkwargs__ + ('left', 'right')
+
+    def __new__(cls, *args, getters=None, left=None, right=None, **kwargs):
```
FabioLuporini, line 29: if we really have to add this `__new__`, which basically doesn't add anything to the super()'s `__new__`, then it looks odd

devito/tools/data_structures.py
```diff
@@ -71,11 +73,24 @@ def __getitem_hook__(self, key):
     def __getnewargs_ex__(self):
         # Bypass default reconstruction logic since this class spawns
         # objects with varying number of attributes
-        return (tuple(self), dict(self.__dict__))
+        kwargs = dict(self.__dict__)
+        kwargs['getters'] = list(kwargs.pop('_getters', []))
```
mloubout, line 77: why  not just `kwargs['getters'] =  list(self.getters)` ?

devito/types/utils.py
enwask: Doesn't need to be added to `__new__` so those parameters are explicit, but it's not strictly necessary since that's covered by `self.__dict__.assign(...` in the parent. I can remove the parameters here if preferred

devito/types/utils.py
```diff
@@ -24,6 +24,8 @@ class Stagger(Tag):
 
 class DimensionTuple(EnrichedTuple):
 
+    __rkwargs__ = EnrichedTuple.__rkwargs__ + ('left', 'right')
```
mloubout, line 27: Can add a comment why this needed?

mloubout (approved):

devito/types/utils.py
FabioLuporini: if really needed, this would be a hack, which we cannot merge

FabioLuporini:
could you also rebase the commit history as a single commit?

let me know if you need help with rebasing

enwask:
> if really needed, it would be a hack, which we cannot merge

@FabioLuporini I mean, is it? I would argue rather what's a hack is overriding instance `__dict__` by `EnrichedTuple.__new__`, and adding arbitrary attributes when constructing it and its children (essentially creating anonymous classes). I saw that `left` and `right` were the two attributes used consistently with `DimensionTuple` (child class) so I thought those attributes should explicitly be a part of the class.

I think ideally the hierarchy would be refactored such that those instance hacks aren't used anywhere and the attributes are explicit; that's why I at least thought the change should be made for `DimensionTuple`

FabioLuporini:
> > if really needed, it would be a hack, which we cannot merge
> 
> @FabioLuporini I mean, is it? Arguably what's more of a hack is [overriding instance `__dict__` by `EnrichedTuple.__new__`](https://github.com/devitocodes/devito/blob/772f1300b37d345d533e56216414ea1d90795aca/devito/tools/data_structures.py#L48), and adding arbitrary attributes when constructing it and its children (essentially creating anonymous classes).

I think it's perfectly legal to alter `__dict__` within `__new__` for classes that have a dynamic number of kwargs.

That being said, I think I agree with the rest you've said. We should maybe just wrap a dict and pull from it each time we enter `__getattr__`. Sounds neater

enwask:
Fair enough. So for now should I leave the explicit `left`/`right` rkwargs for `DimensionTuple` or get rid of them? Probably should have done this before but I can implement an override for `EnrichedTuple._rebuild` to reapply any arbitrary attributes from `__dict__` instead

FabioLuporini (approved):
for me this is now GTG, thanks. @mloubout ?

mloubout:
GTG, need rebase

enwask:
Made some changes requested on Slack; refactored `EnrichedTuple._getters` -> `.getters` for cleaner construction, as well as overriding `_rebuild` so arbitrary attributes aren't discarded (previously rebuilding e.g. `DimensionTuple(..., left=whatever)` dropped the additional attributes).

Also added `test_enrichedtuple_rebuild`.

mloubout (approved):

PR #2425, opened 2024-07-24
dsl: Make CondEq and CondNe easier to import

EdCaunt:
Allow import of `CondEq` and `CondNe` from `devito` rather than requiring `devito.symbolics`. In due course, these should probably move to `devito.types.relational` and use the `AbstractRel` mixin.


mloubout (approved):
need rebase

PR #2429, opened 2024-07-25
api: Make Derivative Reconstructable

mloubout:
Fixes

https://devitocodes.slack.com/archives/C1HSXP6KB/p1721850267638059

### DIFF
--- a/tests/test_derivatives.py
+++ b/tests/test_derivatives.py
@@ -713,6 +713,16 @@ class TestFD:
         assert dxy0.x0 == {y: y+y.spacing/2}
         assert dxy02.x0 == {x: x+x.spacing/2}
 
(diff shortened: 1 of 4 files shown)


tests/test_derivatives.py
```diff
@@ -713,6 +713,16 @@ def test_deriv_spec(self):
         assert dxy0.x0 == {y: y+y.spacing/2}
         assert dxy02.x0 == {x: x+x.spacing/2}
 
+    def test_deriv_stagg_plain(self):
```
EdCaunt, line 716: Would it be worth also testing evaluating `staggered=None` onto `staggered=x`? Also potentially mixing `staggered=None` and `staggered=NODE`

FabioLuporini (approved):
LGTM

PR #2431, opened 2024-07-30
compiler: Remove counters from subdomains and automatically generate unique names

EdCaunt:
Rather than using counters to ensure subdimensions etc are unique, they should be dynamically renamed during compilation.

Supersedes #2405 .


devito/types/grid.py
```diff
@@ -502,7 +502,6 @@ def __subdomain_finalize__(self, grid, **kwargs):
         # Create the SubDomain's SubDimensions
         sub_dimensions = []
         sdshape = []
-        counter = kwargs.get('counter', 0) - 1
```
georgebisbas, line 505: looks like remooving some hack here

EdCaunt: Yeah, that's the idea with the PR. It's also necessary to remove before introducing functions on subdimensions, since otherwise the counter gets even hackier

devito/ir/equations/algorithms.py
```diff
@@ -162,74 +161,112 @@ def concretize_subdims(exprs, **kwargs):
     """
     sregistry = kwargs.get('sregistry')
 
-    # {root Dimension -> {SubDimension -> concrete SubDimension}}
-    mapper = defaultdict(dict)
+    mapper = {}
+    rebuilt = {}  # Rebuilt implicit dims etc which are shared between dimensions
```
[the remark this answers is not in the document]
EdCaunt: This dict is for objects that get rebuilt over the course of the pass but aren't used in the final mapper. It ensures reuse of things like `MultiSubDomain` implicit dimensions, so I think "processed" would be misleading

devito/ir/equations/algorithms.py
```diff
@@ -162,74 +161,112 @@ def concretize_subdims(exprs, **kwargs):
...
+    if idim0 is not None:
+        try:
+            # Get a preexisiting substitution if one exists
+            idim1 = rebuilt[idim0]
+            # If a substitution exists for the implicit dimension,
+            # then there is also one for the function
+            functions = rebuilt[d.functions]
+        except KeyError:
+            iname = sregistry.make_name(prefix=idim0.name)
+            rebuilt[idim0] = idim1 = idim0._rebuild(name=iname)
```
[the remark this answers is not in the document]
EdCaunt: Same thing right?

devito/ir/equations/algorithms.py
```diff
@@ -162,74 +162,124 @@ def concretize_subdims(exprs, **kwargs):
...
     if not mapper:
         return exprs
 
-    subs = {}
-    for i in mapper.values():
-        subs.update(i)
+    # There may be indexed Arrays defined on SubDimensions in the expressions.
+    # These must have their dimensions replaced and their .function attribute
+    # reset to prevent recovery of the original SubDimensions.
+    functions = set().union(*[set(retrieve_functions(e)) for e in exprs])
```
FabioLuporini, line 175:
```
functions = retrieve_functions(exprs)
```

should work, and if not, it should be fixed

devito/ir/equations/algorithms.py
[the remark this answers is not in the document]
FabioLuporini: Ed's version is better

devito/ir/equations/algorithms.py
```diff
@@ -162,74 +162,124 @@ def concretize_subdims(exprs, **kwargs):
...
+             for tkn in tkns0]
 
-    thickness = (ltkn, rtkn)
+    kwargs = {'thickness': tuple(tkns1), 'functions': d.functions}
 
-    subs[d] = d._rebuild(d.name, pd, thickness)
+    idim0 = d.implicit_dimension
+    if idim0 is not None:
+        try:
+            # Get a preexisiting substitution if one exists
```
FabioLuporini, line 267: useless comment

devito/ir/equations/algorithms.py
```diff
@@ -162,74 +162,124 @@ def concretize_subdims(exprs, **kwargs):
...
+
+@_concretize_subdims.register(MultiSubDimension)
+def _(d, mapper, rebuilt, sregistry):
+    if d in mapper:
         # Already have a substitution for this dimension
         return
 
-    name = sregistry.make_name(prefix=d.name)
-    ltkn, rtkn = MultiSubDimension._symbolic_thickness(name)
+    tkns0 = MultiSubDimension._symbolic_thickness(d.name)
```
FabioLuporini, line 258: nitpicking: I'd call this, for clarity, `abstract_tkn` and the one below `concrete_tkn`

devito/ir/equations/algorithms.py
```diff
@@ -162,74 +162,124 @@ def concretize_subdims(exprs, **kwargs):
...
+            # If a substitution exists for the implicit dimension,
+            # then there is also one for the function
+            functions = rebuilt[d.functions]
+        except KeyError:
+            iname = sregistry.make_name(prefix=idim0.name)
+            rebuilt[idim0] = idim1 = idim0._rebuild(name=iname)
+
+            fdims = (idim1,) + (d.functions.dimensions[1:])
+            frebuilt = d.functions._rebuild(dimensions=fdims, function=None,
+                                            halo=None, padding=None,
```
FabioLuporini, line 278: are halo and padding really necessary here

devito/ir/equations/algorithms.py
```diff
@@ -162,74 +162,124 @@ def concretize_subdims(exprs, **kwargs):
...
+                                            initializer=d.functions.data)
```
FabioLuporini, line 279: we need a workaround to this. We cannot create data in the compiler. Here we somehow have to initialize the rebuilt function with a reference to `d.functions.data`. I'd be fine with introducing a new low-level (developer-only) API along the lines of `..., initializer=Reference(d.functions.data), . ..` and then use it right before this `elif`:

https://github.com/devitocodes/devito/blob/master/devito/types/dense.py#L94

with something like

```
if isinstance(initializer, Reference):
     self._data = initializer._data
```

Comments appreciated. This was just off top of my head

devito/ir/iet/utils.py
```diff
@@ -134,7 +134,6 @@ def derive_parameters(iet, drop_locals=False, ordering='default'):
     # using devito as a library-generator to be embedded within legacy codes
     if ordering == 'canonical':
         parameters = sorted(parameters, key=lambda p: str(type(p)))
-
```
FabioLuporini, line 137: bring it back

devito/passes/iet/definitions.py
```diff
@@ -337,6 +337,8 @@ def place_definitions(self, iet, globs=None, **kwargs):
         storage = Storage()
         defines = FindSymbols('defines-aliases|globals').visit(iet)
 
+        # FIXME: Use of findsymbols here is the culprit, pulls the original
```
FabioLuporini, line 340: Drop

devito/types/basic.py
```diff
@@ -345,6 +345,10 @@ def _filter_assumptions(cls, **kwargs):
         for i in list(kwargs):
             if i in _assume_rules.defined_facts:
                 assumptions[i] = kwargs.pop(i)
+        # Handle the nonnegative argument
```
FabioLuporini, line 348: nitpicking: this method can use some blank lines

devito/types/basic.py
```diff
@@ -382,7 +386,8 @@ def __eq__(self, other):
     __hash__ = sympy.Symbol.__hash__
 
     def _hashable_content(self):
-        return super()._hashable_content() + (self.dtype, self.is_const)
+        return super()._hashable_content() \
+            + (self.dtype, self.is_const, self.nonnegative)
```
FabioLuporini, line 390: isn't it part of super() already? can you add a test to confirm this. 

All the assumptions should be  part of the hash clearly, but we don't want to have to list them here explicitly one by one

devito/types/basic.py
```diff
@@ -426,6 +431,13 @@ def is_const(self):
         """
         return self._is_const
 
+    @property
+    def nonnegative(self):
+        """
+        Alias for is_nonnegative
```
FabioLuporini, line 437: why the need? sympy convention is "is_..."

FabioLuporini (commented):
This is looking so good

devito/ir/equations/algorithms.py
EdCaunt: Yeah, I looked into removing them, but the problem is, if you set `function=None` (which is necessary here), there is no way of checking the dimensions have been changed during a rebuild, since you can't check against the `function` kwarg. You need to specify `halo=None` and `padding=None` when changing the dimensions, as a bunch of mappers need reconstructing with the new dimensions. It throws an error otherwise.

devito/ir/equations/algorithms.py
EdCaunt: Yeah, this was the sort of functionality I was thinking of, but had to resort to initializer. I'll create a separate PR for this, then rebase on top of it

devito/types/basic.py
EdCaunt: Because whilst the attribute is `is_nonnegative`, the argument that sets it is just `nonnegative`, so it's needed for rebuilding

devito/ir/equations/algorithms.py
```diff
@@ -162,74 +162,124 @@ def concretize_subdims(exprs, **kwargs):
...
+    _concretize_subdims(exprs, mapper, rebuilt, sregistry)
     if not mapper:
         return exprs
 
-    subs = {}
-    for i in mapper.values():
-        subs.update(i)
+    # There may be indexed Arrays defined on SubDimensions in the expressions.
+    # These must have their dimensions replaced and their .function attribute
+    # reset to prevent recovery of the original SubDimensions.
```
ZoeLeibowitz, line 174: Remove full stop

devito/ir/equations/algorithms.py
```diff
@@ -162,74 +162,124 @@ def concretize_subdims(exprs, **kwargs):
...
+    functions = set().union(*[set(retrieve_functions(e)) for e in exprs])
+    functions = {f for f in functions if f.is_Array}
+    for f in functions:
+        dimensions = tuple(mapper[d] if d in mapper else d for d in f.dimensions)
+        if dimensions != f.dimensions:  # A dimension has been rebuilt
+            # So build a mapper for Indexed
```
ZoeLeibowitz, line 180: Maybe these comments should be stuck together?

tests/test_dimension.py
```diff
@@ -773,9 +776,10 @@ def test_expandingbox_like(self, opt):
         eqn = eqn.subs({x: xi, y: yi})
 
         op = Operator(eqn, opt=opt)
+        print(op.parameters)
```
ZoeLeibowitz, line 779: left over

tests/test_subdomains.py
```diff
@@ -669,10 +671,14 @@ class Dummy(SubDomainSet):
         # Make sure it jit-compiles
         op.cfunction
 
-        assert_structure(op, ['t,n0', 't,n0,xi20_blk0,yi20_blk0,x,y,z'],
-                         't,n0,xi20_blk0,yi20_blk0,x,y,z')
+        assert_structure(op, ['t,n0', 't,n0,ix0_blk0,iy0_blk0,x,y,z'],
+                         't,n0,ix0_blk0,iy0_blk0,x,y,z')
 
-        xi, _, _ = dummy.dimensions
+        # Drag a rebuilt MultiSubDimension out of the operator kicking and screaming
```
ZoeLeibowitz, line 677: kicking and screaming?

devito/ir/equations/algorithms.py
EdCaunt: Actually, I think we already have it:
```
frebuilt = d.functions._rebuild(dimensions=fdims, function=None,
                                            halo=None, padding=None,
                                            allocator=ExternalAllocator(d.functions.data),
                                            initializer=lambda x: None)
```

devito/ir/equations/algorithms.py
```diff
@@ -162,74 +163,123 @@ def concretize_subdims(exprs, **kwargs):
...
 
-    subs = {}
-    for i in mapper.values():
-        subs.update(i)
+    # There may be indexed Arrays defined on SubDimensions in the expressions
+    # These must have their dimensions replaced and their .function attribute
+    # reset to prevent recovery of the original SubDimensions
+    functions = {f for f in retrieve_functions(exprs) if f.is_Array}
+    for f in functions:
+        dimensions = tuple(mapper[d] if d in mapper else d for d in f.dimensions)
```
mloubout, line 178: `mapper.get(d, d)`

devito/ir/equations/algorithms.py
```diff
@@ -162,74 +163,123 @@ def concretize_subdims(exprs, **kwargs):
...
+        if dimensions != f.dimensions:
+            # A dimension has been rebuilt, so build a mapper for Indexed
+            mapper[f.indexed] = f._rebuild(dimensions=dimensions, function=None).indexed
```
mloubout, line 181: Shouldn't `f` already be indexed if getting there?

devito/types/basic.py
mloubout: Not sure I follow why it's needed and why `is_nonnegative` is problematic

devito/ir/equations/algorithms.py
EdCaunt: No, `retrieve_functions` gets the underlying function

mloubout: then shouldn't this use `retrieve_indexed` instead?

devito/types/basic.py
EdCaunt: `is_nonnegative` isn't problematic, but it's set with `Symbol(name='x', nonnegative=True)`. Hence rebuilding requires the alias

devito/types/basic.py
```diff
@@ -341,10 +341,16 @@ def _filter_assumptions(cls, **kwargs):
...
         for key in ('real', 'imaginary', 'commutative'):
             kwargs.pop(key, None)
+
         # extract sympy.Symbol-specific kwargs
         for i in list(kwargs):
             if i in _assume_rules.defined_facts:
                 assumptions[i] = kwargs.pop(i)
+
+        # Handle the nonnegative argument
+        if 'nonnegative' in assumptions and assumptions['nonnegative'] is None:
```
mloubout, line 351: that's just `assumptions.pop('nonnegative', None)`

devito/ir/equations/algorithms.py
EdCaunt: No, because you can `uxreplace` a single `{f.indexed: g.indexed}` to turn all `f[x]`, `f[x+1]` into `g[x]`, `g[x+1]`

devito/types/basic.py
EdCaunt: No it's not. If `nonnegative` is provided as a kwarg, but is not `None` then you don't want to pop it.

devito/ir/equations/algorithms.py
```diff
@@ -162,74 +163,129 @@ def concretize_subdims(exprs, **kwargs):
...
+        fkwargs.update({'dimensions': (idim1,) + (d.functions.dimensions[1:]),
+                        'halo': None,
+                        'padding': None})
+
+    try:
+        functions = rebuilt[d.functions]
+    except KeyError:
+        fkwargs.update({'name': sregistry.make_name(prefix=d.functions.name),
+                        'function': None,
+                        'allocator': ExternalAllocator(d.functions.data),
```
EdCaunt, line 283: Note: `data_with_halo` would be safer here

FabioLuporini, line 283: `._data`

`.data` instead is just a view. Suggests that you need a test with space_order != 0 to trigger what would be an issue if you don't replace `.data` with `._data`

devito/ir/equations/algorithms.py
```diff
@@ -162,74 +163,129 @@ def concretize_subdims(exprs, **kwargs):
...
+                        'initializer': lambda x: None})
```
FabioLuporini, line 284: if the ExternalAllocator is passed, we shouldn't need to pass initializer like that. One should be enough; if it isn't, please tweak it inside `DiscreteFunction.__init_finalize_`

devito/types/dimension.py
```diff
@@ -395,6 +395,22 @@ def __eq__(self, other):
     __hash__ = Symbol.__hash__
 
 
+class UniqueDimension(BasicDimension):
```
FabioLuporini, line 398: I struggle to understand the need for this class

EdCaunt: `Dimension`s with the same names always hash the same, so when used as dictionary keys, each entry replaces the last, rather than creating new entries. The implicit dimension of the `SubDomainSet` always has the name `'n'`

devito/ir/equations/algorithms.py
EdCaunt: You need to set a noop initialiser as per the docstring

devito/types/basic.py
```diff
@@ -879,6 +880,12 @@ def __new__(cls, *args, **kwargs):
             # let's just return `function` itself
             return function
 
+        # If dimensions have been replaced, then it is necessary to remove halo
+        # and padding kwargs so that they are rebuilt with the new dimensions
+        if function is not None and function.dimensions != dimensions:
```
mloubout, line 885: That should not be a valid combination. If `function` is not None it *has* to have the same dimensions

devito/types/dimension.py
FabioLuporini: so the actual error is that it should be named uniquely instead ?

devito/types/basic.py
FabioLuporini: this is just for reconstruction; see one of the latest threads on slack in #development

devito/types/dimension.py
EdCaunt: But then you need a counter again... Although I suppose you could append the `id(self)` to the name when you first create it, before subsequently rebuilding it as `n0`, `n1`, ... Maybe that's the way to go?

mloubout (approved):

devito/ir/support/symregistry.py
```diff
@@ -30,7 +30,7 @@ def __init__(self):
         # passes, to maximize symbol (especially Dimension) reuse
         self.caches = {}
 
-    def make_name(self, prefix=None):
+    def make_name(self, prefix=None, increment_first=True):
```
mloubout, line 33: Any reason to not make the default case to keep first name

FabioLuporini: I guess you would then have

float r = ...
float r1 = ...

and `r` is such a common name

devito/ir/equations/algorithms.py
```diff
@@ -162,74 +163,127 @@ def concretize_subdims(exprs, **kwargs):
...
-        subs.update(i)
+    # There may be indexed Arrays defined on SubDimensions in the expressions
+    # These must have their dimensions replaced and their .function attribute
+    # reset to prevent recovery of the original SubDimensions
+    functions = {f for f in retrieve_functions(exprs) if f.is_Array}
+    for f in functions:
+        dimensions = tuple(mapper.get(d, d) for d in f.dimensions)
+        if dimensions != f.dimensions:
+            # A dimension has been rebuilt, so build a mapper for Indexed
+            mapper[f.indexed] = f._rebuild(dimensions=dimensions, function=None).indexed
```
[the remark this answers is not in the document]
FabioLuporini: I mean, given this:

https://github.com/devitocodes/devito/blob/master/devito/types/basic.py#L886

devito/ir/equations/algorithms.py
```diff
@@ -162,74 +164,139 @@ def concretize_subdims(exprs, **kwargs):
...
-    # Grid.__subdomain_finalize__
-    pass
+def _(d, mapper, rebuilt, sregistry):
+    if d in mapper:
+        # Already have a substitution for this dimension
+        return
 
+    tkns = [tkn._rebuild(name=sregistry.make_name(prefix=tkn.name))
+            for tkn in d.tkns]
+    tkns_subs = {tkn0: tkn1 for tkn0, tkn1 in zip(d.tkns, tkns)}
```
FabioLuporini, line 218: Isn't it as simple as:

```
subs = {i: i._rebuild(name=sregistry.make_name(prefix=i.name) for i in d.tkns}
```

probably even a one liner this way...

devito/ir/equations/algorithms.py
```diff
@@ -162,74 +164,139 @@ def concretize_subdims(exprs, **kwargs):
...
+    if idim0 is not None:
+        if idim0 in rebuilt:
+            idim1 = rebuilt[idim0]
+        else:
+            iname = sregistry.make_name(prefix='n')
+            rebuilt[idim0] = idim1 = idim0._rebuild(name=iname)
+
+        kwargs['implicit_dimension'] = idim1
+        fkwargs.update({'dimensions': (idim1,) + d.functions.dimensions[1:],
+                        'halo': None,
```
FabioLuporini, line 276: I'm not sure these two Nones are actually necessary anymore?

Given https://github.com/devitocodes/devito/blob/master/devito/types/basic.py#L888-L889

devito/ir/equations/algorithms.py
```diff
@@ -162,74 +164,139 @@ def concretize_subdims(exprs, **kwargs):
...
+
+    if d.functions in rebuilt:
+        functions = rebuilt[d.functions]
+    else:
+        # Increment every instance of this name after the first encountered
+        fname = sregistry.make_name(prefix=d.functions.name, increment_first=False)
+        # Warn the user if name has been changed, since this will affect overrides
+        if fname != d.functions.name:
+            fkwargs['name'] = fname
+            warning("%s <%s> renamed as '%s'" %
```
FabioLuporini, line 287: Something more informative about how to work around this warning?

E.g., "... Consider assigning an explicit name to `%s` instead" % d.functions.name"

?

devito/ir/support/symregistry.py
EdCaunt: >I guess you would then have
>
>float r = ...
>float r1 = ...
>
>and r is such a common name
Correct. This behaviour is only desirable in some cases, hence the switch.

devito/ir/equations/algorithms.py
EdCaunt: They are, but I've moved them to make it clear why they're needed (`function=None` is also set further down)

FabioLuporini (approved):

FabioLuporini:
GTG!

PR #2434, opened 2024-07-31
api: revamp custom coefficients API

mloubout:
Superseeds #1644


FabioLuporini (commented):
This is great, thanks!

Couple of questions, mostly related to the aesthetics of the API

* do we now risk a proliferation of `f.d(weights=weights, ...)` in the PDE specification? things might get a little noisy 
  * Is it useless or perhaps difficult to provide, in addition, an API such that if you do `obj = Space(f, w, ...)` then `obj.dx == f.dx(weights=w)` ?
* I think, now that we're revamping the API, we need to add an alias such that `f.dx(weights=w0) == f.dx(w=w0)` , again to minimize verbosity
* EvalDerivatives and IndexDerivatives are all constructed automatically and seamlessly regardless of whether taylor or custom coefficients are use, right?

mloubout:
> do we now risk a proliferation of f.d(weights=weights, ...) in the PDE specification?

Well yes, and no. One thing that will be added to tuto doc is that you can provide your own backend for FD through `fd_weights_registry`  so if you gonna have your own coefficients everywhere you can create your own `my_custom_weight` and then add it to `fd_weights_registry` and then use it 

>we need to add an alias such that f.dx(weights=w0) == f.dx(w=w0) 

Completely fine yes, will add it

>EvalDerivatives and IndexDerivative

Yes, It actually makes everything quite simpler for the compiler, everything now goes through standard FD and creates EvalDerivatives and IndexDerivative. There is no more intricated post-process replacement pass after evaluation.

EdCaunt (commented):
Looks like a pretty big cleanup, much needed.

There should be a test to make sure that changing the weights of a `Derivative` changes the hash, and that two `Derivative`s with identical weights + spec hash the same. Also tests for pickling and unpickling derivatives with custom coefficients.

EdCaunt:
> do we now risk a proliferation of f.d(weights=weights, ...) in the PDE specification?
The user can also do something like:
```
subs = {f.dx: f.dx(weights=weights0),
             f.dy: f.dy(weights=weights0),
             f.dx2: f.dx2(weights=weights1)}
eq0 = Eq(f, f.dx + f.dx2)
eq1 = Eq(f, f.dx + f.dy +1)
eqs = [eq.subs(subs) for eq in (eq0, eq1)]
```
We should probably have a test for this

EdCaunt:
We should also check that:
```
solve(f.dx(weights=weights), f.forward)
```
works correctly

examples/seismic/tutorials/07_DRP_schemes.ipynb
georgebisbas, line 1: Do the big figure changes need to be committed?

examples/seismic/tutorials/07_DRP_schemes.ipynb
```diff
@@ -492,23 +489,21 @@
...
+    "                      1.64875, -0.223818, 0.0274017])\n",
+    "\n",
     "ux_u_coeffs = weights_u/x.spacing**2\n",
     "uz_u_coeffs = weights_u/z.spacing**2\n",
     "ux_l_coeffs = weights_l/x.spacing**2\n",
     "uz_l_coeffs = weights_l/z.spacing**2\n",
     "\n",
     "u_lap = u_DRP.dx2(weights=ux_u_coeffs) + u_DRP.dy2(weights=uz_u_coeffs)\n",
-    "l_lap = u_DRP.dx2(weights=ux_l_coeffs) + u_DRP.dy2(weights=uz_l_coeffs)\n",
+    "l_lap = u_DRP.dx2(weights=ux_l_coeffs, fd_order=order-4) + u_DRP.dy2(weights=uz_l_coeffs, fd_order=order-4)\n",
```
mloubout, line 506: ??

EdCaunt: With the new interface, you can completely truncate the zeroes off the end of the stencil in the lower section of the domain. Think it should reduce computational cost unless there's some symbolic optimisation during lowering which removes the zero coefficients anyway

examples/seismic/tutorials/07_DRP_schemes.ipynb
EdCaunt: Can you avoid committing them somehow?

devito/types/equation.py
```diff
@@ -65,14 +63,43 @@ class Eq(sympy.Eq, Evaluable):
...
 
         return obj
 
+    @classmethod
+    def _apply_coeffs(cls, expr, coefficients):
+        """
+        This process legacy API of Substitution/Coefficients applying the weights
+        to the target Derivatives.
+        """
+        from devito.symbolics import retrieve_derivatives
```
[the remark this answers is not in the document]
mloubout: circular import, this is for legacy support so I think it's ok to keep here

mloubout:
> eqs = [eq.subs(subs) for eq in (eq0, eq1)]

I would prefer to avoid adding these type of constructions that are not really good or clean

EdCaunt:
Notebook needs the text reworking. It still references the old API throughout.

examples/seismic/tutorials/07_DRP_schemes.ipynb
georgebisbas: is the image indeed different?
If not, one thing I do is 'git add -p' 
https://nuclearsquid.com/writings/git-add/

devito/ir/clusters/algorithms.py
```diff
@@ -663,8 +663,18 @@ def _normalize_reductions_dense(cluster, mapper, sregistry, platform):
                 a = mapper[rhs] = Array(name=name, dtype=e.dtype, dimensions=dims,
                                         grid=grid)
 
-                processed.extend([Eq(a.indexify(), rhs),
-                                  e.func(lhs, a.indexify())])
+                # Populate the Array (the "map" part)
+                processed.append(e.func(a.indexify(), rhs, operation=None))
+
+                # Set all untouched entried to the identity value if necessary
```
georgebisbas, line 669: entries

EdCaunt (approved):

devito/finite_differences/tools.py
```diff
@@ -305,6 +283,12 @@ def generate_indices(expr, dim, order, side=None, matvec=None, x0=None):
         else:
             o_min -= 1
 
+    if nweights > 0 and (o_max - o_min + 1) != nweights:
+        assert nweights == (o_max - o_min + 1) + 1
```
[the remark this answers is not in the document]
FabioLuporini: agree

devito/finite_differences/tools.py
```diff
@@ -325,3 +309,16 @@ def make_shift_x0(shift, ndim):
             raise ValueError("ndim length must be equal to 1 or 2")
     raise ValueError("shift parameter must be one of the following options: "
                      "None, float or tuple with shape equal to %s" % (ndim,))
+
+
+def process_weights(weights, expr):
+    if weights is None:
+        return 0, None
+    elif isinstance(weights, Function):
+        wdim = {d for d in weights.dimensions if d not in expr.dimensions}
```
FabioLuporini, line 318: potential nitpicking and therefore ignorable:

is it worth introducing a `SymbolicWeights` subclass (can't call it just `Weights` because it'd be the same name as `finite_differences/differentiable.py::Weights`) and add a few properties such as `.dimension` ? or is this impossible because you need `expr`?

devito/finite_differences/tools.py
```diff
@@ -325,3 +309,16 @@ def make_shift_x0(shift, ndim):
...
+        assert len(wdim) == 1
+        wdim = wdim.pop()
+        shape = weights.shape
+        return shape[weights.dimensions.index(wdim)], wdim
+    else:
+        return len(list(weights)), None
```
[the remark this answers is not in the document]
FabioLuporini: perhaps weights could be a generator?

devito/types/equation.py
```diff
@@ -65,14 +63,43 @@ class Eq(sympy.Eq, Evaluable):
 
     def __new__(cls, lhs, rhs=0, subdomain=None, coefficients=None, implicit_dims=None,
                 **kwargs):
+        if coefficients is not None:
```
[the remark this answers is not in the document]
FabioLuporini: yeah this is basically my comment above, we probably need a global flag somewhere 

or here's another idea...

we could add a `deprecated(..., flag='symbolic_coefficients')` that internally performs voodoo -- example, flagging the `'symbolic_coefficients'` entry in a data structure such that only the first time `deprecated` with that flag is encountered the message is printed
so by using the same flag name in different places we would ensure uniqueness of the message

but yeah, just a random thought

devito/types/equation.py
```diff
@@ -65,14 +63,43 @@ class Eq(sympy.Eq, Evaluable):
...
+            warn("The Substitution API is deprecated and will be removed, "
+                 "coefficients should be passed directly to the derivative object"
+                 " i.e `u.dx(weights=...)",
+                 DeprecationWarning, stacklevel=2)
         kwargs['evaluate'] = False
+        # Backward compat
+        rhs = cls._apply_coeffs(rhs, coefficients)
+        lhs = cls._apply_coeffs(lhs, coefficients)
         obj = sympy.Eq.__new__(cls, lhs, rhs, **kwargs)
```
FabioLuporini, line 75: nitpicking: blank lines above and below?

examples/seismic/model.py
```diff
@@ -356,7 +356,7 @@ def _cfl_coeff(self):
         if 'lam' in self._physical_parameters or 'vs' in self._physical_parameters:
             coeffs = fd_w(1, range(-self.space_order//2+1, self.space_order//2+1), .5)
             c_fd = sum(np.abs(coeffs[-1][-1])) / 2
-            return np.sqrt(self.dim) / self.dim / c_fd
+            return .95 * np.sqrt(self.dim) / self.dim / c_fd
```
FabioLuporini, line 359: so this justifies the change in expected norm values above?

examples/seismic/viscoelastic/viscoelastic_example.py
```diff
@@ -41,8 +41,8 @@ def run(shape=(50, 50), spacing=(20.0, 20.0), tn=1000.0,
 @pytest.mark.parametrize("dtype", [np.float32, np.float64])
 def test_viscoelastic(dtype):
     _, _, _, [rec1, rec2, v, tau] = run(dtype=dtype)
-    assert np.isclose(norm(rec1), 12.30114, atol=1e-3, rtol=0)
-    assert np.isclose(norm(rec2), 0.312462, atol=1e-3, rtol=0)
+    assert np.isclose(norm(rec1), 12.62339, atol=1e-3, rtol=0)
```
FabioLuporini, line 44: same here as above

devito/finite_differences/derivative.py
```diff
@@ -9,13 +9,13 @@
...
 from devito.tools import (as_mapper, as_tuple, filter_ordered, frozendict, is_integer,
-                          Reconstructable)
+                          Pickable)
 from devito.types.utils import DimensionTuple
 
 __all__ = ['Derivative']
 
 
-class Derivative(sympy.Derivative, Differentiable, Reconstructable):
+class Derivative(sympy.Derivative, Differentiable, Pickable):
```
FabioLuporini, line 18: why isn't Reconstructable enough ?

mloubout: You only get sympy's rkwargs at pickle if you only use Reconstructable so you would loose all the x0/weights/side/transpose at pickle, quite weird it didn't trigger a bug before

examples/seismic/model.py
mloubout: Yes. The CFL bound is extremely tight without it so for some complicated models it can become unstable, this is safer

devito/finite_differences/tools.py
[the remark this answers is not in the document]
mloubout: It can also be `Weights` which behave weirdly with `len`

devito/finite_differences/derivative.py
FabioLuporini: I wonder whether it's due to this , which we *might* want to lift in Reconstructable. Because as for the rest a Pickable should reconstruct like a Reconstructable...

anyway, no need to investigate now

devito/finite_differences/coefficients.py
```diff
@@ -1,73 +1,15 @@
...
-        if dimension.is_Time:
-            if wl != function.time_order:
-                raise ValueError("Number of FD weights provided does not "
-                                 "match the functions space_order")
-        elif dimension.is_Space:
-            if wl != function.space_order:
-                raise ValueError("Number of FD weights provided does not "
-                                 "match the functions space_order")
-
+        deprecations.coeff_warn
```
FabioLuporini, line 8: nitpicking: perhaps a memoized_meth would be neater because you're performing an action rather than retrieving a property

but really, this is super-nitpicking. Can revisit another day

FabioLuporini (approved):

PR #2435, opened 2024-07-31
compiler: fix unevaluation of sparse cluster

mloubout:
Fix issue with pickling operator with sparse + subs


FabioLuporini:
Thanks, merged

PR #2438, opened 2024-08-06
dsl: Improve Function rebuilding and use of preexisting arrays as Function data

EdCaunt:


devito/types/basic.py
```diff
@@ -879,6 +879,14 @@ def __new__(cls, *args, **kwargs):
             # let's just return `function` itself
             return function
 
+        # If dimensions have been replaced, then it is necessary to set `function`
+        # to None. It is also necessary to remove halo and padding kwargs so that
+        # they are rebuilt with the new dimensions
+        if function is not None and function.dimensions != dimensions:
```
mloubout, line 885: I'm sorry but no we cannot allow that it doesn't matter if it's for internals, this need to go through a `__new__` not `_rebuild`

EdCaunt: But `_rebuild()` is a `__new__` with some extra convenience around it?

devito/types/dense.py
```diff
@@ -84,6 +85,12 @@ def __init_finalize__(self, *args, function=None, **kwargs):
 
         # Data initialization
         initializer = kwargs.get('initializer')
+
+        # Don't want to reinitialise array if DataReference used as allocator;
+        # create a no-op intialiser to avoid overwriting the original array.
+        if isinstance(self._allocator, DataReference):
```
FabioLuporini, line 91: can this be made part of the if-elif-else cascade below?

tests/test_function.py
```diff
@@ -0,0 +1,102 @@
+import pytest
+import numpy as np
+
+from devito import Dimension, Function, Grid, Eq, Operator
+from devito.data.allocators import DataReference
+
+
+class TestRebuild:
```
FabioLuporini, line 8: I don't think we need a new file for this...

this class can go inside test_symbolics ...

tests/test_function.py
```diff
@@ -0,0 +1,102 @@
...
+        f1 = f._rebuild(dimensions=dims1)
+
+        assert f0.function is f0
+        assert f0.dimensions == dims0
+
+        assert f1.function is f1
+        assert f1.dimensions == dims1
+
+
+class TestDataReference:
```
FabioLuporini, line 32: ... while this class belongs to test_data, where other allocators are tested already

tests/test_function.py
EdCaunt: `test_symbolics` is a mess which really needs tidying up. It's currently an unordered grab bag of miscellaneous tests. Maybe that should be done in this PR?

FabioLuporini: I can do with a test_rebuild.py in which we gather this TestRebuild and all of the rebuilding tests in test_symbolics.py

devito/data/allocators.py
```diff
@@ -337,10 +337,10 @@ def put_local(self):
         return self._node == 'local'
 
 
-class ExternalAllocator(MemoryAllocator):
+class DataReference(MemoryAllocator):
```
mloubout, line 340: Add `ExternalAllocator = DataReference` for backward. compatibility

tests/test_function.py
```diff
@@ -0,0 +1,102 @@
...
+
+        f = Function(name='f', dimensions=(x, y), shape=(11, 11))
+
+        dims0 = (x0, y0)
+        dims1 = (x, y0)
+
+        f0 = f._rebuild(dimensions=dims0)
+        f1 = f._rebuild(dimensions=dims1)
+
+        assert f0.function is f0
```
mloubout, line 25: this needs more checks on underlying data

tests/test_function.py
EdCaunt: Yeah, that's a solution

devito/types/dense.py
EdCaunt: No, several branches in the cascade use `initialiser` which means it needs to be set here to avoid code duplication. Keeping it here also results in more readable code.

tests/test_function.py
```diff
@@ -0,0 +1,102 @@
...
+        assert np.all(f.data_with_halo == g.data_with_halo)
+
+        check = np.array(f.data_with_halo[1:-1, 1:-1])
+
+        # Update both
+        Operator([Eq(f, f+1), Eq(g, g+1)])()
+        assert np.all(f.data_with_halo == g.data_with_halo)
+        # Check that it was incremented by two
+        check += 2
+        assert np.all(f.data == check)
```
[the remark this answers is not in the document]
EdCaunt: It's reshaped, which makes it more fiddly to build the array from scratch (since it will skip various numbers) than just copying the data before the operator is run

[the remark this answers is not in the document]
EdCaunt: Also the amount of skipping changes depending on the MPI config, since there will be different halos

FabioLuporini (approved):

PR #2441, opened 2024-08-12
compiler: respect fetch direction

mloubout:


EdCaunt (approved):
Lgtm

georgebisbas (approved):

devito/passes/clusters/asynchrony.py
```diff
@@ -237,10 +237,11 @@ def _actions_from_update_memcpy(c, d, clusters, actions, sregistry):
     target = e.lhs.function
 
     fetch = e.rhs.indices[d]
-    if direction is Forward:
-        findex = fetch + 1
+    fshift = 1 if direction is Forward else -1
```
FabioLuporini, line 240: random thought: we should just set the direction values to -1, 0 (Any), 1

devito/passes/clusters/asynchrony.py
```diff
@@ -237,10 +237,11 @@ def _actions_from_update_memcpy(c, d, clusters, actions, sregistry):
...
+    if isinstance(fetch, IntDiv) or any(isinstance(a, IntDiv) for a in fetch.args):
```
FabioLuporini, line 241: I don't understand why we need this change.

Searching for IntDiv like that seems fragile to me. Also, either `search(fetch, IntDiv)` or `fetch.find(IntDiv)` should be a more compact and robust way of expressing what this line is trying to express

FabioLuporini (commented):
We'll need a test for this, ideally here in OSS (`test_gpu_common::TestStreaming`) or at worst in PRO would be fine too

FabioLuporini:
> We'll need a test for this, ideally here in OSS (`test_gpu_common::TestStreaming`) or at worst in PRO would be fine too

sorry, there is indeed one in PRO. Sorry!

devito/passes/clusters/asynchrony.py
mloubout: Yeah good point. 0 might lead to scheduling issues though

devito/passes/clusters/asynchrony.py
mloubout: If you have `fetch = IntDiv(time, factor)` or some expression of it, then `fetch._subs(pd, pd + fshift)` will become `IntDiv(time+1, factor)` which basically does nothing because `IntDiv(time, factor) = IntDiv(time+1, factor)`. Ideally would not want that if but rather  find the time dimension in `fetch` then do a `._sub(t, t+t.spacing)` but then need to do an extra `{spacing: 1}` afterwards too. 
THis seemed simpler and less involved and I think covers 99% of the cases

PR #2446, opened 2024-08-21
compiler: Add various performance optimization variants

FabioLuporini:


devito/passes/clusters/cse.py
```diff
@@ -32,21 +28,58 @@ class CTemp(Temp):
...
+        The symbol registry to use for creating temporaries.
+    options : dict
+        The optimization options.
+        Accepted: ['cse-min-cost', 'cse-algo'].
+        * 'cse-min-cost': int. The minimum cost of a common sub-expression to be
+          considered for CSE. Default is 1.
+        * 'cse-algo': str. The CSE algorithm to apply. Accepted: ['basic',
+          'smartsort', 'advanced']. Default is 'basic'.
     """
+    min_cost = options['cse-min-cost']
```
EdCaunt, line 70: Are defaults set on these somewhere else?

devito/passes/clusters/cse.py
```diff
@@ -55,28 +88,29 @@ def _cse(maybe_exprs, make, min_cost=1, mode='default'):
...
-    # - it sometimes "captures too much", losing factorization opportunities;
-    # - very slow
-    # TODO: a second "sympy" mode will be provided, relying on SymPy's CSE() but
-    # also ensuring some form of post-processing
-    assert mode == 'default'  # Only supported mode ATM
+    Notes
+    -----
+    We're not using SymPy's CSE for three reasons:
+
+        * It also captures array index access functions (e.g., i+1 in A[i+1]);
```
EdCaunt, line 97: In future, we will want to capture these for things like `x+1-ltkn` and `INT(abs(x))` etc, but we probably want some kind of filter specifying a minimum verbosity

devito/symbolics/manipulation.py
```diff
@@ -17,7 +18,7 @@
 
 __all__ = ['xreplace_indices', 'pow_to_mul', 'indexify', 'subs_op_args',
            'normalize_args', 'uxreplace', 'Uxmapper', 'reuse_if_untouched',
-           'evalrel']
+           'evalrel', 'flatten_args']
```
EdCaunt, line 21: Why not move this if it causes a circular import?

FabioLuporini: 1) this is where it belongs
2) I would need a local important in this file otherwise

devito/passes/clusters/cse.py
FabioLuporini: yeah

devito/passes/clusters/factorization.py
```diff
@@ -96,7 +110,7 @@ def collect_special(expr):
     # Collect common temporaries (r0, r1, ...)
     w_coeffs = Add(*w_coeffs, evaluate=False)
     symbols = retrieve_symbols(w_coeffs)
-    if symbols:
+    if len(set(symbols)) != len(symbols):
```
mloubout, line 113: retrieve_symbols should return a set already why is this needed?

devito/passes/clusters/utils.py
```diff
@@ -1,44 +1,8 @@
...
-from devito.symbolics import uxreplace
 from devito.tools import as_tuple
-from devito.types import CriticalRegion, Eq, Symbol, Wildcard
+from devito.types import CriticalRegion, Eq, Symbol
 
-__all__ = ['makeit_ssa', 'is_memcpy', 'make_critical_sequence',
-           'in_critical_region']
-
-
-def makeit_ssa(exprs):
```
mloubout, line 10: nice

devito/symbolics/manipulation.py
```diff
@@ -276,6 +279,25 @@ def _eval_numbers(expr, args):
         args[:] = [expr.func(*numbers)] + others
 
 
+def flatten_args(args, op, ignore=None):
```
mloubout, line 282: I know I looked at this recently, can't remember why. I think expr.expand(**hints)` does that but may have limited options

devito/types/parallel.py
```diff
@@ -44,8 +44,10 @@ class NThreadsBase(NThreadsAbstract):
 
     @cached_property
     def default_value(self):
-        return int(os.environ.get('OMP_NUM_THREADS',
-                                  configuration['platform'].cores_physical))
+        return int(os.environ.get(
+            'OMP_NUM_THREADS',
+            configuration['platform'].cores_physical_per_numa_domain
```
mloubout, line 49: Since you touched this..... need a `// mpi_ranks`

devito/passes/clusters/cse.py
```diff
@@ -160,65 +231,181 @@ def _compact_temporaries(exprs, exclude):
...
+@_catch.register(Mul)
 def _(expr):
-    mapper = Counter()
-    for a in expr.args:
-        mapper.update(count(a))
+    mapper = _catch(expr.args)
+
+    mapper[Candidate(expr)].append(expr)
+
+    for n in range(2, len(expr.args)):
```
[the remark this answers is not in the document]
FabioLuporini: yes, this is all carefully explained in the docstring... it's an expensive algorithm and deliberately we're implementing the heuristic I chose (relying already on the canonical ordering of the arguments etc etc)

devito/types/parallel.py
FabioLuporini: why? this is already physical cores "per numa domain"

I could take `.core_physical` and divide it by `mpi_ranks`, which might be slightly better than `cores_physical_per_numa_domain`, for some exotic use cases...

devito/passes/clusters/cse.py
```diff
@@ -160,65 +231,181 @@ def _compact_temporaries(exprs, exclude):
...
+        if not isinstance(e0.lhs, CTemp):
+            continue
+
+        for e1 in exprs:
+            if e0.lhs in e1.rhs.free_symbols:
+                dag.add_edge(e0, e1, force_add=True)
+
+    def choose_element(queue, scheduled):
+        # Try to honor temporary names as much as possible
+        first = sorted(queue, key=lambda i: str(i.lhs)).pop(0)
```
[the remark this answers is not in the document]
FabioLuporini: that's what it was in the beginning... then I realised we had non deterministic code gen. Reason? `ComponentAccess`, used for vector types. If I use `.name` (or better, `.base.name`) , I won't be able to distinguish between `a.x` and `a.y`

devito/passes/clusters/factorization.py
```diff
@@ -72,6 +75,17 @@ def collect_special(expr):
...
+    # Any factorization possible?
+    if not any(len(i) > 1 for i in [w_funcs, w_pows, w_coeffs]):
+        # `evaluate=True` below guarantees de-nesting of Adds
+        # For example:
+        # `args[0] = ((0.1*(a + b)) + (0.2*(c + d)))`
+        # `args[1] = ((0.1*(e + f)) + (0.2*(g + h)))`
+        # ->
+        # `expr = (0.1*(a + b) + 0.2*(c + d) + 0.1*(e + f) + 0.2*(g + h))`
+        # Which is then further factorizable by `collect_const`
+        return reuse_if_untouched(expr, args, evaluate=True)
```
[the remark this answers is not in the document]
FabioLuporini: it's only triggered if any of the args has changed, otherwise `expr` is returned. Yes, this is necessary or we'll miss factorization opportunities down the line

devito/passes/clusters/factorization.py
FabioLuporini: no, it returns a list: https://github.com/devitocodes/devito/blob/master/devito/symbolics/search.py#L157
since it's "mode=all"

we indeed want a list here because it's only when the same symbol appears twice (e.g. `[r42 .... r42 ...`) that we know it's worth attempting a factorization

devito/passes/clusters/factorization.py
```diff
@@ -184,20 +207,21 @@ def _collect_nested(expr):
...
     elif expr.is_Mul:
-        args, candidates = zip(*[_collect_nested(arg) for arg in expr.args])
-        return Mul(*args), ReducerMap.fromdicts(*candidates)
+        args, candidates = zip(*[_collect_nested(a, strategy) for a in expr.args])
+        expr = reuse_if_untouched(expr, args, evaluate=True)
+        return expr, ReducerMap.fromdicts(*candidates)
     elif expr.is_Equality:
-        args, candidates = zip(*[_collect_nested(expr.lhs),
-                                 _collect_nested(expr.rhs)])
+        args, candidates = zip(*[_collect_nested(expr.lhs, strategy),
```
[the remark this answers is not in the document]
FabioLuporini: this is legacy, I can probably drop it; will try

devito/passes/clusters/factorization.py
```diff
@@ -128,6 +142,11 @@ def collect_const(expr):
         else:
             inverse_mapper[-v].append(-k)
 
+    # Any factorization possible?
+    if len(inverse_mapper) == len(expr.args) or \
+       list(inverse_mapper) == [1]:
```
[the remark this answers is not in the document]
FabioLuporini: ok

devito/passes/clusters/cse.py
```diff
@@ -160,65 +231,181 @@ def _compact_temporaries(exprs, exclude):
...
+
+class Candidate(tuple):
+
+    def __new__(cls, expr, conditionals=None, sources=()):
+        conditionals = frozendict(conditionals or {})
+        sources = as_tuple(sources)
+        return tuple.__new__(cls, (expr, conditionals, sources))
+
+    @property
+    def expr(self):
```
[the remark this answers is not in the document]
FabioLuporini: see explanation on slack. Sounds trivial, it's not in my opinion

devito/passes/clusters/cse.py
```diff
@@ -160,65 +231,181 @@ def _compact_temporaries(exprs, exclude):
...
+                    queue.remove(a)
+                except ValueError:
+                    args.append(a)
+            assert not queue
+            subs[i] = self.expr.func(*args)
+
+        return subs
+
+
+def catch(exprs, mode):
```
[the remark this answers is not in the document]
FabioLuporini: I can't use `search` because it'd be an alias of `symbolics.search`, and `retrieve_cse` is misleading given what "cse" means and in particular given that we're just looking for _potential_ common subexpressions (that is, candidates)

I think `catch` (for Candidates) is fine

devito/core/gpu.py
```diff
@@ -196,14 +198,14 @@ def _specialize_clusters(cls, clusters, **kwargs):
 
         # Reduce flops
```
georgebisbas, line 199: Flops minimization and Reduce flops as comments subsections looks duplicate....
I would keep CSE in the first subtitle, line 49

examples/performance/01_gpu.ipynb
georgebisbas, line 1: so, it is not applied here now

tests/test_cse.py
```diff
@@ -0,0 +1,248 @@
...
+from devito.finite_differences.differentiable import diffify
+from devito.ir import DummyEq, FindNodes, FindSymbols, Conditional
+from devito.ir.support import generator
+from devito.passes.clusters.cse import CTemp, _cse
+from devito.symbolics import indexify
+from devito.types import Array, Symbol, Temp
+
+
+@pytest.mark.parametrize('exprs,expected,min_cost', [
+    # Simple cases
```
georgebisbas, line 19: are all these just moved, or edited/enhanced as well?

devito/passes/clusters/cse.py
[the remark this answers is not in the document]
mloubout: Yeah looks like it. Only nitpicking is that I'd rather set `self._sources` , .. rather than args but not important

mloubout (approved):

examples/performance/01_gpu.ipynb
```diff
@@ -307,8 +307,7 @@
       "    {\n",
```
georgebisbas, line 307: <p>so there is diff here, no temp?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/devitocodes/devito/pull/2446/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

devito/core/gpu.py
EdCaunt: Agreed

devito/core/operator.py
```diff
@@ -159,6 +169,9 @@ def _check_kwargs(cls, **kwargs):
         if oo['mpi'] and oo['mpi'] not in cls.MPI_MODES:
             raise InvalidOperator("Unsupported MPI mode `%s`" % oo['mpi'])
 
+        if oo['cse-algo'] not in ('basic', 'smartsort', 'advanced'):
+            raise InvalidArgument("Illegal `cse-algo` value")
```
EdCaunt, line 173: This error message should say what the illegal value was, much like the one above

devito/passes/clusters/cse.py
EdCaunt: We will also probably want a unique kind of variable name to improve readability. Something like `i0` rather than `r0`

devito/passes/clusters/cse.py
```diff
@@ -92,61 +126,98 @@ def _cse(maybe_exprs, make, min_cost=1, mode='default'):
...
+                scheduled[k.key] = v
+
+        processed.append(pe)
+
+    return processed
+
+
+def _compact(exprs, exclude):
+    """
+    Drop useless temporaries:
```
EdCaunt, line 202: Are there also edge cases where it creates expressions like `r1 = r0`? I'm sure we had some issue like this

tests/test_cse.py
FabioLuporini: very few modifications

devito/passes/clusters/cse.py
FabioLuporini: One of the roles of `_compact` is catching them all !

EdCaunt (approved):

PR #2452, opened 2024-09-11
compiler: Hotfix and simplify minimize_symbols

FabioLuporini:
Hotfix for the previous PR

### DIFF
--- a/devito/ir/iet/visitors.py
+++ b/devito/ir/iet/visitors.py
@@ -1229,7 +1229,10 @@ class Uxreplace(Transformer):
         dimension = uxreplace(o.dim, self.mapper)
         limits = [uxreplace(i, self.mapper) for i in o.limits]
         pragmas = self._visit(o.pragmas)
+
         uindices = [uxreplace(i, self.mapper) for i in o.uindices]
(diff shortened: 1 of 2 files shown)


devito/ir/iet/visitors.py
```diff
@@ -1229,7 +1229,10 @@ def visit_Iteration(self, o):
         dimension = uxreplace(o.dim, self.mapper)
         limits = [uxreplace(i, self.mapper) for i in o.limits]
         pragmas = self._visit(o.pragmas)
+
         uindices = [uxreplace(i, self.mapper) for i in o.uindices]
+        uindices = filter_ordered(i for i in uindices if isinstance(i, Dimension))
```
EdCaunt, line 1234: I presume you're liable to get `int` or some other type that doesn't have the attribute `is_Dimension` here?

EdCaunt (approved):

georgebisbas (approved):

devito/ir/iet/visitors.py
FabioLuporini: yes, whatever the caller passes in `self.mapper`

PR #2453, opened 2024-09-11
compiler: Minor tweaks for elastic code gen

FabioLuporini:


devito/arch/compiler.py
```diff
@@ -687,6 +687,10 @@ def __init_finalize__(self, **kwargs):
         if not configuration['safe-math']:
             self.cflags.append('--use_fast_math')
 
+        # Optionally print out per-kernel shared memory and register usage
+        if configuration['profiling'] == 'advanced2':
```
georgebisbas, line 691: cool!

georgebisbas (commented):
How is this affecting elastic?
Code is the same as I understand, do we get faster compilation?

devito/types/basic.py
```diff
@@ -490,6 +491,11 @@ def _cache_key(cls, *args, **kwargs):
         # From the kwargs
         key.update(kwargs)
 
+        # Any missing __rkwargs__ along with their default values
+        params = inspect.signature(cls.__init_finalize__).parameters
```
mloubout, line 495: Ouch how can we end up in such a weird spot

mloubout (commented):
minor comment but looks fine

devito/types/basic.py
FabioLuporini: try the newly added tests without this patch 😬 I don't remember the details, but basically caching bypassed because a different cache key gets computed

devito/finite_differences/tools.py
```diff
@@ -233,7 +233,7 @@ def make_stencil_dimension(expr, _min, _max):
     Create a StencilDimension for `expr` with unique name.
     """
     n = len(expr.find(StencilDimension))
-    return StencilDimension(name='i%d' % n, _min=_min, _max=_max)
+    return StencilDimension('i%d' % n, _min, _max)
```
EdCaunt, line 236: Nitpick: could these just be `min` and `max` now?

devito/symbolics/manipulation.py
```diff
@@ -244,6 +246,20 @@ def add(self, expr, make, terms=None):
...
 
+def subs_if_composite(expr, subs):
+    """
+    Call `expr.subs(subs)` if `subs` contain composite expressions, that is
+    expressions that can be part of larger expressions of the same type (e.g.,
+    `a*b` could be part of `a*b*c`, while `a[1]` cannot be part of a "larger
+    Indexed"). Instead, if `subs` consists of just "primitive" expressions, then
+    resort to the much faster `uxreplace`.
+    """
+    if all(isinstance(i, (Indexed, IndexDerivative)) for i in subs):
```
mloubout, line 257: So why can't this just be moved inside uxreplace?

devito/types/basic.py
mloubout: Sure but I don't get why this needs this elaborate inspect instead of just having StencilDimension implement `_cache_key` and add step/spacing

mloubout (approved):
Some minor comments but looks fine to me

devito/finite_differences/tools.py
FabioLuporini: they are special python keywords so we tend not to as it's not recommended

devito/symbolics/manipulation.py
```diff
@@ -244,6 +246,20 @@ def add(self, expr, make, terms=None):
             self[base] = self.extracted[base] = make()
 
 
+def subs_if_composite(expr, subs):
+    """
+    Call `expr.subs(subs)` if `subs` contain composite expressions, that is
```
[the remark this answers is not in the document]
FabioLuporini: noted

devito/symbolics/manipulation.py
FabioLuporini: because it'd contradict the API -- uxreplace performs no re-simplifications.

devito/types/basic.py
FabioLuporini: Im not sure what makes you think it's due to StencilDimension? the test maybe ? but it was not just that. Maybe it emerged from there, but the problem is way more general. In fact, IIRC the issue was the presence/absence of the `is_const` flag, which pops up after reconstruction but it's not part of the key (without this patch) at first instantiation

PR #2455, opened 2024-09-11
examples: make pytest import optional

mloubout:
Fix #2454

PR #2456, opened 2024-09-11
API: Support harmonic averaging for parameter Function

mloubout:
Needed for accurate elastic simulations

On top of #2455

### DIFF
--- a/devito/types/basic.py
+++ b/devito/types/basic.py
@@ -985,8 +985,9 @@ class AbstractFunction(sympy.Function, Basic, Pickable, Evaluable):
 
         # Averaging mode for off the grid evaluation
         self._avg_mode = kwargs.get('avg_mode', 'arithmetic')
-        assert self._avg_mode in ['arithmetic', 'harmonic'], "Accepted avg_mode " \
-            "values are 'arithmetic' or 'harmonic', invalid %s" % self._avg_mode
+        if self._avg_mode not in ['arithmetic', 'harmonic']:
+            raise ValueError("Invalid averaging mode_mode %s, accepted values are"
+                             " arithmetic or harmonic" % self._avg_mode)
 
     @classmethod
     def __args_setup__(cls, *args, **kwargs):
@@ -1152,10 +1153,13 @@ class AbstractFunction(sympy.Function, Basic, Pickable, Evaluable):
             return self
 
         # Base function
-        retval = 1 / self.function if self._avg_mode == 'harmonic' else self.function
+        if self._avg_mode == 'harmonic':
+            retval = 1 / self.function
+        else:
+            retval = self.function
         # Apply interpolation from inner most dim
         for d, i in self._grid_map.items():
-            retval = retval.diff(d, deriv_order=0, fd_order=2, x0={d: i}).evaluate
+            retval = retval.diff(d, deriv_order=0, fd_order=2, x0={d: i})
         if self._avg_mode == 'harmonic':
             retval = 1 / retval
 
--- a/devito/types/basic.py
+++ b/devito/types/basic.py
@@ -852,7 +852,7 @@ class AbstractFunction(sympy.Function, Basic, Pickable, Evaluable):
     """
 
     __rkwargs__ = ('name', 'dtype', 'grid', 'halo', 'padding', 'ghost',
-                   'alias', 'space', 'function', 'is_transient')
+                   'alias', 'space', 'function', 'is_transient', 'avg_mode')
 
     def __new__(cls, *args, **kwargs):
         # Preprocess arguments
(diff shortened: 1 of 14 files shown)


devito/types/basic.py
```diff
@@ -983,6 +983,11 @@ def __init_finalize__(self, *args, **kwargs):
         # certain optimizations, such as avoiding memory copies
         self._is_transient = kwargs.get('is_transient', False)
 
+        # Averaging mode for off the grid evaluation
+        self._avg_mode = kwargs.get('avg_mode', 'arithmetic')
+        assert self._avg_mode in ['arithmetic', 'harmonic'], "Accepted avg_mode " \
```
FabioLuporini, line 988: this shouldn't be an assert, but rather a `raise ValueError`

devito/types/basic.py
```diff
@@ -1147,13 +1152,16 @@ def _evaluate(self, **kwargs):
             return self
 
         # Base function
-        retval = self.function
+        retval = 1 / self.function if self._avg_mode == 'harmonic' else self.function
         # Apply interpolation from inner most dim
         for d, i in self._grid_map.items():
-            retval = retval.diff(d, 0, fd_order=2, x0={d: i})
+            retval = retval.diff(d, deriv_order=0, fd_order=2, x0={d: i}).evaluate
```
FabioLuporini, line 1158: Always a bit wary of these premature evaluations

devito/types/basic.py
```diff
@@ -1147,13 +1152,16 @@ def _evaluate(self, **kwargs):
...
+        if self._avg_mode == 'harmonic':
+            retval = 1 / retval
+
         # Evaluate. Since we used `self.function` it will be on the grid when evaluate
         # is called again within FD
-        return retval.evaluate
+        return retval.evaluate.expand()
```
FabioLuporini, line 1164: why the need to expand ?

also, not that we're calling `.evaluate` multiple times here -- above in the `for d, i in self._grid_map` loop at each iteration, and then again here.

THere's something weird about this

examples/seismic/acoustic/acoustic_example.py
```diff
@@ -1,5 +1,8 @@
 import numpy as np
-import pytest
+try:
+    import pytest
+except:
```
FabioLuporini, line 4: nitpicking: All these imports should be `except ImportError`

devito/types/basic.py
```diff
@@ -1147,13 +1152,16 @@ def _evaluate(self, **kwargs):
             return self
 
         # Base function
-        retval = self.function
+        retval = 1 / self.function if self._avg_mode == 'harmonic' else self.function
```
FabioLuporini, line 1155: nitpicking: for homogeneity with the rest of the codebase, it should be

```
if cond:
     <then part>
else:
     <else part>
```

even if it's a trivial code fragment. But again, nitpicking...

devito/types/basic.py
```diff
@@ -1147,13 +1152,16 @@ def _evaluate(self, **kwargs):
...
         # Apply interpolation from inner most dim
         for d, i in self._grid_map.items():
-            retval = retval.diff(d, 0, fd_order=2, x0={d: i})
+            retval = retval.diff(d, deriv_order=0, fd_order=2, x0={d: i}).evaluate
+        if self._avg_mode == 'harmonic':
+            retval = 1 / retval
+
         # Evaluate. Since we used `self.function` it will be on the grid when evaluate
```
FabioLuporini, line 1162: does this comment still apply?

examples/seismic/acoustic/acoustic_example.py
georgebisbas: I thought we needed exactly the opposite?..e.g. pass

examples/seismic/model.py
```diff
@@ -304,7 +307,11 @@ def _initialize_physics(self, vp, space_order, **kwargs):
             vs = kwargs.pop('vs')
             self.lam = self._gen_phys_param((vp**2 - 2. * vs**2)/b, 'lam', space_order,
                                             is_param=True)
-            self.mu = self._gen_phys_param(vs**2 / b, 'mu', space_order, is_param=True)
+            # Need to add small value to avoid division by zero
+            if isinstance(vs, np.ndarray):
```
EdCaunt, line 311: Wouldn't this still be the case if `vs` was a float? Where does the div by zero come from?

examples/seismic/acoustic/acoustic_example.py
FabioLuporini: ???

examples/seismic/model.py
FabioLuporini: user input setting vs=0 ?

devito/types/basic.py
mloubout: Without exapnd you get horrible stuff like 05*(0.5*(0.5 ... instead of 0.125.

This is a trivial square/cube averaging that should be a plain expression anything else is just messing up with the compiler for no reason

examples/seismic/model.py
mloubout: >Where does the div by zero come from?

Harmonic mean for half noee evaluation. `1/(.5/mu[x] + .5/mu[x+1]))`

If `vs` is just a float there is no averaging

FabioLuporini (approved):

PR #2457, opened 2024-09-17
misc: Add more helpful error for time bounds edge case

EdCaunt:
Adds a better error message for certain edge cases where `time_M` (or nominally any other dimension) is arg checked inside `Operator._prepare_arguments`.


EdCaunt:
Correct fix appears to be elsewhere. Closing.

PR #2459, opened 2024-09-27
compiler: Propagate metadata down to _arg_defaults

FabioLuporini:
This is necessary for optimization purposes in PRO


mloubout (approved):

devito/passes/iet/definitions.py
```diff
@@ -470,30 +470,22 @@ def _map_function_on_high_bw_mem(self, site, obj, storage, devicerm, read_only=F
...
-                eq = Eq(obj[cdims], 0)
-
-                irs, _ = self.rcompile(eq)
-
-                init = irs.iet.body.body[0]
-
-                name = self.sregistry.make_name(prefix='init')
-                efunc = make_callable(name, init)
-                init = Call(name, efunc.parameters)
+                efuncs, init = make_zero_init(obj, self.rcompile, self.sregistry)
```
mloubout, line 473: Can we get read of `gpu-create` and just do `if is_transient` and have the zero init in the core no matter?

FabioLuporini: as discussed offline, we would break user code

mloubout: We wouldn't we just need `is_gou_create` to return `true` for any `is_tansient` for now so that any transient function is directly allocated in gen code and add a deprecation warning

georgebisbas (approved):

devito/passes/iet/definitions.py
FabioLuporini: if the users who have been using gpu-create don't start creating `is_transient` Functions, how can it possibly work?

I agree though that gpu-create should get deprecated

EdCaunt (approved):

devito/passes/iet/definitions.py
mloubout: Well yes why I said as a first step `is_gpu_create` should behave as now, but returns `true` as well for `is_transient` Function and throw a warning to tell people "stop using gpu-create and set your function with is_transient"

PR #2461, opened 2024-10-03
arch: Avoid deprecated compiler flags

FabioLuporini:
recent NVC has deprecated gpu=pinned and suggested to rather use what this patch uses

### DIFF
--- a/devito/arch/compiler.py
+++ b/devito/arch/compiler.py
@@ -47,6 +47,7 @@ def sniff_compiler_version(cc, allow_fail=False):
         else:
             raise RuntimeError("The `%s` compiler isn't available on this system" % cc)
 
+    ver = ver.strip()
     if ver.startswith("gcc"):
         compiler = "gcc"
     elif ver.startswith("clang"):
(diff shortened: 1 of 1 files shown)


georgebisbas (commented):
WOuld this still work w previous nvc versions I guess?

devito/arch/compiler.py
```diff
@@ -601,7 +601,7 @@ def __init_finalize__(self, **kwargs):
         platform = kwargs.pop('platform', configuration['platform'])
 
         if platform is NVIDIAX:
-            self.cflags.append('-gpu=pinned')
+            self.cflags.append('-gpu=mem:separate:pinnedalloc')
```
mloubout, line 604: probably need a version check like icx for openmp

EdCaunt (approved):

devito/arch/compiler.py
FabioLuporini: (@georgebisbas ) pushed tweak so that older versions are guaranteed to work

georgebisbas (approved):
Thanks!

mloubout (approved):

PR #2470, opened 2024-10-18
dsl: Move SubDimension thickness evaluation to the thicknesses themselves

EdCaunt:
Fixes issues with standalone use of `SubDimension.symbolic_min` or `SubDimension.symbolic_max` in an operator. Previously variants on
```
x = Dimension('x')
ix = SubDimension.left('ix', x, 2)
f = Function(name="f", dimensions=(ix,), shape=(5,), dtype=np.int32)

eqns = Eq(f[ix.symbolic_max], 1)

op = Operator(eqns)
op(x_m=0)
```
would suffer from two issues:

1. The SubDimension thickness (`ix.symbolic_max`) would not be concretised from `x_ltkn` to `x_ltkn0`
2. The SubDimension thickness would never be given a value by `_arg_values` unless the user explicitly passed it to the operator (and even then, it would not be adjusted to reflect any MPI decomposition)

These are rectified by this PR.


devito/operator/operator.py
```diff
@@ -640,6 +642,11 @@ def _prepare_arguments(self, autotune=None, **kwargs):
         for d in reversed(toposort):
             args.update(d._arg_values(self._dspace[d], grid, **kwargs))
 
+        # Process SubDimensionThicknesses
+        for p in self.parameters:
```
FabioLuporini, line 646: I think this should be triggered from within `SubDimension._arg_values` itself

devito/types/dimension.py
```diff
@@ -536,6 +535,81 @@ def _arg_check(self, *args, **kwargs):
 # The Dimensions below are exposed in the user API. They can only be created by
 # the user
 
+class SubDimensionThickness(DataSymbol):
```
FabioLuporini, line 538: this should be called just "Thickness", it's a first-class citizen now

devito/types/dimension.py
```diff
@@ -536,6 +535,81 @@ def _arg_check(self, *args, **kwargs):
...
+    """A DataSymbol to represent a thickness of a SubDimension"""
```
FabioLuporini, line 539: full stop

devito/types/dimension.py
```diff
@@ -536,6 +535,81 @@ def _arg_check(self, *args, **kwargs):
...
+
+    __rkwargs__ = DataSymbol.__rkwargs__ + ('root', 'side', 'local', 'value')
+
+    def __new__(cls, *args, **kwargs):
```
FabioLuporini, line 543: just put root=None etc in the signature instead of popping from kwargs?

devito/types/dimension.py
```diff
@@ -716,89 +801,17 @@ def bound_symbols(self):
...
-            else:
-                # dimension is of type ``middle``
-                ltkn = grid.distributor.glb_to_loc(self.root, r_ltkn, LEFT) or 0
-                rtkn = grid.distributor.glb_to_loc(self.root, r_rtkn, RIGHT) or 0
-        else:
-            ltkn = r_ltkn or 0
-            rtkn = r_rtkn or 0
-
-        return {i.name: v for i, v in zip(self._thickness_map, (ltkn, rtkn))}
+        # SubDimension thicknesses at runtime are calculated by the thicknesses
```
FabioLuporini, line 812: as I wrote in another comment, this SubDimension should trigger the processing of its own arguments, namely its own Thicknesses

FabioLuporini (commented):
Looks like a very nice clean up!

devito/types/dimension.py
EdCaunt: Thought we didn't do those on single line docstrings?

devito/types/dimension.py
EdCaunt: Maybe I should swap the `namedtuple` "Thickness" and `SubDomainThickness` class names?

devito/types/dimension.py
```diff
@@ -536,6 +535,81 @@ def _arg_check(self, *args, **kwargs):
 # The Dimensions below are exposed in the user API. They can only be created by
 # the user
 
+class SubDimensionThickness(DataSymbol):
+    """A DataSymbol to represent a thickness of a SubDimension"""
+
+    __rkwargs__ = DataSymbol.__rkwargs__ + ('root', 'side', 'local', 'value')
+
+    def __new__(cls, *args, **kwargs):
+        root = kwargs.pop('root')
```
mloubout, line 544: just make it explicit `__new__(cls, *args, root=None, local=None, side=None, **kwargs):

devito/types/dimension.py
```diff
@@ -659,49 +735,58 @@ class SubDimension(AbstractSubDimension):
...
-        return cls(name, parent,
-                   left=parent.symbolic_min+lst,
-                   right=parent.symbolic_max-rst,
-                   thickness=((lst, thickness_left), (rst, thickness_right)),
-                   local=local)
+        return cls(name, parent, thickness=(thickness_left, thickness_right), local=local)
+
+    @memoized_meth
+    def _symbolic_thickness(self, thickness=None):
+        kwargs = {'dtype': np.int64, 'is_const': True, 'nonnegative': True,
```
mloubout, line 759: why int64 here and int32 above?

mloubout (commented):
some mnior comments

devito/operator/operator.py
EdCaunt: No, the issue is that the `SubDimension` isn't necessarily available to the operator. You can craft legal `Operator`s which only contain the thickness, not the parent subdimension, hence the reworking in this PR

devito/types/dimension.py
EdCaunt: They shouldn't (or at least it would be redundant if it did), see my other comment

devito/types/dimension.py
EdCaunt: Leftover from an earlier iteration I think

devito/ir/equations/algorithms.py
```diff
@@ -213,18 +213,24 @@ def _(expr, mapper, rebuilt, sregistry):
...
         return
 
-    tkns_subs = {tkn: tkn._rebuild(name=sregistry.make_name(prefix=tkn.name))
-                 for tkn in d.tkns}
-    left, right = [mM.subs(tkns_subs) for mM in (d.symbolic_min, d.symbolic_max)]
-    thickness = tuple((v, d._thickness_map[k]) for k, v in tkns_subs.items())
-
-    mapper[d] = d._rebuild(symbolic_min=left, symbolic_max=right, thickness=thickness)
+    tkns = tuple(t._rebuild(name=sregistry.make_name(prefix=t.name)) for t in d.tkns)
+    mapper.update({t0: t1 for t0, t1 in zip(d.tkns, tkns)})
```
georgebisbas, line 232: I would avoid t0 and t1 due to name clash with time buffers.
It's not about correctness, but mostly when grepping

devito/types/dimension.py
```diff
@@ -536,6 +535,77 @@ def _arg_check(self, *args, **kwargs):
...
+        # So overriding the thickness to a nonzero value should not cause
+        # boundaries to exist between ranks where they did not before
+        r_tkn = kwargs.get(self.name, self.value)
+        if grid is not None and grid.is_distributed(self.root):
+            # Get local thickness
+            if self.local:
+                # Dimension is of type `left`/`right` - compute the offset
+                # and then add 1 to get the appropriate thickness
+                if self.value is not None:
+                    tkn = grid.distributor.glb_to_loc(self.root, r_tkn-1, self.side)
```
georgebisbas, line 590: why r_tkn-1?

devito/types/dimension.py
```diff
@@ -583,28 +663,20 @@ def symbolic_size(self):
...
         return self._thickness
 
-    @cached_property
-    def _thickness_map(self):
-        return dict(self.thickness)
-
-    @property
-    def is_abstract(self):
-        return all(i is None for i in flatten(self.thickness))
+    tkns = thickness  # Shortcut for thickness
```
georgebisbas, line 666: should this not be a cached propertry?

devito/types/dimension.py
EdCaunt: Retained from SubDimension._arg_values iirc

tests/test_caching.py
```diff
@@ -684,14 +684,16 @@ def test_sparse_function(self, operate_on_empty_cache):
...
-        assert len(_SymbolCache) == init_cache_size + 2
+        # pos* that belong to the abstract class and the dimension
+        # bounds (x_m, x_M, y_m, y_M)
+        assert len(_SymbolCache) == init_cache_size + 6
         clear_cache()
-        # Now we should be back to the original state
-        assert len(_SymbolCache) == init_cache_size
+        # Now we should be back to the original state plus the dimension bounds
+        # (x_m, x_M, y_m, y_M)
+        assert len(_SymbolCache) == init_cache_size + 4
```
mloubout, line 696: That's not good, why is there a leftover 4?

EdCaunt: `x_m, x_M, y_m, y_M` are not longer touched at grid instantiation, and so are only created and cached at `i = u.inject(expr=u, field=u)`. However, the grid and subdimension thickness symbols persist and thus these symbols persist in the cache to my understanding

mloubout: But is there a  cache leak? If it needs one more cache clear to remove it that's ok (and you should add an extra line testing it gets back to `init_cache_size ` after one more) but if there is a reference that cannot be reached then it's problematic.

EdCaunt: How is it determined that an object should be removed from cache? If there's no other objects referencing it?

mloubout: Yes. The cache is no different than the standard python GC, it clears objects when there are no more references to it, so if it cannot be cleared then there is unreachable references

EdCaunt: Ok, added additional testing to check that all symbols disappear once the grid goes out of scope

devito/operator/operator.py
mloubout: Why isn't it caught by `self.input` 's default line 487

tests/test_caching.py
```diff
@@ -684,14 +684,21 @@ def test_sparse_function(self, operate_on_empty_cache):
...
+        # Delete the grid and check that all symbols are subsequently garbage collected
+        del grid
+        for n in (10, 3, 0):
```
mloubout, line 699: Is 10 `init_cache_size` ? If so just 
```
del grid
assert len(_SymbolCache) == init_cache_size
```

And everything was deleted that's that's needed to check

devito/operator/operator.py
EdCaunt: Assuming "it" means the subdomain, you can craft a legal operator as in the MFE at the top of this PR where the Operator never sees the SubDimension, only its thickness symbols

tests/test_caching.py
EdCaunt: Init cache size is 11 (and consists of a different set of symbols)

devito/operator/operator.py
mloubout: I don't mean the subdomain no, I mean the thickness, you're saying you can create an operator where the Thickness (p here) is in op.parameters but is not in op.inputs?

If it's the case, i.e if it is *not* in inputs, then it should not need to be in args since it's not an input.
If it's not the case, the line 487 processes all inputs and should cover that case already

EdCaunt: What do you mean? Several things that need `_arg_values` calling on them are in `self.parameters` and not `self.input` (`x_m` and `x_M` for example).  The argument I'm making is the thickness `ltkn` can appear in an operator without its parent subdimension `ix`

FabioLuporini: I **think** if you add `is_Input = True` to `class Thickness`, then you don't need this extra loop

for example  https://github.com/devitocodes/devito/blob/master/devito/types/constant.py#L42

devito/types/dimension.py
```diff
@@ -536,6 +535,77 @@ def _arg_check(self, *args, **kwargs):
 # The Dimensions below are exposed in the user API. They can only be created by
 # the user
 
+class Thickness(DataSymbol):
```
FabioLuporini, line 538: To be extra clear, but not required for this PR, we have two types of thicknesses:

1) those for classic SubDomains (which inherit from DataSymbol)
2) those that are "compiler generated" and therefore invisible to the user, like those created by an ABox (so we may need AbstractThickness, etc etc)

But again, for another day

devito/types/dimension.py
```diff
@@ -536,6 +535,77 @@ def _arg_check(self, *args, **kwargs):
...
+        # TODO: Should check that the thickness isn't larger than the parent
+        # domain
+        pass
+
+    def _arg_values(self, grid=None, **kwargs):
+        # Allow override of thickness values to disable BCs
+        # However, arguments from the user are considered global
+        # So overriding the thickness to a nonzero value should not cause
+        # boundaries to exist between ranks where they did not before
+        r_tkn = kwargs.get(self.name, self.value)
```
FabioLuporini, line 583: nitpicking: for consistency, it should be named "rtkn", not "r_tkn" -- or at least, that's the way I've been naming thickness variables

devito/types/dimension.py
```diff
@@ -536,6 +535,77 @@ def _arg_check(self, *args, **kwargs):
...
+                tkn = grid.distributor.glb_to_loc(self.root, r_tkn, self.side) or 0
+        else:
+            tkn = r_tkn or 0
+
+        return {self.name: tkn}
+
+    def _arg_finalize(self, *args, **kwargs):
+        return {}
+
+    def _arg_apply(self, *args, **kwargs):
```
FabioLuporini, line 605: drop this, irrelevant

devito/types/dimension.py
```diff
@@ -536,6 +535,77 @@ def _arg_check(self, *args, **kwargs):
...
+                    tkn = 0
+            else:
+                # Dimension is of type `middle`
+                tkn = grid.distributor.glb_to_loc(self.root, r_tkn, self.side) or 0
+        else:
+            tkn = r_tkn or 0
+
+        return {self.name: tkn}
+
+    def _arg_finalize(self, *args, **kwargs):
```
FabioLuporini, line 602: drop this, irrelevant

devito/types/dimension.py
FabioLuporini: never mind, ignore the above... I see what you've done below now. Excellent

devito/types/dimension.py
```diff
@@ -659,49 +731,59 @@ class SubDimension(AbstractSubDimension):
...
-                   local=local)
+        return cls(name, parent, thickness=(thickness_left, thickness_right), local=local)
+
+    @memoized_meth
+    def _symbolic_thickness(self, thickness=None):
+        kwargs = {'dtype': np.int32, 'is_const': True, 'nonnegative': True,
+                  'root': self.root, 'local': self.local}
+
+        names = ["%s_%stkn" % (self.parent.name, s) for s in ('l', 'r')]
+        return SubDimensionThickness(*[Thickness(name=n, side=s, value=t, **kwargs)
```
FabioLuporini, line 759: nitpicking, aesthetically this could probably be improved by breaking lines in a smarter way

devito/types/dimension.py
FabioLuporini: gotcha

devito/types/dimension.py
EdCaunt: A no-op `_arg_check`, `_arg_finalize`, and `_arg_apply` are required for `_prepare_arguments`

devito/operator/operator.py
EdCaunt: I tried it, and this is not the case

mloubout: How is that possible? It's clearly in parameters since you added this so if you add is_Input and it doesn't catch it then something is broken

EdCaunt: There's no grid in the kwargs at the point where `_arg_values` is called on the inputs, which makes it no-op. It also needs to be done before objects get `_arg_values` called on them iirc, hence the current position of this loop.

You could reorder `_prepare_arguments` to process the grid earlier, but my functions-on-subdomains branch has a bunch of reworking here and fixing the merge conflicts will be a pain, so best to stick a pin in it and fix down the line imo.

mloubout: Can you open an issue about it then

mloubout (approved):

devito/operator/operator.py
FabioLuporini: I also don't understand this thing honestly. We need an issue and ideally a comment on top of this loop explaining what's going on and what should actually happen instead. You can do it in another branch as it's a minor thing

devito/types/dimension.py
FabioLuporini: this should be fixed at the superclass site then, somehow, but fine..

FabioLuporini (approved):

FabioLuporini:
Merged, thanks

PR #2471, opened 2024-10-20
api: enforce sympy shifts

mloubout:
Was preventing this

https://github.com/devitocodes/devito/blob/779522508a45d677f233811846fcc7eb36cea916/examples/seismic/tti/operators.py#L158

to simplify leading to `u.dx2 + u.dy2 - u.dy2` expressions....


FabioLuporini:
merged, thanks

PR #2474, opened 2024-10-26
Implement snapshotting for the acoustic wave equation

malfarhan7:
Implement snapshotting to save snapshots of the forward wavefield used to compute the gradient to reduce memory usage.

### DIFF
--- a/examples/seismic/acoustic/operators.py
+++ b/examples/seismic/acoustic/operators.py
@@ -1,4 +1,4 @@
-from devito import Eq, Operator, Function, TimeFunction, Inc, solve, sign
+from devito import Eq, Operator, Function, TimeFunction, Inc, solve, sign, ConditionalDimension
 from devito.symbolics import retrieve_functions, INT, retrieve_derivatives
 
 
@@ -108,7 +108,7 @@ def iso_stencil(field, model, kernel, **kwargs):
 
 
 def ForwardOperator(model, geometry, space_order=4,
-                    save=False, kernel='OT2', **kwargs):
+                    save=False, kernel='OT2', factor=None, **kwargs):
     """
     Construct a forward modelling operator in an acoustic medium.
 
@@ -126,6 +126,8 @@ def ForwardOperator(model, geometry, space_order=4,
         Defaults to False.
     kernel : str, optional
         Type of discretization, 'OT2' or 'OT4'.
+    factor : int, optional
+        Downsampling factor to save snapshots of the wavefield.
     """
     m = model.m
 
@@ -144,10 +146,28 @@ def ForwardOperator(model, geometry, space_order=4,
 
     # Create interpolation expression for receivers
     rec_term = rec.interpolate(expr=u)
+    # Build operator equations
+    equations = eqn + src_term + rec_term
 
+    if factor:
+        # Implement snapshotting
+        nsnaps = (geometry.nt + factor - 1) // factor
+        time_subsampled = ConditionalDimension(
+            't_sub', parent=model.grid.time_dim, factor=factor)
+        usnaps = TimeFunction(name='usnaps', grid=model.grid,
+                              time_order=2, space_order=space_order,
+                              save=nsnaps, time_dim=time_subsampled)
+        # Add equation to save snapshots
+        snapshot_eq = Eq(usnaps, u)
+        equations += [snapshot_eq]
+    else:
+        usnaps = None
     # Substitute spacing terms to reduce flops
-    return Operator(eqn + src_term + rec_term, subs=model.spacing_map,
-                    name='Forward', **kwargs)
+    op = Operator(equations, subs=model.spacing_map, name='Forward', **kwargs)
+    if usnaps is not None:
+        return op, usnaps
+    else:
+        return op
 
 
 def AdjointOperator(model, geometry, space_order=4,
@@ -189,8 +209,8 @@ def AdjointOperator(model, geometry, space_order=4,
 
 
 def GradientOperator(model, geometry, space_order=4, save=True,
-                     kernel='OT2', **kwargs):
-    """
+                     kernel='OT2', factor=None, **kwargs):
+     """
     Construct a gradient operator in an acoustic media.
 
     Parameters
@@ -206,30 +226,59 @@ def GradientOperator(model, geometry, space_order=4, save=True,
         Option to store the entire (unrolled) wavefield.
     kernel : str, optional
         Type of discretization, centered or shifted.
+    factor : int, optional
+        Downsampling factor to save snapshots of the wavefield.
     """
     m = model.m
 
-    # Gradient symbol and wavefield symbols
+    # Gradient symbol
     grad = Function(name='grad', grid=model.grid)
-    u = TimeFunction(name='u', grid=model.grid, save=geometry.nt if save
-                     else None, time_order=2, space_order=space_order)
-    v = TimeFunction(name='v', grid=model.grid, save=None,
-                     time_order=2, space_order=space_order)
-    rec = geometry.rec
+
+    # Create the adjoint wavefield
+    v = TimeFunction(name='v', grid=model.grid, time_order=2, space_order=space_order)
 
     s = model.grid.stepping_dim.spacing
     eqn = iso_stencil(v, model, kernel, forward=False)
 
-    if kernel == 'OT2':
-        gradient_update = Inc(grad, - u * v.dt2)
-    elif kernel == 'OT4':
-        gradient_update = Inc(grad, - u * v.dt2 - s**2 / 12.0 * u.biharmonic(m**(-2)) * v)
     # Add expression for receiver injection
+    rec = geometry.rec
     receivers = rec.inject(field=v.backward, expr=rec * s**2 / m)
 
+    time = model.grid.time_dim
+
+    if factor is not None:
+        # Condition to apply gradient update only at snapshot times
+        condition = Eq(time % factor, 0)
+        # Create the ConditionalDimension for subsampling
+        time_subsampled = ConditionalDimension('t_sub', parent=time, factor=factor)
+        # Define usnaps with time_subsampled as its time dimension
+        nsnaps = (geometry.nt + factor - 1) // factor
+        usnaps = TimeFunction(name='usnaps', grid=model.grid,
+                              time_order=2, space_order=space_order,
+                              save=nsnaps, time_dim=time_subsampled)
+        # Gradient update without indexing usnaps
+        if kernel == 'OT2':
+            gradient_update = Inc(grad, - usnaps * v.dt2, implicit_dims=[time_subsampled],
+                                  condition=condition)
+        elif kernel == 'OT4':
+            gradient_update = Inc(grad, - usnaps * v.dt2
+                                  - s**2 / 12.0 * usnaps.biharmonic(m**(-2)) * v,
+                                  implicit_dims=[time_subsampled],
+                                  condition=condition)
+    else:
+        u = TimeFunction(name='u', grid=model.grid,
+                         save=geometry.nt if save else None,
+                         time_order=2, space_order=space_order)
+        if kernel == 'OT2':
+            gradient_update = Inc(grad, - u * v.dt2)
+        elif kernel == 'OT4':
+            gradient_update = Inc(grad, - u * v.dt2
+                                  - s**2 / 12.0 * u.biharmonic(m**(-2)) * v)
+            
     # Substitute spacing terms to reduce flops
-    return Operator(eqn + receivers + [gradient_update], subs=model.spacing_map,
-                    name='Gradient', **kwargs)
+    op = Operator(eqn + receivers + [gradient_update], subs=model.spacing_map,
+                  name='Gradient', **kwargs)
+    return op
 
 
 def BornOperator(model, geometry, space_order=4,
--- a/examples/seismic/acoustic/wavesolver.py
+++ b/examples/seismic/acoustic/wavesolver.py
@@ -1,6 +1,6 @@
 from devito import Function, TimeFunction, DevitoCheckpoint, CheckpointOperator, Revolver
 from devito.tools import memoized_meth
-from examples.seismic.acoustic.operators import (
+from devitofwi.devito.acoustic.operators import (
     ForwardOperator, AdjointOperator, GradientOperator, BornOperator
 )
 
@@ -23,6 +23,7 @@ class AcousticWaveSolver:
     space_order: int, optional
         Order of the spatial stencil discretisation. Defaults to 4.
     """
+
     def __init__(self, model, geometry, kernel='OT2', space_order=4, **kwargs):
         self.model = model
         self.model._initialize_bcs(bcs="damp")
@@ -44,11 +45,11 @@ class AcousticWaveSolver:
         return self.model.critical_dt
 
     @memoized_meth
-    def op_fwd(self, save=None):
+    def op_fwd(self, save=None, factor=None):
         """Cached operator for forward runs with buffered wavefield"""
         return ForwardOperator(self.model, save=save, geometry=self.geometry,
                                kernel=self.kernel, space_order=self.space_order,
-                               **self._kwargs)
+                               factor=factor, **self._kwargs)
 
     @memoized_meth
     def op_adj(self):
@@ -58,11 +59,11 @@ class AcousticWaveSolver:
                                **self._kwargs)
 
     @memoized_meth
-    def op_grad(self, save=True):
+    def op_grad(self, save=True, factor=None):
         """Cached operator for gradient runs"""
         return GradientOperator(self.model, save=save, geometry=self.geometry,
                                 kernel=self.kernel, space_order=self.space_order,
-                                **self._kwargs)
+                                factor=factor, **self._kwargs)
 
     @memoized_meth
     def op_born(self):
@@ -71,7 +72,7 @@ class AcousticWaveSolver:
                             kernel=self.kernel, space_order=self.space_order,
                             **self._kwargs)
 
-    def forward(self, src=None, rec=None, u=None, model=None, save=None, **kwargs):
+    def forward(self, src=None, rec=None, u=None, model=None, save=None, factor=None, **kwargs):
         """
         Forward modelling function that creates the necessary
         data objects for running a forward modelling operator.
@@ -90,6 +91,8 @@ class AcousticWaveSolver:
             The time-constant velocity.
         save : bool, optional
             Whether or not to save the entire (unrolled) wavefield.
+        factor : int, optional
+        Downsampling factor to save snapshots of the wavefield.
 
         Returns
         -------
@@ -108,12 +111,24 @@ class AcousticWaveSolver:
         model = model or self.model
         # Pick vp from model unless explicitly provided
         kwargs.update(model.physical_params(**kwargs))
+        # Get the operator
+        op_fwd = self.op_fwd(save=save, factor=factor)
+        # Prepare parameters for operator apply
+        op_args = {'src': src, 'rec': rec, 'u': u, 'dt': kwargs.pop('dt', self.dt)}
+        op_args.update(kwargs)
 
         # Execute operator and return wavefield and receiver data
-        summary = self.op_fwd(save).apply(src=src, rec=rec, u=u,
-                                          dt=kwargs.pop('dt', self.dt), **kwargs)
-
-        return rec, u, summary
+        if factor:
+            # Operator returned is op, usnaps
+            op, usnaps = op_fwd
+            op_args['usnaps'] = usnaps
+            summary = op.apply(**op_args)
+            
+        else:
+            op = op_fwd
+            usnaps = None
+            summary = op.apply(**op_args)
+        return rec, u, usnaps, summary
 
     def adjoint(self, rec, srca=None, v=None, model=None, **kwargs):
         """
@@ -155,8 +170,8 @@ class AcousticWaveSolver:
                                       dt=kwargs.pop('dt', self.dt), **kwargs)
         return srca, v, summary
 
-    def jacobian_adjoint(self, rec, u, src=None, v=None, grad=None, model=None,
-                         checkpointing=False, **kwargs):
+    def jacobian_adjoint(self, rec, u=None, usnaps=None, src=None, v=None, grad=None, model=None,
+                         factor=None, checkpointing=False, **kwargs):
         """
         Gradient modelling function for computing the adjoint of the
         Linearized Born modelling function, ie. the action of the
@@ -168,6 +183,8 @@ class AcousticWaveSolver:
             Receiver data.
         u : TimeFunction
             Full wavefield `u` (created with save=True).
+        usnaps : TimeFunction
+            Snapshots of the wavefield `u`.
         v : TimeFunction, optional
             Stores the computed wavefield.
         grad : Function, optional
@@ -176,12 +193,22 @@ class AcousticWaveSolver:
             Object containing the physical parameters.
         vp : Function or float, optional
             The time-constant velocity.
+        checkpointing : boolean, optional 
+            Flag to enable checkpointing (default False). 
+            Cannot be used with snapshotting.
+        factor : int, optional
+            Downsampling factor for the saved snapshots of the wavefield `u`.
+            Cannot be used with checkpointing.
 
         Returns
         -------
         Gradient field and performance summary.
         """
         dt = kwargs.pop('dt', self.dt)
+        # Check that snapshotting and checkpointing are not used together
+        if factor is not None and checkpointing:
+            raise ValueError("Cannot use snapshotting (factor) and checkpointing simultaneously.")
+
         # Gradient symbol
         grad = grad or Function(name='grad', grid=self.model.grid)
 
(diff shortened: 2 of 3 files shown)


examples/seismic/acoustic/operators.py
```diff
@@ -144,10 +146,28 @@ def ForwardOperator(model, geometry, space_order=4,
 
     # Create interpolation expression for receivers
     rec_term = rec.interpolate(expr=u)
-
+    # Build operator equations
+    equations = eqn + src_term + rec_term
+
+    if factor:
```
mloubout, line 152: This needs to be wrapped into a utility function as it's duplicated below

examples/seismic/acoustic/operators.py
```diff
@@ -144,10 +146,28 @@ def ForwardOperator(model, geometry, space_order=4,
...
+        # Implement snapshotting
+        nsnaps = (geometry.nt + factor - 1) // factor
+        time_subsampled = ConditionalDimension(
+            't_sub', parent=model.grid.time_dim, factor=factor)
+        usnaps = TimeFunction(name='usnaps', grid=model.grid,
```
mloubout, line 157: You still have `u` with full time saved line 135 you can't have both

examples/seismic/acoustic/operators.py
```diff
@@ -144,10 +146,28 @@ def ForwardOperator(model, geometry, space_order=4,
...
+        snapshot_eq = Eq(usnaps, u)
+        equations += [snapshot_eq]
+    else:
+        usnaps = None
     # Substitute spacing terms to reduce flops
-    return Operator(eqn + src_term + rec_term, subs=model.spacing_map,
-                    name='Forward', **kwargs)
+    op = Operator(equations, subs=model.spacing_map, name='Forward', **kwargs)
+    if usnaps is not None:
+        return op, usnaps
```
mloubout, line 168: No the operator build cannot return objects like that. This is an abstract operator with placeholders that might not be correct for runtime.

examples/seismic/acoustic/operators.py
```diff
@@ -206,30 +226,59 @@ def GradientOperator(model, geometry, space_order=4, save=True,
...
-        gradient_update = Inc(grad, - u * v.dt2 - s**2 / 12.0 * u.biharmonic(m**(-2)) * v)
     # Add expression for receiver injection
+    rec = geometry.rec
     receivers = rec.inject(field=v.backward, expr=rec * s**2 / m)
 
+    time = model.grid.time_dim
+
+    if factor is not None:
+        # Condition to apply gradient update only at snapshot times
+        condition = Eq(time % factor, 0)
```
mloubout, line 251: No you don't need that `usnap` already contains the conditon

examples/seismic/acoustic/operators.py
```diff
@@ -206,30 +226,59 @@ def GradientOperator(model, geometry, space_order=4, save=True,
...
+        elif kernel == 'OT4':
+            gradient_update = Inc(grad, - usnaps * v.dt2
+                                  - s**2 / 12.0 * usnaps.biharmonic(m**(-2)) * v,
+                                  implicit_dims=[time_subsampled],
+                                  condition=condition)
+    else:
+        u = TimeFunction(name='u', grid=model.grid,
+                         save=geometry.nt if save else None,
+                         time_order=2, space_order=space_order)
+        if kernel == 'OT2':
```
mloubout, line 272: Unnecessary duplicate, `u` contains the information you should not need separate cases for `gradient_update`

examples/seismic/acoustic/wavesolver.py
```diff
@@ -1,6 +1,6 @@
 from devito import Function, TimeFunction, DevitoCheckpoint, CheckpointOperator, Revolver
 from devito.tools import memoized_meth
-from examples.seismic.acoustic.operators import (
+from devitofwi.devito.acoustic.operators import (
```
mloubout, line 3: leftover?

examples/seismic/acoustic/wavesolver.py
```diff
@@ -108,12 +111,24 @@ def forward(self, src=None, rec=None, u=None, model=None, save=None, **kwargs):
         model = model or self.model
         # Pick vp from model unless explicitly provided
         kwargs.update(model.physical_params(**kwargs))
+        # Get the operator
+        op_fwd = self.op_fwd(save=save, factor=factor)
+        # Prepare parameters for operator apply
```
mloubout, line 116: Don't know what this is for.

examples/seismic/acoustic/wavesolver.py
```diff
@@ -108,12 +111,24 @@ def forward(self, src=None, rec=None, u=None, model=None, save=None, **kwargs):
...
+        op_args = {'src': src, 'rec': rec, 'u': u, 'dt': kwargs.pop('dt', self.dt)}
+        op_args.update(kwargs)
 
         # Execute operator and return wavefield and receiver data
-        summary = self.op_fwd(save).apply(src=src, rec=rec, u=u,
-                                          dt=kwargs.pop('dt', self.dt), **kwargs)
-
-        return rec, u, summary
+        if factor:
```
mloubout, line 121: No, usnap needs to be create here like `u` then passed as argument

examples/seismic/acoustic/wavesolver.py
```diff
@@ -108,12 +111,24 @@ def forward(self, src=None, rec=None, u=None, model=None, save=None, **kwargs):
...
+            # Operator returned is op, usnaps
+            op, usnaps = op_fwd
+            op_args['usnaps'] = usnaps
+            summary = op.apply(**op_args)
+            
+        else:
```
mloubout, line 127: shouldn't need if else only kwargs

examples/seismic/acoustic/wavesolver.py
```diff
@@ -209,8 +236,17 @@ def jacobian_adjoint(self, rec, u, src=None, v=None, grad=None, model=None,
             wrp.apply_forward()
             summary = wrp.apply_reverse()
         else:
-            summary = self.op_grad().apply(rec=rec, grad=grad, v=v, u=u, dt=dt,
-                                           **kwargs)
+            if factor is not None:
```
mloubout, line 239: Again, not needed, input `u` should contain all metada needed

mloubout (changes requested):
Thank you for the contribution!

I have left some comments as it needs some changes to be mergeable. Some kind of test also needs to be added so that it is maintainable.

examples/seismic/acoustic/wavesolver.py
malfarhan7: Hi Mathias, thank you for your feedback. I have reviewed and cleaned the code to the best of my understanding. I have included a notebook to compare computing the FWI gradient with and without snapshotting and two scripts to calculate the memory usage of both methods. After updating the code, the memory usage for calculating the gradient with snapshotting is more than twice that of the older code version. This reduced memory usage (I guess) because I was passing 'usnaps' with the operator (which is not good practice). I am wondering, is it possible to improve the code more to reduce the memory usage?

examples/seismic/acoustic/operators.py
malfarhan7: I've created a function to construct usnaps.

examples/seismic/acoustic/operators.py
malfarhan7: Removed.

examples/seismic/acoustic/operators.py
malfarhan7: Fixed. The operator build only returns op now.

examples/seismic/acoustic/operators.py
malfarhan7: Fixed.

examples/seismic/acoustic/operators.py
malfarhan7: Fixed. No cases are used.

examples/seismic/acoustic/wavesolver.py
malfarhan7: You are right. I did not catch it.

examples/seismic/acoustic/wavesolver.py
malfarhan7: Removed.

examples/seismic/acoustic/wavesolver.py
malfarhan7: Fixed. usnaps is created now.

examples/seismic/acoustic/wavesolver.py
malfarhan7: Fixed

malfarhan7:
Hi Mathias, thank you for your feedback. I have reviewed and cleaned the code to the best of my understanding. I have included a notebook to compare computing the FWI gradient with and without snapshotting and two scripts to calculate the memory usage of both methods. After updating the code, the memory usage for calculating the gradient with snapshotting is more than twice that of the older code version. This reduced memory usage (I guess) because I was passing 'usnaps' with the operator (which is not good practice). I am wondering, is it possible to improve the code more to reduce the memory usage?

examples/seismic/acoustic/operators.py
```diff
@@ -206,30 +225,38 @@ def GradientOperator(model, geometry, space_order=4, save=True,
...
+                         save=geometry.nt if save else None,
+                         time_order=2, space_order=space_order) 
+    v = TimeFunction(name='v', grid=model.grid, save=None, 
                      time_order=2, space_order=space_order)
     rec = geometry.rec
 
     s = model.grid.stepping_dim.spacing
     eqn = iso_stencil(v, model, kernel, forward=False)
-
+
```
FabioLuporini, line 247: revert change, pep8 violation

examples/seismic/acoustic/operators.py
```diff
@@ -206,30 +225,38 @@ def GradientOperator(model, geometry, space_order=4, save=True,
...
     elif kernel == 'OT4':
-        gradient_update = Inc(grad, - u * v.dt2 - s**2 / 12.0 * u.biharmonic(m**(-2)) * v)
-    # Add expression for receiver injection
+        gradient_update = Inc(grad, - u * v.dt2
+                                - s**2 / 12.0 * u.biharmonic(m**(-2)) * v)
+        
+     # Add expression for receiver injection
     receivers = rec.inject(field=v.backward, expr=rec * s**2 / m)
-
+
```
FabioLuporini, line 256: same as above

examples/seismic/acoustic/operators.py
```diff
@@ -206,30 +225,38 @@ def GradientOperator(model, geometry, space_order=4, save=True,
...
+        Downsampling factor to save snapshots of the wavefield.
     """
     m = model.m
 
     # Gradient symbol and wavefield symbols
     grad = Function(name='grad', grid=model.grid)
-    u = TimeFunction(name='u', grid=model.grid, save=geometry.nt if save
-                     else None, time_order=2, space_order=space_order)
-    v = TimeFunction(name='v', grid=model.grid, save=None,
+    if factor: # Apply the imaging condition at the snapshots of the full wavefield
```
FabioLuporini, line 235: Leave a blank line between the `grad = ` and this `if factor`

FabioLuporini: Move the comment inside the body of the `if`

examples/seismic/acoustic/operators.py
```diff
@@ -206,30 +225,38 @@ def GradientOperator(model, geometry, space_order=4, save=True,
...
+        u = create_snapshot_time_function(model, 'u', geometry, space_order, factor)
+    else:# Apply the imaging condition at every time step of the full wavefield
```
FabioLuporini, line 237: Move the comment inside the body of the `else`

examples/seismic/acoustic/wavesolver.py
```diff
@@ -90,30 +91,38 @@ def forward(self, src=None, rec=None, u=None, model=None, save=None, **kwargs):
             The time-constant velocity.
         save : bool, optional
             Whether or not to save the entire (unrolled) wavefield.
+        factor : int, optional
+        Downsampling factor to save snapshots of the wavefield.
```
FabioLuporini, line 95: Indent

examples/seismic/acoustic/wavesolver.py
```diff
@@ -90,30 +91,38 @@ def forward(self, src=None, rec=None, u=None, model=None, save=None, **kwargs):
...
 
         Returns
         -------
         Receiver, wavefield and performance summary
         """
         # Source term is read-only, so re-use the default
         src = src or self.geometry.src
-        # Create a new receiver object to store the result
+        # Create a new receiver object to store the result
```
FabioLuporini, line 103: drop trailing space

examples/seismic/acoustic/wavesolver.py
```diff
@@ -90,30 +91,38 @@ def forward(self, src=None, rec=None, u=None, model=None, save=None, **kwargs):
...
 
         # Create the forward wavefield if not provided
         u = u or TimeFunction(name='u', grid=self.model.grid,
-                              save=self.geometry.nt if save else None,
-                              time_order=2, space_order=self.space_order)
-
+                                save=self.geometry.nt if save and factor is None else None,
+                                time_order=2, space_order=self.space_order)
+        if factor: # Create snapshots of the forward wavefield
+            usnaps = create_snapshot_time_function(self.model, 'usnaps', self.geometry, self.space_order, factor)
```
FabioLuporini, line 111: variable definitions should not be conditional-dependent

examples/seismic/tutorials/Gradient_full_marmousi.py
```diff
@@ -0,0 +1,163 @@
+r'''
```
FabioLuporini, line 1: nitpicking: filenames start with lower letters

more important: there's way too much duplicated code between this file and `gradient_snap_marmousi`. Can it not be made a single, self-contained example, with argparse used to allow the user to select among different modes

examples/seismic/tutorials/Gradient_full_marmousi.py
```diff
@@ -0,0 +1,163 @@
...
+using the full time axis.
+
+Memory-profiler run as: mprof run Gradient_full_marmousi.py; mprof plot
+'''
+
+import numpy as np
+import matplotlib.pyplot as plt
+from scipy.ndimage import gaussian_filter
+import scipy
+from memory_profiler import memory_usage
```
FabioLuporini, line 12: imports from stdlib at the very top

then blank line

then imports from third parties (eg scipy)

then blank line

then examples imports
then devito imports

examples/seismic/acoustic/operators.py
```diff
@@ -106,9 +106,17 @@ def iso_stencil(field, model, kernel, **kwargs):
         eqns.append(freesurface(model, Eq(unext, eq_time)))
     return eqns
 
+def create_snapshot_time_function(model, name, geometry, space_order, factor, save=True):
+    nsnaps = (geometry.nt + factor - 1) // factor
+    time_subsampled = ConditionalDimension('t_sub', 
+                                            parent=model.grid.time_dim, factor=factor)
+    u_ = TimeFunction(name=name, grid=model.grid,
```
FabioLuporini, line 113: "usnaps" for homogeneity

examples/seismic/acoustic/operators.py
```diff
@@ -126,13 +134,16 @@ def ForwardOperator(model, geometry, space_order=4,
...
         Type of discretization, 'OT2' or 'OT4'.
+    factor : int, optional
+        Downsampling factor to save snapshots of the wavefield.
     """
     m = model.m
 
     # Create symbols for forward wavefield, source and receivers
     u = TimeFunction(name='u', grid=model.grid,
-                     save=geometry.nt if save else None,
+                     save=geometry.nt if save and factor is None else None,
```
FabioLuporini, line 144: I'm not a big fan of this composite conditional involving both `save` and `factor`, which is also repeated across other modules

examples/seismic/acoustic/operators.py
```diff
@@ -206,30 +225,38 @@ def GradientOperator(model, geometry, space_order=4, save=True,
...
+                                - s**2 / 12.0 * u.biharmonic(m**(-2)) * v)
+        
+     # Add expression for receiver injection
     receivers = rec.inject(field=v.backward, expr=rec * s**2 / m)
-
+    
     # Substitute spacing terms to reduce flops
     return Operator(eqn + receivers + [gradient_update], subs=model.spacing_map,
-                    name='Gradient', **kwargs)
+                  name='Gradient', **kwargs)
```
FabioLuporini, line 259: re-indent

examples/seismic/acoustic/wavesolver.py
```diff
@@ -90,30 +91,38 @@ def forward(self, src=None, rec=None, u=None, model=None, save=None, **kwargs):
...
         model = model or self.model
         # Pick vp from model unless explicitly provided
         kwargs.update(model.physical_params(**kwargs))
 
         # Execute operator and return wavefield and receiver data
-        summary = self.op_fwd(save).apply(src=src, rec=rec, u=u,
-                                          dt=kwargs.pop('dt', self.dt), **kwargs)
-
-        return rec, u, summary
+        if factor: # Return snapshots of the forward wavefield
```
FabioLuporini, line 118: since factor is passed down to `op_fwd`, I don't think we need the extra `if factor : .... else: ...` here, somehow it should be avoided and/or it's avoidable

FabioLuporini (changes requested):

examples/seismic/acoustic/operators.py
malfarhan7: Fixed

examples/seismic/acoustic/operators.py
malfarhan7: Fixed

examples/seismic/acoustic/operators.py
malfarhan7: Fixed

examples/seismic/acoustic/operators.py
malfarhan7: Fixed

examples/seismic/acoustic/wavesolver.py
malfarhan7: Fixed

examples/seismic/acoustic/wavesolver.py
malfarhan7: Fixed

examples/seismic/acoustic/wavesolver.py
malfarhan7: Now `usnaps` is defined unconditionally.

examples/seismic/tutorials/Gradient_full_marmousi.py
malfarhan7: Removed `Gradient_full_marmousi.py` and `Gradient_snap_marmousi.py` and added one script `fwi_gradient_marmousi.py` with argparse to choose between full gradient or snapshotting.

examples/seismic/tutorials/Gradient_full_marmousi.py
malfarhan7: Fixed

examples/seismic/acoustic/operators.py
malfarhan7: Fixed

examples/seismic/acoustic/operators.py
malfarhan7: Moved the conditional statement outside of the `u` definition but I do not know if there is a better way to avoid the composite conditional statement.

examples/seismic/acoustic/wavesolver.py
malfarhan7: Is there a better way to do this, as the code did not run correctly without the condition? I made the return statement conditional so as not to break people's code, so I kept the number of returned objects at three.

malfarhan7:
Hi Fabio, thank you for your feedback. I have reviewed and cleaned the code to the best of my understanding.

examples/seismic/acoustic/operators.py
```diff
@@ -126,13 +180,17 @@ def ForwardOperator(model, geometry, space_order=4,
         Defaults to False.
     kernel : str, optional
         Type of discretization, 'OT2' or 'OT4'.
+    factor : int, optional
+        Downsampling factor to save snapshots of the wavefield.
     """
     m = model.m
 
     # Create symbols for forward wavefield, source and receivers
+    save_value = geometry.nt if save and factor is None else None
```
FabioLuporini, line 189: To make the code more homogeneous and easier to read, what we could do, imho, is the following:

* drop `save` from this TimeFunction (which is just legacy behaviour...)
* systematically create `usnaps` and the corresponding equation
* with an `inf` (aka `sys.maxint` ?) `factor`, no snapshots will be saved at runtime
* Tweak [here](https://github.com/devitocodes/devito/blob/master/devito/types/dense.py#L127) adding a line `if self.size == 0: return`, so that Devito avoids allocating memory entirely if pointless

In my opinion, this will dramatically clean up the code here. 

At the moment, the proliferation of `if factor` logic is still affecting maintainability too much

@mloubout is on vacation ATM, but would be useful to hear from his thoughts about this matter

examples/seismic/acoustic/operators.py
```diff
@@ -107,8 +109,60 @@ def iso_stencil(field, model, kernel, **kwargs):
     return eqns
 
 
+def create_snapshot_time_function(model, name, geometry, space_order, factor=None):
```
mloubout, line 112: I would lift that into `seismic/utils.py` or some new file so it can be easily reused from other pdes. It can probably be done directly from the wavefield too i,e

```
def create_snapshot_time_function(u, nsnap):

    name = f"{u.name}_save"
    grid = u.grid
    
    ....
    
 ```

Makes it also decoupled from those geometry/model. THis would also allow to directly return the equation `Eq(u,usave)` as well

File: examples/seismic/tutorials/fwi_gradient_marmousi.py [hunk 26]
```diff
@@ -0,0 +1,118 @@
...
+
+
+# Define the water layer mask
+def mask(model, value):
+    mask = model > value
+    return mask.astype(int)
+
+
+# Compute residual
+def compute_residual(residual, dobs, dsyn):
```
mloubout, line 27: this is a copy paste for the fwi example, import the function from there

examples/seismic/tutorials/fwi_gradient_marmousi.py
```diff
@@ -0,0 +1,118 @@
+import argparse
```
mloubout, line 1: This looks like it has a lot of common with `examples/seismic/inversion` it would be better to avoid duplications and update that example instead.

examples/seismic/tutorials/fwi_gradient_marmousi.py
```diff
@@ -0,0 +1,118 @@
...
+        solver.gradient(rec=residual, u=u0, vp=vp_in, grad=grad, factor=factor)
+
+    return objective, grad
+
+
+def main(mode, factor=None):
+    shape = (601, 221)
+    spacing = (15.0, 15.0)
+    origin = (0.0, 0.0)
+    vel_path = '../../../devito/data/Marm.bin'
```
mloubout, line 72: This needs to be made available for CI somehow

examples/seismic/tutorials/fwi_gradient_marmousi.py
```diff
@@ -0,0 +1,118 @@
...
+
+    source_locations = np.column_stack((
+        np.linspace(0.0, model.domain_size[0], nshots),
+        np.zeros(nshots)
+    ))
+
+    ff, grad = fwi_gradient(mode, model, solver, geometry,
+                            source_locations, model0.vp, factor)
+    mem_usage = memory_usage()[0]
+    print(f"Memory usage at the end of gradient ({mode} mode): {mem_usage:.2f} MiB")
```
mloubout, line 102: Not sure what the goal of this check is here. the gradient is computed in as separate function so all the memory is freed. Showing memory usage difference would need to do this between forward and gradient for a single shot

examples/seismic/tutorials/fwi_gradient_marmousi.py
```diff
@@ -0,0 +1,118 @@
...
+    return residual
+
+
+# FWI Gradient Kernel
+def fwi_gradient(mode, model, solver, geometry, source_locations, vp_in, factor=None):
+    grad = Function(name="grad", grid=model.grid)
+    residual = Receiver(name='residual', grid=model.grid,
+                        time_range=geometry.time_axis,
+                        coordinates=geometry.rec_positions)
+    d_obs = Receiver(name='d_obs', grid=model.grid,
```
mloubout, line 47: `geometry.new_rec(name='d_obs')` same for others

examples/seismic/tutorials/fwi_gradient_marmousi.py
```diff
@@ -0,0 +1,118 @@
...
+                     time_range=geometry.time_axis,
+                     coordinates=geometry.rec_positions)
+    d_syn = Receiver(name='d_syn', grid=model.grid,
+                     time_range=geometry.time_axis,
+                     coordinates=geometry.rec_positions)
+    objective = 0.0
+    for i in range(nshots):
+        geometry.src_positions[0, :] = source_locations[i, :]
```
mloubout, line 55: Modifying geometry isn't good here. Instead get a source `src=geometry.src` and on;y modify the `src` coordinates. (I know this is just taken from the FWI tutorial but it's an old tutorial that needs updates)

examples/seismic/tutorials/fwi_gradient_marmousi.py
```diff
@@ -0,0 +1,118 @@
...
+        solver.forward(vp=model.vp, rec=d_obs)
+        save_value = True if mode == "full" else False
```
mloubout, line 57: this should be just a bool directly at argparse

examples/seismic/acoustic/operators.py
malfarhan7: Hi Fabio, thank you for your feedback.
If I drop `save` from u = TimeFunction(name='u', ...), I won't have access to the entire time axis to sample `u`, right? Can you please point me to some recent code which does not use this legacy behavior?
Thanks!

PR #2476, opened 2024-10-31
compiler: Improve CIRE's cost model

FabioLuporini:


devito/passes/clusters/aliases.py
```diff
@@ -958,15 +958,58 @@ def pick_best(variants):
...
+            grid = ii.function.grid
+            if grid is None:
+                continue
 
-        ntemps = len(i.schedule)
-        nfunctions0 = len({i.function for i in indexeds0})
-        nfunctions1 = len({i.function for i in indexeds1})
+            ntemps = 0
+            for sa in i.schedule:
+                if len(sa.writeto) < grid.dim:
```
mloubout, line 992: `sa.writeto.itdims`?  that comparison seems odd

tests/test_dse.py
```diff
@@ -2153,7 +2217,7 @@ def test_multiple_rotating_dims(self):
         # to jit-compile `op1`. However, we also check numerical correctness
         op1.apply(time_m=0, time_M=nt-2, dt=dt, u=u1, vx=vx1, vy=vy1)
 
-        assert np.allclose(u.data, u1.data, rtol=1e-5)
+        assert np.allclose(u.data, u1.data, rtol=1e-3)
```
mloubout, line 2220: What this much change?

devito/passes/clusters/aliases.py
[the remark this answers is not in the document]
FabioLuporini: what we had before was good already, here I've been surgical...

tests/test_dse.py
FabioLuporini: highly unstable equations and input data, and now more aggressive refactorings

devito/passes/clusters/aliases.py
FabioLuporini: it's ok, grid.dim is a number (should have been called grid.ndim, legacy...) and writeto is an IterationSpace so its length is the number of Dimensions in it

Issue #2448, opened 2024-09-05
compiler: Redundant haloupdate


georgebisbas:
Elastic-like cross-loop dependencies generate a redundant haloupdate.

Problem in Pseudocode:
```
for time
  haloupd vx[t0]
  write to vx[t1] - read from vx[t0]
  haloupd vx[t1]
  read from vx[t1]
  read from vx[t0] 
```
it could be
```
<(haloup vx[t0])>
for time
  ---DROP haloupd vx[t0]---since previously written t1 is now t0 and is already updated
  write to vx[t1] - read from vx[t0]
  haloupd vx[t1]
  read from vx[t1]
  read from vx[t0] 
```

Python script to reproduce:
(reduced, starting from https://github.com/devitocodes/devito/blob/master/examples/seismic/tutorials/06_elastic_varying_parameters.ipynb)

```sh
DEVITO_LOGGING=DEBUG DEVITO_MPI=1 mpirun -n 1 python3 tests/elastic_mfe_1d.py 
DEVITO_LOGGING=DEBUG DEVITO_MPI=1 ../../tmpi/tmpi 1 python3 tests/elastic_mfe_1d.py 
```

`elastic_mfe_1d.py `:
```python
import numpy as np

from devito import (SpaceDimension, Grid, TimeFunction, Eq, Operator,
                    solve, Constant)
from examples.seismic.source import TimeAxis, Receiver

# Space related
extent = (1500., )
shape = (201, )
x = SpaceDimension(name='x', spacing=Constant(name='h_x', value=extent[0]/(shape[0]-1)))
grid = Grid(extent=extent, shape=shape, dimensions=(x, ))

# Time related
t0, tn = 0., 30.
dt = (10. / np.sqrt(2.)) / 6.
time_range = TimeAxis(start=t0, stop=tn, step=dt)

# Velocity and pressure fields
so = 2
to = 1
v = TimeFunction(name='v', grid=grid, space_order=so, time_order=to)
tau = TimeFunction(name='tau', grid=grid, space_order=so, time_order=to)

# The receiver
nrec = 1
rec = Receiver(name="rec", grid=grid, npoint=nrec, time_range=time_range)
rec.coordinates.data[:, 0] = np.linspace(0., extent[0], num=nrec)
rec_term = rec.interpolate(expr=v)

# First order elastic-like dependencies equations
pde_v = v.dt - (tau.dx)
pde_tau = (tau.dt - ((v.forward).dx))

[...]

PR #2477, opened 2024-11-04
compiler: Rework HaloSpot optimization

georgebisbas:
Fix for issue https://github.com/devitocodes/devito/issues/2448, supplemented with some code edits


devito/passes/iet/mpi.py
```diff
@@ -59,120 +59,111 @@ def _hoist_halospots(iet):
...
-    for hs in FindNodes(HaloSpot).visit(iet):
-        if hs.body.is_HaloSpot:
-            halo_scheme = HaloScheme.union([hs.halo_scheme, hs.body.halo_scheme])
-            mapper[hs] = hs._rebuild(halo_scheme=halo_scheme, body=hs.body.body)
     iet = Transformer(mapper, nested=True).visit(iet)
 
     return iet
 
 
-def _merge_halospots(iet):
```
georgebisbas, line 135: This is intact, nothing ahs changed

tests/test_mpi.py
```diff
@@ -2727,6 +2801,85 @@ def test_adjoint_F_no_omp(self, mode):
         self.run_adjoint_F(3)
 
 
+class TestElastic:
```
georgebisbas, line 2804: This is stronger than the mfe, as it also looks for correct grouping of functions in a halospot.
The MFE does not need to merge functions

georgebisbas: No similar test exists in the codebase

devito/mpi/halo_scheme.py
```diff
@@ -28,7 +28,39 @@ class HaloLabel(Tag):
...
+                self.halos == other.halos and
+                self.dims == other.dims)
+
+    def __hash__(self):
+        return hash((frozenset(self.loc_indices.items()),
+                     frozenset(self.loc_dirs.items()),
+                     frozenset(self.halos),
+                     frozenset(self.dims)))
+
+    def rebuild(self, **kwargs):
```
EdCaunt, line 57: Subclass `Reconstructable` and add `rargs`?

devito/mpi/halo_scheme.py
```diff
@@ -95,8 +127,16 @@ def __init__(self, exprs, ispace):
         self._honored = frozendict(self._honored)
 
     def __repr__(self):
```
EdCaunt, line 129: Could this be attached to a mixin class for both `HaloSpot` and `HaloSchemeEntry`?

tests/test_mpi.py
```diff
@@ -1005,6 +1011,75 @@ def test_avoid_haloupdate_if_distr_but_sequential(self, mode):
         calls = FindNodes(Call).visit(op)
         assert len(calls) == 0
 
+    @pytest.mark.parallel(mode=1)
+    def test_issue_2448(self, mode):
+        extent = (10., )
```
EdCaunt, line 1016: Nitpick: remove whitespace

tests/test_mpi.py
```diff
@@ -1005,6 +1011,75 @@ def test_avoid_haloupdate_if_distr_but_sequential(self, mode):
...
+        shape = (2, )
+        so = 2
+        to = 1
+
+        x = SpaceDimension(name='x', spacing=Constant(name='h_x',
```
EdCaunt, line 1021: Do you need this `SpaceDimension`?

tests/test_mpi.py
```diff
@@ -1005,6 +1011,75 @@ def test_avoid_haloupdate_if_distr_but_sequential(self, mode):
...
+                                                      value=extent[0]/(shape[0]-1)))
+        grid = Grid(extent=extent, shape=shape, dimensions=(x, ))
```
EdCaunt, line 1023: Nitpick: remove whitespace from tuple

tests/test_mpi.py
```diff
@@ -1005,6 +1011,75 @@ def test_avoid_haloupdate_if_distr_but_sequential(self, mode):
...
+        # First order elastic-like dependencies equations
+        pde_v = v.dt - (tau.dx)
+        pde_tau = (tau.dt - ((v.forward).dx))
+        u_v = Eq(v.forward, solve(pde_v, v.forward))
+
+        u_tau = Eq(tau.forward, solve(pde_tau, tau.forward))
+
+        # Test two variants of receiver interpolation
+        nrec = 1
+        rec = Receiver(name="rec", grid=grid, npoint=nrec, time_range=time_range)
```
EdCaunt, line 1043: Use `SparseTimeFunction`?

devito/ir/iet/nodes.py
```diff
@@ -1457,8 +1457,17 @@ def __init__(self, body, halo_scheme):
         self._halo_scheme = halo_scheme
 
     def __repr__(self):
-        functions = "(%s)" % ",".join(i.name for i in self.functions)
-        return "<%s%s>" % (self.__class__.__name__, functions)
+        fstrings = []
+        for f in self.functions:
+            loc_indices = set().union(*[self.halo_scheme.fmapper[f].loc_indices.values()])
+            loc_indices = list(loc_indices)
+            loc_indices_str = str(list(loc_indices)) if loc_indices else ""
```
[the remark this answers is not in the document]
georgebisbas: sure!

devito/ir/iet/nodes.py
```diff
@@ -1457,8 +1457,16 @@ def __init__(self, body, halo_scheme):
         self._halo_scheme = halo_scheme
 
     def __repr__(self):
-        functions = "(%s)" % ",".join(i.name for i in self.functions)
-        return "<%s%s>" % (self.__class__.__name__, functions)
+        fstrings = []
+        for f in self.functions:
+            loc_indices = set().union(*[self.halo_scheme.fmapper[f].loc_indices.values()])
```
FabioLuporini, line 1462: since this is a set, the result of the `__repr__` will be non deterministic across different python executions, that is, you may see either [t0,t1] or [t1,t0]

devito/mpi/halo_scheme.py
```diff
@@ -28,7 +28,39 @@ class HaloLabel(Tag):
 STENCIL = HaloLabel('stencil')
 
 
-HaloSchemeEntry = namedtuple('HaloSchemeEntry', 'loc_indices loc_dirs halos dims')
+class HaloSchemeEntry:
```
FabioLuporini, line 31: as Ed wrote below, this should inherit from Reconstructable

devito/mpi/halo_scheme.py
```diff
@@ -28,7 +28,39 @@ class HaloLabel(Tag):
...
+
+    def __init__(self, loc_indices, loc_dirs, halos, dims):
```
FabioLuporini, line 33: can it not inherit, alternatively, from EnrichedTuple?

devito/mpi/halo_scheme.py
FabioLuporini: yes but where would you add it? 

I would just drop all these __repr__ it's a lot of code for what at the end of the day?

tests/test_mpi.py
```diff
@@ -2727,6 +2801,85 @@ def test_adjoint_F_no_omp(self, mode):
...
+        assert calls[2].arguments[0] is tau[0, 1]
+        assert calls[3].arguments[0] is tau[1, 1]
+
+        assert calls[4].arguments[0] is v[0]
+        assert calls[4].arguments[1] is v[1]
+
+
+class TestTTI_w_MPI:
+
+    @pytest.mark.parallel(mode=[(1)])
```
FabioLuporini, line 2869: mode=1

tests/test_mpi.py
```diff
@@ -2727,6 +2801,85 @@ def test_adjoint_F_no_omp(self, mode):
...
+
+        assert calls[1].arguments[0] is tau[0, 0]
+        assert calls[2].arguments[0] is tau[0, 1]
+        assert calls[3].arguments[0] is tau[1, 1]
+
+        assert calls[4].arguments[0] is v[0]
+        assert calls[4].arguments[1] is v[1]
+
+
+class TestTTI_w_MPI:
```
FabioLuporini, line 2867: no underscores in class names

FabioLuporini: probably don't need to add a new class either

tests/test_mpi.py
```diff
@@ -1005,6 +1011,75 @@ def test_avoid_haloupdate_if_distr_but_sequential(self, mode):
...
+        to = 1
+
+        x = SpaceDimension(name='x', spacing=Constant(name='h_x',
+                                                      value=extent[0]/(shape[0]-1)))
+        grid = Grid(extent=extent, shape=shape, dimensions=(x, ))
+
+        # Time related
+        t0, tn = 0., 30.
+        dt = (10. / np.sqrt(2.)) / 6.
+        time_range = TimeAxis(start=t0, stop=tn, step=dt)
```
FabioLuporini, line 1028: just like Ed wrote below, you don't need examples-level abstractions

tests/test_mpi.py
```diff
@@ -1005,6 +1011,75 @@ def test_avoid_haloupdate_if_distr_but_sequential(self, mode):
         calls = FindNodes(Call).visit(op)
         assert len(calls) == 0
 
+    @pytest.mark.parallel(mode=1)
+    def test_issue_2448(self, mode):
+        extent = (10., )
+        shape = (2, )
+        so = 2
+        to = 1
```
FabioLuporini, line 1019: useless probably

tests/test_mpi.py
```diff
@@ -1005,6 +1011,75 @@ def test_avoid_haloupdate_if_distr_but_sequential(self, mode):
         calls = FindNodes(Call).visit(op)
         assert len(calls) == 0
 
+    @pytest.mark.parallel(mode=1)
+    def test_issue_2448(self, mode):
+        extent = (10., )
+        shape = (2, )
```
FabioLuporini, line 1017: if you are not executing the operators, you don't need Grid.extent, so you can probably just inline this variable instead

devito/passes/iet/mpi.py
```diff
@@ -368,3 +359,37 @@ def mpiize(graph, **kwargs):
...
+# Utility functions to avoid code duplication
+
+def _create_cond_mapper(iet):
+    cond_mapper = MapHaloSpots().visit(iet)
+    return {hs: {i for i in v if i.is_Conditional and
+                 not isinstance(i.condition, GuardFactorEq)}
+            for hs, v in cond_mapper.items()}
+
+
+def merge_rules():
```
FabioLuporini, line 373: this shouldn't be a function; a tuple will do

devito/passes/iet/mpi.py
```diff
@@ -368,3 +359,37 @@ def mpiize(graph, **kwargs):
         make_halo_exchanges(graph, mpimode=mpimode, **kwargs)
 
     make_reductions(graph, mpimode=mpimode, **kwargs)
+
+
+# Utility functions to avoid code duplication
```
FabioLuporini, line 364:
```
# *** Utilities
```

for code homogeneity (nitpicking)

devito/passes/iet/mpi.py
```diff
@@ -59,120 +59,116 @@ def _hoist_halospots(iet):
...
-                    for dep in scopes[i].d_flow.project(f):
-                        if not any(r(dep, candidates, loc_dims) for r in rules):
-                            break
-                    else:
-                        hsmapper[hs] = hsmapper[hs].drop(f)
-                        imapper[i].append(hs.halo_scheme.project(f))
-                        break
+                # Compare hs0 to subsequent halo_spots, looking for optimization
+                # possibilities
+                _process_halo_to_halo(hs0, hs1, iters, scopes, hsmapper, imapper)
```
FabioLuporini, line 104: a function call here , but it smells like a disguise to split a very very long code block into two

devito/passes/iet/mpi.py
```diff
@@ -59,120 +59,116 @@ def _hoist_halospots(iet):
...
+                if cond_mapper.get(hs0) != cond_mapper.get(hs1):
+                    continue
 
-                    all_candidates = set().union(*candidates)
-                    reads = scopes[i].getreads(f)
-                    if any(set(a.ispace.dimensions) & all_candidates
-                           for a in reads):
-                        continue
+                # If there are overlapping time accesses, skip
+                if any(i in hs0.halo_scheme.loc_values
```
FabioLuporini, line 98: question: would it be useful to implement `hs0.halo_scheme.intersection(hs1.halo_scheme)` ? Might make this code neater, while being reusable elsewhere

devito/passes/iet/mpi.py
```diff
@@ -59,120 +59,116 @@ def _hoist_halospots(iet):
...
+            # If also merge-able we can start hoisting the latter
+            for dep in scopes[iter].d_flow.project(f):
+                if not any(r(dep, hs1, v.loc_indices) for r in merge_rules()):
+                    break
+            else:
+                hse = hs1.halo_scheme.fmapper[f]
+                raw_loc_indices = {}
+                # Entering here means we can lift, and we need to update
+                # the loc_indices with known values
+                # TODO: Can I get this in a more elegant way?
```
FabioLuporini, line 138: not only that, you should also check the iteration direction

FabioLuporini: which means, btw, you have to add a pass for a gradient-like operator (as long as time flows backwards, we're good)

FabioLuporini: not sure if it's useful or not, but there are various places in which we have to distinguish our actions depending on whether an iteration flows forward or backwards. For example:

https://github.com/devitocodes/devito/blob/master/devito/passes/clusters/asynchrony.py#L231
https://github.com/devitocodes/devito/blob/master/devito/mpi/halo_scheme.py#L533

and a few others (some of these are quite legacy unfortunately)

I wonder whether there's an utility function that we can write and use in more than one place, but I'm far from sure, as each of these *seem* very special cases... but it's worth looking into it

devito/passes/iet/mpi.py
```diff
@@ -59,120 +59,116 @@ def _hoist_halospots(iet):
...
 
     mapper = {}
-    for i, halo_spots in iter_mapper.items():
-        if i is None or len(halo_spots) <= 1:
+    for iter, halo_spots in MapNodes(Iteration, HaloSpot, 'immediate').visit(iet).items():
+        if iter is None or len(halo_spots) <= 1:
             continue
 
-        scope = Scope([e.expr for e in FindNodes(Expression).visit(i)])
+        scope = Scope([e.expr for e in FindNodes(Expression).visit(iter)])
```
FabioLuporini, line 171: don't use the word "iter" since it has a special meaning in python

devito/passes/iet/mpi.py
```diff
@@ -59,120 +59,116 @@ def _hoist_halospots(iet):
...
-    Merge HaloSpots on the same Iteration tree level where all data dependencies
-    would be honored.
-    """
+def _process_halo_to_halo(hs0, hs1, iters, scopes, hsmapper, imapper):
 
-    # Merge rules -- if the retval is True, then it means the input `dep` is not
-    # a stopper to halo merging
+    # Loop over the functions in the HaloSpots
+    for f, v in hs1.fmapper.items():
+        # If no time accesses, skip
```
FabioLuporini, line 120: "E.g., if no `time` accesses, skip"

you have to say it's an example...

devito/passes/iet/mpi.py
```diff
@@ -59,120 +59,116 @@ def _hoist_halospots(iet):
...
+        if not hs1.halo_scheme.fmapper[f].loc_indices:
+            continue
 
-    def rule0(dep, hs, loc_indices):
-        # E.g., `dep=W<f,[t1, x]> -> R<f,[t0, x-1]>` => True
-        return not any(d in hs.dimensions or dep.distance_mapper[d] is S.Infinity
-                       for d in dep.cause)
+        # If the function is not in both HaloSpots, skip
+        if (*hs0.functions, *hs1.functions).count(f) < 2:
```
FabioLuporini, line 125: `.intersection(...)`, again as mentioned above?

devito/passes/iet/mpi.py
```diff
@@ -59,120 +59,116 @@ def _hoist_halospots(iet):
...
-                all(not any(dep.read.touched_halo(d.root)) for d in dep.cause))
+        for iter in iters:
+            # If also merge-able we can start hoisting the latter
+            for dep in scopes[iter].d_flow.project(f):
+                if not any(r(dep, hs1, v.loc_indices) for r in merge_rules()):
+                    break
+            else:
+                hse = hs1.halo_scheme.fmapper[f]
+                raw_loc_indices = {}
+                # Entering here means we can lift, and we need to update
```
FabioLuporini, line 136: move comment two lines above

devito/passes/iet/mpi.py
FabioLuporini: in retrospect, now that I've finished reviewing, we should probably:

* simplify (as much as possible!) what's currently inside `_process_halo_to_halo`
* inline it here directly

devito/passes/iet/mpi.py
```diff
@@ -59,120 +59,116 @@ def _hoist_halospots(iet):
...
+      haloupd v[t0]                  R v[t0]
+      R v[t0]
 
-    rules = [rule0, rule1]
+    """
 
     # Precompute scopes to save time
     scopes = {i: Scope([e.expr for e in v]) for i, v in MapNodes().visit(iet).items()}
 
+    cond_mapper = _create_cond_mapper(iet)
```
FabioLuporini, line 77: typically we use the word "make", so `_make_cond_mapper(iet)`, but obviously I'm nitpicking here

devito/passes/iet/mpi.py
```diff
@@ -59,120 +59,116 @@ def _hoist_halospots(iet):
...
 
-                for n, i in enumerate(iters):
-                    if i not in scopes:
-                        continue
+            # Nothing to do if the HaloSpot is void
+            if hs0.halo_scheme.is_void:
+                continue
 
-                    candidates = [i.dim._defines for i in iters[n:]]
+            for hs1 in halo_spots[i+1:]:
```
FabioLuporini, line 91: maybe you could avoid an indentation level but rather considering all pairs

```
from itertools import combinations

...

for hs0, hs1 in combinations(halo_schemes, 2):
    ...
```

FabioLuporini: btw, are we sure that the current code works if there are *2* hoisting opportunities, instead of *1* ?

could we add a test?

devito/passes/iet/mpi.py
georgebisbas: wdym by 'say it's an example' ?

devito/mpi/halo_scheme.py
georgebisbas: When you print the IET, you see more useful details, it makes dev/debug life easier!

devito/ir/iet/nodes.py
```diff
@@ -1457,8 +1457,17 @@ def __init__(self, body, halo_scheme):
...
+            loc_indices = list(loc_indices)
+            loc_indices_str = str(list(loc_indices)) if loc_indices else ""
+
+            fstrings.append(f"{f.name}{loc_indices_str}")
```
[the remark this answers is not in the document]
georgebisbas: Sure

tests/test_mpi.py
georgebisbas: dropped

tests/test_mpi.py
georgebisbas: done

devito/ir/iet/nodes.py
georgebisbas: Used OrderedSet

tests/test_mpi.py
georgebisbas: done

devito/passes/iet/mpi.py
```diff
@@ -59,120 +59,116 @@ def _hoist_halospots(iet):
...
-                not any(i & dep.cause for i in candidates) and
-                not any(i & loc_dims for i in candidates))
 
-    def rule1(dep, candidates, loc_dims):
-        # A reduction isn't a stopper to hoisting
-        return dep.write is not None and dep.write.is_reduction
+    Example:
+                                   haloupd v[t0]
+    for time                       for time
+      W v[t1]- R v[t0]             W v[t1]- R v[t0]
```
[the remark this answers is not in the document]
georgebisbas: done

devito/passes/iet/mpi.py
```diff
@@ -59,120 +59,116 @@ def _hoist_halospots(iet):
...
+        if (*hs0.functions, *hs1.functions).count(f) < 2:
+            continue
 
-    def rule1(dep, hs, loc_indices):
-        # TODO This is apparently never hit, but feeling uncomfortable to remove it
-        return (dep.is_regular and
-                dep.read is not None and
-                all(not any(dep.read.touched_halo(d.root)) for d in dep.cause))
+        for iter in iters:
+            # If also merge-able we can start hoisting the latter
```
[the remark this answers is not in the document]
georgebisbas: reworked terminology

devito/passes/iet/mpi.py
georgebisbas: fixed everywhere

devito/passes/iet/mpi.py
georgebisbas: no needed anymore

devito/passes/iet/mpi.py
georgebisbas: done

devito/passes/iet/mpi.py
georgebisbas: I am working on this now

devito/passes/iet/mpi.py
georgebisbas: I though the rules there would ensure that we do not hoist some wrong t-slice.
Shouldn;t we just assert:
assert hs0.loc_dirs == hs1.loc_dirs
in order to do any of those transforms?

devito/passes/iet/mpi.py
```diff
@@ -48,157 +49,157 @@ def _drop_halospots(iet):
...
     mapper = {i: HaloSpot(i._rebuild(), HaloScheme.union(hss))
               for i, hss in imapper.items()}
+
     mapper.update({i: i.body if hs.is_void else i._rebuild(halo_scheme=hs)
                    for i, hs in hsmapper.items()})
 
-    # Transform the IET hoisting/dropping HaloSpots as according to the analysis
     iet = Transformer(mapper, nested=True).visit(iet)
-
-    # Clean up: de-nest HaloSpots if necessary
```
georgebisbas, line 124: there is no need for that anymore

devito/passes/iet/mpi.py
georgebisbas: done

devito/passes/iet/mpi.py
```diff
@@ -368,3 +367,42 @@ def mpiize(graph, **kwargs):
...
+    # If there are Conditionals involved, both `hs0` and `hs1` must be
+    # within the same Conditional, otherwise we would break the control
+    """
+    cond0 = cond_mapper.get(hs0)
+    cond1 = cond_mapper.get(hs1)
+
+    return cond0 != cond1
+
+
+def motion_rules():
```
EdCaunt, line 392: Perhaps define a `motion_rules = _motion_rules()`?

devito/mpi/halo_scheme.py
```diff
@@ -28,7 +28,34 @@ class HaloLabel(Tag):
 STENCIL = HaloLabel('stencil')
 
 
-HaloSchemeEntry = namedtuple('HaloSchemeEntry', 'loc_indices loc_dirs halos dims')
+class HaloSchemeEntry(Reconstructable):
+
+    __rkwargs__ = ('loc_indices', 'loc_dirs', 'halos', 'dims')
```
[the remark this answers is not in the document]
FabioLuporini: yes

devito/mpi/halo_scheme.py
FabioLuporini: > that look redundant

with EnrichedTuple, once you properly override `__rargs__` and `__rkwargs__`, you probably don't need to override `__init__` and `__eq__`

devito/mpi/halo_scheme.py
FabioLuporini: anyway, Ed has a point, this is redundant code which should somehow be factorized (a private method of HaloScheme, also called by `HaloSpot.__repr__`? not sure)

devito/passes/iet/mpi.py
FabioLuporini: this should not be a function

devito/passes/iet/mpi.py
```diff
@@ -368,3 +367,42 @@ def mpiize(graph, **kwargs):
...
+def _make_cond_mapper(iet):
+    cond_mapper = MapHaloSpots().visit(iet)
+    return {hs: {i for i in v if i.is_Conditional and
+                 not isinstance(i.condition, GuardFactorEq)}
+            for hs, v in cond_mapper.items()}
+
+
+def ensure_control_flow(hs0, hs1, cond_mapper):
+    """
+    # If there are Conditionals involved, both `hs0` and `hs1` must be
```
FabioLuporini, line 383: Drop "#"

devito/passes/iet/mpi.py
```diff
@@ -368,3 +367,42 @@ def mpiize(graph, **kwargs):
...
+    # within the same Conditional, otherwise we would break the control
```
FabioLuporini, line 384: control flow

devito/passes/iet/mpi.py
```diff
@@ -48,157 +49,155 @@ def _drop_halospots(iet):
...
     # Analysis
-    cond_mapper = MapHaloSpots().visit(iet)
-    cond_mapper = {hs: {i for i in v if i.is_Conditional and
-                        not isinstance(i.condition, GuardFactorEq)}
-                   for hs, v in cond_mapper.items()}
+    cond_mapper = _make_cond_mapper(iet)
+
+    mapper = {}
 
     iter_mapper = MapNodes(Iteration, HaloSpot, 'immediate').visit(iet)
```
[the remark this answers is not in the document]
FabioLuporini: now that I think about it, you can probably also move the `cond_mapper` inside such a new utility function, and do the conditional filtering in there directly, rather than repeating it in both passes (it's basically the same stuff repeated at the moment)

devito/passes/iet/mpi.py
FabioLuporini: And I'd call this simply "rules"

```
def _rule0(...):
     """...."""
     <impl>

def _rule1(...):
     """...."""
     <impl>

rules = (_rule0, _rule1)
```

tests/test_mpi.py
```diff
@@ -2727,6 +2856,79 @@ def test_adjoint_F_no_omp(self, mode):
...
 
 
+class TestElastic:
+
+    @pytest.mark.parallel(mode=[(1, 'diag')])
+    def test_elastic_structure(self, mode):
+
+        so = 4
+        model = demo_model(preset='layers-elastic', nlayers=1,
+                           shape=(301, 301), spacing=(10., 10.),
```
mloubout, line 2866: This is a codegen test there is no need for anything this complicated.

```
grid = Grid((3,3,3))
lam = Function
mu = Function
ro = Function
nrec =1
```

tests/test_mpi.py
```diff
@@ -2727,6 +2856,79 @@ def test_adjoint_F_no_omp(self, mode):
...
+        assert calls[4].arguments[0] is v[0]
+        assert calls[4].arguments[1] is v[1]
+
+
+class TestTTIwMPI:
+
+    @pytest.mark.parallel(mode=1)
+    def test_halo_structure(self, mode):
+
+        mytest = TestTTI()
```
mloubout, line 2921: remove, oneline `TestTTI().tti_operator`

devito/passes/iet/mpi.py
```diff
@@ -48,157 +49,155 @@ def _drop_halospots(iet):
...
 
-    mapper = {}
-    for i, halo_spots in iter_mapper.items():
-        if i is None or len(halo_spots) <= 1:
-            continue
+    # Drop pairs that have keys that are None
+    iter_mapper = {k: v for k, v in iter_mapper.items() if k is not None}
```
mloubout, line 177: This whole section looks like a plain copy paste of 89-119 especially the loop next, `iter_maper` filtering and conditions, can't this be. merged or lifted  in some common piece?

devito/mpi/halo_scheme.py
georgebisbas: I fail to do so, so far,, the already existing code of DimensionTuple does not seem to help me....

devito/passes/iet/mpi.py
georgebisbas: Done

devito/passes/iet/mpi.py
georgebisbas: done

devito/passes/iet/mpi.py
georgebisbas: done

devito/passes/iet/mpi.py
```diff
@@ -48,157 +49,155 @@ def _drop_halospots(iet):
...
+                else:
+                    # hs1 is lifted out of `it`, and we need to get
+                    # the new indexing for the HaloSpot
+                    hse = hs1.halo_scheme.fmapper[f]
+                    raw_loc_indices = {}
+
+                    for d in hse.loc_indices:
+                        md = hse.loc_indices[d]
+                        if md in it.uindices:
+                            new_min = md.symbolic_min.subs(it.dim,
```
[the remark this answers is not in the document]
georgebisbas: fixed+added

tests/test_mpi.py
georgebisbas: done

devito/passes/iet/mpi.py
```diff
@@ -368,3 +350,55 @@ def mpiize(graph, **kwargs):
         make_halo_exchanges(graph, mpimode=mpimode, **kwargs)
 
     make_reductions(graph, mpimode=mpimode, **kwargs)
+
+
+# *** Utilities
+
+def filter_iter_mapper(iet):
```
EdCaunt, line 357: Nitpick: if these aren't going to be exposed, do they need to be private? Either way, they should probably all be consistent with one another

devito/passes/iet/mpi.py
```diff
@@ -368,3 +350,55 @@ def mpiize(graph, **kwargs):
...
+                   for d in dep.cause)
+
+
+def _rule1(dep, hs, loc_indices):
+    # E.g., `dep=W<f,[t1, x+1]> -> R<f,[t1, xl+1]>` and `loc_indices={t: t0}` => True
+    return any(dep.distance_mapper[d] == 0 and dep.source[d] is not v
+               for d, v in loc_indices.items())
+
+
+rules = [_rule0, _rule1]
```
EdCaunt, line 404: Nitpick: Making this a tuple would be safer

tests/test_mpi.py
```diff
@@ -1005,6 +1008,182 @@ def test_avoid_haloupdate_if_distr_but_sequential(self, mode):
...
+        # First order elastic-like dependencies equations
+        pde_v = v.dt - (tau.dx)
+        pde_tau = tau.dt - ((v.forward).dx)
+        u_v = Eq(v.forward, solve(pde_v, v.forward))
+
+        u_tau = Eq(tau.forward, solve(pde_tau, tau.forward))
+
+        # Test two variants of receiver interpolation
+        nrec = 1
+        rec = SparseTimeFunction(name="rec", grid=grid, npoint=nrec, nt=tn)
```
EdCaunt, line 1034: Nitpick: You could remove some lines by having `npoint=1, nt=30` in here. You can do similar stuff throughout this test (for example `Grid(shape=(2,))`.

tests/test_mpi.py
```diff
@@ -1005,6 +1008,182 @@ def test_avoid_haloupdate_if_distr_but_sequential(self, mode):
...
+        rec2.coordinates.data[:, 0] = np.linspace(0., shape[0], num=nrec)
+
+        # The receiver 2
+        rec_term2 = rec2.interpolate(expr=v2)
+
+        # The receiver 3
+        rec_term3 = rec2.interpolate(expr=v2.forward)
+
+        # Test receiver interpolation 0, here we have a halo exchange hoisted
+        op2 = Operator([u_v] + [u_v2] + [u_tau] + [u_tau2] + rec_term0 + rec_term2)
```
EdCaunt, line 1100: Should this and `op3` be split off in their own tests, or is it necessary to have them in this test?

EdCaunt (commented):
Few minor comments. Looks very good to me

examples/seismic/viscoacoustic/operators.py
```diff
@@ -524,7 +524,7 @@ def ForwardOperator(model, geometry, space_order=4, kernel='sls', time_order=2,
 
     # Substitute spacing terms to reduce flops
     return Operator(eqn + src_term + rec_term, subs=model.spacing_map,
-                    name='Forward', **kwargs)
+                    name='ViscoAcForward', **kwargs)
```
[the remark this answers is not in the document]
FabioLuporini: spell it out as ViscoIsoAcousticForward ?

tests/test_mpi.py
```diff
@@ -1005,6 +1008,180 @@ def test_avoid_haloupdate_if_distr_but_sequential(self, mode):
...
+
+        assert len(FindNodes(HaloUpdateCall).visit(op0.body.body[1].body[1].body[0])) == 1
+        assert len(FindNodes(HaloUpdateCall).visit(op0.body.body[1].body[1].body[1])) == 2
+        assert calls[0].arguments[0] is v
+        assert calls[1].arguments[0] is tau
+        assert calls[2].arguments[0] is v
+
+        # Test receiver interpolation 1, here we should not have any halo exchange
+        # hoisted
+        op1 = Operator([u_v] + [u_tau] + rec_term1)
```
FabioLuporini, line 1059: this test is way too long. This thing starting here should be a separate test

tests/test_mpi.py
```diff
@@ -1005,6 +1008,180 @@ def test_avoid_haloupdate_if_distr_but_sequential(self, mode):
...
+        # The correct we want
+        assert len(calls) == 3
+
+        assert len(FindNodes(HaloUpdateCall).visit(op1.body.body[1].body[0])) == 0
+        assert len(FindNodes(HaloUpdateCall).visit(op1.body.body[1].body[1])) == 3
+        assert calls[0].arguments[0] is tau
+        assert calls[1].arguments[0] is v
+        assert calls[2].arguments[0] is v
+
+        # Further complicate/stree-test adding an artifical example
```
FabioLuporini, line 1073: this is yet another test

tests/test_mpi.py
FabioLuporini: ah, yes... I was writing the same above. This must be split over multiple tests

tests/test_mpi.py
```diff
@@ -1357,7 +1534,7 @@ def test_merge_haloupdate_if_diff_locindices_v1(self, mode):
         op = Operator(eqns)
 
         calls = FindNodes(Call).visit(op)
-        assert len(calls) == 1
+        assert len(calls) == 2
```
FabioLuporini, line 1537: you should update the test docstring
you should also ensure that one of these two halo_updates is outside the time loop

tests/test_mpi.py
```diff
@@ -2727,6 +2903,71 @@ def test_adjoint_F_no_omp(self, mode):
         self.run_adjoint_F(3)
 
 
+class TestElastic:
+
+    @pytest.mark.parallel(mode=[(1, 'diag')])
+    def test_elastic_structure(self, mode):
```
FabioLuporini, line 2909: at first glance, it looks like there's a lot of redundant checks that also appear in the huge test above... but I may be wrong

tests/test_mpi.py
```diff
@@ -2727,6 +2903,71 @@ def test_adjoint_F_no_omp(self, mode):
...
+        assert calls[0].arguments[0] is v[0]
+        assert calls[0].arguments[1] is v[1]
+        assert calls[1].arguments[0] is tau[0, 0]
+        assert calls[2].arguments[0] is tau[0, 1]
+        assert calls[3].arguments[0] is tau[1, 1]
+        assert calls[4].arguments[0] is v[0]
+        assert calls[4].arguments[1] is v[1]
+
+
+class TestTTIwMPI:
```
FabioLuporini, line 2956: you can drop "wMPI"

FabioLuporini (commented):
A few more comments in, but it's surely getting closer and closer to a mergeable state

devito/mpi/halo_scheme.py
georgebisbas: What I am trying to say is that in order to factor code out, someone has to factor out the fmapper/functions, but I had failed to do it. Now I think I did it, fix coming

devito/passes/iet/mpi.py
```diff
@@ -48,157 +49,133 @@ def _drop_halospots(iet):
...
+                    if not any(r(dep, hs1, v.loc_indices) for r in rules):
                         break
+                else:
+                    # hs1 can be hoisted out of `it`, but we need to infer valid
+                    # loc_indices
+                    hse = hs1.halo_scheme.fmapper[f]
+                    raw_loc_indices = {}
+
+                    for d in hse.loc_indices:
+                        md = hse.loc_indices[d]
```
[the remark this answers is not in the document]
georgebisbas: You are right!

devito/passes/iet/mpi.py
```diff
@@ -48,157 +49,133 @@ def _drop_halospots(iet):
...
+                        if md in it.uindices:
+                            md_sub = it.start
+                            raw_loc_indices[d] = md.symbolic_min.subs(it.dim, md_sub)
+                        else:
+                            raw_loc_indices[d] = md
```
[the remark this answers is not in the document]
georgebisbas: Is was this tutorial only failing..

https://github.com/devitocodes/devito/blob/master/examples/seismic/tutorials/12_time_blocking.ipynb

need to check closer if I can factor out some mfe I guess

devito/passes/iet/mpi.py
```diff
@@ -48,157 +49,133 @@ def _drop_halospots(iet):
...
+    """
 
     # Analysis
-    cond_mapper = MapHaloSpots().visit(iet)
-    cond_mapper = {hs: {i for i in v if i.is_Conditional and
-                        not isinstance(i.condition, GuardFactorEq)}
-                   for hs, v in cond_mapper.items()}
-
-    iter_mapper = MapNodes(Iteration, HaloSpot, 'immediate').visit(iet)
+    cond_mapper = _make_cond_mapper(iet)
```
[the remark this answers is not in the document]
georgebisbas: sure

devito/passes/iet/mpi.py
```diff
@@ -368,3 +345,55 @@ def mpiize(graph, **kwargs):
...
+
+
+def _make_cond_mapper(iet):
+    cond_mapper = MapHaloSpots().visit(iet)
+    return {hs: {i for i in v if i.is_Conditional and
+                 not isinstance(i.condition, GuardFactorEq)}
+            for hs, v in cond_mapper.items()}
+
+
+def _ensure_control_flow(hs0, hs1, cond_mapper):
```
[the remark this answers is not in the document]
georgebisbas: Yes

examples/seismic/viscoacoustic/operators.py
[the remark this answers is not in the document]
georgebisbas: Yes

examples/seismic/viscoelastic/operators.py
```diff
@@ -64,4 +64,4 @@ def ForwardOperator(model, geometry, space_order=4, save=False, **kwargs):
 
     # Substitute spacing terms to reduce flops
     return Operator([u_v, u_r, u_t] + src_rec_expr, subs=model.spacing_map,
-                    name='Forward', **kwargs)
+                    name='ViscoElForward', **kwargs)
```
georgebisbas, line 67: same way

georgebisbas (commented):
I realized that I had added comments but were not posted, only drafted...releasing them + some new ones

georgebisbas (commented):
More comments and fixes to follow tmr

devito/passes/iet/mpi.py
```diff
@@ -368,3 +345,55 @@ def mpiize(graph, **kwargs):
...
+def _filter_iter_mapper(iet):
+    """
+    Given an IET, return a mapper from Iterations to the HaloSpots.
+    Additionally, filter out Iterations that are not of interest.
+    """
+    iter_mapper = MapNodes(Iteration, HaloSpot, 'immediate').visit(iet)
+    iter_mapper = {k: [hs for hs in v if not hs.halo_scheme.is_void]
+                   for k, v in iter_mapper.items()}
+    iter_mapper = {k: v for k, v in iter_mapper.items() if k is not None}
+    iter_mapper = {k: v for k, v in iter_mapper.items() if len(v) > 1}
```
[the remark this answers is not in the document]
georgebisbas: There is nothing to optimize if you have only one candidate...

devito/passes/iet/mpi.py
```diff
@@ -11,27 +12,27 @@
...
+    iet = _hoist_invariant(iet)
     iet = _merge_halospots(iet)
     iet = _drop_if_unwritten(iet, **kwargs)
     iet = _mark_overlappable(iet)
 
     return iet, {}
 
 
-def _drop_halospots(iet):
+def _drop_reduction_halospots(iet):
```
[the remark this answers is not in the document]
georgebisbas: did it, although docstring was already kinda specific.

devito/ir/iet/nodes.py
```diff
@@ -1430,8 +1438,20 @@ def DummyExpr(*args, init=False):
...
+            loc_indices = OrderedSet(*(self.fmapper[f].loc_indices.values()))
+            loc_indices_str = str(list(loc_indices)) if loc_indices else ""
+            fstrings.append("%s%s" % (f.name, loc_indices_str))
 
-class HaloSpot(Node):
+        functions = ",".join(fstrings)
+        return "<%s(%s)>" % (self.__class__.__name__, functions)
+
+
+class HaloSpot(HaloMixin, Node):
```
FabioLuporini, line 1454: after re-thinking about this change, I struggle to understand.

* HaloScheme (in `mpi.py`) has this new `__repr__`
* HaloSpot (here) carries a `halo_scheme`
* Hence, couldn't `HaloSpot.__repr__` simply resort to `self.halo_scheme.__repr__()` (or a wrapper of it if you want to add more information)

I think this thing that the same exact representation is used in two different places doesn't make much sense

devito/mpi/halo_scheme.py
FabioLuporini: so what's the issue? 

There's a lot of redundant code here -- we should use EnrichedTuple

Then:

* you shouldn't need to override `__eq__`
* you shouldn't need to override `__repr__`
* I have some doubts about overriding `__hash__`, but maybe also see comment below

devito/mpi/halo_scheme.py
```diff
@@ -28,14 +29,41 @@ class HaloLabel(Tag):
...
+
+    def __eq__(self, other):
+        if not isinstance(other, HaloSchemeEntry):
+            return False
+        return (self.loc_indices == other.loc_indices and
+                self.loc_dirs == other.loc_dirs and
+                self.halos == other.halos and
+                self.dims == other.dims)
+
+    def __hash__(self):
```
FabioLuporini, line 50: why do you need to use all these frozentsets?

if `halos` and `dims` are mutuable, they should be turned into immutable inside `__init__`

devito/mpi/halo_scheme.py
```diff
@@ -556,7 +596,7 @@ def process_loc_indices(raw_loc_indices, directions):
     known = set().union(*[i._defines for i in loc_indices])
     loc_dirs = {d: v for d, v in directions.items() if d in known}
 
-    return frozendict(loc_indices), frozendict(loc_dirs)
+    return loc_indices, loc_dirs
```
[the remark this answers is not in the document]
FabioLuporini: ping

devito/passes/iet/mpi.py
```diff
@@ -48,157 +47,129 @@ def _drop_halospots(iet):
...
+                continue
+
+            # If there are overlapping loc_indices, skip
+            hs0_mdims = hs0.halo_scheme.loc_values
+            hs1_mdims = hs1.halo_scheme.loc_values
+            if hs0_mdims.intersection(hs1_mdims):
+                continue
+
+            for f, v in hs1.fmapper.items():
+
```
FabioLuporini, line 97: no need for a blank line here I think

devito/passes/iet/mpi.py
```diff
@@ -48,157 +47,129 @@ def _drop_halospots(iet):
...
+                    raw_loc_indices = {}
+
+                    for d, v in hse.loc_indices.items():
+                        if v in it.uindices:
+                            v_sub = it.start
+                            raw_loc_indices[d] = v.symbolic_min.subs(it.dim, v_sub)
+                        else:
+                            raw_loc_indices[d] = v
+
+                    hse = hse._rebuild(loc_indices=frozendict(raw_loc_indices))
```
FabioLuporini, line 117: frozendict cast should be done inside the HSE `__init__`

devito/passes/iet/mpi.py
```diff
@@ -48,157 +47,129 @@ def _drop_halospots(iet):
...
+                else:
+                    # hs1 can be hoisted out of `it`, but we need to infer valid
+                    # loc_indices
+                    hse = hs1.halo_scheme.fmapper[f]
+                    raw_loc_indices = {}
+
+                    for d, v in hse.loc_indices.items():
+                        if v in it.uindices:
+                            v_sub = it.start
+                            raw_loc_indices[d] = v.symbolic_min.subs(it.dim, v_sub)
```
FabioLuporini, line 113: what can v.symbolic_min be here?

devito/passes/iet/mpi.py
FabioLuporini: good

devito/passes/iet/mpi.py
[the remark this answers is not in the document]
FabioLuporini: ah right oc...

devito/passes/iet/mpi.py
```diff
@@ -368,3 +339,55 @@ def mpiize(graph, **kwargs):
...
+                   for k, v in iter_mapper.items()}
+    iter_mapper = {k: v for k, v in iter_mapper.items() if k is not None}
+    iter_mapper = {k: v for k, v in iter_mapper.items() if len(v) > 1}
+
+    return iter_mapper
+
+
+def _make_cond_mapper(iet):
+    cond_mapper = MapHaloSpots().visit(iet)
+    return {hs: {i for i in v if i.is_Conditional and
```
FabioLuporini, line 362: nitpicking (for code homogeneity), the `if` part of a multi-line comprehension goes on the next line, ie

```
[a for a in X
 if (condA and
     condB and
     condC)]
```

FabioLuporini: actually, this is a nested comprehension, you might actually use an explicit for loop...

devito/passes/iet/mpi.py
```diff
@@ -368,3 +339,55 @@ def mpiize(graph, **kwargs):
...
+
+# *** Utilities
+
+def _filter_iter_mapper(iet):
+    """
+    Given an IET, return a mapper from Iterations to the HaloSpots.
+    Additionally, filter out Iterations that are not of interest.
+    """
+    iter_mapper = MapNodes(Iteration, HaloSpot, 'immediate').visit(iet)
+    iter_mapper = {k: [hs for hs in v if not hs.halo_scheme.is_void]
```
FabioLuporini, line 352: nitpicking: three consecutive comprehensions, you may be better off with an explicit for loop (aside from the performance considerations since these lists are tiny)

examples/seismic/viscoelastic/operators.py
FabioLuporini: the new names aren't consistent since here you don't specify the type of media

tests/test_mpi.py
```diff
@@ -1005,6 +1008,183 @@ def test_avoid_haloupdate_if_distr_but_sequential(self, mode):
...
+        u_tau = Eq(tau.forward, solve(pde_tau, tau.forward))
+
+        # Receiver
+        rec = SparseTimeFunction(name="rec", grid=grid, npoint=1, nt=tn)
+        rec.coordinates.data[:, 0] = np.linspace(0., shape[0], num=1)
+
+        return grid, v, tau, u_v, u_tau, rec
+
+    @pytest.mark.parallel(mode=1)
+    def test_issue_2448_I(self, mode, setup):
```
FabioLuporini, line 1036: nitpicking: we use v0, v1, v2 ... not roman numbers (again, nothing against it, but we need code homogeneity)

tests/test_mpi.py
```diff
@@ -1005,6 +1008,183 @@ def test_avoid_haloupdate_if_distr_but_sequential(self, mode):
...
+        v = TimeFunction(name='v', grid=grid, space_order=so)
+        tau = TimeFunction(name='tau', grid=grid, space_order=so)
+
+        # First order elastic-like dependencies equations
+        pde_v = v.dt - (tau.dx)
+        pde_tau = tau.dt - ((v.forward).dx)
+        u_v = Eq(v.forward, solve(pde_v, v.forward))
+        u_tau = Eq(tau.forward, solve(pde_tau, tau.forward))
+
+        # Receiver
```
FabioLuporini, line 1029: useless

tests/test_mpi.py
```diff
@@ -1005,6 +1008,183 @@ def test_avoid_haloupdate_if_distr_but_sequential(self, mode):
         calls = FindNodes(Call).visit(op)
         assert len(calls) == 0
 
+    @pytest.fixture
+    def setup(self):
```
FabioLuporini, line 1012: too generic if it's specific to a set of tests

tests/test_mpi.py
```diff
@@ -1005,6 +1008,183 @@ def test_avoid_haloupdate_if_distr_but_sequential(self, mode):
...
+
+        assert len(calls) == 3
+        assert len(FindNodes(HaloUpdateCall).visit(op0.body.body[1].body[1].body[0])) == 1
+        assert len(FindNodes(HaloUpdateCall).visit(op0.body.body[1].body[1].body[1])) == 2
+        assert calls[0].arguments[0] is v
+        assert calls[1].arguments[0] is tau
+        assert calls[2].arguments[0] is v
+
+    @pytest.mark.parallel(mode=1)
+    def test_issue_2448_II(self, mode, setup):
```
FabioLuporini, line 1053: leaving the comment here but it applies to this entire new code segment. SInce it's a large batch of tests all around issue 2448, you could think of creating a class `class TestIssue2448` 

Then the name `setup` for the fixture would become acceptable

And you could add a docstring for more general explanation

tests/test_mpi.py
FabioLuporini: memo

tests/test_mpi.py
```diff
@@ -2727,6 +2911,71 @@ def test_adjoint_F_no_omp(self, mode):
...
+
+        v = VectorTimeFunction(name='v', grid=grid, space_order=so)
+        tau = TensorTimeFunction(name='t', grid=grid, space_order=so)
+
+        damp = Function(name='damp', grid=grid)
+        l = Function(name='lam', grid=grid)
+        mu = Function(name='mu', grid=grid)
+        ro = Function(name='b', grid=grid)
+
+        # The receiver
```
FabioLuporini, line 2930: useless

tests/test_mpi.py
FabioLuporini: Another possibility:

I just saw that there's a new `TestElastic`...

maybe call it `TestElasticLike` and move this whole batch of new tests in there?

Up to you

tests/test_mpi.py
```diff
@@ -2727,6 +2911,71 @@ def test_adjoint_F_no_omp(self, mode):
...
+        # Time update
+        u_v = Eq(v.forward, damp * solve(pde_v, v.forward))
+        u_t = Eq(tau.forward, damp * solve(pde_tau, tau.forward))
+
+        op = Operator([u_v] + [u_t] + rec_term)
+        assert len(op._func_table) == 11
+
+        calls = [i for i in FindNodes(Call).visit(op) if isinstance(i, HaloUpdateCall)]
+
+        # The correct we want
```
FabioLuporini, line 2948: useless

tests/test_mpi.py
```diff
@@ -2727,6 +2911,71 @@ def test_adjoint_F_no_omp(self, mode):
...
+        assert len(calls) == 5
+
+        assert len(FindNodes(HaloUpdateCall).visit(op.body.body[1].body[1].body[0])) == 1
```
FabioLuporini, line 2951: are you saying these are not covered already in the new `test_issue_*` that you've added above?

it seems to be the same dependency pattern, hence a redundant test...

tests/test_mpi.py
```diff
@@ -2727,6 +2911,71 @@ def test_adjoint_F_no_omp(self, mode):
...
+        assert calls[3].arguments[0] is tau[1, 1]
+        assert calls[4].arguments[0] is v[0]
+        assert calls[4].arguments[1] is v[1]
+
+
+class TestTTIOp:
+
+    @pytest.mark.parallel(mode=1)
+    def test_halo_structure(self, mode):
+
```
FabioLuporini, line 2968: useless blank line

devito/mpi/halo_scheme.py
georgebisbas: I still do not understand this. SHould I still be able to construct HaloSchemeEntry the same way I do it now?
Will I need to specify getters ?
Like e.g. here:
```
    def glb_shape(self):
        """Shape of the decomposed domain."""
        return EnrichedTuple(*self._glb_shape, getters=self.dimensions)
```

devito/passes/iet/mpi.py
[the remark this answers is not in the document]
georgebisbas: I think this can be resolved

tests/test_mpi.py
georgebisbas: Updated

devito/mpi/halo_scheme.py
georgebisbas: is this fine now, or I still do not understand?

devito/passes/iet/mpi.py
georgebisbas: This is what happens here, in case `v` is not in uindices like the time_blocking example we use the already existing variable. In backwards cases `v_sub` is `time_M`.
```
(Pdb) print(v.symbolic_min)
Mod(time, 2)
(Pdb) print(v_sub)
time_m
(Pdb) print(raw_loc_indices)
{t: Mod(time_m, 2)}
```

devito/ir/iet/nodes.py
```diff
@@ -1492,6 +1495,9 @@ def body(self):
     def functions(self):
         return tuple(self.fmapper)
 
+    def __repr__(self):
```
FabioLuporini, line 1498: show stay below `__init__`, just like all other special `__X` methods

devito/mpi/halo_scheme.py
```diff
@@ -28,14 +28,41 @@ class HaloLabel(Tag):
...
+                f"loc_dirs={self.loc_dirs}, halos={self.halos}, dims={self.dims})")
+
 
 Halo = namedtuple('Halo', 'dim side')
 
 OMapper = namedtuple('OMapper', 'core owned')
 
 
-class HaloScheme:
+class HaloScheme():
```
FabioLuporini, line 65: avoid

devito/ir/iet/nodes.py
```diff
@@ -1492,6 +1495,9 @@ def body(self):
...
+        funcs = self.halo_scheme.__reprfuncs__()
```
FabioLuporini, line 1499: again, it's a useless complication. Just return something along the lines of `HaloSpot(f,g)` and if the user really cares about the underlying `halo_scheme`, they can always do `hs.halo_scheme` and this way get the textual information they are after

so, IOW, drop `__reprfuncs__`

devito/mpi/halo_scheme.py
FabioLuporini: > SHould I still be able to construct HaloSchemeEntry the same way I do it now?

what happens if you try (so yes, w/o specifying any getters)

devito/mpi/halo_scheme.py
```diff
@@ -93,9 +120,17 @@ def __init__(self, exprs, ispace):
             self._honored[i.root] = frozenset([(ltk, rtk)])
         self._honored = frozendict(self._honored)
 
+    def __reprfuncs__(self):
```
FabioLuporini, line 123: as explained above, avoid; put it back inside `__repr__`

devito/mpi/halo_scheme.py
```diff
@@ -121,7 +156,10 @@ def union(self, halo_schemes):
         Create a new HaloScheme from the union of a set of HaloSchemes.
         """
         halo_schemes = [hs for hs in halo_schemes if hs is not None]
-        if not halo_schemes:
+
+        if len(halo_schemes) == 1:
+            return halo_schemes[0]
```
FabioLuporini, line 161: why this change?

devito/mpi/halo_scheme.py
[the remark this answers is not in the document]
FabioLuporini: ok so these two lines above

```
        halos = frozenset(halos)
        dims = frozenset(dims)
```

are droppable as well?

devito/passes/iet/mpi.py
```diff
@@ -48,157 +47,128 @@ def _drop_halospots(iet):
...
+
+            for f, v in hs1.fmapper.items():
+                if f not in hs0.functions:
+                    continue
+
+                for dep in scopes[it].d_flow.project(f):
+                    if not any(r(dep, hs1, v.loc_indices) for r in rules):
                         break
+                else:
+                    # hs1 can be hoisted out of `it`, but we need to infer valid
```
FabioLuporini, line 104: nitpicking: should be \`hs1\`

devito/passes/iet/mpi.py
```diff
@@ -48,157 +47,128 @@ def _drop_halospots(iet):
...
+                    # loc_indices
+                    hse = hs1.halo_scheme.fmapper[f]
+                    raw_loc_indices = {}
+
+                    for d, v in hse.loc_indices.items():
+                        if v in it.uindices:
+                            v_sub = it.start
```
FabioLuporini, line 111: if you call the variable just `loc_indices` instead of `raw_loc_indices` maybe you don't need this extra temporary `v_sub` as the line above will fit on the same line

devito/passes/iet/mpi.py
```diff
@@ -48,157 +47,128 @@ def _drop_halospots(iet):
...
+                            raw_loc_indices[d] = v.symbolic_min.subs(it.dim, v_sub)
+                        else:
+                            raw_loc_indices[d] = v
+
+                    hse = hse._rebuild(loc_indices=raw_loc_indices)
```
FabioLuporini, line 116: again you might not need this extra variable if u just use `loc_indices`

devito/passes/iet/mpi.py
```diff
@@ -368,3 +338,62 @@ def mpiize(graph, **kwargs):
...
+    for k, v in MapNodes(Iteration, HaloSpot, 'immediate').visit(iet).items():
+        filtered_hs = [hs for hs in v if not hs.halo_scheme.is_void]
+        if k is not None and len(filtered_hs) > 1:
+            iter_mapper[k] = filtered_hs
+
+    return iter_mapper
+
+
+def _make_cond_mapper(iet):
+
```
FabioLuporini, line 360: no blank line; add simple docstring

devito/passes/iet/mpi.py
```diff
@@ -368,3 +338,62 @@ def mpiize(graph, **kwargs):
...
+    cond_mapper = {}
+    for hs, v in MapHaloSpots().visit(iet).items():
+        conditionals = set()
+        for i in v:
+            if i.is_Conditional and not isinstance(i.condition, GuardFactorEq):
```
FabioLuporini, line 365: this smells like a hack. What are you trying to accomplish here?

Is this a recent addition or is it something I've missed all along until now?

also, it should have been a set comprehension --

```
conditionals = {i for i in v if i.is_Conditional ...}
```

But again, this line seems a bit dodgy to me

tests/test_mpi.py
```diff
@@ -1220,7 +1223,7 @@ def test_avoid_fullmode_if_crossloop_dep(self, mode):
         assert np.all(f.data[:] == 2.)
 
     @pytest.mark.parallel(mode=2)
-    def test_avoid_haloudate_if_flowdep_along_other_dim(self, mode):
+    def test_avoid_haloupdate_if_flowdep_along_other_dim(self, mode):
```
FabioLuporini, line 1226: why is this being tagged as a diff?

tests/test_mpi.py
```diff
@@ -1333,8 +1336,10 @@ def test_merge_haloupdate_if_diff_locindices_v1(self, mode):
 
         * the second and third Eqs cannot be fused in the same loop
 
-        In the IET we end up with *one* HaloSpots, placed right before the
-        second Eq. The third Eq will seamlessy find its halo up-to-date.
+        In the IET we end up with *two* HaloSpots, one placed before the
```
FabioLuporini, line 1339: does the name of the test still make sense or should it change as well?

devito/mpi/halo_scheme.py
georgebisbas: There is zero work done when you have one haloscheme...so early exit...

devito/passes/iet/mpi.py
georgebisbas: ok

tests/test_mpi.py
georgebisbas: right, renamed

FabioLuporini (commented):
basically ready, only one docstring issue remaining which is a bit meh so maybe we should fix it before merging

EdCaunt (approved):

FabioLuporini (approved):

PR #2482, opened 2024-11-13
dsl: Add ArgProvider inheritance to AbstractSymbol

EdCaunt:
Adds a default no-op `_arg_values`, `_arg_check` etc to any subclasses of `AbstractSymbol`.

### DIFF
--- a/devito/types/basic.py
+++ b/devito/types/basic.py
@@ -297,7 +297,7 @@ class Basic(CodeSymbol):
         return set()
 
 
-class AbstractSymbol(sympy.Symbol, Basic, Pickable, Evaluable, ArgProvider):
+class AbstractSymbol(sympy.Symbol, Basic, Pickable, Evaluable):
 
     """
     Base class for scalar symbols.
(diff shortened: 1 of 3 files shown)


devito/types/basic.py
```diff
@@ -549,7 +549,7 @@ def __new__(cls, *args, **kwargs):
     __hash__ = Uncached.__hash__
 
 
-class Scalar(Symbol, ArgProvider):
+class Scalar(Symbol):
```
mloubout, line 552: Well that was intentional 
- Scalar carry data
- Symbol don't carry data

So the second one is not an arg provider as it does not have values

mloubout (commented):
What issue is this for?

devito/types/basic.py
EdCaunt: It's to remove the extraneous no-op methods in `Thickness`. Maybe `DataSymbol` should get `ArgProvider` instead then?

FabioLuporini (approved):

PR #2484, opened 2024-11-13
ci: Add Flake8 linting to notebooks

EdCaunt:
@ggorman pointed out that you can now Flake8 lint Jupyter Notebooks. We should probably do this.

### DIFF
--- a/.github/workflows/flake8.yml
+++ b/.github/workflows/flake8.yml
@@ -5,14 +5,14 @@ concurrency:
   cancel-in-progress: true
 
 on:
-   # Trigger the workflow on push or pull request,
-   # but only for the master branch
-   push:
-     branches:
-       - master
-   pull_request:
-     branches:
-       - master
+  # Trigger the workflow on push or pull request,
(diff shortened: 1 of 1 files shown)


mloubout:
I personally wouldn't. Flake8 is really not adapted to notebooks, where readable and potentially longer lines are preferred.

EdCaunt:
Looking at it, there are only about 5 "Line too long" warnings in the whole output, and most are only 5 or so characters over.

mloubout (changes requested):
This diff is way too big for just a few line change. Also those depa should be in requirements-testing.txt not hardcoded in the yaml

EdCaunt:
> This diff is way too big for just a few line change. 

Yeah, I messed up adding the changes in one of the notebooks somehow. Will fix in due course

EdCaunt:
> Also those depa should be in requirements-testing.txt not hardcoded in the yaml

I was just going off how the Flake8 workflow already works, can change this

PR #2487, opened 2024-11-15
compiler: Expose Bundle's properties correctly

FabioLuporini:


devito/types/array.py
```diff
@@ -441,12 +443,13 @@ def initvalue(self):
...
 
-    for i in ['_mem_internal_eager', '_mem_internal_lazy', '_mem_local',
+    for i in ('_mem_internal_eager', '_mem_internal_lazy', '_mem_local',
               '_mem_mapped', '_mem_host', '_mem_stack', '_mem_constant',
               '_mem_shared', '__padding_dtype__', '_size_domain', '_size_halo',
               '_size_owned', '_size_padding', '_size_nopad', '_size_nodomain',
               '_offset_domain', '_offset_halo', '_offset_owned', '_dist_dimensions',
-              '_C_get_field', 'grid', 'symbolic_shape']:
+              '_C_get_field', 'grid', 'symbolic_shape',
+              *AbstractFunction.__properties__):
```
EdCaunt, line 452: Nitpick: `('_mem_internal_eager', '_mem_internal_lazy', ...) + AbstractFunction.__properties__` might be tidier imo

devito/types/array.py
```diff
@@ -368,6 +368,8 @@ def __args_setup__(cls, *args, **kwargs):
             raise ValueError("Component type must be subclass of AbstractFunction")
         if len({i.__padding_dtype__ for i in components}) != 1:
             raise ValueError("Components must have the same padding dtype")
+        if len({i.properties for i in components}) != 1:
+            raise ValueError("Components must have the same properties")
```
georgebisbas, line 372: looks more like "counting" rather than "comparing", should you  say, the same number of properties?

georgebisbas (approved):

devito/types/array.py
mloubout: `a + b` creates two tuples a, b then a new third one to merge them while current one only creates one directly.

mloubout (approved):

devito/types/array.py
mloubout: It's not counting properties it's check the set length one, i.e all properties are the same

PR #2488, opened 2024-11-15
api: fix pickling of sparse operations

mloubout:


FabioLuporini (approved):
Uncontroversial, thanks

PR #2501, opened 2024-12-18
compiler: Generate fminf/fmaxf where necessary

FabioLuporini:


tests/test_symbolics.py
```diff
@@ -562,6 +562,31 @@ def test_minmax():
...
+
+    op = Operator(eqn)
+
+    g.data[:] = 3.0
+
+    op.apply()
+
+    # Check generated code -- ensure it's using the fp64 versions of min/max,
+    # that is fminf/fmaxf
+    assert all(i in str(op) for i in expected)
```
EdCaunt, line 585: Would it be good (belt and braces) to also check:
`assert all(i not in str(op) for i in not_expected)`
to ensure that the correct versions are used throughout?

tests/test_symbolics.py
```diff
@@ -562,6 +562,31 @@ def test_minmax():
...
+    (np.float32, ("fmaxf", "fminf")),
+    (np.float64, ("fmax", "fmin")),
+])
+def test_minmax_precision(dtype, expected):
+    grid = Grid(shape=(5, 5), dtype=dtype)
+
+    f = Function(name="f", grid=grid)
+    g = Function(name="g", grid=grid)
+
+    eqn = Eq(f, Min(g, 4.0) + Max(g, 2.0))
```
mloubout, line 575: Would use numpy numbers to check dtype is correct

tests/test_symbolics.py
FabioLuporini: overkill I'd say...

tests/test_symbolics.py
FabioLuporini: wdym exactly? I can think of at least two ways...

I landed on the above because that's the natural way people write the equations and it doesn't matter whether it's 4, 4.0, or 4.0F, because in the end, `g` will drive the `fmin/fmax` generation

georgebisbas (approved):

tests/test_symbolics.py
FabioLuporini: @mloubout I'm gonna merge this because I don't like broken CI, but as soon as you elaborate on how you want this test improved, I'll mkae the change and push it together with one of the upcoming branches (at least one more to go in before christmas)

PR #2502, opened 2024-12-19
compiler: Add 'groupby' mode to MapNodes visitor

ZoeLeibowitz:
Add `MapNodes(..., mode='groupby')` back, which was dropped in #2477 and I didn't see this. It is useful for the PETSc lowering.

### DIFF
--- a/tests/test_visitors.py
+++ b/tests/test_visitors.py
@@ -390,8 +390,7 @@ def test_map_nodes(block1):
     assert len(map_nodes.keys()) == 1
 
     for iters, (expr,) in map_nodes.items():
-        # Replace the outermost `Iteration` with a placeholder, represented in
-        # this example by a `Call` to a `Callable`
+        # Replace the outermost `Iteration` with a `Call`
         callback = Callable('solver', iters[0], 'void', ())
         processed = Transformer({iters[0]: Call(callback.name)}).visit(block1)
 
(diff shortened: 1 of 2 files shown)


tests/test_visitors.py
```diff
@@ -376,3 +376,22 @@ def test_find_symbols_with_duplicates():
...
+    """
+    map_nodes = MapNodes(Iteration, Expression, mode='groupby').visit(block1)
+
+    assert len(map_nodes.keys()) == 1
+
+    for iters, (expr,) in map_nodes.items():
+        # Replace the outermost `Iteration` with a `Call`
+        callback = Callable('solver', iters[0], 'void', ())
+        processed = Transformer({iters[0]: Call(callback.name)}).visit(block1)
+
```
georgebisbas, line 396: I think it is fine!
Nipick: Maybe you could add:

+assert len(iters) == 3
assert str(processed) == 'solver();'

but sure it is not needed.

georgebisbas (approved):

tests/test_visitors.py
ZoeLeibowitz: I have `assert str(processed) == 'solver();'` just below

FabioLuporini (approved):

EdCaunt (approved):

PR #2503, opened 2024-12-20
compiler: Avoid allocating Bundles on the host if transient

FabioLuporini:
Impacted test tweaked in PRO

### DIFF
--- a/devito/passes/iet/definitions.py
+++ b/devito/passes/iet/definitions.py
@@ -10,14 +10,16 @@ from operator import itemgetter
 import numpy as np
 
 from devito.ir import (Block, Call, Definition, DummyExpr, Return, EntryFunction,
-                       FindSymbols, MapExprStmts, Transformer, make_callable)
+                       FindNodes, FindSymbols, MapExprStmts, Transformer,
+                       make_callable)
 from devito.passes import is_gpu_create
 from devito.passes.iet.engine import iet_pass
 from devito.passes.iet.langbase import LangBB
 from devito.symbolics import (Byref, DefFunction, FieldFromPointer, IndexedPointer,
                               SizeOf, VOID, Keyword, pow_to_mul)
 from devito.tools import as_mapper, as_list, as_tuple, filter_sorted, flatten
-from devito.types import Array, CustomDimension, DeviceMap, DeviceRM, Eq, Symbol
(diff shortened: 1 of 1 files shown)


devito/passes/iet/definitions.py
```diff
@@ -214,6 +216,38 @@ def _alloc_mapped_array_on_high_bw_mem(self, site, obj, storage, *args):
 
         storage.update(obj, site, allocs=alloc, frees=free, efuncs=(efunc0, efunc1))
 
+    def _alloc_bundle_struct_on_high_bw_mem(self, site, obj, storage):
```
EdCaunt, line 219: Nitpick: some comments would help reduce the bus factor of this code

devito/passes/iet/definitions.py
```diff
@@ -340,9 +374,22 @@ def place_definitions(self, iet, globs=None, **kwargs):
         for i in FindSymbols().visit(iet):
             if i in defines:
                 continue
+
             elif i.is_LocalObject:
                 self._alloc_object_on_low_lat_mem(iet, i, storage)
-            elif i.is_Array or i.is_Bundle:
+
+            elif i.is_Bundle:
```
EdCaunt, line 381: Is there any way to reduce the "death by conditionals" here?

FabioLuporini: After spending a couple of hours on this very matter this morning, I'm inclined to say that it's pretty challenging to come up with one. The problem is that there's a combinatorial explosion of possibilities (luckily with small exponent), and all those `if`s implement such space

devito/passes/iet/definitions.py
FabioLuporini: I agree, if it turns green I will add it to the to-be-rebased #2500

devito/passes/iet/definitions.py
EdCaunt: I suppose it's all attributes too, which makes it even more challenging without obfuscating what's going on

EdCaunt (approved):

devito/passes/iet/definitions.py
FabioLuporini: if the implicit question was "can you use singledispatch", yeah, you got it right

PR #2504, opened 2024-12-20
compiler: Fix code generation for pow and fabs when using float32 

JDBetteridge:


tests/test_symbolics.py
```diff
@@ -298,7 +298,10 @@ def test_extended_sympy_arithmetic():
 def test_integer_abs():
     i1 = Dimension(name="i1")
     assert ccode(Abs(i1 - 1)) == "abs(i1 - 1)"
-    assert ccode(Abs(i1 - .5)) == "fabs(i1 - 5.0e-1F)"
+    assert ccode(Abs(i1 - .5)) == "fabsf(i1 - 5.0e-1F)"
+    assert ccode(
+        Abs(i1 - Constant('half', dtype=np.float64, default_value=0.5))
+    ) == "fabs(i1 - half)"
```
JDBetteridge, line 304: I'm not happy with this, but don't 100% know how I'm supposed to generate a literal double precision `0.5` symbolically

tests/test_symbolics.py
```diff
@@ -587,6 +590,50 @@ def test_minmax_precision(dtype, expected):
...
+
+    eqn = Eq(f, abs(g))
+
+    op = Operator(eqn)
+
+    g.data[:] = -1.0
+
+    op.apply()
+
+    assert expected in str(op)
```
JDBetteridge, line 633: This seems like a bad way to test that the right thing is happening, especially since `pow` is a substring of `powf` and `abs` is a substring of `absf`. However, I have just cargo-cult'd the above test!

FabioLuporini: that's OK

devito/symbolics/printer.py
```diff
@@ -159,7 +169,16 @@ def _print_Abs(self, expr):
...
+            func = "abs"
+        elif self.single_prec(expr):
+            func = "fabsf"
+        elif any([isinstance(a, Real) for a in expr.args[0].args]):
+            # The previous condition isn't sufficient to detect case with
+            # Python `float`s in that case, fall back to the "default"
+            func = "fabsf" if self.single_prec() else "fabs"
+        else:
+            func = "fabs"
         return "%s(%s)" % (func, self._print(expr.args[0]))
```
FabioLuporini, line 182: f-string, since you're in the neighbour of one old-style string?

FabioLuporini (approved):
LGTM, thanks!

devito/symbolics/printer.py
```diff
@@ -122,15 +125,22 @@ def _print_math_func(self, expr, nest=False, known=None):
...
-            pass
-        return super()._print_Pow(expr)
+        # Completely reimplement `_print_Pow` from sympy, since it doesn't
+        # correctly handle precision
+        if "Pow" in self.known_functions:
+            return self._print_Function(expr)
+        PREC = precedence(expr)
+        suffix = 'f' if self.single_prec(expr) else ''
+        if equal_valued(expr.exp, -1):
+            return f'{self._print_Float(Float(1.0))}/' + \
```
mloubout, line 135: Why do you need the f-string, both those methods shoulf already return a string.

devito/symbolics/printer.py
```diff
@@ -122,15 +125,22 @@ def _print_math_func(self, expr, nest=False, known=None):
...
+                f'{self.parenthesize(expr.base, PREC)}'
+        elif equal_valued(expr.exp, 0.5):
+            return f'{self._ns}sqrt{suffix}({self._print(expr.base)})'
```
mloubout, line 138: What's _ns again?

devito/symbolics/printer.py
```diff
@@ -159,8 +169,17 @@ def _print_Abs(self, expr):
         if isinstance(self.compiler, AOMPCompiler):
             return "fabs(%s)" % self._print(expr.args[0])
         # Check if argument is an integer
-        func = "abs" if has_integer_args(*expr.args[0].args) else "fabs"
-        return "%s(%s)" % (func, self._print(expr.args[0]))
+        if has_integer_args(*expr.args[0].args):
+            func = "abs"
+        elif self.single_prec(expr):
+            func = "fabsf"
+        elif any([isinstance(a, Real) for a in expr.args[0].args]):
```
mloubout, line 176: From the look of it this seems un-necessary. You just need

```
If self.single_prec(expr) or self.single_prec():
    fabsf
else:
    # not integer or f32, default 
    fabs
```

tests/test_symbolics.py
mloubout: Just np.float64(0.5)

devito/symbolics/printer.py
JDBetteridge: The name space, this is just lifted straight from the sympy code

mloubout: I don't think we handle namespaces here through the printer but directly at method level if needed so I would remove it for now. Only std one might be needed but it's default so not really needed

devito/symbolics/printer.py
JDBetteridge: Not so, this is the example that I shared on slack, but will repeat here for reference:
```python
>>> i1 = Dimension(name="i1")
>>> sympy_dtype(i1 - 0.5)
<class 'numpy.int32'>
```
In the second part of the `or` in your example `self.single_prec()` will always return `True`, regardless of the types of the expression (unless the default is changed).

This code is a bit horrible, but ensures that the "default" is only used if there is a floating point number whose type cannot be determined. Otherwise strictly use the correct function call for `float` or `double`.

tests/test_symbolics.py
JDBetteridge:
```python
>>> ccode(Abs(i1 - np.float64(0.5)))
'fabs(i1 - 5.0e-1F)'
```
It looks like the float64 is being demoted to float32, this could point to an issue elsewhere

devito/symbolics/printer.py
JDBetteridge: They do, I'll edit this. It made sense when everything was on one line, but that was failing linting

devito/symbolics/printer.py
JDBetteridge: Ah, but my test is wrong anyway...

tests/test_symbolics.py
```diff
@@ -587,6 +590,50 @@ def test_minmax_precision(dtype, expected):
...
+
+    eqn = Eq(f, g**1.5)
+
+    op = Operator(eqn)
+
+    g.data[:] = 4.0
+
+    op.apply()
+
+    assert expected in str(op)
```
EdCaunt, line 611: If you check for `pow(` and `powf(` in the string, then you can also check that the other is not present

JDBetteridge:
I'm going to add the outstanding unresolved conversations to an issue and potentially tackle it in another PR, after some more thought/discussion.

tests/test_symbolics.py
JDBetteridge: Good point!

devito/symbolics/printer.py
```diff
@@ -159,8 +168,17 @@ def _print_Abs(self, expr):
...
+            # The previous condition isn't sufficient to detect case with
+            # Python `float`s in that case, fall back to the "default"
+            func = "fabsf" if self.single_prec() else "fabs"
```
georgebisbas, line 178: I can see two self.single_prec tests. Isn;t this redundant here?
Can it be one check?

devito/symbolics/printer.py
FabioLuporini: I'm going to merge this, but perhaps in a subsequent PR you could attach a comment to that `if`

PR #2509, opened 2025-01-06
compiler: Concretize SubDimensions to same object across repeated calls to concretize_subdims

ZoeLeibowitz:
Ensure that SubDimensions are consistently concretized to the same object across multiple calls to the function. This is important when using `rcompile` on equations with SubDimensions.

TODO: Enhance `SymbolRegistry` by making it rebuildable, enabling operations such as `rcompile` on specific subsets.

### DIFF
--- a/requirements-optional.txt
+++ b/requirements-optional.txt
@@ -1,4 +1,4 @@
 matplotlib
 pyrevolve==2.2.4
-scipy
+scipy>=1.13.0
 distributed
--- a/requirements-optional.txt
+++ b/requirements-optional.txt
@@ -1,4 +1,4 @@
 matplotlib
 pyrevolve==2.2.4
-scipy>=1.13.0
+scipy
 distributed
--- a/tests/test_builtins.py
+++ b/tests/test_builtins.py
@@ -1,7 +1,10 @@
 import pytest
 import numpy as np
 from scipy.ndimage import gaussian_filter
-from scipy.datasets import ascent
+try:
+    from scipy.datasets import ascent
+except ImportError:
+    from scipy.misc import ascent
 
 from devito import ConditionalDimension, Grid, Function, TimeFunction, switchconfig
 from devito.builtins import (assign, norm, gaussian_smooth, initialize_function,
--- a/tests/test_builtins.py
+++ b/tests/test_builtins.py
@@ -1,7 +1,7 @@
 import pytest
 import numpy as np
 from scipy.ndimage import gaussian_filter
-from scipy.misc import ascent
+from scipy.datasets import ascent
 
 from devito import ConditionalDimension, Grid, Function, TimeFunction, switchconfig
 from devito.builtins import (assign, norm, gaussian_smooth, initialize_function,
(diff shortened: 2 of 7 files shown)


tests/test_builtins.py
```diff
@@ -1,7 +1,7 @@
 import pytest
 import numpy as np
 from scipy.ndimage import gaussian_filter
-from scipy.misc import ascent
+from scipy.datasets import ascent
```
mloubout, line 4: If I remember correctly this one behaves a bit differently and was leading to issues in the tests but if ci is green that's fine. The requirements need to be update though to scipy version with that import

mloubout (commented):
Not completely sure carrying around an extra mapper is the best way but don't have a better option top of my head without lots of rework

EdCaunt (approved):
Looks good to me. As @mloubout mentioned, we should probably revisit and rework this at some point in the future, but it is a relatively minor intervention for now

tests/test_dimension.py
```diff
@@ -2050,8 +2050,31 @@ def test_correct_thicknesses(self):
...
+        `rcompile` on equations with SubDimensions.
+        """
+
+        grid = Grid((2, 2))
+
+        x = Dimension('x')
+        ix = SubDimension.middle('ix', x, 2, 2)
+
+        u = Function(name='u', grid=grid)
+        eq = Eq(u, ix.ltkn)
```
EdCaunt, line 2073: You could strengthen this test by making the rhs `ix + ix.ltkn + ix.rtkn`

tests/test_builtins.py
EdCaunt: CI now fails with the old version. [See here](https://github.com/devitocodes/devito/actions/runs/12635281488/job/35204852799?pr=2510#step:11:32)

requirements-optional.txt
```diff
@@ -1,4 +1,4 @@
 matplotlib
 pyrevolve==2.2.4
-scipy
+scipy>=1.13.0
```
mloubout, line 3: I think it's too restrictive, it's a very new version, would wrap the import in try/except to handle different version import path

mloubout (changes requested):

mloubout (approved):

requirements-optional.txt
```diff
@@ -1,4 +1,4 @@
 matplotlib
 pyrevolve==2.2.4
-scipy>=1.13.0
+scipy
```
georgebisbas, line 3: I think it is better to add an upper bound:

e.g. <=

and then (if needed) do what  @mloubout did here (?):

https://github.com/devitocodes/devito/commit/dee09dda5ad89c72d82ed9a84a839a1ccbbdb6ba

georgebisbas: keep lower bound ofc

requirements-optional.txt
georgebisbas: ah sorry jsut saw that

FabioLuporini:
> Not completely sure carrying around an extra mapper is the best way but don't have a better option top of my head without lots of rework

We discussed this in yesterday's meeting, @mloubout .

My suggestion was to rather use `sregistry`. And, in fact, we could (should) use `sregistry` also in place of that `mapper` we are currently carrying along throughout the whole pass

PR #2510, opened 2025-01-06
compiler: Fix bug in concretization of conditionals containing Thicknesses

EdCaunt:


FabioLuporini (approved):
Thanks, GTG

mloubout (commented):
Can you open an issue that subs is broken with your MFE as it might pop up again since this fix the usage not the bug

Except that GTG

mloubout (approved):

PR #2512, opened 2025-01-07
tests: Add test for using a Constant as a condition in a ConditionalDimension

EdCaunt:
This case is valid but not currently tested. It is useful for creating switches that can be set in Python-land. Example applications include preventing accesses on empty ranks when performing sparse operations on functions defined on subdomains.

### DIFF
--- a/tests/test_dimension.py
+++ b/tests/test_dimension.py
@@ -1288,6 +1288,20 @@ class TestConditionalDimension:
         op = Operator(eq)
         op.cfunction
 
+    @pytest.mark.parametrize('value', [0, 1])
+    def test_constant_as_condition(self, value):
+        x = Dimension('x')
+
+        c = Constant(name="c", dtype=np.int8, value=value)
+        cd = ConditionalDimension(name="cd", parent=x, condition=c)
+
+        f = Function(name='f', dimensions=(x,), shape=(11,), dtype=np.int32)
+
+        op = Operator(Eq(f, 1, implicit_dims=cd))
+        op.apply()
+
+        assert np.all(f.data == value)
+
     def test_symbolic_factor(self):
         """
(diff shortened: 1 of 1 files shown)


tests/test_dimension.py
```diff
@@ -1288,6 +1288,20 @@ def test_no_index_symbolic(self):
         op = Operator(eq)
         op.cfunction
 
+    @pytest.mark.parametrize('value', [0, 1])
+    def test_constant_as_condition(self, value):
+        x = Dimension('x')
+
+        c = Constant(name="c", dtype=np.int8, value=value)
```
mloubout, line 1295: Shouldn't this be a Bool if it's a condition by itself?

mloubout (approved):

tests/test_dimension.py
```diff
@@ -1288,6 +1288,20 @@ def test_no_index_symbolic(self):
...
+        cd = ConditionalDimension(name="cd", parent=x, condition=c)
+
```
georgebisbas, line 1297: can probably drop this empty line

tests/test_dimension.py
georgebisbas: Actually you can enhance the test with a bool as well? Behaviour should be the same?

EdCaunt: But C doesn't have a core bool type? You have to import `stdbool.h`

EdCaunt: Convention is to use an int iirc

FabioLuporini: yes, int is OK

FabioLuporini: int8 btw is converted into a char

FabioLuporini (approved):

PR #2515, opened 2025-01-10
compiler: Check `sympy_type` returns a floating `point type` in the code printer

JDBetteridge:
This aims to fix a bug in code generation for certain platforms

### DIFF
--- a/devito/symbolics/printer.py
+++ b/devito/symbolics/printer.py
@@ -44,7 +44,10 @@ class CodePrinter(C99CodePrinter):
         return self._settings['compiler']
 
     def single_prec(self, expr=None):
+        # Extract the dtype of the expression
         dtype = sympy_dtype(expr) if expr is not None else self.dtype
+        # Check that the dtype is a floating point type
+        dtype = dtype if np.issubdtype(dtype, np.floating) else self.dtype
         return dtype in [np.float32, np.float16]
 
     def parenthesize(self, item, level, strict=False):
--- a/devito/symbolics/printer.py
+++ b/devito/symbolics/printer.py
@@ -170,12 +170,12 @@ class CodePrinter(C99CodePrinter):
         # AOMPCC errors with abs, always use fabs
         if isinstance(self.compiler, AOMPCompiler):
             return "fabs(%s)" % self._print(expr.args[0])
-        # Check if argument is an integer
-        if has_integer_args(*expr.args[0].args):
+        # Check if argument (always exactly one!) is an integer
+        if has_integer_args(expr.args[0]):
             func = "abs"
-        elif self.single_prec(expr):
+        elif self.single_prec(expr.args[0]):
             func = "fabsf"
-        elif any([isinstance(a, Real) for a in expr.args[0].args]):
+        elif isinstance(expr.args[0], Real):
             # The previous condition isn't sufficient to detect case with
             # Python `float`s in that case, fall back to the "default"
             func = "fabsf" if self.single_prec() else "fabs"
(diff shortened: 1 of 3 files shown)


tests/test_symbolics.py
```diff
@@ -313,7 +313,7 @@ def test_cos_vs_cosf():
 
     # Doesn't make much sense, but it's legal
     c = dSymbol('c', dtype=np.int32)
-    assert ccode(cos(c)) == "cos(c)"
+    assert ccode(cos(c)) == "cosf(c)"
```
mloubout, line 316: so `ccode(Abs(c))` will be `absf(c)` which is incorrect it should be `abs(c)`.

devito/symbolics/printer.py
```diff
@@ -44,7 +44,10 @@ def compiler(self):
         return self._settings['compiler']
 
     def single_prec(self, expr=None):
+        # Extract the dtype of the expression
         dtype = sympy_dtype(expr) if expr is not None else self.dtype
+        # Check that the dtype is a floating point type
+        dtype = dtype if np.issubdtype(dtype, np.floating) else self.dtype
```
mloubout, line 50: I don't think this is valid, this will return "is float" for integer and will break a lot of cases. This is probably only necessary for some limited cases like `cos` but not in general

tests/test_symbolics.py
JDBetteridge: You are correct, there is a bug here with your specific example. However, I think I introduced it in my last PR and it is specific to `abs`/`fabsf`/`fabs`, which I will address.

But, in this test, why should the resultant code be `cos(c)` over `cosf(c)` when `c` is an integer type?

mloubout: Well `cos` is technically not defined for integers so it will convert to float or double depending on the function used so I guess the test is broken in itself and should use a function that is valid for integers

devito/symbolics/printer.py
JDBetteridge: I disagree a little bit here, the method is called `single_prec` not `is_float`. If an expression only containing integers is passed, it is unclear that the result should return anything other than the default, as determined here by `self.dtype`.

I could see an argument for following the sympy behaviour of returning `None` if the precision cannot be determined, but this won't change much as `bool(None) == bool(False) == False`.

In the bigger picture, there are a lot of issues coinciding here, it seems that `sympy_dtype` isn't[ quite fit for purpose](https://github.com/devitocodes/devito/pull/2513/files#r1906778054) and the printer is doing some rather rough type inference. Personally, I think home-grown type inference should be avoided here (at least for the purposes of this PR), but I am open to further suggestions.

mloubout: > the method is called single_prec not is_float

Well, sure, that's the name we used when we made it, but this is effectively an `is_float`. This merely checks if it needs the float literal. 

> If an expression only containing integers is passed  it is unclear that the result should return anything other than the default, as determined here by self.dtype

self.dtype is only a last resort default for expression that cannot have a type determined, the type of the expression if determine is always the one that should be used irrespective of what that type is.

mloubout:
@JDBetteridge is this still needed?

JDBetteridge:
Potentially not. I thing the floating point issue has been fixed upstream, I will check the integer argument issue and close

FabioLuporini:
any updates on this one?

JDBetteridge:
I think it can just be closed, `sympy_dtype` still returns the wrong thing in some cases (integer type - 0.5, for instance), but the generated code that was previously incorrect now generates the correct thing on main.

PR #2516, opened 2025-01-14
compiler: Misc improvements to code generation

FabioLuporini:
In essence, this is a batch of tweaks to support GPU features in PRO

### DIFF
--- a/devito/arch/archinfo.py
+++ b/devito/arch/archinfo.py
@@ -31,6 +31,8 @@ __all__ = ['platform_registry', 'get_cpu_info', 'get_gpu_info', 'get_nvidia_cc',
            'POWER8', 'POWER9',
            # Generic GPUs
            'AMDGPUX', 'NVIDIAX', 'INTELGPUX',
+           # Nvidia GPUs
+           'AMPERE',
            # Intel GPUs
            'PVC', 'INTELGPUMAX', 'MAX1100', 'MAX1550']
 
@@ -916,6 +918,38 @@ class NvidiaDevice(Device):
             return False
 
 
+class Volta(NvidiaDevice):
+    pass
+
+
+class Ampere(Volta):
+
+    def supports(self, query, language=None):
+        if language != 'cuda':
+            return False
+
+        if query == 'async-loads':
+            return True
+
+        return super().supports(query, language)
+
+
+class Hopper(Ampere):
+
+    def supports(self, query, language=None):
+        if language != 'cuda':
+            return False
+
+        if query == 'tma':
+            return True
+
+        return super().supports(query, language)
+
+
+class Blackwell(Hopper):
+    pass
+
+
 class AmdDevice(Device):
 
     max_mem_trans_nbytes = 256
@@ -983,6 +1017,10 @@ POWER9 = Power('power9')
 ANYGPU = Cpu64('gpu')
 
 NVIDIAX = NvidiaDevice('nvidiaX')
+VOLTA = Volta('volta')
+AMPERE = Ampere('ampere')
+HOPPER = Hopper('hopper')
+BLACKWELL = Blackwell('blackwell')
 
 AMDGPUX = AmdDevice('amdgpuX')
 
--- a/devito/arch/archinfo.py
+++ b/devito/arch/archinfo.py
@@ -867,6 +867,12 @@ class Device(Platform):
             'max-block-dims': 3,
         }
 
+    def supports(self, query, language=None):
+        """
+        Check if the device supports a given feature.
+        """
+        return False
+
 
 class IntelDevice(Device):
 
@@ -894,6 +900,20 @@ class NvidiaDevice(Device):
             if 'tesla' in architecture.lower():
                 return 'tesla'
         return None
+
+    def supports(self, query, language=None):
+        if language != 'cuda':
+            return False
+
+        cc = get_nvidia_cc()
+        if query == 'async-loads' and cc >= 80:
+            # Asynchronous pipeline loads -- introduced in Ampere
+            return True
+        elif query == 'tma' and cc >= 90:
+            # Tensor Memory Acceleratory -- introduced in Hopper
+            return True
+        else:
+            return False
 
 
 class AmdDevice(Device):
--- a/devito/arch/archinfo.py
+++ b/devito/arch/archinfo.py
@@ -32,7 +32,7 @@ __all__ = ['platform_registry', 'get_cpu_info', 'get_gpu_info', 'get_nvidia_cc',
            # Generic GPUs
            'AMDGPUX', 'NVIDIAX', 'INTELGPUX',
            # Nvidia GPUs
-           'AMPERE',
+           'VOLTA', 'AMPERE', 'HOPPER', 'BLACKWELL',
            # Intel GPUs
            'PVC', 'INTELGPUMAX', 'MAX1100', 'MAX1550']
 
@@ -912,7 +912,7 @@ class NvidiaDevice(Device):
             # Asynchronous pipeline loads -- introduced in Ampere
             return True
         elif query == 'tma' and cc >= 90:
-            # Tensor Memory Acceleratory -- introduced in Hopper
+            # Tensor Memory Accelerator -- introduced in Hopper
             return True
         else:
             return False
--- a/devito/arch/compiler.py
+++ b/devito/arch/compiler.py
@@ -13,8 +13,8 @@ from codepy.jit import compile_from_string
 from codepy.toolchain import (GCCToolchain,
                               call_capture_output as _call_capture_output)
 
-from devito.arch import (AMDGPUX, Cpu64, AppleArm, NVIDIAX, POWER8, POWER9, Graviton,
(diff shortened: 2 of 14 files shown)


devito/ir/clusters/cluster.py
```diff
@@ -540,19 +544,3 @@ def reduce_properties(clusters):
             properties[d] = normalize_properties(properties.get(d, v), v)
 
     return Properties(properties)
-
-
-def tailor_properties(properties, ispace):
```
FabioLuporini, line 545: note for reviewers: finally moved to ir/support/properties as promised in an older PR

devito/passes/clusters/misc.py
```diff
@@ -166,79 +166,76 @@ class Key(tuple):
...
+        A Key contains elements that can logically be split into two groups -- the
+        `strict` and the `weak` components of the Key. Two Clusters (ClusterGroups)
+        having same `strict` but different `weak` parts are, by definition, not
+        fusible; however, since at least their `strict` parts match, they can at
+        least be topologically reordered.
         """
 
-        def __new__(cls, strict, weak):
+        def __new__(cls, itintervals, guards, syncs, weak):
+            strict = [itintervals, guards, syncs]
```
mloubout, line 177: can probably just make i a property returning this three since the object holds those three

devito/passes/clusters/misc.py
```diff
@@ -304,21 +296,35 @@ def choose_element(queue, scheduled):
...
+            if k.guards:
+                compatible = [i for i in candidates if m[i].guards == k.guards]
+                candidates = compatible or candidates
+
+            if k.syncs:
+                compatible = [i for i in candidates if m[i].syncs == k.syncs]
+                candidates = compatible or candidates
+
+            # Process the `weak` part of the key
+            for i in reversed(range(len(k.weak) + 1)):
```
mloubout, line 314: `range(len(k.weak), -1, -1)`

devito/passes/clusters/misc.py
```diff
@@ -304,21 +296,35 @@ def choose_element(queue, scheduled):
...
+                choosable = [e for e in candidates if m[e].weak[:i] == k.weak[:i]]
```
georgebisbas, line 315: I dont like this var name but ok

devito/types/basic.py
```diff
@@ -1511,6 +1516,12 @@ def __new__(cls, label, shape, function=None):
 
     func = Pickable._rebuild
 
+    @sympy.cacheit
```
mloubout, line 1519: why not memoized instead?

devito/arch/archinfo.py
```diff
@@ -31,6 +31,8 @@
            'POWER8', 'POWER9',
            # Generic GPUs
            'AMDGPUX', 'NVIDIAX', 'INTELGPUX',
+           # Nvidia GPUs
+           'AMPERE',
```
mloubout, line 35: Why not Volta/Hopper/Blackwell too?

mloubout (approved):
Some minor comments but looks straightforward to me.

devito/arch/archinfo.py
FabioLuporini: unnecessary. I only need to import AMPERE in PRO for testing

devito/arch/archinfo.py
```diff
@@ -895,6 +903,52 @@ def march(self):
...
+
+        if query == 'async-loads':
+            return True
+
+        return super().supports(query, language)
+
+
+class Hopper(Ampere):
+
+    def supports(self, query, language=None):
```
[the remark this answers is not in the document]
FabioLuporini: no it cannot be a property and these things are architecture specific -- I don't see significant code duplication

devito/arch/compiler.py
```diff
@@ -487,7 +487,7 @@ def __init_finalize__(self, **kwargs):
         language = kwargs.pop('language', configuration['language'])
         platform = kwargs.pop('platform', configuration['platform'])
 
-        if platform is NVIDIAX:
+        if isinstance(platform, NvidiaDevice):
```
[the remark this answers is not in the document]
FabioLuporini: yeah

devito/passes/clusters/misc.py
FabioLuporini: we could, but it's a nested class and I was trying to keep it as small as possible

tests/test_mpi.py
```diff
@@ -21,7 +21,10 @@
 from devito.tools import Bunch
 
 from examples.seismic.acoustic import acoustic_setup
-from tests.test_dse import TestTTI
+try:
+    from tests.test_dse import TestTTI
```
mloubout, line 25: why not just `from .test_dse import TestTTI`. If this file is run, test_dse exists there should not be any case where this cannot be imported here

devito/passes/clusters/misc.py
FabioLuporini: done

devito/types/basic.py
FabioLuporini: mimics 100% what's in sympy

FabioLuporini: can change if you prefer !

devito/passes/clusters/misc.py
FabioLuporini: the choosable among the candidates -- I think it's OK! why don't you like it? Im open to alternatives!

georgebisbas: hm, not sure....maybe like `filtered`, `valid`, `eligible` ?

georgebisbas (approved):

tests/test_mpi.py
mloubout: Actually just `from test_dse import TestTTI` see e.g.

https://github.com/devitocodes/devito/blob/da2c9a4bc6edd795d7cb046e2ac26de2339e0d8b/tests/test_autotuner.py#L153

devito/arch/archinfo.py
```diff
@@ -895,6 +903,52 @@ def march(self):
                 return 'tesla'
         return None
 
+    def supports(self, query, language=None):
+        if language != 'cuda':
+            return False
+
+        cc = get_nvidia_cc()
+        if query == 'async-loads' and cc >= 80:
```
EdCaunt, line 911: Why are the checks for CUDA compiler version not in the `Ampere` and `Hopper` classes?

Suggestion: why not something along the lines of
```python
class NvidiaDevice(Device):
    _supports = ()
    _mincuda = None

    def supports(self, query, language=None):
        if language != 'cuda':
            return False

        cc = get_nvidia_cc()
        return query in _supports and (_mincuda is None or cc >= _mincuda)
 
class Volta(NvidiaDevice):
    pass

class Ampere(Volta):
    _supports = super()._supports + ('async-loads',)
    _mincuda = 80

class Hopper(Ampere):
    _supports = super()._supports + ('tma',)
    _mincuda = 90

class Blackwell(Hopper):
    pass
```
Whilst this could probably be made more elegant, it does reduce the number of lines of code and reduce code duplication.

Maybe this idea is too unsophisticated though, since one could conceivably have high-end hardware but outdated compilers

devito/finite_differences/differentiable.py
```diff
@@ -730,6 +730,12 @@ def __init_finalize__(self, *args, **kwargs):
 
         super().__init_finalize__(*args, **kwargs)
 
+    @classmethod
+    def class_key(cls):
+        # Ensure Weights appear before any other AbstractFunction
```
EdCaunt, line 735: Why is this needed? Genuinely curious

devito/arch/archinfo.py
FabioLuporini: you cannot use super() on class-level modules (attributes & methods)

and yes, you may have older / newer CUDA versions on different architectures

I suggest we refrain from over-engineering this until we'll truly need to do so (if ever, hopefully never) but I will consider push backs

devito/finite_differences/differentiable.py
FabioLuporini: sympy voodoo, sigh

to avoid that e.g. CUDA unvectorized code looks like `a[i]*b[i]` and like `ab[i].y*ab[i].x` once vectorized (ie, with operands flipped). In short, for consistency of the order of arithmetic operations

devito/arch/archinfo.py
```diff
@@ -895,6 +903,52 @@ def march(self):
...
+            # Asynchronous pipeline loads -- introduced in Ampere
+            return True
+        elif query == 'tma' and cc >= 90:
+            # Tensor Memory Acceleratory -- introduced in Hopper
```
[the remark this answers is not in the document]
FabioLuporini: I think I'll postpone it until the next PR unless I'm really convinced a further push is required by this one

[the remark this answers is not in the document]
FabioLuporini: never mind. I had to push. Fixed

georgebisbas (approved):

EdCaunt (approved):

mloubout (approved):

PR #2517, opened 2025-01-14
compiler: fix scalar aliases from conditionals

mloubout:


devito/passes/clusters/aliases.py
```diff
@@ -261,11 +261,14 @@ def callback(self, clusters, prefix, xtracted=None):
...
 
             if made:
+                idx = processed.index(g[0])
+
                 for n, c in enumerate(g, -len(g)):
                     processed[processed.index(c)] = made.pop(n)
-                processed = made + processed
 
                 xtracted.extend(made)
+                while made:
```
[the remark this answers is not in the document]
mloubout: `processed[n:n] = made` is not a valid synthax. You need to insert all the elements of `made` before the cluster that got processed

devito/passes/clusters/misc.py
```diff
@@ -95,7 +95,11 @@ def callback(self, clusters, prefix):
 
             properties = c.properties.filter(key)
 
-            lifted.append(c.rebuild(ispace=ispace, properties=properties))
+            # Lifted scalar clusters cannot be guarded
+            # as they would not be in the scope of the guarded clusters
+            guards = {} if c.guards and c.is_scalar else c.guards
```
FabioLuporini, line 100: explicit if's as by convention, also, you can drop the first part "if c.guards"

so:

```
if c.is_scalar:
    guards = {}
else:
    guards = c.guards
```

mloubout: Done

FabioLuporini (approved):
GTG

PR #2520, opened 2025-01-20
api: Misc fixes for builtins and harmonic averaging

### LINKED ISSUE #1374: Temporary increase in memory when executed pyrevolve.Operator leads to memory error
A python script running a TTI RTM for only one shot with pyrevolve (https://sesibahia-my.sharepoint.com/:u:/g/personal/oscar_ladino_fieb_org_br/EWpX_VT4U3lCqdAArLVaGKABe9oysSb0KKRDlqIyL1XpwA?e=TgKdL4) runs fine for a period of time before crashing with the following error:
```
  File "/home/oscarm/.conda/envs/devito-v4.2.2/lib/python3.8/site-packages/pytools/prefork.py", line 49, in call_capture_output
    popen = Popen(cmdline, cwd=cwd, stdin=PIPE, stdout=PIPE,
  File "/home/oscarm/.conda/envs/devito-v4.2.2/lib/python3.8/subprocess.py", line 854, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/home/oscarm/.conda/envs/devito-v4.2.2/lib/python3.8/subprocess.py", line 1637, in _execute_child
    self.pid = _posixsubprocess.fork_exec(
OSError: [Errno 12] Cannot allocate memory
```
It seems to be that forking to call the compiler when applying the pyrevolve.Operator temporarily doubles the memory of the parent process. If the parent process is using more than half of the system memory, it will lead to run into a memory error. A workaround is to precompile the code before instantiate the CheckpointOperator object, i.e  
```
cp = DevitoCheckpoint([u,v])
op_fwd = solver.op_fwd(save=False)
op_fwd.cfunction
op_imaging.cfunction
wrap_fw = CheckpointOperator(op_fwd, src=geometry.src, u=u, v=v, ... )
wrap_rev = CheckpointOperator(op_imaging, u=u, v=v, ... )
```
However, it would appear that there is another possible solution as suggested in this thread https://devitocodes.slack.com/archives/C7JMLMSG0/p1593739206410500?thread_ts=1593727821.408500&cid=C7JMLMSG0

mloubout:
Fixes #1374

### DIFF
--- a/devito/finite_differences/differentiable.py
+++ b/devito/finite_differences/differentiable.py
@@ -626,15 +626,16 @@ class Mod(DifferentiableOp, sympy.Mod):
 class SafeInv(Differentiable, sympy.core.function.Application):
     _fd_priority = 0
 
-    def __new__(cls, val, base, **kwargs):
-        return super().__new__(cls, val, base)
-
     @property
     def base(self):
         return self.args[1]
 
+    @property
+    def val(self):
+        return self.args[0]
+
     def __str__(self):
-        return f'1/({self.args[0]})'
+        return Pow(self.args[0], -1).__str__()
 
     __repr__ = __str__
 
--- a/devito/finite_differences/differentiable.py
+++ b/devito/finite_differences/differentiable.py
@@ -626,8 +626,8 @@ class Mod(DifferentiableOp, sympy.Mod):
 class SafeInv(Differentiable, sympy.core.function.Application):
     _fd_priority = 0
 
-    def __new__(cls, *args, **kwargs):
-        newobj = super().__new__(cls, *args)
+    def __new__(cls, val, base, **kwargs):
+        newobj = super().__new__(cls, val, base)
         return newobj
 
     def _evaluate(self, **kwargs):
@@ -635,6 +635,10 @@ class SafeInv(Differentiable, sympy.core.function.Application):
 
     def _subs(self, old, new, **hints):
         return self.func(*[i._subs(old, new, **hints) for i in self.args])
+
+    @property
+    def base(self):
+        return self.args[1]
 
 
 class IndexSum(sympy.Expr, Evaluable):
--- a/devito/finite_differences/differentiable.py
+++ b/devito/finite_differences/differentiable.py
@@ -627,18 +627,16 @@ class SafeInv(Differentiable, sympy.core.function.Application):
     _fd_priority = 0
 
     def __new__(cls, val, base, **kwargs):
-        newobj = super().__new__(cls, val, base)
-        return newobj
-
-    def _evaluate(self, **kwargs):
-        return self.func(*[i._evaluated(**kwargs) for i in self.args])
-
-    def _subs(self, old, new, **hints):
-        return self.func(*[i._subs(old, new, **hints) for i in self.args])
+        return super().__new__(cls, val, base)
 
     @property
     def base(self):
         return self.args[1]
+
+    def __str__(self):
+        return f'1/({self.args[0]})'
+
+    __repr__ = __str__
 
 
 class IndexSum(sympy.Expr, Evaluable):
@@ -692,6 +690,8 @@ class IndexSum(sympy.Expr, Evaluable):
 
     def _sympystr(self, printer):
         return str(self)
+
+    _latex = _sympystr
 
     def _hashable_content(self):
         return super()._hashable_content() + (self.dimensions,)
--- a/devito/finite_differences/differentiable.py
+++ b/devito/finite_differences/differentiable.py
@@ -623,6 +623,20 @@ class Mod(DifferentiableOp, sympy.Mod):
     __sympy_class__ = sympy.Mod
 
 
+class SafeInv(Differentiable, sympy.core.function.Application):
+    _fd_priority = 0
+
+    def __new__(cls, *args, **kwargs):
+        newobj = super().__new__(cls, *args)
+        return newobj
+
+    def _evaluate(self, **kwargs):
+        return self.func(*[i._evaluated(**kwargs) for i in self.args])
+
+    def _subs(self, old, new, **hints):
+        return self.func(*[i._subs(old, new, **hints) for i in self.args])
+
+
 class IndexSum(sympy.Expr, Evaluable):
 
     """
--- a/devito/operator/operator.py
+++ b/devito/operator/operator.py
@@ -894,6 +894,13 @@ class Operator(Callable):
         >>> op = Operator(Eq(u3.forward, u3 + 1))
         >>> summary = op.apply(time_M=10)
         """
+        # Compile the operator before building the arguments list
+        # to avoid out of memory with greedy compilers
+        try:
+            cfunction = self.cfunction
+        except Exception as e:
+            raise CompilationError(f"Failed to compile the Operator {self.name}") from e
+
         # Build the arguments list to invoke the kernel function
         with self._profiler.timer_on('arguments'):
             args = self.arguments(**kwargs)
@@ -901,7 +908,6 @@ class Operator(Callable):
         # Invoke kernel function with args
         arg_values = [args[p.name] for p in self.parameters]
         try:
-            cfunction = self.cfunction
             with self._profiler.timer_on('apply', comm=args.comm):
                 retval = cfunction(*arg_values)
         except ctypes.ArgumentError as e:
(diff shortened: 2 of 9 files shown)


devito/finite_differences/differentiable.py
```diff
@@ -623,6 +623,24 @@ class Mod(DifferentiableOp, sympy.Mod):
     __sympy_class__ = sympy.Mod
 
 
+class SafeInv(Differentiable, sympy.core.function.Application):
+    _fd_priority = 0
+
+    def __new__(cls, val, base, **kwargs):
+        newobj = super().__new__(cls, val, base)
```
FabioLuporini, line 630: no need for `newobj` for such small one liner methods

devito/finite_differences/differentiable.py
```diff
@@ -623,6 +623,24 @@ class Mod(DifferentiableOp, sympy.Mod):
...
+        return newobj
+
+    def _evaluate(self, **kwargs):
```
FabioLuporini, line 633: why is this override needed? it should automatically pick up

https://github.com/devitocodes/devito/blob/9b8a144d8418102cc48e0e09fd0be2d53a34bcfe/devito/types/lazy.py#L45

devito/finite_differences/differentiable.py
```diff
@@ -623,6 +623,24 @@ class Mod(DifferentiableOp, sympy.Mod):
...
+        return self.func(*[i._evaluated(**kwargs) for i in self.args])
+
+    def _subs(self, old, new, **hints):
```
FabioLuporini, line 636: similar to above -- why is this `_subs` override needed?

devito/passes/iet/misc.py
```diff
@@ -225,6 +226,11 @@ def _(expr):
         return ()
 
 
+@_lower_macro_math.register(SafeInv)
+def _(expr):
+    return (('SAFEINV(a, b)', '(((a) < 1e-12 || (b) < 1e-12) ? (0.0F) : (1.0F / (a)))'),)
```
FabioLuporini, line 231: the 1e-12 bounds should ideally be pulled from the SafeInv's attributes (default to 1e-12, but allow caller/user to pass a different threshold?)

devito/types/basic.py
```diff
@@ -1167,12 +1167,16 @@ def _evaluate(self, **kwargs):
...
-            retval = 1 / retval
 
         # Evaluate. Since we used `self.function` it will be on the grid when evaluate
         # is called again within FD
-        return retval.evaluate.expand()
+        if self._avg_mode == 'harmonic':
+            from devito.finite_differences.differentiable import SafeInv
+            retval = SafeInv(retval.evaluate, base=self.function)
+        else:
+            retval = retval.evaluate
```
FabioLuporini, line 1177: if there was an expand() before, I'd be surprised no tests failed now

devito/types/basic.py
```diff
@@ -1450,12 +1454,19 @@ def indexify(self, indices=None, subs=None):
         # Indices after substitutions
         indices = []
         for a, d, o, s in zip(self.args, self.dimensions, self.origin, subs):
-            if d in a.free_symbols:
+            if a.is_Function and len(a.args) == 1:
+                # E.g. Abs(expr)
```
FabioLuporini, line 1458: it has to be an integer function, so maybe you can check for Abs; if not Abs but a Function, return an error?

I don't know, maybe nitpicking, feel free to ignore...

mloubout: Could be floor/ceil/INT/...., so lots of options, this just check it's unary Op, multi-arg op are too weird here so maybe should raise an error

devito/types/basic.py
mloubout: the `expand` was only there for the harmonic case to avoid nested `1/ (1/ (1/a + 1/ b)))` and such, now that it goes through it proper safe div it's not needed anymore

devito/passes/iet/misc.py
mloubout: it's a safe threshold. But this is also only needed for single precision, double precision should probably default to standard div (not that we ever run double prec)

devito/finite_differences/differentiable.py
mloubout: I honestly don't rember was having issues without it I'll double check, same below

devito/types/basic.py
```diff
@@ -1167,12 +1167,16 @@ def _evaluate(self, **kwargs):
...
         for d, i in self._grid_map.items():
             retval = retval.diff(d, deriv_order=0, fd_order=2, x0={d: i})
-        if self._avg_mode == 'harmonic':
-            retval = 1 / retval
 
         # Evaluate. Since we used `self.function` it will be on the grid when evaluate
         # is called again within FD
-        return retval.evaluate.expand()
+        if self._avg_mode == 'harmonic':
+            from devito.finite_differences.differentiable import SafeInv
```
EdCaunt, line 1174: Does this avoid a circular import?

mloubout: yes

devito/operator/operator.py
```diff
@@ -894,14 +894,20 @@ def apply(self, **kwargs):
         >>> op = Operator(Eq(u3.forward, u3 + 1))
         >>> summary = op.apply(time_M=10)
         """
+        # Compile the operator before building the arguments list
+        # to avoid out of memory with greedy compilers
```
FabioLuporini, line 898: is this related to

https://github.com/devitocodes/devito/issues/1374 

?

devito/passes/iet/misc.py
FabioLuporini: yeah, it's just that like that it's a magic number buried deep down inside a compiler pass, so it should rather be pulled from `expr` itself (can do as a class attribute)

devito/finite_differences/differentiable.py
```diff
@@ -623,6 +623,23 @@ class Mod(DifferentiableOp, sympy.Mod):
     __sympy_class__ = sympy.Mod
 
 
+class SafeInv(Differentiable, sympy.core.function.Application):
+    _fd_priority = 0
+
+    def __new__(cls, val, base, **kwargs):
```
FabioLuporini, line 629: is this override just to tolerate potential extra `**kwargs` (evaluate?) ?

devito/operator/operator.py
mloubout: Probably yes, tagging it

devito/symbolics/inspection.py
```diff
@@ -117,7 +117,8 @@ def estimate_cost(exprs, estimate=False):
     'div': 5,
     'Abs': 5,
     'floor': 1,
-    'ceil': 1
+    'ceil': 1,
+    'SafeInv': 10,
```
georgebisbas, line 121: should you put it futher up, so they remain in descending order?

devito/passes/iet/misc.py
mloubout: done

devito/operator/operator.py
```diff
@@ -894,14 +894,20 @@ def apply(self, **kwargs):
...
+        try:
+            cfunction = self.cfunction
+        except Exception as e:
```
FabioLuporini, line 901: drop try-except; `compiler.py`'s already throws an exception if compilation fails, and that's what the user should see

devito/finite_differences/differentiable.py
```diff
@@ -623,6 +623,23 @@ class Mod(DifferentiableOp, sympy.Mod):
     __sympy_class__ = sympy.Mod
 
 
+class SafeInv(Differentiable, sympy.core.function.Application):
```
FabioLuporini, line 626: I'm now thinking that this is generated at `_evaluate()`, hence it doesn't (or better maybe, _shouldn't_) need to be a `Differentiable`. It might just have to inherit from sympy.Application and, as such, be moved into `symbolics/extended_sympy.py`

mloubout: No, this is part of `_eval_at` not `_evaluate` so the result needs to be evaluable and differentiable as it could be inside a derivative  as well.

devito/operator/operator.py
mloubout: done

FabioLuporini (approved):
GTG

PR #2521, opened 2025-01-20
compiler: Blocking dims for subdims now use name of root

EdCaunt:
Draft for now. Pushing to spin CI.

Nominally means that introducing a MultiSubDomain won't result in different block names

### DIFF
--- a/devito/passes/clusters/blocking.py
+++ b/devito/passes/clusters/blocking.py
@@ -354,7 +354,7 @@ class SynthesizeBlocking(Queue):
             except KeyError:
                 pass
 
-        base = self.sregistry.make_name(prefix=d.name)
+        base = self.sregistry.make_name(prefix=d.root.name)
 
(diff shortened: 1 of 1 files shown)


devito/passes/clusters/blocking.py
```diff
@@ -366,7 +366,7 @@ def _derive_block_dims(self, clusters, prefix, d, blk_size_gen):
             bd = BlockDimension(name, bd, bd, bd + bd.step - 1, size=step)
             block_dims.append(bd)
 
-        bd = BlockDimension(d.name, bd, bd, bd + bd.step - 1, 1, size=step)
+        bd = BlockDimension(d.root.name, bd, bd, bd + bd.step - 1, 1, size=step)
```
mloubout, line 369: Shouldn't this be done in the abridge pass

georgebisbas (commented):
Tests failing for the names

PR #2522, opened 2025-01-21
api: fix backward compatibility of Substitution, missing function check

mloubout:
Fix backward compat that was applying weights to all derivative missing the `coeff.function` check

### DIFF
--- a/devito/finite_differences/derivative.py
+++ b/devito/finite_differences/derivative.py
@@ -307,7 +307,8 @@ class Derivative(sympy.Derivative, Differentiable, Pickable):
         if self in subs:
             new = subs.pop(self)
             try:
-                return new._xreplace(subs)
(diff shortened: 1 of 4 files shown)


FabioLuporini (approved):
GTG

devito/finite_differences/derivative.py
```diff
@@ -307,7 +307,8 @@ def _xreplace(self, subs):
         if self in subs:
             new = subs.pop(self)
             try:
-                return new._xreplace(subs)
+                new, flag = new._xreplace(subs)
+                return new, True
```
mloubout, line 311: Note: this was actually incorrectly returning `False` flag in lots of cases (since subs is then empy here) so `new._xreplace(subs)` would return `new, False`

PR #2524, opened 2025-01-23
api: Fix custom coefficients inlining

FabioLuporini:
this is only an issue for the legacy custom coefficients API

### DIFF
--- a/devito/types/equation.py
+++ b/devito/types/equation.py
@@ -98,7 +98,7 @@ class Eq(sympy.Eq, Evaluable, Pickable):
         if not mapper:
             return expr
 
-        return expr.xreplace(mapper)
+        return expr.subs(mapper)
 
     def _evaluate(self, **kwargs):
         """
--- a/examples/seismic/tutorials/07_DRP_schemes.ipynb
+++ b/examples/seismic/tutorials/07_DRP_schemes.ipynb
@@ -83,7 +83,7 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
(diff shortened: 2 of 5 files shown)


mloubout (approved):

devito/types/equation.py
```diff
@@ -98,7 +98,7 @@ def _apply_coeffs(cls, expr, coefficients):
         if not mapper:
             return expr
 
-        return expr.xreplace(mapper)
+        return expr.subs(mapper)
```
mloubout, line 101: @FabioLuporini this was the issue. If you had something like `Eq(u, u.dx2 + 2 * u.laplace, substitutions=...)` then `xreplace` only replaces the firs appearance so it was using custom coeffs for some and standard coeffs for other. Code looks much better (less. weights and derivatives) with the fix.

examples/seismic/tutorials/07_DRP_schemes.ipynb
```diff
@@ -83,7 +83,7 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "Eq(-u(t, x, y)/dt + u(t + dt, x, y)/dt + (0.1*u(t, x, y) - 0.6*u(t, x - h_x, y) + 0.6*u(t, x + h_x, y))/h_x, 0)\n"
+    "Eq(-u(t, x, y)/dt + u(t + dt, x, y)/dt + 0.1*u(t, x, y)/h_x - 0.6*u(t, x - h_x, y)/h_x + 0.6*u(t, x + h_x, y)/h_x, 0)\n"
```
georgebisbas, line 86: so we lose the flop reduction here?

mloubout: This is only the symbolic equation, not the operator.

PR #2531, opened 2025-02-06
compiler: Support compiler flags for Cortex

georgebisbas:
compiler: Support compiler flags for Cortex on Raspberry pi

### DIFF
--- a/devito/arch/compiler.py
+++ b/devito/arch/compiler.py
@@ -466,9 +466,7 @@ class GNUCompiler(Compiler):
 
 
 class ArmCompiler(GNUCompiler):
-
-    def __init_finalize__(self, **kwargs):
-        GNUCompiler.__init_finalize__(self, **kwargs)
+    pass
 
 
 class ClangCompiler(Compiler):


devito/arch/compiler.py
```diff
@@ -470,9 +470,11 @@ def __init_finalize__(self, **kwargs):
         GNUCompiler.__init_finalize__(self, **kwargs)
         platform = kwargs.pop('platform', configuration['platform'])
 
-        # Graviton flag
```
georgebisbas, line 473: This was redundant, duplicating `mcpu` flag

mloubout: would probably remove the GNU one instead then. Don't remember if can end up in the GNUCompiler but not in the Arm one for graviton

georgebisbas: This way you are safe for both DEVITO_ARCH={gcc, arm} and PLATFORM=graviton

georgebisbas: right, checked with all possible env combos for compiler{GCC, arm}, and platform={graviton, cortexa76} seems to that all flags are ok

mloubout (approved):

devito/arch/compiler.py
```diff
@@ -443,6 +443,9 @@ def __init_finalize__(self, **kwargs):
             self.cflags = ['-mcpu=native'] + self.cflags
         elif isinstance(platform, Graviton):
             self.cflags = ['-mcpu=%s' % platform.march] + self.cflags
+        elif isinstance(platform, Cortex):
+            self.cflags += ['-march=%s' % platform.march]
```
EdCaunt, line 447: Nitpick: could be fstring

georgebisbas: So I guess for homogeneity I should change the whole file? If this is a transition devito codebase is doing I am happy to do it.

devito/arch/compiler.py
```diff
@@ -468,11 +469,6 @@ class ArmCompiler(GNUCompiler):
 
     def __init_finalize__(self, **kwargs):
```
FabioLuporini, line 470: this constructor can go

georgebisbas: The whole class is redundant now...but keeping for compatibility I guess.

mloubout (approved):

PR #2532, opened 2025-02-07
docker: add option to install gcc 14

mloubout:


.github/workflows/docker-bases.yml
```diff
@@ -17,7 +17,7 @@ on:
         description: 'Build compiler bases'
   schedule:
     # Run once a week
-    - cron: "0 13 * * 1"
+    - cron: "0 0 1 * *"
```
mloubout, line 20: Note: Once a month should be enough overkill to do once a week.

FabioLuporini: Then you should update the comment above

docker/Dockerfile.amd
```diff
@@ -35,6 +35,7 @@ RUN rm /etc/apt/sources.list.d/* \
     gfortran \
     flex \
     hwloc \
+    gdb \
```
FabioLuporini, line 38: Why not in the base dockerfile?

and, random thought, could we also add 

* python3-dbg (for gdb python support)
* valgrind

they should be pretty small, but very helpful.

Thoughts?

docker/Dockerfile.cpu
```diff
@@ -4,60 +4,78 @@
...
 
-ENV DEBIAN_FRONTEND noninteractive
+# Add repo for other python versions
+RUN apt-get update && \
+    apt-get install -y software-properties-common
 
 # Install python
 RUN apt-get update && \
-    apt-get install -y dh-autoreconf python3-venv python3-dev python3-pip
+    apt-get install -y dh-autoreconf python3-venv python3-dev python3-pip gdb
```
FabioLuporini, line 20: as written above, maybe we do this in Dockerfile.devito instead?

docker/Dockerfile.nvidia
```diff
@@ -16,7 +16,7 @@ ENV DEBIAN_FRONTEND noninteractive
 # Install python
 RUN apt-get update && \
     apt-get install -y -q gpg apt-utils curl wget vim libnuma-dev tmux numactl \
-                          dh-autoreconf python3-venv python3-dev python3-pip
+                          dh-autoreconf python3-venv python3-dev python3-pip gdb
```
FabioLuporini, line 19: see comments above

docker/Dockerfile.cpu
```diff
@@ -4,60 +4,78 @@
...
 
 # Install for basic base not containing it
 RUN apt-get install -y vim wget git flex libnuma-dev tmux \
         numactl hwloc curl \
         autoconf libtool build-essential procps software-properties-common
 
+
+# Install compilersx
+RUN if [ -n "$gcc" ]; then \
+        apt-get install libgmp-dev libmpc-dev libmpfr-dev libisl-dev binutils texinfo -y && \
```
FabioLuporini, line 30: do we really have to compile it from source?

docker/Dockerfile.cpu
mloubout: No it's a pain, the `.devito` is only adding devito to the base, and installing things in it defeats the purpose of multistage and minimal final image

FabioLuporini (approved):

PR #2533, opened 2025-02-10
misc: Update advisor with oneAPI 2025

georgebisbas:
Some more changes are needed, but finally works


devito/arch/compiler.py
```diff
@@ -771,6 +771,20 @@ def __init_finalize__(self, **kwargs):
...
+            # add the necessary flags to the compiler
+            from devito.operator.profiling import locate_intel_advisor
+            path = locate_intel_advisor()
+            self.add_include_dirs(path.joinpath('include').as_posix())
+
+            _default_libs = ['ittnotify']
+            self.add_libraries(_default_libs)
+
+            libdir = path.joinpath('lib64').as_posix()
+            self.add_library_dirs(libdir)
```
mloubout, line 785: Just `add_lib_dirs(libdir, rpath=True)`

devito/arch/compiler.py
```diff
@@ -771,6 +771,20 @@ def __init_finalize__(self, **kwargs):
             self.__init_intel_mpi__()
             self.__init_intel_mpi_flags__()
 
+        if configuration['profiling'] == 'advisor':
+            # Locate the Intel Advisor installation and
+            # add the necessary flags to the compiler
+            from devito.operator.profiling import locate_intel_advisor
+            path = locate_intel_advisor()
```
mloubout, line 778: Why aren't all those set by the `setvars.sh` that's what it's supposed to be for

benchmarks/user/advisor/README.md
```diff
@@ -2,35 +2,50 @@ Example runs:
...
+depending on where you installed oneAPI
+
 * In Linux systems you may need to enable system-wide profiling by setting:
-  - `/proc/sys/kernel/yama/ptrace_scope` to `0`
-  - `/proc/sys/kernel/perf_event_paranoid` to `1`
 
-* `numactl` must be available on the system. If not available, install with:
-	`sudo apt-get install numactl`
+```sh
+/proc/sys/kernel/yama/ptrace_scope to 0
```
EdCaunt, line 28: Should these be the actual commands?

benchmarks/user/advisor/roofline.py
```diff
@@ -65,21 +64,25 @@
 def roofline(name, project, scale, precision, mode, th):
     pd.options.display.max_rows = 20
 
-    log('Opening project...')
+    log(f'Opening project {project}...')
     project = advisor.open_project(str(project))
 
     if not project:
-        err('Could not open project %s.' % project)
+        err(f'Could not open project {project}.')
```
EdCaunt, line 71: What does `open_project` return if it fails to open the project? This error message might return something cryptic since `project` gets overwritten

benchmarks/user/advisor/roofline.py
```diff
@@ -169,7 +172,7 @@ def roofline(name, project, scale, precision, mode, th):
             label_y = row.self_gflops
             ax.text(label_x, label_y,
                     'Time: %.2fs\n'
-                    'Incidence: %.0f%%' % (row.self_time, row.percent_weight),
+                    'Peak pct: %.0f%%' % (row.self_time, row.percent_weight),
```
EdCaunt, line 175: Worth converting these to fstring?

devito/arch/compiler.py
georgebisbas: ok

devito/arch/compiler.py
georgebisbas: Unfortunately, this was true with Intel studios, but now they are not by default. You have to link explicitly. At least, this is what I see.

georgebisbas: And this is actually why it stopped working at some point

benchmarks/user/advisor/README.md
georgebisbas: wdym? They may or may not be always needed

benchmarks/user/advisor/roofline.py
georgebisbas: done

examples/performance/02_advisor_roofline.ipynb
```diff
@@ -267,86 +272,6 @@
     "As you can see from this roofline graph, the main point is different from the single point of the previous graph. Moreover, each point is labelled with 'Time' and 'Incidence' indicators. These represent the total execution time of each loop's main body and their percentage incidence on the total execution time of the main time loop."
    ]
   },
-  {
```
georgebisbas, line 270: All this is redundant in my opinion to be treated as-a-worthy to mention API

benchmarks/user/advisor/README.md
JDBetteridge: As root
```bash
echo 0 > /proc/sys/kernel/yama/ptrace_scope
echo 1 > /proc/sys/kernel/perf_event_paranoid
```
As far as I know this is the only way to achieve this. Downside is you need to be root :confused:

georgebisbas: Correct @JDBetteridge.
So we have these command there to the discretization of the user.
They may not be needed. This needs for sure sudo rights. We leave it for the user IFF needed
e.g. for me, it works fine with
```
cat /proc/sys/kernel/yama/ptrace_scope
1
cat /proc/sys/kernel/perf_event_paranoid 
4
```

IIIRC it is more needed for vtune rather than advisor

devito/operator/profiling.py
```diff
@@ -367,19 +364,8 @@ class AdvisorProfiler(AdvancedProfiler):
...
-            # Make sure future compilations will get the proper header and
-            # shared object files
-            compiler = configuration['compiler']
-            compiler.add_include_dirs(self.path.joinpath('include').as_posix())
-            compiler.add_libraries(self._default_libs)
-            libdir = self.path.joinpath('lib64').as_posix()
-            compiler.add_library_dirs(libdir)
-            compiler.add_ldflags('-Wl,-rpath,%s' % libdir)
+        super(AdvisorProfiler, self).__init__(name)
+        return
```
FabioLuporini, line 368: no need

FabioLuporini: AFAICT, this is just yet another empty constructor? droppable?

devito/operator/profiling.py
```diff
@@ -367,19 +364,8 @@ class AdvisorProfiler(AdvancedProfiler):
     _ext_calls = [_api_resume, _api_pause]
 
     def __init__(self, name):
-        self.path = locate_intel_advisor()
-        if self.path is None:
-            self.initialized = False
-        else:
-            super().__init__(name)
```
FabioLuporini, line 374: I prefer to keep this logic here rather than in compiler.py, which shouldn't be aware of profiling

devito/arch/compiler.py
```diff
@@ -771,6 +771,18 @@ def __init_finalize__(self, **kwargs):
             self.__init_intel_mpi__()
             self.__init_intel_mpi_flags__()
 
+        if configuration['profiling'] == 'advisor':
```
FabioLuporini, line 774: see comment below, but basically since "profiling" requires third-party tools (as opposed to MPI which is an actual API) I have a preference for this bit of logic to be kept in profiling

conftest.py
```diff
@@ -79,6 +79,12 @@ def skipif(items, whole_module=False):
            isinstance(configuration['platform'], Cpu64):
             skipit = "`icx+cpu64` won't work with this test"
             break
+        # Skip if icx or advisor are not available
+        if i not in 'only-advisor' or \
```
FabioLuporini, line 83: this is wrong, it either has to be a tuple or something's off

devito/arch/archinfo.py
```diff
@@ -464,6 +465,27 @@ def get_cuda_path():
...
+
+    env_path = os.environ["PATH"]
+    env_path_dirs = env_path.split(":")
+
+    for env_path_dir in env_path_dirs:
+        # intel/oneapi/advisor is the directory for Intel oneAPI
+        if "intel/advisor" in env_path_dir or "intel/oneapi/advisor" in env_path_dir:
+            path = Path(env_path_dir)
+            if path.name.startswith('bin'):
+                path = path.parent
```
FabioLuporini, line 483: return path.parent

devito/operator/profiling.py
```diff
@@ -367,19 +368,12 @@ class AdvisorProfiler(AdvancedProfiler):
...
-            # Make sure future compilations will get the proper header and
-            # shared object files
-            compiler = configuration['compiler']
-            compiler.add_include_dirs(self.path.joinpath('include').as_posix())
-            compiler.add_libraries(self._default_libs)
-            libdir = self.path.joinpath('lib64').as_posix()
-            compiler.add_library_dirs(libdir)
-            compiler.add_ldflags('-Wl,-rpath,%s' % libdir)
+        super(AdvisorProfiler, self).__init__(name)
```
FabioLuporini, line 371: `super().__init__(name)`

devito/operator/operator.py
```diff
@@ -219,6 +219,12 @@ def _build(cls, expressions, **kwargs):
 
         # Required for the jit-compilation
         op._compiler = kwargs['compiler']
+
+        # Add any metadata from the profiler
```
FabioLuporini, line 223: nitpicking, I wouldn't call it "metadata"

devito/arch/archinfo.py
```diff
@@ -464,6 +465,27 @@ def get_cuda_path():
     return None
 
 
+def get_advisor_path():
```
FabioLuporini, line 468: memoize

tests/test_benchmark.py
```diff
@@ -87,3 +87,20 @@ def test_run_mpi(mode):
         'dump_norms': 'norms.txt'
     }
     run('acoustic', **kwargs)
+
+
+@skipif('only-advisor')
```
FabioLuporini, line 92: since it's called "skipif", instead of calling it "only-advisor" I'd use something else, for example "noadvisor" , which would also be similar to the "nodevice" that we already have in place

conftest.py
georgebisbas: yes

devito/arch/archinfo.py
georgebisbas: yes

devito/arch/archinfo.py
georgebisbas: yes

devito/operator/operator.py
georgebisbas: hope is better

tests/test_benchmark.py
georgebisbas: right!

benchmarks/user/advisor/roofline.py
georgebisbas: I was sure I had replied in this, not sure I cannot see my message. This is part of the advisor API, it should already be there 

https://www.intel.com/content/www/us/en/developer/articles/training/how-to-use-the-intel-advisor-python-api.html

conftest.py
```diff
@@ -80,7 +80,7 @@ def skipif(items, whole_module=False):
             skipit = "`icx+cpu64` won't work with this test"
             break
         # Skip if icx or advisor are not available
-        if i not in 'only-advisor' or \
+        if i not in ('noadvisor') or \
```
FabioLuporini, line 83: no this is wrong as it's not a tuple

devito/operator/profiling.py
```diff
@@ -368,12 +368,15 @@ class AdvisorProfiler(AdvancedProfiler):
...
-        self._include_dirs = [path.joinpath('include').as_posix()]
-
-        libdir = path.joinpath('lib64').as_posix()
-        self._lib_dirs = [libdir]
+        if path:
+            self._include_dirs = [path.joinpath('include').as_posix()]
+            libdir = path.joinpath('lib64').as_posix()
+            self._lib_dirs = [libdir]
+        else:
+            warning("Intel Advisor not found; reverting to `advanced`")
```
FabioLuporini, line 378: is it really reverting?

devito/operator/profiling.py
```diff
@@ -514,8 +510,9 @@ def create_profile(name):
     if profiler.initialized:
         return profiler
     else:
-        warning("Couldn't set up `%s` profiler; reverting to `advanced`" % level)
-        profiler = profiler_registry['basic'](name)
```
georgebisbas, line 518: Here, it was reverting to `basic`

devito/operator/profiling.py
```diff
@@ -367,19 +368,14 @@ class AdvisorProfiler(AdvancedProfiler):
...
     def __init__(self, name):
-        self.path = locate_intel_advisor()
-        if self.path is None:
-            self.initialized = False
+        super().__init__(name)
+        path = get_advisor_path()
+        if path:
+            self._include_dirs = [path.joinpath('include').as_posix()]
+            libdir = path.joinpath('lib64').as_posix()
+            self._lib_dirs = [libdir]
```
FabioLuporini, line 376: no, instances shouldn't change class attributes like that....

and btw why is it even a class attribute ?

devito/operator/profiling.py
FabioLuporini: ok but no need for that extra var imho (nitpicking)

FabioLuporini (changes requested):

devito/operator/profiling.py
```diff
@@ -52,8 +50,18 @@ def __init__(self, name):
         # Python-level timers
         self.py_timers = OrderedDict()
 
+        # Instance attributes
```
FabioLuporini, line 53: remove this comment

devito/operator/profiling.py
```diff
@@ -52,8 +50,18 @@ def __init__(self, name):
...
+        self._include_dirs = []
```
FabioLuporini, line 54: since these attributes (and the corresponding _add methods (please make them private)) are used only required by "third party profilers", you should move them into the Advisor class , or inside an intermediate class for ThirdPartyProfilers/ExternalProfilers 

because here they'll never ever be used by the standard Profiler classes

FabioLuporini: oh, I see why now...

because you need these attributes in operator.py

devito/operator/profiling.py
georgebisbas: I'd better change/improve I think this comment

devito/operator/profiling.py
georgebisbas: So, no changes here AFAIU

benchmarks/user/advisor/run_advisor.py
```diff
@@ -1,16 +1,14 @@
...
 from subprocess import check_output, PIPE, Popen
-import sys
 from tempfile import gettempdir, mkdtemp
 
-import click
-
-
-from benchmarks.user.advisor.advisor_logging import (check, log, progress,
-                                                     log_process)
+from advisor_logging import (check, log, progress, log_process)
```
FabioLuporini, line 11: nitpicking, no need for parentheses now

devito/operator/profiling.py
```diff
@@ -367,21 +376,23 @@ class AdvisorProfiler(AdvancedProfiler):
     _ext_calls = [_api_resume, _api_pause]
 
     def __init__(self, name):
-        self.path = locate_intel_advisor()
-        if self.path is None:
-            self.initialized = False
+        if self._attempted_init:
```
FabioLuporini, line 379: shouldn't this go _after_ the `super()` below?

devito/operator/profiling.py
```diff
@@ -367,21 +376,23 @@ class AdvisorProfiler(AdvancedProfiler):
...
-            super().__init__(name)
-            # Make sure future compilations will get the proper header and
-            # shared object files
-            compiler = configuration['compiler']
-            compiler.add_include_dirs(self.path.joinpath('include').as_posix())
-            compiler.add_libraries(self._default_libs)
-            libdir = self.path.joinpath('lib64').as_posix()
-            compiler.add_library_dirs(libdir)
-            compiler.add_ldflags('-Wl,-rpath,%s' % libdir)
+            self._attempted_init = False
```
FabioLuporini, line 390: useless

benchmarks/user/advisor/run_advisor.py
georgebisbas: Done

devito/operator/profiling.py
georgebisbas: No, otherwise, we would change the class attributes again! And then we would have the same problem as why it was not working before. include dirs and libdirs will not be included

devito/operator/profiling.py
georgebisbas: This is useful to be able to revert to `advanced` IFF `icx` is available but `advisor` not.
Check this line:
https://github.com/devitocodes/devito/pull/2533/files#diff-6ef42b440b70eb3098c97da2cf40ef16d030c6566dbdc3aeb2f87cf0313c057dR525

conftest.py
```diff
@@ -79,6 +79,12 @@ def skipif(items, whole_module=False):
            isinstance(configuration['platform'], Cpu64):
             skipit = "`icx+cpu64` won't work with this test"
             break
+        # Skip if icx or advisor are not available
+        if i not in ('noadvisor',) or \
```
FabioLuporini, line 83: shouldn't this rather be

```
if i == 'noadvisor' and (not isinstance(configuration['compiler'], IntelCompiler) or not get_advisor_path())
```

devito/operator/profiling.py
```diff
@@ -35,10 +33,15 @@ class Profiler:
     _default_libs = []
     _ext_calls = []
 
+    _include_dirs = []
+    _lib_dirs = []
```
FabioLuporini, line 37: what's the difference between this and the class attribute above `_default_libs` ?

devito/operator/profiling.py
```diff
@@ -52,7 +55,13 @@ def __init__(self, name):
         # Python-level timers
         self.py_timers = OrderedDict()
 
-        self.initialized = True
+        self._attempted_init = True
+
+    def add_include_dir(self, dir_path):
```
FabioLuporini, line 60: you don't need these public methods since the updated attributes (`_include_dirs` , `_lib_dirs`) are only updated internally from the class itself, not from the outside

devito/operator/profiling.py
```diff
@@ -367,21 +376,23 @@ class AdvisorProfiler(AdvancedProfiler):
...
-        if self.path is None:
-            self.initialized = False
+        if self._attempted_init:
+            return
+
+        super().__init__(name)
+
+        path = get_advisor_path()
+        if path:
+            self.add_include_dir(path.joinpath('include').as_posix())
```
FabioLuporini, line 386: as explained above, this should just be `self._include_dirs.append(...)`

conftest.py
georgebisbas: hm.... you are right.. I mistakenly saw in the docker image the mpi test being executed and thought it was this...(compared to a local run in which the mpi test was skipped....)

VERY good catch may thanks

devito/operator/profiling.py
georgebisbas: yes

conftest.py
```diff
@@ -79,6 +79,12 @@ def skipif(items, whole_module=False):
            isinstance(configuration['platform'], Cpu64):
             skipit = "`icx+cpu64` won't work with this test"
             break
+        # Skip if icx or advisor are not available
+        if i == 'noadvisor' and \
+           not isinstance(configuration['compiler'], IntelCompiler) or \
```
FabioLuporini, line 84: this is wrong again, I think, because of and/or operator precedence...

The fact we got this wrong multiple times and nothing breaks on CI suggests it's not very useful maybe

devito/operator/profiling.py
```diff
@@ -367,21 +367,23 @@ class AdvisorProfiler(AdvancedProfiler):
...
+        if self._attempted_init:
+            return
+
+        super().__init__(name)
+
+        path = get_advisor_path()
+        if path:
+            self._include_dirs.append(path.joinpath('include').as_posix())
+            self._lib_dirs.append(path.joinpath('lib64').as_posix())
+            self._attempted_init = True
```
FabioLuporini, line 379: maybe for another day, but these three lines 

```
     self._attempted_init = True
else:
     self._attempted_init = False
```

could become a single line

```
self._attempted_init = bool(path)
```

but anyway, nitpicking... maybe you can do it in another PR

examples/performance/02_advisor_roofline.ipynb
```diff
@@ -158,9 +155,13 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "\u001b[1;37;32mOpening project...\u001b[0m\n",
+      "\u001b[1;37;32mOpening project /home/gb4018/workspace/devitocodes/devito/examples/performance/profilings/JupyterProfiling...\u001b[0m\n",
```
FabioLuporini, line 158: This is not OK, I don't like personal dirs in logs (ie `gb4018`)

examples/performance/02_advisor_roofline.ipynb
```diff
@@ -218,9 +219,13 @@
...
```
FabioLuporini, line 222: same here

devito/operator/operator.py
```diff
@@ -219,6 +219,12 @@ def _build(cls, expressions, **kwargs):
 
         # Required for the jit-compilation
         op._compiler = kwargs['compiler']
+
+        # Required for compilation by the profiler
```
mloubout, line 223: Why aren't those part of the profller init?

georgebisbas: It was before, and this was the reason it was not working, as a different compiler was build after doing the profiler init. Related discussion here: https://devitocodes.slack.com/archives/C7JMLMSG0/p1739458707616429?thread_ts=1739458681.588539&cid=C7JMLMSG0

devito/operator/profiling.py
georgebisbas: but why append then if path is None? If the compiler reverts to advanced in the case of advisor's absence, the paths will be linked right?

FabioLuporini (approved):

benchmarks/user/advisor/roofline.py
```diff
@@ -198,19 +201,18 @@ def roofline(name, project, scale, precision, mode, th):
     legend = plt.legend(loc='center left', bbox_to_anchor=(1, 0.5),
                         prop={'size': 7}, title='Rooflines')
 
-    # saving the chart in PNG format
-    plt.savefig('%s.png' % name, bbox_extra_artists=(legend,), bbox_inches='tight')
+    # saving the chart in PDF format
+    plt.savefig('%s.pdf' % name, bbox_extra_artists=(legend,), bbox_inches='tight')
```
JDBetteridge, line 205:
```suggestion
    plt.savefig(f'{name}.pdf', bbox_extra_artists=(legend,), bbox_inches='tight')
```

benchmarks/user/advisor/README.md
```diff
@@ -1,48 +1,88 @@
...
+```
+### Limitations:
+
+* Not tested with all possible examples that Devito can support.
+* Running the `tripcounts` analysis is time-consuming, despite starting in paused
   mode. This analysis, together with the `survey` analysis, is necessary to
   generate a roofline. Both are run by `run_advisor.py`.
-* Requires python3, untested in earlier versions of python and conda environments
-* Currently requires download of repository and running `pip3 install .`, the scripts
+* Requires Python3, untested in conda environments
```
JDBetteridge, line 60:
```suggestion
* Requires Python 3.9 or later, untested in conda environments
```
Let's not use unsupported Python versions!

benchmarks/user/advisor/roofline.py
```diff
@@ -198,19 +201,18 @@ def roofline(name, project, scale, precision, mode, th):
...
     figpath = os.path.realpath(__file__).split(os.path.basename(__file__))[0]
-    log('Figure saved in %s%s.png.' % (figpath, name))
+    log('\nFigure saved in %s%s.pdf.' % (figpath, name))
```
JDBetteridge, line 207:
```suggestion
    log(f'\nFigure saved in {figpath}{name}.pdf.')
```

benchmarks/user/advisor/roofline.py
```diff
@@ -198,19 +201,18 @@ def roofline(name, project, scale, precision, mode, th):
...
 
     # Save the JSON file
     with open('%s.json' % name, 'w') as f:
         f.write(json.dumps(roofline_data))
 
-    log('JSON file saved as %s.json.' % name)
+    log('\nJSON file saved as %s.json.' % name)
```
JDBetteridge, line 213:
```suggestion
    log(f'\nJSON file saved as {name}.json.')
```

JDBetteridge (changes requested):
Nitpicks. But otherwise LGTM

JDBetteridge (approved):

FabioLuporini:
Merged, thanks

PR #2537, opened 2025-02-14
arch: Support Grace arm architecture

mloubout:


EdCaunt (approved):

devito/arch/compiler.py
```diff
@@ -614,6 +614,8 @@ def __init_finalize__(self, **kwargs):
         elif isinstance(platform, Cpu64):
             if language == 'openmp':
                 self.cflags.append('-mp')
+            if isinstance(platform, NvidiaArm):
+                self.cflags.append(f'-mcpu={platform.march}')
```
georgebisbas, line 618: I guess you cross-checked that this is not double-printed in the compilation line right? Otherwise OK

georgebisbas (approved):

PR #2538, opened 2025-02-15
compiler: Misc extensions towards supporting thread block clustering

FabioLuporini:


devito/types/basic.py
```diff
@@ -211,6 +211,14 @@ def _mem_shared(self):
         """
         return False
 
+    @property
+    def _mem_shared_remote(self):
```
mloubout, line 215: I assume this is for grace-hoper where the GPU can access the grace's shared memory?

devito/types/misc.py
```diff
@@ -342,3 +342,17 @@ def closing(self):
 A wildcard for use in the RHS of Eqs that encode some kind of semantics
 (e.g., a synchronization operation) but no computation.
 """
+
+
+# *** CXX support types
+
+auto = CustomDtype('auto')
```
mloubout, line 349: I get why it is done like that, but it's not really a type. It's a return qualifier. I'm not sure how else it should be done; maybe add a TODO or issue for it.

mloubout (approved):
Minot comment looks good to go

devito/passes/clusters/cse.py
```diff
@@ -70,6 +70,12 @@ def cse(cluster, sregistry=None, options=None, **kwargs):
     min_cost = options['cse-min-cost']
     mode = options['cse-algo']
 
+    # Unsafe, and almost always uselsss as these tend to be no-op Clusters, to
```
georgebisbas, line 73: useless

devito/types/basic.py
FabioLuporini: no, this is for just hopper where a block X running on SM0 can access the shared memory of block Y on the neighbouring SM1

devito/passes/clusters/cse.py
FabioLuporini: agreed, dropping

devito/types/misc.py
FabioLuporini: done, pushing updated version now

mloubout (approved):

devito/passes/clusters/cse.py
georgebisbas: I actually meant the typo: uselsss, but happy to help indirectly

PR #2539, opened 2025-02-17
tests: Remove floating point equality assertions

JDBetteridge:
I'm guessing we are going to find more

### DIFF
--- a/tests/test_symbolics.py
+++ b/tests/test_symbolics.py
@@ -712,7 +712,7 @@ def test_pow_precision(dtype, expected):
     op.apply()
 
     assert expected in str(op)
-    assert np.all(f.data == 8.0)
+    assert np.allclose(f.data, 8.0, rtol=np.finfo(dtype).eps)
 
 
 @pytest.mark.parametrize('dtype,expected', [


tests/test_symbolics.py
```diff
@@ -620,8 +621,8 @@ def test_minmax_precision(dtype, expected):
     # Check generated code -- ensure it's using the fp64 versions of min/max,
     # that is fminf/fmaxf
     assert all(i in str(op) for i in expected)
-
-    assert np.all(f.data == 6.0)
+    tolerance = 1e-8 if dtype is np.float32 else 1e-16
```
georgebisbas, line 624: I would add tolerance to the parametrization instead of here

georgebisbas: for all the tests

tests/test_symbolics.py
```diff
@@ -643,7 +644,8 @@ def test_pow_precision(dtype, expected):
     op.apply()
 
     assert expected in str(op)
-    assert np.all(f.data == 8.0)
+    tolerance = 1e-8 if dtype is np.float32 else 1e-16
```
georgebisbas, line 647: tolerance = 1e-8 if dtype is np.float32 else 1e-16
```suggestion
    tolerance = 1e-7 if dtype is np.float32 else 1e-16
```

tests/test_symbolics.py
```diff
@@ -705,7 +708,7 @@ def test_multibounds_op(self):
         op.apply(time_M=5)
         fnorm2 = norm(f)
 
-        assert fnorm == fnorm2
+        assert math.isclose(fnorm, fnorm2, rel_tol=1e-8)
```
mloubout, line 711: should use `np.isclose`

tests/test_symbolics.py
```diff
@@ -665,7 +667,8 @@ def test_abs_precision(dtype, expected):
     op.apply()
 
     assert expected in str(op)
-    assert np.all(f.data == 1.0)
+    tolerance = 1e-8 if dtype is np.float32 else 1e-16
```
mloubout, line 670: Don't need those to be hard coded, you can use `np.finfo(dtype).eps` 

```
In [4]: np.finfo(np.float32).eps
Out[4]: np.float32(1.1920929e-07)
```

mloubout: You can also define those eps at the top for the whole file it's constants (or even in conftest)

tests/test_symbolics.py
mloubout: >np.isclose returns a boolean array

only for arrays, it returns a numpy boolean for scalars which those are

tests/test_symbolics.py
JDBetteridge: Used numpy inspection instead

tests/test_symbolics.py
JDBetteridge: So it does

mloubout (approved):

georgebisbas (approved):
passes locally

tests/test_symbolics.py
```diff
@@ -620,8 +620,7 @@ def test_minmax_precision(dtype, expected):
     # Check generated code -- ensure it's using the fp64 versions of min/max,
     # that is fminf/fmaxf
     assert all(i in str(op) for i in expected)
-
-    assert np.all(f.data == 6.0)
+    assert np.allclose(f.data, 6.0, rtol=np.finfo(dtype).eps)
```
FabioLuporini, line 623: why? we expect perfect equality here

tests/test_symbolics.py
```diff
@@ -705,7 +704,7 @@ def test_multibounds_op(self):
         op.apply(time_M=5)
         fnorm2 = norm(f)
 
-        assert fnorm == fnorm2
+        assert np.isclose(fnorm, fnorm2, rtol=np.finfo(fnorm.dtype).eps)
```
FabioLuporini, line 707: lolz, I struggle to understand this test at all... aren't those two Operators equivalent?

and if, again, should we not just expect perfect equality

FabioLuporini (commented):
Yes, I know that fp comparison should go with an eps etc, but imho these cases are quite special and we expect actual equality

tests/test_symbolics.py
JDBetteridge: @georgebisbas reported issues on
> gcc 11.4,  Intel(R) Core(TM) i7-7700K CPU , ubuntu 22, latest master

I opened the PR in response. I wonder if it's still an issue? I couldn't reproduce, but could understand it being platform specific

tests/test_symbolics.py
JDBetteridge: No idea

mloubout (approved):

FabioLuporini (approved):

PR #2556, opened 2025-03-17
install: Add packaging to requirements

JDBetteridge:
I keep getting tripped up by this every time I create a new `venv`


mloubout (approved):
I always thought it was default with python but yeah it keeps popping up

georgebisbas (approved):

PR #2557, opened 2025-03-20
dsl: ensure SparseFunction coordinates and point symbols are always real

EdCaunt:


mloubout (commented):
Looks fine, one small comment.

devito/types/sparse.py
```diff
@@ -202,6 +202,12 @@ def __subfunc_setup__(self, suffix, keys, dtype=None, inkwargs=False, **kwargs):
             else:
                 dtype = dtype or self.dtype
 
+        # Complex coordinates are not valid, so fall back to corresponding
+        # real floating point type if dtype is complex.
+        if issubclass(dtype, np.complexfloating):
```
mloubout, line 207: Weird my comment disapearded. 

`dtype = self.dtype(0).real.__class__` is safer. For example, if someone uses fp16 sparse this would automatically upcast all subfunctions

EdCaunt: Oh that's much better, I was looking for something like that

georgebisbas (approved):

mloubout (approved):

